Issue #526, opened 2020-03-09
Code Standards / Linting checked & applied consistently for all langs


ches:
**Is your feature request related to a problem? Please describe.**

Code standards expectations are by documentation for some project languages, and by automation for others. In cases of the former:

  1. Contributors don't realize they should follow them
  1. Contributors use different versions or settings for tools like formatters
  1. Non-compliant code sometimes gets checked in and merged. When someone later fixes it i.e. by running a formatter, their PR may be inflated with many out-of-scope changes.

**Describe the solution you'd like**

For all of Go, Java, and Python:

- Standards are checked by CI, to address issues 1 and 3.
- Development dependency versions and settings (linting and formatting tools) are managed by build automation, to alleviate issue 2.

Try to fail fast for feedback, but without undue friction on development. For Java for example, we complain on `mvn test` but not `mvn compile` (nor `test-compile`). Perhaps better yet, on CI all the linting could be in one dedicated job / GH status check (though a failure should report all problems, not fail one language, fix and push, fail another language…).

Git hooks could be an optional suggestion.

@woop suggested on #508:

> How about we just have `make lint` or `make format` at the root of the Feast repo? That would be based on hardcoded configuration that everyone can see, and should ideally cover the whole repo (Java, Go, Python).

Sounds good to me.

**Describe alternatives you've considered**

**Additional context**

#487 
#508

woop:
I am closing this PR in favour of https://github.com/gojek/feast/pull/545. 

@ches I think we can just remove the Python generated code for now. Doesnt seem to be worth all the extra trouble. We can assess the impact later. Java and Go remain unchanged.

woop:
Oops, accidentally closed this issue instead of #508.

woop:
I'm pretty sure this is sufficiently resolved for now. Happy to reopen though. Closing for now.

PR #545, opened 2020-03-17
Add Format and Lint to Makefile

woop:

Refactors the root Make file so that we can `format` and `lint` for Java, Python, and Golang.

This PR works towards fixing #526, but does not address all the necessary requirements to close it.

```release-note
None
```


woop:
I think we still need two or three things before we can close #526.
1. A dedicated task that runs these checks and makes sure all failures are visible before exiting
2. A means of pinning versions so that protoc (and its plugins) as well as the other formatters are always consistent. Alternatively we can also remove the auto generated code from the repository.

woop:
The latest commit also removes Python proto generated code.

sdk/python/requirements-ci.txt
```diff
@@ -28,4 +28,11 @@ confluent_kafka
...
-tabulate==0.8.*
\ No newline at end of file
+tabulate==0.8.*
+isort
+grpcio-tools
+mypy
+mypy-protobuf
+pre-commit
+flake8
+black
```
imjuanleonard: would propose to set the version for the linters like grpcio-tools, mypy, black, flake8 etc.
It would be awesome when we can be sure on same version when developing

Rest is LGTM :) The import structure and linter, format would be really helpful for anyone developing next

sdk/python/feast/__init__.py
```diff
@@ -1,14 +0,0 @@
-from pkg_resources import get_distribution, DistributionNotFound
```
imjuanleonard: Why this file get removed?

sdk/python/requirements-ci.txt
woop: I agree, but we also dont want to be too restrictive. If we force everyone to confirm to our versions then its also quite high friction. Only `protoc` was a problem, and with the addition of this PR we won't be version controlling proto generated Python code, so the problem kind of goes away.

I think we can pin the versions once it becomes a problem

sdk/python/feast/__init__.py
woop: Yea this shouldnt have been removed.

sdk/python/requirements-ci.txt
imjuanleonard: Got it

ches: The "compatible" specifier might be a reasonable compromise, e.g. approximately expecting Semantic Versioning if a tool doesn't break behavior within a major version series:

    flake8 ~= 3.0

Makefile
```diff
@@ -14,26 +14,73 @@
...
+# Go SDK
+
+compile-protos-go:
+	@$(foreach dir,$(PROTO_TYPE_SUBDIRS), cd ${ROOT_DIR}/protos; protoc -I/usr/local/include -I. --go_out=plugins=grpc,paths=source_relative:../sdk/go/protos/ feast/$(dir)/*.proto;)
+
+format-go:
+	cd ${ROOT_DIR}/sdk/go; gofmt -s -w *.go
+
+lint-go:
+	cd ${ROOT_DIR}/sdk/go; go vet; golint *.go
```
ches: If my version of `golint` is still current, it requires a flag `-set_exit_status` in order to return an error status to the shell.

A formatting check should probably be included in `lint-go` as well, it is for the other languages. `gofmt` has no such feature at all to return an error exit status, the idiom seems to be to use `test -z $(gofmt -l ./)` when this is desired.

I'm not sure if we're actually wanting it to fail on CI yet, because of the desire to not fail lint checks one-at-a-time, but we probably want to be consistent about whether all the `lint-*` make targets exit nonzero when there's an issue (I haven't checked behavior of all others yet).

Probably `lint-go` will not fail CI for now even if this target is updated to exit nonzero, because `.prow/scripts/test-golang-sdk.sh` does not `set -e`.

Issue #302, opened 2019-11-13
Add job management APIs


zhilingc:
We should add a job management to the core API for 
- tracking status of jobs
- cleaning up jobs

For retrieval, we could possibly retrieve jobs based on the existing GetFeatureSets and GetStores filters:

```
rpc GetJobs(GetJobsRequest) returns (GetJobsResponse);

// Retrieves matching jobs given feature set and/or store filters. If none are provided,
// all active jobs will be returned.
message GetJobsRequest {
    // Filter by feature set name and version
    GetFeatureSetsRequest.Filter feature_set = 1;

    // Filter by store name
    GetStoresRequest.Filter store = 2;
}

message GetJobsResponse {

    enum JobStatus {
        // Job state unknown
        UNKNOWN = 0;

        // New feature set or feature set version created
        PENDING = 1;

        // Error occurred while trying to apply changes
        RUNNING = 2;

        COMPLETED = 3;

        /** When user sent abort command, but it's still running */
        ABORTING = 4;

        /** User initiated abort job */
        ABORTED = 5;

        /**
        * Runner’s reported that the import job failed to run or there is a failure during job
        * submission.
        */
        ERROR = 6;

        /** job has been suspended and waiting for cleanup */
        SUSPENDING = 7;

        /** job has been suspended */
        SUSPENDED = 8;
    }

    message Job {
        string name = 1;
        JobStatus status = 2;
        repeated feast.core.FeatureSet feature_sets = 3;
        feast.core.Store store = 4;
        feast.core.Source source = 5;
    }
    repeated Job jobs = 1;
}
```

Jobs should be aborted by name. 

```
rpc AbortJob(AbortJobRequest) returns (AbortJobResponse);

message AbortJobRequest {
    string job_name = 1;
}

message AbortJobResponse {
    enum Status {
        INVALID = 0;
        SUCCESS = 1;
        ERROR = 2;
    }

    // Feature set response has been enriched with version and source information
    Status status = 1;
}
```

woop:
Thanks for this @zhilingc!

It seems like there are three types of "gets":
1. Get one job
2. Get all jobs
3. Get some jobs (all, filter)

Can I suggest that we start with a Job RPC and a ListJobs RPC instead of a single RPC that does both? [It would mirror this structure from Kubernetes](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cri-api/pkg/apis/runtime/v1alpha2/api.proto#L101)

[...]

PR #548, opened 2020-03-18
Add Ingestion Job management API for Feast Core

mrzzy:
PR to implement a Ingestion Job Management API into feast core.
This PR adds:
- extends CoreService API to  give the ability to list, stop, and restart ingestion jobs.
- extends Python sdk's `Client` to add support for the additions to the core service API.
- extends the CLI with `feast ingest-job list, describe, stop, restart`
- tests for the additions where appropriate.

Fixes:  #302 

**Does this PR introduce a user-facing change?**: Yes

```release-note
Added Ingestion Job Management to Feast Core and the Python SDK:
- Listing Ingestion Jobs with filters
- Restarting Ingestion Jobs
- Stopping Ingestion Jobs

Added ingest-job subcommand to CLI for Ingestion Job Management:
-  feast ingest-job list, describe, stop, restart
```


core/src/main/java/feast/core/service/JobService.java
```diff
@@ -0,0 +1,223 @@
...
+
+  /* Private Utility Methods */
+  /**
+   * Finds &amp; returns featuresets matching the given feature set refererence
+   *
+   * @param fsReference FeatureSetReference that specifies which featuresets to match
+   * @throws UnsupportedOperationException fsReference given is unsupported.
+   * @return Returns a list of matching featuresets
+   */
+  private List<FeatureSet> findFeatureSets(FeatureSetReference fsReference)
```
woop: This should almost certainly be externalized. Can you have a look in the code base to see where else it occurs and see if we can reuse or generalize this?

core/src/main/java/feast/core/service/JobService.java
```diff
@@ -0,0 +1,223 @@
...
+    Optional<Job> getJob = this.jobRepository.findById(request.getId());
+    if (getJob.isEmpty()) {
+      throw new NoSuchElementException(
+          "Attempted to stop nonexistent job with id: " + getJob.get().getId());
+    }
+
+    // check job status is valid for stopping
+    Job job = getJob.get();
+    JobStatus status = job.getStatus();
+    if (status.equals(JobStatus.ABORTED)
```
zhilingc: You can use `JobStatus.TERMINAL_STATE` for this check

core/src/main/java/feast/core/dao/FeatureSetRepository.java
```diff
@@ -36,6 +36,12 @@ FeatureSet findFirstFeatureSetByNameLikeAndProject_NameOrderByVersionDesc(
   // find all feature sets and order by name and version
   List<FeatureSet> findAllByOrderByNameAscVersionAsc();
 
+  // find all feature sets by name and project name
+  List<FeatureSet> findAllByNameAndProject_Name(String name, String projectName);
```
zhilingc: I dont think you need these methods, might be better to inject the `SpecService` and use `ListFeatureSets`

protos/feast/core/IngestionJob.proto
```diff
@@ -0,0 +1,65 @@
...
+    // List of feature sets whose features are populated by this job.
+    repeated feast.core.FeatureSet feature_sets = 4;
+    // Source this job is reading from.
+    feast.core.Source source = 5;
+    // Store this job is writing to.
+    feast.core.Store store = 6;
+}
+
+// Status of a Feast Ingestion Job
+enum IngestionJobStatus {
```
zhilingc: Could you possibly make feast.core.model.JobStatus a facade of this? or update the Job model to use this enum instead of our own

core/src/main/java/feast/core/service/JobService.java
mrzzy: How should calling `stopJob()` in one of the transitional states (ie`ABORTING`, `SUSPENDING` be handled? `JobStatus.TERMINAL_STATE` only the final states.

protos/feast/core/IngestionJob.proto
mrzzy: I added `toIngestionProto()` to `JobStatus` that creates the `IngestionJobStatus` proto.

core/src/main/java/feast/core/dao/FeatureSetRepository.java
mrzzy: I refactored `JobService` to use `SpecService` to get its featuresets. The new methods have been removed.

core/src/main/java/feast/core/service/JobService.java
mrzzy: I updated JobService use `JobStatus.TERMINAL_STATE` to check for final states. I added `JobStatus.TRANSITIONAL_STATES` to track transitional states (ie `PENDING`, `ABORTING`, `SUSPENDING`)

core/src/main/java/feast/core/service/JobService.java
mrzzy: I moved the this method to `SpecService` as `matchFeatureSets()`.

core/src/main/java/feast/core/service/JobService.java
```diff
@@ -0,0 +1,228 @@
...
+  /* Job Service API */
+  /**
+   * List Ingestion Jobs in feast matching the given request
+   *
+   * @param request list ingestion jobs request specifying which jobs to include
+   * @throws InvalidArgumentException when given filter in a unsupported configuration
+   * @throws InvalidProtocolBufferException on error when constructing response protobuf
+   * @return list ingestion jobs response
+   */
+  public ListIngestionJobsResponse listJobs(ListIngestionJobsRequest request)
```
zhilingc: Since this method requires multiple separate calls to the db, it should be annotated with 
`@Transactional(readOnly=True)` to ensure that you're reading from the same snapshot of data

core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -131,6 +132,38 @@ public GetFeatureSetResponse getFeatureSet(GetFeatureSetRequest request)
     return GetFeatureSetResponse.newBuilder().setFeatureSet(featureSet.toProto()).build();
   }
 
+  /**
+   * Finds &amp; returns the featuresets matching the given feature set reference. TODO: merge with
```
zhilingc: Yeah, you should merge this method. Add a private method to the jobservice to convert the feature reference to a filter if necessary.

mrzzy: @woop wanted the method out of `JobService`, so I moved it to `SpecService`:
> This should almost certainly be externalized. Can you have a look in the code base to see where else it occurs and see if we can reuse or generalize this?
Where should the method be?

zhilingc: You can use the existing method in specservice, just add a helper method to convert the `FeatureReference` to `Filter` since the former is a subset of the latter anyway.

core/src/main/java/feast/core/model/Job.java
```diff
@@ -102,4 +106,31 @@ public void updateMetrics(List<Metrics> newMetrics) {
   public String getSinkName() {
     return store.getName();
   }
+
+  /**
+   * Convert a job model to ingestion job proto
+   *
+   * @return Ingestion Job proto derieved from the given job
+   */
+  public IngestionJobProto.IngestionJob toIngestionProto() throws InvalidProtocolBufferException {
```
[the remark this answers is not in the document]
mrzzy: Fixed.

core/src/main/java/feast/core/model/JobStatus.java
```diff
@@ -64,4 +66,38 @@
...
+  public static final Collection<JobStatus> getTransitionalStates() {
+    return TRANSITIONAL_STATES;
+  }
+
+  /**
+   * Convert a Job Status to Ingestion Job Status proto
+   *
+   * @return IngestionJobStatus proto derieved from this job status
+   */
+  public IngestionJobStatus toIngestionProto() {
```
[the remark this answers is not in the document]
mrzzy: Fixed.

core/src/main/java/feast/core/model/JobStatus.java
```diff
@@ -64,4 +66,38 @@
...
+    // maps job models job status to ingestion job status
+    Map<JobStatus, IngestionJobStatus> statusMap =
```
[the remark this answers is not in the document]
mrzzy: I think this should remain non static as it allows us to write: `status.toProto();`
instead of a more verbose 
`JobStatus.toProto(status);`

> assuming the signature of the method becomes `toProto(JobStatus)`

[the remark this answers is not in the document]
zhilingc: No, I mean the statusMap should be static, not the method.

mrzzy:
This PR is non longer a work-in-progress and is ready for review.

I have manually tested to this to work with both Direct and Dataflow runners, although I like to ask a reviewer to double check that the integration with Dataflow is fully working.

woop:
> This PR is non longer a work-in-progress and is ready for review.

Thanks for this @mrzzy. Much needed

> I have manually tested to this to work with both Direct and Dataflow runners, although I like to ask a reviewer to double check that the integration with Dataflow is fully working.

How would the reviewer double check integration with Dataflow is working? Normally a reviewer only reviews the code, and the code should ideally contain all the necessary automation to test the functionality that is implemented.

mrzzy:
I agree that this is the ideal case, however the current end to end tests are only configured to test with the Direct runner, leaving a correctness blind spot with the Dataflow runner that has be addressed with manual testing.

woop:
> I agree that this is the ideal case, however the current end to end tests are only configured to test with the Direct runner, leaving a correctness blind spot with the Dataflow runner that has be addressed with manual testing.

Ideally the tests would be updated to incorporate Dataflow testing as well. We can possibly leave it out of scope for now. The reviewer in this case won't have time to actually spin up Dataflow and test this all out.

@mrzzy would you mind creating a new GitHub issue to add a Dataflow based test?

mrzzy:
I opened issue #595 for Dataflow Test Coverage.

core/src/main/java/feast/core/service/JobService.java
```diff
@@ -0,0 +1,282 @@
...
+    Optional<Job> getJob = this.jobRepository.findById(request.getId());
+    if (getJob.isEmpty()) {
+      throw new NoSuchElementException(
+          "Attempted to stop nonexistent job with id: " + getJob.get().getId());
+    }
+
+    // check job status is valid for restarting
+    Job job = getJob.get();
+    JobStatus status = job.getStatus();
+    if (JobStatus.getTransitionalStates().contains(status) || status.equals(JobStatus.UNKNOWN)) {
```
zhilingc, line 168: It might make sense to disallow restarting of jobs in terminal state as well, in order to prevent race conditions with the job coordinator. Another (smaller) issue with allowing job restarts for terminal jobs is that Dataflow only retains information about stopped jobs for a short period of time, after which the job cannot be retrieved any more.

sdk/python/feast/job.py
```diff
@@ -187,3 +194,98 @@ def to_chunked_dataframe(
 
     def __iter__(self):
         return iter(self.result())
+
+
+class IngestJob:
```
zhilingc, line 199: Can you implement the __str__ method so that the object can be printed out in a human readable format? See feature_set.py:L87

core/src/main/java/feast/core/service/JobService.java
mrzzy: Thanks for pointing the race condition out, I have disabled restarting for terminal job statuses. As for your second point, this would not be a problem as the Dataflow Job Manager is configured to recreate the job once stopped.

sdk/python/feast/job.py
mrzzy: I have added `__str__()` and `__repr__()` to IngestJob to render it in a human readable format.

mrzzy:
@zhilingc Thanks for the review and taking the time out of your day to test the dataflow runner.
I have implemented your comments into the PR:
-  disabled `stopJobs()` for suspended or aborted jobs due to possible race condition.
- added `__str__()` and `__repr__()` to `IngestJob` to render it in a human readable format.
- Added `feast ingest-job list, describe, stop, restart` to feast CLI
- Renamed `Job` to `RetrivalJob` to prevent confusion with `IngestJob` in python sdk.

Issue #489, opened 2020-02-24
Shift ingestion feature metrics to after store write 


zhilingc:
**Is your feature request related to a problem? Please describe.**
Existing metrics collection for ingestion is done in parallel to store writes:
```
               +
               |
       +-------+--------+
       |                |
+------+-----+   +------+-----+
|            |   |            |
|   store    |   |  metrics   |
|   write    |   |   write    |
|            |   |            |
+------------+   +------------+
```

While this does give a rough gauge as to what data looks like in flight, it doesn't actually reflect the status of the data in the store, which is ultimately what we are concerned about, since write failures are common. 

**Describe the solution you'd like**
Ideally we want to write metrics POST store writes, as so:
```
              ...
               +
               |
         +-----+-----+
         |           |
         |   store   |
         |   write   |
         |           |
         +-----+-----+
               |
      +--------+--------+
      +                 +
   success           failure
      +                 +
      |                 |
+-----+-----+     +-----+-----+
|  write    |     |  write    |
|  sucess   |     |  failure  |
|  metrics  |     |  metrics  |
+-----------+     +-----------+
```

Metrics collected post store writes would more accurately reflect the data in the stores.

PR #553, opened 2020-03-20
Change serving and ingestion to use storage API

zhilingc:
Swaps out specific storage implementations in the serving service and ingestion code for the interfaces defined in the serving service.

Additionally addresses Issue #489: the WriteResult is used to write metrics only after successful writes.


serving/src/main/java/feast/serving/configuration/ServingServiceConfig.java
```diff
@@ -79,9 +83,9 @@ public ServingService servingService(
 
     switch (store.getType()) {
       case REDIS:
-        servingService =
-            new RedisServingService(
-                storeConfiguration.getServingRedisConnection(), specService, tracer);
+        OnlineRetriever redisRetriever =
```
woop: Just flagging that we should probably have two "store types" that a user can define, so that we dont have one case statement that deals with both batch and online stores. Right now the user has to know what a store will be used for since we will select online/batch for them behind the scenes. A dual purpose store might be a problem.

serving/src/main/java/feast/serving/service/BatchServingService.java
```diff
@@ -0,0 +1,89 @@
...
+      return GetFeastServingInfoResponse.newBuilder()
+          .setType(FeastServingType.FEAST_SERVING_TYPE_BATCH)
+          .build();
+    }
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public GetOnlineFeaturesResponse getOnlineFeatures(GetOnlineFeaturesRequest getFeaturesRequest) {
+    throw Status.UNIMPLEMENTED.withDescription("Method not implemented").asRuntimeException();
```
woop: I know that we are only instantiating a single serving service per Feast Serving deployment, but I want to point out that that our `ServingService` interface is unnatural if the implementers of that interface will never implement all its methods. I would prefer to have `ServingService` and `ServingServiceImpl`, where we always select `ServingServiceImpl`. If a store supports both online and batch, then it can be queried for both. If it only supports batch, then online will fail, and vice versa.

serving/src/main/java/feast/serving/configuration/ServingServiceConfig.java
zhilingc: We can refactor this in another PR.

Issue #504, opened 2020-02-29
Authentication and authorization


woop:
Creating this issue to track the discussion around user authentication & authorization. 

### 1. Prior Art: Auth as proposed in the Feast Projects RFC
In the [Feast: Projects RFC](https://docs.google.com/document/d/14-QBz9X8zK_aGAY2ti43a7PqMxvqp0035ec7QYgmEBM/edit#heading=h.c29o6bkcurw9) we proposed a basic form of authentication as a "minmal" implementation for Feast 0.5.

**Overview**
At a high level, this proposal adds user authentication based on a new user table that Feast maintains. No API is made available to manage the creation/removal of users, but the actions taken by users is on the Feast Core API is limited to the resources they have access to. Resource access is managed only based on membership of users to projects. 

Note on Roles: "Roles" were originally proposed but will be delayed until a future release can implement them properly. Right now there is only a full "owner" role which is implicit. All users either have ownership of a project, or not.

**Assumptions**:
- User authentication only applies to Feast Core, not ingestion or "serving".

**Out of scope**:
- Full user management (create or remove users, or manage user profiles).
- JWT or functionality like sessions on user auth.
- RBAC

**Data model**
![image](https://user-images.githubusercontent.com/6728866/75601939-4d655880-5afb-11ea-8fc4-249bf3ae0df4.png)

**API**
- Feast Core API
  - New RPCs
    - AddMember: Add a user to a specific project
    - RemoveMember: Remove user from a specific project
    - ListMembers: List members in a project
    - ~~AddRole~~
    - ~~RemoveRole~~
- Feast Core Data Model
  - New table
    - Users: Users table
    - UserProjects: Relation between users & projects
    - ~~Roles: Roles table~~
    - ~~UserProjectRole: Relation between roles, users, and projects~~
  - New classes
    - Users
    - ~~Roles~~
- Python SDK
  - Addition of user auth configuration
  - New methods:
    - add_member()
    - remove_member()
    - list_members()
    - ~~add_role()~~
    - ~~remove_role()~~

### 2. Objectives
The main objective of this issue is to finalize two proposals. The first is a "minimal" set of functionality that we want to support, and the second is a "full" set of functionality. Since Feast 0.5 is our next release, the priority is to finalize the implementation of the minimal release.

* **Minimal**: To implement a basic form of authentication in Feast 0.5. 
[...]

PR #554, opened 2020-03-20
Add support for authentication and authorization

woop:

First implementation of auth for Feast (related to #504 minimal implementation).

1. Adds authentication to Feast Core (with support for different implementations). Currently only supports Google Open ID through gRPC metadata.
2. Adds authorization to Feast Core (with support for different implementations). Currently only supports Ory Keto.
3. Adds authentication to Python SDK/CLI. Users can enable authentication client side and Feast will send their Google Open ID credentials as gRPC metadata to Core. 
4. Refactor the Python SDK/CLI SSL/TLS handling.
5. Prevents unauthorized creation or modification of feature sets in projects that a user does not have membership in.

Limitations
- Does not handle user or role management in authorization provider (creating projects, adding members, removing members, listing members).

Related to #504, but doesn't close the card. This is a minimal implementation.

```release-note
None
```


woop:
testing prow.

dr3s:
Does this really only support Google OpenID?  It shouldn't be any effort to make the OIDC configurable to support any valid implementation such as Auth0.

core/src/main/java/feast/core/auth/authentication/GoogleOID/GoogleOpenIDAuthenticationProvider.java
```diff
@@ -0,0 +1,68 @@
...
+import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.oauth2.jwt.NimbusJwtDecoder;
+import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationConverter;
+import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationProvider;
+
+/**
+ * Google Open ID Authentication Provider. This provider is used to validate incoming requests to
+ * Feast Core.
+ */
+public class GoogleOpenIDAuthenticationProvider implements AuthenticationProvider {
```
dr3s, line 31: This seems like it would support any openid provider.  Any reason the package and class can't be made generic?

dr3s: Actually, why even have this class at all?  Why not just set up the JwtAuthenticationProvider directly?

core/src/main/java/feast/core/config/SecurityConfig.java
```diff
@@ -0,0 +1,144 @@
...
+   * @return AuthenticationManager
+   */
+  @Bean
+  @ConditionalOnProperty(prefix = "feast.security.authentication", name = "enabled")
+  AuthenticationManager authenticationManager() {
+    final List<AuthenticationProvider> providers = new ArrayList<>();
+
+    if (securityProperties.getAuthentication().isEnabled()) {
+      switch (securityProperties.getAuthentication().getProvider()) {
+        case "GoogleOpenID":
```
dr3s, line 64: not to beat a dead horse but this seems to provide little value over directly setting properties of the JwtAuthenticationProvider

core/src/main/java/feast/core/config/SecurityConfig.java
```diff
@@ -0,0 +1,144 @@
...
+          providers.add(
+              new GoogleOpenIDAuthenticationProvider(
```
dr3s, line 66: Why not use the capabilities of spring to configure the providers rather than build another factory method?

core/src/main/java/feast/core/config/SecurityConfig.java
```diff
@@ -0,0 +1,144 @@
...
+   *
+   * @return AuthorizationProvider used to validate access to Feast resources.
+   */
+  @Bean
+  @ConditionalOnProperty(prefix = "feast.security.authorization", name = "enabled")
+  AuthorizationProvider authorizationProvider() {
+    if (securityProperties.getAuthentication().isEnabled()
+        && securityProperties.getAuthorization().isEnabled()) {
+      switch (securityProperties.getAuthorization().getProvider()) {
+        case "KetoAuthorization":
```
dr3s, line 135: same as authentication provider factory comment.

core/src/main/java/feast/core/config/SecurityConfig.java
woop: > Does this really only support Google OpenID? It shouldn't be any effort to make the OIDC configurable to support any valid implementation such as Auth0.

Yip, it should be pretty easy to extend to non-Google through config. This was a minimal implementation. The reason I specified it as Google is because our Python SDK uses OIDC tokens that it gets from a local Gcloud SDK and sends that along to Feast for auth. That is very Google specific.

Server side can easily be generalized.

core/src/main/java/feast/core/config/SecurityConfig.java
woop: How would you use Spring in this case?

woop:
> Does this really only support Google OpenID? It shouldn't be any effort to make the OIDC configurable to support any valid implementation such as Auth0.

Yip, it should be pretty easy to extend to non-Google through config. This was a minimal implementation. The reason I specified it as Google is because our Python SDK uses OIDC tokens that it gets from a local Gcloud SDK and sends that along to Feast for auth. That is very Google specific.

Server side can easily be generalized.

core/src/main/java/feast/core/auth/authentication/GoogleOID/GoogleOpenIDAuthenticationProvider.java
woop: > This seems like it would support any openid provider. Any reason the package and class can't be made generic?

It can.

> Actually, why even have this class at all? Why not just set up the JwtAuthenticationProvider directly?

Honestly I don't recall why this was broken out. It looks like if we generalize cert handling then we can use JwtAuthenticationProvider

core/src/main/java/feast/core/config/SecurityConfig.java
woop: Can be done. See points above.

core/src/main/java/feast/core/config/SecurityConfig.java
dr3s: It's been awhile since I have done this in Spring Config but basically you dynamically create the @Bean which then can accept normal spring config:
```
@Configuration
public class AuthConfiguration
{

     @Value("${feast.auth.authorizationProviderClass}")
     private String authorizationProviderClassName;

    @Bean
     public AuthorizationProvider authorizationProvider()
     {
        Class<?> authClass = Class.forName(authorizationProviderClassName);
        AuthorizationProvider authProvider = AuthorizationProvider.class.cast(authClass.newInstance());
        return authProvider;
     }
```

dr3s:
> > Does this really only support Google OpenID? It shouldn't be any effort to make the OIDC configurable to support any valid implementation such as Auth0.
> 
> Yip, it should be pretty easy to extend to non-Google through config. This was a minimal implementation. The reason I specified it as Google is because our Python SDK uses OIDC tokens that it gets from a local Gcloud SDK and sends that along to Feast for auth. That is very Google specific.
> 
> Server side can easily be generalized.

Got it.  We have it on our list to add an generic oauth2/openid auth module to the client as well.

core/src/main/java/feast/core/config/SecurityConfig.java
woop: Yea that looks great, as long as failure is graceful and easy to debug.

dr3s:
replaced by #793

Issue #556, opened 2020-03-21
Feast CI: Prow vs Gitlab, Circle, Drone


woop:
**Is your feature request related to a problem? Please describe.**

We are currently using [Prow](https://github.com/kubernetes/test-infra/tree/master/prow) for 
- Managing pull request life cycles
- Executing tests and builds

Prow is relatively flexible. It's nothing more than an orchestrator for spinning up Kubernetes pods in a cluster. However, the CI experience can be improved.
- Prow is not intuitive to understand nor is it well documented. This means most contributors don't want to touch prow config nor CI in general.
- The Prow UI is simplistic and functional, but doesn't present logs in a way that is easy to read. Most modern CI/CD systems will aggregate logs as well as apply styling or colour coding. Most modern CI systems also make it easy to understand how a specific step in the pipeline is being executed without having to look at Kubernetes yaml.
- Prow is very low level. A lot of functionality that already exists in other CI systems needs to be manually implemented
    - Different runner classes
    - Protected variables
    - Caching of builds steps and dependencies (this is a big one)
    - Most CI systems have many integrations like Slack, Jira, email notifications, but Prow has none.

**Describe the solution you'd like**
I'd like to propose that we continue to use Prow for managing the PR life cycle, but that we also start to explore the integration of a model CI system like Circle CI, Drone CI, or Gitlab CI. We could introduce these systems for common tasks like `linting`, `unit-tests` or `releases`, and if they meet our needs we could move our integration/e2e tests off of Prow as well.

GitHub supports having multiple CI systems run on the same code, and Prow will wait for all of the CI checks to pass before allowing a merge. As a first cut I will attempt to add linting as through [Gitlab-CI](http://gitlab.com)

woop:
I've set up Gitlab for Feast CI over here: https://gitlab.com/feast-dev/feast/

Pipelines will automatically be triggered and will be visible here: https://gitlab.com/feast-dev/feast/pipelines/

Currently this is limited to linting.

woop:
Closing this issue since the setup is stable with GitHub Actions and Prow.

PR #558, opened 2020-03-21
Add Feast CI image and infra housekeeping

woop:

This is a general infra cleanup task. 
- Move CI scripts into `infra/scripts` folder instead of Prow.
- Adds a Dockerfile for general purpose CI build steps `feast-ci`
- Refactor Makefile a bit

Note: Still some Python formatting/linting that needed to happen. Hopefully the last time.

Helps with #526 and #556 since we can use this image to run parallel lint steps in Gitlab

```release-note
None
```


woop:
Force merging this.

Makefile
```diff
@@ -30,8 +30,15 @@ protos: compile-protos-go compile-protos-python compile-protos-docs
 
 build: protos build-java build-docker build-html
 
+install-ci-dependencies: install-python-ci-dependencies install-go-ci-dependencies
```
ches: Was `install-java-ci-dependencies` needed here too?

woop: It's "unimplemented" right now because we actually need some build artifacts like `ingestion` to be available as dependencies. So if we run `install-java-ci-dependencies` it doesn't actually work. Not sure if there is a nice solution that doesn't involve actually building `ingestion` and `datatypes`.

woop: https://maven.apache.org/plugins/maven-dependency-plugin/go-offline-mojo.html#excludeGroupIds might be a possible solution.

ches: Hmm I see, so you want this image to cache dependencies, ideally.

Good news! After [a decade of the dependency plugin not supporting multi-module builds well for go-offline][1], a swell fellow [fixed it][2] a few months back. The fix is in the latest release, it's working for me after updating, sans `excludeGroupIds`.

I'll try to send a patch when I have a minute to actually test it with the built Docker image, but aside from updating this makefile the patch is just to bump the existing version `3.1.1` in the main `pom.xml` `<pluginManagement>` section to `3.1.2`.

Might need to remove the `cd`s into the subdirectories, I don't think Maven deals with it well—just `mvn dependency:go-offline` from top level should do it.

[1]: https://issues.apache.org/jira/browse/MDEP-204
[2]: https://github.com/apache/maven-dependency-plugin/pull/20

woop: Above doesn't work, also tried https://github.com/qaware/go-offline-maven-plugin

I think I will just use `verify` for now

PR #560, opened 2020-03-21
Implement unit tests on GitHub Actions

woop:

Continuation of #556, introduces GitHub Actions for unit testing Feast.


woop:
After integrating unit tests for Go, Java, Python, I realized that our existing build had a tight coupling to GCP for Python tests. The batch serving test requires a service account. For the time being I have disabled that test since it shouldnt be necessary that people have that in a unit test.

sdk/python/tests/test_client.py
```diff
@@ -307,131 +295,131 @@ def test_get_feature_set(self, mocked_client, mocker):
...
-        actual_dataframe = response.to_dataframe()
-
-        assert actual_dataframe[
-            ["my_project/customer_feature_1:1", "my_project/customer_feature_2:1"]
-        ].equals(
-            expected_dataframe[
-                ["my_project/customer_feature_1:1", "my_project/customer_feature_2:1"]
-            ]
-        )
+    # @pytest.mark.parametrize(
```
khorshuheng: Why is this commented out rather than removed?

woop: Because we need this test. The problem is that the test should be fixed instead of removed entirely, but I think its out of scope for this PR.

woop: Alternatively we can delete it and depend entirely on e2e, but I think having unit tests for these methods makes sense. Some parts of this Python SDK are not properly tested though, given the high amount of IO

infra/docker/ci/Dockerfile
```diff
@@ -23,13 +21,25 @@ RUN apt-get install -y build-essential curl python${PYTHON_VERSION} \
...
-# Add contents of local Feast repository to image (execute from Feast root)
-COPY . /feast/
-
-# Install all dependencies
-RUN cd /feast && make install-ci-dependencies
\ No newline at end of file
+ENV PATH="$HOME/bin:${PATH}"
+
+# Install Protoc and Plugins
+ENV PROTOC_VERSION 3.11.4
```
khorshuheng: Should we use 3.11.4 or 3.10.0?

woop: Yea, 3.10 would be better. Let me see if that works.

PR #565, opened 2020-03-24
Add pollingInterval config option

### LINKED ISSUE #531: Parameterize core async job worker poll interval
It's currently hardcoded to 1 minute. We should make it parameterizable - immediate usecase is when we run end to end tests, we want to set the polling interval to be shorter than that.

Extract that value into the configuration.

zhilingc:
Adds config option for JobCoodinator polling interval, and shortens the interval for e2e tests to reduce flakiness of those tests.

Fixes #531

```release-note
Adds new configuration option feast.jobs.updates.pollingIntervalMillis
```

### DIFF
--- a/core/src/main/resources/application.yml
+++ b/core/src/main/resources/application.yml
@@ -31,6 +31,8 @@ feast:
     # Key-value dict of job options to be passed to the population jobs.
     options: {}
     updates:
+      # Job update polling interval in milliseconds: how often Feast checks if new jobs should be sent to the runner.
+      pollingIntervalMillis: 60000
       # Timeout in seconds for each attempt to update or submit a new job to the runner.
       timeoutSeconds: 240
     metrics:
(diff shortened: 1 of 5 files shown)


core/src/main/resources/application.yml
```diff
@@ -31,6 +31,8 @@ feast:
     # Key-value dict of job options to be passed to the population jobs.
     options: {}
     updates:
+      # Job update polling interval in milliseconds: how often Feast checks if new jobs should be sent to the runner.
+      pollingIntervalMillis: 60000
```
ches: Spring [has support for duration expressions](https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-conversion-duration) to make these things more human-friendly, users could write `pollingInterval: 1m`.

PR #566, opened 2020-03-24
Add index for join table for jobs-featuresets relation

zhilingc:
With a large amount of feature sets and jobs, any operation on the jobs table takes a ridiculous amount of time - due to the fact that the join table is not indexed, resulting in very long JobCoordinator `poll()` calls. This PR specifies indexes for the two columns in the join table, which should speed up queries.

Tested locally and this should be an automatic upgrade. No migration is required.

```release-note
NONE
```


woop:
lgtm

PR #110, opened 2019-01-27
Refactor stores

tims:
This is for #109 

/hold

Issue #525, opened 2020-03-09
Clean up Feast configuration


woop:
Creating this card to clean up the Feast Core and Feast Serving configuration files. The configuration files have the following problems
- No clear separation of static Spring or JVM configuration from user level configuration
- Duplicate configuration points for the same thing (BigQuery, or Dataflow configuration)
- Lack of documentation of the options that are configurable.
- Lack of documentation on where new configuration should be placed.
- [EDIT] Adding more scope to this refactor. Configuration should also be easy to define and pass to stores. Validation should happen at the storage layer, not in Feast Core or Feast Serving. This change should make it easier to support stores as interfaces, possibly with factories building the stores/retrievers based on configuration. 

This issue tracks the refactor, cleanup, and documentation to solve the above.

mrzzy:
Not sure if this should be a separate issue, but this is definitely a pain point. There is almost no documentation on [feast.dev](feast.dev) the configuring and using Dataflow runner except that it is faster and more reliable than DirectRunner. Had to reverse engineer the config file from `core/src/main/java/feast/core/config/JobConfig.java`

woop:
Yea, that is definitely a major pain point. Thanks for pointing it out and being so detailed.

woop:
Throwing this into the ring as well. The following BQ configuration parameters must be set for BQ retrieval to work.
```
bigquery-initial-retry-delay-secs: 
bigquery-total-timeout-secs: 21600
```

woop:
Two more notes. 
- We can generate configuration metadata from @ConfigurationProperties annotated classes (like FeastProperties). This produces a JSON with keys/descriptions/defaults for all `feast` keys in `application.yml`. We could easily convert that into markdown and publish it to docs.feast.dev. I
- We should ensure that our JavaDocs are up to date. It's currently not being generated.

ches:
> * We can generate configuration metadata from @ConfigurationProperties annotated classes (like FeastProperties)

The account which someone apparently created to catch accidental mentions like that one is my favorite thing on the Internet today.

Returning to my seat in the peanut gallery now.

woop:
> > * We can generate configuration metadata from @ConfigurationProperties annotated classes (like FeastProperties)
> 
> The account which someone apparently created to catch accidental mentions like that one is my favorite thing on the Internet today.
> 
> Returning to my seat in the peanut gallery now.

Just noticed it, hahahahahahha

woop:
Keeping this open for the time being until #458 is done.

[...]

Issue #482, opened 2020-02-19
Feast API: Adding a new historical store


woop:
## 1. Introduction
We've had a lot of demand for either open source or AWS batch stores (#367, #259). Folks from the community have asked us how they can contribute code to add their stores types. 

In this issue I will walk through how batch stores are currently being used and how a new batch store type can be added.

## 2. Overview
Feast interacts with a batch store in two places
- **Data ingestion:** Ingestion jobs that load data into stores must be able to locate stores, apply migrations, and write data into feature set tables.
- **Feature serving (batch):** Feast serving executes batch retrieval jobs in order for users to export historical feature data.

## 3. Data ingestion
Feast creates and manages population jobs that stream in data from upstream data sources. Currently Feast only supports Kafka as a data source, meaning these jobs are all long running. Batch ingestion pushes data to Kafka topics after which they are picked up by these "population" jobs.

In order for the ingestion + population flow to complete, the destination store must be writable. This means that Feast must be able to create the appropriate tables/schemas in the store and also write data from the population job into the store.

Currently Feast Core starts and manages these population jobs that ingest data into stores, although we are planning to move this responsibility to the serving layer. Feast Core starts an Apache Beam job which synchronously runs migrations on the destination store and subsequently starts consuming from Kafka and publishing records.

Below is a "happy-path" example of a batch ingestion process:
![Untitled (1)](https://user-images.githubusercontent.com/6728866/74807906-91e73c00-5324-11ea-8ba5-2b43c7c5282b.png)

In order to accommodate a new store type, the Apache Beam job needs to be updated to support
- Setup (create tables/schemas): The current implementation for BigQuery/Redis is captured in  [StoreUtil.java](https://github.com/gojek/feast/blob/master/ingestion/src/main/java/feast/ingestion/utils/StoreUtil.java)
- Writes: A store specific client needs to be implemented that can write to a new store type in [WriteToStore.java](https://github.com/gojek/feast/blob/master/ingestion/src/main/java/feast/ingestion/transform/WriteToStore.java#L90)

## 4. Feature serving (batch)

Feast Serving is a web service that allows for the retrieval of feature data from a batch feature store. Below is a sequence diagram for a typical feature request from a batch store.

![Untitled](https://user-images.githubusercontent.com/6728866/74797157-702a8c80-5305-11ea-8901-bf6f4eb075f9.png)

[...]

PR #567, opened 2020-03-25
Add general storage API and refactor existing store implementations

### LINKED ISSUE #489: Shift ingestion feature metrics to after store write

zhilingc:
This PR adds an API for implementing various backends for feast, adds the implementations for redis and bigquery, and refactors the ingestion and serving code to use this API. It also introduces a fix for issue #489.

Outstanding issues with this PR:
- [ ] Connectors have beam dependencies to implement the write operations. Since serving has no need for these, the beam dependencies are excluded.
- [x] Stores like BQ do not expose success metrics because BQIO does not emit successful rows. As a fix for this, PR #568 adds pre-write metrics to track values for stores that have such limitations.
- [ ] No Service Loading yet
- [ ] Can implement a store that serves both online and batch features, but no way to serve both on Serving
- [x] Deadletter functionality not refactored to use API yet

Fixes #489 
Fixes #110 

```release-note
NONE
```


ches:
I didn't look too closely at the code dependencies yet, but is it feasible to move the metrics namespacing commit out of this PR to later, or could it even apply directly to master directly? It seems useful orthogonally to the storage refactor and would help us keep review focus on the central subject.

zhilingc:
> I didn't look too closely at the code dependencies yet, but is it feasible to move the metrics namespacing commit out of this PR to later, or could it even apply to master directly? It seems useful orthogonally to the storage refactor and would help us keep review focus on the central subject.

It's included in this PR because BQ doesn't support success metrics, so i needed to write metrics elsewhere as well, but i can reverse that merge first until you're done looking through this one.

ches:
> It's included in this PR because BQ doesn't support success metrics, so i needed to write metrics elsewhere as well, but i can reverse that merge first until you're done looking through this one.

I still don't understand how this is specific to the scope of storage modularization effort then—if it doesn't work for BigQuery, then how does it work for current master? If there are no success metrics for BQ on current master, then this is a new feature, not refactoring.

zhilingc:
> > It's included in this PR because BQ doesn't support success metrics, so i needed to write metrics elsewhere as well, but i can reverse that merge first until you're done looking through this one.
> 
> I still don't understand how this is specific to the scope of storage modularization effort then—if it doesn't work for BigQuery, then how does it work for current master? If there are no success metrics for BQ on current master, then this is a new feature, not refactoring.

No, it's because in this storage refactor PR i moved the original metrics from before the write to after, because I had to change the structure of the pipeline. (These post-write metrics are admittedly a new feature). The metrics namespacing PR adds the pre-write metrics back into the pipeline to ensure that users writing to bigquery get at least some visibility into what they're pushing into the sink.

storage/api/src/main/java/feast/storage/api/retrieval/HistoricalRetrievalResult.java
```diff
@@ -0,0 +1,100 @@
...
+
+  public abstract String getId();
+
+  public abstract JobStatus getStatus();
+
+  @Nullable
+  public abstract String getError();
+
+  @Nullable
+  public abstract List<String> getFileUris();
```
ches: Not a full review, but before forgetting I just wanted to carry over a point I had on #569 that's more appropriate to discuss in the wider context here:

> I feel a little reserved though about committing to the pattern of out-of-band file-based result data as API, i.e. `getFileUris()`. Not sure if large data sets will ever be streamed through RPC responses, but either of these options compromise the possibility for user jobs doing retrieval and processing the results to leverage data locality in systems where that could be achieved—see discussion starting at https://github.com/gojek/feast/issues/482#issuecomment-590827523

woop: @ches As you have rightly mentioned in #482, maintaining data locality requires a discussion on its own. I fear that our specific requirements at Gojek have steered us towards an approach that may be less than ideal in an HDFS + Spark world. 

You've mentioned that you have been working on a retrieval layer internally that bypasses our historical retrieval and presumably continues to maintain locality. 

I'd be happy to start a new issue on the requirements we (Gojek) currently have on the historical retrieval side, but ideally we'd have yours as well. Between the two we would cover a large swathe of use cases I believe.

ches: Sure, that's very reasonable, it's a complex subject but one that I think we've framed sufficiently to start that. I'm far from having a design proposal without requirements all laid out, but I can contribute to starting an RFC with problem statements.

For the sake of letting this PR move forward, maybe we can document `getFileUris` as a provisional API with a `@see` pointing to the new issue. In the worst case an implementation for which it doesn't make sense can throw an `UnsupportedOperationException` or the like and we can learn from use as this proceeds.

In our current approach with Hive, most likely we would not implement `HistoricalRetrievalResult` at all—we use Feast Ingestion but not Serving for historical, there is no `GetBatchFeatures` RPC call. It might be interesting to consider whether a similar call could be made mostly for Feast authorization purposes, metrics, etc., without expecting Serving to orchestrate the data staging. But, that's a matter for our new thread.

woop: >  It might be interesting to consider whether a similar call could be made mostly for Feast authorization purposes, metrics, etc., without expecting Serving to orchestrate the data staging. But, that's a matter for our new thread.

I'd really like to see if this is possible. 

By the way, how are you currently doing point in time correct joins? What triggers that execution?

ingestion/src/main/java/feast/ingestion/utils/StoreUtil.java
```diff
@@ -101,155 +78,19 @@
     VALUE_TYPE_TO_STANDARD_SQL_TYPE.put(Enum.BOOL_LIST, StandardSQLTypeName.BOOL);
   }
 
-  public static void setupStore(Store store, FeatureSet featureSet) {
+  public static FeatureSink getFeatureSink(
```
woop, line 81: I have mixed feelings about introducing a new term here (`FeatureSink`). Does it carry any unique meaning about a store?

serving/src/main/java/feast/serving/service/HistoricalServingService.java
```diff
@@ -0,0 +1,128 @@
...
+    this.specService = specService;
+    this.jobService = jobService;
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public GetFeastServingInfoResponse getFeastServingInfo(
+      GetFeastServingInfoRequest getFeastServingInfoRequest) {
+    try {
+      BigQueryHistoricalRetriever bigQueryHistoricalRetriever =
```
woop, line 54: Is it possible for us to isolate the conditional logic that instantiates specific store types and their corresponding retrievers to a specific location, perhaps `ServingServiceConfig.java`? This try/catch seems out of place.

woop:
Your check list (in the PR comment) is a bit ambiguous. It would be clearer if they were phrases as todo actions `To refactor deadletter functionality to use API` for example. Then a tick represents that it is done.

serving/src/main/java/feast/serving/service/HistoricalServingService.java
```diff
@@ -0,0 +1,128 @@
...
+          (BigQueryHistoricalRetriever) retriever;
+      return GetFeastServingInfoResponse.newBuilder()
+          .setType(FeastServingType.FEAST_SERVING_TYPE_BATCH)
+          .setJobStagingLocation(bigQueryHistoricalRetriever.jobStagingLocation())
+          .build();
+    } catch (Exception e) {
```
woop, line 60: This exception seems too broad. If there is a problem with a retriever then we should be logging it out.

serving/src/main/java/feast/serving/service/HistoricalServingService.java
```diff
@@ -0,0 +1,128 @@
...
+  private final JobService jobService;
+
+  public HistoricalServingService(
+      HistoricalRetriever retriever, CachedSpecService specService, JobService jobService) {
+    this.retriever = retriever;
+    this.specService = specService;
+    this.jobService = jobService;
+  }
+
+  /** {@inheritDoc} */
```
woop, line 49: I think it would be useful if we had comments or JavaDocs describing these methods in order for us to know the scope/responsibility.

serving/src/main/java/feast/serving/service/HistoricalServingService.java
```diff
@@ -0,0 +1,128 @@
...
+  @Override
+  public GetOnlineFeaturesResponse getOnlineFeatures(GetOnlineFeaturesRequest getFeaturesRequest) {
+    throw Status.UNIMPLEMENTED.withDescription("Method not implemented").asRuntimeException();
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public GetBatchFeaturesResponse getBatchFeatures(GetBatchFeaturesRequest getFeaturesRequest) {
```
woop, line 75: I'm finding it hard to limit comments here. This is probably out of scope for the storage refactor, but having the serving service do job management here (creating a job Id and kicking off the job) seems to be an incorrect abstraction. I think this should be a part of what the job service does.

serving/src/main/java/feast/serving/service/OnlineServingService.java
```diff
@@ -0,0 +1,162 @@
...
+      GetFeastServingInfoRequest getFeastServingInfoRequest) {
+    return GetFeastServingInfoResponse.newBuilder()
+        .setType(FeastServingType.FEAST_SERVING_TYPE_ONLINE)
+        .build();
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public GetOnlineFeaturesResponse getOnlineFeatures(GetOnlineFeaturesRequest request) {
+    try (Scope scope = tracer.buildSpan("Redis-getOnlineFeatures").startActive(true)) {
```
woop, line 68: Should this trace be made more generic or interpolated?

serving/src/main/java/feast/serving/service/OnlineServingService.java
```diff
@@ -0,0 +1,162 @@
...
+          specService.getFeatureSets(request.getFeaturesList());
+      List<EntityRow> entityRows = request.getEntityRowsList();
+      Map<EntityRow, Map<String, Value>> featureValuesMap =
+          entityRows.stream()
+              .collect(Collectors.toMap(row -> row, row -> Maps.newHashMap(row.getFieldsMap())));
+
+      List<List<FeatureRow>> featureRows =
+          retriever.getOnlineFeatures(entityRows, featureSetRequests);
+
+      for (var fsIdx = 0; fsIdx < featureRows.size(); fsIdx++) {
```
woop, line 81: I think this block of code would benefit from comments that describe what is happening.

serving/src/main/java/feast/serving/service/OnlineServingService.java
```diff
@@ -0,0 +1,162 @@
...
+          featureValuesMap.values().stream()
+              .map(valueMap -> FieldValues.newBuilder().putAllFields(valueMap).build())
+              .collect(Collectors.toList());
+      return getOnlineFeaturesResponseBuilder.addAllFieldValues(fieldValues).build();
+    }
+  }
+
+  @Override
+  public GetBatchFeaturesResponse getBatchFeatures(GetBatchFeaturesRequest getFeaturesRequest) {
+    throw Status.UNIMPLEMENTED.withDescription("Method not implemented").asRuntimeException();
```
woop, line 142: We could maybe use https://docs.oracle.com/javase/6/docs/api/java/lang/UnsupportedOperationException.html

woop:
I think the two key questions are
- The layer of abstraction that the storage API should encapsulate. At a minimum it should be I/O. Right now the scope is more than that if we include stream specific interfaces in anticipation of where and how the storage layer will be used. Making the storage API purely I/O is potentially one way to remove the Beam dependency, but I am not sure if its worth the added complexity it would add to ingestion.
- The second question is what level of modularity the storage API should support. If we want to exclude Beam from serving then we might need multiple submodules within the storage API for ingestion and serving. Right now the ingestion and serving layers are already cherry picking sub-dependencies in the form of `FeatureSink` and `Retriever`, so I think this should be possible. That being said I am not sure if its worth the added complexity, and my understanding of Java package and dependency management is rather limited so I am not sure if there are other technical challenges.

Other things to comment on
- Do we have a definition somewhere for what a `FeatureSink` and `Retriever` is? If we want users to extend these then we should describe these as part of the comments.
- `Retrieval` and `Write` are inconsistent with each other. I would prefer `readers`/`writers` or no abstraction at all. Open to ideas.
- Another inconsistency is that we are prefixing `FeatureSink` with `Feature` but with `Retriever` we aren't. In order to be consistent with `DeadletterSink` and to actually capture the intent here, I would prefer to use something like `StoreSink` or `StorageSink`. This avoids introducing a new concept since everyone already knows what a store is.

The idea of having the storage API consist out of a bunch of loosely defined interfaces is growing on me. If you have a look at the Presto connector interfaces you can see a pretty nice implementation. https://github.com/prestodb/presto/blob/master/presto-spi/src/main/java/com/facebook/presto/spi/connector/Connector.java. 

One of the reasons why I like their approach is because it will read a lot better from the consumer side. Instead of instantiating a retriever in serving you would instantiate a store. You would then ask the store for the retriever (or for the statistics retriever?). It can then let you know whether that functionality is implemented or even if you have a valid configuration to initialize it.

Another thing that we can consider adopting is more explicit interface names with a flat directory structure. Right now we are grouping interfaces into `retrieval` and `write`, but where would we house `statistics` functionality if we wanted to add that? Would we add another folder? It seems like there are overlapping concerns there because you would probably want to instantiate a single store from a single set of configurations and then you want to call it for both features and statistics.

woop:
One more comment. I think we should rename the title of this PR to be more descriptive. It will become a line in the changelog after all. @zhilingc I will leave that up to you.

ingestion/src/main/java/feast/ingestion/utils/StoreUtil.java
zhilingc: It's to distinguish stores that store features v.s. deadletter rows.

serving/src/main/java/feast/serving/service/OnlineServingService.java
zhilingc: No, we should be using the GRPC statuses so that they are correctly thrown back to users.

woop: Ah I didn't realize there was only `asRuntimeException` and nothing more granular.

serving/src/main/java/feast/serving/service/OnlineServingService.java
ches: @zhilingc You added docs that can resolve this comment, but could we now take it a step further and factor some parts out to helpers, or domain methods on classes involved?

Forgive the laundry list, I'm going to make suggestions before I start a line-by-line review that will take me awhile to finish:

- Pull the metrics out to helpers, they eat up a lot of lines especially in deeply-nested code and metrics always add noise distracting readers from the business logic. Style nit while we're on this: I think the static imports are less than helpful, on first pass I was looking for a local for `staleKeyCount`, if the usage site was `Metrics.staleKeyCount` I would not have needed to.
- Scan for variable extractions, there are a few things repeated, e.g. `featureSetRequest.getSpec().getProject()`, `featureSetRequest.getSpec().getMaxAge()`—nicely named locals will reduce the reading fatigue.
- Minor, but the part assigning `Map<String, FeatureReference> featureNames` could perhaps reasonably be an instance method on `FeatureSetRequest`, like `refsByName` or something (also a suggestion for the local identifier name here, `featureNames` sounds like a sequence to me rather than a map).
- `isStale` (which could currently be marked `static` btw) could also _maybe_ be a method on `FeatureSetRequest`, somehow… I don't like that as-is, especially because we've moved it to storage-api in this PR and `EntityRow` is a nested message of `GetOnlineFeaturesRequest`—that does not belong at that layer. Maybe we're missing an abstraction here.

   `isStale` is behavior that should be unit tested. It could be, as a static method of `OnlineServingService`, but it feels wrong: there's nothing binding the arguments as belonging together in an aggregate (in DDD sense of the word). In other words, it's easy for a caller to make a mistake in using it. How could we improve that? I think this belongs in a perhaps-missing domain service layer (again in DDD sense, this `OnlineServingService` is pretty much an MVC controller currently, and that's not where I want to test or rely on core business logic behavior of Feast).

ches: Note that what `isStale` really needs is the feature set spec in question, the `FeatureSetRequest` is incidental. That's a hint at the boundaries of a domain service layer, I'd say.

Somewhat random aside, `FeatureSetReference` is currently defined in `feast.serving.ServingAPIProto`—I think it probably belongs in `feast.types`. It was added for v0.4, unfortunate that moving it is probably a breaking change. At least Maven module-wise it's all one dependency on `datatypes-java`, it'll look out of place as an import in layers far removed from serving but otherwise little consequence.

serving/src/main/java/feast/serving/service/HistoricalServingService.java
ches: FWIW a comment only to do `@inheritdoc` is unnecessary, it is automatic if you don't add any new Javadoc comment. `@inheritdoc` allows you to extend with additions.

serving/src/main/java/feast/serving/service/HistoricalServingService.java
ches: Completely agree on both counts, the abstraction/layer boundaries and the scope of storage refactor.

I'm not sure where/when to address it, if we're going to have all this serving refactoring in the PR too.

woop: Well I just realized that my configuration clean up #525 is the yak shave of the century. I am trying to limit the scope of that as well, but I am touching all of the same places that @zhilingc is because configuration is heavily used by stores. I've even now rebased on top of this very PR.

I think we should try and limit the already large scope of this PR to only the most critical aspects of the storage API.

zhilingc: For now, for historical stores at least, the serving service *only* manages jobs, because historical serving is done async, and historical retrieval is analogous to starting a retrieval job. I'd argue that until there is a need for the two to be defined separately (e.g. if serving does some other job management), it's fine as is.

serving/src/main/java/feast/serving/service/OnlineServingService.java
zhilingc: `isStale` is definitely a bit of a conundrum here, I don't think it belongs to the `FeatureSetRequest` because I dont think its the responsibility of the request info object (which is for the most part a value object) to contain that sort of logic. 

Regarding `FeatureSetReference`, The recent PR #548 adds this as a datatype ([see here](https://github.com/gojek/feast/blob/1dd0f503a56ce4b55706317d5d4acde227305e91/protos/feast/core/FeatureSetReference.proto)). Once that PR has been merged in I will refactor this code to use it.

woop:
Moving ahead with merging this into master after the name changes.

PR #575, opened 2020-03-27
Fix runner to string inconsistency

### LINKED ISSUE #574: Converting Runner to String for assigning to Job.Runner is inconsistent.
## Expected Behavior 
One true way to convert `Runner` to string when assigning to `Job.Runner`.

## Current Behavior
There are multiple ways in which `Runner` is converted to string in feast core to be stored in `Job.runner` :
- `toString()`method which is used by [`JobUpdateTask`](https://github.com/gojek/feast/blob/9facd2e5f00c0616da2bb616664e619ee0100de6/core/src/main/java/feast/core/job/JobUpdateTask.java#L147) and `DirectJob`  runner.
- `getName()` method which is used by [`Dataflow`](https://github.com/gojek/feast/blob/9facd2e5f00c0616da2bb616664e619ee0100de6/core/src/main/java/feast/core/job/dataflow/DataflowJobManager.java#L204) runner

## Steps to reproduce
N.A.

## Possible Solution

Standardise on one true way to convert `Runner` to string when assigning to `Job.Runner`.

mrzzy:
- Fixes the inconsistency of converting `Runner` to string when assigning to `Job.runner` by converting all to use `Runner.name()`.
- Changes `Runner.toString()` to return a human readable string.
- Changes all uses of `Runner.getName()` for all non `Job.runner` applications (ie logging, config file) to use `Runner.toString()`
- Removes `Runner.getName()`

This is necessary to standardise the use of Runner.name() when
passing to the Job.runner, so that code depending on Job.runner
would know what to expect. 

Fixes #574 

```release-note
NONE
```

### DIFF
--- a/core/src/main/java/feast/core/job/Runner.java
+++ b/core/src/main/java/feast/core/job/Runner.java
@@ -29,16 +29,17 @@ public enum Runner {
 
   /**
    * Get the human readable name of this runner. Returns a human readable name of the runner that
-   * can be used for logging/config files.
+   * can be used for logging/config files/etc.
    */
-  public String getName() {
+  @Override
+  public String toString() {
     return name;
   }
 
   /** Parses a runner from its human readable name. */
   public static Runner fromString(String runner) {
     for (Runner r : Runner.values()) {
-      if (r.getName().equals(runner)) {
+      if (r.toString().equals(runner)) {
         return r;
       }
     }
--- a/core/src/main/java/feast/core/job/Runner.java
+++ b/core/src/main/java/feast/core/job/Runner.java
@@ -27,10 +27,15 @@ public enum Runner {
     this.name = name;
   }
 
+  /**
+   * Get the human readable name of this runner. Returns a human readable name of the runner that
+   * can be used for logging/config files.
+   */
   public String getName() {
     return name;
   }
 
+  /** Parses a runner from its human readable name. */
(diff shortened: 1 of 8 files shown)


core/src/main/java/feast/core/job/Runner.java
```diff
@@ -27,10 +27,15 @@
     this.name = name;
   }
 
+  /**
+   * Get the human readable name of this runner. Returns a human readable name of the runner that
+   * can be used for logging/config files.
```
pradithya: I think your approach should be the other way around. As a best practice, it's not a good idea to make your application logic depend on `toString()` . Use `toString()` for the purpose of being read by human

mrzzy: The codebase currently favors using `getName()` for human readable applications like logging and config. However, I think that your point about  `toString()` is valid as we don't want to end in a situation where our code breaks because someone overwrote `toString()` to make it human readable.

Hence I propose that we:
- replace `getName()` with `toString()`
- use `name()` when converting to string  for Job.runner as it cannot be overridden
> Reference: https://stackoverflow.com/questions/13291076/java-enum-why-use-tostring-instead-of-name

mrzzy:
If backporting, please note that this PR introduced a bug that is fixed in PR #578

PR #586, opened 2020-03-29
Change pipeline to use DeadletterSink API

zhilingc:
Changes the population job pipeline to use the deadletter API. Doesn't add option for deadletter type yet, will hold off on that until we implement a new deadletter sink.


woop:
Hmmm thats weird. Your GitHub Action tests aren't running.

PR #593, opened 2020-03-31
Enable Mypy lint

### LINKED ISSUE #594: Enable Mypy linter on the lint-python side
## Expected Behavior 
The mypy python linter is enabled

## Current Behavior
Saw the TODO on the makefile

## Steps to reproduce
- run `make lint-python` should show mypy run

## Possible Solution
- Enable the mypy linter, and add type hint for all the functions. 
- Make implicit type hint become explicit

imjuanleonard:
Will help to know the type of functions explicitly

Fixes #594

```
None
```

### DIFF
--- a/sdk/python/feast/client.py
+++ b/sdk/python/feast/client.py
@@ -821,8 +821,8 @@ def _build_feature_references(
 
         feature_split = feature_version.split(":")
         if len(feature_split) == 2:
-            name, version = feature_split
-            version = int(version)
+            name = feature_split[0]
+            version = int(feature_split[1])
         elif len(feature_split) == 1:
             name = feature_split[0]
         else:
--- a/sdk/python/feast/client.py
+++ b/sdk/python/feast/client.py
@@ -20,7 +20,7 @@ import tempfile
 import time
 from collections import OrderedDict
 from math import ceil
-from typing import Dict, List, Optional, Tuple, Union
+from typing import cast, Dict, List, Optional, Tuple, Union
 
 import grpc
 import pandas as pd
@@ -728,7 +728,7 @@ class Client:
         while True:
             if timeout is not None and time.time() - current_time >= timeout:
                 raise TimeoutError("Timed out waiting for feature set to be ready")
-            feature_set = self.get_feature_set(name, version)
+            feature_set = cast(FeatureSet, self.get_feature_set(name, version))
             if (
                 feature_set is not None
                 and feature_set.status == FeatureSetStatus.STATUS_READY
--- a/sdk/python/feast/client.py
+++ b/sdk/python/feast/client.py
@@ -73,7 +73,20 @@ from feast.serving.ServingService_pb2_grpc import ServingServiceStub
 
 _logger = logging.getLogger(__name__)
 
-CPU_COUNT = os.cpu_count()  # type: int
+
+def get_max_cpu_workers() -> int:
+    """
+    Get the number of CPU count in an OS
+
+    Returns:
+        int:
+            the number of CPU worker in OS, the minimum value is 1, and maximum is total CPU count - 1
+    """
+    cpu_count = os.cpu_count()
+    if isinstance(cpu_count, int) and (cpu_count - 1 > 0):
+        return cpu_count - 1
+    else:
+        return 1
 
 
 class Client:
@@ -655,7 +668,7 @@ class Client:
         chunk_size: int = 10000,
         version: int = None,
         force_update: bool = False,
-        max_workers: int = max(CPU_COUNT - 1, 1),
+        max_workers: int = get_max_cpu_workers(),
         disable_progress_bar: bool = False,
         timeout: int = KAFKA_CHUNK_PRODUCTION_TIMEOUT,
     ) -> None:
--- a/sdk/python/feast/client.py
+++ b/sdk/python/feast/client.py
@@ -737,7 +737,7 @@ class Client:
                 )
                 self.apply(feature_set)
             else:
-                raise Exception("Use feature set object to do force_update")
+                raise TypeError("Use FeatureSet Object to do force_update")
         current_time = time.time()
 
         print("Waiting for feature set to be ready for ingestion...")
--- a/sdk/python/feast/client.py
+++ b/sdk/python/feast/client.py
@@ -729,12 +729,15 @@ class Client:
 
         # Update the feature set based on PyArrow table of first row group
         if force_update:
-            feature_set.infer_fields_from_pa(
-                table=pq_file.read_row_group(0),
-                discard_unused_fields=True,
-                replace_existing_features=True,
-            )
-            self.apply(feature_set)
+            if isinstance(feature_set, FeatureSet):
+                feature_set.infer_fields_from_pa(
+                    table=pq_file.read_row_group(0),
+                    discard_unused_fields=True,
+                    replace_existing_features=True,
+                )
+                self.apply(feature_set)
+            else:
+                raise Exception("Use feature set object to do force_update")
         current_time = time.time()
 
         print("Waiting for feature set to be ready for ingestion...")
--- a/sdk/python/feast/client.py
+++ b/sdk/python/feast/client.py
@@ -576,7 +576,7 @@ class Client:
 
         if serving_info.type != FeastServingType.FEAST_SERVING_TYPE_BATCH:
             raise Exception(
-                f'You are connected to a store "{self._serving_url}" which '
+                f'You are connected to a store "{self.serving_url}" which '
                 f"does not support batch retrieval "
             )
 
--- a/sdk/python/feast/client.py
+++ b/sdk/python/feast/client.py
@@ -74,21 +74,6 @@ from feast.serving.ServingService_pb2_grpc import ServingServiceStub
 _logger = logging.getLogger(__name__)
 
 
-def get_max_cpu_workers() -> int:
-    """
-    Get the number of CPU count in an OS
-
-    Returns:
-        int:
-            the number of CPU worker in OS, the minimum value is 1, and maximum is total CPU count - 1
-    """
-    cpu_count = os.cpu_count()
-    if isinstance(cpu_count, int) and (cpu_count - 1 > 0):
-        return cpu_count - 1
-    else:
-        return 1
-
-
 class Client:
     """
     Feast Client: Used for creating, managing, and retrieving features.
@@ -668,7 +653,7 @@ class Client:
         chunk_size: int = 10000,
         version: int = None,
         force_update: bool = False,
-        max_workers: int = get_max_cpu_workers(),
+        max_workers: int = None,
         disable_progress_bar: bool = False,
         timeout: int = KAFKA_CHUNK_PRODUCTION_TIMEOUT,
     ) -> None:
@@ -710,6 +695,9 @@ class Client:
             None:
                 None
         """
+
+        if max_workers is None:
+            max_workers = _get_max_cpu_workers()
 
         if isinstance(feature_set, FeatureSet):
             name = feature_set.name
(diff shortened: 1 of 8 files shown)


sdk/python/feast/client.py
```diff
@@ -73,7 +73,20 @@
 
 _logger = logging.getLogger(__name__)
 
-CPU_COUNT = os.cpu_count()  # type: int
+
+def get_max_cpu_workers() -> int:
```
woop, line 77: Can we move this to another file in the package?

sdk/python/feast/client.py
```diff
@@ -716,22 +729,25 @@ def ingest(
...
+            else:
+                raise Exception("Use feature set object to do force_update")
         current_time = time.time()
 
         print("Waiting for feature set to be ready for ingestion...")
         while True:
             if timeout is not None and time.time() - current_time >= timeout:
                 raise TimeoutError("Timed out waiting for feature set to be ready")
-            feature_set = self.get_feature_set(name, version)
+            feature_set_spec = self.get_feature_set(name, version)
```
woop, line 747: why the rename?

sdk/python/feast/feature_set.py
```diff
@@ -296,7 +296,7 @@ def drop(self, name: str):
             del self._fields[name]
             return
 
-    def _add_fields(self, fields: List[Field]):
+    def _add_fields(self, fields):
```
woop, line 299: What happened to the type?

sdk/python/feast/feature_set.py
```diff
@@ -306,6 +306,33 @@ def _add_fields(self, fields: List[Field]):
         for field in fields:
             self.add(field)
 
+    def _entities_and_features_to_fields(
```
woop, line 309: Why is this a method and not a function?

woop: It doesn't seem like this will be reused.

sdk/python/feast/feature_set.py
```diff
@@ -361,8 +388,9 @@ def infer_fields_from_df(
 
         # Create dictionary of fields that will not be inferred (manually set)
         provided_fields = OrderedDict()
+        fields = self._entities_and_features_to_fields(entities, features)
 
-        for field in entities + features:
+        for field in fields:
```
woop, line 393: You are testing all of the fields twice...

sdk/python/feast/job.py
```diff
@@ -172,8 +173,11 @@ def to_chunked_dataframe(
             pd.DataFrame:
                 Pandas DataFrame of the feature values.
         """
+
+        # Object is Avro row type object, refer to self.result function for this type
+        records: List[object] = []
```
woop, line 178: Why are you using object instead of the type?

sdk/python/feast/type_map.py
```diff
@@ -444,10 +444,14 @@ def pa_to_feast_value_type(value: object) -> ValueType:
         "list<item: binary>": ValueType.BYTES_LIST,
         "list<item: bool>": ValueType.BOOL_LIST,
     }
-    return type_map[value.type.__str__()]
+
+    if isinstance(value, pa.lib.ChunkedArray):
```
woop, line 448: Why isn't this check done in the first line (or even in the argument as a mypy type)?

sdk/python/feast/client.py
imjuanleonard: Do you have any package suggestions? 

To be honest, I am new to the python code structure standard(if there is any), so I keep the function on the same place as it was before

sdk/python/feast/client.py
imjuanleonard: The `feature_set` is a parameter already defined on the function arguments.

Mypy considers it is a bad practice when we reuse the same variable but assign it to a different type. We will lose track of what the `feature_set` type is. Hence the new variable creation

sdk/python/feast/feature_set.py
imjuanleonard: There are 2 way of doing this previously:

1. Whenever we create the `_add_fields` the List [feature/entity] need to be converted to field first
2. Remove the type check for the fields part, and let the add handles it. Inside this function, there is a single add function where it checks the type of the field.

sdk/python/feast/feature_set.py
imjuanleonard: Ok, at first putting underscore is a private function. Because it is only being used inside this class, have just re-read about method and function, I am rewriting it to a function.

Thanks

sdk/python/feast/feature_set.py
imjuanleonard: Ah by the way, can you elaborate more on this? what do you mean?

Is it the `is_instance` check?

sdk/python/feast/type_map.py
imjuanleonard: Ah okay, because it is only getting used in 1 place, I have moved the `pa.lib.ChunkedArray` to the arguments. Have rebased the commit

sdk/python/feast/job.py
imjuanleonard: The idea was:
1. The Avro Row will be populated to object thus overriding the type
2. The records don't call any method, it is just being passed around, so it is safe
3. After checking all the tests pass (need to tests for the e2e as well thought)
4. I cannot find what is AvroRow filetype, do you know?
I have tried to dig deeper into the class until I found it was reading only from `args` and `kwargs` without return type, is it byte?

sdk/python/feast/client.py
woop: Please come up with proposals

sdk/python/feast/client.py
```diff
@@ -716,22 +729,25 @@ def ingest(
...
-            self.apply(feature_set)
+            if isinstance(feature_set, FeatureSet):
+                feature_set.infer_fields_from_pa(
+                    table=pq_file.read_row_group(0),
+                    discard_unused_fields=True,
+                    replace_existing_features=True,
+                )
+                self.apply(feature_set)
+            else:
+                raise Exception("Use feature set object to do force_update")
```
woop, line 740: We can use TypeError here.

sdk/python/feast/client.py
```diff
@@ -716,22 +729,25 @@ def ingest(
 
         # Update the feature set based on PyArrow table of first row group
         if force_update:
-            feature_set.infer_fields_from_pa(
-                table=pq_file.read_row_group(0),
-                discard_unused_fields=True,
-                replace_existing_features=True,
-            )
-            self.apply(feature_set)
+            if isinstance(feature_set, FeatureSet):
```
woop, line 732: I think we can probably safely remove force_update here. Its better if people run that as a separate step. Not sure if anyone is using it right now internally, and I think as written it was partly broken any ways.

sdk/python/feast/client.py
woop: The spec is used typically for the proto representation. Can you think of a different way to solve this? I think we should be fetching the feature set from Feast Core if they provide a string. That would also solve the exception above.

sdk/python/feast/feature_set.py
woop: Fields contains all the features and entities. You are looping over it twice. I dont think you need `entities_and_features_to_fields`. You can just loop over entities and features.

sdk/python/feast/job.py
woop: Types are there to add safety. It seems like you are looking for a shortcut here.

sdk/python/feast/feature_set.py
imjuanleonard: The mypy convention was because it cannot handle the `entities + features` syntax where we process both of them at the same time, separating into two iteration also require both the feature and entity to be aggregated, which would create the same number of iteration.
So the field converter was created. 

I don't think it would affect the performance, as field and entity is limited, the maximum of N in the O(N) is very less (less than 1000 feature + entity) also 2 converter iteration is very fast process

sdk/python/feast/client.py
imjuanleonard: okay, adding the type error

sdk/python/feast/job.py
imjuanleonard: Okay, let me try to simulate this function call to Avro row, and print the type, let me see what is the type.

Although I am certain object should be sufficient, we can see the type to make sure

sdk/python/feast/client.py
imjuanleonard: I think we can raise this as a new Issue? so the linter error will be separate case

sdk/python/feast/client.py
imjuanleonard: I have tried to use the above solution actually. Even though I add the check for the `feature_set` to make sure that it was `FeatureSet` object, the linter won't acknowledge the new type, because the type was defined in the parameter, so reusing the same variable for different type was not allowed.

I think we can change the name? Are you happy if we add `_` in front of the variable?
 ```suggestion
            _feature_set = self.get_feature_set(name, version)
```

sdk/python/feast/client.py
imjuanleonard: Because this is only one function According [Python Packaging Authority](https://www.pypa.io/en/latest/) and their sample project pip https://github.com/pypa/pip/tree/master/src/pip/_internal/utils

We can create have `utils` folder. Will create `utils` folder and inside it, there would be `os.py` which will  contain this function

Most of the other references allow it to be this way first because it is still 1 small function and only getting used in 1 place. When we have multiple functions then it would be a good time to think about the structure

Do let me know which one you're good, I am happy with any of the options

woop: How about we make its access private and move it to the bottom of the file?

sdk/python/feast/client.py
woop: Ok, can you do that please.

sdk/python/feast/client.py
imjuanleonard: cool, I am happy with that as well

sdk/python/feast/client.py
woop: Can you either use `cast` or `assert isinstance`?

https://github.com/python/mypy/issues/2589#issuecomment-268044680
https://mypy.readthedocs.io/en/stable/casts.html

sdk/python/feast/feature_set.py
woop: I am going to increase the scope here a bit, so I apologies for that, but I'd like you to do a quick refactor here.

In a new PR I'd like to make the following changes to remove redundant methods that just add complexity

Can you remove
- `_add_fields()` 
- `add()`

Move the logic from `add` into `entities` and `features` setter.

What do you think? that would make things a lot cleaner, although it will be a breaking change.

sdk/python/feast/job.py
woop: > Although I am certain object should be sufficient
Sufficient to do what?

sdk/python/feast/feature_set.py
woop: Over time we will move away from Fields because we want to treat entities and features separately. Which is why I would prefer to do validation and building of this `dict` from within the method itself, and not to create a new method. For the time being this proposal is ok

Can we at least rename this `entities_and_features_to_fields` so that it is more descriptive?
`_create_field_list` or something?

sdk/python/feast/client.py
imjuanleonard: Cool! Thanks, the cast works like charm, will keep it notes

sdk/python/feast/feature_set.py
imjuanleonard: Ah okay, got your point. Is it, we want to move away from Fields by separating Entities and Features?

Yeah definitely it is way cleaner, but will probably have breaking changes if not careful. 
Will raise the issue to keep track in a new PR

woop: Ok, please be descriptive in your issue.

sdk/python/feast/job.py
imjuanleonard: > Sufficient to do what?

O yeah, the object is not sufficient, I thought it will inherit everything, but it doesn't allow you to inherit the function as well.

 I just checked it out, the fastreader.reader will return `List of dict` hence, have rebased the commit. 

Thanks

sdk/python/feast/client.py
imjuanleonard: This is done

woop:
Closing in favor of #749

PR #596, opened 2020-04-02
Create project if not exists on applyFeatureSet

### LINKED ISSUE #453: Allow create_project to rerun safely
Having the `create_project` method on the Feast Client fail when a project exists is important, but its also a pain in development. It means one part of the code in a notebook should be skipped every time.

It would be convenient to be able to run `create_project` multiple times without having the second call throw an exception.

Something like
```
client.create_project("my-project", skip_if_exists=True)
```

Alternatively we can use `set_project` and only create the project lazily when its needed.

Joostrothweiler:
Allow create_project to rerun safely #453. Instead of having to run `create_project` before apply, the apply method can create the project if it does not already exist.

Fixes #453 

```release-note
When applying a featureSet to a project that does not yet exist, Feast create the project on apply.
```

### DIFF
--- a/core/src/main/java/feast/core/service/SpecService.java
+++ b/core/src/main/java/feast/core/service/SpecService.java
@@ -288,7 +288,7 @@ public class SpecService {
     // Validate incoming feature set
     FeatureSetValidator.validateSpec(newFeatureSet);
 
-    // Find project or create new one if not exists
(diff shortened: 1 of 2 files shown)


woop:
Awesome, thanks @Joostrothweiler!

PR #601, opened 2020-04-03
Add better code docs to storage refactor

zhilingc:
Reopening this after rebasing on master.
Addresses most of @woop 's comments on PR #567

### DIFF
--- a/serving/src/main/java/feast/serving/configuration/ServingServiceConfig.java
+++ b/serving/src/main/java/feast/serving/configuration/ServingServiceConfig.java
@@ -117,8 +117,7 @@ public class ServingServiceConfig {
                 .setStorage(storage)
                 .build();
 
-        servingService =
-            new HistoricalServingService(bqRetriever, specService, jobService);
+        servingService = new HistoricalServingService(bqRetriever, specService, jobService);
         break;
       case CASSANDRA:
       case UNRECOGNIZED:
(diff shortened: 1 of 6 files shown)


woop:
Let me know if you want this reviewed or not.

zhilingc:
> Let me know if you want this reviewed or not.

Yeah, I had to close the other one because of some issues rebasing. I added changes based on your comment already

serving/src/main/java/feast/serving/configuration/ServingServiceConfig.java
```diff
@@ -117,7 +117,8 @@ public ServingService servingService(
                 .setStorage(storage)
                 .build();
 
-        servingService = new HistoricalServingService(bqRetriever, specService, jobService);
+        servingService =
+            new HistoricalServingService(bqRetriever, specService, jobService, jobStagingLocation);
```
woop, line 121: Trust me I really want to greenlight this PR, but I don't think adding `stagingLocation` directly to the `HistoricalServingService` is correct here. That state is already stored within the retriever, which means it is also available to the HistoricalServingService if we expose it.

woop: If you are happy with me making the change then I can quickly do it and you can see if you agree with the structure?

PR #609, opened 2020-04-07
Backport ci changes

khorshuheng:
Move CI scripts from .prow to infra


woop:
This is a lot more than just CI, it also contains Python SDK changes.

khorshuheng:
> This is a lot more than just CI, it also contains Python SDK changes.

I ended up doing this manually rather than cherry pick the commits, as there are quite a lot of conflicts. I noticed that Python and Golang test no longer appears, not too sure why is that the case though.

PR #610, opened 2020-04-08
Add Format and Lint to Makefile (#545)

khorshuheng:
Backport #545


sdk/go/protos/feast/core/Store.pb.go
```diff
@@ -296,6 +301,20 @@ func (m *Store_RedisConfig) GetPort() int32 {
 	return 0
 }
 
+func (m *Store_RedisConfig) GetInitialBackoffMs() int32 {
```
ches, line 304: There was a bit of codegen that was stale here I guess. Also a few Go module version updates, I think the Go SDK is kind of dormant so those aren't significant.

ches:
Ah, looks like Google credentials are a problem for the Actions build though.

woop:
> Ah, looks like Google credentials are a problem for the Actions build though.

On master I have removed those tests. I know that isn't the ideal situation, but the functionality should still be covered by e2e

https://github.com/gojek/feast/blob/master/sdk/python/tests/test_client.py#L407

Relevant issue: https://github.com/gojek/feast/issues/562

PR #533, opened 2020-03-13
Refactor Feast Helm charts for better end user install experience

davidheryanto:
**What this PR does / why we need it**:

1. Use --spring.config.additional-location to specify Spring application config
   so the default application.yaml bundled in the jar can be used, and users
   only need to override required config. This is to prevent chart users being
   overwhelmed by the amount of config to set.

2. Use less templating, if else, and functions in configmap.yaml in Feast
   charts. Instead set a reasonable default and allow users to override
   in application-override.yaml. This makes it easier to manage when
   application.yaml structure changes and documentation for application.yaml
   can be delegated to:
   https://github.com/gojek/feast/blob/master/core/src/main/resources/application.yml
   https://github.com/gojek/feast/blob/master/serving/src/main/resources/application.yml

3. Remove double nesting of charts. Previously, Postgresql and Kafka subcharts
   are bundled in Feast Core suchart. Now all 3 subcharts are sibling charts.
   This reduces coupling between charts and it's easier to configure chart
   values one at a time than one big values yaml. Also, documentation for
   third party subcharts can be delegated to the original authors.

4. Use helm-docs to generate README for Feast charts so the documentation
   is reproducible and easier to keep up to date when values.yaml changes.
   It is also easier to follow best practices when writing values.yaml.
   https://github.com/norwoodj/helm-docs

5. Add prometheus and grafana charts to support out of the box metrics
   collector and visualization when users install Feast with this chart.

6. Add tests for users to verify installation:

7. Add examples to install Feast with 3 different configurations (from simple to complex):
   - online serving only with direct runner
   - online and batch serving with DirectRunner
   - online and batch serving with DataflowRunner

> To check how the Chart documentation will look like
> https://github.com/davidheryanto/feast/tree/helm-chart-v0.4-refactor/infra/charts/feast

**Which issue(s) this PR fixes**:

NONE

**Does this PR introduce a user-facing change?**:

Yes, the format of Helm values passed during installation has changed, from nested to flatter structure. But in general, the accepted values in each chart are the same.

PR #611, opened 2020-04-11
Clean up Feast configuration

### LINKED ISSUE #525: Clean up Feast configuration

woop:

This PR is meant to resolve #525 and should be merged on only after the #567 and #533 PRs have been merged in.

The major changes will be that it will restructure and clean up Feast configuration properties in order for developers to find it easier to configure a Feast deployment and easier to know when it is misconfigured.

Additionally I am attempting to create a consistent way to instantiate and configure stores.

Fixes #525

```release-note
Both Feast Serving and Feast Core have breaking changes to the structure of their application.yml files.
```


woop:
Still to do: Need to make all keys in application.yml use a consistent separator `_` vs `-`

protos/feast/core/Runner.proto
```diff
@@ -0,0 +1,73 @@
...
+//
+
+syntax = "proto3";
+package feast.core;
+
+option java_package = "feast.core";
+option java_outer_classname = "RunnerProto";
+option go_package = "github.com/gojek/feast/sdk/go/protos/feast/core";
+
+message DirectRunnerConfigOptions {
```
zhilingc, line 24: Is there a reason why runner configuration is defined as a proto? Not that its wrong -- it's just strange that it's the only piece of application configuration that is

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/RedisOnlineRetriever.java
```diff
@@ -45,6 +49,13 @@ public RedisOnlineRetriever(StatefulRedisConnection<byte[], byte[]> connection)
     this.syncCommands = connection.sync();
   }
 
+  public RedisOnlineRetriever(RedisConfig config) {
```
zhilingc: Should this follow the BQ pattern of `RedisOnlineRetriever.fromConfig(config)`?

core/src/main/java/feast/core/config/FeastProperties.java
```diff
@@ -16,53 +16,217 @@
...
   public static class JobProperties {
 
-    private String runner;
-    private Map<String, String> options;
+    @NotBlank
+    /* The active Apache Beam runner name. This name references one instance of the Runner class */
+    private String activeRunner;
+
+    /** List of configured job runners. */
+    private List<Runner> runners = new ArrayList<>();
```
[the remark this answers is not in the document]
woop: Forward compatibility.

core/src/main/java/feast/core/config/FeastProperties.java
```diff
@@ -16,53 +16,217 @@
...
+      @Positive private int partitions = 1;
+    }
   }
 
+  /**
+   * Validates whether stream options are correct.
+   *
+   * @return Boolean used for assertion
+   */
+  @AssertTrue
```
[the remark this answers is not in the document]
woop: That's nice. Will see if it works.

serving/src/main/resources/application.yml
```diff
@@ -1,12 +1,51 @@
...
   # Feast Serving requires connection to Feast Core to retrieve and reload Feast metadata (e.g. FeatureSpecs, Store information)
   core-host: ${FEAST_CORE_HOST:localhost}
   core-grpc-port: ${FEAST_CORE_GRPC_PORT:6565}
 
+  # Indicates the active store. Only a single store in the last can be active at one time. In the future this key
+  # will be deprecated in order to allow multiple stores to be served from a single serving instance
+  active_store: online
+
+  # List of store configurations
+  stores:
```
khorshuheng, line 12: As there are several changes in configurations, should this still be ported into v0.4 branch? On the other hand, if we don't port this, any PR that follows after this PR is merged will have a much higher chance of conflicting with the v0.4 branch.

protos/feast/core/Runner.proto
woop: So my rough thoughts on this are:
- If the configuration is defined clearly in FeastProperties then we probably want to make that the source of truth. 
- If the configuration in FeastProperties contains a generic container like `Map<String, String>` then we need to somehow document that. For validation I want to delegate that to the implementation layer like to the stores themselves. So the user would go to the stores documentation.
- For runners we are also using a `Map<String, String>` and we have a validation class, but its not really easy to find. Ideally we want documentation to point to.

So the easiest way to document this is to define it through a proto, because the proto will generate documentation that we can point users to. We are already doing this for `Source` and `Stores`, so I think adding runners is natural for now. The options will show up here: https://api.docs.feast.dev/grpc/feast.core.pb.html

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/RedisOnlineRetriever.java
woop: Yip, I think you might have been reviewing the commit prior to this one. The change has been made.

serving/src/main/resources/application.yml
woop: There are loads of breaking changes. If we backport this then we will have a lot of problems. 

> any PR that follows after this PR is merged will have a much higher chance of conflicting with the v0.4 branch

That is true, but that is why we maintain a separate branch for 0.4. We should ideally move on from 0.4 because the amount of changes is high and its not worth the upkeep.

khorshuheng: I agree that we should move on from 0.4. But in that case, do we still want to release 0.4.8? And if we do, what should be the last commit that we should stop at? Storage API refactoring?

woop: The point is that we dont backport anything significant. Only critical functionality like hotfixes. Otherwise there is no point in having releases if we just backport everything.

core/src/main/java/feast/core/config/FeastProperties.java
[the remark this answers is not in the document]
woop: Added: https://github.com/gojek/feast/pull/611/commits/6015ec7de3716c8c565d67ec9db46e2f0d9c34b8

protos/feast/core/Runner.proto
zhilingc: ok

PR #438, opened 2020-01-20
Update protos with Tensorflow data validation schema

davidheryanto:
**What this PR does / why we need it**:
This pull request is regarding feature validation in Feast. It updates `EntitySpec` and `FeatureSpec` protos with fields to specify constaints for the values. The constraint specs are adapted from Tensorflow metadata schema: https://github.com/tensorflow/metadata/blob/3b9d4114c6856ce18b7819023acb94f07183d348/tensorflow_metadata/proto/v0/schema.proto.  

Only `presence_constraints`, `shape_type` and `domain_info` are currently added to `EntitySpec` and `FeatureSpec` because it is adequate for the current validation requirement in Feast and to ensure integration with Tensorflow data validation works correctly, before adding more features supported in Tensorflow validation.

**Which issue(s) this PR fixes**:
Partially addresses  #172 

**Does this PR introduce a user-facing change?**:
No, user-facing change with updates in Feast SDK will follow in subsequent pull requests.

Issue #405, opened 2020-01-04
Entity types as a higher-level concept


woop:
## Introduction
Currently an entity, or more formally an entity type, is treated as a special type of field within a feature set. There has been an attempt to simplify the creation and management of entities and to keep them consistent with features, however some challenges exist with our current approach.

Note: The terms entity and entity type will be used interchangeable in the following issue.

## How are entities created?
- Users define an entity as part of a feature set. An entity in this case is a field like any other within the feature set. More than one entity can exist within a feature set.
- An entity's name must be unique within a feature set.
- There are no constraints on entities outside of a feature set, either at the project or global level. This means that multiple feature sets can define the same entities again.

## How are entities used?
- Retrieving feature values: Entities are used as a key for retrieving features. In order to retrieve feature values within a feature set, all entities must be provided as part of the lookup.
- Joining feature sets: In the event that feature values are being retrieved from multiple feature sets, entities are used to look up these feature values. Entities are also used to join across these feature sets to construct a single result set.

## What is the problem?
1. Discovery: It seems intuitive that users would start their discovery experience from the point of view of an entity type, since their business problem is generally framed around one or more entities. By nesting entities within feature sets and within projects and not providing a discovery means, it makes discovery harder. 
2. Consistency: Entities are typically consistent across all projects and systems in most organizations. This consistency is not enforced in Feast at the moment. Users are bound to redefine entities in their local projects if no consistency is enforced at an organizational level. Failure would occur when lookups happen or when joins happen across feature sets, especially when joins need to happen across projects. 
3. Key building: If entities and features maintain mutual compatibility in terms of supported data types, then support must be maintained for building keys from all feature value types. This adds a lot of complexity to key building since support must be maintained to serialize complex composite data structures in order to build these keys. 

### 1. Project-level entities
Functionality
- Entities are created outside of feature sets, but they still reside in a specific project namespace. 
- Entities have their own distinct API and supported data types (which may be more limited than features)
- Entities must be unique within a project namespace, but can be duplicated across an organization. Uniqueness is ensured through a full entity reference (`gojek/customer`).
- Entities are still defined as part of a feature set, but this is a selection process instead of creation.

Advantages
- Entities receive all the sharing and isolation benefits of "projects". Entities would not have to be treated separately from a logical and/or development standpoint. There would also be no explosion of a global entity namespace
- Users are free to experiment and develop within their projects without affecting other users, since duplication is allowed across projects.
- No need for a central team to gate-keep the creation of entities.

[...]

PR #612, opened 2020-04-13
Add computation and retrieval of batch feature statistics

zhilingc:
This PR adds support for retrieval of batch statistics over data ingested into a feast warehouse store, as proposed in M2 of the [Feature Validation RFC](https://docs.google.com/document/d/1TPmd7r4mniL9Y-V_glZaWNo5LMXLshEAUpYsohojZ-8/edit?pli=1#). 

Note that it deviates from the RFC in the following ways:
- Statistics are computed using SQL. This is because TFDV is unfortunately, only available in python, and Multi-SDK connectors for Beam is still a [work in progress](https://beam.apache.org/roadmap/connectors-multi-sdk/). Computing the statistics using SQL will be the compromise until either TFDV is available in Java, or cross-language execution is supported.
- Statistics can only be computed over a single feature-set at a time. This is mostly to reduce complexity in implementation. Since datasets are unable to span multiple feature sets, it makes sense to have this restriction in place. 

This is a bit of a chonky PR, and the code itself requires a bit of cleaning up, hence the WIP status, but refer to the attached [notebook](https://github.com/gojek/feast/blob/16867806944a9f8634c8503a118db6b1e9ed1ed9/examples/statistics/Historical%20Feature%20Statistics%20with%20Feast%2C%20TFDV%20and%20Facets.ipynb) for how this implementation looks like for a user of Feast.

```release-note
- Adds GetFeatureStatistics to the CoreService
- Adds get_statistics method to the client in the python SDK
```


core/src/main/java/feast/core/model/FieldId.java
```diff
@@ -0,0 +1,68 @@
...
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+@Embeddable
+@NoArgsConstructor
+@AllArgsConstructor
+@Getter
+@Setter
+public class FieldId implements Serializable {
```
woop, line 33: Perhaps FieldRef or FeatureRef?

woop:
Will try to review it tomorrow morning @zhilingc

core/src/main/java/feast/core/grpc/CoreServiceImpl.java
```diff
@@ -113,6 +95,21 @@ public void listFeatureSets(
...
+      StreamObserver<GetFeatureStatisticsResponse> responseObserver) {
+    try {
+      GetFeatureStatisticsResponse response = statsService.getFeatureStatistics(request);
+      responseObserver.onNext(response);
+      responseObserver.onCompleted();
+    } catch (RetrievalException | IllegalArgumentException | IOException e) {
+      log.error("Exception has occurred in GetFeatureStatistics method: ", e);
+      responseObserver.onError(
+          Status.INTERNAL.withDescription(e.getMessage()).withCause(e).asRuntimeException());
+    }
```
woop, line 110: Would it make sense to also return the message to the end user of other exceptions? It's obviously not a good practice long term, but for the time being it would make debugging easier.

core/src/main/java/feast/core/model/Entity.java
```diff
@@ -0,0 +1,127 @@
...
+
+/** Feast entity object. Contains name, type as well as domain metadata about the entity. */
+@Getter
+@Setter
+@javax.persistence.Entity
+public class Entity extends Field {
+
+  public Entity() {}
+
+  public Entity(FieldId fieldId) {
```
woop, line 33: We should possibly make these static factory methods like `fromProto`. It would make the API a bit cleaner.

sdk/python/feast/client.py
```diff
@@ -851,6 +858,78 @@ def ingest(
...
+                )
+            if end_date is not None:
+                request.end_date.CopyFrom(
+                    Timestamp(seconds=int(end_date.strftime("%s")))
+                )
+
+        return self._core_service_stub.GetFeatureStatistics(
+            request
+        ).dataset_feature_statistics_list
         return None
```
woop, line 933: Should this `return None` be deleted?

sdk/python/feast/client.py
```diff
@@ -908,6 +987,20 @@ def _build_feature_references(
     return features
 
 
+def _generate_dataset_id(feature_set: FeatureSet) -> str:
```
woop, line 990: The term dataset still seems a bit off to me.

The `id` that we are returning is a `job_id` or an `ingestion_id` since it is unique to every invocatio of `ingest()`. If we refer to it as `dataset_id` then I think users would expect it to be associated with their dataset, which means they provide the value or at the very least the value doesnt change in subsequent runs.

sdk/python/feast/client.py
zhilingc: oops. might have been from rebasing

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/statistics/FeatureSetStatisticsQueryInfo.java
```diff
@@ -0,0 +1,101 @@
...
+ */
+public class FeatureSetStatisticsQueryInfo {
+  private final String project;
+  private final String name;
+  private final int version;
+  private String datasetId = "";
+  private String date = "";
+  private final List<FieldStatisticsQueryInfo> features;
+
+  public FeatureSetStatisticsQueryInfo(
```
woop, line 41: Is this constructor being used?

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/statistics/FieldStatisticsQueryInfo.java
```diff
@@ -0,0 +1,94 @@
...
+package feast.storage.connectors.bigquery.statistics;
+
+import feast.core.FeatureSetProto.EntitySpec;
+import feast.core.FeatureSetProto.FeatureSpec;
+import feast.types.ValueProto.ValueType.Enum;
+
+/**
+ * Value class for Features containing information necessary to template stats-retrieving queries.
+ */
+public class FieldStatisticsQueryInfo {
```
woop, line 26: Just my own curiousity. How are we deciding when to use AutoValue and when to not use AutoValue?

In this case we have a value class which seems like it fits the AutoValue use case well, but we opt not to use it. Then for complex classes like `BigQueryStatisticsRetriever` we do opt for AutoValue.

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/statistics/FieldStatisticsQueryInfo.java
```diff
@@ -0,0 +1,94 @@
...
+      case BYTES_LIST:
+      case BOOL_LIST:
+      case FLOAT_LIST:
+      case INT32_LIST:
+      case INT64_LIST:
+      case DOUBLE_LIST:
+      case STRING_LIST:
+        return new FieldStatisticsQueryInfo(entitySpec.getName(), "LIST");
+      default:
+        throw new IllegalArgumentException("Invalid entity type provided");
```
woop, line 83: Can we return the `valueType` as a string?

tests/e2e/feature-validation.py
```diff
@@ -0,0 +1,99 @@
...
+            "datetime": [time_offset] * N_ROWS,
+            "entity_id": [i for i in range(N_ROWS)],
+            "strings": ["a", "b"] * int(N_ROWS / 2),
+            "ints": [int(i) for i in range(N_ROWS)],
+            "floats": [10.5 - i for i in range(N_ROWS)],
+        }
+    )
+
+    expected_stats = statistics_pb2.DatasetFeatureStatisticsList()
+    with open("statistics/expected_output_basic_dataset.json", "r") as fo:
```
woop, line 69: I think it would be less fragile if we generated the statistics from the dataframe using TFDV instead of reading it from a file. That would make it a lot easier to ensure that we have and can maintain compatibility.

tests/e2e/feature-validation.py
```diff
@@ -0,0 +1,99 @@
...
+    )
+
+    assert stats == dataset_basic["stats"]
+
+
+def test_basic_by_date(client, dataset_basic):
+    stats = client.get_statistics(
+        feature_refs=["strings", "ints", "floats"],
+        store="bigquery",
+        start_date=dataset_basic["date"],
```
woop, line 96: This just tests a single day right?

woop:
I made a first run through.

The only big thing that stands out is the duplication between entity level statistics and feature level statistics. I'll think about how we can simplify this implementation because as written it is a lot of scary to merge this in.

Other than that I am finding it a bit slow going to review the PR because some of the methods require comments. I'd rather not ding the methods and ask for JavaDocs or comments, so would you mind just giving the PR a scan through and adding them? It would be especially useful for `StatsService.java` as well as the model fields.

tests/e2e/feature-validation.py
zhilingc: We don't support the full set of TFDV stats, so generating it on the fly wouldn't allow the test to pass.

tests/e2e/feature-validation.py
zhilingc: yes

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/statistics/FieldStatisticsQueryInfo.java
zhilingc: I personally use autovalue when there are classes that require more than 2 elements to construct, but pebble is unfortunately incompatible with autovalue classes, so in this case I can't.

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/statistics/FeatureSetStatisticsQueryInfo.java
zhilingc: Nope. Removed.

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/statistics/FieldStatisticsQueryInfo.java
zhilingc: Sorry, what do you mean?

woop: I mean can we add the case value into the exception so that it's easier to debug.

tests/e2e/feature-validation.py
woop: Possibly to filter out the ones we don't need?

sdk/python/feast/client.py
zhilingc: I was wary of letting users set this value since there is a chance they would assign the same value twice. And running the ingestion twice SHOULD have distinct ids because users probably don't want to treat the data as belonging to the same set... I can add the option to override the dataset id, but it really opens up the possibility of users breaking the system.

I don't think it should be a `job_id`, which is already an extremely overloaded term, and retrieval of data by `ingestion_id` doesn't seem as intuitive as something like a `dataset_id`

woop: The reason I brought this up is because the current implementation does not fit the typical 
SOP for our users where they build idempotent data system. They expect to be able to create a dataset and ingest it based on a name, and have the data system upsert. So basically they just run ingestion until it succeeds and then they are done. They don't care, and often don't know, that we are running Kafka.

And our users definitely don't want to maintain a list of UUIDs. They probably just want to query their datasets by the names they are using like "2020417-mydataset`

Regardless, I think your current approach is fine for now because nothing prevents us from eventually allowing them to set this Id if we allow idempotency. I would probably reuse this same `dataset_id` though.

zhilingc: ok. I've added the option for people to supply their own dataset id, and fixed the sql so that it only retrieves the latest value for each dataset/entity/event timestamp. This should ensure idempotency.

woop:
I'll talk about the data model first because that is the most important part to get right now.

**Assumptions**
- Entities are primarily used as keys
- Features are primarily used as values in the context of models. Our users care more about statistics on features than on entities.
- Features need references (like feature references).
- Entities need references, but these references are different from features.
- Entities will also be treated differently in the future (See #405). Constraints on features and constraints on entities are different. Entities will be reused and features must be unique by name.
- Nothing prevents a user from ingesting entity data as a feature. 

**Thoughts**
- Although there is a use case for having entity level statistics, I don't think it warrants the amount of code we'd need to maintain here.
- There are pre-existing problems with the data model that should be solved now before 0.5 is released. Normally that would happen in a separate PR, but we can also reuse this PR if that is easier for you. 
- One of the problems that I see is the fact that `Field` and its inheritence adds complexity without any real value.
- The second is that there is a coupling between the references of features and entities that I don't think should exist.
- Lastly, and this is definitely out of scope for your work, is that the source table is hardcoded to Kafka (thanks for the reminder @mrzzy . Will add a separate issue for that.)

**Recommendation**
- Remove `Field` and its inheritance
- Keep `FieldId` but rename it to `FeatureReference`
- Have `Feature` reference `FeatureReference` as its primary key
- Create an `EntityReference` (for the time being this can be the same as `FeatureReference`)
- Keep `FeatureStatistics` and have it reference `FeatureReference`
- Remove `EntityStatistics` completely as well as statistics methods on Entity.

cc @ches @Yanson

zhilingc:
@woop 

> Features are primarily used as values in the context of models. Our users care more about statistics on features than on entities.
> Although there is a use case for having entity level statistics, I don't think it warrants the amount of code we'd need to maintain here.

I think that’s a very strong assumption to make. 

I’ve been back and forth about entity statistics the past few weeks; my first iteration only involved feature statistics, but i ended up adding support for entity statistics later because i think it is something that is truly valuable to users, for checking id distributions and bounds, etc. I don’t think it’s 

To your point:

> Nothing prevents a user from ingesting entity data as a feature.

This doesn’t make sense when you consider the user experience. Do they double-up on entity columns so that they can be ingested as features JUST to retrieve statistics? Do they ingest it as a separate dataset? Why do users have to perform hacks to transform a logical unit (the dataset) into something else just to do something that is expectedly a basic requirement?

> One of the problems that I see is the fact that Field and its inheritance adds complexity without any real value.

I’m not sure what sort of complexity you mean here. I think if we do have statistics for entities it’s a lot easier to maintain compatibility with TFDV fields with inheritance, which is the main reason why the class was introduced in the first place. Additional fields that differ between features and entities can be defined in their concrete classes.

> The second is that there is a coupling between the references of features and entities that I don't think should exist.

Do you mean the fact that they use the same object? I guess it’s fine to separate them for the sake of forward compatibility…

> Have Feature reference FeatureReference as its primary key
Do you mean have FeatureReference as a distinct entity and have feature sets, features and statistics reference that entity? I’m not sure if that’s the best idea.

> Keep FeatureStatistics and have it reference FeatureReference
FeatureReference is part of the definition of Feature and FeatureStatistics cannot form a relation directly to it. To be exact, with the current implementation, the table definition for Features looks like:

```
-- Table Definition ----------------------------------------------

CREATE TABLE feature (
    feature_set character varying(255),
    name character varying(255),
    project character varying(255),
    version integer,
    bool_domain bytea,
    domain character varying(255),
    float_domain bytea,
    group_presence bytea,
    image_domain bytea,
    int_domain bytea,
    mid_domain bytea,
    natural_language_domain bytea,
    presence bytea,
    shape bytea,
    string_domain bytea,
    struct_domain bytea,
    time_domain bytea,
    time_of_day_domain bytea,
    type character varying(255),
    url_domain bytea,
    value_count bytea,
    CONSTRAINT feature_pkey PRIMARY KEY (feature_set, name, project, version)
);
```

And FeatureStatistics has the following constraint:
```
    CONSTRAINT fksrjnxx9em2xg2cqgd6ifm4lll FOREIGN KEY (feature_set, name, project, version) REFERENCES feature(feature_set, name, project, version)
```

zhilingc:
Ok, I polled a few data scientists and they acknowledged that there is no need to have entity statistics except for occasional once-off checks, which can be done directly to the db anyway. I'll remove the entity statistics - but does that mean i remove the entity field schemas introduced in #438 as well?

woop:
> I’ve been back and forth about entity statistics the past few weeks; my first iteration only involved feature statistics, but i ended up adding support for entity statistics later because i think it is something that is truly valuable to users, for checking id distributions and bounds, etc. I don’t think it’s

I think your sentence cut off there.

I agree that entity statistics are very useful, don't get me wrong, which is why we wanted them in the first place.

What doesn't line up for me is this implementation. As it stands, Feast does not support entity statistics. If I am forced to choose between having only feature statistics and merging in all of this code then I will definitely chose the former.

However, the idea case is that we can **cleanly** support entity statistics.

> This doesn’t make sense when you consider the user experience. Do they double-up on entity columns so that they can be ingested as features JUST to retrieve statistics? Do they ingest it as a separate dataset? Why do users have to perform hacks to transform a logical unit (the dataset) into something else just to do something that is expectedly a basic requirement?

My point here is that we can add feature statistics safely and support it properly without immediately supporting entity statistics. If users really want a work-around to get an idea of entity statistics then they can ingest them as features, which means adding additional columns. That would not be something we would officially support, and they don't lose anything over what they have today.

Duplicating this code for both entities and features is after all tech debt. Another Feast engineer will have to clean this up later. I would rather solve this properly or scope it down.

> I’m not sure what sort of complexity you mean here. I think if we do have statistics for entities it’s a lot easier to maintain compatibility with TFDV fields with inheritance, which is the main reason why the class was introduced in the first place. Additional fields that differ between features and entities can be defined in their concrete classes.

I stand corrected.

> Do you mean the fact that they use the same object? I guess it’s fine to separate them for the sake of forward compatibility…

Yes. Their constraints are distinct.

> Do you mean have FeatureReference as a distinct entity and have feature sets, features and statistics reference that entity? I’m not sure if that’s the best idea.

Exactly how it is right now for FieldId, but it's just FeatureReference. Entity doesn't reference FeatureReference. It doesn't have to be a distinct entity, it can be an embeddedId.

> FeatureReference is part of the definition of Feature and FeatureStatistics cannot form a relation directly to it. To be exact, with the current implementation, the table definition for Features looks like:

The table definition you are showing me is exactly what I want. It's a foreign key based on multiple foreign columns on the feature reference (name, feature_set, project, version). I guess it just comes down to a naming difference (`FieldId` -> `FeatureReference`). 

--------------

It's very clear that our existing data model is wrong because it forces your statistics design to be split between features and entities. Statistics are ultimately about data, and the only common concepts in our data layer is the `Field`.

image

So our options are

**1. Map statistics to Fields instead of Entities and Features.**
This does make sense at the level of feature statistics. Features map to data through fields. Entities map to data through fields. Statistics occur on the data/field.

The problem with our current `Field` is that the parent child relationship of fields to features and entities doesn't work. `Field` as a means of sharing common Java fields through inheritance makes sense, but `Fields` shouldn't be the parents of `Entities` and `Features` because `Entities` don't share an identity with a `Field`. Each `Field` in a `FeatureRow` is unique, and that `Field`, even if its an `Entity`, has a different identity for each `Feature Set`. 

One way to make this work is invert the relationship between Feature/Entity and Field to one that is consistent with [FeatureRow.proto](https://github.com/gojek/feast/blob/master/protos/feast/types/FeatureRow.proto)

* `Fields` have an identity (`FieldRef`) and a table.
* `Feature set` has a single collection of `Fields` (instead of `Features` and `Entities`). This creates a many to many mapping between `Feature Set` and `Field`
* A `Field` can then be configured to be one of two types, `Entity` or a `Feature`. 
* `Fields` that have a `Feature` type don't have an external identity. They share an identity with the `Field`, `FieldRef`. There may or may not be a `Feature` table that references this `Field` table. Not sure if it's necessary.
* `Fields` that have an `Entity` type reference an external `Entity`. Meaning an `Entity` has a table. These entities can occur in multiple `Feature Sets`. This might be out of scope for your PR though.
* Statistics is successfully de-duplicated and only reference a `Field`

**2. Replace Field with Feature** 
The thing with Proposal (1) is that the identity of `Field` will probably be `FieldId`/`FieldRef` which is 100% identical to `FeatureRef`. That raises the question of what differentiates these two (Field/Feature). Do we really need `FieldId` and `FeatureRef`? 

If we want to treat all `Fields` as `Features` (generating statistics or feeding them into models), then we could simplify the data model by replacing `Field` with `Feature` in proposal 1. That removes the duplication between `FieldId` and `FeatureRef`. 

**3. Just remove Field** 
The final proposal is the one I gave you previously. We remove `Field` and keep `Entities` and `Features` separate and allow statistics only on `Features`. This is the smallest cut that we can make right now, since the data model doesn't dramatically change.

Ok, so there you have three proposals and my logic behind it. Please come up with a solution

zhilingc:
The issue with (1) is that features were split to be distinct from entities because while they're currently mostly congruent to each other, that won't be the case in the future. 
I'll do as you suggested in (3), but does that mean i remove the entity field schemas introduced in #438 as well?

woop:
> The issue with (1) is that features were split to be distinct from entities because while they're currently mostly congruent to each other, that won't be the case in the future.

Ok, how does (1) prevent that? All of the options keep entities and features disconnected.

> I'll do as you suggested in (3), but does that mean i remove the entity field schemas introduced in #438 as well?

You would need to remove entity schemas yes. I am fine with going for (3), but I would still like to hear your thoughts on what I wrote above.

core/src/main/java/feast/core/model/Entity.java
```diff
@@ -19,94 +19,35 @@
...
   private Entity(String name, ValueType.Enum type) {
-    this.setId(new FieldId());
-    this.setName(name);
+    this.setReference(new EntityReference());
+    this.reference.setName(name);
     this.setType(type.toString());
   }
 
-  public static Entity withId(FieldId fieldId) {
+  public static Entity withRef(EntityReference entityId) {
```
woop, line 43: `entityRef` perhaps?

core/src/main/java/feast/core/model/Feature.java
```diff
@@ -32,19 +32,49 @@
...
+  private byte[] urlDomain;
+  private byte[] timeDomain;
+  private byte[] timeOfDayDomain;
 
   public Feature() {}
 
   private Feature(String name, ValueType.Enum type) {
-    this.setId(new FieldId());
-    this.setName(name);
+    this.setReference(new FeatureReference());
```
woop, line 70: Is there a reason to have a no args constructor for `FeatureReference`? If there isn't then I think it might be safer to require all the fields/args that make up the reference.

core/src/main/java/feast/core/model/Feature.java
```diff
@@ -32,19 +32,49 @@
...
+    this.reference.setName(name);
     this.setType(type.toString());
   }
 
-  public static Feature withId(FieldId fieldId) {
+  public static Feature withReference(FeatureReference featureId) {
```
woop, line 75: is `with` used in practice? How about `fromRef`?

core/src/main/java/feast/core/service/StatsService.java
```diff
@@ -0,0 +1,587 @@
...
+    }
+
+    // 3. Retrieve the statistics from the StatsRetriever.
+    List<List<FeatureNameStatistics>> featureNameStatisticsList = new ArrayList<>();
+    if (request.getDatasetIdsCount() == 0) {
+      // If no dataset provided, retrieve by date
+      long timestamp = request.getStartDate().getSeconds();
+      while (timestamp < request.getEndDate().getSeconds()) {
+        List<FeatureNameStatistics> featureNameStatistics =
+            getFeatureNameStatisticsByDate(
```
woop, line 125: Would this results in multiple calls to the Feast db and multiple calls to BigQuery? If so, can we use the full time range?

core/src/main/java/feast/core/model/Feature.java
zhilingc: `fromRef` doesn't convey the meaning though. It's called a `reference` but its more of a property of the class

core/src/main/java/feast/core/service/StatsService.java
zhilingc: No, if the full time range is used, the values cannot be cached.

core/src/main/java/feast/core/service/StatsService.java
```diff
@@ -0,0 +1,587 @@
...
+  private StatisticsRetriever getStatisticsRetriever(String storeName)
+      throws InvalidProtocolBufferException {
+    Store store = storeRepository.getOne(storeName).toProto();
+    if (store.getType() != StoreType.BIGQUERY) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Invalid store %s specified. Batch statistics are only supported for BigQuery stores",
+              store.getType()));
+    }
+    return BigQueryStatisticsRetriever.newBuilder()
```
woop, line 326: Can we please use a static factory method to make this consistent with the other retrievers?

`fromConfig(Map<String, String> config)`?

core/src/main/java/feast/core/model/FeatureStatistics.java
```diff
@@ -0,0 +1,269 @@
...
+  @ManyToOne
+  @JoinColumns({
+    @JoinColumn(name = "project", referencedColumnName = "project"),
+    @JoinColumn(name = "feature_set", referencedColumnName = "feature_set"),
+    @JoinColumn(name = "version", referencedColumnName = "version"),
+    @JoinColumn(name = "name", referencedColumnName = "name")
+  })
+  private Feature feature;
+
+  // Only one of these fields should be populated.
```
woop, line 57: Previously when we discussed this we said we would store statistics with a key of `day` + `dataset`, which would allow us to roll it up to any time range and to any dataset.

Is there a reason why both won't be set here? Are we still doing that?

zhilingc: a large portion of statistics are lost in aggregation, so we want to avoid aggregating within datasets, which should be the logical unit for retrieval here anyway.

core/src/main/java/feast/core/service/StatsService.java
zhilingc: done

core/src/main/java/feast/core/model/Feature.java
zhilingc: Using a constructor with just name now. The previous model refactor you did injects the project and feature set name independently of the initialisation of the Feature/Entity class. I don't think it's within scope to change that right now.

core/src/main/java/feast/core/model/FeatureStatistics.java
```diff
@@ -0,0 +1,269 @@
...
+  private byte[] topValues;
+
+  // Byte statistics
+  private float minBytes;
+  private float maxBytes;
+  private float avgBytes;
+
+  // Instantiates a Statistics object from a tensorflow metadata FeatureNameStatistics object and a
+  // dataset ID.
+  public static FeatureStatistics fromProto(
```
woop, line 96: I think we should perhaps change the method names in this class. `fromProto` and `toProto` are normally reserved for `domain model` conversion to and from our proto representation. We are passing non-proto methods here.

core/src/main/java/feast/core/model/FeatureStatistics.java
woop: Have we documented the list of statistics that are lost? How are we dealing with that in the case of a time range query?

core/src/main/java/feast/core/model/FeatureStatistics.java
```diff
@@ -0,0 +1,269 @@
...
+      case STRUCT:
+        StructStatistics structStats =
+            StructStatistics.newBuilder().setCommonStats(commonStatistics).build();
+        featureNameStatisticsBuilder.setStructStats(structStats);
+        break;
+    }
+    return featureNameStatisticsBuilder.build();
+  }
+
+  private static FeatureStatistics fromProto(FeatureNameStatistics featureNameStatistics)
```
woop, line 205: It's confusing that this overlaps with `fromProtos` above. It could have been a non-static method like `updateFromFeatureNameStatistics`

core/src/main/java/feast/core/model/FeatureStatistics.java
```diff
@@ -0,0 +1,269 @@
...
+    featureStatistics.setDate(date);
+    Feature feature =
+        Feature.withReference(
+            new FeatureReference(
+                project, featureSetName, version, featureNameStatistics.getPath().getStep(0)));
+    featureStatistics.setFeature(feature);
+    return featureStatistics;
+  }
+
+  public FeatureNameStatistics toProto() throws InvalidProtocolBufferException {
```
woop, line 132: Same as the other comments. I think it should be renamed to indicate that this is exporting a TensorFlow `FeatureNameStatistics` object.

core/src/main/java/feast/core/service/StatsService.java
woop: Correct me if I am wrong, but wouldn't the only requirement that the query we send to BigQuery returns multiple results? 

Meaning we check our cache (in one query), we ask BQ for the days that we don't have (in one query), we cache them, and then we return the full set.

core/src/main/java/feast/core/service/StatsService.java
```diff
@@ -0,0 +1,582 @@
...
+          String.format(
+              "Invalid store %s specified. Batch statistics are only supported for BigQuery stores",
+              store.getType()));
+    }
+    return BigQueryStatisticsRetriever.create(store.getBigqueryConfig());
+  }
+
+  private FeatureSetSpec getFeatureSetSpec(String featureSetId)
+      throws InvalidProtocolBufferException, IllegalArgumentException, RetrievalException {
+    String[] split = featureSetId.split("/");
```
woop, line 330: Is it possible for us to dedupe this feature ref conversion? Can we move it to `FeatureReference.java` so that other parts of the code base can just reuse that `toString`, `fromString` maybe?

core/src/main/java/feast/core/model/FeatureStatistics.java
zhilingc: `FeatureNameStatistics` is a proto.

woop:
Questions
* Do we have or can we add a multiple day (time) statistics aggregation test?
* Do we have a definitive list of statistics that we provide users?
* Do we have a list of statistics that they lose when they do aggregation?
* Are we testing that the results contain the above statistics?
* One of the biggest pains with our current bigquery storage implementation is that failures don't bubble up to the user. Can we also test that our exceptions bubble up? Especially if there is a problem with BigQuery configuration.
* Is there a test for the force refresh?
* Is there a test for the user providing no dataset/time range or providing both?

zhilingc:
> Do we have or can we add a multiple day (time) statistics aggregation test?

Added to the end to end tests.

> Do we have a definitive list of statistics that we provide users?
> Do we have a list of statistics that they lose when they do aggregation?

All in the gitbook documentation already. See the table at the bottom of the page.

> Are we testing that the results contain the above statistics?

Yes, in the end to end test.

> One of the biggest pains with our current bigquery storage implementation is that failures don't bubble up to the user. Can we also test that our exceptions bubble up? Especially if there is a problem with BigQuery configuration.

The grpc method catches all exceptions, so it will be bubbled up. I've tested this personally.

> Is there a test for the force refresh?

No, i've only tested this locally. I can add the unit test, but it will take time.

> Is there a test for the user providing no dataset/time range or providing both?

Added.

core/src/main/java/feast/core/service/StatsService.java
zhilingc: It's for the FeatureSet, not Feature. And doing a code search finds only a single usage of this in feast core, so I don't think refactoring it out is too important right now

protos/feast/core/CoreService.proto
```diff
@@ -274,3 +281,42 @@ message StopIngestionJobRequest {
...
 message StopIngestionJobResponse {}
+
+message GetFeatureStatisticsRequest {
+    // Feature set to retrieve the statistics for. A fully qualified feature set
+    // id in the format of project/feature_set:version must be provided.
+    string feature_set_id = 1;
+
+    // Optional filter which filters returned statistics by selected features. These
+    // features must be present in the data that is being processed.
+    repeated string feature_ids = 2;
```
woop, line 292: Feature references?

zhilingc: feature names actually

woop: Alright, so we change it to feature_names then?

woop: I guess it doesn't matter much since we can always change a field name

zhilingc: I think since we're providing `feature_set_id` `features` should logically refer to feature names within that feature set.

ches:
Would it be feasible to break this pull request up into a few somehow? At a first glance there's a lot that I think is generally beneficial independent of the feature addition of batch statistics, like entity/domain model additions and updates like `Entity`, feature and entity references, etc.

I have little doubt that the batch statistics feature addition in itself will necessarily be a large PR no matter how we slice it, but I see the 5,000 line diffstat here and my spirits get deflated that I will need to reserve the majority of a weekend to it sometime in the next few weeks, by which point it will be merged (which is totally fine, just that it feels like a drag to bring people back to it later if there are questions & comments).

ches:
Some products alternate technical and feature releases, starting to feel Feast 0.5 could be split on its two roadmap categories that way… speaking from the perspective of someone who will upgrade a deployment at some point (i.e. merging down a branch), [the scope is getting pretty huge](https://github.com/gojek/feast/compare/v0.4.7...master) (and that's not including this, #554, or #533 yet). It's great work, it's just creating some anxiety for me about the effort and risks of upgrading.

Anyway, I'm off topic for this PR, I'll take this elsewhere…

zhilingc:
@ches That's fair. I'll do what I can.

woop:
> Some products alternate technical and feature releases, starting to feel Feast 0.5 could be split on its two roadmap categories that way

Yea the scope of 0.5 has increased a lot, which is partly good and bad. By splitting do you mean based on the existing master branch versus the additional functionality we are introducing with open PRs, or would you move out some functionality also merged into master?

My intuition is to put our efforts towards making 0.5 as stable and well tested as possible, so we would probably deploy it side by side and migrate folks over slowly. I would want to include all breaking changes for that matter. I think its possible to have 0.5 and 0.6 be technical and feature releases, but we probably will roll up to 0.6 quite quickly, which means we won't have much time to maintain the 0.5 branch.

Issue #562, opened 2020-03-23
Remove GCP dependency in test_get_batch_features


woop:
## Expected Behavior 

Python unit tests should not depend on GCP service accounts being available

## Current Behavior

Currently the test `test_get_batch_features` in the Python SDK leads to a failure when it is run from an environment where GOOGLE_APPLICATION_CREDENTIALS are not available, or it cannot infer a local Gcloud configuration. This is triggered because the `Job` object initializes itself with a storage client that requires GCP credentials.

For the time being this test has been commented out, please see this PR https://github.com/gojek/feast/pull/560/files#r396175767

## Steps to reproduce

1. Uncomment above test
2. Execute from a clean environment (kf-feast/feast-ci:latest docker image)

## Possible Solution

- Stub out the storage client in the unit test.
- Alternatively we should make the Job object not initialize a Storage Client.

PR #613, opened 2020-04-14
Remove test that requires GCP credentials

zhilingc:
Backporting removal of a test that requires gcp access. #562


khorshuheng:
Going to merge this despite the test failure as this is fixed on https://github.com/gojek/feast/pull/610/files. At the same time, this PR is needed for 610 to pass.

PR #627, opened 2020-04-15
Correct links to why-feast and concepts doc site

anderseriksson:


woop:
Thanks @anderseriksson

PR #628, opened 2020-04-16
Add ingestion tests to validate field handling

junhui096:
This PR refactors the initial test for feast ingestions and adds additional unit tests.The following tests are added:

1. Verifies that the ingestion job ignores unknown fields in the incoming feature row and still writes the row to the feature store.
2. Verifies that in the event of missing fields in the feature rows, the remaing fields are still written to the store, with the missing fields set to null.
3. Verifies that if an entity field is missing from a feature row, the entire row is not added to the store.

General tests.

```release-note
NONE
```


woop:
@junhui096 Please make sure to have atomic and descriptive commits.

image

These are not descriptive enough for us to walk through

ingestion/src/test/java/feast/ingestion/ImportJobTest.java
```diff
@@ -180,91 +173,320 @@ public void runPipeline_ShouldWriteToRedisCorrectlyGivenValidSpecAndFeatureRow()
...
+  public void runPipeline_ShouldWriteFeatureRowsWithMissingEntityFieldsToDeadLetterQueue()
+      throws IOException, InterruptedException {
+
+    List<FeatureRow> input = new ArrayList<>();
+
+    LOGGER.info("Generating test data ...");
+    IntStream.range(0, IMPORT_JOB_SAMPLE_FEATURE_ROW_SIZE - FAILED_JOB_SAMPLE_FEATURE_ROW_SIZE)
+        .forEach(
+            i -> {
+              FeatureRow randomRow = TestUtil.createRandomFeatureRow(spec);
```
woop, line 349: I'd like you to remove the non-determinism from your tests please. I know that this was already in use by whoever wrote these tests previously, but I don't want us to use any randomness. Ideally you would hardcode the tests somewhere here, while still abstracting the boilerplate as much as possible.

ingestion/src/test/java/feast/ingestion/ImportJobTest.java
```diff
@@ -180,91 +173,320 @@ public void runPipeline_ShouldWriteToRedisCorrectlyGivenValidSpecAndFeatureRow()
...
+          FeatureRow actualValue = getFeatureRowFromByte(key, actualByteValue);
+
+          // Ensure the retrieved FeatureRow is equal to the ingested FeatureRow.
+          Assert.assertEquals(expectedValue, actualValue);
+        });
+    redisClient.shutdown();
+  }
+
+  @Test
+  public void runPipeline_ShouldWriteFeatureRowsWithMissingEntityFieldsToDeadLetterQueue()
```
woop, line 340: Would you mind adding a few comments throughout this test to make it clearer what you are doing? It's a little bit cryptic as is.

woop: Same for your other test.

ingestion/src/test/java/feast/ingestion/ImportJobTest.java
```diff
@@ -180,91 +185,332 @@ public void runPipeline_ShouldWriteToRedisCorrectlyGivenValidSpecAndFeatureRow()
...
+      Assert.fail(
+          String.format(
+              "Actual Redis value cannot be parsed as FeatureRow, key: %s, value :%s",
+              key, new String(actualByteValue, StandardCharsets.UTF_8)));
+    }
+    return actualValue;
+  }
+
+  // Convert feature rows to a list of values for each non-entity feature.
+  private FeatureRow getFieldsFromFeatureRow(FeatureSetSpec spec, FeatureRow row) {
```
davidheryanto, line 463: Can we rename and update comments for this method so the name and comments are more representative of what this method is doing?

e.g.
- name is `getFields...` but it actually returns `FeatureRow`
- `spec` is passed in as parameter, but it's not clear without looking at the implementation details how this spec argument is used

ingestion/src/test/java/feast/ingestion/ImportJobTest.java
```diff
@@ -180,91 +185,332 @@ public void runPipeline_ShouldWriteToRedisCorrectlyGivenValidSpecAndFeatureRow()
...
+          FeatureRow actualValue = getFeatureRowFromByte(key, actualByteValue);
+
           // Ensure the retrieved FeatureRow is equal to the ingested FeatureRow.
           Assert.assertEquals(expectedValue, actualValue);
         });
     redisClient.shutdown();
   }
+
+  // get the expected rows from the input, and default values.
+  private Map<RedisKey, FeatureRow> getExpectedRows(List<FeatureRow> input, FeatureSetSpec spec) {
```
davidheryanto, line 414: I think it's good to briefly mention in comments how the `spec` argument is used to get the expected rows.
Also the method comment mention `default` values but this `default` values is nowhere in this method.

ingestion/src/test/java/feast/ingestion/ImportJobTest.java
```diff
@@ -180,91 +185,332 @@ public void runPipeline_ShouldWriteToRedisCorrectlyGivenValidSpecAndFeatureRow()
...
     List<FeatureRow> input = new ArrayList<>();
-    Map<RedisKey, FeatureRow> expected = new HashMap<>();
 
     LOGGER.info("Generating test data ...");
     IntStream.range(0, IMPORT_JOB_SAMPLE_FEATURE_ROW_SIZE)
         .forEach(
             i -> {
-              FeatureRow randomRow = TestUtil.createRandomFeatureRow(featureSet.getSpec());
-              RedisKey redisKey = TestUtil.createRedisKey(featureSet.getSpec(), randomRow);
+              FeatureRow randomRow = TestUtil.createFeatureRowWithDefaults(spec, defaultValues);
```
davidheryanto, line 219: Since this created FeatureRow is not random (the value is `default`) maybe the variable name should be updated.

ingestion/src/test/java/feast/ingestion/ImportJobTest.java
davidheryanto: Can we also update this thes method name?
Because it is NOT asserting that the invalid rows are written to a deadletter queue, it is only asserting that the rows with missing entity are not ingested.

Either that or I think we should update the test implemention to actually check that the ingestion job is writing to a deadletter queue.

ingestion/src/test/java/feast/ingestion/ImportJobTest.java
davidheryanto: Also for other similar variables named `randomRow`

woop:
@junhui096 please have a look at the code over [here](https://github.com/gojek/feast/pull/676#discussion_r419907248).

ingestion/src/test/java/feast/ingestion/ImportJobTest.java
```diff
@@ -180,91 +200,376 @@ public void runPipeline_ShouldWriteToRedisCorrectlyGivenValidSpecAndFeatureRow()
...
+  */
+
+  @Test
+  public void runPipeline_ShouldWriteToRedisCorrectlyGivenMissingFieldsInFeatureRow()
+      throws IOException, InterruptedException {
+
+    List<FeatureRow> input = new ArrayList<>();
+
+    LOGGER.info("Generating test data ...");
+    IntStream.range(0, IMPORT_JOB_SAMPLE_FEATURE_ROW_SIZE)
```
woop, line 326: Can we just use a single FeatureRow here that is constructed by hand? That would make it a lot easier to understand what is going on. I think we should be testing missing fields inbetween other fields. 

Imagine a good row
TS E1 E2 F1 F2 F3 F4

Take away one feature
TS E1 E2 F1 __ F3 F4

The row should be stored as is above, and should be readable

ingestion/src/test/java/feast/ingestion/TestFeatureRow.json
```diff
@@ -0,0 +1,27 @@
+{
```
zhilingc, line 1: These test files should not be here. If you do need to have test files (although i think this is unnecessary since feature rows could easily be declared in code) they should be placed in the `test/resources` folder.

ingestion/src/main/java/feast/ingestion/transform/ReadFromSource.java
```diff
@@ -73,27 +75,50 @@ public ReadFromSource build() {
...
+                    .commitOffsetsInFinalize())
+            .apply(
+                "KafkaRecordToFeatureRow",
+                ParDo.of(
+                        KafkaRecordToFeatureRowDoFn.newBuilder()
+                            .setSuccessTag(getSuccessTag())
+                            .setFailureTag(getFailureTag())
+                            .build())
+                    .withOutputTags(getSuccessTag(), TupleTagList.of(getFailureTag())));
+    PCollection<FeatureRow> validRows =
```
zhilingc, line 100: This not supposed to be in this PR, I think

ingestion/src/test/java/feast/ingestion/ImportJobTest.java
```diff
@@ -94,30 +106,26 @@
...
-    LocalRedis.stop();
-    LocalKafka.stop();
-  }
+  private static List<FeatureRow> defaultValues = new ArrayList<>();
 
-  @Test
-  public void runPipeline_ShouldWriteToRedisCorrectlyGivenValidSpecAndFeatureRow()
-      throws IOException, InterruptedException {
-    FeatureSetSpec spec =
+  private static FeatureSetSpec spec;
```
zhilingc, line 116: All these don't have to be/shouldn't be static

ingestion/src/test/java/feast/ingestion/ImportJobTest.java
```diff
@@ -180,91 +199,367 @@ public void runPipeline_ShouldWriteToRedisCorrectlyGivenValidSpecAndFeatureRow()
...
-                      .map(field -> field.toBuilder().clearName().build())
-                      .collect(Collectors.toList());
-              randomRow =
-                  randomRow
-                      .toBuilder()
-                      .clearFields()
-                      .addAllFields(fields)
-                      .clearFeatureSet()
+              FeatureRow testRow = defaultValues.get(i);
+              input.add(testRow);
```
zhilingc, line 233: just use `Arrays.copyOfRange` or `.sublist` here

ingestion/src/test/java/feast/ingestion/ImportJobTest.java
```diff
@@ -180,91 +199,367 @@ public void runPipeline_ShouldWriteToRedisCorrectlyGivenValidSpecAndFeatureRow()
...
+    these features are ignored in ingestion.
+  */
+
+  @Test
+  public void runPipeline_ShouldWriteToRedisCorrectlyGivenUnknownFeatureFields()
+      throws IOException, InterruptedException {
+
+    List<FeatureRow> input = new ArrayList<>();
+
+    IntStream.range(0, IMPORT_JOB_SAMPLE_FEATURE_ROW_SIZE)
```
zhilingc, line 272: again, `Arrays.copyOfRange().stream().map(...)` should make this much cleaner.

ingestion/src/test/java/feast/ingestion/ImportJobTest.java
```diff
@@ -180,91 +199,367 @@ public void runPipeline_ShouldWriteToRedisCorrectlyGivenValidSpecAndFeatureRow()
...
+   If there are missing fields in an incoming feature row, this test verifies that their values are set to null during ingestion.
+  */
+
+  @Test
+  public void runPipeline_ShouldWriteToRedisCorrectlyGivenMissingFieldsInFeatureRow()
+      throws IOException, InterruptedException {
+
+    List<FeatureRow> input = new ArrayList<>();
+    FeatureRow.Builder builder = FeatureRow.newBuilder();
+    FileReader reader = new FileReader("src/test/java/feast/ingestion/TestFeatureRow.json");
```
zhilingc, line 325: If this is a single row, it's better to just declare it in code.

ingestion/src/test/java/feast/ingestion/ImportJobTest.java
```diff
@@ -180,91 +199,367 @@ public void runPipeline_ShouldWriteToRedisCorrectlyGivenValidSpecAndFeatureRow()
...
         });
     redisClient.shutdown();
   }
+
+  /* This test verifies that the feature set reference is prepended with default in the ingestion pipeline
+     if the reference is not namespaced with its project.
+  */
+
+  @Test
+  public void runPipeline_ShouldPrependDefaultProjectToFeatureSetReference() {
```
zhilingc, line 417: Doesn't belong in this PR

junhui096:
> Would it be better to use `TestUtil.createRandomFeatureRow` to generate the rows rather than persisting a text file of rows?

I think the reason I replaced `TestUtil.createRandomFeatureRow` earlier, is because of an earlier comment about avoiding non-determinism in the tests.

woop:
Closing this PR since there is no movement.

PR #636, opened 2020-04-20
Kafka producer should raise an exception when it fails to connect to broker

### LINKED ISSUE #626: No exception when connecting to Kafka fails
Hey guys,

I have installed feast(0.4.4) via Helm3 on GKE 

The Basic example is working until the ingestion (online serving) part

There I get ->>

Waiting for feature set to be ready for ingestion...
  0%|          | 0/15 [00:00<?, ?rows/s]

but there is no progress.

When I look on GCPs bigquery interface I can see that the project "customer_project" is created with the correct columns in "customer_transactions".
But for sure no data

get_feature_set gives me 

{
  "spec": {
    "name": "customer_transactions",
    "version": 1,
    "entities": [
      {
        "name": "customer_id",
        "valueType": "INT64"
      }
    ],
    "features": [
      {
        "name": "daily_transactions",
        "valueType": "DOUBLE"
      },
      {
        "name": "total_transactions",
        "valueType": "INT64"
      }
    ],
    "maxAge": "432000s",
    "source": {
      "type": "KAFKA",
      "kafkaSourceConfig": {
        "bootstrapServers": "feast-kafka:9092",
        "topic": "feast"
      }
    },
    "project": "customer_project_1"
  },
  "meta": {
    "createdTimestamp": "2020-04-15T10:26:51Z",
    "status": "STATUS_READY"
  }
}

I had to modify some port service setup in the chart so it can be that some of feast have connection issues between kafka etc.
But There are no errors in the logs of the core and serving pod.

What can be the problem and how is a way to debug that?

junhui096:
In an ingestion job from the Python SDK, the producer should throw an exception if it fails to push data to the Kafka broker.

Fixes #626

```release-note
NONE
```


sdk/python/feast/loaders/abstract_producer.py
```diff
@@ -129,7 +129,12 @@ def flush(self, timeout: Optional[int]):
         Returns:
             int: Number of messages still in queue.
         """
-        return self.producer.flush(timeout=timeout)
+        messages = self.producer.flush(timeout=timeout)
```
zhilingc, line 132: Might be better to fail fast using produce() callbacks rather than wait for the timeout/queue to clear.

sdk/python/feast/loaders/abstract_producer.py
```diff
@@ -47,6 +47,7 @@ def _inc_pbar(self, meta):
     def _set_error(self, exception: str):
         self.error_count += 1
```
zhilingc, line 48: This line can be removed if youre throwing an exception, might be a little redundant

PR #650, opened 2020-04-25
JobUpdateTask cleanups

ches:

Cleanup, this is purely refactoring aiming to make code easier to read, understand, and modify.

I've been doing some work—and reviews of work—around this code, and found it a bit hard to follow and work with. This is my attempt at improving that.

This PR is meant to be reviewed commit-by-commit, it should hopefully make sense quickly that way. Tests are untouched until the final commit and should pass at every step.

I have a question or two though that I'll ask inline, so I'm marking this WIP, I may update a bit more based on the answers, especially docstrings where warranted.

No issue

```release-note
NONE
```

### DIFF
--- a/core/src/main/java/feast/core/job/JobUpdateTask.java
+++ b/core/src/main/java/feast/core/job/JobUpdateTask.java
@@ -56,8 +56,9 @@ public class JobUpdateTask implements Callable<Job> {
   private final SourceProto.Source sourceSpec;
   private final StoreProto.Store store;
   private final Optional<Job> currentJob;
-  private JobManager jobManager;
-  private long jobUpdateTimeoutSeconds;
+  private final JobManager jobManager;
+  private final long jobUpdateTimeoutSeconds;
+  private final String runnerType;
 
   public JobUpdateTask(
       List<FeatureSetProto.FeatureSet> featureSets,
@@ -73,6 +74,7 @@ public class JobUpdateTask implements Callable<Job> {
     this.currentJob = currentJob;
     this.jobManager = jobManager;
     this.jobUpdateTimeoutSeconds = jobUpdateTimeoutSeconds;
+    this.runnerType = jobManager.getRunnerType().toString();
   }
 
   @Override
@@ -89,19 +91,16 @@ public class JobUpdateTask implements Callable<Job> {
           featureSets.stream()
               .map(fs -> FeatureSet.fromProto(fs).getId())
               .collect(Collectors.toSet());
+
       if (newFeatureSetsPopulatedByJob.equals(existingFeatureSetsPopulatedByJob)) {
-        JobStatus currentJobStatus = job.getStatus();
-        JobStatus newJobStatus = jobManager.getJobStatus(job);
-        if (newJobStatus != currentJobStatus) {
-          AuditLogger.log(
-              Resource.JOB,
-              job.getId(),
-              Action.STATUS_CHANGE,
-              "Job status updated: changed from %s to %s",
-              currentJobStatus,
-              newJobStatus);
+        JobStatus currentStatus = job.getStatus();
+        JobStatus newStatus = jobManager.getJobStatus(job);
+        if (newStatus != currentStatus) {
+          var auditMessage = "Job status updated: changed from %s to %s";
+          logAudit(Action.STATUS_CHANGE, job, auditMessage, currentStatus, newStatus);
         }
-        job.setStatus(newJobStatus);
+
+        job.setStatus(newStatus);
         return job;
       } else {
         submittedJob = executorService.submit(() -> updateJob(job, featureSets, store));
@@ -138,36 +137,23 @@ public class JobUpdateTask implements Callable<Job> {
             featureSetsFromProto(featureSetProtos),
             JobStatus.PENDING);
     try {
-      AuditLogger.log(
-          Resource.JOB,
-          jobId,
-          Action.SUBMIT,
-          "Building graph and submitting to %s",
-          jobManager.getRunnerType().toString());
+      logAudit(Action.SUBMIT, job, "Building graph and submitting to %s", runnerType);
 
       job = jobManager.startJob(job);
-      if (job.getExtId().isEmpty()) {
+      var extId = job.getExtId();
+      if (extId.isEmpty()) {
         throw new RuntimeException(
             String.format("Could not submit job: \n%s", "unable to retrieve job external id"));
       }
 
-      AuditLogger.log(
-          Resource.JOB,
-          jobId,
-          Action.STATUS_CHANGE,
-          "Job submitted to runner %s with ext id %s.",
-          jobManager.getRunnerType().toString(),
-          job.getExtId());
+      var auditMessage = "Job submitted to runner %s with ext id %s.";
+      logAudit(Action.STATUS_CHANGE, job, auditMessage, runnerType, extId);
 
       return job;
     } catch (Exception e) {
       log.error(e.getMessage());
-      AuditLogger.log(
-          Resource.JOB,
-          jobId,
-          Action.STATUS_CHANGE,
-          "Job failed to be submitted to runner %s. Job status changed to ERROR.",
-          jobManager.getRunnerType().toString());
+      var auditMessage = "Job failed to be submitted to runner %s. Job status changed to ERROR.";
+      logAudit(Action.STATUS_CHANGE, job, auditMessage, runnerType);
 
       job.setStatus(JobStatus.ERROR);
       return job;
@@ -178,14 +164,8 @@ public class JobUpdateTask implements Callable<Job> {
   private Job updateJob(
       Job job, List<FeatureSetProto.FeatureSet> featureSets, StoreProto.Store store) {
     job.setFeatureSets(featureSetsFromProto(featureSets));
-    job.setStore(feast.core.model.Store.fromProto(store));
-    AuditLogger.log(
-        Resource.JOB,
-        job.getId(),
-        Action.UPDATE,
-        "Updating job %s for runner %s",
-        job.getId(),
-        jobManager.getRunnerType().toString());
+    job.setStore(Store.fromProto(store));
+    logAudit(Action.UPDATE, job, "Updating job %s for runner %s", job.getId(), runnerType);
     return jobManager.updateJob(job);
   }
 
@@ -207,5 +187,9 @@ public class JobUpdateTask implements Callable<Job> {
                         .setMeta(fsp.getMeta())
                         .build()))
         .collect(Collectors.toList());
+  }
+
+  private void logAudit(Action action, Job job, String detail, Object... args) {
+    AuditLogger.log(Resource.JOB, job.getId(), action, detail, args);
   }
 }
--- a/core/src/main/java/feast/core/job/JobUpdateTask.java
+++ b/core/src/main/java/feast/core/job/JobUpdateTask.java
@@ -80,39 +80,24 @@ public class JobUpdateTask implements Callable<Job> {
   @Override
   public Job call() {
     ExecutorService executorService = Executors.newSingleThreadExecutor();
-    Source source = Source.fromProto(sourceSpec);
     Future<Job> submittedJob;
 
-    if (currentJob.isPresent()) {
-      Job job = currentJob.get();
-      Set<String> existingFeatureSetsPopulatedByJob =
-          job.getFeatureSets().stream().map(FeatureSet::getId).collect(Collectors.toSet());
-      Set<String> newFeatureSetsPopulatedByJob =
-          featureSets.stream()
-              .map(fs -> FeatureSet.fromProto(fs).getId())
-              .collect(Collectors.toSet());
-
-      if (newFeatureSetsPopulatedByJob.equals(existingFeatureSetsPopulatedByJob)) {
-        JobStatus currentStatus = job.getStatus();
-        JobStatus newStatus = jobManager.getJobStatus(job);
-        if (newStatus != currentStatus) {
-          var auditMessage = "Job status updated: changed from %s to %s";
-          logAudit(Action.STATUS_CHANGE, job, auditMessage, currentStatus, newStatus);
-        }
-
-        job.setStatus(newStatus);
-        return job;
-      } else {
-        submittedJob = executorService.submit(() -> updateJob(job, featureSets, store));
-      }
+    if (currentJob.isEmpty()) {
+      submittedJob = executorService.submit(this::createJob);
     } else {
-      String jobId = createJobId(source.getId(), store.getName());
-      submittedJob = executorService.submit(() -> startJob(jobId, featureSets, sourceSpec, store));
+      Job job = currentJob.get();
+
+      if (featureSetsChangedFor(job)) {
+        submittedJob = executorService.submit(() -> updateJob(job));
+      } else {
+        return updateStatus(job);
+      }
     }
 
     try {
       return submittedJob.get(getJobUpdateTimeoutSeconds(), TimeUnit.SECONDS);
     } catch (InterruptedException | ExecutionException | TimeoutException e) {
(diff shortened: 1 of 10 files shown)


core/src/main/java/feast/core/service/JobService.java
```diff
@@ -158,16 +158,15 @@ public RestartIngestionJobResponse restartJob(RestartIngestionJobRequest request
     // check job exists
     Optional<Job> getJob = this.jobRepository.findById(request.getId());
     if (getJob.isEmpty()) {
+      // FIXME: if getJob.isEmpty then constructing this error message will always throw an error...
       throw new NoSuchElementException(
           "Attempted to stop nonexistent job with id: " + getJob.get().getId());
```
ches, line 163: This is bug, I'll raise it in a separate issue/PR, didn't want to change it here because it should probably have a hotfix instead of waiting on this refactoring PR to get through. There are two or three instances of the same problem in this file.

The intention here was probably to use `request.getId()` for the exception message. `getJob.get()` will throw `NoSuchElementException` by definition. @mrzzy

core/src/main/java/feast/core/job/JobUpdateTask.java
```diff
@@ -73,147 +74,131 @@ public JobUpdateTask(
     this.currentJob = currentJob;
     this.jobManager = jobManager;
     this.jobUpdateTimeoutSeconds = jobUpdateTimeoutSeconds;
+    this.runnerType = jobManager.getRunnerType().toString();
   }
 
   @Override
   public Job call() {
     ExecutorService executorService = Executors.newSingleThreadExecutor();
```
ches, line 82: @zhilingc I'm trying to understand why this is used—`JobUpdateTask` is a `Callable` and its instances are all dispatched on separate threads by `JobCoordinatorService`, it seems redundant to have the tasks then use another thread again where they invoke a job manager (that will [usually] do network I/O). If anything, it would seem that futures should be at the layer of the job manager instead. `call()` submits and then blocks on them before returning.

One more question for you is about the parameters of helper methods like `startJob` and `updateJob`: most of them except the job itself are instance state of the `JobUpdateTask` itself, and they were able to be `final`, so it seems like there's no need for passing them around through arguments. I removed them in the process of working out what `call()` actually needed to touch. Am I missing anything, or were they just a vestige of some refactoring in your initial work?

core/src/main/java/feast/core/job/JobUpdateTask.java
```diff
@@ -73,147 +74,131 @@ public JobUpdateTask(
     this.currentJob = currentJob;
     this.jobManager = jobManager;
     this.jobUpdateTimeoutSeconds = jobUpdateTimeoutSeconds;
+    this.runnerType = jobManager.getRunnerType().toString();
```
ches, line 77: We had confusion and almost a bug before around `Job#runner` being a string, which I _think_ #575 has fixed, but I feel this is still very error-prone. I think I want to also refactor `Job#runner` to be the enum type instead of stringly-typed programming, I'll see how much that change hurts but it should probably be another PR anyway.

core/src/main/java/feast/core/job/JobUpdateTask.java
```diff
@@ -73,147 +74,131 @@ public JobUpdateTask(
...
-      Set<String> existingFeatureSetsPopulatedByJob =
-          currentJob.get().getFeatureSets().stream()
-              .map(FeatureSet::getId)
-              .collect(Collectors.toSet());
-      Set<String> newFeatureSetsPopulatedByJob =
-          featureSets.stream()
-              .map(fs -> FeatureSet.fromProto(fs).getId())
-              .collect(Collectors.toSet());
-      if (existingFeatureSetsPopulatedByJob.size() == newFeatureSetsPopulatedByJob.size()
-          && existingFeatureSetsPopulatedByJob.containsAll(newFeatureSetsPopulatedByJob)) {
```
ches, line 93: This was [the definition of `Set#equals`][1] ☺️ 

[1]: https://github.com/openjdk/jdk/blob/5fc5cb9f252310634e36ed2ce2e09fa42e430bcf/src/java.base/share/classes/java/util/AbstractSet.java#L85-L99

core/src/main/java/feast/core/model/JobStatus.java
```diff
@@ -53,33 +51,41 @@
...
-   * <p>Terminal job state is final and will not change to any other state.
+   * <p>A terminal job state is final and will not change to any other state.
    *
-   * @return collection of terminal job state.
+   * @return set of terminal job states.
    */
-  public static Collection<JobStatus> getTerminalState() {
-    return TERMINAL_STATE;
+  public static Set<JobStatus> getTerminalStates() {
+    return TERMINAL_STATES;
```
ches, line 64: Got a little carried away, `JobStatus` changes are further than I meant to go for now… but I started touching `JobCoordinatorService` for `filter(job -> !job.hasTerminated())` and I ended up here…

These semantically are/should be Sets I think. I'm not sure I see value in these getters wrapping constants, after this PR their only remaining use is in a few tests. Considered removing them, but thought I'd leave it to discuss first.

core/src/main/java/feast/core/job/JobUpdateTask.java
ches: The enum matter is raised in #651.

zhilingc:
Thanks for the PR, I'll have a look tomorrow!

core/src/main/java/feast/core/job/JobUpdateTask.java
zhilingc: After reading through the code I'm not sure what my past self was thinking tbqh :( You're right, its completely redundant. 

As for the second point, you're right. I must've missed it.

core/src/main/java/feast/core/model/JobStatus.java
zhilingc: Makes sense.

zhilingc:
No complaints here, thanks so much for the PR, and sorry for the mess :/

core/src/main/java/feast/core/service/JobService.java
mrzzy: Good catch. It should be `request.getId()` instead.

core/src/main/java/feast/core/job/JobUpdateTask.java
```diff
@@ -73,147 +74,131 @@ public JobUpdateTask(
...
     Job job =
         new Job(
             jobId,
             "",
             jobManager.getRunnerType().name(),
-            Source.fromProto(source),
-            Store.fromProto(sinkSpec),
-            featureSets,
+            Source.fromProto(sourceSpec),
+            Store.fromProto(store),
```
mrzzy, line 134: Nitpick here but we should probably make our naming consistent:
- suffix `Spec` (ie `sourceSpec` & `storeSpec`) for protobuf versions of the object.
- reserve non suffixed `source` & `store` for native version of the object

core/src/main/java/feast/core/job/JobUpdateTask.java
```diff
@@ -73,147 +74,131 @@ public JobUpdateTask(
...
+  private static List<FeatureSet> featureSetsFromProto(List<FeatureSetProto.FeatureSet> protos) {
+    return protos.stream()
+        .map(
+            fsp ->
+                FeatureSet.fromProto(
+                    FeatureSetProto.FeatureSet.newBuilder()
+                        .setSpec(fsp.getSpec())
+                        .setMeta(fsp.getMeta())
+                        .build()))
+        .collect(Collectors.toList());
```
mrzzy, line 198: @zhilingc is the difference between calling:
1. ```
FeatureSet.fromProto(
                    FeatureSetProto.FeatureSet.newBuilder()
                        .setSpec(fsp.getSpec())
                        .setMeta(fsp.getMeta())
                        .build()))
```
and 2. `FeatureSet.fromProto(fsp)`?

If not we just do 2. and simplify the code.

Comment (woop) [reply core/src/main/java/feast/core/job/JobUpdateTask.java:198]: Right now it doesn't seem like there is any difference.

Comment (zhilingc) [reply core/src/main/java/feast/core/job/JobUpdateTask.java:198]: No. I think this was leftover from when we were deliberating whether to pass only specs to the job (or the full proto object)

File: core/src/main/java/feast/core/job/JobUpdateTask.java (hunk 6 above)
Comment (woop) [reply core/src/main/java/feast/core/job/JobUpdateTask.java:134]: Agreed. That was the original intention, but it seems like some kind of drift has occurred.

**Comment by woop** [comment]
Thanks @ches, looks great. The only open item from my side is @mrzzy's [comment](https://github.com/gojek/feast/pull/650#discussion_r415501353). Otherwise it looks good to me.

File: core/src/main/java/feast/core/job/JobUpdateTask.java (hunk 6 above)
Comment (ches) [reply core/src/main/java/feast/core/job/JobUpdateTask.java:134]: I can do that. In light of @zhilingc's [answer above][1] I'll update the PR with removing the Executor usage in `JobUpdateTask`, so I'll add a commit making the naming consistent too.

I wasn't fully noticing this intended convention myself—"feature set spec" (especially since v0.4) I got because of `spec` and `meta` as children of `FeatureSet` in the schema, and generally the distinction between the "registry specification" and a concrete set of values, an "instance" of a feature set you might say. Makes sense that the same logic applies to sources and stores, and hopefully as more domain model classes are getting defined now we will start to see fewer of the "specs" in core logic and that naming will make them stand out.

[1]: https://github.com/gojek/feast/pull/650#discussion_r415472243

**Comment by woop** [comment]
Just FYI we are considering enabling the e2e tests again by default. The added benefit of not having to wait for them in small PRs isn't worth the manual effort for big PRs.

**Comment by ches** [comment]
Okay, so I did the spec renaming for consistency [as @mrzzy suggested][1], and when I stepped back and looked at it, all the instance variables of `Job` were protobuf types and it made me sad. So in the last commit I went for the full monty and converted them all to domain types, `JobCoordinatorService` now unmarshals protos to domain models as soon as it takes them off the wire so they're not passed around the program anywhere. I think the result is much nicer, no to and fro conversions happening in several places across classes.

I tried [removing the redundant `Executor` in `JobUpdateTask`][2] and made a realization: it currently serves to impose the per-task timeout. In `JobCoordinatorService` we do `CompletionService#take()` which can block indefinitely, so it relies on the tasks timing out internally in which case the callable returns null. Changing from `take()` to `poll(long timeout, TimeUnit unit)` would require updating logic further in the coordinator to preserve existing behavior, so I'm going to leave it alone for now.

I'm leaning more toward my instinct in the original comment/question for @zhilingc being the right answer eventually: the asynchrony (and possibility of failure) should be expressed at the `JobManager` interface layer, indeed [there is an existing TODO][3] that it needs a timeout.

[1]: https://github.com/gojek/feast/pull/650#discussion_r415501353
[2]: https://github.com/gojek/feast/pull/650#discussion_r415130044
[3]: https://github.com/gojek/feast/blob/45e6500ca5dfa78c36be95069863c515d762dc1b/core/src/main/java/feast/core/job/dataflow/DataflowJobManager.java#L303

**Comment by ches** [comment]
The job manager soon after converts things back to proto again, but that's natural, these are I/O boundaries of the system. The overhead of these conversions is the price we pay if we don't want wire format in business logic layer, which is probably worth it. It ought to be less than the actual de/serialization anyway, and this stuff isn't a hot path/tight loop 🤞

**Comment by ches** [comment]
Oh, finally, let me know if you'd like me to try to squash a few commits. I think you're typically squashing merges anyway so it'd not be time well spent in that case.

File: core/src/main/java/feast/core/service/JobCoordinatorService.java [hunk 8]
```diff
@@ -148,27 +150,23 @@ public void Poll() throws InvalidProtocolBufferException {
...
     }
-
-    log.info("Updating feature set status");
-    updateFeatureSetStatuses(jobUpdateTasks);
+    executorService.shutdown();
   }
 
   // TODO: make this more efficient
   private void updateFeatureSetStatuses(List<JobUpdateTask> jobUpdateTasks) {
     Set<FeatureSet> ready = new HashSet<>();
```
woop, line 158: Not really related to this PR per say, but would it be possible for us to remove this `updateFeatureSetStatuses` call and use the real job state?

core/src/main/java/feast/core/job/JobUpdateTask.java
```diff
@@ -52,168 +49,130 @@
...
       long jobUpdateTimeoutSeconds) {
 
     this.featureSets = featureSets;
-    this.sourceSpec = sourceSpec;
+    this.source = source;
     this.store = store;
     this.currentJob = currentJob;
     this.jobManager = jobManager;
     this.jobUpdateTimeoutSeconds = jobUpdateTimeoutSeconds;
+    this.runnerName = jobManager.getRunnerType().toString();
```
woop, line 74: The `runnerName` as we have it above refers to the `runnerType` right? Should we deduplicate that from the configuration (we could do that at the configuration level as well)

```
feast:
  jobs:
    polling_interval_milliseconds: 60000
    job_update_timeout_seconds: 240

    active_runner: my_direct_runner

    runners:
      - name: my_direct_runner
        type: DirectRunner
        options: {}
```

https://github.com/gojek/feast/blame/ded7ca59a2bd5b40715af73fdaa4e4f19ad0b915/core/src/main/java/feast/core/config/FeastProperties.java#L95

woop: The idea with the configuration was to provide a more forward compatible configuration schema, but we could just as easily use type there instead of name.

core/src/main/java/feast/core/service/JobCoordinatorService.java
ches: I think you're asking: rather than cache feature set statuses in bulk, could status be returned on-demand by checking job state for one given feature set, when requested? I guess the team will have more context on the design.

As it stands, I'd have to get my head more into functional consideration of what it's doing beyond the mechanical refactoring I've done to this class, but it seems like there could be more to look at later:

1. In the `Poll()` loop we are:
   - At the beginning, doing `getJob` for the source/store combo of all feature sets that are subscribed to
   - Then, doing the `startOrUpdate` of tasks for all those, and waiting on all those tasks
   - Throwing away the results of waiting (updated `Job` references), because the `startOrUpdate` returns `void`
   - Doing `getJob` again for each of the tasks, in the `updateFeatureSetStatuses`

   It seems like this second round of DB lookups could be avoided by keeping and using the `Job`s we just got back. Unless I'm missing something, if it's trying to avoid a race condition or something.
2. `getJob` does `jobRepository.findBySourceIdAndStoreNameOrderByLastUpdatedDesc` returning all matching records, and then takes the first (most recent) one client-side instead of having SQL do it. And this happens in loops, 2x per polling loop as noted in 1.

core/src/main/java/feast/core/job/JobUpdateTask.java
ches: Yes, from your example config, `jobManager.getRunnerType().toString()` will be `"DirectRunner"` (whereas `jobManager.getRunnerType().name()` would be `"DIRECT"`).

In `JobUpdateTask` this value is being used only for log messages.

ches:
Is there anything further I should update on this one?

core/src/main/java/feast/core/service/JobCoordinatorService.java
woop: Tracking your comments here: #664 

> and then takes the first (most recent) one client-side instead of having SQL do it. 

That's a very low hanging fruit.

woop:
> I'm leaning more toward my instinct in the original comment/question for @zhilingc being the right answer eventually: the asynchrony (and possibility of failure) should be expressed at the JobManager interface layer, indeed there is an existing TODO that it needs a timeout.

This seems reasonable, although I have only superficially looked at job management. I do think we need to look at this in a bit more detail soon though, especially if folks are bringing their own runners.

PR #651, opened 2020-04-26
core: Use Runner enum type instead of string for Job model

ches:

This is refactoring aiming to remove a burden on the programmer to get something right.

#575 sought to clear up inconsistencies between uses of `Runner#name()` (the standard final method of `java.lang.Enum` that returns the value's enum constant name) and the riskily-named `Runner#getName()` defined in Feast for human-readable Beam Runner names.

The latter is used as runner name users can set in config (now as `Runner#toString` since #575). The former is used for values of the runner column of the jobs table in SQL (as it should be). But it relied on careful coding to use the right one when constructing `Job` instances. This is error prone, as #578 demonstrates.

There is a more robust way: use the enum instead of stringly-typed programming. It's one of the reasons we have enums ☺️ 

This also renames the internal identifier in the Runner definition to `humanName`, to distinguish it further from `Enum#name()`.

No issue. Relates to #575, #578, brought up in https://github.com/gojek/feast/pull/650#discussion_r415132957

```release-note
NONE
```

### DIFF
--- a/core/src/main/java/feast/core/job/Runner.java
+++ b/core/src/main/java/feast/core/job/Runner.java
@@ -16,33 +16,37 @@
  */
 package feast.core.job;
 
+import java.util.NoSuchElementException;
+
+/**
+ * An Apache Beam Runner, for which Feast Core supports managing ingestion jobs.
+ *
+ * @see <a href="https://beam.apache.org/documentation/#runners">Beam Runners</a>
+ */
 public enum Runner {
   DATAFLOW("DataflowRunner"),
   FLINK("FlinkRunner"),
   DIRECT("DirectRunner");
 
-  private final String name;
+  private final String humanName;
 
-  Runner(String name) {
-    this.name = name;
+  Runner(String humanName) {
+    this.humanName = humanName;
   }
 
-  /**
-   * Get the human readable name of this runner. Returns a human readable name of the runner that
-   * can be used for logging/config files/etc.
-   */
+  /** Returns the human readable name of this runner, usable in logging, config files, etc. */
   @Override
   public String toString() {
-    return name;
+    return humanName;
   }
 
   /** Parses a runner from its human readable name. */
-  public static Runner fromString(String runner) {
+  public static Runner fromString(String humanName) {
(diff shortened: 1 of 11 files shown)


core/src/main/java/feast/core/job/Runner.java
```diff
@@ -16,33 +16,37 @@
...
-  public static Runner fromString(String runner) {
+  public static Runner fromString(String humanName) {
     for (Runner r : Runner.values()) {
-      if (r.toString().equals(runner)) {
+      if (r.toString().equals(humanName)) {
         return r;
       }
     }
-    throw new IllegalArgumentException("Unknown value: " + runner);
+    throw new NoSuchElementException("Unknown Runner value: " + humanName);
```
ches, line 50: I felt this is perhaps a more fitting exception type for this case.

core/src/main/java/feast/core/model/Job.java
```diff
@@ -55,9 +56,9 @@
   private String extId;
 
   // Runner type
-  // Use Runner.name() when converting a Runner to string to assign to this property.
+  @Enumerated(EnumType.STRING)
   @Column(name = "runner")
-  private String runner;
+  private Runner runner;
```
ches, line 61: The removed comment sums up the motivation of the PR.

core/src/main/java/feast/core/service/JobService.java
```diff
@@ -50,13 +51,13 @@
...
+/** A Job Management Service that allows users to manage Feast ingestion jobs. */
 @Slf4j
 @Service
 public class JobService {
-  private JobRepository jobRepository;
-  private SpecService specService;
-  private Map<String, JobManager> jobManagers;
+  private final JobRepository jobRepository;
+  private final SpecService specService;
+  private final Map<Runner, JobManager> jobManagers;
```
woop, line 60: Out of scope for this PR, but I still feel like JobManager can be changed to something closer to Runner. In fact it can probably be `Runner` itself. Meaning this code would become `private final Map<RunnerType, Runner>`

Issue #479, opened 2020-02-18
Feast API: Feature references, concept hierarchy, and data model


woop:
This issue is meant to be a discussion of the current Feast API as it relates to `feature references`,  a key component of the user facing API. Additionally, it will also discuss the current data model and our concept hierarchy. 

## 1. Background
The Feast user facing API and data model changed dramatically from 0.1 to 0.2+. The original intention was to simplify the API as much as possible and gradually evolve it as new user requirements available.

Two important reference documents on this topic are
* [Feast 0.3 RFC](https://docs.google.com/document/d/1QnUQWhwJ1fDnMQ4sZdUdLa_4C-x5Hhpw3C8nYLg4nsY/edit)
* [Feast Projects RFC](https://docs.google.com/document/d/14-QBz9X8zK_aGAY2ti43a7PqMxvqp0035ec7QYgmEBM/edit#)

## 2. Problem statement
The Feast API is evolving as more and more teams adopt the software and share their requirements with us. In most cases this means an expansion of the API, but in some cases it means a reversal. 

With the introduction of projects into Feast ([Feast Projects RFC](https://docs.google.com/document/d/14-QBz9X8zK_aGAY2ti43a7PqMxvqp0035ec7QYgmEBM/edit#)), our API has evolved again. This change has affected feature references, the data model, and concept hierarchy.

The most critical feedback on this change has been that it introduces unnecessary complexity to address problems (isolation, namespacing, security), that could be solved in a different way.

## 3. Objective
The point of this GitHub issue is to settle our API for feature references, our concept hierarchy, and data model in such a way that we
* Meet all our known requirements for future development
* Minimize user facing changes and migration requirements
* Maintain flexibility in accepting new user requirements and evolving our API

Put simply, we want to make sure that we are on the right path and make the necessary changes now when its least disruptive.

## 4. What are feature references?
Feature references (previously Feature Ids) are strings/objects within Feast that allows Feast and users of Feast to reference specific features. Feature references are primarily used as a means of indicating to Feast which features a user would like to retrieve.

Originally, feature references were defined as follows
`<feature-set>:<feature-name>:<feature-version>`
All parts of the above reference were required at the time.

Feature references have recently been updated (as part of the [Projects RFC](https://docs.google.com/document/d/14-QBz9X8zK_aGAY2ti43a7PqMxvqp0035ec7QYgmEBM/edit#))

The move towards project namespaces now moves feature sets and features/entities into the following hierarchy
[...]

PR #655, opened 2020-04-26
Split Field model into distinct Feature and Entity objects

zhilingc:
This is a split-off off #612 that introduces the model changes made in that PR in a more digestible chunk.

This PR includes:
- Removal of Field object
- Addition of distinct Feature and Entity objects
- Removal of TFX fields on entities

SQL changes:
- Surrogate `long` ids for feature sets, features and entities
- drop TFX constraints from entities table

```release-note
Model changes to FeatureSets, Features and Entities. Requires Migration.
```


ches:
Could you please summarize in the description what the SQL schema changes are that are implied with this branch? Should the PR include a SQL migration script?

woop:
> Could you please summarize in the description what the SQL schema changes are that are implied with this branch? Should the PR include a SQL migration script?

For the script, should it perhaps be at the release level, or is there an advantage to having it at the PR level?

ches:
> For the script, should it perhaps be at the release level, or is there an advantage to having it at the PR level?

Could be at release. With the PR, reviewers could run it on an existing development environment with data in it and a.) see that the migration script works, and b.) experiment with the PR with some existing data. Manual of course, but it's something.

This would be out of scope for this PR for sure, but separately maybe we could consider integrating something like Flyway into the project, both for development convenience and for shipping migration scripts with Feast releases that operators can have a process to apply. Could be used to load seed data for automated integration tests too.

woop:
> This would be out of scope for this PR for sure, but separately maybe we could consider integrating something like Flyway into the project, both for development convenience and for shipping migration scripts with Feast releases that operators can have a process to apply. Could be used to load seed data for automated integration tests too.

Yea I really like Flyway. I think it makes sense. I have only used it as an external tool in the past in non-JVM projects. It seems like integration here would mean that migrations could be triggered manually using the CLI or `mvn`.

The value add would be mostly in the migration scripts themselves, so perhaps we could start there I think, and add documentation and Flyway around it.

zhilingc:
@ches @woop The diff between the 0.4 schema and 0.5 is the following:
```
create table entity
(
	feature_set varchar(255) not null,
	name varchar(255) not null,
	project varchar(255) not null,
	version integer not null,
	type varchar(255),
	constraint entity_pkey
		primary key (feature_set, name, project, version)
);

create table feature
(
	feature_set varchar(255) not null,
	name varchar(255) not null,
	project varchar(255) not null,
	version integer not null,
	bool_domain bytea,
	domain varchar(255),
	float_domain bytea,
	group_presence bytea,
	image_domain bytea,
	int_domain bytea,
	mid_domain bytea,
	natural_language_domain bytea,
	presence bytea,
	shape bytea,
	string_domain bytea,
	struct_domain bytea,
	time_domain bytea,
	time_of_day_domain bytea,
	type varchar(255),
	url_domain bytea,
	value_count bytea,
	constraint feature_pkey
		primary key (feature_set, name, project, version)
);

create table feature_set_entities
(
	feature_set_id varchar(255) not null
		constraint fkbax98goqv127qc5su483xjhq9
			references feature_sets,
	entities_name varchar(255) not null,
	entities_project varchar(255) not null,
	entities_feature_set_id varchar(255) not null,
	entities_version integer not null,
	constraint feature_set_entities_pkey
		primary key (feature_set_id, entities_name, entities_project, entities_feature_set_id, entities_version),
	constraint uk_byydfwkove4rtygt5pydandx4
		unique (entities_name, entities_project, entities_feature_set_id, entities_version),
	constraint fktm3p2cp241tbqfb7r1ptfegdx
		foreign key (entities_name, entities_project, entities_feature_set_id, entities_version) references entity
);

create index idx_jobs_feature_set_entities_feature_set_id
	on feature_set_entities (feature_set_id);

create table feature_set_features
(
	feature_set_id varchar(255) not null
		constraint fk2w40q1cwd6pnqsbv5rvjkx5wj
			references feature_sets,
	features_name varchar(255) not null,
	features_project varchar(255) not null,
	features_feature_set_id varchar(255) not null,
	features_version integer not null,
	constraint feature_set_features_pkey
		primary key (feature_set_id, features_name, features_project, features_feature_set_id, features_version),
	constraint uk_tngc6xfm924rl6wxhqkb97jdl
		unique (features_name, features_project, features_feature_set_id, features_version),
	constraint fk1qcxf8joepru1qpyg1wu4lqfb
		foreign key (features_name, features_project, features_feature_set_id, features_version) references feature
);

create index idx_jobs_feature_set_features_feature_set_id
	on feature_set_features (feature_set_id);

create index idx_jobs_feature_sets_job_id
	on jobs_feature_sets (job_id);

create index idx_jobs_feature_sets_feature_sets_id
	on jobs_feature_sets (feature_sets_id);

drop table features;

drop table entities;

drop table metrics;
```

I can add migration scripts to this PR if necessary

woop:
Thanks @zhilingc 

From an upgrade perspective I think people care about retaining their existing data. 

>  I'm not sure if column orders can be altered :/

Are any of our systems (like Spring) picky about column order?

zhilingc:
> Are any of our systems (like Spring) picky about column order?

It's not spring, spring doesn't care, but it's not directly supported in postgres.

woop:
> > Are any of our systems (like Spring) picky about column order?
> 
> It's not spring, spring doesn't care, but it's not directly supported in postgres.

Ok, so where does the failure happen if we lose column ordering? I am trying to understand the comment you made above.

zhilingc:
Oh, actually I suppose its not necessary to reorder columns.
```
drop sequence hibernate_sequence;

alter table entities
	add feature_set varchar(255) not null;

-- column reordering is not supported entities.feature_set

create unique index entities_pkey
	on entities (feature_set, name, project, version);

alter table entities
	add constraint entities_pkey
		primary key (feature_set, name, project, version);

create table feature_set_entities
(
	feature_set_id varchar(255) not null
		constraint fkbax98goqv127qc5su483xjhq9
			references feature_sets,
	entities_name varchar(255) not null,
	entities_project varchar(255) not null,
	entities_feature_set_id varchar(255) not null,
	entities_version integer not null,
	constraint feature_set_entities_pkey
		primary key (feature_set_id, entities_name, entities_project, entities_feature_set_id, entities_version),
	constraint uk_byydfwkove4rtygt5pydandx4
		unique (entities_name, entities_project, entities_feature_set_id, entities_version),
	constraint fkqr5sj1yp7uunob8nf8fuxwfmw
		foreign key (entities_name, entities_project, entities_feature_set_id, entities_version) references entities
);

create index idx_jobs_feature_set_entities_feature_set_id
	on feature_set_entities (feature_set_id);

alter table features
	add feature_set varchar(255) not null;

-- column reordering is not supported features.feature_set

alter table features
	add constraint features_pkey
		primary key (feature_set, name, project, version);

create table feature_set_features
(
	feature_set_id varchar(255) not null
		constraint fk2w40q1cwd6pnqsbv5rvjkx5wj
			references feature_sets,
	features_name varchar(255) not null,
	features_project varchar(255) not null,
	features_feature_set_id varchar(255) not null,
	features_version integer not null,
	constraint feature_set_features_pkey
		primary key (feature_set_id, features_name, features_project, features_feature_set_id, features_version),
	constraint uk_tngc6xfm924rl6wxhqkb97jdl
		unique (features_name, features_project, features_feature_set_id, features_version),
	constraint fk9dltpwl1lu6w7cqxyhod9sk7t
		foreign key (features_name, features_project, features_feature_set_id, features_version) references features
);

create index idx_jobs_feature_set_features_feature_set_id
	on feature_set_features (feature_set_id);

alter table entities drop column bool_domain;

alter table entities drop column domain;

alter table entities drop column float_domain;

alter table entities drop column group_presence;

alter table entities drop column image_domain;

alter table entities drop column int_domain;

alter table entities drop column mid_domain;

alter table entities drop column natural_language_domain;

alter table entities drop column presence;

alter table entities alter column project set not null;

alter table entities drop column shape;

alter table entities drop column string_domain;

alter table entities drop column struct_domain;

alter table entities drop column time_domain;

alter table entities drop column time_of_day_domain;

alter table entities drop column url_domain;

alter table entities drop column value_count;

-- column reordering is not supported entities.version

alter table entities alter column version set not null;

alter table entities alter column type drop not null;

alter table entities drop constraint fkhyblh5sfunv00a8ums8ms9otq;

alter table entities drop column feature_set_id;

alter table features alter column project set not null;

alter table features alter column type drop not null;

-- column reordering is not supported features.version

alter table features alter column version set not null;

drop index uk5trg3fpcjbjw5w2dpuuqswdxh;

alter table features drop constraint uk5trg3fpcjbjw5w2dpuuqswdxh;

alter table features drop constraint fkfxcpsscvj0g89o4p5dx4insb1;

alter table features drop column feature_set_id;

drop table metrics;
```

zhilingc:
> An aside about SQL data model in general, not specific to this PR: any particular reasons for the heavy use of string keys? There are several cases where a string PK is entirely made up of data in other columns in the table, they could possibly be composite keys instead of surrogates. If there's going to be a surrogate anyway, I feel like auto-generated numeric IDs would get us into less trouble down the road than "project/feature_set_name:42" when versions are going away and projects may have changes coming…

No excuses here, it's an artifact from when we initially wrote Feast, and it's unfortunately stuck around since. It's going to be hell to migrate existing feature sets if not version controlled because of this, you're right :( I'm happy to change it in this PR, but would that expand this PR's scope too much? 

Any opinions @woop, @khorshuheng ?

woop:
> > An aside about SQL data model in general, not specific to this PR: any particular reasons for the heavy use of string keys? There are several cases where a string PK is entirely made up of data in other columns in the table, they could possibly be composite keys instead of surrogates. If there's going to be a surrogate anyway, I feel like auto-generated numeric IDs would get us into less trouble down the road than "project/feature_set_name:42" when versions are going away and projects may have changes coming…
> 
> No excuses here, it's an artifact from when we initially wrote Feast, and it's unfortunately stuck around since. It's going to be hell to migrate existing feature sets if not version controlled because of this, you're right :( I'm happy to change it in this PR, but would that expand this PR's scope too much?
> 
> Any opinions @woop, @khorshuheng ?

Not much to say around the actual issue. I only see downsides to composed keys like these. 

Same goes for sources by the way

image

>  If there's going to be a surrogate anyway, I feel like auto-generated numeric IDs would get us into less trouble

Yea I am in favor of surrogates with a unique constraint on the candidate key (project, name, version), even if we don't see the optimization benefits, its probably safer long term. I will defer to @zhilingc around the effort in implementing this with surrogates vs composite primary key (both of which I prefer to the string approach).

I do think it would be a separate PR and I think if we are going to make data model changes then now would be the time (for 0.5). I guess this all comes down to how blocked you are right now and how much work you think this is.

ches:
> I do think it would be a separate PR and I think if we are going to make data model changes then now would be the time (for 0.5).

Agreed, wouldn't want to hold this one up further and especially because of @woop's point that it applies to areas like Sources too, which are not the scope of this PR. Probably makes sense to confront them all at once, unless @zhilingc thinks this is the best time/place to do so for the tables in this PR. If there is wide agreement this is something to fix, then seems best to do it in a release that's already breaking schema. I feel like I keep dog-piling your release scope though… I'll see where I can help.

zhilingc:
It's ok, i can make the change here. Give me a bit..

ches:
Another question that comes to mind as I'm integrating labels:

With this PR killing the `Field` class, are we going to remove all the TFDV presence & domain stats fields in the `EntitySpec` proto, until that feature is reconsidered for entities? Those fields have not been in a release yet, but `labels` is going to be field number 19 in my schema so I either want to renumber now if Feast is going to, or we decide to set a range as reserved.

By the way, according to [this][1], I wonder if we should consider putting the stats in a higher range of field numbers.

[1]: https://developers.google.com/protocol-buffers/docs/proto#assigning-field-numbers

woop:
> With this PR killing the `Field` class, are we going to remove all the TFDV presence & domain stats fields in the `EntitySpec` proto, until that feature is reconsidered for entities? Those fields have not been in a release yet, but `labels` is going to be field number 19 in my schema so I either want to renumber now if Feast is going to, or we decide to set a range as reserved.

Yes, we shouldn't have any statistics fields on entities. You can renumber now. 

> By the way, according to [this](https://developers.google.com/protocol-buffers/docs/proto3#assigning-field-numbers), I wonder if we should consider putting the stats in a higher range of field numbers.

Good point. I think we should.

ches:
My mistake, `labels` as field number 19 is on `FeatureSpec`, not `EntitySpec`, however that still might move if the stats fields are moved on `FeatureSpec`. I'll open an issue about the latter.

zhilingc:
I was looking at the code and I think for sources it's a little more complicated since sources are not just KafkaSources, and most of the object is part of its id, so to speak. I've changed the feature set id though.

ches:
> With this PR killing the `Field` class, are we going to remove all the TFDV presence & domain stats fields in the `EntitySpec` proto, until that feature is reconsidered for entities?

Oh, @zhilingc already did it in this PR. I could swear I looked before I said this…

core/src/main/java/feast/core/service/JobCoordinatorService.java
```diff
@@ -195,8 +195,13 @@ private void updateFeatureSetStatuses(List<JobUpdateTask> jobUpdateTasks) {
...
-  // TODO: Put in a util somewhere?
-  private static List<FeatureSet> featureSetsFromProto(List<FeatureSetProto.FeatureSet> protos) {
-    return protos.stream().map(FeatureSet::fromProto).collect(Collectors.toList());
+  private List<FeatureSet> featureSetsFromProto(List<FeatureSetProto.FeatureSet> protos) {
+    return protos.stream()
+        .map(FeatureSetProto.FeatureSet::getSpec)
+        .map(
+            fs ->
+                featureSetRepository.findFeatureSetByNameAndProject_NameAndVersion(
+                    fs.getName(), fs.getProject(), fs.getVersion()))
```
ches, line 204: Could we do this without N+1 queries?

core/src/main/java/feast/core/job/JobUpdateTask.java
```diff
@@ -103,9 +103,13 @@ public Job call() {
 
   boolean featureSetsChangedFor(Job job) {
     Set<String> existingFeatureSetsPopulatedByJob =
-        job.getFeatureSets().stream().map(FeatureSet::getId).collect(Collectors.toSet());
+        job.getFeatureSets().stream()
+            .map(fs -> fs.getProject() + "/" + fs.getName())
```
ches, line 107: Seems like it should be encapsulated in a model, not another encoding the programmer needs knowledge of at every use site.

protos/feast/core/FeatureSet.proto
```diff
@@ -71,46 +71,6 @@ message EntitySpec {
 
     // Value type of the feature.
```
ches, line 72: `s/feature/entity`

ches:
> I am in favor of surrogates with a unique constraint on the candidate key (project, name, version)

@zhilingc I think we've missed this constraint bit.

core/src/main/java/feast/core/service/JobCoordinatorService.java
zhilingc: It's a little hard unless i refactor the spec service so that the `listFeatureSets` logic is available to the job service :/ Any suggestions here?

zhilingc: I'm not that keen on creating a composite id for featuresets since many other tables join on featuresets, and join tables really add to the complexity of the data model.

woop: > I'm not that keen on creating a composite id for featuresets since many other tables join on featuresets, and join tables really add to the complexity of the data model.

Was this comment meant to be somewhere else?

ches: > It's a little hard unless i refactor the spec service so that the `listFeatureSets` logic is available to the job service :/ Any suggestions here?

Maybe let's just stick an `OPTIMIZE:` comment on it, and add it to the scope of #664.

zhilingc: ok cool

woop: > It's a little hard unless i refactor the spec service so that the listFeatureSets logic is available to the job service :/ Any suggestions here?

I think we should be able to create another query right? You already have the `featureSetRepository`,

core/src/main/java/feast/core/model/FeatureSet.java
```diff
@@ -94,19 +61,40 @@
...
+        @JoinColumn(name = "features_name"),
+        @JoinColumn(name = "features_project"),
+        @JoinColumn(name = "features_feature_set_id"),
+        @JoinColumn(name = "features_version")
+      },
+      indexes = {
+        @Index(
+            name = "idx_jobs_feature_set_features_feature_set_id",
+            columnList = "feature_set_id"),
+      })
```
woop, line 96: I really don't want to delay this PR any more because we do have some time pressure, but since its a data model I think its worth making sure that we are getting it right. In previous discussions ([Add computation and retrieval PR](https://github.com/gojek/feast/pull/612#issuecomment-616520831)) we agreed on the removal of Fields in a kind of evolution of the existing schema.

However, looking at the schema now it still seems kind of complicated (needing join tables). I am not sure if it is necessarily complicated or unnecessarily. I don't know all the Springisms and how that affects our data model, but just going by how this could be modeled in SQL without Spring. Is there anything that prevents us from following the more normalized approach as follows?

```
Project
- id (PK)
- name
- ...

Feature set
- id (PK)
- project_id (unique 1/3 )
- name (unique 1/3 )
- version (unique 1/3 )
- ...

Feature (or Entity)
- id (PK)
- feature_set_id (unique 1/2 )
- name (unique 1/2 )
```

All of the existing functionality like feature references would still hold, I am thinking purely from a persistence standpoint.

zhilingc: Having a placeholder id means that whenever you want to join something to that object you will have to look up the actual object in the database to get the pkey instead of letting the orm do the heavy lifting. (case in point, how this PR joins job <> feature set).

ches: > (case in point, how this PR joins job <> feature set).

Could you elaborate a bit more with an example and counterexample? It's hard to extrapolate from code what ORMs will do with a bunch of annotations, or their entirely implicit behavior, without running the code oneself with SQL logging turned up.

zhilingc: With a proper id (containing the actual identifying elements of an object, like `FeatureReference` in this PR), you are able to reference that object to perform joins on or mutate the object using those identifying elements. This is useful because we get these references from either the protos or user requests.

If feature set had an id, [this](https://github.com/gojek/feast/blob/9ba8fee75a3b5976bb819a1bee8ab5f7fd9c484e/core/src/main/java/feast/core/service/JobCoordinatorService.java#L204) doesn't need to be done, because i could just instantiate the FeatureSet from the proto and then make the status changes, and the correct FeatureSet would be updated.

In the absence of such an id, the placeholder id has to be retrieved each time to point to that exact object whenever you want to join something else to it.

zhilingc: I'm sorry if what i said is a little incoherent :( the model proposed by willem is possible, and there are ways around what i've mentioned, but having natural ids would allow us to move between protos and db objects more naturally when forming relationships between feast objects is what I mean

zhilingc: But i'm not an expert, so if you guys think it's the better approach to use placeholder ids then i'd be happy to implement it

woop: I'm happy to figure out the pros/cons of the natural/surrogate key approach, but do we agree the model above makes sense?

I mean specifically in the sense of not having join tables between features/entities and feature sets, and not persisting parent fields like `project` or `feature set` in the children (entities, features).

woop: > elaborate a bit more with an example and counterexample

Yea I think this would be super useful. I understand the reasoning above, but I am trying to find out what this actually looks like in practice.

zhilingc:
Removed the embedded ids in favor of surrogate ids. I've retained the child -> parent relationship because [bidirectional relationships are more efficient](https://vladmihalcea.com/the-best-way-to-map-a-onetomany-association-with-jpa-and-hibernate/) than unidirectional ones.

core/src/main/java/feast/core/model/FeatureSet.java
```diff
@@ -94,19 +59,20 @@
   private long maxAgeSeconds;
 
   // Entity fields inside this feature set
-  @ElementCollection(fetch = FetchType.EAGER)
-  @CollectionTable(name = "entities", joinColumns = @JoinColumn(name = "feature_set_id"))
-  @Fetch(FetchMode.SUBSELECT)
-  private Set<Field> entities;
+  @OneToMany(
+      mappedBy = "featureSet",
```
woop, line 63: Should this be `feature_set`?

zhilingc: no this is correct

woop:
This looks good to me. Will leave it open until EOD.

ches:
I'm still not crystal clear on what tradeoffs we've just made in the last commit. Could we update examples of the SQL schema and/or migration?

Roughly speaking, I get that with the prior `@EmbeddedId` approach you get a join model, a Java entity class representing join table records. This can have value especially if that model deserves first-class status in the object model—concept + behavior of its own, potential of new non-join columns in the join table in the future that enrich the join model, etc.—as references probably do, viz #674 and its cousins. This is part of what @zhilingc is getting at in https://github.com/gojek/feast/pull/655#discussion_r419201646 I think, although also an efficiency/access pattern angle.

I'm not sure whether they need/warrant a table though (objectively neutral statement here pending pros and cons to weigh, _not_ an "I'm not sure" to be read as "I don't think they do"). Or if there is a join table, should it also use surrogate rather than composite keys because of open questions for the future about reference components, in #479?

Assuming we have simpler schema as per the most recent change, and assuming we need/want references to have first class types so we soon define canonical domain models for them without SQL backing, what have we lost or gained with the resulting design?

Feel somewhat as if we exhausted @zhilingc into submission without hearing some valid considerations articulated.

Sorry to drag out any further. This has begged for a whiteboard session.

zhilingc:
The join tables have been removed. The generated SQL for migration is following:
```
alter sequence hibernate_sequence nominvalue;

alter table entities
	add id bigint not null;

-- column reordering is not supported entities.id

create unique index entities_pkey
	on entities (id);

create unique index uk4hredqqfh86prhp1hf08nofvk
	on entities (name, feature_set_id);

alter table entities
	add constraint entities_pkey
		primary key (id);

alter table entities
	add constraint uk4hredqqfh86prhp1hf08nofvk
		unique (name, feature_set_id);

alter table feature_sets
	add labels text;

create unique index ukei1j8q7sfkjlhuxstxs7s3c6
	on feature_sets (name, version, project_name);

alter table feature_sets
	add constraint ukei1j8q7sfkjlhuxstxs7s3c6
		unique (name, version, project_name);

alter table features
	add id bigint not null;

-- column reordering is not supported features.id

alter table features
	add labels text;

-- column reordering is not supported features.labels

alter table features
	add constraint features_pkey
		primary key (id);

alter table features
	add constraint ukedouxmpcoev743cmstfwq25yp
		unique (name, feature_set_id);

alter table entities drop column bool_domain;

alter table entities drop column domain;

alter table entities drop column float_domain;

alter table entities drop column group_presence;

alter table entities drop column image_domain;

alter table entities drop column int_domain;

alter table entities drop column mid_domain;

alter table entities alter column name drop not null;

alter table entities drop column natural_language_domain;

alter table entities drop column presence;

alter table entities drop column project;

alter table entities drop column shape;

alter table entities drop column string_domain;

alter table entities drop column struct_domain;

alter table entities drop column time_domain;

alter table entities drop column time_of_day_domain;

alter table entities alter column type drop not null;

-- column reordering is not supported entities.feature_set_id

alter table entities alter column feature_set_id type integer using feature_set_id::integer;

alter table entities alter column feature_set_id drop not null;

alter table entities drop column url_domain;

alter table entities drop column value_count;

alter table entities drop column version;

alter table feature_sets alter column id type integer using id::integer;

-- column reordering is not supported features.feature_set_id

alter table features alter column feature_set_id type integer using feature_set_id::integer;

alter table features alter column feature_set_id drop not null;

alter table features alter column name drop not null;

alter table features alter column type drop not null;

drop index uk5trg3fpcjbjw5w2dpuuqswdxh;

alter table features drop constraint uk5trg3fpcjbjw5w2dpuuqswdxh;

alter table features drop column project;

alter table features drop column version;

alter table jobs_feature_sets alter column feature_sets_id type integer using feature_sets_id::integer;

drop table metrics;

```

woop:
Disclaimer 1: My gripe is mostly on the join tables and their value in a one-to-many relationship, and not so much whether surrogates or composite keys are preferable. We don't need join tables to use composite keys, but I understand how they help when you want to propagate all the fields down to children. 

**In favour of hierarchical/normalized approach**
- Surrogates allow the data model to be more flexible and less brittle by allowing the composite fields to change independently of the relationship. 
- Surrogates are more efficient to store and index
- They don't require each table with a foreign key reference to have all the columns in the EmbeddedId, which currently stands at four. This bloats the data model and gives me anxiety.
- Specifically if we only store a field once and don't use EmbeddedIds, then moving that relationship around seems a bit easier. For example we can make features globally unique by removing the unique constraint on the feature set (and possibly removing the column altogether). This would be harder with the composite key approach since we’d need to drop and create new join tables and move the feature set out of the embeddedId. 

**In favour of EmbeddedId approach**
- Migrations are harder with surrogates (at least in my SQL Server experience where the numbers don't matter and the rules are made up)
- Possibly one less query to the database required since Ids can be composed outside of the database.
- EmbeddedIds/Composites are easier to read since you can find the natural keys in each table that reference them.
- EmbeddedIds in our case can map naturally to our Refs. FeatureRef, EntityRef, FeatureSetRef.

> assuming we need/want references to have first class types so we soon define canonical domain models for them without SQL backing

I take it you mean we could have FeatureRefs, FeatureSetRefs, and EntityRefs floating around without each one of them being persisted. Something like GetOnlineFeatures where no persistence layer is touched for these references, but they are still functionally used?

If that is the case then it seems like we have three possible implementations areas:
1. Proto implementation used during cross service communication
2. Canonical domain model used for getting shit done
3. Spring data model used for persistence

It seems like one risk that you are flagging is a potential disconnect between (2) and (3) that may not exist if we followed the EmbeddedId approach where they could be the same thing? 

Whether or not we follow the EmbeddedId (FeatureRef, EntityRef) approach or not, we will still have to make separate classes for FeatureRef, and EntityRef.  I like the idea of these classes being used for querying Features/Entities with [Spring’s Specifications Interface](https://www.baeldung.com/rest-api-search-language-spring-data-specifications). So we basically can get the benefits of composite keys without the horrible (in my opinion) data model, and we’d also not need to query for surrogates. 

It's clear that there are unknowns here and so we should probably err on the side of simplicity and delaying unnecessary complexity. So I think the normalized/surrogate approach has the edge here, but perhaps I am missing some obvious advantage.

Useful links
- Why are composite keys discouraged https://stackoverflow.com/questions/14112839/why-are-composite-keys-discouraged-in-hibernate
- https://stackoverflow.com/questions/23850396/composite-vs-surrogate-keys-for-referential-integrity-in-6nf

woop:
Composite/EmbeddedId (d579424b)
image

Surrogate/Normalized (935602e5)
image

woop:
> Sorry to drag out any further. This has begged for a whiteboard session.

Always happy with good feedback, but we are under time pressure to cut this release in order to roll it out. We still need to get through version removal as well, which is based on this PR.

Async communication seems to be working well here, but if you think jumping on a quick call to discuss would be better then we could do that as well.

ches:
Thanks a lot for the detail @woop, that's the analysis I was after and the ERDs are essentially the whiteboard session I was looking for.

I'm sold on the model without the join tables.

> > assuming we need/want references to have first class types so we soon define canonical domain models for them without SQL backing
> 
> I take it you mean we could have FeatureRefs, FeatureSetRefs, and EntityRefs floating around without each one of them being persisted. Something like GetOnlineFeatures where no persistence layer is touched for these references, but they are still functionally used?

Yes, exactly, and that's precisely a case where looking them up from persistence would be unnecessary, and undesirable unless maybe for a SQL online store.

> 1. Proto implementation used during cross service communication
> 2. Canonical domain model used for getting shit done
> 3. Spring data model used for persistence
> 
> It seems like one risk that you are flagging is a potential disconnect between (2) and (3) that may not exist if we followed the EmbeddedId approach where they could be the same thing?

I hadn't thought very deeply about it yet, but no, so far I'm not worried that maintaining appropriate invariants in code will be a problem here.

I suggest being conservative about committing to proto schemas until we're more certain about further outlook of #479, I think that's discussion for #674.

> Whether or not we follow the EmbeddedId (FeatureRef, EntityRef) approach or not, we will still have to make separate classes for FeatureRef, and EntityRef… This also keeps the FeatureRef and EntityRef pure (annotation free).

Indeed, this is part of what I had in mind in raising discussion of whether join models would have value or not—if they did, it would still likely be cumbersome because of how widely these are going to be needed, beyond SQL persistence contexts as above.

PR #656, opened 2020-04-26
Add unique ingestion id for all batch ingestions

zhilingc:
This is a split-off off #612 that introduces the model changes made in that PR in a more digestible chunk.

Adds a unique identifier for all discrete batch loads of data into feast for re-referencing the set of data later on.

```release-note
Adds dataset_id field into BQ
```

### DIFF
--- a/protos/feast/core/Store.proto
+++ b/protos/feast/core/Store.proto
@@ -69,6 +69,7 @@ message Store {
     // ====================|==================|================================
     // - event_timestamp   | TIMESTAMP        | event time of the FeatureRow
     // - created_timestamp | TIMESTAMP        | processing time of the ingestion of the FeatureRow
+    // - dataset_id        | STRING           | identifier of the batch dataset a row belongs to
     // - job_id            | STRING           | identifier for the job that writes the FeatureRow to the corresponding BigQuery table
     //
     // BigQuery table created will be partitioned by the field "event_timestamp"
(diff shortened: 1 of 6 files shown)


woop:
I am personally happy to merge this at any time. I'll leave it open for a little bit.

zhilingc:
@ches I just felt it was more intuitive to provide a dataset id over an ingestion id, and in the old PR I originally offered the functionality for the user to be able to set it, but it was punted to when we integrate these ids into batch retrieval. But i can't say i'm too opinionated about this. if you guys think it should be `ingestion_id` I'm happy to change it.

woop:
> @ches I just felt it was more intuitive to provide a dataset id over an ingestion id, and in the old PR I originally offered the functionality for the user to be able to set it, but it was punted to when we integrate these ids into batch retrieval. But i can't say i'm too opinionated about this. if you guys think it should be `ingestion_id` I'm happy to change it.

I think your previous concern was more around the retrieval of statistics based on an `ingestion` filter being unintuitive? I can see that.

We are starting to overload this `ingestion` term as well, with `IngestionJob` in the Python SDK referring to the population jobs. There is a bit of disconnect between those two ingestion flows right now, but I do think they can be one flow (ingest, populate) in the future so keeping the names the same seems reasonable to me. 

So anyway, my preference is `ingestion_id`, but I also don't think `dataset_id` would work in some sense.

zhilingc:
Renamed to `ingestion_id` and rebased on master.

PR #657, opened 2020-04-27
Move e2e test scripts from .prow to infra

khorshuheng:
In master branch, prow has been configured such that the scripts are run from the infra directory rather than .prow. This change was back-ported to v0.4 branch, but not for v0.3 branch, which will cause the prow tests to fail.

.prow config has been updated in v0.3 branch for clarity purpose.


ches:
/lgtm

Thanks for quickly doing the backport!

PR #536, opened 2020-03-16
Add feature and feature set labels, for metadata

imjuanleonard:
**What this PR does / why we need it**:
-> Extension for #463 

Update (@suwik): 
- Addressed previous review comments
- Added labels on the feature set level
- Removed the python SDK changes related to labels implemented inside field.py (30d63fa0fa)
- Made a small unit test refactoring on the way.

**Which issue(s) this PR fixes**:
Fixes #463 

**Does this PR introduce a user-facing change?**:
```release-note
Feature spec and Feature set spec will both have a new field called labels
```

PR #658, opened 2020-04-27
Add Field Status metadata to Online Serving

### LINKED ISSUE #278: Return missing feature metadata in response
There is currently no simple mechanism for differentiating between missing and null features. A missing feature would be one where the feature creation pipeline for the corresponding timestamp has not been run yet, thus feature retrieval yields no results. A null feature would be one where the pipeline has been executed, but the feature value for that entity is null e.g. event counts for an entity would be 0 if the said entity did not trigger any events within the time window.

It would be useful if Feast returned metadata on features that have not been ingested yet. This could take on the form of a boolean mask that is applied to the resultant dataframe.

mrzzy:
Adds field status metadata to Online Serving so that users can debug Serving returning missing values.

Field status metadata returned (`FieldStatus`)  classifies the missing values into 3 possible cases:
- `NULL_VALUE` - the value is missing because the user did not specify a value on ingestion.
- `NOT_FOUND` - the value is missing because it could not be retrieved from the online store (ie feature data key not found Redis store.)
- `OUTSIDE_MAX_AGE` - the value is missing because it was purposely discarded as the time since the value was ingested and retrieval has exceeded the FeatureSet's max age.

Updates SDKs (Go, Java, Python) to support field status metadata.

Fixes #278 

```release-note
Add Field Status metadata to Online Serving:
* changes to ServingService Protobuf:
  - Added statuses map to FieldValues.
  - Added FieldStatus enum to define Field Status states.
* Updated SDKs to support Field Status Metadata.
```


ches:
Just a very, um, meta thing—if I saw this changelog and #536 in the same release notes I would think they're related. Should we find some disambiguating terminology? Maybe "registry metadata" and "serving metadata"?

ches:
I guess "feature labels" could already suffice unambiguously for what I called "registry metadata". The feature in this PR being called "feature metadata" is problematic though, I think.

woop:
> Just a very, um, meta thing—if I saw this changelog and #536 in the same release notes I would think they're related. Should we find some disambiguating terminology? Maybe "registry metadata" and "serving metadata"?

Yea I agree :) 

Actually the Feast project has a long history of abusing the term metadata. It's just such a convenient word. It's everything `Util`, `Spec`, and `Info` aspire to be, and more.

How about
* Feature status metrics
* Feature status metadata
* Feature value statuses

ches:
> It's everything `Util`, `Spec`, and `Info` aspire to be, and more.

I might steal that.

I think "feature value statuses" is apt.

ches:
Can we introduce this change/functionality in an evolutionary-friendly way? Like introduce it as a new RPC and new messages, with the old ones remaining supported, possibly going through a deprecation period before Feast 1.0?

Breaking the contract is extremely disruptive here. Upgrading the Serving API will break existing online clients in production, they will all have to upgrade their SDK, and it's not possible to roll out a forward-compatible SDK upgrade to fleets in advance. This breaks all the rules of API evolution.

woop:
@mrzzy any objections to `feature value statuses`?

> Breaking the contract is extremely disruptive here. Upgrading the Serving API will break existing online clients in production, they will all have to upgrade their SDK, and it's not possible to roll out a forward-compatible SDK upgrade to fleets in advance. This breaks all the rules of API evolution.

We can. [Some ideas on gRPC versioning from Microsoft](https://docs.microsoft.com/en-us/aspnet/core/grpc/versioning?view=aspnetcore-3.1). 

We could add a new RPC with `V2`?

mrzzy:
I renaming "Feature Metadata" to "Field Status" method instead of "Feature Status" as it also returns entity statuses in addition to feature statuses.

zhilingc:
The golang code has to be versioned too, since version pinning is a non-goal of go modules :/

ches:
This is a good example I think in favor of SDKs providing higher-level API wrapping protobuf types and request/response objects—this Java API was seamlessly upgraded in the PR, and the same could apply to using a new RPC underneath and phasing out the old one:

```java
public List<Row> getOnlineFeatures(List<String> features, List<Row> rows, String defaultProject)
```

while this is very susceptible to source-incompatible breakage, at least if user code touches the (public) underlying `RawResponse`:

```go
func (fc *GrpcClient) GetOnlineFeatures(ctx context.Context, req *OnlineFeaturesRequest) (
	*OnlineFeaturesResponse, error) {
```

zhilingc:
The golang SDK follows a similar pattern, but unfortunately for golang the proto-generated code is persisted (and versioned) together with the sdk code :(

mrzzy:
I'm picking this up again for Feast v0.6.
In light of the API compatibility issue with this PR, I have drafted a [proposal to version Serving Serving](https://docs.google.com/document/d/1VQngwBcx-yWgGpAbsFVdth9GnjL8q-ZgUNBGv57R0Fk/edit#bookmark=id.83v4b664gjf0) to maintain some API Compatibility in the spirit of #687.
@ches @woop @zhilingc

mrzzy:
After discussing with @woop yesterday. I have introduced a [new proposal to introduce the changes without versioning Serving](https://docs.google.com/document/d/1VQngwBcx-yWgGpAbsFVdth9GnjL8q-ZgUNBGv57R0Fk/edit#bookmark=id.yth8okbfghgx):
- Versioning serving would add alot of complexity may not justified for simply adding online metadata.
- Backwards compatibility will still be maintain between v0.5 SDK and v0.6 Feast.

On the SDK side, I still think its valuable to introduce the [Row abstraction for Python/Golang SDKs](https://docs.google.com/document/d/1VQngwBcx-yWgGpAbsFVdth9GnjL8q-ZgUNBGv57R0Fk/edit#bookmark=id.7b18ijf0mesv) as a breaking change from v0.5 -> v0.6.
- Higher level Row API can act as a bastion between Serving Protobuf API changes in the future, without breaking existing clients. As @ches has pointed out.
- Conceptual compatibility between SDKs.

Currently, I am leaning towards this.

woop:
I agree with the proposed approach.

> On the SDK side, I still think its valuable to introduce the Row abstraction for Python/Golang SDKs as a breaking change from v0.5 -> v0.6.

How would this be a breaking change? Do you mean if users implement the new client they would need to restructure their data? Other than that, it would not break existing clients, right?

woop:
Also, would it make sense to fix all of these user side proto dependencies? For example using the Duration proto for the Python SDK is a poor experience.

mrzzy:
> How would this be a breaking change? Do you mean if users implement the new client they would need to restructure their data? Other than that, it would not break existing clients, right?
This would not break existing clients. However it is labeled as a breaking change as v0.5 SDK to v0.6 SDK has action required.
> Also, would it make sense to fix all of these user side proto dependencies? For example using the Duration proto for the Python SDK is a poor experience.
Agreed. We should stick with native equivalents whenever possible.

The Row API proposal seems to have expanded in scope quite abit. I think it should be a seperate PR/issue at this point.

mrzzy:
> How would this be a breaking change? Do you mean if users implement the new client they would need to restructure their data? Other than that, it would not break existing clients, right?

This would not break existing clients. However it is labeled as a breaking change as v0.5 SDK to v0.6 SDK changes to the user API that requires user action. 

> Also, would it make sense to fix all of these user side proto dependencies? For example using the Duration proto for the Python SDK is a poor experience.
Agreed. We should stick with native equivalents whenever possible.

The Row API proposal seems to have expanded in scope quite abit. I think it makes sense to do that overhaul in a separate PR instead. I've create an issue for this #766.

protos/feast/serving/ServingService.proto
```diff
@@ -91,35 +91,58 @@ message GetOnlineFeaturesRequest {
...
+        map<string, feast.types.Value> fields = 1;
+        // Map of feature or entity name to feature/entity statuses/metadata.
+        map<string, FieldStatus> statuses = 2;
+    }
+   
+    enum FieldStatus {
+        // Status is unset for this field.
+        INVALID = 0;
+
+        // Field value is present for this field and within maximum allowable range
```
woop, line 119: This range is the age?

sdk/python/feast/client.py
```diff
@@ -882,6 +893,46 @@ def _build_feature_references(
...
+
+    # collect entity row refs
+    entity_refs = set()
+    for entity_row in entity_rows:
+        entity_refs.update(entity_row.fields.keys())
+
+    # strip the project part the string feature references returned from serving
+    strip_field_values = []
+    for field_value in response.field_values:
+        strip_fields, strip_statuses = {}, {}
```
pyalex, line 917: another ~functional~ pythonic suggestion. highly subjective though. but I was trying to get rid off strip_ref_str & ref_str. it's just too many of them

```suggestion
strip = lambda ref: repr(FeatureRef.from_str(ref, ignore_project=True)) if ref not in entity_refs else ref
for field_value in response.field_values:
    fields_and_statuses = zip(sorted(field_value.fields.items()), 
                              sorted(field_value.statuses.items())
    fields_and_statuses = [(strip(key), (field, status)) for (key, field), (_, status) in fields_and_statuses]
    keys, fields, statuses = zip(*fields_and_statuses)
    GetOnlineFeaturesResponse.FieldValues(
                fields=dict(zip(keys, fields)), statuses=dict(zip(keys, statuses)),
            )
```

sdk/go/client.go
```diff
@@ -50,33 +50,48 @@ func (fc *GrpcClient) GetOnlineFeatures(ctx context.Context, req *OnlineFeatures
...
-			if !isEntity { // is feature ref
-				featureRef, err := parseFeatureRef(refStr, true)
-				if err != nil {
-					return nil, err
+	return r, err
+}
+
+// Wait for and get online feature data from serving.
+// Polls serving for feature data with an exponential backoff if features are not found.
+func (fc *GrpcClient) WaitGetOnlineFeatures(ctx context.Context, req *OnlineFeaturesRequest, waitForSecs float64) (
```
woop, line 68: I am wondering if this shouldn't be outside of Feast. Is there a reason why whatever is wrapping the client couldn't retry?

sdk/go/types.go
```diff
@@ -5,17 +5,15 @@ import (
...
+// Row is a map of fields
 type Row map[string]*types.Value
 
 func (r Row) equalTo(other Row) bool {
-	for k, v := range r {
-		if otherV, ok := other[k]; !ok {
+	for k, field := range r {
+		if otherField, ok := other[k]; !ok {
+			return false
+		} else if !proto.Equal(field, otherField) {
```
woop, line 15: Do we need this else?

sdk/java/src/main/java/com/gojek/feast/BackOffSupplier.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+package com.gojek.feast;
+
+import java.time.Duration;
+import java.time.LocalDateTime;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+/** Backoff Supplier wraps a existing {@link Supplier} with backoff retries. */
+public class BackOffSupplier<T> {
```
woop, line 25: I am having second thoughts about adding all this code into Feast. Can we leave this up to the user? What value does this bring to us?

sdk/go/client.go
mrzzy: No. The point here is to abstract the quirk how Feast works, remove need for users to retry so that they can call one method and just get features, instead having to worry about writing a retry loop to wait for feature data because Feast ingestion is eventually consistent.

sdk/java/src/main/java/com/gojek/feast/BackOffSupplier.java
mrzzy: The value it gives us is that we can greatly clean up the end to end tests, which have copy pasted implementations of poll implementations. Now that Feast SDKs supports waiting internally, we can remove the unsightly polling from end to end tests.

mrzzy: Removed.

sdk/go/client.go
mrzzy: Removed.

protos/feast/serving/ServingService.proto
mrzzy: Yes its max age. I have clarified in the documentation.

sdk/java/src/main/java/com/gojek/feast/FeastClient.java
```diff
@@ -133,35 +145,97 @@ public GetFeastServingInfoResponse getFeastServingInfo() {
...
+                        .build());
+            response = FeastClient.stripProjectsInResponse(entityRefs, response);
+
+            if (waitForSecs == 0.0) {
+              // Waiting for feature data disabled. return response as is.
+              return response;
+            }
+
+            // Check field status of feature data retrieved.
+            for (FieldValues fieldValues : response.getFieldValuesList()) {
```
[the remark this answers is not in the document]
mrzzy: Removed polling, code so this part of the code is now gone.

sdk/go/types.go
mrzzy: Corrected.

sdk/python/feast/client.py
```diff
@@ -882,6 +893,46 @@ def _build_feature_references(
...
+    Strip the project part the string feature references returned in the given response.
+    Args:
+        entity_row: entity_rows passed when making the get features request
+        response: response obtained from a get online features request.
+    Returns:
+        The given response with projects stripped from string feature references
+    """
+
+    # collect entity row refs
+    entity_refs = set()
```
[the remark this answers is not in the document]
mrzzy: Corrected.

sdk/python/feast/client.py
mrzzy: Thanks for this. I implement a version of suggestion without a `sorted()` call.

sdk/python/feast/client.py
```diff
@@ -620,6 +620,8 @@ def get_online_features(
         feature_refs: List[str],
         entity_rows: List[GetOnlineFeaturesRequest.EntityRow],
         project: Optional[str] = None,
+        omit_entities: bool = False,
+        include_meta: bool = False,
```
woop, line 624: You missed the docstring.

serving/src/main/java/feast/serving/service/OnlineServingService.java
```diff
@@ -64,110 +62,211 @@ public GetFeastServingInfoResponse getFeastServingInfo(
...
-          retriever.getOnlineFeatures(entityRows, featureSetRequests);
-      if (scope != null) {
-        scope.span().log(ImmutableMap.of("event", "featureRows", "value", featureRows));
+      // Collect the feature/entity value for each entity row in entityValueMap
+      Map<EntityRow, Map<String, Value>> entityValuesMap =
+          entityRows.stream().collect(Collectors.toMap(row -> row, row -> new HashMap<>()));
+      // Collect the feature/entity status metadata for each entity row in entityValueMap
+      Map<EntityRow, Map<String, FieldStatus>> entityStatusesMap =
+          entityRows.stream().collect(Collectors.toMap(row -> row, row -> new HashMap<>()));
+
```
woop, line 72: What happened to the trace?

serving/src/main/java/feast/serving/service/OnlineServingService.java
```diff
@@ -64,110 +62,211 @@ public GetFeastServingInfoResponse getFeastServingInfo(
...
+        // Check that feature row returned corresponds to a given entity row.
+        if (featureRows.size() != entityRows.size()) {
+          throw Status.INTERNAL
+              .withDescription(
+                  "The no. of FeatureRow obtained from OnlineRetriever"
+                      + "does not match no. of entityRow passed.")
+              .asRuntimeException();
         }
+
+        for (var i = 0; i < entityRows.size(); i++) {
```
woop, line 99: Can we use a more modern means of iteration for readability?

serving/src/main/java/feast/serving/service/OnlineServingService.java
```diff
@@ -64,110 +62,211 @@ public GetFeastServingInfoResponse getFeastServingInfo(
...
+    return valueMap;
   }
 
-  @Override
-  public GetJobResponse getJob(GetJobRequest getJobRequest) {
-    throw Status.UNIMPLEMENTED.withDescription("Method not implemented").asRuntimeException();
+  /**
+   * Generate Field level Status metadata for the given valueMap.
+   *
+   * @param isNotFound whether the given valueMap represents values that were not found in the
```
woop, line 187: Why does the method have awareness of what is calling it? Wouldn't it make more sense for the caller to provide the status?

serving/src/main/java/feast/serving/service/OnlineServingService.java
```diff
@@ -64,110 +62,211 @@ public GetFeastServingInfoResponse getFeastServingInfo(
...
-  public GetBatchFeaturesResponse getBatchFeatures(GetBatchFeaturesRequest getFeaturesRequest) {
-    throw Status.UNIMPLEMENTED.withDescription("Method not implemented").asRuntimeException();
+    return valueMap;
   }
 
-  @Override
-  public GetJobResponse getJob(GetJobRequest getJobRequest) {
-    throw Status.UNIMPLEMENTED.withDescription("Method not implemented").asRuntimeException();
+  /**
+   * Generate Field level Status metadata for the given valueMap.
```
woop, line 185: is `valueMap` defined?

serving/src/main/java/feast/serving/service/OnlineServingService.java
```diff
@@ -64,110 +62,211 @@ public GetFeastServingInfoResponse getFeastServingInfo(
...
-      return getOnlineFeaturesResponseBuilder.addAllFieldValues(fieldValues).build();
+      return GetOnlineFeaturesResponse.newBuilder().addAllFieldValues(fieldValuesList).build();
     }
   }
 
-  private void populateStaleKeyCountMetrics(String project, FeatureReference ref) {
-    Metrics.staleKeyCount.labels(project, ref.getName()).inc();
-  }
+  /**
+   * Unpack feature values using data from the given feature row for features specified in the given
```
woop, line 137: What does `unpack` mean?

serving/src/main/java/feast/serving/service/OnlineServingService.java
```diff
@@ -64,110 +62,211 @@ public GetFeastServingInfoResponse getFeastServingInfo(
...
+   * feature set request.
+   *
+   * @param featureRow to unpack for feature values.
+   * @param featureSetRequest for which the feature row was retrieved.
+   * @param isStaleValues whether which the feature row contains stale values.
+   * @return valueMap mapping string feature reference to feature valuyesaq
```
woop, line 143: For a single entity?

serving/src/main/java/feast/serving/service/OnlineServingService.java
```diff
@@ -64,110 +62,211 @@ public GetFeastServingInfoResponse getFeastServingInfo(
...
+              .asRuntimeException();
         }
+
+        for (var i = 0; i < entityRows.size(); i++) {
+          FeatureRow featureRow = featureRows.get(i);
+          EntityRow entityRow = entityRows.get(i);
+          // Unpack feature field values and merge into entityValueMap
+          boolean isStaleValues = this.isStale(featureSetRequest, entityRow, featureRow);
+          Map<String, Value> valueMap =
+              this.unpackValueMap(featureRow, featureSetRequest, isStaleValues);
```
woop, line 105: Can this be a static method?

serving/src/main/java/feast/serving/service/OnlineServingService.java
```diff
@@ -64,110 +62,211 @@ public GetFeastServingInfoResponse getFeastServingInfo(
...
+                    return FieldStatus.OUTSIDE_MAX_AGE;
+                  } else if (fieldValue.getValCase().equals(Value.ValCase.VAL_NOT_SET)) {
+                    return FieldStatus.NULL_VALUE;
+                  }
+                  return FieldStatus.PRESENT;
+                }));
   }
 
+  /**
+   * Determine if the feature data in the given feature row is considered stale. Data is considered
```
woop, line 212: We should really stop using the term stale because it is too overloaded. `isOutsideMaxAge`?

serving/src/main/java/feast/serving/service/OnlineServingService.java
```diff
@@ -64,110 +62,211 @@ public GetFeastServingInfoResponse getFeastServingInfo(
...
-  private void populateStaleKeyCountMetrics(String project, FeatureReference ref) {
-    Metrics.staleKeyCount.labels(project, ref.getName()).inc();
-  }
+  /**
+   * Unpack feature values using data from the given feature row for features specified in the given
+   * feature set request.
+   *
+   * @param featureRow to unpack for feature values.
+   * @param featureSetRequest for which the feature row was retrieved.
+   * @param isStaleValues whether which the feature row contains stale values.
```
woop, line 142: Why do we pass this as an argument here? Seems like a code smell if the unpack needs to know something about staleness.

serving/src/main/java/feast/serving/service/OnlineServingService.java
```diff
@@ -64,110 +62,211 @@ public GetFeastServingInfoResponse getFeastServingInfo(
...
+                  Collectors.toMap(
+                      featureRowField -> {
+                        FeatureReference featureRef = nameRefMap.get(featureRowField.getName());
+                        return RefUtil.generateFeatureStringRef(featureRef);
+                      },
+                      featureRowField -> {
+                        // drop stale feature values.
+                        if (isStaleValues) {
+                          return Value.newBuilder().build();
+                        } else {
```
woop, line 166: You dont need an else if you return.

serving/src/main/java/feast/serving/service/OnlineServingService.java
mrzzy: Updated.

serving/src/main/java/feast/serving/service/OnlineServingService.java
mrzzy: Yes. Updated.

serving/src/main/java/feast/serving/service/OnlineServingService.java
mrzzy: For a single feature set request. Updated docs.

serving/src/main/java/feast/serving/service/OnlineServingService.java
mrzzy: `valueMap` is passed as the first argument of this method. I missed out its `@param` in the documentation.

serving/src/main/java/feast/serving/service/OnlineServingService.java
mrzzy: Changed.

serving/src/main/java/feast/serving/service/OnlineServingService.java
mrzzy: Readded.

sdk/python/feast/client.py
mrzzy: Corrected.

serving/src/main/java/feast/serving/service/OnlineServingService.java
mrzzy: Corrected.

PR #502, opened 2020-02-27
Add Support for Redis Cluster

lavkesh:
This pull request adds REDIS_CLUSTER as a store. 

Fixes #478

PR #659, opened 2020-04-28
Fix redis cluster e2e

terryyylim:
Fixes end to end test scripts to use config introduced by #502 

```release-note
NONE
```


woop:
Need to change `.prow/scripts/download-maven-cache.sh` to `infra/scripts/download-maven-cache.sh`

Issue #652, opened 2020-04-26
Docker Compose setup is broken on master


ches:
## Expected Behavior

Following https://docs.feast.dev/installation/docker-compose works successfully.

## Current Behavior

Spring configurations fail to wire, e.g.

```
[...]

PR #661, opened 2020-04-29
Docker compose bug fix 

woop:

* ~Change docker compose file to use host based. This avoids having an inconsistency between default configuration and docker compose configuration.~
* Fix bug with application.yml typo
* Fix port configuration bug for postgres

Hopefully fixes all bugs in #652, but more changes are required to close the issue

```release-note
None
```

### DIFF
--- a/infra/docker-compose/docker-compose.yml
+++ b/infra/docker-compose/docker-compose.yml
@@ -19,7 +19,7 @@ services:
       - java
       - -jar
       - /opt/feast/feast-core.jar
-      - --spring.config.location=classpath:/application.yml,file:/etc/feast/application.yaml
+      - --spring.config.location=classpath:/application.yml,file:/etc/feast/application.yml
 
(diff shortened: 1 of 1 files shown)


khorshuheng:
There is one concern though - host network mode only works on Linux host machine. Mac users will not be able to use this.

woop:
> There is one concern though - host network mode only works on Linux host machine. Mac users will not be able to use this.

Damnit I keep forgetting about that limitation. Isn't there a workaround that doesnt involve using private networks?

khorshuheng:
The workaround i used personally, is to do port mapping on 9092 for Kafka. If i want to test using local jupyter notebook, i edit my etc hosts such that kafka:9092 points to localhost:9092 (so that it matches the advertised listener host). I do agree this is not a pretty workaround, though.

woop:
Removed host networking.

PR #668, opened 2020-05-01
Clean up Docker Compose and add test

### LINKED ISSUE #652: Docker Compose setup is broken on master

woop:

* Creates a GitHub Actions test for testing the current `Docker Compose` configuration. This **won't** test the latest commit's code (on PR or push). It will test based on what is configured on that commit. Currently master  is set to use the `latest` docker image for the docker-compose setup, with the SDK using `v0.4.7`.
* Modularizes the docker compose files into three separate files. One for the Core and dependencies, one for online, and one for the batch/historical store. Otherwise users will experience a failure on startup since jobs won't start if Feast Batch Serving registers itself without Feast Core having a GCP service account.
* Uses a minimal notebook that is faster to download

Fixes #652

```release-note
NONE
```

**Problems with the current setup**
A new release has to be made to backport the tutorials from master into 0.4. That would allow us to clean the tutorials up a bit to match the Docker Compose. The current tutorial has unnecessary steps like reinstalling the SDK.


infra/docker-compose/docker-compose.batch.yml
```diff
@@ -0,0 +1,25 @@
+version: "3.7"
+
+services:
+  batch-serving:
+    image: ${FEAST_SERVING_IMAGE}:${FEAST_VERSION}
+    volumes:
+      - ./serving/${FEAST_BATCH_SERVING_CONFIG}:/etc/feast/application.yml
+      - ./serving/${FEAST_BATCH_STORE_CONFIG}:/etc/feast/store.yml
```
zhilingc, line 8: I think this is not necessary anymore after the config refactor

infra/docker-compose/docker-compose.online.yml
```diff
@@ -0,0 +1,19 @@
+version: "3.7"
+
+services:
+  online-serving:
+    image: ${FEAST_SERVING_IMAGE}:${FEAST_VERSION}
+    volumes:
+      - ./serving/${FEAST_ONLINE_SERVING_CONFIG}:/etc/feast/application.yml
+      - ./serving/${FEAST_ONLINE_STORE_CONFIG}:/etc/feast/store.yml
```
zhilingc, line 8: Same for this

woop:
> Does this require any updates to the documentation?

Yes it does, but I will update that through `gitbook` since it is more convenient.

infra/docker-compose/docker-compose.batch.yml
woop: We don't run `master` on Docker Compose, so the configuration refactor shouldn't affect things here.

infra/scripts/test-docker-compose.sh
```diff
@@ -0,0 +1,33 @@
...
+# Get Docker Container IP address
+export JUPYTER_DOCKER_CONTAINER_IP_ADDRESS=$(docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' feast_jupyter_1)
+
+
+# Print Jupyter container information
+docker inspect feast_jupyter_1
+docker logs feast_jupyter_1
+
+# Wait for Jupyter Notebook Container to come online
+timeout 5m bash -c 'until [ $(curl -s ${JUPYTER_DOCKER_CONTAINER_IP_ADDRESS}:8888/tree | grep notebook | wc -l) -gt 0 ]; do echo waiting for Jupyter to start...; echo; sleep 5; done'
```
woop, line 30: @khorshuheng @davidheryanto must say that I quite like `timeout` as opposed to sleep. As long as you can create a nice little logical test. We can replace sleeps with this maybe.

infra/docker-compose/docker-compose.batch.yml
zhilingc: ooh okay. got it

woop: Thanks for taking a look though :)

Going to resolve these...

infra/scripts/test-docker-compose.sh
mrzzy: Maybe we can replace this with [`wait-for-it.sh`](https://github.com/vishnubob/wait-for-it)? So that we don't have to nest a while loop here. 
Its a pure bash script that blocks until a specific port comes alive, so in this case we can wait for `8888` to come up:
```
./wait-for-it.sh ${JUPYTER_DOCKER_CONTAINER_IP_ADDRESS}:8888 --timeout=300 
```
Although I'm not sure if the explicit check for the notebook  
(`curl -s ${JUPYTER_DOCKER_CONTAINER_IP_ADDRESS}:8888/tree | grep notebook | wc -l) -gt 0 `)  
is necessary to after `8888` comes up.

mrzzy:
>Modularizes the docker compose files into three separate files. One for the Core and dependencies, one for online, and one for the batch/historical store. 

I am in favor of a singular docker-compose file if possible in the future. Maybe we could ensure that all the dependencies is loaded before exposing its gRPC port? Then we can synchronize the order that Core/Dependencies and Serving has to started up in the future with [wait-for-it](https://github.com/vishnubob/wait-for-it) automatically.

>Otherwise users will experience a failure on startup since jobs won't start if Feast Batch Serving registers itself without Feast Core having a GCP service account.

Perhaps we should make this this failure explicit by making Batch Serving to refuse to run if Feast core does not have the required service account. Seems like otherwise this could be a silent error that can catch new users by surprise.

woop:
> I am in favor of a singular docker-compose file if possible in the future.

So am I. Right now I am just trying to make the startup process explicit. It's still a one-liner so I dont see it being too painful.

> Then we can synchronize the order that Core/Dependencies and Serving has to started up in the future with wait-for-it automatically.

Makes sense. Quite like the `wait-for-it` script!

> Perhaps we should make this this failure explicit by making Batch Serving to refuse to run if Feast core does not have the required service account. 

Yes, ideally that would be the case. The system shouldnt load if the configuration isn't set up correctly.

However, this would require a moderate amount of code changes because job management only detects the problem when it spins up new jobs after feature set registration. The fact that Feast Core needs the service account is the bigger problem. Which is why we want to move job management to Serving so that failure can be immediate and localized. I am not in favour of adding more complexity between these two services until we have done that.

infra/scripts/test-docker-compose.sh
woop: Oooh, this is nice!

infra/docker-compose/docker-compose.online.yml
```diff
@@ -0,0 +1,19 @@
...
+    depends_on:
+      - core
+      - redis
```
mrzzy, line 11: This `depends_on` dependency should be removed as `core`  and `redis` are in separate docker-compose file.
`depends_on` only works for services on the same docker-compose file.

woop: Are you sure about this? Doesn't compose merge the files first?

mrzzy: I see, if we are going for this `docker-compose` command, this would not be a problem.
`docker-compose -f docker-compose.yml -f docker-compose.online.yml`

infra/docker-compose/jupyter/startup.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -ex
+
+# Clone Feast repository into Jupyter container
+git clone -b ${FEAST_REPOSITORY_VERSION} --single-branch https://github.com/gojek/feast.git
```
mrzzy, line 6: This will fail on multiple `docker-compose up` as Feast repo already exists.
We should add a test for the `feast` directory an not clone feast if it already exists.

woop: Do you mean if they reconfigure `jupyter` and then want to restart it? Running `docker-compose up` twice doesnt restart the container. To be safe we can add `|| true` to ensure that the script doesnt fail here

mrzzy: The workflow that would fail is:
- `docker-compose up` - when they first try out Feast's docker-compose setup.
- `docker-compose down` - after they are trying out.
- `docker-compose up` - when they to use it again. The failure happens here.

woop: Good point.

woop: Actually, I can't produce this error using those commands. It clones successfully every time.

mrzzy: The exact commands i used to create the error:
- `docker-compose -f docker-compose.yml -f docker-compose.online.yml up`
- Ctrl-C to bring down docker-compose stack
- `docker-compose -f docker-compose.yml -f docker-compose.online.yml up`

This is the error I got:
```
jupyter_1         | + git clone -b v0.4.7 --single-branch https://github.com/gojek/feast.git
jupyter_1         | fatal: destination path 'feast' already exists and is not an empty directory.
feast_jupyter_1 exited with code 128
```

woop: Ok, "works" on my side. Added the change.

PR #675, opened 2020-05-04
Add end-to-end Dataflow test

terryyylim:
Currently the tests are using DirectRunner. The goal of this PR is to
1. Stage Feast and its infrastructure inside the kf-feast project
2. Start an e2e (bq-batch-retrieval-dataflow.py) with Dataflow runner for the new PR's commits
3. Tear down the infrastructure

Triggering of this test will happen through Prow, and should allow multiple of such tests to run concurrently.

```release-note
NONE
```


.prow/config.yaml
```diff
@@ -225,6 +225,48 @@ presubmits:
     branches:
     - ^v0\.(3|4)-branch$
 
+    - name: test-end-to-end-batch-dataflow
+      decorate: true
+      always_run: true
```
woop, line 230: `always_run` should probably be off by default. Otherwise the service account can be printed out by any random person.

infra/scripts/test-end-to-end-batch-dataflow.sh
```diff
@@ -0,0 +1,269 @@
+#!/usr/bin/env bash
+echo "Preparing environment variables..."
+
+set -e
+set -o pipefail
+
+test -z ${GOOGLE_APPLICATION_CREDENTIALS} && GOOGLE_APPLICATION_CREDENTIALS="/etc/service-account/service-account-df.json"
+test -z ${GCLOUD_PROJECT} && GCLOUD_PROJECT="kf-feast"
+test -z ${GCLOUD_REGION} && GCLOUD_REGION="asia-east1"
```
zhilingc, line 9: kf-feast uses us-central1

infra/scripts/test-end-to-end-batch-dataflow.sh
```diff
@@ -0,0 +1,269 @@
...
+Helm install Feast and its dependencies
+============================================================
+"
+cd $ORIGINAL_DIR/infra/scripts/test-templates
+envsubst $'$TEMP_BUCKET $DATASET_NAME $GCLOUD_PROJECT $GCLOUD_NETWORK \
+  $GCLOUD_SUBNET $GCLOUD_REGION $feast_kafka_1_ip
+  $feast_kafka_2_ip $feast_kafka_3_ip $feast_redis_ip $feast_statsd_ip' < values-end-to-end-batch-dataflow.yaml > $ORIGINAL_DIR/infra/charts/feast/values-end-to-end-batch-dataflow-updated.yaml
+
+cd $ORIGINAL_DIR/infra/charts/feast
+helm install --wait --timeout 600s --values="values-end-to-end-batch-dataflow-updated.yaml" ${HELM_RELEASE_NAME} .
```
zhilingc, line 192: Would be great if you could add a helm template command before this for visibility!

tests/e2e/bq-batch-retrieval-dataflow.py
```diff
@@ -0,0 +1,374 @@
+import random
```
zhilingc, line 1: How is this file different from the regular batch retrieval test file?

terryyylim: https://github.com/gojek/feast/pull/675/files/7a35f2ab27cd54f44af557cde28b7135b55d14af#diff-26a2640f3c2a41f5f484d457155d99b9R367

It runs a new test to generate the Dataflow Jobs IDs in a text file for deletion after shutting down Feast infra.

Also, I've increased ingestion timeout for the first 2 tests to 480 and 360 respectively.

infra/scripts/test-end-to-end-batch-dataflow.sh
terryyylim: Fixed

infra/scripts/test-end-to-end-batch-dataflow.sh
terryyylim: Added! Let me know if its better now.

tests/e2e/bq-batch-retrieval-dataflow.py
zhilingc: I wonder if its not possible to use the same file with a flag for type of runner, then maybe this test can run just a smaller subset of the original suite of tests? Since it's intention is not really to test batch retrieval behavior, jsut to ensure that ingestion works on dataflow.

woop: Yea I agree. We actually want to have a test suite that is parameterized across runners, sources, and stores. So having to maintain only a single test suite would be ideal. That is out of scope for this PR, but at the same time we should try and avoid adding more tests here that would be harder to remove in the future.

Also, our tests right now are 90% staging a system, and 10% actually testing. Standing up and tearing down the system should be separated from test execution ideally. Perhaps we can start with this PR?

terryyylim: Utilising pytest markers to group tests before running them.

infra/scripts/test-end-to-end-batch-dataflow.sh
```diff
@@ -0,0 +1,272 @@
...
+Cleaning up
+============================================================
+"
+cd $ORIGINAL_DIR/tests/e2e
+
+# Remove BQ Dataset
+bq rm -r -f ${GCLOUD_PROJECT}:${DATASET_NAME}
+
+# Clean up GKE cluster
+kubectl delete services --all
```
woop, line 255: Which namespace is this limited to? How do you know you are deleting the right services?

woop, line 255: Arent these services deleted by the cluster deletion (same for helm)?

tests/e2e/bq-batch-retrieval.py
```diff
@@ -359,4 +372,15 @@ def test_no_max_age(client):
...
+    assert output["entity_id"].to_list() == output["feature_value8"].to_list()
+
+
+@pytest.mark.dataflow_runner
+def test_write_terminate_jobs(client):
+    ingest_jobs = client.list_ingest_jobs()
+    ingest_jobs = [client.list_ingest_jobs(job.id)[0].external_id for job in ingest_jobs if job.status == IngestionJobStatus.RUNNING]
+
+    cwd = os.getcwd()
+    with open(f"{cwd}/ingesting_jobs.txt", "w+") as output:
```
woop, line 384: This isn't really a test right?

How are you going to ensure that this code always runs at the end? I think it might be safer to use https://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code

You want to make sure that the ingestion_jobs.txt is not created until right at the end.

infra/scripts/test-end-to-end-batch-dataflow.sh
```diff
@@ -0,0 +1,272 @@
...
+gcloud auth activate-service-account --key-file ${GOOGLE_APPLICATION_CREDENTIALS}
+
+gcloud config set project ${GCLOUD_PROJECT}
+gcloud config set compute/region ${GCLOUD_REGION}
+gcloud config list
+
+echo "
+============================================================
+Build docker image based on current commit SHA
+============================================================
```
woop, line 61: This should be a separate workflow. Publishing images shouldnt be a part of a test, unless you are naming it something different and cleaning it up afterwards. I vote for externalizing this.

infra/scripts/test-end-to-end-batch-dataflow.sh
terryyylim: Have removed redundant command.

tests/e2e/bq-batch-retrieval.py
terryyylim: Updated test with fixture for tearing down infra. It would only run for dataflow_runner marker.

infra/scripts/test-end-to-end-batch-dataflow.sh
terryyylim: Shifted docker image creation to use prow's presubmit job.

PR #676, opened 2020-05-04
Add support for feature set updates and remove versions

### LINKED ISSUE #386: Feature set versioning
This issue is created to compile the feedbacks from users regarding Feature set versioning in Feast.

   **Feedback**
- Some users do not care about feature versioning at all and would like to use latest feature set all the time. The current workaround is to make an API call to retrieve the latest version of a feature set, then use it to construct the correct feature id string.
- Bumping up feature set version might be redundant in the case of backward compatible changes, as neither the ingestion and retrieval workflow should be affected.

**Potential solution and associated complexity**
- We can make feature set version optional for feature id. In the case where feature set version is not specified, we should use the latest version. However, this begs the question on whether this should be handled on the SDK side or the server side. Either way, the parsing of the feature id string will become significantly more complicated.
- We can scrap feature set versioning, and in return, does not allow any changes in Feature Set that is not backward incompatible. For example, there should not be any change in data type nor column drops. A column can be marked as deprecated, however, and will be filtered out during ingestion and retrieval. The risk of removing feature set versioning, is that there might be legitimate use case for versioning, such as when multiple teams are sharing feature set and not all the team might want to use the latest feature set.

Comments from more Feast users will be useful to determine which direction should we pursue for future version of Feast. In particular, it will be good to know if how are the existing users utilizing the feature set versioning.

zhilingc:
Addresses #386, and implemented in accordance with this [RFC](https://docs.google.com/document/d/1P44LHd724JloQtpn10naAg5MlzcrkeOlN3QzAn9O8uY).

For backward compatibility, user facing code still allows users to provide versions, but they will be ignored.

This PR is built on top of #655.

Fixes #386

```release-note
Elements in the DB will have to be migrated to only retain the latest feature sets.
```


ingestion/src/main/java/feast/ingestion/transform/fn/ValidateFeatureRowDoFn.java
```diff
@@ -58,7 +58,9 @@ public static Builder newBuilder() {
   public void processElement(ProcessContext context) {
     String error = null;
     FeatureRow featureRow = context.element();
-    FeatureSet featureSet = getFeatureSets().getOrDefault(featureRow.getFeatureSet(), null);
+    featureRow =
+        featureRow.toBuilder().setFeatureSet(stripVersion(featureRow.getFeatureSet())).build();
+    FeatureSet featureSet = getFeatureSets().get(featureRow.getFeatureSet());
     List<FieldProto.Field> fields = new ArrayList<>();
     if (featureSet != null) {
       for (FieldProto.Field field : featureRow.getFieldsList()) {
```
woop, line 66: Should this for loop have `break` or `continue` if we are tolerant of unknown fields?

protos/feast/core/CoreService.proto
```diff
@@ -163,23 +154,24 @@ message ListStoresResponse {
...
 message ApplyFeatureSetRequest {
-    // Feature set version and source will be ignored
     feast.core.FeatureSet feature_set = 1;
 }
 
 message ApplyFeatureSetResponse {
     enum Status {
-        // Latest feature set version is consistent with provided feature set
+        // Latest feature set is consistent with provided feature set
         NO_CHANGE = 0;
```
woop, line 163: Out of scope for this PR, but should this be `UNKNOWN` with `NO_CHANGE` moved to a higher number?

protos/feast/core/CoreService.proto
```diff
@@ -52,8 +52,11 @@ service CoreService {
     // Create or update and existing feature set.
     //
     // This function is idempotent - it will not create a new feature set if schema does not change.
-    // If an existing feature set is updated, core will advance the version number, which will be
-    // returned in response.
+    // Schema changes will update the feature set if the changes are valid.
+    // All changes except the following are valid:
+    // - Changes to feature set id (name, project)
+    // - Changes to entities
+    // - Changes to feature type
```
woop, line 59: Feature name changes?

ingestion/src/main/java/feast/ingestion/transform/fn/ValidateFeatureRowDoFn.java
```diff
@@ -58,7 +58,9 @@ public static Builder newBuilder() {
   public void processElement(ProcessContext context) {
     String error = null;
     FeatureRow featureRow = context.element();
-    FeatureSet featureSet = getFeatureSets().getOrDefault(featureRow.getFeatureSet(), null);
+    featureRow =
+        featureRow.toBuilder().setFeatureSet(stripVersion(featureRow.getFeatureSet())).build();
```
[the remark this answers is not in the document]
zhilingc: No, versions should be stripped for rows being ingested into the systems because the feature set string is still used as part of the redis key.

protos/feast/core/CoreService.proto
zhilingc: I dont want to break this enum

protos/feast/core/CoreService.proto
zhilingc: Moot, because a feature with a different name is a different feature

core/src/test/java/feast/core/service/SpecServiceTest.java
```diff
@@ -100,49 +92,37 @@
   private List<Store> stores;
   private Source defaultSource;
 
+  // TODO: Updates update features in place, so if tests follow the wrong order they might break.
```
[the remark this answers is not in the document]
zhilingc: For cloning to be possible the FeatureSet object has to either 
1. Be serializable, which extends to all child objects (Source, Feature, Entity, Project). 
or 
2. Implement a copy method (also extends to all child objects) 

which i felt was redundant considering in the actual application each instantiation of the FeatureSet object is scoped to a single request anyway, which is why updateFromProto updates the feature set in place.
But if that is not agreeable I will implement the former.

core/src/main/java/feast/core/model/FeatureSet.java
```diff
@@ -192,9 +258,11 @@ public void addFeature(Feature feature) {
 
     List<FeatureSpec> featureSpecs = new ArrayList<>();
     for (Feature featureField : features) {
-      FeatureSpec.Builder featureSpecBuilder = FeatureSpec.newBuilder();
-      setFeatureSpecFields(featureSpecBuilder, featureField);
-      featureSpecs.add(featureSpecBuilder.build());
+      if (!featureField.isArchived()) {
```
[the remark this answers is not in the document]
zhilingc: Sure, but is it too much information for the user?

core/src/main/java/feast/core/model/FeatureSet.java
```diff
@@ -160,6 +150,82 @@ public static FeatureSet fromProto(FeatureSetProto.FeatureSet featureSetProto) {
...
+    }
+
+    Map<String, Entity> existingEntities =
+        entities.stream().collect(Collectors.toMap(entity -> entity.getName(), entity -> entity));
+
+    // 1b. check no change to entities
+    if (spec.getEntitiesList().size() != existingEntities.keySet().size()) {
+      throw new IllegalArgumentException(
+          String.format("Given set of entities do not match size of existing entities."));
+    } else {
```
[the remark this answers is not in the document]
zhilingc: I don't think a size comparison is sufficient?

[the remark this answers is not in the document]
zhilingc: Im changing it to a set comparison, but we lose the niceties of telling the user exactly which entity is problematic

ingestion/src/main/java/feast/ingestion/transform/fn/ValidateFeatureRowDoFn.java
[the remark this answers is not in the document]
woop: Ok, so wouldnt his proposal work? It happens in a prior stage.

protos/feast/core/CoreService.proto
woop: We know that, but do our users? If you have a `Feature()` and you change the name and apply, you might expect it to rename the feature.

core/src/main/java/feast/core/model/FeatureSet.java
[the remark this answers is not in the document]
woop: I was too cryptic, sorry. I mean you could have
```
if(predicate){
exception
}
rest...
```
If the exception triggers then the rest of the method doesnt matter.

ingestion/src/main/java/feast/ingestion/transform/fn/ValidateFeatureRowDoFn.java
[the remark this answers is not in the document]
zhilingc: oh okay. I see. Sure, why not

protos/feast/core/CoreService.proto
woop: Fair enough, but should this be tracked in an issue so that it can be fixed over time?

zhilingc: okay

woop: A `TODO` would also suffice.

core/src/test/java/feast/core/service/SpecServiceTest.java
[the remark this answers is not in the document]
woop: Can we do something like create a new feature set and just update that from the original one? Then pass that second feature set to the tests. I admit I havent spent much brain power on this `TODO` so I trust your judgement here.

core/src/main/java/feast/core/model/FeatureSet.java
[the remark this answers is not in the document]
woop: I mean if you do something like

`my_feature_set.list_features()` it shouldn't print archived features, but if you do something like `my_feature_set.list_features(show_archived=True)` it should.

We dont have to implement the client side functionality, but it might make sense to still add the archived flag to the message so that its simply a matter of exposing it.

core/src/test/java/feast/core/service/SpecServiceTest.java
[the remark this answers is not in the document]
zhilingc: The update is done by the code though, on the object returned by the mock, not the object passed to the method

ingestion/src/main/java/feast/ingestion/transform/fn/ValidateFeatureRowDoFn.java
zhilingc: Out of scope for this PR.

woop: Ok, but it seems like a problem because it is highlighted in the [Removal of Versions RFC](https://docs.google.com/document/d/1P44LHd724JloQtpn10naAg5MlzcrkeOlN3QzAn9O8uY/edit)?

 `Rows that are short of any fields, or have additional fields, will still be written to the stores`

Should we have an additional PR?

core/src/main/java/feast/core/model/FeatureSet.java
```diff
@@ -160,6 +152,65 @@ public static FeatureSet fromProto(FeatureSetProto.FeatureSet featureSetProto) {
...
+    status = FeatureSetStatus.STATUS_PENDING;
+    // 4. Update max age and source.
+    maxAgeSeconds = spec.getMaxAge().getSeconds();
+    source = Source.fromProto(spec.getSource());
+
+    Map<String, FeatureSpec> updatedFeatures =
+        spec.getFeaturesList().stream().collect(Collectors.toMap(FeatureSpec::getName, fs -> fs));
+
+    // 3. Tombstone features that are gone, update features that have changed
+    for (Feature feature : features) {
```
mrzzy, line 196: Its unclear just by looking at the code where `features` came from. Maybe `this.features`?

protos/feast/core/CoreService.proto
zhilingc: Ok, updated.

core/src/main/java/feast/core/model/FeatureSet.java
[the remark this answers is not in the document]
zhilingc: I think exposing archived features isn't that trivial, and should be saved for a subsequent PR. I've updated the exception thrown so that users know when they're applying a feature that was previously archived.

core/src/main/java/feast/core/model/FeatureSet.java
```diff
@@ -160,6 +151,65 @@ public static FeatureSet fromProto(FeatureSetProto.FeatureSet featureSetProto) {
...
+
+    // 1b. check no change to entities
+    if (!Sets.newHashSet(spec.getEntitiesList()).equals(existingEntities)) {
+      throw new IllegalArgumentException(
+          String.format(
+              "You are attempting to change the entities of this feature set: Given set of entities \n{%s}\n does not match existing set of entities\n {%s}. This isn't allowed. Please create a new feature set. ",
+              spec.getEntitiesList(), existingEntities));
+    }
+
+    status = FeatureSetStatus.STATUS_PENDING;
```
mrzzy, line 186: Is this a duplicate of `SpecService.java:281`?
```java
     featureSet.setStatus(FeatureSetStatus.STATUS_PENDING); 
```

core/src/main/java/feast/core/service/JobCoordinatorService.java
```diff
@@ -99,16 +97,11 @@ public void Poll() throws InvalidProtocolBufferException {
...
-                        .setFeatureSetVersion(subscription.getVersion())
-                        .setProject(subscription.getProject())
-                        .build())
-                .getFeatureSetsList();
-        featureSets.addAll(featureSetsFromProto(featureSetSpecs));
+        List<FeatureSet> featureSetsForSub =
+            featureSetRepository.findAllByNameLikeAndProject_NameLikeOrderByNameAsc(
+                subscription.getName().replace('*', '%'),
+                subscription.getProject().replace('*', '%'));
+        featureSets.addAll(featureSetsForSub);
```
mrzzy, line 104: Not sure of the benefits of bypassing `SpecService` here. 
If we centralize listing of FeatureSets through `ListFeatureSets`,  changes to `listFeatureSets()` can propagate upstream (ie `Subscriptions`, `JobService`, etc.).

core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -144,82 +123,54 @@ public GetFeatureSetResponse getFeatureSet(GetFeatureSetRequest request)
...
     }
 
     checkValidCharactersAllowAsterisk(name, "featureSetName");
     checkValidCharactersAllowAsterisk(project, "projectName");
 
     List<FeatureSet> featureSets = new ArrayList<FeatureSet>() {};
 
-    if (project.equals("*")) {
+    if (project.contains("*")) {
       // Matching all projects
```
mrzzy, line 145: Misleading comment.

woop:
`Elements in the DB will have to be migrated to only retain the latest feature sets.`

Should we add the note that versions will be ignored hence forth?

core/src/main/java/feast/core/model/FeatureSet.java
zhilingc: Oops. Removed.

core/src/main/java/feast/core/model/FeatureSet.java
```diff
@@ -359,9 +346,4 @@ public boolean equals(Object obj) {
     }
     return this.equalTo(((FeatureSet) obj));
```
ches, line 347: A random individual comment here, because I was just now looking at the labels PR when pulled in for our internal backport, and noticing that we did not update `FeatureSet`'s equality for labels.

I don't have this branch checked out at the moment, but I believe `equalTo` is effectively unused now because it's only usage was in `SpecService` for determining if version should increment—`equalTo` can probably be folded into the `equals` method now.

We can send a quick patch for including labels in `equals`, unless you want to do it here if you go ahead and consolidate the two methods.

ingestion/src/main/java/feast/ingestion/transform/fn/ValidateFeatureRowDoFn.java
zhilingc: oh right. I forgot. I'll add it to this pr, sorry!

protos/feast/core/CoreService.proto
```diff
@@ -102,9 +105,6 @@ message GetFeatureSetRequest {
 
     // Name of feature set (required).
     string name = 1;
-
-    // Version of feature set (optional). If omitted then latest feature set will be returned.
-    int32 version = 2;
```
woop, line 107: Should we add a reserved keyword here?

https://developers.google.com/protocol-buffers/docs/proto#reserved

sdk/python/feast/client.py
```diff
@@ -883,39 +861,27 @@ def _build_feature_references(
...
 
-        if len(project) == 0 or len(name) == 0 or version < 0:
+        if len(project) == 0 or len(name) == 0:
             raise ValueError(
-                f'Could not parse feature ref {feature_ref}, expecting "project/feature:version"'
+                f'Could not parse feature ref {feature_ref}, expecting "project/feature"'
             )
 
-        features.append(FeatureReference(project=project, name=name, version=version))
+        features.append(FeatureReference(project=project, name=name.split(":")[0]))
```
woop, line 884: Would it be better to just allow the name to be sent as is and give the user an exception?

If people are bumping their SDK version then we can encourage them to use the new feature references. I think we should just ensure that old clients are still supported by the serving layer, which you have already done.

woop: cc @mrzzy

ingestion/src/main/java/feast/ingestion/transform/fn/ValidateFeatureRowDoFn.java
zhilingc: But the code already has that behavior?

woop: I'm just eyeballing it, so maybe I am missing something. Wouldn't the `break` statement stop the for loop entirely? Meaning if it finds an unknown field then it will stop processing further (possibly good) fields.

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/writer/BigQueryFeatureSink.java
```diff
@@ -189,6 +195,18 @@ private TableDefinition createBigQueryTableDefinition(FeatureSetProto.FeatureSet
         TimePartitioning.newBuilder(TimePartitioning.Type.DAY).setField("event_timestamp").build();
     log.info("Table partitioning: " + timePartitioning.toString());
 
+    FieldList fieldsList = FieldList.of(Collections.emptyList());
```
davidheryanto, line 198: I think `fieldList` variable here is somehow unused i.e. not used when creating the latest BigQuery `TableDefinition`.

This I believe will result in the existing table schema (if exists) to be ignored?

Update of `FeatureSet` with BigQuery store will fail in the following case?
> Error: Field 'col1' is missing in new schema

Existing schema in BigQuery:
| column name | type |
| ----- | -------|
| name | STRING |
| col1 | INT64 |
| col2 | FLOAT |

New `FeatureSet` to apply:
```yaml
kind: feature_set
spec:
  project: myproject
  maxAge: 3600s
  name: myfeatureset
  entities:
  - name: name
    valueType: STRING
  features:
  - name: col2
    valueType: FLOAT
  - name: col3
    valueType: INT64
```

davidheryanto: Unless this is the intended behaviour. Although `createBigQueryTableDefinition` method has this in the comment:
> ... If a table already exists, existing fields will be retained ...

sdk/python/feast/client.py
mrzzy: In this case, since versions are completely removed from the model and `fs:1` & `fs:2` no longer refer to different feature sets I think we should make it clear cut in SDKs (ie make the breaking chnage) that versions are no longer supported, otherwise we mind run into queries about why updating `fs:2` into `fs:3` broke some part of their code that depended on `fs:1` etc.

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/writer/BigQueryFeatureSink.java
zhilingc: Yep, found this oversight today while implementing the e2e test and fixed it :) nice catch

sdk/python/feast/client.py
zhilingc: I think for the user 0.5 should only phase out versions in a soft way, throwing deprecation warnings if versions are provided, and only start throwing exceptions in subsequent versions.

protos/feast/core/CoreService.proto
zhilingc: donezo

core/src/main/java/feast/core/model/FeatureSet.java
zhilingc: I think i'd love for it to be a patch just so this PR is a little less overwhelming to digest, haha.

ingestion/src/main/java/feast/ingestion/transform/fn/ValidateFeatureRowDoFn.java
zhilingc: ah crap sorry brain farted, you're right, it should be continue

sdk/python/feast/client.py
woop: Is there a specific scenario that the above code would help? It seems like we are already doing this on the server side. The upgrade path we are planning is to bump the server (because clients are more likely to be out of date). Then notify all clients to upgrade to the latest version.

0.4 client to 0.4 server works and 0.4 client to 0.5 server works (ignores version). So it seems like what you are optimizing for is a 0.5 client to a 0.4 server?

ingestion/src/main/java/feast/ingestion/transform/fn/ValidateFeatureRowDoFn.java
woop: Also related to #628 

cc @junhui096

sdk/python/feast/client.py
zhilingc: Fair point, I've removed the code stripping versions from the sdks

zhilingc:
Pending fix for #677 to pass end to end tests

sdk/python/feast/client.py
mrzzy: I've created an issue to discuss this in more detail #687. We should have a consistent policy for API compatibility across all PRs moving forward.

core/src/main/java/feast/core/model/FeatureSet.java
zhilingc: Patch is here: #690 
I'll consolidate the equality methods in this PR after that PR is merged in

tests/e2e/bq-batch-retrieval.py
```diff
@@ -509,17 +511,14 @@ def test_update_featureset_update_featureset_and_ingest_second_subset(
...
     while True:
-        fs = client.get_feature_set(name=update_fs.name, project=update_fs.project)
-        if fs.status == FeatureSetStatus.STATUS_READY:
-            # Give BQ table schema some time to update
-            time.sleep(120)
+        buf_size = get_streaming_buffer_size(client, update_fs)
+        if buf_size == 10:
             break
-        time.sleep(5)
+        time.sleep(15)
```
woop: This is very nit picky, but I am thinking of removing these sleeps over time and adding timeouts/retries. Inevitably the sleeps just keep getting bumped up and up over time.

Should we maybe start incorporating https://github.com/rholder/retrying?

woop: I guess you dont have to be the one to do it, but if you end up touching the sleeps again then I think it would be nice to consider it.

zhilingc: This is a proper retry with a 15s backoff though. I havent touched the other sleeps.

core/src/main/java/feast/core/model/FeatureSet.java
zhilingc: This is done

tests/e2e/bq-batch-retrieval.py
```diff
@@ -392,3 +440,163 @@ def infra_teardown(pytestconfig, core_url, serving_url):
...
+
+def get_rows_ingested(client: Client, feature_set: FeatureSet, ingestion_id: str) -> int:
+    response = client._core_service_stub.ListStores(ListStoresRequest(filter=ListStoresRequest.Filter(name="historical")))
+    bq_config = response.store[0].bigquery_config
+    project = bq_config.project_id
+    dataset = bq_config.dataset_id
+    table = f"{PROJECT_NAME}_{feature_set.name}"
+
+    bq_client = bigquery.Client(project=project)
+    rows = bq_client.query(f'SELECT * FROM `{project}.{dataset}.{table}` WHERE ingestion_id = "{ingestion_id}"').result()
```
woop, line 597: In order to save money, could this be a `count()`?

tests/e2e/bq-batch-retrieval.py
```diff
@@ -392,3 +440,163 @@ def infra_teardown(pytestconfig, core_url, serving_url):
...
+    client.apply(update_fs)
+
+    # We keep retrying this ingestion until all values make it into the buffer.
+    # This is a necessary step because bigquery streaming caches table schemas
+    # and as a result, rows may be lost.
+    while True:
+        ingestion_id = client.ingest(feature_set=update_fs, source=subset_df)
+        time.sleep(15)  # wait for rows to get written to bq
+        rows_ingested = get_rows_ingested(client, update_fs, ingestion_id)
+        if rows_ingested == len(subset_df):
```
woop, line 523: If there is a transient failure, would we ever ingest the same rows twice? Should we dedupe in the select statement below perhaps?

zhilingc: I think its enough of an edge case to ignore it in this query, especially if we want to keep it cheap and simple

protos/feast/core/CoreService.proto
ches: Is it? Not seeing it for this or `feature_set_version` on `ListFeatureSetsRequest`

protos/feast/serving/ServingService.proto
```diff
@@ -69,9 +69,6 @@ message FeatureReference {
     // Feature name
     string name = 2;
 
-    // Feature version
-    int32 version = 3;
```
ches, line 73: `reserved` for this one too?

core/src/main/java/feast/core/job/JobUpdateTask.java
```diff
@@ -101,11 +101,19 @@ public Job call() {
...
-    Set<FeatureSet> existingFeatureSetsPopulatedByJob = Sets.newHashSet(job.getFeatureSets());
-    Set<FeatureSet> newFeatureSetsPopulatedByJob = Sets.newHashSet(featureSets);
+  boolean requiresUpdate(Job job) {
+    // If set of feature sets has changed
+    if (!Sets.newHashSet(featureSets).equals(Sets.newHashSet(job.getFeatureSets()))) {
+      return true;
+    }
 
-    return !newFeatureSetsPopulatedByJob.equals(existingFeatureSetsPopulatedByJob);
+    // If any existing feature set populated by the job has its status as pending
```
ches, line 110: Curious: was this addition a semantic fix independent of versions? More practical side of the question: should it be backported to v0.4?

ches: Related: #708, probably also/instead the backport candidate.

ches: Or rather, #708 obviates my question so there is no backport need.

PR #680, opened 2020-05-06
Add grpc health probe implementation to core

### LINKED ISSUE #514: Kubernetes liveness / readiness probes

Currently we've got, as default [1], for **both liveness and readiness** probes:
* in `feast-core`, an HTTP probe on the `/healthz` endpoint that checks the metadata DB [2]
* in `feast-serving`, a "grpc probe" (using exec) that does not seem to check any store connection, although there is an action item there to do so [3]

Raising this issue to discuss:
- Should we have feast-core liveness / readiness probes also implemented through Grpc? `HealthController` is the only RestController in core, and it seems better to check the status of the Grpc endpoint instead. 
- Should we have database checks in liveness probes? Restarting the pod will have no impact in DB / Store unavailability, and checking external dependencies in a probe can potentially cause a single point of failure (see https://blog.colinbreck.com/kubernetes-liveness-and-readiness-probes-how-to-avoid-shooting-yourself-in-the-foot)

Grpc based probes for both components.
No DB checks in liveness probes by default.

**Additional context**

Refs:
[1] https://github.com/gojek/feast/blob/master/infra/charts/feast/charts/feast-core/templates/deployment.yaml#L107-L129
https://github.com/gojek/feast/blob/master/infra/charts/feast/charts/feast-serving/templates/deployment.yaml#L99-L119
[2] https://github.com/gojek/feast/blob/master/core/src/main/java/feast/core/http/HealthController.java
[3] https://github.com/gojek/feast/blob/master/serving/src/main/java/feast/serving/controller/HealthServiceController.java

zhilingc:

Changing the health probe for feast core to use grpc, and test simple functionality (retrieving projects) instead of just pinging the db.

This works well as a readiness probe, but not so much as a liveness probe since even in the event it fails, restarting the service will not help solve the problem. Additionally, [grpc health probe](https://github.com/grpc-ecosystem/grpc-health-probe/) does not allow us to differentiate between the two types of probes.

For now I have the liveness probe disabled by default in the helm chart, but I'm happy to take any suggestions regarding the problems above ^

Fixes #514

```release-note

```


ches:
We should include in changelog / a migration guide that `/healthz` will be gone in the release that includes this, and point to the gRPC replacement.

ches:
For now, I've called it out as a note in the Breaking Changes section of [the GitHub Release for v0.5.0](https://github.com/feast-dev/feast/releases/tag/v0.5.0)

PR #682, opened 2020-05-07
Add Docker Compose configuration for development purposes

dr3s:
**What this PR does / why we need it**: Simple dev workflow enhancement to allow running core with its dependent services.  In the future, I hope to make this debuggable.
I'm open to changing it the PR as it stands was expecting:

1. make changes (IDE would likely incremental compile)
2. docker-compose up would compile as necessary and run
3. manual testing whatever
4. shut down and repeat

I haven't yet built a whole dev workflow, so there may be an alternative that is better.  One idea from @mrzzy is to keep the code container with just a shell running and use `docker exec` to run `mvn` or other commands.

Eventually I would like a whole debugging setup but this seemed worthwhile as a separate PR.

**Which issue(s) this PR fixes**:  N/A

```release-note
NONE
```

### DIFF
--- a/core/src/main/resources/application.yml
+++ b/core/src/main/resources/application.yml
@@ -73,7 +73,7 @@ feast:
     # See the following for options https://api.docs.feast.dev/grpc/feast.core.pb.html#KafkaSourceConfig
     options:
       topic: feast-features
-      bootstrapServers: localhost:9092
+      bootstrapServers: ${STREAM_HOST:localhost}:${STREAM_PORT:9092}
       replicationFactor: 1
       partitions: 1
 
--- a/core/src/main/resources/application.yml
+++ b/core/src/main/resources/application.yml
@@ -73,7 +73,7 @@ feast:
     # See the following for options https://api.docs.feast.dev/grpc/feast.core.pb.html#KafkaSourceConfig
     options:
       topic: feast-features
-      bootstrapServers: ${STREAM_HOST:localhost}:${STREAM_PORT:9092}
+      bootstrapServers: localhost:9092
       replicationFactor: 1
       partitions: 1
 
--- a/infra/docker-compose/docker-compose.dev.yml
+++ b/infra/docker-compose/docker-compose.dev.yml
@@ -1,0 +1,70 @@
+version: "3.7"
+
+services:
+  core:
+    image: maven:3.6-openjdk-11
+    volumes:
+      - ${HOME}/.m2:/root/.m2:delegated
+      - ../../.:/code:cached
+    environment:
+      DB_HOST: db
+      STREAM_HOST: kafka
+      GOOGLE_APPLICATION_CREDENTIALS: /etc/gcloud/service-accounts/key.json
+    restart: on-failure
+    depends_on:
+      - db
+      - kafka
+    ports:
+      - 6565:6565
+
+    working_dir: /code
+    command:
+      - mvn
+      - -pl
+      - core
+      - spring-boot:run
+
+  jupyter:
+    image: jupyter/minimal-notebook:619e9cc2fc07
+    volumes:
+      - ./gcp-service-accounts/${FEAST_JUPYTER_GCP_SERVICE_ACCOUNT_KEY}:/etc/gcloud/service-accounts/key.json
+      - ./jupyter/startup.sh:/etc/startup.sh
+    depends_on:
+      - core
+    environment:
+      FEAST_CORE_URL: core:6565
+      FEAST_ONLINE_SERVING_URL: online-serving:6566
+      FEAST_BATCH_SERVING_URL: batch-serving:6567
+      GOOGLE_APPLICATION_CREDENTIALS: /etc/gcloud/service-accounts/key.json
+      FEAST_REPOSITORY_VERSION: ${FEAST_REPOSITORY_VERSION}
+    ports:
+      - 8888:8888
+    command: ["/etc/startup.sh"]
+
+  kafka:
+    image: confluentinc/cp-kafka:5.2.1
+    environment:
+      KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
+      KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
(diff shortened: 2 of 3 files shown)


core/src/main/resources/application.yml
```diff
@@ -73,7 +73,7 @@ feast:
     # See the following for options https://api.docs.feast.dev/grpc/feast.core.pb.html#KafkaSourceConfig
     options:
       topic: feast-features
-      bootstrapServers: localhost:9092
+      bootstrapServers: ${STREAM_HOST:localhost}:${STREAM_PORT:9092}
```
woop, line 76: You should be able to set this configuration without introducing new env vars. Have a look at Spring environmental variables. https://stackoverflow.com/a/48893053

dr3s: thanks, updated.

infra/docker-compose/docker-compose.dev.yml
```diff
@@ -0,0 +1,70 @@
+version: "3.7"
+
+services:
+  core:
+    image: maven:3.6-openjdk-11
+    volumes:
+      - ${HOME}/.m2:/root/.m2:delegated
+      - ../../.:/code:cached
```
mrzzy, line 8: Not sure about using `cached` here. According to the [docs](https://docs.docker.com/docker-for-mac/osxfs-caching/#cached):
> For directories mounted with cached, the host’s view of the file system is authoritative; writes performed by containers are immediately visible to the host, but there may be a delay before writes performed on the host are visible within containers.

This can be confusing if I make a change in the code in the host but when I run with `mvn -pl core spring-boot run` the change is not reflected.

infra/docker-compose/docker-compose.dev.yml
```diff
@@ -0,0 +1,70 @@
...
+      - kafka
+    ports:
+      - 6565:6565
+
+    working_dir: /code
+    command:
+      - mvn
+      - -pl
+      - core
+      - spring-boot:run
```
mrzzy, line 25: If the aim is to `docker-compose exec core` and do `mvn -pl core spring-boot:run`
to test changes after you make changes in the core's code then maybe we shouldn't 
run core on `docker-compose up`, as the container will stop as soon as `mvn -pl core spring-boot:run` completes.

`sleep 9999d` can suffice here.

mrzzy:
👍 Thanks for the PR. I can finally retire my hacked together `docker-compose` dev setup to develop Feast Core. It would be nice see to something similar for Feast Serving.

infra/docker-compose/docker-compose.dev.yml
dr3s: indeed, it's something that I thought about but I don't know if it's an issue.  All changes should be reflected on startup but changes from the host while the container is running may be delayed.  I'm not sure the performance is worth it but I'm not sure when it will be a problem.  I only did the bare minimum to get my dev setup.  I'm planning on further changes as I haven't even completed a full build, debug, change, build, cycle yet.  I opened this since I struggled so much to get everything running that I thought it would be worth it and it's low risk.

infra/docker-compose/docker-compose.dev.yml
dr3s: interesting.  I need to try out some approaches but that is good feedback.

dr3s: swapping out the command for
```
    entrypoint: /bin/sh
    stdin_open: true
    tty: true
```
does the trick.  Then you can run:
`docker-compose -f docker-compose.dev.yml exec core mvn -pl core spring-boot:run` or whatever you want.  I can update the PR, if this is what you prefer.  I like it but it definitely needs some docs for first-time users.  I think a bit of reorganization of directories and files could make it easier to grok and use.

infra/docker-compose/docker-compose.dev.yml
dr3s: https://docs.docker.com/docker-for-mac/osxfs-caching/ gives an overview of the issue with fs performance on OSX.  I could also see an argument for reversing the settings (.m2 cached and /code delegated).  The reason I did it this way was I expected most code changes would be on the host and m2 changes during build on the container.  Given your other comment about using `docker-compose exec` to keep the container running, I could see reversing these settings.  I'm not really sure and haven't run this setup before, so open to any advice.

infra/docker-compose/docker-compose.dev.yml
mrzzy: @dr3s I think we have different ideas of how this PR should work. Seems like you where going for a more `mvn compile` in host & `docker-compose up` kind of workflow from your discussion. Could you update the PR description with a ideal flow of how this PR should work?

dr3s: sure I will do that.  I'm open to changing it the PR as it stands was expecting:

1. make changes (IDE would likely incremental compile)
2. `docker-compose up` would compile as necessary and run
3. manual testing whatever

The comment above about leaving the container running was in response to your feedback.  I can go that direction.  I just haven't yet built a whole workflow.  This can wait until then but I thought it valuable as-is for first-time users.

infra/docker-compose/docker-compose.dev.yml
dr3s: With more reading, I think this is the right approach because the performance seems pretty bad.  The way I interpret it is, `delegated` for stuff you can lose, `cached` for faster reads, `consistent` when you can't deal with the delay.  I don't know what the delay is but I prefer this unless there is an objection.  I will resolve this but feel free to unresolve if you think this is not the right approach.

infra/docker-compose/docker-compose.dev.yml
mrzzy: In that case I think this is fine.

woop:
Thanks for this again. I'm wondering if it makes sense to update the [Development Guide](https://github.com/gojek/feast/blob/v0.4-branch/docs/contributing/development-guide.md) to include this flow? That could make it easier for new users to get up to speed.

dr3s:
@woop sure, I'll add the docs if you are ok with the approach.

woop:
This flow isn't typically one that I would use. @khorshuheng would you mind reviewing? You had thoughts on this before.

PR #686, opened 2020-05-08
Add Java code coverage reporting

ches:

I'm interested in tracking coverage data. I thought you might be too.

This adds [the JaCoCo Maven plugin][2], with a module dedicated to reporting aggregate coverage, which is said to be [the most reliable way to achieve that][1]. It is currently under `docs/coverage/java` as a tidy place to tuck it away, but GitBook might barf on a `pom.xml` being in there, not sure, so I'll move it if you suggest someplace else.

We use Codecov privately, it should slurp up JaCoCo data readily if you're interested in setting it up, and/or I can look into adding a GitHub Action with JaCoCo's stock HTML report kept as an artifact.

I don't have convenient access to a public web host at the moment to drop the static browseable report as an example, so here's a screenshot of the root page for where we stand:

Screen Shot 2020-05-09 at 2 00 40 AM

You can generate it yourself on the branch with:

    $ mvn test jacoco:report-aggregate
    $ open docs/coverage/java/target/site/jacoco-aggregate/index.html

```release-note
NONE
```

[1]: https://github.com/jacoco/jacoco/wiki/MavenMultiModule#implementation-options
[2]: https://www.eclemma.org/jacoco/trunk/doc/maven.html

### DIFF
--- a/serving/pom.xml
+++ b/serving/pom.xml
@@ -41,6 +41,11 @@
   <build>
     <plugins>
       <plugin>
+        <groupId>org.jacoco</groupId>
+        <artifactId>jacoco-maven-plugin</artifactId>
+      </plugin>
+
+      <plugin>
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-maven-plugin</artifactId>
         <configuration>
@@ -248,11 +253,9 @@
       <scope>test</scope>
     </dependency>
 
-    <!-- TODO: fix version discrepancy with managed version -->
     <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
-      <version>2.23.0</version>
       <scope>test</scope>
     </dependency>
 
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,3 @@
 docs
+!docs/coverage
 charts
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -16,7 +16,11 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-maven-
       - name: test java
-        run: make test-java
+        run: make test-java-with-coverage
+      - uses: actions/upload-artifact@v2
+        with:
+          name: java-coverage-report
+          path: ${{ github.workspace }}/docs/coverage/java/target/site/jacoco-aggregate/
 
   unit-test-python:
     runs-on: ubuntu-latest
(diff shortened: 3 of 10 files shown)


serving/pom.xml
```diff
@@ -248,11 +253,9 @@
       <scope>test</scope>
     </dependency>
 
-    <!-- TODO: fix version discrepancy with managed version -->
     <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
-      <version>2.23.0</version>
```
ches, line 255: A recent Byte Buddy is needed for some functionality of Mockito with Java 11, which is probably why we already have a 1.9.x version set in root POM dependency management. For some reason this older version of Mockito overriding in a couple of modules breaks even though the new Byte Buddy version should be forced, specifically when JaCoCo instrumentation is running. Letting these use newer Mockito set in dependency management already fixed it. [This seems to be a thing][1].

It appears there is no issue with the newer Mockito on these snowflake modules anymore anyway, so good to drop the special cases.

[1]: https://github.com/mockito/mockito/issues/1483#issuecomment-615477170

woop:
Thanks for this @ches. Coverage is pretty poor. That BQ connector :(

>  don't have convenient access to a public web host at the moment

Anything you add to `dist/` as part of `make html` will get served on `docs.feast.dev`

Relevant code https://github.com/gojek/feast/blob/master/Makefile#L153

I am happy to merge as is. Would be nice to have a report somewhere, but not sure if I want to advertise this badge yet.

ches:
I added simple persistence of the report as a build artifact. They are [not conveniently browseable][1], but it's something at least, I'm happier to have that than not.

I'm more interested in knowing _what_ is covered and what isn't, more than an aggregate percentage number, so I don't much care about a badge anyway. We should also bear in mind this is unit tests only (and some tests with in-memory DBs that I'd consider a level of integration).

The lint-python check is failing on master and not affected by this PR.

[1]: https://github.community/t5/GitHub-Actions/Browsing-artifacts/td-p/38817

ches:
/retest

Even though e2e shouldn't be failing from this PR, I guess having it pass can eventually let Prow do the merge…

ches:
Rebased with master for the Python linting fix.

ches:
/retest
I guess

woop:
We are going back to peak flakiness. Apologies, having people look at this. At least 3 different flaky tests.

PR #690, opened 2020-05-11
Check for labels in FeatureSet equality

zhilingc:

Feature set equality is missing check for labels. This PR adds it.


ches:
I noticed this when we we merged it internally, should have filed/fixed it here—glad you found it!

Issue #674, opened 2020-05-03
Replace string references used in non user facing components in Feast


mrzzy:
## Background

Currently Feast uses string references in 3 areas:
- Feature Set References
- Feature References
- Subscriptions

As a result many implementations of string reference parsing and formatting has cropped up all over Feast's code:
- Feast Serving's [`RefUtil`](https://github.com/gojek/feast/blob/e4f8fe905d3c182416e1b69bf60bfbc0e8348f8a/serving/src/main/java/feast/serving/util/RefUtil.java)
- Feast Core's [`Store`](https://github.com/gojek/feast/blob/7a0ff91839757233d3af957b84096f9e2836386b/core/src/main/java/feast/core/model/Store.java#L127)
- Feast Ingestion's [`SpecUtil`](https://github.com/gojek/feast/blob/7a0ff91839757233d3af957b84096f9e2836386b/ingestion/src/main/java/feast/ingestion/utils/SpecUtil.java#L34)
- Feast Storage Writers: [`Redis`](https://github.com/gojek/feast/blob/7a0ff91839757233d3af957b84096f9e2836386b/storage/connectors/redis/src/main/java/feast/storage/connectors/redis/writer/RedisCustomIO.java#L185) [`RedisCluster`](https://github.com/gojek/feast/blob/7a0ff91839757233d3af957b84096f9e2836386b/storage/connectors/rediscluster/src/main/java/feast/storage/connectors/rediscluster/writer/RedisClusterCustomIO.java#L187).
- Feast Storage Retrievers [`Redis`](https://github.com/gojek/feast/blob/7a0ff91839757233d3af957b84096f9e2836386b/storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/RedisOnlineRetriever.java#L214) [`RedisCluster`](https://github.com/gojek/feast/blob/7a0ff91839757233d3af957b84096f9e2836386b/storage/connectors/rediscluster/src/main/java/feast/storage/connectors/rediscluster/retriever/RedisClusterOnlineRetriever.java#L218)
- etc...

## Problem

Changes to the feature/feature set references/subscriptions are difficult as we have to track down all the different string parsing/formatting implementations. Changes to these areas should not require a comb over the entire codebase to rectify all these parsing and formatting implementations.

Examples  of currently WIP and potential changes that is currently affected by this and why the problem is relevant:
- Feature Set version removal #386 
- Changes to the Feast concept & hierarchy and data model #479 
[...]

PR #693, opened 2020-05-12
Make Projects optional & Update Feature References

### LINKED ISSUE #631: Validate feature reference and project functionality
This card is a follow up on #479, and specifically [this comment ](https://github.com/gojek/feast/issues/479#issuecomment-615544705)

The goal is to add tests and/or the functionality referenced in the above comment.

- <s>Feature references cannot contain a project when it is in string form. This is to prevent multiple projects from being selected in a single feature request (for retrieval).</s>
- Feature References can contain projects to prevent breaking changes to existing clients. An additional field `feature_set_name` is used to reference a feature when a single feature name is ambiguous.
- <s>Projects can only be defined once per incoming request. There should be a single key on each Feast client to set the project, since projects aren't defined in the feature reference any more.</s>
- SDKs to add a field `project` to allow users to override projects for all given feature references. Moving forward this would be the preferred way to set projects, not in feature references, although that is still accepted for backwards compatibility.
- If a project isn't defined then the default project is used. The default project is just called, "default".
- The default project is set on the server side, not the client side. Clients should not have to provide a project.
- Feature names may be duplicated across feature sets in the same project. Currently they have to be unique in a project.
- All tests need to be updated according to the above functionality.
    - Specifically we should make sure that we do not set or create a project by default in any test, unless we are testing project functionality specifically. Projects should be invisible if the user is just using the default project.

mrzzy:
Makes Projects optional by introducing a default project `default`:
- Feast Core automatically creates a non archivable `default` project. 
- `default` project would be used where users do not explicit specify a project in  CoreService's `listFeatureSets(), getFeatureSet(), applyFeatureSet()`.
- Feast Serving resolves Feature References without projects specified to Features in the `default` projects. The `project` field is now optional.

Update Feature References to support duplicate Features names within project but unique within Feature Set:
- Introduces optional `feature_set` field to FeatureReference proto.
- Allows users to specify feature set name by extending string Feature References to the format: `[project/][feature_set:]feature_name` with Feast Serving
- Feast Serving attempts to resolve the feature references with missing `project`, and /or `feature_set_name`. If multiple features match, returns an error to mark the ambiguity.
- Versions in Feature References are ignored by serving if specified.

> When specifying Feature Set name in string feature references to get feature data (ie `feature_set:feature`)
> - Online Serving returns the feature data keyed by the submitted string feature reference.
> - Batch Serving returns the feature data in columns named in the format `feature_set__feature`  

> Serving attempts to maintain backwards compatibility with existing feature references in format: `[project/]feature` 

Update Feast Go, Java, Python SDKs to transition to string Feature References without Projects:
- Remove support for specifying projects in string feature references in SDKs. Specifying project in string Feature References supplied is now an error.
- Add project field to get batch/online features methods to allow users to explicitly specify project.
- String Feature References returned from get online features methods no longer contain project.

Fixes #631 

```release-note
Projects are now optional in Feast APIs (except ingestion).
- Feast creates a "default" project automatically.
- Feast uses the the "default" project where a project is not explicitly specified. 

Feast now namespaces Feature names within its Feature Set.
- Added optional "feature_set_name" field to FeatureReference proto.
- Feast supports FeatureSet specifying Feature Sets Name in string feature references: "driver_fs:driver_feature"
- Feast now allows features to have a unique name within a FeatureSet, but duplicated across a Project

Action Required: Dropping Support for specifying projects in string Feature References in Feast SDKs:
- Projects now cannot be specified in string Feature References passed to get online/batch features methods.
- Use the project parameter in those methods to specify projects instead. 
- String Feature References returned in get online features will not contain projects.
```


core/src/main/java/feast/core/service/AccessManagementService.java
```diff
@@ -60,6 +63,8 @@ public void archiveProject(String name) {
     Optional<Project> project = projectRepository.findById(name);
     if (!project.isPresent()) {
       throw new IllegalArgumentException(String.format("Could not find project: \"%s\"", name));
+    } else if (name.equals(Project.DEFAULT_NAME)) {
+      throw new IllegalArgumentException("Archiving the default project is not allowed.");
```
woop, line 67: Nitpick, but can we use guard clauses? https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html

core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -134,11 +137,12 @@ public GetFeatureSetResponse getFeatureSet(GetFeatureSetRequest request)
 
   /**
    * Return a list of feature sets matching the feature set name, version, and project provided in
-   * the filter. All fields are requried. Use '*' for all three arguments in order to return all
-   * feature sets and versions in all projects.
+   * the filter. Use '*' for all three arguments in order to return all feature sets and versions in
+   * all projects.
```
woop, line 141: Shouldn't this be limited to the default project?

core/src/test/java/feast/core/service/JobServiceTest.java
```diff
@@ -134,14 +134,14 @@ public void setupJobRepository() {
     when(this.jobRepository.findAll()).thenReturn(Arrays.asList(this.job));
   }
 
-  // TODO: setup fake job manager
+  // setup fake job manager
```
woop, line 137: Can we please use Javadoc style comments like this: https://stackoverflow.com/questions/20869421/is-there-a-way-to-write-a-single-line-javadoc-comment-like-in-doxygen?

core/src/test/java/feast/core/service/SpecServiceTest.java
```diff
@@ -808,10 +828,22 @@ public void shouldDoNothingIfNoChange() throws InvalidProtocolBufferException {
   }
 
   @Test
-  public void shouldFailIfGetFeatureSetWithoutProject() throws InvalidProtocolBufferException {
-    expectedException.expect(IllegalArgumentException.class);
-    expectedException.expectMessage("No project provided");
-    specService.getFeatureSet(GetFeatureSetRequest.newBuilder().setName("f1").build());
+  public void shouldUseDefaultProjectIfProjectUnspecified() throws InvalidProtocolBufferException {
```
woop, line 831: Maybe `getOrListFeatureSetShouldUseDefaultProjectIfProjectUnspecified`?

protos/feast/serving/ServingService.proto
```diff
@@ -63,14 +63,21 @@ message GetFeastServingInfoResponse {
...
+    // TODO: change to ignore post #676
     int32 version = 3;
+    
+    // Feature set name specifying the feature set of this referenced feature.
+    // This field is optional if the feature referenced is unique across the project
+    // in which case the feature set would be automatically infered
+    string feature_set_name = 5;
 
     // The features will be retrieved if:
     // entity_timestamp - max_age  <= event_timestamp <= entity_timestamp
```
woop, line 83: Is there any reason why we can't remove this max age as well? I know its not listed explicitly in your issue, but its kind of bothering me.

sdk/go/request.go
```diff
@@ -45,57 +45,53 @@ func (r OnlineFeaturesRequest) buildRequest() (*serving.GetOnlineFeaturesRequest
...
-			featureSplit = strings.Split(projectSplit[1], ":")
-		} else if len(projectSplit) == 1 {
-			project = defaultProject
-			featureSplit = strings.Split(projectSplit[0], ":")
-		} else {
-			return nil, fmt.Errorf(ErrInvalidFeatureName, featureRef)
+// Creates a slice of FeatureReferences from string representation in
+// the format featureset:feature.
+// featureRefStrs - string feature references to parse.
+// project - Sets the project in parsed FeatureReferences. Otherwise ""
```
woop, line 56: Think you have a type there with `""`

sdk/go/request.go
```diff
@@ -2,37 +2,37 @@ package feast
...
+	"github.com/gojek/feast/sdk/go/protos/feast/serving"
 )
 
 var (
-	// ErrInvalidFeatureName indicates that the user has provided a feature reference with the wrong structure or contents
-	ErrInvalidFeatureName = "invalid feature references %s provided, feature names must be in the format <project>/<feature>:<version>"
+	// ErrInvalidFeatureRef indicates that the user has provided a feature reference
+	// with the wrong structure or contents
+	ErrInvalidFeatureRef = "Invalid Feature Reference %s provided, " +
+		"feature reference must be in the format featureset:]name"
```
woop, line 16: I think there is another typo here.

sdk/go/request.go
```diff
@@ -45,57 +45,53 @@ func (r OnlineFeaturesRequest) buildRequest() (*serving.GetOnlineFeaturesRequest
...
-		if len(featureSplit) == 2 {
-			name = featureSplit[0]
-			v, err := strconv.Atoi(featureSplit[1])
+		var featureRef serving.FeatureReference
+		if strings.Contains(parseStr, "/") {
+			return nil, fmt.Errorf(ErrInvalidFeatureRef, featureRefStr)
+		}
+		// parse version if specified
+		hasVersion, _ := regexp.MatchString(".*:\\d+", parseStr)
+		if hasVersion {
```
woop, line 74: Would you mind shining some light on this code? Wouldn't it be easier to just ignore versions and have Core deal with it? So if you set a version it fails at Core, if you don't set it then it succeeds. So upgrading to Feast 0.5 requires you to remove versions.

sdk/python/feast/client.py
```diff
@@ -339,13 +341,17 @@ def archive_project(self, project):
...
+                ArchiveProjectRequest(name=project),
+                timeout=self._config.getint(CONFIG_GRPC_CONNECTION_TIMEOUT_DEFAULT_KEY),
+            )  # type: ArchiveProjectResponse
+        except grpc.RpcError as e:
+            raise grpc.RpcError(e.details())
 
+        # revert to the default project
         if self._project == project:
-            self._project = ""
+            self._project = FEAST_DEFAULT_OPTIONS[CONFIG_PROJECT_KEY]
```
woop, line 354: I think its cleaner to reset the project instead of setting it. Otherwise we don't know if the active config is the default or not.

sdk/go/request.go
mrzzy: Corrected.

sdk/go/request.go
mrzzy: Not a typo. Its an optional parameter which users should pass `""` if they project is not specified as golang does not support default parameters.

core/src/main/java/feast/core/service/SpecService.java
mrzzy: It will default to the default project if its omitted.

protos/feast/serving/ServingService.proto
mrzzy: I think we can remove this safely. Does not seem to be referenced anywhere. @zhilingc

core/src/main/java/feast/core/service/AccessManagementService.java
mrzzy: Rectified

sdk/go/request.go
mrzzy: This code is added as a band aid for testing. Post #676 would be removed.

sdk/python/feast/client.py
mrzzy: When the user first runs, post `pip install feast`:
`
from feast import Client
client = Client()
`
It imports default options `FEAST_DEFAULT_OPTIONS` which sets the project to `default`.
This change makes it consistent that client will revert to `default` when project is otherwise unspecified or an existing one archived.

core/src/main/java/feast/core/service/AccessManagementService.java
mrzzy: Corrected to `if (name.equals(Project.DEFAULT_NAME)) {`

core/src/test/java/feast/core/service/JobServiceTest.java
mrzzy: Removed offending comments.

core/src/test/java/feast/core/service/SpecServiceTest.java
mrzzy: Renamed.

core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -82,7 +82,8 @@ public SpecService(
   /**
    * Get a feature set matching the feature name and version and project. The feature set name and
    * project are required, but version can be omitted by providing 0 for its value. If the version
-   * is omitted, the latest feature set will be provided.
+   * is omitted, the latest feature set will be provided. If the project is ommited, the default
```
zhilingc, line 85: typo in omitted :)

core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -134,11 +137,12 @@ public GetFeatureSetResponse getFeatureSet(GetFeatureSetRequest request)
...
    * Return a list of feature sets matching the feature set name, version, and project provided in
-   * the filter. All fields are requried. Use '*' for all three arguments in order to return all
-   * feature sets and versions in all projects.
+   * the filter. Use '*' for all three arguments in order to return all feature sets and versions in
+   * all projects.
    *
    * <p>Project name can be explicitly provided, or an asterisk can be provided to match all
-   * projects. It is not possible to provide a combination of asterisks/wildcards and text.
+   * projects. It is not possible to provide a combination of asterisks/wildcards and text. If the
+   * project name is ommited, the default would be used.
```
zhilingc, line 145: typo here as well

core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -275,7 +284,8 @@ public ListStoresResponse listStores(ListStoresRequest.Filter filter) {
 
   /**
    * Creates or updates a feature set in the repository. If there is a change in the feature set
-   * schema, then the feature set version will be incremented.
+   * schema, then the feature set version will be incremented. If there a project is not specified
```
zhilingc, line 287: If the project is not specified in the given feature set, it will be assigned to the default project.

protos/feast/serving/ServingService.proto
zhilingc: Yeah it's not used at all

sdk/go/client.go
```diff
@@ -49,6 +50,31 @@ func (fc *GrpcClient) GetOnlineFeatures(ctx context.Context, req *OnlineFeatures
...
+	
+	// collect unqiue entity refs from entity rows
+	var entityRefs map[string]struct{}
+	for _, entityRows := range req.Entities {
+		for ref, _ := range entityRows {
+			entityRefs[ref] = struct{}{}
+		}
+	}
+
+	// strip projects from to projects
```
zhilingc, line 62: Why does the SDK strip projects from serving's response?

core/src/main/java/feast/core/model/Project.java
```diff
@@ -34,6 +34,7 @@
 @Entity
 @Table(name = "projects")
 public class Project {
+  public static final String DEFAULT_NAME = "default";
```
zhilingc, line 37: Also, is the default project not configurable?

core/src/main/java/feast/core/service/SpecService.java
mrzzy: Corrected.

core/src/main/java/feast/core/service/SpecService.java
mrzzy: Corrected.

core/src/main/java/feast/core/service/SpecService.java
mrzzy: Not sure why the comment is there. Currently, `listStore.Filter` has only one field: store name. Should we still be able to list stores by project? @woop

sdk/go/client.go
mrzzy: We want to remove projects in Feature References in user facing SDK API's completely.
- If the user makes a request and specifies the `project` parameter on get online/batch features, the SDK will add the project to the Feature Reference submitted to serving.
- Serving processes these Feature References with projects and keys the fields with Feature References with Projects unaware that the project was added by the SDK not the user.
- The SDK strips the Feature References of their projects and returns them to the user.

The stripping of projects is done so that when the user specifies `driver_fs:name` with `project="rides
` they would get `driver_fs:name` back instead of `rides/driver_fs:name` returned from serving.

core/src/main/java/feast/core/model/Project.java
mrzzy: Yes its not configurable.

serving/src/main/java/feast/serving/specs/CachedSpecService.java
```diff
@@ -191,21 +220,81 @@ public void scheduledPopulateCache() {
...
-                .setProject(featureSetSpec.getProject())
-                .setName(featureSpec.getName())
-                .build();
-        mapping.put(
-            generateFeatureStringRef(featureRef), generateFeatureSetStringRef(featureSetSpec));
-      }
-    }
+    Map<String, String> mapping = new HashMap<>();
+
+    featureSets.values().stream()
```
zhilingc, line 235: This means that how users are able to reference features varies from store to store, right?

serving/src/main/java/feast/serving/util/RefUtil.java
```diff
@@ -21,7 +21,13 @@
 
 public class RefUtil {
   public static String generateFeatureStringRef(FeatureReference featureReference) {
```
zhilingc, line 23: I think we can do away with these string references and straight up use `FeatureSetReference`s and `FeatureReference`s in the maps.

sdk/go/client.go
zhilingc: I think rather than strip projects completely it would be a lot more intuitive to the user if the returned feature reference were consistent with what they provided (if a user provides `rides/driver_fs:name` they would expect to retrieve their values with the reference `rides/driver_fs:name` over `driver_fs:name`)

core/src/main/java/feast/core/service/SpecService.java
zhilingc: No, sorry, i'm correcting the comment since it's a little hard to understand

sdk/go/client.go
mrzzy: Yes that would be indeed clearer if that is the case.

However this PR attempts to address #631 which stipulates that we want to move towards to project less Feature References as we figure what place projects has in Feast core model in #479. 

As such in the new `v0.5` SDKs, I have dropped projects in Feature References entirely,  users can only specify `driver_fs:name` or `name`. Putting in feature references (ie `rides/driver_fs:name` would result in an error in the SDK. If users want to specify a project, they have to do so via a dedicated parameter in get online/batch features methods. 

This is done in the SDK instead of Serving as I also want to maintain some form of compatibility with the old v0.4 SDK which does have project in its feature references.

serving/src/main/java/feast/serving/util/RefUtil.java
mrzzy: Yeah I think its a good idea. Although I think we should address in a separate dedicated techdebt/refactoring PR. Related issue: #674

core/src/main/java/feast/core/service/SpecService.java
mrzzy: Corrected.

protos/feast/serving/ServingService.proto
```diff
@@ -63,12 +63,21 @@ message GetFeastServingInfoResponse {
...
-    // Project name
+    // Project name. This field is optional, if unspecified will default to 'default'.
     string project = 1;
 
     // Feature name
     string name = 2;
 
+    // Feature version. Ignored by Feast for backwards compatibility.
+    // NOTE: This field is deprecated and will be removed in a future version
+    int32 version = 3 [deprecated = true];
```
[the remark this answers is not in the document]
mrzzy: Removed

core/src/main/java/feast/core/service/SpecService.java
mrzzy: Corrected.

core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -94,8 +95,9 @@ public GetFeatureSetResponse getFeatureSet(GetFeatureSetRequest request)
     if (request.getName().isEmpty()) {
       throw new IllegalArgumentException("No feature set name provided");
     }
+    // Autofill default project if project is not specified
     if (request.getProject().isEmpty()) {
-      throw new IllegalArgumentException("No project provided");
+      request = request.toBuilder().setProject("default").build();
```
woop, line 100: Should this be using one global project constant?

core/src/main/java/feast/core/service/AccessManagementService.java
```diff
@@ -61,6 +64,9 @@ public void archiveProject(String name) {
     if (!project.isPresent()) {
       throw new IllegalArgumentException(String.format("Could not find project: \"%s\"", name));
     }
+    if (name.equals(Project.DEFAULT_NAME)) {
+      throw new IllegalArgumentException("Archiving the default project is not allowed.");
```
woop, line 68: How about `UnsupportedOperationException`?

core/src/test/java/feast/core/service/SpecServiceTest.java
```diff
@@ -661,10 +675,20 @@ public void shouldDoNothingIfNoChange() throws InvalidProtocolBufferException {
   }
 
   @Test
-  public void shouldFailIfGetFeatureSetWithoutProject() throws InvalidProtocolBufferException {
-    expectedException.expect(IllegalArgumentException.class);
-    expectedException.expectMessage("No project provided");
-    specService.getFeatureSet(GetFeatureSetRequest.newBuilder().setName("f1").build());
+  public void getOrListFeatureSetShouldUseDefaultProjectIfProjectUnspecified()
+      throws InvalidProtocolBufferException {
+    when(featureSetRepository.findFeatureSetByNameAndProject_Name("f4", Project.DEFAULT_NAME))
```
woop, line 680: Should we be testing the list/find/get/apply without providing the `project` at all? Not sure if I am misreading this, but it seems like you are providing it.

protos/feast/core/CoreService.proto
```diff
@@ -154,6 +155,9 @@ message ListStoresResponse {
 }
 
 message ApplyFeatureSetRequest {
+    // Feature set version and source will be ignored.
```
woop, line 158: Can you elaborate on the `source` being ignored?

protos/feast/serving/ServingService.proto
woop: Keeping this open until its resolved.

sdk/java/src/main/java/com/gojek/feast/FeastClient.java
```diff
@@ -125,7 +143,18 @@ public GetFeastServingInfoResponse getFeastServingInfo() {
         .map(
             field -> {
               Row row = Row.create();
-              field.getFieldsMap().forEach(row::set);
+              field
+                  .getFieldsMap()
+                  .forEach(
+                      (String name, Object value) -> {
```
woop, line 149: Are you intentionally using `Object` here or is there a more specific type we can use?

sdk/java/src/main/java/com/gojek/feast/RequestUtil.java
```diff
@@ -17,47 +17,88 @@
...
+    if (featureRefString.contains("/")) {
+      if (ignoreProject) {
+        String[] projectSplit = featureRefString.split("/");
+        featureRefString = projectSplit[1];
       } else {
         throw new IllegalArgumentException(
-            String.format(
-                "Feature id '%s' has invalid format. Expected format: <project>/<feature-name>.",
-                featureRefString));
+            String.format("Unsupported feature reference: %s", featureRefString));
```
woop, line 74: how does `ignoreProject` relate to the incoming request. It seems like `project1/feature_set:feature` would raise an exception every time if `ignoreProject` is true.

sdk/java/src/main/java/com/gojek/feast/RequestUtil.java
```diff
@@ -17,47 +17,88 @@
...
-        throw new IllegalArgumentException(
-            String.format(
-                "Feature id '%s' has invalid format. Expected format: <project>/<feature-name>.",
-                featureRefString));
-      }
+    // parse featureset if specified
+    if (featureRefString.contains(":")) {
+      String[] featureSetSplit = featureRefString.split(":");
+      featureRef.setFeatureSetName(featureSetSplit[0]);
+      featureRefString = featureSetSplit[1];
```
woop, line 82: Can we create a new local variable instead of reusing this one? Hard to follow...

sdk/java/src/main/java/com/gojek/feast/RequestUtil.java
```diff
@@ -17,47 +17,88 @@
...
+  }
 
-      featureRefs.add(FeatureReference.newBuilder().setName(name).setProject(project).build());
+  /**
+   * Render a feature reference as string.
+   *
+   * @param featureReference to render as string
+   * @return string represenation of feature reference.
+   */
+  public static String renderFeatureRef(FeatureReference featureReference) {
```
woop, line 94: Is `render` == `toString()`? I think if you want to introduce a new term then you should flesh out the comment a bit. It's a bit too vague.

sdk/java/src/test/java/com/gojek/feast/RequestUtilTest.java
```diff
@@ -89,23 +66,35 @@ void createFeatureSets_ShouldReturnFeatureSetsForValidFeatureIds(
...
+    assertEquals(expected.size(), actual.size());
+    for (int i = 0; i < expected.size(); i++) {
+      assertEquals(expected.get(i), actual.get(i));
+    }
+  }
+
   private static Stream<Arguments> provideInvalidFeatureRefs() {
-    return Stream.of(
-        Arguments.of(Collections.singletonList("/noproject")),
-        Arguments.of(Collections.singletonList("")));
```
woop, line 95: Is there a reason `""` was removed?

sdk/python/feast/client.py
```diff
@@ -834,50 +864,25 @@ def ingest(
...
+        feature_ref_strs: List of string feature references
+        project: Optionally specifies the project in the parsed feature references.
 
-        features.append(FeatureReference(project=project, name=name))
-    return features
+    Returns:
+        A list of FeatureReference protos parsed from args.
+    """
+    feature_refs = [FeatureRef.from_str(ref_str) for ref_str in feature_ref_strs]
+    ref_protos = [ref.to_proto() for ref in feature_refs]
```
woop, line 880: I'd prefer `feature_ref_proto` or `feature_ref_spec` just to disambiguate with `feature_set_ref`

sdk/python/feast/feature.py
```diff
@@ -62,3 +63,86 @@ def from_proto(cls, feature_proto: FeatureProto):
         feature.update_shape_type(feature_proto)
         feature.update_domain_info(feature_proto)
         return feature
+
+
+class FeatureRef:
+    """ Feature Reference represents a reference to a specific feature.  """
+
+    def __init__(self, name: str, feature_set_name: str = None):
```
woop, line 71: I know you are matching the proto field names here, but eventually I think we should rename these to `feature` and `feature_set`. Now might be the time.

sdk/python/feast/client.py
```diff
@@ -635,14 +642,37 @@ def get_online_features(
...
                     ),
                     entity_rows=entity_rows,
                 )
             )
+            # collect entity row refs
+            entity_refs = set()
+            for entity_row in entity_rows:
+                entity_refs.update(entity_row.fields.keys())
+
+            def strip_project(field_value):
```
woop, line 656: What is the advantage of nesting this method as opposed to in-lining the code in a loop or moving the method out? `get_online_features` is already quite big, so perhaps we should handle the response cleanup elsewhere.

sdk/python/feast/feature.py
```diff
@@ -62,3 +63,86 @@ def from_proto(cls, feature_proto: FeatureProto):
...
+        proto = FeatureRefProto()
+        if "/" in feature_ref_str:
+            if ignore_project:
+                _, feature_ref_str = feature_ref_str.split("/")
+            else:
+                raise ValueError(f"Unsupported feature reference: {feature_ref_str}")
+
+        # parse feature set name if specified
+        if ":" in feature_ref_str:
+            proto.feature_set_name, feature_ref_str = feature_ref_str.split(":")
```
woop, line 112: Same comment as above. I think it would be clearer if we didn't reuse the `feature_ref_str` name and instead used something like `feature_name`

sdk/python/feast/test_feature.py
```diff
@@ -0,0 +1,23 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from feast.feature import FeatureRef
+
+
+class TestFeatureRef:
+    def test_str_ref(self):
+        original_ref = FeatureRef(project="test", name="test", version=2)
```
woop, line 20: Should this version still be here?

sdk/python/tests/test_client.py
```diff
@@ -400,130 +409,125 @@ def test_stop_ingest_job(self, mocked_client, mocker):
...
+        )
+
+        mocker.patch.object(
+            mocked_client._core_service_stub,
+            "GetFeatureSet",
+            return_value=GetFeatureSetResponse(
+                feature_set=FeatureSetProto(
+                    spec=FeatureSetSpecProto(
+                        name="driver",
+                        project="driver_project",
```
woop, line 432: Would it make sense to have an online and batch test where no project is defined? Unless that is being tested elsewhere.

serving/src/main/java/feast/serving/specs/CachedSpecService.java
```diff
@@ -126,6 +139,19 @@ public FeatureSetSpec getFeatureSetSpec(String featureSetRef) throws ExecutionEx
                 FeatureSetSpec featureSetSpec = featureSetCache.get(fsName);
                 List<FeatureReference> requestedFeatures =
                     featureRefs.stream().map(Pair::getRight).collect(Collectors.toList());
+
+                // check that requested features reference point to different features in the
+                // featureset.
+                HashSet<String> featureNames = new HashSet<>();
```
woop, line 145: What is the downside of allowing this? (I am just thinking of the performance hit of keeping this in, even if its small).

woop:
Just spotted something. We are clearing this map, then building it again. It's not clear how long the time between `clear` and `putAll` is, but this might be a minor bug that we can resolve. `featureToFeatureSetMapping` is being used for every incoming request.

```
featureToFeatureSetMapping.clear();
featureToFeatureSetMapping.putAll(getFeatureToFeatureSetMapping(featureSetMap));
```

https://github.com/feast-dev/feast/blob/master/serving/src/main/java/feast/serving/specs/CachedSpecService.java#L153

cc @zhilingc

serving/src/main/java/feast/serving/specs/CachedSpecService.java
```diff
@@ -191,21 +217,81 @@ public void scheduledPopulateCache() {
...
+              for (FeatureSpec featureSpec : featureSetSpec.getFeaturesList()) {
+                // Register the different permutations of string feature references
+                // that refers to this feature in the feature to featureset mapping.
+
+                // Features in FeatureSets in default project can be referenced without project.
+                boolean isInDefaultProject =
+                    featureSetSpec.getProject().equals(DEFAULT_PROJECT_NAME);
+
+                for (boolean hasProject : new boolean[] {true, false}) {
+                  if (!isInDefaultProject && hasProject == false) continue;
```
woop, line 244: `hasProject == false` == `!hasProject`

serving/src/main/java/feast/serving/specs/CachedSpecService.java
woop: How so?

serving/src/test/java/feast/serving/service/CachedSpecServiceTest.java
```diff
@@ -63,70 +60,60 @@
...
-            .setRedisConfig(RedisConfig.newBuilder().setHost("localhost").setPort(6379))
-            .addSubscriptions(
-                Subscription.newBuilder().setProject("project").setName("fs1").build())
-            .addSubscriptions(
-                Subscription.newBuilder().setProject("project").setName("fs2").build())
-            .build();
-
-    when(coreService.registerStore(store)).thenReturn(store);
+    this.store = Store.newBuilder().build();
+    this.featureSetSpecs = new HashMap<>();
```
woop, line 64: Was the change from `LinkedHashMap` to `HashMap` intentional? The one is ordered.

serving/src/test/java/feast/serving/service/OnlineServingServiceTest.java
```diff
@@ -181,7 +103,7 @@ public void shouldReturnKeysWithoutVersionIfNotProvided() {
                         Field.newBuilder().setName("entity2").setValue(strValue("b")).build(),
                         Field.newBuilder().setName("feature1").setValue(intValue(2)).build(),
                         Field.newBuilder().setName("feature2").setValue(intValue(2)).build()))
-                .setFeatureSet("featureSet")
+                .setFeatureSet("project/featureSet")
```
woop, line 106: Would the default project result in a feature set of `featureSet` or `default/featureSet`? Do we have a test for the default case?

tests/e2e/basic-ingest-redis-serving.py
```diff
@@ -63,101 +63,164 @@ def client(core_url, serving_url, allow_dirty):
...
-    if cust_trans_fs_actual is None:
-        raise Exception(
-            "Client cannot retrieve 'customer_transactions' FeatureSet "
-            "after registration. Either Feast Core does not save the "
-            "FeatureSet correctly or the client needs to wait longer for FeatureSet "
-            "to be committed."
-        )
-
+    # reset client's project to default for other tests
+    client.set_project(DEFAULT_PROJECT)
```
woop, line 119: I want to make sure that users dont have to set the project. Can we perhaps recreate the client or force clear this project?

tests/e2e/basic-ingest-redis-serving.py
```diff
@@ -63,101 +63,164 @@ def client(core_url, serving_url, allow_dirty):
...
 
+@pytest.mark.timeout(45)
+@pytest.mark.run(order=13)
+def test_basic_retrieve_online_multiple_featureset(client, cust_trans_df, driver_df):
+    # Poll serving for feature values until the correct values are returned
+    while True:
+        time.sleep(1)
+        feature_refs = [
+            "customer_transactions:daily_transactions",
+            "driver:rating",
```
woop, line 183: Should we maybe omit one feature set here?

core/src/test/java/feast/core/service/SpecServiceTest.java
mrzzy: Notice that I'm not providing the default project when I'm building list/get/apply request (ie no calls to `.setProject()`). Here I'm only provide the default project to mocks to verify that list/get/apply calls will map to the default project in SpecService and correspondingly makes the correct calls to  the`featureSetRepository` mock with the correct project: `default`

woop: Alright, my bad :)

sdk/python/feast/feature.py
mrzzy: The `feature_ref_str` reuse is imposed  by the flexibility of feature references. I have used this pattern in various  string Feature Reference parsing methods throughout the codebase, not just in the python SDK

serving/src/test/java/feast/serving/service/CachedSpecServiceTest.java
mrzzy: The original never used the ordered trait of `LinkedHashMap`, only calling `.get()` on it, hence the change to `HashMap`.

serving/src/test/java/feast/serving/service/OnlineServingServiceTest.java
mrzzy: > Would the default project result in a feature set of featureSet or default/featureSet? 

Pending  @zhilingc 's PR  #701 

> Do we have a test for the default case?

We not need to test for the default case here as `OnlineServingService` never interacts with the `feature_set` field of the feature row. (ie The tests would pass even if `feature_set` is unset in the `FeatureRow`)

core/src/main/java/feast/core/service/SpecService.java
mrzzy: Corrected.

core/src/main/java/feast/core/service/AccessManagementService.java
mrzzy: Corrected.

protos/feast/core/CoreService.proto
mrzzy: Sorry this documentation is just plainly incorrect. Source is not ignored if specified

sdk/java/src/main/java/com/gojek/feast/FeastClient.java
mrzzy: We can use `ValueProto.Value` here. I used `Object` as `Row.set(String, Object)` used `Object`

sdk/java/src/main/java/com/gojek/feast/RequestUtil.java
mrzzy: Clarified.

sdk/java/src/main/java/com/gojek/feast/RequestUtil.java
mrzzy: Corrected

sdk/java/src/test/java/com/gojek/feast/RequestUtilTest.java
mrzzy: Nope. Corrected.

sdk/python/feast/client.py
mrzzy: Corrected.

sdk/python/feast/feature.py
mrzzy: Changed `FeatureReference`'s `feature_set_name` field to `feature_set`

sdk/python/feast/client.py
mrzzy: Removed nesting in favor of a loop.

sdk/python/feast/test_feature.py
mrzzy: Nope removed.

sdk/python/tests/test_client.py
mrzzy: That should be tested in the e2e tests

serving/src/main/java/feast/serving/specs/CachedSpecService.java
mrzzy: Its a limitation posed by this `FeatureSetRequest`:
https://github.com/feast-dev/feast/blob/14cae741068b4e18a56aaea211099564d78a703b/storage/api/src/main/java/feast/storage/api/retriever/FeatureSetRequest.java#L58

serving/src/main/java/feast/serving/specs/CachedSpecService.java
mrzzy: Changed

tests/e2e/basic-ingest-redis-serving.py
mrzzy: Changed this to `client.set_project()` to reset to default project.

tests/e2e/basic-ingest-redis-serving.py
mrzzy: Corrected to test Feature Set inference.

protos/feast/serving/ServingService.proto
mrzzy: Corrected

sdk/go/client.go
zhilingc: So that means in this transitional state retrieving across projects is not possible?

mrzzy: Yes

storage/connectors/bigquery/src/main/resources/templates/join_featuresets.sql
```diff
@@ -17,8 +17,8 @@ LEFT JOIN (
     event_timestamp,
     {{ entities | join(', ') }}
     {% for featureSet in featureSets %}
-    {% for featureName in featureSet.features %}
-    ,{{ featureSet.project }}_{{ featureName }} as {{ featureName }}
+    {% for feature in featureSet.features %}
+	,{{ featureSet.project }}__{{ featureSet.name }}__{{ feature.name }} as {% if feature.featureSet != "" %}{{ featureSet.name }}__{% endif %}{{ feature.name }}
```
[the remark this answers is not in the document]
mrzzy: This should not be a concern as v0.4 was limited project-namespace feature names that were unique across the project. Hence the feature set + feature collision should not happen as v0.4 feature names are unique anyway.

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/retriever/QueryTemplater.java
```diff
@@ -79,10 +79,7 @@ public static String createEntityTableUUIDQuery(String leftTableName) {
       Duration maxAge = spec.getMaxAge();
       List<String> fsEntities =
           spec.getEntitiesList().stream().map(EntitySpec::getName).collect(Collectors.toList());
-      List<String> features =
-          featureSetRequest.getFeatureReferences().stream()
-              .map(FeatureReference::getName)
-              .collect(Collectors.toList());
+      List<FeatureReference> features = List.copyOf(featureSetRequest.getFeatureReferences());
```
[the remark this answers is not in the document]
mrzzy: Corrected

PR #695, opened 2020-05-12
Clean up documentation

heysachin:

This brings more clarity on, Point-in-time-correct Join and a typo correction.


woop:
Thanks @heysachin

PR #698, opened 2020-05-13
Clean up documentation

heysachin:

This brings more clarity on, Point-in-time-correct Join and a typo correction.


heysachin:
@woop This is a copy of an old pull request you approved, the older one had some conflicts and the merging of them cause the pull to be a mess and thus I had to create a new one.

PR #700, opened 2020-05-14
Add .proto to packages of Protobuf generated Java classes

### LINKED ISSUE #684: Package conflicts exist between proto and non-proto modules.
Java classes that are currently generated from Protobuf definitions have overlapping package names with Feast modules. This causes conflicts when trying to import classes from Feast modules.

For example the [CoreService.proto](https://github.com/gojek/feast/blob/master/protos/feast/core/CoreService.proto#L23) uses `feast.core`, which is also in use by the `Feast Core` module: [example](https://github.com/gojek/feast/blob/master/core/src/main/java/feast/core/CoreApplication.java#L17).

Proposal is to prefix all proto generated packages with `.proto`

woop:

Adds `.proto` to the package of Java generated proto code in order to differentiate it from non-proto code.

Example: `feast.core.CoreServiceGrpc` to `feast.proto.core.CoreServiceGrpc`

Fixes #684 

```release-note
Java classes generated from Protobuf files have had .proto added to their package path.
```


ches:
/lgtm

Hope there isn't a reason to hold this—there's no hold label—but it's potentially a big merge conflict pain waiting for anyone who branches before this gets in, so if we want let's get it done. Anyone with already open PRs, yell at @woop not me 😉

woop:
> /lgtm
> 
> Hope there isn't a reason to hold this—there's no hold label—but it's potentially a big merge conflict pain waiting for anyone who branches before this gets in, so if we want it let's get it done. Anyone with already open PRs, yell at @woop not me

I think there will be yelling, but it won't be because of this :)

We've had a long running Mexican standoff between PRs for a while now, so I have asked @mrzzy to just squash his history into a [single commit (well it's 4 now)](https://github.com/feast-dev/feast/pull/693/commits)so that he doesn't lose years of his life rebasing. 

Anyway, thanks for getting this merged in!

PR #701, opened 2020-05-14
Apply default project to rows without project during ingestion

zhilingc:

Extension to #693. Feast Core deploys Feast ingestion jobs with a parameter that sets its default project. This default project is just a string (using “default” for now).

When a FeatureRow is read from Kafka with with a complete feature set reference (`project/feature_set`) then it is processed as normal. However, when a FeatureRow is read in with a feature set reference missing project (just `featureset`), then Feast Ingestion prepends the default project string to this reference.

```release-note
Feature rows can be ingested with just the feature set name. Feast core will apply the default project to those feature rows.
```

### DIFF
--- a/ingestion/src/test/java/feast/ingestion/transform/ProcessAndValidateFeatureRowsTest.java
+++ b/ingestion/src/test/java/feast/ingestion/transform/ProcessAndValidateFeatureRowsTest.java
@@ -109,6 +109,7 @@ public class ProcessAndValidateFeatureRowsTest {
             .setCoder(ProtoCoder.of(FeatureRow.class))
             .apply(
                 ProcessAndValidateFeatureRows.newBuilder()
+                    .setDefaultProject("myproject")
                     .setFailureTag(FAILURE_TAG)
                     .setSuccessTag(SUCCESS_TAG)
                     .setFeatureSetSpecs(featureSetSpecs)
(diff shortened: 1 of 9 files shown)


ingestion/src/test/java/feast/ingestion/transform/ProcessAndValidateFeatureRowsTest.java
```diff
@@ -158,6 +159,56 @@ public void shouldStripVersions() {
...
+
+    p.run();
+  }
+
+  @Test
+  public void shouldApplyDefaultProject() {
+    FeatureSetSpec fs1 =
+        FeatureSetSpec.newBuilder()
+            .setName("feature_set")
+            .setProject("myproject")
```
woop, line 178: Should you be providing the `project` here?

zhilingc: No, it only autofills project for incoming rows, not the feature spec

woop: Yes, but isn't this feature spec used to create the row in the first place?

zhilingc: It's manually set a few lines down

woop: Oops I misread.

PR #707, opened 2020-05-14
Add support for labels on feature sets

### LINKED ISSUE #663: Update Python SDK to support labels
This issue follows #463 which has a partial implementation in #536. 

One additional PR is needed to get #436 into a consistent and complete state for end users. Currently #536 makes changes to Feast Core by adding support for labels to feature sets and features, but it doesn't add support for labels to feature sets in the Python SDK.

The task here is to add this support and update any end-to-end tests that test both the Python SDK and Feast Core w.r.t label metadata.

Joostrothweiler:

Adding labels to Python sdk with relevant FeatureSet class functions:
```python
set_label(key: str, value: str)
remove_label(key: str)
```

Fixes #663 

```release-note
Users can use the Python SDK to add metadata to feature sets in the form of labels.
```

### DIFF
--- a/sdk/python/feast/feature_set.py
+++ b/sdk/python/feast/feature_set.py
@@ -86,7 +86,8 @@ class FeatureSet:
                 return False
 
         if (
-            self.name != other.name
+            self.labels != other.labels
+            or self.name != other.name
             or self.project != other.project
             or self.max_age != other.max_age
         ):
--- a/sdk/python/feast/feature_set.py
+++ b/sdk/python/feast/feature_set.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 import warnings
 from collections import OrderedDict
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, MutableMapping
 
 import pandas as pd
 from google.protobuf import json_format
@@ -56,6 +56,7 @@ class FeatureSet:
         entities: List[Entity] = None,
         source: Source = None,
         max_age: Optional[Duration] = None,
+        labels: Optional[MutableMapping[str, str]] = None
     ):
         self._name = name
         self._project = project
@@ -69,6 +70,7 @@ class FeatureSet:
         else:
             self._source = source
         self._max_age = max_age
+        self._labels = labels
         self._status = None
         self._created_timestamp = None
 
@@ -218,6 +220,21 @@ class FeatureSet:
         self._max_age = max_age
 
     @property
+    def labels(self):
+        """
+        Returns the labels of this feature set. This is the user defined metadata
+        defined as a dictionary.
+        """
+        return self._labels
+
+    @labels.setter
+    def labels(self, labels: MutableMapping[str, str]):
+        """
+        Set the labels for this feature set
+        """
+        self._labels = labels
+
+    @property
     def status(self):
         """
         Returns the status of this feature set
@@ -244,6 +261,24 @@ class FeatureSet:
         Sets the status of this feature set
         """
         self._created_timestamp = created_timestamp
+
+    def set_label(self, key: str, value: str):
+        """
+        Sets the label value for a given key
+        """
+        if not self.labels:
+            self.labels = {key: value}
+        else:
+            self.labels[key] = value
+
+    def remove_label(self, key: str):
+        """
+        Removes a label based on key
+        """
+        if not self.labels or key not in self.labels.keys():
+            raise ValueError("Could not find label key " + key + ", no action taken")
+        elif key in self.labels.keys():
+            del self.labels[key]
 
     def add(self, resource):
         """
@@ -796,6 +831,7 @@ class FeatureSet:
                 and feature_set_proto.spec.max_age.nanos == 0
                 else feature_set_proto.spec.max_age
             ),
+            labels=feature_set_proto.spec.labels,
             source=(
                 None
                 if feature_set_proto.spec.source.type == 0
@@ -825,6 +861,7 @@ class FeatureSet:
             name=self.name,
             project=self.project,
             max_age=self.max_age,
+            labels=self.labels,
             source=self.source.to_proto() if self.source is not None else None,
             features=[
                 field.to_proto()
--- a/sdk/python/feast/feature_set.py
+++ b/sdk/python/feast/feature_set.py
@@ -69,8 +69,11 @@ class FeatureSet:
             self._source = None
         else:
             self._source = source
+        if labels is None:
+            self._labels = OrderedDict()
+        else:
+            self._labels = labels
         self._max_age = max_age
-        self._labels = labels
         self._status = None
         self._created_timestamp = None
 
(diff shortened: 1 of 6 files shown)


ches:
Thanks for the contribution @Joostrothweiler!

I haven't reviewed yet, but something I was just thinking of that I want to remind us all of:

Verify that this enables the Feast CLI to handle labels applied with YAML feature set specifications.

It needs a bit of extra work I think, so this could be separate task/issue if you can't get to it. Just want to track it if we decide that, because I think the YAML becomes especially helpful if using quite a bit of metadata.

It may raise questions like if "applying" a feature set should nullify existing label fields if labels aren't given every time, which would be a usability annoyance IMO. But this definitely goes into new discussion territory outside this PR.

sdk/python/feast/feature_set.py
```diff
@@ -238,6 +256,24 @@ def created_timestamp(self, created_timestamp):
...
+            self.labels[key] = value
+
+    def remove_label(self, key: str):
+        """
+        Removes a label based on key
+        """
+        if not self.labels or key not in self.labels.keys():
+            raise ValueError("Could not find label key " + key + ", no action taken")
+        elif key in self.labels.keys():
+            del self.labels[key]
```
ches, line 275: This is consistent with how the `drop` method behaves for features and entities on the feature set, so the `ValueError` might be the right thing for the scope of this PR. Separately though, I wonder if `KeyError` would be more appropriate for both, if not an application-specific error.

I might be inclined to initialize labels as an empty dictionary instead of the type being `Optional` though—unless there's a clear application-specific semantic reason for distinguishing no value from empty, I think empty is preferable design for collections. Give a stronger contract and spare both us and callers from null checks.

In that case, the method implementation could be reduced to `del self.labels[key]` with `KeyError` propagating, unless we want to give it a more descriptive error message.

sdk/python/feast/field.py
```diff
@@ -65,6 +71,13 @@ def dtype(self) -> ValueType:
         """
         return self._dtype
 
+    @property
+    def labels(self) -> MutableMapping[str, str]:
+        """
+        Getter for labels of this field
+        """
+        return self._presence
```
ches, line 79: Copypasta mistake, should be `_labels` 😇 

Calls to attention that we're missing label tests for fields, although I wonder if we should just be removing `Field` from the Python SDK at this point.

Joostrothweiler:
> There is a bug on `Field` so that is critical.

Missed this one.. But indeed this also showed I missed some tests for Feature. Made the change and added some tests.

> Empty dict vs. `Optional` is a design choice that I'll leave the maintainers to weigh in on, I'm not too familiar with the Python SDK and my sensibility of Pythonic is rusty these days.

I agree that an empty dict would make more sense. Changed it to initialize it to `OrderedDict` by default, similar to `_fields`. 

> I do think there is desire to kill `Field` in Python now as it has been in Java, and to update E2E tests for labels added in #536 to use this new API. These could be follow-ups, again I'll leave to others to weigh in.

Have not made any changes regarding this. If it's something that should be part of this PR I'm happy to make the changes.

sdk/python/feast/feature_set.py
Joostrothweiler: Indeed it makes more sense. Changed it to just use `del`. Shall I include a change for the `drop` method to raise a `KeyErrors` in this PR as well?

sdk/python/feast/field.py
Joostrothweiler: Good catch. Changed it and added some test cases.

sdk/python/feast/feature_set.py
woop: Yea, makes sense.

woop:
> Verify that this enables the Feast CLI to handle labels applied with YAML feature set specifications.

Which raises another good point. We don't have any kind of e2e test coverage for the CLI. 

> It may raise questions like if "applying" a feature set should nullify existing label fields if labels aren't given every time, which would be a usability annoyance IMO. But this definitely goes into new discussion territory outside this PR.

I think we should probably just try to model ourselves after Kubernetes here, although I admit I am not sure how this is handled. I do know that there are various ways to update a resource though.

> Have not made any changes regarding this. If it's something that should be part of this PR I'm happy to make the changes.

I think it's out of scope for your PR. The desire is strong though :)

Thanks for the PR @Joostrothweiler

woop:
@Joostrothweiler I do think updating the e2e tests to make sure that labels works is important here. There was some coverage, but not for this new API obviously.

Joostrothweiler:
@woop I updated the existing e2e test cases to take the labels directly from the `Feature(Set)` class and use `to_proto` function.

sdk/python/feast/feature_set.py
ches: Do we care much that labels are an `OrderedDict`? It's not expressed by the type `MutableMapping` either. After updates we still have this signature:

```python
labels: Optional[MutableMapping[str, str]] = None
```

which to me would be nice if were simpler:

```python
labels: MutableMapping[str, str] = {}
```

Joostrothweiler: > Do we care much that labels are an `OrderedDict`? It's not expressed by the type `MutableMapping` either (this is okay if ordering is used as an implementation detail but isn't part of contract, not sure that distinction exists here though).

I don't think we care much. I just aligned this with `self._fields`. Is there a particular reason why we would want it in this case but not for labels? I can change it to initialize with a simple empty dict if this is preferred.

> After updates we still have this signature:
> 
> ```python
> labels: Optional[MutableMapping[str, str]] = None
> ```
> 
> which to me would be nice if were simpler:
> 
> ```python
> labels: MutableMapping[str, str] = {}
> ```

This would set the default value to a mutable object, which Pycharm elegantly suggests not to do:
_"Default argument values are evaluated only once at function definition time, which means that modifying the default value of the argument will affect all subsequent calls of the function."_

ches: Ahh yes I forgot about this quirk in Python, I'm sorry. Now I'm reminded of the idiom for argument types like `features: List[Feature] = None` and `if features is None` to initialize. So that pattern should be followed, sans `Optional` type.

Regarding `OrderedDict` I'll leave it to @woop or others, I'm not sure if there was reason behind for fields and if it should apply to labels or not.

woop:
@Joostrothweiler Apologies for taking so long to review this. Will try to get to it as soon as possible.

Joostrothweiler:
@woop no problem. Let me know if there's anything I can do

sdk/python/feast/feature_set.py
woop: We do not have a special ordered contract with end users. I think when I used OrderedDict in the Python SDK I was trying to have my cake and eat it as well. 

For Fields/Features/Entities we should be using Lists (as users see them) and dicts as its implemented inside the FeatureSet class. We do not have an ordered contract with users. So technically speaking they should be maps/dicts across Feast, but from a user perspective I feel like its cleaner to expose lists, especially when persisting as YAML/JSON.

For labels I am happy to commit to simply a `dict()` instead of an OrderedDict.

Other than that the PR looks good @Joostrothweiler

woop:
I will merge this in. We can resolve the issues around Dicts later.

PR #709, opened 2020-05-15
Extract fs update tests so ci doesn't run it

zhilingc:
The tests take too long, and timeout frequently because the bq streaming cache refresh doesn't happen within a deterministic amount of time.

The tests can still be run locally if the functionality needs to be tested.

Fixes #

```release-note

```


khorshuheng:
tests/e2e/bq-batch-fs-updates.py can be included in prow, but with auto run set as false. That way, we can trigger the test from prow manually if needed.

PR #710, opened 2020-05-15
Fix kubectl installation on dataflow e2e test

khorshuheng:
The current dataflow e2e test attempts to install kubectl via apt, but it will fail because the package is not available. This PR uses alternative way to install kubectl.

Fixes #

```release-note

```


khorshuheng:
test-end-to-end-batch-dataflow

woop:
@khorshuheng should this still be open given #716?

khorshuheng:
Can be close.

PR #714, opened 2020-05-18
Add feature set status JOB_STARTING to denote feature sets waiting for job to get to RUNNING state

zhilingc:
Removal of PENDING status check in #708 causes jobs to be unable to track feature set updates.

This PR adds a new status STATUS_JOB_STARTING to denote that a feature set was updated, but a job has already been initialised, preventing the JobCoordinator from spinning up new jobs in its place.

Fixes #

```release-note

```

### DIFF
--- a/protos/feast/core/FeatureSet.proto
+++ b/protos/feast/core/FeatureSet.proto
@@ -147,5 +147,6 @@ message FeatureSetMeta {
 enum FeatureSetStatus {
     STATUS_INVALID = 0;
     STATUS_PENDING = 1;
+    STATUS_JOB_STARTING = 3;
     STATUS_READY = 2;
 }
--- a/core/src/main/java/feast/core/job/JobUpdateTask.java
+++ b/core/src/main/java/feast/core/job/JobUpdateTask.java
@@ -25,6 +25,7 @@ import feast.core.model.Job;
 import feast.core.model.JobStatus;
 import feast.core.model.Source;
 import feast.core.model.Store;
+import feast.proto.core.FeatureSetProto.FeatureSetStatus;
 import java.time.Instant;
 import java.util.List;
 import java.util.Optional;
@@ -102,7 +103,16 @@ public class JobUpdateTask implements Callable<Job> {
 
   boolean requiresUpdate(Job job) {
     // If set of feature sets has changed
-    return !Sets.newHashSet(featureSets).equals(Sets.newHashSet(job.getFeatureSets()));
+    if (!Sets.newHashSet(featureSets).equals(Sets.newHashSet(job.getFeatureSets()))) {
+      return true;
+    }
+    // If any of the incoming feature sets were updated
+    for (FeatureSet featureSet : featureSets) {
+      if (featureSet.getStatus() == FeatureSetStatus.STATUS_PENDING) {
+        return true;
+      }
+    }
+    return false;
   }
 
   private Job createJob() {
--- a/core/src/main/java/feast/core/job/dataflow/DataflowJobManager.java
+++ b/core/src/main/java/feast/core/job/dataflow/DataflowJobManager.java
@@ -162,6 +162,7 @@ public class DataflowJobManager implements JobManager {
               true);
 
       job.setExtId(extId);
+      job.setStatus(JobStatus.PENDING);
       return job;
     } catch (InvalidProtocolBufferException e) {
       log.error(e.getMessage());
--- a/core/src/main/java/feast/core/service/JobCoordinatorService.java
+++ b/core/src/main/java/feast/core/service/JobCoordinatorService.java
@@ -132,17 +132,19 @@ public class JobCoordinatorService {
     tasks.forEach(ecs::submit);
 
     int completedTasks = 0;
+    List<Job> startedJobs = new ArrayList<>();
     while (completedTasks < tasks.size()) {
       try {
         Job job = ecs.take().get();
         if (job != null) {
-          jobRepository.saveAndFlush(job);
+          startedJobs.add(job);
         }
       } catch (ExecutionException | InterruptedException e) {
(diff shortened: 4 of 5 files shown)


protos/feast/core/FeatureSet.proto
```diff
@@ -147,5 +147,6 @@ message FeatureSetMeta {
 enum FeatureSetStatus {
     STATUS_INVALID = 0;
     STATUS_PENDING = 1;
```
woop, line 149: I prefer `NOT_READY` and `PENDING`, so pending means the job is starting. It's confusing to have pending and job_starting.

zhilingc: imo `NOT_READY` could mean anything that's... not `READY`. in this context `PENDING` means pending job initialization and `JOB_STARTING`  means the job is in the initializing state, which is less confusing to me

woop: I disagree, this is super confusing to me.

> imo NOT_READY could mean anything that's... not READY

Exactly. It's more future proof.

> in this context PENDING means pending job initialization and

Then why not call it `PENDING_JOB_INITIALIZATION`? `PENDING` is meaningless.

woop: Also, it's a very leaky abstraction to have job statuses on a feature set. Ideally these statuses would be on the job, not the feature set. Although I am not sure if that kind of change is in scope given our timelines.

zhilingc: I don't think it's future proof, it's nebulous and doesn't convey any meaning to the user (or developer). I'd rather be more explicit with `PENDING_JOB_INITIALIZATION` in that case.

woop: > I don't think it's future proof, it's nebulous and doesn't convey any meaning to the user (or developer).

`NOT_READY` is the opposite of `READY`. Somehow we are happy having a `READY` status but not a `NOT_READY` status. I really fail to see how `PENDING` is somehow less nebulous than `NOT READY`. `PENDING` what? It's not clear. Remember we aren't talking about a job, something that an action should be taken on, we are talking about a schema/specification. 

I question the very notion that a job status should be attached to a feature set in the first place, but I will try and limit the scope here to only solving the problem.

Let's go ahead as is. `PENDING`. We can fix this later.

zhilingc: Exactly. `NOT_READY` is the opposite of `READY`. Which means that `PENDING` is a subset of `NOT_READY` - what's the difference between the two states then?

woop: > Exactly. NOT_READY is the opposite of READY. Which means that PENDING is a subset of NOT_READY

This seems like a non sequitur if ever I have seen one.

To me the fundamental problem here is that `PENDING` is ambiguous in meaning, and in fact we use both meanings in an overlapping way. In Feast we are using it as both `REGISTERED BUT NOT READY` as well as `JOB BEING PROCESSED`. So does `PENDING` mean pending the creation of the job, or pending the ready status of the feature set? I guess it means both, which is where the ambiguity comes in.

Then we are saying we will remove one meaning `JOB BEING PROCESSED` and give it an official status called `JOB_STARTING`, so the only function that the `PENDING` status now has is `REGISTERED BUT NOT_READY`. Hence my proposal of a `NOT READY` status.

Just to reiterate, I don't think feature sets should have job statuses. So I dont feel too strongly about this naming because it will probably be changed in the future.

zhilingc: I mean, while jobs are being processed a feature set is also in the `REGISTERED BUT NOT READY` state, which is why I'm saying `NOT_READY` is ambiguous here. :/

zhilingc: To add on, as long as feature sets remain the main source of information for users, it should have informative (read: actionable) statuses for whoever is querying for them, rather than masking them behind catch-alls for the sake of masking the existence of jobs from users. As a user i'd definitely prefer Feast being debuggable over being magical.

woop: > masking them behind catch-alls for the sake of masking

I love to sit on a Monday night reviewing people's code just so that I can `mask for the sake of masking`...

> As a user i'd definitely prefer Feast being debuggable over being magical.

Taken to an extreme, should we open up the Feast internals like our database to our users? If not, then you agree that a line should be drawn somewhere.

That line is called abstraction, not "magic".

Being conservative in revealing system internals through your API is the safer approach, unless you want to have frequent breaking changes. We've had enough of those of late. If you'd read what I had said you'd have seen that I don't believe that job statuses should be attached to feature sets, and that this is a leaky abstraction. 

The whole argument falls flat when you consider that a mapping from a feature set to a job, where the job has an appropriate status, is sufficient for debugging. 

No magic involved, just common sense.

zhilingc: Nothing hinges on the fact that `STATUS_JOB_STARTING` is named as such, so I don't really see any problem with it introducing any APIs that could break in the future. Rather, the point of the name is to be distinct from the existing status `PENDING`, which `NOT_READY` really isn't. 

And sure, users are able to query for the jobs, but the feature set should hold the first hand information as to where to look. I don't think it's common sense to look up the job if the status returned is `NOT_READY`.

woop: > the point of the name is to be distinct from the existing status PENDING, which NOT_READY really isn't.

What I was looking for in `NOT_READY` was a state that represented the fact that a spec was registered but not started. If `NOT_READY` is ambiguous for you then we can find another name (`NOT_INITIALIZED`?). 

To be clear, this is a separate problem from the one I listed when I started this thread, which was the overlap between `PENDING` and `JOB_STARTED`.

PR #715, opened 2020-05-18
Increase Docker Compose test timeout

SwampertX:
The Docker Compose tests have been failing sparingly but in a non-deterministic manner, due to timeout errors, [here](https://github.com/feast-dev/feast/runs/684970318?check_suite_focus=true#step:3:469) is an example of the latest of such failure.

NA

```release-note
NONE
```


woop:
Quick tip, PR titles go into our change log. They should be descriptive and as short as possible.

PR #718, opened 2020-05-19
Refactor runner configuration, add labels to dataflow options

zhilingc:

Adds option of setting labels for dataflow jobs, and cleans up the job configuration code a bit, also ties the runner configuration classes to the protos. Both runners extend the `RunnerConfig` class which support conversion of the config fields to beam args.

### DIFF
--- a/core/src/main/java/feast/core/job/option/RunnerConfig.java
+++ b/core/src/main/java/feast/core/job/option/RunnerConfig.java
@@ -1,0 +1,60 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright 2018-2020 The Feast Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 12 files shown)


core/src/main/java/feast/core/job/option/RunnerConfig.java
```diff
@@ -0,0 +1,60 @@
...
+package feast.core.job.option;
+
+import feast.core.util.TypeConversion;
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public abstract class RunnerConfig {
+  public String[] toArgs() throws IllegalAccessException {
```
ches, line 26: Javadoc pls ☺️

zhilingc: yessir

woop:
@zhilingc is this going to be backported to 0.5?

PR #720, opened 2020-05-20
Expand Go and Java SDK Test Coverage

mrzzy:

Expand SDKs Test Coverage:
- Added unit test to test client.go in Go SDK
- Added unit test to test `FeastClient` in Java SDK.
- Expanded unit test for `Client.get_online_features()` in Python SDK to verify the correct request is submitted.

```release-note
NONE
```


sdk/python/tests/test_client.py
```diff
@@ -201,7 +202,7 @@ def test_version(self, mocked_client, mocker):
         [pytest.lazy_fixture("mock_client"), pytest.lazy_fixture("secure_mock_client")],
     )
     def test_get_online_features(self, mocked_client, mocker):
-        ROW_COUNT = 300
+        ROW_COUNT = 1
```
woop, line 205: Why was the count changed here?

sdk/go/client_test.go
```diff
@@ -0,0 +1,97 @@
...
+	tt := []struct {
+		name    string
+		req     OnlineFeaturesRequest
+		recieve OnlineFeaturesResponse
+		want    OnlineFeaturesResponse
+		wantErr bool
+		err     error
+	}{
+		{
+			name: "valid",
```
zhilingc, line 25: This name can be more informative if you're testing for more than just a "valid" request

sdk/java/src/main/java/com/gojek/feast/FeastClient.java
```diff
@@ -39,8 +40,8 @@
 
   private static final int CHANNEL_SHUTDOWN_TIMEOUT_SEC = 5;
 
-  private final ManagedChannel channel;
-  private final ServingServiceGrpc.ServingServiceBlockingStub stub;
+  private ManagedChannel channel;
```
zhilingc, line 43: I think you should keep these as is (`final`). if you want to repalce the channel and stub with mocks for testing purposes, you might want to make a protected all args constructor instead.

sdk/java/src/main/java/com/gojek/feast/FeastClient.java
```diff
@@ -161,9 +162,9 @@ public GetFeastServingInfoResponse getFeastServingInfo() {
         .collect(Collectors.toList());
   }
 
-  private FeastClient(ManagedChannel channel) {
+  protected FeastClient(ManagedChannel channel) {
```
zhilingc, line 165: same here, leave it as private

sdk/python/tests/test_client.py
mrzzy: Corrected.

sdk/java/src/main/java/com/gojek/feast/FeastClient.java
mrzzy: I only need to mock `channel`. Mocking `stub` is not possible as its a `final` class.

sdk/java/src/main/java/com/gojek/feast/FeastClient.java
mrzzy: ^ Refer to above.

mrzzy: I only need to mock `channel`. Mocking `stub` is not possible as its a `final` class.

sdk/java/src/main/java/com/gojek/feast/FeastClient.java
mrzzy: Corrected.

sdk/go/client_test.go
mrzzy: Corrected.

sdk/go/client.go
```diff
@@ -61,17 +61,17 @@ func (fc *GrpcClient) GetOnlineFeatures(ctx context.Context, req *OnlineFeatures
 
 	// strip projects from to projects
 	for _, fieldValue := range resp.GetFieldValues() {
-		var stripFields map[string]*types.Value
+		stripFields := make(map[string]*types.Value)
```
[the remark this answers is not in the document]
mrzzy: Corrected.

sdk/java/src/main/java/com/gojek/feast/FeastClient.java
zhilingc: okayy

sdk/python/tests/test_client.py
woop: Hmmm, it is still showing as `ROW_COUNT = 1` to me. Is that expected?

mrzzy: :disappointed_relieved: Should be reverted.

PR #721, opened 2020-05-21
Remove unused Hibernate dep from Serving

ches:

Removes an unused dependency. Just a little housekeeping noticed in the context of something related. Core and Serving were using different versions of Hibernate, which I first wanted to align, then thought to myself, "wait, how does Serving use Hibernate?"

Maybe someone knows the background of what "Hibernate for formatting SQL string" meant, but grepping/`mvn dependency:analyze`/building suggests it is indeed unused.

No issue

```release-note
NONE
```

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -149,11 +149,9 @@
             <version>v1b3-rev266-1.25.0</version>
         </dependency>
 
-        <!--compile 'org.hibernate:hibernate-core:5.3.6.Final'-->
         <dependency>
             <groupId>org.hibernate</groupId>
             <artifactId>hibernate-core</artifactId>
-            <version>5.3.6.Final</version>
         </dependency>
 
         <!--compile 'org.postgresql:postgresql:42.2.5'-->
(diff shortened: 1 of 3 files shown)


woop:
Flaky tests are killing me.

ches:
I tried one rerun of the Docker Compose workflow, still got timeouts.

woop:
> I tried one rerun of the Docker Compose workflow, still got timeouts. Looks like it's failing on every recent PR.

Ah, it's supposed to be failing. Yay. So it's not flakiness, it's due to not updating the SDK to the latest version. Will have a look a bit later.

davidheryanto:
/lgtm
I think the Hibernate library was used last time to format the SQL query we generate for querying historical storage (i.e BigQuery). But yeah it seems we no longer use it.

woop:
> I tried one rerun of the Docker Compose workflow, still got timeouts. Looks like it's failing on every recent PR.

I believe I have fixed the docker compose test, but for some reason your test isnt using the code that is on the master branch. It should be cloning the `v0.5-branch` into the docker compose tests.

I think you should maybe just rebase and try again.

Might be smarter to only run these Docker Compose tests on master and not on PRs.

ches:
/retest

Docker Compose Actions workflow still appears to time out.

PR #518, opened 2020-03-07
Enforce JDK 11 for development, but build Ingestion to target Java 8

ches:
**What this PR does / why we need it**:

This is a PoC for the idea in #517 that Ingestion could be easily built for Java 8 while other components continue to target 11.

```
$ mvn clean compile

$ javap -v ingestion/target/classes/feast/ingestion/ImportJob.class | grep major
  major version: 52

$ javap -v serving/target/classes/feast/serving/ServingApplication.class | grep major
  major version: 55
```

**Which issue(s) this PR fixes**:

Addresses #517, speculatively and not completely.

**Does this PR introduce a user-facing change?**:

```release-note
Feast Ingestion is again built for Java 8, for more certain Beam runner compatibility.
```

PR #451, opened 2020-01-30
Upgrade to Java 11

Yanson:
**What this PR does / why we need it**:

Upgrade Feast to next LTS - Java 11 - since Java 8 is outdated and no longer supported.

Our internal build and testing of Feast has been on Java 11 all along, so it's fair to say there shouldn't be any problems. I'm not able to fully test because 1) slight variation in base-images for building/packaging, 2) we use a different Dockerfile that uses pre-built Jars from another step rather than a 'builder' pattern

**Which issue(s) this PR fixes**:
NONE

**Does this PR introduce a user-facing change?**:
```release-note
Feast now requires a Java 11 runtime. JVM options should be set accordingly.
```

PR #722, opened 2020-05-21
Restore Feast Java SDK and Ingestion compatibility with Java 8 runtimes

ches:

Restores Feast Ingestion compatibility with Java 8 runtimes.

With the addition of the Storage APIs and connector implementations, we did not specify that their build should target Java 8 bytecode. Because Ingestion depends on them and bundles them in an uber jar, we regressed on keeping Ingestion compatible with Java 8 runtimes according to #518.

This could be confirmed prior to this change on master with, for example:

    $ mvn -pl ingestion clean compile
    $ javap -v storage/connectors/redis/target/classes/feast/storage/connectors/redis/writer/RedisCustomIO.class | grep major
      major version: 55

I take this as proof that a blacklist approach with 11 as the default in our root POM is error prone, so this change flips it so that 8 is default and our leaf applications that are able use 11 opt into it in whitelist fashion.

This should be backported to `v0.5-branch`, I believe.

None, perhaps should have filed a bug separately, but this can be bug and fix in one ☺️ 

```release-note
- Restores Feast Ingestion compatibility with Java 8 runtimes
- Restores Feast Java SDK compatibility with Java 8 runtimes
```

### DIFF
--- a/ingestion/pom.xml
+++ b/ingestion/pom.xml
@@ -33,15 +33,6 @@
     <plugins>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-compiler-plugin</artifactId>
-        <configuration>
-          <!-- To ensure Ingestion remains compatible for Java 8-only Beam runners -->
-          <release>8</release>
-        </configuration>
-      </plugin>
-
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-shade-plugin</artifactId>
         <version>3.2.1</version>
         <executions>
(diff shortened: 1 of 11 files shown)


ingestion/pom.xml
```diff
@@ -84,6 +75,31 @@
...
       </plugin>
 
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <executions>
+          <!--
+            We want to keep Ingestion compatible with Java 8 runtime for Beam runners, #518.
+            Unfortunately this doesn't seem to work on Reactor modules, though:
+            https://github.com/mojohaus/mojo-parent/issues/85
```
ches, line 85: I hoped that this would fail before applying the fix, verifying that it could catch this problem, but it didn't. I kept this in, because if it does fail on third-party dependencies, we would probably like to know if/when it changes in the future. This rule is from `extra-enforcer-rules` which was added #451 but not actually used for anything so far, I think.

ches:
/hold

Tests need compilation fixes

sdk/java/pom.xml
```diff
@@ -85,6 +85,14 @@
 
   <build>
     <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <release>11</release>
```
ches, line 92: Somewhat of a question whether SDK should build for 8 or 11 perhaps. It was implicitly 11 since that become default in the root, so I preserved that, but I can also change it now if there is a desire to. Needs a small code change in tests I think.

zhilingc: haha. @khorshuheng just ran into this during our upgrade today.

ches: I'll take that as a real-world indicator that yes, we should keep it Java 8-compatible. Pushed this change.

ches:
Rebased so #725 should fix the build.

ches:
/retest

gamble

woop:
It's not clear why these e2e tests are failing inside Docker Compose, yet they work when run outside of it. I have removed the from PRs and allow them only to run on master for the time being. Unfortunately I cant remove it from the list of checks, so we probably have to force merge once you are ready.

Created to track the issue #726

ches:
> Unfortunately I cant remove it from the list of checks, so we probably have to force merge once you are ready.

This is ready from my side, if it makes sense to everyone else. Ditto for #721 blocked by the misbehaved check.

PR #729, opened 2020-05-22
Fix Go and Java SDK Regressions

mrzzy:
This is a split off containing solely the bug fixes of PR #720 

Fixes SDK bugs introduced in PR #693 & #700 
- Go SDK crashes due to trying to add to a `nil` map. (PR #693 ).
- Go SDK obmitted entity values when in values returned from `GetOnlineFeatures()` (PR #693 )
- Java SDK's `Row.set()` rejects `feast.proto.types.ValueProto.Value` as its hardcoded class path to `feast.types.ValueProto.Value`.( PR #700 )

```release-note
Fixed Go SDK error due to trying to add to a nil map.
Fixed Go SDK omitted entity values when get online features.
Fixed Java SDK's Row.set() does not accept feast Value type
```

### DIFF
--- a/sdk/go/client.go
+++ b/sdk/go/client.go
@@ -52,7 +52,7 @@ func (fc *GrpcClient) GetOnlineFeatures(ctx context.Context, req *OnlineFeatures
 	resp, err := fc.cli.GetOnlineFeatures(ctx, featuresRequest)
 
 	// collect unqiue entity refs from entity rows
-	var entityRefs map[string]struct{}
+	entityRefs := make(map[string]struct{})
 	for _, entityRows := range req.Entities {
 		for ref, _ := range entityRows {
 			entityRefs[ref] = struct{}{}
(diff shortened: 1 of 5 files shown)


ches:
> This is a split off containing solely the bug fixes of PR #725

Is the correct issue number reference? I'm not finding where these come from and it definitely doesn't look like #725 ☺️

ches:
> **Does this PR introduce a user-facing change?**:
> 
> ```
> NONE
> ```

Bug fixes should definitely have a changelog IMO, although I guess we currently use PR titles rather than this field. The PR number references where issues were introduced are super helpful, the extra mile would be if we also note the Feast version number where a bug was introduced, if it has been released yet. Good for users to quickly know if they're affected, good for us to know where to backport.

woop:
> > **Does this PR introduce a user-facing change?**:
> > ```
> > NONE
> > ```
> 
> Bug fixes should definitely have a changelog IMO, although I guess we currently use PR titles rather than this field. The PR number references where issues were introduced are super helpful, the extra mile would be if we also note the Feast version number where a bug was introduced, if it has been released yet. Good for users to quickly know if they're affected, good for us to know where to backport.

Yea we are using the PR titles right now. Ideal case would be to use the release note if available, and fall back to the PR. I dont think our changelog generator supports that right now though. So we might need to update our release process.

By the way, we can enforce release notes using a Prow plugin as well: http://prow.feast.ai/plugins

woop:
@mrzzy mind updating the release note so that I can merge this?

PR #731, opened 2020-05-22
Fix project argument for feature set describe in CLI

### LINKED ISSUE #727: feast cli feature-sets describe broken
## Expected Behavior 
`feast feature-sets describe <name>` cannot be used because it doesn't allow the user to set a project, nor does it default to any value.

## Steps to reproduce
Call `feast feature-sets describe <some_feature_set>`

### Specifications

- Version: 0.5.0

## Possible Solution

The method calls `fs = feast_client.get_feature_set(name=name)`. Since no project is provided to `get_feature_set`, a default project needs to be set in the client. 

Either 
1. Allow users to pass feature set ids with projects specified (`project/feature_set`) or
2. Allow users to set a default project.

The method should fall back to a default project (`default`) should one not be provided.

terryyylim:
`feast feature-sets describe <name>` cannot be used because it doesn't allow the user to set a project, nor does it default to any value.

Fixes #727 

```release-note
Users can now pass in --project parameter to retrieve information about a feature set using cli.
```

### DIFF
--- a/sdk/python/feast/cli.py
+++ b/sdk/python/feast/cli.py
@@ -156,12 +156,20 @@ def feature_set_create(filename):
 
 @feature_set.command("describe")
 @click.argument("name", type=click.STRING)
-def feature_set_describe(name: str):
+@click.option(
+    "--project",
+    "-p",
+    help="Project that feature set belongs to",
+    type=click.STRING,
+    default="default",
+)
+def feature_set_describe(name: str, project: str):
     """
     Describe a feature set
     """
     feast_client = Client()  # type: Client
-    fs = feast_client.get_feature_set(name=name)
+    fs = feast_client.get_feature_set(name=name, project=project)
+
     if not fs:
         print(f'Feature set with name "{name}" could not be found')
         return


sdk/python/feast/cli.py
```diff
@@ -155,13 +155,23 @@ def feature_set_create(filename):
...
 @feature_set.command("describe")
-@click.argument("name", type=click.STRING)
-def feature_set_describe(name: str):
+@click.option(
+    "--name", "-n", help="Name of feature set", required=True, type=click.STRING
+)
+@click.option(
+    "--project",
+    "-p",
+    help="Project where feature set can be found",
```
woop, line 164: Perhaps `Project that feature set belongs to`?

sdk/python/feast/cli.py
```diff
@@ -155,13 +155,23 @@ def feature_set_create(filename):
 
 
 @feature_set.command("describe")
-@click.argument("name", type=click.STRING)
-def feature_set_describe(name: str):
+@click.option(
+    "--name", "-n", help="Name of feature set", required=True, type=click.STRING
```
woop, line 159: I think a positional argument for the name makes a bit more sense here. Did you have a reason for making this change?

woop:
Oh nice, thanks for this @terryyylim!

I think `Fix project argument for feature set describe in CLI` would be a better title. The titles are used for our change log so we want them to be human readable.

Also, you need to add a `kind` label to all PRs. This one requires `kind/bug`.

sdk/python/feast/cli.py
terryyylim: Not exactly other than offering the `help` option to explain what name refers to. (though I guess it's self-explanatory)

sdk/python/feast/cli.py
terryyylim: 👌 i'll make that change

woop:
@terryyylim Remember to update the release note. Other than that it looks good.

PR #734, opened 2020-05-23
v0.3 backport: Add Java coverage reporting

ches:
**What this PR does / why we need it**:

Backports #686 to `v0.3-branch`. Also backports #410, which was a final missing piece for v0.3 after #407—this fixes the Prow job for `publish-docker-images` on the branch.

I have possibly a few of these housekeeping-ish things that I want to bring in internally, and as much as possible / whenever valuable for mainline Feast I'm trying to do things as *upstream master => upstream backport => bring backport in-house*. If PR reviews and so forth become a burden though, if you aren't supporting any v0.3 deployments yourselves anymore, I'm happy for us to "formally" take ownership of the maintenance branch in whatever ways make sense.

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #737, opened 2020-05-23
v0.3 backport: Add feature and feature set labels

ches:

Backports #536 to v0.3, courtesy of @suwik's work.

We don't typically backport features, but I'm proposing such in this case for practical reasons:

- It's a small, useful feature addition and its scope is confined mostly to registry metadata.
- We already use this backport internally with v0.3.
- Sticky matter of logistics on our (Agoda's) side: since we use this, it means we internally publish a v0.3-based version of `datatypes-java` with the labels fields added to the protos. Our Feast SDK is hosted on github.com pending being open sourced, and its GitHub Actions CI build cannot reach our behind-the-firewall Artifactory to resolve the private dependency version. We will live on v0.3 a little while longer, so I'd like to not have to relegate "v0.3 with labels" to an SDK branch where CI isn't usable. We could alternatively publish the datatypes to Maven Central under our own group ID, but it's rather roundabout, so I wanted to try reception in mainline Feast before turning to that.

This is not just housekeeping, but my note about maintenance on #734 stands.

The largest contributor to the size of this PR is Python code generated from protobufs. In v0.3 the generated code is still tracked in Git, and it seems like it was stale before this change, because running the codegen produced a lot more changes than just the label additions.

```release-note
Added key/value labels to feature and feature set specs for Feast v0.3, as in #536 for v0.5
```


core/src/test/java/feast/core/service/SpecServiceTest.java
```diff
@@ -367,6 +355,100 @@ public void applyFeatureSetShouldNotCreateFeatureSetIfFieldsUnordered()
...
+    // follow the order in the request
+    List<FeatureSpec> appliedFeatureSpecs =
+        new ArrayList<>(appliedFeatureSetSpec.getFeaturesList());
+    appliedFeatureSpecs.sort(Comparator.comparing(FeatureSpec::getName));
+
+    List<Map<String, String>> featureSpecsLabels =
+        featureSpecs.stream().map(e -> e.getLabelsMap()).collect(Collectors.toList());
+    assertThat(appliedEntitySpecs, equalTo(entitySpecs));
+    assertThat(appliedFeatureSpecs, equalTo(featureSpecs));
+    assertThat(featureSpecsLabels, equalTo(featureLabels));
```
ches, line 426: I just noticed this when raising this PR: I believe that the `featureSpecsLabels` here should be assigned from `appliedFeatureSpecs.stream()` instead of `featureSpecs.stream()`, otherwise we are asserting on test setup code instead of covering actual roundtrip through the feature set being applied. I've asked @suwik to double-check my reasoning.

I've made that change locally and the test passes, so I believe it proves there is no functional issue with feature-level labels, we just aren't covering what we want to with the test. I'll include the update in this PR if @suwik corroborates, but the same change should be forward-ported to master, so I want to track that.

ches:
`publish-docker-images` will have the same problem as #734 for `v0.3-branch`, looking into that for both now.

woop:
@ches do you want this merged or do you want to wait for @suwik? I have heavy fingers and they tend to fall on `/lgtm` if I don't see problems with a PR.

ches:
😄 

I went ahead and pushed the change that I think makes sense, so we can forward-port the commit. I'm happy to go ahead with it so I can get a few more things done by Monday toward an 0.3.8 release we can use.

PR #739, opened 2020-05-23
v0.5 backport: Restore SDK and Ingestion compatibility with Java 8 runtimes

ches:

Backports #722 to v0.5.

```release-note
- Restores Feast Ingestion compatibility with Java 8 runtimes
- Restores Feast Java SDK compatibility with Java 8 runtimes
```

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -33,6 +33,14 @@
     <build>
         <plugins>
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <release>11</release>
+                </configuration>
+            </plugin>
+
+            <plugin>
                 <groupId>org.jacoco</groupId>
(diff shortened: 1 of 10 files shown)


ches:
Hmm probably we also need to backport #725, and perhaps #728.

ches:
/retest

Maven Central connection blip it seems

woop:
Those tests dont add any value to you though, I think. They are testing the current master branch. I've removed them from all non-master branches and all PRs. So they only trigger on a push to master and we can track them through the front page badge.

ches:
> Wait a minute, they shouldn't have run on a PR in the first place. Strange.

I'd guess it's that b8456c1bdc1a4 isn't on `v0.5-branch`—my interpretation of [this] is that a pull request event is probably associated with the target branch as its Git ref, so the workflow configurations from the branch head will apply.

[this]: https://help.github.com/en/actions/reference/events-that-trigger-workflows#about-workflow-events

PR #740, opened 2020-05-25
Labels unit tests improvement

suwik:
The last assertion was only asserting against the object constructed in the test code and not the object created by production code. This PR fixes it. 
Also removed entities as I don't think they're relevant to this test.

NONE

no 

```release-note
NONE
```


woop:
You arent properly introduced to Feast without a `test-end-to-end-batch` failure on your name.

PR #748, opened 2020-05-27
Redis host fix

AnujaVane:
**Why we need it**: 
1. The online-serving.yml file does not effectively set the host to redis (although it has the correct value in the config) but instead tries to connect to localhost and errors out. 
2. The sample_redis_config.yml file specified in the development guide does not exist and a default config seems to point to it because of which it errors.

**What this PR does**:
1.  Creates a sample_redis_config.yml from the one mentioned in the development guide. Changes the host from "localhost" to "redis".
2. Mounts the yml file to the host filesystem.
3. Sets the config path to point to it.

**Which issue(s) this PR fixes**: Redis connection 

```release-note

```

### DIFF
--- a/infra/docker-compose/docker-compose.online.yml
+++ b/infra/docker-compose/docker-compose.online.yml
@@ -5,6 +5,7 @@ services:
     image: ${FEAST_SERVING_IMAGE}:${FEAST_VERSION}
     volumes:
       - ./serving/${FEAST_ONLINE_SERVING_CONFIG}:/etc/feast/application.yml
+      - ./serving/sample_redis_config.yml:/etc/feast/sample_redis_config.yml
     depends_on:
       - redis
     ports:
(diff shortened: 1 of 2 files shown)


woop:
@AnujaVane are you using Feast 0.4 or Feast 0.5. Feast 0.5 doesn't use external store configuration any more. I think the problem here is that the development guide is out of date.

woop:
We have a docker compose test that is supposed to ensure that the public configuration for docker compose is functional and working. Can you confirm whether running that test works for you locally? https://github.com/feast-dev/feast/actions?query=workflow%3A%22docker+compose+tests%22+branch%3Amaster

AnujaVane:
@woop I could get Feast up and running after setting the FEAST_HOME_DIR environment variable. I'm gonna go ahead and take the PR down. Thanks!

PR #749, opened 2020-05-28
Enable Mypy static type checking

### LINKED ISSUE #594: Enable Mypy linter on the lint-python side

SwampertX:
Supercedes #593.

Fixes #594 .

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/client.py
+++ b/sdk/python/feast/client.py
@@ -22,7 +22,7 @@ import time
 import uuid
 from collections import OrderedDict
 from math import ceil
-from typing import Dict, List, Optional, Set, Tuple, Union
+from typing import Dict, List, Optional, Tuple, Union
 
 import grpc
 import pandas as pd
@@ -48,7 +48,9 @@ from feast.core.CoreService_pb2 import (
     ApplyFeatureSetRequest,
     ApplyFeatureSetResponse,
     ArchiveProjectRequest,
+    ArchiveProjectResponse,
     CreateProjectRequest,
+    CreateProjectResponse,
     GetFeastCoreVersionRequest,
     GetFeatureSetRequest,
     GetFeatureSetResponse,
--- a/sdk/python/feast/client.py
+++ b/sdk/python/feast/client.py
@@ -22,7 +22,7 @@ import time
 import uuid
 from collections import OrderedDict
 from math import ceil
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Dict, List, Optional, Set, Tuple, Union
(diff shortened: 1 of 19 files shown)


sdk/python/feast/client.py
```diff
@@ -311,8 +311,8 @@ def list_projects(self) -> List[str]:
             List of project names
 
         """
-        self._connect_core()
-        response = self._core_service_stub.ListProjects(
+        self._connect_core()  # ensures self._core_service_stub is defined
+        response = self._core_service_stub.ListProjects(  # type: ignore
```
woop, line 315: Why are you using `type: ignore` so frequently? Can we set these types?

SwampertX: `_core_service_stub` has type `Optional[CoreServiceStub]`, which Mypy throws error when we call it as it could be `None`. However `_connect_core()` guarantees that `_core_service_stub` is non-null, but Mypy could not detect.

SwampertX: I wrote it in the commit message here too: https://github.com/feast-dev/feast/pull/749/commits/ffbe15bf5a95b658e388387c7f4220230a82b2fb

SwampertX:
I need some help in these areas:
1. https://github.com/feast-dev/feast/pull/749/checks?check_run_id=726605113#step:5:68 this error pops up quite often, and I could not figure out where `FeatureTypeValue` comes from, since [the attribute `feature.type`](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/feature_set.py#L729) should have type `FeatureType` according to [the protobuf](https://github.com/feast-dev/feast/blob/master/protos/tensorflow_metadata/proto/v0/schema.proto#L148). 
  Same errors in https://github.com/feast-dev/feast/pull/749/commits/6e024a47a3ed7d3c3529eec3c916abe18413089f are solved by casting (needs review/advice), however casting cannot solve the current error.

2.  https://github.com/feast-dev/feast/pull/749/commits/36940588f04b78d574646cbf8e910024a603a4e9. I removed `EntitySpec` entirely. `EntitySpec` is listed as one of the possible types of the parameter `feature` but is found by Mypy to not have the attributes/methods as the other types. I henced removed the type as a possiblility.

PR #751, opened 2020-05-28
Remove usage of parallel stream for feature value map generation

khorshuheng:
Feaute value map is not thread safe. Using parallel stream here can leads to racing condition, which results in incorrect feature value map being computed.

```release-note
Fix the bug where feature value map is generated incorrectly intermittently due to racing condition, when the feature is stale.
```


woop:
@khorshuheng this should probably be backported to 0.5 as well right?

PR #752, opened 2020-05-28
Merge Redis cluster connector with Redis connector

pyalex:
Huge part of redis-cluster storage code was just copy-pasted from redis. In order to simplify further development we need to consolidate those storages.

Fixes #

```release-note

```


storage/connectors/redis/src/main/java/feast/storage/connectors/redis/writer/RedisFeatureSink.java
```diff
@@ -60,29 +71,42 @@ public static Builder builder() {
...
-    RedisClient redisClient =
-        RedisClient.create(RedisURI.create(getRedisConfig().getHost(), getRedisConfig().getPort()));
-    try {
-      redisClient.connect();
-    } catch (RedisConnectionException e) {
-      throw new RuntimeException(
-          String.format(
-              "Failed to connect to Redis at host: '%s' port: '%d'. Please check that your Redis is running and accessible from Feast.",
-              getRedisConfig().getHost(), getRedisConfig().getPort()));
+    if (getRedisConfig() != null) {
```
woop, line 83: Should we handle the null case?

pyalex: Not sure, how exactly. I just mimiced the behavior of RedisFeatureSink and RedisClusterFeatureSink. So the latest do nothing in this method. So when RedisConfig is null I do the same.
Do you want to have exception here?

pyalex: added throw

Issue #646, opened 2020-04-23
What is the need for using unique Kafka Consumer Group for ingestion job?


bharathelangovan:
We are trying to figure out why a new consumer group is created when an ingestion job is started. This means that if there is a gap between the termination and starting of a Dataflow job there will be a data within the Kafka topic that is not ingested
https://github.com/gojek/feast/blob/196f7112f0c4ba58f2417e0c9856dac07435e0be/ingestion/src/main/java/feast/ingestion/transform/ReadFromSource.java#L86
We would like to know if this was intentional and if so what are the ways to avoid data loss?

ches:
Glad someone asked, we’ve patched this internally and I hadn’t thought to raise the question upstream yet.

Essentially we base the consumer group names on `Source` ID and the target `Store` name. It’s not quite a cherry-pick for me to bring it to an upstream PR because there’s one extra element specific to our use that probably is not generally appropriate, but it’s a small patch to extract if no one sees problems with the approach.

davidheryanto:
Thanks for bringing this up.

I think early on Feast Core didn't have the ability to ensure jobs are **always** running to ingest the FeatureSets applied.

That is, if a job is stopped, it won't be automatically restarted until the FeatureSet is reapplied again.

In such situation, it is possible that a job is stopped, then only after a "long" period of time it is started again (when the FeatureSet is reapplied).

For a FeatureSet with a high throughput Source, it is likely that during the period the job is stopped, the lag becomes incredibly high that if the job is to continue from the last offset it may not be able to keep up. Also, this very old data may have less value than the most recent one. But because the job cannot catch up, it cannot ingest fresh data as a result. That's why when a job is stopped, it is assigned a new consumer group id when started again so it will read from the latest offset in the Source.

However, now that Feast Core always ensures there are jobs running to ingest the FeatureSet applied, I think it makes sense for us to change the behaviour such that the same consumer group id is used when the job is stopped, and then automatically restarted by Feast Core. Because Feast
Core will very quickly attempt to restart the job, there will be no problem of the job not being able to catch up with the Source.

> As a side note, currently if a currently running job is restarted via an **update** i.e. reapplying an existing FeatureSet with same the Source and Store but new features, the **same** consumer group id will be retained. (This is different from stopping a job and then applying the FeatureSet because the updated FeatureSet is applied when the job is **still** running). And I think currently it only works with DataflowRunner.

woop:
[...]

PR #757, opened 2020-05-29
Upgrade ingestion to allow for in-flight updates to feature sets for sinks

pyalex:

There are several improvements:

* Sink API accepts `PCollection<FeatureSetSpec>` and both BQ and redis can change their processing according to new specs in-flight
* BQ sink writes through file loads (to gs) - that enables schema update & also might be less expensive
* BQ sink produces sucessful inserts stream - enables metrics on this sink
* Dataflow JobManager stop+start job when specs is updated. This is caused by the bug in Dataflow that is losing state of PCollectionView on update.
* Kafka Reader in ingestion now has stable group id (across restarting jobs)

Unblock #761 

```release-note

```


core/src/main/java/feast/core/job/dataflow/DataflowJobManager.java
```diff
@@ -137,38 +137,15 @@ public Job startJob(Job job) {
...
-      return job;
-    } catch (InvalidProtocolBufferException e) {
-      log.error(e.getMessage());
-      throw new IllegalArgumentException(
-          String.format(
-              "DataflowJobManager failed to UPDATE job with id '%s' because the job"
-                  + "has an invalid spec. Please check the FeatureSet, Source and Store specs. Actual error message: %s",
-              job.getId(), e.getMessage()));
-    }
+    abortJob(job.getExtId());
```
woop, line 147: Do we really need the `updateJob()` still if it can be replaced with `abortJob`?

core/src/main/java/feast/core/job/dataflow/DataflowJobManager.java
```diff
@@ -283,14 +260,17 @@ private ImportOptions getPipelineOptions(
     pipelineOptions.setJobName(jobName);
     pipelineOptions.setFilesToStage(
         detectClassPathResourcesToStage(DataflowRunner.class.getClassLoader()));
-
+    if (pipelineOptions.getTempLocation() == null && sink.getBigqueryConfig() != null) {
+      pipelineOptions.setTempLocation(sink.getBigqueryConfig().getStagingLocation());
```
woop, line 264: Is there a way to make this more generic (non-BQ specific)?

infra/scripts/test-templates/values-end-to-end-batch-dataflow.yaml
```diff
@@ -30,7 +30,7 @@ feast-core:
             subnetwork: regions/$GCLOUD_REGION/subnetworks/$GCLOUD_SUBNET
             maxNumWorkers: 1
             autoscalingAlgorithm: THROUGHPUT_BASED
-            usePublicIps: false
+            usePublicIps: true
```
woop, line 33: How are we using public IPs here?

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/writer/BigQuerySinkHelpers.java
```diff
@@ -17,34 +17,30 @@
...
-  public GetTableDestination(String projectId, String datasetId) {
-    this.projectId = projectId;
-    this.datasetId = datasetId;
-  }
-
-  @Override
-  public TableDestination apply(ValueInSingleWindow<FeatureRow> input) {
-    String[] splitName = input.getValue().getFeatureSet().split("/");
+    if (splitName.length != 2) {
+      throw new RuntimeException("FeatureRow key is invalid. Must be project/feature-set");
```
woop, line 28: If the `project` is not set then it should be configured to write to the `default` project. Meaning `project` is optional.

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/writer/BigQueryWrite.java
```diff
@@ -16,92 +16,201 @@
...
-            ParDo.of(
-                new DoFn<FeatureRowProto.FeatureRow, FeatureRowProto.FeatureRow>() {
-                  @ProcessElement
-                  public void processElement(ProcessContext context) {}
-                }));
-
     return WriteResult.in(input.getPipeline(), successfulInserts, failedElements);
   }
+
+  private PCollection<FeatureRow> mergeInputWithResult(
```
woop, line 142: Would appreciate a few comments on this method.

woop:
@pyalex would you mind pointing out all the code that we pulled up from Apache Beam? I would want to make sure that we give them proper attribution and that we don't put our license on their work.

core/src/main/java/feast/core/job/dataflow/DataflowJobManager.java
pyalex: I guess, there's nothing bad in having updateJob in the JobManager interface. There's still a chance that Dataflow update would be fixed in future and we'll switch back and also there could be other backends which support update w/o stop+start

core/src/main/java/feast/core/job/dataflow/DataflowJobManager.java
pyalex: we actually don't need this part at all, since now it's
`@NotBlank public String tempLocation;`

pyalex:
> @pyalex would you mind pointing out all the code that we pulled up from Apache Beam? I would want to make sure that we give them proper attribution and that we don't put our license on their work.

sure, it's just one file https://github.com/feast-dev/feast/pull/757/files#diff-27d2cb5b830d488116fc8c364639b498

And I also added exclusion to spotless configuration https://github.com/feast-dev/feast/pull/757/files#diff-600376dffeb79835ede4a0b285078036R402 to avoid license addition. Not sure if there's better way to do it? @woop

pyalex:
> I've left some comments, but they aren't fundamental changes. Just small nitpicks. Overall the PR looks good to me. One thing that I would ask for is improved JavaDoc coverage. This part of the code base can be hard to parse, so comments can only help the next person.

Added some details on big query writing flow as well

core/src/main/resources/application.yml
```diff
@@ -38,7 +38,8 @@ feast:
     runners:
       - name: direct
         type: DirectRunner
-        options: {}
+        options:
+          tempLocation: gs://bucket/tempLocation
```
woop, line 42: Is this supposed to be under DataflowRunner? Or added here: https://github.com/feast-dev/feast/blob/master/protos/feast/core/Runner.proto#L33

docs/roadmap.md
```diff
@@ -1,5 +1,24 @@
 # Roadmap
 
+## Feast 0.6 \(Feature Release\)
```
woop, line 3: Not sure where this came from, bad rebase?

core/src/main/resources/application.yml
pyalex: it affects Direct as well, and was added here https://github.com/feast-dev/feast/pull/757/files#diff-ed1aa58bf58c59fe211c82b6df89182eR37

docs/roadmap.md
pyalex: Apparently, I don't see any difference in my branch vs master on this file. So have no idea why it shows it here

core/src/main/resources/application.yml
woop: Oops, I missed that. Thanks for pointing it out.

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/writer/BigQueryFeatureSink.java
```diff
@@ -16,61 +16,99 @@
...
    * @param config {@link BigQueryConfig}
    * @param featureSetSpecs
-   * @return {@link BigQueryFeatureSink.Builder}
+   * @param bigQueryServices {@link BigQueryServices} - internal beam BigQuery wrapper
+   * @param bigQueryOptions {@link ValueProvider} value provider of BigQuery - from gcloud sdk
+   * @return
    */
-  public static FeatureSink fromConfig(
-      BigQueryConfig config, Map<String, FeatureSetProto.FeatureSetSpec> featureSetSpecs) {
+  static FeatureSink fromConfig(
```
woop, line 67: Would it be possible to use the `Builder` directly instead of adding this `fromConfig`?

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/writer/BigQueryFeatureSink.java
```diff
@@ -16,61 +16,99 @@
...
+  @Nullable
+  public abstract ValueProvider<BigQuery> getBigQueryOptions();
+
+  public abstract PCollection<KV<String, FeatureSetProto.FeatureSetSpec>> getFeatureSetSpecs();
 
   /**
-   * Initialize a {@link BigQueryFeatureSink.Builder} from a {@link BigQueryConfig}. This method
-   * initializes a {@link BigQuery} client with default options. Use the builder method to inject
-   * your own client.
+   * Mostly for internal use in tests Same as public version but allows to specify bigQuery clients
```
woop, line 58: This is very nitpicky, but can you fix the casing/formatting of this sentence please.

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/writer/BigQuerySinkHelpers.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+package feast.storage.connectors.bigquery.writer;
+
+import com.google.api.services.bigquery.model.TimePartitioning;
+import com.google.cloud.bigquery.DatasetId;
+import org.apache.beam.sdk.io.gcp.bigquery.TableDestination;
+
+public class BigQuerySinkHelpers {
+  /**
+   * Generating BQ table destination from dataset reference and featuresSet's project and name If
```
woop, line 25: You are missing a period before the `If`

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/writer/BigQuerySinkHelpers.java
```diff
@@ -0,0 +1,59 @@
...
+   * project is undefined "default" would be selected
+   *
+   * @param dataset {@link DatasetId} reference to bq project and dataset
+   * @param featureSetKey featureSet key in format project-name/set-name
```
woop, line 29: `featureSetKey featureSet key in format project-name/set-name` > `featureSetRef Feature set reference with format <project>/<feature-set-name>`

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/writer/BigQueryWrite.java
```diff
@@ -16,92 +16,227 @@
...
     this.destination = destination;
+    this.schemas = schemas;
+  }
+
+  public BigQueryWrite withTriggeringFrequency(Duration triggeringFrequency) {
+    this.triggeringFrequency = triggeringFrequency;
+    return this;
+  }
+
+  public BigQueryWrite withTestServices(BigQueryServices testServices) {
```
woop, line 74: Should this be called something like `withBigQueryService` in case it is used outside the testing context?

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/writer/BigQueryWrite.java
```diff
@@ -16,92 +16,227 @@
...
+              public void process(ProcessContext c) {
+                CoGbkResult result = c.element().getValue();
+                result
+                    .getAll(successTag)
+                    .forEach(success -> result.getAll(inputTag).forEach(c::output));
+              }
+            }));
+  }
+
+  private class FeatureDynamicDestinations extends DynamicDestinations<FeatureRow, String> {
```
woop, line 216: Can we get a JavaDoc here please.

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/writer/BigQueryWrite.java
pyalex: It's basically same method as here https://github.com/apache/beam/blob/b12a38aa2156c2652937669aeb2e46be15eca06c/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIO.java#L2388
and there's no other (public) implementation except the test one, and it was done intentionally

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/writer/BigQuerySinkHelpers.java
```diff
@@ -0,0 +1,59 @@
...
+   */
+  public static TableDestination getTableDestination(DatasetId dataset, String featureSetKey) {
+    String[] splitName = featureSetKey.split("/");
+    String projectName, setName;
+
+    if (splitName.length == 2) {
+      projectName = splitName[0];
+      setName = splitName[1];
+    } else {
+      projectName = "default";
```
Yanson, line 40:
```suggestion
      projectName = Poject.DEFAULT_NAME;
```

pyalex: unfortunately Project model is in core, and there's reverse dependency core -> ingestion. maybe, it would be a good idea to move all models to `datatypes`, but I'm not sure it should be done within this PR. what about just local class constant for now?

woop: > what about just local class constant for now?

This makes sense. I dont think it should be a part of this PR, but it's definitely a tech debt that should be cleared up.

ches:
> * Dataflow JobManager stop+start job when specs is updated. This is caused by the bug in Dataflow that is losing state of PCollectionView on update.

Is there some further reference to what this about, that can be included in the description?

ches:
> * Kafka Reader in ingestion now has stable group id (across restarting jobs)

Does this obviate #760 and close #646? It'd be ideal if this sort of change could be unbundled from a big PR like this—easier to track the status/closure of issues, easier to backport fixes versus enhancements, etc.

woop:
> > * Kafka Reader in ingestion now has stable group id (across restarting jobs)
> 
> Does this obviate #760 and close #646? It'd be ideal if this sort of change could be unbundled from a big PR like this—easier to track the status/closure of issues, easier to backport fixes versus enhancements, etc.

I believe it does close those, will let @pyalex confirm. Yea, ideally those changes would not be bundled into one PR.

pyalex:
@ches @woop Correct, issue with kafka consumer group should be solved and it was rather side effect not a main purpose of this PR

ches:
> it was rather side effect not a main purpose of this PR

Understandable, glad for a fix getting mainlined either way!

PR #760, opened 2020-06-01
Use stable Kafka consumer group ID for job

### LINKED ISSUE #646: What is the need for using unique Kafka Consumer Group for ingestion job?

kishanpradhan:

Fixes #646 

```release-note

```

### Design
This PR will have the functionality to have same consumer group for same job even after restart. Currently you can set consumer group from the ImportJob and it is optional.

But from issue discussion,

> For a FeatureSet with a high throughput Source, it is likely that during the period the job is stopped, the lag becomes incredibly high that if the job is to continue from the last offset it may not be able to keep up. Also, this very old data may have less value than the most recent one. But because the job cannot catch up, it cannot ingest fresh data as a result. That's why when a job is stopped, it is assigned a new consumer group id when started again so it will read from the latest offset in the Source.

the current PR will not be that useful.

### Solutions
1. Use labels to  point out that one feature set will use different consumer groups at all time or same consumer groups. User can add label like `high_throughput: true` and based on it we will set the consumer group from ImportJob.
 ```  
 labels {
    key: "high_throughput"
    value: "true"
  }
 ```
2. We can add consumer group value to Kafka config while registering a feature set. We make this consumer group as optional so if not provided we always use the new consumer group as it is now. The Kafka config will look like
  ```
  kafka_source_config {
      bootstrap_servers: "localhost:9092"
      topic: "feast-features"
      consumer_group: "some_group_name"
    }
 ```

### DIFF
--- a/ingestion/src/main/java/feast/ingestion/ImportJob.java
+++ b/ingestion/src/main/java/feast/ingestion/ImportJob.java
@@ -120,6 +120,8 @@ public class ImportJob {
               "ReadFeatureRowFromSource",
               ReadFromSource.newBuilder()
                   .setSource(source)
+                  .setConsumerGroupId(
+                      ReadFromSource.generateConsumerGroupId(source.getType(), store.getName()))
                   .setSuccessTag(FEATURE_ROW_OUT)
                   .setFailureTag(DEADLETTER_OUT)
                   .build());
--- a/ingestion/src/main/java/feast/ingestion/transform/ReadFromSource.java
+++ b/ingestion/src/main/java/feast/ingestion/transform/ReadFromSource.java
@@ -23,6 +23,7 @@ import feast.core.SourceProto.SourceType;
 import feast.ingestion.transform.fn.KafkaRecordToFeatureRowDoFn;
 import feast.storage.api.writer.FailedElement;
 import feast.types.FeatureRowProto.FeatureRow;
+import java.util.Optional;
 import org.apache.beam.sdk.io.kafka.KafkaIO;
 import org.apache.beam.sdk.transforms.PTransform;
 import org.apache.beam.sdk.transforms.ParDo;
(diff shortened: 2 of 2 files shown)


woop:
> We can add consumer group value to Kafka config while registering a feature set. We make this consumer group as optional so if not provided we always use the new consumer group as it is now. The Kafka config will look like

Having the consumer group defined on the source seems a bit dangerous to me. This means that multiple different serving deployments will share a consumer group, which would lead to missing data. Unless I am missing something, this doesnt seem viable.

Yanson:
I agree setting the consumer on the source doesn't seem to work since you can have multiple sinks.

This relates to the Slack conversation you (@woop) had with @algattik - I think we should have stable consumer groups for each source+store combination (which may span multiple jobs).

If there are concerns that ingestion will lag after it was down for a while, the sink(?) could have a flag such as in _Solution 1_ to indicate it should reset to latest offset (either by fast forwarding or using a new group ID, but I think it's cleaner to maintain the group ID).

woop:
> I agree setting the consumer on the source doesn't seem to work since you can have multiple sinks.
> 
> This relates to the Slack conversation you (@woop) had with @algattik - I think we should have stable consumer groups for each source+store combination (which may span multiple jobs).
> 
> If there are concerns that ingestion will lag after it was down for a while, the sink(?) could have a flag such as in _Solution 1_ to indicate it should reset to latest offset (either by fast forwarding or using a new group ID, but I think it's cleaner to maintain the group ID).

I agree. The group Id should be maintained in my opinion. Skipping old data should be the edge case.

woop:
#757 resolves this I believe (although its bundled in a much larger PR).

Issue #747, opened 2020-05-27
Feast Ingestion not exporting success metrics to statsd


pmjacinto:
## Expected Behavior 

The dataflow step `WriteSuccessMetrics` should be executed and produce metrics.

## Current Behavior

In the previous step, the "dummy" PTransform always seems to output zero rows.
https://github.com/feast-dev/feast/blob/v0.5.0/storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/writer/BigQueryWrite.java#L93-L105

This leads to the metrics transform never getting any inputs.

Is this an expected behaviour?

<img width="825" alt="Screenshot 2020-05-27 at 16 32 13" src="https://user-images.githubusercontent.com/29306657/83040934-c7907500-a037-11ea-9a72-5fd186e9b3c7.png">
---
<img width="825" alt="Screenshot 2020-05-27 at 16 32 34" src="https://user-images.githubusercontent.com/29306657/83040944-cb23fc00-a037-11ea-9beb-f30fb8ba538a.png">

## Steps to reproduce

- Create an ingestion pipeline
- Ingest a row
- Check dataflow metrics

### Specifications

- Version: 0.5.0
- Platform: GKE, Dataflow
- Subsystem: Feast Ingestion

woop:
Hi @pmjacinto 

Thanks for raising this. I believe the comment that you linked to does indicate the problem: BigQueryIO does not emit successful writes, meaning we can't emit metrics. `WriteSuccessMetrics` is an IO agnostic step that is always generated. For the time being we emit total rows processed and failed metrics, the difference of which should give you an estimation of rows written.

woop:
@Yanson @pmjacinto for your information, the latest addition to BQ ingestion will have success metrics again #757

PR #763, opened 2020-06-02
Compute and write metrics for rows prior to store writes

zhilingc:

In PR #568, ingestion metrics were shifted to be computed and written post-store-writes, in order to accurately reflect the values that are actually available to users. 

However, for certain stores (like BQ), successfully written rows cannot be retrieved post-write, and as a result feature row metrics are not produced for ingestion jobs writing to such stores. This PR re-introduces inflight feature row metrics as an alternative source of monitoring.

Fixes #

```release-note
Adds the tag "metrics_namespace" to the following metrics:
- feature_row_lag_ms_*
- feature_value_lag_ms_*
- feature_row_ingested_count
- feature_value_missing_count
- feature_value_*

The tag can contain either of:
- `Inflight` 
- `WriteToStoreSuccess`
```


Yanson:
Does this fix #747?

zhilingc:
> Does this fix #747?

Sort of. It provides an alternative means to monitor elements being ingested when the sink is BigQuery.

woop:
@zhilingc can we add this change in metrics to our documentation please?

PR #769, opened 2020-06-04
Add support to Python SDK for staging files on Amazon S3

### LINKED ISSUE #562: Remove GCP dependency in test_get_batch_features

### LINKED ISSUE #706: Support S3 staging
Currently Feast cannot stage data anywhere except GCP storage.

Implement a handler for s3 staging URLs, so that data can be staged in AWS

we could use kafka and beam to provide an alternative to file based staging.  considering that it is used to ingest the same features, there shouldn't be any reason it couldn't be used for the entity rows or the retrieved features.  It would be nice to allow for multiple transfer methods encapsulated behind an pluggable API.

Part of https://github.com/gojek/feast/issues/367

jmelinav:
Currently Feast client cannot stage data anywhere except GCP storage. The PR adds S3 staging support on client.
**Which issue(s) this PR fixes**: Client can stage files on s3. 

Python Client changes for #706.
Edit: Fixes #562 

```release-note
Added support to Python SDK for staging files on Amazon S3
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [v0.6.0](https://github.com/feast-dev/feast/tree/v0.6.0) (2020-MM-DD)
+
+[Full Changelog](https://github.com/feast-dev/feast/compare/v0.5.1...v0.6.0)
+
+**Implemented enhancements:**
+- Add support for staging files on S3 [\#706](https://github.com/feast-dev/feast/pull/769) ([jmelinav](https://github.com/jmelinav))
+
 ## [0.5.1](https://github.com/feast-dev/feast/tree/0.5.1) (2020-06-06)
 
 [Full Changelog](https://github.com/feast-dev/feast/compare/v0.5.0...v0.5.1)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,5 @@
 # Changelog
 
-## [v0.6.0](https://github.com/feast-dev/feast/tree/v0.6.0) (2020-MM-DD)
-
-[Full Changelog](https://github.com/feast-dev/feast/compare/v0.5.1...v0.6.0)
-
-**Implemented enhancements:**
-- Add support for staging files on S3 [\#706](https://github.com/feast-dev/feast/pull/769) ([jmelinav](https://github.com/jmelinav))
-
 ## [0.5.1](https://github.com/feast-dev/feast/tree/0.5.1) (2020-06-06)
 
 [Full Changelog](https://github.com/feast-dev/feast/compare/v0.5.0...v0.5.1)
--- a/sdk/python/feast/job.py
+++ b/sdk/python/feast/job.py
@@ -21,7 +21,7 @@ from feast.serving.ServingService_pb2 import (
 from feast.serving.ServingService_pb2 import Job as JobProto
 from feast.serving.ServingService_pb2_grpc import ServingServiceStub
 from feast.source import Source
-from feast.staging.storage_client import StorageClient
+from feast.staging.storage_client import get_staging_client
 from feast.wait import wait_retry_backoff
 from tensorflow_metadata.proto.v0 import statistics_pb2
 
@@ -48,7 +48,6 @@ class RetrievalJob:
         """
         self.job_proto = job_proto
         self.serving_stub = serving_stub
-        self.storage_client = StorageClient()
 
     @property
     def id(self):
@@ -122,7 +121,7 @@ class RetrievalJob:
         """
         uris = self.get_avro_files(timeout_sec)
         for file_uri in uris:
-            file_obj = self.storage_client.execute_file_download(file_uri)
+            file_obj = get_staging_client(file_uri.scheme).download_file(file_uri)
             file_obj.seek(0)
             avro_reader = fastavro.reader(file_obj)
 
--- a/sdk/python/feast/job.py
+++ b/sdk/python/feast/job.py
@@ -24,8 +24,8 @@ from feast.source import Source
 from feast.wait import wait_retry_backoff
 from tensorflow_metadata.proto.v0 import statistics_pb2
 
-from feast.staging.staging_strategy import StagingStrategy
 
+from feast.staging.storage_client import StorageClient
 # Maximum no of seconds to wait until the retrieval jobs status is DONE in Feast
 # Currently set to the maximum query execution time limit in BigQuery
 DEFAULT_TIMEOUT_SEC: int = 21600
@@ -49,7 +49,7 @@ class RetrievalJob:
         """
         self.job_proto = job_proto
         self.serving_stub = serving_stub
-        self.staging_strategy = StagingStrategy()
+        self.storage_client = StorageClient()
 
     @property
     def id(self):
@@ -123,7 +123,7 @@ class RetrievalJob:
         """
         uris = self.get_avro_files(timeout_sec)
         for file_uri in uris:
-            file_obj = self.staging_strategy.execute_file_download(file_uri)
+            file_obj = self.storage_client.execute_file_download(file_uri)
             file_obj.seek(0)
             avro_reader = fastavro.reader(file_obj)
 
--- a/sdk/python/feast/job.py
+++ b/sdk/python/feast/job.py
@@ -21,11 +21,10 @@ from feast.serving.ServingService_pb2 import (
 from feast.serving.ServingService_pb2 import Job as JobProto
 from feast.serving.ServingService_pb2_grpc import ServingServiceStub
 from feast.source import Source
+from feast.staging.storage_client import StorageClient
 from feast.wait import wait_retry_backoff
 from tensorflow_metadata.proto.v0 import statistics_pb2
 
-
-from feast.staging.storage_client import StorageClient
 # Maximum no of seconds to wait until the retrieval jobs status is DONE in Feast
 # Currently set to the maximum query execution time limit in BigQuery
 DEFAULT_TIMEOUT_SEC: int = 21600
--- a/sdk/python/feast/job.py
+++ b/sdk/python/feast/job.py
@@ -1,10 +1,8 @@
-import tempfile
 from typing import List
 from urllib.parse import urlparse
 
 import fastavro
 import pandas as pd
-from google.cloud import storage
 from google.protobuf.json_format import MessageToJson
 
 from feast.constants import CONFIG_TIMEOUT_KEY
@@ -26,6 +24,15 @@ from feast.source import Source
 from feast.wait import wait_retry_backoff
 from tensorflow_metadata.proto.v0 import statistics_pb2
 
+from feast.staging.staging_strategy import StagingStrategy
+
+# Maximum no of seconds to wait until the retrieval jobs status is DONE in Feast
+# Currently set to the maximum query execution time limit in BigQuery
+DEFAULT_TIMEOUT_SEC: int = 21600
+
+# Maximum no of seconds to wait before reloading the job status in Feast
+MAX_WAIT_INTERVAL_SEC: int = 60
+
 
 class RetrievalJob:
     """
@@ -42,8 +49,7 @@ class RetrievalJob:
         """
         self.job_proto = job_proto
         self.serving_stub = serving_stub
-        # TODO: abstract away GCP depedency
-        self.gcs_client = storage.Client(project=None)
+        self.staging_strategy = StagingStrategy()
 
     @property
     def id(self):
@@ -117,16 +123,7 @@ class RetrievalJob:
         """
         uris = self.get_avro_files(timeout_sec)
         for file_uri in uris:
-            if file_uri.scheme == "gs":
-                file_obj = tempfile.TemporaryFile()
-                self.gcs_client.download_blob_to_file(file_uri.geturl(), file_obj)
-            elif file_uri.scheme == "file":
-                file_obj = open(file_uri.path, "rb")
-            else:
-                raise Exception(
-                    f"Could not identify file URI {file_uri}. Only gs:// and file:// supported"
-                )
-
+            file_obj = self.staging_strategy.execute_file_download(file_uri)
             file_obj.seek(0)
             avro_reader = fastavro.reader(file_obj)
 
--- a/sdk/python/feast/loaders/file.py
+++ b/sdk/python/feast/loaders/file.py
@@ -76,7 +76,7 @@ def export_source_to_staging_location(
         source_uri = urlparse(source)
         if source_uri.scheme in ["", "file"]:
             # Local file provided as a source
-            dir_path = None
+            dir_path = ""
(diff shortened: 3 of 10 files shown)


woop:
Can you please also add a note in the release log that you have added `S3` support?

sdk/python/feast/job.py
```diff
@@ -47,8 +46,7 @@ def __init__(
         """
         self.job_proto = job_proto
         self.serving_stub = serving_stub
-        # TODO: abstract away GCP depedency
-        self.gcs_client = storage.Client(project=None)
+        self.staging_strategy = StagingStrategy()
```
woop, line 49: Somehow `StagingStrategy` does not fit for me here. If I read this I wouldn't know what this class does. Would it be better if we called it something like `ObjectStoreClient`? or `StorageClient`?

sdk/python/feast/staging/staging_strategy.py
```diff
@@ -0,0 +1,269 @@
+import re
+from abc import ABC, ABCMeta, abstractmethod
+from enum import Enum
+from tempfile import TemporaryFile
+from typing import List
+from urllib.parse import ParseResult, urlparse
+
+import boto3
```
woop, line 8: We should probably make these conditional imports, and not force users to import them.

sdk/python/feast/staging/staging_strategy.py
```diff
@@ -0,0 +1,269 @@
...
+    """
+
+    def __init__(self):
+        self._protocol_dict = dict()
+
+    def execute_file_download(self, file_uri: ParseResult) -> TemporaryFile:
+        """
+        Downloads a file from the uri location and returns a TemporaryFile object
+
+        :param file_uri: Parsed uri of the file ex: urlparse("gs://bucket/file.avro")
```
woop, line 34: Can we please use Google style docstrings?

sdk/python/setup.py
```diff
@@ -46,6 +46,7 @@
     "numpy",
     "google",
     "confluent_kafka",
+    'boto3'
```
woop, line 49: I think we should remove both the GCP and AWS clients here and ask users to install them with descriptive exceptions. @jmelinav what do you think?

sdk/python/feast/staging/staging_strategy.py
```diff
@@ -0,0 +1,269 @@
...
+        self.s3_client = boto3.client("s3")
+
+    def download_file(self, uri: ParseResult) -> TemporaryFile:
+        """
+        Downloads a file from AWS s3 storage and returns a TemporaryFile object
+
+        :param uri: Parsed uri of the file ex: urlparse("s3://bucket/file.avro")
+        :return: TemporaryFile object
+        """
+        url = uri.path[1:]  # removing leading / from the path
```
woop, line 200: Would `lstrip('/')` be safer?

woop:
@jmelinav looks good overall, thanks for the contribution!

Just one final comment: Can you please just double check that we have sufficient test coverage for the `file` and `GCS` clients in the rest of the test suite for the parts that you changed? My intuition is that they would be covered by the existing test cases, but would still be appreciated if you could eye ball it.

algattik:
UTs failing with `AssertionError: DataFrame.columns are different` https://gist.github.com/algattik/98937ec1edbb09d8eab3e24b323615c1

algattik:
When running `make test-python`:

================================================================================================= FAILURES =================================================================================================
_________________________________________________________________ test_export_source_to_staging_location_s3_wildcard_as_source_should_pass _________________________________________________________________

E           botocore.exceptions.ClientError: An error occurred (InvalidAccessKeyId) when calling the ListObjects operation: The AWS Access Key Id you provided does not exist in our records.

jmelinav:
> UTs failing with `AssertionError: DataFrame.columns are different` https://gist.github.com/algattik/98937ec1edbb09d8eab3e24b323615c1

@algattik Seems like problem is column ordering, I will change assert to ignore column ordering. thanks for finding that.

sdk/python/feast/job.py
jmelinav: I will rename class as `StorageClient`.

sdk/python/setup.py
jmelinav: This change will force an adoption for the existing users. is that okay?

algattik:
I'm building on this to add Azure storage support, so it definitely makes sense to make imports conditional

sdk/python/setup.py
jmelinav: @woop I went ahead and removed only boto3, and left google cloud storage dependency. so that existing users are not affected.

jmelinav:
> > UTs failing with `AssertionError: DataFrame.columns are different` https://gist.github.com/algattik/98937ec1edbb09d8eab3e24b323615c1
> 
> @algattik Seems like problem is column ordering, I will change assert to ignore column ordering. thanks for finding that.

@algattik I have updated the test cases to ignore column ordering, Let me know if it works for you now. Also the issue below must be cascading effect of other test failure. Let me know if this issue comes up again.
 ```E botocore.exceptions.ClientError: An error occurred (InvalidAccessKeyId) when calling the ListObjects operation: The AWS Access Key Id you provided does not exist in our records.```

jmelinav:
> Can you please also add a note in the release log that you have added `S3` support?

Added  a note in CHANGELOG.md.

jmelinav:
> @jmelinav looks good overall, thanks for the contribution!
> 
> Just one final comment: Can you please just double check that we have sufficient test coverage for the `file` and `GCS` clients in the rest of the test suite for the parts that you changed? My intuition is that they would be covered by the existing test cases, but would still be appreciated if you could eye ball it.

@woop A quick glance through bq-batch-retrieval.py says it has good coverage for file and gcs around the changes I have made. Also I have added unit tests for `file` and `s3` clients.

jmelinav:
> @jmelinav looks good overall, thanks for the contribution!
> 
> Just one final comment: Can you please just double check that we have sufficient test coverage for the `file` and `GCS` clients in the rest of the test suite for the parts that you changed? My intuition is that they would be covered by the existing test cases, but would still be appreciated if you could eye ball it.

@woop A quick glance through bq-batch-retrieval.py says it has good coverage for file and gcs around the changes I have made. Also I have added unit tests for `file` and `s3` clients.

sdk/python/setup.py
woop: That's ok. I think we should phase out our GCP requirement over time thought.

[I quite like how Airflow does this](https://airflow.apache.org/docs/stable/installation.html)

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## [v0.6.0](https://github.com/feast-dev/feast/tree/v0.6.0) (2020-MM-DD)
+
+[Full Changelog](https://github.com/feast-dev/feast/compare/v0.5.1...v0.6.0)
+
+**Implemented enhancements:**
+- Add support for staging files on S3 [\#706](https://github.com/feast-dev/feast/pull/769) ([jmelinav](https://github.com/jmelinav))
```
woop, line 8: You don't need to add this yet. We will add it when cutting the release.

sdk/python/feast/staging/storage_client.py
```diff
@@ -0,0 +1,312 @@
+#
+# Copyright 2020 Cimpress
```
woop, line 2: You'll have to update this to `The Feast Authors`

sdk/python/feast/staging/storage_client.py
```diff
@@ -0,0 +1,312 @@
...
+
+        Returns:
+             TemporaryFile object
+        """
+        url = uri.geturl()
+        file_obj = TemporaryFile()
+        self.gcs_client.download_blob_to_file(url, file_obj)
+        return file_obj
+
+    def list_files(self, bucket: str, uri: ParseResult) -> List[str]:
```
woop, line 171: Is there an example of `list_files` being tested or used? Are we enforcing the use of wildcards? Normally people dont use wildcards for listing objects on GCS.

CHANGELOG.md
jmelinav: Oh, I thought this is what you meant by "add a note in the release. I will revert this change."

sdk/python/feast/staging/storage_client.py
jmelinav: This was an existing piece of code which has been moved here. (`def _get_files(bucket: str, uri: ParseResult) -> List[str]:` method in `file.py`).

sdk/python/feast/staging/storage_client.py
```diff
@@ -0,0 +1,312 @@
...
+    """
+    Currently supported protocols enum class.
+    """
+
+    GS = "gs"
+    S3 = "s3"
+    LOCAL_FILE = "file"
+
+
+class StorageClient:
```
mrzzy, line 35: This class seems to be redundant. Its whole purpose is to figure out which `PROTOCOL` maps to which `StagingProtocol` implementation. Maybe we could expose `_protocol_dict` without a wrapping class?

sdk/python/feast/staging/storage_client.py
```diff
@@ -0,0 +1,312 @@
...
+            elif PROTOCOL(protocol) == PROTOCOL.LOCAL_FILE:
+                self._protocol_dict[protocol] = LocalFSProtocol()
+            else:
+                raise Exception(
+                    f"Could not identify file protocol {protocol}. Only gs:// and file:// and s3:// supported"
+                )
+            return self._protocol_dict[protocol]
+
+
+class AbstractStagingProtocol(ABC):
```
mrzzy, line 117: Not sure about naming this `Protocol` when it clearly defines methods that relate storage client methods. Especially since we already have `PROTOCOL` enum, it might cause confusion.

sdk/python/feast/loaders/file.py
```diff
@@ -66,47 +69,34 @@ def export_source_to_staging_location(
         uri_path = None  # type: Optional[str]
         if uri.scheme == "file":
```
mrzzy, line 70: Can the handling of file:/// type URLs be abstracted away inside the StorageClient via LocalFSProtocol?

jmelinav: @mrzzy I tried while refactoring this file. This particular case didn't fit well with the abstraction. This part of the code is saving the data frame into a temp location and then it will be uploaded into a staging location. I saw those two as separate concerns as staging location itself could be a local file system.

sdk/python/feast/staging/storage_client.py
jmelinav: @mrzzy Can be done, But one thing that `StorageClient` is doing, adding a place holder and lazy initialization of StagingProtocol. Lazy initialization can be done with just the method `_get_staging_protocol` but this method will have to be invoked before use every invocation. StorageClient cannot be a dependency in class. Let me know what you think. I am open to removing this class and exposing  _protocol_dict

sdk/python/feast/staging/storage_client.py
jmelinav: @mrzzy `AbstractStagingClient` would be appropriate?

sdk/python/feast/staging/storage_client.py
mrzzy: > , adding a place holder 

Not sure what you mean here.

> lazy initialization of StagingProtocol. 

Lazy initialization makes sense as the python package already has a terrible start up time and we won't want to worsen it. However I don't think we need a wrapper class to do it, we can just delay calling `__init__()` on the staging protocol classes:
```
storage_protocols = {
    "gcs": GCSProtocol,
    "s3": S3Protocol,
    ...
}
# when we actually use it
staging_protocol = staging_protocol["s3"]()
```

We can wrap this in a `get_staging_protocol(protocol)` to get human readable error messages when a bad protocol is  used while making sure we keep boilerplate to a minimum.

sdk/python/feast/staging/storage_client.py
mrzzy: I think storage client/backend might be more suitable as generic list/upload/download methods (we expose here don't really have anything to do with staging even though we use it for staging

sdk/python/feast/loaders/file.py
mrzzy: @jmelinav makes sense if we see in a staging protocol perspective using the current abstraction. See comment on `StagingProtocol`. If we consider this as a storage client/backend, integrating this might make more sense.

sdk/python/tests/test_job.py
```diff
@@ -0,0 +1,144 @@
...
+                        status=JobStatus.JOB_STATUS_DONE,
+                        data_format=DataFormat.DATA_FORMAT_INVALID,
+                    )
+                ),
+                Exception,
+            ),
+        ],
+        ids=["when_retrieve_job_fails", "when_data_format_is_not_avro"],
+    )
+    def test_to_dataframe_s3_file_staging_should_raise(
```
mrzzy, line 137: Might be out of scope for this PR but it would be nice if we can add test for GCS staging as well.

mrzzy:
Since this PR only addresses the support on the Python SDK client side, should we:
- the retitle the PR as such lest people start thinking that full S3 staging support is addressed in the PR
- not close the S3 staging issue until Serving side has something that supports S3 staging.

sdk/python/feast/staging/storage_client.py
jmelinav: Sounds good.

jmelinav:
> Since this PR only addresses the support on the Python SDK client side, should we:
> 
> * the retitle the PR as such lest people start thinking that full S3 staging support is addressed in the PR
> * not close the S3 staging issue with this PR until Serving side has something that supports S3 staging.

Sounds good to me, I have removed the fixes #706 from the PR description and mentioned it addresses only python client changes. Also I have appended to `Client side only(python sdk)`. Feel free to retitle.

sdk/python/tests/test_job.py
jmelinav: @mrzzy for sure, I did try for gcs unit tests, I couldn't figure out the errors I was running into during gcs mocks, so I gave up.

woop: Lets track it over here. https://github.com/feast-dev/feast/issues/831

PR #785, opened 2020-06-09
Add API for listing feature sets using labels

terryyylim:
Adds listing functionality using labels for FeatureSet

N/A

```release-note
User can now list feature sets by labels
```


core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -189,7 +191,17 @@ public ListFeatureSetsResponse listFeatureSets(ListFeatureSetsRequest.Filter fil
     ListFeatureSetsResponse.Builder response = ListFeatureSetsResponse.newBuilder();
     if (featureSets.size() > 0) {
       for (FeatureSet featureSet : featureSets) {
-        response.addFeatureSets(featureSet.toProto());
+        if (labels != null && !labels.isEmpty()) {
```
woop, line 194: This can be made more concise by using guard clauses. 
```
If( label filter ){
test if label in feature set, if not, "continue"
}
response.addFeatureSets(...);
```

sdk/python/feast/cli.py
```diff
@@ -121,14 +121,48 @@ def feature_set():
...
+    "--project",
+    "-p",
+    help="Project that feature set belongs to",
+    type=click.STRING,
+    default="*",
+)
+@click.option(
+    "--name",
+    "-n",
+    help="Name that feature set belongs to, may include * to match multiple feature sets eg. driver_*",
```
woop, line 134: A feature set doesn't belong to a name. I think this can be rephrased.

sdk/python/feast/cli.py
```diff
@@ -121,14 +121,48 @@ def feature_set():
...
 
+    labels_kv = labels.split(",")
+    labels_dict = {}
+    if labels == "":
+        pass
+    elif len(labels_kv) % 2 == 0:
+        for k, v in zip(labels_kv[0::2], labels_kv[1::2]):
+            labels_dict[k] = v
+    else:
+        print("Uneven key-value label pairs were entered")
```
woop, line 159: You should raise an exception, not print.

sdk/python/feast/cli.py
```diff
@@ -121,14 +121,48 @@ def feature_set():
...
+def feature_set_list(project: str, name: str, labels: str):
     """
     List all feature sets
     """
     feast_client = Client()  # type: Client
 
+    labels_kv = labels.split(",")
+    labels_dict = {}
+    if labels == "":
+        pass
```
woop, line 154: What are you trying to achieve with this if statement?

sdk/python/feast/client.py
```diff
@@ -448,7 +448,12 @@ def list_feature_sets(
         if name is None:
             name = "*"
 
-        filter = ListFeatureSetsRequest.Filter(project=project, feature_set_name=name)
+        if labels is None:
```
woop, line 451: Would it make more sense to set the default to `dict()`? You are setting the default to `None` and then setting it to `{}`

sdk/python/feast/cli.py
terryyylim: Would this be clearer? `Name of feature set, may include * to match multiple feature sets eg. driver_*`

tests/e2e/redis/basic-ingest-redis-serving.py
```diff
@@ -28,27 +28,26 @@
 import uuid
 
 FLOAT_TOLERANCE = 0.00001
-PROJECT_NAME = 'basic_' + uuid.uuid4().hex.upper()[0:6]
+PROJECT_NAME = "basic_" + uuid.uuid4().hex.upper()[0:6]
```
woop, line 31: Why did all of this formatting change? Can you reverse this please. We can have another PR if we need to update formatting.

tests/e2e/redis/basic-ingest-redis-serving.py
```diff
@@ -119,13 +123,29 @@ def test_basic_register_feature_set_success(client):
...
     client.apply(cust_trans_fs_expected)
-    cust_trans_fs_actual = client.get_feature_set("customer_transactions",
-                                                  project=PROJECT_NAME)
+    cust_trans_fs_actual = client.get_feature_set(
+        "customer_transactions", project=PROJECT_NAME
+    )
     assert cust_trans_fs_actual == cust_trans_fs_expected
 
+    # Register feature set with labels
+    driver_fs_expected = FeatureSet.from_yaml(f"{DIR_PATH}/basic/driver_fs.yaml")
```
woop, line 137: This test is too brittle. I think it would be better to move all of the logic into code instead of using the YAML files. Otherwise if somebody modifies driver_fs then this test will break.

sdk/python/feast/cli.py
terryyylim: For a default case of no labels being passed, I'll use an empty dictionary.

sdk/python/feast/client.py
terryyylim: Yup that makes sense, will change to default {}

tests/e2e/redis/basic-ingest-redis-serving.py
terryyylim: Okay. It changed cause I ran `black` formatting on it.

sdk/python/feast/cli.py
woop: What about

`Filters feature sets by name. Wildcards may be included to match multiple feature sets...`

sdk/python/feast/cli.py
woop: What about setting the default argument to be a dict()?

core/src/main/java/feast/core/model/FeatureSet.java
```diff
@@ -121,6 +121,17 @@ private String getProjectName() {
     }
   }
 
+  public boolean hasAllLabels(Map<String, String> labels) throws InvalidProtocolBufferException {
+    Map<String, String> featureSetLabelsMap = this.toProto().getSpec().getLabelsMap();
```
woop, line 125: I think you should implement a `getLabelsMap` on this class instead of converting to proto here.

sdk/python/feast/cli.py
terryyylim: Updated.

sdk/python/feast/cli.py
terryyylim: Updated.

core/src/main/java/feast/core/service/SpecService.java
terryyylim: Updated.

tests/e2e/redis/basic-ingest-redis-serving.py
terryyylim: Updated.

tests/e2e/redis/basic-ingest-redis-serving.py
terryyylim: Reverted changes.

sdk/python/feast/client.py

sdk/python/feast/cli.py
terryyylim: Added helper function to manage logic.

sdk/python/feast/cli.py
```diff
@@ -120,6 +120,18 @@ def feature_set():
     pass
 
 
+def _get_labels_dict(label_str: str):
+    labels_dict = {}
+    labels_kv = label_str.split(",")
+    if label_str == "":
+        return labels_dict
+    if len(labels_kv) % 2 == 1:
```
woop, line 128: This should still raise a value error

core/src/main/java/feast/core/model/FeatureSet.java
terryyylim: Updated.

sdk/python/feast/cli.py
terryyylim: Updated.

core/src/main/java/feast/core/model/FeatureSet.java
```diff
@@ -121,6 +121,17 @@ private String getProjectName() {
     }
   }
 
+  public boolean hasAllLabels(Map<String, String> labels) {
```
woop, line 124: Please add JavaDocs.

core/src/main/java/feast/core/model/FeatureSet.java
```diff
@@ -262,6 +273,10 @@ public void addFeature(Feature feature) {
     return FeatureSetProto.FeatureSet.newBuilder().setMeta(meta).setSpec(spec).build();
   }
 
+  public Map<String, String> getLabelsMap() {
+    return TypeConversion.convertJsonStringToMap(this.labels);
```
woop, line 277: should this be `this.getLabels()`?

core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -126,13 +128,17 @@ public GetFeatureSetResponse getFeatureSet(GetFeatureSetRequest request)
...
+   * will be returned.
+   *
    * @param filter filter containing the desired featureSet name
    * @return ListFeatureSetsResponse with list of featureSets found matching the filter
    */
   public ListFeatureSetsResponse listFeatureSets(ListFeatureSetsRequest.Filter filter)
       throws InvalidProtocolBufferException {
     String name = filter.getFeatureSetName();
     String project = filter.getProject();
+    Map<String, String> labels = filter.getLabelsMap();
```
woop, line 141: I think we should call this `labelFilter` to be clearer.

sdk/python/feast/cli.py
woop: Please print out what the user entered as part of exception.

sdk/python/feast/cli.py
```diff
@@ -120,15 +120,52 @@ def feature_set():
...
+    "--project",
+    "-p",
+    help="Project that feature set belongs to",
+    type=click.STRING,
+    default="*",
+)
+@click.option(
+    "--name",
+    "-n",
+    help="Filters feature sets by name. Wildcards may be included to match multiple feature sets",
```
woop, line 146: You should indicate the wildcard (`*`) since people may not know the character.

sdk/python/feast/cli.py
```diff
@@ -120,15 +120,52 @@ def feature_set():
     pass
 
 
+def _get_labels_dict(label_str: str):
```
woop, line 123: We should start adding comments everywhere, this private method is no different. Please add Google Docstring (Intellij can generate).

core/src/main/java/feast/core/model/FeatureSet.java
terryyylim: There's a current helper function that converts this, used in a similar way [here](https://github.com/feast-dev/feast/blob/master/core/src/main/java/feast/core/model/Feature.java#L241). Should I avoid writing it that way?

woop: yea you should try to use methods as much as possible because you can add logic there later.

PR #792, opened 2020-06-11
FeatureSets are delivered to Ingestion Job through Kafka

### LINKED ISSUE #761: Job Coordination Improvement Proposal

There's some technical debt in ingestion part mostly related to Job Coordination: job is being restarted on schema update, this restart can take arbitrary amount of time, hence we had to split jobs as much as possible to minimize mutual disruption (on restart). This negatively affects our UX and also restrict us from using Dataflow resources more efficiently, which currently leads to very high spendings on Dataflow.

I drafter proposal that addresses those issues and suggest new design in communications between core service and ingestion pipeline
https://docs.google.com/document/d/1gqkCWZUyVBIU8OKhxIhIf1BBd3JrbOx2WnXrpGLFCAc/edit#heading=h.wigpvke4im4g

pyalex:

SpecService & IngestionJob are now communicate through kafka topics, which makes job restarts on FeatureSet change obsolete. Now job restarted only when subscription configuration of store was changed.

Communication Flow:
1. SpecService.applyFeatureSet publish FeatureSetSpec to `specs-topic` and set FeatureSet status to Pending
2. IngestionJob reads from this topic (all history of changes + recent updates) & build materialized view of Specs in memory
3. IngestionJob sends acknowledgment back to SpecService via `specs-ack-topic`
4. SpecService collects acknowledgments from all related jobs (see FeatureSetJobStatus) and when all running jobs acknowledged FeatureSet status is changed to Ready

Fixes #761 

No

```release-note

```

### DIFF
--- a/core/src/main/java/feast/core/config/FeastProperties.java
+++ b/core/src/main/java/feast/core/config/FeastProperties.java
@@ -139,6 +139,10 @@ public class FeastProperties {
     /* Feature stream options */
     @NotNull private FeatureStreamOptions options;
 
+    /* FeatureSetSpec stream options - communication channel between SpecService and IngestionJob
+     * to update Spec inside job w/o restart */
+    @NotNull private FeatureSetSpecStreamProperties specsOptions;
+
     /** Feature stream options */
     @Getter
     @Setter
@@ -158,6 +162,16 @@ public class FeastProperties {
 
       /* Number of Kafka partitions to to use for managed feature stream. */
       @Positive private int partitions = 1;
+    }
+
+    @Getter
+    @Setter
+    public static class FeatureSetSpecStreamProperties {
+      /* Kafka topic to send feature set spec to ingestion streaming job */
+      @NotBlank private String specsTopic = "feast-feature-set-specs";
+
+      /* Kafka topic to receive acknowledgment from ingestion job on successful processing of new specs */
+      @NotBlank private String specsAckTopic = "feast-feature-set-specs-ack";
     }
   }
 
--- a/core/src/main/java/feast/core/config/FeatureStreamConfig.java
+++ b/core/src/main/java/feast/core/config/FeatureStreamConfig.java
@@ -16,30 +16,119 @@
  */
 package feast.core.config;
 
-import com.google.common.base.Strings;
 import feast.core.config.FeastProperties.StreamProperties;
 import feast.core.model.Source;
+import feast.core.util.KafkaSerialization;
+import feast.proto.core.FeatureSetProto;
+import feast.proto.core.IngestionJobProto;
 import feast.proto.core.SourceProto.KafkaSourceConfig;
 import feast.proto.core.SourceProto.SourceType;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.concurrent.ExecutionException;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.kafka.clients.admin.AdminClient;
 import org.apache.kafka.clients.admin.AdminClientConfig;
-import org.apache.kafka.clients.admin.CreateTopicsResult;
 import org.apache.kafka.clients.admin.NewTopic;
-import org.apache.kafka.common.errors.TopicExistsException;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
+import org.apache.kafka.clients.producer.ProducerConfig;
+import org.apache.kafka.common.config.TopicConfig;
+import org.apache.kafka.common.serialization.StringDeserializer;
+import org.apache.kafka.common.serialization.StringSerializer;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.kafka.core.*;
 
 @Slf4j
 @Configuration
 public class FeatureStreamConfig {
 
   String DEFAULT_KAFKA_REQUEST_TIMEOUT_MS_CONFIG = "15000";
+  int DEFAULT_SPECS_TOPIC_PARTITIONING = 1;
+  short DEFAULT_SPECS_TOPIC_REPLICATION = 3;
+
+  @Bean
+  public KafkaAdmin admin(FeastProperties feastProperties) {
+    String bootstrapServers = feastProperties.getStream().getOptions().getBootstrapServers();
+
+    Map<String, Object> configs = new HashMap<>();
+    configs.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServers);
+    configs.put(
+        AdminClientConfig.REQUEST_TIMEOUT_MS_CONFIG, DEFAULT_KAFKA_REQUEST_TIMEOUT_MS_CONFIG);
+    return new KafkaAdmin(configs);
+  }
+
+  @Bean
+  public NewTopic featureRowsTopic(FeastProperties feastProperties) {
+    StreamProperties streamProperties = feastProperties.getStream();
+
+    return new NewTopic(
+        streamProperties.getOptions().getTopic(),
+        streamProperties.getOptions().getPartitions(),
+        streamProperties.getOptions().getReplicationFactor());
+  }
+
+  @Bean
+  public NewTopic featureSetSpecsTopic(FeastProperties feastProperties) {
+    StreamProperties streamProperties = feastProperties.getStream();
+    Map<String, String> configs = new HashMap<>();
+    configs.put(TopicConfig.CLEANUP_POLICY_CONFIG, TopicConfig.CLEANUP_POLICY_COMPACT);
+
+    NewTopic topic =
+        new NewTopic(
+            streamProperties.getSpecsOptions().getSpecsTopic(),
+            DEFAULT_SPECS_TOPIC_PARTITIONING,
+            DEFAULT_SPECS_TOPIC_REPLICATION);
+
+    topic.configs(configs);
+    return topic;
+  }
+
+  @Bean
+  public NewTopic featureSetSpecsAckTopic(FeastProperties feastProperties) {
+    StreamProperties streamProperties = feastProperties.getStream();
+
+    return new NewTopic(
+        streamProperties.getSpecsOptions().getSpecsAckTopic(),
+        DEFAULT_SPECS_TOPIC_PARTITIONING,
+        (short) 1);
+  }
+
+  @Bean
+  public KafkaTemplate<String, FeatureSetProto.FeatureSetSpec> specKafkaTemplate(
+      FeastProperties feastProperties) {
+    StreamProperties streamProperties = feastProperties.getStream();
+    Map<String, Object> props = new HashMap<>();
+
+    props.put(
+        ProducerConfig.BOOTSTRAP_SERVERS_CONFIG,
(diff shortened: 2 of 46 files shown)


core/src/main/java/feast/core/config/FeastProperties.java
```diff
@@ -159,6 +161,19 @@ public Runner getActiveRunner() {
...
+    @Setter
+    public static class FeatureSetSpecStreamProperties {
+      /* Kafka topic to send feature set spec to ingestion streaming job */
+      @NotBlank private String specsTopic = "feast-feature-set-specs";
+
+      /* Kafka topic to receive acknowledgment from ingestion job on successful processing of new specs */
+      @NotBlank private String specsAckTopic = "feast-feature-set-specs-ack";
+
+      /*Comma separated list of Kafka bootstrap servers */
+      @NotBlank private String bootstrapServers = "localhost:9092";
```
woop, line 175: Will we have a use case for splitting the spec stream and feature stream? I think we can maybe roll these options into a single class.

core/src/main/java/feast/core/config/FeatureStreamConfig.java
```diff
@@ -16,30 +16,119 @@
...
 @Slf4j
 @Configuration
 public class FeatureStreamConfig {
 
   String DEFAULT_KAFKA_REQUEST_TIMEOUT_MS_CONFIG = "15000";
+  int DEFAULT_SPECS_TOPIC_PARTITIONING = 1;
+  short DEFAULT_SPECS_TOPIC_REPLICATION = 3;
+
+  @Bean
+  public KafkaAdmin admin(FeastProperties feastProperties) {
```
woop, line 50: Where is this method being used?

core/src/main/java/feast/core/config/JobConfig.java
```diff
@@ -54,19 +56,34 @@ public JobManager getJobManager(FeastProperties feastProperties)
     String configJson = gson.toJson(runnerConfigOptions);
 
     FeastProperties.MetricsProperties metrics = jobProperties.getMetrics();
+    FeastProperties.StreamProperties streamProperties = feastProperties.getStream();
```
woop, line 59: Would it be cleaner to move this out into a method?

core/src/main/java/feast/core/model/FeatureSet.java
```diff
@@ -125,6 +132,24 @@ public void setProject(Project project) {
...
+  public String getReference() {
+    return String.format("%s/%s", getProjectName(), getName());
+  }
+
+  /** @return Pair &lt;ProjectName, FeatureSetName&gt; */
+  public static Pair<String, String> parseReference(String reference) {
+    String[] split = reference.split("/", 2);
+    if (split.length == 1) {
+      return Pair.of(Project.DEFAULT_NAME, split[0]);
+    } else {
```
woop, line 144: I dont think we need an `else` if we are returning, and we should probably raise an exception in the case of >2.

core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -60,21 +64,26 @@
...
+  private final KafkaTemplate<String, FeatureSetProto.FeatureSetSpec> specPublisher;
 
   @Autowired
   public SpecService(
       FeatureSetRepository featureSetRepository,
       StoreRepository storeRepository,
       ProjectRepository projectRepository,
-      Source defaultSource) {
+      Source defaultSource,
+      KafkaTemplate<String, FeatureSetProto.FeatureSetSpec> specPublisher) {
```
woop, line 81: I just want to flag that even though I agree with the approach we are taking with Spring and Kafka, I do consider it technical debt that we are going into. Ideally the life cycle of jobs and the updates of feature sets to those jobs would fully encapsulated in the job management layer, especially if we ever want to separate job management from core.

core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -299,6 +309,26 @@ public ApplyFeatureSetResponse applyFeatureSet(FeatureSetProto.FeatureSet newFea
...
 
+    featureSet.incVersion();
+
+    try {
+      specPublisher
+          .sendDefault(featureSet.getReference(), featureSet.toProto().getSpec())
+          .get(SPEC_PUBLISHING_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+    } catch (Exception e) {
+      throw io.grpc.Status.UNAVAILABLE
+          .withDescription("Unable to publish FeatureSet to Kafka")
```
woop, line 320: I think we should return the exception message so that users (mostly us) can figure out what is wrong. In the future we can abstract that away.

core/src/main/java/feast/core/service/SpecService.java
pyalex: I would be more than happy to move this communication responsibility to JobService, but right now JobService is dependant on SpecService and I need publishing to kafka be synchronous part of applyFeatureSet. So yeah, currently it's a tech debt.

core/src/main/java/feast/core/config/JobConfig.java
pyalex: Changed to SpecsStreamingUpdateConfig

core/src/main/java/feast/core/config/FeastProperties.java
```diff
@@ -139,6 +139,8 @@ public Runner getActiveRunner() {
     /* Feature stream options */
     @NotNull private FeatureStreamOptions options;
 
+    @NotNull private FeatureSetSpecStreamProperties specsOptions;
```
woop, line 142: Can we add a comment here please.

core/src/main/java/feast/core/model/Job.java
```diff
@@ -59,17 +61,9 @@
...
-      joinColumns = @JoinColumn(name = "job_id"),
-      inverseJoinColumns = @JoinColumn(name = "feature_sets_id"),
-      indexes = {
-        @Index(name = "idx_jobs_feature_sets_job_id", columnList = "job_id"),
-        @Index(name = "idx_jobs_feature_sets_feature_sets_id", columnList = "feature_sets_id")
-      })
-  private List<FeatureSet> featureSets;
+  // FeatureSets populated by the job via intermediate FeatureSetJobStatus model
+  @OneToMany(mappedBy = "job", cascade = CascadeType.ALL)
+  private Set<FeatureSetJobStatus> featureSets = new HashSet<>();
```
woop, line 66: Shouldn't this be `featureSetJobStatus`?

core/src/main/java/feast/core/config/JobConfig.java
```diff
@@ -38,14 +40,34 @@
 public class JobConfig {
   private final Gson gson = new Gson();
 
+  @Bean
+  public IngestionJobProto.SpecsStreamingUpdateConfig specsStreamingUpdateConfig(
```
woop, line 44: Can we add a comment here please and use a verb? `create` or `get`?

core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -299,6 +309,29 @@ public ApplyFeatureSetResponse applyFeatureSet(FeatureSetProto.FeatureSet newFea
       status = Status.UPDATED;
     }
 
+    featureSet.incVersion();
+
+    try {
+      specPublisher
```
woop, line 315: Comment would be nice here.

core/src/test/java/feast/core/service/SpecServiceTest.java
```diff
@@ -318,6 +337,74 @@ public void applyFeatureSetShouldUpdateAndSaveFeatureSetIfAlreadyExists()
...
         FeatureSet.fromProto(expected));
+
+    assertThat(applyFeatureSetResponse.getFeatureSet().getSpec().getVersion(), equalTo(2));
+    verify(kafkaTemplate)
+        .sendDefault(eq(featureSets.get(0).getReference()), any(FeatureSetSpec.class));
+  }
+
+  @Test
+  @SneakyThrows
+  public void applyFeatureSetShouldNtWorkWithoutProducingToKafka() {
```
woop, line 348: Typo in name.

ingestion/src/main/java/feast/ingestion/transform/specs/FilterRelevantFunction.java
```diff
@@ -0,0 +1,48 @@
...
+
+import feast.ingestion.utils.SpecUtil;
+import feast.proto.core.FeatureSetProto;
+import feast.proto.core.SourceProto;
+import feast.proto.core.StoreProto;
+import java.util.List;
+import org.apache.beam.sdk.transforms.ProcessFunction;
+import org.apache.beam.sdk.values.KV;
+
+public class FilterRelevantFunction
```
woop, line 27: Can we get a comment on this class please.

ingestion/src/main/java/feast/ingestion/transform/specs/KafkaRecordToFeatureSetSpec.java
```diff
@@ -0,0 +1,43 @@
...
+
+import com.google.protobuf.InvalidProtocolBufferException;
+import feast.proto.core.FeatureSetProto;
+import org.apache.beam.sdk.io.kafka.KafkaRecord;
+import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.values.KV;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class KafkaRecordToFeatureSetSpec
```
woop, line 27: Same here, just a brief one liner would be nice, to disambiguate it from the data stream.

core/src/main/java/feast/core/model/Job.java
pyalex: Don't we call models in plural?

ingestion/src/test/java/feast/ingestion/transform/specs/FilterRelevantTest.java
```diff
@@ -0,0 +1,109 @@
...
+import org.apache.beam.sdk.values.KV;
+import org.apache.beam.sdk.values.PCollection;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class FilterRelevantTest {
+  @Rule public transient TestPipeline p = TestPipeline.create();
+
+  @Test
+  public void onlySpecsThatMatchStoreSubscriptionShouldPas() {
```
woop, line 38: Typo.

protos/feast/core/FeatureSet.proto
```diff
@@ -63,6 +63,11 @@ message FeatureSetSpec {
 
     // User defined metadata
     map<string,string> labels = 8;
+
+    // incremental version
+    // This options is read-only (for clients) -
+    // only changed inside SpecService when FeatureSet is updated
```
woop, line 69: How about
```
Read-only self-incrementing version that increases monotonically when changes are made to a feature set```

File: protos/feast/core/IngestionJob.proto [hunk 16]
```diff
@@ -63,3 +63,14 @@ enum IngestionJobStatus {
     // job has been suspended
     SUSPENDED = 8;
 }
+
+message SpecsStreamingUpdateConfig {
```
woop, line 67: These proto messages definitely need comments on each field. It's quite useful since we will produce documentation from it here automatically: https://api.docs.feast.dev/grpc/feast.core.pb.html#IngestionJobStatus

PR #793, opened 2020-06-11
Authentication and Authorization

dr3s:

First implementation of auth for Feast (related to #504 minimal implementation).

  

1.   Adds authentication to Feast Core (with support for different implementations). Currently any JWT bearer token through gRPC metadata.
2.  Adds authorization to Feast Core (with support for different implementations). Currently only supports Ory Keto.  A follow up PR will add an HTTP authorization adapter.
3.  Adds authentication to Python SDK/CLI. Two implementations included: users can enable authentication client side and Feast will send their Google Open ID credentials as gRPC metadata to Core, or they can provide client credentials and OAuth2 provider and the JWT will be fetched for them.
4.  Refactored the Python SDK/CLI SSL/TLS handling.
5.  Prevents unauthorized creation or modification of feature sets in projects that a user does not have membership in.

Limitations

Does not handle user or role management in authorization provider (creating projects, adding members, removing members, listing members).

Which issue(s) this PR fixes:

Related to #504, but doesn't close the card. This is a minimal implementation.  Replaces https://github.com/feast-dev/feast/pull/554

Does this PR introduce a user-facing change?:

Yes, documentation will be needed:

- The Python Client SDK has a constructor now to pass authentication configuration.   
- The Core Service API requires GRPC metadata when authentication is enabled.
- Configuration for Core has been extended to enable authentication and authorization.


core/pom.xml
```diff
@@ -31,15 +31,36 @@
...
+              <groupId>org.springframework.boot</groupId>
+              <artifactId>spring-boot-maven-plugin</artifactId>
+              <configuration>
+                  <skip>false</skip>
+              </configuration>
+              <executions>
+                  <execution>
+                      <id>build-info</id>
+                      <goals>
+                          <goal>build-info</goal>
```
woop, line 58: What is the reason for adding this exclusion?

core/pom.xml
```diff
@@ -56,6 +77,19 @@
       <artifactId>spring-boot-devtools</artifactId>
       <optional>true</optional>
     </dependency>
+    
+        <dependency>
+            <groupId>com.google.auto.value</groupId>
```
woop, line 82: We don't want to add AutoValue until Lombok has been removed because they don't play well together.

core/pom.xml
```diff
@@ -127,6 +161,10 @@
       <version>5.3.0.RELEASE</version>
     </dependency>
 
+    <dependency>
+        <groupId>io.github.lognet</groupId>
```
woop, line 165: I am using https://github.com/feast-dev/feast/pull/793/commits/639c6079005e095837001d2394b65e1d1eaa7fdc#diff-357e4854869b2e21c38b1b437f11095aR116 for the spring boot gRPC starter, because Lognet has a security vulnerability. 

Is there a reason  you added it back?

sdk/python/feast/client.py
```diff
@@ -494,7 +492,7 @@ def get_batch_features(
         self,
         feature_refs: List[str],
         entity_rows: Union[pd.DataFrame, str],
-        project: str = None,
+        default_project: str = None,
```
woop, line 495: Why was this change made?

core/pom.xml
dr3s: execution not exclusion.  I think it came from master.  I'm not really sure about much in the pom.  Feel free to push changes if you know better.

core/pom.xml
dr3s: good to know.  I can use lombok in the meantime.

core/pom.xml
dr3s: just the merge.  I'll try to remove.

sdk/python/feast/client.py
jmelinav: A merge conflict started causing an error in test cases. To fix that, I did this change, But i missed checking which one is latest. I will revert this change and also update test cases.

core/pom.xml
woop: > execution not exclusion. I think it came from master. I'm not really sure about much in the pom. Feel free to push changes if you know better.

Oops, misread.

woop:
I'd just like to repeat a message I sent @dr3s privately. In order for us to ensure that auth is functioning properly, I think we should be extending the basic e2e test suite.

1. If you look at this script: https://github.com/feast-dev/feast/blob/master/.prow/config.yaml#L144
1. It runs this pytest command https://github.com/feast-dev/feast/blob/master/infra/scripts/test-end-to-end.sh#L56
1. Which in turn runs this e2e test https://github.com/feast-dev/feast/tree/master/tests/e2e/redis

What I suggest we do
1. Add a new Prow script called `test-end-to-end-auth`. 
1. Let that script execute (2) above with a flag that enables auth. This flag is passed on through the Pytest command.
1. Pytest should run the normal basic redis e2e test fully with auth. 
1. Extend the basic redis e2e test with auth specific tests that you think are missing. Use marks as we have with Dataflow tests in order to NOT run the auth tests when the auth flag is not passed to Pytest. You can see an [example here](https://github.com/feast-dev/feast/blob/master/tests/e2e/bq/bq-batch-retrieval.py#L85)

core/pom.xml
```diff
@@ -18,228 +18,318 @@
...
-            <version>6.1.2.Final</version>
-        </dependency>
-        <dependency>
-            <groupId>org.hibernate.validator</groupId>
-            <artifactId>hibernate-validator-annotation-processor</artifactId>
-            <version>6.1.2.Final</version>
-        </dependency>
-
-    </dependencies>
+  <modelVersion>4.0.0</modelVersion>
```
woop, line 21: Can I please ask that we remove this formattin/indentation change.

core/src/main/java/feast/core/auth/authorization/AuthorizationResult.java
```diff
@@ -0,0 +1,43 @@
...
+package feast.core.auth.authorization;
+
+import java.util.Optional;
+import javax.annotation.Nullable;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+@Getter
+@AllArgsConstructor
+public class AuthorizationResult {
```
woop, line 26: A small javadoc comment for this class would be appreciated.

core/pom.xml
dr3s: I guess I'm not sure how this occurred because it was in your original branch.  We didn't do anything to format the file.  Is there a formatting standard that we can just configure in the IDE?  I'm not sure how it is expected to be formatted.

woop: Ah ok, well some dependencies have changed while others have not changed. The indentation change makes it hard to see what the difference is.

infra/scripts/test-end-to-end-batch.sh
```diff
@@ -56,7 +56,27 @@ else
   echo "[DEBUG] Skipping building jars"
 fi
 
-cat <<EOF > /tmp/core.warehouse.application.yml
+echo "
+============================================================
+Starting Feast Core
+============================================================
```
woop, line 62: Do we need to make changes to this file?

sdk/python/feast/client.py
```diff
@@ -540,8 +526,26 @@ def get_batch_features(
...
+            >>> # Client with insecure core
             >>> feast_client = Client(core_url="localhost:6565", serving_url="localhost:6566")
+            >>> # Client with secure core (Auth provider: Google OpenID)
+            >>> feast_client = Client(
+            >>>     core_url="localhost:6565",
+            >>>     serving_url="localhost:6566",
+            >>>     core_enable_auth=True,
+            >>>     core_auth_provider="google")
+            >>> # Client with secure core (Auth provider: OAuth)
+            >>> feast_client = Client(
```
woop, line 539: I am not sure how I feel about dumping all this security information here in `get_batch_features`. I think it can be left for a separate section in our documentation or it can be documented as part of Client initialization.

sdk/python/feast/client.py
```diff
@@ -598,10 +604,7 @@ def get_batch_features(
             entity_rows, serving_info.job_staging_location
         )  # type: List[str]
         request = GetBatchFeaturesRequest(
-            features=_build_feature_references(
-                feature_ref_strs=feature_refs,
-                project=project if project is not None else self.project,
```
woop, line 603: Is there a reason for this code to move/change?

tests/e2e/requirements.txt
```diff
@@ -9,5 +9,5 @@ pytest-timeout==1.3.3
 pytest-ordering==0.6.*
 tensorflow-data-validation==0.21.2
 deepdiff==4.3.2
-tensorflow==2.1.0
+tensorflow==2.2.0
```
woop, line 12: This should probably not have been changed right?

jmelinav: will be reverted, It was an effort to fix dependency error that appears in end-to-end tests.

sdk/python/feast/client.py
jmelinav: Will remove this, Client initialization has the documentation already.

core/pom.xml
jmelinav: Done, Please have a look at it and confirm that I haven't missed any of the dependencies that you had included.

woop:
I think we have to get the test scripts merged in through a different PR if we want to execute them on this PR. Would you mind moving the Prow config and test script updates (test-end-to-end.sh) in a separate PR? I can get that merged in and we can quickly run a round of tests.

core/src/main/java/feast/core/service/ProjectService.java
```diff
@@ -16,35 +16,58 @@
...
+import org.springframework.security.access.AccessDeniedException;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.context.SecurityContext;
 import org.springframework.stereotype.Service;
-import org.springframework.transaction.annotation.Transactional;
 
 @Slf4j
 @Service
-public class AccessManagementService {
+public class ProjectService {
```
mrzzy, line 37: Not sure about the change of naming here. Here we are mostly dealing with projects in the context of authentication so it makes more sense to be `AccessManagementService` or `AuthorizationService` instead of `ProjectService`.

core/src/main/resources/application.yml
```diff
@@ -77,12 +70,30 @@ feast:
...
       specsAckTopic: feast-specs-ack
       notifyIntervalMilliseconds: 1000
+      
+  security:
+    authentication:
+      enabled: false
+      provider: jwt
+    authorization:
+      enabled: false
+      provider: none
```
mrzzy, line 84: We should implements some sensible defaults here for `authorization` section with the `keto` authorization provider. The rule of the thumb is that user has to change the least amount of config for `Keto` auth to work.

core/src/main/java/feast/core/service/ProjectService.java
```diff
@@ -16,35 +16,58 @@
...
+    this.authorizationProvider = authorizationProvider.getIfUnique();
+    this.securityProperties = feastProperties.getSecurity();
   }
 
   /**
    * Creates a project
    *
    * @param name Name of project to be created
    */
-  @Transactional
```
mrzzy, line 47: Why is `@Transactional` removed here? Calling a `projectRepository` writes to the database which want to be atomic.

core/src/main/java/feast/core/service/ProjectService.java
```diff
@@ -16,35 +16,58 @@
...
   public void createProject(String name) {
```
mrzzy, line 71: Might be out of scope for this PR it would be nice if the user creating the project will be automatically enrolled in the Authorization provider so that they can delete their own projects.

core/src/main/java/feast/core/auth/authentication/DefaultJwtAuthenticationProvider.java
```diff
@@ -0,0 +1,68 @@
...
+import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.oauth2.jwt.NimbusJwtDecoder;
+import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationConverter;
+import org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationProvider;
+
+/**
+ * Google Open ID Authentication Provider. This provider is used to validate incoming requests to
+ * Feast Core.
+ */
+public class DefaultJwtAuthenticationProvider implements AuthenticationProvider {
```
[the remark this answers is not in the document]
jmelinav: @mrzzy It is not specific to google, if you set `jwkEndpointURI` in `application.yaml`, that will be picked up. Only default URI is google specific.

core/src/main/java/feast/core/config/FeastProperties.java
```diff
@@ -31,9 +39,11 @@
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.info.BuildProperties;
+import org.springframework.stereotype.Component;
 
 @Getter
 @Setter
+@Component
```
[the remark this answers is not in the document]
jmelinav: Good to know. I will remove @Component and if everything works fine.

core/src/main/java/feast/core/service/ProjectService.java
jmelinav: @woop was this change done by you? if not it must be merge/rebase problem.

core/src/main/java/feast/core/service/ProjectService.java
jmelinav: @woop again, If it's not your change should be reverted. Let me know.

core/pom.xml
```diff
@@ -185,7 +228,11 @@
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
         </dependency>
-
+	
+		<dependency>
```
pyalex, line 232: Not sure how kafka is utilized in this PR, but we already have spring-kafka which depends on kafka-client

core/src/main/java/feast/core/service/ProjectService.java
woop: This could have been a change from my side. Let's revert it.

core/src/main/resources/application.yml
woop: Yea that makes sense, but what does this tangibly mean? That we will deploy a Keto instance? If not, then the default configuration won't work.

core/src/main/java/feast/core/service/ProjectService.java
woop: This was done from my side. I am still torn between these two. `ProjectService` is more explicit, but the idea was always that it would be generalized later. I just felt like `AccessManagementService` is a premature generalization.

core/src/main/java/feast/core/auth/authentication/DefaultJwtAuthenticationProvider.java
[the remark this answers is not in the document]
woop: Lets remove the default here and move it to FeastProperties.

PR #798, opened 2020-06-15
Compression of FeatureRows collection in memory

pyalex:

Recently we changed BigQuery writes to use BatchLoad method to be able to update table schema in-flight. In order to produce "successful inserts" output for metrics calculation we keep in memory all FeatureRows until BigQuery load job succeed.
To use memory more efficient this PR proposes to compress FeatureRows that awaiting for bq job to finish by converting them from row-orientation into column-orientation and using generic compression on those columns.

Fixes #

```release-note

```

### DIFF
--- a/storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/compression/FeatureRowsBatch.java
+++ b/storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/compression/FeatureRowsBatch.java
@@ -1,0 +1,228 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright 2018-2020 The Feast Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package feast.storage.connectors.bigquery.compression;
+
+import static feast.proto.types.ValueProto.Value.ValCase.*;
+import static feast.storage.connectors.bigquery.common.TypeUtil.*;
+
+import feast.proto.types.FeatureRowProto;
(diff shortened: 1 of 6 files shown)


storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/compression/FeatureRowsBatch.java
```diff
@@ -0,0 +1,228 @@
...
+    protoToSchemaTypes.put(DOUBLE_LIST_VAL, Schema.FieldType.array(Schema.FieldType.DOUBLE));
+
+    schemaToProtoTypes =
+        protoToSchemaTypes.entrySet().stream()
+            .collect(Collectors.toMap(Map.Entry::getValue, Map.Entry::getKey));
+  }
+
+  public FeatureRowsBatch(Iterable<FeatureRowProto.FeatureRow> featureRows) {
+    schema = inferCommonSchema(featureRows);
+    initValues();
```
woop, line 75: Can we use `this.`, small nitpick.

woop: Also, is there a reason that functionality is not in the constructor?

storage/connectors/bigquery/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BatchLoadsWithResult.java
```diff
@@ -135,6 +142,12 @@
 
     PCollection<WriteBundlesToFiles.Result<DestinationT>> results =
         input
+            .apply("WindowWithTrigger",
+                Window.<KV<DestinationT, TableRow>>configure()
+                    .triggering(
+                        Repeatedly.forever(
+                            AfterPane.elementCountAtLeast(FILE_TRIGGERING_RECORD_COUNT)))
```
woop, line 149: The trigger still happens after some amount of time regardless of whether this count is reached, right?

pyalex: yes, end-of-window trigger is implied here

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/compression/FeatureRowsBatch.java
pyalex: it was just too many lines of code, decided to split them into methods

PR #799, opened 2020-06-16
Allow users compute statistics over retrieved batch datasets

zhilingc:
This PR adds the functionality to allow users to optionally compute tfdv statistics over retrieved batch datasets:

```
# compute_statistics defaults to false. If false, statistics() will always return an empty object.
dataset = client.get_batch_features(feature_refs, entity_rows, compute_statistics=True)
# waits for retrieval job to be complete, then returns tfdv statistics
stats = dataset.statistics()
```
Like with the feature set statistics, these statstics can then be fed into tfdv for validation, or facets to explore the dataset.

Fixes #

```release-note
- Adds the parameter compute_statistics to client.get_batch_statistics()
- Adds the method statistics() which returns statistics from retrieval job
```


protos/feast/serving/ServingService.proto
```diff
@@ -173,6 +177,9 @@ message Job {
     // Output only. The data format for all the files.
     // For CSV format, the files contain both feature values and a column header.
     DataFormat data_format = 6;
+    // Output only. The statistics computed over
+    // the retrieved dataset. Only available for BigQuery stores.
+    tensorflow.metadata.v0.DatasetFeatureStatisticsList dataset_feature_statistics_list = 7;
```
woop, line 182: After this PR, maybe after 0.6, we should definitely encapsulate these final fields (file_uris, data_format, and stats) in a dataset message.

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/statistics/BigQueryStatisticsRetriever.java
```diff
@@ -71,51 +70,48 @@ static Builder newBuilder() {
...
-  public FeatureSetStatistics getFeatureStatistics(
-      FeatureSetSpec featureSetSpec, List<String> features, String dataset) {
-    FeatureSetStatisticsQueryInfo featureSetStatisticsQueryInfo =
-        new FeatureSetStatisticsQueryInfo(
-            featureSetSpec.getProject(), featureSetSpec.getName(), dataset);
-    return getFeatureSetStatistics(featureSetSpec, features, featureSetStatisticsQueryInfo);
+  public FeatureStatistics getFeatureStatistics(
+      FeatureSetSpec featureSetSpec, List<String> features, String ingestionId) {
+    String featureSetTableName =
+        String.format("%s_%s", featureSetSpec.getProject(), featureSetSpec.getName());
```
woop, line 76: Is it possible to reduce duplication between these two methods?

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/statistics/Dataset.java
```diff
@@ -0,0 +1,66 @@
...
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+import org.joda.time.format.DateTimeFormat;
+import org.joda.time.format.DateTimeFormatter;
+
+/**
+ * Dataset represents a subset of data within a table to retrieve statistics over. Data can be
+ * subset by either ingestionId or date.
+ */
+public class Dataset {
```
woop, line 32: I'm a bit iffy about dataset being used as the term for what is more comparable to your `Info` value types. That being said I think this is ok for now but this would be tech debt that we will clear later once we have proper datasets as a concept.

tests/e2e/bq/bq-batch-retrieval.py
```diff
@@ -676,13 +679,50 @@ def test_update_featureset_retrieve_valid_fields(client, update_featureset_dataf
...
+            "entity_id": [i + id_offset for i in range(N_ROWS)],
+            "feature_value3": ["CORRECT"] * N_ROWS,
+        }
     )
+    client.ingest(proc_time_fs, df)
+
+    def check():
+        feature_retrieval_job = client.get_batch_features(
+            entity_rows=df[["datetime", "entity_id"]],
+            feature_refs=["feature_value3"],
```
woop, line 707: Can we test feature refs with the feature reference in the name as well?

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/statistics/Dataset.java
zhilingc: I don't really like it either. Any suggestions? It was initially DataSubset but the two are pretty analogous to each other

woop: Dunno, maybe just taint it with some prefix. We can always pull something out of our bag of generic suffixes.

`StatsDatasetInfo`?

Sounds horrible, but at least it doesn't name squat `Dataset` :)

PR #801, opened 2020-06-18
Add Common module

terryyylim:
Since common functionality can be found across existing modules (core, serving, ingestion, SDK), a common module has also been created in this PR, with tests added. The refactoring of parts of the codebase to be aligned with usage of common module will be incrementally done moving forward.

```release-note
NONE
```


common/.gitignore
```diff
@@ -0,0 +1,36 @@
+### Scratch files ###
```
woop, line 1: Why do we need a new .gitignore?

common/src/main/java/feast/common/function/StringUtils.java
```diff
@@ -0,0 +1,211 @@
...
+import feast.proto.core.FeatureSetProto.FeatureSetSpec;
+import feast.proto.core.FeatureSetReferenceProto.FeatureSetReference;
+import feast.proto.core.StoreProto;
+import feast.proto.serving.ServingAPIProto.FeatureReference;
+import java.util.Arrays;
+import java.util.List;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+public class StringUtils {
```
pyalex, line 28: I would rather have those functions grouped by domain. Also name "StringUtils" doesn't bring clarity here

common/src/main/java/feast/common/function/StringUtils.java
```diff
@@ -0,0 +1,211 @@
...
+  // Subscription-related functionality
+  /**
+   * Accepts a comma-delimited Subscriptions that is string-formatted and converts it to a list of
+   * Subscription class objects.
+   *
+   * @param subscriptions String formatted Subscriptions, comma delimited.
+   * @param exclude flag to determine if subscriptions with exclusion flag should be returned
+   * @return List of Subscription class objects
+   */
+  public static List<StoreProto.Store.Subscription> getSubscriptionsByStr(
```
pyalex, line 83: maybe better name "parseSubscriptionFrom"? similarly to protobuf api https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/Parser

common/src/main/java/feast/common/function/StringUtils.java
terryyylim: I was expecting this haha, had a hard time thinking for a name so pushed for a review to have you guys help brainstorm 😂

terryyylim: By domain you mean `FeatureSet`, `Store` etc?

woop: My trigger words are
* Info
* Spec
* Util
* Helper
* Metadata

Sometimes unavoidable, but we should pay attention when/why they are introduced.

woop:
I don't know why, but I am irrationally excited about finally having a common module. Way overdue.

woop:
@terryyylim I think we should find a balance here between creating a common module and introducing functionality. Introducing the subscription blacklist and refactoring into a common module is ok, but it looks like we also have new methods introduced that are not being used by the rest of the code base. I think those should be excluded until this PR lands, and then introduced as part of a refactor.

common/src/main/java/feast/common/function/StringUtils.java
pyalex: > By domain you mean `FeatureSet`, `Store` etc?

exactly

common/.gitignore
terryyylim: Removed.

common/src/main/java/feast/common/function/StringUtils.java
terryyylim: Updated classes based on domain.

common/src/main/java/feast/common/function/StringUtils.java
terryyylim: Renamed function to follow protobuf api convention.

common/src/main/java/feast/common/function/Feature.java
```diff
@@ -14,25 +14,28 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package feast.serving.util;
+package feast.common.function;
```
woop, line 17: Why do we have to nest this inside of `function`? I thought we wanted to have the functionality be domain specific. How about `models`?

common/src/main/java/feast/common/function/Feature.java
```diff
@@ -14,25 +14,28 @@
...
 
-public class RefUtil {
-  public static String generateFeatureStringRef(FeatureReference featureReference) {
+public class Feature {
+
+  /**
+   * Accepts FeatureReference object and returns its reference in String
+   * "project/featureset_name:feature_name".
+   *
+   * @param featureReference
```
woop, line 27: If you arent going to write a comment then you should link to the type https://stackoverflow.com/questions/5915992/how-to-reference-a-method-in-javadoc

woop: You should also talk about required or optional properties and how they will be handled.

common/src/main/java/feast/common/function/Store.java
```diff
@@ -0,0 +1,163 @@
...
+        .collect(Collectors.toList());
+  }
+
+  /**
+   * Accepts a Subscription class object and returns it in string format
+   *
+   * @param sub Subscription class to be converted to string format
+   * @return String formatted Subscription class
+   */
+  public static String parseSubscriptionFrom(StoreProto.Store.Subscription sub) {
```
woop, line 48: Why do you use `sub` here but `subscription` in the previous method? can you choose one?

common/src/main/java/feast/common/function/Store.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    return String.format("%s:%s:%s", sub.getProject(), sub.getName(), sub.getExclude());
+  }
+
+  /**
+   * Accepts a exclude parameter to determine whether to return subscriptions that are excluded.
+   *
+   * @param sub String formatted Subscription to be converted to Subscription class
+   * @param exclude flag to determine if subscriptions with exclusion flag should be returned
```
woop, line 61: Why is it the responsibility of this method to have this `excluded` flag? Why cant the caller filter the list?

common/src/main/java/feast/common/function/Store.java
```diff
@@ -0,0 +1,163 @@
...
+   *
+   * @param subscriptions List of Subscriptions available in Store
+   * @param projectName Project name used for matching Subscription's Project
+   * @param featureSetName Featureset name used for matching Subscription's Featureset
+   * @return boolean flag to signify if FeatureRow is subscribed to Featureset
+   */
+  public static boolean isSubscribedToFeatureSet(
+      List<StoreProto.Store.Subscription> subscriptions,
+      String projectName,
+      String featureSetName) {
```
woop, line 104: Isn't a feature reference as an argument more natural here?

common/src/test/java/feast/common/function/CommonFunctionsTest.java
```diff
@@ -0,0 +1,197 @@
...
+import java.util.List;
+import org.junit.Before;
+import org.junit.Test;
+import org.tensorflow.metadata.v0.FeaturePresence;
+import org.tensorflow.metadata.v0.FeaturePresenceWithinGroup;
+import org.tensorflow.metadata.v0.FixedShape;
+import org.tensorflow.metadata.v0.IntDomain;
+import org.tensorflow.metadata.v0.ValueCount;
+
+public class CommonFunctionsTest {
```
woop, line 40: This is a meaningless class name, can it be more domain specific or split according tot he classes it tests?

core/src/main/java/feast/core/model/Store.java
```diff
@@ -119,26 +120,18 @@ public static Store fromProto(StoreProto.Store storeProto) throws IllegalArgumen
...
-  }
+  /**
+   * Returns a List of Subscriptions based on whether to include those with excluded flag.
+   * Currently, store's subscriptions string will store ALL subscriptions, but during retrieval for
+   * subscriptions, the exclude flag provided by this function will be used for filtering.
+   *
+   * @param exclude boolean flag to signify whether to return subscriptions which have excluded flag
+   * @return List of Subscription that based on exclusion filter
+   */
+  public List<Subscription> getSubscriptionsByStr(boolean exclude) {
```
woop, line 131: This method makes the caller think that they will pass a string. `ByStr` implies that. Also, this exclusion flag probably is introduced to provide backward compatibility, but it seems out of place. Ideally we would not be filtering the subscription list at this level, but we should have a means to apply that subscription list for detection, for example as you have done with `isSubscribedToFeatureSet`. 

What would the impact be of removing this field?

common/src/main/java/feast/common/function/Store.java
terryyylim: So, for this function, its written to abstract out the usage in these 2 places, [1](https://github.com/feast-dev/feast/blob/master/ingestion/src/main/java/feast/ingestion/transform/FeatureRowToStoreAllocator.java#L78) and [2](https://github.com/feast-dev/feast/blob/master/ingestion/src/main/java/feast/ingestion/transform/specs/FilterRelevantFunction.java#L46). Wanted to keep the original intent of the function, unless I try to refactor the parts which are using this as well.

common/src/main/java/feast/common/models/Feature.java
```diff
@@ -14,25 +14,28 @@
...
+
+  /**
+   * Accepts FeatureReference object and returns its reference in String
+   * "project/featureset_name:feature_name".
+   *
+   * @param featureReference {@link FeatureReference}
+   * @return String format of FeatureReference
+   */
+  public static String getFeatureStringRef(
+      FeatureReference featureReference, boolean ignoreProject) {
```
woop, line 31: `ignoreProject` is undocumented. Is there a reason the change is in this PR?

terryyylim: Previously in RequestUtil [here](https://github.com/feast-dev/feast/pull/801/files#diff-04c8ea2f153f6312e191d7cbace3d298L95), this returns a string without the project, whereas other use-cases return with project in the string.

woop: Ok, this needs to be cleaned up later though. I would rather use overloaded methods here in any case.

Please make sure to update the comments.

terryyylim: Ok, will clean this up. I've updated the docs to explain the parameter for now.

Issue #742, opened 2020-05-26
Fix Telecom Customer Churn Prediction Example Notebook


bradaallen:
Three small issues with the telco churn walkthrough:
1. The docker container used for loading the example notebooks has all of the packages for the "basic" notebook environment, but not for the "telco" notebook. (eg, `pillow`, `python-slugify`, etc.)
2. The environmental variables do not need to be set, as they are part of the config in the the docker-compose. The environmental variables set to `localhost`, but in the config are `core`. Using the code to set the environmental variables in the notebook results in an error.
3. For making predictions, there is an `entity_df` variable that is referenced in the telco notebook - but that variable is not instantiated anywhere.

woop:
Thanks for this @bradaallen

woop:
Should be fixed in #803

PR #803, opened 2020-06-18
Build Feast Jupyter image and clean up examples

woop:

* Builds a base image for Jupyter (`gcr.io/kf-feast/feast-jupyter:latest`) so that it's faster for new users to start the Jupyter container for Docker Compose. Feast dependencies are still installed upon startup, but most of them should now already be available.
* Cleaned up the Basic example a little bit.
* Fixed `entity_df` bug in Telco example and added more details

Fixes: #742


algattik:
Funny, I've done exactly the same thing today on our fork :) 

Here's my dockerfile (maximizing layer reuse). It seems to me that there is no need anymore for the startup script to download and install stuff, and downloading release versions in the dev cycle makes config management harder?

```
FROM jupyter/minimal-notebook:619e9cc2fc07

USER root

WORKDIR /feast

# Init empty .git settings (required by SDK setup.py)
RUN git init .

COPY sdk/python sdk/python
COPY Makefile Makefile
COPY protos protos
COPY tests tests
COPY README.md README.md

# Install Python dependencies
RUN make compile-protos-python

# Install CI requirements (only needed for running tests)
RUN pip install -r sdk/python/requirements-ci.txt

# Install Feast SDK
RUN pip install -e sdk/python -U

# Switch back to original user and workdir
USER $NB_UID
WORKDIR $HOME

CMD start-notebook.sh --NotebookApp.token=''
```

woop:
@algattik neat!

I might reuse some of yours. I will still be cloning the repository since the image I am building is not meant for development, it's meant to have the latest stable dependencies preinstalled, and I guess we can have the actual SDK also preinstalled.

I don't like mapping in the examples directly from the repo because master is normally not consistent with the stable branch, and if we pin versions then its very hard to backport tutorials or make small tweaks.

woop:
@algattik Made some changes to find a better middle ground. Hope it looks better :)

I didn't attribute the Dockerfile to you, but normally people don't care :medal_military:

infra/docker-compose/docker-compose.yml
```diff
@@ -22,21 +22,23 @@ services:
...
       FEAST_ONLINE_SERVING_URL: online-serving:6566
       FEAST_BATCH_SERVING_URL: batch-serving:6567
       GOOGLE_APPLICATION_CREDENTIALS: /etc/gcloud/service-accounts/key.json
-      FEAST_REPOSITORY_VERSION: ${FEAST_REPOSITORY_VERSION}
     ports:
       - 8888:8888
-    command: ["/etc/startup.sh"]
+    command: >
+      bash -c "
+      git clone https://github.com/feast-dev/feast.git || true
```
mrzzy, line 39: Is there a reason why can't we bake in the Feast repo into the docker image?

infra/docker/jupyter/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM jupyter/minimal-notebook:619e9cc2fc07
+
+USER root
+WORKDIR /feast
+
+# .git (required by SDK setup.py)
```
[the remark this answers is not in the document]
woop: This image should also be built for unreleased Python SDK versions, so master for example. The image version should correspond to the SDK version, so we can't just use latest.

infra/docker-compose/docker-compose.yml
woop: The `latest` image will only be built on a release, which means that if we want to roll out the latest tutorials then we have to cut a release. Often I find myself updating docs/tutorials between releases, which is why I have pinned this `clone` to the branch instead. In theory we could build the `latest` image on every commit, but that image will go stale very quickly on the user's side (they have to repull the whole time).

infra/docker/jupyter/Dockerfile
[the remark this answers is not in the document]
zhilingc: Then I think i'd prefer if you used a builder image to build, then copy just the sdk folder over.  Seems cleaner.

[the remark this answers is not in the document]
woop: Alright, are you trying to avoid having unnecessary folders or unnecessary Python dependencies?

infra/docker/jupyter/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+COPY .git .git
```
algattik, line 7: .git and README:md are only needed by " pip install -e sdk/python" IIRC, so if you move them down you can reuse layer with requirements installed. (.git folder changes with every commit)

algattik:
> @algattik Made some changes to find a better middle ground. Hope it looks better :)
> 
> I didn't attribute the Dockerfile to you, but normally people don't care 🎖️

My ego shall survive :)

infra/docker/jupyter/Dockerfile
woop: Corrected, thanks.

PR #804, opened 2020-06-18
Fix Optional#get() and string comparison bugs in JobService

ches:

Fixes some bugs in the Ingestion Job Management service.

Cleaning house a bit and I had a todo [from here][1] about the `get()` calls that will always fail. These don't actually have much impact since `get()` already throws `NoSuchElementException` which we're doing anyway, but we'd lose the message we try to set.

Actually perhaps more concerning is a bunch of `==` comparisons on strings, those might be a source of more functionally significant bugs for filtering in the list jobs API. I'm surprised that no tests trip something here—a little short on time to look more closely at the existing tests now, apologies.

These issues are all flagged by IntelliJ. The `==` and `this.` receiver for all class member references smell like the author was writing mostly Python at the time 😉 I'm sympathetic to not always liking IDEs, but if you use a lighter editor I'd recommend setting up something like the SpotBugs Maven plugin, which would be nice to do anyway IMO.

[1]: https://github.com/feast-dev/feast/pull/650#discussion_r415125907

```release-note
NONE
```

### DIFF
--- a/core/src/test/java/feast/core/service/JobServiceTest.java
+++ b/core/src/test/java/feast/core/service/JobServiceTest.java
@@ -46,11 +46,7 @@ import feast.proto.core.StoreProto.Store.RedisConfig;
 import feast.proto.core.StoreProto.Store.StoreType;
 import feast.proto.types.ValueProto.ValueType.Enum;
 import java.time.Instant;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Date;
-import java.util.List;
-import java.util.Optional;
+import java.util.*;
 import org.junit.Before;
 import org.junit.Test;
 import org.mockito.Mock;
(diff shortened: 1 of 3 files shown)


core/src/test/java/feast/core/service/JobServiceTest.java
```diff
@@ -345,6 +340,12 @@ public void testStopUnsupportedError() {
     this.job.setStatus(prevStatus);
   }
 
+  @Test(expected = NoSuchElementException.class)
+  public void testStopJobForUnknownId() {
+    var request = StopIngestionJobRequest.newBuilder().setId("bogusJobId").build();
+    jobService.stopJob(request);
```
ches, line 346: This didn't fail before the fix since it was the same exception type anyway. I don't think there's an easy way to assert on a thrown exception's message in JUnit 4.

ches: I think Gojek folks tried/wanted to before—the Java SDK already uses it—but there was some reason it could not or should not be done? IIRC either @davidheryanto or @zhilingc mentioned it to me somewhere awhile back.

mrzzy:
Thanks for fixing my bad code practices. :disappointed_relieved: 

> These issues are all flagged by IntelliJ. The == and this. receiver for all class member references smell like the author was writing mostly Python at the time.

This was pretty much on point :open_mouth:, this is my first stab at writing Java in a production system.

> I'm sympathetic to not always liking IDEs, but if you use a lighter editor I'd recommend setting up something like the SpotBugs Maven plugin, which would be nice to do anyway IMO.
Will check that out.

PR #807, opened 2020-06-19
Test end to end with authentication.

jmelinav:
Enables end to end test with authentication. Related to #504 and PR #793

Fixes #

```release-note
NONE
```

### DIFF
--- a/tests/e2e/redis/basic-ingest-redis-serving.py
+++ b/tests/e2e/redis/basic-ingest-redis-serving.py
@@ -50,9 +50,21 @@ def allow_dirty(pytestconfig):
 
 
 @pytest.fixture(scope='module')
+def enable_auth(pytestconfig):
+    return True if pytestconfig.getoption(
+        "enable_auth").lower() == "true" else False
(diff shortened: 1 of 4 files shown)


tests/e2e/redis/basic-ingest-redis-serving.py
```diff
@@ -49,10 +49,22 @@ def allow_dirty(pytestconfig):
...
+        "enable_auth").lower() == "true" else False
+        
+        
 @pytest.fixture(scope='module')
 def client(core_url, serving_url, allow_dirty):
     # Get client for core and serving
     client = Client(core_url=core_url, serving_url=serving_url)
+    # if enable_auth is True, Google Id token will be 
+    # passed in the metadata for authentication. 
+    client = Client(core_url=core_url,
```
woop, line 64: It's not clear why the Client is created twice here. Should the first one be removed?

jmelinav: It should be.

PR #809, opened 2020-06-19
Introducing Flyway as DB schema migration manager

pyalex:

Adding Flyway to project.
Existing DB schema added as V1. It's not possible to sync migration version with our release versions, since first baseline version name will be always V1.
Fake migration will be run to sync with current schema on first core start.

Fixes #

```release-note

```

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -235,5 +235,10 @@
             <version>6.1.2.Final</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.flywaydb</groupId>
+            <artifactId>flyway-core</artifactId>
+            <version>${flyway.version}</version>
+        </dependency>
     </dependencies>
 </project>
--- a/core/src/main/resources/application.yml
+++ b/core/src/main/resources/application.yml
@@ -85,12 +85,14 @@ spring:
         merge:
           entity_copy_observer: allow
     hibernate.naming.physical-strategy=org.hibernate.boot.model.naming: PhysicalNamingStrategyStandardImpl
-    hibernate.ddl-auto: update
+    hibernate.ddl-auto: validate
   datasource:
     driverClassName: org.postgresql.Driver
     url: jdbc:postgresql://${DB_HOST:127.0.0.1}:${DB_PORT:5432}/${DB_DATABASE:postgres}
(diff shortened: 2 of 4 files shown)


woop:
@pyalex We probably need a new page underneath `Troubleshooting` ([link](https://docs.feast.dev/administration/troubleshooting)) for `Upgrading Feast`. We can just keep it simple for now, but what I was thinking is just having a release specific process (steps to upgrade from 0.4 to 0.5), and a section for common tasks like applying a migration. This PR should probably introduce the latter.

pyalex:
@woop flyway is integrated with spring-boot and all migrations will be happen automatically with start of `core`. There's no need for any manual steps. However such page in documentation still maybe useful to explain what was changed and what issues may occur

ches:
Any comment on why this was closed, change of plans?

ches:
Superseded by #810?

ches:
> flyway is integrated with spring-boot and all migrations will be happen automatically with start of `core`. There's no need for any manual steps.

Some organizations will have database operations managed by dedicated DBA teams, production DDL needs to run through formal change management processes, etc. (Speaking from experience here). Flyway can still help prepare scripts for such procedures, and Feast docs don't need to duplicate things that Flyway docs cover about how this might work, but I absolutely agree that Feast docs should cover what will happen, ideally with some acknowledgement that implicit auto will not be usable by everyone.

This ought to cover initial deployment of Feast in the future too IMO, where environments again may not allow automatic creation of schema by Hibernate or Flyway (these kinds of things are all disabled by Hibernate's defaults, Feast config sets Spring's `hibernate.ddl-auto: validate`, still—which is something that might need to be reviewed given Flyway should own management of schema now).

woop:
@ches I believe the idea was that the backport wasn't necessary since Spring ignores the baseline migration if a schema already exists, so folks coming from 0.5 would just get the diffs in 0.6.

You have a valid point though, in that not all organizations have the freedom to use implicit auto. This seems to only be a problem if you are baselined prior to 0.6.

What would be an acceptable minimum here? Should the migrations start from an earlier version? Should we document a means of upgrading versions (without migrations, perhaps import/export of schemas)?

PR #815, opened 2020-06-22
Create table in BigQuery if doesn't exists when new FeatureSetSpec arrived to IngestionJob

pyalex:

Currently we create table in BQ only with first batch insert, when first data is arrived.
To speed up e2e tests this PR changes it, so table would be created inside IngestionJob when new FeatureSetSpec is received.

Fixes #

```release-note

```

### DIFF
--- a/storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/writer/FeatureSetSpecToTableSchema.java
+++ b/storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/writer/FeatureSetSpecToTableSchema.java
@@ -42,6 +42,14 @@ import org.apache.beam.sdk.transforms.DoFn;
 import org.apache.beam.sdk.values.KV;
 import org.slf4j.Logger;
 
+/**
+ * Converts {@link feast.proto.core.FeatureSetProto.FeatureSetSpec} into BigQuery schema Serializes
+ * it into json-like format {@link TableSchema} Fetches existing schema to merge existing fields
+ * with new ones.
+ *
+ * <p>As a side effect this Operation may create bq table (if it doesn't exist) to make
+ * bootstrapping faster
+ */
 public class FeatureSetSpecToTableSchema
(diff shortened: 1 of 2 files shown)


storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/writer/FeatureSetSpecToTableSchema.java
```diff
@@ -76,19 +84,42 @@ public void processElement(
...
+        tableReference.getProjectId(), tableReference.getDatasetId(), tableReference.getTableId());
+  }
+
+  private Table getExistingTable(String specKey) {
+    return bqService.getTable(generateTableId(specKey));
+  }
+
+  private void createTable(String specKey, Schema schema) {
+    TimePartitioning timePartitioning =
+        TimePartitioning.newBuilder(TimePartitioning.Type.DAY).setField("event_timestamp").build();
```
zhilingc, line 112: Maybe use the constant `EVENT_TIMESTAMP_COLUMN` here instead?

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/writer/FeatureSetSpecToTableSchema.java
```diff
@@ -76,19 +84,42 @@ public void processElement(
       @Element KV<String, FeatureSetProto.FeatureSetSpec> element,
       OutputReceiver<KV<String, TableSchema>> output,
       ProcessContext context) {
-    Schema schema = createSchemaFromSpec(element.getValue(), element.getKey());
-    output.output(KV.of(element.getKey(), serializeSchema(schema)));
+    String specKey = element.getKey();
+
+    Table existingTable = getExistingTable(specKey);
+    Schema schema = createSchemaFromSpec(element.getValue(), specKey, existingTable);
```
zhilingc, line 90: should building the schema be moved down before `createTable(specKey, schema);`? just for a bit more efficiency

PR #816, opened 2020-06-22
Add missing key count metric

terryyylim:
[missingKeyCount](https://github.com/feast-dev/feast/blob/master/serving/src/main/java/feast/serving/util/Metrics.java#L41 ) metric used to be available in Prometheus but was removed during refactor to OnlineServingService and RedisOnlineRetriever, from [RedisServingService](https://github.com/feast-dev/feast/blob/cc884b4480c50555d8a9f28e0629d138d0d9bd36/serving/src/main/java/feast/serving/service/RedisServingService.java#L220).

```release-note
NONE
```

### DIFF
--- a/serving/src/main/java/feast/serving/util/Metrics.java
+++ b/serving/src/main/java/feast/serving/util/Metrics.java
@@ -46,14 +46,6 @@ public class Metrics {
           .labelNames("project", "feature_name")
           .register();
 
-  public static final Counter invalidEncodingCount =
-      Counter.build()
-          .name("invalid_encoding_feature_count")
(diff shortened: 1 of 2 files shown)


serving/src/main/java/feast/serving/util/Metrics.java
```diff
@@ -40,20 +40,12 @@
 
   public static final Counter missingKeyCount =
```
zhilingc, line 41: Maybe just rename this to match the metric name?

terryyylim: Renamed.

PR #817, opened 2020-06-22
Consolidate ingestion jobs to one job per source

pyalex:

All IngestionJobs assigned to the same physical source are now consolidated into one that writes to all sinks subscribed to this source.

Refactoring:
* JobUpdateTask is dead
* Job to Stores - is now m2m relationship

Fixes #

```release-note

```


woop:
To be clear, the consolidation is to reduce the amount of simultaneous jobs.

core/src/main/java/feast/core/job/CheckJobStatusTask.java
```diff
@@ -0,0 +1,46 @@
...
+import feast.core.model.Job;
+import feast.core.model.JobStatus;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+@Builder(setterPrefix = "set")
+public class CheckJobStatusTask implements JobTask {
```
woop, line 29: Should this be check or update?

core/src/main/java/feast/core/job/CreateJobTask.java
```diff
@@ -0,0 +1,80 @@
...
+import lombok.Builder;
+import lombok.Getter;
+import lombok.Setter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Getter
+@Setter
+@Builder(setterPrefix = "set")
+public class CreateJobTask implements JobTask {
```
woop, line 34: It would be great if we could get Javadocs for these job task classes, even if only a brief description.

core/src/main/java/feast/core/util/StreamUtil.java
```diff
@@ -0,0 +1,37 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package feast.core.util;
+
+import java.util.function.Function;
+
+public class StreamUtil {
```
woop, line 21: Can we add documentation to this class? It seems specific to exception handling not streams per say.

core/src/main/java/feast/core/service/JobCoordinatorService.java
```diff
@@ -143,68 +137,111 @@ void startOrUpdateJobs(List<JobUpdateTask> tasks) {
...
+        // (another words, it exists in kafka)
+        // so we expect Job will ack latest version based on history from kafka topic
+        status.setVersion(fs.getVersion());
+      }
+      status.setDeliveryStatus(FeatureSetProto.FeatureSetJobDeliveryStatus.STATUS_IN_PROGRESS);
+      job.getFeatureSetJobStatuses().add(status);
+    }
   }
 
   /** Get the non terminated ingestion job ingesting between source and source if exists. */
```
woop, line 232: I think this comment should be source and stores.

core/src/main/java/feast/core/job/UpgradeJobTask.java
```diff
@@ -22,6 +22,7 @@
 import lombok.Getter;
 import lombok.Setter;
 
+/** Task that upgrade given {@link Job} by restarting it in {@link JobManager} */
```
woop, line 25: Is there a difference between an update and an upgrade, and what does an `upgrade` actually mean?

pyalex: I was looking for separation between update status (pooling smth) and upgrade job (pushing changes). Naming smth is hardest programming task

core/src/main/java/feast/core/model/Job.java
```diff
@@ -98,8 +111,8 @@ public boolean isRunning() {
     return getStatus() == JobStatus.RUNNING;
   }
 
-  public String getSinkName() {
-    return store.getName();
+  public boolean isDeployed() {
```
woop, line 114: Does `isDeployed` mean it is running? Or could have an `id` without actually being live?

core/src/main/java/feast/core/service/JobCoordinatorService.java
```diff
@@ -143,68 +137,111 @@ void startOrUpdateJobs(List<JobUpdateTask> tasks) {
...
     }
 
-    return jobUpdateTasks;
+    return jobTasks;
   }
 
-  /** Get the non terminated ingestion job ingesting between source and source if exists. */
-  @Transactional
-  public Optional<Job> getJob(Source source, Store store) {
+  private boolean jobRequiresUpgrade(Job job, Set<Store> stores) {
```
woop, line 199: For both `jobRequiresUpgrade` and `allocateFeatureSets` it would be great to have a short comment on what they are doing. Does `jobRequiresUpgrade` only care about feature set consistency, or also sinks or stores or other configuration? What does allocation mean?

core/src/main/java/feast/core/job/UpgradeJobTask.java
woop: Yea that make sense, but this method is called `upgradeJob` but it calls `updateJob`, which is why I asked. Should updateJob be renamed as well? Not sure I feel too strongly on this, just asking.

core/src/main/java/feast/core/model/Job.java
pyalex: means it has extId, so it could be running, or could be aborting

PR #819, opened 2020-06-23
Prevent reserved fields from being registered

### LINKED ISSUE #818: Feature set spec with reserved spec should not be allowed to be registered
## Expected Behavior 
If a feature spec contains reserved column name such as `created_timestamp`, Feast should not allow the feature set to be registered at all.

## Current Behavior
The feature set get registered, and job update fails due to duplicated column in bigquery schema.

## Steps to reproduce
Create a feature spec that contains reserved column name.

## Possible Solution
Feast client should perform verification on feature set spec prior to registering the feature set to Feast core.

terryyylim:
This PR prevents applying of Featuresets with Feature names that are reserved.

Fixes #818

```release-note
NONE
```

### DIFF
--- a/core/src/main/java/feast/core/validators/FeatureSetValidator.java
+++ b/core/src/main/java/feast/core/validators/FeatureSetValidator.java
@@ -30,6 +30,9 @@ import org.apache.commons.lang3.StringUtils;
 
 public class FeatureSetValidator {
 
+  private static List<String> reservedNames =
+      Arrays.asList("created_timestamp", "event_timestamp", "ingestion_id", "job_id");
+
   public static void validateSpec(FeatureSet featureSet) {
     if (featureSet.getSpec().getProject().isEmpty()) {
       throw new IllegalArgumentException("Project name must be provided");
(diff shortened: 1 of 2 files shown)


core/src/main/java/feast/core/validators/FeatureSetValidator.java
```diff
@@ -64,4 +67,18 @@ private static void checkUniqueColumns(
           String.format("fields within a featureset must be unique."));
     }
   }
+
+  private static void checkReservedColumns(List<FeatureSpec> featureSpecs) {
+    List<String> reservedNames =
+        Arrays.asList("created_timestamp", "event_timestamp", "ingestion_id", "job_id");
```
woop, line 73: These should be defined elsewhere as constants, not within the method.

core/src/main/java/feast/core/validators/FeatureSetValidator.java
```diff
@@ -64,4 +67,18 @@ private static void checkUniqueColumns(
...
+    String reservedNamesString = StringUtils.join(reservedNames, ", ");
+    for (FeatureSpec featureSpec : featureSpecs) {
+      if (reservedNames.contains(featureSpec.getName())) {
+        throw new IllegalArgumentException(
+            String.format(
+                "Reserved feature names have been used, which are not allowed. These names include %s.",
```
woop, line 79: Please include the field the user tried to register as well.

PR #820, opened 2020-06-23
Remove Duplicated Strip Projects Code from SDKs

mrzzy:

## Problem
Continuation of: https://github.com/feast-dev/feast/pull/693.

In v0.5 SDK, projects stripped from string Feature References in individual SDKs. This is a source of code duplication as the same strip project code is implemented and re implemented in the 3 SDKs. This duplication makes maintenance difficult as we have to maintain the same logic in all 3 SDKs.

## What does the PR do
- Add `project` to `GetOnlineFeaturesRequest` as optional field in ServingService.proto. 
    - When the user specifies the `project` with project name, Serving get online features from the specified `project`.
    - The `GetOnlineFeaturesRequest`'s `project` field will take precedence over `project` specified in `FeatureReferences` if both are specified.
- Update the new v0.6 SDK to use `GetOnlineFeatureRequest`'s  `project` field.
- Remove for v0.6 Strip projects code inf SDKs.

Fixes #

```release-note
Added project field to GetOnlineFeaturesRequest in ServingService proto to override project in request.
```


sdk/java/src/main/java/com/gojek/feast/RequestUtil.java
```diff
@@ -53,29 +46,19 @@
...
     if (featureRefString.contains("/")) {
-      if (ignoreProject) {
-        String[] projectSplit = featureRefString.split("/");
-        featureRefString = projectSplit[1];
-      } else {
-        throw new IllegalArgumentException(
-            String.format("Unsupported feature reference: %s", featureRefString));
-      }
+      throw new IllegalArgumentException(
+          String.format("Unsupported feature reference: %s", featureRefString));
```
woop, line 58: Can we be more specific with this exception?

sdk/python/feast/client.py
```diff
@@ -671,10 +671,9 @@ def get_online_features(
             entity_rows: List of GetFeaturesRequest.EntityRow where each row
                 contains entities. Timestamp should not be set for online
                 retrieval. All entity types within a feature
-            project: Specifies the project which contain the FeatureSets
-                which the requested features belong to.
+            project: Optionally specify the the project override. If specified, uses given project for retrieval.
+                Overrides the projects specified in Feature References if also are specified.
```
woop, line 675: I think you can remove that final `are`

serving/src/main/java/feast/serving/specs/CachedSpecService.java
```diff
@@ -103,69 +103,99 @@ public FeatureSetSpec getFeatureSetSpec(String featureSetRef) throws ExecutionEx
     return featureSetCache.get(featureSetRef);
   }
 
+  /**
+   * Get FeatureSetSpecs for the given features references. See {@link #getFeatureSets(List,
+   * String)} for full documentation.
+   */
+  public List<FeatureSetRequest> getFeatureSets(List<FeatureReference> featureReferences) {
+    return getFeatureSets(featureReferences, "");
```
woop, line 111: Out of curiosity, is there a reason that the default project could not be used here?

serving/src/main/java/feast/serving/specs/CachedSpecService.java
```diff
@@ -103,69 +103,99 @@ public FeatureSetSpec getFeatureSetSpec(String featureSetRef) throws ExecutionEx
...
-              }
+            (fsName, fsNamesAndfeatureRefs) -> {
+              List<FeatureReference> featureRefs =
+                  fsNamesAndfeatureRefs.stream().map(Pair::getRight).collect(Collectors.toList());
+              featureSetRequests.add(buildFeatureSetRequest(fsName, featureRefs));
             });
     return featureSetRequests;
   }
 
+  /** Build a Feature Set request from the Feature Set specified by name and Feature References */
```
woop, line 149: A full Javadoc comment would be nice.

serving/src/main/java/feast/serving/specs/CachedSpecService.java
```diff
@@ -103,69 +103,99 @@ public FeatureSetSpec getFeatureSetSpec(String featureSetRef) throws ExecutionEx
...
+  private FeatureSetRequest buildFeatureSetRequest(
+      String featureSetName, List<FeatureReference> featureReferences) {
+    // get feature set for name
+    FeatureSetSpec featureSetSpec;
+    try {
+      featureSetSpec = featureSetCache.get(featureSetName);
+    } catch (ExecutionException e) {
```
woop, line 156: One of the major pain point that I have seen is when users are requesting features that dont exist because a store is not subscribed to it. It maybe out of scope for this PR, but is there a way that we can address that problem?

Basically they should get an exception that a feature set exists but is not being populated, instead of just "this feature or feature set doesnt exist".

mrzzy: I think one way we can resolve this populate the feature set cache from core wholesale and apply the subscription checking only when request is made. This way we be able to determine if the request is rejected due missing feature set or because subscription was not applied properly.

serving/src/main/java/feast/serving/specs/CachedSpecService.java
mrzzy: Corrected.

sdk/java/src/main/java/com/gojek/feast/RequestUtil.java
mrzzy: Corrected.

serving/src/main/java/feast/serving/specs/CachedSpecService.java
mrzzy: When `project` is `""` we want to revert the existing behavior of inspecting the Feature References for the project use in the request, in addition to the default project.

Updated docs to clarify this.

PR #822, opened 2020-06-23
Send acknowledgment on Spec Update only after sinks are ready

pyalex:

Fixes #

```release-note

```

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -50,6 +50,11 @@
             <version>${project.version}</version>
             <scope>compile</scope>
         </dependency>
+        <!--compileOnly 'org.projectlombok:lombok:1.18.12'-->
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+        </dependency>
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
--- a/common/src/main/java/feast/common/models/FeatureSetReference.java
+++ b/common/src/main/java/feast/common/models/FeatureSetReference.java
@@ -1,0 +1,32 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright 2018-2020 The Feast Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package feast.common.models;
+
+import java.io.Serializable;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@Data
+@AllArgsConstructor
+public class FeatureSetReference implements Serializable {
+  private String reference;
+  private Integer version;
+
+  // Empty constructor required for Avro decoding.
+  @SuppressWarnings("unused")
+  public FeatureSetReference() {}
+}
--- a/ingestion/src/main/java/feast/ingestion/transform/specs/ReadFeatureSetSpecs.java
+++ b/ingestion/src/main/java/feast/ingestion/transform/specs/ReadFeatureSetSpecs.java
@@ -19,6 +19,8 @@ package feast.ingestion.transform.specs;
 import com.google.auto.value.AutoValue;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
+import feast.common.models.FeatureSetReference;
+import feast.proto.core.FeatureSetProto;
 import feast.proto.core.FeatureSetProto.FeatureSetSpec;
 import feast.proto.core.IngestionJobProto;
 import feast.proto.core.SourceProto;
@@ -26,6 +28,9 @@ import feast.proto.core.StoreProto;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
+import org.apache.beam.sdk.coders.AvroCoder;
+import org.apache.beam.sdk.coders.KvCoder;
+import org.apache.beam.sdk.extensions.protobuf.ProtoCoder;
 import org.apache.beam.sdk.io.kafka.KafkaIO;
 import org.apache.beam.sdk.transforms.*;
 import org.apache.beam.sdk.transforms.windowing.AfterPane;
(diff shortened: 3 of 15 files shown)


common/src/main/java/feast/common/models/FeatureSetReference.java
```diff
@@ -0,0 +1,32 @@
...
+package feast.common.models;
+
+import java.io.Serializable;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@Data
+@AllArgsConstructor
+public class FeatureSetReference implements Serializable {
+  private String reference;
```
woop, line 26: Is this the feature set reference in string format? Could it be individual fields?

woop:
Let me know once this is ready for review, still missing a PR description.

common/pom.xml
```diff
@@ -50,6 +50,11 @@
             <version>${project.version}</version>
             <scope>compile</scope>
         </dependency>
+        <!--compileOnly 'org.projectlombok:lombok:1.18.12'-->
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
```
woop, line 56: We need to make a decision on Lombok. Personally I am open to using it, but the team previously decided to phase it out and to use purely AutoValue. Apparently they are not compatible.

common/src/main/java/feast/common/models/FeatureSetReference.java
```diff
@@ -0,0 +1,37 @@
...
+import java.io.Serializable;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@Data
+@AllArgsConstructor
+public class FeatureSetReference implements Serializable {
+  private String projectName;
+  private String featureSetName;
+  private Integer version;
```
woop, line 28: Ideally we should have comments on all fields and classes inside `models` because they should rarely change and will be important for understanding Feast, but especially this version since it overlaps with the old feature set versions.

ingestion/src/main/java/feast/ingestion/transform/specs/ReadFeatureSetSpecs.java
```diff
@@ -102,6 +110,26 @@ public abstract Builder setSpecsStreamingUpdateConfig(
                       featureSetSpecs.sort(
                           Comparator.comparing(FeatureSetSpec::getVersion).reversed());
                       return featureSetSpecs.get(0);
-                    }));
+                    }))
+        .apply("FeatureSetReferenceKey", ParDo.of(new CreateFeatureSetReference()))
```
woop, line 114: Should this be a verb perhaps? ` `CreateFeatureSetReferenceKey`?

ingestion/src/main/java/feast/ingestion/transform/specs/WriteFeatureSetSpecAck.java
```diff
@@ -17,29 +17,38 @@
...
 import org.apache.beam.sdk.values.PDone;
 import org.apache.kafka.common.serialization.ByteArraySerializer;
 import org.apache.kafka.common.serialization.StringSerializer;
+import org.joda.time.Duration;
 
 /**
- * Converts input {@link feast.proto.core.FeatureSetProto.FeatureSetSpec} into {@link
+ * Collects output from sinks prepareWrite (several streams flatten into one). As soon as count of
+ * each FeatureSetReference reach getSinksCount() - it means that enough amount of sinks updated its
+ * state - ack is pushed.
```
woop, line 40: What happens if that getSinksCount() is not reached? (in case of failure)

pyalex: Spec wouldn't be acknowledged. FeatureSet would remain in PENDING state.
Issue must be fixed and Job must be restarted - it will reread specs and acknowledge. Currently all issues are gonna be related to BigQuery: either bq unavailable (very low probability) or we couldn't merge schema (so it makes sense not to ack such spec).

common/pom.xml
pyalex: Right, so I think AutoValue is more heavily used, since it's everywhere in Beam and I also applied it a lot in ingestion. So I changed to AutoValue.

woop: Ok, so lets continue to use Lombok for the time being. I hope this doesnt interfere with the rest of the code base.

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/writer/FeatureSetSpecToTableSchema.java
```diff
@@ -99,7 +102,7 @@ public void processElement(
       createTable(specKey, schema);
```
zhilingc, line 102: This comment probably comes late, but since table updates are purely additive (and backward compatible), is there a downside to running `createTable` even if the table already exists?

common/pom.xml
pyalex: "as key" I meant for grouping operations of course

ingestion/src/main/java/feast/ingestion/transform/specs/WriteFeatureSetSpecAck.java
```diff
@@ -66,20 +78,50 @@ public PDone expand(PCollection<KV<String, FeatureSetProto.FeatureSetSpec>> inpu
...
+
+    @Override
+    public PCollection<FeatureSetReference> expand(PCollection<FeatureSetReference> input) {
+      return input
+          .apply(
+              "OnEveryElementTrigger",
+              Window.<FeatureSetReference>into(new GlobalWindows())
+                  .accumulatingFiredPanes()
+                  .triggering(Repeatedly.forever(AfterPane.elementCountAtLeast(1)))
+                  .withAllowedLateness(Duration.ZERO))
```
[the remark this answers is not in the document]
pyalex: I'm not quite sure, but Beam demands specified lateness

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/writer/FeatureSetSpecToTableSchema.java
pyalex: I suppose it's no point to run createTable if table already exists. I guess you meant updateTable, but we're already doing it as part of batch load process, which 100% guaranteed to be consistent approach and doing it here - I'm not sure about consistency.

common/pom.xml
pyalex: ok, I guess @zhilingc is right. I could use `SchemaCoder`. Of course not through decorator `@DefaultSchema()` since it's common module and we don't want to have dependency on beam here. But that's still feasible. I guess we can move towards AutoValue.

ingestion/src/main/java/feast/ingestion/transform/specs/WriteFeatureSetSpecAck.java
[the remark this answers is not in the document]
zhilingc: oke

storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/writer/FeatureSetSpecToTableSchema.java
zhilingc: Yeah, i meant updateTable. By sketchy consistency do you mean the race condition between in-progress writes and feature-set updates?

PR #825, opened 2020-06-24
Allow ingestion job grouping/consolidation to be configurable

pyalex:

Decision regarding how to group sources and stores into IngestionJobs is extracted from `JobCoordinatorService` to `JobGroupingStrategy`. Thus it's now configurable with `feast.jobs.consolidate-jobs-per-source` whether to create one job per Source (and write to all stores from one job) or one job per pair Source-Store and write only to one store in job. The latter is default.

Fixes #

```release-note
It is possible to enable a job grouping strategy which allows ingestion jobs to be consolidated by source.
```

### DIFF
--- a/core/src/main/java/feast/core/config/JobConfig.java
+++ b/core/src/main/java/feast/core/config/JobConfig.java
@@ -68,10 +68,20 @@ public class JobConfig {
         .build();
   }
 
+  /**
+   * Returns Grouping Strategy which is responsible for how Ingestion would be split across job
+   * instances (or how Sources and Stores would be grouped together). Choosing strategy depends on
+   * FeastProperties config "feast.jobs.consolidate-jobs-per-source".
+   *
+   * @param feastProperties feast config properties
+   * @param jobRepository repository required by strategy
+   * @return JobGroupingStrategy
+   */
   @Bean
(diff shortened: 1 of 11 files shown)


core/src/main/java/feast/core/config/JobConfig.java
```diff
@@ -64,6 +68,16 @@
         .build();
   }
 
+  @Bean
+  public JobGroupingStrategy getJobGroupingStrategy(
+      FeastProperties feastProperties, JobRepository jobRepository) {
+    if (feastProperties.getJobs().getConsolidateJobsPerSource()) {
```
woop, line 74: Small nitpick, for readability we are trying to pop out the nested config into separate variables instead of referencing it directly. So

```
String shouldConsolidateJobsPerSource = feastProperties.getJobs().getConsolidateJobsPerSource();

if (shouldConsolidateJobsPerSource){...}
```

core/src/main/java/feast/core/config/JobConfig.java
```diff
@@ -64,6 +68,16 @@
         .build();
   }
 
+  @Bean
+  public JobGroupingStrategy getJobGroupingStrategy(
```
woop, line 72: Small comment would be nice here.

PR #826, opened 2020-06-25
Add native types for Python SDK online retrieval

### LINKED ISSUE #766: Migrate SDKs to use Native types instead of Protobuf types
### Background
One of the stumbling blocks working on PR #658   with the SDKs is that the Python and Golang SDKs directly expose the Serving API to users. This is a problem changes with the Serving protobuf API will propagate to user API changes on the Python and Java client. This makes API compatibility hard to track and maintain. ([see @ches's comment about it here](https://github.com/feast-dev/feast/pull/658#issuecomment-621681310))

### Problem 
Currently the SDKs expose internal Protobuf types directly to users.
For example:
- Python SDK
https://github.com/feast-dev/feast/blob/a33883fbbf0c8d4b27711d85bca313ab11da81b6/sdk/python/feast/client.py#L619
- Golang SDK
https://github.com/feast-dev/feast/blob/a33883fbbf0c8d4b27711d85bca313ab11da81b6/sdk/go/client.go#L84

This poses a problem as:
- This makes the gRPC API exposed by Feast Serving/Core a user facing API, affecting downstream systems if changes are made to it.
- Users prefer to deal with native types instead of Protobuf types. For example, Python users prefer to interact with `datetime.timedelta` instead of Protobuf's `Duration` type.

### Proposed Solution
1. Row API (similar to Java SDK’s Row)  to be introduced to Python and Go SDKs as an higher level API. 
- ie Row would be used to abstract away `GetOnlineFeaturesResponse` etc.
```python
feature_rows = client.get_online_features(....)
```

2. Use native types in User Facing SDK APIs
- ie:
```golang
func (fc *GrpcClient) GetFeastServingInfo(....) map[string]string
```

terryyylim:
Allows user to pass in a more readable form of `EntityRow`, using Python native form, for get_online_features method. Also, users are now able to convert the response in a more readable form by using `to_dict()`.

Fixes #766 

```release-note
Users can now use `to_dict()` after getting the response from python SDK's get_online_features to convert response to a more human readable form.
```


woop:
> Allows user to pass in a more readable form of EntityRow
> Does this PR introduce a user-facing change?: NONE

Seems contradictory x_X

woop:
Please update the Gitbook examples and tutorials as well to this new format.

sdk/python/feast/client.py
```diff
@@ -682,6 +689,8 @@ def get_online_features(
         """
 
         try:
+            if entity_rows and isinstance(entity_rows[0], dict):
```
woop, line 692: We should have a deprecation warning for the old format I think

sdk/python/feast/client.py
```diff
@@ -993,6 +1004,32 @@ def _get_grpc_metadata(self):
...
+
+    Returns:
+        A list of EntityRow protos parsed from args.
+    """
+    entity_row_list = []
+
+    for entity in entities:
+        for key, value in entity.items():
+            # Infer the specific type for this row
+            current_dtype = python_type_to_feast_value_type(name=key, value=value)
```
woop, line 1025: Can we calculate the current_dtype once instead of every time? It should be consistent for a column. If the casting fails then we can just fail.

tests/e2e/redis/basic-ingest-redis-serving.py
```diff
@@ -741,6 +741,77 @@ def test_list_entities_and_features(client):
...
+        features=[
+            Feature(name="rating", dtype=ValueType.INT64, labels={"key1":"val1"}),
+            Feature(name="cost", dtype=ValueType.FLOAT),
+            Feature(name="past_transactions_int", dtype=ValueType.INT64_LIST),
+            Feature(name="past_transactions_double", dtype=ValueType.DOUBLE_LIST),
+            Feature(name="past_transactions_float", dtype=ValueType.FLOAT_LIST),
+            Feature(name="past_transactions_string", dtype=ValueType.STRING_LIST),
+            Feature(name="past_transactions_bool", dtype=ValueType.BOOL_LIST)
+        ],
+        entities=[Entity("customer_id", ValueType.INT64)],
```
woop, line 779: Where are you testing the new entity retrieval?

tests/e2e/redis/basic-ingest-redis-serving.py
```diff
@@ -741,6 +741,77 @@ def test_list_entities_and_features(client):
...
+    )
+
+    client.set_project(PROJECT_NAME)
+    client.apply(customer_fs)
+
+    customer_fs = client.get_feature_set(name="customer2")
+    client.ingest(customer_fs, list_type_dataframe, timeout=300)
+    time.sleep(15)
+
+    online_request_entity = [{"customer_id": 1}]
```
woop, line 790: Can we try multiple entities please and we also need to test failure, what if mixed types are provided or mixed list types.

sdk/python/feast/client.py
```diff
@@ -687,7 +690,10 @@ def get_online_features(
             Each EntityRow provided will yield one record, which contains
             data fields with data value and field status metadata (if included).
         """
-
+        warnings.warn(
```
woop, line 693: Does this warning run for every single call?

terryyylim: yes for every call to get_online_features()

sdk/python/feast/client.py
terryyylim: Using this loop currently to check if mixed types are present.

sdk/python/feast/client.py
```diff
@@ -1018,11 +1030,29 @@ def _infer_entity_rows(
...
 
+    is_mixed_type = False
     for entity in entities:
         for key, value in entity.items():
+            if isinstance(value, list):
+                is_mixed_type = _is_mixed_type(value)
             # Infer the specific type for this row
             current_dtype = python_type_to_feast_value_type(name=key, value=value)
+
+            if is_mixed_type:
```
woop, line 1043: I don't think this approach is optimal. I think we should just try and detect the first type and cast all remaining values. If there is a failure then there is a failure.

sdk/python/feast/client.py
```diff
@@ -1018,11 +1030,29 @@ def _infer_entity_rows(
...
+                raise TypeError(
+                    f"Input entity {key} of List type has mixed types and that is not allowed. "
+                )
+            if key not in temp_dtype_storage:
+                temp_dtype_storage[key] = current_dtype
+            else:
+                if current_dtype == temp_dtype_storage[key]:
```
woop, line 1050: Why not invert this statement so that you dont have the pass.

sdk/python/feast/client.py
terryyylim: hmm you mean letting this [function](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/type_map.py#L233) handle the failure?

woop: Yes

terryyylim: Yup, that can handle it. If that's the case, we would not need `_is_mixed_type` function anymore. Will remove that then.

sdk/python/feast/client.py
woop: That might overload the system. Can we try and run it only once?

terryyylim: I think a way is to keep track of how many times we've warned and if the number of times is 1 then we stop warning. But I'm not sure if that's a good approach since users may sometimes not see the warning on the first time.

sdk/python/feast/client.py
mrzzy: Not sure if we want to deprecate Entity Rows entirely as I don't we have a native Python equivalent of corner types like `int32` or `float`. Entity Rows can act like a cop out when users want to specify type information.

sdk/python/feast/client.py
```diff
@@ -960,6 +976,41 @@ def _get_grpc_metadata(self):
...
+
+    Args:
+        entities: List of Dict[str, Union[bool,bytes,float,int,str,List[bool,bytes,float,int,str]]]]
+            where each key represents the entity name and value is feast.types.Value in Python native form.
+
+    Returns:
+        A list of EntityRow protos parsed from args.
+    """
+    entity_row_list = []
+    temp_dtype_storage = dict()
```
mrzzy, line 993: `entity_type_map` or `entity_types` makes for better naming IMO.

sdk/python/feast/type_map.py
```diff
@@ -38,6 +39,44 @@
...
+def feast_value_type_to_python_type(field_value_proto: ProtoValue) -> Any:
+    """
+    Converts field value Proto to Dict and returns each field's Feast Value Type value
+    in their respective Python value.
+
+    Args:
+        field_value_proto: Field value Proto
+
+    Returns:
+        Python native type based on Feast Value Type
```
mrzzy, line 51: Clarify the docs: 
Python native type representation/version of the given `field_value_proto`

sdk/python/feast/type_map.py
```diff
@@ -38,6 +39,44 @@
...
+        if (k == "floatListVal") or (k == "doubleListVal"):
+            return [float(item) for item in v["val"]]
+        if k == "stringListVal":
+            return [str(item) for item in v["val"]]
+        if k == "bytesListVal":
+            return [bytes(item) for item in v["val"]]
+        if k == "boolListVal":
+            return [bool(item) for item in v["val"]]
+
+        if k in ["int32Val", "floatVal", "doubleVal", "stringVal", "boolVal"]:
```
mrzzy, line 71: Is there a reason why these types can't be converted into python native types?

sdk/python/feast/client.py
```diff
@@ -668,9 +673,13 @@ def get_online_features(
                 "feature_set:feature" where "feature_set" & "feature" refer to
                 the feature and feature set names respectively.
                 Only the feature name is required.
-            entity_rows: List of GetFeaturesRequest.EntityRow where each row
+            entity_rows:
+                List of GetFeaturesRequest.EntityRow where each row
                 contains entities. Timestamp should not be set for online
                 retrieval. All entity types within a feature
+                OR
+                List of Dict[str, Union[bool,bytes,float,int,str,List[bool,bytes,float,int,str]]]]
```
mrzzy, line 681: Can also allow users to pass `Dict[str,feast.types.Value]` when they want to specify type information post v0.7, as we don't have a native Python equivalent of corner types like `int32` or `float`.

sdk/python/feast/type_map.py
terryyylim: They can and are automatically converted as shown [here](https://developers.google.com/protocol-buffers/docs/proto3#json). So don't need to explicitly do so.

sdk/python/feast/client.py
terryyylim: So explicit declaration of something like `Value(int64_val=1001)`? Should that be included in this PR?

sdk/python/feast/type_map.py
```diff
@@ -56,6 +95,9 @@ def python_type_to_feast_value_type(
     """
 
     type_name = type(value).__name__
+    if isinstance(value, list):
+        type_name = "ndarray"
+        value = np.asarray(value)
```
mrzzy, line 100: This piece of code seems out of place. Can move closer/merge with the block starting at line 125?

tests/e2e/redis/basic-ingest-redis-serving.py
```diff
@@ -299,6 +299,157 @@ def try_get_features():
...
+            "customer2_past_transactions_double": [[float(i)+0.5,float(i)+2] for i in range(N_ROWS)],
+            "customer2_past_transactions_float": [[float(i)+0.5,float(i)+2] for i in range(N_ROWS)],
+            "customer2_past_transactions_string": [['first_'+str(i),'second_'+str(i)] for i in range(N_ROWS)],
+            "customer2_past_transactions_bool": [[True,False] for _ in range(N_ROWS)]
+        }
+    )
+    return customer_df
+
+@pytest.fixture(scope='module')
+def entity_list_type_dataframe():
```
mrzzy, line 322: `list_type_dataframe` and `entity_list_type_dataframe` naming tell us nothing about what the dataframes are used for/meant to be ingested to etc.

terryyylim: That's why I added "cases" in `test_basic_retrieve_online_dict`. But yup, I can be more descriptive with the dataframes, will add comments in those fixtures.

sdk/python/feast/response.py
```diff
@@ -0,0 +1,54 @@
...
+        """
+        Getter for GetOnlineResponse's field_values.
+        """
+        return self.proto.field_values
+
+    def to_dict(self) -> Dict[str, Any]:
+        """
+        Converts GetOnlineFeaturesResponse features into a dictionary form.
+        """
+        features = [k for row in self.field_values for k, _ in row.fields.items()]
```
mrzzy, line 46: Nitpick. Can we use `fields` instead of `features` as field values might contain entity values as well as feature values?

tests/e2e/redis/basic-ingest-redis-serving.py
```diff
@@ -299,6 +299,157 @@ def try_get_features():
...
+        "district_rating",
+        "district_cost",
+        "district_past_transactions_int",
+        "district_past_transactions_double",
+        "district_past_transactions_float",
+        "district_past_transactions_string",
+        "district_past_transactions_bool"
+    ]
+
+    online_features_actual2 = client.get_online_features(entity_rows=online_request_entity2, feature_refs=online_request_features2)
```
mrzzy, line 424: Unfortunately, I think you need to poll `get_online_features` for feature values as they might not be immediately available. See [here](https://github.com/feast-dev/feast/blob/b461f4dea517de5e5ad7c9114e069081d9437810/tests/e2e/redis/basic-ingest-redis-serving.py#L261) for how to do it.

tests/e2e/redis/basic-ingest-redis-serving.py
```diff
@@ -299,6 +299,157 @@ def try_get_features():
...
+        "customer2_rating",
+        "customer2_cost",
+        "customer2_past_transactions_int",
+        "customer2_past_transactions_double",
+        "customer2_past_transactions_float",
+        "customer2_past_transactions_string",
+        "customer2_past_transactions_bool"
+    ]
+
+    online_features_actual = client.get_online_features(entity_rows=online_request_entity, feature_refs=online_request_features)
```
mrzzy, line 378: Unfortunately, I think you need to poll `get_online_features` for feature values as they might not be immediately available (even after your `time.sleep`). See [here](https://github.com/feast-dev/feast/blob/b461f4dea517de5e5ad7c9114e069081d9437810/tests/e2e/redis/basic-ingest-redis-serving.py#L261) for how to do it. Same for `online_features_actual2`.

sdk/python/feast/client.py
mrzzy: Yes so something like:
```python
client.get_online_features(
    ...,
    entity_rows={
        "customer_id": Value(int32_val=1001),
    }
)
```
so we support `Value` as well as native types, as there is no native equivalent for some types ie `float` and `int32`.

woop: I think this is fair.

sdk/python/feast/client.py
mrzzy: We can have a flag on client ie `self._has_warned_entity_row_deprecation` so that we warn the first time they call `get_online_features` per script run, but not every time they call `get_online_features()`. Users will see the warning every time they run their script.

sdk/python/feast/type_map.py
```diff
@@ -80,6 +119,10 @@ def python_type_to_feast_value_type(
     if type_name in type_map:
         return type_map[type_name]
 
+    if isinstance(value, list):
```
mrzzy, line 122: `pd.core.series.Series` accepts a Python as input. The `np.array` cast is redundant here.
we can combine `if isinstance(value, list):` and `if type_name == "ndarray":` into a single if:

```python
if isinstance(value, list) or type_name == "ndarray":
    if recurse:
            # Convert to list type
            list_items = pd.core.series.Series(value)
            ...
```

sdk/python/feast/client.py
```diff
@@ -994,17 +995,21 @@ def _infer_entity_rows(
...
-            proto_value = _python_value_to_proto_value(current_dtype, value)
+                # Infer the specific type for this row
+                current_dtype = python_type_to_feast_value_type(name=key, value=value)
+
+                if key not in entity_type_map:
+                    entity_type_map[key] = current_dtype
+                else:
+                    if current_dtype != entity_type_map[key]:
+                        raise TypeError(
+                            f"Input entity {key} has mixed types and that is not allowed. "
```
woop, line 1010: Please print the types in the exception.

sdk/python/feast/client.py
woop: @terryyylim can you please add a test for thes Value() types?

woop:
@terryyylim You should definitely update our documentation to ensure that all references to the old API are changed (for entity rows) and also please make sure that the example notebooks are updated.

examples/basic/basic.ipynb
```diff
@@ -469,15 +470,41 @@
...
     "    ],\n",
     "    entity_rows=[\n",
-    "        GetOnlineFeaturesRequest.EntityRow(\n",
-    "            fields={\n",
-    "                \"customer_id\": Value(\n",
-    "                    int64_val=1001)\n",
-    "            }\n",
-    "        )\n",
+    "        {\n",
+    "            \"customer_id\": Value(int64_val=1001)\n",
```
woop, line 474: I dont want to recommend the use of this Proto API if we have a Python native API

examples/feast-xgboost-churn-prediction-tutorial/Telecom Customer Churn Prediction (with Feast and XGBoost).ipynb
```diff
@@ -7219,7 +7240,7 @@
     "        entity_rows = []\n",
     "        for customer_id in customer_ids:\n",
     "            entity_rows.append(\n",
-    "                GetOnlineFeaturesRequest.EntityRow(fields={'customer_id': Value(string_val=customer_id)})\n",
+    "                {'customer_id': Value(string_val=customer_id)}\n",
```
woop, line 7243: Same problem here, can we make it 

```
{'customer_id': customer_id}
```

sdk/python/feast/client.py
```diff
@@ -668,9 +674,13 @@ def get_online_features(
                 "feature_set:feature" where "feature_set" & "feature" refer to
                 the feature and feature set names respectively.
                 Only the feature name is required.
-            entity_rows: List of GetFeaturesRequest.EntityRow where each row
+            entity_rows:
+                List of GetFeaturesRequest.EntityRow where each row
                 contains entities. Timestamp should not be set for online
                 retrieval. All entity types within a feature
+                OR
+                List of Dict[str, Union[bool,bytes,float,int,str,List[bool,bytes,float,int,str]]]]
```
woop, line 682: This explanation is super confusing. Can we just explain this in a human readable way? 

A list of dictionaries where the key is an entity and the value is either a Python native type or feast.types.Value? You can remove all mention of EntityRow since we dont want people to use it.

sdk/python/feast/client.py
```diff
@@ -655,10 +661,10 @@ def get_batch_features(
     def get_online_features(
         self,
         feature_refs: List[str],
-        entity_rows: List[GetOnlineFeaturesRequest.EntityRow],
+        entity_rows: List[Union[GetOnlineFeaturesRequest.EntityRow, Dict[str, Any]]],
         project: Optional[str] = None,
         omit_entities: bool = False,
-    ) -> GetOnlineFeaturesResponse:
+    ) -> OnlineResponse:
```
woop, line 667: Can you please add an example of retrieval here. Have a look here.https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

sdk/python/feast/client.py
```diff
@@ -679,19 +689,26 @@ def get_online_features(
...
+        )
         try:
+            if entity_rows and isinstance(entity_rows[0], dict):
+                entity_rows = _infer_entity_rows(entity_rows)  # type: ignore
             response = self._serving_service.GetOnlineFeatures(
                 GetOnlineFeaturesRequest(
                     omit_entities_in_response=omit_entities,
                     features=_build_feature_references(feature_ref_strs=feature_refs),
-                    entity_rows=entity_rows,
+                    entity_rows=entity_rows,  # type: ignore
```
woop, line 703: Why are we ignoring the type, what is the conflict?

sdk/python/feast/response.py
```diff
@@ -0,0 +1,54 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Any, Dict, List
+
+from feast.serving.ServingService_pb2 import GetOnlineFeaturesResponse
+from feast.type_map import feast_value_type_to_python_type
+
+
+class OnlineResponse:
```
woop, line 21: Can we change the file name to `online_response.py`?

PR #829, opened 2020-06-25
JobCoordinator may be turned off by configuration

pyalex:

We need option to deploy core without job management

Fixes #

```release-note
Added the option of turning ingestion job management on or off.
```

### DIFF
--- a/core/src/main/java/feast/core/config/FeastProperties.java
+++ b/core/src/main/java/feast/core/config/FeastProperties.java
@@ -76,6 +76,8 @@ public class FeastProperties {
   @Getter
   @Setter
   public static class JobProperties {
+    /* Toggle for enabling/disabling job management */
+    private Boolean enabled = true;
 
     @NotBlank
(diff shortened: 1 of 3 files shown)


ches:
We pretty much coincidentally implemented the same thing at the same time.

Illuminating a use case for anyone following along: you probably want to deploy redundant Core instances for availability. They will not play well together if they all try to manage jobs, duplicating things and stomping on each other.

A boolean flag serves as poor man's leader election, enable it on only one instance in a deployment. If the active coordinator fails, remediate by bringing it back to life or deploying a new instance with this flag enabled.

Although not specific to the job coordination, a related concern is that Serving instances talk to Core to load and cache registered feature set specifications. The caching provides a layer of resilience if Core instances fail. Enabling a means of load balancing / round-robin for the gRPC connections from Serving to Core instances may be desirable however.

Some of this may have a home in Feast operations guide documentation.

PR #833, opened 2020-06-29
Add Support For Bigtable as Online Store

Wirick:

This implements support for using bigtable to ingest and serve low latency features, while having a fully managed and scalable database backend.

This design will be a jumping off point, as I see it as easily adaptable to different use cases of a service like bigtable. I would be interested in exploring if we could maybe have different couplings of serving and ingestion code so that we could do specific actions with keys. For our use case it may require a different bigtable instance, since it can't use a multi cluster routing app profile but that seems like a worthwhile investment. https://cloud.google.com/bigtable/docs/writes#increments

This way we could potentially have features that are incremented or appended to (like lists of values for computational/aggregation purposes) Certainly having to serialize lots of values may not be performant, but if we are careful about it I could see that we could compute aggregations on the fly. This may even be possible using the same sort of configurations as this implementation uses.

Fixes #

```release-note
Add Bigtable Store
```


Wirick:
@ches and @zhilingc wondering if you have any advice for me given this error that I'm getting on startup for core and all my unit tests seem to be passing
```
╰─ kcd logs feast-core-77d587895-zdlg6
Exception in thread "main" java.lang.IllegalStateException: Failed to get nested archive for entry BOOT-INF/lib/feast-ingestion-dev.jar
	at org.springframework.boot.loader.archive.JarFileArchive.getNestedArchive(JarFileArchive.java:108)
	at org.springframework.boot.loader.archive.JarFileArchive.getNestedArchives(JarFileArchive.java:86)
	at org.springframework.boot.loader.ExecutableArchiveLauncher.getClassPathArchives(ExecutableArchiveLauncher.java:70)
	at org.springframework.boot.loader.Launcher.launch(Launcher.java:49)
	at org.springframework.boot.loader.JarLauncher.main(JarLauncher.java:51)
Caused by: java.io.IOException: Unable to open nested jar file 'BOOT-INF/lib/feast-ingestion-dev.jar'
	at org.springframework.boot.loader.jar.JarFile.getNestedJarFile(JarFile.java:254)
	at org.springframework.boot.loader.jar.JarFile.getNestedJarFile(JarFile.java:239)
	at org.springframework.boot.loader.archive.JarFileArchive.getNestedArchive(JarFileArchive.java:103)
	... 4 more
Caused by: java.lang.IllegalStateException: Zip64 archives are not supported
	at org.springframework.boot.loader.jar.CentralDirectoryEndRecord.getNumberOfRecords(CentralDirectoryEndRecord.java:124)
	at org.springframework.boot.loader.jar.JarFileEntries.visitStart(JarFileEntries.java:91)
	at org.springframework.boot.loader.jar.CentralDirectoryParser.visitStart(CentralDirectoryParser.java:89)
	at org.springframework.boot.loader.jar.CentralDirectoryParser.parse(CentralDirectoryParser.java:57)
	at org.springframework.boot.loader.jar.JarFile.<init>(JarFile.java:121)
	at org.springframework.boot.loader.jar.JarFile.<init>(JarFile.java:109)
	at org.springframework.boot.loader.jar.JarFile.createJarFileFromFileEntry(JarFile.java:287)
	at org.springframework.boot.loader.jar.JarFile.createJarFileFromEntry(JarFile.java:262)
	at org.springframework.boot.loader.jar.JarFile.getNestedJarFile(JarFile.java:250)
	... 6 more
```

Seems like there may be too many files in the shaded jar? Should the approach be to move away from shade or upgrade/align dependencies so there are less files. See https://github.com/spring-projects/spring-boot/issues/2895 and https://github.com/spring-projects/spring-boot/pull/16091 for context)

PR #841, opened 2020-06-30
Add disk size as Dataflow Job Configuration

khorshuheng:
After #718, it's not possible to have arbitrary options set for runner configuration. Instead, it has to be one of the options supported by the protos defined in `DataflowRunnerConfigOptions` .

`diskSizeGb` is an important option which wasn't available in the proto. This PR added the configuration.

Fixes #

```release-note
Users will now be able to specify disk size for dataflow workers.
```

### DIFF
--- a/core/src/test/java/feast/core/job/dataflow/DataflowRunnerConfigTest.java
+++ b/core/src/test/java/feast/core/job/dataflow/DataflowRunnerConfigTest.java
@@ -42,6 +42,7 @@ public class DataflowRunnerConfigTest {
             .setUsePublicIps(false)
             .setWorkerMachineType("n1-standard-1")
             .setDeadLetterTableSpec("project_id:dataset_id.table_id")
+            .setDiskSizeGb(100)
             .putLabels("key", "value")
             .build();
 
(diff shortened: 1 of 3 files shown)


core/src/test/java/feast/core/job/dataflow/DataflowRunnerConfigTest.java
```diff
@@ -38,10 +38,12 @@ public void shouldConvertToPipelineArgs() throws IllegalAccessException {
             .setNetwork("default")
             .setSubnetwork("regions/asia-east1/subnetworks/mysubnetwork")
             .setMaxNumWorkers(1)
+            .setServiceAccount("serviceaccount")
```
woop, line 41: Why is service account being modified here? Is this a mistake?

protos/feast/core/Runner.proto
```diff
@@ -75,6 +75,9 @@ message DataflowRunnerConfigOptions {
     /* BigQuery table specification, e.g. PROJECT_ID:DATASET_ID.PROJECT_ID */
     string deadLetterTableSpec = 12;
 
+    /* Disk size to use on each remote Compute Engine worker instance */
+    int32 diskSizeGb = 13;
+
     /* Labels to apply to the dataflow job */
-    map<string, string> labels = 13;
+    map<string, string> labels = 14;
```
woop, line 82: Why are the proto numbers being changed?

core/src/test/java/feast/core/job/dataflow/DataflowRunnerConfigTest.java
khorshuheng: Service account is part of the available configurations, but it was never tested in the test. We can have this in another PR is it is preferable.

woop: Its still not clear if its necessary or not.

khorshuheng: I have removed the service account argument from the test.

PR #842, opened 2020-06-30
Replace batch with historical for Python SDK retrieval

woop:

We are slowly starting to rename API methods from `batch` to `historical` for all historical retrieval. This is one step towards deprecating the method `get_batch_features()` for the identical `get_historical_features()`.

```release-note
Users can now use get_historical_features() instead of get_batch_features() on the Python SDK
```

### DIFF
--- a/examples/basic/basic.ipynb
+++ b/examples/basic/basic.ipynb
@@ -572,7 +572,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "By calling the `get_batch_features` method we are able to retrieve a `job` object. This object can be used to retrieve the resulting training dataset that is exported by Feast. \n",
+    "By calling the `get_historical_features` method we are able to retrieve a `job` object. This object can be used to retrieve the resulting training dataset that is exported by Feast. \n",
     "\n",
     "The dataset that is returned will contain feature values for each entity and timestamp combination in `entity_rows`."
    ]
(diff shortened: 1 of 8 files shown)


sdk/python/feast/client.py
```diff
@@ -547,6 +548,25 @@ def get_batch_features(
...
+        return self.get_historical_features(
+            feature_refs, entity_rows, compute_statistics, project
+        )
+
+    def get_historical_features(
+        self,
+        feature_refs: List[str],
+        entity_rows: Union[pd.DataFrame, str],
+        compute_statistics: bool = False,
+        project: str = None,
```
terryyylim, line 569: Should also update the example to use `get_historical_features()` instead of `get_batch_features()`.

terryyylim:
The example notebooks should be updated as well to encourage users to use `get_historical_features()` instead.

sdk/python/feast/client.py
woop: Done.

woop:
> The example notebooks should be updated as well to encourage users to use `get_historical_features()` instead.

Done.

examples/basic/basic.ipynb
```diff
@@ -585,7 +585,7 @@
    },
    "outputs": [],
    "source": [
-    "job = batch_client.get_batch_features(\n",
+    "job = batch_client.get_historical_features(\n",
```
terryyylim, line 588: `By calling the get_batch_features method we are able to retrieve a job object. This object can be used to retrieve the resulting training dataset that is exported by Feast.`

This portion can be updated as well.

terryyylim:
lgtm!

PR #845, opened 2020-06-30
Don't send unrecognized featureSets to deadletter in IngestionJob

pyalex:

Currently we overpopulating deadletter sink (in bq) by sending all FeatureRows that coming in with unknown FeatureSetReference. That could create billions of rows within hours and create long queue in IngestionJob, which disrupt normal processing (OOM, eg). We should just write smth to log instead.

Fixes #

```release-note

```

### DIFF
--- a/ingestion/src/main/java/feast/ingestion/transform/fn/ValidateFeatureRowDoFn.java
+++ b/ingestion/src/main/java/feast/ingestion/transform/fn/ValidateFeatureRowDoFn.java
@@ -31,9 +31,12 @@ import java.util.Map;
 import org.apache.beam.sdk.transforms.DoFn;
 import org.apache.beam.sdk.values.PCollectionView;
 import org.apache.beam.sdk.values.TupleTag;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @AutoValue
 public abstract class ValidateFeatureRowDoFn extends DoFn<FeatureRow, FeatureRow> {
+  private static final Logger log = LoggerFactory.getLogger(ValidateFeatureRowDoFn.class);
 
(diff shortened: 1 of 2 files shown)


ingestion/src/main/java/feast/ingestion/transform/fn/ValidateFeatureRowDoFn.java
```diff
@@ -65,40 +68,41 @@ public void processElement(ProcessContext context) {
     FeatureRow featureRow = context.element();
     Iterable<FeatureSetProto.FeatureSetSpec> featureSetSpecs =
         context.sideInput(getFeatureSets()).get(featureRow.getFeatureSet());
+    if (featureSetSpecs == null) {
+      log.warn(
+          String.format(
+              "FeatureRow contains invalid feature set id %s. Please check that the feature rows are being published to the correct topic on the feature stream.",
```
woop, line 74: We are now calling it a `feature set reference`.

ches:
This feels like a somewhat arbitrary technical fix to an operational problem. Semantically I would consider unrecognized feature set a "dead letter" just as much as the remaining issues that route there.

For us, this could be a (SQL) replication delay from the registry when a producer already started publishing data, a case where we may want to be able to reprocess. On the other hand Feast does need a means of protecting itself from misuse, e.g. data being erroneously sent without any intended registration (deployment in wrong environment, etc.). Perhaps separate error queues.

I don't strongly object to this going in, especially as it's currently with dead letters coupled to BigQuery, but it does bring to mind that this is an area of debt for the backlog IMO: the BigQuery coupling, and new design considerations that could accompany tackling that.

woop:
> this is an area of debt for the backlog IMO: the BigQuery coupling, and new design considerations that could accompany tackling that.

Agreed on that. The whole deadletter and logging area needs TLC. So yes, this is a bit of a workaround until its properly solved.

PR #846, opened 2020-06-30
Replace IngestionJob when store was updated

pyalex:

TestCase that fails:
1. There's existing consolidated job with state RUNNING
2. Job has attached Store with subscription project_1:*
3. We update Store config (on serving layer) by adding new subscription project_2:*
4. After serving layer sent Store to Core - Job should be replaced with new one that has both subscriptions

Fixes #

```release-note

```

### DIFF
--- a/core/src/main/java/feast/core/service/JobCoordinatorService.java
+++ b/core/src/main/java/feast/core/service/JobCoordinatorService.java
@@ -223,6 +223,10 @@ public class JobCoordinatorService {
       return true;
     }
 
+    if (stores.stream().anyMatch(s -> s.getLastUpdated().after(job.getCreated()))) {
+      return true;
+    }
+
     return false;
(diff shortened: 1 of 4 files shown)


core/src/main/java/feast/core/service/JobCoordinatorService.java
```diff
@@ -223,6 +223,10 @@ private boolean jobRequiresUpgrade(Job job, Set<Store> stores) {
       return true;
     }
 
+    if (stores.stream().anyMatch(s -> s.getLastUpdated().after(job.getCreated()))) {
```
pyalex, line 226: This is a hacky solution of the case when store's subscription was changed but it didn't affect configuration (added or removed Store from the Set, see previous line).
Requires better solution in future.

PR #848, opened 2020-06-30
Fix FeatureSetJobStatus removal

pyalex:

Currently due to cascade relations from both Job and FeatureSet to FeatureSetJobStatus it's impossible to delete FeatureSetJobStatus. As a solution I propose to delete cascade on FeatureSet side because:
1. We manage FeatureSetJobStatus from Job side
2. Ideally when we completely decouple JobCoordination from Core there will be no foreign key (relationship in general) to FeatureSet in FeatureSetJobStatus (see my other PR #852 )
3. Currently we not planning to physically delete feature sets. Even if we will delete them - only by marking as deleted.

Fixes #

```release-note

```

### DIFF
--- a/core/src/main/java/feast/core/model/FeatureSet.java
+++ b/core/src/main/java/feast/core/model/FeatureSet.java
@@ -93,7 +93,7 @@ public class FeatureSet extends AbstractTimestampEntity {
   @Column(name = "version", columnDefinition = "integer default 0")
   private int version;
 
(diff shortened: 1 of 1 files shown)


mrzzy:
Its not clear what this problem this solves. @pyalex mind updating PR description?

pyalex:
@mrzzy updated PR description

woop:
Ok makes sense. Deletion is hard. It will cause inconsistency here. Lets assume feature sets dont get deleted completely but only marked as deleted.

PR #850, opened 2020-06-30
OSX doesn't always have sched_getaffinity

Wirick:
1. The python client breaks when calling
 CPU_COUNT: int = len(os.sched_getaffinity(0))

When importing the python client I encounter an error
```
  File "<stdin>", line 1, in <module>
AttributeError: module 'os' has no attribute 'sched_getaffinity'
```
We should fallback or use another trick to compute this value, it seems to be a known issue
https://stackoverflow.com/questions/42538153/python-3-6-0-os-module-does-not-have-sched-getaffinity-method

### DIFF
--- a/sdk/python/feast/client.py
+++ b/sdk/python/feast/client.py
@@ -93,7 +93,10 @@ from feast.types.Value_pb2 import Value as Value
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


mrzzy:
@Wirick Thanks for this. However, this seems to be a duplicate of #839?

Wirick:
didn't even see it just fixed in the dev process

PR #851, opened 2020-07-01
Bump spring to 2.3.0

Wirick:
1. This induces upgrades for a number of different packages in order to be
compatible, including kafka and the jakarta validation-api
2. Resolved logging errors in core/serving from seeing different logger
   classes in the classpath by excluding them

I am able to build and run feast core and feast serving, along with bigquery batch ingestion with these updated dependencies. However there was an upgrade needed in how we mock the url for the test database. Context: https://stackoverflow.com/questions/49088847/after-spring-boot-2-0-migration-jdbcurl-is-required-with-driverclassname

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,20 +49,24 @@
         <grpcVersion>1.17.1</grpcVersion>
         <protocVersion>3.10.0</protocVersion>
         <protobufVersion>3.10.0</protobufVersion>
-        <springBootVersion>2.0.9.RELEASE</springBootVersion>
+        <springBootVersion>2.3.0.RELEASE</springBootVersion>
+        <spring.security.version>5.3.0.RELEASE</spring.security.version>
+        <spring.security.oauth.version>2.4.0.RELEASE</spring.security.oauth.version>
         <org.apache.beam.version>2.18.0</org.apache.beam.version>
         <com.google.cloud.version>1.91.0</com.google.cloud.version>
         <io.prometheus.version>0.8.0</io.prometheus.version>
         <byte-buddy.version>1.9.10</byte-buddy.version>
         <hamcrest.version>1.3</hamcrest.version>
         <hibernate.version>5.3.6.Final</hibernate.version>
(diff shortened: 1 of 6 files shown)


ches:
The re-indentation is a bit distracting, I know some POM files in the project are indented with two spaces and others four, and GitHub has the option to ignore whitespace in diffs, but if we're going to do mass reformats it'd be nice to do them in separate commits and ideally with a means to check/automate the convention going forward.

Something that could have been brought up in #793 I think: we use an imported Spring bill of materials from the root `pom.xml` which sets managed versions of dependencies for Spring ecosystem—it'd be preferable IMO to let that drive versions of things like spring-security deps (i.e. just drop the `<version>` tags). Unless we have a specific reason for deviating from the BOM, which should be documented with a comment.

Spring and Beam or proto/gRPC eco dep versions may be at odds sometimes so may necessitate overrides, but at least as long as Core depends on Ingestion I believe it's worth maintaining the BOM usage. If/when that situation changes it might make sense to move service API modules under their own parent POM holding the Spring BOM.

pom.xml
```diff
@@ -366,6 +374,9 @@
                         <goals>
                             <goal>jar</goal>
                         </goals>
+                        <configuration>
+                            <javadocExecutable>${java.home}/bin/javadoc</javadocExecutable>
```
ches, line 378: Any reason this needed to set specifically?

Wirick:
Closing in favor of https://github.com/feast-dev/feast/pull/854

pom.xml
Wirick: Yea my intellij environment ran into the issue described and solved here https://intellij-support.jetbrains.com/hc/en-us/community/posts/206880855-MavenReportException-Error-while-creating-archive-Unable-to-find-javadoc-command-The-environment-variable-JAVA-HOME-is-not-correctly-set- so thats what allowed me to execute this in intellij. We can continue the convo here https://github.com/feast-dev/feast/pull/854

PR #852, opened 2020-07-01
Keep StoreProto inside JobStore to decouple JobCoordination from SpecService internals

pyalex:

This PR reverts #846 to keep Job and Store more decoupled (and not use StoreRepository from JobCoordinator).
Now Job keeps full copy of related Stores (in proto) in `jobs_stores` table. It allows us to compare deployed stores with new one coming from `SpecService` and upgrade IngestionJob if needed.

Fixes #

```release-note

```

### DIFF
--- a/core/src/main/java/feast/core/dao/JobRepository.java
+++ b/core/src/main/java/feast/core/dao/JobRepository.java
@@ -41,5 +41,5 @@ public interface JobRepository extends JpaRepository<Job, String> {
   List<Job> findByFeatureSetJobStatusesIn(List<FeatureSetJobStatus> featureSetsJobStatuses);
 
   // find jobs by feast store name
-  List<Job> findByStoresName(String storeName);
+  List<Job> findByJobStoresIdStoreName(String storeName);
 }
--- a/core/src/main/java/feast/core/dao/JobRepository.java
+++ b/core/src/main/java/feast/core/dao/JobRepository.java
@@ -40,6 +40,6 @@ public interface JobRepository extends JpaRepository<Job, String> {
 
   List<Job> findByFeatureSetJobStatusesIn(List<FeatureSetJobStatus> featureSetsJobStatuses);
 
-  // find jobs by feast store name
+  // find jobs that have at least one store with given name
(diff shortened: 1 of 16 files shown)


core/src/main/java/feast/core/dao/JobRepository.java
```diff
@@ -41,5 +41,5 @@
   List<Job> findByFeatureSetJobStatusesIn(List<FeatureSetJobStatus> featureSetsJobStatuses);
 
   // find jobs by feast store name
-  List<Job> findByStoresName(String storeName);
+  List<Job> findByJobStoresIdStoreName(String storeName);
```
woop, line 44: I think the comment is out of date.

core/src/main/java/feast/core/model/Job.java
```diff
@@ -144,6 +136,20 @@ public void addAllFeatureSets(Set<FeatureSet> featureSets) {
     }
   }
 
+  public Set<Store> getStores() {
```
woop, line 139: Need comments here please :)

core/src/main/java/feast/core/dao/JobRepository.java
pyalex: why? it still looking by store name, but it could be one of stores

core/src/main/java/feast/core/model/JobStore.java
```diff
@@ -0,0 +1,110 @@
...
+
+    this.storeProto = output.toByteArray();
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
+    JobStore jobStore = (JobStore) o;
+    return Objects.equal(id, jobStore.id) && Objects.equal(storeProto, jobStore.storeProto);
```
woop, line 103: In general we are trying to always use `this.` and a getter for retrieving and setting fields, especially in these kinds of methods where we are comparing external and local objects. Would you mind scanning through the code and adding that where it makes sense? So `this.getStoreProto()` as opposed to `storeProto`. A bit more verbose, but easier to understand.

Issue #823, opened 2020-06-23
Authentication and Authorization for feast serving


jmelinav:
**Is your feature request related to a problem? Please describe.**
Currently user authentication and authorization limits access to Feast core(#793 ), but doesn't secure feast serving. Since feast client connects to core and serving directly, authentication and authorization on feast serving is a necessary feature. 
**Describe the solution you'd like**
The PR #793, implements jwt authenticator and Keto authorizer for feast core. Same implementation will be done on the feast serving. Instead of replicating the code, Authorization and authentication will be moved separate module feast-auth, which will be included as a dependency in feast core and feast serving. Rest of the configuration will remain same as described in the documentation for #793. Python sdk will be updated to use secure channel to connect to feast serving. 

**Assumptions:**
- User authentication and authorization only applies to Feast Core and serving not ingestion.

**Out of scope:**
- Go and java clients.

**Describe alternatives you've considered**
Refer to issue #793 for the alternatives considered.

jmelinav:
As part of this, I will be moving the below files into feast-auth package, 
/src/main/java/feast/core/config/SecurityConfig.java
/src/main/java/feast/core/auth/authentication/DefaultJwtAuthenticationProvider.java
/src/main/java/feast/core/auth/authorization/AuthorizationProvider.java
/src/main/java/feast/core/auth/authorization/AuthorizationResult.java
/src/main/java/feast/core/auth/authorization/Keto/KetoAuthorizationProvider.java
`SecurityProperties` class which is in /src/main/java/feast/core/config/FeastProperties.java

Also I feel the generic validators can be moved to the feast-common package since it is used in core, will be used in auth and may be useful in other modules too. So for now I will move src/main/java/feast/core/validators/OneOfStrings.java validator into feast-common. Let me know your thoughts. I will update the list if I come across other files that needs to be moved.

woop:
Sounds good.

mrzzy:
Working on getting Authentication (only) working on Feast Serving.

jmelinav:
> Working on getting Authentication (only) working on Feast Serving.

@mrzzy I'm almost ready with for the PR for authentication. I'm working on authorization in the same PR. I will create a wip PR.

woop:
@jmelinav also not the changes here #864

PR #856, opened 2020-07-01
Authentication and Authorization into feast-auth module.

jmelinav:
Authentication and authorization code is currently part of core, To enable authentication and authorization on Serving, and in other modules on need basis, it is ideal to have it in a separate module. 

Fixes #
Related to #823 

```release-note
None
```

### DIFF
--- a/core/src/main/java/feast/core/config/FeastProperties.java
+++ b/core/src/main/java/feast/core/config/FeastProperties.java
@@ -76,7 +76,7 @@ public class FeastProperties {
 
   @Bean
   SecurityProperties securityProperties() {
-    return this.security;
+    return this.getSecurity();
   }
 
   /** Feast job properties. These properties are used for ingestion jobs. */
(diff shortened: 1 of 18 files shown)


core/src/main/java/feast/core/config/FeastProperties.java
```diff
@@ -39,11 +40,12 @@
...
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.info.BuildProperties;
-import org.springframework.stereotype.Component;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
 
 @Getter
 @Setter
-@Component
+@Configuration
```
woop, line 48: Can you shine some light on this change please?

jmelinav: I added a security bean definition in the class. 
Below is what java doc says for `@Configuration`.
`Indicates that a class declares one or more @Bean methods and may be processed by the Spring container to generate bean definitions and service requests for those beans at runtime.`
I'm not sure if `@Component` does the same. Java doc didn't say it explicitly. If both are same it can be reverted.

PR #857, opened 2020-07-02
Bugfix: Re-applying of featuresets does not update label changes

terryyylim:
Currently, when we reapply featuresets to add/remove labels from featureset or features, the changes made to labels are not propagated.

Fixes #

```release-note
NONE
```

### DIFF
--- a/core/src/main/java/feast/core/model/Feature.java
+++ b/core/src/main/java/feast/core/model/Feature.java
@@ -240,7 +240,6 @@ public class Feature {
    * @param featureSpec {@link FeatureSpec} containing schema changes.
    */
   public void updateFromProto(FeatureSpec featureSpec) {
-    this.setLabels(TypeConversion.convertMapToJsonString(featureSpec.getLabelsMap()));
     if (isArchived()) {
       throw new IllegalArgumentException(
           String.format(
@@ -253,6 +252,7 @@ public class Feature {
               "You are attempting to change the type of feature %s from %s to %s. This isn't allowed. Please create a new feature.",
               featureSpec.getName(), type, featureSpec.getValueType()));
     }
+    this.setLabels(TypeConversion.convertMapToJsonString(featureSpec.getLabelsMap()));
     updateSchema(featureSpec);
   }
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/feast/core/model/Feature.java
```diff
@@ -235,11 +235,12 @@ public void archive() {
   }
 
   /**
-   * Update the feature object with a valid feature spec. Only schema changes are allowed.
+   * Update the feature object with a valid feature spec.
    *
    * @param featureSpec {@link FeatureSpec} containing schema changes.
    */
   public void updateFromProto(FeatureSpec featureSpec) {
+    this.setLabels(TypeConversion.convertMapToJsonString(featureSpec.getLabelsMap()));
```
pyalex, line 243: shouldn't we apply update after isArchived check?

core/src/main/java/feast/core/model/FeatureSet.java
```diff
@@ -314,9 +314,10 @@ public void updateFromProto(FeatureSetProto.FeatureSet featureSetProto)
               spec.getEntitiesList(), existingEntities));
     }
 
-    // 4. Update max age and source.
+    // 4. Update max age, source and labels.
     maxAgeSeconds = spec.getMaxAge().getSeconds();
```
pyalex, line 318: could please add `this.` here as well

core/src/main/java/feast/core/model/FeatureSet.java
```diff
@@ -314,9 +314,10 @@ public void updateFromProto(FeatureSetProto.FeatureSet featureSetProto)
               spec.getEntitiesList(), existingEntities));
     }
 
-    // 4. Update max age and source.
+    // 4. Update max age, source and labels.
```
pyalex, line 317: looks like numbering is broken. should be `2.`

core/src/main/java/feast/core/model/FeatureSet.java
terryyylim: adding for source as well.

PR #863, opened 2020-07-08
Bugfix: Blacklisted subscriptions are currently not updated in store

terryyylim:
This PR fixes the issue where updates to subscriptions exclude flag does not propagate to the stores. The caveat in this PR is that all blacklisted subscriptions will apply to both online and historical stores.

Fixes #

```release-note
NONE
```

### DIFF
--- a/core/src/main/java/feast/core/service/SpecService.java
+++ b/core/src/main/java/feast/core/service/SpecService.java
@@ -346,20 +346,24 @@ public class SpecService {
               .build();
     }
 
+    String projectName = newFeatureSet.getSpec().getProject();
+    String featureSetName = newFeatureSet.getSpec().getName();
+    List<Boolean> isSubscribedToStores = new ArrayList<>() {};
     for (Store store : storeRepository.findAll()) {
       List<Subscription> subscriptionList = store.getSubscriptions();
-      String projectName = newFeatureSet.getSpec().getProject();
-      String featureSetName = newFeatureSet.getSpec().getName();
       boolean isSubscribed =
           isSubscribedToFeatureSet(subscriptionList, projectName, featureSetName);
-
-      if (!isSubscribed) {
-        throw new RegistrationException(
-            String.format(
-                "The supplied Project and FeatureSet, %s/%s has been blacklisted and is not available for registration.",
-                projectName, featureSetName));
-      }
+      isSubscribedToStores.add(isSubscribed);
     }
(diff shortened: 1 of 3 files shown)


core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -321,6 +323,20 @@ public ApplyFeatureSetResponse applyFeatureSet(FeatureSetProto.FeatureSet newFea
...
 
+    for (Store store : storeRepository.findAll()) {
+      List<Subscription> subscriptionList = store.getSubscriptions();
+      String projectName = newFeatureSet.getSpec().getProject();
+      String featureSetName = newFeatureSet.getSpec().getName();
+      boolean isSubscribed =
+          isSubscribedToFeatureSet(subscriptionList, projectName, featureSetName);
+
+      if (!isSubscribed) {
+        throw new RegistrationException(
```
pyalex, line 334: I don't understand, why featureSet blacklisted by one storage cannot be used by another?
Here we allow only feature sets subscripted by all stores, which is not true already: we have sauron features sets, and shared redis is not subscribed to them.

core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -335,6 +337,33 @@ public ListStoresResponse listStores(ListStoresRequest.Filter filter) {
...
+      boolean isSubscribed =
+          isSubscribedToFeatureSet(subscriptionList, projectName, featureSetName);
+      isSubscribedToStores.add(isSubscribed);
+    }
+    // Only throw error if FeatureSet is not subscribed by ALL stores
+    if (!isSubscribedToStores.isEmpty()
+        && isSubscribedToStores.stream().allMatch(x -> x == false)) {
+      throw new RegistrationException(
+          String.format(
+              "The supplied Project and FeatureSet, %s/%s has been blacklisted and is not available for registration.",
```
pyalex, line 363: It doesn't necessarily mean that it was blacklisted. There simply can be no subscribers.

pyalex: I think we need to give some hint here, like `ask your administrator to add subscription in store configuration on serving layer`

terryyylim: Would this be clearer and more accurate then `has been blacklisted and is not...` -> `is either blacklist or not subscribed to and is not...`?

pyalex (approved):

PR #864, opened 2020-07-08
Replace Keto Authorization with External HTTP Authorization

woop:

We need to extend Feast Auth towards Feast Serving, but pulling in more and more dependencies in Serving (and throughout Feast) is becoming a big pain. This PR replaces the existing Keto Authorization Provider with an HTTP Authorization Provider. It also comes with an Open API specification that users can implement as a web service in order to roll their own authorization layer. 

```release-note
Users can now provide their own external authorization service for use with Feast
```


auth/src/main/java/feast/auth/authorization/AuthUtil.java
```diff
@@ -0,0 +1,72 @@
...
+  /**
+   * Get user email from their authentication object.
+   *
+   * @param authentication Spring Security Authentication object, used to extract user details
+   * @return String user email
+   */
+  public static String getEmailFromAuth(Authentication authentication) {
+    Jwt principle = ((Jwt) authentication.getPrincipal());
+    Map<String, Object> claims = principle.getClaims();
+    String email = (String) claims.get("email");
```
dr3s, line 40: this claim should be configurable and it should be renamed `getPrincipalFromAuth`

woop: Makes sense.

Alternatively, I think the `getEmailFromAuth` method can probably be removed entirely. Right now that `email` is only used for logging purposes.

woop: I have removed it entirely.

auth/pom.xml
```diff
@@ -46,6 +52,85 @@
...
+                            </configOptions>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-javadoc-plugin</artifactId>
+                <configuration>
+                    <excludePackageNames>feast.auth.generated.client.api</excludePackageNames>
```
woop, line 131: See https://github.com/OpenAPITools/openapi-generator/issues/3272 for why this exclusion was added. I dont think the Javadocs are super critical on the client since the service API is documented in the OpenAPI spec.

auth/src/main/resources/api.yaml
```diff
@@ -0,0 +1,76 @@
...
+        "500":
+          description: Offline
+  /readiness:
+    get:
+      responses:
+        "200":
+          description: Ready
+        "500":
+          description: Not Ready
+  /checkProjectAccess:
```
dr3s, line 26: it might be better to keep this API more generic to resources and separate the authorization decision from the ACL retrieval.  What we are finding in our implementation (to be shared shortly), is that the authorization decision can be generalized for different ACL providers (Keto, homegrown, etc).  In fact our initial implementation is to just create an adapter between our homegrown system and the Keto API.

I see value in pluggable authorization logic as well but I'd rather not take on that scope until there is a concrete need.  So my preference would be for:
`/principals/{principal}/acls?resourceType=project&resourceId=[projectId}`
but if we wanted to take on the auth decision then at least:
`/checkResourceAccess`

core/src/main/resources/application.yml
```diff
@@ -91,9 +91,9 @@ feast:
 
     authorization:
       enabled: false
-      provider: none
+      provider: http
       options:
-        basePath: http://localhost:3000
+        externalAuthUrl: http://localhost:8082
```
dr3s, line 96: maybe `authorizationUrl` would be more clear?

auth/src/main/resources/api.yaml
woop: I considered generalizing instead of creating a very specific endpoint, but I felt like it would increase the scope of the PR significantly. Also, nothing prevents us from adding more generalized endpoints over time.

In terms of taking on the authorization decision: Part of the reason that is within Feast is because Keto doesn't make that easy. Keto has a pretty low level API which requires pre- and post- processing of results in order to determine whether a user is authorized to access a resource. Many other solutions are more full fledged. 

What would be the response object in the first `GET` request? And how would that be used?

`checkResourceAccess` should be pretty straightforward to add, I'm assuming `resourceId` and `resourceType` would make its way into the query?

core/src/main/resources/application.yml
woop: I added the external because the configuration options are shared across all providers, and it might not be clear that we are talking about a downstream service as opposed to an endpoint on Feast. I dont feel too strongly about this though.

auth/src/main/java/feast/auth/authorization/HttpAuthorizationProvider.java
```diff
@@ -0,0 +1,141 @@
...
+   * @return AuthorizationResult result of authorization query
+   */
+  public AuthorizationResult checkAccessToProject(String projectId, Authentication authentication) {
+
+    CheckAccessRequest checkAccessRequest = new CheckAccessRequest();
+    Object context = getContext(authentication);
+    String subject = getSubjectFromAuth(authentication, DEFAULT_SUBJECT_CLAIM);
+    checkAccessRequest.setAction("ALL");
+    checkAccessRequest.setContext(context);
+    checkAccessRequest.setResource(projectId);
```
dr3s, line 77: we probably need to change this to include the resourcetype like `org.feast.project:{projectId}` or something

woop: Oops, I fixed it but havent pushed yet.

woop: I had the do-not-merge on for protection ;)

But no worries, this at least allows your team to move ahead. I'll submit a patch tomorrow. I am going to bed now.

PR #865, opened 2020-07-08
Add Authentication and Authorization for feast serving

### LINKED ISSUE #823: Authentication and Authorization for feast serving

jmelinav:
PR enables authentication and authorization for feast serving, configurations remain same as feast core's authentication and authorization configuration.

Fixes #823 

**Does this PR introduce a user-facing change?**:Yes

```release-note
The following params on python sdk "Client" object has been renamed. 
params "core_enable_auth", "core_auth_provider" and "core_auth_token" 
has been renamed to "enable_auth", "auth_provider" and "auth_token" respectively. 
Now same params are used for enabling auth on both core and serving connections.

It doesn't break if auth is enabled only on core and not on serving. 
However client will pass the auth metadata to serving as well.

action required : If deployment has auth enabled, 
then python sdk "Client" object references should be updated to use new params. 
```


serving/src/main/java/feast/serving/config/ServingSecurityConfig.java
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Creates a SecurityMetadataSource when authentication is enabled. This allows for the
+   * configuration of endpoint level security rules.
+   *
+   * @return GrpcSecurityMetadataSource
+   */
+  @Bean
+  @ConditionalOnProperty(prefix = "feast.security.authentication", name = "enabled")
+  GrpcSecurityMetadataSource grpcSecurityMetadataSource() {
```
woop, line 55: Is there a reason why this Bean has to be in Core/Serving as opposed to the Auth module?

auth/src/main/java/feast/auth/service/AuthorizationService.java
```diff
@@ -0,0 +1,62 @@
...
+   * @param securityContext User's Spring Security Context. Used to identify user.
+   * @param project Name of the project for which membership should be tested.
+   */
+  public void authorizeRequest(SecurityContext securityContext, String project) {
+    Authentication authentication = securityContext.getAuthentication();
+    if (!this.securityProperties.getAuthorization().isEnabled()) {
+      return;
+    }
+
+    AuthorizationResult result = this.authorizationProvider.checkAccess(project, authentication);
```
woop, line 57: We will have to make sure that certificate lookup is cached in order to prevent latency spikes in serving.

jmelinav: @woop I don't get what you mean? are you taking about caching across the requests?

woop: Yip, not exactly in scope for your PR. But the keys used to verify JWTs should be cached. There are libraries that make this easy. I think we should just ensure that there are no production impacts from moving authentication to serving.

jmelinav: @woop I checked the flow in the code, `NimbusJwtDecoder.withJwkSetUri(jwkEndpointURI).build();` I do see caching in the library.

auth/src/main/java/feast/auth/service/AuthorizationService.java
```diff
@@ -0,0 +1,63 @@
...
+
+  private final SecurityProperties securityProperties;
+  private final AuthorizationProvider authorizationProvider;
+
+  @Autowired
+  public AuthorizationService(
+      SecurityProperties securityProperties,
+      ObjectProvider<AuthorizationProvider> authorizationProvider) {
+    this.securityProperties = securityProperties;
+    this.authorizationProvider = authorizationProvider.getIfUnique();
```
mrzzy, line 42: Nitpick but `getIfAvailable()` is more meaningful.

serving/src/main/java/feast/auth/credentials/CoreAuthenticationProperties.java
```diff
@@ -0,0 +1,56 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package feast.auth.credentials;
+
+import feast.common.validators.OneOfStrings;
+import java.util.Map;
+
+public class CoreAuthenticationProperties {
```
mrzzy, line 22: Might be out of scope but is it possible to get some/all of these properties from Core instead of forcing the user to specify them manually. Serving already has a dependency on Core running when starting.

serving/src/main/java/feast/auth/credentials/JwtAuthCredentials.java
```diff
@@ -0,0 +1,110 @@
...
+    this.request =
+        new Request.Builder()
+            .url(options.get(OAUTH_URL))
+            .addHeader(CONTENT_TYPE, APPLICATION_JSON)
+            .post(requestBody)
+            .build();
+  }
+
+  @Override
+  public void thisUsesUnstableApi() {
```
mrzzy, line 82: Should this be removed?

serving/src/main/java/feast/serving/config/FeastProperties.java
```diff
@@ -61,6 +66,41 @@ public FeastProperties() {}
...
+
+  /**
+   * Setter for SecurityProperties
+   *
+   * @param security :input {@link SecurityProperties} object
+   */
+  public void setSecurity(SecurityProperties security) {
+    this.security = security;
+  }
+
```
mrzzy, line 103: cmiiw @woop, but there needs to be `@PostConstruct` `validate()` method in order for validation to work. Frame of reference [Core's FeastProperties](https://github.com/feast-dev/feast/blob/dc159e484766467bca4c14610883461e6be080fa/core/src/main/java/feast/core/config/FeastProperties.java#L231)

serving/src/main/java/feast/serving/specs/CoreSpecService.java
```diff
@@ -24,20 +24,28 @@
...
   private static final Logger log = org.slf4j.LoggerFactory.getLogger(CoreSpecService.class);
   private final CoreServiceGrpc.CoreServiceBlockingStub blockingStub;
 
-  public CoreSpecService(String feastCoreHost, int feastCorePort) {
+  public CoreSpecService(
+      String feastCoreHost, int feastCorePort, ObjectProvider<CallCredentials> callCredentials) {
     ManagedChannel channel =
         ManagedChannelBuilder.forAddress(feastCoreHost, feastCorePort).usePlaintext().build();
-    blockingStub = CoreServiceGrpc.newBlockingStub(channel);
+    CallCredentials creds = callCredentials.getIfUnique();
```
mrzzy, line 43: `getIfAvailable()`?

serving/src/test/java/feast/serving/controller/ServingServiceGRpcControllerTest.java
```diff
@@ -59,23 +76,66 @@ public void setUp() {
...
+        .when(authProvider)
+        .checkAccessToProject(anyString(), any(Authentication.class));
+    service.getOnlineFeatures(validRequest, mockStreamObserver);
+    Mockito.verify(mockServingService).getOnlineFeatures(validRequest);
+  }
+
+  @Test(expected = AccessDeniedException.class)
+  public void shouldThrowErrorOnValidRequestIfRequestIsUnauthorized() {
+    service = getServingServiceGRpcController(true);
+    Authentication auth = mock(Authentication.class);
```
mrzzy, line 132: Nitpick but can we move the minor duplication here up to the `setUp()` method or a common method?

serving/src/test/java/feast/serving/controller/ServingServiceGRpcControllerTest.java
```diff
@@ -59,23 +76,66 @@ public void setUp() {
...
+    SecurityContext context = mock(SecurityContext.class);
+    SecurityContextHolder.setContext(context);
+    when(context.getAuthentication()).thenReturn(auth);
+    doReturn(AuthorizationResult.failed(null))
+        .when(authProvider)
+        .checkAccessToProject(anyString(), any(Authentication.class));
+    service.getOnlineFeatures(validRequest, mockStreamObserver);
```
mrzzy, line 139: Is there supposed to be a verify check here?

serving/src/main/java/feast/serving/controller/ServingServiceGRpcController.java
```diff
@@ -70,6 +77,8 @@ public void getFeastServingInfo(
   public void getOnlineFeatures(
       GetOnlineFeaturesRequest request,
       StreamObserver<GetOnlineFeaturesResponse> responseObserver) {
+    this.authorizationService.authorizeRequest(
```
mrzzy, line 80: ### Multiple Projects
Online Serving (`getOnlineFeatures()`) support 2 cases for supplying projects:
- Project is specified in the request. In this case only one project can be retrieved per request. This code handles that case properly
- Project is not specified in the request. Leading `getOnlineFeatures()` to look at FeatureReferences in the request for the project. Since multiple unique projects can be referenced in the Feature references , and we need to ensure that the user is authorized for all unique projects specified in the request.

### Latency
This is probably a duplicate of @woop comment on latency.
Online Serving is latency sensitive, Perhaps we would to cache authorization requests to Core for a duration of time. (ie authorization grant is valid for a duration of time before having to re verify with Core). See `CachedSpecService` for a similar cache implementation for reference

serving/src/main/java/feast/serving/controller/ServingServiceGRpcController.java
```diff
@@ -92,6 +101,10 @@ public void getBatchFeatures(
       GetBatchFeaturesRequest request, StreamObserver<GetBatchFeaturesResponse> responseObserver) {
     try {
       RequestHelper.validateBatchRequest(request);
+      // Authorize only valid request, Get the project from 1st feature.
+      // "Project" remains same in the request for all features.
+      this.authorizationService.authorizeRequest(
+          SecurityContextHolder.getContext(), request.getFeatures(0).getProject());
```
mrzzy, line 107: See comment above about **multiple projects** second case

serving/src/main/java/feast/serving/controller/ServingServiceGRpcController.java
jmelinav: From what I understood from the code, if project is not specified on the api `client.get_online_features() or client.get_historical_features()` a default value set in the config will be passed. I see that requested features could be in different feature sets but I don't see an option to refer to multiple projects in a single request. Correct me if my understanding is wrong. Below is what the pydoc says, `feature_refs: List of feature references that will be returned for each entity.
                Each feature reference should have the following format:
                "feature_set:feature" where "feature_set" & "feature" refer to
                the feature and feature set names respectively.
                Only the feature name is required.`

serving/src/main/java/feast/auth/credentials/JwtAuthCredentials.java
jmelinav: no, it's unimplemented method in the abstract class. I believe it will be removed in future io.grpc releases.

serving/src/test/java/feast/serving/controller/ServingServiceGRpcControllerTest.java
jmelinav: annotation `@Test(expected = AccessDeniedException.class)` is verifying the exception thrown.

serving/src/main/java/feast/auth/credentials/CoreAuthenticationProperties.java
dr3s: Wouldn't it be nice if there wasn't a connection dependency though?  Maybe these properties are only useful for core but if there is a direct need for Serving to know about them, I would argue that the dependency to them should be direct.

I guess the argument for getting them from core would be if serving is just delegating and it is feasible that serving could connect to multiple core instances someday and the settings would be specific to each.

serving/src/main/java/feast/serving/controller/ServingServiceGRpcController.java
dr3s: I definitely don't want to break existing clients but feast is 0.x, so we should weigh the tradeoff of more complexity vs a breaking change.

serving/src/main/java/feast/serving/controller/ServingServiceGRpcController.java
dr3s: this needs safe handling regardless.

serving/src/main/java/feast/serving/controller/ServingServiceGRpcController.java
dr3s: Caching is definitely desirable but we just made the move to externalize authorization decisions to another service.  There could also be caching there but I think given the nature of online serving, it makes sense to also cache the results from that service in serving directly.  It would have to be a pretty short and small cache though I would think.  Hopefully the number of users x projects is small for online serving.  In our case it should be.

jmelinav: @mrzzy I see what you mean, I had a look at the v0.4 sdk. Caching is definitely one solution, How about we propose an adaption for the clients fetching features from multiple projects. The adoption would be to add access to `default` project for all such clients. I see that client passes `project=default` if not set. That being said, irrespective of we propose adoption or not, caching would be helpful.

mrzzy: > The adoption would be to add access to default project for all such clients. I see that client passes project=default if not set.

This seems to be a hacky way of solving things. I don't think we want a situation where we have to deal with special project `default` with special properties.

> There could also be caching there but I think given the nature of online serving, it makes sense to also cache the results from that service in serving directly.

Yeah, I am agreement on this one. Feast Serving already caches the Feature Set specs it pulls from Feast Core in [`CachedSpecService`](https://github.com/feast-dev/feast/blob/master/serving/src/main/java/feast/serving/specs/CachedSpecService.java). We could have something similar for Feast Serving.

auth/src/main/java/feast/auth/service/AuthorizationService.java
```diff
@@ -0,0 +1,63 @@
...
+import org.springframework.beans.factory.ObjectProvider;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.access.AccessDeniedException;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.context.SecurityContext;
+import org.springframework.stereotype.Service;
+
+@AllArgsConstructor
+@Service
+public class AuthorizationService {
```
woop, line 32: Doesn't this overlap with https://github.com/feast-dev/feast/blob/3926ee96f9e73ad12508fe5a249ba49e5c37be60/core/src/main/java/feast/core/service/AccessManagementService.java#L117?

Should they be deduped?

infra/scripts/test-end-to-end-batch.sh
```diff
@@ -113,9 +113,21 @@ feast:
   tracing:
     enabled: false
 
+  security:
```
woop, line 116: Is it really necessary to add these values? Can defaults not be used?

serving/pom.xml
```diff
@@ -278,12 +278,56 @@
...
-
     <dependency>
       <groupId>jakarta.validation</groupId>
       <artifactId>jakarta.validation-api</artifactId>
       <version>${jakarta.validation.api.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.springframework.security</groupId>
+      <artifactId>spring-security-core</artifactId>
+      <version>5.3.0.RELEASE</version>
```
woop, line 289: These versions need to be updated, and ideally we should set versions in the root pom.xml.

serving/src/main/java/feast/auth/credentials/CoreAuthenticationProperties.java
woop: I want to see if we can remove the dependency on Core, not make matters worse. Further coupling these systems is not a good idea in my opinion.

serving/src/main/java/feast/auth/credentials/JwtAuthCredentials.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @Override
+  public void applyRequestMetadata(
+      RequestInfo requestInfo, Executor appExecutor, MetadataApplier applier) {
+    appExecutor.execute(
+        new Runnable() {
+          @Override
+          public void run() {
+            try {
+              Response response = httpClient.newCall(request).execute();
```
woop, line 95: Does this mean that for each incoming request we will have an outgoing request for a token?

serving/src/main/java/feast/serving/config/FeastProperties.java
woop: Validation here would be very useful.

serving/src/main/java/feast/serving/config/ServingSecurityConfig.java
```diff
@@ -0,0 +1,90 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@Configuration
+@ComponentScan("feast.auth")
+public class ServingSecurityConfig {
+
+  private final FeastProperties feastProperties;
+  private static String GOOGLE = "google";
```
woop, line 53: What value does this `GOOGLE` add?

serving/src/main/java/feast/serving/controller/ServingServiceGRpcController.java
woop: Ok so if I can summarize, we will have a cache at the `authorizeReques` level?

serving/src/main/java/feast/auth/credentials/CoreAuthenticationProperties.java
dr3s: we can do that but not in this PR.  These are just options to authenticate against core which is currently broken because either authentication is not turned on when you think it is or when it is turned on serving doesn't start. 

Maybe we need a new issue for the dependency on core.  I'm curious what you think that would like like.  I think it's weird that the client has two servers to connect to.  I would rather see one API that serves all clients and put job mgmt in it's own microservice.  In that way the core API that serving is currently using becomes part of serving.

serving/src/main/java/feast/auth/credentials/JwtAuthCredentials.java
dr3s: access tokens need to be cached until near expiry, if not done already.  @jmelinav let me know if you need an example.

auth/src/main/java/feast/auth/service/AuthorizationService.java
jmelinav: Yes, it should be. I wanted to get your input before that. does it makes sense to rename it to ProjectService after removing the method?

infra/scripts/test-end-to-end-batch.sh
jmelinav: The `classpath:/application.yml` was not being passed on the feast-serving start up, when I made this change. I see that it's been added. defaults can be used now.

serving/src/main/java/feast/auth/credentials/JwtAuthCredentials.java
jmelinav: Right now yes, I will push an update with caching.

serving/src/main/java/feast/serving/config/ServingSecurityConfig.java
jmelinav: Just a constant used for comparison here, `CallCredentials CoreGrpcAuthenticationCredentials()`. It can be hardcoded since it's used only once.

serving/src/main/java/feast/serving/controller/ServingServiceGRpcController.java
jmelinav: Yes, that is correct.

jmelinav: @mrzzy Have a look at the updated implementation and let me know it makes sense.

serving/src/main/java/feast/serving/config/ServingSecurityConfig.java
```diff
@@ -0,0 +1,95 @@
...
+   * connect to core with CallCredentials
+   *
+   * @return CallCredentials
+   */
+  @Bean
+  @ConditionalOnProperty(prefix = "feast.core-authentication", name = "enabled")
+  CallCredentials CoreGrpcAuthenticationCredentials() throws IOException {
+    switch (feastProperties.getCoreAuthentication().getProvider()) {
+      case "google":
+        return MoreCallCredentials.from(GoogleCredentials.getApplicationDefault());
```
jmelinav, line 85: After reading a bit, This needs transport layer security enabled. I will do the metadata implementation similar to what we have in client.

auth/src/main/java/feast/auth/service/AuthorizationService.java
woop: Do your idea is to have one ProjectService and one AuthorizationService?

core/pom.xml
```diff
@@ -127,7 +127,7 @@
 		<dependency>
 			<groupId>org.springframework.security.oauth</groupId>
 			<artifactId>spring-security-oauth2</artifactId>
-			<version>2.5.0.RELEASE</version>
+			<version>${spring.security.oauth2.version}</version>
```
woop, line 130: Can the version be defined in the parent pom?

auth/src/main/java/feast/auth/service/AuthorizationService.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.Set;
+import lombok.AllArgsConstructor;
+import org.springframework.beans.factory.ObjectProvider;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.access.AccessDeniedException;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.context.SecurityContext;
+import org.springframework.stereotype.Service;
+
+@AllArgsConstructor
```
woop, line 31: Is this being used?

auth/src/main/java/feast/auth/service/AuthorizationService.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+  }
+
+  /**
+   * Determine whether a user has access to a set of project
+   *
+   * @param securityContext User's Spring Security Context. Used to identify user.
+   * @param projectSet set of the project for which membership should be tested.
+   */
+  public void authorizeRequest(SecurityContext securityContext, Set<String> projectSet) {
```
woop, line 71: Do we really need this method? Woudn't it be a cleaner abstraction to loop over the single project method instead of having another method here?

core/src/main/java/feast/core/config/FeastProperties.java
```diff
@@ -278,5 +280,19 @@ public void validate() {
                 + e.getMessage());
       }
     }
+
+    // Validate AuthenticationProperties
+    Set<ConstraintViolation<AuthenticationProperties>> authenticationPropsViolations =
```
woop, line 285: How do we know whether we are actually validating anything here and whether failures are being detected?

infra/scripts/test-end-to-end.sh
```diff
@@ -86,8 +76,30 @@ feast:
...
+      enabled: $ENABLE_AUTH
+      provider: jwt
+    authorization:
+      enabled: false
+      provider: none
 EOF
 
+if [[ ${ENABLE_AUTH} = "true" ]]; 
+  then
+    print_banner "Starting 'Feast core with auth'."
```
woop, line 93: Can we remote the nested quotes here and the full stop.

sdk/python/feast/client.py
```diff
@@ -165,11 +165,11 @@ def _serving_service(self):
             channel = create_grpc_channel(
                 url=self._config.get(CONFIG_SERVING_URL_KEY),
                 enable_ssl=self._config.getboolean(CONFIG_SERVING_ENABLE_SSL_KEY),
-                enable_auth=False,
+                enable_auth=self._config.getboolean(CONFIG_CORE_ENABLE_AUTH_KEY),
```
woop, line 168: Shouldnt this be `CONFIG_SERVING_ENABLE_AUTH_KEY`?

sdk/python/tests/test_client.py
```diff
@@ -103,6 +104,32 @@ def mock_client(self):
...
+        return client
+
+    @pytest.fixture
+    def secure_mock_client_with_auth(self):
+        client = Client(
+            core_url=CORE_URL,
+            serving_url=SERVING_URL,
+            core_enable_ssl=True,
+            serving_enable_ssl=True,
+            core_enable_auth=True,
```
woop, line 126: I think you should be enabling security separately here for serving.

serving/pom.xml
```diff
@@ -278,12 +278,56 @@
...
+    </dependency>
+    <dependency>
+      <groupId>com.nimbusds</groupId>
+      <artifactId>nimbus-jose-jwt</artifactId>
+      <version>8.2.1</version>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.security</groupId>
+      <artifactId>spring-security-oauth2-core</artifactId>
+      <version>${spring.security.version}</version>
```
woop, line 329: Is it possible for us to move all of these security dependency versions up to the parent?

tests/e2e/redis/basic-ingest-redis-serving.py
```diff
@@ -1091,48 +1095,59 @@ def core_service_stub(self, core_url):
         core_service_stub = CoreServiceStub(core_channel)
         return core_service_stub
 
-    def apply_feature_set(self, core_service_stub, feature_set_proto):
+    @pytest.fixture(scope="module")
+    def auth_meta_data(self, enable_auth):
```
woop, line 1099: It's not clear to me why we need this method. Wouldn't this be handled automatically by the client library?

serving/src/main/java/feast/auth/credentials/JwtAuthCredentials.java
```diff
@@ -0,0 +1,120 @@
...
+import okhttp3.RequestBody;
+import okhttp3.Response;
+import org.springframework.boot.configurationprocessor.json.JSONObject;
+import org.springframework.security.oauth2.jwt.NimbusJwtDecoder;
+
+/*
+ * Oauth Credentials Implementation for serving.
+ * Used by CoreSpecService to connect to core.
+ */
+public class JwtAuthCredentials extends CallCredentials {
```
woop, line 40: Would this class make more sense within the auth module? It would be great if we could localize auth dependencies in one module.

serving/src/main/java/feast/auth/credentials/JwtAuthCredentials.java
```diff
@@ -0,0 +1,120 @@
...
+
+  @Override
+  public void thisUsesUnstableApi() {
+    // TODO Auto-generated method stub
+
+  }
+
+  @Override
+  public void applyRequestMetadata(
+      RequestInfo requestInfo, Executor appExecutor, MetadataApplier applier) {
```
woop, line 97: I think it's still hard for me to validate whatever we are doing here without reference implementations or services. @jmelinav has shared https://auth0.com/docs/api-auth/tutorials/client-credentials, but ideally we would have more than one example to know whether what we are building is generalized. I would hate to have breaking changes in the future if we didn't get this right, and yes I know this functionality has already shipped in the Python SDK.

Have we looked at whether Keycloak supports this? Perhaps we can build an integration test around it.

tests/e2e/redis/basic-ingest-redis-serving.py
```diff
@@ -103,7 +107,7 @@ def allow_dirty(pytestconfig):
 
 @pytest.fixture(scope="module")
 def enable_auth(pytestconfig):
-    return pytestconfig.getoption("enable_auth")
+    return True if pytestconfig.getoption("enable_auth").lower() == "true" else False
```
woop, line 110: I think we need at least a single test that fails if it expects a failure due to authentication being enabled. So basically it should send no auth details and get an UNAUTHENTICATED response. That way we know that something is wrong.

serving/src/main/java/feast/auth/credentials/JwtAuthCredentials.java
jmelinav: Yes Keycloak supports client credentials flow. It's an RFC defined in Oauth2 framework. 

- https://tools.ietf.org/html/rfc6749#section-4.4 
- https://www.keycloak.org/docs/latest/server_development/index.html#authenticate-with-a-service-account

woop: Hey @jmelinav. I think Keycloak would work, but it might be overkill. What if we tried to use Spring Security for this? https://docs.spring.io/spring-security-oauth2-boot/docs/current/reference/html5/

infra/scripts/test-end-to-end.sh
dr3s: remove quotes and period

serving/src/main/java/feast/auth/credentials/JwtAuthCredentials.java
dr3s: That one is deprecated but I think spring security would be best https://github.com/spring-projects/spring-security/wiki/OAuth-2.0-Migration-Guide

auth/src/main/java/feast/auth/service/AuthorizationService.java
jmelinav: It is being used in `CoreServiceAuthTest`.

core/pom.xml
jmelinav: It is defined in parent pom.

serving/pom.xml
jmelinav: It is referring to versions defined in parent pom.

core/src/main/java/feast/core/config/FeastProperties.java
jmelinav: This is validating `OneOfStrings` validation annotations we have on providers for authentication and authorization.

sdk/python/feast/client.py
jmelinav: Should we rename this to `CONFIG_ENABLE_AUTH_KEY` instead of core or serving? 
Since, enabling it just passes metadata and doesn't hurt even if it is disabled on serving and enabled on core. The reason being introduction of more variables.
For ex, since Core and Serving could be set up to have two different auth providers we might have to add variables for that too. I feel at some point we will have to start merging some of the variables which are common. If we want Keep them separate, it would make more sense to separate them as `CoreClient` and `ServingClient`.

sdk/python/tests/test_client.py
jmelinav: Please see my comments above.

tests/e2e/redis/basic-ingest-redis-serving.py
jmelinav: `TestsBasedOnGrpc` class in the file is testing some of the functionality using grpc without using client.

sdk/python/feast/client.py
woop: Ok, lets merge them. That makes sense to me.

serving/src/main/java/feast/serving/controller/ServingServiceGRpcController.java
```diff
@@ -92,6 +106,10 @@ public void getBatchFeatures(
       GetBatchFeaturesRequest request, StreamObserver<GetBatchFeaturesResponse> responseObserver) {
     try {
       RequestHelper.validateBatchRequest(request);
+      // Authorize only valid request, Get the project from 1st feature.
+      // "Project" remains same in the request for all features.
```
mrzzy: Does the comment still reflect the code? It seems to check all the `projects` in the request now.

serving/pom.xml
woop: If the dependency versions are defined in the parent, do we have to redefine them in the child?

sdk/python/feast/client.py
jmelinav: Yes I will update the release notes.

serving/src/main/java/feast/serving/controller/ServingServiceGRpcController.java
jmelinav: Has been removed.

auth/src/main/java/feast/auth/credentials/GoogleAuthCredentials.java
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  @Override
+  public void applyRequestMetadata(
+      RequestInfo requestInfo, Executor appExecutor, MetadataApplier applier) {
+    appExecutor.execute(
+        () -> {
+          try {
+            // Fetches new token if it is not available or if token has expired.
+            if (this.accessToken == null || Instant.now().isAfter(this.tokenExpiryTime)) {
```
woop, line 66: It's strange that the Google gRPC library doesn't have refreshing the credentials built into it.

serving/src/main/java/feast/serving/controller/ServingServiceGRpcController.java
jmelinav: has been removed.

auth/src/main/java/feast/auth/credentials/GoogleAuthCredentials.java
jmelinav: It has `refreshIfExpired` which I'm using. I will check if that condition is good enough.

serving/pom.xml
dr3s: we're not defining the versions here but referencing them to declare the dependency.  there is a way to inherit dependencies but it's not best practice.  The simplest is what has been done to this point, declare versions in dependency mgmt of parent pom to keep them in sync across projects but declare only direct dependencies in child poms.

auth/src/main/java/feast/auth/service/AuthorizationService.java
dr3s: Is it needed though?  Seems like a bad practice.  Unless this is a convention for the project, constructors should exist for a reason other than a test.

tests/e2e/redis/basic-ingest-redis-serving.py
```diff
@@ -162,6 +166,13 @@ def test_version_returns_results(client):
     assert not version_info["serving"] == "not configured"
 
 
+def test_list_feature_sets_when_auth_enabled_should_raise(enable_auth):
```
jmelinav, line 169: I wanted to use `pytest.mark`. It was making it mutually exclusive. Instead of running this along with others when auth is enabled. It was either running this or others based on `-m` options. I didn't find options to pass multiple markers.

tests/e2e/redis/basic-ingest-redis-serving.py
mrzzy: Hmm, this class is tech debt waiting to be moved back into client based tests. Is there a need to have GRPC only tests where client tests do not cover the test scenario ?

mrzzy: Hmm, this is tech debt waiting to be moved back into to the client based tests. Not in scope for this PR.

serving/src/test/resources/docker-compose/docker-compose-it-core.yml
```diff
@@ -0,0 +1,53 @@
+version: '3'
```
woop, line 1: Would it be possible for us to use the docker-compose file in `infra/docker-compose`? Is there anything set differently here?

jmelinav: Only change is, this doesn't contain jupyter container. Also there was an issue with the context(relative path) for `DockerComposeContainer`, when you have multiple docker-compose files. I could define multiple `DockerComposeContainer` but that increased the boot time and also adds an extra container(I guess for management/networking) for each DockerComposeContainer.

serving/src/test/java/feast/serving/it/ServingServiceOauthAuthroizationIT.java
```diff
@@ -0,0 +1,204 @@
...
+@ActiveProfiles("it")
+@SpringBootTest(
+    properties = {
+      "feast.core-authentication.enabled=true",
+      "feast.core-authentication.provider=oauth",
+      "feast.security.authentication.enabled=true",
+      "feast.security.authorization.enabled=true"
+    })
+@Testcontainers
+public class ServingServiceOauthAuthroizationIT extends BaseAuthIT {
```
dr3s, line 58: typo

infra/docker-compose/docker-compose.online.yml
```diff
@@ -5,11 +5,16 @@ services:
     image: ${FEAST_SERVING_IMAGE}:${FEAST_VERSION}
     volumes:
       - ./serving/${FEAST_ONLINE_SERVING_CONFIG}:/etc/feast/application.yml
+      # Required if authentication is enabled on core and
+      # provider is 'google'. GOOGLE_APPLICATION_CREDENTIALS is used for connecting to core.
+      - ./gcp-service-accounts/${FEAST_BATCH_SERVING_GCP_SERVICE_ACCOUNT_KEY}:/etc/gcloud/service-accounts/key.json
```
woop, line 10: I really don't want to force users to use GCP in order to try out Feast. We already have that problem with historical serving. Is there a reason to set this as the default option? Will it fail if an account isnt provided?

jmelinav: No, It doesn't fail if it isn't provided, If auth is enabled and set to `google`, then this will be used otherwise it is ignored.

infra/scripts/test-end-to-end.sh
```diff
@@ -86,8 +76,30 @@ feast:
...
+    start_feast_core /tmp/core.warehouse.application.yml
+    print_banner "Starting Feast Serving with auth"
+  else
+    print_banner "Starting Feast core without auth"
+    start_feast_core
+    print_banner "Starting Feast Serving without auth"
+fi
+
+
 start_feast_serving /tmp/serving.warehouse.application.yml
```
mrzzy, line 103: Should this be inside the `ENABLE_AUTH` if statement?

PR #866, opened 2020-07-09
Improve parallelization in Redis Sink

pyalex:

In redis sink we pack rows into batches before actually flush them to redis. Current implementation use GroupBy null (all rows into one group) which cause a lot of shuffling. It could lead to significant throughput degradation. This PR replaces GroupBy with GroupIntoBatches with FeatureReference used as key. This should improve parallelization of Redis Sink

Fixes #

```release-note
Users can now configure the flush frequency to write to a Redis store.
```

### DIFF
--- a/storage/connectors/redis/src/main/java/feast/storage/connectors/redis/writer/RedisCustomIO.java
+++ b/storage/connectors/redis/src/main/java/feast/storage/connectors/redis/writer/RedisCustomIO.java
@@ -40,13 +40,14 @@ import org.apache.beam.sdk.transforms.windowing.*;
 import org.apache.beam.sdk.values.*;
 import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.joda.time.Duration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class RedisCustomIO {
 
-  private static final int DEFAULT_BATCH_SIZE = 1000;
-  private static final int DEFAULT_TIMEOUT = 2000;
+  private static final int DEFAULT_BATCH_SIZE = 10000;
+  private static final int DEFAULT_FREQUENCY_SECONDS = 30;
 
   private static TupleTag<FeatureRow> successfulInsertsTag =
       new TupleTag<FeatureRow>("successfulInserts") {};
@@ -68,8 +69,8 @@ public class RedisCustomIO {
 
     private PCollectionView<Map<String, Iterable<FeatureSetSpec>>> featureSetSpecs;
     private RedisIngestionClient redisIngestionClient;
-    private int batchSize;
-    private int timeout;
+    private int batchSize = DEFAULT_BATCH_SIZE;
+    private Duration flushFrequency = Duration.standardSeconds(DEFAULT_FREQUENCY_SECONDS);
 
     public Write(
         RedisIngestionClient redisIngestionClient,
@@ -83,8 +84,8 @@ public class RedisCustomIO {
       return this;
     }
 
-    public Write withTimeout(int timeout) {
-      this.timeout = timeout;
+    public Write withFlushFrequency(Duration frequency) {
+      this.flushFrequency = flushFrequency;
       return this;
     }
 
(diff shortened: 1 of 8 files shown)


storage/connectors/redis/src/main/java/feast/storage/connectors/redis/writer/RedisCustomIO.java
```diff
@@ -83,23 +84,28 @@ public Write withBatchSize(int batchSize) {
...
-                      .triggering(Repeatedly.forever(AfterPane.elementCountAtLeast(batchSize)))
-                      .discardingFiredPanes())
-              .apply("AttachSingletonKey", WithKeys.of((Void) null))
-              .apply("GroupOntoSingleton", GroupByKey.create())
+                  "AttachFeatureReferenceKey",
+                  ParDo.of(
+                      new DoFn<FeatureRow, KV<String, FeatureRow>>() {
+                        @ProcessElement
+                        public void process(ProcessContext c) {
+                          c.output(KV.of(c.element().getFeatureSet(), c.element()));
```
woop, line 103: What if the rows themselves are split across workers prior to grouping. Meaning the act of grouping them by this key forces them to be collocated on a worker which causes shuffling. I think the approach makes sense, but I wonder if this grouping shouldnt be done as the first step in the Beam pipeline instead of here at the back during writes. So basically as soon as we identify the key (feature set) then we can group/collocate those elements and do all other processes on them (metrics, validation, writes).

pyalex: 1. True. shuffling didn't disappear completely but with null key it was much worse.
2. I like the idea with keeping groups through the whole pipeline, but I don't know how to implement it and I'm not sure it feasible at all, since it's runner's optimization responsibility.

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/writer/RedisFeatureSink.java
```diff
@@ -113,11 +114,29 @@ public RedisFeatureSink withSpecsView(
...
-      return new RedisCustomIO.Write(
-          new RedisClusterIngestionClient(getRedisClusterConfig()), getSpecsView());
+      RedisCustomIO.Write writer =
+          new RedisCustomIO.Write(
+              new RedisClusterIngestionClient(getRedisClusterConfig()), getSpecsView());
+
+      if (getRedisClusterConfig().getFlushFrequencySeconds() > 0) {
+        writer =
+            writer.withFlushFrequency(
+                Duration.standardSeconds(getRedisClusterConfig().getFlushFrequencySeconds()));
```
woop, line 124: Should there be a default here or an exception if it isnt set? In which case would you not have a flush frequency used?

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/writer/RedisCustomIO.java
woop: Ok, lets just keep this in mind. Dont think we have to optimize it now.

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/writer/RedisFeatureSink.java
pyalex: Default is specified inside RedisCustomIO https://github.com/feast-dev/feast/pull/866/files#diff-f780c1a978d6503861ed53a1084d79d9R50
Do you want to pull it here?

woop: I dont think it needs to be pulled into this location, I just want to make sure there are no silent exceptions. If this value always needs to be set then we need an exception if the predicate fails.

pyalex: I pulled default values to RedisFeatureSink, so all configuration would happen in one place

pyalex: So we have default values & we won't fail in case some config wasn't set

PR #867, opened 2020-07-09
Add IngestionId & EventTimestamp to FeatureRowBatch to calculate lag metric correctly

pyalex:

Currently row based metrics (like lag) are calculated incorrectly due to absence of correct timestamp in FeatureRow.

Fixes #

```release-note

```


storage/connectors/bigquery/src/main/java/feast/storage/connectors/bigquery/compression/FeatureRowsBatch.java
```diff
@@ -132,13 +139,24 @@ private void initValues() {
   }
 
   private void toColumnar(Iterable<FeatureRowProto.FeatureRow> featureRows) {
+    int timestampColumnIdx = schema.getFieldCount() - 2;
```
khorshuheng, line 142: If someone change the schema ordering / add new fields, this part of the code will need to be changed. Otherwise the application might fail silently. Should we change this part such that it doesn't depend on schema ordering?

PR #868, opened 2020-07-09
Fix Python SDK ingestion for featureset name that exist in multiple projects

terryyylim:
Current, Python SDK client will always ingest into `default` project's featureset, even if project is known in FeatureSet object.

Fixes #

```release-note
Fixed a bug where the default project was always selected during ingestion from Python SDK
```


sdk/python/feast/client.py
```diff
@@ -836,12 +836,39 @@ def ingest(
...
+            >>>
+            >>> feast_client = Client(core_url="localhost:6565")
+            >>> fs_df = pd.DataFrame(
+            >>>         {
+            >>>            "datetime": [pd.datetime.now()],
+            >>>            "driver": [1001],
+            >>>            "rating": [4.3],
+            >>>         }
+            >>>     )
+            >>> client.ingest("project1:driver", fs_df)
```
woop, line 851: `project:feature_set` is not a valid feature reference. It would be better to have a `project` key on the method, or even look at the project that the user has configured in their environment.

woop:
All bug fixes need release notes. Also, please dont use `Bugfix: ` in the title. That is what `kind` is for. 

Finally, you also need to have tests to ensure that this functionality is working.

PR #870, opened 2020-07-09
Configurable Keto variables

AnujaVane:

1. Configurable permission role templates. 
2. Configurable subject from the JWT token.

### DIFF
--- a/auth/src/main/java/feast/auth/authorization/Keto/KetoAuthorizationProvider.java
+++ b/auth/src/main/java/feast/auth/authorization/Keto/KetoAuthorizationProvider.java
@@ -59,21 +59,24 @@ public class KetoAuthorizationProvider implements AuthorizationProvider {
    * @return AuthorizationResult result of authorization query
    */
   public AuthorizationResult checkAccess(String project, Authentication authentication) {
-    String email = getClaimFromAuth(authentication);
+    String subject = (String)this.options.get("subject");
+    if ((subject == null) || (subject.isEmpty()))
+      subject = "email";
+    String subjectValue = getSubjectFromAuth(authentication, subject);
     try {
       // Get all roles from Keto
       List<OryAccessControlPolicyRole> roles =
-          this.apiInstance.listOryAccessControlPolicyRoles("glob", 500L, 500L, email);
-      List<String> writeRoles = (List<String>) this.options.get("writeRoles");
+          this.apiInstance.listOryAccessControlPolicyRoles("glob", 500L, 500L, subjectValue);
+      List<String> roleTemplates = (List<String>) this.options.get("roles");
 
-      for (String writeRole : writeRoles) {
-        if (writeRole.contains("{PROJECT}")) {
-          writeRole = writeRole.replace("{PROJECT}", project);
+      for (String roleTemplate : roleTemplates) {
+        if (roleTemplate.contains("{PROJECT}")) {
+          roleTemplate = roleTemplate.replace("{PROJECT}", project);
         }
         // Loop through all roles the user has
         for (OryAccessControlPolicyRole role : roles) {
-          // If the user has an admin or project specific role, return.
-          if ((writeRole).equals(role.getId())) {
+          // If the user has a role matching one of the templates, return.
+          if ((roleTemplate).equals(role.getId())) {
             return AuthorizationResult.success();
           }
         }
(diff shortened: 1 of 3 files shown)


dr3s:
run `mvn spotless:apply`

auth/src/main/java/feast/auth/authorization/Keto/KetoAuthorizationProvider.java
```diff
@@ -33,20 +33,22 @@
...
    *     "basePath" for the provider URL
    */
-  public KetoAuthorizationProvider(Map<String, String> options) {
+  public KetoAuthorizationProvider(Map<String, Object> options) {
     if (options == null) {
       throw new IllegalArgumentException("Cannot pass empty or null options to KetoAuth");
     }
     ApiClient defaultClient = Configuration.getDefaultApiClient();
-    defaultClient.setBasePath(options.get("basePath"));
+    defaultClient.setBasePath((String) options.get("basePath"));
```
dr3s, line 49: a better name for the option is `authorizationUrl`

auth/src/main/java/feast/auth/authorization/Keto/KetoAuthorizationProvider.java
```diff
@@ -57,20 +59,28 @@ public KetoAuthorizationProvider(Map<String, String> options) {
    * @return AuthorizationResult result of authorization query
    */
   public AuthorizationResult checkAccess(String project, Authentication authentication) {
-    String email = getEmailFromAuth(authentication);
+    String subject = (String) this.options.get("subject");
+    if ((subject == null) || (subject.isEmpty())) subject = "email";
```
dr3s, line 63: `email` should be a constant like `DEFAULT_SUBJECT_CLAIM`

auth/src/main/java/feast/auth/authorization/Keto/KetoAuthorizationProvider.java
```diff
@@ -57,20 +59,28 @@ public KetoAuthorizationProvider(Map<String, String> options) {
    * @return AuthorizationResult result of authorization query
    */
   public AuthorizationResult checkAccess(String project, Authentication authentication) {
-    String email = getEmailFromAuth(authentication);
+    String subject = (String) this.options.get("subject");
```
dr3s, line 62: A better name for this option is `subjectClaim`

auth/src/main/java/feast/auth/authorization/Keto/KetoAuthorizationProvider.java
```diff
@@ -57,20 +59,28 @@ public KetoAuthorizationProvider(Map<String, String> options) {
...
+    if ((subject == null) || (subject.isEmpty())) subject = "email";
+    String subjectValue = getSubjectFromAuth(authentication, subject);
```
dr3s, line 64: no need to pass in the claim name since `this.options` can be referenced in the method

auth/src/main/java/feast/auth/authorization/Keto/KetoAuthorizationProvider.java
dr3s: rename the option and the variable actually

auth/src/main/java/feast/auth/authorization/Keto/KetoAuthorizationProvider.java
```diff
@@ -80,27 +90,32 @@ public AuthorizationResult checkAccess(String project, Authentication authentica
...
 
   /**
    * Get user email from their authentication object.
    *
    * @param authentication Spring Security Authentication object, used to extract user details
+   * @param subject Subject from the Auth token
    * @return String user email
    */
-  private String getEmailFromAuth(Authentication authentication) {
+  private String getSubjectFromAuth(Authentication authentication, String subject) {
```
dr3s, line 103: rename `subject` to `subjectClaim`

auth/src/main/java/feast/auth/authorization/Keto/KetoAuthorizationProvider.java
```diff
@@ -80,27 +90,32 @@ public AuthorizationResult checkAccess(String project, Authentication authentica
...
     Jwt principle = ((Jwt) authentication.getPrincipal());
     Map<String, Object> claims = principle.getClaims();
-    String email = (String) claims.get("email");
+    String subjectValue = (String) claims.get(subject);
```
dr3s, line 106: `subjectValue` can then be named `subject`

auth/src/main/java/feast/auth/config/SecurityProperties.java
```diff
@@ -54,6 +54,6 @@
     private String provider;
 
     // K/V options to initialize the provider with
```
dr3s, line 56: this is no longer a k/V pair since we allow a list for roles.  may be worth a conversation

auth/src/main/java/feast/auth/authorization/Keto/KetoAuthorizationProvider.java
```diff
@@ -57,20 +59,28 @@ public KetoAuthorizationProvider(Map<String, String> options) {
...
+      List<String> roleTemplates = (List<String>) this.options.get("roles");
 
-      // Loop through all roles the user has
-      for (OryAccessControlPolicyRole role : roles) {
-        // If the user has an admin or project specific role, return.
-        if (("roles:admin").equals(role.getId())
-            || (String.format("roles:feast:%s-member", project)).equals(role.getId())) {
-          return AuthorizationResult.success();
+      for (String roleTemplate : roleTemplates) {
+        if (roleTemplate.contains("{PROJECT}")) {
```
dr3s, line 72: `PROJECT` could be a constant and actually could be in a list of constants that could be replaced in the role templates.

core/src/main/resources/application.yml
```diff
@@ -94,6 +94,10 @@ feast:
       provider: none
       options:
         basePath: http://localhost:3000
+        subject: email
+        roles:
+          - role:feast:admin
+          - role:feast:{PROJECT}-member
```
woop, line 100: I am a bit wary about using complex objects as options. I would strongly prefer to stick to K/V here, so perhaps we can turn roles into a CSV.

woop:
I think a fundamental problem that I have with the current `checkAccess` implementation for Keto is that it doesn't use Keto as intended. Yes, we can retrieve the roles and parse it client-side to determine access, but that is what the [/allowed](https://www.ory.sh/keto/docs/reference/api/#check-if-a-request-is-allowed) endpoint on Keto is for. In a sense, I think we are duplicating work by adding role templating into Feast. Keto already supports this through its pattern matching.

dr3s:
replaced with https://github.com/feast-dev/feast/pull/864

PR #874, opened 2020-07-13
Add IngestionId & EventTimestamp to FeatureRowBatch to calculate lag metric correctly

pyalex:

Currently row based metrics (like lag) are calculated incorrectly due to absence of correct timestamp in FeatureRow.

Fixes #

```release-note

```


woop:
Good feedback that we have received in the past is that we should be updating the release note with the bug as well as mentioning when the bug was introduced.

pyalex:
> Good feedback that we have received in the past is that we should be updating the release note with the bug as well as mentioning when the bug was introduced.

So metrics on BQ never worked, since it didn't produce output before 0.6. But we can say that bug was introduced with 0.6

Issue #834, opened 2020-06-29
Feast 0.7 Release


woop:
Kicking off this issue to document the upcoming 0.7 release.

At the moment the roadmap is not 100% solidified yet, but we are proposing the following:

#### New Functionality
1. Entities as a first-class concept #405 
1. ~Datasets as a first-class concept~ (pushed out)
1. Feast UI (MVP)
1. Native SDK types instead of proto types
1. Audit log support
1. Request/response logging for online serving

#### Technical debt, refactoring, or housekeeping
1. Improved integration testing framework
1. Rectify all flaky batch tests
1. Decouple job management from Feast Core
1. More descriptive error logs in Feast online serving

#### Proposals
1. Training-serving skew detection proposal

duongnt:
A Feast UI would be awesome! I'm also curious on the skew detection proposal.

dr3s:
It would be great if we could complete auth in 0.7.  In addition to authentication and authorization for both core and serving, we are opening another issue for auth when serving connects to core.

woop:
> It would be great if we could complete auth in 0.7. In addition to authentication and authorization for both core and serving, we are opening another issue for auth when serving connects to core.

Hi @dr3s, would it make sense to swing back now and figure out what is still missing on the auth side? I think we are lacking in documentation, but it would still be important to list out any specific functionality that we need to add.

jmelinav:
> > It would be great if we could complete auth in 0.7. In addition to authentication and authorization for both core and serving, we are opening another issue for auth when serving connects to core.
> 
> Hi @dr3s, would it make sense to swing back now and figure out what is still missing on the auth side? I think we are lacking in documentation, but it would still be important to list out any specific functionality that we need to add.

Hi @woop , Some of the improvements we could make are:

- Removing serving dependency from core.  
- Api Gateway is an other idea @dr3s has. We could do auth at api gateway level and other service could run and connect to each other without auth. Api gateway will perform auth and route requests to services. This also eliminates the configuration of auth on every exposed service and the services connect to it. Client will have configuration for a single service. 
- Securing ingestion, we talked about adding s3/gcs as source, other option is to proxy through api gateway for small data sets.

Let me know your views.

woop:
> > > It would be great if we could complete auth in 0.7. In addition to authentication and authorization for both core and serving, we are opening another issue for auth when serving connects to core.
> > 
> > 
[...]

PR #878, opened 2020-07-14
Add REST endpoints for Feast UI

SwampertX:
**What this PR does / why we need it**: In preparation for Feast UI (#834), we
need REST endpoints for Feast UI to consume. This PR creates such endpoints.

The below HTTP endpoints have been added to Feast Core:

| HTTP Endpoint                | Method | Accept Query Parameters? |
|------------------------------|--------|--------------------------|
| `/api/v1/version`            | `GET`  | No                       |
| `/api/v1/projects`           | `GET`  | No                       |
| `/api/v1/feature-sets`       | `GET`  | Yes                      |
| `/api/v1/features`           | `GET`  | Yes                      |
| `/api/v1/feature-statistics` | `GET`  | Yes                      |

For query parameters on specific endpoints, please refer to the javadoc on each
method on the controller.

Fixes #

Yes

```release-note
The below HTTP endpoints have been added to Feast Core:

| HTTP Endpoint                | Method | Accept Query Parameters? |
|------------------------------|--------|--------------------------|
| `/api/v1/version`            | `GET`  | No                       |
| `/api/v1/projects`           | `GET`  | No                       |
| `/api/v1/feature-sets`       | `GET`  | Yes                      |
| `/api/v1/features`           | `GET`  | Yes                      |
| `/api/v1/feature-statistics` | `GET`  | Yes                      |

For query parameters on specific endpoints, please refer to the javadoc on each
method on the controller.
```


Yanson:
Do we really need a rest controller? How about [grpc-web](https://github.com/grpc/grpc-web) or [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway)?

woop:
> Do we really need a rest controller? How about [grpc-web](https://github.com/grpc/grpc-web) or [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway)?

@Yanson I am a little bit conflicted myself on which approach to take. I originally asked @SwampertX to use grpc-gateway, but after some discussion we decided to implement this as a REST controller. The biggest reason being that we didn't want to maintain another proxy and keep it in sync with our APIs, and I wasn't sure how well a gRPC proxy would play with our auth setup.

I dont feel too strongly either way though. Right now the goal is just to get a REST endpoint up so that we can continue development on the UI, however that is done.

Yanson:
> Is your concern around code maintenance or dependencies?

I'm not very opinionated on this as I haven't used either of my suggestions. It just seems like we might be creating work for ourselves and require maintaining parity with two sets of APIs.

I thought grpc-gateway was more automatic than you suggest, but my inclination would still have been to see if grpc-web gets us what we need (again, never used it myself).

core/src/main/java/feast/core/grpc/CoreServiceResponseGenerator.java
```diff
@@ -0,0 +1,155 @@
...
+import java.util.stream.Collectors;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/**
+ * Logic for feast core GRPC service. Contains methods that take in Request messages and return
+ * Response messages. This class does not take responsibility for accessManagement.
+ */
+@Component
+public class CoreServiceResponseGenerator {
```
woop, line 63: I dont understand the purpose of this class. It seems like we need a gRPC service and a REST service, and they can both share services like `SpecService`. Why do we need another layer?

woop:
> > Is your concern around code maintenance or dependencies?
> 
> I'm not very opinionated on this as I haven't used either of my suggestions. It just seems like we might be creating work for ourselves and require maintaining parity with two sets of APIs.
> 
> I thought grpc-gateway was more automatic than you suggest, but my inclination would still have been to see if grpc-web gets us what we need (again, never used it myself).

Well looking at the code right now it does seem like its way more verbose than I would have liked.

core/src/main/java/feast/core/grpc/CoreServiceResponseGenerator.java
SwampertX: I was hoping to abstract out the specific interactions with the different services, in pursuit of consistency between gRPC and REST services (by maintaining a single source of truth for their behaviours). This layer is pretty thin, I created it because I was hesitant to replicate the gRPC implementations into REST, and when one changes the other has to be manually changed as well.

core/src/main/java/feast/core/controller/CoreServiceRestController.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  @RequestMapping(
+      value = "/api/v1/projects/{project}/feature-sets/{featureSet}",
+      params = {"project", "featureSet"},
+      method = RequestMethod.GET)
+  @ResponseBody
+  public String getFeatureSet(
+      @PathVariable("project") String project, @PathVariable("featureSet") String featureSetName)
+      throws InvalidProtocolBufferException {
```
mrzzy, line 61: What does the user see if we encounter an exception?

core/src/main/java/feast/core/controller/CoreServiceRestController.java
```diff
@@ -0,0 +1,101 @@
...
+
+  @RequestMapping(value = "/api/v1/version", method = RequestMethod.GET)
+  @ResponseBody
+  public String getVersion() throws InvalidProtocolBufferException {
+    GetFeastCoreVersionResponse response = responseGenerator.generateGetFeastCoreVersionResponse();
+    return jsonPrinter.print(response);
+  }
+
+  @RequestMapping(
+      value = "/api/v1/projects/{project}/feature-sets/{featureSet}",
```
mrzzy, line 55: Nitpick. Lets reserve plurals for when talking about mutliple `projects`  or `feature-sets`. Use `project` and `feature-set` respectively.

core/src/main/java/feast/core/controller/CoreServiceRestController.java
```diff
@@ -0,0 +1,101 @@
...
+    return jsonPrinter.print(responseGenerator.generateListProjectsResponse());
+  }
+
+  @RequestMapping(value = "/api/v1/features", method = RequestMethod.GET)
+  @ResponseBody
+  public String listFeatures(@RequestParam String[] entities, @RequestParam String project)
+      throws InvalidProtocolBufferException {
+    ListFeaturesRequest.Filter.Builder filterBuilder =
+        ListFeaturesRequest.Filter.newBuilder().setProject(project);
+    for (int i = 0; i < entities.length; i++) {
```
mrzzy, line 94: Use
```
for (String entity: entities) {
//.....
}
```
instead of manual indexing.

core/src/main/java/feast/core/controller/exception/handler/RestResponseEntityExceptionHandler.java
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.web.context.request.WebRequest;
+import org.springframework.web.servlet.mvc.method.annotation.ResponseEntityExceptionHandler;
+
+@ControllerAdvice
+public class RestResponseEntityExceptionHandler extends ResponseEntityExceptionHandler {
+
+  @ExceptionHandler(value = {InvalidProtocolBufferException.class})
+  protected ResponseEntity<Object> handleInvalidProtocolBufferException(
+      RuntimeException ex, WebRequest request) {
+    String bodyOfResponse = "An invalid response was received from Feast Core.";
```
mrzzy, line 34: Seems like "Unexpected error occurred in Feast Core" might make more sense.
Also this should be returned as JSON instead of a plain string as the client has been talking to Feast Core in JSON all along using JSON.

core/src/main/java/feast/core/grpc/CoreServiceResponseGenerator.java
woop: What can this class do that `specService` can't do? I mean what functionality would go here that can't go into the `service` classes like `specService`? Seems like we already have a clear abstraction. gRPC/Rest for I/O, services for shared logic.

core/src/main/java/feast/core/controller/CoreServiceRestController.java
```diff
@@ -0,0 +1,152 @@
...
+  @ResponseBody
+  public String getVersion() throws InvalidProtocolBufferException {
+    GetFeastCoreVersionResponse response = GetFeastCoreVersionResponse.newBuilder()
+        .setVersion(feastProperties.getVersion()).build();
+    return jsonPrinter.print(response);
+  }
+
+  @RequestMapping(
+      value = "/project/{project}/feature-set/{featureSetName}",
+      params = {"project", "featureSet"},
```
terryyylim, line 80: If we're using path variables for this GET request, I don't think we need these params here.

core/src/main/java/feast/core/controller/CoreServiceRestController.java
```diff
@@ -0,0 +1,152 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping(value = "/api/v1", produces = "application/json")
```
terryyylim, line 47: @CrossOrigin annotation should be added here.

core/src/main/java/feast/core/controller/CoreServiceRestController.java
woop: can we be consistent with plural/singular?

SwampertX: @mrzzy @woop should we use plural or singular here for the path? @mrzzy [suggested singular](https://github.com/feast-dev/feast/pull/878#discussion_r455502972).

core/src/main/java/feast/core/grpc/CoreServiceResponseGenerator.java
SwampertX: It was an oversight from me due to unfamiliarity with the codebase. Removed the class.

core/src/main/java/feast/core/controller/CoreServiceRestController.java
```diff
@@ -0,0 +1,159 @@
...
+
+  private final JsonFormat.Printer jsonPrinter;
+  private final FeastProperties feastProperties;
+  private SpecService specService;
+  private JobService jobService;
+  private StatsService statsService;
+  private AccessManagementService accessManagementService;
+
+  @Autowired
+  public CoreServiceRestController(
```
woop, line 62: Please add JavaDocs to your classes and methods.

Also, it is very important that you indicate that these endpoints are experimental. Minimally that should be within the comments.

core/src/main/java/feast/core/controller/CoreServiceRestController.java
```diff
@@ -0,0 +1,159 @@
...
+    this.accessManagementService = accessManagementService;
+    this.jsonPrinter = JsonFormat.printer();
+  }
+
+  @RequestMapping(value = "/version", method = RequestMethod.GET)
+  @ResponseBody
+  public String getVersion() throws InvalidProtocolBufferException {
+    GetFeastCoreVersionResponse response =
+        GetFeastCoreVersionResponse.newBuilder().setVersion(feastProperties.getVersion()).build();
+    return jsonPrinter.print(response);
```
woop, line 81: Do we really need a jsonPrinter here? At the very least this can be externalized into a single method right?

Have a look at [ProtobufJsonFormatHttpMessageConverter](https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/http/converter/protobuf/ProtobufJsonFormatHttpMessageConverter.html)  and https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/util/JsonFormat

core/src/main/java/feast/core/controller/CoreServiceRestController.java
```diff
@@ -0,0 +1,159 @@
...
+
+  @RequestMapping(
+      value = "/project/{project}/feature-set/{featureSetName}",
+      method = RequestMethod.GET)
+  @ResponseBody
+  public String getFeatureSet(@PathVariable String project, @PathVariable String featureSetName)
+      throws InvalidProtocolBufferException {
+    GetFeatureSetRequest request =
+        GetFeatureSetRequest.newBuilder().setProject(project).setName(featureSetName).build();
+    return jsonPrinter.print(specService.getFeatureSet(request));
```
woop, line 92: How are exceptions being handled? Do we have tests for that?

core/src/main/java/feast/core/controller/CoreServiceRestController.java
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  @RequestMapping(value = "/feature-sets", method = RequestMethod.GET)
+  @ResponseBody
+  public String listFeatureSets(
+      @RequestParam String project, @RequestParam(defaultValue = "*") String name)
```
woop, line 98: Do we have to set a default value here? Wouldn't this cause duplication? Can we try to move the defaults into the services?

core/src/main/java/feast/core/controller/CoreServiceRestController.java
```diff
@@ -0,0 +1,159 @@
...
+    return jsonPrinter.print(specService.listFeatures(filter));
+  }
+
+  @RequestMapping(value = "/feature-statistics", method = RequestMethod.GET)
+  @ResponseBody
+  public String getFeatureStatistics(
+      @RequestParam String feature_set_id,
+      @RequestParam(required = false) Optional<String[]> features,
+      @RequestParam(required = false) Optional<String> store,
+      @RequestParam(name = "start_date", required = false) Optional<Timestamp> startDate,
```
woop, line 127: Is there a reason you are mixing camel and snake case?

core/src/main/java/feast/core/controller/CoreServiceRestController.java
SwampertX: Done

core/src/main/java/feast/core/controller/CoreServiceRestController.java
woop: @mrzzy how would we list all feature-sets? Wouldn't that be at `/api/v1/projects/{project}/feature-sets`?

If we only look at feature-sets then it would be weird to have `/api/v1/projects/{project}/feature-set` return multiple feature sets, so I think that is less bad than having `/api/v1/projects/{project}/feature-sets/123` point to a single feature set.

I dont think either option is wrong though.

Possibly another question that we have to ask is whether `/projects/{project}` has to be there. I am not sure if we will ever list `/projects` out, although it could theoretically happen. Raises the question of whether we should follow the Google style query parameter for projects instead.

core/src/main/java/feast/core/controller/CoreServiceRestController.java
woop: My preference and what is most commonly followed for collections is plural, although neither is wrong as long as we are consistent.

woop: You are being inconsistent by having a `/feature-sets` endpoint and then a different endpoint in `/projects/{project}/feature-set/{featureSetName}`. https://github.com/feast-dev/feast/pull/878/files#diff-8921713d5b60ce56305727b7497dd3d0R130

I think that should be a single path.

woop: If we want to be able to list all feature sets across all projects optionally then we should remove `projects` from the hierarchy and make it a query parameter

SwampertX: Will be removing this endpoint since can be done in https://github.com/feast-dev/feast/pull/878/files#diff-8921713d5b60ce56305727b7497dd3d0R128.

core/src/main/java/feast/core/controller/CoreServiceRestController.java
SwampertX: That is true. Removed and handled automatically by https://github.com/feast-dev/feast/blob/master/core/src/main/java/feast/core/config/WebMvcConfig.java#L28

core/src/main/java/feast/core/controller/CoreServiceRestController.java
SwampertX: An oversight here. Changed to camel case.

SwampertX: Currently the query params are in snake case while as parameters for the controller methods, camel case is used.

core/src/main/java/feast/core/controller/CoreServiceRestController.java
mrzzy: Hmm, the code for this diff is gone. I think what I was referring to 
> /api/v1/projects/{project}/feature-sets/123

as being unsavory uses purals.

I am OK with what is currently implemented. 
https://github.com/feast-dev/feast/blob/f666287128f7cb3eed8406d83a0494fe5653f718/core/src/main/java/feast/core/controller/CoreServiceRestController.java#L106

Although its more technically tricky this is what I favor IMO:
```
/api/v1/feature-set/{feature-set-ref} 
```
or we can replace with the Feature Set's auto generated id
```
/api/v1/feature-set/42
```

core/src/main/java/feast/core/controller/CoreServiceRestController.java
mrzzy: +1 for consistency. Leaning towards `/feature-sets` for list and `/feature-set` for get.

core/src/main/java/feast/core/controller/CoreServiceRestController.java
mrzzy: :+1: very clear javadocs

core/src/main/java/feast/core/controller/CoreServiceRestController.java
SwampertX: This endpoint will be removed as `/feature-sets` can do the same and more.

core/src/test/java/feast/core/controller/CoreServiceRestControllerTest.java
```diff
@@ -0,0 +1,62 @@
...
+  @Mock SpecService specService;
+  @Mock JobService jobService;
+  @Mock StatsService statsService;
+  @Mock AccessManagementService accessManagementService;
+
+  @InjectMocks CoreServiceRestController controller;
+
+  @Test
+  public void getVersion() {
+    String version = "v0.7";
```
woop, line 47: What happens when Feast 0.8 is released?

core/src/test/java/feast/core/controller/CoreServiceRestControllerTest.java
```diff
@@ -0,0 +1,62 @@
...
+    GetFeastCoreVersionResponse response =
+        GetFeastCoreVersionResponse.newBuilder().setVersion(version).build();
+    doReturn(version).when(feastProperties).getVersion();
+    assertEquals(response, controller.getVersion());
+  }
+
+  @Test
+  public void listProjects() {
+    String version = "v0.7";
```
woop, line 56: Should this be testing for projects?

SwampertX: Oh my, I was too sloppy when committing. I should just remove this unit test since it is kind of pointless.

core/src/main/java/feast/core/controller/CoreServiceRestController.java
SwampertX: Yep, will keep http and grpc behaviour consistent on which parameters are optional. Removed the default value here.

core/src/main/java/feast/core/controller/CoreServiceRestController.java
```diff
@@ -0,0 +1,199 @@
...
+      AccessManagementService accessManagementService) {
+    this.feastProperties = feastProperties;
+    this.specService = specService;
+    this.jobService = jobService;
+    this.statsService = statsService;
+    this.accessManagementService = accessManagementService;
+  }
+
+  /**
+   * GET /version : Fetches the version of Feast Core.
```
woop, line 82: Is there a reason to write out the path and verb in the comments? Seems like extra documentation to maintain when its already defined in code.

core/src/main/java/feast/core/controller/exception/handler/RestResponseEntityExceptionHandler.java
```diff
@@ -0,0 +1,60 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.MissingServletRequestParameterException;
+import org.springframework.web.bind.UnsatisfiedServletRequestParameterException;
+import org.springframework.web.bind.annotation.ControllerAdvice;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.context.request.WebRequest;
+import org.springframework.web.servlet.mvc.method.annotation.ResponseEntityExceptionHandler;
+
+@ControllerAdvice
+public class RestResponseEntityExceptionHandler extends ResponseEntityExceptionHandler {
```
woop, line 32: Can you please add JavaDocs to the class and methods?

core/src/main/java/feast/core/controller/CoreServiceRestController.java
SwampertX: If the javadoc will eventually be user-facing, including the path and verb in the javadoc would be useful for end-users. However if it is to annotate the code then yep it could feel redundant.

core/src/main/java/feast/core/controller/CoreServiceRestController.java
```diff
@@ -0,0 +1,199 @@
...
+@CrossOrigin
+@Slf4j
+@RequestMapping(value = "/api/v1", produces = "application/json")
+public class CoreServiceRestController {
+
+  private final FeastProperties feastProperties;
+  private SpecService specService;
+  private JobService jobService;
+  private StatsService statsService;
+  private AccessManagementService accessManagementService;
```
mrzzy, line 65: Apologies for throwing code water at the last minute, and not in scope for this PR.
Is it possible for the CoreServiceRestController to call CoreServiceImpl?
This would remove much of the code duplication required to achieve feature parity.

mrzzy:
Apologies for throwing code water at the last minute, and not in scope for this PR.
As a next step, Is it possible for the CoreServiceRestController to delegate CoreServiceImpl?

> I'm not very opinionated on this as I haven't used either of my suggestions. It just seems like we might be creating work for ourselves and require maintaining parity with two sets of APIs.
> Well looking at the code right now it does seem like its way more verbose than I would have liked.

This would remove much of the code duplication required to achieve feature parity.

dr3s:
A gateway approach should be pretty simple wrt to auth.  Just take the Authorization header and add it to the grpc metadata. But I personally like this approach better because you can customize the payload for the content type without a lot of mapping.

core/src/main/java/feast/core/config/WebSecurityConfig.java
```diff
@@ -40,7 +40,7 @@ protected void configure(HttpSecurity http) throws Exception {
 
     // Bypasses security/authentication for the following paths
     http.authorizeRequests()
-        .antMatchers("/actuator/**", "/metrics/**")
+        .antMatchers("/actuator/**", "/metrics/**", "/api/v1/**")
```
dr3s, line 43: I think this needs to be dealt with.  It's a lot of work to add authentication and prevent it from being routed around.  I haven't located all the holes but it would be nice to ensure that any new routes are secure by default.

core/src/main/java/feast/core/controller/CoreServiceRestController.java
dr3s: there are definitely tools to generate the openapi yaml and use that for documentation rather than depending upon comments for the routes, response codes, etc...

core/src/main/java/feast/core/controller/exception/handler/RestResponseEntityExceptionHandler.java
SwampertX: added, please take a look

woop:
> Not in scope for this PR. But as a next step, Is it possible for the CoreServiceRestController to delegate to CoreServiceImpl?
> 
> > I'm not very opinionated on this as I haven't used either of my suggestions. It just seems like we might be creating work for ourselves and require maintaining parity with two sets of APIs.
> > Well looking at the code right now it does seem like its way more verbose than I would have liked.
> 
> This would remove much of the code duplication required to achieve feature parity:
> 
> * Monitoring and (future) Audit Logging is centered on CoreServiceImpl.
> * Defaults set on CoreServiceImpl would propagate to CoreServiceRestController, also ensuring that the behave identically.

I like what you are going for here, but wouldn't that be akin to using an external service to translate HTTP to gRPC? In my mind we would have

1. Common services like SpecService: Contains all business logic and domain models
2. gRPC Service: Wraps common services and applies only gRPC casting, exception handling, logging.
3. REST Service: Wraps common services applies only HTTP casting (Json), exception handling, logging.

Can you point out the specific instances of duplication here? I am happy to keep things simple for a start, but it seems like we want to keep the gRPC/REST services as siblings and not parent/child. I am also concerned that whenever we want to implement a REST specific endpoint that we will then need to create a gRPC endpoint.

core/src/main/java/feast/core/config/WebSecurityConfig.java
woop: We are considering disabling http by default, and adding a flag for teams to enable it with a documented caveat that auth has not been implemented and that this endpoint is experimental.

core/src/main/java/feast/core/config/WebSecurityConfig.java
```diff
@@ -37,10 +47,15 @@
    */
   @Override
   protected void configure(HttpSecurity http) throws Exception {
+    List<String> matchersToBypass = new ArrayList<>(List.of("/actuator/**", "/metrics/**"));
+
+    if (feastProperties.securityProperties().getAuthentication().isDisableRestControllerAuth()) {
```
terryyylim, line 52: Should be `if (!feastProperties.securityProperties().getAuthentication().isDisableRestControllerAuth())` here.

core/src/test/resources/application.properties
```diff
@@ -17,6 +17,7 @@
 grpc.port=${GRPC_PORT:6565}
 
 feast.security.authentication.enabled = true
+feast.security.authentication.disableRestControlerAuth = true
```
terryyylim, line 20: `disableRestControllerAuth` missing l

SwampertX: fixed

core/src/main/java/feast/core/controller/CoreServiceRestController.java
SwampertX: It is definitely good to have. Opened an issue (https://github.com/feast-dev/feast/issues/939) to track this.

core/src/main/java/feast/core/config/WebSecurityConfig.java
SwampertX: @dr3s good point here, but I'm afraid I still don't have much idea how to configure auth on this controller. As a workaround, I have added a flag that disables auth on the Rest Controller's endpoints, which defaults to enabling auth that temporarily deals with the problem. Will open another issue for this.

core/src/main/resources/application.yml
```diff
@@ -90,6 +90,7 @@ feast:
       provider: jwt
       options:
         jwkEndpointURI: "https://www.googleapis.com/oauth2/v3/certs"
+      disableRestControllerAuth: false
```
terryyylim, line 93: Is this better to be just under `security`, instead of `security.authentication`?

SwampertX: I think better to be under `security` since we bypass both authentication and authorization.

core/src/main/resources/application.yml
```diff
@@ -98,6 +98,8 @@ feast:
         authorizationUrl: http://localhost:8082
         subjectClaim: email
         
+    disableRestControllerAuth: false
```
mrzzy, line 101: @SwampertX can we add a comment here to document what enabling and disabling the rest controller does (especially security implications ie rest controller is not authenticated.)

mrzzy, line 101: @SwampertX can we add comments documenting:
- what is this flag for?
- how does it relate to `authentication` & `authorization` flag
- Rest Controller usability if not enabled?

woop: It allows unauthenticated access to the REST controller. Perhaps `bypass` would be a better description?

mrzzy: This is more a documentation problem than a naming problem. It is not immediately clear to the user what this flag does and when they should use it. (ie need to use `disableRestControllerAuth` in order to use rest controller cause auth is not yet completed should be documented somewhere). A comment providing usage info and some context would suffice.

SwampertX: Added, please take a look @mrzzy .

mrzzy:
-r

mrzzy:
/test test-end-to-end
--

PR #883, opened 2020-07-15
Fix unit tests not running in feast.core package.

### LINKED ISSUE #882: Most Unit Tests are not Running on feast.core package.
## Expected Behavior 
When running `mvn test` to run tests, only one test class was running under the `feast.core` package:
```
[INFO]                                                                                                                                                                                                       
[INFO] --- spotless-maven-plugin:1.26.1:check (spotless-check) @ feast-core ---                                                                                                                              
[INFO]                                                                                                                                                                                                       
[INFO] --- maven-surefire-plugin:3.0.0-M4:test (default-test) @ feast-core ---                                                                                                                               
[INFO]                                                                                                                                                                                                       
[INFO] -------------------------------------------------------                                                                                                                                               
[INFO]  T E S T S                                                                                                                                                                                            
[INFO] -------------------------------------------------------                                                                                                                                               
[INFO] Running feast.core.grpc.CoreServiceAuthTest                                                                                                                                                           
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 3.255 s - in feast.core.grpc.CoreServiceAuthTest                                                                                      
[INFO]                                                                                                                                                                                                       
[INFO] Results:                                                                                                                                                                                              
[INFO]                                                                                                                                                                                                       
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0                                                                                                                                                      
[INFO]                                                                                                                                                                                                       
````
This is unexpected as there a ton more tests under [core/src/test/java/feast/core/](https://github.com/feast-dev/feast/tree/master/core/src/test/java/feast/core). The rest of the unit tests should be running.

## Cause
Investigation points to the new JUnit 5 (`org.junit.jupiter`) test harness added to Core's `pom.xml` in #793 
- `CoreServiceAuthTest` is annotated with `@Test` from `org.junpiter.api.Test`
- All other tests are  using the old `@Test`: `org.junit.Test` 

## Proposed Solution
Revert addition `JUnit` 5 on Feast Core and update `CoreServiceAuthTest` to use old `@Test`.

### PR DESCRIPTION
**Pull request by mrzzy**
Fixes most `feast.core` unit tests not running as described in issue  #882 
- Reverts JUnit 5 dependency in `core/pom.xml`
- Fix any broken unit tests. 

Fixes #882 

```release-note
NONE
```

### DIFF
--- a/core/src/main/java/feast/core/service/AccessManagementService.java
+++ b/core/src/main/java/feast/core/service/AccessManagementService.java
@@ -42,11 +42,16 @@ public class AccessManagementService {
   private AuthorizationProvider authorizationProvider;
   private ProjectRepository projectRepository;
 
+  // TODO: Remove duplication of constructor
   public AccessManagementService(
       FeastProperties feastProperties,
       ProjectRepository projectRepository,
       AuthorizationProvider authorizationProvider) {
     this.projectRepository = projectRepository;
+    // create default project if it does not yet exist.
+    if (!projectRepository.existsById(Project.DEFAULT_NAME)) {
+      this.createProject(Project.DEFAULT_NAME);
+    }
     this.authorizationProvider = authorizationProvider;
     this.securityProperties = feastProperties.getSecurity();
   }
@@ -76,6 +81,7 @@ public class AccessManagementService {
       throw new IllegalArgumentException(String.format("Project already exists: %s", name));
     }
     Project project = new Project(name);
+    System.out.println(project.getName());
     projectRepository.saveAndFlush(project);
   }
 
--- a/core/src/main/java/feast/core/service/AccessManagementService.java
+++ b/core/src/main/java/feast/core/service/AccessManagementService.java
@@ -48,10 +48,6 @@ public class AccessManagementService {
       ProjectRepository projectRepository,
       AuthorizationProvider authorizationProvider) {
     this.projectRepository = projectRepository;
-    // create default project if it does not yet exist.
-    if (!projectRepository.existsById(Project.DEFAULT_NAME)) {
-      this.createProject(Project.DEFAULT_NAME);
-    }
     this.authorizationProvider = authorizationProvider;
     this.securityProperties = feastProperties.getSecurity();
   }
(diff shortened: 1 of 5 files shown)


core/src/main/java/feast/core/service/AccessManagementService.java
```diff
@@ -42,11 +42,16 @@
   private AuthorizationProvider authorizationProvider;
   private ProjectRepository projectRepository;
 
+  // TODO: Remove duplication of constructor
```
mrzzy, line 45: @dr3s is it possible to remove the `ObjectProvider` here to remove the code duplication in two nearly identical `AccessManagementService()` constructors? This makes the code really error prone. 
(:disappointed_relieved: I was stuck debugging because I didn't realize there were two constructors.)
It seems that one constructor is used by Spring and one used by tests.

core/src/main/java/feast/core/service/AccessManagementService.java
```diff
@@ -76,6 +81,7 @@ public void createProject(String name) {
       throw new IllegalArgumentException(String.format("Project already exists: %s", name));
     }
     Project project = new Project(name);
+    System.out.println(project.getName());
```
woop, line 84: What is this about?

core/src/main/java/feast/core/service/AccessManagementService.java
```diff
@@ -42,11 +42,16 @@
...
   private ProjectRepository projectRepository;
 
+  // TODO: Remove duplication of constructor
   public AccessManagementService(
       FeastProperties feastProperties,
       ProjectRepository projectRepository,
       AuthorizationProvider authorizationProvider) {
     this.projectRepository = projectRepository;
+    // create default project if it does not yet exist.
+    if (!projectRepository.existsById(Project.DEFAULT_NAME)) {
```
woop, line 52: I wonder if this should be a part of our migration scripts instead of code.

core/src/main/java/feast/core/service/AccessManagementService.java
mrzzy: Corrected.

core/src/main/java/feast/core/service/AccessManagementService.java
mrzzy: Added migration to create the default project.

PR #884, opened 2020-07-15
Add caching to authorization

jmelinav:
Currently all requests are authorized against an external server, as this can cause latency issues in serving, Caching the authorization for a short span of time will avoid external call on subsequent requests. 

related to #823 

```release-note
Project level authorization access requests are now cached.
```

### DIFF
--- a/auth/pom.xml
+++ b/auth/pom.xml
@@ -28,10 +28,10 @@
             <artifactId>feast-common</artifactId>
             <version>${project.version}</version>
         </dependency>
-		<dependency>
-		    <groupId>org.springframework</groupId>
-		    <artifactId>spring-context-support</artifactId>
-		</dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-context-support</artifactId>
+        </dependency>
         <dependency>
             <groupId>net.devh</groupId>
             <artifactId>grpc-server-spring-boot-starter</artifactId>
@@ -96,16 +96,16 @@
             <version>3.0.2</version>
         </dependency>
         <dependency>
-			<groupId>org.springframework</groupId>
-			<artifactId>spring-test</artifactId>
-			<scope>test</scope>
-		</dependency>
-		<dependency>
-			<groupId>org.mockito</groupId>
-			<artifactId>mockito-core</artifactId>
-			<version>${mockito.version}</version>
-			<scope>test</scope>
-		</dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>${mockito.version}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
     <build>
         <plugins>
--- a/auth/pom.xml
+++ b/auth/pom.xml
@@ -28,6 +28,10 @@
             <artifactId>feast-common</artifactId>
             <version>${project.version}</version>
         </dependency>
+		<dependency>
+		    <groupId>org.springframework</groupId>
+		    <artifactId>spring-context-support</artifactId>
+		</dependency>
         <dependency>
             <groupId>net.devh</groupId>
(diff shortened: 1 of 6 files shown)


auth/pom.xml
```diff
@@ -91,6 +95,17 @@
             <artifactId>jsr305</artifactId>
             <version>3.0.2</version>
         </dependency>
+        <dependency>
+			<groupId>org.springframework</groupId>
```
woop, line 99: Can we fix the formatting here please?

auth/src/main/java/feast/auth/config/CacheConfiguration.java
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  @Override
+  public CacheResolver cacheResolver() {
+    // TODO Auto-generated method stub
+    return null;
+  }
+
+  /*
+   * KeyGenerator for HttpAuthorizationProvider.checkAccessToProject() Method.
```
woop, line 82: Can you update this comment to be a bit more descriptive?

auth/src/main/java/feast/auth/config/CacheConfiguration.java
```diff
@@ -0,0 +1,102 @@
...
+   * Key format : checkAccessToProject-<projectId>-<subjectClaim>
+   */
+  @Override
+  public KeyGenerator keyGenerator() {
+    return (Object target, Method method, Object... params) -> {
+      String projectId = (String) params[0];
```
woop, line 88: Would it be better to try and have a conditional based on the method name?

auth/src/main/java/feast/auth/utils/AuthUtils.java
```diff
@@ -0,0 +1,54 @@
...
+   * Get user email from their authentication object.
+   *
+   * @param authentication Spring Security Authentication object, used to extract user details
+   * @param subjectClaim Indicates the claim where the subject can be found
+   * @return String user email
+   */
+  public static String getSubjectFromAuth(Authentication authentication, String subjectClaim) {
+    Jwt principle = ((Jwt) authentication.getPrincipal());
+    Map<String, Object> claims = principle.getClaims();
+    String subjectValue = (String) claims.getOrDefault(subjectClaim, "");
```
woop, line 39: Isn't it better to raise an exception if we cant find the claim? In which case do we want an empty string?

auth/src/main/java/feast/auth/config/CacheConfiguration.java
jmelinav: @cacheable, By default uses concatenated method params as key. For this case since it made sense to cache based on specific field in `Authentication` object(subClaim). Also we could set `KeyGenerator` in  `@cacheable`. I could have made this a bean and mentioned it in `@cacheable`. I choose to make it default, since i didn't see any other caches that we would add to auth module.

auth/src/main/java/feast/auth/config/CacheConfiguration.java
jmelinav: Do you mean use `subClaim` if it is `checkAccessToProject` and default key generator otherwise?. It would be cleaner to write an other key generator or use spring's default key generator if we add other cache in future.

auth/src/main/java/feast/auth/config/CacheConfiguration.java
woop: Yes, but my point is that `CacheConfiguration` is not `ProjectAuhorizationCacheConfiguration`. It's a general purpose class. This method is called `keyGenerator` which is also generic. However, the implementation is specific to a single caller (project authorization).

I understand your reasoning, but I would prefer that we either create a specific bean and select `keyGenerator` in `@cacheable` or have conditional logic here in the `KeyGenerator` with specific key generation methods.

auth/src/main/java/feast/auth/config/CacheConfiguration.java
woop: Please see my comment above.

auth/pom.xml
jmelinav: Do you guys use any specific formatter? I remember it happened in my last PR too.

PR #885, opened 2020-07-15
Add Auth header

AnujaVane:
**What this PR does / why we need it**:  Added Auth token to the header of the request sent to the external Authorization API.

### DIFF
--- a/auth/src/main/resources/api.yaml
+++ b/auth/src/main/resources/api.yaml
@@ -29,7 +29,6 @@ paths:
         - name: Authorization
           in: header
           description: Auth token
-          required: true
           schema:
             type: string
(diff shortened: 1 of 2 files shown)


auth/src/main/resources/api.yaml
```diff
@@ -25,6 +25,13 @@ paths:
           description: Not Ready
   /checkAccess:
     post:
+      parameters:
+        - name: Authorization
+          in: header
+          description: Auth token
+          required: true
```
woop, line 32: Should this really be required?

AnujaVane: @woop I'll take it out, thanks!

Issue #277, opened 2019-10-29
Standardize logging libraries for Java components


woop:
Currently there is a mix of logging libraries being used in the Java components (ingestion, serving, core). We currently used both SLF4J and Log4j2.

The plan is to standardize on only SLF4J, and this issue tracks the work item.

ches:
Bringing some rationale into context here, @davidheryanto pointed out that SLF4J is [recommended for Beam][1].

Practically speaking I believe this means library modules in Feast should use the SLF4J API (only `slf4j-api` dependency), and the executable applications use Logback as the de facto implementation.

[1]: https://cloud.google.com/dataflow/docs/guides/logging

ches:
Reopening as I think #430 hasn't completely addressed this:

1. That PR added `log4j-slf4j-impl` which ["allows applications coded to the SLF4J API to use Log4j 2 as the implementation"][1]. I think (?) what we decided is that we want the opposite, SLF4J as the implementation (_and_ API), if we are already committed to it as the "lowest common denominator" by Beam. [Good summary info about all this on SO][2].
1. We still have direct usage of both APIs in Feast. The last holdout of Log4j2 is `core/src/main/java/feast/core/log/AuditLogger.java`. SLF4J is used a lot "by hand" and with the Lombok annotation.

I guess my acceptance criterion for closing this issue would be:

> A Feast operator does not need to configure both a `logback.xml` and `log4j2.xml` to change logging behavior for their deployment. (A Feast developer shouldn't either).

If there are Feast dependencies which use the Log4j (2) APIs, we should take care of the bridging so that they don't need to think about all this.

[1]: https://logging.apache.org/log4j/2.0/log4j-slf4j-impl/index.html
[2]: https://stackoverflow.com/questions/41498021/is-it-worth-to-use-slf4j-with-log4j2

Yanson:
Agree that #430 doesn't fully fix this issue. All that did was remove any impl from ingestion and standardise on log4j impl for core and serving to give them both JSON logging and env conf.

The link you posted suggests using log4j2 API, not _"SLF4J as the implementation (and API)"_.

Given Dataflow will work with log4j _or_ slf4j, I don't think it matters which we standardise on. However, since log4j2 offers more functionality (which we are using in `AuditLogger`), I think that is the more sensible choice.

Alternatively we could stick to slf4j _just for ingestion_ and use log4j for core + serving.

[...]

PR #891, opened 2020-07-18
Add Structured Audit Logging

mrzzy:
**Why we need this PR**
Feast behavior and usage be opaque due to a lack of consistent logging of:
- request/response messages handled by Feast Core/Serving.
- The user identity that was adapted.

Logging in Feast is also inconsistent making it difficult to parsed by third party logging systems.

**What this PR does**:
Adds Audit Logging to Feast Core and Feast Serving:
- Added `AuditLogger` that exposes structured logging methods `logMessage(), logAction()` etc.
    - Make `AuditLogger` disable/configurable from Core/Serving's `application.yml`
    - Log entries produced by `AuditLogger` are structured JSON and machine parsable.
- Added `AuditLogEntry` and subclasses that define the structure of each log entries and provides JSON conversion.
- Added `GrpcMessageInterceptor` to intercept incoming/request or outgoing/response and log them to the Audit Log for both Core and Serving Services. 
    -  full request/response available in JSON audit log.
    - support for displaying authenticated identity (ie authenticated user).

Refactor/Tech debt cleanups:
- Make `JobService` perform actions on jobs via `JobTask` instead of directly with `JobManager` to be consistent with `JobCoordinatorService`

Fixes #

```release-note
Add Structured Audit Logging
- Audit Log Entries are produced in structured JSON format.
- Audit Logger can be configured from application.yml:
     - feast.logging.audit.enabled - enables and disables audit logging
     - feast.logging.audit.messageEnabled - enables and disables request/response message audit logging.

### CODE REVIEW

File: common/src/main/java/feast/common/interceptors/GrpcExceptionInterceptor.java [hunk 1]
```diff
@@ -0,0 +1,102 @@
...
+
+    // Register listener to intercept uncaught exceptions
+    ServerCall.Listener<ReqT> listener = next.startCall(call, headers);
+    return new SimpleForwardingServerCallListener<ReqT>(listener) {
+      @Override
+      public void onHalfClose() {
+        try {
+          super.onHalfClose();
+        } catch (AuthenticationException e) {
+          // TODO: figure out of catching exception breaks spring security
```
mrzzy, line 56: @woop Does the current E2E auth tests verify that there are no false positives for authentication? (ie should fail to authenticate due to wrong credentials. )

common/src/main/java/feast/common/logging/AuditLogger.java
```diff
@@ -0,0 +1,79 @@
...
+import feast.common.config.LoggingProperties.AuditLogProperties;
+import feast.common.logging.AuditLogEntry.FeastInstance;
+import java.util.Collection;
+import lombok.Getter;
+import lombok.extern.log4j.Log4j2;
+import org.apache.logging.log4j.Level;
+import org.apache.logging.log4j.Marker;
+import org.apache.logging.log4j.MarkerManager;
+
+@Log4j2
```
woop, line 28: Aren't we moving away from Log4j2?

common/src/main/java/feast/common/logging/AuditLogger.java
```diff
@@ -0,0 +1,79 @@
...
+@Getter
+public class AuditLogger {
+  /** Defines the output formats that the Audit Logger can produce logs in */
+  public static enum LogFormat {
```
woop, line 32: Shouldn't the format be a part of the logger configuration xml instead of a part of the code base?

common/src/main/java/feast/common/interceptors/GrpcExceptionInterceptor.java
woop: #892 These tests will fail your authentication if you try to sign in with a JWT that cannot be verified.

common/src/main/java/feast/common/interceptors/GrpcExceptionInterceptor.java
```diff
@@ -0,0 +1,102 @@
...
+import io.grpc.StatusRuntimeException;
+import java.util.NoSuchElementException;
+import org.apache.logging.log4j.Level;
+import org.springframework.security.core.AuthenticationException;
+
+/**
+ * ExceptionInterceptor intercepts exceptions throw when handling GRPC messages. Intercepts uncaught
+ * exceptions and logs to audit logger and returns the appropriate Grpc statuses.
+ */
+public class GrpcExceptionInterceptor implements ServerInterceptor {
```
woop, line 37: Should this be in a separate PR?

common/src/main/java/feast/common/interceptors/GrpcMessageInterceptor.java
```diff
@@ -0,0 +1,82 @@
...
+  @Override
+  public <ReqT, RespT> Listener<ReqT> interceptCall(
+      ServerCall<ReqT, RespT> call, Metadata headers, ServerCallHandler<ReqT, RespT> next) {
+    // Unpack method name from call
+    String fullMethodName = call.getMethodDescriptor().getFullMethodName();
+    String methodName = fullMethodName.substring(fullMethodName.indexOf("/") + 1);
+
+    // Attempt Extract current authenticated identity.
+    Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
+    String identity = (authentication == null) ? "" : authentication.getPrincipal().toString();
```
woop, line 55: Does this work for you? I have been pulling the identity from claims.

common/src/main/java/feast/common/interceptors/GrpcMessageInterceptor.java
```diff
@@ -0,0 +1,82 @@
...
+
+    // Register forwarding call to intercept outgoing response and log to audit log
+    call =
+        new SimpleForwardingServerCall<ReqT, RespT>(call) {
+          @Override
+          public void sendMessage(RespT message) {
```
woop, line 61: Is this supposed to apply to GetOnlineFeatures as well, because in that case we will probably experience a serious performance hit. Audit logging is normally distinct from request/response logging, since with the former you expect durability and consistency but with the latter you want availability and performance.

common/src/main/java/feast/common/logging/LogEvent.java
```diff
@@ -0,0 +1,135 @@
...
+
+import com.google.auto.value.AutoOneOf;
+import com.google.auto.value.AutoValue;
+import com.google.protobuf.Message;
+import java.io.PrintWriter;
+import java.io.StringWriter;
+
+/** LogEvent describes a specific event that a log is capturing */
+@AutoOneOf(LogEvent.Kind.class)
+public abstract class LogEvent {
```
woop, line 27: My initial reaction when seeing this code is that it seems quite overwhelming. A lot of decisions have been made here in terms of the structure of our audit logging and I am not sure if we are being too ambitious here. The amount of code that needs to be maintained here is quite intense, and I am quite worried that these abstractions will end up changing quite soon.

Did you follow some kind of convention when structuring the audit logger and its associated types, or was everything custom made?

common/src/test/java/feast/common/logging/AuditLogEntryTest.java
```diff
@@ -0,0 +1,81 @@
...
+            .setOmitEntitiesInResponse(false)
+            .addAllFeatures(
+                Arrays.asList(
+                    FeatureReference.newBuilder().setName("feature1").build(),
+                    FeatureReference.newBuilder().setName("feature2").build()))
+            .build();
+
+    return Arrays.asList(
+        AuditLogEntry.newBuilder()
+            .setInstance(FeastInstance.of(FeastComponent.SERVING, "0.6.1"))
```
woop, line 43: Why is the user expected to set the version and component instead of inferring that from the environment?

common/src/test/java/feast/common/logging/AuditLogEntryTest.java
```diff
@@ -0,0 +1,81 @@
...
+        AuditLogEntry.newBuilder()
+            .setInstance(FeastInstance.of(FeastComponent.CORE, "0.6.1"))
+            .setEvent(LogEvent.ofException(new IllegalArgumentException("Bad Argument")))
+            .build());
+  }
+
+  @Test
+  public void shouldReturnStringRepresentationOfAuditLog() {
+    for (AuditLogEntry auditLog : getTestAuditLogs()) {
+      System.out.println(auditLog.toString());
```
woop, line 67: What is this testing? Wouldnt toString() always return something?

core/src/main/java/feast/core/grpc/CoreServiceImpl.java
```diff
@@ -68,110 +69,52 @@ public CoreServiceImpl(
...
-      GetFeastCoreVersionResponse response =
-          GetFeastCoreVersionResponse.newBuilder().setVersion(feastProperties.getVersion()).build();
-      responseObserver.onNext(response);
-      responseObserver.onCompleted();
-    } catch (RetrievalException | StatusRuntimeException e) {
-      log.error("Could not determine Feast Core version: ", e);
-      responseObserver.onError(
-          Status.INTERNAL.withDescription(e.getMessage()).withCause(e).asRuntimeException());
-    }
+    GetFeastCoreVersionResponse response =
```
woop, line 72: This code looks much better with the exception handling in the interceptor, but I don't think its in scope for this PR is it?

core/src/main/java/feast/core/job/task/CreateJobTask.java
```diff
@@ -0,0 +1,67 @@
...
+import feast.core.model.JobStatus;
+import lombok.extern.log4j.Log4j2;
+import org.apache.logging.log4j.Level;
+
+/** Task that starts recently created {@link Job} by using {@link feast.core.job.JobManager}. */
+@Log4j2
+public class CreateJobTask extends JobTask {
+
+  public CreateJobTask(Job job, JobManager jobManager, AuditLogger auditLog) {
+    super(job, jobManager, auditLog);
```
woop, line 34: Why did we add this constructor?

core/src/main/java/feast/core/job/task/JobTask.java
```diff
@@ -0,0 +1,72 @@
...
+import feast.common.logging.LogResource;
+import feast.common.logging.LogResource.ResourceType;
+import feast.core.job.JobManager;
+import feast.core.model.Job;
+import feast.core.model.JobStatus;
+import java.util.concurrent.Callable;
+import lombok.Getter;
+import lombok.Setter;
+import lombok.extern.log4j.Log4j2;
+import org.apache.logging.log4j.Level;
```
woop, line 30: It bothers me a bit that we are pulling in logging dependencies directly into our code. Shouldn't this be abstracted away?

common/src/main/java/feast/common/interceptors/GrpcMessageInterceptor.java
```diff
@@ -0,0 +1,82 @@
...
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.context.SecurityContextHolder;
+
+/**
+ * GrpcMessageInterceptor intercepts a GRPC calls to log handling of GRPC messages to the Audit Log.
+ * Intercepts the incoming and outgoing messages logs them to the audit log, together with method
+ * name and assumed authenticated identity (if authentication is enabled).
+ */
+public class GrpcMessageInterceptor implements ServerInterceptor {
+  private AuditLogger auditLog;
```
woop, line 40: Is it possible for us to keep things simpler by using SLF4J for logging instead of having our own AuditLogger? Example: https://github.com/yidongnan/grpc-spring-boot-starter/blob/master/examples/local-grpc-server/src/main/java/net/devh/boot/grpc/examples/local/server/LogGrpcInterceptor.java#L34

I can see some value in having a way to standardize the message format through static methods, especially if we need to have actions/resources/identities, but having to pass around an audit logger when SLF4J is already available seems like an unnecessary overhead to me.

serving/pom.xml
```diff
@@ -155,10 +155,10 @@
       <optional>true</optional>
     </dependency>
 
-    <!--compile io.github.lognet:grpc-spring-boot-starter:3.0.2'-->
     <dependency>
-      <groupId>io.github.lognet</groupId>
-      <artifactId>grpc-spring-boot-starter</artifactId>
+        <groupId>net.devh</groupId>
+        <artifactId>grpc-server-spring-boot-starter</artifactId>
+        <version>${grpc.spring.boot.starter.version}</version>
```
woop, line 161: Do we really need to set this? Shouldnt this be set in the parent-pom?

woop:
I think it is unlikely that we will get this PR merged in any time soon in its current state. It's trying to do too much.

1. New AuditLogger/LogEvents with custom messages
1. Global exception handler
1. Request/response logging
1. Job transition logging
1. Replacement of exception logging 

The MVP we need for audit logging is not much different from what we have in our existing AuditLogger. We need a way to log incoming requests to track user actions. Tracking internal state changes of jobs are not necessary yet, and even in that case I think it can be tracked at the [entity level](https://www.baeldung.com/database-auditing-jpa). 

I would recommend that you try and separate these into different PRs. The first and most important PR implements a logging interceptor that logs authenticated user requests and the incoming payload using a standardized audit log format (hopefully with SLF4J). This might serve as inspiration https://cloud.google.com/logging/docs/reference/audit/auditlog/rest/Shared.Types/AuditLog

common/src/main/java/feast/common/config/LoggingProperties.java
```diff
@@ -0,0 +1,48 @@
...
+
+    // Defines the Log Event kinds are logged with the audit log.
+    // Audit logging is disabled for the Log Event kinds that are not specified here.
+    private Set<LogEvent.Kind> kinds;
+
+    // Configures the output formats that are produced by the audit log.
+    private Set<LogFormat> formats;
+
+    // Configures the directions of messages of which matchingo messages are logged
+    // are logged to the audit log. Requires message event kind to be enabled in kinds./
```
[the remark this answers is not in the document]
mrzzy: Direction represents the flow of messages relative to the service:
- incoming - matches requests messages
- outgoing - matches response messages.

Users can use this to configure for which request "direction" message audit logging is enabled. (ie leave in only `incoming` to enable request logging)

common/src/main/java/feast/common/logging/AuditLogger.java
mrzzy: The idea was to expose the ability to configure the logger with the most common use cases without interacting with the logging configuration.

common/src/main/java/feast/common/interceptors/GrpcExceptionInterceptor.java
mrzzy: Yeah, will do.

common/src/main/java/feast/common/interceptors/GrpcMessageInterceptor.java
mrzzy: The `AuditLogger` can be configured via `application.yml` via `AuditLogProperties` and thus can be disabled for the Online Serving instance from `application.yml` for cases where latency is more important that visibility.

core/src/main/java/feast/core/grpc/CoreServiceImpl.java
mrzzy: Yeah, moving out of this PR.

common/src/main/java/feast/common/logging/LogEvent.java
mrzzy: > Did you follow some kind of convention when structuring the audit logger and its associated types, or was everything custom made?

No, Yes.


> I was hoping we could start with a simpler key/value API.

Reasons that we should go with a Key Value API.
- Infinitely more flexible and extensible as no constraints are applied.

Reasons that we should not go with a Key Value API:
- Java is statically typed. Going with a generic Key Value would mean `Map<String, Object>` which implies lots of ugly casting.
- With no enforced structure in place it will be difficult to parse the logs in upstream systems as there is no expected structure.
- Enforces some rigidity to logging so that we don't break upstream systems unknowing built upon audit logging.

The structured log will be depended on by upstream logging systems and should be treated as a user facing API that should be clearly defined, the same way Core and Serving Service API are clearly defined.

>  I am quite worried that these abstractions will end up changing quite soon once people start trying to use the audit logger

The important thing is that we are fully aware of changes instead of being blind sighted by a `map.put()`

common/src/test/java/feast/common/logging/AuditLogEntryTest.java
mrzzy: Yeah, I guess I was just trying to exercise the code path. Would remove.

serving/pom.xml
mrzzy: This was set as the interceptors do not work properly under the `lognet` grpc starter.

common/src/main/java/feast/common/interceptors/GrpcMessageInterceptor.java
mrzzy: > Is it possible for us to keep things simpler by using SLF4J for logging instead of having our own AuditLogger? Example: https://github.com/yidongnan/grpc-spring-boot-starter/blob/master/examples/local-grpc-server/src/main/java/net/devh/boot/grpc/examples/local/server/LogGrpcInterceptor.java#L34

Why we have our own Audit Logger:
- Allow us to configure the audit logger from `application.yml` which is valuable as:
     - Allows us to configure the logger via `application.yml` (ie disable request logging for Online Serving without hard coding something.)
     - We can inject Feast specific elements into the Audit Log (ie Feast Component,  Feast Version, or in the future git hash.)

mrzzy:
> 4. Job transition logging.

This already exists with the [current](https://github.com/feast-dev/feast/blob/4ea8d49d185f49246c461586ed72caeb7896e22d/core/src/main/java/feast/core/job/CreateJobTask.java#L62) `AuditLogger` 

Next steps:
- Move exception logging/Global exception handler to a separate PR. 
- Make logging code dependent on `SLF4J` instead of `Log4j2` #277

core/src/main/java/feast/core/config/CoreLoggingConfig.java
```diff
@@ -0,0 +1,36 @@
...
+import feast.common.logging.AuditLogEntry.FeastInstance;
+import feast.common.logging.AuditLogEntry.FeastInstance.FeastComponent;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.info.BuildProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ComponentScan("feast.common")
```
woop, line 28: Which components are being picked up here?

mrzzy: The `AuditLogger` and `GrpcMessageInterceptor`. https://github.com/feast-dev/feast/pull/891/files#diff-e5ba13ce49d6f28bcc91d30359f8daa4R35

common/src/main/java/feast/common/interceptors/GrpcMessageInterceptor.java
woop: > Allows us to configure the logger via application.yml (ie disable request logging for Online Serving without hard coding something.)

What have we implemented that can't be implemented using the normal logging configuration https://howtodoinjava.com/spring-boot2/logging/configure-logging-application-yml/

Also, it seems like your answer focuses on configuration but the AuditLogger is more than just a configuration holder right? We can have custom configuration in the FeastProperties for enabling/disabling request/response logging if needed, but that is a separate question to whether we need to maintain the AuditLogger and pass it around.

> We can inject Feast specific elements into the Audit Log

Since we have an application context just floating around inside of Feast, we can easily grab these settings at any time without needing to pass around an AuditLogger right? I am just trying to make sure we arent reinventing something that already exists.

core/src/main/java/feast/core/job/task/CreateJobTask.java
mrzzy: Originally this used the `@Builder` constructor from lombok and had fields duplicated across all 4 Job Task implementations. I move the fields up to `JobTask` abstract class however the `@builder` annotation has problems with inherited fields, hence the added constructor.

common/src/main/java/feast/common/logging/AuditLogger.java
woop: As mentioned in the response above, I think we can use the application.yml for configuring logging without writing our own logformat configuration.

core/src/main/java/feast/core/job/task/JobTask.java
mrzzy: Replaced with `SL4J`.

common/src/main/java/feast/common/logging/LogEvent.java
woop: My suggestion is not to throw away all static typing and just accept arbitrary string/object keys and values. Parts of what you have implemented makes sense, but other parts do not.

From a functional perspective what we want is for specific parts of the Feast code base to create a message that can be serialized and printed/sent to the audit log. so having static methods like `ofMessage(MessageLogEvent.Kind kind, String method, Message message, String identity)` makes sense. 

However, I question the value of this `LogEvent` as a container. It seems like a catchall class that provides little in terms of a contract. When would you have both a `MessageLogEvent` and `ActionLogEvent` in the same object? It doesnt seem like that should be possible and so its not clear what the scope of this `LogEvent` class should be. Will it just grow infinitely as a catchall class that can contain any subclass?

If we agree that maintaining state is a bad thing, then do we need to maintain LogEvents as an object? What is the SLF4J contract that we need to meet. Does it just expect strings?

What prevents us from having an `AuditLogger` class with static methods like `ofMessage` that converts a tight contract of objects into a Map or string that can be serialized?

core/src/main/java/feast/core/config/CoreLoggingConfig.java
woop: Ok. Just wondering if we should be a bit more specific in how we structure the common lib. So that we can have something like `feast.common.config.logging` as the scan target, but I guess this is fine for now.

core/src/main/java/feast/core/config/FeastProperties.java
```diff
@@ -46,7 +47,7 @@
 @Getter
 @Setter
 @Configuration
-@ConfigurationProperties(prefix = "feast", ignoreInvalidFields = true)
+@ConfigurationProperties(prefix = "feast", ignoreInvalidFields = false)
```
woop, line 50: Why was this changed?

common/src/main/java/feast/common/logging/AuditLogger.java
mrzzy: Used `SLF4J` instead.

common/src/main/java/feast/common/logging/AuditLogger.java
mrzzy: Removed.

common/src/main/java/feast/common/interceptors/GrpcMessageInterceptor.java
mrzzy: Yes.

core/src/main/java/feast/core/config/CoreLoggingConfig.java
mrzzy: Made `feast.common.logging` as a scan target.

core/src/main/java/feast/core/config/FeastProperties.java
mrzzy: Reverted.

common/src/main/java/feast/common/interceptors/GrpcMessageInterceptor.java
mrzzy: Updated `AuditLogger` to become a static class.

common/src/main/java/feast/common/interceptors/GrpcMessageInterceptor.java
pyalex: Yep, but we need request/response logging in online serving as well. We just want to make sure that this logging happens in async fashion and doesn't affect latency of the method

common/src/main/java/feast/common/interceptors/GrpcExceptionInterceptor.java
mrzzy: No longer relevant  as this was deemed out of scope for this PR.

core/src/main/java/feast/core/job/dataflow/DataflowJobManager.java
```diff
@@ -183,7 +182,6 @@ public Job abortJob(Job job) {
           Strings.lenientFormat("Unable to drain job with id: %s", dataflowJobId), e);
     }
 
-    job.setStatus(JobStatus.ABORTING);
```
pyalex, line 186: This API is also called from JobService, not only JobTask. Not sure it's good idea to move it to JobTask only

pyalex: ok, I see you reusing Tasks in JobService

serving/pom.xml
mrzzy: Removed version part to be set in the parent POM.

common/src/main/java/feast/common/logging/AuditLogger.java
```diff
@@ -0,0 +1,149 @@
...
+import feast.common.logging.entry.MessageAuditLogEntry;
+import feast.common.logging.entry.TransitionAuditLogEntry;
+import lombok.extern.slf4j.Slf4j;
+import org.slf4j.Marker;
+import org.slf4j.MarkerFactory;
+import org.slf4j.event.Level;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Slf4j
```
woop, line 37: I'm not expert in SLF4J, but wouldnt creating the logger here cause it to always use this class AuditLogger as the place of logging. Isn't it better to have this AuditLogger return objects from its methods and to have the logger defined in the calling class?

woop: Also, by wrapping the logging class we remove control from developers. Every time they want to add custom functionality they have to come and hack this class.

common/src/main/java/feast/common/interceptors/GrpcMessageInterceptor.java
```diff
@@ -0,0 +1,98 @@
...
+
+    MessageCollector.Builder messageCollector = MessageCollector.newBuilder();
+    // Register forwarding call to intercept outgoing response and log to audit log
+    call =
+        new SimpleForwardingServerCall<ReqT, RespT>(call) {
+          @Override
+          public void sendMessage(RespT message) {
+            super.sendMessage(message);
+            // 2. Track the response.
+            messageCollector.setResponse((Message) message);
```
woop, line 79: What happens in the pathological case where no responses are returned and the API is receiving large amount of requests?

common/src/main/java/feast/common/logging/AuditLogger.java
```diff
@@ -0,0 +1,149 @@
...
+@Component
+public class AuditLogger {
+  private static final Marker AUDIT_MARKER = MarkerFactory.getMarker("AUDIT_MARKER");
+  private static FeastInstance instance;
+  private static AuditLogProperties properties;
+
+  @Autowired
+  public AuditLogger(FeastInstance instance, LoggingProperties properties) {
```
woop, line 45: Can this be simplified by pulling in FeastProperties instead of needing a FeastInstance and LoggingProperties and a class to create a FeastInstance Bean?

common/src/main/java/feast/common/logging/entry/AuditLogEntry.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+public abstract class AuditLogEntry {
+  /** Declare Log Type to allow external Logging systems to filter out {@link AuditLogEntry} */
+  public final String logType = "FeastAuditLogEntry";
+
+  /**
+   * FeastInstance provides metadata on the running instance which created the {@link AuditLogEntry}
+   */
+  @AutoValue
+  public abstract static class FeastInstance {
```
woop, line 35: I'm still hopeful that we can replace this class with just FeastProperties or one of its subclasses.

common/src/main/java/feast/common/logging/AuditLogger.java
```diff
@@ -0,0 +1,149 @@
...
+   * @param level describing the severity of the log.
+   * @param method the name of the method handling the service call
+   * @param request protobuf message in the service call.
+   * @param response protobuf message in the servicef call.
+   */
+  public static void logMessage(
+      Level level, String method, Message request, Message response, String identity) {
+    log(
+        level,
+        MessageAuditLogEntry.newBuilder()
```
woop, line 65: If we don't allow direct access to logging AuditLogEntry classes (or its children), then is there an advantage to using the builder pattern?

common/src/main/java/feast/common/logging/config/LoggingProperties.java
```diff
@@ -14,19 +14,24 @@
...
-package feast.core.exception;
+package feast.common.logging.config;
 
-/** Exception thrown when retrieval of a spec from the registry fails. */
-public class RetrievalException extends RuntimeException {
-  public RetrievalException() {
-    super();
-  }
+import javax.validation.constraints.NotNull;
+import lombok.Getter;
```
pyalex, line 20: there was kinda agreement to move from lombok to autovalue, and delete former completely

core/src/main/java/feast/core/config/FeastProperties.java
```diff
@@ -72,11 +73,19 @@ public FeastProperties() {}
...
   @Bean
   SecurityProperties securityProperties() {
-    return this.getSecurity();
+    return getSecurity();
+  }
+
+  /* Feast Audit Logging properties */
+  @NotNull private LoggingProperties logging;
+
+  @Bean
```
woop, line 86: I know you are modeling this after the SecurityProperties, but do we really need Beans here? For example, why does JobProperties work?

common/src/main/java/feast/common/logging/AuditLogger.java
```diff
@@ -0,0 +1,149 @@
...
+   * @param resourceId resource specific identifier identifing the instance of the resource.
+   */
+  public static void logTransition(
+      Level level, String status, ResourceType resourceType, String resourceId) {
+    log(
+        level,
+        TransitionAuditLogEntry.newBuilder()
+            .setInstance(instance)
+            .setStatus(status)
+            .setResource(LogResource.of(resourceType, resourceId))
```
woop, line 108: Isn't it cleaner to have a ResourceType and a String resourceId, and not a `LogResource`? Seems like an unnecessary abstraction.

core/src/main/java/feast/core/job/task/CreateJobTask.java
```diff
@@ -0,0 +1,64 @@
...
+import feast.common.logging.AuditLogger;
+import feast.common.logging.entry.LogResource.ResourceType;
+import feast.core.job.JobManager;
+import feast.core.model.Job;
+import feast.core.model.JobStatus;
+import lombok.extern.slf4j.Slf4j;
+import org.slf4j.event.Level;
+
+/** Task that starts recently created {@link Job} by using {@link feast.core.job.JobManager}. */
+@Slf4j
```
woop, line 28: Is this annotation used?

common/src/main/java/feast/common/logging/LogEvent.java
mrzzy: Refactored away LogEvent into MessageAuditLogEntry, ActionAuditLogEntry, TransitionAuditLogEntry subclasses.

common/src/main/java/feast/common/logging/AuditLogger.java
mrzzy: > I'm not expert in SLF4J, but wouldnt creating the logger here cause it to always use this class AuditLogger as the logging class (as opposed to the class that calls the logger)? 
>  Also, by wrapping the logging class we remove control from developers. Every time they want to add custom functionality they have to come and hack this class.

One way we can workaround is to allow the user to provide their own logger when logging.

>Isn't it better to have this AuditLogger return objects from its methods and to have the logger defined in the calling class?

The `AuditLogger` serves the following purposes:
- tags the logs with `AUDIT_MARK` marker, which is used by [log4j2 logging backend]  to determine the right logging pattern to render the audit logs in.
- ability to transparently config audit logger.  (ie enable/disable message level audit logging.) from `application.yml`

common/src/main/java/feast/common/logging/config/LoggingProperties.java
mrzzy: Auto value only provides immutable classes. Spring requires mutable POJOs with setters to inject values into these properties. Without lombok we will have setter hell ala Serving's [FeastProperties](https://github.com/feast-dev/feast/blob/d07875d4efa13c573bffebadf5206d72b7042cf7/serving/src/main/java/feast/serving/config/FeastProperties.java#L199)

common/src/main/java/feast/common/logging/AuditLogger.java
mrzzy: The idea was that we consolidate referencing resources into this format across AuditLogEntry subclasses. Both `Action` and `Transition` `AuditLogEntry`s use LogResource to reference a resource.

common/src/main/java/feast/common/logging/AuditLogger.java
mrzzy: The builder pattern is useful for `MessageAuditLogEntry` in specific as it can replace this class. https://github.com/feast-dev/feast/blob/d03a46523e4894806f407a4db290c4d27511cfec/common/src/main/java/feast/common/interceptors/GrpcMessageInterceptor.java#L41.

We can reintroduce `of...()` construction methods to use where verbosity is more of a concern.

common/src/main/java/feast/common/logging/AuditLogger.java
mrzzy: `FeastProperties` are actually separate classes each coupled to Feast Core or Feast Serving. To do this we would have to introduce a base class/interface that the base classes inherit from.

core/src/main/java/feast/core/config/FeastProperties.java
mrzzy: See above comment about consolidating behind `FeastProperties`

common/src/main/java/feast/common/logging/AuditLogger.java
mrzzy: `LogResource` consolidates the resource referencing fields together. It also looks cleaner on the JSON ie 
`entry.resource.id` and `entry.resource.type` instead of `entry.resourceId` and `entry.resourceType`

common/src/main/java/feast/common/logging/entry/AuditLogEntry.java
mrzzy: See above comment about consolidating behind FeastProperties

common/src/main/java/feast/common/logging/AuditLogger.java
woop: Yea lets just use logging properties, but we don't need to have FeastInstance any more right?

common/src/main/java/feast/common/logging/AuditLogger.java
woop: Makes sense. We can keep it for now.

common/src/main/java/feast/common/interceptors/GrpcMessageInterceptor.java
mrzzy: Adding error handling and make sure that the response would default to a empty Message if not message is returned.

common/src/main/java/feast/common/logging/AuditLogger.java
mrzzy: FeastInstance will still be constructed in the individual FeastProperties classes

core/src/main/java/feast/core/job/task/CreateJobTask.java
mrzzy: Yes [here](https://github.com/mrzzy/feast/blob/9ff000e4391ffa6be6bb5496aa85719d492adae1/core/src/main/java/feast/core/job/task/CreateJobTask.java#L45)

common/src/main/java/feast/common/logging/AuditLogger.java
mrzzy: Reintroduced of() methods for `TransitionAuditLogEntry` and `ActionAuditLogEntry`.

common/src/main/java/feast/common/logging/config/LoggingProperties.java
```diff
@@ -14,30 +14,27 @@
...
+  @Getter
+  @Setter
+  public static class AuditLogProperties {
+    private FeastInstance instance;
+
+    // Whether to enable/disable audit logging entirely.
+    private boolean enabled;
+
+    // Whether to enable/disable message level (ie request/response) audit logging.
+    private boolean messageEnabled;
```
woop, line 38: Can we call this something more descriptive? `enableMessageLogging`?

mrzzy: Ok. updated.

common/src/main/java/feast/common/logging/AuditLogger.java
```diff
@@ -30,20 +31,23 @@
...
+  private static FeastInstance instance;
 
   @Autowired
-  public AuditLogger(LoggingProperties loggingProperties) {
+  public AuditLogger(LoggingProperties loggingProperties, BuildProperties buildProperties) {
     // Spring runs this constructor when creating the AuditLogger bean,
     // which allows us to populate the AuditLogger class with dependencies.
     // This allows us to use the dependencies in the AuditLogger's static methods
     AuditLogger.properties = loggingProperties.getAudit();
+    instance = FeastInstance.of(buildProperties.getArtifact(), buildProperties.getVersion());
```
woop, line 50: Still not sure why we need to persist with FeastInstance.

mrzzy: `FeastInstance` consolidates fields common to describing a feast instance and prevents a field explosion in AuditLogEntry sub classes:
https://github.com/feast-dev/feast/blob/c8271ed40e6203009f7fc436b0e13cf52e8cdd7e/common/src/main/java/feast/common/logging/entry/ActionAuditLogEntry.java#L39

Issue #860, opened 2020-07-04
Extend auth integration tests


woop:
We currently have testing of authentication and authorization through unit tests on Feast Core and as part of the Python SDK e2e tests. 

This issue is to track the extension of Feast Core tests to include proper client/server integration tests with a real database (postgres) as opposed to mocks.

woop:
Lets use this issue to list out the specific test cases that we have to have in order to ensure Feast auth is minimally covered.

We should test the following as integration tests
1. Feast Core and Feast Serving with authentication enabled but no authorization. Feast Serving should send authorized requests to Feast Core. Clients to both should be forced to provide auth details.
1. Feast Core and Feast Serving with authentication and authorization enabled. We should ensure that users cannot modify feature sets in projects for which they are not members. All other functionality should remain unaffected.
1. The functionality at SDK level should be tested, including setting Google credentials or Oauth2 creds. Both towards Feast Core and Feast Serving.
1. We should have a basic load test on Feast Serving to show that there is no negative impact based on the changes we are incorporating here.

PR #892, opened 2020-07-18
Add auth integration tests

woop:

Adds authentication and authorization integration tests. These tests use actual backing services like Postgres, Kafka, Keto, and a fake JWKs server to simulate a real deployment. 

Will be updated and merged once #886 is merged in.

Partly addresses #860, but not enough to close.

```release-note
None
```


auth/src/main/java/feast/auth/config/SecurityConfig.java
```diff
@@ -89,7 +89,7 @@ GrpcAuthenticationReader authenticationReader() {
    * @return AccessDecisionManager
    */
   @Bean
-  @ConditionalOnProperty(prefix = "feast.security.authorization", name = "enabled")
+  @ConditionalOnProperty(prefix = "feast.security.authentication", name = "enabled")
```
jmelinav, line 92: We might have to revert this change till #865 gets merged in. As you already know, If any deployment has authentication enabled but not authorization, serving boot up will fail due to connection dependency on core.

woop: Correct.

core/src/test/java/feast/core/auth/CoreServiceAuthorizationIT.java
```diff
@@ -0,0 +1,362 @@
...
+    Channel insecureChannel =
+        ManagedChannelBuilder.forAddress("localhost", feast_core_port).usePlaintext().build();
+    insecureCoreService = CoreServiceGrpc.newBlockingStub(insecureChannel);
+  }
+
+  @AfterAll
+  static void tearDown() {
+    environment.stop();
+    wireMockRule.stop();
+    CollectorRegistry.defaultRegistry.clear();
```
pyalex, line 157: this is being called in BaseIT.tearDown, could be omitted here

core/pom.xml
```diff
@@ -315,12 +314,24 @@
...
+            <version>5.6.2</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <version>5.6.2</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
```
pyalex, line 329: junit-jupiter-engine is being installed as failsafe dependency to prevent conflicts with junit-vintage-engine, so not sure it needed here, but on the other hand everything works

PR #899, opened 2020-07-27
Apply labels when creating Dataflow jobs for Ingestion

davidheryanto:

This PR sets the `label` options for Dataflow jobs created when `FeatureSets` are applied. The labels are useful for **organizing** Dataflow jobs and breaking down **costs** for Dataflow jobs in Google Cloud. 

The labels for the Dataflow jobs are derived from the `FeatureSets` the Dataflow job is expected to process when the Dataflow job is _created/updated_. This means that _new_ FeatureSets that the Dataflow job processes after it was first created (Dataflow  job listens to the message broker for _new_ FeatureSets after it is created) will not be considered in the labels unless the Dataflow job is restarted. (This is mainly due to the limitation in Dataflow jobs API in Google that only allows the setting of labels at jobs creation/update time)

When a Dataflow job is processing 2 FeatureSets with the same label _key_ but different label _value_, the resulting label value for the Dataflow job will be the concatenation of the label value from both FeatureSets. The label keys and values to be applied to the Dataflow jobs will be validated so it can meet the requirement for Dataflow job labels https://cloud.google.com/resource-manager/docs/creating-managing-labels#requirements. If it does not meet the requirement, those labels will be ignored, so that the Dataflow job creation can still proceed. (This is because labels are good to have but should not cause _disruption_ in job creation)

None

None

### DIFF
--- a/core/src/main/java/feast/core/job/dataflow/DataflowJobManager.java
+++ b/core/src/main/java/feast/core/job/dataflow/DataflowJobManager.java
@@ -41,7 +41,12 @@ import feast.proto.core.SourceProto;
 import feast.proto.core.StoreProto;
 import java.io.IOException;
 import java.security.GeneralSecurityException;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Set;
+import java.util.TreeSet;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.beam.runners.dataflow.DataflowPipelineJob;
@@ -60,6 +65,15 @@ public class DataflowJobManager implements JobManager {
   private final DataflowRunnerConfig defaultOptions;
   private final MetricsProperties metrics;
   private final IngestionJobProto.SpecsStreamingUpdateConfig specsStreamingUpdateConfig;
+
+  // For Dataflow jobs with multiple FeatureSets with same label key, the label values from
+  // different FeatureSets will be joined by JOB_LABEL_VALUE_JOIN_CHAR character
+  static final String JOB_LABEL_VALUE_JOIN_CHAR = "_";
+  // Label key and value must start with a lowercase letter and contains only letters, numbers,
+  // dashes or underscores. The total length of label key or value is at most 63 characters.
+  // Adapted from:
+  // https://cloud.google.com/resource-manager/docs/creating-managing-labels#requirements
+  static final Pattern JOB_LABEL_VALID_PATTERN = Pattern.compile("^[a-z][a-z0-9-_]{0,62}$");
 
   public DataflowJobManager(
       DataflowRunnerConfigOptions runnerConfigOptions,
@@ -114,6 +128,7 @@ public class DataflowJobManager implements JobManager {
   @Override
   public Job startJob(Job job) {
     try {
+      Map<String, String> labels = getJobLabels(job.getFeatureSets());
       String extId =
           submitDataflowJob(
               job.getId(),
@@ -121,7 +136,8 @@ public class DataflowJobManager implements JobManager {
               job.getStores().stream()
                   .map(wrapException(Store::toProto))
                   .collect(Collectors.toSet()),
-              false);
+              false,
+              labels);
       job.setExtId(extId);
       job.setStatus(JobStatus.RUNNING);
       return job;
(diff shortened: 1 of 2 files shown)


core/src/main/java/feast/core/job/dataflow/DataflowJobManager.java
```diff
@@ -298,4 +323,75 @@ private String waitForJobToRun(DataflowPipelineJob pipelineResult)
...
+   * thrown. This is because invalid labels should not result in failed Dataflow jobs and disruption
+   * to the ingestion of feature data.
+   *
+   * <p>Warning logs from Feast Core can be inspected to debug cases where expected labels for
+   * Dataflow jobs are not applied.
+   *
+   * @param featureSets FeatureSets that the DataflowJob expects to ingest from
+   * @return Labels that should be applied to the Dataflow job
+   */
+  private Map<String, String> getJobLabels(Collection<FeatureSet> featureSets) {
```
pyalex, line 346: 1. Currently we're decoupling jobs from core and job doesn't have access to models anymore. At best it should be FeatureSetProto
2. JobManager knows nothing about FeatureSets as for now, and it's not his responsibility. I suggest to move this label calculation to JobCoordinator. (but it will only have FeatureSetProto as well)

core/src/main/java/feast/core/job/dataflow/DataflowJobManager.java
```diff
@@ -262,6 +286,7 @@ private ImportOptions getPipelineOptions(
     pipelineOptions.setUpdate(update);
     pipelineOptions.setRunner(DataflowRunner.class);
     pipelineOptions.setJobName(jobName);
+    pipelineOptions.setLabels(labels);
```
pyalex, line 289: Cmiiw, but this overwrites all labels that could be set in application properties. Shouldn't we extend them instead?

core/src/main/java/feast/core/job/dataflow/DataflowJobManager.java
davidheryanto: 1. @pyalex Do you mean this field and method on the Job class are going to be removed? 
https://github.com/feast-dev/feast/blob/master/core/src/main/java/feast/core/model/Job.java#L88
https://github.com/feast-dev/feast/blob/master/core/src/main/java/feast/core/model/Job.java#L110

> By the way, I have run end-to-end test manually with this update and I can confirm the labels get propagated from the FeatureSets to the Dataflow jobs. So currently the the FeatureSets information can be retrieved from the Job.

2. I agree that the job label calculation can be moved to JobCoordinator. However, I believe this makes more sense if we have made either changes:
- add labels field to `IngestionJob.proto` so when the JobManager submit the job it knows what labels to apply
OR
- update JobManager interface with labels argument i.e  
 from `Job startJob(Job job)` to maybe  `Job startJob(Job job, Map<String,String> labels)`

However, I don't feel like making changes to the interface and Job proto, so that they are more stable and because we don't really have a clear use case yet of adding labels to Job proto aside from being used in Dataflow jobs.

So currently, I made the job labels application Dataflow job specific to avoid unnecessary changes and also because the label requirements for Dataflow job will be different than for other kinds of Job (if we have more in the future). Other kind of jobs may not even support labels at all. 

Do you have any thoughts on this @woop? Whether we should update the JobManager interface or IngestionJob proto.

core/src/main/java/feast/core/job/dataflow/DataflowJobManager.java
davidheryanto: Yes I think we should extend them. Thanks for checking

core/src/main/java/feast/core/job/dataflow/DataflowJobManager.java
pyalex: There's actually bigger problem that I forgot to mention. Job's featureSets will change during job lifetime without restarting/updating job since we delivery featureSets through kafka right now in-flight.
I'm not sure we want to restart job just to update it's labels.
As alternative proposal: I currently work on separating JobManagement from core, thus we will be able to run many JM services and each can be configured to use it's own set of labels. Hence, we will have groups of feature sets labeled.

pyalex: 1. Yes I meant those fields. Please check this [PR](https://github.com/feast-dev/feast/pull/903) however it's still under review. But in general we have plan to completely separate JobManagement & Core models. So there won't be only one direction dependency JM -> Core. Thus Core's models don't know about jobs.

pyalex: 2. We can add labels to the Job model itself (calculated on JobCoordinator level). And each JobManager can decide what to do with them.

davidheryanto:
Closing because there is going to be change in job lifecycle in https://github.com/feast-dev/feast/pull/903

PR #903, opened 2020-07-29
Use JobManager's backend as persistent storage and source of truth

pyalex:

As part of the plan to decouple JobCoordinator & Core and eventually move Job Management to dedicated service this PR is the first step in this direction.

JobCoordinator doesn't rely on any persistent storage anymore. We store temporary state in memory (see `InMemoryJobRepository`) and fetch this state each time application is being started (see `JobManager.listJobs`).

For `DataflowJobManager` it means that we fetch list of jobs directly from Dataflow. We distinguish jobs that are being managed by current application by job's labels (see property `feast.jobs.coordinator.jobSelector`). Same set of labels is being applied to job on creation.
For `DirectJobManager` we just start with empty job list.

Refactoring:
1. JobServiceTest was replaced with JobServiceIT

Fixes #

```release-note
API:
1. listIngestionJobs now returns IngestionJob with list of FeatureSetReference instead of list of FeatureSet

Properties (core):
feast:
  jobs:
    coordinator:
      # moved here from feast.jobs
      consolidate-jobs-per-source: false
      
      # labels that being applied to new jobs and used to filter existing jobs
      jobSelector:
        application: feast
```


core/pom.xml
```diff
@@ -237,6 +237,7 @@
         <dependency>
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
+            <version>1.18.12</version>
```
woop, line 240: Should these versions perhaps be set in `pom.xml` in the root?

core/src/main/java/feast/core/job/dataflow/DataflowJobManager.java
```diff
@@ -231,6 +230,78 @@ public JobStatus getJobStatus(Job job) {
...
+              } catch (IOException e) {
+                return null;
+              }
+            })
+        .filter(
+            dfJob ->
+                dfJob != null
+                    && dfJob.getLabels() != null
+                    && this.jobSelector.entrySet().stream()
+                        .allMatch(
```
woop, line 272: If `jobSelector` is empty, would this predicate return true?

core/src/main/java/feast/core/config/FeastProperties.java
```diff
@@ -90,8 +90,18 @@ SecurityProperties securityProperties() {
...
+    /* Job Coordinator related properties */
+    private CoordinatorProperties coordinator;
+
+    @Getter
+    @Setter
+    public static class CoordinatorProperties {
+      /* If true only one IngestionJob would be created per source with all subscribed stores in it */
+      private Boolean consolidateJobsPerSource = false;
+
+      /* Labels to filter by jobs managed by current instance */
```
woop, line 102: `Labels to filter by jobs managed by current instance` is a bit hard to undestand. What about `Labels to identify jobs managed by this job coordinator`?

core/src/main/java/feast/core/job/dataflow/DataflowJobManager.java
```diff
@@ -231,6 +230,78 @@ public JobStatus getJobStatus(Job job) {
...
+              try {
+                return dataflow
+                    .projects()
+                    .locations()
+                    .jobs()
+                    .get(projectId, location, dfJob.getId())
+                    .setView("JOB_VIEW_ALL")
+                    .execute();
+              } catch (IOException e) {
+                return null;
```
woop, line 264: What kind of exceptions are being skipped here?  Should we perhaps be logging a message out?

core/src/main/java/feast/core/job/dataflow/DataflowJobManager.java
pyalex: yes

pyalex: Added tests for this case

core/src/main/java/feast/core/config/FeastProperties.java
```diff
@@ -103,8 +103,18 @@ LoggingProperties loggingProperties() {
...
+    private CoordinatorProperties coordinator;
+
+    @Getter
+    @Setter
+    public static class CoordinatorProperties {
+      /* If true only one IngestionJob would be created per source with all subscribed stores in it */
+      private Boolean consolidateJobsPerSource = false;
+
+      /* Labels to identify jobs managed by this job coordinator */
+      private Map<String, String> jobSelector = new HashMap<>();
```
woop, line 116: Would it be possible to consolidate these options in the Job Runner configuration? The relationship is 1:1 right?

core/src/main/java/feast/core/job/InMemoryJobRepository.java
```diff
@@ -0,0 +1,103 @@
...
+import feast.proto.core.SourceProto;
+import java.util.*;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/**
+ * Keeps state of all jobs managed by current application in memory. On start loads persistent state
+ * through JobManager from JobManager backend.
```
woop, line 32: I think we should add an example here like `e.g., using Dataflow to maintain state about jobs` so that it is less abstract.

core/src/main/java/feast/core/job/InMemoryJobRepository.java
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * <p>Provides flexible access to objects via JPA-like filtering API.
+ */
+@Component
+public class InMemoryJobRepository implements JobRepository {
+  private final JobManager jobManager;
+
+  private Map<String, Job> storage;
```
woop, line 40: Can we get comments for all the non-obvious fields/methods please. `storage`, `findFirstBySourceAndStoreNameAndStatusNotInOrderByLastUpdatedDesc` at least.

core/src/main/java/feast/core/job/JobManager.java
```diff
@@ -70,4 +71,11 @@
    * @return job status.
    */
   JobStatus getJobStatus(Job job);
+
+  /**
+   * List of RUNNING jobs
```
woop, line 76: If it's only running jobs, then should it be `listRunningJobs`?

core/src/main/java/feast/core/job/task/RestartJobTask.java
```diff
@@ -34,11 +34,12 @@ public RestartJobTask(Job job, JobManager jobManager) {
   @Override
   public Job call() {
     try {
-      job = jobManager.restartJob(job);
-      log.info("Restart job %s for runner %s", job.getId(), job.getRunner().toString());
+      // abort job and expect replacement will be spawned
+      job = jobManager.abortJob(job);
+      log.info("Restart job {} for runner {}", job.getId(), jobManager.getRunnerType().toString());
       AuditLogger.logAction(Level.INFO, JobTasks.RESTART.name(), ResourceType.JOB, job.getId());
```
woop, line 40: Should this be changed to JobTakss.ABORT?

core/src/main/java/feast/core/model/FeatureSetDeliveryStatus.java
```diff
@@ -0,0 +1,46 @@
...
+
+import com.google.common.base.Objects;
+import feast.common.models.FeatureSetReference;
+import feast.proto.core.FeatureSetProto.FeatureSetJobDeliveryStatus;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class FeatureSetDeliveryStatus {
```
woop, line 27: Can we get JavaDocs here please.

core/src/main/java/feast/core/config/FeastProperties.java
pyalex: Not sure wdym. This selector is meant to be used by all runners that support labels/filters or have any state. And job runner has only `options` that is used to spawn new job and kinda platform specific.
I was also planning to add here feature-sets selector to specify which feature sets this instance is responsible for. So idea was to keep those configs together.

core/src/main/java/feast/core/service/JobService.java
```diff
@@ -99,19 +93,23 @@ public ListIngestionJobsResponse listJobs(ListIngestionJobsRequest request) {
...
 
           // find jobs for the matching featuresets
           Collection<Job> matchingJobs =
-              this.jobRepository
-                  .findByFeatureSetJobStatusesFeatureSetNameAndFeatureSetJobStatusesFeatureSetProjectName(
-                      fsReference.getName(), fsReference.getProject());
+              this.jobRepository.findByFeatureSetReference(
+                  FeatureSetReference.of(
+                      fsReference.getProject().isEmpty()
+                          ? Project.DEFAULT_NAME
```
woop, line 110: Should this be moved into the reference? So that the caller doesnt have to implement the handling of default project name.

core/src/main/resources/application.yml
```diff
@@ -65,9 +65,13 @@ feast:
...
 
-    # if true one job per source with many stores would be created
-    # if false one job per source-store pair would be created
-    consolidate-jobs-per-source: false
+    coordinator:
+      # if true one job per source with many stores would be created
+      # if false one job per source-store pair would be created
+      consolidate-jobs-per-source: false
+
+      jobSelector:
```
woop, line 73: Think we need to add comments here to describe exactly how this will be used.

PR #905, opened 2020-07-29
Spark + Databricks support

### LINKED ISSUE #764: Databricks Spark runner
Running on non-GCP clouds is a common request (#367) but the lack of a fully managed Beam service comparable to Google Dataflow makes this hard. The Feast code is tied to GCP in some places.

Databricks is a popular managed service available on Azure and AWS. It offers fully managed and optimized Spark environments with a REST API on top. Databricks is the main committer on Apache Spark, and on Azure the service is directly offered and supported by Microsoft, making it available to all Azure customers without an extra contract needed. The Spark clusters run in the customer's environment as fully managed VMs. The customer only pays while clusters are running, storage at rest has very little cost.

The Databricks runtime includes the open-source [Delta Lake](http://delta.io) storage layer which allows efficiently using cloud storage as a repository for historical serving.

We're starting work on a Databricks runner that we would like to submit as a PR. This issue is a place to discuss and align with the community upfront, to ensure this PR will be accepted.

- The Beam Spark runner (#362) has limitations with Structured Streaming. Spark doesn't have a standard client-server API, and causes classpath issues.
- Running Beam on Flink introduces additional infrastructure management, and requires use of community (unsupported) Flink operator. We would also have to introduce a component to replace BigQuery.

**Additional context**

### Baseline

We will baseline the work on the 0.5.0 release.

### Ingestion - Feast Core

- A class (feast.core.job.databricks.DatabricksJobManager) implementing a Databricks job manager to create, monitor and control jobs. In contrast to preexisting feast.core.job packages, it does not use the feast.ingestion Beam implementation classes, nor the Feast storage connectors, but will call out to the Databricks API to run jobs, passing job parameters such as Kafka topic name. The Databricks Run ID will be tracked as Feast job `extId`.
- Databricks jobs will be run with `max_retries=-1` to retry indefinitely in case of job failure (e.g. VM host failure).
- The JobManager `updateJob` and `restartJob` will be implemented as stop+start.
- The job definition Protobuf will be extended to allow the user to specify the Databricks cluster configuration for the job (e.g. Azure F3sv2 VM, 3 worker nodes)

### Ingestion - Spark job

- A Spark job JAR (new Feast Maven module) will be developed. The JAR should be small and have no dependencies beyond the Spark and Delta runtimes + the Spark Redis connector. Scala is usually the preferred programming language for concision, but Java can be used as well if required. The CI/CD process (and not the Feast runtime) shall deploy the JAR to Databricks' storage.
- The Spark job will connect directly to Kafka to retrieve features, using the Kafka Spark connector.
- Feast’s ValidateFeatureRowDoFn class, which is tied to Beam, will be split into a core logic class that only has dependencies on common classes (feast-datatypes, and feast-storage-api for the FailedElement class). The core logic class must be placed into a new Feast module, so that it can be used by both the Beam Ingestion module and the Spark job. The Spark job will apply the core logic validation on each incoming feature in the stream, using a UDF, and store dead lettered messages into a specific Delta table.
- The Spark job will store features into Delta tables (one table per FeatureSet) on cloud storage, for historical serving.
- The Spark job will connect directly to Redis to populate data for online serving, using the Redis Spark connector.

### Historical Retriever - API change

- The Retriever API is defined to return a list of blobs containing Avro data. Feast server-side code currently enumerates blobs on GCP storage to return a list to the client.  We want to change this, to return a cloud storage path (directory) containing Avro files, and move the blob enumeration work to the client. Motivation:
   - Do not tie server-side code to specific GCP/Azure storage client libraries
   - Efficiency, e.g. when the client does not access the data directly but triggers a remote ML job
   - Support Parquet format in the future, where the directory is the parquet "file"
- In a future step we would like to add Parquet support instead of/in addition to Avro. We expect this will lead to a significant size reduction in many cases, due to columnar compression.

### Historical Retriever - Feast Serving

- We will develop a retriever as a FEAST Storage connector (reader only). We will build a class DatabricksHistoricalRetriever that runs a Databricks job, similarly to the ingestion job.
- The class will generate a new temporary cloud storage location in which the job output will be stored (for download by the SDK client). The implementation will call the Databricks Jobs API to run the job, and then busy-loop, calling the API until the job is completed.

### Historical Retriever - Spark job

- A Spark job JAR (new Feast Maven module) will be developed, similarly to the Spark ingestion job. It will replicate the logic done in the [SQL templates for BigQuery](https://github.com/feast-dev/feast/tree/v0.5.0/storage/connectors/bigquery/src/main/resources/templates).
- Depending on the complexity of this, we might propose an initial version that is not yet fully featured (e.g. working only a single featureset).

### Historical Retriever - Python SDK

- The Python SDK is currently directly tied to the GCP Storage library.  This will be replaced with [Apache Libcloud](https://libcloud.apache.org/) to make the SDK compatible with Azure/AWS without adding complexity.

### Secrets management

- Feast to Databricks: a Personal Access Token must be passed along with REST API operations. This token will be retrieved from the application environment by the DatabricksJobManager.
- Databricks to Cloud Storage/Redis/Kafka: secrets should be preprovisioned into Databricks (Secrets API), and made available to the jobs via the Spark environment.

### Integration testing

- A Databricks API emulator (new Feast Maven module) will be developed. The emulator will be packaged as a Docker image, running a Java application exposing a REST API that is a compatible subset of the Databricks API (but only covering the jobs operations used in Feast, so a very small surface). The Java application will run real local Spark applications, so it should be able to test the real Feast Spark jobs (at small scale, and using local mounted storage instead of cloud storage).
    - I have a prototype of this working already, using the [Spark REST framework](http://sparkjava.com/) (confusing naming - this is unrelated to Apache Spark!) for simplicity and small image size. We can also refactor that to use Spring REST if desired.

- In addition, the Databricks setup can be integration tested against the real Databricks, similar to what the BigQuery test scripts are doing.

Yanson:

This PR introduces support for Spark running Databricks for ingestion with Delta Lake as the warehouse storage.

It's not our intention to merge this PR directly, rather use it for feedback on which components can be extracted into smaller PRs and what changes might be needed beforehand.

Fixes #764

```release-note
TBD
```


.github/workflows/docker_build_test_push.yml
```diff
@@ -0,0 +1,312 @@
...
+        terraform_wrapper: false
+
+    - name: Set Terraform variables infra
+      run: |
+        echo ::set-env name=ARM_CLIENT_ID::${{ secrets.ARM_CLIENT_ID }}
+        echo ::set-env name=ARM_CLIENT_SECRET::${{ secrets.ARM_CLIENT_SECRET }}
+        echo ::set-env name=ARM_SUBSCRIPTION_ID::${{ secrets.ARM_SUBSCRIPTION_ID }}
+        echo ::set-env name=ARM_TENANT_ID::${{ secrets.ARM_TENANT_ID }}
+
+    - name: Terraform init infra
```
dr3s, line 71: Why introduce terraform given the other myriad of tools already used?

infra/terraform/app/providers.tf
```diff
@@ -0,0 +1,25 @@
...
+
+  required_providers {
+    azurerm    = "= 2.13.0"
+    databricks = "= 0.2.0"
+    helm       = "~> 1.2.2"
+    kubernetes = "~> 1.11.3"
+  }
+}
+
+provider "azurerm" {
```
dr3s, line 14: I don't think any more cloud specific content should be added.  There is already a large amount of GCP specific stuff.  We are trying to add AWS support in a way that doesn't couple us to AWS.  At some point cloud providers need to be abstracted away or isolated, so they don't permeate the dependencies.

dr3s:
I wonder if there is a way to separate your PR for databricks support from your specific implementation on Azure.  Databricks support would be great but it seems like this PR has a lot of code/config tied to your specific implementation.

woop:
> I wonder if there is a way to separate your PR for databricks support from your specific implementation on Azure. Databricks support would be great but it seems like this PR has a lot of code/config tied to your specific implementation.

So this PR is more of a strawman for us to have a look at before the submit the real PRs. Right now its just about identifying any problem areas that we want to change, and then agreeing on the set of PRs that would make sense.

storage/connectors/bigquery/src/test/java/feast/storage/connectors/bigquery/writer/BigQuerySinkTest.java
```diff
@@ -368,6 +369,7 @@ public void updateSchemaWithExistingTable() {
   }
 
   @Test
+  @Ignore
```
pyalex, line 372: It shouldn't be here

pyalex:
Currently there's some difference in ingestion pipelines (beam vs spark). Not sure if there's plan to change it. Just want to save note somewhere.
1. spark doesn't accept in-flight updates of spec. specs are being loaded once on job start
2. sinks don't support store subscriptions: all valid rows go to all stores independently of store configuration

.github/workflows/docker_build_test_push.yml
Yanson: Sorry for any confusion about what's included in this PR. Terraform, Docker GitHub Actions, etc code as been left in to demonstrate our internal build + test process and help understand what needs to be integrated into the main Feast testing strategy. We don't really intend on including that in final PRs.

infra/terraform/app/providers.tf
Yanson: Again, this is to demonstrate our internal approach. There's no expectation for it to be merged.

Yanson:
> Currently there's some difference in ingestion pipelines (beam vs spark). Not sure if there's plan to change it. Just want to save note somewhere.
> 
> 1. spark doesn't accept in-flight updates of spec. specs are being loaded once on job start
> 2. sinks don't support store subscriptions: all valid rows go to all stores independently of store configuration
> 3. spark doesn't write metrics
> 4. spark doesn't send to deadletters rows that failed on sink stage
> 
> My concern is that if ingestion jobs would be developed in different directions - it would be unfeasible to test them together by some common integration tests. Hence needed support effort would increase.

Thank you for this, we intend to match existing functionality. I didn't think all of that was missing, but it's possible it's still in PR in our fork. I'll update this branch or provide clarity in due course.

mohit-udaan:
@Yanson @woop we have been waiting for the release of Feast on Azure. Any estimate on ETA for merging of this PR?

Yanson:
> @Yanson @woop we have been waiting for the release of Feast on Azure. Any estimate on ETA for merging of this PR?

This PR will not be merged directly - the features will be extracted into smaller PRs over the coming weeks. The first step is for this PR to reach feature parity and be analysed fully.

Yanson:
Closing in favour of https://github.com/feast-dev/feast/pull/1076.

PR #906, opened 2020-07-29
Dataflow runner options: disk type & streaming engine

pyalex:

This PR extends execution parameteres for Dataflow Job according to [official documentation](https://cloud.google.com/dataflow/docs/guides/specifying-exec-params#setting-other-cloud-dataflow-pipeline-options).

Next parameter were added:
* workerDiskType
* enableStreamingEngine

Next parameter were updated:
* zone -> workerZone (due to deprecation)

Fixes #

```release-note

new options introduced in application.yaml

feast:
  jobs:
    runners:
    - name: dataflow
      options:
        enableStreamingEngine: false
        workerDiskType: compute.googleapis.com/projects/asia-east1-a/diskTypes/pd-ssd
        workerZone: asia-east1-a  # renamed "zone" due to deprecation

```

### DIFF
--- a/core/src/main/resources/application.yml
+++ b/core/src/main/resources/application.yml
@@ -42,11 +42,13 @@ feast:
         options:
           project: my_gcp_project
           region: asia-east1
-          zone: asia-east1-a
+          workerZone: asia-east1-a
           tempLocation: gs://bucket/tempLocation
           network: default
           subnetwork: regions/asia-east1/subnetworks/mysubnetwork
           maxNumWorkers: 1
+          enableStreamingEngine: false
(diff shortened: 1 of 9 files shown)


core/src/main/resources/application.yml
```diff
@@ -42,11 +42,13 @@ feast:
         options:
           project: my_gcp_project
           region: asia-east1
-          zone: asia-east1-a
+          workerZone: asia-east1-a
```
woop, line 45: Should this be in the release notes?

pyalex: added

PR #908, opened 2020-07-30
Fix Online Serving unable to retrieve feature data after Feature Set update.

mrzzy:
## Problem
Feast Online Serving will throw an when the user attempts  Feature Retrieval in the following scenario:
- user applies feature set.
- user ingests feature data for that feature set.
- user updates to feature set to add or archive features.
- user is unable to retrieve Feature Data from Online Serving and gets the following error:
```
*status.statusError: rpc error: code = DataLoss desc = Failed to decode FeatureRow from bytes retrieved from redis: Possible data corruption
``` 

Update Ingestion's `RedisCustomIO` to encode feature rows by setting field name to a  hash of the actual name.
- Used hash to reduce the increase in storage used when storing encoded feature rows in redis.

> This changes the encoding of Feature Rows stored in Redis.

Update Online Serving's `FeatureRowDecoder` to support decoding Feature Rows by name hash.
- Missing fields in encoded Feature Row would be decoded in as empty values.
- Extra Fields in encoded Feature Row would be omitted from the decoded feature row.

> `FeatureRowDecoder` will continue to support decoding existing Feature Row in existing encoding stored in Redis.

Fixes #

```release-note
Encoding of Feature Rows stored in Redis hash changed:
- Fields in encoded Feature Row are now set to a hash of the field's actual name.
```

### DIFF
--- a/storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/FeatureRowDecoder.java
+++ b/storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/FeatureRowDecoder.java
@@ -105,7 +105,8 @@ public class FeatureRowDecoder {
           .addAllFields(fields)
           .build();
 
-    } else if (isEncodedV2(encodedFeatureRow)) {
+    }
+    if (isEncodedV2(encodedFeatureRow)) {
       // Feature Row v2 encoding uses a hashed name as the field name and does not have feature set
       // reference set.
       // Decoding reverts the field name to a unhashed string and set feature set reference.
--- a/storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/FeatureRowDecoder.java
+++ b/storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/FeatureRowDecoder.java
@@ -64,7 +64,7 @@ public class FeatureRowDecoder {
    * @return boolean
    */
   private boolean isEncodedV2(FeatureRow featureRow) {
-    return featureRow.getFieldsList().stream().allMatch(field -> !field.getName().isEmpty());
+    return featureRow.getFieldsList().stream().anyMatch(field -> field.getName().isEmpty());
   }
 
   /**
--- a/storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/FeatureRowDecoder.java
+++ b/storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/FeatureRowDecoder.java
@@ -41,10 +41,10 @@ public class FeatureRowDecoder {
   }
 
   /**
-   * Check if encoded feature row v1 is encoded. A Feature Row v1 is considered encoded if both it'
-   * feature set reference and fields names are not set. The no. of fields in the feature row should
-   * also match up with the number of fields in the Feature Set spec. NOTE: This method is
-   * deprecated and will be removed in Feast v0.7.
+   * Check if encoded feature row v1 is encoded. The Feature Row v1 encoding defines that a Feature
+   * Row is is considered encoded if both it's feature set reference and fields names are not set .
+   * The no. of fields in the feature row should also match up with the number of fields in the
+   * Feature Set spec. NOTE: This method is deprecated and will be removed in Feast v0.7.
    *
    * @param featureRow Feature row
    * @return boolean
@@ -57,8 +57,8 @@ public class FeatureRowDecoder {
   }
 
   /**
-   * Check if encoded feature row v2 is encoded. A Feature Row v2 is considered encoded if it's both
-   * it feature set reference and fields names are set.
+   * Check if encoded feature row v2 is encoded. The Feature Row v2 encoding defines that a Feature
+   * Row is f is considered encoded if it's both it feature set reference and fields names are set.
    *
    * @param featureRow Feature row
    * @return boolean
@@ -68,8 +68,8 @@ public class FeatureRowDecoder {
   }
 
   /**
-   * Decode feature row encoded by {@link RedisCustomIO}. NOTE: support for decoding feature row v1
-   * will be dropped in Feast 0.7
+   * Decode feature row encoded by {@link RedisCustomIO}. NOTE: support for decoding Feature Row v1
+   * encoding will be dropped in Feast 0.7
    *
    * @throws IllegalArgumentException if unable to the decode the given feature row
    * @param encodedFeatureRow Feature row
@@ -106,7 +106,7 @@ public class FeatureRowDecoder {
           .build();
 
     } else if (isEncodedV2(encodedFeatureRow)) {
-      // Encoded Feature Row v2 uses a hashed name as the field name and does not have feature set
+      // Feature Row v2 encoding uses a hashed name as the field name and does not have feature set
       // reference set.
       // Decoding reverts the field name to a unhashed string and set feature set reference.
       Map<String, Value> nameHashValueMap =
--- a/storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/FeatureRowDecoder.java
+++ b/storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/FeatureRowDecoder.java
@@ -41,10 +41,10 @@ public class FeatureRowDecoder {
   }
 
   /**
-   * Check if encoded feature row v1 is encoded. The Feature Row v1 encoding defines that a Feature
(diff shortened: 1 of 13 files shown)


storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/FeatureRowDecoder.java
```diff
@@ -36,60 +41,100 @@ public FeatureRowDecoder(String featureSetRef, FeatureSetSpec spec) {
   }
 
   /**
-   * A feature row is considered encoded if the feature set and field names are not set. This method
-   * is required for backward compatibility purposes, to allow Feast serving to continue serving non
-   * encoded Feature Row ingested by an older version of Feast.
+   * Check if encoded feature row v1 is encoded. A Feature Row v1 is considered encoded if both it'
```
woop, line 44: Does the concept of a Feature Row V1 extend beyond the Redis IO?

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/FeatureRowDecoder.java
```diff
@@ -36,60 +41,100 @@ public FeatureRowDecoder(String featureSetRef, FeatureSetSpec spec) {
...
+   * Check if encoded feature row v2 is encoded. The Feature Row v2 encoding defines that a Feature
+   * Row is f is considered encoded if it's both it feature set reference and fields names are set.
    *
    * @param featureRow Feature row
    * @return boolean
    */
-  public boolean isEncodingValid(FeatureRow featureRow) {
-    return featureRow.getFieldsList().size() == spec.getFeaturesList().size();
+  private boolean isEncodedV2(FeatureRow featureRow) {
+    return featureRow.getFieldsList().stream().allMatch(field -> !field.getName().isEmpty());
```
woop, line 67: From a performance perspective, would it make sense to rather use `anyMatch()` or test the first field?

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/FeatureRowDecoder.java
```diff
@@ -36,60 +41,100 @@ public FeatureRowDecoder(String featureSetRef, FeatureSetSpec spec) {
...
+              .collect(Collectors.toList());
+
+      return encodedFeatureRow
+          .toBuilder()
+          .clearFields()
+          .setFeatureSet(featureSetRef)
+          .addAllFields(fields)
+          .build();
+
+    } else if (isEncodedV2(encodedFeatureRow)) {
```
woop, line 108: Do you need an `else` statement here? I think two `if` statements would be cleaner.

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/FeatureRowDecoder.java
mrzzy: Beyond `RedisCustomIO`and `FeatureRowDecoder` I am not aware of any such encoding storage space optimization in Feast.

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/FeatureRowDecoder.java
```diff
@@ -36,60 +41,100 @@ public FeatureRowDecoder(String featureSetRef, FeatureSetSpec spec) {
...
+
+      List<String> featureNames =
+          spec.getFeaturesList().stream().map(FeatureSpec::getName).collect(Collectors.toList());
+
+      List<Field> fields =
+          featureNames.stream()
+              .map(
+                  name -> {
+                    String nameHash =
+                        Hashing.murmur3_32().hashString(name, StandardCharsets.UTF_8).toString();
```
woop, line 124: What is the length of the hash? Just want to make sure its as small as possible.

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/FeatureRowDecoder.java
```diff
@@ -36,60 +41,100 @@ public FeatureRowDecoder(String featureSetRef, FeatureSetSpec spec) {
...
+                    Value value =
+                        nameHashValueMap.getOrDefault(nameHash, Value.newBuilder().build());
+                    return Field.newBuilder().setName(name).setValue(value).build();
+                  })
+              .collect(Collectors.toList());
+
+      return encodedFeatureRow
```
woop, line 131: Hmmm, isn't there also a timestamp that is being lost here?

tests/e2e/redis/basic-ingest-redis-serving.py
```diff
@@ -675,6 +675,90 @@ def try_get_features():
...
+@pytest.mark.timeout(600)
+@pytest.mark.run(order=18)
+def test_basic_retrieve_feature_row_missing_fields(client, cust_trans_df):
+    # reset the project to default
+    client.set_project()
+
+    feature_refs = ["daily_transactions", "total_transactions", "null_values"]
+    # update cust_trans_fs with one additional feature.
+    # feature rows ingested before the feature set update will be missing a field.
+    old_cust_trans_fs = client.get_feature_set(name="customer_transactions")
```
woop, line 687: Our tests are becoming more and more brittle over time if we have these expectations of feature sets being registered already. One change upstream breaks multiple tests. Over time I want to make them a bit more atomic, which would also allow for higher parallelism.

Possible to add your own `apply` here for the first feature set?

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/FeatureRowDecoder.java
mrzzy: The code is simply overriding fields here. `encodedFeatureRow` is coverted to builder which means its values are preserved in builder form. Whatever that is not `set` is preserved from `encodedFeatureRow`.

mrzzy: 8 characters.

tests/e2e/redis/basic-ingest-redis-serving.py
mrzzy: Noted. Added.

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/FeatureRowDecoder.java
mrzzy: Updated

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/FeatureRowDecoder.java
mrzzy: Updated

mrzzy: Changed to `anyMatch()`

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/FeatureRowDecoder.java
woop: I mean I am asking because it seems like the versioning is on the encoder/decoder and not the feature row. It's a bit ambiguous as written, since there isn't a mention of the feature row versions elsewhere.

mrzzy: The versions are introduced to disambiguate between the different feature row encodings. We can call it legacy/new if need be.

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/FeatureRowDecoder.java
mrzzy: Assuming that the fields stored in the feature row are float values (32 bit, 4 bytes), this would mean a ~3x increase in space consumption.
@woop @pyalex @khorshuheng

pyalex: Not all fields that are stored are float, there're a lot string as well and int64. So everything is not so bad

pyalex: But I guess that we can safely cut hash string to 4-5 chars

PR #913, opened 2020-08-01
Redis sink flushes only rows that have more recent eventTimestamp

pyalex:

Since we keep only one row per entity in Redis storage previously it was unpredictable which row will be written, especially in batch ingestion. In this PR we read existing value from redis first to compare eventTimestamps and only if row from stream has more recent eventTimestamp it will be written to store. This guarantees that more recent row won't be overwritten with older value. Now redis will have the latest value for entity.

Also, I refactored tests for redis sink to remove duplicated parts of code and use `@Parametrized` tests instead.

Fixes #

```release-note
Redis sink has now more predictable order of writes. Value is being overwritten only by more recent row from stream.
```


storage/connectors/redis/src/main/java/feast/storage/connectors/redis/writer/BatchDoFnWithRedis.java
```diff
@@ -0,0 +1,88 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Base class for redis-related DoFns. Assumes that operations will be batched. Prepares redisClient
+ * on DoFn.Setup stage and close it on DoFn.Teardown stage.
+ *
+ * @param <Input>
+ * @param <Output>
+ */
+public class BatchDoFnWithRedis<Input, Output> extends DoFn<Input, Output> {
```
mrzzy, line 35: There seems to be only one extension of this base class (`RedisCustomIO.WriteDoFn`). Is this base class necessary?

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/writer/RedisCustomIO.java
```diff
@@ -250,35 +211,101 @@ private FailedElement toFailedElement(
...
 
         try {
-          executeBatch(featureRows, latestSpecs);
-          featureRows.forEach(row -> context.output(successfulInsertsTag, row));
+          executeBatch(
+              (redisIngestionClient) ->
+                  deduplicatedRows.entrySet().stream()
+                      .map(
+                          entry ->
+                              redisIngestionClient
```
woop, line 232: As an eventual optimization, would it make sense to cache the value that is returned from Redis? So that if we know that all values that we want to write is older than what is cached, we can then discard it all without even looking at Redis. Might be a premature optimization though.

tests/e2e/redis/basic-ingest-redis-serving.py
```diff
@@ -1273,6 +1273,58 @@ def get_running_jobs():
         time.sleep(1)
 
 
+@pytest.mark.run(order=18)
```
woop, line 1276: How are you selecting the order here? It doesn't seem to fit in with the rest of the tests. My original idea with the order was something like

10-19 (one set of tests), but maybe only 10-14 in use
20-29 (another set of tests
etc...

Currently it doesnt seem like the order is used consistently though

pyalex: I will move it. I just had top my test up, because after this
```
@pytest.mark.run(order=30)
def test_large_volume_register_feature_set_success(client):
```
nothing succeed. It fill queue for minutes ahead

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/writer/RedisCustomIO.java
pyalex: Within batch rows are already deduplicated and on this moment it's only one value per key. So each redis key being fetched only once.
Keep this cache as state between batches might be too expensive.

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/writer/BatchDoFnWithRedis.java
pyalex: Just wanted to separate redis wrapper code from business logic

tests/e2e/redis/basic-ingest-redis-serving.py
pyalex: done

PR #926, opened 2020-08-03
Impute default project if empty before authorization is called

jmelinav:
if `project` is not set in the `featureSet`, which is in the request for `applyFeatureSet` method, an empty project will be passed for authorization. Instead it should authorize against `default` project. 

Related to #793 

**Does this PR introduce a user-facing change?**:no

```release-note
In 'applyFeatureSet' api, project name imputation happens before invoking authorization service. 
Previous behavior, only 'admin' user could create feature sets in 'default' project, 
and authorization was failing for users who have access to 'default' project, 
since authorization was being invoked with empty 'projectId'. 
```

### DIFF
--- a/core/src/main/java/feast/core/grpc/CoreServiceImpl.java
+++ b/core/src/main/java/feast/core/grpc/CoreServiceImpl.java
@@ -30,6 +30,7 @@ import feast.core.service.SpecService;
 import feast.core.service.StatsService;
 import feast.proto.core.CoreServiceGrpc.CoreServiceImplBase;
 import feast.proto.core.CoreServiceProto.*;
+import feast.proto.core.FeatureSetProto.FeatureSet;
 import io.grpc.Status;
 import io.grpc.StatusRuntimeException;
 import io.grpc.stub.StreamObserver;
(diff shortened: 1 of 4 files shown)


core/src/main/java/feast/core/grpc/CoreServiceImpl.java
```diff
@@ -187,9 +188,10 @@ public void applyFeatureSet(
     String projectId = null;
 
     try {
-      projectId = request.getFeatureSet().getSpec().getProject();
+      FeatureSet featureSet = specService.imputeProjectName(request.getFeatureSet());
+      projectId = featureSet.getSpec().getProject();
```
woop, line 192: The PR makes sense, I just have a question on our long term plan for where this logic should live. Instead of pulling logic from the services up into the gRPC handler, would it make sense to push authorization down into the services like `specService`? That way we could keep the HTTP and gRPC handlers slim and deduplicated.

jmelinav: It makes sense in general but not sure if we should do that for authorization. for example, if an api in handler invokes multiple services or if services are nested, authorization would happen more than once. May be a facade below handlers will help in keeping them slim? However it will be better if impute is the 1st step in the process, that way down stream doesn't have to worry missing fields.

woop: > It makes sense in general but not sure if we should do that for authorization. for example, if an api in handler invokes multiple services or if services are nested, authorization would happen more than once. 

In general authorization is localized. I am not sure that it is possible to even have authorization up here in the gRPC service long term. For example if we want to take more granular actions as part of a method.

Do you have an example of authorization that would happen multiple times?

I mean a gRPC handler should always just call a single service and get a response. If we place the authorization at the top of that service (like specService), then both the REST and gRPC controllers/services will be able to use it.

jmelinav: If we are sure that only one service is invoked per api/request and services will not be nested, it might work. I don't know all the service and context around them in feast. The scenario I was thinking of was, let's say we add ingestion service in future, which uses SpecService to read the feature sets(may be to check if feature set is ready or it exists etc.) and job service to create a new ingestion job. In which case authorization may be invoked more than once if both service methods require authorization. 

You would know better if service will interact with each other or not.

woop: I'm going to let this merge in, but I think we push it down in the future.

woop: @jmelinav will you please add a kind and add a release note?

PR #930, opened 2020-08-05
Fix empty array when retrieving stats data

terryyylim:
These 2 methods, [this](https://github.com/feast-dev/feast/blob/master/core/src/main/java/feast/core/service/StatsService.java#L254) and [that](https://github.com/feast-dev/feast/blob/master/core/src/main/java/feast/core/service/StatsService.java#L185) will always be returning at least an empty arraylist. So essentially, when being called in [getFeatureStatistics](https://github.com/feast-dev/feast/blob/master/core/src/main/java/feast/core/service/StatsService.java#L92), these conditions, [this](https://github.com/feast-dev/feast/blob/master/core/src/main/java/feast/core/service/StatsService.java#L135) and [that](https://github.com/feast-dev/feast/blob/master/core/src/main/java/feast/core/service/StatsService.java#L151) will never be triggered, since length of arraylist will always at least be 1.

Fixes #

```release-note
N/A
```

### DIFF
--- a/core/src/main/java/feast/core/service/StatsService.java
+++ b/core/src/main/java/feast/core/service/StatsService.java
@@ -129,6 +129,9 @@ public class StatsService {
         List<FeatureNameStatistics> featureNameStatistics =
             getFeatureNameStatisticsByDate(
                 statisticsRetriever, featureSet, features, timestamp, request.getForceRefresh());
+        if (featureNameStatistics.size() != 0) {
+          featureNameStatisticsList.add(featureNameStatistics);
+        }
         featureNameStatisticsList.add(featureNameStatistics);
         timestamp += 86400; // advance by a day
       }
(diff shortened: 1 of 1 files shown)


core/src/main/java/feast/core/service/StatsService.java
```diff
@@ -129,6 +129,9 @@ public GetFeatureStatisticsResponse getFeatureStatistics(GetFeatureStatisticsReq
         List<FeatureNameStatistics> featureNameStatistics =
             getFeatureNameStatisticsByDate(
                 statisticsRetriever, featureSet, features, timestamp, request.getForceRefresh());
+        if (featureNameStatistics.size() != 0) {
+          featureNameStatisticsList.add(featureNameStatistics);
```
woop, line 133: Why isnt this in `getFeatureNameStatisticsByDate`?

terryyylim: Would you mean returning null if the size is 0 in `getFeatureNameStatisticsByDate`?

PR #935, opened 2020-08-05
Fix grpc security variables name and missing exec qualifier in docker.dev

jmelinav:
Grpc security variable names in application.yml are incorrect. Also exec qualifiers are missing from DockerFile.dev

Related #823 

```release-note
previous behavior : enabling grpc security and setting valid server.crt and server.key, throws an error. 
Also docker images build using DockerFile.dev will not boot.

With this PR, enabling security shouldn't throw any error if valid server.crt and server.key are set.
Images build using DockerFile.dev will boot without any errors.
```

### DIFF
--- a/core/src/main/resources/application.yml
+++ b/core/src/main/resources/application.yml
@@ -113,8 +113,8 @@ grpc:
     port: 6565
     security:
       enabled: false
-      certificateChainPath: server.crt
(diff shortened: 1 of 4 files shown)


woop:
Remember to add a release note to bugs, thanks for the contribution @jmelinav!

PR #938, opened 2020-08-07
Fix Audit Message Logging Interceptor Race Condition

mrzzy:
**Why we need it**:
Under load the `GrpcMessageInterceptor` will throw `IllegalStateException` due to a race condition:
- `GrpcMessageInterceptor` expects that `onMessage()`, `onHalfClose()`, `close()` are called in that order.
- Under load this does not hold and we observe that `close()` is being called before `onMessage()`, causing the `MessageAuditLogEntry` to be missing `request`, causing `IllegalStateException` when  `.build()` is called.

**What this PR does**:
Fixes  `GrpcMessageInterceptor`'s race condition by allowing `closed()` to be called before `onMessage()`
- defaults message to an empty protobuf if not set by `onMessage()`.

Adds an integration test `CoreLoggingIT` to verify that Message Audit Logs are produced correctly.

Fixes #

```release-note
NONE
```


common/src/main/java/feast/common/interceptors/GrpcMessageInterceptor.java
```diff
@@ -59,22 +55,25 @@
...
+    // As sendMessage(), close() and onMessage() may not be called in order (async)
+    // add a try log message call to each so that message logging would not
+    // depend on the calls to made in a specific order.
     call =
         new SimpleForwardingServerCall<ReqT, RespT>(call) {
           @Override
           public void sendMessage(RespT message) {
-            // 2. Track the response & Log entry to audit logger
+            // Track the response & Log entry to audit logger
             super.sendMessage(message);
```
woop, line 66: Should this `super.sendMessage(message);` not be the last call that is made?

mrzzy: This was done to ensure that logging code, as much as possible stays out of the way of the rest of message (ie exceptions in logging code do not disrupt message handling.) Going to try as last call to see if it fixes the concurrency issue.

mrzzy: From my testing, moving the location of statements does not resolve the issue.

core/src/test/java/feast/core/logging/CoreLoggingIT.java
```diff
@@ -0,0 +1,201 @@
...
+  public void shouldProduceMessageAuditLogsOnCall()
+      throws InterruptedException, InvalidProtocolBufferException {
+    // Generate artifical load on feast core.
+    UpdateStoreRequest request =
+        UpdateStoreRequest.newBuilder().setStore(DataGenerator.getDefaultStore()).build();
+    UpdateStoreResponse response = coreService.updateStore(request);
+
+    // Wait required to ensure audit logs are flushed into test audit log appender
+    Thread.sleep(1000);
+    // Check message audit logs are produced for each audit log.
```
woop, line 98: Thanks for throwing in some comments. Makes it easy to read :+1:

core/src/test/java/feast/core/logging/CoreLoggingIT.java
```diff
@@ -0,0 +1,201 @@
...
+
+    assertEquals(1, logJsonObjects.size());
+    JsonObject logJsonObject = logJsonObjects.get(0);
+    // Check correct status code is tracked on error.
+    assertEquals(logJsonObject.get("statusCode").getAsString(), statusCode.toString());
+  }
+
+  /** Check that expected message audit logs are produced when under load. */
+  @Test
+  public void shouldProduceExpectedNumberOfAuditLogs()
```
woop, line 156: I thought the point of the race was that it would always log something but would just have some missing data. Is the test you are doing here only to make sure something was logged? Seems like a good idea to make sure the contents itself is correct.

mrzzy: Updated the test to check  the request and response are correct.

woop (approved):

PR #943, opened 2020-08-11
JobCoordinator use public API to communicate with Core

pyalex:

This PR continues to decouple JobCoordinator from core. JC doesn't use repositories (JPA) from Core anymore and get/update feature sets through Spec API. All models objects were replaced with respected protobuf objects.

SpecService API was extended with:
1. Filter FeatureSets by Status
2. Update status of FeatureSet

Also, as part of test refactoring SpecService unit tests were converted into integration tests.

Fixes #

```release-note
User can filter feature sets by status via ListFeatureSets method (not implemented in SDK)
```


protos/feast/core/CoreService.proto
```diff
@@ -349,3 +354,10 @@ message GetFeatureStatisticsResponse {
     // despite the message being of list type.
     tensorflow.metadata.v0.DatasetFeatureStatisticsList dataset_feature_statistics_list = 1;
 }
+
+message UpdateFeatureSetStatusRequest {
```
woop, line 358: Need comments please.

protos/feast/core/CoreService.proto
```diff
@@ -150,6 +153,8 @@ message ListFeatureSetsRequest {
         // User defined metadata for feature set.
         // Feature sets with all matching labels will be returned.
         map<string,string> labels = 4;
+
+        FeatureSetStatus status = 5;
```
woop, line 157: Comment here as well please.

PR #946, opened 2020-08-12
Output Subject Claim as Identity in Logging interceptor

mrzzy:
- Moves `subjectClaim` option from `feast.authorization.options` to `feast.authentication.options`
  -  Since `subjectClaim` options is used by `GrpcMessageInterceptor` even when only authentication is enabled.
- Update `GrpcMessageInterceptor` to output subject claim as identity.
   - ie for the `google` authentication provider, this would output the users email as identity.
  - uses subject claim configed via `feast.authentication.options.subjectClaim`.
- Merges `feast-auth` module into `feast-common` module to prevent a circular dependency.
- Disables Spring Security autoconfiguration beans as security is not yet implemented for Job Controller.

Fixes #

```release-note
* Moved `subjectClaim` option from feast.authorization.options to feast.authentication.options
* Message Audit Logs will now output configured Subject Claim as Identity instead of Subject when feast.authentication.options.subjectClaim is set
```


auth/src/main/java/feast/auth/interceptors/GrpcMessageInterceptor.java
```diff
@@ -55,7 +67,14 @@
 
     // Attempt Extract current authenticated identity.
     Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
-    String identity = (authentication == null) ? "" : authentication.getName();
+    String identity = "";
+    if (authentication != null) {
+      System.out.println(securityProperties);
```
woop, line 72: Should this line be here?

mrzzy: Corrected.

auth/src/main/java/feast/auth/config/SecurityProperties.java
```diff
@@ -54,6 +53,7 @@
     private String provider;
 
     // K/V options to initialize the provider with
+    public static final String SUBJECT_CLAIM = "subjectClaim";
```
woop, line 56: Can we add a dedicated comment to this field?

mrzzy: Added.

auth/src/main/java/feast/auth/config/SecurityProperties.java
```diff
@@ -54,6 +52,8 @@
     private String provider;
 
     // K/V options to initialize the provider with
+    // Key for Subject Claim option which sets the name of the subject claim field in tokens.
```
woop, line 55: Should the comment above this one be below at the options?

job-controller/src/main/java/feast/jobcontroller/config/WebSecurityConfig.java
```diff
@@ -0,0 +1,53 @@
...
+  @Override
+  protected void configure(HttpSecurity http) throws Exception {
+    // Bypasses security/authentication for the following paths
+    http.authorizeRequests()
+        // TODO: Currently allows access to all endpoints as Security has not been implemented for
+        // JobController yet.
+        // When security is enabled, should only allow unauthenticated access to actuator and
+        // metrics endpoints.
+        .antMatchers("/")
+        // .antMatchers("/actuator/**", "/metrics/**")
```
woop, line 45: Does this still have to be so permissive?

pom.xml
```diff
@@ -731,6 +814,9 @@
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-javadoc-plugin</artifactId>
                     <version>3.1.1</version>
+                    <configuration>
+                        <doclint>all</doclint>
```
woop, line 818: What is this for?

mrzzy: Originally `doclint` is enabled via the [`maven-compiler-plugin` option here](https://github.com/feast-dev/feast/pull/946/files#diff-600376dffeb79835ede4a0b285078036L712). This enables `doclint` from `maven-javadoc-plugin` so that we can disable doclint for classpath of the OpenAPI client which fails `doclint`.

woop: But there is an exclusion. Why did this work in the past but not any more?

mrzzy: I'm am not entirely sure myself.
The effect of this move is that doclint moves from the compile maven lifecycle to the test maven lifecycle.
Doclint still works and flags issues with the docs.

job-controller/src/main/java/feast/jobcontroller/config/WebSecurityConfig.java
mrzzy: Opted for disabling Spring Security autoconfiguration beans instead of hardcoding endpoints to disable in `WebSecurityConfig`.

mrzzy:
/test  @feast-ci-bot
test-end-to-end-redis-cluster

woop (approved):

PR #947, opened 2020-08-13
JobCoordinator can be configured to use only subset of FeatureSets or Stores

pyalex:
This PR adds ability to configure via application properties which FeatureSets and Stores will be available in (and managed by) current instance of JobManager.

This should be useful when you have several deployments of JobManager and each will be responsible for only subset of FeatureSets.

Fixes #

```release-note
# application.yaml
feast:
  jobs:
    coordinator:
      # Specify feature sets that should be handled by current instance of JobManager
      featureSetSelector:
        - project: "*"
          name: "*"
      # Stores names that won't be used by current instance of JobManager
      whitelisted-stores: [online, online_cluster, historical]
```

### DIFF
--- a/core/src/main/java/feast/core/config/FeastProperties.java
+++ b/core/src/main/java/feast/core/config/FeastProperties.java
@@ -22,6 +22,7 @@ import feast.auth.config.SecurityProperties.AuthorizationProperties;
 import feast.common.logging.config.LoggingProperties;
 import feast.common.validators.OneOfStrings;
 import feast.core.config.FeastProperties.StreamProperties.FeatureStreamOptions;
+import feast.proto.core.StoreProto;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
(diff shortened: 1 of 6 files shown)


core/src/main/java/feast/core/config/FeastProperties.java
```diff
@@ -114,6 +115,30 @@ LoggingProperties loggingProperties() {
 
       /* Labels to identify jobs managed by this job coordinator */
       private Map<String, String> jobSelector = new HashMap<>();
+
+      /* Selectors to define featureSets that are responsibility of current JobManager */
+      private List<Selector> featureSetSelector = new ArrayList<>();
+
+      /* Specify names of stores that must not be used by current JobManager */
+      private List<String> blacklistedStores = new ArrayList<>();
```
woop, line 123: Is it easier to go the reverse? stores that are whitelisted?

I can imagine that a JobCoordinator will be deployed only for a list of stores, and they will probably be hardcoded as part of deployment.

core/src/main/java/feast/core/config/FeastProperties.java
```diff
@@ -114,6 +115,30 @@ LoggingProperties loggingProperties() {
...
+
+      /**
+       * Similarly to Store's subscription this selector defines set of FeatureSets. All FeatureSets
+       * that match both project and name will be selected. Project and name may use *
+       */
+      @Getter
+      @Setter
+      public static class Selector {
```
woop, line 131: We should maybe make this `FeatureSelector` to differentiate it from other selectors.

core/src/main/java/feast/core/config/FeastProperties.java
pyalex: Use case I had in my mind was that we don't want to have two job coordinators writing to bigquery. So we turning off just one this store.
In reverse case we need to specify all stores names that are distributed across all serving configs.

pyalex: But right now it's not a lot of them. So I guess both options would work

woop: > Use case I had in my mind was that we don't want to have two job coordinators writing to bigquery. So we turning off just one this store.

Are you sure there isnt a use case like this? I think it might theoretically be useful during rollouts, even though its not a part of our plan now. There is also no downside to having duplicated rows if they are deduplicated at query time.

> In reverse case we need to specify all stores names that are distributed across all serving configs.

Yea, but understanding which stores to exclude is a lot harder. I know my store, but I dont want to keep track of other people's stores. So lets say we have 5-6 job coordinators across multiple VPC. They each have 1 store. Does this mean that they need to exclude all stores except their own?

PR #949, opened 2020-08-14
Restart Ingestion Job on code version update

pyalex:
1. Each Job can have individual labels that are being merged with common labels (see `feast.jobs.runners.options.labels`) before job created. Job labels are being generated & managed by `JobCoordinator`. Currently labels consists of `feast.job.coordinator.jobSelector` and feast version. Job labels overwrite common labels for the same keys.
 
2. Added label `feast_version` that keeps version of feast code attached to the job. As soon as new version is being released - JobCoordinator automatically restarts jobs with older version in this label.

Fixes #

```release-note

```

### DIFF
--- a/core/src/main/java/feast/core/service/JobCoordinatorService.java
+++ b/core/src/main/java/feast/core/service/JobCoordinatorService.java
@@ -59,7 +59,7 @@ import org.springframework.stereotype.Service;
 public class JobCoordinatorService {
 
   private final int SPEC_PUBLISHING_TIMEOUT_SECONDS = 5;
-  public static final String VERSION_LABEL = "FeastVersion";
+  public static final String VERSION_LABEL = "feast_version";
 
   private final JobRepository jobRepository;
   private final SpecService specService;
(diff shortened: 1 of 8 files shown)


core/src/main/java/feast/core/service/JobCoordinatorService.java
```diff
@@ -81,6 +85,14 @@ public JobCoordinatorService(
...
     this.specPublisher = specPublisher;
     this.groupingStrategy = groupingStrategy;
+    this.featureSetSubscriptions =
+        feastProperties.getJobs().getCoordinator().getFeatureSetSelector().stream()
+            .map(JobProperties.CoordinatorProperties.Selector::toSubscription)
+            .collect(Collectors.toList());
+    this.blacklistedStores = feastProperties.getJobs().getCoordinator().getBlacklistedStores();
+    this.currentVersion = feastProperties.getVersion();
+    this.jobLabels = new HashMap<>(feastProperties.getJobs().getCoordinator().getJobSelector());
+    this.jobLabels.put("version", this.currentVersion);
```
woop, line 95: should `this.currentVersion` be a getter?

pyalex: this is private final property, not sure how getter could be useful?

PR #953, opened 2020-08-21
Fix flaky tests

pyalex:
Some refactoring / bug fixes to make tests more stable:

1. Use testcontainers in ingestion unit-tests
2. Better isolation for e2e tests
3. Fix rounding error in timestamp for batch tests
4. Fix Segmentation Fault occurring in Python SDK
5. Dynamic port allocation in python SDK unit tests

Fixes #

```release-note

```

### DIFF
--- a/sdk/python/tests/test_client.py
+++ b/sdk/python/tests/test_client.py
@@ -156,7 +156,7 @@ class TestClient:
         server = grpc.server(futures.ThreadPoolExecutor(max_workers=10))
         Core.add_CoreServiceServicer_to_server(CoreServicer(), server)
         port = find_free_port()
-        server.add_insecure_port("[::]:{}".format(port))
+        server.add_insecure_port(f"[::]:{port}")
         server.start()
         yield port
         server.stop(0)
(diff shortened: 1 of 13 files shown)


sdk/python/tests/test_client.py
```diff
@@ -187,9 +198,10 @@ def secure_core_server_with_auth(self, server_credentials):
             interceptors=(AllowAuthInterceptor(),),
         )
         Core.add_CoreServiceServicer_to_server(CoreServicer(), server)
-        server.add_secure_port("[::]:50055", server_credentials)
+        port = find_free_port()
+        server.add_secure_port("[::]:{}".format(port), server_credentials)
```
woop, line 202: Can we use f-strings? I quite like how they read.

tests/e2e/bq/bq-batch-retrieval.py
```diff
@@ -186,7 +186,7 @@ def test_batch_get_historical_features_with_file(client):
     client.ingest(file_fs1, features_1_df, timeout=480)
 
     # Rename column (datetime -> event_timestamp)
-    features_1_df["datetime"] + pd.Timedelta(
+    features_1_df["datetime"] += pd.Timedelta(
```
woop, line 189: I know you didn't add this buffer, but isnt it counter-intuitive that it needs to be added? Shouldn't the tests be representative of how users will be using the system?

woop (approved):

Issue #555, opened 2020-03-20
No TTL on Redis keys


idahoakl:
## Expected Behavior 
TTL set on values stored within Redis.

## Current Behavior
No TTL is set on values stored within Redis.  For FeatureSets that have short lived Entity values this leads to Redis filling with old data that is never referenced.  Once Redis is full it either slows down while trying to expire keys or stops ingesting new data.

## Steps to reproduce
1.  Install Feast with an online serving component backed by Redis
1.  Create any FeatureSet
1.  Ingest data into the FeatureSet
1.  Check the TTL on a key within Redis, TTL value of -1 

### Specifications

- Version: 0.4.6
- Platform: 
- Subsystem: ingestion

## Possible Solution
Potentially set the Redis key TTL to be the max_age on the FeatureSet.

woop:
This has come up in the past as well as functionality that we should introduce. I am more included to label this as a feature request though, as opposed to a bug. 

I think setting it to max_age makes sense, but for a lot of folks that parameter is still quite vaguely defined. We might be evicting keys quite aggressively if they max_age is too low, but I guess that is up to the user to configure correct then.

idahoakl:
I debated between a feature request vs a bug.  I decided on a bug given the fact that if the data is older than max_age it can't be retrieved from online serving but there is no easy way to remove it from taking up space within Redis.

I think adding an option to ingest to set a TTL (based on max_age or static) or not would be good to have.  There may be use cases where using on of Redis' maxmemory eviction policies is the desired behavior.  In that case setting no TTL would remain available.

ches:
FWIW we implemented max age as the TTL for the Cassandra serving implementation that is in v0.3 only currently. There is no such general LRU expulsion mode functionality in it and other potential serving store backends, so it’s essential to choose some strategy.

We do expect to have to police that max age cannot be set “unreasonably low”, and may add some constant factor like one day to the TTL from the max age, to mitigate pressure on the storage engine without changing a semantic requirement expected by the user.

zhilingc:
> I debated between a feature request vs a bug. I decided on a bug given the fact that if the data is older than max_age it can't be retrieved from online serving but there is no easy way to remove it from taking up space within Redis.
> 
> I think adding an option to ingest to set a TTL (based on max_age or static) or not would be good to have. There may be use cases where using on of Redis' maxmemory eviction policies is the desired behavior. In that case setting no TTL would remain available.

[...]

PR #955, opened 2020-08-21
Set TTL on Redis values

ravdin:
# Background

Currently no TTL is set on values stored in Redis. This results in the database storing stale values that are never referenced.

# Solution

This change allows an import job to be configured to set the TTL on Redis values based on the FeatureSet's max age. If the TTL is less than or equal to zero, the value will not be written to Redis. The configuration settings are:

* enableRedisTTL: Set to true to enable setting the Redis TTL. Defaults to false.
* maxRedisTTLJitterSeconds: Adds a random value to the TTL, up to this value. Defaults to 0.

https://github.com/feast-dev/feast/issues/555


ingestion/src/main/java/feast/ingestion/ImportJob.java
```diff
@@ -152,7 +152,8 @@ public static PipelineResult runPipeline(ImportOptions options) throws IOExcepti
     PCollectionList<FeatureSetReference> sinkReadiness = PCollectionList.empty(pipeline);
 
     for (Store store : stores) {
-      FeatureSink featureSink = getFeatureSink(store);
+      FeatureSink featureSink =
+          getFeatureSink(store, options.getEnableRedisTTL(), options.getMaxRedisTTLJitterSeconds());
```
pyalex, line 156: we have sink options grouped per specific type (see `protos/feast/core/Store.proto`) and this one is generic API. Those options should be passed via `RedisConfig` & `RedisClusterConfig`

woop: @ravdin To restate what Oleksii said, you should be putting this configuration here: https://github.com/feast-dev/feast/blob/master/protos/feast/core/Store.proto#L108

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/writer/RedisCustomIO.java
```diff
@@ -307,6 +375,49 @@ boolean rowShouldBeWritten(FeatureRow newRow, byte[] currentValue) {
...
+          } else {
+            skippedFeatureRows.add(newRow);
+            return false;
+          }
+        } else {
+          // Log an error instead of throwing an exception since
+          // setting a TTL is an
+          // optimization
+          // for storage space usage.
+          log.error("Unable to find FeatureSet to set Redis TTL  featureSet={}", spec.getName());
```
woop, line 416: Does this mean we will always log errors for users that have feature sets with default max age of 0?

woop:
Thanks for the contribution @ravdin 

So our summary on the conversation from the community call
1. Max age for TTL makes sense
2. You need to move the Redis specific config into the Redis or RedisCluster store configuration protos
3. We should have a maximum TTL. If the user does not have a valid max age, or if the max age is larger than the maximum, then we use the maximum. This is also configured in the Redis configuration.

Does that make sense?

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/writer/RedisCustomIO.java
ravdin: No, it shouldn't. The intention is that we'll log if no max age is set at all.

ingestion/src/main/java/feast/ingestion/ImportJob.java
ravdin: Done.

ravdin:
@woop @pyalex I added a maximum Redis TTL and two unit tests (test that we're using the maximum if there is no max age and confirm that we're overriding the max age with the configured maximum). If the maximum is set to 0 (i.e. not set), then the behavior will be as if there is no maximum.

Can you confirm that we want to override the feature set max age with this maximum if the max age is larger? I'm not sure that I see the use case for this.

protos/feast/core/Store.proto
```diff
@@ -134,6 +140,12 @@ message Store {
     int32 max_retries = 3;
     // Optional. How often flush data to redis
     int32 flush_frequency_seconds = 4;
+    // Set the Redis key TTL based on the max age of the FeatureSet that data is being written for.
+    bool enable_redis_ttl = 5;
+    // Optional. Set the maximum amount of jitter in seconds added to Redis TTL. This has no effect if Redis TLL is disabled.
+    int32 max_redis_ttl_jitter_seconds = 6;
+    // Optional. Set the maximum Redis TTL. This has no effect if Redis TTL is disabled.
```
woop, line 147: maybe we should specify that this is in seconds?

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/writer/RedisCustomIO.java
woop: Not setting a max age is actually very common. What if we only log a warning here, and we only log if neither a max age has been configured on the feature set or as part of the redis config (maximum max age).

I think as administrators we must be able to set the expiration, even if users dont set the max age.

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/writer/RedisCustomIO.java
```diff
@@ -307,6 +356,56 @@ boolean rowShouldBeWritten(FeatureRow newRow, byte[] currentValue) {
...
+            keysWithTtl.put(key, ttlSeconds);
+          } else {
+            skippedFeatureRows.add(newRow);
+            return false;
+          }
+        } else {
+          if (maxRedisTtl > 0) {
+            keysWithTtl.put(key, maxRedisTtl);
+          }
+          // Log an error instead of throwing an exception since
```
woop, line 400: Would you mind fixing the comments. Not sure why its broken up.

woop:
> @woop @pyalex I added a maximum Redis TTL and two unit tests (test that we're using the maximum if there is no max age and confirm that we're overriding the max age with the configured maximum). If the maximum is set to 0 (i.e. not set), then the behavior will be as if there is no maximum.
> 
Looks good overall!


> Can you confirm that we want to override the feature set max age with this maximum if the max age is larger? I'm not sure that I see the use case for this.

I dont see a need to make changes to the feature set, but I do see a need to change the TTL/expiry in Redis. We shouldnt allow users to set 9999999999 as a max age, just so that they can store their keys forever. An admin should be able to control expiry.

idahoakl:
> I dont see a need to make changes to the feature set, but I do see a need to change the TTL/expiry in Redis. We shouldnt allow users to set 9999999999 as a max age, just so that they can store their keys forever. An admin should be able to control expiry

I see the reason to allow an admin to be able to set an expiration to override to control data retention but setting the override retention at a specific store seems like a poor user experience since the user wouldn't be aware of the store specific setting when store or retrieving their data.  

For example if a user sets a max age on a FeatureSet of 30 days the expectation would be that any data stored with a timestamp of less than 30 days old would be retrievable from Feast Serving.  If the admin sets a maximum of 10 days then data stored with a timestamp of 20 days old would be expired yet according to what the max age that a user has specified the data should be available.  The problem would be more confusing if there are multiple stores that subscribe to the same FeatureSets but have different maximums.  One store would have the data while the other would not.

My thought is that if there is a desire to control a max retention period for a FeatureSet that setting should be moved further upstream so that users are aware of the restriction.  This is probably out of scope for this PR though.

woop:
> > I dont see a need to make changes to the feature set, but I do see a need to change the TTL/expiry in Redis. We shouldnt allow users to set 9999999999 as a max age, just so that they can store their keys forever. An admin should be able to control expiry
> 
> I see the reason to allow an admin to be able to set an expiration to override to control data retention but setting the override retention at a specific store seems like a poor user experience since the user wouldn't be aware of the store specific setting when store or retrieving their data.
> 
> For example if a user sets a max age on a FeatureSet of 30 days the expectation would be that any data stored with a timestamp of less than 30 days old would be retrievable from Feast Serving. If the admin sets a maximum of 10 days then data stored with a timestamp of 20 days old would be expired yet according to what the max age that a user has specified the data should be available. The problem would be more confusing if there are multiple stores that subscribe to the same FeatureSets but have different maximums. One store would have the data while the other would not.
> 
> My thought is that if there is a desire to control a max retention period for a FeatureSet that setting should be moved further upstream so that users are aware of the restriction. This is probably out of scope for this PR though.

I agree on your points, but I also somewhat disagree. I mean a counterpoint: The team provisioning a Redis Cluster are the ones sizing it and probably setting an LRU expiry for when a maximum memory limit is hit. That is already an implicit eviction that the user has no control over and is invisible to them.

So the way I am reasoning about this TTL at the store level is that its a different place to configure the absolute maximum TTL that a store should have. It's the same as going to the store and setting some default TTL, or sizing it to a specific size of memory or nodes. 

I think the idea of exposing TTL control to users is a mistake, but I dont really have a better idea yet. It's a leaky abstraction, so I am hoping we can move the other way where the user doesnt have to think about it. Ultimately what we want is to be able to have some form of guarantee that their data will be available, but TTL/MaxAge isnt enough for that to be guaranteed.

ravdin:
@woop I pushed the changes that you suggested.

woop:
@ravdin thanks, would you mind just resolving the conflicts so that we can get this merged in?

ravdin:
@woop I resolved the merge conflicts.

ingestion/src/main/java/feast/ingestion/options/ImportOptions.java
```diff
@@ -119,4 +119,18 @@
   int getWindowSizeInSecForFeatureValueMetric();
 
   void setWindowSizeInSecForFeatureValueMetric(int seconds);
+
+  @Description(
+      "Set the Redis key TTL based on the max age of the FeatureSet that data is being written for.")
+  @Default.Boolean(false)
+  boolean getEnableRedisTTL();
```
pyalex, line 126: this changes can be reverted, since configuration moved to `RedisClusterConfig`

ravdin: @pyalex Done.

storage/connectors/redis/src/main/java/feast/storage/connectors/redis/writer/RedisCustomIO.java
```diff
@@ -242,7 +266,22 @@ public void processElement(ProcessContext context) {
                                   .thenAccept(
                                       currentValue -> {
                                         FeatureRow newRow = entry.getValue();
+                                        FeatureSetSpec spec =
+                                            latestSpecs.get(newRow.getFeatureSet());
                                         if (rowShouldBeWritten(newRow, currentValue)) {
+                                          if (redisIngestionClient.getEnableRedisTtl()) {
```
pyalex, line 272: I don't understand why we need here `setRedisTtl` check. and how it affect our decision write row to redis or not?

pyalex: Does it mean that feature sets without max-age won't be written to redis?

pyalex: I think this part should be to line 298 `if (keysWithTtl.containsKey(key))`, bc here it clearly creates a bug:
1. this lambda function is being called from multiple threads, and `keysWithTtl` is not thread safe
2. if row don't pass `setRedisTtl` it's not being written at all (unit test needed)

pyalex: Ok, I get it. In this part we checking whether `eventTimestamp + ttl` < currentTime. Thus if row is outdated already, we won't write it at all. But implementation is very confusing and `setRedisTtl` interface and name as well.
Also this function is being called from many threads. Are you sure `keysWithTtl ` is thread safe?

pyalex: OK, to prevent method that modifies its arguments I suggest next
```
List<Triple<RedisKey, FeatureRow, lon>> rowsWithTtl = 
   deduplicatedRows.entrySet.stream()
        .map((key, row) -> 
           Triple(key, row, calculateRowTtl(row, latestSpecs.get(row.getFeatureSet()), redisIngestionClient.getMaxRedisTtlJitterSeconds(), redisIngestionClient.getMaxRedisTtlSeconds())
        .filter(triple -> triple._3 > 0).collect(...)

....
long calculateRowTtl(FeatureRow, FeatureSetSpec spec, int maxRedisTtlJitterSeconds, long maxRedisTtl)

Comment (pyalex) [reply storage/connectors/redis/src/main/java/feast/storage/connectors/redis/writer/RedisCustomIO.java:272]:
```
skippedFeatureRow = deduplicatedRows.filter(not in rowsWithTtl)
```

ravdin: Got it, thank you for the feedback. I'll address this in the next couple of days.

ravdin: @pyalex I made a few changes to avoid the call by reference issues you brought up. You're correct that we wanted to avoid writes if they are unnecessary (i.e. if the TTL <= 0). We thought it would be helpful to log when this happens, which is why we're keeping track of the rows that are skipped for this reason. We also need to distinguish where we are not using a TTL (either because TTL has not been enabled or the feature row does not have a max age we can use).

I created a `calculateRowTtl` function as you suggested that returns the row TTL or null if TTL should not be used. If the TTL is not a positive number, then the caller is responsible for skipping the write. Hopefully this makes sense!

pyalex: @radvin I agree that we need to log skipped row. Visibility is always good. My question is: do we really need to read from redis before skip row with `ttl < 0`?

My feeling is that independently of what stored in redis we won't want to write outdated (`ttl < 0`) value. This means that we can filtered out those rows earlier and make read batch smaller.

Also I still concerned about this line
```keysWithTtl.put(entry.getLeft(), ttl);```
This is inside lambda function, that is being called by lettuce in potentially multi-threaded executor and `keysWithTtl` is not thread-afe.

ravdin: @pyalex I made these changes as suggested:

1. We're no longer reading from redis if the TTL is calculated to be <= 0. This means that these rows will be ignored regardless of what is already in the database.
2. I made the `keysWithTtl` hashmap thread safe by moving writes outside the lambda function.

pyalex (approved):

ravdin:
@pyalex I did a rebase on the branch (no other changes though). Do you have timing on when you're looking to merge?

pyalex:
@ravdin can you please fix this DCO check? I guess this is only thing which blocks merge

ravdin:
@pyalex Done.

woop:
@ravdin can we retarget this to 0.7 please?

woop:
Closing this PR for the time being. @ravdin please let me know if we should reopen it or target a different branch/version.

PR #961, opened 2020-08-24
Add request response logging via fluentd

terryyylim:
This PR introduces the option to enable request/response logging to an external fluentd service. This is an addition to the existing Audit logging capabilities which logs to the console.

N/A

```release-note
Users can now enable forwarding of logs to an external fluentd service that's been setup.
```

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -158,6 +158,11 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.fluentd</groupId>
+            <artifactId>fluent-logger</artifactId>
+            <version>0.3.1</version>
+        </dependency>
 
         <!-- Bean Validation API support -->
         <dependency>
--- a/common/src/main/java/feast/common/logging/AuditLogger.java
+++ b/common/src/main/java/feast/common/logging/AuditLogger.java
@@ -26,6 +26,7 @@ import feast.common.logging.entry.LogResource;
 import feast.common.logging.entry.LogResource.ResourceType;
 import feast.common.logging.entry.MessageAuditLogEntry;
 import feast.common.logging.entry.TransitionAuditLogEntry;
+import io.grpc.Status;
 import java.util.HashMap;
 import java.util.Map;
 import lombok.extern.slf4j.Slf4j;
@@ -41,7 +42,8 @@ import org.springframework.stereotype.Component;
 @Slf4j
 @Component
 public class AuditLogger {
-  private static final String destination = "fluentd";
+  private static final String DESTINATION = "fluentd";
+  private static final String DEFAULT_RELEASE_NAME = "feast-0-0";
   private static final Marker AUDIT_MARKER = MarkerFactory.getMarker("AUDIT_MARK");
   private static FluentLogger fluentLogger;
   private static AuditLogProperties properties;
@@ -128,31 +130,33 @@ public class AuditLogger {
       return;
     }
     if (entry.getKind().equals(AuditLogEntryKind.MESSAGE)
+        && properties.getMessageLogging() != null
         && !properties.getMessageLogging().isEnabled()) {
       return;
     }
 
     // Either forward log to logging layer or log to console
-    if (properties.getMessageLogging().getDestination().equals(destination)) {
+    if (properties.getMessageLogging().getDestination().equals(DESTINATION)) {
       Map<String, Object> fluentdLogs = new HashMap<>();
-      String releaseTag = StringUtils.defaultIfEmpty(System.getenv("RELEASE_TAG"), "default_tag");
+      String releaseName =
+          StringUtils.defaultIfEmpty(System.getenv("RELEASE_NAME"), DEFAULT_RELEASE_NAME);
       MessageAuditLogEntry messageAuditLogEntry = (MessageAuditLogEntry) entry;
 
       fluentdLogs.put("id", messageAuditLogEntry.getId());
       fluentdLogs.put("service", messageAuditLogEntry.getService());
       fluentdLogs.put("status_code", messageAuditLogEntry.getStatusCode());
       fluentdLogs.put("method", messageAuditLogEntry.getMethod());
-      fluentdLogs.put("release_tag", releaseTag);
+      fluentdLogs.put("release_name", releaseName);
       try {
         fluentdLogs.put("request", JsonFormat.printer().print(messageAuditLogEntry.getRequest()));
         fluentdLogs.put("response", JsonFormat.printer().print(messageAuditLogEntry.getResponse()));
       } catch (InvalidProtocolBufferException e) {
-        throw io.grpc.Status.NOT_FOUND
-            .withDescription("Unable to forward logs to logging service")
+        throw Status.INTERNAL
+            .withDescription(
+                "Request/Response log conversion to JSON failed. Unable to forward logs to logging service.")
             .withCause(e)
             .asRuntimeException();
       }
-      // tag can be used as "service" to split rr logs into separate tables
       fluentLogger.log("fluentd", fluentdLogs);
     } else {
       // Log event to audit log through enabled formats
--- a/common/src/main/java/feast/common/logging/AuditLogger.java
+++ b/common/src/main/java/feast/common/logging/AuditLogger.java
@@ -18,13 +18,17 @@ package feast.common.logging;
 
 import feast.common.logging.config.LoggingProperties;
 import feast.common.logging.config.LoggingProperties.AuditLogProperties;
+import feast.common.logging.config.LoggingProperties.LogForwarderProperties;
 import feast.common.logging.entry.ActionAuditLogEntry;
 import feast.common.logging.entry.AuditLogEntry;
 import feast.common.logging.entry.LogResource;
 import feast.common.logging.entry.LogResource.ResourceType;
 import feast.common.logging.entry.MessageAuditLogEntry;
 import feast.common.logging.entry.TransitionAuditLogEntry;
+import java.util.HashMap;
+import java.util.Map;
 import lombok.extern.slf4j.Slf4j;
+import org.fluentd.logger.FluentLogger;
 import org.slf4j.Marker;
 import org.slf4j.MarkerFactory;
 import org.slf4j.event.Level;
@@ -36,8 +40,10 @@ import org.springframework.stereotype.Component;
 @Component
 public class AuditLogger {
   private static final Marker AUDIT_MARKER = MarkerFactory.getMarker("AUDIT_MARK");
+  private static FluentLogger FLUENT_LOGGER;
   private static AuditLogProperties properties;
   private static BuildProperties buildProperties;
+  private static LogForwarderProperties logForwarderProperties;
 
   @Autowired
   public AuditLogger(LoggingProperties loggingProperties, BuildProperties buildProperties) {
@@ -46,6 +52,7 @@ public class AuditLogger {
     // This allows us to use the dependencies in the AuditLogger's static methods
     AuditLogger.properties = loggingProperties.getAudit();
     AuditLogger.buildProperties = buildProperties;
+    AuditLogger.logForwarderProperties = loggingProperties.getForwarder();
   }
 
   /**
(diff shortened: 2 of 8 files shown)


common/pom.xml
```diff
@@ -114,5 +114,10 @@
             <version>4.0</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.fluentd</groupId>
```
woop, line 118: Do we really need to pull this into all modules? Can this be moved to serving only?

common/src/main/java/feast/common/logging/AuditLogger.java
```diff
@@ -37,8 +41,10 @@
 @Component
 public class AuditLogger {
   private static final Marker AUDIT_MARKER = MarkerFactory.getMarker("AUDIT_MARK");
+  private static FluentLogger FLUENT_LOGGER;
```
woop, line 44: Why is `FLUENT_LOGGER` capitalized?

common/src/main/java/feast/common/logging/AuditLogger.java
```diff
@@ -61,6 +68,23 @@ public static void logMessage(Level level, MessageAuditLogEntry.Builder entryBui
             .setComponent(buildProperties.getArtifact())
             .setVersion(buildProperties.getVersion())
             .build());
+    Map<String, Object> fluentDLogs = new HashMap<>();
```
woop, line 71: Nitpick, but I think I prefer `fluentdLogs` so that we dont have two capital letters after each other.

common/src/main/java/feast/common/logging/AuditLogger.java
```diff
@@ -61,6 +68,23 @@ public static void logMessage(Level level, MessageAuditLogEntry.Builder entryBui
...
+
+    if (logForwarderProperties.isEnabled()) {
+      FLUENT_LOGGER =
+          FluentLogger.getLogger(
```
woop, line 75: `To create Fluent Logger instances, users need to invoke getLogger method in FluentLogger class like org.slf4j, org.log4j logging libraries. The method should be called only once`

Does this mean you are creating a new logger every time? Why did you create a static instance?

common/src/main/java/feast/common/logging/AuditLogger.java
```diff
@@ -61,6 +68,23 @@ public static void logMessage(Level level, MessageAuditLogEntry.Builder entryBui
...
+              "feast",
+              logForwarderProperties.getFluentdHost(),
+              logForwarderProperties.getFluentdPort());
+      fluentDLogs.put("id", entryBuilder.build().getId());
+      fluentDLogs.put("service", entryBuilder.build().getService());
+      fluentDLogs.put("status_code", entryBuilder.build().getStatusCode());
+      fluentDLogs.put("method", entryBuilder.build().getMethod());
+      fluentDLogs.put("request", entryBuilder.build().getRequest());
```
woop, line 83: Request and response should be in Json, just like the existing audit logger. Otherwise its very hard to parse in BQ.

woop: https://github.com/feast-dev/feast/blob/21509be8717674575db6edb6f67378b40043e1e3/common/src/main/java/feast/common/logging/entry/MessageAuditLogEntry.java#L97

common/src/main/java/feast/common/logging/config/LoggingProperties.java
```diff
@@ -24,6 +24,20 @@
 @Setter
 public class LoggingProperties {
   @NotNull private AuditLogProperties audit;
+  @NotNull private LogForwarderProperties forwarder;
```
woop, line 27: `logForwarderProperties`

common/src/main/java/feast/common/logging/config/LoggingProperties.java
```diff
@@ -24,6 +24,20 @@
...
+
+  @Getter
+  @Setter
+  public static class LogForwarderProperties {
+    // Whether to enable/disable (request/response) log forwarding entirely.
```
woop, line 32: If you enable this, and disable audit logging, what happens?

common/pom.xml
terryyylim: AuditLogger is used in both Core and Serving, and it exists in common module, so I think that's not possible.

common/src/main/java/feast/common/logging/AuditLogger.java
terryyylim: Updated.

common/src/main/java/feast/common/logging/AuditLogger.java
terryyylim: Updated.

common/src/main/java/feast/common/logging/config/LoggingProperties.java
terryyylim: Updated.

common/src/main/java/feast/common/logging/AuditLogger.java
terryyylim: Updated, creating through constructor now.

common/src/main/java/feast/common/logging/interceptors/GrpcMessageInterceptor.java
```diff
@@ -86,6 +87,7 @@ public void sendMessage(RespT message) {
             entryBuilder.setResponse((Message) message);
           }
 
+          @SneakyThrows
```
woop, line 90: Please don't use this lombok annotation. We are only using Lombok for getters and setters.

core/src/main/resources/application.yml
```diff
@@ -51,6 +51,12 @@ feast:
     disableRestControllerAuth: false
 
   logging:
+    # Logging forwarder currently provides a machine readable structured JSON log to an
+    # external fluentd service that can give better insight into what is happening in Feast.
+    logForwarderProperties:
+      enabled: false
```
woop, line 57: What configuration is necessary in order to log using the forwarder but **NOT** log to the console?

common/src/main/java/feast/common/logging/AuditLogger.java
```diff
@@ -47,6 +53,13 @@ public AuditLogger(LoggingProperties loggingProperties, BuildProperties buildPro
     // This allows us to use the dependencies in the AuditLogger's static methods
     AuditLogger.properties = loggingProperties.getAudit();
     AuditLogger.buildProperties = buildProperties;
+    if (AuditLogger.properties.getMessageLogging().isEnabled()) {
```
pyalex, line 56: can `properties.getMessageLogging()` return null?

common/src/main/java/feast/common/logging/AuditLogger.java
```diff
@@ -113,28 +126,49 @@ private static void log(Level level, AuditLogEntry entry) {
...
-        break;
-      case WARN:
-        log.warn(AUDIT_MARKER, entryJSON);
-        break;
-      case ERROR:
-        log.error(AUDIT_MARKER, entryJSON);
-        break;
+    // Either forward log to logging layer or log to console
+    Map<String, Object> fluentdLogs = new HashMap<>();
+    if (properties.getMessageLogging().getDestination().equals("fluentd")) {
```
pyalex, line 135: maybe `fluentd` to constant?

common/src/main/java/feast/common/logging/AuditLogger.java
```diff
@@ -113,28 +126,49 @@ private static void log(Level level, AuditLogEntry entry) {
...
+      MessageAuditLogEntry messageAuditLogEntry = (MessageAuditLogEntry) entry;
+      try {
+        fluentdLogs.put("id", messageAuditLogEntry.getId());
```
pyalex, line 138: which one of this throws exception? do we need to put all of them into `try`?

common/src/main/java/feast/common/logging/AuditLogger.java
```diff
@@ -113,28 +126,49 @@ private static void log(Level level, AuditLogEntry entry) {
...
-        log.info(AUDIT_MARKER, entryJSON);
-        break;
-      case WARN:
-        log.warn(AUDIT_MARKER, entryJSON);
-        break;
-      case ERROR:
-        log.error(AUDIT_MARKER, entryJSON);
-        break;
+    // Either forward log to logging layer or log to console
+    Map<String, Object> fluentdLogs = new HashMap<>();
```
pyalex, line 134: I suggest to replace this with `ImmutableMap` since there's no need for mutability and code would contain less of `fluentdLogs.put`

common/src/main/java/feast/common/logging/AuditLogger.java
```diff
@@ -113,28 +130,52 @@ private static void log(Level level, AuditLogEntry entry) {
...
+      try {
+        fluentdLogs.put("request", JsonFormat.printer().print(messageAuditLogEntry.getRequest()));
+        fluentdLogs.put("response", JsonFormat.printer().print(messageAuditLogEntry.getResponse()));
+      } catch (InvalidProtocolBufferException e) {
+        throw io.grpc.Status.NOT_FOUND
+            .withDescription("Unable to forward logs to logging service")
+            .withCause(e)
+            .asRuntimeException();
+      }
+      // tag can be used as "service" to split rr logs into separate tables
```
pyalex, line 157: not sure this comment related here

common/src/main/java/feast/common/logging/AuditLogger.java
```diff
@@ -36,7 +42,10 @@
 @Slf4j
 @Component
 public class AuditLogger {
+  private static final String DESTINATION = "fluentd";
+  private static final String DEFAULT_NAME = "feast-0-0";
```
pyalex, line 46: probably `DEFAULT_RELEASE_NAME`

common/src/main/java/feast/common/logging/AuditLogger.java
```diff
@@ -113,28 +130,52 @@ private static void log(Level level, AuditLogEntry entry) {
...
+      fluentdLogs.put("id", messageAuditLogEntry.getId());
+      fluentdLogs.put("service", messageAuditLogEntry.getService());
+      fluentdLogs.put("status_code", messageAuditLogEntry.getStatusCode());
+      fluentdLogs.put("method", messageAuditLogEntry.getMethod());
+      fluentdLogs.put("release_name", releaseName);
+      try {
+        fluentdLogs.put("request", JsonFormat.printer().print(messageAuditLogEntry.getRequest()));
+        fluentdLogs.put("response", JsonFormat.printer().print(messageAuditLogEntry.getResponse()));
+      } catch (InvalidProtocolBufferException e) {
+        throw io.grpc.Status.NOT_FOUND
```
pyalex, line 153: NOT_FOUND and description are not really represent the cause of the exception that may happen here

common/src/main/java/feast/common/logging/AuditLogger.java
```diff
@@ -113,28 +130,52 @@ private static void log(Level level, AuditLogEntry entry) {
       return;
     }
     if (entry.getKind().equals(AuditLogEntryKind.MESSAGE)
-        && !properties.isMessageLoggingEnabled()) {
+        && !properties.getMessageLogging().isEnabled()) {
```
pyalex, line 133: similar issue with NPE as on line 59

common/src/main/java/feast/common/logging/AuditLogger.java
terryyylim: Updated description and status type.

common/src/main/java/feast/common/logging/config/LoggingProperties.java
```diff
@@ -31,7 +31,22 @@
...
+    private MessageLogging messageLogging;
+
+    @Getter
+    @Setter
+    public static class MessageLogging {
+      // Whether to enable/disable message level (ie request/response) audit logging.
+      private boolean enabled;
+
+      // Whether to log to console or fluentd
+      private String destination;
```
woop, line 43: Can we use the `OneOf` annotation here for safety?

common/src/main/java/feast/common/logging/AuditLogger.java
```diff
@@ -112,24 +129,47 @@ private static void log(Level level, AuditLogEntry entry) {
...
-        log.info(AUDIT_MARKER, entryJSON);
-        break;
-      case WARN:
-        log.warn(AUDIT_MARKER, entryJSON);
-        break;
-      case ERROR:
-        log.error(AUDIT_MARKER, entryJSON);
-        break;
+    // Either forward log to logging layer or log to console
+    if (properties.getMessageLogging().getDestination().equals(FLUENTD_DESTINATION)) {
```
woop, line 133: What prevents this code from executing if message Logging is disabled?

terryyylim: That's handled [here](https://github.com/feast-dev/feast/blob/master/common/src/main/java/feast/common/logging/interceptors/GrpcMessageInterceptor.java#L73) from #940.

woop: By the way, as a convention we try and instantiate variables from getters in order to make code more readable. So something like
```
String messageLogDestination = properties.getMessageLogging().getDestination();
if (message.equals(FLUENTD_DESTINATION)) {

}
```

woop (approved):

PR #962, opened 2020-08-25
Use grpc interceptor for open tracing and ensure that trace id is propagated

khorshuheng:
Feast Serving and Feast Golang SDK uses Jaeger tracing. However, the spans are collected in such a way that the propagation context are ignored. As a result, the trace from client and server side exists as separated, independent spans.

In addition, the tracer instance is not initialized properly in the Feast Golang SDK. The SDK users should have been allowed to pass an existing tracer instance to the client, so that the Feast spans can be connected with the existing traces from the client side.

This PR use third party interceptor libraries to capture the trace for both client and server, which takes care of the trace id propagation.

Tracing support has also been added to Java and Python SDK.

Fixes #

```release-note
Tracing support has been added to Feast SDK. Users are now able to pass existing instance of Open Tracing Tracer to Feast clients.

The span operation name will be different, as the interceptor library use a different naming convention than the name that was hard coded previously.
```

### DIFF
--- a/sdk/go/client.go
+++ b/sdk/go/client.go
@@ -8,6 +8,7 @@ import (
 	"github.com/feast-dev/feast/sdk/go/protos/feast/serving"
 	"google.golang.org/grpc"
 
+	"github.com/opentracing-contrib/go-grpc"
 	"go.opencensus.io/plugin/ocgrpc"
 )
 
@@ -24,12 +25,35 @@ type GrpcClient struct {
 	conn *grpc.ClientConn
 }
 
-// NewGrpcClient constructs a client that can interact via grpc with the feast serving instance at the given host:port.
-func NewGrpcClient(host string, port int) (*GrpcClient, error) {
-	feastCli := &GrpcClient{}
+// GrpcClientBuilder is used to construct the client with more advanced options.
+type GrpcClientBuilder struct {
+	host   string
+	port   int
+	tracer opentracing.Tracer
+}
 
-	adr := fmt.Sprintf("%s:%d", host, port)
-	conn, err := grpc.Dial(adr, grpc.WithStatsHandler(&ocgrpc.ClientHandler{}), grpc.WithInsecure())
+// WithTracer supplies an existing tracer instance to Feast client, so that the gRPC calls
+// to Feast Online Serving can be traced. If no tracer is supplied, a no ops tracer will
+// be used instead.
+func (builder *GrpcClientBuilder) WithTracer(tracer opentracing.Tracer) *GrpcClientBuilder {
+	builder.tracer = tracer
+	return builder
+}
+
+// Build build the client.
+func (builder *GrpcClientBuilder) Build() (*GrpcClient, error) {
+
+	feastCli := &GrpcClient{}
+	if builder.tracer == nil {
+		builder.tracer = opentracing.NoopTracer{}
+	}
+
+	adr := fmt.Sprintf("%s:%d", builder.host, builder.port)
+	conn, err := grpc.Dial(adr, grpc.WithStatsHandler(&ocgrpc.ClientHandler{}), grpc.WithInsecure(),
+		grpc.WithUnaryInterceptor(
+			otgrpc.OpenTracingClientInterceptor(builder.tracer)),
+		grpc.WithStreamInterceptor(
+			otgrpc.OpenTracingStreamClientInterceptor(builder.tracer)))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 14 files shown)


sdk/go/client.go
```diff
@@ -24,12 +25,56 @@ type GrpcClient struct {
...
+
+
+// funcClientOption wraps a function that modifies clientOptions into an
+// implementation of the ClientOption interface.
+type funcClientOption struct {
+	f func(*clientOptions)
+}
+
+func (fdo *funcClientOption) apply(do *clientOptions) {
+	fdo.f(do)
```
woop, line 54: Is all this complexity around ClientOptions just so that we can have an optional argument? Wouldn't it be cleaner to use some kind of getter/setter or builder pattern?

sdk/java/src/main/java/com/gojek/feast/FeastClient.java
```diff
@@ -47,11 +50,23 @@
    *
    * @param host hostname or ip address of Feast serving GRPC server
    * @param port port number of Feast serving GRPC server
+   * @param tracer tracer instance to be used for tracing grpc calls to Feast
    * @return {@link FeastClient}
    */
-  public static FeastClient create(String host, int port) {
+  public static FeastClient create(String host, int port, Tracer tracer) {
```
woop, line 56: Shouldn't we have a method without a tracer as well?

sdk/java/src/main/java/com/gojek/feast/FeastClient.java
```diff
@@ -47,11 +50,23 @@
...
     ManagedChannel channel = ManagedChannelBuilder.forAddress(host, port).usePlaintext().build();
-    return new FeastClient(channel);
+    return new FeastClient(channel, tracer);
+  }
+
+  /**
+   * Create a client to access Feast
```
woop, line 62: Can we please be more descriptive in comments. This is very vague.

sdk/python/setup.py
```diff
@@ -46,6 +46,8 @@
     "numpy",
     "google",
     "confluent_kafka",
+    "grpcio-opentracing",
+    "opentracing"
```
woop, line 50: Do we need to force users to have this dependency?

sdk/python/feast/client.py
```diff
@@ -282,6 +297,16 @@ def version(self):
 
         return result
 
+    def set_tracer(self, tracer: Tracer):
+        """
+        Set the OpenTracing Tracer instance used to trace the gRPC calls to Feast Core
```
woop, line 302: Do we need to trace to Core as well? Is that even implemented?

serving/src/main/java/feast/serving/controller/ServingServiceGRpcController.java
```diff
@@ -96,7 +100,11 @@ public void getOnlineFeatures(
         this.checkProjectAccess(request.getFeaturesList());
       }
       RequestHelper.validateOnlineRequest(request);
+      Span span = tracer.buildSpan("getOnlineFeatures").start();
```
woop, line 103: Can you explain the motivation/difference in this change? Where would a spike in authorization show up?

khorshuheng: The interceptor will automatically create a span surrounding the entire GetOnlineFeatureRequest calls. Any span that is created within will be a child span. The end result looks something like:

GetOnlineFeatureRequest  |------------------------------------------------------------------|
getOnlineFeature                                    |-------------------------------------------|

So the time between the start of GetOnlineFeatureRequest and the start of getOnlineFeature will be the time spent on authorization, validation, etc.

We can add spans for authorization as well, in which case it will look like:

GetOnlineFeatureRequest  |------------------------------------------------------------------|
authorization                            |-----|
getOnlineFeature                                    |-------------------------------------------|

So it really depends on how fine grained we want the span to be like. I am fine with adding more span and possibly renaming the operation name, if you think that's better?

sdk/python/feast/client.py
khorshuheng: It is not implemented on server side. So, if we include the interceptor here, it will only capture the latency as observed by the client side (and ingress gateway, if set up)

The reason i include Feast core, is because Feast client is shared by both serving and core, and it seems a little unnatural that setting a tracer affects only serving but not core. Either way, I can rename set_tracer to set_serving_tracer, and remove the interceptor for Feast core, if we don't want to trace the call to Feast Core?

sdk/java/src/main/java/com/gojek/feast/FeastClient.java
khorshuheng: The original method wasn't removed: https://github.com/feast-dev/feast/pull/962/files#diff-091f65854936e6e5c42af98a3893a2dfR70

sdk/python/setup.py
khorshuheng: This is a little tricky. `opentracing` is only used because i want to use type hint for the tracer argument. So, if i remove the type hint, we don't really need `opentracing`, and the dependency can be safely removed.

As for `grpcio-opentracing`, that is used for the interceptor. I can move the import within the method (and only if tracer is set), so that the SDK doesn't run into runtime error. But, how should the user know that they need to include `grpcio-opentracing` in their dependencies, if they want to enable tracing? Would returning an error message and document this explicitly in the method comment suffice?

PR #969, opened 2020-09-01
Add configurable prefix to Consumer Group in IngestionJob's Kafka reader

terryyylim:
To allow multiple deployments with jobs running in parallel, we should separate their Kafka consumer groups which is based on the JobId. This would prevent overwriting each other's offsets.

Fixes #

```release-note
NONE
```

### DIFF
--- a/job-controller/src/main/java/feast/jobcontroller/config/JobControllerConfig.java
+++ b/job-controller/src/main/java/feast/jobcontroller/config/JobControllerConfig.java
@@ -87,9 +87,9 @@ public class JobControllerConfig {
     Boolean shouldConsolidateJobs =
         feastProperties.getJobs().getController().getConsolidateJobsPerSource();
     if (shouldConsolidateJobs) {
-      return new ConsolidatedJobStrategy(jobRepository);
+      return new ConsolidatedJobStrategy(jobRepository, feastProperties.getJobs());
     } else {
-      return new JobPerStoreStrategy(jobRepository);
+      return new JobPerStoreStrategy(jobRepository, feastProperties.getJobs());
     }
   }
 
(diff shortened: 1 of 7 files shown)


job-controller/src/main/resources/application.yml
```diff
@@ -23,6 +23,9 @@ feast:
     # Enabling JobManagement
     enabled: true
 
+    # Prefix for JobId
```
woop, line 26: This comment says absolutely nothing.

job-controller/src/main/java/feast/jobcontroller/runner/ConsolidatedJobStrategy.java
```diff
@@ -71,6 +74,9 @@ private String createJobId(SourceProto.Source source) {
                 source.getKafkaSourceConfig().getBootstrapServers(),
                 source.getKafkaSourceConfig().getTopic()),
             dateSuffix);
+    if (!this.jobProperties.getJobIdPrefix().isEmpty()) {
```
woop, line 77: What happens when jobProperties is null?

job-controller/src/main/java/feast/jobcontroller/config/JobControllerConfig.java
```diff
@@ -87,9 +87,9 @@ public JobGroupingStrategy getJobGroupingStrategy(
     Boolean shouldConsolidateJobs =
         feastProperties.getJobs().getController().getConsolidateJobsPerSource();
     if (shouldConsolidateJobs) {
-      return new ConsolidatedJobStrategy(jobRepository);
+      return new ConsolidatedJobStrategy(jobRepository, feastProperties.getJobs());
```
woop, line 90: Can we extract the jobProperties before passing it? It's a convention that I am trying to get us to follow, even if it takes more lines of code

job-controller/src/main/java/feast/jobcontroller/runner/ConsolidatedJobStrategy.java
terryyylim: Addressed case.

job-controller/src/main/resources/application.yml
terryyylim: Updated comment.

job-controller/src/main/java/feast/jobcontroller/config/JobControllerConfig.java
terryyylim: Extracted before passing jobProperties.

pyalex (approved):

PR #975, opened 2020-09-01
Add redis key prefix as an option to Redis cluster

khorshuheng:
There are scenarios where multiple Feast deployment are sharing the same Redis cluster, but key collision between the different deployment should be avoided. Since Redis cluster doesn't support database, attaching prefix to keys is a good workaround to achive namespacing.

To facilitate migration of redis keys from one prefix to another, fallback mechanism has also been introduced. If enabled, when a key is not found in the database, a fallback prefix can be used instead. However, this will results in more calls to Redis, and additional latency.

Fixes #

```release-note
Additional optional configuration for Redis cluster, where user can specified a prefix to be prepended to the Redis keys written / retrieved.
```

### DIFF
--- a/storage/connectors/redis/src/test/java/feast/storage/connectors/redis/retriever/RedisClusterOnlineRetrieverTest.java
+++ b/storage/connectors/redis/src/test/java/feast/storage/connectors/redis/retriever/RedisClusterOnlineRetrieverTest.java
@@ -23,7 +23,6 @@ import static org.mockito.MockitoAnnotations.initMocks;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
-import com.google.protobuf.AbstractMessageLite;
 import com.google.protobuf.Duration;
 import com.google.protobuf.Timestamp;
 import feast.proto.core.FeatureSetProto.EntitySpec;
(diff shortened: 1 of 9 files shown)


pyalex:
Since we may change prefix during project lifetime (with new release, for example), falling mechanism is needed to use both new & old prefixes.

storage/connectors/redis/src/test/java/feast/storage/connectors/redis/retriever/RedisClusterOnlineRetrieverTest.java
```diff
@@ -160,46 +180,21 @@ public void shouldReturnResponseWithValuesIfKeysPresent() {
...
-    List<FeatureRow> featureRows =
-        Lists.newArrayList(
-            FeatureRow.newBuilder()
-                .setEventTimestamp(Timestamp.newBuilder().setSeconds(100))
-                .addAllFields(
-                    Lists.newArrayList(
-                        Field.newBuilder().setValue(intValue(1)).build(),
-                        Field.newBuilder().setValue(intValue(1)).build()))
-                .build());
+    when(serializer.serialize(redisKeys.get(0))).thenReturn(serializedKey1);
```
pyalex, line 186: do we need to mock serializer? why not to use actual implementation?

khorshuheng: I have replaced the mock serializer with actual implementation.

pyalex (approved):

PR #976, opened 2020-09-01
Fix invalid characters for project, featureset, entity and features creation

terryyylim:
Currently - in names is being replaced to _. This is hacky solution that need to be implemented in many places. It also creates potential bug since project-1 and project_1 (feature-set and feature_set) will be using the same bq table.

It's better to validate this (and not allow such characters) on feature set creation. This PR addresses the above issue for project, featureset, features and entities.

Fixes #

```release-note
Users will no longer be allowed to include dashes in project, featureset, entity and feature name.
```

### DIFF
--- a/core/src/main/java/feast/core/validators/FeatureSetValidator.java
+++ b/core/src/main/java/feast/core/validators/FeatureSetValidator.java
@@ -44,8 +44,8 @@ public class FeatureSetValidator {
       throw new IllegalArgumentException("Feature set label keys must not be empty");
     }
 
-    checkValidCharacters(featureSet.getSpec().getProject(), "project");
-    checkValidCharacters(featureSet.getSpec().getName(), "name");
+    checkValidCharacters(featureSet.getSpec().getProject(), "project::name");
+    checkValidCharacters(featureSet.getSpec().getName(), "featureset::name");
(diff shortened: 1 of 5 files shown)


core/src/main/java/feast/core/validators/FeatureSetValidator.java
```diff
@@ -44,8 +44,8 @@ public static void validateSpec(FeatureSet featureSet) {
       throw new IllegalArgumentException("Feature set label keys must not be empty");
     }
 
-    checkValidCharacters(featureSet.getSpec().getProject(), "project");
-    checkValidCharacters(featureSet.getSpec().getName(), "name");
+    checkValidCharacters(featureSet.getSpec().getProject(), "project::name");
+    checkValidCharacters(featureSet.getSpec().getName(), "featureset::name");
```
woop, line 48: Would it make more sense to have a new argument like `resource` and `field` instead of `::`

terryyylim: I added resource. I think it's clearer to show what they inputted as well, so I replaced field with the actual input itself in the error message.

pyalex (approved):

PR #977, opened 2020-09-02
Fix Python native types multiple entities online retrieval

terryyylim:
Online retrieval with python native types was introduced in 0.6 and a bug was introduced. Specifically, retrieving features with more than 1 entity would fail since only 1 entity is converted during the online retrieval call.

Fixes #

```release-note

```

### DIFF
--- a/tests/e2e/redis/basic-ingest-redis-serving.py
+++ b/tests/e2e/redis/basic-ingest-redis-serving.py
@@ -709,6 +709,62 @@ def test_basic_ingest_retrieval_str(client):
 
 @pytest.mark.timeout(600)
 @pytest.mark.run(order=18)
+def test_basic_ingest_retrieval_multi_entities(client):
+    # Set to another project to test ingestion based on current project context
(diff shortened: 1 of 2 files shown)


tests/e2e/redis/basic-ingest-redis-serving.py
```diff
@@ -709,6 +709,63 @@ def try_get_features():
...
+    merchant_df = pd.DataFrame(
+        {
+            "datetime": [time_offset] * N_ROWS,
+            "driver_id": [i for i in range(N_ROWS)],
+            "merchant_id": [i for i in range(N_ROWS)],
+            "merchant_sales": [float(i) + 0.5 for i in range(N_ROWS)],
+        }
+    )
+    client.ingest("merchant_fs", merchant_df, timeout=600)
+    time.sleep(15)
```
woop, line 737: What does this sleep do by the way?

terryyylim: Not required, removed.

pyalex (approved):

PR #971, opened 2020-09-01
Authentication Support for Java & Go SDKs

mrzzy:
**What this PR does / why we need it**:

Implements support for authentication to Go and Java SDKs, which is required for interacting with a authentication enabled Feast Serving.
- Update Feast clients to attach OIDC token used for authentication into metadata when enabled.
- Allows users to provide credentials via Google, OAuth Client Credentials request, or a Static token.
  - Google: Uses user provided Service Account Credentials (ie JSON) to obtain token.
  - OAuth Client Credentials: Makes a client credentials request to obtain token.
  - Static: Uses a user provided static token.

**Which issue(s) this PR fixes**:

Fixes #950 

**Does this PR introduce a user-facing change?**:

```release-note
Adds authentication support to Go and Java SDKs via Google, OAuth Client Credentials request, or a Static token.
```

PR #986, opened 2020-09-07
Fix Java & Go SDK TLS support

mrzzy:
Follow up of #971:
- implements missing TLS support for Java SDK.
- Fixes Go SDK issue where it does not use system certificates if TLS enabled but no certificate path is specified.
- Rename Go and Java SDK secure client creation methods to include "secure" instead of "authenticated"
     - Go SDK: `NewAuthGrpcClient()` -> `NewSecureGrpcClient()`  
     - Java SDK: `createdAuthenticated()` -> `createSecure()`

Continuation of #504

Fixes #

```release-note
implements TLS support for Java SDK.
 Rename Go and Java SDK secure client creation methods to include "secure" instead of "authenticated"
     - Go SDK: `NewAuthGrpcClient()` -> `NewSecureGrpcClient()`  
     - Java SDK: `createdAuthenticated()` -> `createSecure()`
```

### DIFF
--- a/sdk/go/client.go
+++ b/sdk/go/client.go
@@ -63,7 +63,7 @@ func NewSecureGrpcClient(host string, port int, security SecurityConfig) (*GrpcC
 			return nil, err
 		}
 		options = append(options, grpc.WithTransportCredentials(tlsCreds))
-	} else if security.EnableTLS {
+	} else {
 		// Use system TLS certificate pool.
 		certPool, err := x509.SystemCertPool()
 		if err != nil {
--- a/sdk/go/client.go
+++ b/sdk/go/client.go
@@ -27,7 +27,7 @@ type GrpcClient struct {
 type SecurityConfig struct {
 	// Whether to enable TLS SSL trasnport security if true.
 	EnableTLS bool
-	// Optional: Provides path to TLS certificate use the verify Service identity.
+	// Optional: Provides path to TLS certificate used the verify Service identity.
 	TLSCertPath string
 	// Optional: Credential used for authentication.
 	// Disables authentication if unspecified.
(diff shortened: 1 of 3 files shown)


pyalex (approved):

sdk/go/client.go
```diff
@@ -36,33 +37,42 @@ type SecurityConfig struct {
...
-	// Read TLS certificate from given path instead of using system certs if specified.
-	if security.EnableTLS && security.TLSCertPath != "" {
+	} else if security.EnableTLS && security.TLSCertPath != "" {
+		// Read TLS certificate from given path.
 		tlsCreds, err := credentials.NewClientTLSFromFile(security.TLSCertPath, "")
 		if err != nil {
 			return nil, err
 		}
 		options = append(options, grpc.WithTransportCredentials(tlsCreds))
+	} else if security.EnableTLS {
```
woop, line 66: Should this maybe be an `else`? the predicate is redundant.

sdk/java/src/main/java/com/gojek/feast/FeastClient.java
```diff
@@ -52,8 +56,12 @@
    * @return {@link FeastClient}
    */
   public static FeastClient create(String host, int port) {
-    ManagedChannel channel = ManagedChannelBuilder.forAddress(host, port).usePlaintext().build();
-    return new FeastClient(channel, Optional.empty());
+    try {
+      // configure client with no security config.
+      return FeastClient.createSecure(host, port, SecurityConfig.newBuilder().build());
+    } catch (SSLException e) {
+      throw new RuntimeException(e);
```
woop, line 63: Is there a reason to rethrow without any attached message?

sdk/go/client.go
mrzzy: Updated.

sdk/java/src/main/java/com/gojek/feast/FeastClient.java
mrzzy: Updated `createSecure()` to throw `IllegalArgumentException` when provided with bad certificate.
- Removed redundant try catch in `create()`

woop (approved):

PR #996, opened 2020-09-10
Remove and update deprecation functions

terryyylim:
This PR removes functions from Python SDK which have been marked to be deprecated in future releases.

Fixes #

```release-note
Users should use get_historical_features instead of get_batch_features, and only pass in list of dictionaries for get_online_features instead of passing in proto EntityRow.
```

### DIFF
--- a/infra/charts/feast/templates/tests/test-feast-batch-serving.yaml
+++ b/infra/charts/feast/templates/tests/test-feast-batch-serving.yaml
@@ -15,7 +15,7 @@ spec:
     - bash
     - -c
     - |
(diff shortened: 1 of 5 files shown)


mrzzy:
Since we are using a new API, maybe we should be switching SDKs?

https://github.com/terryyylim/feast/blob/f004268d7d03e10d696d1adf53661934b03a8e6c/infra/charts/feast/templates/tests/test-feast-online-serving.yaml#L18

khorshuheng (approved):

PR #1000, opened 2020-09-13
Fixed SA Key permission denied in Jupyter Server deployed using Docker Compose

mrzzy:
**Why we need it**:
Using the docker-compose deploy, users will encounter permission denied error when running an example notebook
when the Google SDK attempts to access credentials at `/etc/gcloud/service-accounts/key.json`

**What this PR does**
Adds an entrypoint.sh script to fix permissions for the SA key on container startup.

Fixes #

```release-note

```

### DIFF
--- a/infra/docker/jupyter/entrypoint.sh
+++ b/infra/docker/jupyter/entrypoint.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+#
+# Feast Jupyter Server
+# Entrypoint Script
+#
+
+# Update permissions to allow notebook user access to - Google SA key 
+if [ -f ${GOOGLE_APPLICATION_CREDENTIALS} ]
+then
+    chown -R $NB_UID /etc/gcloud
+fi
+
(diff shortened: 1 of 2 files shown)


infra/docker/jupyter/entrypoint.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+#
+# Feast Jupyter Server
+# Entrypoint Script
+#
+
+# Update permissions to allow notebook user access to - Google SA key
```
woop, line 7: Isn't the NB_UID always the same? Could we statically provide access to this service account to the user by granting ownership ahead of time?

mrzzy: Its possible, we can grant permissions by `chown` before `docker-compose up`:
```sh
chown 1000:1000 infra/docker-compose/gcp-service-accounts/key.json
docker-compose up
```
closing this PR in favour of docs to direct users to do this.

PR #1001, opened 2020-09-13
Fix Go SDK extra colon in metadata header for Authentication

mrzzy:
Fixes Go SDK outputting a extra `:` in `Authorization` grpc metadata header used for authentication:
- `Bearer: TOKEN` instead of `Bearer TOKEN`, the latter which is correct.
- Bug causes Go SDK authentication with Feast to fail.

Fixes #

```release-note

```


mrzzy:
/

woop (approved):

PR #1003, opened 2020-09-15
Fix listing of Ingestion Jobs with Error Status

### LINKED ISSUE #895: Ingestion Job API List should show Ingestion Jobs with error JobStatus
## Current Behavior 
Currently Ingestion Job API `JobService`'s list ingestion job operation skips over Ingestion Jobs with an `ERROR` `JobStatus` state.
https://github.com/feast-dev/feast/blob/72cc12098fa2cdbc41cec11ec93c45cd65cd513a/core/src/main/java/feast/core/service/JobService.java#L140-L144

## Expected Behavior
Users expect that Ingestion Job List should show all ingestion Jobs irregardless of `JobStatus`.

If verbosity is a concern a flag could be added to filter out jobs by `JobStatus` (ie `ERROR`),
but by default all Ingestion Jobs should be shown.

terryyylim:
Listing of ingestion jobs skip over the ones with `ERROR` job status.

Fixes #895 

```release-note
NONE
```

### DIFF
--- a/job-controller/src/main/java/feast/jobcontroller/service/JobService.java
+++ b/job-controller/src/main/java/feast/jobcontroller/service/JobService.java
@@ -112,11 +112,6 @@ public class JobService {
     List<IngestionJobProto.IngestionJob> ingestJobs = new ArrayList<>();
     for (String jobId : matchingJobIds) {
       Job job = this.jobRepository.findById(jobId).orElseThrow();
-      // job that failed on start won't be converted toProto successfully
-      // and they're irrelevant here
-      if (job.getStatus() == JobStatus.ERROR) {
-        continue;
-      }
       ingestJobs.add(job.toProto());
     }
 


job-controller/src/main/java/feast/jobcontroller/service/JobService.java
```diff
@@ -112,11 +112,6 @@ public ListIngestionJobsResponse listJobs(ListIngestionJobsRequest request) {
     List<IngestionJobProto.IngestionJob> ingestJobs = new ArrayList<>();
     for (String jobId : matchingJobIds) {
       Job job = this.jobRepository.findById(jobId).orElseThrow();
-      // job that failed on start won't be converted toProto successfully
```
pyalex, line 115: Have this `job that failed on start won't be converted toProto successfully` been fixed?

PR #1014, opened 2020-09-23
Introduce Entity as higher-level concept

### LINKED ISSUE #405: Entity types as a higher-level concept

terryyylim:
Introducing entities as a higher-level concept is the first step in generalizing Feast. The Feast v0.8 proposal which seeks to do so can be found [here](https://docs.google.com/document/d/1snRxVb8ipWZjCiLlfkR4Oc28p7Fkv_UXjvxBFWjRBj4/edit#).

This PR does not include versioning of protos nor removal of existing EntitySpec concept. This will be done in a separate PR in the near future.

Fixes #405 

```release-note
Entities (as a higher-level concept) can now be registered and retrieved from Feast Core.
```


protos/feast/core/CoreService.proto
```diff
@@ -202,6 +269,19 @@ message ListStoresResponse {
     repeated feast.core.Store store = 1;
 }
 
+message ApplyEntityRequest {
+    // If project is unspecified, will default to 'default' project.
+    // If project specified does not exist, the project would be automatically created.
+    feast.core.Entity entity = 1;
```
mrzzy, line 275: Which parts of the entity can be updated?

protos/feast/core/CoreService.proto
```diff
@@ -202,6 +269,19 @@ message ListStoresResponse {
...
+
+    // Name of project that this entity belongs to.
+    string project = 2;
+}
+
+message ApplyEntityResponse {
+    feast.core.Entity entity = 1;
```
mrzzy, line 282: Not sure if returning the entity would be useful.
Would returning a status be useful instead?

terryyylim: Not too sure what you mean here. Aren't we getting rid of the use of Status?

protos/feast/core/CoreService.proto
terryyylim: That can be found [here](https://github.com/feast-dev/feast/pull/1014/files#diff-431a7153babcc8eca8290e9578cd7996R82).

protos/feast/core/CoreService.proto
mrzzy: Hmm ok that I will get rid of statuses on my end as well.

mrzzy:
Make sure to update go language definitions your new protobufs with `make compile-go-protos` and commit them.

protos/feast/core/CoreService.proto
mrzzy: Does it make more sense to expose only `EntitySpecV2` instead to users, as the other (meta) fields are set by Feast?

core/src/main/java/feast/core/grpc/CoreServiceImpl.java
```diff
@@ -191,6 +243,40 @@ public void listStores(
     }
   }
 
+  @Override
+  public void applyEntity(
```
woop, line 247: I think we should start to layer on comments here.

core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -65,23 +73,63 @@
...
+    if (projectName.isEmpty()) {
+      projectName = Project.DEFAULT_NAME;
+    }
+
+    checkValidCharacters(projectName, "project");
+    checkValidCharacters(entityName, "entity");
+
+    EntityV2 entity;
+
+    entity = entityRepository.findEntityByNameAndProject_Name(entityName, projectName);
```
woop, line 120: Any reason why we cant initialize and instantiate in one line?

core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -277,6 +325,94 @@ public ListFeaturesResponse listFeatures(ListFeaturesRequest.Filter filter) {
...
+   * be returned. Regex is not supported. Explicitly defining an entity name is not possible if a
+   * project name is not set explicitly.
+   *
+   * <p>The labels in the filter accepts a map. All entities which contain every provided label will
+   * be returned.
+   *
+   * @param filter Filter containing the desired entity name, project and labels
+   * @return ListEntitiesResponse with list of entities found matching the filter
+   */
+  public ListEntitiesResponse listEntities(ListEntitiesRequest.Filter filter) {
```
woop, line 347: Do we have a strong use for this wildcard filters or can we remove it?

core/src/main/java/feast/core/grpc/CoreServiceImpl.java
```diff
@@ -191,6 +243,40 @@ public void listStores(
...
+      ApplyEntityRequest request, StreamObserver<ApplyEntityResponse> responseObserver) {
+
+    String projectId = null;
+
+    try {
+      Entity entity = request.getEntity();
```
pyalex, line 253: do all this lines need to be in try catch?

core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -324,6 +460,61 @@ public ListStoresResponse listStores(ListStoresRequest.Filter filter) {
...
+    // Find project or create new one if it does not exist
+    Project project = projectRepository.findById(projectName).orElse(new Project(projectName));
+
+    // Ensure that the project retrieved from repository is not archived
+    if (project.isArchived()) {
+      throw new IllegalArgumentException(String.format("Project is archived: %s", projectName));
+    }
+
+    // Retrieve existing Entity
+    EntityV2 entity =
```
woop, line 492: Should this be a transaction to avoid a race condition?

core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -324,6 +460,61 @@ public ListStoresResponse listStores(ListStoresRequest.Filter filter) {
...
+        entityRepository.findEntityByNameAndProject_Name(
+            newEntity.getSpec().getName(), projectName);
+
+    if (entity == null) {
+      // Create new entity since it doesn't exist
+      newEntity = newEntity.toBuilder().setSpec(newEntity.getSpec()).build();
+      entity = EntityV2.fromProto(newEntity);
+    } else {
+      // If the entity remains unchanged, we do nothing.
+      if (entity.toProto().getSpec().equals(newEntity.getSpec())) {
```
woop, line 502: Does this work for labels as well? So maps will be equal?

core/src/main/resources/db/migration/V2.7__Entities_Higher_Level_Concept.sql
```diff
@@ -0,0 +1,15 @@
...
+    last_updated timestamp without time zone NOT NULL,
+    name character varying(255),
+    project_name character varying(255),
+    type character varying(255),
+    description text,
+    labels text
+);
+
+ALTER TABLE ONLY entities_v2
+    ADD CONSTRAINT entities_v2_pkey PRIMARY KEY (id);
```
woop, line 15: Should there be a `project` foreign key/

core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -65,23 +73,63 @@
...
+
+    if (entityName.isEmpty()) {
+      throw new IllegalArgumentException("No entity name provided");
+    }
+    // Autofill default project if project is not specified
+    if (projectName.isEmpty()) {
+      projectName = Project.DEFAULT_NAME;
+    }
+
+    checkValidCharacters(projectName, "project");
```
pyalex, line 115: do we need to validate it here?
we assume that entity with invalid characters cannot be created

protos/feast/core/CoreService.proto
```diff
@@ -71,6 +75,22 @@ service CoreService {
...
     rpc ApplyFeatureSet (ApplyFeatureSetRequest) returns (ApplyFeatureSetResponse);
 
+    // Create or update and existing entity.
+    //
+    // This function is idempotent - it will not create a new entity if schema does not change.
+    // Schema changes will update the entity if the changes are valid.
+    // Following changes are not valid:
+    // - Changes to name
+    // - Changes to type
+    // - Changes to description
```
woop, line 85: We cant change the description? is this a mistake?

core/src/main/resources/db/migration/V2.7__Entities_Higher_Level_Concept.sql
```diff
@@ -0,0 +1,15 @@
...
+    created timestamp without time zone NOT NULL,
+    last_updated timestamp without time zone NOT NULL,
+    name character varying(255),
+    project_name character varying(255),
+    type character varying(255),
+    description text,
+    labels text
+);
+
+ALTER TABLE ONLY entities_v2
```
pyalex, line 14: can we also have unique constraint (project_name, name) here?

sdk/python/feast/client.py
```diff
@@ -353,6 +360,145 @@ def archive_project(self, project):
         if self._project == project:
             self._project = FEAST_DEFAULT_OPTIONS[CONFIG_PROJECT_KEY]
 
+    def apply_entity(
+        self, entities: Union[List[EntityV2], EntityV2], project: str = None
+    ):
+        """
+        Idempotently registers entities with Feast Core. Either a single
+        entity or a list can be provided.
```
woop, line 368: What if we just make this a single entity? Do we need to support a list?

sdk/python/feast/client.py
```diff
@@ -353,6 +360,145 @@ def archive_project(self, project):
...
+            >>>     value_type=ValueType.STRING,
+            >>>     labels={
+            >>>         "key": "val"
+            >>>     }
+            >>> )
+            >>> feast_client.apply_entity(entity)
+        """
+
+        if project is None:
+            if self.project is not None:
```
woop, line 391: Can we move this if/else into the project property so that the ValueError is thrown there? Otherwise people will continue to have to implement this conditional. `self.property` should always return something or throw an exception.

protos/feast/core/CoreService.proto
pyalex: agreed, `EntitySpecV2` should be here instead

tests/e2e/redis/basic-ingest-redis-serving.py
```diff
@@ -1390,6 +1390,50 @@ def is_different_jobs():
         time.sleep(1)
 
 
+@pytest.mark.run(order=80)
+def test_basic_register_entity_success(client):
```
pyalex, line 1394: I'm not sure we need entity tests in e2e and especially redis-serving e2e

sdk/python/tests/test_entity.py
```diff
@@ -0,0 +1,78 @@
...
+CORE_URL = "core.feast.local"
+SERVING_URL = "serving.feast.local"
+
+
+class TestEntity:
+    @pytest.fixture(scope="function")
+    def server(self):
+        server = grpc.server(futures.ThreadPoolExecutor(max_workers=10))
+        Core.add_CoreServiceServicer_to_server(CoreServicer(), server)
+        server.add_insecure_port("[::]:50051")
```
pyalex, line 36: dynamic port generation please

sdk/python/feast/client.py
```diff
@@ -353,6 +360,145 @@ def archive_project(self, project):
...
+        except grpc.RpcError as e:
+            raise grpc.RpcError(e.details())
+
+        # Extract the returned entity
+        applied_entity = EntityV2.from_proto(apply_entity_response.entity)
+
+        # Deep copy from the returned entity to the local entity
+        entity._update_from_entity(applied_entity)
+
+    def list_entities_v2(
```
pyalex, line 431: I guess in SDK we can replace previous methods, as we discussed

sdk/python/feast/entity.py
```diff
@@ -44,3 +56,246 @@ def from_proto(cls, entity_proto: EntityProto):
...
+        name: str,
+        description: str,
+        value_type: ValueType,
+        labels: Optional[MutableMapping[str, str]] = None,
+    ):
+        self._name = name
+        self._description = description
+        self._value_type = value_type
+        if labels is None:
+            self._labels = OrderedDict()  # type: MutableMapping[str, str]
```
woop, line 77: Do labels have to be ordered?

sdk/python/feast/entity.py
```diff
@@ -44,3 +56,246 @@ def from_proto(cls, entity_proto: EntityProto):
...
+        else:
+            self._labels = labels
+
+        self._created_timestamp: Optional[Timestamp] = None
+
+    def __eq__(self, other):
+        if not isinstance(other, EntityV2):
+            return NotImplemented
```
woop, line 85: Should this not be a descriptive exceptioN?

sdk/python/tests/test_entity.py
```diff
@@ -0,0 +1,78 @@
...
+import grpc
+import pytest
+
+from feast.client import Client
+from feast.core import CoreService_pb2_grpc as Core
+from feast.entity import EntityV2
+from feast.value_type import ValueType
+from feast_core_server import CoreServicer
+
+CORE_URL = "core.feast.local"
```
woop, line 27: Where is this used?

sdk/python/feast/entity.py
```diff
@@ -44,3 +56,246 @@ def from_proto(cls, entity_proto: EntityProto):
...
+
+    @property
+    def created_timestamp(self):
+        """
+        Returns the created_timestamp of this entity
+        """
+        return self._created_timestamp
+
+    @created_timestamp.setter
+    def created_timestamp(self, created_timestamp):
```
woop, line 170: Why do we expose a setter for this meta field?

core/src/main/java/feast/core/model/EntityV2.java
```diff
@@ -0,0 +1,136 @@
...
+              "You are attempting to change the type of this entity in %s project from %s to %s. This isn't allowed. Please create a new entity.",
+              projectName, getType(), spec.getValueType().toString()));
+    }
+
+    // 2. Update labels
+    this.setLabels(TypeConversion.convertMapToJsonString(spec.getLabelsMap()));
+  }
+
+  /**
+   * Return a boolean to facilitate streaming elements on the basis of given predicate.
```
woop, line 118: What does this even mean?

core/src/main/java/feast/core/model/EntityV2.java
```diff
@@ -0,0 +1,136 @@
...
+            .setValueType(ValueType.Enum.valueOf(getType()))
+            .putAllLabels(TypeConversion.convertJsonStringToMap(labels));
+
+    // Build Entity
+    EntityProto.Entity entity = EntityProto.Entity.newBuilder().setMeta(meta).setSpec(spec).build();
+    return entity;
+  }
+
+  /**
+   * Updates the existing entity from a proto.
```
woop, line 97: Does this only update labels?

core/src/main/java/feast/core/model/EntityV2.java
```diff
@@ -0,0 +1,136 @@
...
+
+  // User defined metadata
+  @Column(name = "labels", columnDefinition = "text")
+  private String labels;
+
+  public EntityV2() {
+    super();
+  }
+
+  public EntityV2(
```
woop, line 63: Would you please add comments to describe the model?

core/src/main/java/feast/core/service/SpecService.java
terryyylim: I guess it'll be easier to locate a set of entities that are registered with a certain pattern within a project.

core/src/main/java/feast/core/service/SpecService.java
terryyylim: Not necessary as we could just return nothing found. Just thought this would be clearer to the user, telling them that they're retrieving entity with invalid characters.

protos/feast/core/CoreService.proto
terryyylim: I was thinking in the perspective of sharing the use of an entity. If the description is changed after another user has started using it, it may confused him/her about the entity being used. We could allow changing the description, but we'll need to find a way to ensure the description is an appropriate one each time it's changed and also made known to users.

sdk/python/feast/client.py
terryyylim: I think it's better to support a list, especially in the case if we only allow users to register entity through yaml config. This may require us to have a new yaml file for every new entity that's to be registered. Unless we're okay with that.

woop:
Doesnt this PR introduce a user facing change?

core/src/main/java/feast/core/service/SpecService.java
woop: Yes, but is there a need for it?

protos/feast/core/CoreService.proto
woop: How do we ensure the first description is appropriate? Are we supposed to police that? I dont think this is Feast's problem.

terryyylim: Yeah, would be difficult. I'll amend this.

core/src/main/java/feast/core/model/EntityV2.java
terryyylim: It should update description as well. Will add.

sdk/python/tests/test_entity.py
terryyylim: Removed.

sdk/python/feast/entity.py
terryyylim: Added.

sdk/python/feast/entity.py
terryyylim: Removed.

sdk/python/feast/client.py
terryyylim: Replaced.

sdk/python/feast/client.py

core/src/main/resources/db/migration/V2.7__Entities_Higher_Level_Concept.sql
terryyylim: Added.

core/src/main/resources/db/migration/V2.7__Entities_Higher_Level_Concept.sql
terryyylim: Updated.

core/src/main/java/feast/core/service/SpecService.java
terryyylim: No reason, updated.

core/src/main/java/feast/core/model/EntityV2.java
```diff
@@ -60,6 +60,12 @@ public EntityV2() {
     super();
   }
 
+  /**
+   * EntityV2 object supports Entity registration outside of FeatureSet.
```
woop, line 64: It only applies to Feature Tables right?

core/src/main/java/feast/core/grpc/CoreServiceImpl.java
terryyylim: Added a small comment, I guess SpecService function has docs on it, not sure if we want to add too much here.

core/src/main/java/feast/core/model/EntityV2.java
terryyylim: Updated.

core/src/main/java/feast/core/model/EntityV2.java
terryyylim: Updated doc.

core/src/main/java/feast/core/service/SpecService.java
terryyylim: Not a strong need, have removed it.

core/src/main/java/feast/core/service/SpecService.java
terryyylim: Yup, tested by updating labels vs. no updates.

tests/e2e/redis/basic-ingest-redis-serving.py
terryyylim: Removed.

sdk/python/feast/entity.py
terryyylim: Removed.

core/src/main/java/feast/core/model/EntityV2.java
terryyylim: Added.

core/src/main/java/feast/core/model/EntityV2.java
terryyylim: Updated.

protos/feast/core/CoreService.proto
terryyylim: Updated.

sdk/python/feast/entity.py
```diff
@@ -44,3 +55,273 @@ def from_proto(cls, entity_proto: EntityProto):
         """
         entity = cls(name=entity_proto.name, dtype=ValueType(entity_proto.value_type))
         return entity
+
+
+class EntityV2:
```
woop, line 60: Do we really need to expose the `V2` to users? I think we can replace the existing Entity with this new one. Any concerns with that @mrzzy ?

core/src/main/java/feast/core/service/SpecService.java
woop: Hmmm, what have you removed? I still see `name.replace('*', '%'), project.replace('*', '%'))`

woop: If its easier to keep in then we can keep it in, but this seems like a lot of complexity to keep if we dont actually need it.

terryyylim: Oh, I disallowed searching for wildcard names in a specific project. So if we're removing wildcard filters, should we update the proto to only consider project and labels in the Filter then?

terryyylim: Doesn't seem very useful to have entity name in the labels since it'll just return that entity in a list. The user might be better off using getEntity in this case.

sdk/python/feast/entity.py
terryyylim: I think that's okay, but we'll need to update (perhaps use another name?) the previous Entity implementation used in tests and feature set. However, this would introduce breaking change to existing users who are still using the featureset-entity concept.

eg. if they do something like
```
fs = FeatureSet(
        "featureset",
        features=[
            Feature("feat1", ValueType.STRING)
        ],
        entities=[Entity("entity_id", ValueType.INT64)],
        max_age=Duration(seconds=100),
    )
```

terryyylim: Will be updating SDK in separate PR, along with replacing existing e2e tests.

woop:
Looks good to me. I will approve. @pyalex can you lgtm when you are ready.

woop (approved):

PR #1016, opened 2020-09-24
Added kustomization templates

### LINKED ISSUE #258: Create Kustomize installation for Feast 0.3
The primary means of deploying Feast right now is [Helm](helm.sh). Helm templates create the Kubernetes manifests that deploy the various components like Feast Core, Serving, and optionally some infrastructural components like Redis, PostgreSQL, and Kafka.

Helm charts for Feast 0.3 installation can be found [here](https://github.com/gojek/feast/tree/0.3-dev/infra/charts). 

Helm is becoming harder and harder to maintain, even though it's the defacto package manager in the Kubernetes ecosystem. We'd like to add support for installing Feast through [Kustomize](https://github.com/kubernetes-sigs/kustomize). This would hopefully allow us to have a simpler and more human readable way to install Feast, and would allow for integration of Feast onto platforms like Kubeflow.

In order to complete this issue, we would like to have a Kustomize installation script that has feature parity with the existing Helm installation. Meaning you can configure and deploy a complete Feast 0.3 installation into a Kubernetes cluster.

ravdin:
Currently Feast is primarily installed through Helm templates. We would also like to support installing the various Feast components with Kustomize. This update adds Kustomize templates for installing the various Feast services. We have implemented a similar solution internally.

Fixes #258. 

```release-note
NONE
```

### DIFF
--- a/infra/kustomization/README.md
+++ b/infra/kustomization/README.md
@@ -1,0 +1,52 @@
+feast-kustomization contains open source feast-core kustomization and will contain feast-serving kustomization
+
+├── README.md
+├── base
+│   ├── feast-core
+│   │   ├── application.yml
+│   │   ├── deployment.yaml
+│   │   ├── kustomization.yaml
+│   │   └── service.yaml
+│   ├── kafka
+│   │   └── kustomization.yaml
+│   ├── kafka-rest-proxy
+│   │   ├── 10-schema-registry-deployment.yaml
+│   │   ├── 11-schema-registry-service.yaml
+│   │   ├── 20-rest-proxy.yaml
+│   │   ├── 21-rest-service.yaml
+│   │   └── kustomization.yaml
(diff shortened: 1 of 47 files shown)


terryyylim:
@ravdin Hope you don't mind, updated the comment to follow the template we use.

woop:
@ravdin I think you should be targeting Feast 0.7 here. There will be some large changes in the way deployments work from 0.8 onward, so I think we need to forward port these kustomizations explicitly. I also think that most teams running the current Feast will stay on 0.7 for a while.

woop:
@ravdin I dont think we have the bandwidth to support this config (Kustomize). I still think your PR here is great and do wish that people can actually get access to it. Would it be possible for you to push it to a repository? I can link to the repository as a community contrib installation from our docs.

PR #1019, opened 2020-09-29
Add Feature Tables API to Core & Python SDK

mrzzy:
Implements the Feature Table API as defined in the [0.8 RFC](https://docs.google.com/document/d/1snRxVb8ipWZjCiLlfkR4Oc28p7Fkv_UXjvxBFWjRBj4/edit#)
- Adds Feature Tables API to Core: `ApplyFeatureTable()`, `ListFeatureTables()`, `GetFeatureTable()`
- Add Feature Tables support to Python SDK: `apply()`, `list_feature_tables()`, `get_feature_table()` (credits to @terryyylim).

```release-note
Add Feature Table apply, list, get to Feast Core and Python SDK.
```


core/src/main/java/feast/core/grpc/CoreServiceImpl.java
```diff
@@ -391,4 +392,103 @@ public void listProjects(
...
+      // Check if user has authorization to apply feature table
+      authorizationService.authorizeRequest(SecurityContextHolder.getContext(), projectName);
+
+      ApplyFeatureTableResponse response = specService.applyFeatureTable(request);
+      responseObserver.onNext(response);
+      responseObserver.onCompleted();
+    } catch (AccessDeniedException e) {
+      log.info(
+          String.format(
+              "ApplyFeatureTable: wNot authorized to access project to apply: %s", projectName));
```
woop, line 413: Typo

core/src/main/java/feast/core/model/FeatureSource.java
```diff
@@ -0,0 +1,202 @@
...
+import javax.persistence.Table;
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.Setter;
+
+@Entity
+@Getter
+@Setter(AccessLevel.PRIVATE)
+@Table(name = "feature_sources")
+public class FeatureSource {
```
woop, line 44: This table is going to become crazy big if we keep it denormalized, but I guess its ok for now.

core/src/main/java/feast/core/model/FeatureTable.java
```diff
@@ -0,0 +1,323 @@
...
+  private long maxAgeSecs;
+
+  // Streaming Feature Source used to obtain data for features from a stream
+  @OneToOne(cascade = CascadeType.ALL)
+  @JoinColumn(name = "stream_source_id", nullable = true)
+  private FeatureSource streamSource;
+
+  // Batched Feature Source used to obtain data for features from a batch of data
+  @OneToOne(cascade = CascadeType.ALL)
+  @JoinColumn(name = "batch_source_id", nullable = true)
```
woop, line 102: Why is this nullable?

core/src/main/java/feast/core/model/FeatureTable.java
```diff
@@ -0,0 +1,323 @@
...
+  private FeatureSource batchSource;
+
+  // Autoincrementing version no. of this Feature Table.
+  // Autoincrements every update made to the feature set.
+  @Column(name = "revision", nullable = false)
+  private int revision;
```
woop, line 108: Why not set the default value here?

core/src/main/java/feast/core/model/FeatureTable.java
```diff
@@ -0,0 +1,323 @@
...
+
+  public FeatureTable() {
+    this.revision = 1;
+  };
+
+  /**
+   * Construct Feature from Protobuf spec representation in the given project with entities
```
woop, line 115: `Feature`?

core/src/main/java/feast/core/model/FeatureTable.java
```diff
@@ -0,0 +1,323 @@
...
+    table.setProject(new Project(projectName));
+
+    Set<FeatureV2> features =
+        spec.getFeaturesList().stream()
+            .map(featureSpec -> FeatureV2.fromProto(table, featureSpec))
+            .collect(Collectors.toSet());
+    table.setFeatures(features);
+
+    Set<EntityV2> entities =
+        FeatureTable.resolveEntities(
```
woop, line 137: I dont understand why we cant create an object straight from a proto. Why do we have to query the database? Shouldnt the object have enough unique information (keys)?

core/src/main/java/feast/core/model/FeatureTable.java
```diff
@@ -0,0 +1,323 @@
...
+            projectName, spec.getName(), entityRepo, spec.getEntitiesList());
+    table.setEntities(entities);
+
+    String labelsJSON = TypeConversion.convertMapToJsonString(spec.getLabelsMap());
+    table.setLabelsJSON(labelsJSON);
+
+    table.setMaxAgeSecs(spec.getMaxAge().getSeconds());
+
+    // Configure sources only if set
+    if (!spec.getStreamSource().equals(FeatureSourceSpec.getDefaultInstance())) {
```
woop, line 147: You always need a batch source.

core/src/main/java/feast/core/model/FeatureTable.java
```diff
@@ -0,0 +1,323 @@
...
+      table.setStreamSource(FeatureSource.fromProto(spec.getStreamSource()));
+    }
+    if (!spec.getBatchSource().equals(FeatureSourceSpec.getDefaultInstance())) {
```
woop, line 150: You always need a batch source.

core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -625,4 +634,110 @@ public UpdateStoreResponse updateStore(UpdateStoreRequest updateStoreRequest)
         .setStore(updateStoreRequest.getStore())
         .build();
   }
+
+  /**
+   * Applies the given Feature Table to the Feature Table registry. Creates the Feature Table if
+   * does not exist, otherwise updates the existing feature Table. Applies feature table in project
```
woop, line 640: Can we be consistent with casing.

core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -625,4 +634,110 @@ public UpdateStoreResponse updateStore(UpdateStoreRequest updateStoreRequest)
...
+   * if specified, otherwise in default project.
+   *
+   * @param request contain apply feature Table parameters.
```
woop, line 643: `contain apply feature Table parameters.` is quite poorly worded

sdk/python/feast/feature_source.py
```diff
@@ -0,0 +1,532 @@
...
+    """
+    FileOptions value type. Used to define source types in FeatureSource.
+    """
+
+    UNKNOWN = 0
+    PARQUET = 1
+    AVRO = 2
+
+
+class FeatureSourceSpecOptions:
```
woop, line 44: What is the point of this class?

terryyylim: Not much other than to state the need to create `to_proto` and `from_proto` method.

woop: Doesnt seem useful.

sdk/python/feast/feature_source.py
```diff
@@ -0,0 +1,558 @@
...
+
+    @project_id.setter
+    def project_id(self, project_id):
+        """
+        Sets the project id of this file
+        """
+        self._project_id = project_id
+
+    @property
+    def sql_query(self):
```
woop, line 161: This doesnt match the RFC I believe.

protos/feast/core/FeatureSource.proto
```diff
@@ -0,0 +1,111 @@
...
+    string file_url  = 2;
+  }
+
+  // Defines options for Feature source that sources features from a BigQuery Query
+  message BigQueryOptions {
+    // GCP Project ID
+    string project_id = 1;
+
+    // SQL Query used by BQ Source to query feature data.
+    string sql_query = 2;
```
khorshuheng, line 71: Spark BQ Connector doesn't support arbitrary sql query. So we should change this to table reference instead. You can refer to the updated Feast 0.8 RFC.

core/src/main/java/feast/core/model/FeatureSource.java
pyalex: Can we not flatten all the fields?
I'm for the option with kafkaConfig, bqConfig, fileConfig columns instead with json inside

core/src/main/java/feast/core/model/FeatureTable.java
```diff
@@ -0,0 +1,323 @@
...
+    if (!getName().equals(spec.getName())) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Updating the name of a registered FeatureTable is not allowed: %s to %s",
+              getName(), spec.getName()));
+    }
+    // - Entities cannot be changed
+    List<String> entityNames =
+        getEntities().stream().map(EntityV2::getName).collect(Collectors.toList());
+    if (!entityNames.containsAll(spec.getEntitiesList())) {
```
pyalex, line 175: does it check for deletion of entity?

core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -625,4 +634,110 @@ public UpdateStoreResponse updateStore(UpdateStoreRequest updateStoreRequest)
...
+      throw new IllegalArgumentException(
+          String.format(
+              "Cannot apply Feature Table to archived Project: (table: %s, project: %s)",
+              applySpec.getName(), projectName));
+    }
+
+    // Create or update depending on whether there is an existing Feature Table
+    Optional<FeatureTable> existingTable =
+        tableRepository.findFeatureTableByNameAndProject_Name(applySpec.getName(), projectName);
+    FeatureTable applyTable = FeatureTable.fromProto(projectName, applySpec, entityRepository);
```
pyalex, line 667: `applyTable` name seems confusing

core/src/main/java/feast/core/model/FeatureSource.java
woop: So each of those is a map?

protos/feast/core/CoreService.proto
```diff
@@ -432,4 +453,50 @@ message UpdateFeatureSetStatusRequest {
...
+    // Name of the FeatureTable to retrieve.
+    string name = 2;
+}
+
+message GetFeatureTableResponse {
+    // The Feature Table retrieved.
+    FeatureTable table = 1;
+}
+
+message ListFeatureTablesRequest {
```
pyalex, line 484: why we dropped wildcard search? it was used in JobController to retrieve all feature-sets by store subscription and I guess it could be useful for JobService as well if we won't drop subscription mechanism in stores.

pyalex: @woop will we support several online stores with different subscriptions in new architecture?

terryyylim: So you mean we should support single project, wildcard feature table name matches?

woop: We only have a single online store. There is no such thing as a store subscription any more.

woop: When we implement a `stream sink` in the future, then we will need to create some kind of `ingestion process` that hooks up to a stream based on configuration to update multiple online stores. But Feast doesnt know about these in the form of subscriptions. There is also no call to Feast Core to get this information.

core/src/main/java/feast/core/validators/FeatureTableValidator.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class FeatureTableValidator {
+  protected static final Set<String> RESERVED_NAMES =
+      Set.of("created_timestamp", "event_timestamp");
+
+  public static void validateSpec(FeatureTableSpec spec) {
```
pyalex, line 32: does it check for empty features/entities list?

core/src/main/java/feast/core/model/FeatureTable.java
terryyylim: By deletion, you mean from [entity1, entity2] to [entity1]?

core/src/main/java/feast/core/validators/FeatureTableValidator.java
terryyylim: Added check in same file.

core/src/main/java/feast/core/service/SpecService.java
terryyylim: Renamed to table.

protos/feast/core/FeatureSource.proto
terryyylim: Updated proto to use reference.

sdk/python/feast/feature_source.py
terryyylim: Updated proto to use reference.

sdk/python/feast/feature_source.py
terryyylim: Removed.

core/src/main/java/feast/core/service/SpecService.java
terryyylim: Re-worded the description.

core/src/main/java/feast/core/service/SpecService.java
terryyylim: Made casing consistent throughout with `FeatureTable`. Let me know if you spot anymore inconsistent casing.

core/src/main/java/feast/core/model/FeatureTable.java
terryyylim: Updated to make batch source compulsory.

core/src/main/java/feast/core/model/FeatureTable.java
terryyylim: Updated to `FeatureTable`.

core/src/main/java/feast/core/model/FeatureTable.java
terryyylim: Updated to make batch source compulsory.

core/src/main/java/feast/core/grpc/CoreServiceImpl.java
terryyylim: Fixed typo.

core/src/main/java/feast/core/model/FeatureTable.java
terryyylim: I think the use of `resolveEntities` is to first check that the entity we're trying to use exist, since the understanding is that an entity should exist prior to being used in a FeatureTable.

core/src/test/java/feast/core/model/FeatureTableTest.java
```diff
@@ -0,0 +1,139 @@
...
+import feast.proto.core.FeatureTableProto.FeatureTableSpec;
+import java.util.List;
+import java.util.Map;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+
+public class FeatureTableTest {
+  public static final String PROJECT_NAME = "project";
+  @Mock EntityRepository entityRepo;
```
pyalex, line 40: Please replace this test with IT. No more repository mocking

core/src/main/java/feast/core/model/FeatureTable.java
pyalex: yes

core/src/test/java/feast/core/model/FeatureTableTest.java
terryyylim: Replaced.

core/src/main/java/feast/core/model/FeatureTable.java
terryyylim: Good catch, it should have been `containsAll` both ways. Updated this check. Added a test [here](https://github.com/feast-dev/feast/pull/1019/files#diff-5dcec7c4d3f743c906004faffa9b5018R1121) as well.

core/src/main/java/feast/core/model/FeatureTable.java
terryyylim: Updated.

core/src/main/java/feast/core/model/FeatureSource.java
terryyylim: Updated to only using `config` column.

core/src/main/java/feast/core/model/DataSource.java
```diff
@@ -0,0 +1,192 @@
...
+  @Column(name = "config")
+  private String configJSON;
+
+  // Field mapping between sourced fields (key) and feature fields (value).
+  // Stored as serialized JSON string.
+  @Column(name = "field_mapping", columnDefinition = "text")
+  private String fieldMapJSON;
+
+  @Column(name = "ts_column")
+  private String tsColumn;
```
woop, line 66: should we make this `timestamp_column` maybe?

sdk/python/feast/data_source.py
```diff
@@ -0,0 +1,512 @@
...
+
+class DataSource:
+    """
+    DataSource that can be used source features
+    """
+
+    def __init__(
+        self,
+        type: str,
+        field_mapping: MutableMapping[str, str],
```
woop, line 356: why MutableMapping?

terryyylim: Updated to use Dict

core/src/main/java/feast/core/model/DataSource.java
terryyylim: Replace all ts_column/tsColumn with timestamp_column/timestampColumn respectively.

woop (approved):

PR #1021, opened 2020-09-30
Pyspark job for feature batch retrieval

khorshuheng:
This is the prerequisite to use PySpark for batch retrieval job instead of BQ Client / Feast Batch Serving.

This PR contains a standalone pyspark job (i.e. has not Feast or other external dependencies)  to perform batch feature retrieval. Feast users are not expected to use this pyspark script directly. Rather, Feast SDK will be responsible for submitting the pyspark job to a spark cluster, which will be implemented in a separate PR. 

```release-note
NONE
```


sdk/python/feast/pyspark/batch_retrieval_job.py
```diff
@@ -0,0 +1,203 @@
...
+from pyspark.sql import DataFrame, SparkSession, Window
+from pyspark.sql.functions import col, expr, monotonically_increasing_id, row_number
+
+
+def as_of_join(
+    entity: DataFrame,
+    join_keys: List[str],
+    feature_table: DataFrame,
+    features: List[str],
+    feature_prefix: str = "",
```
woop, line 14: Would you mind explaining this please?

sdk/python/feast/pyspark/batch_retrieval_job.py
```diff
@@ -0,0 +1,203 @@
...
+        features (List[str]):
+            The feature columns which should be present in the result dataframe.
+        feature_prefix (str):
+            Feature column prefix for the result dataframe.
+        max_age (str):
+            Tolerance for the feature event timestamp recency.
+
+    Returns:
+        DataFrame: Join result.
+
```
woop, line 42: I think we should add an example in the documentation

sdk/python/feast/pyspark/batch_retrieval_job.py
```diff
@@ -0,0 +1,203 @@
...
+def batch_retrieval(spark: SparkSession, conf: Dict) -> DataFrame:
+    """Retrieve batch features based on given configuration.
+
+    Args:
+        spark (SparkSession):
+            Spark session.
+        conf (Dict):
+            Configuration for the retrieval job, in json format. Sample configuration as follows:
+
+            sample_conf = {
```
woop, line 130: This can be moved into an example https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

sdk/python/feast/pyspark/batch_retrieval_job.py
```diff
@@ -0,0 +1,378 @@
...
+    entity: DataFrame,
+    join_keys: List[str],
+    feature_table: DataFrame,
+    features: List[str],
+    feature_prefix: str = "",
+    max_age: str = None,
+) -> DataFrame:
+    """Perform an as of join between entity and feature table, given a maximum age tolerance.
+    Join conditions:
+    1. Join keys values match.
```
woop, line 19: This sentence isnt clear.

woop, line 19: Should we talk about entities here?

sdk/python/feast/pyspark/batch_retrieval_job.py
```diff
@@ -0,0 +1,378 @@
...
+    2. Feature event timestamp is the closest match possible to the entity event timestamp,
+       but must not be more recent than the entity event timestamp, and the difference must
+       not be greater than max_age, unless max_age is not specified.
+    3. If more than one feature table rows satisfy condition 1 and 2, feature row with the
```
woop, line 23: What happens if there is no match?

sdk/python/feast/pyspark/batch_retrieval_job.py
```diff
@@ -0,0 +1,378 @@
...
+        entity (DataFrame):
+            Entity dataframe. Must contain the column event_timestamp.
+        join_keys (List[str]):
+            Column names for the join keys.
+        feature_table (DataFrame):
+            Feature table dataframe. Must contain the columns event_timestamp and created_timestamp.
+        features (List[str]):
+            The feature columns which should be present in the result dataframe.
+        feature_prefix (str):
+            Feature column prefix for the result dataframe.
```
woop, line 36: Can you explain WHY this would need to be used?

sdk/python/feast/pyspark/batch_retrieval_job.py
```diff
@@ -0,0 +1,378 @@
...
+
+class MissingColumnError(Exception):
+    pass
+
+
+class TimestampColumnError(Exception):
+    pass
+
+
+def verify_schema(
```
woop, line 135: Please add a comment here

sdk/python/feast/pyspark/batch_retrieval_job.py
```diff
@@ -0,0 +1,378 @@
...
+            raise MissingColumnError(
+                f"{timestamp_col} not found. Input columns: {', '.join(df.columns)}"
+            )
+        elif actual_dtypes[timestamp_col] != "timestamp":
+            raise TimestampColumnError(
+                f"{timestamp_col} is expected to be of timestamp type. Actual type: {actual_dtypes[timestamp_col]}"
+            )
+
+
+def join_entity_to_feature_tables(
```
woop, line 165: Please add a comment. Also, should all these methods be public?

sdk/python/feast/pyspark/batch_retrieval_job.py
```diff
@@ -0,0 +1,378 @@
...
+            query["join"],
+            tables[query["table"]],
+            query["features"],
+            feature_prefix=f"{query['table']}__",
+            max_age=query.get("max_age"),
+        )
+    return joined
+
+
+def batch_retrieval(spark: SparkSession, conf: Dict) -> DataFrame:
```
woop, line 245: Should we use a verb in this method `retrieve_historical_features`? We have moved on from the term `batch`

sdk/python/tests/data/column_mapping_test_entity.csv
```diff
@@ -0,0 +1,2 @@
+id,event_timestamp
+1001,2020-09-02T00:00:00.000
```
woop, line 2: Can we please add a few more rows, like 5-10. One row is asking for trouble.

sdk/python/feast/pyspark/batch_retrieval_job.py
```diff
@@ -0,0 +1,378 @@
...
+            Entity dataframe. Must contain the column event_timestamp.
+        join_keys (List[str]):
+            Column names for the join keys.
+        feature_table (DataFrame):
+            Feature table dataframe. Must contain the columns event_timestamp and created_timestamp.
+        features (List[str]):
+            The feature columns which should be present in the result dataframe.
+        feature_prefix (str):
+            Feature column prefix for the result dataframe.
+        max_age (str):
```
pyalex, line 37: why it's `str`? what's the format?

pyalex:
@khorshuheng I think we need to add minimum lower boundary
`min(event_timestamp) of all entities` - `max_age`
and apply this as default filter to feature_table. Otherwise there's no filters pushed to BQ, and we load too much unnecessary data

sdk/python/tests/data/column_mapping_test_entity.csv
pyalex: Even better to generate thousands to test how multiple partitions can affect result

sdk/python/tests/test_as_of_join.py
```diff
@@ -0,0 +1,670 @@
...
+        (2001, 8001, datetime(year=2020, month=9, day=2), 4.0, 4.5,),
+    ]
+    expected_joined_df = spark.createDataFrame(
+        spark.sparkContext.parallelize(expected_joined_data), expected_joined_schema
+    )
+
+    assert_dataframe_equal(joined_df, expected_joined_df)
+
+
+def test_entity_filter(
```
pyalex, line 336: Not sure what is tested here. Description would be helpful

sdk/python/tests/data/column_mapping_test_entity.csv
woop: Yes, using code to generate data isn't a bad idea, as long as we avoid too much `rand()`

sdk/python/feast/pyspark/batch_retrieval_job.py
khorshuheng: This is for the Pyspark `interval` expression. For example, 13 hour, 6 day, 60 second, 3 month. I can also change this to integer type instead of string, in which case the max age would be in seconds, similar to the current behaviour. Just thought that using a string might be more convenient for users, since they don't need to convert days / months to seconds, though that would be a breaking change. So should i use seconds and integer instead?

khorshuheng:
> @khorshuheng I think we need to add minimum lower boundary
> `min(event_timestamp) of all entities` - `max_age`
> and apply this as default filter to feature_table. Otherwise there's no filters pushed to BQ, and we load too much unnecessary data

While i agree with that, i am not sure how should the minimum lower boundary be specified. Should it be hard coded within the SDK? If not, where should we retrieve this lower boundary?

pyalex (approved):

PR #1022, opened 2020-09-30
Dataproc and Standalone Cluster Spark Job launcher

khorshuheng:
This PR depends on https://github.com/feast-dev/feast/pull/1021. Two types of Spark Job launchers have been included: Dataproc cluster and standalone cluster. Launchers for Yarn cluster, Kubernetes, Amazon EMR will be implemented at a later stage, in separate PRs.

Users are not expected to use these launchers directly. Instead, Feask SDK will use this launchers to submit batch feature retrieval job. This will be implemented in another PR.

Fixes #

```release-note
NONE
```


sdk/python/feast/pyspark/historical_feature_retrieval_job.py
```diff
@@ -270,14 +270,19 @@ def join_entity_to_feature_tables(
...
     """Retrieve batch features based on given configuration.
 
     Args:
         spark (SparkSession):
             Spark session.
         conf (Dict):
-            Configuration for the retrieval job, in json format. Sample configuration as follows:
+            Configuration for the retrieval job, in json format.
+        entity_df (DataFrame):
+            Optional. If provided, the entity will be used directly and conf["entity"] will be ignored.
```
woop, line 284: What does `entity will be used directly` mean. Not clear why `entity_df` can be optional.

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
```diff
@@ -412,10 +420,12 @@ def start_job(spark: SparkSession, conf: Dict):
...
-    spark = SparkSession.builder.appName("Batch Retrieval").getOrCreate()
-    spark.sparkContext.addFile("config.json")
-    config_file_path = SparkFiles.get("config.json")
-    with open(config_file_path, "r") as config_file:
-        conf = json.load(config_file)
-        start_job(spark, conf)
+    spark = SparkSession.builder.getOrCreate()
+    parser = argparse.ArgumentParser(description="Retrieval job arguments")
+    parser.add_argument(
+        "config_json", type=str, help="Configuration in json string format"
```
woop, line 426: So this is a json blob right?

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
khorshuheng: It's mainly to support the scenario below:
`client.get_historical_features_df(["rating"], entity_df)`

Where a user passed in a Spark dataframe / Pandas dataframe directly as entity, instead of using `FileSource` , `BigQuerySource`.

If we mandate user to always use `FileSource` or `BigQuerySource`, then we don't need this additional argument.

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
khorshuheng: Yes, it will be a json string, similar approach to the Spark jobs written by @pyalex . Originally i chose to use an actual config file, but it makes the job launcher much more complicated due to uploading and retrieval of configuration file.

pyalex: In case of `ingestion job` there're several json argument and I suggest you can split your config as well. It's just easier to work with job that has separated parameters, that one big json.
So in `ingestion-job` there are
`--feature-table` - json with project, name, entities, features
`--source` separate source to be able to run custom source
I assume in historical retrieval you would also need
`--entity-source`

Here's example of arguments for `ingestion job`
```

--source "{"kafka":{"bootstrapServers":"10.200.219.238:6668","topic":"supply-aggregator-sauron","classpath":"com.gojek.esb.aggregate.supply.AggregatedSupplyMessage","mapping":{},"timestampColumn":"event_timestamp"}}" 
--feature-table "{"name":"file","project":"default","entities":[{"name":"customer","type":"STRING"}],"features":[{"name":"feature1","type":"INT32"}]}" 
```

pyalex: I don't think that our jobs arguments will be completely the same. Historical job receives several `feature tables` for example. But at least we can try to keep it similar

khorshuheng: `--source separate source to be able to run custom source`

@pyalex How would the custom source applies in the context of historical feature retrieval? Is it something like a user input to override the Feature Table source registered on Feast Core?

khorshuheng: Or do you mean that the source of the feature tables should be it's own parameter (`--source`), rather than as a field under `--feature-table` ?

oavdeev: Just to throw in another idea, for feature_table and entity we could also pass the protobuf in JSON format. You'd still to have to manually parse it, but on the caller side it would be easier as we could replace [custom serialization code](https://github.com/feast-dev/feast/pull/1026/files#diff-7bc0d6431261a5482c3078341f05acf6R134-R156) with e.g. `json.dumps(MessageToDict(feature_table.to_proto()))`

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
woop: I dont understand why its part of this method. Doesnt the entity_df get uploaded prior to calling this method, meaning it should be configured through the conf?

sdk/python/feast/pyspark/launchers.py
```diff
@@ -0,0 +1,274 @@
...
+        )
+
+    def _stage_files(self, pyspark_script: str) -> str:
+        from google.cloud import storage
+
+        client = storage.Client()
+        bucket = client.get_bucket(self.staging_bucket)
+        blob_path = os.path.join(
+            self.remote_path,
+            "temp",
```
woop, line 249: why temp?

sdk/python/feast/pyspark/launchers.py
```diff
@@ -0,0 +1,274 @@
...
+        )
+        if scheme != "gs":
+            raise ValueError(
+                "Only GCS staging location is supported for DataprocLauncher."
+            )
+        self.project_id = project_id
+        self.region = region
+        self.job_client = dataproc_v1.JobControllerClient(
+            client_options={
+                "api_endpoint": "{}-dataproc.googleapis.com:443".format(region)
```
woop, line 238: can we use f-strings please

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
pyalex: @khorshuheng I just provided example of ingestion job args. It may be not very relevant to Historical directly. I'm just saying we can try to converge arguments format.

pyalex: @oavdeev we can move towards protobuf-like format. But it's already can't be 100% compatible. FeatureTable, for example, has entities as strings. Whereas most jobs need type as well. So some massaging would still be required

oavdeev: I see, didn't realize that. Let's just go with bespoke json then like in the ingestion job

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
```diff
@@ -1,19 +1,318 @@
...
+    """
+
+    format: str
+    path: str
+    timestamp_column: str
+    mapping: Dict[str, str] = {}
+    options: Dict[str, str] = {}
+
+
+class FeatureTableSource(NamedTuple):
```
woop, line 33: I don't understand why we have two sources, nor why `created_timestamp` is needed for the `FeatureTableSource`

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
```diff
@@ -1,19 +1,318 @@
...
+    Attributes:
+        name (str): Field name.
+        type (str): Feast type name.
+    """
+
+    name: str
+    type: str
+
+    @property
+    def spark_type(self):
```
woop, line 71: should this be `to_spark_type`?

khorshuheng: `spark_type` would be more natural, since the intended way to call them would be:

`>>> entity = Field(name="driver", type="int32")`
`>>> entity.type`
"int32"
`>>>  entity.spark_type`
"int"

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
```diff
@@ -1,19 +1,433 @@
...
+        mapped_entity_df = _map_column(df, source.mapping)
+
+        return EntityDataframe(
+            df=mapped_entity_df, timestamp_column=source.timestamp_column
+        )
+
+
+class FeatureTableDataframe(NamedTuple):
+    """
+    Feature table dataframe with specification.
```
woop, line 289: Can you add a comment here that describes this class. This comment says nearly nothing.

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
```diff
@@ -1,19 +1,433 @@
...
 from pyspark.sql import DataFrame, SparkSession, Window
 from pyspark.sql.functions import col, expr, monotonically_increasing_id, row_number
 
 
+class Source(abc.ABC):
+    """
+    Source for an entity or feature dataframe.
+
+    Attributes:
+        timestamp_column (str): Column representing the event timestamp.
```
woop, line 16: can we make this `event_timestamp_column`?

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
```diff
@@ -1,19 +1,433 @@
...
+        created_timestamp_column (str): Column representing the creation timestamp. Required
+            only if the source corresponds to a feature table.
+        mapping (Optional[Dict[str, str]]): If present, the source column will be renamed
```
woop, line 19: should this be `field_map`?

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
woop: Ok. Doesnt seem wrong per say, but is there a reason for this approach? Isn't it more pythonic to use getters and setters, like `entity.to_spark_type()`?

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
```diff
@@ -1,19 +1,433 @@
...
+    Attributes:
+        format (str): Output format.
+        path (str): Output uri.
+    """
+
+    format: str
+    path: str
+
+
+class EntityDataframe(NamedTuple):
```
woop, line 249: I dont understand the point of this class.

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
woop: I dont understand why we need this class

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
```diff
@@ -258,164 +621,158 @@ def join_entity_to_feature_tables(
...
+    feature_tables_sources: List[Source],
+    feature_tables: List[FeatureTable],
+) -> DataFrame:
+    """Retrieve historical features based on given configurations.
 
-        `options` is optional. If present, the options will be used when reading / writing the input / output.
+    Args:
+        spark (SparkSession): Spark session.
+        entity_source (Source): Entity data source.
+        feature_tables_sources (Source): List of feature tables data sources.
```
woop, line 651: Typo on source

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
```diff
@@ -258,164 +621,158 @@ def join_entity_to_feature_tables(
...
+        feature_tables (List[FeatureTable]): List of feature table specification.
+            The order of the feature table must correspond to that of feature_tables_sources.
 
-        `max_age` is in seconds, and determines the lower bound of the timestamp of the retrieved feature.
-        If not specified, this would be unbounded.
+    Returns:
+        DataFrame: Join result.
```
woop, line 656: is `join result` sufficient documentation?

sdk/python/feast/pyspark/launchers.py
```diff
@@ -0,0 +1,304 @@
...
+            SparkJobFailure: The spark job submission failed, encountered error
+                during execution, or timeout.
+
+        Returns:
+            str: file uri to the result file.
+        """
+        raise NotImplementedError
+
+
+class StandaloneCluster(JobLauncher):
```
woop, line 160: Should this be `StandaloneClusterJobLauncher`?

sdk/python/feast/pyspark/launchers.py
```diff
@@ -0,0 +1,304 @@
...
+        **kwargs,
+    ) -> RetrievalJob:
+        """
+        Submits a historical feature retrieval job to a Spark cluster.
+
+        Args:
+            pyspark_script (str): Local file path to the pyspark script for historical feature
+                retrieval.
+            entity_source_conf (List[Dict]): Entity data source configuration.
+            feature_tables_sources_conf (Dict): List of feature tables data sources configurations.
```
woop, line 144: Where have you documented where `feature_tables_sources_conf` will be used vs sources found in `feature_tables_conf`

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
khorshuheng: I am fine with making this `field_map`, but currently Oleksi's ingestion job is using `mapping` instead. Which one should i use then?

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
```diff
@@ -258,164 +621,158 @@ def join_entity_to_feature_tables(
...
+    entity_source: Source,
+    feature_tables_sources: List[Source],
+    feature_tables: List[FeatureTable],
+) -> DataFrame:
+    """Retrieve historical features based on given configurations.
 
-        `options` is optional. If present, the options will be used when reading / writing the input / output.
+    Args:
+        spark (SparkSession): Spark session.
+        entity_source (Source): Entity data source.
```
oavdeev, line 650: I don't think "entity data source" is a concept that has been used before in the codebase, maybe worth adding an explanation what it is. Or maybe call it `entity_df_source`? It is somewhat confusing that "entities" is used interchangeably for the entity objects themselves and the dataframe for point-in-time joins.

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
khorshuheng: This class contains information from both FeatureTable and Source, excluding the location of the feature table. In my original implementation, both timestamp columns and feature table specification are within the same level of configuration. But for the implementation in ingestion job, timestamp columns is in Source, whereas feature table specification is in FeatureTable.

I can see two choices here:
1. Eliminate FeatureTableDataframe, and pass both Source and FeatureTable into `as_of_join` and `join_entity_to_feature_tables`, eventhough the two methods don't exactly need to know the format and path of the Feature Table source.

2. Move timestamp columns to FeatureTable instead of Source, but that would mean the configuration is different from the Ingestion Job.

sdk/python/feast/pyspark/launchers.py
```diff
@@ -0,0 +1,304 @@
...
+        job_id: str,
+        **kwargs,
+    ) -> RetrievalJob:
+
+        submission_cmd = [
+            self.spark_submit_script_path,
+            "--master",
+            self.master_url,
+            "--name",
+            job_id,
```
oavdeev, line 200: Why is `job_id` externally configurable here? (as opposed to generating one inside JobLauncher)

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
khorshuheng: I prefer `event_timestamp_column` as well, though there are other instances in Feast SDK where `timestamp_column` is used instead (Datasource, BatchSource, for example).

I am happy to make the change though. Should i change it only for `historical_feature_retrieval_job`?

sdk/python/feast/pyspark/launchers.py
khorshuheng: Feast client will be the one generating the `job_id`, rather than user specified. As to why the job id is generated in Feast Client, rather than JobLauncher:

Prior to launching the job, JobLauncher will be doing some preparatory task, such as uploading Pandas dataframe to GCS / S3 (if user input is a pandas dataframe rather than a uri pointing to the file) , and define the output path based on Feast client configuration. In both cases, we can use job id as part of the GCS / S3 path name, which would be useful for tracking purposes.

That being said, i am fine with moving the job id generation within JobLauncher instead, if there is a good reason as to why it's better to have the job id generation within Job Launcher instead of Feast Client?

sdk/python/feast/pyspark/launchers.py
khorshuheng: Sources would not be part of `feature_tables_conf` , as per the design of configuration found here: https://github.com/feast-dev/feast/blob/master/spark/ingestion/src/main/scala/feast/ingestion/IngestionJobConfig.scala#L74-L79

Though, i understand that the above behaviour is actually inconsistent with the way we currently define our FeatureTable proto.

Should i revamp the configuration such that `Source` is part of Feature Table conf? In which case, it would not be necessary to have both `feature_tables_conf` and `feature_tables_sources_conf`.

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
woop: Lets just make the change here please.

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
woop: Mapping seems a bit vague. The type is a map, so its already a mapping. @pyalex

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
woop: > This class contains information from both FeatureTable and Source, excluding the location of the feature table. In my original implementation, both timestamp columns and feature table specification are within the same level of configuration. But for the implementation in ingestion job, timestamp columns is in Source, whereas feature table specification is in FeatureTable.

Ok, but it's not clear what this class is for. You're talking about your past implementations and implementation details but what problem are you trying to solve. This class is called FeatureTableDataframe, could it be replaced by using a FeatureTable and a DataFrame?


> I can see two choices here:
> 
> 1. Eliminate FeatureTableDataframe, and pass both Source and FeatureTable into `as_of_join` and `join_entity_to_feature_tables`, eventhough the two methods don't exactly need to know the format and path of the Feature Table source.
> 2. Move timestamp columns to FeatureTable instead of Source, but that would mean the configuration is different from the Ingestion Job.

You are presenting solutions but I dont know what problem you are trying to solve.

sdk/python/feast/pyspark/launchers.py
oavdeev: I see. I don't feel super strongly either way, but you could say that job id generation is a part of JobLauncher internal logic. That way the cloud-specific implementation would be able to use a specific jobid format for convenience. There might be cloud-specific limitations on job id length / charset that Feast Client is not aware of.

It may come in handy given that we don't plan to store jobs in a central database, they will have to be stored as some attribute or tag in dataproc/EMR job metadata itself.

sdk/python/feast/pyspark/launchers.py
oavdeev: Maybe we can tweak the format in a separate diff? Either way works, but above everything else i'd def prefer things to be consistent between this and @pyalex ingestion job.

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
khorshuheng: Feature table class doesn't contain timestamp column and created timestamp column. If it does, then yes, I don't need this class and I can easily replace this with a FeatureTable and a Dataframe, which is my preferred solution. The only reason why I haven't done this is because of the previous discussion where we want to use similar configuration for ingestion job.

The problem I am trying to solve here: I want  neither as_of_join method nor join_entity_features method to have any argument that contains file format and file path information, so that means I can't pass Source as an argument. But timestamp column information is required. Without creating a new class, that means my function arguments for as of join methods, will need to have list of Feature Table, List or dictionary of event timestamp columns, and the dataframes. The new class acts as a container that store all these methods together.

That being say, I don't like this class myself either. I will move the timestamp column to FeatureTable and remove FeatureTableDataframe.

woop (approved):

sdk/python/feast/pyspark/launchers.py
woop: Is this `job_id` the internal `job_id` or the external one? They might be different. The latter should be encapsulated, the former is internal to Feast and would be exposed to users. Or are we going to make them one and the same?

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
khorshuheng: Renamed to `field_mapping` instead because that's what being used in the protobuf and the other part of the Feast SDK right now.

PR #1023, opened 2020-10-05
Refactor Python SDK to remove v1 concepts

terryyylim:
In upcoming releases, we'll be slowly shifting towards a different architecture as defined in this [RFC](https://docs.google.com/document/d/1snRxVb8ipWZjCiLlfkR4Oc28p7Fkv_UXjvxBFWjRBj4/edit#). This PR is the first of a series that'll be related to cleaning up older concepts of Feast in the codebase.

The following changes would be included in subsequent PRs, likely when new retrieval method has been implemented.
- helm charts
- load test
- docker-compose test

Fixes #

```release-note
Python SDK `ingest` method now supports ingesting data from a batch source (parquet file/bigquery table) specified by the user, and written to the specified FeatureTable.
```


sdk/python/feast/client.py
```diff
@@ -368,9 +343,7 @@ def archive_project(self, project):
         if self._project == project:
             self._project = FEAST_DEFAULT_OPTIONS[CONFIG_PROJECT_KEY]
 
-    def apply_entity(
-        self, entities: Union[List[EntityV2], EntityV2], project: str = None
-    ):
+    def apply_entity(self, entities: Union[List[Entity], Entity], project: str = None):
```
woop, line 346: should this be `apply_entities`?

sdk/python/feast/client.py
```diff
@@ -977,197 +601,128 @@ def ingest(
...
             >>>            "driver": [1001],
             >>>            "rating": [4.3],
             >>>         }
             >>>     )
             >>> client.set_project("project1")
-            >>> client.ingest("driver", fs_df)
             >>>
-            >>> driver_fs = client.get_feature_set(name="driver", project="project1")
-            >>> client.ingest(driver_fs, fs_df)
+            >>> driver_ft = client.get_feature_table(name="driver")
```
woop, line 628: I think `get_feature_table("driver")` should also work.

sdk/python/feast/loaders/ingest.py
```diff
@@ -1,165 +1,39 @@
...
 FEAST_CORE_URL_ENV_KEY = "FEAST_CORE_URL"  # type: str
 BATCH_FEATURE_REQUEST_WAIT_TIME_SECONDS = 300
-KAFKA_CHUNK_PRODUCTION_TIMEOUT = 120  # type: int
+BATCH_INGESTION_PRODUCTION_TIMEOUT = 120  # type: int
 
 
-def _encode_pa_tables(
-    file: str, feature_set: str, fields: dict, ingestion_id: str, row_group_idx: int
-) -> List[bytes]:
+def check_field_mappings(
```
woop, line 11: should this be private?

sdk/python/feast/client.py
terryyylim: If we change this, then we should also change `apply_feature_table`. Do we prefer plural?

sdk/python/feast/client.py
terryyylim: Yup, that works. I'll update the example.

tests/e2e/redis/parallel-ingest-redis-serving.py
```diff
@@ -0,0 +1,185 @@
...
+            description="Driver entity for car rides",
+            value_type=ValueType.STRING,
+            labels={"team": "matchmaking", "common_key": "common_val"},
+        )
+
+        cls.basic_ft_spec = FeatureTable.from_yaml(
+            f"{DIR_PATH}/specifications/dev_ft.yaml"
+        )
+
+    def test_discovery(self, client):
```
woop, line 57: Discovery as a name is overloaded. Can we be more specific on what we are testing?

tests/e2e/redis/parallel-ingest-redis-serving.py
```diff
@@ -0,0 +1,185 @@
...
+        # TODO: Add ingest and retrieval check
+        pass
+
+
+@pytest.mark.incremental
+class TestAllTypesIngestionRetrieval:
+    def setup_class(cls):
+        prefix = "alltypes_ingestion"
+        suffix = str(int(datetime.now().timestamp()))
+        batch_source = DataSource(
```
woop, line 101: Is this setup the way you think users of Feast will end up using the API?

terryyylim: This way, we can isolate the objects we need for this test suite in the `setup_class` method. Using classes also allow us to more easily plan which tests to run in sequence and which tests to run in parallel.

woop: The API should match commands that users are running. Is there any reason we cant just use a single big function if we want to run it in sequence?

woop: Also, `SourceType(1).name` doesnt look like something we would expose to users.

tests/e2e/redis/parallel-ingest-redis-serving.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def test_basic_retrieval(self, client):
+        # TODO: Add ingest and retrieval check
+        pass
+
+
+@pytest.mark.incremental
+class TestAllTypesIngestionRetrieval:
+    def setup_class(cls):
+        prefix = "alltypes_ingestion"
```
woop, line 99: What is this prefix for?

tests/e2e/redis/parallel-ingest-redis-serving.py
```diff
@@ -0,0 +1,185 @@
...
+            value_type=ValueType.STRING,
+            labels={"cat": "alltypes"},
+        )
+
+        cls.alltypes_ft_name = f"{prefix}_alltypes_{suffix}"
+        cls.alltypes_ft_spec = FeatureTable(
+            name="alltypes",
+            entities=["alltypes_id"],
+            features=[
+                Feature(name="float_feature", dtype=ValueType.FLOAT).to_proto(),
```
woop, line 124: Why do we have to convert to_proto()? are users expected to do this as well?

tests/e2e/redis/parallel-ingest-redis-serving.py
```diff
@@ -0,0 +1,185 @@
...
+
+        actual_common_entities = client.list_entities(labels=common_filtering_labels)
+        actual_matchmaking_entities = client.list_entities(
+            labels=matchmaking_filtering_labels
+        )
+        assert len(actual_common_entities) == 2
+        assert len(actual_matchmaking_entities) == 1
+
+        # ApplyFeatureTable
+        client.apply_feature_table(self.basic_ft_spec, PROJECT_NAME)
```
woop, line 79: Do we need to use the PROJECT_NAME again here if its already in the client?

sdk/python/tests/test_client.py
```diff
@@ -76,7 +56,6 @@
 
 CORE_URL = "core.feast.example.com"
```
woop, line 57: do we need this CORE_URL?

sdk/python/tests/test_client.py
```diff
@@ -495,516 +534,102 @@ def test_list_feature_sets(self, mocked_client, mocker):
...
+            field_mapping={
+                "ride_distance": "ride_distance",
+                "ride_duration": "ride_duration",
+            },
+            options=FileOptions(file_format="avro", file_url="data/test.avro"),
+            timestamp_column="ts_col",
+            date_partition_column="date_partition_col",
+        )
+
+        stream_source = DataSource(
```
woop, line 582: We should be using specific sources like `FileSource` here and not the underlying base class. Otherwise the API is too low level.

terryyylim: In protos, we have the different DataSources identified via its options [here](https://github.com/feast-dev/feast/blob/master/protos/feast/core/DataSource.proto#L97). Thus, in the sdk, I only exposed DataSource, and let the different options be configurable via native classes `FileOptions`, `BigQueryOptions`, `KafkaOptions`, `KinesisOptions`.

terryyylim: Is our preference here to create native classes `FileSource`, `BigQuerySource`, `KafkaSource` and `KinesisSource`? I felt that following how protos were defined would be a more standardized approach for the codebase.

sdk/python/tests/test_client.py
terryyylim: Yes, it's required to instantiate the Client.

tests/e2e/redis/parallel-ingest-redis-serving.py
terryyylim: Removed.

tests/e2e/redis/parallel-ingest-redis-serving.py
terryyylim: Removed.

tests/e2e/redis/parallel-ingest-redis-serving.py
terryyylim: Removed `SourceType(1).name`.

Refactored pytests to use fixtures instead of class-based approach.

tests/e2e/redis/parallel-ingest-redis-serving.py
terryyylim: Renamed the tests.

sdk/python/feast/loaders/ingest.py
terryyylim: Updated.

infra/scripts/test-docker-compose.sh
```diff
@@ -63,4 +63,4 @@ export FEAST_ONLINE_SERVING_CONTAINER_IP_ADDRESS=$(docker inspect -f '{{range .N
 ${PROJECT_ROOT_DIR}/infra/scripts/wait-for-it.sh ${FEAST_ONLINE_SERVING_CONTAINER_IP_ADDRESS}:6566 --timeout=120
 
 # Run e2e tests for Redis
-docker exec feast_jupyter_1 bash -c 'cd /feast/tests/e2e/redis && pytest --verbose -rs basic-ingest-redis-serving.py --core_url core:6565 --serving_url=online_serving:6566 --jobcontroller_url=jobcontroller:6570 --kafka_brokers=kafka:9092'
+docker exec feast_jupyter_1 bash -c 'cd /feast/tests/e2e/redis && pytest -n 1 --dist=loadscope parallel-ingest-redis-serving.py --core_url core:6565 --serving_url=online_serving:6566 --kafka_brokers=kafka:9092'
```
woop, line 66: Should this still be testing redis?

infra/scripts/test-end-to-end-redis-cluster.sh
```diff
@@ -101,7 +101,8 @@ ORIGINAL_DIR=$(pwd)
 cd tests/e2e
 
 set +e
-pytest redis/* --junitxml=${LOGS_ARTIFACT_PATH}/python-sdk-test-report.xml
+CORE_NO=$(nproc --all)
+pytest src/* -n ${CORE_NO} --dist=loadscope --junitxml=${LOGS_ARTIFACT_PATH}/python-sdk-test-report.xml
```
woop, line 105: wait, what does this test?

sdk/python/feast/cli.py
```diff
@@ -217,6 +212,21 @@ def feature_table():
     pass
 
 
+def _get_labels_dict(label_str: str) -> Dict[str, str]:
+    """
+    Converts CLI input labels string to dictionary format if provided string is valid.
```
woop, line 217: Please add full docstring.

sdk/python/feast/client.py
```diff
@@ -605,370 +578,21 @@ def get_feature_table(
...
         chunk_size: int = 10000,
         max_workers: int = max(CPU_COUNT - 1, 1),
-        disable_progress_bar: bool = False,
-        timeout: int = KAFKA_CHUNK_PRODUCTION_TIMEOUT,
-    ) -> str:
+        timeout: int = BATCH_INGESTION_PRODUCTION_TIMEOUT,
+    ) -> None:
         """
-        Loads feature data into Feast for a specific feature set.
+        Batch load feature data into batch source of a specific feature table.
```
woop, line 591: We should just say `into a Feature Table`. In the future it would go into the offline store.

sdk/python/feast/client.py
```diff
@@ -977,197 +601,128 @@ def ingest(
...
-                project = "default"
-            name = feature_set
-        else:
-            raise Exception("Feature set name must be provided")
+        if project is None:
+            project = self.project
+        if isinstance(feature_table, FeatureTable):
+            name = feature_table.name
 
         # Read table and get row count
```
woop, line 637: Should this be removed?

sdk/python/feast/client.py
```diff
@@ -977,197 +601,128 @@ def ingest(
...
-
-        row_count = pq_file.metadata.num_rows
+        dir_path, dest_path, column_names = _read_table_from_source(
+            source, chunk_size, max_workers
+        )
 
         current_time = time.time()
-
-        print("Waiting for feature set to be ready for ingestion...")
+        print("Waiting for feature table to be ready for ingestion...")
```
woop, line 643: What are you waiting for here? There is no status on a feature table

sdk/python/feast/loaders/ingest.py
```diff
@@ -1,165 +1,39 @@
...
-    proto_items = proto_columns.items()
-    append = feature_rows.append
-
-    # Iterate through the rows
-    for row_idx in range(table.num_rows):
-        feature_row = FeatureRow(
-            event_timestamp=datetime_col[row_idx],
-            feature_set=feature_set,
-            ingestion_id=ingestion_id,
+    if "datetime" not in column_names:
```
woop, line 26: Should this be hardcoded? Dont they set the timestamp column themselves?

sdk/python/feast/loaders/ingest.py
```diff
@@ -1,165 +1,39 @@
...
-
-        # Insert field from each column
-        for k, v in proto_items:
-            ext([field(name=k, value=v[row_idx])])
-
-        # Append FeatureRow in byte string form
-        append(feature_row.SerializeToString())
 
-    return feature_rows
+    specified_field_mappings = [v for k, v in feature_table_field_mappings.items()]
```
woop, line 31: Hard to read line of code.

sdk/python/feast/client.py
```diff
@@ -977,197 +601,128 @@ def ingest(
...
-        ingestion_ids: Optional[List[str]] = None,
-        start_date: Optional[datetime.datetime] = None,
-        end_date: Optional[datetime.datetime] = None,
-        force_refresh: bool = False,
-        project: Optional[str] = None,
-    ) -> statistics_pb2.DatasetFeatureStatisticsList:
-        """
-        Retrieves the feature featureStatistics computed over the data in the batch
-        stores.
+        # Check 2) Check if FeatureTable batch source field mappings can be found in provided source table
```
woop, line 672: So does this support not providing field mappings, or only providing some field mappings?

sdk/python/feast/loaders/ingest.py
terryyylim: You mean the user themselves should set the timestamp column?

sdk/python/feast/client.py
terryyylim: Not at the moment, it only succeeds if all field mappings are provided in the dataset.

sdk/python/tests/test_client.py
```diff
@@ -324,83 +298,138 @@ def test_version(self, mocked_client, mocker):
...
         )
-        recieve_response = GetOnlineFeaturesResponse()
-        entity_rows = []
-        for row_number in range(1, ROW_COUNT + 1):
-            request.entity_rows.append(
-                GetOnlineFeaturesRequest.EntityRow(
-                    fields={"driver_id": int_val(row_number)}
-                )
+        mocked_client.set_project("my_project")
+        entity = mocked_client.get_entity("my_entity")
```
woop, line 340: What are we testing here? All these tests seem like they set the response and then run the method that gets that response.

sdk/python/tests/test_client.py
```diff
@@ -409,11 +438,11 @@ def test_get_feature_set(self, mocked_client, mocker):
 
         mocker.patch.object(
             mocked_client._core_service_stub,
-            "GetFeatureSet",
-            return_value=GetFeatureSetResponse(
-                feature_set=FeatureSetProto(
-                    spec=FeatureSetSpecProto(
-                        name="my_feature_set",
+            "GetFeatureTable",
```
woop, line 441: Are we testing anything here?

sdk/python/feast/loaders/ingest.py
pyalex: I guess you can use `timestampColumn` from FeatureTable spec

pyalex: The only thing we may wanna add `date_partition_column` ourselves here as
`df[<date_partition_column>] = to_date(df[<timestampColumn>])`
if it's specified and then use this column for partitioning

sdk/python/feast/client.py
terryyylim: Removed.

sdk/python/feast/client.py
terryyylim: Removed.

sdk/python/feast/client.py

sdk/python/feast/cli.py
terryyylim: Updated.

infra/scripts/test-end-to-end-redis-cluster.sh
terryyylim: Used the wrong path, updated.

infra/scripts/test-docker-compose.sh
terryyylim: Used the wrong path, updated.

sdk/python/feast/loaders/ingest.py
terryyylim: Removed list comprehension.

sdk/python/tests/test_client.py
woop: The `Source` based approach (as in the RFC) is a more abstracted view. It allows us to define a source specific constructor. So they dont have to define the type, we can define that. We can also have input arguments that are specific to each source, and we dont require them to define `FileOptions` which is basically the same as having `FileSource` at the end of the day. So its better to just have a single constructor instead of two.

sdk/python/feast/client.py
```diff
@@ -977,197 +601,126 @@ def ingest(
...
 
-        row_count = pq_file.metadata.num_rows
+        dir_path, dest_path, column_names = _read_table_from_source(
+            source, chunk_size, max_workers
+        )
 
         current_time = time.time()
-
-        print("Waiting for feature set to be ready for ingestion...")
         while True:
```
pyalex, line 642: do we still need this loop? what are we waiting?

sdk/python/feast/loaders/ingest.py
terryyylim: You mean initialize `df[<date_partition_column>]` column if `date_partition_column` is specified in the FeatureTableSpec but not found in the dataframe?

sdk/python/feast/client.py
```diff
@@ -1181,85 +734,9 @@ def _get_grpc_metadata(self):
...
-        feature_set: Feature set of the dataset to be ingested.
-
-    Returns:
-        UUID unique to current time and the feature set provided.
-    """
-    uuid_str = f"{feature_set.name}_{int(time.time())}"
-    return str(uuid.uuid3(uuid.NAMESPACE_DNS, uuid_str))
-
-
 def _read_table_from_source(
```
pyalex, line 737: Should write data partitioned by `date_partition_column` if it's specified.

sdk/python/feast/client.py
```diff
@@ -977,197 +601,126 @@ def ingest(
...
-        if ingestion_ids is not None:
-            request.ingestion_ids.extend(ingestion_ids)
-        else:
-            if start_date is not None:
-                request.start_date.CopyFrom(
-                    Timestamp(seconds=int(start_date.timestamp()))
+                staging_client.upload_file(
+                    dest_path,
+                    uri.hostname,
+                    str(uri.path).strip("/") + "/" + f"date={date_today}/" + file_name,
```
pyalex, line 691: we should partitioned by `date(event_timestamp)` instead of today.
`eventTimestamp` and `datePartitioned` columns are specified in FeatureTable.
With this partitioning it could several directories that we need to upload.

sdk/python/tests/test_client.py
terryyylim: Yup, its testing that the method returns the expected response type and value. Actually, most of these flow are already tested in the e2e tests. Previously we have all these unit tests and e2e tests, but they essentially test the same flow. Should we just keep e2e tests in these cases?

sdk/python/feast/client.py
```diff
@@ -977,197 +601,126 @@ def ingest(
...
                 )
-            if end_date is not None:
-                request.end_date.CopyFrom(Timestamp(seconds=int(end_date.timestamp())))
+            if batch_source_type == "BATCH_BIGQUERY":
+                from google.cloud import bigquery
+
+                bq_table_ref = feature_table.batch_source.bigquery_options.table_ref
+                gcp_project, dataset_table = bq_table_ref.split(":")
```
pyalex, line 697: There should a parser for that inside bigquery sdk. Do we need to implement it ourselves?

sdk/python/feast/client.py
```diff
@@ -977,197 +601,126 @@ def ingest(
...
+                client = bigquery.Client(project=gcp_project)
+
+                table_ref = client.dataset(dataset).table(table)
+                job_config = bigquery.LoadJobConfig()
+                job_config.source_format = bigquery.SourceFormat.PARQUET
+
+                # Check for date partitioning column in FeatureTable spec
+                if feature_table.batch_source.date_partition_column:
+                    time_partitioning_obj = bigquery.table.TimePartitioning(
+                        field=feature_table.batch_source.date_partition_column
```
pyalex, line 709: `date_partition_column` most probably would be specified only for files. Since BQ is smart enough to partition on timestamp column

pyalex: so we can use eventTimestamp here instead

sdk/python/tests/test_client.py
```diff
@@ -495,516 +534,100 @@ def test_list_feature_sets(self, mocked_client, mocker):
...
-        [
-            (dataframes.BAD_NO_DATETIME, Exception, lazy_fixture("client")),
-            (
-                dataframes.BAD_INCORRECT_DATETIME_TYPE,
-                Exception,
-                lazy_fixture("client"),
+        # Create Feature Tables
+        batch_source = DataSource(
+            type="BATCH_FILE",
+            field_mapping={
```
woop, line 573: What is the point of this field mapping?

tests/e2e/test-register-ingest.py
```diff
@@ -0,0 +1,158 @@
+import os
```
woop, line 1: Where is the ingestion (name of the file)?

tests/e2e/test-register-ingest.py
```diff
@@ -0,0 +1,158 @@
...
+            Feature(name="bool_list_feature", dtype=ValueType.BOOL_LIST),
+        ],
+        max_age=Duration(seconds=3600),
+        batch_source=batch_source,
+        labels={"cat": "alltypes"},
+    )
+
+
+def test_get_list_basic(client: Client, customer_entity: Entity, driver_entity: Entity):
+    basic_ft_spec = FeatureTable.from_yaml(f"{DIR_PATH}/specifications/dev_ft.yaml")
```
woop, line 98: I think a better test would be `from_yaml` -> register -> get feature table -> to_yaml -> compare yamls

I also think that using `from_yaml` here makes the test a bit brittle. I would prefer that we move all of the test logic into Python code unless we want to test the yaml functionality itself.

sdk/python/tests/test_client.py
terryyylim: Added as field mapping is compulsory parameter.

tests/e2e/test-register-ingest.py
terryyylim: Renamed.

sdk/python/feast/client.py
terryyylim: Removed.

sdk/python/tests/test_client.py
terryyylim: Agreed that would be cleaner. I've updated the FeatureTable class and only expose `FileSource`, `BigQuerySource`, `KafkaSource` and `KinesisSource` instead of the respective options native classes.

sdk/python/feast/client.py
terryyylim: Updated.

sdk/python/feast/client.py

sdk/python/feast/client.py
terryyylim: Updated.

sdk/python/feast/client.py

tests/e2e/test-register.py
```diff
@@ -0,0 +1,203 @@
...
+        description="Driver entity for car rides",
+        value_type=ValueType.STRING,
+        labels={"team": "matchmaking", "common_key": "common_val"},
+    )
+
+
+@pytest.fixture
+def basic_featuretable():
+    batch_source = FileSource(
+        type="BATCH_FILE",
```
woop, line 53: Why do we need a type here?

sdk/python/feast/client.py
```diff
@@ -977,197 +601,131 @@ def ingest(
...
+                    file_paths += [os.path.join(dirpath, file) for file in filenames]
+                for path in file_paths:
+                    file_name = path.split("/")[-1]
+                    partition_date = path.split("/")[-2].split("=")[-1]
+                    staging_client.upload_file(
+                        path,
+                        uri.hostname,
+                        str(uri.path).strip("/")
+                        + "/"
+                        + f"date={partition_date}/"
```
pyalex, line 696: parquet structure should be used instead

tests/e2e/test-register.py
terryyylim: Removed.

sdk/python/tests/test_client.py
terryyylim: Removed.

sdk/python/tests/test_client.py
terryyylim: Removed.

sdk/python/feast/feature_table.py
```diff
@@ -153,7 +166,7 @@ def stream_source(self):
         return self._stream_source
 
     @stream_source.setter
-    def stream_source(self, stream_source: DataSource):
+    def stream_source(self, stream_source: DataSourceProto):
```
woop, line 169: Why does this take a proto?

sdk/python/feast/feature_table.py
```diff
@@ -43,18 +44,28 @@ class FeatureTable:
...
+        stream_source: Optional[Union[KafkaSource, KinesisSource]] = None,
         max_age: Optional[Duration] = None,
         labels: Optional[MutableMapping[str, str]] = None,
     ):
         self._name = name
         self._entities = entities
-        self._features = features
-        self._batch_source = batch_source
-        self._stream_source = stream_source
+        self._features = [
```
woop, line 56: Can you explain what is happening here?

sdk/python/feast/data_source.py
```diff
@@ -437,74 +405,211 @@ def date_partition_column(self, date_partition_column):
...
-            )
+
+class FileSource(DataSource):
+    def __init__(
+        self,
+        field_mapping,
+        timestamp_column,
+        file_format,
+        file_url,
+        date_partition_column="",
```
woop, line 426: Why is this set to `""`?

sdk/python/feast/feature_table.py
```diff
@@ -254,54 +267,50 @@ def _to_data_source(cls, data_source):
         Convert dict to data source.
```
woop, line 267: This method is mysterious. Can you please add documentation so that its clear what its purpose is. By looking at the name I would expect it to convert a feature table into a data source.

sdk/python/feast/feature_table.py
terryyylim: Yup, so when generating FeatureTableSpec in `to_proto` method, it requires `self.features` in proto form, this is because we're expecting `FeatureSpecV2` object as defined in FeatureTable.proto file.

sdk/python/feast/feature_table.py
terryyylim: Similar to expecting FeatureSpecV2, we're expecting DataSourceProto when generating FeatureTableSpec in `to_proto` method. This allows the user to supply just the native DataSource class when using the SDK.

sdk/python/feast/feature_table.py
terryyylim: Added more description and renamed the function.

sdk/python/feast/data_source.py
terryyylim: It is an optional field, added typings to the respective native classes.

sdk/python/feast/feature_table.py
terryyylim: I've cleaned this up, to shift conversion inside `to_proto` method, this makes typings and initialization of class object code alot cleaner.

sdk/python/feast/feature_table.py
terryyylim: I've cleaned this up, to shift conversion inside `to_proto` method, this makes typings and initialization of class object code alot cleaner.

sdk/python/feast/client.py
```diff
@@ -605,370 +578,21 @@ def get_feature_table(
...
-                ),
-                metadata=self._get_grpc_metadata(),
-            )
-        except grpc.RpcError as e:
-            raise grpc.RpcError(e.details())
-
-        response = OnlineResponse(response)
-        return response
-
     def ingest(
```
pyalex, line 581: Please split this function in parts

sdk/python/feast/client.py
```diff
@@ -977,197 +601,145 @@ def ingest(
...
+            if issubclass(type(feature_table.batch_source), FileSource):
+                from urllib.parse import urlparse
+
+                file_url = feature_table.batch_source.file_options.file_url[:-1]
+                uri = urlparse(file_url)
+                staging_client = get_staging_client(uri.scheme)
+
+                if date_partition_dest_path is not None:
+                    file_paths = list()
+                    for (dirpath, dirnames, filenames) in os.walk(
```
pyalex, line 690: It could be one loop
```
for path in glob.glob(date_partition_dest_path + "**/*"):
   upload path

File: sdk/python/feast/client.py (hunk 23 above)
Comment (pyalex) [reply sdk/python/feast/client.py:756]: Sorry, I wasn't clear enough.
So current flow is:
1. Read from file / pandas
2. Write to parquet
3. Read from parquet
4. Write to parquet with partitions

Shouldn't it be only 1 and 4 when partitioning column specified?
Can all this flow be in one function?

File: sdk/python/feast/client.py [hunk 38]
```diff
@@ -977,197 +601,145 @@ def ingest(
...
-            producer.print_results()
-        finally:
-            # Remove parquet file(s) that were created earlier
-            print("Removing temporary file(s)...")
-            shutil.rmtree(dir_path)
+                bq_table_ref = feature_table.batch_source.bigquery_options.table_ref
+                gcp_project, dataset_table = bq_table_ref.split(":")
 
-        return ingestion_id
+                client = bigquery.Client(project=gcp_project)
```
pyalex, line 721: bq_client? too many clients

sdk/python/tests/test_client.py
```diff
@@ -324,687 +299,182 @@ def test_version(self, mocked_client, mocker):
...
-            lazy_fixture("secure_mock_client_with_auth"),
-        ],
+        "mocked_client", [lazy_fixture("mock_client")],
     )
-    def test_get_historical_features(self, mocked_client, mocker):
-
-        mocked_client._serving_service_stub = Serving.ServingServiceStub(
-            grpc.insecure_channel("")
-        )
+    def test_ingest(self, mocked_client, mocker):
```
pyalex, line 403: can we please have all cases coverage?
w / wo partition column
pandas / file input
file / bq source

sdk/python/tests/test_client.py
```diff
@@ -324,687 +299,182 @@ def test_version(self, mocked_client, mocker):
...
         mocked_client._core_service_stub = Core.CoreServiceStub(
             grpc.insecure_channel("")
         )
 
+        N_ROWS = 100
```
pyalex, line 408: This data generation should be moved out from test

sdk/python/feast/__init__.py
```diff
@@ -16,9 +23,13 @@
 __all__ = [
     "Client",
     "Entity",
-    "Feature",
-    "FeatureSet",
-    "Source",
+    "DataSource",
```
woop, line 26: Why is DataSource exported?

sdk/python/feast/feature_table.py
```diff
@@ -80,7 +80,9 @@ def __eq__(self, other):
         ):
             return False
 
-        if self.entities != other.entities:
+        if sorted(self.entities) != sorted(other.entities):
+            return False
+        if self.features != other.features:
```
woop, line 85: Why do we sort the one list but not the other?

sdk/python/feast/client.py
pyalex: OK, I agree, we already have big functions that do a lot of stuff.
So what about? 
_read_table_from_source() -> returns Dataframe
if file:
     _write_to_parquet(df, timestamp_column, date_partitioning_column=None) -> returns directory
     recursively copy directory
if bq:
    load table from dataframe

sdk/python/feast/feature_table.py
```diff
@@ -333,12 +335,12 @@ def from_proto(cls, feature_table_proto: FeatureTableProto):
             batch_source=(
                 None
                 if not feature_table_proto.spec.batch_source.ByteSize()
```
woop, line 337: What ByteSize?

sdk/python/feast/feature_table.py
```diff
@@ -362,11 +364,23 @@ def to_proto(self) -> FeatureTableProto:
         spec = FeatureTableSpecProto(
             name=self.name,
             entities=self.entities,
-            features=self.features,
+            features=[
+                feature.to_proto()
+                for feature in self.features
+                if type(feature) == Feature
```
woop, line 370: What happens if this if condition fails?

sdk/python/feast/feature_table.py
terryyylim: Removed for batch_source, only kept for stream_source since it's optional.

sdk/python/feast/client.py
terryyylim: Current flow:

If FileSource and partition:
  1. Read from file/pandas
  2. Write to parquet with partitions
Else:
  1. Read from file / pandas
  2. Write to parquet

sdk/python/feast/feature_table.py
```diff
@@ -43,10 +43,10 @@ class FeatureTable:
     def __init__(
         self,
         name: str,
-        entities: Union[str, List[str]],
-        features: Union[FeatureV2, List[FeatureV2]],
-        batch_source: Optional[DataSource] = None,
-        stream_source: Optional[DataSource] = None,
+        entities: List[str],
```
woop, line 46: Shouldnt this be `Entity`?

woop (approved):

PR #1037, opened 2020-10-11
Return all tables if project unspecified

terryyylim:
Currently, in Serving, [when caching Feature sets](https://github.com/feast-dev/feast/blob/master/serving/src/main/java/feast/serving/specs/CachedSpecService.java#L239), we allow wildcard match to return all Feature sets spec in one go. Likewise, as we update Serving to handle Feature tables, we'll require an easy way to retrieve all Feature tables. 

When project is not specified in list Feature tables call, all Feature tables will be returned.

Fixes #

```release-note
NONE
```

### DIFF
--- a/core/src/main/java/feast/core/service/SpecService.java
+++ b/core/src/main/java/feast/core/service/SpecService.java
@@ -682,19 +682,28 @@ public class SpecService {
 
   /**
    * List the FeatureTables matching the filter in the given filter. Scopes down listing to project
-   * if specified, the default project otherwise.
+   * if specified, otherwise all FeatureTables will be returned.
    *
    * @param filter Filter containing the desired project and labels
    * @return ListFeatureTablesResponse with list of FeatureTables found matching the filter
    */
   @Transactional
   public ListFeatureTablesResponse listFeatureTables(ListFeatureTablesRequest.Filter filter) {
-    String projectName = resolveProjectName(filter.getProject());
+    String projectName = filter.getProject();
     Map<String, String> labelsFilter = filter.getLabelsMap();
 
-    checkValidCharacters(projectName, "project");
+    if (projectName.isEmpty()) {
+      projectName = "*";
+    }
 
-    List<FeatureTable> matchingTables = tableRepository.findAllByProject_Name(projectName);
+    List<FeatureTable> matchingTables;
+    if (projectName.equals("*")) {
+      matchingTables =
+          tableRepository.findAllByProject_NameLikeOrderByNameAsc(projectName.replace('*', '%'));
+    } else {
+      checkValidCharacters(projectName, "project");
+      matchingTables = tableRepository.findAllByProject_Name(projectName);
+    }
 
     ListFeatureTablesResponse.Builder response = ListFeatureTablesResponse.newBuilder();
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -682,19 +682,28 @@ public ApplyFeatureTableResponse applyFeatureTable(ApplyFeatureTableRequest requ
...
    */
   @Transactional
   public ListFeatureTablesResponse listFeatureTables(ListFeatureTablesRequest.Filter filter) {
-    String projectName = resolveProjectName(filter.getProject());
+    String projectName = filter.getProject();
     Map<String, String> labelsFilter = filter.getLabelsMap();
 
-    checkValidCharacters(projectName, "project");
+    if (projectName.isEmpty()) {
+      projectName = "*";
```
woop, line 696: We do we persist with these wildcard patterns?

core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -682,19 +682,28 @@ public ApplyFeatureTableResponse applyFeatureTable(ApplyFeatureTableRequest requ
...
+    }
 
-    List<FeatureTable> matchingTables = tableRepository.findAllByProject_Name(projectName);
+    List<FeatureTable> matchingTables;
+    if (projectName.equals("*")) {
+      matchingTables =
+          tableRepository.findAllByProject_NameLikeOrderByNameAsc(projectName.replace('*', '%'));
```
woop, line 702: We don't need wildcard matching any more.

core/src/main/java/feast/core/service/SpecService.java
terryyylim: Since we're keeping FeatureSets functionality inside the Serving implementation, we need to manage the use of wildcard in application.yml when starting Serving service. An example would be our integration tests which uses [these settings](https://github.com/feast-dev/feast/blob/master/serving/src/test/java/feast/serving/it/BaseAuthIT.java#L75).

woop: It's not clear how FeatureSets and FeatureTables are connected here. Aren't they completely separate?

terryyylim: They are but we're using the same set of subscriptions from the application.yml.

woop: Why do we need a subscription for FeatureTables again?

terryyylim: So we're doing away with using subscriptions from application.yml like the logic [here](https://github.com/feast-dev/feast/blob/master/serving/src/main/java/feast/serving/specs/CachedSpecService.java#L235) and directly list all tables?

woop: We will only have a single online store, or a single stream. So I dont think we need subs for `feature tables` any more. 

What you have linked to is for `feature sets`. We still need that right.

terryyylim: Yup we still need it for `feature sets`. Okay, I'll remove the `*` and use `%` directly then.

woop: Why do we need to use `%`?

PR #1040, opened 2020-10-12
CLI command to start/stop/list streaming ingestion job on emr

oavdeev:

Add SDK glue code and the corresponding CLI command to kick off the streaming ingestion on EMR, check if it is already running, stop it and list running jobs.

```release-note
Added CLI commands `start-stream-to-online` and `stop-stream-to-online` to kick off stream ingestion jobs 
```

### DIFF
--- a/sdk/python/feast/cli.py
+++ b/sdk/python/feast/cli.py
@@ -371,5 +371,58 @@ def sync_offline_to_online(feature_table: str, start_time: str, end_time: str):
     feast.pyspark.aws.jobs.sync_offline_to_online(client, table, start_time, end_time)
 
 
+@cli.command()
+@click.option(
+    "--feature-table",
+    "-t",
+    help="Feature table name to ingest data into",
+    required=True,
+)
+@click.option(
+    "--jar", "-j", help="Feature table name to ingest data into", default="",
+)
+def start_stream_to_online(feature_table: str, jar: str):
+    """
+    Start stream to online sync job.
+    """
+    import feast.pyspark.aws.jobs
+
+    client = Client()
+    table = client.get_feature_table(feature_table)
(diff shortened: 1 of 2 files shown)


sdk/python/feast/cli.py
```diff
@@ -371,5 +371,58 @@ def sync_offline_to_online(feature_table: str, start_time: str, end_time: str):
...
+)
+def start_stream_to_online(feature_table: str, jar: str):
+    """
+    Start stream to online sync job.
+    """
+    import feast.pyspark.aws.jobs
+
+    client = Client()
+    table = client.get_feature_table(feature_table)
+    feast.pyspark.aws.jobs.start_stream_to_online(client, table, [jar] if jar else [])
```
woop, line 392: Should we be bypassing the client here? What would happen in the case where we use the job service? or will that be added later?

woop:
@oavdeev can we add a release note to this PR please.

woop:
Did you rebase this PR @oavdeev. It's strange that its looking for a job controller.

sdk/python/feast/cli.py
oavdeev: The job service will have to talk to the core service to get the metadata too, so it would need a client (for example, to find the feature tables so it can kick off streaming ingestion jobs)

oavdeev:
> @oavdeev can we add a release note to this PR please (as in PR template)

Added but i expect the actual interface may still change a bit before the release

woop (approved):

sdk/python/feast/cli.py
woop: Oh I meant moving all requests through the client, even starting jobs.

PR #1049, opened 2020-10-13
Replace Data Source specific format options with DataFormat message

mrzzy:
Add data format messages to represent as data formats:
- `FileFormat` will replace `file_format` in FileOptions
- `StreamFormat `class_path` in Kafka and KinesisOptions.
- This would allow us to support multiple data formats for each `SourceType` (ie `AvroFormat` and `ProtoFormat` for Kafka).
- Added `AvroFormat` Protobuf message

Fixes #

```release-note
Add `FileFormat` to replace `file_format` in FileOptions and `StreamFormat` to replace `class_path` in Kafka and KinesisOptions.
```


core/src/main/java/feast/core/model/DataSource.java
```diff
@@ -88,27 +89,35 @@ public static DataSource fromProto(DataSourceProto.DataSource spec) {
...
+          dataSourceConfigMap.put(
+              "record_format", json.print(spec.getKinesisOptions().getRecordFormat()));
+          dataSourceConfigMap.put("region", spec.getKinesisOptions().getRegion());
+          dataSourceConfigMap.put("stream_name", spec.getKinesisOptions().getStreamName());
+          break;
+        default:
+          throw new UnsupportedOperationException(
+              String.format("Unsupported Feature Store Type: %s", spec.getType()));
+      }
+    } catch (InvalidProtocolBufferException e) {
```
pyalex, line 119: where exactly this exception happens, do we need to wrap it all in `try`?

protos/feast/core/DataSource.proto
```diff
@@ -76,9 +104,9 @@ message DataSource {
     // Kafka topic to collect feature data from.
     string topic = 2;
 
-    // Classpath to the generated Java Protobuf class that can be used to decode
-    // Feature data from the obtained Kafka message
-    string class_path = 3; 
+    // Defines the data format encoding feature/entity data in Kafka messages.
+    // Kafka Data Sources support Avro and Proto as data formats.
+    DataFormat message_format = 3;
```
pyalex, line 109: So now we can should expect parquet from Kafka?
Sounds a little bit strange

pyalex: Maybe we should separate batch formats / streaming formats?

core/src/main/java/feast/core/model/DataSource.java
mrzzy: The exception happens when JSON printer is unable to print the DataFormat Protobuf as JSON. I wrapped the block with try as there are multiple calls to json.print that need a try.

protos/feast/core/DataSource.proto
mrzzy: Add `DataSourceValidator` to make  sure bad data source/data format combinations are rejected by Core  on apply.

pyalex (approved):

PR #1051, opened 2020-10-14
"Start Offline-to-online ingestion" method in Python SDK

pyalex:

This PR aim to glue together various backend implementations of job submission and Feast SDK by introducing some cleaner structure with `SparkJob`, `SparkJobParameters` interfaces and separate `launch` modules.
Also `offline_to_online_ingestion` is added to `JobLauncher` interface. Implementations for standalone & dataproc launchers provided as well.

Fixes #

```release-note

```

### DIFF
--- a/sdk/python/feast/pyspark/abc.py
+++ b/sdk/python/feast/pyspark/abc.py
@@ -95,7 +95,7 @@ class RetrievalJobParameters(SparkJobParameters):
 
     def get_name(self) -> str:
         all_feature_tables_names = [ft["name"] for ft in self._feature_tables]
-        return f"HistoryRetrieval-{all_feature_tables_names}"
+        return f"HistoryRetrieval-{'-'.join(all_feature_tables_names)}"
 
     def get_main_file_path(self) -> str:
         return os.path.join(
--- a/sdk/python/feast/pyspark/abc.py
+++ b/sdk/python/feast/pyspark/abc.py
@@ -331,4 +331,7 @@ class JobLauncher(abc.ABC):
         start: datetime,
         end: datetime,
     ) -> IngestionJob:
+        """
+        Submits a batch ingestion job to a Spark cluster.
+        """
         raise NotImplementedError
--- a/sdk/python/feast/pyspark/abc.py
+++ b/sdk/python/feast/pyspark/abc.py
@@ -1,9 +1,9 @@
 import abc
-import os
 import json
+import os
 from datetime import datetime
-from typing import Dict, List, Optional
 from enum import Enum
+from typing import Dict, List, Optional
 
 
 class SparkJobFailure(Exception):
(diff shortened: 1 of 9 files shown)


oavdeev (commented):
I wonder if there is a way to structure this without the diamond inheritance? After re-reading this a couple times, subjectively i have a pretty hard time following what implements what methods, and what gets called when.

One way I can think of is to maybe split `SparkJob` interface into two:
* `SparkJobParams` that deals with preparing the job to launch (so it provides file path / class name / args) 
* `SparkJob` that deals purely with the job lifecycle after it has been launched (so, id/name/status)

pyalex:
@oavdeev that makes sense

sdk/python/feast/pyspark/abc.py
```diff
@@ -25,6 +35,85 @@ def get_id(self) -> str:
...
+        **kwargs,
+    ):
+        self._feature_tables = feature_tables
+        self._feature_tables_sources = feature_tables_sources
+        self._entity_source = entity_source
+        self._destination = destination
+
+    def get_name(self) -> str:
+        all_feature_tables_names = [ft["name"] for ft in self._feature_tables]
+        return f"HistoryRetrieval-{all_feature_tables_names}"
```
oavdeev, line 98:
```suggestion
        return f"HistoryRetrieval-{'-'.join(all_feature_tables_names)}"
```

sdk/python/feast/pyspark/abc.py
```diff
@@ -25,6 +35,85 @@ def get_id(self) -> str:
...
+        entity_source: Dict,
+        destination: Dict,
+        **kwargs,
+    ):
+        self._feature_tables = feature_tables
+        self._feature_tables_sources = feature_tables_sources
+        self._entity_source = entity_source
+        self._destination = destination
+
+    def get_name(self) -> str:
```
oavdeev, line 96: there is a chance that we'll run into limits on name length on different platforms, on EMR it is 256 chars max

sdk/python/feast/pyspark/abc.py
```diff
@@ -53,8 +142,53 @@ def get_output_file_uri(self, timeout_sec=None):
...
 
+class IngestionJobParameters(SparkJobParameters):
+    def __init__(
+        self,
+        feature_table: Dict,
+        source: Dict,
+        start: datetime,
+        end: datetime,
+        jar: str,
+        **kwargs,
```
oavdeev, line 153: doesn't look like we use kwargs

khorshuheng (approved):

sdk/python/feast/pyspark/launcher.py
```diff
@@ -123,22 +129,45 @@ def start_historical_feature_retrieval_job(
     job_id: str,
```
oavdeev, line 129: i think it would make sense to remove `job_id` here as well

oavdeev (approved):

sdk/python/feast/pyspark/launcher.py
khorshuheng: Noted. Will take note of this in my follow up PR.

khorshuheng: Removed the job_id

PR #1053, opened 2020-10-14
Running historical retrieval on emr

oavdeev:
This contains the code to run historical retrieval job on EMR and a CLI command to invoke it. The CLI command is for debugging only; I expect it'll go away or at least will change to not call EMR code directly.

After landing PR this I plan to refactor EMR code to convert it to a JobLauncher and make it much more similar to the GCP code path, and then we can hook it up to feast Client class.

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/pyspark/aws/jobs.py
+++ b/sdk/python/feast/pyspark/aws/jobs.py
@@ -4,11 +4,16 @@ import logging
 import os
 import random
 import string
+import tempfile
 import time
-from typing import Any, Dict, List, NamedTuple, Optional, Tuple
+from collections import defaultdict
+from io import BytesIO
+from pathlib import Path
+from typing import IO, Any, Dict, List, NamedTuple, Optional, Tuple
 
 import boto3
 import botocore
+import pandas
 import yaml
 
 from feast.client import Client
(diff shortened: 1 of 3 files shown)


woop:
> This contains the code to run historical retrieval job on EMR and a CLI command to invoke it. The CLI command is for debugging only; I expect it'll go away or at least will change to not call EMR code directly.
> 
> After landing PR this I plan to refactor EMR code to convert it to a JobLauncher and make it much more similar to the GCP code path, and then we can hook it up to feast Client class.

Please remember to include a release note so that when we build a change log we know whether there were user facing changes or not.

sdk/python/feast/pyspark/aws/jobs.py
```diff
@@ -516,3 +556,147 @@ def stop_stream_to_online(table_name: str):
...
+            f.name
+            for f in table.features
+            if f.name in feature_filter or "*" in feature_filter
+        ]
+
+        res_dicts.append(_feature_table_to_json(client, table, features))
+    return res_tables, res_dicts
+
+
+def historical_feature_retrieval(
```
woop, line 662: Can we please get comments here. Google style docstring.

oavdeev:
> Please remember to include a release note so that when we build a change log we know whether there were user facing changes or not.

I wasn't sure what to put here since this is just a smallish part of AWS support that doesn't exist in 0.7. I imagine the changelog will have something higher level like "feast now runs on EMR".

woop:
> > Please remember to include a release note so that when we build a change log we know whether there were user facing changes or not.
> 
> I wasn't sure what to put here since this is just a smallish part of AWS support that doesn't exist in 0.7. I imagine the changelog will have something higher level like "feast now runs on EMR".

Yip, but you can just put `NONE` in there so that we know that we can ignore this PR.

PR #1054, opened 2020-10-15
Feast SDK integration for historical feature retrieval using Spark

khorshuheng:
Add historical feature retrieval job back to the Feast client again.

Fixes #

```release-note

```


sdk/python/feast/pyspark/launcher.py
```diff
@@ -0,0 +1,145 @@
...
+    feature_tables: List[FeatureTable],
+    output_format: str,
+    output_path: str,
+    job_id: str,
+) -> RetrievalJob:
+    launcher = resolve_launcher(client._config)
+    retrieval_job_pyspark_script = os.path.join(
+        pathlib.Path(__file__).parent.absolute(),
+        "pyspark",
+        "historical_feature_retrieval_job.py",
```
oavdeev, line 130: nit: if we're using pathlib we could as well use pathlib style path joining with `/`

sdk/python/feast/pyspark/launcher.py
```diff
@@ -0,0 +1,145 @@
...
+        "name": feature_table.name,
+        "entities": [
+            {"name": n, "type": client.get_entity(n).value_type}
+            for n in feature_table.entities
+        ],
+        "max_age": feature_table.max_age.ToSeconds() if feature_table.max_age else None,
+    }
+
+
+def start_historical_feature_retrieval_spark_session(
```
oavdeev, line 92: I'm not 100% follow what is this for? I'd expect this to be within the local launcher as it assumes that you have Spark env setup.

khorshuheng: This function is useful for cases where, the user has their own SparkSession, in a jupyter notebook for example. In such case, it might be more convenient to return a Spark dataframe, rather than the actual output.

pyalex (approved):

PR #1057, opened 2020-10-15
Bump pyarrow and kafka-python versions in v0.7-branch

masonlr:

Allows newer versions of kafka-python and pyarrow in the python sdk.

For pyarrow, the exisitng pyarrow restrictions can fail to install when using python 3.8:

```
$ pip install "pyarrow<0.16.0,>=0.15.1"
...
  -- Configuring incomplete, errors occurred!
  See also "/private/var/folders/b4/tg23_67x57x3kk06w29jry2h0000gn/T/pip-install-n4albg8c/pyarrow/build/temp.macosx-10.9-x86_64-3.8/CMakeFiles/CMakeOutput.log".
  See also "/private/var/folders/b4/tg23_67x57x3kk06w29jry2h0000gn/T/pip-install-n4albg8c/pyarrow/build/temp.macosx-10.9-x86_64-3.8/CMakeFiles/CMakeError.log".
  error: command 'cmake' failed with exit status 1
  ----------------------------------------
  ERROR: Failed building wheel for pyarrow
Failed to build pyarrow
ERROR: Could not build wheels for pyarrow which use PEP 517 and cannot be installed directly
```

For existing projects that use aiokafka, there is a hard requirement for [kafka-python>=2.0.0](https://github.com/aio-libs/aiokafka/blob/84e2a981e9ae83ff24b8de8e43c8b299ee932861/setup.py#L101). Hence, it would be convenient if feast supported the newer version of kafka-python.

No issue, though have discussed with @woop on the kubeflow feast slack channel.

```release-note
NONE
```


woop:
I think there is an incompatibility @masonlr. We should maybe close this for now and swing back to it.

PR #1059, opened 2020-10-15
Create empty Job Service

tsotnet:
**What this PR does / why we need it**: This PR creates an empty Job Service with and the Docker image creation step.


infra/docker/jobservice/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM python:3.7-slim-buster
+
+USER root
+WORKDIR /feast
+
+COPY sdk/python sdk/python
+COPY Makefile Makefile
+COPY protos protos
+COPY tests tests
```
woop, line 9: Do we need the tests?

protos/feast/core/JobService.proto
```diff
@@ -0,0 +1,203 @@
...
+import "feast/serving/ServingService.proto";
+
+service JobService {
+    // Start job to ingest data from offline store into online store
+    rpc StartOfflineToOnlineIngestionJob (StartOfflineToOnlineIngestionJobRequest) returns (StartOfflineToOnlineIngestionJobResponse);
+
+    // Produce a training dataset, return a job id that will provide a file reference
+    rpc GetHistoricalFeatures (GetHistoricalFeaturesRequest) returns (GetHistoricalFeaturesResponse);
+
+    //   IMPORTANT: The following RPCs are not must-haves for the MVP. They are stretch goals. All of these methods would exist internally in the job service and would be called by a control loop to start/stop/restart streaming ingestion jobs.
```
woop, line 35: We can probably just remove this.

protos/feast/core/JobService.proto
```diff
@@ -0,0 +1,203 @@
...
+    // The Historical Features request that triggered this export job
+    GetHistoricalFeaturesRequest request = 2;
+  }
+
+  message OfflineToOnlineMeta {
+    // Reference to the Feature Table being populated by this job
+    string project = 1;
+    string table_name = 2;
+    // Connection String to the Redis Online Store that the job is writing to.
+    string redis_connection_string = 3;
```
oavdeev, line 97: we should make this a struct for consistency with the[ ingestion job parameter format](https://github.com/feast-dev/feast/blob/master/spark/ingestion/src/main/scala/feast/ingestion/IngestionJobConfig.scala#L29)

terryyylim (commented):
I think we should standardize the naming to `created_timestamp`, `stop_timestamp` since those are use eg in FeatureTable proto as well.

protos/feast/core/JobService.proto
```diff
@@ -0,0 +1,203 @@
...
+  // Batch DataSource that can be used to obtain entity values for historical retrieval.
+  // For each entity value, a feature value will be retrieved for that value/timestamp
+  // Only 'BATCH_*' source types are supported.
+  // Currently only BATCH_FILE source type is supported.
+  DataSource entities_source = 2;
+
+  // Optional field to specify project name override. If specified, uses the
+  // given project for retrieval. Overrides the projects specified in
+  // Feature References if both are specified.
+  string project_name = 3;
```
terryyylim, line 147: Maybe standardize to `project` here?

sdk/python/feast/job_service.py
```diff
@@ -0,0 +1,67 @@
+from concurrent.futures import ThreadPoolExecutor
+
+import grpc
+
+import feast
+from feast.core import JobService_pb2_grpc
+
+
+class JobServiceServicer(JobService_pb2_grpc.JobServiceServicer):
```
terryyylim, line 9: I recommend we follow pep8 style guide as stated [here](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names), to utilize lowercase separated by underscores for function names.

tsotnet:
I just added the helm charts, I'll be working on handling the above comments

infra/charts/feast/charts/feast-jobservice/templates/deployment.yaml
```diff
@@ -109,22 +109,21 @@ spec:
...
+        #   {{- end }}
+        #   {{- if index .Values "application-generated.yaml" "enabled" -}}
+        #   ,file:/etc/feast/application-generated.yaml
+        #   {{- end }}
+        #   {{- if index .Values "application-secret.yaml" "enabled" -}}
+        #   ,file:/etc/secrets/feast/application-secret.yaml
+        #   {{- end }}
+        #   {{- if index .Values "application-override.yaml" "enabled" -}}
+        #   ,file:/etc/feast/application-override.yaml
+        #   {{- end }}
```
oavdeev, line 126: remove commented out stuff?

tsotnet: Will do in a separate diff

sdk/python/feast/job_service.py
tsotnet: This is autogenerated code by grpc, I don't think we have a control to modify it.

protos/feast/core/JobService.proto
oavdeev-tt: Actually lets just drop it from here for now. The only way this entire `Job` proto is used is to list jobs, i think it is fine if you just get the basic info like job type and table name there, but not redis connection details. 

We can always add more stuff later.

protos/feast/core/JobService.proto
```diff
@@ -0,0 +1,197 @@
...
+    // Get details of a single job
+    rpc GetJob (GetJobRequest) returns (GetJobResponse);
+}
+
+
+enum JobType {
+	INVALID_JOB = 0;
+	OFFLINE_TO_ONLINE_JOB = 1;
+	STREAM_TO_ONLINE_JOB = 2;
+	EXPORT_JOB = 4;
```
khorshuheng, line 53: Historical feature retrieval job would be classify as an export job?

infra/docker/jobservice/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+USER root
+WORKDIR /feast
+
+COPY sdk/python sdk/python
+COPY Makefile Makefile
+COPY protos protos
+
+# Install make
+RUN apt-get update && apt-get -y install make git
```
khorshuheng, line 11: Why do we need git for the dockerfile?

infra/docker/jobservice/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+# Install Python dependencies
+RUN make compile-protos-python
+
+# Install Feast SDK
+COPY .git .git
```
khorshuheng, line 17: Is there a reason why we copy git into the docker file?

woop (approved):

infra/docker/jobservice/Dockerfile
tsotnet: Looks like git is used by sdk/python/setup.py here https://github.com/feast-dev/feast/blob/master/sdk/python/setup.py#L52. If I remove this line docker build step errors out.

Also I copied this mostly from the jupyter dockerfile, which contains the similar build step. We could improve docker build files by removing the dependency on git (seems like it's only used to copy the readme file from the git repo root), but I think that's probably better handled by another PR.

PR #1061, opened 2020-10-15
EMR launcher

oavdeev:
This adds a EMR launcher in line with the new launcher interface. It does historical retrieval and offline-to-online retrieval. I'll do stream ingestion next, figured this is large enough.

A few notable bits (aside from just moving things around):
* this adds redis configuration to the SDK client, which exists solely to be passed to the spark job. We might want to make it optional
* I've converted a lot of the existing EMR code to aws/emr_utils.py, but left the old code in for now -- will clean it up in a separate PR
* added a convenience method to upload a pandas DF to s3 (currently only supported by EMR launcher). It is not strictly necessary but comes in handy for testing/cli

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/cli.py
+++ b/sdk/python/feast/cli.py
@@ -364,11 +364,13 @@ def sync_offline_to_online(feature_table: str, start_time: str, end_time: str):
     """
     Sync offline store to online.
     """
-    import feast.pyspark.aws.jobs
+    from datetime import datetime
 
     client = Client()
     table = client.get_feature_table(feature_table)
-    feast.pyspark.aws.jobs.sync_offline_to_online(client, table, start_time, end_time)
+    client.start_offline_to_online_ingestion(
+        table, datetime.fromisoformat(start_time), datetime.fromisoformat(end_time)
+    )
 
 
 @cli.command()
@@ -422,6 +424,41 @@ def list_emr_jobs():
     print(
         tabulate(jobs, headers=feast.pyspark.aws.jobs.JobInfo._fields, tablefmt="plain")
     )
+
+
+@cli.command()
+@click.option(
+    "--features",
+    "-f",
+    help="Features in feature_table:feature format, comma separated",
+    required=True,
+)
+@click.option(
+    "--entity-df-path",
+    "-e",
+    help="Path to entity df in CSV format. It is assumed to have event_timestamp column and a header.",
+    required=True,
+)
+@click.option("--destination", "-d", help="Destination", default="")
+def get_historical_features(features: str, entity_df_path: str, destination: str):
+    """
+    Get historical features
+    """
+    import pandas
+
+    client = Client()
+
+    # TODO: clean this up
+    entity_df = pandas.read_csv(entity_df_path, sep=None, engine="python",)
+
+    entity_df["event_timestamp"] = pandas.to_datetime(entity_df["event_timestamp"])
+
+    uploaded_df = client.stage_dataframe(
+        entity_df, "event_timestamp", "created_timestamp"
+    )
+
+    job = client.get_historical_features(features.split(","), uploaded_df,)
+    print(job.get_output_file_uri())
 
 
 if __name__ == "__main__":
--- a/sdk/python/feast/client.py
+++ b/sdk/python/feast/client.py
@@ -76,6 +76,7 @@ from feast.loaders.ingest import (
 from feast.online_response import OnlineResponse, _infer_online_entity_rows
 from feast.pyspark.abc import RetrievalJob, SparkJob
 from feast.pyspark.launcher import (
+    stage_dataframe,
     start_historical_feature_retrieval_job,
     start_historical_feature_retrieval_spark_session,
     start_offline_to_online_ingestion,
@@ -885,9 +886,16 @@ class Client:
         return feature_tables
 
     def start_offline_to_online_ingestion(
-        self,
-        feature_table: Union[FeatureTable, str],
-        start: Union[datetime, str],
-        end: Union[datetime, str],
+        self, feature_table: Union[FeatureTable, str], start: datetime, end: datetime,
     ) -> SparkJob:
         return start_offline_to_online_ingestion(feature_table, start, end, self)  # type: ignore
+
+    def stage_dataframe(
+        self,
+        df: pd.DataFrame,
+        event_timestamp_column: str,
+        created_timestamp_column: str,
+    ) -> FileSource:
+        return stage_dataframe(
+            df, event_timestamp_column, created_timestamp_column, self
+        )
--- a/sdk/python/feast/pyspark/abc.py
+++ b/sdk/python/feast/pyspark/abc.py
@@ -5,6 +5,10 @@ from datetime import datetime
 from enum import Enum
 from typing import Dict, List, Optional
 
+import pandas
+
+from feast.data_source import FileSource
+
 
 class SparkJobFailure(Exception):
     """
@@ -258,18 +262,30 @@ class IngestionJobParameters(SparkJobParameters):
         start: datetime,
         end: datetime,
         jar: str,
+        redis_host: str,
+        redis_port: int,
+        redis_ssl: bool,
     ):
         self._feature_table = feature_table
         self._source = source
         self._start = start
         self._end = end
         self._jar = jar
+        self._redis_host = redis_host
+        self._redis_port = redis_port
+        self._redis_ssl = redis_ssl
 
     def get_name(self) -> str:
         return (
-            f"BatchIngestion-{self._feature_table['name']}-"
+            f"BatchIngestion-{self.get_feature_table_name()}-"
             f"{self._start.strftime('%Y-%m-%d')}-{self._end.strftime('%Y-%m-%d')}"
         )
+
+    def _get_redis_config(self):
+        return dict(host=self._redis_host, port=self._redis_port, ssl=self._redis_ssl)
+
+    def get_feature_table_name(self) -> str:
+        return self._feature_table["name"]
 
     def get_main_file_path(self) -> str:
         return self._jar
@@ -289,6 +305,8 @@ class IngestionJobParameters(SparkJobParameters):
             self._start.strftime("%Y-%m-%dT%H:%M:%S"),
             "--end",
             self._end.strftime("%Y-%m-%dT%H:%M:%S"),
+            "--redis",
+            json.dumps(self._get_redis_config()),
         ]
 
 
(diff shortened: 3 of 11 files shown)


sdk/python/feast/cli.py
```diff
@@ -424,5 +426,44 @@ def list_emr_jobs():
...
+    "--features",
+    "-f",
+    help="Features in feature_table:feature format, comma separated",
+    required=True,
+)
+@click.option(
+    "--entity-df-path",
+    "-e",
+    help="Path to entity df in CSV format. It is assumed to have event_timestamp column and a header.",
+    required=True,
```
khorshuheng, line 440: I am wondering if it might be better if the users are expected to provide a uri that is recognizable by the Spark Launcher, such as s3:// for EMR, gs:// for Dataproc, and file:// for standalone cluster launchers running locally. That way, we skip the process of reading to Pandas dataframe and convert the file again.

Staged panda dataframe is still a useful method to have though, because we plan to add support to pandas dataframe as input argument for historical feature retrieval method, for Feast Client.

sdk/python/feast/pyspark/launchers/aws/emr.py
```diff
@@ -0,0 +1,244 @@
...
+        staging_location: str,
+        emr_log_location: str,
+    ):
+        """
+        Initialize a dataproc job controller client, used internally for job submission and result
+        retrieval.
+
+        Args:
+            cluster_name (str):
+                Dataproc cluster name.
```
khorshuheng, line 119: nit: This doc string is for dataproc, need to convert this for EMR launcher.

sdk/python/feast/pyspark/abc.py
```diff
@@ -334,3 +352,18 @@ def offline_to_online_ingestion(
             IngestionJob: wrapper around remote job that can be used to check when job completed.
         """
         raise NotImplementedError
+
+    @abc.abstractmethod
```
khorshuheng, line 356: Since Dataproc Launcher and Standalone Launcher hasn't implemented this change, this currently breaks the python linter CI stage and also breaks Dataproc and Standalone Launcher functionalities since they can't be instantiated.

Perhaps we can create the methods for the two Launchers, and raise NotImplementedError for the time being? Then we can either add the implementation directly via this PR (either me or @pyalex), or with a follow up PR?

sdk/python/feast/pyspark/historical_feature_retrieval_job.py
```diff
@@ -773,7 +773,7 @@ def _feature_table_from_dict(dct: Dict[str, Any]) -> FeatureTable:
     spark = SparkSession.builder.getOrCreate()
     args = _get_args()
     feature_tables_conf = json.loads(args.feature_tables)
-    feature_tables_sources_conf = json.loads(args.feature_tables_source)
+    feature_tables_sources_conf = json.loads(args.feature_tables_sources)
```
khorshuheng, line 776: Thanks for catching the typo

sdk/python/feast/pyspark/launchers/aws/emr.py
```diff
@@ -0,0 +1,244 @@
...
+
+        job_ref = self._submit_emr_job(step)
+
+        return EmrIngestionJob(self._emr_client(), job_ref)
+
+    def stage_dataframe(
+        self, df: pandas.DataFrame, event_timestamp: str, created_timestamp_column: str
+    ) -> FileSource:
+        with tempfile.NamedTemporaryFile() as f:
+            df.to_parquet(f)
```
khorshuheng, line 232: @pyalex As of Feast 0.7 the dataframe is typically converted to pyarrow / Avro format. I am wondering, if we should make parquet the standard staging format as oppose to pyarrow / Avro?  Since Spark doesn't support Avro format out of the box.

sdk/python/feast/cli.py
```diff
@@ -424,5 +426,44 @@ def list_emr_jobs():
...
+
+    # TODO: clean this up
+    entity_df = pandas.read_csv(
+        entity_df_path, sep=None, engine="python", dtype={"user_id": str}
+    )
+
+    entity_df["event_timestamp"] = pandas.to_datetime(entity_df["event_timestamp"])
+
+    uploaded_df = client.stage_dataframe(
+        entity_df, "event_timestamp", "created_timestamp"
```
khorshuheng, line 461: Just a heads up: "created_timestamp" is actually supposed to be optional for the entity, so it's a bug that we need to resolve in another PR.

sdk/python/feast/pyspark/launchers/aws/emr.py
```diff
@@ -0,0 +1,244 @@
...
+    _sync_offline_to_online_step,
+    _upload_jar,
+    _wait_for_job_state,
+)
+
+
+class EmrJobMixin:
+    def __init__(self, emr_client, job_ref: EmrJobRef):
+        """
+        :param operation: (google.api.core.operation.Operation): A Future for the spark job result,
```
woop, line 43: Is this Google docstring style?

sdk/python/feast/pyspark/launchers/aws/emr.py
```diff
@@ -0,0 +1,244 @@
...
+    def get_output_file_uri(self, timeout_sec=None):
+        _wait_for_job_state(
+            self._emr_client, self._job_ref, TERMINAL_STEP_STATES, timeout_sec
+        )
+        return self._output_file_uri
+
+
+class EmrIngestionJob(EmrJobMixin, IngestionJob):
+    """
+    Ingestion job result for a Dataproc cluster
```
woop, line 86: Typo

sdk/python/feast/pyspark/launchers/aws/emr.py
```diff
@@ -0,0 +1,244 @@
...
+
+        emr = self._emr_client()
+
+        if self._existing_cluster_id:
+            step["ActionOnFailure"] = "CONTINUE"
+            step_ids = emr.add_job_flow_steps(
+                JobFlowId=self._existing_cluster_id, Steps=[step],
+            )
+            return EmrJobRef(self._existing_cluster_id, step_ids["StepIds"][0])
+        else:
```
woop, line 160: Personal preference, but I prefer to not have an "else" if the "if" returns.

sdk/python/feast/pyspark/launchers/aws/emr.py
woop: > As of Feast 0.7 the dataframe is typically converted to pyarrow / Avro format

In which instances do you see that? Uploading to BQ?

sdk/python/feast/client.py
```diff
@@ -885,9 +886,16 @@ def _get_feature_tables_from_feature_refs(
...
     def start_offline_to_online_ingestion(
-        self,
-        feature_table: Union[FeatureTable, str],
-        start: Union[datetime, str],
-        end: Union[datetime, str],
+        self, feature_table: Union[FeatureTable, str], start: datetime, end: datetime,
     ) -> SparkJob:
         return start_offline_to_online_ingestion(feature_table, start, end, self)  # type: ignore
+
+    def stage_dataframe(
```
pyalex, line 893: We actually already have method in the Feast Client that takes dataframe and put it in offline storage. Currently it called `ingest` but I guess we'll rename it. Anyway, this shouldn't be a part of spark interop, since it's not related to spark

sdk/python/feast/pyspark/launchers/aws/emr.py
khorshuheng: https://github.com/feast-dev/feast/blob/v0.7.1/sdk/python/feast/loaders/file.py#L29

pyalex: we already write parquet in `client.ingest` but with DataFormat PR we'll have to support all formats that user may specify as Batch Source format, since `client.ingest` writes to existing batch source

khorshuheng: @pyalex  In the case of historical feature retrieval, where the user input is Panda dataframe, do we actually want to make the user specify the format of the entity source file? Or simply standardized to parquet?

sdk/python/feast/pyspark/launchers/aws/emr_utils.py
```diff
@@ -0,0 +1,363 @@
...
+    """ Compute sha256 hash of a file. File pointer will be reset to 0 on return. """
+    fileobj.seek(0)
+    h = hashlib.sha256()
+    for block in iter(lambda: fileobj.read(2 ** 20), b""):
+        h.update(block)
+    fileobj.seek(0)
+    return h.hexdigest()
+
+
+def _s3_upload(
```
pyalex, line 83: Can this be moved to staging client?

sdk/python/feast/client.py
oavdeev: In my mind this is somewhat different from ingest. It is not intended for permanent storage, this is a convenience method "take this dataframe, put it in some temp location where Spark can access it". I agree launcher might not be the best place for it, just gotta be some code that can read staging_location from the config to construct the temp path.

sdk/python/feast/pyspark/launchers/aws/emr_utils.py
oavdeev: Yes, if it is ok i'd do it in a separate PR though. It is not exactly 1-1 with the current staging client interface.

sdk/python/feast/cli.py
oavdeev: I agree, it is mostly for convenience/testing for now, to reduce a # of steps that someone need to do to see if historical retrieval works. I wouldn't expect people to normally use local CSV for entity dfs. I'd tweak this interface in later PRs though.

sdk/python/feast/pyspark/launchers/aws/emr.py
oavdeev: i'll leave it as is here, if it was a "short circuit" early return I'd agree, but here it is symmetric since two branches are equivalent alternative code paths (using an existing cluster `else` using a new cluster)

sdk/python/feast/pyspark/launchers/aws/emr.py
woop: Didn't we previously agree to only support parquet?

sdk/python/feast/pyspark/launchers/aws/emr_utils.py
khorshuheng: I submitted a PR here: https://github.com/feast-dev/feast/pull/1063/files

sdk/python/feast/client.py
pyalex: if it's some upload-to-temp-location function - it's probably shouldn't be part of Feast Client API. maybe contrib?
or just keep it internally. What's the user use case?

oavdeev: Just the user convenience - if you're getting started with feast and want to run historical retrieval, we can upload your pandas entity dataframe to S3 so you don't have to think about how to upload it and what bucket to use. We'll just put it in staging location for you. Basically trying to remove an extra friction point in onboarding and tutorials. Right now it is only used for CLI historical-retrieval command.

I agree it may not be the best place for it, but at the same time it needs to have access to the config to figure out _where_ to upload the dataframe. So i can't make it completely detached from the client (that has the config object)

pyalex (approved):

pyalex:
We agreed with @oavdeev to refactor staging client in separate PR and merge this to unblock further development.

PR #1067, opened 2020-10-19
Historical feature retrieval e2e test

khorshuheng:

This PR add end to end test for historical feature retrieval. The new test has been tested locally against a local spark cluster, but at the moment it will fail during the CI as the corresponding test infrastructure has not been set up yet.

Fixes #

```release-note
NONE
```

### DIFF
--- a/tests/e2e/test_historical_features.py
+++ b/tests/e2e/test_historical_features.py
@@ -1,0 +1,125 @@
+import os
+import tempfile
+import uuid
+from datetime import datetime, timedelta
+from urllib.parse import urlparse
+
+import numpy as np
+import pandas as pd
+import pytest
+from google.protobuf.duration_pb2 import Duration
+from pandas._testing import assert_frame_equal
+
+from feast import Client, Entity, Feature, FeatureTable, FileSource, ValueType
+from feast.data_format import ParquetFormat
+from feast.staging.storage_client import get_staging_client
+
+
+@pytest.fixture(scope="function")
+def staging_path(pytestconfig, tmp_path):
+    if pytestconfig.getoption("env") == "local":
+        return f"file://{tmp_path}"
+
+    staging_path = pytestconfig.getoption("staging_path")
+    return os.path.join(staging_path, str(uuid.uuid4()))
+
+
+def test_historical_features(feast_client: Client, staging_path: str):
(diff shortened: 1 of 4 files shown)


tests/e2e/test_historical_features.py
```diff
@@ -0,0 +1,124 @@
...
+    max_age.FromSeconds(2 * 86400)
+
+    transactions_feature_table = FeatureTable(
+        name="transactions",
+        entities=["customer_id"],
+        features=[
+            Feature("daily_transactions", ValueType.DOUBLE),
+            Feature("total_transactions", ValueType.DOUBLE),
+        ],
+        batch_source=FileSource(
```
woop, line 42: I think we need to change the file source to have named arguments and not positional, even if they are required. The only positional argument should be the path, and it should come first. Otherwise its hard to understand what is being defined here.

tests/e2e/test_historical_features.py
```diff
@@ -0,0 +1,124 @@
...
+        datetime.utcnow()
+        .replace(hour=0, minute=0, second=0, microsecond=0)
+        .replace(tzinfo=None)
+    )
+    retrieval_outside_max_age_date = retrieval_date + timedelta(1)
+    event_date = retrieval_date - timedelta(2)
+    creation_date = retrieval_date - timedelta(1)
+
+    customers = [1001, 1002, 1003, 1004, 1005]
+    daily_transactions = [np.random.rand() * 10 for _ in customers]
```
woop, line 63: please set a fixed seed.

tests/e2e/test_historical_features.py
```diff
@@ -0,0 +1,124 @@
...
+
+        job = feast_client.get_historical_features(feature_refs, customer_source)
+        output_dir = job.get_output_file_uri()
+
+        _, _, joined_df_destination_path, _, _, _ = urlparse(output_dir)
+        joined_df = pd.read_parquet(joined_df_destination_path)
+
+        expected_joined_df = pd.DataFrame(
+            {
+                "event_timestamp": [retrieval_date for _ in customers]
```
woop, line 109: Are you at all testing the as_of join functionality? Looks like you just have one timestamp in the range?

PR #1072, opened 2020-10-19
Streaming Ingestion Job supports AVRO format as input

pyalex:

Add avro parser to Streaming Ingestion Job if `AvroFormat` was set in `Source`. Relies on `jsonSchema` passed with `AvroFormat` to generate `DataFrame` structure.

Fixes #

```release-note

```

### DIFF
--- a/spark/ingestion/src/test/scala/feast/ingestion/StreamingPipelineIT.scala
+++ b/spark/ingestion/src/test/scala/feast/ingestion/StreamingPipelineIT.scala
@@ -38,12 +38,17 @@ import feast.ingestion.helpers.RedisStorageHelper._
 import feast.ingestion.helpers.DataHelper._
 import feast.proto.storage.RedisProto.RedisKeyV2
 import feast.proto.types.ValueProto
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.avro.to_avro
+import org.apache.spark.sql.functions.{col, struct}
+import org.apache.spark.sql.types.StructType
 
 class StreamingPipelineIT extends SparkSpec with ForAllTestContainer {
   val redisContainer = GenericContainer("redis:6.0.8", exposedPorts = Seq(6379))
   val kafkaContainer = KafkaContainer()
 
   override val container = MultipleContainers(redisContainer, kafkaContainer)
+
   override def withSparkConfOverrides(conf: SparkConf): SparkConf = conf
     .set("spark.redis.host", redisContainer.host)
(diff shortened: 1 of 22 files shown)


spark/ingestion/src/test/scala/feast/ingestion/StreamingPipelineIT.scala
```diff
@@ -291,4 +296,75 @@ class StreamingPipelineIT extends SparkSpec with ForAllTestContainer {
...
+        bootstrapServers = kafkaContainer.bootstrapServers,
+        topic = "avro",
+        format = AvroFormat(schemaJson = """{
+            |"type": "record",
+            |"name": "TestMessage",
+            |"fields": [
+            |{"name": "customer", "type": ["string","null"]},
+            |{"name": "feature1", "type": "int"},
+            |{"name": "feature2", "type": "float"},
+            |{"name": "eventTimestamp", "type": [{"type": "long", "logicalType": "timestamp-micros"}, "null"]}
```
khorshuheng, line 321: Generate data creates the event timestamp as datetime, but the type is defined as long here. Seems like this is what causing the e2e test to fail.

tests/e2e/test_online_features.py
```diff
@@ -128,3 +137,111 @@ def test_offline_ingestion(feast_client: Client, staging_path: str):
...
+
+    wait_retry_backoff(
+        lambda: (None, job.get_status() == SparkJobStatus.IN_PROGRESS), 60
+    )
+
+    time.sleep(5)
+
+    original = generate_data()[["s2id", "unique_drivers", "event_timestamp"]]
+    for record in original.to_dict("records"):
+        record["event_timestamp"] = int(record["event_timestamp"].timestamp() * 1e6)
```
khorshuheng, line 177: Based on the error message here: http://prow.feast.ai/view/gcs/feast-templocation-kf-feast/pr-logs/pull/feast-dev_feast/1072/test-end-to-end/1318191468216586242

It seems like a `datetime` object input is expected, which will then be converted to long.

tests/e2e/requirements.txt
```diff
@@ -8,7 +8,9 @@ pytest-mock==1.10.4
 pytest-timeout==1.3.3
 pytest-ordering==0.6.*
 pytest-xdist==2.1.0
-tensorflow-data-validation==0.21.2
+# tensorflow-data-validation==0.21.2
```
khorshuheng, line 11: Should we just delete this rather than commenting it out?

khorshuheng (approved):

PR #1083, opened 2020-10-21
Multiple tiny AWS related fixes

oavdeev:
* Fix object path math in client staging uploader
* Add random suffix to the historical retrieval output location
* EMR job now actually checks for status
* add boto3 dep

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/client.py
+++ b/sdk/python/feast/client.py
@@ -927,7 +927,7 @@ class Client:
             entity_source,
             feature_tables,
             output_format,
-            os.path.join(output_location, str(uuid.uuid4)),
+            os.path.join(output_location, str(uuid.uuid4())),
         )
 
     def get_historical_features_df(
--- a/sdk/python/feast/client.py
+++ b/sdk/python/feast/client.py
@@ -913,7 +913,7 @@ class Client:
                     else entity_staging_uri.netloc
                 )
                 staging_client.upload_file(
-                    df_export_path.name, bucket, entity_staging_uri.path
+                    df_export_path.name, bucket, entity_staging_uri.path.lstrip("/")
                 )
                 entity_source = FileSource(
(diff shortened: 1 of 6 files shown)


sdk/python/setup.py
```diff
@@ -45,6 +45,9 @@
     "pyarrow<0.16.0,>=0.15.1",
     "numpy",
     "google",
+    "s3fs",
+    "boto3",
```
woop, line 49: shouldn't AWS dependencies be optional?

sdk/python/feast/client.py
```diff
@@ -912,7 +912,7 @@ def get_historical_features(
                     else entity_staging_uri.netloc
                 )
                 staging_client.upload_file(
-                    df_export_path.name, bucket, entity_staging_uri.path
+                    df_export_path.name, bucket, entity_staging_uri.path.lstrip("/")
```
pyalex, line 915: shouldn't that `strip` be part of implementation (GS, s3)?

sdk/python/setup.py
oavdeev: Since GCP is not i figured we could as well bundle AWS deps too. Otherwise in each tutorial we'd have to tell people to install them separately. We'd have also have to build two docker images (GCP and AWS flavored) at least for the job service.

sdk/python/feast/client.py
oavdeev: We can put `strip` into both implementations too (gcp actually has it already). It just goes a bit against S3 (and i suspect GCP) semantics where `/a` and `a` are different objects

khorshuheng: I will look into the GCP part again.

khorshuheng (approved):

PR #1092, opened 2020-10-22
support redis ssl in feast-serving

oavdeev:
Support connecting to redis with ssl in feast-serving

```release-note
NONE
```

### DIFF
--- a/protos/feast/core/Store.proto
+++ b/protos/feast/core/Store.proto
@@ -110,6 +110,8 @@ message Store {
     int32 max_retries = 4;
     // Optional. How often flush data to redis
     int32 flush_frequency_seconds = 5;
+    // Optional. Connect over SSL.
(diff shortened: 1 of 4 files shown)


protos/feast/core/Store.proto
```diff
@@ -110,6 +110,8 @@ message Store {
     int32 max_retries = 4;
     // Optional. How often flush data to redis
     int32 flush_frequency_seconds = 5;
+
+    bool ssl = 6;
```
woop, line 114: You are missing a comment, and should this be `enable_ssl`?

oavdeev: Added a comment. This parameter is called just `ssl`/`setSsl` both in python and java client so I left it that way for consistency

woop (approved):

PR #1094, opened 2020-10-23
Refactored end-to-end tests fully orchestrated by pytest

pyalex:
Reworked e2e tests consist of:
* Registering definitions in Core
* Retrieve historical features
* Retrieve online features (ingested by batch/stream)

Runners regrouped by environment:
* standalone (run spark locally, all services (kafka, redis, postgres, incl feast core & serving) spawned by pytest as subprocesses)
* gcp (dataproc, redis-cluster & kafka in kubernetes)
* (potentially) aws

Auth (enabled/disabled) is now just a parameterised fixture.

Fixes #

```release-note

```


.github/workflows/complete.yml
```diff
@@ -156,3 +156,25 @@ jobs:
       - name: copy to gs
         run: gsutil cp ./spark/ingestion/target/feast-ingestion-spark-${GITHUB_SHA}.jar gs://feast-jobs/spark/ingestion/
 
+  test-end-to-end:
+    runs-on: [self-hosted]
```
woop, line 160: The self-hosted runners don't clean up. If you aren't using docker then all the state will remain. I dont think this test is safe, it's going to fall over for our current runners because they are already so unstable.

pyalex: I don't use docker. It's regular test, which github runner was designed for, hence there should be way to easily clean up. Right now for example python w/ environment (and installed packages) being cleaned up by python action

pyalex: Also, we're on preemptible nodes right now. So we have additional clean up

woop: > I don't use docker. It's regular test, which github runner was designed for, hence there should be way to easily clean up. Right now for example python w/ environment (and installed packages) being cleaned up by python action

ok, but self-hosted runners don't clean up right? state just hangs around as far as I know.

infra/scripts/setup-e2e-env-gcp.sh
```diff
@@ -0,0 +1,36 @@
...
+      $feast_kafka_ip_name $feast_redis_1_ip_name $feast_redis_2_ip_name $feast_redis_3_ip_name \
+      --region ${GCLOUD_REGION} --subnet ${GCLOUD_SUBNET}
+
+export feast_kafka_ip=$(gcloud compute addresses describe $feast_kafka_ip_name --region=${GCLOUD_REGION} --format "value(address)")
+export feast_redis_1_ip=$(gcloud compute addresses describe $feast_redis_1_ip_name --region=${GCLOUD_REGION} --format "value(address)")
+export feast_redis_2_ip=$(gcloud compute addresses describe $feast_redis_2_ip_name --region=${GCLOUD_REGION} --format "value(address)")
+export feast_redis_3_ip=$(gcloud compute addresses describe $feast_redis_3_ip_name --region=${GCLOUD_REGION} --format "value(address)")
+
+
+envsubst '$feast_kafka_ip' < helm/kafka-values.tpl.yaml > helm/kafka-values.yaml
```
woop, line 28: Do we have to use double templating (envsubst, helm)? What if we just used `--set` CLI command for Kafka IP and redis IP?

PR #1098, opened 2020-10-26
Add gh workflow for dockerhub

terryyylim:

This PR adds a github workflow to build and push to Dockerhub for every new release.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,6 +21,62 @@ jobs:
         echo $RELEASE_VERSION
         echo ${{ steps.vars.outputs.release_version }}
 
+  build-publish-docker-images:
+    runs-on: [ubuntu-latest]
+    strategy:
+      matrix:
+        component: [core, serving, jupyter]
+    env:
+      MAVEN_CACHE: gs://feast-templocation-kf-feast/.m2.2020-08-19.tar
+      DOCKER_BUILDKIT: '1'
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -21,6 +21,62 @@ jobs:
...
+      - name: Login to DockerHub
+        uses: docker/login-action@v1 
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
+        with:
+          version: '290.0.1'
+          export_default_credentials: true
+      - run: gcloud auth configure-docker --quiet
```
woop, line 47: What is this for?

terryyylim: If I'm understanding it correctly, it configures the credentials for the github action since we're not providing the SA key explicitly in the previous step as shown in the example [here](https://github.com/GoogleCloudPlatform/github-actions/tree/master/setup-gcloud#usage).

khorshuheng: @terryyylim gcloud auth configure-docker is only used when you need to authenticate docker to push to Google container registry.

woop (approved):

PR #1102, opened 2020-10-27
Add validation when feature type is changed

terryyylim:

In the past, we didn't allow changes to feature types so this validation was not necessary. As we think about implementing deletion of feature tables, we'll need to consider how old values stored in redis will be retrieved. This PR is the first step to handling older values that are based of older types.

Fixes #

```release-note
NONE
```

### DIFF
--- a/serving/src/main/java/feast/serving/service/OnlineServingServiceV2.java
+++ b/serving/src/main/java/feast/serving/service/OnlineServingServiceV2.java
@@ -104,30 +104,19 @@ public class OnlineServingServiceV2 implements ServingServiceV2 {
             FeatureTableSpec featureTableSpec =
                 specService.getFeatureTableSpec(projectName, feature.get().getFeatureReference());
             FeatureProto.FeatureSpecV2 tempFeatureSpecV2 =
-                featureTableSpec.getFeaturesList().stream()
-                    .filter(
-                        featureSpecV2 ->
-                            featureSpecV2
-                                .getName()
-                                .equals(feature.get().getFeatureReference().getName()))
-                    .collect(Collectors.toList())
-                    .get(0);
-            String valueTypeString = tempFeatureSpecV2.getValueType().toString();
-            String valueCaseString = feature.get().getFeatureValue().getValCase().toString();
-            boolean isSameType = checkSameType(valueTypeString, valueCaseString);
+                specService.getFeatureSpec(feature.get().getFeatureReference());
+            ValueProto.ValueType.Enum valueTypeEnum = tempFeatureSpecV2.getValueType();
+            ValueProto.Value.ValCase valueCase = feature.get().getFeatureValue().getValCase();
+            boolean isSameFeatureType = checkSameType(valueTypeEnum, valueCase);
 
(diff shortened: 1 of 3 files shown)


serving/src/main/java/feast/serving/service/OnlineServingServiceV2.java
```diff
@@ -102,14 +103,31 @@ public GetOnlineFeaturesResponse getOnlineFeatures(GetOnlineFeaturesRequestV2 re
 
             FeatureTableSpec featureTableSpec =
                 specService.getFeatureTableSpec(projectName, feature.get().getFeatureReference());
+            FeatureProto.FeatureSpecV2 tempFeatureSpecV2 =
```
pyalex, line 106: Can we build some map outside of the loop to avoid this filtering every iteration?

serving/src/main/java/feast/serving/service/OnlineServingServiceV2.java
```diff
@@ -102,14 +103,31 @@ public GetOnlineFeaturesResponse getOnlineFeatures(GetOnlineFeaturesRequestV2 re
...
+            String valueCaseString = feature.get().getFeatureValue().getValCase().toString();
+            boolean isSameType = checkSameType(valueTypeString, valueCaseString);
 
             boolean isOutsideMaxAge = checkOutsideMaxAge(featureTableSpec, entityRow, feature);
-            Map<String, ValueProto.Value> valueMap = unpackValueMap(feature, isOutsideMaxAge);
+            Map<String, ValueProto.Value> valueMap =
+                unpackValueMap(feature, isOutsideMaxAge, isSameType);
             allValueMaps.putAll(valueMap);
 
+            boolean isNotFound = false;
```
pyalex, line 124: isn't it the same as
`isNotFound = !isSameType`

serving/src/main/java/feast/serving/service/OnlineServingServiceV2.java
```diff
@@ -153,6 +171,33 @@ public GetOnlineFeaturesResponse getOnlineFeatures(GetOnlineFeaturesRequestV2 re
     }
   }
 
+  private boolean checkSameType(String valueTypeString, String valueCaseString) {
+    HashMap<String, String> typingMap =
+        new HashMap<>() {
+          {
+            put("BYTES", "BYTES_VAL");
```
pyalex, line 178: why convert to string? Can we use original enum values instead?

serving/src/main/java/feast/serving/service/OnlineServingServiceV2.java
```diff
@@ -196,11 +241,11 @@ public GetOnlineFeaturesResponse getOnlineFeatures(GetOnlineFeaturesRequestV2 re
   }
 
   private static Map<String, ValueProto.Value> unpackValueMap(
-      Optional<Feature> feature, boolean isOutsideMaxAge) {
+      Optional<Feature> feature, boolean isOutsideMaxAge, boolean isSameType) {
```
pyalex, line 244: `sameType` name seems not very clear here

pyalex (approved):

PR #1110, opened 2020-10-29
In Historical Retrieval from BQ join between source & entities is performed inside BQ

pyalex:

This PR prevents load of full (batch source) table from BQ to join it with entities in Spark.
Instead entities are being uploaded to BQ and original BQ sources are being replaced by views (with join) before passing batch source to historical retrieval job.

This optimization is only applied when user pass entities to method `get_historical_features` as pandas dataframe or as BigQuerySource.

Fixes #

```release-note

```


protos/feast/core/JobService.proto
```diff
@@ -128,6 +128,12 @@ message GetHistoricalFeaturesRequest {
   // Export to AWS S3 - s3://path/to/features
   // Export to GCP GCS -  gs://path/to/features
   string output_location = 4;
+
+  // Specify format name for output, eg. parquet
+  string output_format = 5;
+
+  // Data sources to read features from
+  repeated DataSource data_sources = 6;
```
woop, line 136: It's not clear what this data source is, nor how its used.

sdk/python/feast/staging/entities.py
```diff
@@ -0,0 +1,129 @@
...
+  source.*
+FROM
+  `{entities.project}.{entities.dataset_id}.{entities.table_id}` entities
+JOIN
+  `{source.project}.{source.dataset_id}.{source.table_id}` source
+ON
+  ({entity_key})"""
+
+
+def create_view_to_source_with_joined_entities(
```
woop, line 97: You should have bigquery in the name of this function if its hardcoded to BQ

sdk/python/feast/staging/entities.py
```diff
@@ -0,0 +1,129 @@
...
+    destination_ref = bigquery.TableReference(
+        bigquery.DatasetReference(source_ref.project, source_ref.dataset_id),
+        f"_view_{source_ref.table_id}_{datetime.now():%Y%m%d%H%M%s}",
+    )
+
+    view = bigquery.Table(destination_ref)
+    view.view_query = JOIN_TEMPLATE.format(
+        entities=entities_ref,
+        source=source_ref,
+        entity_key=",".join([f"source.{e} = entities.{e}" for e in entity_names]),
```
khorshuheng, line 118: What happens if field_mappings is defined for the BigQuery table, and one or more of the mapped fields involves the entities?

pyalex: Added check: if any entity is in field_mapping - optimization not applied

khorshuheng (approved):

PR #1111, opened 2020-10-29
Implement JobService API calls & connect it to SDK

tsotnet:

- Implement start_offline_to_online_ingestion, start_historical_feature_retrieval_job, start_stream_to_online_ingestion SDK calls with JobService (including branching logic depending on whether job service is enabled or not)
- Set correct environment for JobService (Helm Charts & Dockerfile)
- Update JobService protos as necessary (according to Slack discussion)


sdk/python/feast/constants.py
```diff
@@ -115,7 +116,15 @@ class AuthProvider(Enum):
...
     CONFIG_SERVING_SERVER_SSL_CERT_KEY: "",
-    # Default connection timeout to Feast Serving and Feast Core (in seconds)
+    # Default Feast Job Service URL
+    CONFIG_JOB_SERVICE_URL_KEY: "localhost:6568",
+    # Enable or disable TLS/SSL to Feast Job Service
+    CONFIG_JOB_SERVICE_ENABLE_SSL_KEY: "False",
+    # Path to certificate(s) to secure connection to Feast Job Service
+    CONFIG_JOB_SERVICE_SERVER_SSL_CERT_KEY: "",
+    # Enable or disable Feast Job Service
+    CONFIG_JOB_SERVICE_ENABLED: "False",
```
woop, line 126: Do we need this? Can we enable the JS if the url is provided?

woop (changes requested):

sdk/python/feast/constants.py
oavdeev: @tsotnet see my PR i had to implement the same thing

infra/docker/jobservice/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.7-slim-buster
+FROM jupyter/pyspark-notebook:ae5f7e104dd5
```
oavdeev, line 1: Hmm this is not ideal, that image is a bit of a kitchen sink (and more than a GB).

But this does raise an interesting question, for best local experience we want to bundle job service with spark, but for production use we could use a much more minimal image. Thoughts (@woop ?)

(not opposed to merging this as is but we'd need to come up with a better way later)

For context, I'm not very concerned about the image _size_ per se, but those jupyter images have a bunch of stuff installed in them in different ways (conda/pip/apt) so it may make this setup more fragile than it needs to.

infra/docker/jobservice/Dockerfile
```diff
@@ -27,4 +27,10 @@ RUN wget -q https://github.com/grpc-ecosystem/grpc-health-probe/releases/downloa
          -O /usr/bin/grpc-health-probe && \
     chmod +x /usr/bin/grpc-health-probe
 
+ENV FEAST_SPARK_LAUNCHER standalone
+ENV FEAST_SPARK_STANDALONE_MASTER "local[*]"
+ENV FEAST_SPARK_HOME $SPARK_HOME
+ENV FEAST_SPARK_EXTRA_OPTIONS "--jars https://storage.googleapis.com/hadoop-lib/gcs/gcs-connector-hadoop2-latest.jar \
+--conf spark.hadoop.fs.gs.impl=com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystem"
+
```
oavdeev, line 35: Can we not hardcode those here and set them from docker-compose instead?

sdk/python/feast/constants.py
tsotnet: Got it, merged

infra/docker/jobservice/Dockerfile
tsotnet: I agree on your points... I spent some time on this and couldn't easily get both proper python and spark versions working properly with each other. Since we already had this image for jupyter notebook I just reused it, but we should definitely have a better base for this... Not sure it's very high priority to figure it out right now though.

infra/docker/jobservice/Dockerfile
tsotnet: I think that might break the GKE deployment too

infra/docker/jobservice/Dockerfile
woop: @tsotnet I think this is the type of tech debt that slowly erodes the product experience. Aren't we already installing this 1+GB image for the Jupyter image? I would love if we could make this one more slim. I am also worried that we will keep delaying cleaning this up until its merged in, so I am more in favour of building our own light weight image.

tsotnet: I don't remember anymore, just some runtime or build time errors when installing things. Nothing that can't be resolved though, just a time sink.

tsotnet: @woop created an issue here https://github.com/feast-dev/feast/issues/1117

sdk/python/feast/pyspark/launchers/standalone/local.py
```diff
@@ -22,6 +22,10 @@
     StreamIngestionJobParameters,
 )
 
+# In-memory cache of Spark jobs
+# This is necessary since we can't query Spark jobs in local mode
+JOB_CACHE = {}
```
oavdeev, line 27: I'd check that python grpc server doesn't use a process pool or fork or something like that, so this is indeed the only copy of JOB_CACHE. I know in e.g. uWSGI it may be an issue, since it runs multiple python processes for you.

tsotnet: I haven't found any related documentation regarding multiprocessing & grpc. Also, when testing things locally I didn't see inconsistent outputs in list_jobs. I should have seen different answers on different requests if it was using multiprocessing.

infra/docker/jobservice/Dockerfile
woop: Is this still pending an update?

tsotnet: I do think this shouldn't be here now that we have both gcp and aws, but I'm not comfortable making changes to this since I'm not able to test if things continue working on gcp. Also it's not just this local change, we need to look into how we configure helm charts as well, since there are bunch of gcp-related flags in helm charts where they shouldn't be. I think cleaning gcp infra setup to make it similar to current AWS setup (where everything aws-related is set from terraform) should be a separate piece of work.

woop: Are you saying that the Feast client doesn't work without these vars set? @pyalex can you confirm?

sdk/python/feast/remote_job.py
```diff
@@ -128,3 +130,21 @@ def __init__(
         job_id: str,
     ):
         super().__init__(service, grpc_extra_param_provider, job_id)
+
+
+def get_remote_job_from_proto(
```
woop, line 135: Can you please add a full docstring?

sdk/python/feast/pyspark/launchers/standalone/local.py
```diff
@@ -215,42 +219,56 @@ def historical_feature_retrieval(
...
     ) -> RetrievalJob:
         job_id = str(uuid.uuid4())
-        return StandaloneClusterRetrievalJob(
+        job = StandaloneClusterRetrievalJob(
             job_id,
             job_params.get_name(),
             self.spark_submit(job_params),
             job_params.get_destination_path(),
         )
+        JOB_CACHE[job_id] = job  # type: ignore
```
woop, line 228: Do we really have to ignore types?

infra/docker/jobservice/Dockerfile
woop: @tsotnet What prevents us from
1. Adding this configuration to Docker Compose
2. Adding this configuration to a Helm values.yaml file (not the chart)

At least all tech debt is easy to find and clean up in the same place, right?

sdk/python/feast/pyspark/launchers/standalone/local.py
tsotnet: Yes, mypy is acting up because it's detecting the JOB_CACHE by first assignment of job and it's complaining on the second assignment (they have different types, e.g. RetrievalJob vs BatchIngestionJob). mypy isn't able to detect the shared parent class unfortunately.

sdk/python/feast/remote_job.py
tsotnet: ok

sdk/python/feast/pyspark/launchers/standalone/local.py
woop: Does Union[TypeA, TypeB] work?

tsotnet: I thought it wasn't possible to define types of global variables, seems like I was incorrect. Defining the type explicitly helped. Will change this too.

tsotnet: What I meant about inheritance is that it's not able to "update" the type dynamically. e.g. if mypy infers from 1 assignment that the type is Dict[str, B], and from 2nd assignment it sees that type is Dict[str, C], while B & C are subclasses of A, it's not able to update the type to Dict[str, A].

infra/docker/jobservice/Dockerfile
tsotnet: Discussed this in the standup; resolving the conversation

sdk/python/feast/job_service.py
```diff
@@ -62,9 +68,15 @@ def _job_to_proto(self, spark_job: SparkJob) -> JobProto:
 
     def StartOfflineToOnlineIngestionJob(self, request, context):
         """Start job to ingest data from offline store into online store"""
-        context.set_code(grpc.StatusCode.UNIMPLEMENTED)
-        context.set_details("Method not implemented!")
-        raise NotImplementedError("Method not implemented!")
+        feature_table = self.client.get_feature_table(
+            request.table_name, request.project
+        )
+        job = self.client.start_offline_to_online_ingestion(
```
pyalex, line 74: can we use `launcher.start_offline_to_online` instead of client here?

pyalex: the reason is: client may have different (user-oriented) API (like accepting strings instead of feature table, dataframes instead of source) and it's not really idempotent (and being already called once)

tsotnet: Done in the fork: https://github.com/feast-dev/feast/pull/1129

tsotnet:
Closing this and instead opening a PR in a fork: https://github.com/feast-dev/feast/pull/1129

PR #1114, opened 2020-10-30
Add feature table deletion

terryyylim:

This PR enables feature table deletion. The following has been implemented:
- `is_deleted` flag to check if a table is deleted or not
- Updated feature table metadata to include hash to determine if job needs to be restarted
- Added `delete_feature_table` method to Python SDK

Fixes #

```release-note
Users can now delete feature tables.
```

### DIFF
--- a/core/src/main/java/feast/core/model/FeatureTable.java
+++ b/core/src/main/java/feast/core/model/FeatureTable.java
@@ -16,6 +16,8 @@
  */
 package feast.core.model;
 
+import com.google.common.primitives.Longs;
+import com.google.protobuf.ByteString;
 import com.google.protobuf.Duration;
 import com.google.protobuf.Timestamp;
 import feast.core.dao.EntityRepository;
@@ -107,6 +109,12 @@ public class FeatureTable extends AbstractTimestampEntity {
   @Column(name = "revision", nullable = false)
   private int revision;
 
+  @Column(name = "is_deleted", nullable = false)
+  private boolean isDeleted;
+
+  @Column(name = "metadata_hash", nullable = false)
+  private long metadataHash;
+
   public FeatureTable() {};
 
   /**
@@ -146,6 +154,13 @@ public class FeatureTable extends AbstractTimestampEntity {
     if (!spec.getStreamSource().equals(DataSourceProto.DataSource.getDefaultInstance())) {
       table.setStreamSource(DataSource.fromProto(spec.getStreamSource()));
     }
+
+    table.setMetadataHash(
+        Objects.hash(
+            table.getEntities(),
+            table.getFeatures(),
+            table.getBatchSource(),
+            table.getStreamSource()));
 
     return table;
   }
@@ -202,6 +217,14 @@ public class FeatureTable extends AbstractTimestampEntity {
       this.streamSource = null;
     }
 
+    // Update hash
+    this.setMetadataHash(
+        Objects.hash(
+            this.getEntities(), this.getFeatures(), this.getBatchSource(), this.getStreamSource()));
+
+    // Set isDeleted to false
+    this.setDeleted(false);
+
     // Bump revision no.
     this.revision++;
   }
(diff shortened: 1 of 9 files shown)


core/src/main/java/feast/core/model/FeatureTable.java
```diff
@@ -107,6 +109,12 @@
   @Column(name = "revision", nullable = false)
   private int revision;
 
+  @Column(name = "is_deleted", nullable = false)
+  private boolean isDeleted;
+
+  @Column(name = "metadata_hash", nullable = false)
```
pyalex, line 115: Can we just calculate this hash when needed? why store it in postgres?

core/src/main/java/feast/core/model/FeatureTable.java
```diff
@@ -147,6 +155,13 @@ public static FeatureTable fromProto(
       table.setStreamSource(DataSource.fromProto(spec.getStreamSource()));
     }
 
+    table.setMetadataHash(
+        Objects.hash(
```
pyalex, line 159: won't it be more stable to use hash of proto?
of course we would need to presort all lists

core/src/main/java/feast/core/model/FeatureTable.java
```diff
@@ -202,6 +217,14 @@ public void updateFromProto(
       this.streamSource = null;
     }
 
+    // Update hash
+    this.setMetadataHash(
```
pyalex, line 221: it's better to define hash function in one place

core/src/main/java/feast/core/model/FeatureTable.java
```diff
@@ -279,6 +311,11 @@ public boolean hasAllLabels(Map<String, String> labelsFilter) {
     return TypeConversion.convertJsonStringToMap(getLabelsJSON());
   }
 
+  public void delete() {
+    this.setDeleted(true);
+    this.setRevision(0);
```
pyalex, line 316: how revision is being used?

core/src/main/java/feast/core/model/FeatureTable.java
woop: I was just about to say the same thing.

core/src/main/java/feast/core/model/FeatureTable.java
terryyylim: Revision is initialized whenever a feature table is created. As our deletion keeps the entry in the database, we should reset it to 0 since future apply to this deleted feature table would yield a new one.

core/src/main/java/feast/core/model/FeatureTable.java
```diff
@@ -279,6 +283,35 @@ public boolean hasAllLabels(Map<String, String> labelsFilter) {
...
+    }
+
+    FeatureTableSpec featureTableSpec =
+        FeatureTableSpec.newBuilder()
+            .addAllEntities(sortedEntities)
+            .addAllFeatures(sortedFeatureSpecs)
+            .setBatchSource(getBatchSource().toProto())
+            .setStreamSource(streamSource)
+            .build();
+    return featureTableSpec.toByteString();
```
pyalex, line 312: hash function is missing

core/src/main/java/feast/core/model/FeatureTable.java
```diff
@@ -279,6 +283,35 @@ public boolean hasAllLabels(Map<String, String> labelsFilter) {
...
+    DataSourceProto.DataSource streamSource = DataSourceProto.DataSource.getDefaultInstance();
+    if (getStreamSource() != null) {
+      streamSource = getStreamSource().toProto();
+    }
+
+    FeatureTableSpec featureTableSpec =
+        FeatureTableSpec.newBuilder()
+            .addAllEntities(sortedEntities)
+            .addAllFeatures(sortedFeatureSpecs)
+            .setBatchSource(getBatchSource().toProto())
```
pyalex, line 309: max age also should be here

terryyylim: Added.

pyalex (approved):

PR #1140, opened 2020-11-05
Implement Job Service control loop for stream ingestion jobs

tsotnet:
In this PR we implement the control loop in Job Service, which runs in the background thread and ensures that stream ingestion jobs are always running. It does this by generating a deterministic hash for each stream ingestion job and ensuring that all jobs with these hashes are in STARTING or RUNNING states, and also by ensuring that all other Spark jobs are terminated.
The implementation includes standalone & EMR modes. I haven't added the implementation for dataproc yet.

```release-note
You can now enable Feast Job Service's control loop by setting `FEAST_JOB_SERVICE_ENABLE_CONTROL_LOOP` env variable to `True`. This will start a background thread in Job Service, ensuring that all stream ingestion jobs are automatically running. This mean you won't have to call `client.start_stream_to_online_ingestion` manually again.
```

### DIFF
--- a/sdk/python/feast/client.py
+++ b/sdk/python/feast/client.py
@@ -18,7 +18,7 @@ import shutil
 import uuid
 from datetime import datetime
 from itertools import groupby
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import grpc
 import pandas as pd
@@ -90,9 +90,10 @@ from feast.loaders.ingest import (
     _write_partitioned_table_from_source,
 )
 from feast.online_response import OnlineResponse, _infer_online_entity_rows
-from feast.pyspark.abc import RetrievalJob, SparkJob
+from feast.pyspark.abc import RetrievalJob, SparkJob, StreamIngestionJob
 from feast.pyspark.launcher import (
     get_job_by_id,
+    get_stream_to_online_ingestion_params,
     list_jobs,
     stage_dataframe,
     start_historical_feature_retrieval_job,
@@ -1098,12 +1099,15 @@ class Client:
             )
 
     def start_stream_to_online_ingestion(
-        self, feature_table: FeatureTable, extra_jars: Optional[List[str]] = None,
+        self,
+        feature_table: FeatureTable,
+        extra_jars: Optional[List[str]] = None,
+        project: str = None,
     ) -> SparkJob:
         if not self._use_job_service:
             return start_stream_to_online_ingestion(
                 client=self,
-                project=self.project,
+                project=project or self.project,
                 feature_table=feature_table,
                 extra_jars=extra_jars or [],
             )
@@ -1113,8 +1117,86 @@ class Client:
             )
             response = self._job_service.StartStreamToOnlineIngestionJob(request)
             return RemoteStreamIngestionJob(
-                self._job_service, self._extra_grpc_params, response.id,
+                self._job_service, self._extra_grpc_params, response.id
             )
+
+    def _get_expected_job_hash_to_table_refs(
+        self, all_projects: bool
+    ) -> Dict[str, Tuple[str, str]]:
+        """
+        Checks all feature tables for the requires project(s) and determines all required stream
+        ingestion jobs from them. Outputs a map of the expected job_hash to a tuple of (project, table_name).
+
+        Args:
+            all_projects (bool): If true, runs the check for all project.
+                Otherwise only checks the current project.
+
+        Returns:
+            Dict[str, Tuple[str, str]]: Map of job_hash -> (project, table_name) for expected stream ingestion jobs
+        """
+        job_hash_to_table_refs = {}
+
+        projects = self.list_projects() if all_projects else [self.project]
+        for project in projects:
+            feature_tables = self.list_feature_tables(project)
+            for feature_table in feature_tables:
+                if feature_table.stream_source is not None:
+                    params = get_stream_to_online_ingestion_params(
+                        self, project, feature_table, []
+                    )
+                    job_hash = params.get_job_hash()
+                    job_hash_to_table_refs[job_hash] = (project, feature_table.name)
+
+        return job_hash_to_table_refs
+
+    def ensure_stream_ingestion_jobs(self, all_projects: bool = False):
+        """Ensures all required stream ingestion jobs are running and cleans up the unnecessary jobs.
+
+        More concretely, it will determine
+        - which stream ingestion jobs are running
+        - which stream ingestion jobs should be running
+        And it'll do 2 kinds of operations
+        - Cancel all running jobs that should not be running
+        - Start all non-existent jobs that should be running
+
+        Args:
+            all_projects (bool, optional): If true, runs the check for all project.
+                Otherwise only checks the current project. Defaults to False.
+        """
+        expected_job_hash_to_table_refs = self._get_expected_job_hash_to_table_refs(
+            all_projects
+        )
+        expected_job_hashes = set(expected_job_hash_to_table_refs.keys())
+
+        jobs_by_hash: Dict[str, StreamIngestionJob] = {}
+        for job in self.list_jobs(include_terminated=False):
+            if isinstance(job, StreamIngestionJob):
+                jobs_by_hash[job.get_hash()] = job
+
+        existing_job_hashes = set(jobs_by_hash.keys())
+
+        job_hashes_to_cancel = existing_job_hashes - expected_job_hashes
+        job_hashes_to_start = expected_job_hashes - existing_job_hashes
+
+        logging.info(
+            f"existing_job_hashes = {sorted(list(existing_job_hashes))} expected_job_hashes = {sorted(list(expected_job_hashes))}"
+        )
+
+        for job_hash in job_hashes_to_cancel:
+            job = jobs_by_hash[job_hash]
+            logging.info(
+                f"Cancelling a stream ingestion job with job_hash={job_hash} job_id={job.get_id()} status={job.get_status()}"
+            )
+            job.cancel()
+
+        for job_hash in job_hashes_to_start:
+            # Any job that we wish to start should be among expected table refs map
+            project, table_name = expected_job_hash_to_table_refs[job_hash]
+            logging.info(
+                f"Starting a stream ingestion job for project={project}, table_name={table_name} with job_hash={job_hash}"
+            )
+            feature_table = self.get_feature_table(name=table_name, project=project)
+            self.start_stream_to_online_ingestion(feature_table, [], project=project)
 
     def list_jobs(self, include_terminated: bool) -> List[SparkJob]:
         if not self._use_job_service:
--- a/sdk/python/feast/job_service.py
+++ b/sdk/python/feast/job_service.py
@@ -236,7 +236,7 @@ def start_job_service():
 
     client = feast.Client()
 
-    if client._config.get(CONFIG_JOB_SERVICE_ENABLE_CONTROL_LOOP) != "False":
+    if client._config.getboolean(CONFIG_JOB_SERVICE_ENABLE_CONTROL_LOOP):
         # Start the control loop thread only if it's enabled from configs
         thread = threading.Thread(target=run_control_loop, daemon=True)
         thread.start()
--- a/sdk/python/feast/job_service.py
+++ b/sdk/python/feast/job_service.py
@@ -5,7 +5,6 @@ import threading
 import time
 import traceback
 from concurrent.futures import ThreadPoolExecutor
-from typing import Dict, Tuple
 
 import grpc
 
@@ -17,11 +16,6 @@ from feast.core.JobService_pb2 import (
     GetHistoricalFeaturesRequest,
     GetHistoricalFeaturesResponse,
     GetJobResponse,
-)
-from feast.core.JobService_pb2 import Job as JobProto
-from feast.core.JobService_pb2 import (
-    JobStatus,
-    JobType,
     ListJobsResponse,
     StartOfflineToOnlineIngestionJobRequest,
     StartOfflineToOnlineIngestionJobResponse,
@@ -29,18 +23,11 @@ from feast.core.JobService_pb2 import (
     StartStreamToOnlineIngestionJobResponse,
 )
 from feast.data_source import DataSource
-from feast.pyspark.abc import (
-    BatchIngestionJob,
-    RetrievalJob,
-    SparkJob,
-    SparkJobStatus,
-    StreamIngestionJob,
-)
+from feast.pyspark.abc import StreamIngestionJob
 from feast.pyspark.launcher import (
     get_job_by_id,
     get_stream_to_online_ingestion_params,
     list_jobs,
-    list_jobs_by_hash,
     start_historical_feature_retrieval_job,
     start_offline_to_online_ingestion,
     start_stream_to_online_ingestion,
@@ -55,33 +42,6 @@ from feast.third_party.grpc.health.v1.HealthService_pb2 import (
 class JobServiceServicer(JobService_pb2_grpc.JobServiceServicer):
     def __init__(self, client):
         self.client = client
-
-    def _job_to_proto(self, spark_job: SparkJob) -> JobProto:
-        job = JobProto()
-        job.id = spark_job.get_id()
-        status = spark_job.get_status()
-        if status == SparkJobStatus.COMPLETED:
-            job.status = JobStatus.JOB_STATUS_DONE
-        elif status == SparkJobStatus.IN_PROGRESS:
-            job.status = JobStatus.JOB_STATUS_RUNNING
-        elif status == SparkJobStatus.FAILED:
-            job.status = JobStatus.JOB_STATUS_ERROR
-        elif status == SparkJobStatus.STARTING:
-            job.status = JobStatus.JOB_STATUS_PENDING
-        else:
-            raise ValueError(f"Invalid job status {status}")
-
-        if isinstance(spark_job, RetrievalJob):
-            job.type = JobType.RETRIEVAL_JOB
-            job.retrieval.output_location = spark_job.get_output_file_uri(block=False)
-        elif isinstance(spark_job, BatchIngestionJob):
-            job.type = JobType.BATCH_INGESTION_JOB
-        elif isinstance(spark_job, StreamIngestionJob):
-            job.type = JobType.STREAM_INGESTION_JOB
-        else:
-            raise ValueError(f"Invalid job type {job}")
-
-        return job
 
     def StartOfflineToOnlineIngestionJob(
         self, request: StartOfflineToOnlineIngestionJobRequest, context
@@ -133,13 +93,11 @@ class JobServiceServicer(JobService_pb2_grpc.JobServiceServicer):
                 self.client, request.project, feature_table, []
             )
             job_hash = params.get_job_hash()
-            jobs_by_hash = list_jobs_by_hash(self.client, include_terminated=False)
-            if job_hash not in jobs_by_hash:
-                raise RuntimeError(
-                    "Feast Job Service has control loop enabled, but couldn't find the existing stream ingestion job for the given FeatureTable"
-                )
-            return StartStreamToOnlineIngestionJobResponse(
-                id=jobs_by_hash[job_hash].get_id()
+            for job in list_jobs(include_terminated=True, client=self.client):
+                if isinstance(job, StreamIngestionJob) and job.get_hash() == job_hash:
+                    return StartStreamToOnlineIngestionJobResponse(id=job.get_id())
+            raise RuntimeError(
+                "Feast Job Service has control loop enabled, but couldn't find the existing stream ingestion job for the given FeatureTable"
             )
 
         # TODO: add extra_jars to request
@@ -156,7 +114,7 @@ class JobServiceServicer(JobService_pb2_grpc.JobServiceServicer):
         jobs = list_jobs(
             include_terminated=request.include_terminated, client=self.client
         )
-        return ListJobsResponse(jobs=[self._job_to_proto(job) for job in jobs])
+        return ListJobsResponse(jobs=[job.to_proto() for job in jobs])
 
     def CancelJob(self, request, context):
         """Stop a single job"""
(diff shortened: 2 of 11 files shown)


sdk/python/feast/job_service.py
```diff
@@ -117,6 +123,22 @@ def StartStreamToOnlineIngestionJob(
         feature_table = self.client.get_feature_table(
             request.table_name, request.project
         )
+
+        if self.client._config.get(CONFIG_JOB_SERVICE_ENABLE_CONTROL_LOOP) != "False":
```
oavdeev, line 127:
```suggestion
        if self.client._config.getboolean(CONFIG_JOB_SERVICE_ENABLE_CONTROL_LOOP):
```

sdk/python/feast/job_service.py
```diff
@@ -164,10 +234,19 @@ def start_job_service():
     log_fmt = "%(asctime)s %(levelname)s %(message)s"
     logging.basicConfig(level=logging.INFO, format=log_fmt)
 
+    client = feast.Client()
+
+    if client._config.get(CONFIG_JOB_SERVICE_ENABLE_CONTROL_LOOP) != "False":
```
oavdeev, line 239:
```suggestion
    if client._config.getboolean(CONFIG_JOB_SERVICE_ENABLE_CONTROL_LOOP):
```

sdk/python/feast/pyspark/abc.py
```diff
@@ -536,3 +547,7 @@ def get_job_by_id(self, job_id: str) -> SparkJob:
     @abc.abstractmethod
     def list_jobs(self, include_terminated: bool) -> List[SparkJob]:
         raise NotImplementedError
+
+    @abc.abstractmethod
+    def list_jobs_by_hash(self, include_terminated: bool) -> Dict[str, SparkJob]:
+        raise NotImplementedError
```
oavdeev, line 553: I don't feel super strongly about this, but it may be a bit more elegant to have `StreamIngestionJob` interface implement `get_job_hash()`, and not extend `JobLauncher` interface at all. 

Otherwise we have a situation where some job attributes are part of the XXJob interface (like output location for `RetrievalJob`) and some are not (like job hash here).

sdk/python/feast/job_service.py
```diff
@@ -145,6 +167,54 @@ def GetJob(self, request, context):
...
+
+        for job_hash in job_hashes_to_start:
+            project, table_name = table_metadata[job_hash]
+            logging.info(
+                f"Starting a stream ingestion job for project={project}, table_name={table_name} with job_hash={job_hash}"
+            )
+            feature_table = client.get_feature_table(name=table_name, project=project)
+            start_stream_to_online_ingestion(client, project, feature_table, [])
+
+        time.sleep(1)
```
oavdeev, line 215: One recurring issue I've seen with daemon threads is obscure bugs when the thread dies for whatever reason (unexpected exceptions an whatnot). The main process will happily chug along, and you may only notice days later that your control loop isn't running anymore, or not notice this at all.

Is there any robust way to avoid that here? Maybe something like `try ... finally: sys.exit(1)` ?

sdk/python/feast/job_service.py
```diff
@@ -145,6 +167,54 @@ def GetJob(self, request, context):
         return GetJobResponse(job=self._job_to_proto(job))
 
 
+def run_control_loop():
```
woop, line 170: should this be `run` or `start`?

sdk/python/feast/job_service.py
```diff
@@ -145,6 +167,54 @@ def GetJob(self, request, context):
...
+    """Runs control loop that continuously ensures that correct jobs are being run."""
```
woop, line 171: I think we need to be more explicit about which `jobs` are being run, and what the control loop will do (killing, restarting)

sdk/python/feast/job_service.py
```diff
@@ -145,6 +167,54 @@ def GetJob(self, request, context):
...
+    logging.info(
+        "Feast Job Service is starting a control loop in a background thread..."
```
woop, line 173: This log is a bit vague? What do I as a a Feast admin understand from this?

sdk/python/feast/pyspark/launchers/standalone/local.py
```diff
@@ -230,7 +293,7 @@ def historical_feature_retrieval(
             self.spark_submit(job_params),
             job_params.get_destination_path(),
         )
-        JOB_CACHE[job_id] = job
+        JOB_CACHE.add_job(job_id, None, job)
```
woop, line 296: By the way, why is `JOB_CACHE` all caps?

sdk/python/feast/job_service.py
```diff
@@ -145,6 +167,54 @@ def GetJob(self, request, context):
...
+    )
+    client = feast.Client()
```
woop, line 175: So how is this client different from `self.client`?

sdk/python/feast/job_service.py
```diff
@@ -145,6 +167,54 @@ def GetJob(self, request, context):
...
+    while True:
+        # Map of job hash -> (project, table_name)
+        table_metadata: Dict[str, Tuple[str, str]] = dict()
+        # Job hashes that should exist
+        final_job_hashes: Set[str] = set()
```
woop, line 180: Would `expected_job_hashes` be a better name?

sdk/python/feast/job_service.py
```diff
@@ -145,6 +167,54 @@ def GetJob(self, request, context):
...
 
+def run_control_loop():
+    """Runs control loop that continuously ensures that correct jobs are being run."""
+    logging.info(
+        "Feast Job Service is starting a control loop in a background thread..."
+    )
+    client = feast.Client()
+    while True:
+        # Map of job hash -> (project, table_name)
+        table_metadata: Dict[str, Tuple[str, str]] = dict()
```
woop, line 178: We try and avoid using the word `metadata` since its typically used when people cant think of a good name. Can we be more descriptive with this variable? `job_hash_to_table_references`?

sdk/python/feast/job_service.py
```diff
@@ -145,6 +167,54 @@ def GetJob(self, request, context):
...
+        # Job hashes that should exist
+        final_job_hashes: Set[str] = set()
+
+        for project in client.list_projects():
```
woop, line 182: I think this `run_control_loop` function can be written a little bit cleaner by pulling out some parts of these deeply nested loops into their own private methods

oavdeev: Just as another 2c, at least until this becomes too unwieldy, I would rather err on the side of having more logic inline rather than extracting it into helper functions here ([see this](http://number-none.com/blow/blog/programming/2014/09/26/carmack-on-inlined-code.html) )

sdk/python/feast/job_service.py
tsotnet: Didn't want to use the same object from 2 different threads, and it didn't look like having 2 different client objects added a lot of overhead.

sdk/python/feast/pyspark/launchers/standalone/local.py
tsotnet: It's a global constant variable (in a sense that we assign value to it once and never change the reference). Do you have a different naming convention for such variables?

sdk/python/feast/job_service.py
tsotnet: Suggestion?

sdk/python/feast/job_service.py
tsotnet: got it

sdk/python/feast/job_service.py
tsotnet: either way is fine for me, I can change it to start

sdk/python/feast/pyspark/abc.py
tsotnet: sure, I can do that

sdk/python/feast/job_service.py
tsotnet: yeah we need to keep it alive, good catch...

sdk/python/feast/pyspark/abc.py
tsotnet: hmm I thought about this and I think one big difference is that get_output_file_uri is a user-facing feature, while job hash is purely implementation, so I think it's better to not expose that as part of the job api. It'll also be more work to properly expose that method (e.g. we'd need to implement RemoteStreamIngestionJob too)

oavdeev: I mean `StreamIngestionJob` interface is not really public anyway right now. It is just a bit unorthogonal to have two `list_jobs*` almost-identical methods in the launcher interface where we could've made that part of the job itself.

This said, I don't feel super strongly so we can leave it as is too.

tsotnet: What do you mean it's not public? `client.start_stream_to_online_ingestion` returns that right?

sdk/python/feast/job_service.py
tsotnet: How about job_hash_to_table_refs to make it little shorter?

sdk/python/feast/pyspark/abc.py
oavdeev: ah you're right

sdk/python/feast/job_service.py
```diff
@@ -44,8 +53,8 @@
 
 
 class JobServiceServicer(JobService_pb2_grpc.JobServiceServicer):
-    def __init__(self):
-        self.client = feast.Client()
+    def __init__(self, client):
+        self.client = client
 
     def _job_to_proto(self, spark_job: SparkJob) -> JobProto:
```
woop, line 59: Question: Why is `_job_to_proto` a method and not a function?

sdk/python/feast/job_service.py
```diff
@@ -117,6 +126,22 @@ def StartStreamToOnlineIngestionJob(
...
+        if self.client._config.getboolean(CONFIG_JOB_SERVICE_ENABLE_CONTROL_LOOP):
+            # If the control loop is enabled, return existing stream ingestion job id instead of starting a new one
+            params = get_stream_to_online_ingestion_params(
+                self.client, request.project, feature_table, []
+            )
+            job_hash = params.get_job_hash()
+            jobs_by_hash = list_jobs_by_hash(self.client, include_terminated=False)
+            if job_hash not in jobs_by_hash:
+                raise RuntimeError(
+                    "Feast Job Service has control loop enabled, but couldn't find the existing stream ingestion job for the given FeatureTable"
```
woop, line 139: Can we use lower case for concepts as a rule of thumb? `feature table`.

sdk/python/feast/job_service.py
```diff
@@ -117,6 +126,22 @@ def StartStreamToOnlineIngestionJob(
         feature_table = self.client.get_feature_table(
             request.table_name, request.project
         )
+
+        if self.client._config.getboolean(CONFIG_JOB_SERVICE_ENABLE_CONTROL_LOOP):
```
woop, line 130: I'm looking at the code now and just wondering if it shouldn't be restructured in some ways. We have two approaches.
1. Keep as much logic and decision-making within the client. Make the Job Service a thin wrapper on the client. Don't break abstraction by accessing internal logic within the client.
2. Make the service/client modes operate at the same abstraction, but then don't use the client at all (create our own config objects, access methods directly).

@tsotnet it seems like this implementation mixes these two approaches. We create a `client` and use it for configuration, but then import internal methods (PySpark launcher and job methods) directly.

woop, line 130: @tsotnet it seems like we can take two approaches with the whole job service structure.
1. A thin wrapper on the Feast Client. Don't access internal methods directly and maintain abstraction. Don't duplicate logic in two places (inside and outside the client). 
2. Provide a shared library as the backend for both the client and service. So the service doesnt have any client instances, but it creates its own config, launches its own jobs, etc. It still uses the same methods as the client, but it just doesnt call the client.

It seems like the current approach is a mix of these two. We pull in the client for config, and then we access internal methods directly. Are we concerned about the internal dependencies we are creating here?

sdk/python/feast/job_service.py
```diff
@@ -117,6 +126,22 @@ def StartStreamToOnlineIngestionJob(
...
+            # If the control loop is enabled, return existing stream ingestion job id instead of starting a new one
+            params = get_stream_to_online_ingestion_params(
+                self.client, request.project, feature_table, []
+            )
+            job_hash = params.get_job_hash()
```
woop, line 135: Would it be better to use `__hash__()` on Python objects and to use a class of the job here instead of params?

sdk/python/feast/job_service.py
woop: Also, are we really going to raise an exception if the job doesn't exist when they run a `start` method? Shouldn't they be able to idempotently run this method?

sdk/python/feast/job_service.py
tsotnet: That wasn't written by me, but I can change it if necessary. @oavdeev any thoughts / reasons not to change this?

p.s. I don't like this becoming a public function on SparkJob because the users will see it, and it's not a publicly facing functionality. I think a private function on SparkJob would be better.

sdk/python/feast/job_service.py
```diff
@@ -145,6 +170,78 @@ def GetJob(self, request, context):
...
+                params = get_stream_to_online_ingestion_params(
+                    client, project, feature_table, []
+                )
+                job_hash = params.get_job_hash()
+                job_hash_to_table_refs[job_hash] = (project, feature_table.name)
+
+    return job_hash_to_table_refs
+
+
+def ensure_stream_ingestion_jobs(client: feast.Client):
```
woop, line 190: Please add docstrings to methods with types.

sdk/python/feast/job_service.py
```diff
@@ -145,6 +170,78 @@ def GetJob(self, request, context):
         return GetJobResponse(job=self._job_to_proto(job))
 
 
+def get_expected_job_hash_to_table_refs(client) -> Dict[str, Tuple[str, str]]:
```
woop, line 173: You are missing a type definition here and a complete docstring. Also, should this be a public method?

sdk/python/feast/job_service.py
woop: It really seems like this `ensure` function should be a method on the client itself and not specific to the job service.

sdk/python/feast/job_service.py
oavdeev: Yeah it could be a standalone function too. I didn't want to put it on `SparkJob` to keep it a pure interface. Besides, it is only used by JobServiceServicer.

sdk/python/feast/job_service.py
tsotnet: Not sure I agree with this. The launcher layer also extensively uses client, so they aren't clearly decoupled from each other. If we really want the good decoupling, not only should job service not use client, but launcher should also not use client at all. We can do that, but not sure this PR is the right place for it.

sdk/python/feast/job_service.py
tsotnet: Didn't want to reuse __hash__ which is intended for a different purpose (e.g. deduping in dictionaries & sets). The params class is more general to define its __hash__ for this purpose only.

sdk/python/feast/job_service.py
tsotnet: Can you elaborate, why? It seems like this is a specific functionality to Job Service. Why should client worry about this? Also, this'll make it difficult to decouple client & job service.

sdk/python/feast/job_service.py
tsotnet: Probably shouldn't be a public method. Do we need all functions to be compliant to the generated docstrings (with parameters & return type). Sometimes it's overly verbose when the function is simple to describe.

sdk/python/feast/job_service.py
woop: > I didn't want to put it on SparkJob to keep it a pure interface

Wouldn't it still be a pure interface? I mean we don't have to implement `to_proto()` on the ABC, but only the children.

sdk/python/feast/job_service.py
woop: > The launcher layer also extensively uses client, so they aren't clearly decoupled from each other.

Yes. I don't think it's a good idea to have the client in the launcher.

> If we really want the good decoupling, not only should job service not use client, but launcher should also not use client at all.

Well which approach do you think is preferrable? I am ok with the JS using the client, I just dont think we should mix the two approaches (use client and then bypass client).

> but not sure this PR is the right place for it.

Knowing that we don't want to increase the internal dependencies (bypassing Client), can we try and not make things worse with this PR at least?

sdk/python/feast/job_service.py
woop: We don't have to follow it religiously. For simple methods we can just have a sentence explaining the method.

sdk/python/feast/job_service.py
woop: `ensure` is the same as `start all jobs necessary to be started`. If we believe that starting a specific job is important (from client) then ensure is also important. In fact there is already a use case for this when you cant run stateful processes but you can launch jobs (Databricks scheduler or Github Actions launching Spark jobs on Databricks/EMR).

> Also, this'll make it difficult to decouple client & job service.
How so? Wouldn't it improve the decoupling if the logic is only in the client?

sdk/python/feast/pyspark/launchers/standalone/local.py
```diff
@@ -22,9 +23,71 @@
...
+
+    # This reentrant lock is necessary for multi-threading access
+    lock: threading.RLock
+
+    def __init__(self):
+        self.job_by_id = {}
+        self.hash_by_id = {}
+        self.lock = threading.RLock()
+
+    def add_job(self, job_id: str, job_hash: Optional[str], job: SparkJob) -> None:
```
woop, line 49: @oavdeev Is't it possible for us to simplify things to `def add_job(self, job: SparkJob)`? And have the `id` and `hash` be a property of the job? It seems like we are doing a lot of bookkeeping here with state that is somewhere derived from each other (id and hash from job)

sdk/python/feast/job_service.py
oavdeev: `__hash__` wouldn't really work here anyway since it would change if you restart the pod/process

sdk/python/feast/job_service.py
tsotnet: > I dont think it should be a method.
What do you mean? If it's not a method, you mean it should be a standalone function? Or do you mean that it shouldn't be a method of JobServiceServicer class specifically?

sdk/python/feast/job_service.py
tsotnet: Can you describe the race condition? The idea is that if the job doesn't already exist and the control loop is working in the background, it'll probably start this job soon. So here we don't start the spark job, hence no race condition. We would have the race condition if we tried to start the stream ingestion job here, since the control loop could also be trying to do the same at the same time.

Decided to throw an exception instead of returning response with empty job_id, because even if we return the response the `job.get_status()` and other methods will fail anyways, since job_id is empty. And that error will be more difficult to debug to users than the explicit exception.

sdk/python/feast/job_service.py
tsotnet: fyi for others: we discussed this in Zoom. Decided to move `ensure_stream_ingestion_jobs` in client and job service will call that. In a separate PR we will remove usage of client in launcher and replace it with config reader. In job service as well, we'll replace client with config object whenever possible.

sdk/python/feast/job_service.py
tsotnet: Oh I thought Willem meant overriding & implementing `__hash__` function instead of `get_job_hash`. Using existing `__hash__` directly would just not work as Oleg stated.

sdk/python/feast/job_service.py
tsotnet: tl;dr: discussed in Zoom, see above comment.

sdk/python/feast/pyspark/launchers/standalone/local.py
tsotnet: It's possible - I'm working on it. I'm combining this idea with what I've discussed with Oleg before. Basically, job hash should be part of `StreamIngestionJob`. In this case a lot of things become simpler to do - e.g. launcher won't have the ugly new `list_jobs_by_hash` method anymore.

sdk/python/feast/job_service.py
tsotnet: Got it, I'll add a more extensive comment since I'm moving this to the client and modifying it a little.

sdk/python/feast/client.py
```diff
@@ -1113,8 +1117,86 @@ def start_stream_to_online_ingestion(
...
+
+        More concretely, it will determine
+        - which stream ingestion jobs are running
+        - which stream ingestion jobs should be running
+        And it'll do 2 kinds of operations
+        - Cancel all running jobs that should not be running
+        - Start all non-existent jobs that should be running
+
+        Args:
+            all_projects (bool, optional): If true, runs the check for all project.
```
woop, line 1163: `run the check` is a bit vague, maybe `ensures stream ingestion jobs are running for all projects`?

sdk/python/feast/client.py
```diff
@@ -1113,8 +1117,86 @@ def start_stream_to_online_ingestion(
...
+            if isinstance(job, StreamIngestionJob):
+                jobs_by_hash[job.get_hash()] = job
+
+        existing_job_hashes = set(jobs_by_hash.keys())
+
+        job_hashes_to_cancel = existing_job_hashes - expected_job_hashes
+        job_hashes_to_start = expected_job_hashes - existing_job_hashes
+
+        logging.info(
+            f"existing_job_hashes = {sorted(list(existing_job_hashes))} expected_job_hashes = {sorted(list(expected_job_hashes))}"
```
woop, line 1182: How do you see a Feast admin using this information @tsotnet?

sdk/python/feast/job_service.py
woop: I meant overriding the __hash__ function so that the unique (and unchanging) properties of the job are used to create a hash. That way we can have `Job1 == Job1` and get away from writing custom getters/setters and storage for hash.

sdk/python/feast/pyspark/abc.py
```diff
@@ -451,18 +482,47 @@ def get_arguments(self) -> List[str]:
...
 
 class StreamIngestionJob(SparkJob):
     """
     Container for the streaming ingestion job result
     """
 
+    def get_hash(self) -> str:
+        """Gets the consistent hash of this stream ingestion job.
+
+        The hash needs to be persisted at the data processing layer, s.t.
```
woop, line 512: s.t?

sdk/python/feast/pyspark/abc.py
```diff
@@ -66,6 +70,27 @@ def cancel(self):
...
+    def to_proto(self) -> JobProto:
+        """Converts SparkJob to the protobuf object.
+
+        Returns:
+            JobProto: Converted protobuf object.
+        """
+        job = JobProto()
+        job.id = self.get_id()
+        status = self.get_status()
+        if status == SparkJobStatus.COMPLETED:
```
pyalex, line 82: What the difference between SparkJobStatus & JobStatusProto?
Why can't we have only proto version used everywhere instead of SparkJobStatus and omit this mapping?

sdk/python/feast/pyspark/abc.py
```diff
@@ -451,18 +482,47 @@ def get_arguments(self) -> List[str]:
...
+        return hashlib.md5(job_json.encode()).hexdigest()
+
 
 class BatchIngestionJob(SparkJob):
     """
     Container for the ingestion job result
     """
 
+    def to_proto(self) -> JobProto:
+        job = super().to_proto()
```
pyalex, line 499: Not really important, but I think start & end date also should be used in hash calculation for batch ingestion if we want to represent job identity by it, even though we wan't run this job in streaming loop

sdk/python/feast/job_service.py
oavdeev: I'd def prefer an explicit `get_job_hash()` and make it part of the params interface (like it is rn). If I was reading feast code, overriding `__hash__` in this matter to do this would be surprising to me.

sdk/python/feast/pyspark/abc.py
oavdeev: Let's keep `SparkJob` a pure abstract class, even if have to duplicate a couple lines of code. This version is substantially less readable to me than the previous one, since now `JobProto` construction logic is spread across the class hierarchy.

oavdeev: @pyalex i agree seems like there no reason why we shouldn't unify those. I'd address this in a separate PR though (after all, that enum wasn't introduced or even changed in this PR).

PR #1156, opened 2020-11-11
Jobservice control loop (based on #1140)

oavdeev:
See #1140 for start of the discussion. Compared to that PR, this addresses some comments and adds tests for the control loop logic.

```release-note
NONE
```

### DIFF
--- a/sdk/python/tests/test_streaming_control_loop.py
+++ b/sdk/python/tests/test_streaming_control_loop.py
@@ -16,7 +16,10 @@ from feast.entity import Entity
 from feast.feature import Feature
 from feast.feature_table import FeatureTable
 from feast.job_service import ensure_stream_ingestion_jobs
-from feast.pyspark.launchers.standalone import StandaloneClusterLauncher
+from feast.pyspark.launchers.standalone import (
+    StandaloneClusterLauncher,
+    reset_job_cache,
+)
 from feast.value_type import ValueType
 from tests.feast_core_server import CoreServicer as MockCoreServicer
 
@@ -37,7 +40,6 @@ def mock_server(servicer, add_fn):
         server.stop(None)
 
 
-CORE_URL = "core.feast.example.com"
 SERVING_URL = "serving.example.com"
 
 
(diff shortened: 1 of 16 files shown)


sdk/python/tests/test_streaming_control_loop.py
```diff
@@ -0,0 +1,184 @@
...
+
+    try:
+        address = "localhost:%d" % port
+        with grpc.insecure_channel(address):
+            yield address
+    finally:
+        server.stop(None)
+
+
+CORE_URL = "core.feast.example.com"
```
woop, line 40: Do we need these?

sdk/python/tests/test_streaming_control_loop.py
```diff
@@ -0,0 +1,184 @@
...
+            # We expect a streaming job to be created for the new Feature Table.
+            assert len(client.list_jobs(include_terminated=False)) == 1
+            assert len(processes) == 1
+
+            first_job_id = client.list_jobs(include_terminated=False)[0].get_id()
+
+            # Pretend that the streaming job has terminated for no reason.
+            processes[0].kill()
+
+            # The control loop is expected to notice the killed job and start it again.
```
woop, line 150: Should we perhaps add one assert here to test that a job has terminated?

sdk/python/tests/test_streaming_control_loop.py
oavdeev-tt: You're right we don't

sdk/python/tests/test_streaming_control_loop.py
oavdeev-tt: Hmm not sure what code path would that help test. We just terminated it one line above this so we know it is dead. The assert below checks that there is one terminated and one started job after we ran `ensure_stream_ingestion_jobs`

woop: Fair enough.

woop (approved):

PR #1158, opened 2020-11-12
Fix github workflow deprecating env variable

terryyylim:

Github actions will be deprecating `set-env` as described in this [blog post](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/). Thus, we need to update it before it causes issues to our ci/cd.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/master_only.yml
+++ b/.github/workflows/master_only.yml
@@ -28,7 +28,7 @@ jobs:
           --archive-uri ${MAVEN_CACHE} \
           --output-dir .
       - name: Get version
-        run: echo ::set-env name=RELEASE_VERSION::${GITHUB_REF#refs/*/}
+        run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
       - name: Build image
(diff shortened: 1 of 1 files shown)


.github/workflows/master_only.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           --archive-uri ${MAVEN_CACHE} \
           --output-dir .
       - name: Get version
-        run: echo ::set-env name=RELEASE_VERSION::${GITHUB_REF#refs/*/}
+        run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
```
woop, line 31: Not really sure what this command now does...

terryyylim: Recommended upgrade from [github workflow docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable).

pyalex (approved):

PR #1160, opened 2020-11-13
Add dataproc executor resource config

terryyylim:

Currently one job can allocate too many resources (by default it will take amount of executors equal to partitions), which can break parallelization. This PR allows configuring of resource allocation per one job, so that many jobs can share a cluster.

Fixes #

```release-note
Users can now configure amount of resources to allocate for their spark ingestion jobs.
```


sdk/python/feast/pyspark/launchers/gcloud/dataproc.py
```diff
@@ -186,7 +186,14 @@ class DataprocClusterLauncher(JobLauncher):
     JOB_TYPE_LABEL_KEY = "feast_job_type"
 
     def __init__(
-        self, cluster_name: str, staging_location: str, region: str, project_id: str,
+        self,
+        cluster_name: str,
+        staging_location: str,
+        region: str,
+        project_id: str,
+        executor_instances: str,
```
woop, line 194: Why positional arguments? Are these required? We don't have defaults?

woop:
`Users can now configure amount of resources to allocate for their spark ingestion jobs.`

How do I as a user know how to set this configuration and what the possible options are?

pyalex:
@woop I think it's not for users, but rather for feast admins and those options will be configured on jobservice side

woop:
> @woop I think it's not for users, but rather for feast admins and those options will be configured on jobservice side

Yea I know, but my question is how do I know what configuration can be set? We need to start documenting the configuration options.

sdk/python/feast/pyspark/launchers/gcloud/dataproc.py
```diff
@@ -200,7 +200,14 @@ class DataprocClusterLauncher(JobLauncher):
     JOB_HASH_LABEL_KEY = "feast_job_hash"
 
     def __init__(
-        self, cluster_name: str, staging_location: str, region: str, project_id: str,
+        self,
+        cluster_name: str,
+        staging_location: str,
+        region: str,
+        project_id: str,
+        executor_instances: str = "2",
```
pyalex, line 208: can we rather rely on defaults in `constants.py` than here?

sdk/python/feast/pyspark/launcher.py
```diff
@@ -59,6 +62,9 @@ def _dataproc_launcher(config: Config) -> JobLauncher:
         config.get(CONFIG_SPARK_STAGING_LOCATION),
         config.get(CONFIG_SPARK_DATAPROC_REGION),
         config.get(CONFIG_SPARK_DATAPROC_PROJECT),
+        config.get(CONFIG_SPARK_DATAPROC_EXECUTOR_INSTANCES),
```
pyalex, line 65: please use named arguments when amount of arguments is so high

pyalex (approved):

PR #1161, opened 2020-11-13
Expires Redis Keys based on Feature Table Max Age

khorshuheng:

Since the release of Feast 0.8, the Redis ingestion is performed via Spark ingestion module rather than the existing Redis storage module. Hence, #955 is not applicable to Feast 0.8. 

For the current implementation, instead of a global value of ttl which would be applicable to all Redis key, the ttl will be computed based on the max ages of all the feature tables associated with the Redis Key. For example, if two feature table are associated with the Redis key, the ttl will be computed as follows:

```
ttl = max(event_timestamp_feature_table_1 + max_age_feature_table_1, event_timestamp_feature_table_2 + max_age_feature_table_2) - current time
```

In the event where any of the feature table associated with the Redis key has max age set to 0, ttl will not be set.

Fixes #

```release-note
After this PR, the key stored on Redis will be deleted when it is older than event_timestamp + max age, unless the key is also associated with another feature table where max age is not specified, or the expiry time is in the future.
```

### DIFF
--- a/spark/ingestion/src/main/scala/feast/ingestion/stores/redis/HashTypePersistence.scala
+++ b/spark/ingestion/src/main/scala/feast/ingestion/stores/redis/HashTypePersistence.scala
@@ -16,16 +16,18 @@
  */
 package feast.ingestion.stores.redis
 
+import java.nio.charset.StandardCharsets
+import java.util
+
+import com.google.common.hash.Hashing
+import com.google.protobuf.Timestamp
+import com.google.protobuf.util.Timestamps
+import feast.ingestion.utils.TypeConversion
 import org.apache.spark.sql.Row
 import org.apache.spark.sql.types._
 import redis.clients.jedis.{Pipeline, Response}
-import java.nio.charset.StandardCharsets
-
-import com.google.common.hash.Hashing
 
 import scala.jdk.CollectionConverters._
-import com.google.protobuf.Timestamp
-import feast.ingestion.utils.TypeConversion
 
 /**
   * Use Redis hash type as storage layout. Every feature is stored as separate entry in Hash.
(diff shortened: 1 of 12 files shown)


woop:
@khorshuheng can you please add a description? This PR has no information.

How does this fit into what @ravdin is going? in #955

spark/ingestion/src/test/scala/feast/ingestion/StreamingPipelineIT.scala
```diff
@@ -136,16 +136,91 @@ class StreamingPipelineIT extends SparkSpec with ForAllTestContainer {
...
+            "_ex:driver-fs" -> new java.sql.Timestamp(
+              (r.getEventTimestamp.getSeconds + maxAge) * 1000
+            )
+          )
+        )
+        val keyTTL = jedis.ttl(encodedEntityKey).toLong
+        keyTTL should (be <= (r.getEventTimestamp.getSeconds + maxAge - ingestionTimeUnix / 1000) and be > 0L)
+      }
+
+      val increasedMaxAge               = 86400 * 2
```
pyalex, line 186: This seems like two tests in one. Can we keep split them?

spark/ingestion/src/main/scala/feast/ingestion/stores/redis/HashTypePersistence.scala
```diff
@@ -51,49 +57,104 @@ class HashTypePersistence(config: SparkRedisConfig) extends Persistence with Ser
...
+  ): Option[java.sql.Timestamp] = {
+    value.asScala.toMap
+      .map { case (key, value) =>
+        (key.map(_.toChar).mkString, value)
+      }
+      .get(timestampHashKey(config.namespace))
+      .map(value => decodeTimestamp(value))
   }
 
+  override def newExpiryTimestamp(
```
pyalex, line 135: This is not really persistence related method. Maybe move it to `RedisSinkRelation`?

pyalex: you can move `MAX_EXPIRED_TIMESTAMP` to package level

spark/ingestion/src/test/scala/feast/ingestion/StreamingPipelineIT.scala
pyalex, line 186: Can we have test for max age = 0 instead, since this flow was already tested with Batch Pipeline?

woop:
> In the event where any of the feature table associated with the Redis key has max age set to 0, ttl will not be set.

So how does an administrator force a TTL?

pyalex (approved):

PR #1166, opened 2020-11-16
Fix missing name variable instantiation

terryyylim:

Currently, in `ingest` method of python SDK client, if `feature_table` parameter is a str, the `name` variable will not be instantiated correctly. This PR addresses that.

Fixes #

```release-note
NONE
```


woop:
Should we add a test somewhere as well to prevent regression?

woop (approved):

PR #1172, opened 2020-11-16
S3 endpoint configuration #1169

### LINKED ISSUE #1169: S3 endpoint configuration
I want to use my on-premise installment of s3 (minio in my case). For this I need a way to provide custom `endpoint_url` for `boto3.client` initialization.

New configuration option like `FEAST_S3_ENDPOINT_URL`. And use it's value in `feast.staging.storage_client.S3Client.__init__` when creating s3 client. Default `None` value will ensure backward compatibility. 
To configure this option for spark jobs,  #1168 can be implemented

mike0sv:

Fixes #1169

```release-note
Added an option to configure S3 endpoint url
```


woop:
Thanks for this @mike0sv!

/ok-to-test

mike0sv:
Hey, @woop, some tests failed because I provided `None` as default value. There are two possible solutions, 1) add `allow_no_value=True` to `ConfigParser` so it will tolerate `None`s 2) Set default to `https://s3.amazonaws.com`, which is default value of `._endpoing` when you create `boto3.client('s3')`. Which do you prefer?

woop:
> allow_no_value

I prefer `allow_no_value=True`. Thanks for giving us options :100:

mike0sv:
Not sure, why this was closed and not sure why tests failed (seems to be not related to my changes). Please send help :)

woop:
> Not sure, why this was closed and not sure why tests failed (seems to be not related to my changes). Please send help :)

Can you please run `make format` prior to pushing your commit? That should fix `lint-python`

@pyalex is trying to fix the other test here: https://github.com/feast-dev/feast/pull/1173/files

mike0sv:
`lint-python` fails because of the `None`, that changed `DEAFULTS` type from `Dict[str, str]` to `Dict[str, Optional[str]]` and mypy raises an error for `deafults` argument for `ConfigParser`. Not sure where it gets that it should be `Mapping[str, str]` thou

mike0sv:
Mypy gets type annotation from here https://github.com/python/typeshed/blob/master/stdlib/3/configparser.pyi
So there is nothing we can do about it.

We can use the `https://s3.amazonaws.com` solution, or remove the default value from the `DEFAULTS` and add new argument `default` to `Config.get` to utilize `ConfigParser.get(..., fallback=x)` logic (basically like `dict.get`)

mike0sv:
...or maybe we can https://github.com/python/typeshed/issues/4773

woop:
@mike0sv I haven't had time to look at this in detail. One of the problems (that is somewhat related) that we want to still solve is that `DEFAULTS` are being persisted to `~/.feast/config` when users make changes to the configuration. So basically we need the defaults to be hardcoded in the Feast Python code base, not persisted to disk. So perhaps this is an opportunity to pull out the defaults somehow, instead of setting them as a section.

The trick, however, is to ensure that defaults are still returned when the user looks up keys using the getters.

I can have a look at this over the weekend, unless you want to think of a solution

> We can use the https://s3.amazonaws.com solution
I don't really like this approach. It seems brittle.

> and add new argument default to Config.get to utilize ConfigParser.get(..., fallback=x)

Not 100% clear what you mean, but does this line up what I said above?

mike0sv:
@woop not sure I get what you mean. 
What I see in code now: on `Config` initialization (when we create `Client` for example), a `ConfigParser` instance is created with defaults from `DEFAULTS` and this is written to `~/.feast/config` (only defaults). Then, when we need to get option value, `Config.get` is used, which uses value that were fromvided to `Client` constructor. If value was not provided, it uses defaults as a fallback.
Now, we can alter values using cli `feast config set` (the only usage of `Config.save` method)
After that, on `Config` initialization, it will read this value from `~/.feast/config`
Is this the desired behavior or is there something wrong that you want to change?

woop:
> Is this the desired behavior or is there something wrong that you want to change?

No, the desired behavior is to not write the defaults to disk. It should stay in the application. That allows us to change defaults for users without having the defaults pinned for users on disk.

mike0sv:
I proposed to add the possibility to define default values for some options not in `DEFAULTS` dict, but wherever you are tying to get configuration option. This is not very good in terms that you will have multiple locations for default values, but is more flexible. 
To achieve that, I proposed to rewrite `Config.get` like this
```
def get(self, option, default=...):
    try:
        .... # old code
     except option_not_found_or_something:
        if default is not ...:
            return default
         raise
```

If we add the logic to also get default value from `DEFAULTS` here, and will not use `defaults` argument of `ConfigParser`, this may solve this problem.
We will have our own "layer" of default values and they will not be written to disk

woop:
> I proposed to add the possibility to define default values for some options not in `DEFAULTS` dict, but wherever you are tying to get configuration option. This is not very good in terms that you will have multiple locations for default values, but is more flexible.
> To achieve that, I proposed to rewrite `Config.get` like this
> 
> ```
> def get(self, option, default=...):
>     try:
>         .... # old code
>      except option_not_found_or_something:
>         if default is not ...:
>             return default
>          raise
> ```
> 
> If we add the logic to also get default value from `DEFAULTS` here, and will not use `defaults` argument of `ConfigParser`, this may solve this problem.
> We will have our own "layer" of default values and they will not be written to disk

I'm happy with this approach. Although it would be great if we didn't have to implement it as a try/except. Could we implement a third layer of `vars=` https://github.com/feast-dev/feast/blob/3b5d30f81691f3ecf72ebb989f30fee8875dd94a/sdk/python/feast/config.py#L135?

so maybe something like `vars= {**DEFAULT, **FILE, **ENV, **CONSTRUCTOR}` (not real variable names of course)

Also, please note the changes in #1174

mike0sv:
Sounds reasonable. I'll wait for #1174 and merge it (together with #1173) then.

woop:
> Sounds reasonable. I'll wait for #1174 and merge it (together with #1173) then.

By the way, #1174 has been merged :)

mike0sv:
Unfortunately, the `vars` solution was not working, because `vars` takes precedence before the actual config, so passing `DEFAULTS` there were never gonna work. After that I tried to utilize `fallback` option, but this way I lost free type casting the `configparser` module provided. So I decided to approach the "defaults are saved to file" problem by hacking `_defaults` attribute for the time of writing. Added some tests and it seems to work

sdk/python/feast/config.py
```diff
@@ -113,24 +114,32 @@ def __init__(
...
 
         Returns: String option that is returned
 
         """
+        default = {option: default} if default is not _UNSET else {}
         return self._config.get(
             CONFIG_FILE_SECTION,
             option,
-            vars={**_get_feast_env_vars(), **self._options},
+            vars={
```
woop: By the way, I think you should submit this change as part of a separate PR please.

mike0sv:
.....

mike0sv:
...

woop:
> Unfortunately, the `vars` solution was not working, because `vars` takes precedence before the actual config, so passing `DEFAULTS` there were never gonna work. After that I tried to utilize `fallback` option, but this way I lost free type casting the `configparser` module provided. So I decided to approach the "defaults are saved to file" problem by hacking `_defaults` attribute for the time of writing. Added some tests and it seems to work

Noted, didn't realize that.

sdk/python/feast/config.py
mike0sv: Created new PR, it seems  it needs to be merged before this one

sdk/python/feast/loaders/ingest.py
```diff
@@ -166,7 +167,7 @@ def _read_table_from_source(
 
 
 def _upload_to_file_source(
-    file_url: str, with_partitions: bool, dest_path: str
+    file_url: str, with_partitions: bool, dest_path: str, config: Config
```
woop, line 170: I think the arguments in the docstring is out of date. Would you mind updating?

woop (approved):

PR #1174, opened 2020-11-18
Refactor configurable options and add sphinx docs

terryyylim:

Currently, constants used for `Config` class are declared separately from default values and the term is used interchangeably with config options, making it difficult to maintain as the number of configurable options increase.

This PR addresses the following:
- Address constants which have different parameter name to be the same
eg. ```CONFIG_SPARK_HISTORICAL_FEATURE_OUTPUT_FORMAT = "historical_feature_output_format"``` to
```CONFIG_HISTORICAL_FEATURE_OUTPUT_FORMAT = "historical_feature_output_format"```
- Add documentation for configurable options.
image

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/cli.py
+++ b/sdk/python/feast/cli.py
@@ -23,7 +23,7 @@ import yaml
 
 from feast.client import Client
 from feast.config import Config
-from feast.constants import ConfigOptions
+from feast.constants import ConfigOptions as opt
 from feast.entity import Entity
 from feast.feature_table import FeatureTable
 from feast.job_service import start_job_service
@@ -422,7 +422,7 @@ def stop_stream_to_online(feature_table: str):
     Stop stream to online sync job
     """
 
-    spark_launcher = Config().get(ConfigOptions.SPARK_LAUNCHER)
+    spark_launcher = Config().get(opt.SPARK_LAUNCHER)
 
     if spark_launcher == "emr":
         import feast.pyspark.aws.jobs
@@ -441,7 +441,7 @@ def list_jobs():
     """
     from tabulate import tabulate
 
-    spark_launcher = Config().get(ConfigOptions.SPARK_LAUNCHER)
+    spark_launcher = Config().get(opt.SPARK_LAUNCHER)
 
     if spark_launcher == "emr":
         import feast.pyspark.aws.jobs
--- a/sdk/python/feast/cli.py
+++ b/sdk/python/feast/cli.py
@@ -23,7 +23,7 @@ import yaml
 
 from feast.client import Client
 from feast.config import Config
-from feast.constants import CONFIG_SPARK_LAUNCHER
+from feast.constants import ConfigOptions
 from feast.entity import Entity
 from feast.feature_table import FeatureTable
 from feast.job_service import start_job_service
@@ -422,7 +422,7 @@ def stop_stream_to_online(feature_table: str):
     Stop stream to online sync job
     """
 
-    spark_launcher = Config().get(CONFIG_SPARK_LAUNCHER)
+    spark_launcher = Config().get(ConfigOptions.SPARK_LAUNCHER)
 
     if spark_launcher == "emr":
         import feast.pyspark.aws.jobs
@@ -441,7 +441,7 @@ def list_jobs():
     """
     from tabulate import tabulate
 
-    spark_launcher = Config().get(CONFIG_SPARK_LAUNCHER)
+    spark_launcher = Config().get(ConfigOptions.SPARK_LAUNCHER)
 
     if spark_launcher == "emr":
         import feast.pyspark.aws.jobs
--- a/sdk/python/feast/client.py
+++ b/sdk/python/feast/client.py
@@ -24,24 +24,7 @@ import grpc
 import pandas as pd
 
 from feast.config import Config
-from feast.constants import (
-    CONFIG_CORE_ENABLE_SSL_KEY,
-    CONFIG_CORE_SERVER_SSL_CERT_KEY,
-    CONFIG_CORE_URL_KEY,
-    CONFIG_ENABLE_AUTH_KEY,
-    CONFIG_GRPC_CONNECTION_TIMEOUT_DEFAULT_KEY,
-    CONFIG_JOB_SERVICE_ENABLE_SSL_KEY,
-    CONFIG_JOB_SERVICE_SERVER_SSL_CERT_KEY,
-    CONFIG_JOB_SERVICE_URL_KEY,
-    CONFIG_PROJECT_KEY,
-    CONFIG_SERVING_ENABLE_SSL_KEY,
-    CONFIG_SERVING_SERVER_SSL_CERT_KEY,
-    CONFIG_SERVING_URL_KEY,
-    CONFIG_SPARK_HISTORICAL_FEATURE_OUTPUT_FORMAT,
-    CONFIG_SPARK_HISTORICAL_FEATURE_OUTPUT_LOCATION,
-    CONFIG_SPARK_STAGING_LOCATION,
-    FEAST_DEFAULT_OPTIONS,
-)
+from feast.constants import ConfigOptions
 from feast.core.CoreService_pb2 import (
     ApplyEntityRequest,
     ApplyEntityResponse,
@@ -81,7 +64,6 @@ from feast.feature_table import FeatureTable
 from feast.grpc import auth as feast_auth
 from feast.grpc.grpc import create_grpc_channel
 from feast.loaders.ingest import (
-    BATCH_INGESTION_PRODUCTION_TIMEOUT,
     _check_field_mappings,
     _read_table_from_source,
     _upload_to_bq_source,
@@ -158,7 +140,7 @@ class Client:
         self._auth_metadata: Optional[grpc.AuthMetadataPlugin] = None
 
         # Configure Auth Metadata Plugin if auth is enabled
-        if self._config.getboolean(CONFIG_ENABLE_AUTH_KEY):
+        if self._config.getboolean(ConfigOptions.ENABLE_AUTH):
             self._auth_metadata = feast_auth.get_auth_metadata_plugin(self._config)
 
     @property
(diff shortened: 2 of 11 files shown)


sdk/python/feast/constants.py
```diff
@@ -21,135 +21,188 @@ class AuthProvider(Enum):
...
+    #: Oauth client secret
+    OAUTH_CLIENT_SECRET: str = ""
+
+    #: Oauth intended recipients
+    OAUTH_AUDIENCE: str = ""
+
+    #: Oauth token request url
+    OAUTH_TOKEN_REQUEST_URL: str = ""
+
+    MAX_WAIT_INTERVAL: str = "60"
```
terryyylim, line 205: Excluded docs for this on purpose since it's only used internally.

sdk/python/feast/cli.py
```diff
@@ -23,7 +23,7 @@
 
 from feast.client import Client
 from feast.config import Config
-from feast.constants import CONFIG_SPARK_LAUNCHER
+from feast.constants import ConfigOptions
```
woop, line 26: maybe just use an alias here like `opt`

sdk/python/feast/client.py
```diff
@@ -189,21 +173,21 @@ def _serving_service(self):
...
-                timeout=self._config.getint(CONFIG_GRPC_CONNECTION_TIMEOUT_DEFAULT_KEY),
+                timeout=self._config.getint(ConfigOptions.GRPC_CONNECTION_TIMEOUT),
             )
             self._serving_service_stub = ServingServiceStub(channel)
         return self._serving_service_stub
 
     @property
     def _use_job_service(self) -> bool:
-        return self._config.exists(CONFIG_JOB_SERVICE_URL_KEY)
+        return ConfigOptions.JOB_SERVICE_URL is not None
```
woop, line 190: wait, shouldnt this be using a getter?

sdk/python/feast/constants.py
```diff
@@ -14,142 +14,196 @@
...
+        self._default = default
+
+    def __get__(self, instance, owner):
+        if instance is None:
+            return self._name.lower()
+
+        return self._default
+
+
+class ConfigMeta(type):
```
woop, line 37: Can you add a short description to explain what you are doing here?

sdk/python/feast/client.py
pyalex: should be `self._config.get(ConfigOptions.JOB_SERVICE_URL) is not None`

sdk/python/feast/constants.py
woop: What do you mean by `internally`?

sdk/python/feast/client.py
terryyylim: My bad, typo.

woop:
I love this PR.

sdk/python/feast/constants.py
```diff
@@ -14,142 +14,196 @@
...
+    """ Feast Configuration Options """
+
+    #: Default datetime column name for point-in-time join
+    DATETIME_COLUMN: str = "datetime"
+
+    #: Environmental variable to specify Feast configuration file location
+    FEAST_CONFIG_FILE_ENV: str = "FEAST_CONFIG"
+
+    #: Default prefix to Feast environmental variables
+    CONFIG_FEAST_ENV_VAR_PREFIX: str = "FEAST_"
```
woop, line 63: Some of these options should not be defaults. They are more like traditional constants. For example, you cant set this `CONFIG_FEAST_ENV_VAR_PREFIX` because you need it to find the `CONFIG_FEAST_ENV_VAR_PREFIX` config in the first place.

sdk/python/feast/constants.py
woop: It's just strange how we allow users to set it but we dont document it. I think it should be documented.

sdk/python/feast/constants.py
```diff
@@ -21,135 +21,194 @@ class AuthProvider(Enum):
...
+    #: Authentication Provider - Google OpenID/OAuth
+    AUTH_PROVIDER: str = "google"
+
+    #: Spark Job launcher
+    SPARK_LAUNCHER: str = "dataproc"  # standalone, dataproc, emr
+
+    #: Feast Spark Job ingestion jobs staging location
+    SPARK_STAGING_LOCATION: str = ""
+
+    #: Feast Spark Job ingestion jar file
```
woop, line 135: Is this http only?

sdk/python/feast/constants.py
```diff
@@ -21,135 +21,194 @@ class AuthProvider(Enum):
...
+    #: Time to wait for historical feature requests before timing out.
+    BATCH_FEATURE_REQUEST_WAIT_TIME_SECONDS: str = "600"
+
+    #: Authentication Provider - Google OpenID/OAuth
+    AUTH_PROVIDER: str = "google"
+
+    #: Spark Job launcher
+    SPARK_LAUNCHER: str = "dataproc"  # standalone, dataproc, emr
+
+    #: Feast Spark Job ingestion jobs staging location
```
woop, line 132: What are the possible options?

sdk/python/feast/constants.py
```diff
@@ -21,135 +21,194 @@ class AuthProvider(Enum):
...
+    #: Default gRPC connection timeout when sending an ApplyFeatureSet command to Feast Core (in seconds)
+    GRPC_CONNECTION_TIMEOUT_APPLY: str = "600"
+
+    #: Default timeout when running batch ingestion
+    BATCH_INGESTION_PRODUCTION_TIMEOUT: str = "120"
+
+    #: Time to wait for historical feature requests before timing out.
+    BATCH_FEATURE_REQUEST_WAIT_TIME_SECONDS: str = "600"
+
+    #: Authentication Provider - Google OpenID/OAuth
```
woop, line 126: What are the possible options?

sdk/python/feast/constants.py
```diff
@@ -21,135 +21,194 @@ class AuthProvider(Enum):
...
+    #: Default Feast Core URL
+    CORE_URL: str = "localhost:6565"
+
+    #: Enable or disable TLS/SSL to Feast Core
+    CORE_ENABLE_SSL: str = "False"
+
+    #: Enable user authentication to Feast Core
+    ENABLE_AUTH: str = "False"
+
+    #: Auth token for user authentication to Feast
```
woop, line 87: What kind of auth? What does this mean?

sdk/python/feast/constants.py
```diff
@@ -21,135 +21,194 @@ class AuthProvider(Enum):
...
+    #: Enable or disable StatsD
+    STATSD_ENABLED: str = "False"
+
+    #: Default StatsD port
+    STATSD_HOST: str = ""
+
+    #: Default StatsD port
+    STATSD_PORT: str = ""
+
+    #: IngestionJob DeadLetter Destination
```
woop, line 177: What are the possible options?

sdk/python/feast/constants.py
```diff
@@ -21,135 +21,194 @@ class AuthProvider(Enum):
...
+    BATCH_INGESTION_PRODUCTION_TIMEOUT: str = "120"
+
+    #: Time to wait for historical feature requests before timing out.
+    BATCH_FEATURE_REQUEST_WAIT_TIME_SECONDS: str = "600"
+
+    #: Authentication Provider - Google OpenID/OAuth
+    AUTH_PROVIDER: str = "google"
+
+    #: Spark Job launcher
+    SPARK_LAUNCHER: str = "dataproc"  # standalone, dataproc, emr
```
woop, line 130: Should we change this to `standalone` so that users arent forced to use dataproc by default?

sdk/python/feast/constants.py
terryyylim: Yup I think so.

sdk/python/feast/constants.py
terryyylim: Updated.

sdk/python/feast/constants.py
terryyylim: Updated.

sdk/python/feast/constants.py
terryyylim: Updated.

sdk/python/feast/constants.py
terryyylim: Updated.

sdk/python/feast/constants.py
terryyylim: Updated.

sdk/python/feast/constants.py
terryyylim: Updated.

sdk/python/feast/cli.py
terryyylim: Updated.

sdk/python/feast/constants.py
woop: how do users know that?

sdk/python/feast/constants.py
```diff
@@ -14,142 +14,212 @@
...
+        if instance is None:
+            return self._name.lower()
+
+        return self._default
+
+
+class ConfigMeta(type):
+    """
+    Class factory which customizes ConfigOptions class instantiation.
+    Specifically, setting its name to lowercase of capitalized variable.
```
woop, line 40: `setting its name` what is "it"?

sdk/python/feast/constants.py
terryyylim: Defaults to None. Users should set this on their own.

sdk/python/feast/constants.py
terryyylim: Shifted out of configurable options.

sdk/python/feast/constants.py
pyalex: It currently depends on spark launcher: in dataproc `http` and `gs` is supported; in emr `http` and `s3`, in local mode `http` and `file`

sdk/python/feast/constants.py
pyalex: Again `gs` supported by dataproc launcher, `s3` by emr.
We do not include s3 support in dataproc launcher and vice versa

pyalex (approved):

PR #1176, opened 2020-11-19
Update Feast Core list features method

terryyylim:
Currently, `listFeatures` method is not updated to return features from feature tables, but rather feature sets.
This PR includes:
- Updated `ListFeaturesResponse` proto
- Updated `SpecService` to retrieve features from feature tables
- Updated REST (GET) endpoint for listing features

Fixes #

```release-note
Users can now retrieve features registered in feature tables.
```

### DIFF
--- a/core/src/main/java/feast/core/model/FeatureTable.java
+++ b/core/src/main/java/feast/core/model/FeatureTable.java
@@ -273,9 +273,9 @@ public class FeatureTable extends AbstractTimestampEntity {
    * @return boolean True if FeatureTable contains all entities in the entitiesFilter
    */
   public boolean hasAllEntities(List<String> entitiesFilter) {
-    List<String> allEntitiesName =
-        this.getEntities().stream().map(entity -> entity.getName()).collect(Collectors.toList());
-    return allEntitiesName.equals(entitiesFilter);
+    Set<String> allEntitiesName =
+        this.getEntities().stream().map(entity -> entity.getName()).collect(Collectors.toSet());
+    return allEntitiesName.equals(new HashSet<>(entitiesFilter));
   }
 
   /**
@@ -285,7 +285,7 @@ public class FeatureTable extends AbstractTimestampEntity {
    * @param labelsFilter contain labels that should be attached to FeatureTable's features
    * @return Map of Feature references and Features
    */
-  public Map<String, FeatureV2> getFeaturesByRef(Map<String, String> labelsFilter) {
+  public Map<String, FeatureV2> getFeaturesByLabels(Map<String, String> labelsFilter) {
     Map<String, FeatureV2> validFeaturesMap;
     List<FeatureV2> validFeatures;
     if (labelsFilter.size() > 0) {
@@ -303,7 +303,7 @@ public class FeatureTable extends AbstractTimestampEntity {
    * @param features List of features to insert to map.
    * @return Map of featureRef:feature.
    */
-  public Map<String, FeatureV2> getFeaturesRefToFeaturesMap(List<FeatureV2> features) {
+  private Map<String, FeatureV2> getFeaturesRefToFeaturesMap(List<FeatureV2> features) {
     Map<String, FeatureV2> validFeaturesMap = new HashMap<>();
     for (FeatureV2 feature : features) {
       ServingAPIProto.FeatureReferenceV2 featureRef =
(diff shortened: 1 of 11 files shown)


core/src/test/java/feast/core/service/SpecServiceIT.java
```diff
@@ -1359,7 +1386,7 @@ public void shouldReturnNoTables() {
               StatusRuntimeException.class,
               () -> apiClient.simpleGetFeatureTable(projectName, featureTableName));
 
-      assertThat(featureTables.size(), equalTo(0));
+      assertThat(featureTables.size(), equalTo(1));
```
pyalex, line 1389: The test is called "should ReturnNoTables", isn't there contradiction?

core/src/main/java/feast/core/service/SpecService.java
```diff
@@ -297,46 +297,37 @@ public ListFeatureSetsResponse listFeatureSets(ListFeatureSetsRequest.Filter fil
...
-        featureSets =
-            featureSets.stream()
-                .filter(featureSet -> featureSet.hasAllEntities(entities))
-                .collect(Collectors.toList());
-      }
+    ListFeaturesResponse.Builder response = ListFeaturesResponse.newBuilder();
+    if (entities.size() > 0) {
+      featureTables =
+          featureTables.stream()
+              .filter(featureTable -> featureTable.hasAllEntities(entities))
```
pyalex, line 318: We should check that entities are exactly match to whatever specified in filter, not a subset

core/src/main/java/feast/core/model/FeatureTable.java
```diff
@@ -263,6 +266,84 @@ public void updateFromProto(
         .collect(Collectors.toSet());
   }
 
+  /**
+   * Return a boolean to indicate if FeatureTable contains all specified entities.
+   *
+   * @param entitiesFilter contain entities that should be attached to the FeatureTable
+   * @return boolean True if FeatureTable contains all entities in the entitiesFilter
+   */
+  public boolean hasAllEntities(List<String> entitiesFilter) {
```
pyalex, line 275: Comparison of lists here is error prone, since list has order. I think we should rather compare sets

core/src/main/java/feast/core/model/FeatureTable.java
```diff
@@ -263,6 +266,84 @@ public void updateFromProto(
...
+    return validFeaturesMap;
+  }
+
+  /**
+   * Returns map for accessing features using their respective feature reference.
+   *
+   * @param features List of features to insert to map.
+   * @return Map of featureRef:feature.
+   */
+  public Map<String, FeatureV2> getFeaturesRefToFeaturesMap(List<FeatureV2> features) {
```
pyalex, line 306: why this method is public? Seems that it's being only used by `getFeaturesByRef`

core/src/main/java/feast/core/model/FeatureTable.java
```diff
@@ -263,6 +266,84 @@ public void updateFromProto(
...
+    return validFeatures;
+  }
+
+  /**
+   * Render a feature reference as string.
+   *
+   * @param featureReference to render as string
+   * @return string representation of feature reference.
+   */
+  public static String renderFeatureRef(ServingAPIProto.FeatureReferenceV2 featureReference) {
```
pyalex, line 341: Do we need this method? It just wraps single function call

core/src/main/java/feast/core/model/FeatureTable.java
```diff
@@ -263,6 +266,84 @@ public void updateFromProto(
...
+  }
+
+  /**
+   * Returns a map of Feature references and Features if FeatureTable's Feature contains all labels
+   * in the labelsFilter
+   *
+   * @param labelsFilter contain labels that should be attached to FeatureTable's features
+   * @return Map of Feature references and Features
+   */
+  public Map<String, FeatureV2> getFeaturesByRef(Map<String, String> labelsFilter) {
```
pyalex, line 288: Name is a little bit confusing. It looks more like `getFeaturesByLabels`

core/src/test/java/feast/core/service/SpecServiceIT.java
```diff
@@ -311,11 +328,15 @@ public void shouldUseDefaultProjectIfProjectUnspecified() {
               .build();
       List<FeatureTableProto.FeatureTable> featureTables =
           apiClient.simpleListFeatureTables(filter);
+      System.out.println(featureTables);
```
pyalex, line 331: debug?

pyalex (approved):

PR #1187, opened 2020-11-22
With defaults

mike0sv:
Stops default config values from being written to config file (`~/.feast/config`) and adds a possibility to provide default value when extracting configuration option. The disscussion about this change is in #1172 

Fixes #

```release-note

```


mike0sv:
sorry wrong branch

PR #1188, opened 2020-11-22
Dont write defaults to config

mike0sv:
Stops default config values from being written to config file (~/.feast/config) and adds a possibility to provide default value when extracting configuration option. The discussion about this change is in #1172

Fixes #

```release-note

```


sdk/python/feast/config.py
```diff
@@ -117,69 +115,66 @@ def __init__(
...
+    def _get(self, option, default, get_method):
+        fallback = {} if default is _UNSET else {"fallback": default}
+        return get_method(
+            CONFIG_FILE_SECTION,
+            option,
+            vars={**_get_feast_env_vars(), **self._options},
+            **fallback,
+        )
+
+    def get(self, option, default=_UNSET):
```
woop, line 127: @mike0sv How are you planning for people to use this `default` argument? Do you want them to define defaults throughout the code base, for example `.get(key, default="foo")`?

sdk/python/feast/config.py
```diff
@@ -29,6 +29,7 @@
 from feast.constants import ConfigOptions as opt
 
 _logger = logging.getLogger(__name__)
+_UNSET = object()
```
woop, line 32: How is the `object()` type used here? Is there any reason why we can't use `default=None`?

sdk/python/feast/config.py
mike0sv: Yes, his option can be used if there is no value even in `.constants`. Also this allows to get arbitrary options from config (it's a side effect, so I don't know if cases where this is needed exists). 
For now, if you try to get an option, which is declared as `Optional` in `.constants` (more precisely, that has `None` as default value), you'll get an error (because of the filtering in the `defaults` method)

sdk/python/feast/config.py
mike0sv: Is a signleton, just like `configparser._UNSET`. If you use `None`, either you can't use `None` as default value, or you'll get `None` even if you don't want default value and do expect an exception

woop (approved):

woop:
You need to add a release note

PR #1194, opened 2020-11-24
Delete v1 concepts

terryyylim:
This is a follow-up PR from #1186, to remove all related v1 concepts from common, core, serving and storage modules. 

Fixes #

```release-note
Administrators can no longer support v1 concepts anymore on latest version deployments while using older SDKs.
```


core/src/main/resources/db/migration/V3.1__Remove_V1_Concepts.sql
```diff
@@ -0,0 +1,15 @@
+DROP TABLE IF EXISTS feature_sets CASCADE;
```
pyalex, line 1: Can we postpone table removal? Maybe move it to separate PR and not merge for now

pyalex (approved):

PR #1198, opened 2020-11-26
Option for additional spark configuration

### LINKED ISSUE #1168: Additional configuration options for Spark jobs
Now there is no way to provide additional configuration options for spark jobs. For example, this is required to provide credentials for s3 storage. 

I'm using python sdk, so I dont know if this can be replicated in java and go, but my proposal is to add a configuration option, something like `FEAST_SPARK_ADDITIONAL_OPTS` which can be set to something like `"option1=value1;option2=value2"` and this options would be added in `spark_submit` method for standalone, `dataproc_sumbit` for dataproc and `_submit_emr_job` for emr.
We need to carefully pick the right format for this configuration option, considering there might be `;` or `=` in option names or values.

Maybe explicitly provide those options with every `client.start....` call, but it would break existing api, so it is way worse in my opinion.

mike0sv:
Signed-off-by: mike0sv <mike0sv@gmail.com>

Adds an option for additional spark configuration

Fixes #1168

```release-note
Added SPARK_ADDITIONAL_OPTS config option to enable users to provide additional config for Spark jobs
```


mike0sv:
Hey guys! I've been pretty preoccupied lately and didnt have the time to work on this. But I'll return to this as soon as I have the time

woop:
@mike0sv apologies about the slow feedback on this PR. We need to move it to https://github.com/feast-dev/feast-spark though, since that will be the new home for Spark related code.

PR #1209, opened 2020-12-02
Features are not being ingested due to max age overflow

pyalex:

Fixes #

With big values of max age (like 1 month) stored as Integer (2628000) converting to milliseconds that happens during writing to redis fails with overflow `2628000 * 1000 = -1666967296`. Thus all rows are being stored with negative expiry time and being deleted right away.

This PR suggests use Long instead of Int for max age. Alternative would be to do `maxAge.toLong` before converting to milliseconds, but it's very easy to miss that part.

```release-note
NONE
```


woop:
@pyalex you can just type `NONE` into the release note box if you want `release-note-none` label to be added.

pyalex:
@woop sure, I was not aware of this option

PR #1212, opened 2020-12-03
Tracing GRPC interceptor in serving & go sdk

pyalex:

Adding grpc interceptors to serving and go client instead of manual tracing. That would allow cross-service tracing.

Fixes #

```release-note
`NewTraceableGrpcClient` constructor was added to go sdk. It allows to pass Tracer on client creation phase
```


woop:
@pyalex is the plan to only implement this functionality in Golang, and not Java?

pyalex:
@woop the plan was to introduce minimal changes. do you think java sdk needs to have this functionality as well?

woop:
> @woop the plan was to introduce minimal changes. do you think java sdk needs to have this functionality as well?

It's fine to have it only for Go, but perhaps we shouldnt document this in feast.dev but only in the API docs.

khorshuheng (approved):

PR #1214, opened 2020-12-03
deprecate apply_entity and apply_feature_table for apply

### LINKED ISSUE #1183: Add support for feature tables to the apply method
We need to deprecate [client.apply_entity()](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/client.py#L443) and [client.apply_feature_table()](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/client.py#L565). Both of these methods were added because `client.apply()` was previously used for applying feature sets.

In Feast 0.9 we should
- Add support for applying feature tables and entities through `client.apply()`. This method doesn't currently exist. Please have a look at [Feast 0.7](https://github.com/feast-dev/feast/blob/v0.7.1/sdk/python/feast/client.py#L359) for a reference implementation. We can follow a similar approach.
- Add deprecation warnings on `apply_feature_table()` and `apply_entity()`
- In Feast 0.10 we will remove `apply_feature_table()` and `apply_entity()`, and `apply()` will only work on feature tables and entities.

jklegar:
deprecate client.apply_entity() and client.apply_feature_table() for client.apply(). Client.apply() was previously used for applying feature sets

Fixes #1183

```release-note
Deprecated client.apply_entity() and client.apply_feature_table(). Users should switch to client.apply() as these methods will be removed in Feast 0.10
```

### DIFF
--- a/sdk/python/feast/client.py
+++ b/sdk/python/feast/client.py
@@ -17,6 +17,7 @@ import os
 import shutil
 import uuid
 from datetime import datetime
+from deprecated import deprecated
 from itertools import groupby
 from typing import Any, Dict, List, Optional, Union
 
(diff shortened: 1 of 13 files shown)


sdk/python/feast/client.py
```diff
@@ -442,6 +443,46 @@ def archive_project(self, project):
...
+        if project is None:
+            project = self.project
+
+        if not isinstance(objects, list):
+            objects = [objects]
+        for obj in objects:
+            if isinstance(obj, Entity):
+                self._apply_entity(project, obj)  # type: ignore
+                continue
+            elif isinstance(obj, FeatureTable):
```
woop, line 480: Is there a reason to use `continue` if you are using an `elif`?

I would expect either 
```
if a:
  do
  continue

if b:
  do 
  continue

raise
```
or
```
if a:
  do
elif b:
  do
else:
 raise
```

sdk/python/feast/client.py
```diff
@@ -442,6 +443,46 @@ def archive_project(self, project):
...
+                self._apply_feature_table(project, obj)  # type: ignore
+                continue
+            raise ValueError(f"Could not determine entity type to apply {obj}")
```
woop, line 483: we should maybe add `type(obj)`

woop:
Can you please add a deprecation warning to the old methods?

https://docs.python.org/3/library/warnings.html

woop (approved):

Issue #1181, opened 2020-11-19
Add Azure support to Feast


woop:
This card tracks the addition of Azure support to Feast.

* Spark Job Runner: The current proposal is to implement support for the [Databricks Jobs API on Azure](https://docs.microsoft.com/en-us/azure/databricks/dev-tools/api/latest/jobs)
* File system: We will need to add support for either DBFS or directly accessing Azure Blob Storage. It seems like DBFS using [mount points](https://docs.databricks.com/data/data-sources/azure/azure-storage.html#mount-an-azure-blob-storage-container) is the simplest way to implement the file system.
* Terraform Installation: We will need to add support for deploying Feast using Terraform to Azure Kubernetes Service (AKS).
* End-to-end tests: We will need to set up end-to-end tests that run all critical flows through a full Azure deployment.

cc @Yanson

woop:
[Feast on Azure RFC](https://docs.google.com/document/d/1uzSeVedDtxUYil_qlvmjnhbxbta_PXcWaEYj__Nvd3w/edit#)
Noteworthy changes to the above
* We are adding support for Azure Blob Storage
* We did not pick up adding Databricks support for Spark Jobs
* @jklegar is working on e2e test support

gramhagen:
Hey this looks really awesome. I'm super new to the project but was looking for feature store solutions that could be used in Azure and would love to help out.

Is there anything specific I can help with? Or another place I should use to discuss this?

woop:
Hey @gramhagen, I somehow didn't notice your comment here! If you are still interested, do let us know over in the Kubeflow Slack (linked in our readme).

PR #1218, opened 2020-12-08
implement abstractstagingclient for azure blob storage

jklegar:

**What this PR does / why we need it**: Supports Azure Blob Storage as an object store and batch data source as part of #1181 

Fixes #

```release-note
Users can now use Azure blob storage as a staging location and as a batch data source. To configure it, they will need to set the environment variables FEAST_AZURE_BLOB_ACCOUNT_NAME and FEAST_AZURE_BLOB_ACCOUNT_ACCESS_KEY 
```

### DIFF
--- a/sdk/python/feast/constants.py
+++ b/sdk/python/feast/constants.py
@@ -128,6 +128,12 @@ class ConfigOptions(metaclass=ConfigMeta):
     #: Endpoint URL for S3 storage_client
     S3_ENDPOINT_URL: Optional[str] = None
 
+    #: Account name for Azure blob storage_client
+    AZURE_BLOB_ACCOUNT_NAME: Optional[str] = None
+
+    #: Account access key for Azure blob storage_client
+    AZURE_BLOB_ACCOUNT_ACCESS_KEY: Optional[str] = None
+
     #: Authentication Provider - Google OpenID/OAuth
     #:
     #: Options: "google" / "oauth"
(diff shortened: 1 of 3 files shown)


sdk/python/feast/loaders/file.py
```diff
@@ -76,10 +78,10 @@ def export_source_to_staging_location(
                 os.path.join(source_uri.netloc, source_uri.path)
             )
         else:
-            # gs, s3 file provided as a source.
+            # gs, s3, azure blob file provided as a source.
             assert source_uri.hostname is not None
             return get_staging_client(source_uri.scheme).list_files(
```
jklegar: seems like the Config object isn't available to be passed to get_staging_client here, so is there a better way to pass in the required parameters (account name and access key) for the Azure client than requiring them to be passed in in the config?

jklegar:
I've done some basic manual testing, not sure if there's more that should be done outside of end-to-end Azure tests - what is the case for S3/GCS?

sdk/python/feast/constants.py
```diff
@@ -128,6 +128,12 @@ class ConfigOptions(metaclass=ConfigMeta):
     #: Endpoint URL for S3 storage_client
     S3_ENDPOINT_URL: Optional[str] = None
 
+    #: Account name for Azure blob storage_client
+    AZURE_BLOB_ACCOUNT_NAME: Optional[str] = None
+
+    #: Account access key for Azure blob storage_client
+    AZURE_BLOB_ACCOUNT_ACCESS_KEY: Optional[str] = None
```
woop, line 135: It's not possible for this configuration to be read from environmental variables by the azure client?

woop:
> what is the case for S3/GCS?

they are covered by e2e tests

woop (approved):

PR #1220, opened 2020-12-08
remove stage_dataframe from the launcher interface

oavdeev:
It was only implemented for EMR anyway. Also now that all upload paths are using the staging client, we can implement this once without having launcher-specific implementations.

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/pyspark/launcher.py
+++ b/sdk/python/feast/pyspark/launcher.py
@@ -303,6 +303,14 @@ def get_job_by_id(job_id: str, client: "Client") -> SparkJob:
 
 
 def stage_dataframe(df, event_timestamp_column: str, config: Config) -> FileSource:
+    """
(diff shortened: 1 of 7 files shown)


sdk/python/feast/pyspark/launcher.py
```diff
@@ -297,6 +302,24 @@ def get_job_by_id(job_id: str, client: "Client") -> SparkJob:
     return launcher.get_job_by_id(job_id)
 
 
-def stage_dataframe(df, event_timestamp_column: str, client: "Client") -> FileSource:
-    launcher = resolve_launcher(client._config)
-    return launcher.stage_dataframe(df, event_timestamp_column)
+def stage_dataframe(df, event_timestamp_column: str, config: Config) -> FileSource:
```
woop, line 305: Can we add a basic description/comment?

woop (approved):

PR #1222, opened 2020-12-09
Support TFRecord as one of the output formats for historical feature retrieval

khorshuheng:
This PR allows user to specify `tfrecord` as the output format of historical feature retrieval, which is useful in cases where the users wish to generate statistics from the retrieved dataset using tfdv, or if the user's machine learning model is based on tensorflow.

Fixes #

```release-note
User can now specify `tfrecord` as the output format of historical feature retrieval.
```

### DIFF
--- a/sdk/python/feast/pyspark/launchers/aws/emr_utils.py
+++ b/sdk/python/feast/pyspark/launchers/aws/emr_utils.py
@@ -302,9 +302,7 @@ def _historical_retrieval_step(
             ],
             "Args": ["spark-submit", pyspark_script_path]
             + args
-            + ["--packages", ",".join(packages)]
-            if packages
-            else [],
+            + (["--packages", ",".join(packages)] if packages else []),
             "Jar": "command-runner.jar",
         },
     }
--- a/sdk/python/feast/pyspark/launchers/aws/emr_utils.py
+++ b/sdk/python/feast/pyspark/launchers/aws/emr_utils.py
@@ -300,9 +300,10 @@ def _historical_retrieval_step(
                     "Value": output_file_uri,
                 },
             ],
-            "Args": ["spark-submit", pyspark_script_path]
-            + args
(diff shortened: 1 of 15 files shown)


tests/integration/test_launchers.py
```diff
@@ -37,3 +46,14 @@ def test_dataproc_job_api(
     wait_for_job_status(retrieved_job, SparkJobStatus.IN_PROGRESS)
     retrieved_job.cancel()
     assert retrieved_job.get_status() == SparkJobStatus.FAILED
+
+
+def test_dataproc_job_tfrecord_output(
+    dataproc_launcher: DataprocClusterLauncher,  # noqa: F811
```
woop, line 52: What is with all these `noqa: F811`? Can it be fixed please?

khorshuheng: That's because we are importing the fixtures from another module, and flake8 doesn't understand pytest fixtures.

One way to fix it would be not to import the fixture and include it directly in `test_launchers`, though that would mean it's harder for us to reuse the fixture across different test files.

I will research for way on how this can be circumvented without explicitly adding `# noqa: F811`, but i am not certain if it is possible.

sdk/python/feast/pyspark/launchers/aws/emr_utils.py
```diff
@@ -297,7 +300,11 @@ def _historical_retrieval_step(
                     "Value": output_file_uri,
                 },
             ],
-            "Args": ["spark-submit", pyspark_script_path] + args,
+            "Args": ["spark-submit", pyspark_script_path]
+            + args
+            + ["--packages", ",".join(packages)]
+            if packages
+            else [],
```
oavdeev-tt, line 307:
```suggestion
            "Args": ["spark-submit", pyspark_script_path]
            + args
            + (["--packages", ",".join(packages)]
            if packages
            else []),
```

PR #1225, opened 2020-12-11
Add spark k8s operator launcher

oavdeev:
This allows Feast to run Spark jobs using spark-k8s-operator. 

To enable it:
1.  set `spark_launcher="k8s"`
2. set `spark_k8s_use_incluster_config` to `True` or `False`, depending on whether spark is running in the same k8s cluster or not. Make sure that feast serviceaccount has permissions to create `SparkApplication` resources. Provide KUBECONFIG if running Feast outside of the cluster.
3. Typically you need `spark_staging_location` and `historical_feature_output_location` to be set and use `s3a://` URL scheme if using S3.
4. Lastly, set `spark_k8s_job_template_path` to point to YAML template containing spark application configuration. Feast comes with one out of the box, but in production you'll likely want to provide a custom one.

**Additional changes**:
* Storage client now supports `s3a://` URL scheme (used by OSS Spark to access S3, we didn't need this for EMR since it understands `s3://` scheme just fine)
* For historical pyspark jobs, arguments are now base64+JSON encoded. Previously they were just JSON encoded, but with OSS spark that caused issues with proper quoting (I _think_ the issue is somewhere in the docker entrypoint script provided with Spark, not the k8s operator itself).

```release-note
Feast now supports launching Spark jobs using k8s operator.
```

### DIFF
--- a/sdk/python/feast/pyspark/launcher.py
+++ b/sdk/python/feast/pyspark/launcher.py
@@ -65,10 +65,26 @@ def _emr_launcher(config: Config) -> JobLauncher:
     )
 
 
+def _k8s_launcher(config: Config) -> JobLauncher:
+    from feast.pyspark.launchers import k8s
+
+    def _get_optional(option):
+        if config.exists(option):
+            return config.get(option)
+
+    return k8s.KubernetesJobLauncher(
+        namespace=config.get(opt.SPARK_K8S_NAMESPACE),
+        resource_template_path=_get_optional(opt.SPARK_K8S_JOB_TEMPLATE_PATH),
+        staging_location=config.get(opt.SPARK_STAGING_LOCATION),
+        incluster=config.getboolean(opt.SPARK_K8S_USE_INCLUSTER_CONFIG),
+    )
+
+
 _launchers = {
     "standalone": _standalone_launcher,
     "dataproc": _dataproc_launcher,
     "emr": _emr_launcher,
+    "k8s": _k8s_launcher,
 }
 
 
--- a/sdk/python/feast/pyspark/launcher.py
+++ b/sdk/python/feast/pyspark/launcher.py
@@ -68,13 +68,9 @@ def _emr_launcher(config: Config) -> JobLauncher:
 def _k8s_launcher(config: Config) -> JobLauncher:
     from feast.pyspark.launchers import k8s
 
-    def _get_optional(option):
-        if config.exists(option):
-            return config.get(option)
-
     return k8s.KubernetesJobLauncher(
         namespace=config.get(opt.SPARK_K8S_NAMESPACE),
-        resource_template_path=_get_optional(opt.SPARK_K8S_JOB_TEMPLATE_PATH),
+        resource_template_path=config.get(opt.SPARK_K8S_JOB_TEMPLATE_PATH, None),
         staging_location=config.get(opt.SPARK_STAGING_LOCATION),
         incluster=config.getboolean(opt.SPARK_K8S_USE_INCLUSTER_CONFIG),
     )
--- a/sdk/python/feast/pyspark/launchers/gcloud/dataproc.py
+++ b/sdk/python/feast/pyspark/launchers/gcloud/dataproc.py
@@ -1,4 +1,3 @@
-import json
 import os
 import time
 import uuid
@@ -265,7 +264,7 @@ class DataprocClusterLauncher(JobLauncher):
         return blob_uri_str
 
     def dataproc_submit(
-        self, job_params: SparkJobParameters
+        self, job_params: SparkJobParameters, extra_properties: Dict[str, str]
     ) -> Tuple[Job, Callable[[], Job], Callable[[], None]]:
         local_job_id = str(uuid.uuid4())
         main_file_uri = self._stage_file(job_params.get_main_file_path(), local_job_id)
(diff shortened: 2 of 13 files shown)


sdk/python/feast/pyspark/launchers/k8s/k8s.py
```diff
@@ -0,0 +1,344 @@
...
+        jar_s3_path = self._upload_jar(ingestion_job_params.get_main_file_path())
+
+        job_id = _generate_job_id()
+
+        resource = _prepare_job_resource(
+            job_template=self._resource_template,
+            job_id=job_id,
+            job_type=OFFLINE_TO_ONLINE_JOB_TYPE,
+            main_application_file=jar_s3_path,
+            main_class="feast.ingestion.IngestionJob",
```
pyalex, line 273: shouldn't main_class be retrieved from `job_params`?

sdk/python/feast/pyspark/launchers/k8s/k8s.py
```diff
@@ -0,0 +1,344 @@
...
+            packages=[
+                "com.google.cloud.spark:spark-bigquery-with-dependencies_2.12:0.17.2"
```
pyalex, line 275: It's `0.18.0` with spark 3 support

pyalex: Shouldn't we move that to some constant shared by whole project?

sdk/python/feast/pyspark/launcher.py
```diff
@@ -65,10 +65,26 @@ def _get_optional(option):
...
+def _k8s_launcher(config: Config) -> JobLauncher:
+    from feast.pyspark.launchers import k8s
+
+    def _get_optional(option):
+        if config.exists(option):
+            return config.get(option)
+
+    return k8s.KubernetesJobLauncher(
+        namespace=config.get(opt.SPARK_K8S_NAMESPACE),
+        resource_template_path=_get_optional(opt.SPARK_K8S_JOB_TEMPLATE_PATH),
```
woop, line 77: Would it be better to use something like `config.get(option, default=None)`, instead of using _get_optional?

sdk/python/feast/pyspark/launchers/gcloud/dataproc.py
```diff
@@ -265,7 +264,7 @@ def _stage_file(self, file_path: str, job_id: str) -> str:
         return blob_uri_str
 
     def dataproc_submit(
-        self, job_params: SparkJobParameters
+        self, job_params: SparkJobParameters, extra_properties: Dict[str, str]
```
woop, line 267: I'm not sure if you saw #1198. Should we close that PR after yours has been incorporated, or first merge that one?

sdk/python/feast/pyspark/launchers/k8s/k8s.py
```diff
@@ -0,0 +1,341 @@
...
+                job_id=job_info.job_id,
+                output_file_uri=job_info.extra_metadata[METADATA_OUTPUT_URI],
+            )
+        elif job_info.job_type == OFFLINE_TO_ONLINE_JOB_TYPE:
+            return KubernetesBatchIngestionJob(
+                api=self._api, namespace=job_info.namespace, job_id=job_info.job_id,
+            )
+        elif job_info.job_type == STREAM_TO_ONLINE_JOB_TYPE:
+            # job_hash must not be None for stream ingestion jobs
+            assert METADATA_JOBHASH in job_info.extra_metadata is not None
```
woop, line 173: By the way, are we fine with these asserts disappearing when Python code optimization is enabled? I like the conciseness of the asserts, but I am not sure if they are safe to use.

woop:
@oavdeev looks good overall :+1: 

The only things that stand out are docs (tracked elsewhere in #1231) and tests. I'd feel a bit more comfortable having the tests as part of this PR, or did you have another plan?

sdk/python/feast/pyspark/launchers/gcloud/dataproc.py
oavdeev: No i haven't. It looks like it is solving a slightly different problem, namely _user-specified_ extra options. Don't feel strongly which one to merge first (and which one to rebase).

sdk/python/feast/pyspark/launchers/k8s/k8s.py
oavdeev: In this case it is mostly to make mypy happy, as the code would crash on the next line anyway even without this check.

sdk/python/feast/pyspark/launcher.py
oavdeev: fixed

sdk/python/feast/pyspark/launchers/k8s/k8s.py
oavdeev: fixed

sdk/python/feast/pyspark/launchers/k8s/k8s.py
oavdeev: fixed

oavdeev:
>I'd feel a bit more comfortable having the tests as part of this PR, or did you have another plan (minikube/kind based)?

I'd rather add tests in a followup PR. I plan to add another integration test, alongside with `test-end-to-end-[aws,gcp]`, and setting this stuff up tends to make a PR nearly unreadable due to lots of tweaks/rebases and retests.

woop (approved):

PR #1234, opened 2020-12-16
Python UDF in Ingestion being used for feature validation

### LINKED ISSUE #1230: Apply custom Python code during ingestion for data validation (Pandas UDF)

Validating data on ingestion is both non-trivial and important. It's important since Machine Learning models that consuming those features are usually highly sensible to any shifts in the data. It's non-trivial since every data model is unique and creating general solution would be not very effective. Our current validation includes only type checks and not-null restrictions applied to entities columns. Even if we will extend those check - that still would be row level validation. What we need is to look on data from distance - in batches - to detect that some column has too many null values (eg, more than 5% on average) and other column has too many outliers (distribution shift). Those checks involves more complicated data transformation, that shouldn't be a part of Feast and instead should be implemented by domain experts.

Thus we need to give users ability to use custom code inside ingestion that will be responsible for such transformations & checks. It should receive data in controlled batches and be able to call specific libraries like `great-expectations`, `tfdv`, `scipy`.

The solution consists of two parts:

1. In Python SDK we create UDFs and pickle the code to pass it to IngestionJob.
1a. In POC we don't allow to pass arbitrary code but rather have some templates that wraps specific solutions (great-expectations, tfdv, etc)
2. Inside Ingestion (only streaming in POC) we gather all columns in batches and pass it as single Pandas Dataframe to the UDF. UDF is being applied right after reading data from source and before saving to storage (before ingestion).
2a. At start we expect UDF to return single column that would mark valid/invalid rows with flag.

SDK side (great-expectations example):

```
you_training_df = client.get_historical_features(...)

from great_expectations.dataset import PandasDataset
ds = PandasDataset.from_dataset(you_training_df)
ds.expect_column_values_to_be_between('column', 0, 100)

expectations = ds.get_expectation_suite()

from feast.contrib.validation.ge import create_validation_udf, apply_validation

validation_udf = create_validation_udf('myUDF', expectations)
apply_validation(client, feature_table, validation_udf, validation_window_secs=30)
```

Inside ingestion Job
```
validationUDF: StructType -> BooleanType

val rowsAfterValidation = batchDF.withColumn("_isValid", validationUDF(struct(columns:_*)))

rowsAfterValidation
     .filter("!_isValid")
     .map(updateMetrics)      
     .writeToDeadLetter

...
     .trigger(ProcessingTimeTrigger.create(config.ValidationWindow, TimeUnit.SECONDS))
     .start
```

**Additional context**
Previous iteration [RFC](https://docs.google.com/document/d/1C24LVnw0qz9N_22ysPingE6qFkv4zOxUCe8FZ2GZ-yI/edit)

pyalex:

This PR introduces **experimental** feature that enables using custom python code inside ingestion job. See #1230 for motivation.

Technical details:
1. Python code being wrapped and pickled inside Python SDK.
2. Pickled object being staged to one of supported storages (gs / s3)
3. Reference to stored objects added to FeatureTable's labels
4. IngestionJob currently only supports `_validationUDF` label and use provided code for feature validation. The code is being called right after reading data from source and before writing it to store. IngestionJob can optionally (set by flag) drop rows that do not pass validation.

Limitations:
Since python pickle happens on customer's machine and unpickle on Spark worker, there might be issues related to incompatibility of pickle protocols even customer's and worker's python versions are different. Ideally to avoid that the minor part of python versions should match (3.**7**, eg). However, it was confirmed by tests that lower version on SDK is fine (3.6 on SDK, 3.7 on worker), but not other way.

Fixes #1230 

```release-note
**experimental** ability to run custom python code inside streaming ingestion job
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ test-java-integration:
 	${MVN} --no-transfer-progress -Dmaven.javadoc.skip=true -Dgpg.skip -DskipUTs=true clean verify
 
 test-java-with-coverage:
-	${MVN} --no-transfer-progress test jacoco:report-aggregate
+	${MVN} --no-transfer-progress -DskipITs=true test jacoco:report-aggregate
 
 build-java:
 	${MVN} clean verify
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ lint-java:
 	${MVN} --no-transfer-progress spotless:check
 
 test-java:
-	${MVN} --no-transfer-progress test
+	${MVN} --no-transfer-progress -DskipITs=true test
 
 test-java-integration:
 	${MVN} --no-transfer-progress -Dmaven.javadoc.skip=true -Dgpg.skip -DskipUTs=true clean verify
@@ -57,7 +57,7 @@ build-java:
 	${MVN} clean verify
 
 build-java-no-tests:
-	${MVN} --no-transfer-progress -Dmaven.javadoc.skip=true -Dgpg.skip -DskipUTs=true -Drevision=${REVISION} clean package
+	${MVN} --no-transfer-progress -Dmaven.javadoc.skip=true -Dgpg.skip -DskipUTs=true -DskipITs=true -Drevision=${REVISION} clean package
 
 # Python SDK
 
--- a/infra/scripts/build-ingestion-py-dependencies.sh
+++ b/infra/scripts/build-ingestion-py-dependencies.sh
@@ -11,7 +11,7 @@ pip3 install -t ${tmp_dir}/libs $PACKAGES
 cd $tmp_dir
 tar -czf pylibs-ge-$PLATFORM.tar.gz libs/
 if [[ $DESTINATION == gs* ]]; then
-  gsutil cp pylibs-ge-$PLATFORM.tar.gz $GS_DESTINATION/
+  gsutil cp pylibs-ge-$PLATFORM.tar.gz $GS_DESTINATION
 else
   mv pylibs-ge-$PLATFORM.tar.gz $DESTINATION
 fi
--- a/infra/scripts/build-ingestion-py-dependencies.sh
+++ b/infra/scripts/build-ingestion-py-dependencies.sh
@@ -2,7 +2,7 @@
 
 PLATFORM=$1
 DESTINATION=$2
-PACKAGES=${PACKAGES:-"great-expectations==0.13.2"}
+PACKAGES=${PACKAGES:-"great-expectations==0.13.2 pyarrow==2.0.0"}
 
 tmp_dir=$(mktemp -d)
 
--- a/infra/scripts/build-ingestion-py-dependencies.sh
+++ b/infra/scripts/build-ingestion-py-dependencies.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-VERSION=$1
+PLATFORM=$1
 DESTINATION=$2
 PACKAGES=${PACKAGES:-"great-expectations==0.13.2"}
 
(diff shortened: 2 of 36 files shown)


sdk/python/feast/pyspark/launchers/aws/emr_utils.py
```diff
@@ -332,6 +332,7 @@ def _stream_ingestion_step(
             ],
             "Args": ["spark-submit", "--class", "feast.ingestion.IngestionJob"]
             + jars_args
+            + ["--conf", "spark.yarn.isPython=true"]
```
oavdeev, line 335: Is there any downside of doing this? Do you know why it isn't it always set it to true in Spark?

spark/ingestion/src/main/scala/feast/ingestion/sources/bq/BigQueryReader.scala
```diff
@@ -46,6 +46,6 @@ object BigQueryReader {
     reader
       .load(s"${source.project}.${source.dataset}.${source.table}")
       .filter(col(source.eventTimestampColumn) >= new Timestamp(start.getMillis))
-      .filter(col(source.eventTimestampColumn) < new Timestamp(end.getMillis))
+      .filter(col(source.eventTimestampColumn) <= new Timestamp(end.getMillis))
```
oavdeev, line 49: Why this change? When I see `start` and `end` i'd generally expect the convention to be [inclusive,exclusive).

Makefile
```diff
@@ -45,19 +45,19 @@ lint-java:
 	${MVN} --no-transfer-progress spotless:check
 
 test-java:
-	${MVN} --no-transfer-progress test
+	${MVN} --no-transfer-progress -DskipITs=true test
```
oavdeev, line 48: why? (i'm not super familiar with our java test machinery, does it now skip some tests it didn't skip before?)

infra/scripts/build-ingestion-py-dependencies.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
```
oavdeev, line 2:
```suggestion
set -euo pipefail
```

sdk/python/feast/contrib/validation/ge.py
```diff
@@ -0,0 +1,126 @@
...
+        "pyspark must be installed to enable validation functionality. "
+        "Please install feast[validation]"
+    )
+
+
+if TYPE_CHECKING:
+    from feast import Client, FeatureTable
+
+
+GE_PACKED_ARCHIVE = "https://storage.googleapis.com/feast-jobs/spark/validation/pylibs-ge-%(platform)s.tar.gz"
```
oavdeev, line 34: How does this work from distribution perspective? I worry that if this is not in any way tied to Feast version, we can't upgrade GE without breaking older versions of ingestion job.

Maybe we don't have to address this now, especially while this is contrib, but sometime down the road we probably need to pin the version of this tarball to a Feast version somehow.

sdk/python/feast/pyspark/launchers/aws/emr_utils.py
pyalex: It's yarn specific and from what I found it's used mostly to enable distribution of python related stuff (like pyspark.zip) to yarn workers. It's being set by `spark-submit` when main file is py-file, which is not the case for our IngestionJob.

Makefile
pyalex: we have two toggles `skipITs` and `skipUTs`- means `skip Integration Tests` and `skip Unit Tests`, we need them to run thoses test suites separately. Before there was no need in `skipITs`, since in maven pipeline `verify` triggered after `test`, but in spark part we have some additional steps (`generate-test-source` phase) that are required only by integration tests, and don't needed by unit tests.

sdk/python/feast/contrib/validation/ge.py
pyalex: I know that smells. But since feature is experimental I think it's ok for now.
As an option for future - we could put this archive inside ingestion jar.

sdk/python/feast/pyspark/launchers/aws/emr_utils.py
oavdeev: Makes sense. Maybe add a comment there for the future generations?

oavdeev (approved):

khorshuheng (approved):
/lgtm

PR #1241, opened 2020-12-22
azure e2e test

### LINKED ISSUE #1181: Add Azure support to Feast

jklegar:

**What this PR does / why we need it**: Runs e2e tests on Azure using the new spark operator and Azure blob storage support. Note it therefore requires #1218 

Fixes #1181

```release-note
NONE
```

### DIFF
--- a/infra/docker/ci/Dockerfile
+++ b/infra/docker/ci/Dockerfile
@@ -83,3 +83,6 @@ RUN apt-get install -y kubectl
 RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 && \
     chmod 700 get_helm.sh && \
     ./get_helm.sh
+
+# Install jq
+RUN apt-get install -y jq
--- a/infra/docker/ci/Dockerfile
+++ b/infra/docker/ci/Dockerfile
@@ -77,12 +77,12 @@ RUN PROTOC_ZIP=protoc-${PROTOC_VERSION}-linux-x86_64.zip && \
 RUN curl -sL https://aka.ms/InstallAzureCLIDeb | bash
 
 # Install kubectl
-RUN apt-get install -y kubectl
+RUN apt-get install -y kubectl=1.20.0-00
 
 # Install helm
 RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 && \
     chmod 700 get_helm.sh && \
-    ./get_helm.sh
+    ./get_helm.sh --version v3.4.2
 
 # Install jq
 RUN apt-get install -y jq
--- a/infra/docker/ci/Dockerfile
+++ b/infra/docker/ci/Dockerfile
@@ -72,3 +72,14 @@ RUN PROTOC_ZIP=protoc-${PROTOC_VERSION}-linux-x86_64.zip && \
 	  git clone https://github.com/istio/tools/ && \
 	  cd tools/cmd/protoc-gen-docs && \
 	  go build && mkdir -p $HOME/bin && cp protoc-gen-docs $HOME/bin
+
+# Install AZ CLI
+RUN curl -sL https://aka.ms/InstallAzureCLIDeb | bash
+
+# Install kubectl
+RUN apt-get install -y kubectl
+
+# Install helm
+RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 && \
+    chmod 700 get_helm.sh && \
+    ./get_helm.sh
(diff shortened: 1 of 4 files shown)


infra/docker/ci/Dockerfile
```diff
@@ -71,4 +71,15 @@ RUN PROTOC_ZIP=protoc-${PROTOC_VERSION}-linux-x86_64.zip && \
...
\ No newline at end of file
+	  go build && mkdir -p $HOME/bin && cp protoc-gen-docs $HOME/bin
+
+# Install AZ CLI
+RUN curl -sL https://aka.ms/InstallAzureCLIDeb | bash
+
+# Install kubectl
+RUN apt-get install -y kubectl
+
+# Install helm
```
jklegar, line 82: not sure if the feast-ci image already has kubectl and/or helm when it is running tests, so it's possible these installs are not needed. Also lmk if there's a better place to install the azure CLI

woop: I think its fine, but should the version be pinned?

infra/scripts/azure-runner.sh
```diff
@@ -0,0 +1,158 @@
...
+        --set "feast-online-serving.image.repository=${DOCKER_REPOSITORY}/feast-serving" \
+        --set "feast-online-serving.image.tag=${GIT_TAG}" \
+        --set "feast-jobservice.image.repository=${DOCKER_REPOSITORY}/feast-jobservice" \
+        --set "feast-jobservice.image.tag=${GIT_TAG}" \
+        --set "feast-core.image.repository=${DOCKER_REPOSITORY}/feast-core" \
+        --set "feast-core.image.tag=${GIT_TAG}" \
+        --set "prometheus-statsd-exporter.enabled=false" \
+        --set "prometheus.enabled=false" \
+        --set "grafana.enabled=false" \
+        --set "feast-jobservice.enabled=false" \
```
woop, line 63: Should we be testing the job service here as well?

infra/scripts/azure-runner.sh
```diff
@@ -0,0 +1,158 @@
...
+# Log into k8s.
+echo "${STEP_BREADCRUMB} Updating kubeconfig"
+az login --service-principal -u "$AZ_SERVICE_PRINCIPAL_ID" -p "$AZ_SERVICE_PRINCIPAL_PASS" --tenant "$AZ_SERVICE_PRINCIPAL_TENANT_ID"
+az aks get-credentials --resource-group "$RESOURCE_GROUP" --name "$AKS_CLUSTER_NAME"
+
+# Sanity check that kubectl is working.
+echo "${STEP_BREADCRUMB} k8s sanity check"
+kubectl get pods
+
+# e2e test - runs in sparkop namespace for consistency with AWS sparkop test.
```
woop, line 130: Is there a reason to keep these consistent? No conflict will occur?

infra/scripts/azure-runner.sh
```diff
@@ -0,0 +1,158 @@
...
+  name: use-spark-operator
+roleRef:
+  kind: Role
+  name: use-spark-operator
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+  - kind: ServiceAccount
+    name: default
+EOF
+}
```
oavdeev, line 107: it may make sense to put these functions in a separate file since they're used by the AWS test scripts as well

infra/scripts/azure-runner.sh
jklegar: it's just so I can reuse the existing sparkop test code as is, and it needs to run in some namespace so why not "sparkop"

infra/docker/ci/Dockerfile
jklegar: the version for which of these?

oavdeev: I'd pin both kubectl and helm. kubectl is [only compatible within one minor version](https://kubernetes.io/docs/setup/release/version-skew-policy/#kubectl) so it is conceivable that one day this will suddenly stop working with our k8s clusters.

woop (approved):

PR #1254, opened 2021-01-06
docs, fixes and scripts to run e2e tests in minikube

oavdeev:
This contains a doc describing how to run e2e tests in minikube, and some fixes and scripts to make that work.

The fixes are to make sure custom S3 endpoints are used throughout by Feast and tests, so it can run against minio.

```release-note
NONE
```


tests/README.md
```diff
@@ -0,0 +1,90 @@
...
+5. Copy secret from minio into new sparkop namespace.
+```bash
+kubectl get secret minio --namespace=minio -oyaml | grep -v '^\s*namespace:\s' | kubectl apply --namespace=sparkop -f -
+```
+
+6. Install Feast. That may fail due to timeout, rerun again in that case
+```bash
+#
+# NB!! make sure to use bash 5.0+ for the next step. You'd need to brew install it on OSX
+DOCKER_REPOSITORY=gcr.io/kf-feast GIT_TAG=develop bash ./infra/scripts/setuplocal.sh
```
woop, line 55: Should this be setup-e2e-local.sh?

oavdeev: yep good catch, that was a last minute rename

infra/scripts/k8s-common-functions.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
```
jklegar, line 1: can we combine this with infra/scripts/runner-helper.sh

oavdeev-tt: good point, done

woop (approved):

infra/scripts/k8s-common-functions.sh
```diff
@@ -30,36 +43,34 @@ function helm_install {
...
 
     # We skip statsd exporter and other metrics stuff since we're not using it anyway, and it
     # has some issues with unbound PVCs (that cause kubectl delete pvc to hang).
-    echo "${STEP_BREADCRUMB} Helm installing feast"
+    echo "${STEP_BREADCRUMB:-} Helm installing feast"
 
-    if ! time helm install --wait "$RELEASE" infra/charts/feast \
-        --timeout 15m \
+    if ! time helm install --wait "$RELEASE" ./infra/charts/feast \
+        --timeout 5m \
```
jklegar, line 60: the azure test requires >5min (15 to be safe), I believe because of some services requiring others to be up + crashloopbackoff. Probably could be fixed but I had just increased the timeout

oavdeev: Ok rolled this back, but this is peculiar since on EKS it usually takes maybe a minute, and you'd think this is pure k8s stuff that shouldn't depend on the cloud provider much.

sdk/python/feast/pyspark/launchers/k8s/k8s.py
```diff
@@ -180,10 +184,6 @@ def _job_from_job_info(self, job_info: JobInfo) -> SparkJob:
             # We should never get here
             raise ValueError(f"Unknown job type {job_info.job_type}")
 
-    def _get_staging_client(self):
-        uri = urlparse(self._staging_location)
-        return get_staging_client(uri.scheme, self._config)
-
     def _get_azure_credentials(self):
```
jklegar, line 187: I had changed it recently to just pass in the whole config but if we don't want to do that, then this function needs the Azure blob account name and access key passed in

PR #1256, opened 2021-01-07
Reporting metrics from validation UDF

pyalex:

In this PR previously added Python UDF for feature validation reports amount of unexpected rows (that didn't pass validation) as statsd counter. That will allow to monitor percentage of invalid data and create alert when it reach some threshold.

Fixes #

```release-note
NONE
```


khorshuheng (approved):
/lgtm

PR #1259, opened 2021-01-08
Enhance job api to return associated feature table and start time

khorshuheng:

As part of an effort to create an UI for Feast, the JobService API needs to return more information, in particular, job start time and the feature table associated with the ingestion jobs.

Fixes #

```release-note
gRPC responses for submitting spark jobs via job service now have additional fields: start time for all types of jobs, and feature table for ingestion jobs. These fields have also been added to the corresponding Job object returned by list_jobs and get_job API.
```

### DIFF
--- a/protos/feast/core/JobService.proto
+++ b/protos/feast/core/JobService.proto
@@ -73,27 +73,27 @@ message Job {
   // Current job status
   JobStatus status = 3;
   // Deterministic hash of the Job
-  string hash = 8;
+  string hash = 4;
   // Start time of the Job
-  google.protobuf.Timestamp start_time = 9;
+  google.protobuf.Timestamp start_time = 5;
 
   message RetrievalJobMeta {
-    string output_location = 4;
+    string output_location = 1;
   }
 
   message OfflineToOnlineMeta {
-      string feature_table = 10;
+      string table_name = 1;
   }
 
   message StreamToOnlineMeta {
-      string feature_table = 11;
+      string table_name = 1;
   }
 
   // JobType specific metadata on the job
   oneof meta {
-    RetrievalJobMeta retrieval = 5;
-    OfflineToOnlineMeta batch_ingestion = 6;
-    StreamToOnlineMeta stream_ingestion = 7;
+    RetrievalJobMeta retrieval = 6;
+    OfflineToOnlineMeta batch_ingestion = 7;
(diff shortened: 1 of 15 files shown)


protos/feast/core/JobService.proto
```diff
@@ -74,15 +74,19 @@ message Job {
...
   string hash = 8;
+  // Start time of the Job
+  google.protobuf.Timestamp start_time = 9;
 
   message RetrievalJobMeta {
     string output_location = 4;
   }
 
   message OfflineToOnlineMeta {
+      string feature_table = 10;
```
pyalex, line 85: Since it's nested message not sure indexing should be dependent on parent?
`string feature_table = 1;` should work just fine

protos/feast/core/JobService.proto
```diff
@@ -109,6 +113,12 @@ message StartOfflineToOnlineIngestionJobRequest {
 message StartOfflineToOnlineIngestionJobResponse {
     // Job ID assigned by Feast
     string id = 1;
+
+    // Job start time
+    google.protobuf.Timestamp job_start_time = 2;
+
+    // Feature table associated with the job
+    string feature_table = 3;
```
pyalex, line 121: Shouldn't naming be somehow in sync in request / response. In request there are two fields `project` & `table_name`. Is this one a `table_name`?

sdk/python/feast/job_service.py
```diff
@@ -55,6 +55,10 @@
 def _job_to_proto(spark_job: SparkJob) -> JobProto:
     job = JobProto()
     job.id = spark_job.get_id()
+    log_uri = spark_job.get_log_uri()
```
pyalex, line 58: looks strange.
I suggest one-liner
`log_uri = spark_job.get_log_uri() or ""`
or
`log_uri = bool(spark_job.get_log_uri()) and spark_job.get_log_uri(`

khorshuheng: I chose to do the current approach rather than the one liner approach because the latter failed the python linting

khorshuheng: Specifically, get_log_uri returns Optional str, hence the linter consider it as type mismatch, even though the condition ensure that it will never be None

pyalex (approved):

PR #1266, opened 2021-01-14
allow https url for spark ingestion jar

jklegar:

**What this PR does / why we need it**: previously if the spark ingestion jar was a https url (which is the default), ingestion would fail because Feast would try to read and upload the jar as if the url were a local path

Fixes #

```release-note
NONE
```


woop:
@jklegar so this change doesn't apply to dataproc?

jklegar:
I believe on dataproc https should already work due to https://github.com/feast-dev/feast/blob/eaaf23372db7b85d27194e6fdbe9f109f30f7c7c/sdk/python/feast/pyspark/launchers/gcloud/dataproc.py#L278

oavdeev (approved):

PR #1289, opened 2021-01-26
Add telemetry to Python SDK

jklegar:
**What this PR does / why we need it**: adds basic telemetry

Fixes #

```release-note
Starting in v0.9, Feast will begin to collect anonymized usage statistics. To opt out or learn more see https://docs.feast.dev/v/master/advanced/telemetry
```

### DIFF
--- a/docs/advanced/telemetry.md
+++ b/docs/advanced/telemetry.md
@@ -6,5 +6,5 @@ The Feast maintainers use anonymous usage statistics to help shape the Feast roa
 
 ### How to disable telemetry
 
-To opt out of telemetry, simply set the environment variable `FEAST_TELEMETRY` to `False` in the environment in which the Feast client is run. You can check telemetry is disabled by ensuring there is no file named `~/.feast/telemetry`.
+To opt out of telemetry, simply set the environment variable `FEAST_TELEMETRY` to `False` in the environment in which the Feast client is run.
 
--- a/docs/advanced/telemetry.md
+++ b/docs/advanced/telemetry.md
@@ -1,0 +1,10 @@
+# Telemetry
+
+### How telemetry is used
+
+The Feast maintainers use anonymous usage statistics to help shape the Feast roadmap. Several client methods are tracked, beginning in Feast 0.9. Users are assigned a UUID which is sent along with the name of the method, the Feast version, the OS (using `sys.platform`), and the current time. For more detailed information see [the source code](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/telemetry.py).
+
+### How to disable telemetry
+
+To opt out of telemetry, simply set the environment variable `FEAST_TELEMETRY` to `False` in the environment in which the Feast client is run. You can check telemetry is disabled by ensuring there is no file named `~/.feast/telemetry`.
+
--- a/sdk/python/feast/client.py
+++ b/sdk/python/feast/client.py
@@ -388,7 +388,9 @@ class Client:
 
     def _configure_telemetry(self):
         telemetry_filepath = join(expanduser("~"), ".feast", "telemetry")
-        self._telemetry_enabled = self._config.get(opt.TELEMETRY, "True") == "True" # written this way to turn the env var string into a boolean
+        self._telemetry_enabled = (
+            self._config.get(opt.TELEMETRY, "True") == "True"
+        )  # written this way to turn the env var string into a boolean
         if self._telemetry_enabled:
             self._telemetry_counter = {"get_online_features": 0}
             if os.path.exists(telemetry_filepath):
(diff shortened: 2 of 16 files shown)


sdk/python/feast/client.py
```diff
@@ -959,6 +997,15 @@ def get_online_features(
             {'sales:daily_transactions': [1.1,1.2], 'sales:customer_id': [0,1]}
         """
 
+        if self._telemetry_enabled:
+            if self._telemetry_counter["get_online_features"] % 100 == 0:
```
woop, line 1001: I think we can bump this to `% 1000`.

docs/advanced/telemetry.md
```diff
@@ -0,0 +1,10 @@
+# Telemetry
+
+### How telemetry is used
+
+The Feast maintainers use anonymous usage statistics to help shape the Feast roadmap. Several client methods are tracked, beginning in Feast 0.9. Users are assigned a UUID which is sent along with the name of the method, the Feast version, the OS (using `sys.platform`), and the current time. For more detailed information see [the source code](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/telemetry.py).
+
+### How to disable telemetry
+
+To opt out of telemetry, simply set the environment variable `FEAST_TELEMETRY` to `False` in the environment in which the Feast client is run. You can check telemetry is disabled by ensuring there is no file named `~/.feast/telemetry`.
```
woop, line 9: So we are providing two approaches to users? Which one should they use (deletion or env var)?

sdk/python/feast/constants.py
```diff
@@ -259,6 +259,9 @@ class ConfigOptions(metaclass=ConfigMeta):
     #: Oauth token request url
     OAUTH_TOKEN_REQUEST_URL: Optional[str] = None
 
+    #: Telemetry enabled
+    TELEMETRY = "True"
```
woop, line 263: How is this constant set to false?

sdk/python/feast/client.py
```diff
@@ -382,6 +386,25 @@ def version(self):
 
         return result
 
+    def _configure_telemetry(self):
```
woop, line 389: It's not clear to me why this code needs to be on a method in the Client? Would it make more sense to have it in the `telemetry.py` file?

sdk/python/feast/client.py
```diff
@@ -382,6 +386,25 @@ def version(self):
...
+        telemetry_filepath = join(expanduser("~"), ".feast", "telemetry")
+        self._telemetry_enabled = self._config.get(opt.TELEMETRY, "True") == "True"
+        if self._telemetry_enabled:
+            self._telemetry_counter = {"get_online_features": 0}
+            if os.path.exists(telemetry_filepath):
+                with open(telemetry_filepath, "r") as f:
+                    self._telemetry_id = f.read()
+            else:
+                self._telemetry_id = str(uuid.uuid4())
```
woop, line 398: So even if they delete the file, we will continue to recreate it?

sdk/python/feast/client.py
```diff
@@ -382,6 +386,25 @@ def version(self):
...
+                print(
+                    "Feast is an open-source project that collects anonymized usage statistics. To opt out or learn more see https://docs.feast.dev/v/master/advanced/telemetry"
```
woop, line 400: Nit: I've been told not to use `open-source` and instead use `open source` https://developers.google.com/style/word-list#letter-o

sdk/python/telemetry_tests/test_telemetry.py
```diff
@@ -0,0 +1,70 @@
...
+
+    timestamp = datetime.utcnow()
+    try:
+        test_client.apply(entity)
+    except Exception:
+        pass
+
+    sleep(30)
+    bq_client = bigquery.Client()
+    query = f"select * from `kf-feast.feast_telemetry.cloudfunctions_googleapis_com_cloud_functions_*` where JSON_EXTRACT(textPayload, '$.is_test')='\"True\"' and JSON_EXTRACT(textPayload, '$.timestamp')>'\"{timestamp.isoformat()}\"'"
```
woop, line 43: Can we add a time range filter here please? This query can become very expensive if we have a lot of data. We should also ensure that there is day partitioning on the BQ table I think.

sdk/python/feast/client.py
jklegar: made it a method here because it uses/sets a number of instance variables, but could also move it

sdk/python/feast/client.py
jklegar: yes, I don't see a way around that because the file won't exist the very first time. I clarified the docs a bit

woop (approved):

PR #1311, opened 2021-02-03
Autoreload jobservice when configmap changed

khorshuheng:
Currently, job service is not restarted when the configmap changes. As a result, the config map changes won't take effect. This PR enables the job service to restart automatically.

Fixes #

```release-note
Jobservice will now be restarted when the associated configmap is changed.
```

### DIFF
--- a/infra/charts/feast/charts/feast-jobservice/templates/deployment.yaml
+++ b/infra/charts/feast/charts/feast-jobservice/templates/deployment.yaml
@@ -18,8 +18,11 @@ spec:
       release: {{ .Release.Name }}
   template:
     metadata:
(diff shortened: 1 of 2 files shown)


infra/charts/feast/charts/feast-jobservice/templates/deployment.yaml
```diff
@@ -18,8 +18,11 @@ spec:
       release: {{ .Release.Name }}
   template:
     metadata:
-      {{- if .Values.prometheus.enabled }}
       annotations:
+      {{- if .Values.sparkOperator.enabled }}
```
woop, line 22: Is there a reason for this conditional?

woop (approved):

PR #1312, opened 2021-02-03
Fix in install instructions

jonasdebeukelaer:
improves docs.

Use label select instead of grep.
grep was returning multiple pods and crashing for me.


woop:
Thanks @jonasdebeukelaer!

Would you mind targeting the master branch? We can backport the change for you.

jonasdebeukelaer:
@woop np, fixed

woop (approved):

PR #1317, opened 2021-02-05
Add project filter to list jobs

terryyylim:
Currently, list jobs api does not support project filter, and this may result in potential feature table name collisions across different feast projects.

Fixes #

```release-note
Users can now list jobs using project as a filter.
```

### DIFF
--- a/sdk/python/feast/client.py
+++ b/sdk/python/feast/client.py
@@ -1288,13 +1288,18 @@ class Client:
             )
 
     def list_jobs(
-        self, include_terminated: bool, table_name: Optional[str] = None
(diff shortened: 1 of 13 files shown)


sdk/python/feast/client.py
```diff
@@ -1288,13 +1288,18 @@ def start_stream_to_online_ingestion(
             )
 
     def list_jobs(
```
woop, line 1290: Should we have docs for this method @terryyylim?

terryyylim: Yeah we should, added!

PR #1330, opened 2021-02-09
Code cleanup with idiomatic java8 stream

YikSanChan:

I think there was a bug: `sortedFeatures` are not really "sorted" when we make the `sortedFeatureSpecs`.

Cleanup the code with idiomatic java8 stream and fix the bug. Fix the below compilation warning as well.

```
[WARNING] /Users/chenyisheng/source/yiksanchan/feast/core/src/main/java/feast/core/model/FeatureTable.java:[365,38] unchecked call to ArrayList(java.util.Collection<? extends E>) as a member of the raw type java.util.ArrayList
[WARNING] /Users/chenyisheng/source/yiksanchan/feast/core/src/main/java/feast/core/model/FeatureTable.java:[365,38] unchecked conversion
  required: java.util.List<feast.core.model.FeatureV2>
  found:    java.util.ArrayList
```

Fixes #

```release-note

```

### DIFF
--- a/core/src/main/java/feast/core/model/FeatureTable.java
+++ b/core/src/main/java/feast/core/model/FeatureTable.java
@@ -359,13 +359,13 @@ public class FeatureTable extends AbstractTimestampEntity {
 
   public String protoHash() {
     List<String> sortedEntities =
-        this.getEntities().stream().map(entity -> entity.getName()).collect(Collectors.toList());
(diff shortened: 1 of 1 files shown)


woop:
Core will be moved out into https://github.com/feast-dev/feast-java. Please move this PR to that repository going forward.

PR #1336, opened 2021-02-10
WIP Add option to use object store metadata registry

jklegar:

**What this PR does / why we need it**: Add support for an object store registry to the SDK. However, note that this should not be used until serving also supports this registry.

Fixes #

```release-note

```

### DIFF
--- a/protos/feast/core/Registry.proto
+++ b/protos/feast/core/Registry.proto
@@ -29,7 +29,7 @@ message Registry {
     repeated Entity entities = 1;
     repeated FeatureTable feature_tables = 2;
 
-    string registry_schema_version = 3; // to support migrations
-    string version = 4; // version id, random string generated on each update
+    string registry_schema_version = 3; // to support migrations; incremented when schema is changed
+    string version_id = 4; // version id, random string generated on each update of the data; now used only for debugging purposes
     google.protobuf.Timestamp last_updated = 5;
 }
--- a/protos/feast/core/Registry.proto
+++ b/protos/feast/core/Registry.proto
@@ -1,0 +1,35 @@
+//
+// * Copyright 2020 The Feast Authors
+// *
+// * Licensed under the Apache License, Version 2.0 (the "License");
+// * you may not use this file except in compliance with the License.
+// * You may obtain a copy of the License at
+// *
+// *     https://www.apache.org/licenses/LICENSE-2.0
+// *
+// * Unless required by applicable law or agreed to in writing, software
+// * distributed under the License is distributed on an "AS IS" BASIS,
+// * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// * See the License for the specific language governing permissions and
+// * limitations under the License.
+//
+
+syntax = "proto3";
+
+package feast.core;
+option java_package = "feast.proto.core";
+option java_outer_classname = "RegistryProto";
+option go_package = "github.com/feast-dev/feast/sdk/go/protos/feast/core";
+
+import "feast/core/Entity.proto";
+import "feast/core/FeatureTable.proto";
+import "google/protobuf/timestamp.proto";
+
+message Registry {
+    repeated Entity entities = 1;
+    repeated FeatureTable feature_tables = 2;
+
+    string registry_schema_version = 3; // to support migrations
+    string version = 4; // version id, random string generated on each update
+    google.protobuf.Timestamp last_updated = 5;
+}
--- a/sdk/python/feast/client.py
+++ b/sdk/python/feast/client.py
@@ -407,7 +407,9 @@ class Client:
         """
 
         if self._use_object_store_registry:
-            raise NotImplementedError("Projects are not implemented for object store registry.")
+            raise NotImplementedError(
+                "Projects are not implemented for object store registry."
+            )
         else:
             response = self._core_service.ListProjects(
                 ListProjectsRequest(),
@@ -425,7 +427,9 @@ class Client:
         """
 
         if self._use_object_store_registry:
-            raise NotImplementedError("Projects are not implemented for object store registry.")
+            raise NotImplementedError(
+                "Projects are not implemented for object store registry."
+            )
         else:
             self._core_service.CreateProject(
                 CreateProjectRequest(name=project),
@@ -444,7 +448,9 @@ class Client:
         """
 
         if self._use_object_store_registry:
-            raise NotImplementedError("Projects are not implemented for object store registry.")
+            raise NotImplementedError(
+                "Projects are not implemented for object store registry."
+            )
         else:
             try:
                 self._core_service_stub.ArchiveProject(
@@ -619,7 +625,6 @@ class Client:
         if project is None:
             project = self.project
 
-
         if self._use_object_store_registry:
             return self._registry.get_entity(name, project)
         else:
@@ -715,7 +720,8 @@ class Client:
 
             # Get latest feature tables from Feast Core
             feature_table_protos = self._core_service.ListFeatureTables(
-                ListFeatureTablesRequest(filter=filter), metadata=self._get_grpc_metadata(),
+                ListFeatureTablesRequest(filter=filter),
+                metadata=self._get_grpc_metadata(),
             )  # type: ListFeatureTablesResponse
 
             # Extract feature tables and return
@@ -811,7 +817,9 @@ class Client:
         """
 
         if self._use_object_store_registry:
-            raise NotImplementedError("This function is not implemented for object store registry.")
+            raise NotImplementedError(
+                "This function is not implemented for object store registry."
+            )
         else:
             if project is None:
                 project = self.project
--- a/sdk/python/feast/client.py
+++ b/sdk/python/feast/client.py
@@ -539,7 +539,7 @@ class Client:
         """
 
         if self._use_object_store_registry:
-            return self._registry.apply_entity(entity)
+            return self._registry.apply_entity(entity, project)
         else:
             entity.is_valid()
             entity_proto = entity.to_spec_proto()
@@ -574,12 +574,12 @@ class Client:
             List of entities
         """
 
-        if self._use_object_store_registry:
-            return self._registry.list_entities()
-        else:
-            if project is None:
-                project = self.project
+        if project is None:
+            project = self.project
 
+        if self._use_object_store_registry:
+            return self._registry.list_entities(project)
+        else:
             filter = ListEntitiesRequest.Filter(project=project, labels=labels)
 
             # Get latest entities from Feast Core
@@ -616,12 +616,13 @@ class Client:
                 self.version(sdk_only=True),
             )
 
-        if self._use_object_store_registry:
-            return self._registry.get_entity(name)
-        else:
-            if project is None:
-                project = self.project
+        if project is None:
+            project = self.project
 
+
+        if self._use_object_store_registry:
+            return self._registry.get_entity(name, project)
+        else:
             try:
                 get_entity_response = self._core_service.GetEntity(
                     GetEntityRequest(project=project, name=name.strip()),
@@ -668,7 +669,7 @@ class Client:
         """
 
         if self._use_object_store_registry:
-            return self._registry.apply_feature_table(feature_table)
+            return self._registry.apply_feature_table(feature_table, project)
         else:
             feature_table.is_valid()
             feature_table_proto = feature_table.to_spec_proto()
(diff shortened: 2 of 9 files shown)


protos/feast/core/Registry.proto
```diff
@@ -0,0 +1,35 @@
...
+import "feast/core/Entity.proto";
+import "feast/core/FeatureTable.proto";
+import "google/protobuf/timestamp.proto";
+
+message Registry {
+    repeated Entity entities = 1;
+    repeated FeatureTable feature_tables = 2;
+
+    string registry_schema_version = 3; // to support migrations
+    string version = 4; // version id, random string generated on each update
```
woop, line 33: What problem does this random string solve, and can it be replaced with a hash of the objects?

protos/feast/core/Registry.proto
```diff
@@ -0,0 +1,35 @@
...
+option java_outer_classname = "RegistryProto";
+option go_package = "github.com/feast-dev/feast/sdk/go/protos/feast/core";
+
+import "feast/core/Entity.proto";
+import "feast/core/FeatureTable.proto";
+import "google/protobuf/timestamp.proto";
+
+message Registry {
+    repeated Entity entities = 1;
+    repeated FeatureTable feature_tables = 2;
```
woop, line 30: Is the idea that we will first implement this registry and then come back to add FeatureViews?

sdk/python/feast/client.py
```diff
@@ -394,12 +404,15 @@ def list_projects(self) -> List[str]:
...
         """
 
-        response = self._core_service.ListProjects(
-            ListProjectsRequest(),
-            timeout=self._config.getint(opt.GRPC_CONNECTION_TIMEOUT),
-            metadata=self._get_grpc_metadata(),
-        )  # type: ListProjectsResponse
-        return list(response.projects)
+        if self._use_object_store_registry:
+            raise NotImplementedError("Projects are not implemented for object store registry.")
```
woop, line 408: How will we accommodate teams that already have projects implemented. Will the migration script remove the project prefix?

sdk/python/feast/client.py
```diff
@@ -146,6 +147,15 @@ def _core_service(self):
             self._core_service_stub = CoreServiceStub(channel)
         return self._core_service_stub
 
+    @property
+    def _use_object_store_registry(self) -> bool:
```
woop, line 151: I think it's important to have a clean migration path for teams that are adopting the object store registry. If we don't implement the Core reader, then my suggested approach is something like
1. Add object store registry support to SDK (what you are doing)
2. Add a conditional flag like `FALL_BACK_TO_CORE_REGISTRY`. That flag is `True` by default, meaning that  clients will connect to and use the core registry if no registry file is found in the object store during read methods. During create/update operations I think we will have to always create the file, but most of the time the apply methods are controlled by a single team, not by many clients.
3. Teams would migrate by upgrading all client SDKs to the latest version
4. The core registry owners would then shut down their core registry and migrate their configuration to the object store
5. All clients in (3) would then see this new registry file and start using it.
6. The core registry can then be removed entirely.

woop: cc @oavdeev

sdk/python/feast/registry.py
```diff
@@ -0,0 +1,123 @@
...
+            registry_proto.last_updated.FromDatetime(datetime.utcnow())
+            self._filepath.write_bytes(registry_proto.SerializeToString())
+        return
+
+    def _get_registry(self):
+        registry_proto = RegistryProto()
+        registry_proto.ParseFromString(self._filepath.read_bytes())
+        return registry_proto
+
+    def _update_registry(self, updater: Callable[[RegistryProto], RegistryProto]):
```
woop, line 117: I need to come back to this code, but I find the classes and methods here surprisingly complex for what is essentially CRUD.

Is there a way that we can simplify the code and make it more readable/explicit by not passing methods around? I'd be quite happy with a simple base class that indicates to implementers what is expected (acts almost like an interface), with an implementation that contains the majority of the logic. 

MLflow for inspiration: https://github.com/mlflow/mlflow/blob/master/mlflow/store/tracking/abstract_store.py

sdk/python/feast/registry.py
```diff
@@ -0,0 +1,123 @@
...
+    @abstractmethod
+    def _get_registry(self):
+        pass
+
+    @abstractmethod
+    def _update_registry(self, updater: Callable[[RegistryProto], RegistryProto]):
+        pass
+
+
+class LocalRegistry(Registry):
```
oavdeev, line 101: I'd avoid having class hierarchies (except for pure interfaces) as a general rule for better readability. They tend to make things unnecessarily hard to follow when the control flow jumps between the descendant class and the parent. Right now it is fairly simple but as this code will evolve it will likely make things complicated.

One idea would be to separate this into a Registry class (that wouldn't inherit from anything) and maybe RegistryStore(pure ABC), and then LocalRegistryStore would implement get_registry and update_registry. And Registry would use RegistryStore.

sdk/python/feast/registry.py
oavdeev: As someone who propsed this interface in the first place; the idea with this was to make it possible for the implementation to use optimistic locking for read-update-write, and to make the interface as minimal as possible. This is a low level class that would have a nicer higher level wrapper (like the Registry above)

protos/feast/core/Registry.proto
oavdeev: It could be a hash, random string is just easier to implement. Properly hashing protos isn't super straightforward

sdk/python/feast/registry.py
woop: > This is a low level class that would have a nicer higher level wrapper 

Which class are you referring to here, the `LocalRegistry`?

protos/feast/core/Registry.proto
woop: How will this version be used?

sdk/python/feast/registry.py
oavdeev: Yes (i wouldn't have it inherit from Registry tho)

protos/feast/core/Registry.proto
oavdeev: Just debug info for now. For example I would maybe propagate this to tags when creating cloud resources. It is generally nice to have a notion of a "version" for the registry state

sdk/python/feast/registry.py
```diff
@@ -0,0 +1,123 @@
...
+        registry_proto = RegistryProto()
+        registry_proto.ParseFromString(self._filepath.read_bytes())
+        return registry_proto
+
+    def _update_registry(self, updater: Callable[[RegistryProto], RegistryProto]):
+        registry_proto = self._get_registry()
+        registry_proto = updater(registry_proto)
+        registry_proto.version = str(uuid.uuid4())
+        registry_proto.last_updated.FromDatetime(datetime.utcnow())
+        self._filepath.write_bytes(registry_proto.SerializeToString())
```
woop, line 122: Is it a bit weird to use `_get_registry()` which does reading from a file path, but then inside of the `_update_registry()` we have direct access to the file path? Why do we even need _get_registry then?

sdk/python/feast/registry.py
woop: @jklegar can we start with this proposal ^?

protos/feast/core/Registry.proto
woop: Ok, so it's almost closer to an id?

oavdeev: I see, maybe `version_id` would be a clearer name then?

sdk/python/feast/client.py
jklegar: can we do something simpler like:
1. Team that owns Feast installation creates object store by using the (soon-coming) migration script
2. All clients upgraded to use the object store; either no applies are allowed during this period or any applies done to the core registry must be reflected manually to the object store registry
3. Team removes core registry

protos/feast/core/Registry.proto
```diff
@@ -0,0 +1,35 @@
...
+
+    string registry_schema_version = 3; // to support migrations; incremented when schema is changed
+    string version_id = 4; // version id, random string generated on each update of the data; now used only for debugging purposes
```
woop, line 33: Are we actually using it today?

sdk/python/feast/client.py
woop: It's kind of a hairy upgrade path. What if one team (a Feast user, not owners of system) says that they have a critical production system and they can't upgrade their client for at least 2-3 weeks? That means you need to take Feast Core down for 2-3 weeks while you wait for that team to upgrade. It's not always possible to get everyone to upgrade at the same point in time, so naturally different teams will upgrade at different points in time. The easiest is to provide some kind of middle ground.

sdk/python/feast/registry.py
```diff
@@ -0,0 +1,226 @@
...
+from google.auth.exceptions import DefaultCredentialsError
+
+from feast.core.Registry_pb2 import Registry as RegistryProto
+from feast.entity import Entity
+from feast.feature_table import FeatureTable
+
+REGISTRY_SCHEMA_VERSION = "1"
+
+
+class Registry:
```
woop, line 32: Would you mind please adding [Python docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) to classes and methods? You can use your judgement on how much/little detail you want to provide, since some of the methods are documented elsewhere.

sdk/python/feast/registry.py
```diff
@@ -0,0 +1,226 @@
...
+        return
+
+    def _write_registry(self, registry_proto: RegistryProto):
+        registry_proto.version_id = str(uuid.uuid4())
+        registry_proto.last_updated.FromDatetime(datetime.utcnow())
+        self._filepath.write_bytes(registry_proto.SerializeToString())
+        return
+
+
+class GCPRegistryStore(RegistryStore):
```
woop, line 175: Should this be GCS Registry Store?

sdk/python/feast/registry.py
```diff
@@ -0,0 +1,226 @@
...
+        try:
+            from google.cloud import storage
+            from google.cloud.exceptions import NotFound
+        except ImportError:
+            raise ImportError(
+                "Install package google-cloud-storage==1.20.* for gcs support"
+                "run ```pip install google-cloud-storage==1.20.*```"
+            )
+        try:
+            self.gcs_client = storage.Client(project="kf-feast")
```
woop, line 186: Is this `kf-feast` a mistake?

sdk/python/feast/registry.py
```diff
@@ -0,0 +1,226 @@
...
+        self._blob = self._uri.path.lstrip("/")
+        try:
+            bucket = self.gcs_client.get_bucket(self._bucket)
+        except NotFound:
+            bucket = self.gcs_client.create_bucket(self._bucket)
+            print(bucket)
+        if not storage.Blob(bucket=bucket, name=self._blob).exists(self.gcs_client):
+            registry_proto = RegistryProto()
+            registry_proto.registry_schema_version = REGISTRY_SCHEMA_VERSION
+            self._write_registry(registry_proto)
```
woop, line 200: I think it's a bit of a trade off on when we would try and read/write to the registry. My intuition is to do this only when the user runs the `apply()` commands. Does writing the proto here mean that when the user creates an instance of `Client`, we will start writing to the object store?

sdk/python/tests/test_client.py
```diff
@@ -448,6 +515,82 @@ def test_apply_feature_table_success(self, test_client):
             and feature_tables[0].entities[0] == "fs1-my-entity-1"
         )
 
+    @pytest.mark.integration
+    @pytest.mark.parametrize(
+        "test_client", [lazy_fixture("client_with_object_registry_gcs")],
+    )
+    def test_apply_feature_table_integration(self, test_client):
+
+        test_client.set_project("project1")
```
woop, line 524: Is `project1` supposed to be used somewhere?

protos/feast/core/Registry.proto
jklegar: defer to @oavdeev here but I think the idea is we can see it when we look at the stored proto if we need to debug something

sdk/python/tests/test_client.py
jklegar: this is copied from `test_apply_feature_table_success`; I checked and it's not used/needed in either

sdk/python/feast/registry.py
```diff
@@ -30,10 +30,21 @@
 
 
 class Registry:
+    """
+    Registry: Used for interfacing with the object store registry.
```
woop, line 34: Sorry, another nitpick. `Foo is an interface for interacting with a foo` is not really descriptive enough. And calling it an interface is also a little bit misleading.

How about `A registry allows for the management and persistence of feature definitions and related metadata`?

sdk/python/feast/registry.py
```diff
@@ -136,27 +209,44 @@ def updater(registry_proto: RegistryProto):
 
 
 class RegistryStore(ABC):
+    """
+    RegistryStore: specific implementations of the object store registry for local file system and GCS.
```
woop, line 213: Is this a typo? If this is an ABC is it really a specific implementation?

woop: Also, it seems like the Registry and Registry Store comments are overlapping.

PR #1360, opened 2021-03-01
(WIP) Support Keto authorization directly

khorshuheng:
The purpose of this PR is to demonstrate the feasibility of supporting Keto as authorization provider directly without going through feast-keto-auth-server. Further discussion will be needed before deciding whether this feature should be included in Feast or not.

Fixes #

```release-note

```

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -125,6 +125,11 @@
             <groupId>com.google.auth</groupId>
             <artifactId>google-auth-library-oauth2-http</artifactId>
         </dependency>
+        <dependency>
+            <groupId>sh.ory.keto</groupId>
+            <artifactId>keto-client</artifactId>
+            <version>0.5.7-alpha.1.pre.0</version>
+        </dependency>
         
         <!-- HTTP/REST -->
         <dependency>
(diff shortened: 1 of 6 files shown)


woop:
I dont understand why we need this functionality when we already have an auth server implementation that works.

khorshuheng:
> I dont understand why we need this functionality when we already have an auth server implementation that works.

The issue is that, feast auth server would be an additional service to be deployed and maintain. I agree that there are pros and cons between communicating with Keto server directly, or use feast auth server as a middle man. Internally we are still discussing what's the best way to deal with this, hence the WIP progress label.

woop:
> > I dont understand why we need this functionality when we already have an auth server implementation that works.
> 
> The issue is that, feast auth server would be an additional service to be deployed and maintain. I agree that there are pros and cons between communicating with Keto server directly, or use feast auth server as a middle man. Internally we are still discussing what's the best way to deal with this, hence the WIP progress label.

The middleman service is a stateless service that just acts as a proxy. It's as low maintenance as it gets. We cant maintain every implementation for each team. Ultimately we need to allow teams to maintain their own implementations, which is why we created the adapter in the first place.

woop:
Also, you should be targeting https://github.com/feast-dev/feast-java, not the 0.9 branch.

woop:
I'm closing this for now but I am happy to discuss it in our weekly call.

PR #1362, opened 2021-03-01
Python-centric feast deploy CLI

oavdeev:
Signed-off-by: Oleg Avdeev <oleg.v.avdeev@gmail.com>

Add `feast apply` and `feast-teardown` commands, supporting new feature repo functionality and GCP Firestore as an online feature store.

```release-note
NONE
```


sdk/python/feast/cli.py
```diff
@@ -353,5 +356,38 @@ def project_list():
     print(tabulate(table, headers=["NAME"], tablefmt="plain"))
 
 
+@cli.command("apply")
+@click.argument("repo_path", type=click.Path(dir_okay=True, exists=True))
+def apply_total_command(repo_path: str):
+    """
+    Applies a feature repo
+    """
+    repo_config = load_repo_config(Path(repo_path))
```
woop, line 365: The plan is to eventually use the FeatureStore class here right?

sdk/python/feast/infra/provider.py
```diff
@@ -0,0 +1,44 @@
+import abc
+from typing import List
+
+from feast import FeatureTable
+from feast.repo_config import FirestoreConfig, OnlineStoreConfig
+
+
+class Provider(abc.ABC):
```
woop, line 8: Can you define a `provider` please? It seems like it is synonymous with an online store here. The original idea was that it is either related to the whole environment (`aws`) or related to your compute, not a specific store.

sdk/python/feast/infra/firestore.py
```diff
@@ -0,0 +1,65 @@
+from typing import List
+
+from feast import FeatureTable
+from feast.infra.provider import Provider
```
woop, line 4: also a bit of a nitpick, but does `infra` add value here? What do we lose if we remove `infra` as a submodule?

sdk/python/feast/cli.py
oavdeev: Yeah I imagine there will be a nicer interface for notebooks. I just tend to implement bits of functionality as pure functions by default (where possible), so they're easier to unit test.

sdk/python/tests/cli/test_cli_local.py
```diff
@@ -1,3 +1,54 @@
...
+
+
+class CliRunner:
+    """
+    NB. We can't use test runner helper from click here, since it doesn't start a new Python
+    interpreter. And we need a new interpreter for each test since we dynamically import
+    modules from the feature repo, and it is hard to clean up that state otherwise.
+    """
+
+    def run(self, args: List[str], cwd: Path) -> subprocess.CompletedProcess:
```
woop, line 18: Do we need a class + method, or can this be a function?

sdk/python/feast/repo_config.py
```diff
@@ -0,0 +1,26 @@
...
+
+class LocalOnlineStoreConfig(NamedTuple):
+    path: str
+
+
+class OnlineStoreConfig(NamedTuple):
+    local: Optional[LocalOnlineStoreConfig] = None
+
+
+class RepoConfig(NamedTuple):
```
woop, line 16: What if we just called this `Config`?

sdk/python/feast/cli.py
woop: Yea I agree with that approach, I just want to make sure that we dont have two code paths that function differently.

sdk/python/feast/repo_config.py
oavdeev: I think we already have a Config or two so I wanted to be a bit more specific here

jklegar: We should reconcile https://github.com/feast-dev/feast/blob/master/sdk/python/feast/feature_store_config.py and this since I think they're the same thing

sdk/python/tests/cli/test_cli_local.py
oavdeev: My plan was to make this class _also_ set up a temporary repo directory and maybe include some other helpers but i didn't get around to it yet

sdk/python/feast/infra/firestore.py
oavdeev: i wanted all cloud specific stuff to be in a subpackage, could rename the subpackage to "provider" too

sdk/python/feast/cli.py
```diff
@@ -353,5 +356,38 @@ def project_list():
...
+
+    apply_total(repo_config, Path(repo_path).resolve())
+
+
+@cli.command("teardown")
+@click.argument("repo_path", type=click.Path(dir_okay=True, exists=True))
+def teardown_command(repo_path: str):
+    """
+    Tear down infra for a a feature repo
```
woop, line 374: Typo

sdk/python/feast/cli.py
```diff
@@ -26,6 +27,8 @@
 from feast.entity import Entity
 from feast.feature_table import FeatureTable
 from feast.loaders.yaml import yaml_loader
+from feast.repo_config import load_repo_config
```
woop, line 30: Is the plan for @jklegar to dedupe repo_config and Config?

sdk/python/feast/cli.py
```diff
@@ -353,5 +356,38 @@ def project_list():
...
+def teardown_command(repo_path: str):
+    """
+    Tear down infra for a a feature repo
+    """
+    repo_config = load_repo_config(Path(repo_path))
+
+    teardown(repo_config, Path(repo_path).resolve())
+
+
+@cli.command("registry-dump")
```
woop, line 381: `export registry`? What format are we supporting here?

sdk/python/feast/feature_store.py
```diff
@@ -22,13 +22,13 @@ class FeatureStore:
     """
 
     def __init__(
-        self, config_path: Optional[str], config: Optional[Config],
+        self, config_path: Optional[str], config: Optional[RepoConfig],
     ):
         if config_path is None or config is None:
```
woop, line 27: Wait, isn't this a bug? If we set config_path=None then this exception will raise.

sdk/python/feast/infra/gcp.py
```diff
@@ -0,0 +1,74 @@
...
+from typing import List, Optional
+
+from feast import FeatureTable
+from feast.infra.provider import Provider
+from feast.repo_config import DatastoreOnlineStoreConfig
+
+
+def _delete_all_values(client, key) -> None:
+    """
+    Delete all keys under the key path in firestore.
```
woop, line 11: Nit: `firestore` -> `Firestore`.

sdk/python/feast/infra/gcp.py
```diff
@@ -0,0 +1,74 @@
...
+        entities = list(query.fetch(limit=1000))
+        if not entities:
+            return
+
+        for entity in entities:
+            print("Deleting: {}".format(entity))
+            client.delete(entity.key)
+
+
+class Gcp(Provider):
```
woop, line 24: Do we have a plan for reconciling the `OnlineStore` ABC and `OfflineStore` ABC approach that @jklegar is taking with the approach here (provider centric)?

sdk/python/feast/infra/gcp.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def _initialize_client(self):
+        from google.cloud import datastore
+
+        if self._project_id is not None:
+            return datastore.Client(self.project_id)
+        else:
+            return datastore.Client()
+
+    def update_infra(
```
woop, line 41: Once we've swept through PR comments, can we get Google style docstrings for all public methods?

sdk/python/feast/infra/gcp.py
```diff
@@ -0,0 +1,74 @@
...
+
+        for table in tables_to_delete:
+            _delete_all_values(
+                client, client.key("FeastProject", project, "FeatureTable", table.name)
+            )
+
+            key = client.key("FeastProject", project, "FeatureTable", table.name)
+            client.delete(key)
+
+    def teardown_infra(self, project: str, tables: List[FeatureTable]) -> None:
```
woop, line 65: Does `infra` include the registry? It seems like its only the online store.

sdk/python/feast/infra/provider.py
```diff
@@ -0,0 +1,49 @@
...
+
+def get_provider(config: RepoConfig) -> Provider:
+    if config.provider == "gcp":
+        from feast.infra.gcp import Gcp
+
+        return Gcp(config.online_store.datastore)
+    elif config.provider == "local":
+        from feast.infra.local_sqlite import LocalSqlite
+
+        assert config.online_store.local is not None
```
woop, line 46: What would set `config.online_store.local`? Is it a default value?

woop:
I also think that we should be adding docs as part of this PR, even though I have't had time to scaffold a location for the documentation as part of docs.feast.dev.

sdk/python/feast/cli.py
oavdeev: It prints json, for now it is purely a small utility for debugging. I didn't call it "export" because you can't really import the result to anything

sdk/python/feast/infra/gcp.py
oavdeev: now that you mention it, it should actually be "datastore" :) let me fix that

sdk/python/feast/infra/gcp.py
oavdeev: This one has a comment on the base abstract interface, so i didn't duplicate it here (it is just a gcp specific implementation)

sdk/python/feast/infra/gcp.py
oavdeev: No, in my mind registry is a separate thing.

sdk/python/feast/infra/provider.py
oavdeev: it is read from the config, there is no default right now but we can add that

oavdeev:
>Would it make sense to have some minimal documentation in a .md file? Or do you think we have to wait for docs.feast.dev to be ready?

It is not that I want to wait but i have more followup PRs for reading and writing on top of this, so I'd rather merge this quickly and add docs separately

sdk/python/feast/infra/provider.py
woop: I think we need to have conventions for `local` and `gcp`. That way `assert` should never actually be hit, otherwise we should probably use a normal exception here.

oavdeev: This assert is not supposed to be hit, it is there to make mypy happy. Config should be validated when it is being parsed, but due to how parsing works rn this member has to be `Optional`, so I need a type assertion here for mypy

sdk/python/feast/feature_store.py
oavdeev: I think it is fixed in #1366 ?

sdk/python/feast/infra/gcp.py
```diff
@@ -0,0 +1,74 @@
+from datetime import datetime
+from typing import List, Optional
+
+from feast import FeatureTable
+from feast.infra.provider import Provider
```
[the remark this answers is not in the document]
oavdeev: Yea i don't feel strongly, was hoping to have provider specific stuff in a submodule but not sure if there ever will be a nice split. Maybe let's see after this and online store RW path is done, if there's no clear line, I can fold it into main module (or name it differently?)

sdk/python/feast/infra/gcp.py
oavdeev: I'd see once both PRs land, I don't think they're in the way of each other right now but there may be a nicer structure

sdk/python/feast/infra/firestore.py
woop: I guess its not a big deal for now. let's just leave it as is.

sdk/python/feast/infra/provider.py
woop: Ok, noted.

woop (approved):

PR #1364, opened 2021-03-02
FeatureStore, FeatureView, Config, and BIgQuerySource classes for updated SDK

jklegar:

**What this PR does / why we need it**: Scaffold classes for updated SDK; FeatureStore will replace the Client class but these are currently just mostly-empty classes.

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/big_query_source.py
+++ b/sdk/python/feast/big_query_source.py
@@ -15,6 +15,10 @@ from typing import Dict, Optional
 
 
 class BigQuerySource:
+    """
+    Represents a BigQuery table reference or BigQuery query that returns a set of features.
+    """
+
     def __init__(
         self,
         table_ref: Optional[str],
@@ -23,7 +27,8 @@ class BigQuerySource:
         field_mapping: Optional[Dict[str, str]],
         query: Optional[str],
     ):
-        assert (table_ref is None) != (query is None), "Exactly one of table_ref and query should be specified"
+        if (table_ref is None) != (query is None):
+            raise Exception("Exactly one of table_ref and query should be specified")
         self.table_ref = table_ref
         self.event_timestamp_column = event_timestamp_column
         self.created_timestamp_column = created_timestamp_column
--- a/sdk/python/feast/big_query_source.py
+++ b/sdk/python/feast/big_query_source.py
@@ -1,0 +1,32 @@
+# Copyright 2019 The Feast Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import Dict, Optional
+
+
+class BigQuerySource:
+    def __init__(
+        self,
+        table_ref: Optional[str],
+        event_timestamp_column: str,
+        created_timestamp_column: Optional[str],
+        field_mapping: Optional[Dict[str, str]],
+        query: Optional[str],
+    ):
+        assert (table_ref is None) != (query is None), "Exactly one of table_ref and query should be specified"
+        self.table_ref = table_ref
+        self.event_timestamp_column = event_timestamp_column
+        self.created_timestamp_column = created_timestamp_column
+        self.field_mapping = field_mapping
+        self.query = query
+        return
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -12,20 +12,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from typing import Optional
+
 from feast.feature_store_config import Config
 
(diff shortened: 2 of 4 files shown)


sdk/python/feast/big_query_source.py
```diff
@@ -0,0 +1,32 @@
...
+class BigQuerySource:
+    def __init__(
+        self,
+        table_ref: Optional[str],
+        event_timestamp_column: str,
+        created_timestamp_column: Optional[str],
+        field_mapping: Optional[Dict[str, str]],
+        query: Optional[str],
+    ):
+        assert (table_ref is None) != (query is None), "Exactly one of table_ref and query should be specified"
```
woop, line 26: Small nitpick, but generally I think we should avoid using assert for errors that we expect users will make in lieu of normal exceptions

https://stackoverflow.com/questions/40182944/whats-the-difference-between-raise-try-and-assert#:~:text=raise%20is%20typically%20used%20when,if%20a%20condition%20is%20met.&text=If%20production%20quality%20code%20raises,ve%20got%20a%20bigger%20problem.

Assert should normally be used for impossible to reach situations.

sdk/python/feast/feature_store.py
```diff
@@ -0,0 +1,31 @@
...
+        config: Optional[Config],
+    ):
+        assert (config_path is None or config is None), "You cannot specify both config_path and config"
+        if config is not None:
+            self.config = config
+        elif config_path is not None:
+            self.config = Config.from_config_path(config_path)
+        else:
+            self.config = Config()
+        return
```
woop, line 31: Do we need a return here?

sdk/python/feast/feature_store.py
```diff
@@ -0,0 +1,31 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import Optional
+from feast.feature_store_config import Config
+
+
+class FeatureStore:
```
woop, line 18: I think we should add a comment to the FeatureStore class.

sdk/python/feast/feature_store_config.py
```diff
@@ -0,0 +1,39 @@
...
+class Config:
+    def __init__(
+        self,
+        provider: Optional[str],
+        online_store: Optional[str],
+        metadata_store: Optional[str],
+    ):
+        self.provider = provider if (provider is not None) else "local"
+        self.online_store = online_store if (online_store is not None) else "local"
+        self.metadata_store = metadata_store if (metadata_store is not None) else join(expanduser("~"), ".feast", "metadata_store")
```
woop, line 28: Would it make sense to rather use the project repository as the local metadata store location? Using a shared folder across multiple projects seems like it can cause problems.

sdk/python/feast/feature_store.py
woop: I'm not sure how useful `fat client` is to users. How about `A FeatureStore object is used to define, create, and retrieve features.`

sdk/python/feast/feature_store_config.py
```diff
@@ -0,0 +1,44 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import Optional
+
+import yaml
+
+
+class Config:
+    """
+    Python representation of the FeatureStore config that the user can specify via a yaml file.
```
woop, line 21: Nit: If the user doesnt know what the yaml file is then this comment doesnt add value. How about `Configuration object that contains all possible configuration options for a FeatureStore`?

sdk/python/feast/feature_view.py
```diff
@@ -0,0 +1,46 @@
...
+from typing import Dict, List
+
+from feast.big_query_source import BigQuerySource
+from feast.entity import Entity
+from feast.feature import Feature
+
+
+class FeatureView:
+    """
+    Represents a collection of features that will be served online.
```
woop, line 24: We could in theory also have batch only users. How about `A FeatureView defines a logical grouping of servable features`?

sdk/python/feast/feature_store_config.py
```diff
@@ -0,0 +1,44 @@
...
+        metadata_store: Optional[str],
+    ):
+        self.provider = provider if (provider is not None) else "local"
+        self.online_store = online_store if (online_store is not None) else "local"
+        self.metadata_store = (
+            metadata_store if (metadata_store is not None) else "./metadata_store"
+        )
+
+    @classmethod
+    def from_config_path(cls, config_path: str):
```
woop, line 37: `from_path` perhaps?

sdk/python/feast/feature_store_config.py
```diff
@@ -0,0 +1,44 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import Optional
+
+import yaml
+
+
+class Config:
```
woop, line 19: Is this `Config` overlapping with https://github.com/feast-dev/feast/pull/1362/files#diff-c90c6781dcec3cdcc26843e013d3d6e9c1ce89d084b3a67e818fe76a0c7d0436R16?

woop (approved):

PR #1366, opened 2021-03-05
Download BigQuery table to pyarrow table for python-based ingestion flow

jklegar:
**What this PR does / why we need it**: Implements downloading a BQ table to an in-memory pyarrow table, which will be part of the python-based ingestion flow

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/big_query_source.py
+++ b/sdk/python/feast/big_query_source.py
@@ -27,8 +27,28 @@ class BigQuerySource:
         field_mapping: Optional[Dict[str, str]],
         query: Optional[str],
     ):
-        if (table_ref is None) != (query is None):
+        if (table_ref is None) == (query is None):
             raise Exception("Exactly one of table_ref and query should be specified")
+        if field_mapping is not None:
+            for value in field_mapping.values():
+                if list(field_mapping.values()).count(value) > 1:
+                    raise Exception(
+                        f"Two fields cannot be mapped to the same name {value}"
+                    )
+
+            if event_timestamp_column in field_mapping.keys():
+                raise Exception(
+                    f"The field {event_timestamp_column} is mapped to {field_mapping[event_timestamp_column]}. Please either remove this field mapping or use {field_mapping[event_timestamp_column]} as the event_timestamp_column."
+                )
+
+            if (
+                created_timestamp_column is not None
+                and created_timestamp_column in field_mapping.keys()
+            ):
+                raise Exception(
+                    f"The field {created_timestamp_column} is mapped to {field_mapping[created_timestamp_column]}. Please either remove this field mapping or use {field_mapping[created_timestamp_column]} as the _timestamp_column."
+                )
+
         self.table_ref = table_ref
         self.event_timestamp_column = event_timestamp_column
         self.created_timestamp_column = created_timestamp_column
--- a/sdk/python/feast/big_query_source.py
+++ b/sdk/python/feast/big_query_source.py
@@ -21,23 +21,23 @@ class BigQuerySource:
 
     def __init__(
         self,
-        table_ref: Optional[str],
         event_timestamp_column: str,
-        created_timestamp_column: Optional[str],
-        field_mapping: Optional[Dict[str, str]],
-        query: Optional[str],
+        table_ref: Optional[str] = None,
+        created_timestamp_column: Optional[str] = None,
+        field_mapping: Optional[Dict[str, str]] = None,
+        query: Optional[str] = None,
     ):
         if (table_ref is None) == (query is None):
-            raise Exception("Exactly one of table_ref and query should be specified")
+            raise ValueError("Exactly one of table_ref and query should be specified")
         if field_mapping is not None:
             for value in field_mapping.values():
                 if list(field_mapping.values()).count(value) > 1:
-                    raise Exception(
+                    raise ValueError(
                         f"Two fields cannot be mapped to the same name {value}"
                     )
 
             if event_timestamp_column in field_mapping.keys():
-                raise Exception(
+                raise ValueError(
                     f"The field {event_timestamp_column} is mapped to {field_mapping[event_timestamp_column]}. Please either remove this field mapping or use {field_mapping[event_timestamp_column]} as the event_timestamp_column."
                 )
 
@@ -45,7 +45,7 @@ class BigQuerySource:
                 created_timestamp_column is not None
                 and created_timestamp_column in field_mapping.keys()
             ):
-                raise Exception(
+                raise ValueError(
                     f"The field {created_timestamp_column} is mapped to {field_mapping[created_timestamp_column]}. Please either remove this field mapping or use {field_mapping[created_timestamp_column]} as the _timestamp_column."
                 )
 
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -14,7 +14,6 @@
 from pathlib import Path
 from typing import Optional
 
-
 from feast.infra.provider import Provider, get_provider
 from feast.registry import Registry
 from feast.repo_config import (
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -71,13 +71,9 @@ class FeatureStore:
 
         # if we have mapped fields, use the original field names in the call to the offline store
         event_timestamp_column = feature_view.inputs.event_timestamp_column
-        fields = (
-            [entity.name for entity in feature_view.entities]
-            + [feature.name for feature in feature_view.features]
-            + [feature_view.inputs.event_timestamp_column]
-        )
-        if feature_view.inputs.created_timestamp_column is not None:
-            fields.append(feature_view.inputs.created_timestamp_column)
+        entity_names = [entity.name for entity in feature_view.entities]
+        feature_names = [feature.name for feature in feature_view.features]
+        created_timestamp_column = feature_view.inputs.created_timestamp_column
         if feature_view.inputs.field_mapping is not None:
             reverse_field_mapping = {
                 v: k for k, v in feature_view.inputs.field_mapping.items()
@@ -87,17 +83,31 @@ class FeatureStore:
                 if event_timestamp_column in reverse_field_mapping.keys()
                 else event_timestamp_column
             )
-            fields = [
+            created_timestamp_column = (
+                reverse_field_mapping[created_timestamp_column]
+                if created_timestamp_column is not None
+                and created_timestamp_column in reverse_field_mapping.keys()
+                else created_timestamp_column
+            )
+            entity_names = [
                 reverse_field_mapping[col]
                 if col in reverse_field_mapping.keys()
                 else col
-                for col in fields
+                for col in entity_names
+            ]
+            feature_names = [
+                reverse_field_mapping[col]
+                if col in reverse_field_mapping.keys()
+                else col
+                for col in feature_names
             ]
 
         table = BigQueryOfflineStore.pull_table(
             feature_view.inputs.table_ref,
-            fields,
+            entity_names,
+            feature_names,
             event_timestamp_column,
+            created_timestamp_column,
             start_date,
             end_date,
         )
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -76,6 +76,8 @@ class FeatureStore:
             + [feature.name for feature in feature_view.features]
             + [feature_view.inputs.event_timestamp_column]
         )
+        if feature_view.inputs.created_timestamp_column is not None:
+            fields.append(feature_view.inputs.created_timestamp_column)
         if feature_view.inputs.field_mapping is not None:
             reverse_field_mapping = {
                 v: k for k, v in feature_view.inputs.field_mapping.items()
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -11,15 +11,11 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from datetime import datetime
 from pathlib import Path
 from typing import Optional
 
-import pyarrow
 
-from feast.feature_view import FeatureView
 from feast.infra.provider import Provider, get_provider
-from feast.offline_store import BigQueryOfflineStore
 from feast.registry import Registry
 from feast.repo_config import (
     LocalOnlineStoreConfig,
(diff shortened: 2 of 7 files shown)


sdk/python/requirements-dev.txt
```diff
@@ -2,7 +2,7 @@ Click==7.*
 google-api-core==1.22.4
 google-auth==1.22.1
 google-cloud-bigquery==1.18
-google-cloud-bigquery-storage==0.7.0
+google-cloud-bigquery-storage==2.3.0
```
jklegar, line 5: these do not seem to be used so upgrading should be fine

sdk/python/feast/offline_store.py
```diff
@@ -0,0 +1,100 @@
...
+        project, dataset, table = feature_view.inputs.table_ref.split(".")
+        client = BigQueryReadClient()
+
+        requested_session = types.ReadSession()
+        requested_session.table = (
+            f"projects/{project}/datasets/{dataset}/tables/{table}"
+        )
+        requested_session.data_format = types.DataFormat.ARROW
+
+        # if we have mapped fields, use the original field names in the call to BigQuery
```
jklegar, line 56: I put some logic to deal with field mappings in here but plan on moving it out into materialize() in the next diff

jklegar:
btw this is stacked on #1364 but I don't think GH PR's support stacked diffs

sdk/python/feast/big_query_source.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    Represents a BigQuery table reference or BigQuery query that returns a set of features.
+    """
+
+    def __init__(
+        self,
+        table_ref: Optional[str],
+        event_timestamp_column: str,
+        created_timestamp_column: Optional[str],
+        field_mapping: Optional[Dict[str, str]],
```
woop, line 27: I think what isn't clear here is what direction of the field mapping.

sdk/python/feast/big_query_source.py
```diff
@@ -0,0 +1,57 @@
...
+        query: Optional[str],
+    ):
+        if (table_ref is None) == (query is None):
+            raise Exception("Exactly one of table_ref and query should be specified")
```
woop, line 31: https://docs.python.org/3/library/exceptions.html#ValueError is preferred I think.

sdk/python/feast/offline_store.py
```diff
@@ -0,0 +1,100 @@
...
+        parent = f"projects/{project}"
+        session = client.create_read_session(
+            parent=parent, read_session=requested_session, max_stream_count=1,
+        )
+
+        if len(session.streams) == 0:
+            # will sometimes happen, indicates no data
+            return None
+        reader = client.read_rows(session.streams[0].name)
+        table = reader.to_arrow(session)
```
woop, line 90: Does this mean that we are loading all source data into memory (not paging)?

sdk/python/setup.py
```diff
@@ -31,7 +31,7 @@
     "google-cloud-storage==1.20.*",
     "google-cloud-core==1.0.*",
     "googleapis-common-protos==1.52.*",
-    "google-cloud-bigquery-storage==0.7.*",
+    "google-cloud-bigquery-storage==2.3.*",
```
woop, line 34: I think this is too tight. How about >?

sdk/python/feast/offline_store.py
```diff
@@ -0,0 +1,100 @@
...
+    ):
+        pass
+
+
+class BigQueryOfflineStore:
+    """
+    BigQueryOfflineStore is a non-user-facing object used for all interaction between Feast and BigQuery.
+    """
+
+    def pull_table(
```
woop, line 42: `pull_table` seems like it squats generic functionality, but it's actually very specific (it contains a FeatureView argument and time range). I would expect this method to have a query interface like `pull_table(self, query)`. That way I can also reuse this method for generating a training dataset. I also think the method can be cleaner if the query building was done separately in a helper method.

Would it make sense to have two higher level methods for us in that case? Something like `export_feature_view_to_arrow(self, feature_view: FeatureView, start_date: datetime, end_date: datetime,
) -> Optional[pyarrow.Table]`
and 
`export_training_features_to_arrow(self, entity_df_query, features, feature_views) -> Optional[pyarrow.Table]`
and both of these would use
`_export_bigquery_table_to_arrow(self, query) -> Optional[pyarrow.Table] `

sdk/python/feast/offline_store.py
```diff
@@ -0,0 +1,100 @@
...
+
+        requested_session = types.ReadSession()
+        requested_session.table = (
+            f"projects/{project}/datasets/{dataset}/tables/{table}"
+        )
+        requested_session.data_format = types.DataFormat.ARROW
+
+        # if we have mapped fields, use the original field names in the call to BigQuery
+        event_timestamp_column = feature_view.inputs.event_timestamp_column
+        fields = (
```
woop, line 58: Can we please add a few comments to this code so that its easier for a reader to parse

sdk/python/feast/big_query_source.py
jklegar: oh yeah I've been assuming it's key=table, value=feature

woop: Ah, I was thinking the reverse.

sdk/python/feast/offline_store.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+
+    @abstractmethod
+    def pull_table(
+        self, feature_view: FeatureView, start_date: datetime, end_date: datetime,
+    ):
+        pass
+
+
+class BigQueryOfflineStore:
```
woop, line 37: Should there be inheritance here?

sdk/python/feast/big_query_source.py
jklegar: Hm, intuitively table->feature makes more sense to me but I may be in the minority with that viewpoint - do we do the reverse in Feast today?

sdk/python/feast/offline_store.py
jklegar: I need to think through paging a little bit more but think it shouldn't be in this function since we'd want to alternate reading a page into memory and writing it to the online store, so I'm deferring implementing paging until we have the full ingestion workflow

sdk/python/feast/offline_store.py
jklegar: changed the interface - would this be reusable for historical feature retrieval now?

sdk/python/feast/big_query_source.py
woop: I dont think it matters, as long as we document it.

sdk/python/feast/feature_store_config.py
```diff
@@ -23,9 +23,9 @@ class Config:
 
     def __init__(
         self,
-        provider: Optional[str],
-        online_store: Optional[str],
-        metadata_store: Optional[str],
+        provider: Optional[str] = None,
```
woop, line 26: Shouldn't we be defining the default provider here instead of in the constructor?

sdk/python/feast/feature_store_config.py
```diff
@@ -23,9 +23,9 @@ class Config:
...
+        online_store: Optional[str] = None,
+        metadata_store: Optional[str] = None,
     ):
         self.provider = provider if (provider is not None) else "local"
         self.online_store = online_store if (online_store is not None) else "local"
```
woop, line 31: I think we need a default `online_store` for the local provider, not a default online store regardless of provider.

sdk/python/feast/offline_store.py
woop: Feedback on the entity query was been quite positive. I dont think we can get away with asking users to provide a table for the entity df, even if they can write a view. It's a poor development experience, unless we provide other functionality like row filters, column filters, sampling, etc.

sdk/python/feast/feature_store.py
```diff
@@ -22,13 +27,64 @@ class FeatureStore:
...
+        if config_path is not None and config is not None:
+            raise ValueError("You cannot specify both config_path and config")
         if config is not None:
             self.config = config
         elif config_path is not None:
             self.config = Config.from_path(config_path)
         else:
             self.config = Config()
+
+    def _pull_table(
```
woop, line 41: It seems like `_pull_table` would be better placed within the storage layer (`BigQueryStore`) itself, and not the `FeatureStore` class. I can't see this `feature_view.inputs.table_ref` holding up long term. Every data source will have their own implementation details. Like lets say we are pulling a parquet file that is compressed, where would the configuration for compression go?

I think the pull table argument should maybe take a `Union[DataSource, EntityDataFrame]`, and so the BigQueryOfflineStore should take `Union[BigQueryDataSource, BigQueryEntityDataFrame]` or something. At least we can fully retrieve the data from within the storage layer.

woop: Also, `_pull_table()` seems like it should be a function and not a method, so that it is easier to test.

sdk/python/feast/offline_store.py
jklegar: Oh not saying to get rid of the entity query, the question was meant to ask is this method reusable in historical feature retrieval as a wrapper around the BQ storage api? The `export_training_features` method would still need to run the query outside of this method (and the BQ storage api doesn't support arbitrary queries anyway) to produce the table/view

woop: I'll try to make it work, but I am not sure about the table_ref. I can change it later if necessary though.

sdk/python/feast/feature_store.py
```diff
@@ -22,13 +27,64 @@ class FeatureStore:
...
+                else event_timestamp_column
+            )
+            fields = [
+                reverse_field_mapping[col]
+                if col in reverse_field_mapping.keys()
+                else col
+                for col in fields
+            ]
+
+        table = BigQueryOfflineStore.pull_table(
```
woop, line 72: One thing that we should remember to implement is to only return the latest feature data from the source table for each key.

woop (changes requested):

sdk/python/feast/feature_store.py
jklegar: Right, easiest place to do this seems to be when ingesting into the online store but I wasn't grabbing the created_timestamp_column from BigQuery so I'm doing that now

sdk/python/feast/repo_config.py
```diff
@@ -19,10 +19,10 @@ class OnlineStoreConfig(NamedTuple):
 
 
 class RepoConfig(NamedTuple):
-    metadata_store: str
-    project: str
-    provider: str
-    online_store: OnlineStoreConfig
+    metadata_store: str = "./metadata_store"
+    project: str = "default"
```
jklegar, line 23: @oavdeev is this meant to be the GCP project?

woop: No I am pretty sure it's the Feast project.

sdk/python/feast/offline_store.py
```diff
@@ -0,0 +1,140 @@
...
+from typing import List, Optional, Tuple
+
+import pyarrow
+
+from feast.feature_view import FeatureView
+
+
+class OfflineStore(ABC):
+    """
+    OfflineStore is a non-user-facing object used for all interaction between Feast and the service used for offline storage of features. Currently BigQuery is supported.
```
woop, line 25: Nit: I think we can simplify this. Non-user-facing isnt really relevant.

What about `An OfflineStore is used for the storage and retrieval of historical feature data`

sdk/python/feast/offline_store.py
```diff
@@ -0,0 +1,140 @@
...
+        created_timestamp_column: Optional[str],
+        start_date: datetime,
+        end_date: datetime,
+    ) -> Optional[pyarrow.Table]:
+        pass
+
+
+class BigQueryOfflineStore(OfflineStore):
+    """
+    BigQueryOfflineStore is a non-user-facing object used for all interaction between Feast and BigQuery.
```
woop, line 44: I think we can maybe just leave out this comment and inherit from OfflineStore for now.

sdk/python/feast/offline_store.py
```diff
@@ -0,0 +1,140 @@
...
+    @staticmethod
+    def _pull_query(query: str) -> pyarrow.Table:
+        from google.cloud import bigquery
+
+        client = bigquery.Client()
+        query_job = client.query(query)
+        return query_job.to_arrow()
+
+
+def run_reverse_field_mapping(
```
woop, line 88: This method feels a bit magical. Would you mind adding a docstring here with a description of the inputs and outputs and what they represent?

woop: I also don't see where `run_reverse_field_mapping` is being used. Maybe I'm blind :)

sdk/python/feast/offline_store.py
```diff
@@ -0,0 +1,140 @@
...
+        ]
+    return (
+        entity_names,
+        feature_names,
+        event_timestamp_column,
+        created_timestamp_column,
+    )
+
+
+def run_forward_field_mapping(
```
woop, line 127: What about `forward_map_columns`? (map as verb)

sdk/python/feast/offline_store.py
jklegar: I think non-user-facing is important since these docstrings get turned into our API documentation - we don't want users to use this class to trigger ingestion or historical retrieval

sdk/python/feast/offline_store.py
jklegar: Added some description - it'll be used in materialize() to calculate arguments for pull_latest_from_table but that function's not written yet, so it's just been tested manually

sdk/python/feast/offline_store.py
woop: But wouldn't that apply to the majority of our code?

sdk/python/feast/offline_store.py
jklegar: I'd like to keep it like this to match the other function

sdk/python/feast/offline_store.py
jklegar: I'm not sure why we have API documentation for our internal classes tbh

sdk/python/feast/offline_store.py
woop: Ah ok. I would have expected this method to be called from `BigQueryStore.pull_latest_from_table()`, which simplifies the OfflineStore contract a bit imho

sdk/python/feast/offline_store.py
woop: In this specific case we expect actual developers to work with these classes. Documenting it is a very small amount of effort for us and makes it a lot more pleasant to navigate and read a code base.

woop (approved):

PR #1367, opened 2021-03-05
Add Firestore online format specification

oavdeev:
We'll be adding support for Google Cloud Firestore as an online store to Feast. This document is the part of our API, a specification on how exactly we'll store feature values in the Firestore.

```release-note
Added Firestore online store format specification
```


docs/specs/online_store_format.md
```diff
@@ -59,6 +59,34 @@ Here's an example of how the entire thing looks like:
...
+We use the following structure to store feature data in the Firestore:
+* at the first level, there is a collection for each Feast project
+* second level, in each project-collection, there is a Firebase document for each Feature Table
+* third level, in the document for the Feature Table, there is a subcollection called `values` that contain a document per feature row. That document contains the following fields:
+  * `key` contains entity key as serialized `feast.types.EntityKey` proto
+  * `value` contains value as serialized `feast.types.Value` proto
+  * `event_ts` contains event timestamp (in the native firestore timestamp format)
+  * `created_ts` contains write timestamp (in the native firestore timestamp format)
+
+Document id for the feature document is computed by hashing entity key using murmurhash3_128 algorithm as follows:
```
woop, line 75: How is the Document Id used?

woop (approved):

docs/specs/online_store_format.md
oavdeev: Document ID is basically the primary key, so on read path you'd compute it and use to retrieve the doc containing feature values.

Another thing I've considered here is to use key field in the doc itself, since Firestore automatically creates indices it would also be pretty efficient. However, in that case it would be theoretically possible to have two docs containing the same entity key, and I wanted to avoid thinking through an additional failure mode.

woop: Makes sense.

woop (approved):

PR #1376, opened 2021-03-09
Read and write path for Datastore and SQLite

oavdeev:
Implement read and write path for online store, both for local (SQLite) mode and GCP (Datastore).

It only implements a lower-level interface for reading and writing to be used internally in Feast, I expect we'll later have nicer wrappers over those methods in the SDK.

Fixes #

```release-note
NONE
```


sdk/python/feast/infra/provider.py
```diff
@@ -0,0 +1,49 @@
...
+
+from feast import FeatureTable
+from feast.repo_config import RepoConfig
+
+
+class Provider(abc.ABC):
+    @abc.abstractmethod
+    def update_infra(
+        self,
+        project: str,
```
woop, line 12: @jklegar @oavdeev @jparthasarthy can we maybe make a decision on something. Is `project` === `repo`? If not, what is the relationship.

oavdeev: Roughly yes, the idea is that you can set project field to something in the repo config. That way all the infra created by `feast deploy` will be namespaced† by the project name. So you'd be able, for example, to `feast deploy` multiple independent feature repos to the same AWS account and they won't interfere with each other.

† exact meaning of "namespaced" is provider-dependent, could be a resource name prefix or some kind of tag

woop: I agree with this approach, but does this imply that we should remove projects from our API itself? Because we would never by able to have an operation like `list_projects()` that returns more than one project right?

oavdeev: I think so, yes

docs/specs/online_store_format.md
```diff
@@ -59,20 +59,20 @@ Here's an example of how the entire thing looks like:
...
-We use the following structure to store feature data in the Firestore:
-* at the first level, there is a collection for each Feast project
-* second level, in each project-collection, there is a Firebase document for each Feature Table
-* third level, in the document for the Feature Table, there is a subcollection called `values` that contain a document per feature row. That document contains the following fields:
-  * `key` contains entity key as serialized `feast.types.EntityKey` proto
-  * `values` contains feature name to value map, values serialized as `feast.types.Value` proto
-  * `event_ts` contains event timestamp (in the native firestore timestamp format)
-  * `created_ts` contains write timestamp (in the native firestore timestamp format)
+We use the following structure to store feature data in Datastore:
+* there is a Datastore Entity for each Feast Project, with Kind `Project`
```
woop, line 67: Can we capitalize the first letters here?

docs/specs/online_store_format.md
```diff
@@ -59,20 +59,20 @@ Here's an example of how the entire thing looks like:
...
+* under that Datastore Entity, there is a Datastore Entity for each Feast Feature Table or View, with Kind `Table`. That contains one additional field, `created_ts` that contains the timestamp when this Datastore Entity was created.
+* under that Datastore Entity, there is a Datastore Entity for each Feast Entity Key with Kind `Row`. That contains the following fields:
+	 * `key` contains entity key as serialized `feast.types.EntityKey` proto
+	 * `values` contains feature name to value map, values serialized as `feast.types.Value` proto
+   * `event_ts` contains event timestamp (in the datastore timestamp format)
```
woop, line 72: Do we know the ts granularity? Same as the proto?

sdk/python/feast/feature_store.py
```diff
@@ -11,24 +11,47 @@
...
+            raise Exception("You cannot specify both repo_path and config")
         if config is not None:
             self.config = config
-        elif config_path is not None:
-            self.config = load_repo_config(config_path)
+        elif repo_path is not None:
+            self.config = load_repo_config(Path(repo_path))
         else:
-            self.config = RepoConfig()
+            self.config = RepoConfig(
```
woop, line 44: Is this the only location when the local defaults are set? Does the CLI also follow this code path?

sdk/python/feast/infra/gcp.py
```diff
@@ -21,19 +28,37 @@ def _delete_all_values(client, key) -> None:
...
+
+
 class Gcp(Provider):
-    _project_id: Optional[str]
+    _gcp_project_id: Optional[str]
 
     def __init__(self, config: Optional[DatastoreOnlineStoreConfig]):
         if config:
-            self._project_id = config.project_id
+            self._gcp_project_id = config.project_id
```
woop, line 54: Wait, I dont think we should be calling this `project_id` if it references a `gcp project id`? Can we differentiate those so that its less confusing?

woop: Also, do we infer the GCP project when the user doesn't set it? We dont want them to have to go and manually set it if we can get it from the env.

sdk/python/feast/infra/local_sqlite.py
```diff
@@ -17,20 +21,91 @@ class LocalSqlite(Provider):
...
         project: str,
         tables_to_delete: List[FeatureTable],
         tables_to_keep: List[FeatureTable],
     ):
-        conn = sqlite3.connect(self._db_path)
+        conn = self._get_conn()
         for table in tables_to_keep:
             conn.execute(
-                f"CREATE TABLE IF NOT EXISTS {_table_id(project, table)} (key BLOB, value BLOB)"
+                f"CREATE TABLE IF NOT EXISTS {_table_id(project, table)} (entity_key BLOB, feature_name TEXT, value BLOB, event_ts timestamp, created_ts timestamp,  PRIMARY KEY(entity_key, feature_name))"
```
woop, line 38: Do we have an update/migrate?

sdk/python/feast/infra/gcp.py
oavdeev: Yea it can be omitted. I can call it `gcp_project_id` too but it goes under `datastore` section in the config so one could argue it is clear that it pertains to GCP

docs/specs/online_store_format.md
oavdeev: No it looks like it is microseconds while proto is technically nanoseconds. And native python datetimes are microseconds too. Not sure if it needs to be duplicated in the spec though, as it is already in the datastore docs

sdk/python/feast/feature_store.py
oavdeev: This and repo struct definition. This is config stuff that'll need later polish

sdk/python/feast/infra/gcp.py
woop: wait, but why do we need to have the project defined there in any case? shouldnt we only allow one project to be used globally?

oavdeev: I'm not sure how gcloud auth works exactly, but wouldn't it be useful to be able to choose a specific GCP project here?

woop: I'm not opposed to it, I just want to keep things as simple as possible. We already have config hell with Spark. I dont feel too strongly here though.

woop (approved):

PR #1379, opened 2021-03-11
Implement materialize method

jklegar:
**What this PR does / why we need it**: Implements ingestion into Firestore, which will be part of the python-based ingestion flow

Fixes #

```release-note
Adds materialize method to FeatureStore class to load data from the offline store into the online store for a specified time interval.
```


jklegar:
Note this is stacked on #1366

sdk/python/feast/feature_store.py
```diff
@@ -58,6 +68,9 @@ def __init__(
     def _get_provider(self) -> Provider:
         return get_provider(self.config)
 
+    def _get_offline_store(self) -> Type[OfflineStore]:
```
woop, line 71: Do we need a method? Any reason we can't use the function directly?

sdk/python/feast/feature_store.py
```diff
@@ -153,6 +166,57 @@ def get_historical_features(
         )
         return job
 
+    @property
+    def project(self) -> str:
```
woop, line 170: Would this be better as a config option? `config.get_project()`?

sdk/python/feast/feature_store.py
```diff
@@ -153,6 +166,57 @@ def get_historical_features(
...
+        return "default"
+
+    def materialize(
```
woop, line 173: Adding a note here to add a docstring.

sdk/python/feast/feature_store.py
```diff
@@ -153,6 +166,57 @@ def get_historical_features(
...
+        self, feature_views: List[str], start_date: datetime, end_date: datetime
```
woop, line 174: `feature_views` should be optional I believe.

sdk/python/feast/feature_store.py
```diff
@@ -153,6 +166,57 @@ def get_historical_features(
...
+    ):
+        full_feature_views = []
```
woop, line 176: Nitpick on the naming: What does `full` represent here?

sdk/python/feast/infra/gcp.py
```diff
@@ -38,7 +38,9 @@ def compute_datastore_entity_id(entity_key: EntityKeyProto) -> str:
     return mmh3.hash_bytes(serialize_entity_key(entity_key)).hex()
 
 
-def _make_tzaware(t: datetime):
+def _make_tzaware(t: Optional[datetime]):
+    if t is None:
+        return t
```
oavdeev, line 43: hmm that seems not great from readability perspective, can be explicitly handle `None` in the caller?

In fact, after looking closer, it looks like there is already no chance of this being called with None

sdk/python/feast/feature_store.py
```diff
@@ -153,6 +166,57 @@ def get_historical_features(
...
+        registry = self._get_registry()
+        for name in feature_views:
+            feature_view = registry.get_feature_view(name, self.project)
```
woop, line 179: Would `list_feature_views()` be a bit faster?

sdk/python/feast/feature_store.py
```diff
@@ -153,6 +166,57 @@ def get_historical_features(
...
+            full_feature_views.append(feature_view)
+
+        # TODO paging large loads
+        for feature_view in full_feature_views:
+            if isinstance(feature_view.input, FileSource):
+                raise NotImplementedError(
+                    "This function is not yet implemented for File data sources"
+                )
+            if feature_view.input.table_ref is None:
```
woop, line 188:
```suggestion
            if not feature_view.input.table_ref:
```

sdk/python/feast/feature_store.py
```diff
@@ -176,3 +240,113 @@ def _get_requested_feature_views(
...
+    feature_view: FeatureView,
+) -> Tuple[List[str], List[str], str, Optional[str]]:
+    """
+    If a field mapping exists, run it in reverse on the entity names, feature names, event timestamp column, and created timestamp column to get the names of the relevant columns in the BigQuery table.
+
+    Args:
+        feature_view: FeatureView object containing the field mapping as well as the names to reverse-map.
+    Returns:
+        Tuple containing the list of reverse-mapped entity names, reverse-mapped feature names, reverse-mapped event timestamp column, and reverse-mapped created timestamp column that will be passed into the query to the offline store.
+    """
```
oavdeev, line 255: formatter nit: not sure if we have a linter rule for line length but let's make it at least under 120 chars per line

sdk/python/feast/offline_store.py
```diff
@@ -87,8 +87,14 @@ class OfflineStore(ABC):
     @staticmethod
     @abstractmethod
     def pull_latest_from_table(
-        feature_view: FeatureView, start_date: datetime, end_date: datetime,
-    ) -> Optional[pyarrow.Table]:
+        table_ref: str,
```
woop, line 90: Is there a reason why we have added all of the fields to the `pull_latest_from_table()` method instead of using `feature_view` which already contains all of the information?

sdk/python/feast/feature_store.py
```diff
@@ -176,3 +240,113 @@ def _get_requested_feature_views(
...
+            v: k for k, v in feature_view.input.field_mapping.items()
+        }
+        event_timestamp_column = (
+            reverse_field_mapping[event_timestamp_column]
+            if event_timestamp_column in reverse_field_mapping.keys()
+            else event_timestamp_column
+        )
+        created_timestamp_column = (
+            reverse_field_mapping[created_timestamp_column]
+            if created_timestamp_column is not None
```
woop, line 272: It's possible for `created_timestamp_column` to be `""`. We should us `not created_timestamp_column`

sdk/python/feast/feature_store.py
```diff
@@ -176,3 +240,113 @@ def _get_requested_feature_views(
...
+        else:
+            created_timestamp = None
+
+        rows_to_write.append(
+            (entity_key, feature_dict, event_timestamp, created_timestamp)
+        )
+    return rows_to_write
+
+
+def _convert_to_proto(value: Any) -> ValueProto:
```
woop, line 338: I believe we're still missing types here. Also, we should move this method to `value_types.py` and give it a descriptive name. I would also confirm whether we dont already have a type map for this mapping.

sdk/python/tests/test_bigquery_ingestion.py
```diff
@@ -0,0 +1,76 @@
...
+    df = pd.DataFrame.from_dict(data)
+
+    # load dataset into BigQuery
+    client = bigquery.Client()
+    job_config = bigquery.LoadJobConfig()
+    gcp_project = client.project
+    bigquery_dataset = "test_ingestion"
+    dataset = bigquery.Dataset(f"{gcp_project}.{bigquery_dataset}")
+    client.create_dataset(dataset, exists_ok=True)
+    table_id = f"{gcp_project}.{bigquery_dataset}.table_{int(time.time())}"
```
woop, line 37: Are you cleaning up these tables?

sdk/python/tests/test_bigquery_ingestion.py
```diff
@@ -0,0 +1,76 @@
...
+from feast.feature_store import FeatureStore
+from feast.feature_view import FeatureView
+from feast.protos.feast.types.EntityKey_pb2 import EntityKey as EntityKeyProto
+from feast.protos.feast.types.Value_pb2 import Value as ValueProto
+from feast.repo_config import LocalOnlineStoreConfig, OnlineStoreConfig, RepoConfig
+from feast.value_type import ValueType
+
+
+@pytest.mark.integration
+def test_bigquery_ingestion():
```
woop, line 19: We should probably also have a more representative ingestion, something like 10M rows.

sdk/python/tests/test_bigquery_ingestion.py
```diff
@@ -0,0 +1,76 @@
...
+    fv = FeatureView(
+        "test_fv",
+        ["id_mapped"],
+        [Feature("value", ValueType.FLOAT)],
+        timedelta(minutes=5),
+        BigQuerySource(
+            "timestamp_mapped",
+            table_id,
+            "created_ts",
+            {"ts": "timestamp_mapped", "id": "id_mapped"},
```
woop, line 51: It's not clear to me how or why `timestamp_mapped` occurs twice. Is that supposed to work?

sdk/python/tests/test_bigquery_ingestion.py
```diff
@@ -0,0 +1,76 @@
...
+    table_id = f"{gcp_project}.{bigquery_dataset}.table_{int(time.time())}"
+    job = client.load_table_from_dataframe(df, table_id, job_config=job_config)
+    job.result()
+
+    # create FeatureView
+    fv = FeatureView(
+        "test_fv",
+        ["id_mapped"],
+        [Feature("value", ValueType.FLOAT)],
+        timedelta(minutes=5),
```
woop, line 46: For clarity we should probably use key word arguments as well.

sdk/python/feast/offline_store.py
```diff
@@ -612,3 +558,10 @@ def _get_requested_feature_views_to_features_dict(
 {% endfor %}
 ORDER BY event_timestamp
 """
+
+
+def get_offline_store(config: RepoConfig) -> Type[OfflineStore]:
+    if config.provider == "gcp":
```
woop, line 564: What should the user select if they want to pull data from BigQuery and write to a local store?

sdk/python/tests/test_bigquery_ingestion.py
```diff
@@ -0,0 +1,76 @@
...
+            "created_ts",
+            {"ts": "timestamp_mapped", "id": "id_mapped"},
+            "",
+        ),
+    )
+    config = RepoConfig(
+        metadata_store="./metadata.db",
+        project="default",
+        provider="gcp",
+        online_store=OnlineStoreConfig(local=LocalOnlineStoreConfig("online_store.db")),
```
woop, line 59: I think this is an example of the confusion I am experiencing around providers (that isn't really a result of this PR, but it's emerging here). I dont think the provider model holds up if we have sources (BigQuery, Parquet) and sinks (Sqlite, Firestore) that can change independently. We can easily support any pairwise combination there, but we only have 2 options for the provider at the moment (gcp, local)

sdk/python/feast/feature_store.py
jklegar: we need a method to specify which implementation of OfflineStore we're using

sdk/python/feast/feature_store.py
jklegar: This is a list of the FeatureView objects instead of just the names, and `feature_views` was already taken; I can change if you have a suggestion

sdk/python/feast/feature_store.py
jklegar: maybe, but we'd have to repeat the logic of returning an error if one of the feature views doesn't exist; I'd prefer to do this and optimize later

sdk/python/feast/offline_store.py
jklegar: My goal is for this function to do just the backend-specific logic, so I don't want to run the field mappings in this function since anyone building a new connector would need to know they have to call the field mapping methods in their implementation of this method. But you're right that table_ref is too tightly coupled to BQ, so I replaced it with the data source, which I think should work

sdk/python/tests/test_bigquery_ingestion.py
jklegar: added a ttl of 14 days on the dataset (which is applied to any table created in that dataset)

sdk/python/tests/test_bigquery_ingestion.py
jklegar: agreed. Haven't tried it but I imagine this won't go over well without paging so I'm going to defer this

woop: Can we perhaps just add the test and optimize it later? I believe @oavdeev will be looking at optimizing it a bit in the next week.

sdk/python/feast/feature_store.py
woop: I mean can we just call `get_offline_store(self.config)` and remove some of these private methods?

sdk/python/feast/feature_store.py
```diff
@@ -153,6 +167,74 @@ def get_historical_features(
...
+    def materialize(
+        self,
+        feature_views: Optional[List[str]],
+        start_date: datetime,
+        end_date: datetime,
+    ) -> None:
+        """
+        Materialize data from the offline store into the online store.
+
+        This method materializes feature data in the specified interval from either
```
woop, line 179: Let's try to use `loads` here maybe so that the user has a different verb instead of repeating `materialize` which they may not understand.

sdk/python/feast/feature_store.py
```diff
@@ -153,6 +167,74 @@ def get_historical_features(
...
+        the specified feature views or all feature views if none are specified
```
woop, line 180: Missing comma :)

sdk/python/feast/feature_store.py
woop: `feature_views_to_materialize`?

sdk/python/tests/test_bigquery_ingestion.py
```diff
@@ -16,65 +17,75 @@
...
-    # create dataset
-    ts = pd.Timestamp.now(tz="UTC").round("ms")
-    data = {
-        "id": [1, 2, 1],
-        "value": [0.1, 0.2, 0.3],
-        "ts_1": [ts - timedelta(minutes=2), ts, ts],
-        "created_ts": [ts, ts, ts],
-    }
-    df = pd.DataFrame.from_dict(data)
+class TestBigQueryIngestion:
```
woop, line 20: Just a naming nitpick, but should this be called something like `TestMaterializeFromBigQueryToLocal`? `TestBigQueryIngestion` makes it seem like we're testing a ingestion API for batch data into BQ.

sdk/python/feast/feature_store.py
```diff
@@ -153,6 +167,76 @@ def get_historical_features(
...
+
+        This method loads feature data in the specified interval from either
+        the specified feature views, or all feature views if none are specified,
+        into the online store where it is available for online serving.
+
+        Args:
+            feature_views (List[str]): Optional list of feature view names. If selected, will only run
+                materialization for the specified feature views.
+            start_date (datetime): Start date of query
+            end_date (datetime): End date of query
```
woop, line 187: Can we please add an example here?

sdk/python/feast/feature_store.py
```diff
@@ -153,6 +167,76 @@ def get_historical_features(
...
+        Materialize data from the offline store into the online store.
+
+        This method loads feature data in the specified interval from either
+        the specified feature views, or all feature views if none are specified,
+        into the online store where it is available for online serving.
+
+        Args:
+            feature_views (List[str]): Optional list of feature view names. If selected, will only run
+                materialization for the specified feature views.
+            start_date (datetime): Start date of query
```
woop, line 186: Nitpick: `Start date of query` isn't really clear. How about `Start date for time range of data to materialize into the online store`

sdk/python/feast/feature_store.py
```diff
@@ -153,6 +167,76 @@ def get_historical_features(
...
+                event_timestamp_column,
+                created_timestamp_column,
+            ) = _run_reverse_field_mapping(feature_view)
+
+            offline_store = self._get_offline_store()
+            table = offline_store.pull_latest_from_table(
+                feature_view.input,
+                entity_names,
+                feature_names,
+                event_timestamp_column,
```
woop, line 222: Is there still a reason to have the columns as part of this method? (`event_timestamp_column`, `created_timestamp_column`)

jklegar: yes, these are the post-feature-mapping names, and I want the feature-mapping to happen outside the BigQuery-specific method which means these names have to be passed in

sdk/python/feast/offline_store.py
```diff
@@ -87,8 +87,14 @@ class OfflineStore(ABC):
...
-        feature_view: FeatureView, start_date: datetime, end_date: datetime,
-    ) -> Optional[pyarrow.Table]:
+        data_source: DataSource,
+        entity_names: List[str],
+        feature_names: List[str],
+        event_timestamp_column: str,
+        created_timestamp_column: Optional[str],
+        start_date: datetime,
+        end_date: datetime,
+    ) -> pyarrow.Table:
```
woop, line 97: I think we need comments here to explain the fields. It's a bit confusing because `feature_names` are actually column names as they are found in the remote table, and that isnt immediately obvious.

woop (approved):

PR #1380, opened 2021-03-11
Allow Feast Client to use REST as transport

randxie:
Send out the PR for discussion purpose. This PR introduces two classes CoreServiceRESTStub and ServingServiceRESTStub to allow the client using REST as transport method. It uses decorator to convert gRPC calls into REST calls, then convert the response back to protobuf messages. Here's an example usage of this feature:

```
import feast
feast_client = feast.Client(core_url="http://localhost:8000", serving_url="http://localhost:8000", transport_name="rest")
feast_client.list_projects()
```

Allow client to use REST as transport.

None

For more information about release notes, see kubernetes' guide here:
http://git.k8s.io/community/contributors/guide/release-notes.md
-->
```release-note
Allow client to use REST as transport method.
```

### DIFF
--- a/sdk/python/feast/client.py
+++ b/sdk/python/feast/client.py
@@ -144,25 +144,26 @@ class Client:
 
         Returns: CoreServiceStub
         """
-        if not self._core_service_stub:
-            if self._transport_name == TransportName.grpc:
-                channel = create_grpc_channel(
-                    url=self._config.get(opt.CORE_URL),
-                    enable_ssl=self._config.getboolean(opt.CORE_ENABLE_SSL),
-                    enable_auth=self._config.getboolean(opt.ENABLE_AUTH),
-                    ssl_server_cert_path=self._config.get(opt.CORE_SERVER_SSL_CERT),
-                    auth_metadata_plugin=self._auth_metadata,
-                    timeout=self._config.getint(opt.GRPC_CONNECTION_TIMEOUT),
-                )
-                self._core_service_stub = CoreServiceStub(channel)
-            elif self._transport_name == TransportName.rest:
-                self._core_service_stub = CoreServiceRESTStub(
-                    self._config.get(opt.CORE_URL)
-                )
-            else:
-                raise ValueError(
-                    "Unknown transport name: {}".format(self._transport_name)
-                )
+        if self._core_service_stub:
+            return self._core_service_stub
+
+        if self._transport_name == TransportName.grpc:
+            channel = create_grpc_channel(
+                url=self._config.get(opt.CORE_URL),
+                enable_ssl=self._config.getboolean(opt.CORE_ENABLE_SSL),
+                enable_auth=self._config.getboolean(opt.ENABLE_AUTH),
+                ssl_server_cert_path=self._config.get(opt.CORE_SERVER_SSL_CERT),
+                auth_metadata_plugin=self._auth_metadata,
+                timeout=self._config.getint(opt.GRPC_CONNECTION_TIMEOUT),
+            )
+            self._core_service_stub = CoreServiceStub(channel)
+        elif self._transport_name == TransportName.rest:
+            self._core_service_stub = CoreServiceRESTStub(
+                self._config.get(opt.CORE_URL)
+            )
+        else:
+            raise ValueError("Unknown transport name: {}".format(self._transport_name))
+
         return self._core_service_stub
 
     @property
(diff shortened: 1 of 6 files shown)


randxie:
cc: @chen-hung-wu

woop:
@randxie Thanks for the contribution!

I'll have a look at the contribution in a bit, but my initial reaction is that this code may have to live in [Feast Spark](https://github.com/feast-dev/feast-spark/). 

/ok-to-test

sdk/python/feast/client.py
```diff
@@ -135,15 +146,24 @@ def _core_service(self):
         Returns: CoreServiceStub
         """
         if not self._core_service_stub:
```
mazinbokhari, line 148: Nit - unnest if statements with
```
if self._core_service_stub:
    return
```

woop:
@randxie At a high level the change looks good, but I am trying to figure out the bigger picture. How do you imagine users will eventually benefit from this change, and what are the other changes that you'd want to introduce alongside this PR?

randxie:
@woop I learnt that Feast has plans to add REST supports. If that's available, users will be able to use/test the REST APIs with this REST transport feature.

I will have follow-up PRs & design docs if Feast team feels adding REST support can be added to the roadmap. I will be happy to contribute to the REST effort (I believe REST APIs also work the best for UI, if UI is also on the roadmap).

woop:
> @woop I learnt that Feast has plans to add REST supports. If that's available, users will be able to use/test the REST APIs with this REST transport feature.
> 
> I will have follow-up PRs & design docs if Feast team feels adding REST support can be added to the roadmap. I will be happy to contribute to the REST effort (I believe REST APIs also work the best for UI, if UI is also on the roadmap).

Great, thanks for clarifying @randxie!

The current state of REST is as follows.
- We have REST controllers on or Core and Serving API. [Example](https://github.com/feast-dev/feast-java/blob/master/core/src/main/java/feast/core/controller/CoreServiceRestController.java)
- They are basically experimental and undocumented right now, but they should still be working. 
- The way in which most teams are enabling REST right now is through [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway). That works fine for Web UIs, and in fact the Gojek team (like @terryyylim @khorshuheng) have already built a Feast UI on GRPC gateway.
- I like the idea of shipping a REST client as part of the Client class, but it's not 100% clear to me what the use case is. I can understand a team wanting to retrieve feature values from a REST API or building a UI on top of our Core API, but if the Client already supports gRPC, why woulndn't they just use that? 
- Our move towards REST long term will start with our Feast 0.10 API. So teams will begin with an object store registry (which doesn't include a REST API), then we will add a REST API for online serving, and then we will consider adding more methods to the REST server over time. It's not clear if we want to expose all functionality from day one.

randxie:
@woop Totally agree with the statement that gRPC client is sufficient for 99.9% use cases and REST client is a bit redundant. It might be a specific requirement for us that gRPC is not well supported and we have to use REST. And therefore, we will need to build REST support in a private fork and it's best for us to contribute back to Feast appropriately.

We also use gRPC-gateway to convert gRPC service into REST service. And I think that's cleaner than the REST controllers. Given that REST controllers are experimental and undocumented, do you think we can make the grpc-gateway path officially supported? Happy to discuss more in the community calls.

Also, is there any plan for open-sourcing the Feast UI?

woop:
> @woop Totally agree with the statement that gRPC client is sufficient for 99.9% use cases and REST client is a bit redundant. It might be a specific requirement for us that gRPC is not well supported and we have to use REST. And therefore, we will need to build REST support in a private fork and it's best for us to contribute back to Feast appropriately.

In this case you are specifically referring to the Client right? That should be pretty easy to maintain internally.

> We also use gRPC-gateway to convert gRPC service into REST service. And I think that's cleaner than the REST controllers. Given that REST controllers are experimental and undocumented, do you think we can make the grpc-gateway path officially supported? Happy to discuss more in the community calls.

Yea I am happy with that. We don't have an agenda for the call today, but you can add it here if you'd like https://docs.google.com/document/d/1MqNNhPCi1vvcYv71j3N0uBb4qIbl2ikaFO8tKohu12I/edit#heading=h.6js8vamr57ls

> Also, is there any plan for open-sourcing the Feast UI?

Yes, but the timeline is totally up to the Gojek contributors. We're not driving that process.

randxie:
@woop This client change is not too bad to maintain. The code change in Feast-java can be a bit tricky though. Let's discuss the REST support in the community call tonigh. For this PR, I will close it for now.

woop:
> t change is not too bad to maintain

Sounds good!

PR #1386, opened 2021-03-12
Add feature views

oavdeev:
Introduces FeatureView object as planned for .10

```release-note
Introduce FeatureView object
```


sdk/python/feast/registry.py
```diff
@@ -117,7 +118,7 @@ def get_entity(self, name: str, project: str) -> Entity:
 
     def apply_feature_table(self, feature_table: FeatureTable, project: str):
         """
-        Registers a single feature table with Feast
+        Registers a single feature table or feature view with Feast
```
woop, line 121: Is this comment change a mistake?

oavdeev: oops yes, originally i tried to do one apply method for both but then changed my mind

woop: For a moment there I thought you were pulling some series magic shoehorning feature views into feature tables :)

protos/feast/core/FeatureView.proto
```diff
@@ -0,0 +1,70 @@
...
+import "feast/core/DataSource.proto";
+import "feast/core/Feature.proto";
+
+message FeatureView {
+    // User-specified specifications of this feature table.
+    FeatureViewSpec spec = 1;
+
+    // System-populated metadata for this feature table.
+    FeatureViewMeta meta = 2;
+    string project = 3;
```
woop, line 36: One thing that stands out here. Why is the project not defined as part of the spec?

woop: I think we should move the `project` into the spec itself. I dont think its a breaking change since the `project` id was only recently introduced on master. @oavdeev what do you think? (it can be a separate PR).

oavdeev: yeah I just copied FeatureTable, not opposed to moving it, but we'd have to move it in both places then

woop: Ok, follow up it is then.

woop (approved):

PR #1389, opened 2021-03-14
Add historical retrieval for BigQuery and Parquet

woop:

* Adds support for file (Parquet) based historical retrieval
* Adds support for BigQuery based historical retrieval

Limitations
* The original SQL query to generate a training dataset was executed as a multi-job query that produced tables. This query does not produce tables, but it will probably run out of memory at some point. It needs to be optimized eventually.
* We are not applying time range filters prior to doing joins. 
* Async (RetrievalJob) behaviour isn't consistent across stores (Parquet, BigQuery). Sometimes work only starts when a user runs (`to_df()`), other times it starts when `get_historical_features()` is run. 
* `apply()` has been added but doesnt yet provision infrastructure. Considering that functionality out of scope for this PR.

```release-note
Added historical retrieval for file and BigQuery sources
```


sdk/python/feast/offline_store.py
```diff
@@ -11,18 +11,76 @@
...
+        pass
+
+    @abstractmethod
+    def to_df(self):
+        """Return dataset as Pandas DataFrame synchronously"""
+        pass
+
+
+class FileRetrievalJob(RetrievalJob):
+    def __init__(self, evaluation_function):
```
oavdeev, line 45: I'd add a type signature for `evaluation_function` here, and a comment explaining what it is for.

sdk/python/feast/offline_store.py
```diff
@@ -11,18 +11,76 @@
...
+from feast.repo_config import RepoConfig
+
+ENTITY_DF_EVENT_TIMESTAMP_COL = "event_timestamp"
+
+
+class RetrievalJob(ABC):
+    """RetrievalJob is used to manage the execution of a historical feature retrieval"""
+
+    def __init__(self):
+        pass
```
oavdeev, line 36: does it need a constructor?

sdk/python/feast/offline_store.py
```diff
@@ -11,18 +11,76 @@
...
+        self.query = query
+
+    def to_df(self):
+        # TODO: Ideally only start this job when the user runs "get_historical_features", not when they run to_df()
+        client = bigquery.Client()
+        df = client.query(self.query).to_dataframe(create_bqstorage_client=True)
+        return df
+
+
+@dataclass
```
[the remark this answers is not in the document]
oavdeev: doesn't look like anything in this context object is supposed to be mutable

sdk/python/feast/offline_store.py
```diff
@@ -82,6 +151,237 @@ def _pull_query(query: str) -> pyarrow.Table:
...
+        else:
+            raise ValueError(
+                f"The entity dataframe you have provided must be a Pandas DataFrame or BigQuery SQL query, "
+                f"but we found: {type(entity_df)} "
+            )
+
+        # Build a query context containing all information required to template the BigQuery SQL query
+        query_context = get_feature_view_query_context(feature_refs, feature_views)
+
+        # TODO: Get timestamps from entity_df
```
oavdeev, line 177:
```suggestion
        # TODO: Infer min_timestamp and max_timestamp from entity_df
```

oavdeev (approved):

sdk/python/feast/offline_store.py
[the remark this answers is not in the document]
woop: Ok, so what change would you recommend?

PR #1392, opened 2021-03-16
Add nicer validation for repo config

oavdeev:
This adds nicer error messages for the config reader. This PR also adds an additional dep on `jsonschema` package. This is already a popular dep (for example, for `jupyter` and `great-expectations`) so that is arguably not too bad.

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/repo_config.py
+++ b/sdk/python/feast/repo_config.py
@@ -7,23 +7,34 @@ from jsonschema import ValidationError, validate
 
 
 class LocalOnlineStoreConfig(NamedTuple):
+    """ Online store config for local (SQLite-based) online store """
     path: str
+    """ str: Path to sqlite db """
 
 
 class DatastoreOnlineStoreConfig(NamedTuple):
+    """ Online store config for GCP Datastore """
     project_id: str
+    """ str: GCP Project Id """
 
 
 class OnlineStoreConfig(NamedTuple):
     datastore: Optional[DatastoreOnlineStoreConfig] = None
+    """ DatastoreOnlineStoreConfig: Optional DatastoreConfig """
     local: Optional[LocalOnlineStoreConfig] = None
+    """ LocalOnlineStoreConfig: Optional local online store config """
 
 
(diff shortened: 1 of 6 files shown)


sdk/python/feast/repo_config.py
```diff
@@ -25,7 +26,64 @@ class RepoConfig(NamedTuple):
...
+# for config validation, something that bindr unfortunately doesn't provide out of the box.
+#
+# The schema should match the namedtuple structure above. It could technically even be inferred from
+# the types above automatically; but for now we choose a more tedious but less magic path of
+# providing the schema manually.
+
+config_schema = {
+    "type": "object",
+    "properties": {
+        "project": {"type": "string"},
```
woop, line 39: This json schema is arguably going to be quite important for people to understand how they can configure Feast. Can we maybe 
1. Add descriptions to the fields?
2. Move it out to a top level file (`config_schema.json`?)

sdk/python/feast/repo_config.py
```diff
@@ -7,23 +7,34 @@
...
     datastore: Optional[DatastoreOnlineStoreConfig] = None
+    """ DatastoreOnlineStoreConfig: Optional DatastoreConfig """
     local: Optional[LocalOnlineStoreConfig] = None
+    """ LocalOnlineStoreConfig: Optional local online store config """
 
 
 class RepoConfig(NamedTuple):
+    """ Repo config. Typically loaded from `feature_store.yaml` """
     metadata_store: str
+    """ str: Path to metadata store. Can be a local path, or remote object storage path """
```
woop, line 34: I think we have to have examples here.

sdk/python/feast/repo_config.py
```diff
@@ -7,23 +7,34 @@
...
     project: str
+    """ str: Feast project id. This can be any alphanumeric string up to 16 characters. """
```
woop, line 36: I think it's important to add how this string is used.

woop (approved):

PR #1396, opened 2021-03-18
Fix a bug in client archive_project method and fix lint in grpc auth

randxie:

Fix bug in client.archive_project. It should use the core_service method instead of raw core_service_stub. Also fix lint in grpc/auth.py.

NONE

```release-note
Fix bug in client.archive_project and fix lint in grpc auth
```

### DIFF
--- a/sdk/python/feast/client.py
+++ b/sdk/python/feast/client.py
@@ -451,7 +451,7 @@ class Client:
             )
         else:
             try:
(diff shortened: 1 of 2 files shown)


woop:
Thanks @randxie, remember to sign your commits please :)

woop (approved):

PR #1399, opened 2021-03-20
Add online read API to FeatureStore class

oavdeev:
Added python-online read API to the FeatureStore class. It closely mimicks the existing API on feast client.

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -11,9 +11,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from collections import defaultdict
 from datetime import datetime
 from pathlib import Path
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import pandas as pd
 import pyarrow
@@ -26,6 +27,11 @@ from feast.offline_store import (
     RetrievalJob,
     get_offline_store,
     get_offline_store_for_retrieval,
+)
+from feast.online_response import OnlineResponse, _infer_online_entity_rows
+from feast.protos.feast.serving.ServingService_pb2 import (
+    GetOnlineFeaturesRequestV2,
+    GetOnlineFeaturesResponse,
 )
 from feast.protos.feast.types.EntityKey_pb2 import EntityKey as EntityKeyProto
 from feast.protos.feast.types.Value_pb2 import Value as ValueProto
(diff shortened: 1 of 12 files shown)


sdk/python/feast/feature_store.py
```diff
@@ -153,26 +160,131 @@ def get_historical_features(
...
+    return EntityKeyProto(entity_names=names, entity_values=values)
+
+
+def _entity_row_to_field_values(
+    row: GetOnlineFeaturesRequestV2.EntityRow,
+) -> GetOnlineFeaturesResponse.FieldValues:
+    result = GetOnlineFeaturesResponse.FieldValues()
+    for k in row.fields:
+        result.fields[k].CopyFrom(row.fields[k])
+        result.statuses[k] = GetOnlineFeaturesResponse.FieldStatus.PRESENT
```
woop, line 258: Is there a reason for the hardcoding here?

sdk/python/feast/feature_store.py
```diff
@@ -153,26 +160,131 @@ def get_historical_features(
...
-def _get_requested_feature_views(
+        response = self._get_online_features(
+            feature_refs=feature_refs,
+            entity_rows=_infer_online_entity_rows(entity_rows),
+            project=self.config.project,
+        )
+
+        return OnlineResponse(response)
+
+    def _get_online_features(
```
woop, line 199: Possible to make this a function so that it's easier to test?

`_get_online_features(provider, all_feature_views, entity_rows, feature_refs, project)`?

sdk/python/feast/feature_store.py
oavdeev: If I read the previous implementation correctly, we return feature values as well as entity field values in `FieldValues`. This code populates entity field values, and unlike features, these are always "present" since they're part of the request.

sdk/python/feast/feature_store.py
oavdeev: Hmm I'm generally a fan of standalone functions wherever possible too, but here I'd need to pass it basically every property of `FeatureStore` (`provider`, `project` and `registry`). So I figured it is okay to keep it as a method? Don't feel strongly though.

woop: Youd not need to pass project/registry right? You only need to provide the list of feature views. 

I just want to keep the FeatureStore class as simple as possible. It's starting to bloat already. If we're going to add an OnlineStore class as well, then we'll have 3x `get_online_features()` methods. Seems like having this function in the online_store.py would make more sense (I guess the same applies to my historical retriever).

oavdeev: Ah i see you don't want another giant Client, makes sense. How about I do another PR to split out OnlineStore and I can move this stuff there then?

woop: Cool, thanks!

PR #1403, opened 2021-03-22
Support pulling query from BigQuery

jklegar:
**What this PR does / why we need it**: Add support to materialize a specified BigQuery query into the online store

Fixes #

```release-note
Adds support to materialize a specified BigQuery query into the online store
```

### DIFF
--- a/sdk/python/tests/test_bigquery_ingestion.py
+++ b/sdk/python/tests/test_bigquery_ingestion.py
@@ -29,7 +29,7 @@ class TestMaterializeFromBigQueryToDatastore:
         )  # 2 weeks in milliseconds
         self.client.update_dataset(dataset, ["default_table_expiration_ms"])
 
-    def test_bigquery_ingestion_correctness(self):
+    def test_bigquery_table_to_datastore_correctness(self):
         # create dataset
         ts = pd.Timestamp.now(tz="UTC").round("ms")
         checked_value = (
@@ -45,15 +45,13 @@ class TestMaterializeFromBigQueryToDatastore:
 
         # load dataset into BigQuery
         job_config = bigquery.LoadJobConfig()
-        table_id = (
-            f"{self.gcp_project}.{self.bigquery_dataset}.correctness_{int(time.time())}"
-        )
+        table_id = f"{self.gcp_project}.{self.bigquery_dataset}.table_correctness_{int(time.time())}"
         job = self.client.load_table_from_dataframe(df, table_id, job_config=job_config)
         job.result()
 
         # create FeatureView
         fv = FeatureView(
-            name="test_bq_correctness",
+            name="test_bq_table_correctness",
             entities=["driver_id"],
             features=[Feature("value", ValueType.FLOAT)],
             ttl=timedelta(minutes=5),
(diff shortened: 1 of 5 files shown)


jklegar:
Note this is stacked on #1379

woop:
@jklegar can you please rename `test_bigquery_ingestion.py` to `test_materialize_from_bigquery_into_datastore.py` or something?

sdk/python/tests/test_bigquery_ingestion.py
```diff
@@ -0,0 +1,149 @@
...
+        dataset.default_table_expiration_ms = (
+            1000 * 60 * 60 * 24 * 14
+        )  # 2 weeks in milliseconds
+        self.client.update_dataset(dataset, ["default_table_expiration_ms"])
+
+    def test_bigquery_table_to_datastore_correctness(self):
+        # create dataset
+        ts = pd.Timestamp.now(tz="UTC").round("ms")
+        checked_value = (
+            random.random()
```
woop, line 36: Can you please make sure we have a fixed seed? I really want to avoid flaky tests.

sdk/python/tests/test_bigquery_ingestion.py
```diff
@@ -0,0 +1,149 @@
...
+from feast.repo_config import LocalOnlineStoreConfig, OnlineStoreConfig, RepoConfig
+from feast.value_type import ValueType
+
+
+@pytest.mark.integration
+class TestMaterializeFromBigQueryToDatastore:
+    def setup_method(self):
+        self.client = bigquery.Client()
+        self.gcp_project = self.client.project
+        self.bigquery_dataset = "test_ingestion"
```
woop, line 24: We should make a new dataset every time. Have a look here: https://github.com/feast-dev/feast/blob/master/sdk/python/tests/test_historical_retrieval.py#L381

sdk/python/tests/test_bigquery_ingestion.py
jklegar: this is a random value so that if nothing is written to the online store, we don't get a false positive through reading the result of the previous test; a fixed seed would get rid of that. I don't expect this to be flaky

woop: I know we arent planning to have flakyness, but sometimes tests evolve and non-determinism can creep in over time. Like if we start testing more types and there is conversion/casting. 

> we don't get a false positive through reading the result of the previous test

Can we isolate the tests to avoid this problem without using random values? Should we use different projects?

sdk/python/tests/test_materialize_from_bigquery_to_datastore.py
```diff
@@ -78,16 +74,72 @@ def test_bigquery_ingestion_correctness(self):
...
+        assert abs(response_dict[f"{fv.name}:value"][0] - checked_value) < 1e-6
+
+    def test_bigquery_query_to_datastore_correctness(self):
+        # create dataset
+        ts = pd.Timestamp.now(tz="UTC").round("ms")
+        checked_value = (
+            random.random()
+        )  # random value so test doesn't still work if no values written to online store
+        data = {
+            "id": [1, 2, 1],
```
woop, line 95: Is the entity key space shared with other tests?

sdk/python/tests/test_materialize_from_bigquery_to_datastore.py
```diff
@@ -78,16 +74,72 @@ def test_bigquery_ingestion_correctness(self):
...
+        table_id = f"{self.gcp_project}.{self.bigquery_dataset}.query_correctness_{int(time.time())}"
+        query = f"SELECT * FROM `{table_id}`"
+        job = self.client.load_table_from_dataframe(df, table_id, job_config=job_config)
+        job.result()
+
+        # create FeatureView
+        fv = FeatureView(
+            name="test_bq_query_correctness",
+            entities=["driver_id"],
+            features=[Feature("value", ValueType.FLOAT)],
```
woop, line 113: By the way, neither your tests (nor my historical retrieval tests) provide coverage of all types that we support. We should add explicit tests for all types as cards into our backlog.

woop (approved):

PR #1404, opened 2021-03-22
use xargs -I so that it works in both Mac and Linux

randxie:

The Makefile only works in Linux, use `xargs -I` so that it works in both Mac and Linux.

None

For more information about release notes, see kubernetes' guide here:
http://git.k8s.io/community/contributors/guide/release-notes.md
-->
```release-note
use xargs -I instead of xargs -i
```


woop:
@randxie the code is already resolved here: https://github.com/feast-dev/feast/pull/1399/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R51

(we should technically have a separate PR, but we're trying to move quickly right now).

woop:
Think there is something wrong with our credential handling. Checking.

oavdeev:
thanks @randxie! I kinda fixed it in #1399 before i saw your PR so closing this one

PR #1411, opened 2021-03-25
Firestore ingestion perf improvements + benchmark script

oavdeev:
* Use a transaction per (mini)batch of data, to improve write performance.
* Use a thread pool when writing.
* Make sure to retry failed transactions.
* Add a script to benchmark the performance.

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/infra/gcp.py
+++ b/sdk/python/feast/infra/gcp.py
@@ -1,5 +1,7 @@
+import itertools
 from datetime import datetime
-from typing import Dict, List, Optional, Sequence, Tuple, Union
+from multiprocessing.pool import ThreadPool
+from typing import Any, Callable, Dict, Iterator, List, Optional, Sequence, Tuple, Union
 
 import mmh3
 from pytz import utc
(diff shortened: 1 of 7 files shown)


sdk/python/feast/infra/gcp.py
```diff
@@ -178,3 +146,89 @@ def online_read(
...
                 result.append((None, None))
         return result
+
+
+ProtoBatch = Sequence[
+    Tuple[EntityKeyProto, Dict[str, ValueProto], datetime, Optional[datetime]]
+]
+
+
+def _to_minibatches(data: ProtoBatch, batch_size=50) -> Iterator[ProtoBatch]:
```
woop, line 156: Should this batch size be closer to (500/operations)? https://stackoverflow.com/questions/52165352/how-can-i-update-more-than-500-docs-in-firestore-using-batch

oavdeev: Definitely something that we could experiment more with. I mainly wanted to be on the safer side, and given that there are multiple threads pushing batches, having a smaller batch size gives you a less spiky throughput. Also less chance of transaction conflict with smaller batch so we have to retry less.

Empirically, when pushing from from my laptop to firestore, 50 didn't seem to be worse than 250 in terms of throughput, so I left it there.

woop (approved):

PR #1413, opened 2021-03-26
[Python SDK][Auth] Refresh token id w/o gcloud cli

pyalex:

Currently python SDK requires `gcloud` cli to be installed in order to extract `id_token` from ServiceAccount. This is very inconvenient. This PR uses `google.oauth2` instead.

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/grpc/auth.py
+++ b/sdk/python/feast/grpc/auth.py
@@ -12,10 +12,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import json
 from http import HTTPStatus
 
 import grpc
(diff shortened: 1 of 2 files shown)


woop:
So you are able to strip the oid token with the oauth module? I believe I added that code because it didn't work without the gcloud CLI. 

> Currently python SDK requires gcloud cli to be installed in order to extract id_token from ServiceAccount.
It doesnt seem to be a requirement. It's wrapped in a try/except.

pyalex:
> It doesnt seem to be a requirement. It's wrapped in a try/except.
Yes, but `google_auth.default` doesn't refresh id_token for service accounts. Thus the only way to authorize with service account is to have `gcloud cli` installed.

khorshuheng (approved):
/lgtm

PR #1414, opened 2021-03-27
Add Feast init command

oavdeev:
Feast init creates a config file and a sample feature repo, populated with some data in parquet format.

```release-note
NONE
```


sdk/python/feast/repo_operations.py
```diff
@@ -132,3 +137,82 @@ def registry_dump(repo_config: RepoConfig):
...
+        metadata_store: /path/to/metadata.db
+        provider: local
+        online_store:
+            local:
+                path: /path/to/online_store.db
+        """
+            )
+        )
+        print(
+            "Generated example feature_store.yaml. Please edit metadata_store"
```
woop, line 177: Why do we need to edit the metadata store location before running apply?

sdk/python/feast/repo_operations.py
```diff
@@ -132,3 +137,82 @@ def registry_dump(repo_config: RepoConfig):
...
+
+        with open(repo_path / "example.py", "wt") as f:
+            f.write(example_py.replace("%PARQUET_PATH%", str(driver_stats_path)))
+
+        # Generate config
+        repo_config.write_text(
+            dedent(
+                f"""
+        project: {project_id}
+        metadata_store: {data_path / "metadata.db"}
```
woop, line 206: I believe the `data_path` is broken here. It's an absolute url, which will break when the repo is run from CI or elsewhere. We should be using a relative path.

sdk/python/feast/repo_operations.py
woop: Is there any reason we cant use a local path? `metadata.db`?

sdk/python/feast/repo_operations.py
```diff
@@ -132,3 +137,82 @@ def registry_dump(repo_config: RepoConfig):
...
+
+def init_repo(repo_path: Path, minimal: bool):
+
+    repo_config = repo_path / "feature_store.yaml"
+
+    if repo_config.exists():
+        print("Feature repository is already initalized, nothing to do.")
+        sys.exit(1)
+
+    project_id = "proj" + "".join(
```
woop, line 159: I've added this btw https://github.com/feast-dev/feast/pull/1417

sdk/python/feast/repo_operations.py
oavdeev: We could but you'll have to edit `online_store` there anyway. The idea here, unlike non-minimal init, you know what you're doing, you'll want to edit the config file anyway. Maybe change provider too. So I don't want to provide "example" values that may not be what you want but are valid.

woop: Ok, I see what you're trying to do. I just don't know how I feel about that. If the user knows what they are doing then presumably they can create a new `.yaml` by hand. I guess its not a big deal right now.

woop (approved):

PR #1415, opened 2021-03-27
Split Provider interface into OnlineStore and InfraProvisioner

oavdeev:
As discussed in Slack, trying to make the concepts a bit more clear. Now there is an `OnlineStore` class that handles reads and writes, and `InfraProvisioner` that is responsible for creating cloud resources. `Provider` concept only exists as a field in the config file.

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/infra/gcp.py
+++ b/sdk/python/feast/infra/gcp.py
@@ -7,7 +7,8 @@ import mmh3
 from pytz import utc
 
 from feast import FeatureTable, FeatureView
-from feast.infra.provider import Provider
+from feast.infra_provisioner import InfraProvisioner
+from feast.online_store import OnlineStore
 from feast.protos.feast.types.EntityKey_pb2 import EntityKey as EntityKeyProto
 from feast.protos.feast.types.Value_pb2 import Value as ValueProto
 from feast.repo_config import DatastoreOnlineStoreConfig
@@ -47,7 +48,7 @@ def _make_tzaware(t: datetime):
         return t
 
 
-class Gcp(Provider):
+class Gcp(InfraProvisioner):
     _gcp_project_id: Optional[str]
 
     def __init__(self, config: Optional[DatastoreOnlineStoreConfig]):
@@ -68,14 +69,14 @@ class Gcp(Provider):
         self,
         project: str,
         tables_to_delete: Sequence[Union[FeatureTable, FeatureView]],
-        tables_to_keep: Sequence[Union[FeatureTable, FeatureView]],
+        tables_to_have: Sequence[Union[FeatureTable, FeatureView]],
         partial: bool,
     ):
         from google.cloud import datastore
 
         client = self._initialize_client()
 
-        for table in tables_to_keep:
+        for table in tables_to_have:
             key = client.key("Project", project, "Table", table.name)
             entity = datastore.Entity(key=key)
             entity.update({"created_ts": datetime.utcnow()})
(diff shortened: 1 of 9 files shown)


sdk/python/feast/infra/gcp.py
```diff
@@ -104,6 +105,24 @@ def teardown_infra(
...
+class OnlineStoreDatastore(OnlineStore):
+    _gcp_project_id: Optional[str]
+
+    def __init__(self, config: Optional[DatastoreOnlineStoreConfig]):
+        if config:
+            self._gcp_project_id = config.project_id
+        else:
+            self._gcp_project_id = None
+
+    def _initialize_client(self):
```
woop, line 118: It's not really clear to me what purpose the `initialize_client` serves? Could this logic be moved to the constructor? Also, is there a reason the `from google.cloud import datastore` import is called twice? Here, and also at the call site/

sdk/python/feast/infra/local_sqlite.py
```diff
@@ -24,7 +25,7 @@ def _to_naive_utc(ts: datetime):
         return ts.astimezone(pytz.utc).replace(tzinfo=None)
 
 
-class LocalSqlite(Provider):
+class LocalSqlite(InfraProvisioner):
```
woop, line 28: I'm still a bit confused about what the goal of the `InfraProvisioner` is. It doesn't seem to different from the `Provider`, except in name. Why would it be called `LocalSqlite`?

sdk/python/feast/infra/gcp.py
```diff
@@ -104,6 +105,24 @@ def teardown_infra(
             key = client.key("Project", project, "Table", table.name)
             client.delete(key)
 
+
+class OnlineStoreDatastore(OnlineStore):
```
woop, line 109: Nit: Would it be easier to read to start with the name `DatastoreOnlineStore`?

sdk/python/feast/infra/gcp.py
```diff
@@ -68,14 +69,14 @@ def update_infra(
...
+        tables_to_have: Sequence[Union[FeatureTable, FeatureView]],
         partial: bool,
     ):
         from google.cloud import datastore
 
         client = self._initialize_client()
 
-        for table in tables_to_keep:
+        for table in tables_to_have:
             key = client.key("Project", project, "Table", table.name)
```
woop, line 80: The `OnlineStore` and `OfflineStore` abstractions still seem off to me. It seems like we are splitting out the online store's reading/writing functionality, but we want to leave it's schema management inside the provisioner. This essentially splits the life cycle of the online store but retains a hard coupling between the components. Wouldn't it be cleaner to go all out? 

More specifically
1. Move the apply functionality into the OnlineStore and OfflineStore implementation classes (with read/write)
2. Let the `Provider` contain its own workflows like
```python
def apply(...):
   online_store.apply()
   # push_api.apply()
   offline_store.apply()
   # serving_api.apply()

def get_online_store()
   return self._online_store()
```
If we don't go this route, then it's not clear to me why we need a provisioner/provider. We can just work with the online store and offline store directly it seems.

sdk/python/feast/infra/gcp.py
oavdeev: It is juts an utility function, I generally try not to have constructor to do any complex initialization that can fail. Not sure how to avoid the import? Both sites need this module but I don't want to do a top level import to avoid hard dependency.

sdk/python/feast/infra/gcp.py
oavdeev: The assumption that the provisioner for some backends could need to provision more things than just tables for online serving, such as (maybe) iam roles and security groups and things of that nature.

woop: Yea, that's why I listed non-serving examples above (like the push_api), implying that the provisioner/provider would have custom logic somewhere.

oavdeev:
After discussing this on slack, maybe this is not a great idea after all

PR #1420, opened 2021-03-28
Fix broken Terraform installation files

### LINKED ISSUE #1419: Issues following the docs for deploying feast on GCP using Terraform
Hello!

I'm trying to test feast but I've found few issues when trying to deploy the platform on GCP through Terraform following the [docs](https://docs.feast.dev/getting-started/install-feast/google-cloud-gke-with-terraform). I'll list them below.

1. Helm fails to find the feast chart (since it was changed to a different [repository](https://github.com/feast-dev/feast-helm-charts). You can see [here](https://github.com/feast-dev/feast/blob/master/infra/terraform/gcp/feast.tf#L116) a reference to the old path.
2. Once I manually changed variable to point to the right location of the chart, however on the last terraform step (helm release) it's giving me timeout. When doing a small research I found that there's a Pod (feast-serving) entering a `CrashLoopBackOff` state, when looking at the logs I found the following:

> `Factory method 'servingServiceV2' threw exception; nested exception is io.lettuce.core.RedisConnectionException: Unable to connect to feast-0-8-redis-master:6379`

Doing some more research I found that the host should not be `feast-0-8-redis-master` since it should be overwritten [here](https://github.com/feast-dev/feast/blob/master/infra/terraform/gcp/feast.tf#L43) and that's not happening since the name of the variables to overwrite (found [here](https://github.com/feast-dev/feast/blob/master/infra/terraform/gcp/feast.tf#L39)) is `feast-online-serving` instead of `feast-serving` (see [here](https://github.com/feast-dev/feast-helm-charts/tree/main/charts/feast-serving)).

## Expected Behavior 

The terraform apply works completely.

## Current Behavior

Failing on the last step (helm release)

## Steps to reproduce

Basically follow [this](https://docs.feast.dev/getting-started/install-feast/google-cloud-gke-with-terraform) documentation from scratch.

### Specifications

- Version: 0.9.3
- Platform: GKE

## Possible Solution

I think the solution comes in 3 steps:
1. Update docs to reflect the new feast helm repo structure
2. Update terraform file to point to the new feast helm repo
3. Update overwrite vars with the proper chart name (`feast-serving`) so the redis ip is properly set and `feast-serving` pod can connect to redis.

I've tested (2 and 3) locally and finally got feast working on my GCP account. I'll create a sample fix PR since it's not too complicated

josegpg:
This PR only fixes small issues with Terraform reference configuration for GCP so the Quickstart guide actually works.

Fixes #1419 

```release-note
NONE
```

### DIFF
--- a/infra/terraform/gcp/feast.tf
+++ b/infra/terraform/gcp/feast.tf
@@ -36,7 +36,7 @@ locals {
       }
     }
 
-    feast-online-serving = {
+    feast-serving = {
       enabled = true
       "application-override.yaml" = {
         feast = {
@@ -113,7 +113,7 @@ resource "helm_release" "feast" {
   depends_on = [kubernetes_secret.feast-postgres-secret, kubernetes_secret.feast_sa_secret]
 
   name  = var.name_prefix
-  chart = "../../charts/feast"
+  chart = "feast-charts/feast"
 
   values = [
     yamlencode(local.feast_helm_values)
--- a/infra/terraform/gcp/feast.tf
+++ b/infra/terraform/gcp/feast.tf
@@ -113,7 +113,7 @@ resource "helm_release" "feast" {
   depends_on = [kubernetes_secret.feast-postgres-secret, kubernetes_secret.feast_sa_secret]
 
   name  = var.name_prefix
-  chart = "feast-charts/feast"
+  chart = "https://feast-helm-charts.storage.googleapis.com/feast-0.100.4.tgz"
 
   values = [
     yamlencode(local.feast_helm_values)
--- a/docs/getting-started/install-feast/google-cloud-gke-with-terraform.md
+++ b/docs/getting-started/install-feast/google-cloud-gke-with-terraform.md
@@ -22,16 +22,7 @@ This Terraform configuration creates the following resources:
 * Install [Helm](https://helm.sh/docs/intro/install/) \(tested with v3.3.4\)
 * GCP [authentication](https://cloud.google.com/docs/authentication) and sufficient [privilege](https://cloud.google.com/iam/docs/understanding-roles) to create the resources listed above.
 
-### 2. Configure Feast Helm Repository
-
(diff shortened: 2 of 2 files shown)


infra/terraform/gcp/feast.tf
```diff
@@ -113,7 +113,7 @@ resource "helm_release" "feast" {
   depends_on = [kubernetes_secret.feast-postgres-secret, kubernetes_secret.feast_sa_secret]
 
   name  = var.name_prefix
-  chart = "../../charts/feast"
+  chart = "feast-charts/feast"
```
woop, line 116: Would it be easier to use a remote repository? 

Something like `chart = "https://feast-helm-charts.storage.googleapis.com/feast-0.100.4.tgz"`?

woop:
Thanks for this @josegpg!

infra/terraform/gcp/feast.tf
josegpg: It's easier I agree, with this doc changes are not necessary. However this solution fixes the version and with every new helm chart release you'll have to keep these terraform files up to date.

Tell me what you prefer, I'm good with both

woop: > However this solution fixes the version and with every new helm chart release you'll have to keep these terraform files up to date

I actually think this is a feature, not a bug :)

We'd essentially be pinning the dependencies to a specific version, which I think is ideal. We want to provide an opinionated set of components that we know work well together. So I think pointing to a specific chart makes the most sense.

We could use this argument layout to specific the repo + chart + version
```
resource "helm_release" "example" {
  name       = "my-redis-release"
  repository = "https://charts.bitnami.com/bitnami"
  chart      = "redis"
  version    = "6.0.1"
}
```

josegpg: @woop Done!

woop (approved):

PR #1429, opened 2021-03-31
Add entity join key and fix entity references

jklegar:
**What this PR does / why we need it**: Add entity join key and fix FeatureStore methods to use join keys instead of input entity names

Fixes #

```release-note
Add optional join_key parameter to Entity that defaults to the entity name. Previously, the entity name was both the name of the entity and the name of the column it referred to; now one can set join_key to the column name and set name to a more human-readable string.
```

### DIFF
--- a/protos/feast/core/Entity.proto
+++ b/protos/feast/core/Entity.proto
@@ -44,6 +44,9 @@ message EntitySpecV2 {
     // Description of the entity.
     string description = 3;
 
+    // Join key for the entity (i.e. name of the column the entity maps to).
+    string join_key = 4;
+
     // User defined metadata
     map<string,string> labels = 8;
 }
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -110,16 +110,19 @@ class FeatureStore:
         )
         self._registry.refresh()
 
-    def list_entities(self) -> List[Entity]:
+    def list_entities(self, allow_cache: bool = False) -> List[Entity]:
         """
         Retrieve a list of entities from the registry
+
+        Args:
+            allow_cache (bool): Whether to allow returning entities from a cached registry
 
         Returns:
             List of entities
         """
         self._tele.log("list_entities")
 
-        return self._registry.list_entities(self.project)
+        return self._registry.list_entities(self.project, allow_cache=allow_cache)
 
     def list_feature_views(self) -> List[FeatureView]:
         """
@@ -418,9 +421,11 @@ class FeatureStore:
         """
         self._tele.log("get_online_features")
 
-        entity_name_to_join_key_map = self._registry.get_entity_name_to_join_key_map(
-            self.project, allow_cache=True
-        )
+        provider = self._get_provider()
+        entities = self.list_entities(allow_cache=True)
+        entity_name_to_join_key_map = {}
+        for entity in entities:
+            entity_name_to_join_key_map[entity.name] = entity.join_key
 
         join_key_rows = []
         for row in entity_rows:
@@ -435,31 +440,14 @@ class FeatureStore:
                 join_key_row[join_key] = entity_value
             join_key_rows.append(join_key_row)
 
-        response = self._get_online_features(
-            feature_refs=feature_refs,
-            entity_rows=_infer_online_entity_rows(join_key_rows),
-            project=self.config.project,
-            join_key_map=entity_name_to_join_key_map,
-        )
-
-        return OnlineResponse(response)
-
-    def _get_online_features(
-        self,
-        entity_rows: List[GetOnlineFeaturesRequestV2.EntityRow],
-        feature_refs: List[str],
-        project: str,
-        join_key_map: Dict[str, str],
-    ) -> GetOnlineFeaturesResponse:
-
-        provider = self._get_provider()
+        entity_row_proto_list = _infer_online_entity_rows(join_key_rows)
 
         union_of_entity_keys = []
         result_rows: List[GetOnlineFeaturesResponse.FieldValues] = []
 
-        for row in entity_rows:
-            union_of_entity_keys.append(_entity_row_to_key(row))
-            result_rows.append(_entity_row_to_field_values(row))
+        for entity_row_proto in entity_row_proto_list:
+            union_of_entity_keys.append(_entity_row_to_key(entity_row_proto))
+            result_rows.append(_entity_row_to_field_values(entity_row_proto))
 
         all_feature_views = self._registry.list_feature_views(
             project=self.config.project, allow_cache=True
@@ -468,10 +456,10 @@ class FeatureStore:
         grouped_refs = _group_refs(feature_refs, all_feature_views)
         for table, requested_features in grouped_refs:
             entity_keys = _get_table_entity_keys(
-                table, union_of_entity_keys, join_key_map
+                table, union_of_entity_keys, entity_name_to_join_key_map
             )
             read_rows = provider.online_read(
-                project=project, table=table, entity_keys=entity_keys,
+                project=self.project, table=table, entity_keys=entity_keys,
             )
             for row_idx, read_row in enumerate(read_rows):
                 row_ts, feature_data = read_row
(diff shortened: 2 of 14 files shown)


protos/feast/core/Entity.proto
```diff
@@ -44,6 +44,9 @@ message EntitySpecV2 {
     // Description of the entity.
     string description = 3;
 
+    // Join key for the entity (i.e. name of the column it represents).
```
woop, line 47: Should we call this `default_join_key`, or is it impossible to override?

sdk/python/feast/feature_store.py
```diff
@@ -333,17 +333,23 @@ def materialize(
...
     ) -> None:
+        join_keys = []
+        registry = self._get_registry()
+        for entity_name in feature_view.entities:
+            entity = registry.get_entity(entity_name, self.config.project)
+            join_keys.append(entity.join_key)
+
         (
-            entity_names,
+            join_key_columns,
```
woop, line 343: If you are going to rename entity_names to join_key_columns then you should maybe make feature_names feature_name_columns

sdk/python/tests/test_materialize_from_bigquery_to_datastore.py
```diff
@@ -50,10 +52,18 @@ def test_bigquery_table_to_datastore_correctness(self):
...
+            description="id for driver",
+            join_key="driver_id",
+            value_type=ValueType.INT32,
+        )
+
         # create FeatureView
         fv = FeatureView(
             name="test_bq_table_correctness",
-            entities=["driver_id"],
+            entities=["Driver ID"],
```
woop, line 66: Wait, we support arbitrary text for the name? I don't think thats a good idea. We should probably just be supporting lowercase alphanumeric with underscore.

sdk/python/feast/feature_store.py
```diff
@@ -388,9 +392,17 @@ def get_online_features(
             {'sales:daily_transactions': [1.1,1.2], 'sales:customer_id': [0,1]}
         """
 
+        join_key_rows = []
+        for row in entity_rows:
+            join_key_row = {}
+            for key, value in row.items():
```
woop, line 396: Nit: Would you mind naming these variables instead of using `key` and `value`?

sdk/python/feast/feature_store.py
```diff
@@ -388,9 +392,17 @@ def get_online_features(
...
+                entity = self._registry.get_entity(key, self.project, allow_cache=True)
```
woop, line 397: Another nitpick, but since we are on `get_online_features()` i'd want us to be conscious of performance.

It seems a little bit unnatural that we are looping over `get_entity` twice, whether or not we are using the cached copy or not. The flow I was expecting was FeatureView -> Entity, use Entity to find Join Key. We have the FeatureViews as the requirements essentially, and this is what you implemented https://github.com/feast-dev/feast/pull/1429/files/c2d11871da3f9507a734d267f63b4719baf3df30..deb84094071761fd0eba3dc39ffd7ed7175b45fd#diff-ff4e2a65c8c29b7971edc77a61a7e970d410d77d9f173f72714d1bb49e2509a2R435

Ideally we'd not scan over all rows and for all keys in order to find the same entities. It seems more efficient to only call the registry once for each FeatureView+Entity combination, create a mapping of entity name to join_key, and reuse that mapping.

sdk/python/feast/feature_store.py
```diff
@@ -507,7 +525,7 @@ def _get_table_entity_keys(
             if entity_name in required_entities_to_values:
                 if required_entities_to_values[entity_name] is not None:
                     raise ValueError(
-                        f"Duplicate entity keys detected. Table {table.name} expects {table.entities}. The entity "
+                        f"Duplicate entity keys detected. Table expects {table_join_keys}. The entity "
```
woop, line 526: It would be great if we didn't lose the exception context here.

sdk/python/feast/feature_view.py
```diff
@@ -176,3 +177,65 @@ def most_recent_end_time(self) -> Optional[datetime]:
...
             return None
         return max([interval[1] for interval in self.materialization_intervals])
+
+    def run_reverse_field_mapping(
+        self, join_keys: List[str],
+    ) -> Tuple[List[str], List[str], str, Optional[str]]:
+        """
+        If a field mapping exists, run it in reverse on the join keys,
+        feature names, event timestamp column, and created timestamp column
+        to get the names of the relevant columns in the BigQuery table.
```
woop, line 187: I dont think we should be talking about BigQuery here. This is a general class. Unless we are referencing is as an example.

sdk/python/feast/feature_view.py
```diff
@@ -176,3 +177,65 @@ def most_recent_end_time(self) -> Optional[datetime]:
         if len(self.materialization_intervals) == 0:
             return None
         return max([interval[1] for interval in self.materialization_intervals])
+
+    def run_reverse_field_mapping(
+        self, join_keys: List[str],
```
woop, line 182: The name `run_reverse_field_mapping` provides almost no intent to the user. Would `get_field_names()` be better?

sdk/python/feast/infra/gcp.py
```diff
@@ -126,41 +128,63 @@ def online_read(
...
+        for entity_name in feature_view.entities:
+            entity = registry.get_entity(entity_name, project)
+            join_keys.append(entity.join_key)
+
+        (
+            join_key_columns,
+            feature_name_columns,
+            event_timestamp_column,
+            created_timestamp_column,
+        ) = feature_view.run_reverse_field_mapping(join_keys)
```
woop, line 151: Not sure if its a good practice to query the registry here and pass in the join keys. Its almost a half step. We should be pulling the feature views from the registry here, based on their names. Then we should either 
* Pull the entities as you've done it here and have a helper function like `run_reverse_field_mapping(feature_view, entities)
or
* When we pull the FeatureViews from the registry they get a reference to the `FeatureStore` stored inside of them. Then we can add a method like `FeatureView.get_entities()` which returns its Entity objects. May need to change the previous one to `FeatureView.entity_names` so that it isnt confusing. Anyway, then when you run `FeatureView.run_reverse_field_mapping()` it will find its own entities using the reference it has to the FeatureStore class, as opposed to it being passed in.
or
* We remove this List[str] as entities and replace it with `List[Entity]` so that when we do a `get_feature_view()` we actually get the entity details back in the first place. Its also much cleaner for end users when registering FeatureViews since they dont have to deal with string references.

protos/feast/core/Entity.proto
jklegar: currently you would use a feature map to map the column to this value rather than override the join key

sdk/python/feast/feature_store.py
jklegar: good call - we now create a map from entity names to join keys that we reuse

sdk/python/feast/feature_view.py
jklegar: changed run_reverse_field_mapping to _get_column_names; I think returning a tuple vs a dictionary are both fine, and inclined to just leave as is until we have a reason to change things. BTW this method is not user-facing and now starts with an underscore

sdk/python/feast/infra/gcp.py
jklegar: Yeah there's definitely tech debt here - IMO the main problem is getting the feature view from the registry doesn't also get the entities from the registry and we should do something like your third option (in a separate diff). Happy to do the first option here though not sure how much it actually gets us since we need join_keys in this method further down anyway

woop: I guess my concern is just that we are making the code base more complicated (using a method) but we're not gaining anything from that since the method acts like a function.

> need join_keys in this method further down anyway

Wouldnt we be able to use this list of entities to get the join keys? I think what I am trying to avoid as much as possible is to have methods with I/O. I want to us to try and follow a more immutable/functional approach as much as we can (which I realize isnt 100% possible). So get_entities/get_feature_views etc should ideally be run only once near the start of one of our methods and then the state should just propagate down. It'll be way easier to reason about and test long term.

woop (approved):

PR #1431, opened 2021-04-02
Add registry refreshing and caching

woop:
* Adds support for caching of the RegistryProto as part of the Registry. Used during `get_online_features()` to prevent users from having to download the registry on every call.
* Also adds a `refresh_registry()` method to the FeatureStore class for users that want to explicitly update their registries.

```release-note
Added support for registry caching and added a refresh_registry() method to the FeatureStore class
```


woop:
Blocked by #1432

sdk/python/feast/feature_store.py
```diff
@@ -72,15 +73,31 @@ def __init__(
...
     def _get_provider(self) -> Provider:
         return get_provider(self.config)
 
-    def _get_registry(self) -> Registry:
-        return Registry(self.config.metadata_store)
+    def refresh_registry(self):
+        """Fetches and caches a copy of the feature registry in memory.
+
+        Explicitly calling this method makes it possible for methods like get_online_features to use cached registry
+        state instead of retrieving the registry state at request-time, thus reducing latency."""
```
jklegar, line 93: can we clarify a little more how this works, i.e. if this function is never called then every cache_ttl seconds the get_online_features request will refresh the cache but if this function is called it refreshes the cache and resets the timer back to cache_ttl seconds

sdk/python/feast/registry.py
```diff
@@ -333,16 +372,22 @@ def __init__(self, filepath: str):
...
-        else:
+            return registry_proto
+        raise FileNotFoundError(
+            f'Registry not found at path "{self._filepath}". Have you run "feast apply"?'
+        )
+
+    def update_registry_proto(self, updater: Callable[[RegistryProto], RegistryProto]):
+        try:
+            registry_proto = self.get_registry_proto()
+        except FileNotFoundError:
```
jklegar, line 387: nice

jklegar (approved):
LGTM with one comment

PR #1436, opened 2021-04-03
Add materialize-incremental to feast cli

tsotnet:

**What this PR does / why we need it**: We have `fs.materialize_incremental` functionality in the `FeatureStore`, but don't have the equivalent function in the CLI. This PR adds that command, while also fixing a bug with timezones. By default, the timezone information was not included in the generated `start_time` and that led to `TypeError: Cannot compare tz-naive and tz-aware datetime-like objects` errors.

Fixes #

```release-note
Add `feast materialize-incremental` command to the CLI
```

### DIFF
--- a/sdk/python/feast/cli.py
+++ b/sdk/python/feast/cli.py
@@ -430,6 +430,30 @@ def materialize_command(start_ts: str, end_ts: str, repo_path: str, views: List[
     )
 
 
+@cli.command("materialize-incremental")
+@click.argument("end_ts")
+@click.argument(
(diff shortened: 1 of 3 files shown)


woop:
Thanks @tsotnet.

Would you mind adding a test? I believe we have a CLI test runner.

> By default, the timezone information was not included in the generated start_time and that led to TypeError: Cannot compare tz-naive and tz-aware datetime-like objects errors.

We're tracking this card elsewhere. I dont think it should be a part of this PR.

tsotnet:
I broke out timezone changes in a separate PR. I'll close this and re-open another PR after I land that bug fix.

PR #1439, opened 2021-04-04
Fix timezone issue in materialize & materialize_incremental

tsotnet:

**What this PR does / why we need it**: Fixes an issue of mixing tz-aware and tz-naive timestamps. This was happening in fs.materialize & fs.materialize_incremental (both call _materialize_single_feature_view where I applied the fix), in FeatureView where the time zones aren't persisted between to_proto & from_proto, and in LocalSqlite:pull_latest_from_table_or_query where the source dataframe could have some tz-naive timestamps.

I also refactored `test_materialize_from_bigquery_to_datastore.py` into `test_materialize.py` where I maintain both existing test cases but also add a unit test that runs in the local mode. Lastly, I made sure to mix in tz-naive and tz-aware timestamps so that we make sure this bug is now well-tested.

Fixes #

```release-note
Fixes tz-naive / tz-aware bug with materialization
```


sdk/python/tests/test_materialize.py
```diff
@@ -0,0 +1,168 @@
...
+                online_store=OnlineStoreConfig(
+                    local=LocalOnlineStoreConfig(
+                        path=str(Path(data_dir_name) / "online_store.db")
+                    )
+                ),
+            )
+            fs = FeatureStore(config=config)
+            fs.apply([fv])
+
+            self.run_materialization_test(fs, fv)
```
woop, line 168: This is a bit of a nitpick, but what do you think of restructuring the tests a little bit? I dont really see the need for a class and methods here when we can use functions. I'm worried that the class will create unnecessary scoping or tests will evolve to influence each other over time. I also see you are mixing graph building in a push (`self.run_materialization_test(fs, fv)`) and pull (`@pytest.mark.parametrize`) manner. We should probably just pick one of the two, and the pull based graph building seems cleaner to me.

something like

def other_functions()

def prep_bq_fs()
  do all bq setup

def prep_local_fs()
  do all local setup

@pytest.mark.parametrize( depends on prep_bq_fs and prep_local_fs)
def test_materialization()

woop (changes requested):
Mostly looks good, just some comments on the test structure

sdk/python/tests/test_materialize.py
tsotnet: Most of this makes sense, but the problem with
> @pytest.mark.parametrize( depends on prep_bq_fs and prep_local_fs)
def test_materialization()

is that I either have to make the entire test function as integration test or unit test. There's no way to selectively mark prep_bq_fs as integration and prep_local_fs as unit test (I just tried this locally).

tsotnet: @woop I went ahead with the push based approach. Basically make everything functions that we call in order to not duplicate anything

woop (approved):

PR #1440, opened 2021-04-06
Support join keys in historical feature retrieval

jklegar:
**What this PR does / why we need it**: Use join keys from the registry in historical feature retrieval rather than using the entity name strings in the input.

Fixes #

```release-note
NONE
```


jklegar:
Note this is stacked on #1429

sdk/python/feast/entity.py
```diff
@@ -36,11 +36,17 @@ def __init__(
         name: str,
         value_type: ValueType,
         description: str = "",
+        join_key: Optional[str] = None,
```
woop, line 39: I think it makes sense to add a docstring on Entity. Seems to be missing.

PR #1442, opened 2021-04-07
Add materialize-incremental cli command

tsotnet:

**What this PR does / why we need it**: We have fs.materialize_incremental functionality in the FeatureStore, but don't have the equivalent function in the CLI. This PR adds that command.

Fixes #

```release-note
Add `feast materialize-incremental` command to the CLI
```

### DIFF
--- a/sdk/python/feast/cli.py
+++ b/sdk/python/feast/cli.py
@@ -416,7 +416,7 @@ def materialize_command(start_ts: str, end_ts: str, repo_path: str, views: List[
     """
     Run a (non-incremental) materialization job to ingest data into the online store. Feast
     will read all data between START_TS and END_TS from the offline store and write it to the
-    online store. If you don't specify feature view names using --views, all registred Feature
+    online store. If you don't specify feature view names using --views, all registered Feature
     Views will be materialized.
 
     START_TS and END_TS should be in ISO 8601 format, e.g. '2021-07-16T19:20:01'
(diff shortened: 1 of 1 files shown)


sdk/python/feast/cli.py
```diff
@@ -429,6 +429,30 @@ def materialize_command(start_ts: str, end_ts: str, repo_path: str, views: List[
     )
 
 
+@cli.command("materialize-incremental")
+@click.argument("end_ts")
+@click.argument(
+    "repo_path", type=click.Path(dir_okay=True, exists=True,), default=Path.cwd
```
woop, line 435: Wait, do we still support `repo_path`? I think we said that we would just allow running Feast commands from the CLI in a repo (not remotely).

tsotnet: I haven't been part of that conversation, just copied from materialize command. Should I delete it in both places and replace `FeatureStore(repo_path=repo_path)` with `FeatureStore(repo_path=Path.cwd)`?

woop: @tsotnet can you maybe just create a card to remove it? I think it can be a separate PR

woop (approved):

sdk/python/feast/cli.py
tsotnet: Done https://tecton.atlassian.net/browse/TEC-4292

PR #1458, opened 2021-04-12
Optimize write rate in Gcp Firestore

tsotnet:
**What this PR does / why we need it**: We used to have less than 100 QPS on writing data to the online store in Gcp Firestore. This was due to using non-batch requests (get & put) instead of batch variants (get_multi & put_multi). Moreover, we completely removed the read part of the functionality and directly write the new data without comparing timestamps. So, it's up to the user now to make sure they're not overwriting old data over the new data. In future we'll add a flag that if enabled by the user, will do this check on every write, but the performance hit is too much to leave it on by default.

The new functionality benchmarks at ~1.95K writes per second.

Fixes #

```release-note
Improve Firestore write speed by 2 orders of magnitude
```

### DIFF
--- a/sdk/python/feast/infra/gcp.py
+++ b/sdk/python/feast/infra/gcp.py
@@ -93,7 +93,7 @@ class GcpProvider(Provider):
     ) -> None:
         client = self._initialize_client()
 
-        pool = ThreadPool(processes=10)
+        pool = ThreadPool(processes=40)
         pool.map(
             lambda b: _write_minibatch(client, project, table, b, progress),
             _to_minibatches(data),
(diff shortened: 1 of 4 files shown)


sdk/python/feast/infra/gcp.py
```diff
@@ -93,7 +93,7 @@ def online_write_batch(
     ) -> None:
         client = self._initialize_client()
 
-        pool = ThreadPool(processes=10)
+        pool = ThreadPool(processes=40)
```
woop, line 96: Do we need 40 here? Where does the ROI drop off?

sdk/python/feast/infra/gcp.py
```diff
@@ -231,62 +231,32 @@ def _write_minibatch(
...
+                key=entity_key.SerializeToString(),
+                values={k: v.SerializeToString() for k, v in features.items()},
+                event_ts=utils.make_tzaware(timestamp),
+                created_ts=(
+                    utils.make_tzaware(created_ts) if created_ts is not None else None
+                ),
+            )
+        )
+        entities.append(entity)
+    with client.transaction():
```
woop, line 255: Do we have to use a transaction?

sdk/python/feast/infra/gcp.py
tsotnet: I saw an improvement from 20 -> 40, but didn't see any improvement from 40 -> 50.

sdk/python/feast/infra/gcp.py
tsotnet: Yes we have to, otherwise we get the following exception:
```
google.api_core.exceptions.InvalidArgument: 400 A non-transactional commit may not contain multiple mutations affecting the same entity.
```

woop (approved):

PR #1459, opened 2021-04-13
Flatten configuration structure for online store

woop:

The current configuration structure requires a nested `online_store` configuration like this
```
project: foo
registry: "registry.db"
provider: local
online_store:
    local:
        path: online_store.db
```
This refactor moves to 
```
project: foo
registry: "registry.db"
provider: local
online_store:
    type: sqlite
    path: online_store.db
```
but allows for a default type to be set, based on the provider. 
```
project: foo
registry: "registry.db"
provider: local
online_store:
    path: online_store.db
```
Much of the work in this PR is just to ensure that exceptions are clear to the user.

```release-note
Flattened configuration structure for online store
```


sdk/python/feast/repo_config.py
```diff
@@ -72,40 +74,54 @@ def get_registry_config(self):
...
-        },
-    },
-    "required": ["project"],
-    "additionalProperties": False,
-}
+    @root_validator(pre=True)
+    def _validate_online_store_config(cls, values):
+        # This method will validate whether the online store configurations are set correctly. This explicit validation
+        # is necessary because Pydantic Unions throw very verbose and cryptic exceptions. We also use this method to
+        # impute the default online store type based on the selected provider. For the time this method should be
```
woop, line 81:
```suggestion
        # impute the default online store type based on the selected provider. For the time being this method should be
```

sdk/python/feast/repo_config.py
```diff
@@ -63,7 +65,7 @@ class RepoConfig(FeastBaseModel):
     provider: StrictStr
```
jklegar, line 65: I think we'll need defaults for provider and project in order for `RepoConfig()` with no arguments to work

jklegar:
fyi there's an empty file `sdk/python/telemetry_tests/data/online.db` included in here

jklegar (approved):

PR #1462, opened 2021-04-13
Change FeatureStore default config to read from current directory

jklegar:

**What this PR does / why we need it**: Change FeatureStore default config to read from current directory instead of using a fixed default config

Fixes #

```release-note
Running `FeatureStore()` without any arguments now looks in the current working directory for a `feature_store.yaml` file instead of using a fixed default config.
```


woop:
@jklegar its not clear what the default behaviour is here. Is it

```
# reads from repository at current location?
fs = FeatureStore()
```
wouldn't that fail in most cases if the user hasn't run `init`?

jklegar:
covered by #1459

PR #1466, opened 2021-04-14
Fix offline store (tz-naive & field_mapping issues)

tsotnet:

**What this PR does / why we need it**: Discovered 2 different bugs in get_historical_features. 1) if the entity_df contains tz-naive timestamps, point-in-time merging doesn't work in local mode. 2) field_mapping is not being used in offline mode (it's used in online mode). Meaning that if you define field_mapping, get_historical_features simply fails in both local and gcp modes.

Fixes #

```release-note
Fix tz-naive issue with offline store and apply field_mapping during historical feature retrieval
```

### DIFF
--- a/sdk/python/tests/test_materialize.py
+++ b/sdk/python/tests/test_materialize.py
@@ -83,20 +83,20 @@ def prep_bq_fs_and_fv(
         event_timestamp_column="ts",
         created_timestamp_column="created_ts",
         date_partition_column="",
-        field_mapping={"ts_1": "ts", "id": "driver_ident"},
+        field_mapping={"ts_1": "ts", "id": "driver_id"},
     )
 
     fv = get_feature_view(bigquery_source)
     e = Entity(
         name="driver_id",
         description="id for driver",
-        join_key="driver_ident",
+        join_key="driver_id",
         value_type=ValueType.INT32,
     )
     with tempfile.TemporaryDirectory() as repo_dir_name:
         config = RepoConfig(
             registry=str(Path(repo_dir_name) / "registry.db"),
-            project=f"test_bq_correctness_{uuid.uuid4()}",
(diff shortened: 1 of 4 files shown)


sdk/python/tests/test_materialize.py
```diff
@@ -83,20 +83,20 @@ def prep_bq_fs_and_fv(
...
         date_partition_column="",
-        field_mapping={"ts_1": "ts", "id": "driver_ident"},
+        field_mapping={"ts_1": "ts", "id": "driver_id"},
     )
 
     fv = get_feature_view(bigquery_source)
     e = Entity(
         name="driver_id",
         description="id for driver",
-        join_key="driver_ident",
```
jklegar, line 93: this tests that things still work when the join_key is different from the name of the entity (although it should probably be in its own separate test) so can we keep it in?

sdk/python/tests/test_materialize.py
```diff
@@ -138,6 +138,31 @@ def prep_local_fs_and_fv() -> Iterator[Tuple[FeatureStore, FeatureView]]:
...
+    event_timestamp: datetime,
+    expected_value: float,
+) -> None:
+    # Check online store
+    response_dict = fs.get_online_features(
+        [f"{fv.name}:value"], [{"driver_id": driver_id}]
+    ).to_dict()
+    assert abs(response_dict[f"{fv.name}__value"][0] - expected_value) < 1e-6
+
+    # Check offline store
```
jklegar, line 155: we do have sdk/python/tests/test_historical_retrieval.py - this works but it may be more natural to cover the field mapping cases there?

woop: Agree

jklegar (approved):

PR #1470, opened 2021-04-15
Change GCP template names to match local template

jklegar:

**What this PR does / why we need it**: Running through the quickstart should now also work with `feast init -t gcp`

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/templates/gcp/driver_repo.py
+++ b/sdk/python/feast/templates/gcp/driver_repo.py
@@ -7,7 +7,7 @@ from feast import BigQuerySource, Entity, Feature, FeatureView, ValueType
 # construction of feature vectors
 driver = Entity(
     # Name of the entity. Must be unique within a project
-    name="driver",
+    name="driver_id",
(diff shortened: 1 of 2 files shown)


sdk/python/feast/templates/gcp/driver_repo.py
```diff
@@ -7,7 +7,7 @@
 # construction of feature vectors
 driver = Entity(
     # Name of the entity. Must be unique within a project
-    name="driver",
+    name="driver_id",
```
woop, line 10: was this intentional?

jklegar: Yeah, `get_online_features` uses entity name and the quickstart has `driver_id` in the input so the name has to be `driver_id` for the GCP template to work with the quickstart. When we change `get_online_features` to use the join_key instead of name this won't matter, but it's probably a good idea for the GCP template to match the local template here anyway

woop (approved):

PR #1473, opened 2021-04-16
Python docs formatting fixes

jklegar:

**What this PR does / why we need it**: Updates and fixes formatting issues in the python SDK reference documentation

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/repo_config.py
+++ b/sdk/python/feast/repo_config.py
@@ -41,7 +41,7 @@ class RegistryConfig(FeastBaseModel):
     """ Metadata Store Configuration. Configuration that relates to reading from and writing to the Feast registry."""
 
     path: StrictStr
-    """ str: Path to metadata store. Can be a local path, or remote object storage path, e.g. gcs://foo/bar """
+    """ str: Path to metadata store. Can be a local path, or remote object storage path, e.g. a GCS URI """
 
     cache_ttl_seconds: StrictInt = 600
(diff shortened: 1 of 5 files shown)


sdk/python/feast/repo_config.py
```diff
@@ -54,7 +54,7 @@ class RepoConfig(FeastBaseModel):
     """ Repo config. Typically loaded from `feature_store.yaml` """
 
     registry: Union[StrictStr, RegistryConfig] = "data/registry.db"
-    """ str: Path to metadata store. Can be a local path, or remote object storage path, e.g. gcs://foo/bar """
+    """ str: Path to metadata store. Can be a local path, or remote object storage path, e.g. a GCS URI """
```
woop, line 57: Is there a reason we can't have an example like `gs://my_bucket/my_path`?

woop, line 57:
```suggestion
    """ str: Path to registry. Can be a local path, or remote object storage path, e.g. a GCS URI """
```

jklegar: I assume there is a way but I spent way too long trying things and googling things to get the literal colon to show up instead of it getting interpreted as a special character like the colon after str. So eventually I gave up and just wrote GCS URI :)

woop (changes requested):
Build still failing btw

sdk/python/feast/repo_config.py
woop: Would you mind creating a card to track this then? Perhaps a card to ensure that configurations come with a real example?

PR #1474, opened 2021-04-16
Fix created timestamp related errors for BigQuery source

jklegar:

**What this PR does / why we need it**: Fix errors in historical retrieval when created_timestamp_column is not set on a BigQuery data source

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/tests/test_historical_retrieval.py
+++ b/sdk/python/tests/test_historical_retrieval.py
@@ -39,11 +39,15 @@ def generate_entities(date):
     return customer_entities, driver_entities, end_date, orders_df, start_date
 
 
-def stage_driver_hourly_stats_parquet_source(directory, df):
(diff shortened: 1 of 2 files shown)


sdk/python/tests/test_historical_retrieval.py
```diff
@@ -281,7 +296,12 @@ def test_historical_features_from_parquet_sources():
 @pytest.mark.parametrize(
     "provider_type", ["local", "gcp"],
 )
-def test_historical_features_from_bigquery_sources(provider_type):
+@pytest.mark.parametrize(
+    "created_timestamp_column", ["created", ""],
```
woop, line 300: My only feedback is that these tests tend to run long, and will run even longer in the future the more we start to parameterize it. Would it be better to just have a single source (out of multiple) that doesnt have a created TS?

woop (commented):
Just one comment, otherwise it looks good

woop (approved):

PR #1475, opened 2021-04-17
Fix time zone issue with get_historical_features

tsotnet:

**What this PR does / why we need it**: There was an issue with FeatureStore.get_historical_features when you pass records with timestamp with non-UTC time zones (pandas merge_asof was throwing an exception). This PR fixes that issue by converting all timestamps in the entity to UTC. If timestamp is tz-naive, we assume it's UTC. If it's tz-aware, we localize to UTC.

I also made significant changes to the tests. First, when generating entities we now generate timestamps in all kinds of different formats. Second, `test_historical_retrieval.py:get_expected_training_df` function used a very similar logic to the local offline store (based on pandas merge_asof). Meaning that if there are bugs in the local offline store, we replicate them in the test. Instead of adding the identical logic to this test, I rewrote this method to do manual (non-pandas) join. This should give us more confidence in the offline store correctness.

Fixes #

```release-note
Fix time zone issue with get_historical_features
```

### DIFF
--- a/sdk/python/tests/test_historical_retrieval.py
+++ b/sdk/python/tests/test_historical_retrieval.py
@@ -178,6 +178,15 @@ def get_expected_training_df(
     current_cols.remove(ENTITY_DF_EVENT_TIMESTAMP_COL)
     expected_df = expected_df[[ENTITY_DF_EVENT_TIMESTAMP_COL] + current_cols]
 
+    # Cast some columns to expected types, since we lose information when converting pandas DFs into Python objects.
+    expected_df["order_is_success"] = expected_df["order_is_success"].astype("int32")
+    expected_df["customer_profile__current_balance"] = expected_df[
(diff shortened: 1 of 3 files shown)


sdk/python/tests/test_historical_retrieval.py
```diff
@@ -274,6 +286,7 @@ def test_historical_features_from_parquet_sources():
                     "customer_id",
                 ]
             ).reset_index(drop=True),
+            check_dtype=False,
```
woop, line 289: Is there a reason for this change?

tsotnet: There were errors for 32bit vs 64 bit stuff, and didn't think it was important to change. I can look into it if you're concerned though

woop (approved):

PR #1479, opened 2021-04-19
Get historical features from the Provider level + Ability to add any OfflineStore

MattDelac:
Everything around "get_historical_features()" is actually retrieve from the source of the FeatureView. Therefore this PR aims to remove the dependency at the _"Provider"_ level

There are also a lot of assumptions in the code about GCP or LocalFile. The long term goal being to accept any kind of custom Provider or Source, we should avoid hard coded "if/elif" statements and make the code more modular from that perspective

Did not create any issue ...
Fixes #

NONE
```release-note
Move get_historical_features() off the Provider interface
A DataSource can now have an offline_store assigned to it
```

### DIFF
--- a/sdk/python/feast/data_source.py
+++ b/sdk/python/feast/data_source.py
@@ -446,6 +446,12 @@ class DataSource:
         """
         self._date_partition_column = date_partition_column
 
+    @property
+    def offline_store(self):
+        # Return OfflineStore() object
+        # Cannot be imported because of cyclic dependency
+        ...
+
     @staticmethod
     def from_proto(data_source):
         """
@@ -596,6 +602,12 @@ class FileSource(DataSource):
         """
         return self._file_options.file_url
 
+    @property
+    def offline_store(self):
+        from feast.infra.offline_stores.file import FileOfflineStore
+
+        return FileOfflineStore()
+
     def to_proto(self) -> DataSourceProto:
         data_source_proto = DataSourceProto(
             type=DataSourceProto.BATCH_FILE,
@@ -663,6 +675,12 @@ class BigQuerySource(DataSource):
         Sets the bigquery options of this data source
         """
         self._bigquery_options = bigquery_options
+
+    @property
+    def offline_store(self):
+        from feast.infra.offline_stores.bigquery import BigQueryOfflineStore
+
+        return BigQueryOfflineStore()
 
     def to_proto(self) -> DataSourceProto:
         data_source_proto = DataSourceProto(
--- a/sdk/python/tests/infra/offline_stores/test_helpers.py
+++ b/sdk/python/tests/infra/offline_stores/test_helpers.py
@@ -1,0 +1,31 @@
+import pytest
+
+from feast.data_source import BigQuerySource, FileSource
+from feast.infra.offline_stores.bigquery import BigQueryOfflineStore
+from feast.infra.offline_stores.file import FileOfflineStore
+from feast.infra.offline_stores.helpers import get_offline_store_from_sources
+
+
+@pytest.fixture
+def bigquery_source():
+    return BigQuerySource(event_timestamp_column="events")
+
+
+@pytest.fixture
+def file_source():
+    return FileSource(event_timestamp_column="events", path="tmp/")
+
+
+def test_retrieve_bigquery_offline_store(bigquery_source):
+    offline_store = get_offline_store_from_sources([bigquery_source, bigquery_source])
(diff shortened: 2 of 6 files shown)


sdk/python/tests/infra/offline_stores/test_helpers.py
```diff
@@ -0,0 +1,24 @@
+import pytest
```
MattDelac, line 1: Unsure why I get an import statement when adding those tests ... 🤔 
```
sdk/python/feast/client.py:26: in <module>
    from feast.data_source import BigQuerySource, FileSource
sdk/python/feast/data_source.py:20: in <module>
    from feast.infra.offline_stores.bigquery import BigQueryOfflineStore
sdk/python/feast/infra/offline_stores/bigquery.py:11: in <module>
    from feast.data_source import BigQuerySource, DataSource
E   ImportError: cannot import name 'BigQuerySource' from 'feast.data_source' (/Users/mattdelacour/src/github.com/Shopify/feast/sdk/python/feast/data_source.py)
```

woop: Cycling dependency?

MattDelac: Ho yeah it sounds like it ... 😞 

Will have a look tomorrow

MattDelac: Done

MattDelac: So my understanding is 
DataSource (from `data_source.py`) is meant to be a low level interface. Then "stores" lile BigQueryOfflineStore can use data sources.

But when I look at the references, BigQuerySource is used everywhere. So it was a bit tricky to bypass the cyclic dependency (I basically make the import statement at the property level which is not great but could not think about another option)

Happy to get you opinion on this one 🙂

MattDelac:
Sorry for the noise earlier. This PR is now ready to be reviewed (all checks are green)

Thanks in advance 🙏

woop:
> Everything around "get_historical_features()" is actually retrieve from the source of the FeatureView. Therefore this PR aims to remove the dependency at the "Provider" level

I think we need more discussion about this. The reason we have `get_historical_features()` on the provider interface is because (for the moment) we want to provider to be able to control which offline stores it can interact with. For example its possible that a local provider may be able to access local files like `file://my/file`, but that may not be the case for a cloud provider. 

> There are also a lot of assumptions in the code about GCP or LocalFile. The long term goal being to accept any kind of custom Provider or Source, we should avoid hard coded "if/elif" statements and make the code more modular from that perspective

Agreed. As a first cut it will always be our approach to use hardcoded logic to keep code simple. Given that we're reaching the point where you'd like to extend Feast for your internal stack, it now makes sense to think through what generalization is needed.

sdk/python/feast/data_source.py
```diff
@@ -446,6 +446,12 @@ def date_partition_column(self, date_partition_column):
         """
         self._date_partition_column = date_partition_column
 
+    @property
+    def offline_store(self):
```
woop, line 450: should this be an abstract method?

sdk/python/feast/data_source.py
```diff
@@ -596,6 +602,12 @@ def path(self):
         """
         return self._file_options.file_url
 
+    @property
+    def offline_store(self):
+        from feast.infra.offline_stores.file import FileOfflineStore
+
+        return FileOfflineStore()
```
woop, line 609: I am not sure if the approach being followed here is the correct one. Here we are saying "A source knows which store serves it", but wouldn't it be the reverse, like "A store knows which sources it can serve". So the way I was expecting the implementation to move towards was that the provider is given a list of sources and it decides based on the stores available to it which store to use.

Put differently, how would you allow two different store types to handle FileSource? Presumably we could have a Pandas implementation and a Dask implementation.

sdk/python/tests/infra/offline_stores/test_helpers.py
woop: > But when I look at the references, BigQuerySource is used everywhere. So it was a bit tricky to bypass the cyclic dependency (I basically make the import statement at the property level which is not great but could not think about another option)

Are there any areas where you see us using BigQuerySource where you think it should not occur?

Cyclic dependencies generally (although not always) point to a problem with code.

MattDelac:
Closing it as I wanted to fire a PR to start a discussion about the Provider concept in Feast (should have been a draft PR, sorry for the noise)

PR #1481, opened 2021-04-20
Fix materialize for None

### LINKED ISSUE #1480: Error during feast materialize-incremental: AttributeError: 'NoneType' object has no attribute 'dtype'
## Expected Behavior 

It should be possible to materialize TableView if some values for some rows are None

## Current Behavior

In the current version error appears when any value in historical store is None.
On command:
```
feast materialize-incremental 2021-05-21T23:46:43
```

I have error:
```
Materializing feature view app_features from 2021-04-19 00:03:27.010110+00:00 to 2021-04-21 21:55:52+00:00Traceback (most recent call last):

  File "/opt/conda/bin/feast", line 8, in <module>

    sys.exit(cli())

  File "/opt/conda/lib/python3.8/site-packages/click/core.py", line 829, in __call__

    return self.main(*args, **kwargs)

  File "/opt/conda/lib/python3.8/site-packages/click/core.py", line 782, in main

    rv = self.invoke(ctx)

  File "/opt/conda/lib/python3.8/site-packages/click/core.py", line 1259, in invoke

    return _process_result(sub_ctx.command.invoke(sub_ctx))

  File "/opt/conda/lib/python3.8/site-packages/click/core.py", line 1066, in invoke

    return ctx.invoke(self.callback, **ctx.params)

  File "/opt/conda/lib/python3.8/site-packages/click/core.py", line 610, in invoke

    return callback(*args, **kwargs)

  File "/opt/conda/lib/python3.8/site-packages/feast/cli.py", line 247, in materialize_incremental_command

    store.materialize_incremental(

  File "/opt/conda/lib/python3.8/site-packages/feast/feature_store.py", line 337, in materialize_incremental

    provider.materialize_single_feature_view(

  File "/opt/conda/lib/python3.8/site-packages/feast/infra/local.py", line 188, in materialize_single_feature_view

    rows_to_write = _convert_arrow_to_proto(table, feature_view, join_keys)

  File "/opt/conda/lib/python3.8/site-packages/feast/infra/provider.py", line 260, in _convert_arrow_to_proto

    value = python_value_to_proto_value(row[idx])

  File "/opt/conda/lib/python3.8/site-packages/feast/type_map.py", line 317, in python_value_to_proto_value

    value_type = python_type_to_feast_value_type("", value)

  File "/opt/conda/lib/python3.8/site-packages/feast/type_map.py", line 160, in python_type_to_feast_value_type

    return type_map[value.dtype.__str__()]

AttributeError: 'NoneType' object has no attribute 'dtype'
```

## Steps to reproduce

The error will appear for any input which contains None value in any row.

### Specifications

- Version: Feast SDK Version: "feast 0.10.0"
- Platform: Python
- Subsystem: Linux

## Possible Solution

We can use feature ValueType instead of parsing type from value if value is None

qooba:
Signed-off-by: qooba <dev@qooba.net>

This pull request fixes the issue during materialize view.
If in input data there is a None value feast can't parse its type in python_type_to_feast_value_type method

Fixes #1480

```release-note
Added support for materializing sources with None types.
```

### DIFF
--- a/sdk/python/feast/infra/provider.py
+++ b/sdk/python/feast/infra/provider.py
@@ -257,7 +257,7 @@ def _convert_arrow_to_proto(
         feature_dict = {}
         for feature in feature_view.features:
             idx = table.column_names.index(feature.name)
-            value = python_value_to_proto_value(row[idx])
+            value = python_value_to_proto_value(row[idx],feature.dtype)
             feature_dict[feature.name] = value
         event_timestamp_idx = table.column_names.index(
             feature_view.input.event_timestamp_column
--- a/sdk/python/feast/infra/provider.py
+++ b/sdk/python/feast/infra/provider.py
@@ -257,7 +257,7 @@ def _convert_arrow_to_proto(
         feature_dict = {}
         for feature in feature_view.features:
             idx = table.column_names.index(feature.name)
-            value = python_value_to_proto_value(row[idx],feature.dtype)
+            value = python_value_to_proto_value(row[idx], feature.dtype)
             feature_dict[feature.name] = value
         event_timestamp_idx = table.column_names.index(
             feature_view.input.event_timestamp_column
--- a/sdk/python/feast/type_map.py
+++ b/sdk/python/feast/type_map.py
@@ -313,8 +313,8 @@ def _python_value_to_proto_value(feast_value_type, value) -> ProtoValue:
     raise Exception(f"Unsupported data type: ${str(type(value))}")
 
 
-def python_value_to_proto_value(value: Any) -> ProtoValue:
-    value_type = python_type_to_feast_value_type("", value)
+def python_value_to_proto_value(value: Any, feature_type: ValueType = None) -> ProtoValue:
+    value_type = python_type_to_feast_value_type("", value) if value else feature_type
     return _python_value_to_proto_value(value_type, value)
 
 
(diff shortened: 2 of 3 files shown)


woop:
Hey @qooba. Thanks again for the PR. This change seems reasonable. FYI you can run `make format` and `make lint` to format your code and test whether linting is passing.

woop:
@qooba would you mind updating one of our tests so that we explicitly test for this condition?

qooba:
Sure :) @woop I have added checks to existing tests.
Please notice that pandas during parquet read/write converts None to NaN which itself causes inconsistency.
Thus I have to add conversion in test:
```
df = df.where(pd.notnull(df), None)
```

I'm not sure if such conversion should be added to feast itself (on the other hand different offline stores can behave differently and it should be consistent).

jklegar (commented):
Hey @qooba, this is great. Good catch on the None to NaN conversion that happens with pandas->parquet->pandas. Since we want results from the online and offline stores to be consistent, I think we should have the `get_historical_features` call just return Nones rather than converting NaNs to Nones within the test, i.e., we'll want to convert in `sdk/python/feast/infra/offline_stores/file.py`. I think for the BigQuery offline store we're ok since BigQuery nulls will turn into pandas Nones. With that change this should be good to go, thanks for the contribution!

qooba:
Hi, @jklegar . I have checked native pandas methods ```df.where(pd.notnull(df), None)``` and ```df.where((pd.notnull(df)), None).replace({np.inf: None})``` but both are changing the value dtype which introduces another inconsistency and breaks other tests. I think that using custom written methods rather than native pandas will affect historical fetch performance thus I think it is better to leave this and add appropriate comment in the documentation (anyway pandas users also live with this :) ). What do you think, is that solution is OK ?

jklegar (approved):
Cool it seems like that's the best thing to go with right now then. Thanks @qooba!

PR #1483, opened 2021-04-21
Add support for DynamoDB and S3 registry

### LINKED ISSUE #1409: DynamoDB support for online feature store
We need to have an online serving backend with Dynamo to support customer deployments on AWS.

Implementation of `Provider` similar to [/sdk/python/feast/infra/local_sqlite.py](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/local_sqlite.py) with DynamoDB backend

leonid133:
We need to have an online serving backend with Dynamo to support customer deployments on AWS.

Fixes #1409

NONE

```release-note
* Add support for S3 registry
* Add aws provider with Dynamo support
```


sdk/python/setup.py
```diff
@@ -61,6 +61,7 @@
     "tabulate==0.8.*",
     "toml==0.10.*",
     "tqdm==4.*",
+    "boto3==1.17.54",
```
woop, line 64: We should probably pull out the GCP and AWS dependencies into `extras`

sdk/python/feast/templates/aws_dynamo/feature_store.yaml
```diff
@@ -0,0 +1,3 @@
+project: my_project
+registry: s3://feast-provectus-demo/data/registry.db
```
woop, line 2: Missed this I think.

sdk/python/feast/registry.py
```diff
@@ -477,3 +479,73 @@ def _write_registry(self, registry_proto: RegistryProto):
...
+        s3 = boto3.resource("s3")
+        try:
+            bucket = s3.Bucket(self._bucket)
+            s3.meta.client.head_bucket(Bucket=bucket.name)
+        except botocore.client.ClientError as e:
+            # If a client error is thrown, then check that it was a 404 error.
+            # If it was a 404 error, then the bucket does not exist.
+            error_code = int(e.response["Error"]["Code"])
+            if error_code == 404:
+                raise Exception(
```
woop, line 506: We should probably have a custom exception. See https://github.com/feast-dev/feast/blob/master/sdk/python/feast/errors.py

sdk/python/feast/cli.py
```diff
@@ -236,7 +236,7 @@ def materialize_incremental_command(end_ts: str, views: List[str]):
 @click.option(
     "--template",
     "-t",
-    type=click.Choice(["local", "gcp"], case_sensitive=False),
+    type=click.Choice(["local", "gcp", "aws_dynamo"], case_sensitive=False),
```
woop, line 239: I think we can just call it `aws` for now. Dynamo can be an implicit default.

woop:
Thanks for this @leonid133!

woop:
Also, we certainly should add the data model for Dynamo here: https://github.com/feast-dev/feast/tree/master/docs/specs

sdk/python/setup.py
woop: Also, the version shouldn't be pinned.

sdk/python/feast/infra/aws_dynamo_provider.py
```diff
@@ -0,0 +1,201 @@
...
+    _get_column_names,
+    _run_field_mapping,
+)
+from feast.protos.feast.types.EntityKey_pb2 import EntityKey as EntityKeyProto
+from feast.protos.feast.types.Value_pb2 import Value as ValueProto
+from feast.registry import Registry
+from feast.repo_config import DynamoOnlineStoreConfig, RepoConfig
+
+
+class AwsDynamoProvider(Provider):
```
YikSanChan, line 26: shall we name this AwsDynamodbProvider instead? [Dynamo](https://www.allthingsdistributed.com/2007/10/amazons_dynamo.html) is not exactly the same as Dynamo DB

sdk/python/feast/repo_config.py
```diff
@@ -100,18 +109,20 @@ def _validate_online_store_config(cls, values):
...
                 elif values["provider"] == "gcp":
                     values["online_store"]["type"] = "datastore"
-
+                elif values["provider"] == "aws_dynamo":
+                    values["online_store"]["type"] = "dynamo"
             online_store_type = values["online_store"]["type"]
-
             # Make sure the user hasn't provided the wrong type
-            assert online_store_type in ["datastore", "sqlite"]
+            assert online_store_type in ["datastore", "sqlite", "dynamo"]
```
blvp, line 116: I believe we can also modify it to dynamodb to distinguish between dynamo and dynamodb.

sdk/python/feast/registry.py
blvp: Should we also add similar for GCP as well or it would be better to open a new ticket?

sdk/python/feast/infra/aws_dynamo_provider.py
```diff
@@ -0,0 +1,201 @@
...
+                for feature_name, value_bin in value["values"].items():
+                    val = ValueProto()
+                    val.ParseFromString(value_bin.value)
+                    res[feature_name] = val
+                result.append((value["event_ts"], res))
+            else:
+                result.append((None, None))
+        return result
+
+    def materialize_single_feature_view(
```
blvp, line 137: @woop This method was duplicated from GCP, should we make abstract Provider because this part is very well generalized already or this is fine to keep it as is for now and then refactor?

sdk/python/feast/infra/aws_dynamo_provider.py
```diff
@@ -0,0 +1,201 @@
...
+                        {"AttributeName": "Row", "KeyType": "HASH"},
+                        {"AttributeName": "Project", "KeyType": "RANGE"},
+                    ],
+                    AttributeDefinitions=[
+                        {"AttributeName": "Row", "AttributeType": "S"},
+                        {"AttributeName": "Project", "AttributeType": "S"},
+                    ],
+                    ProvisionedThroughput={
+                        "ReadCapacityUnits": 5,
+                        "WriteCapacityUnits": 5,
```
blvp, line 57: Probably it would be good to set this number as a parameter of dynamodb online store.

woop:
Also, we will definitely need to have integrations tests with DynamoDB in order to have confidence in this code. We're happy to use our own AWS account, just let me know when you're ready the code and I can help set things up.

sdk/python/feast/registry.py
woop: New PR

sdk/python/feast/registry.py
```diff
@@ -495,3 +499,72 @@ def _write_registry(self, registry_proto: RegistryProto):
         file_obj.seek(0)
         blob.upload_from_file(file_obj)
         return
+
+
+class AwsS3RegistryStore(RegistryStore):
```
tedhtchang, line 504: S3 is a standard protocol. Should we call it `S3RegistryStore` instead ?

sdk/python/feast/templates/aws_dynamo/feature_store.yaml
leonid133: yep

sdk/python/feast/cli.py
woop: Sorry about this, but I think what you had here was correct. Would you mind using `aws_dynamo` until we have Redshift support?

jklegar (commented):
We'll want integration tests to make sure these methods work in the same way as their GCP equivalents and don't rot in the future - I think it should be good enough to add a `sdk/python/tests/test_cli_aws.py` that is the equivalent of `sdk/python/tests/test_cli_gcp.py`, and a pytest fixture `feature_store_with_s3_registry` analogous to `feature_store_with_gcp_registry` in `sdk/python/tests/test_feature_store.py` and used in the same tests in that file. I don't think we've set up AWS authentication in GitHub actions yet so I'll work on getting that set up and we can just run these integration tests locally in the meantime.

jklegar:
FYI all our integration tests use pytest so to run integration tests locally you can just run `pytest sdk/python` for all local tests and `pytest sdk/python --integration` for all tests including those with external dependencies (e.g. GCP resources). This'll run the same tests that are run in CI on GitHub actions

sdk/python/feast/templates/aws/test.py
```diff
@@ -0,0 +1,41 @@
...
+    feature_refs = ["driver_hourly_stats:conv_rate", "driver_hourly_stats:acc_rate"]
+
+    #     print()
+    print("Loading features into the online store...")
+    fs.materialize_incremental(end_date=datetime.now())
+
+    #     print()
+    print("Retrieving online features...")
+
+    #     Retrieve features from the online store (Firestore)
```
tedhtchang, line 30: replace the Firestore with Dynamodb?

sdk/python/setup.py
```diff
@@ -61,6 +61,7 @@
     "tabulate==0.8.*",
     "toml==0.10.*",
     "tqdm==4.*",
+    "boto3==1.17.*",
```
woop, line 64: If this dependency is aws specific then it needs to be in its own `extras` object so that it isnt installed by default

sdk/python/setup.py
```diff
@@ -100,6 +100,10 @@
     "google-cloud-core==1.4.*",
 ]
```
tedhtchang, line 101:
```suggestion
    "boto3==1.17.*"
]    
```

sdk/python/feast/infra/aws_dynamodb_provider.py
```diff
@@ -0,0 +1,232 @@
+import os
+from datetime import datetime
+from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
+
+import boto3
+import mmh3
+import pandas
+from botocore.exceptions import ClientError
```
tedhtchang, line 8:
```suggestion
try:
    import boto3
    from botocore.exceptions import ClientError
except ImportError as e:
    from feast.errors import FeastExtrasDependencyImportError
    raise FeastExtrasDependencyImportError("aws", str(e))
```
perhaps adapt this pattern https://github.com/feast-dev/feast/blob/73ac3de613a729da41b7c832bb4a520132399339/sdk/python/feast/infra/gcp.py#L29-L35

sdk/python/feast/infra/aws_dynamodb_provider.py
```diff
@@ -0,0 +1,232 @@
+import os
+from datetime import datetime
+from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
+
+import boto3
```
tedhtchang, line 5:
```suggestion
```

sdk/python/feast/registry.py
```diff
@@ -490,3 +495,72 @@ def _write_registry(self, registry_proto: RegistryProto):
...
         blob.upload_from_file(file_obj)
         return
+
+
+class S3RegistryStore(RegistryStore):
+    def __init__(self, uri: str):
+        self._uri = urlparse(uri)
+        self._bucket = self._uri.hostname
+        self._key = self._uri.path.lstrip("/")
+        return
```
tedhtchang, line 505:
```suggestion
        try:
            import boto3
            import botocore
        except ImportError as e:
            from feast.errors import FeastExtrasDependencyImportError
            raise FeastExtrasDependencyImportError("aws", str(e))
        self.s3 = boto3.resource("s3", endpoint_url=os.environ.get("FEAST_S3_ENDPOINT_URL"))
        return
```
perhaps adapt this pattern https://github.com/feast-dev/feast/blob/73ac3de613a729da41b7c832bb4a520132399339/sdk/python/feast/registry.py#L433-L440

sdk/python/feast/registry.py
```diff
@@ -490,3 +495,72 @@ def _write_registry(self, registry_proto: RegistryProto):
...
+
+    def get_registry_proto(self):
+        import boto3
+        import botocore
```
tedhtchang, line 509:
```suggestion
```

sdk/python/feast/registry.py
```diff
@@ -490,3 +495,72 @@ def _write_registry(self, registry_proto: RegistryProto):
...
+
+        file_obj = TemporaryFile()
+        registry_proto = RegistryProto()
+        s3 = boto3.resource("s3", endpoint_url=os.environ.get("FEAST_S3_ENDPOINT_URL"))
```
tedhtchang, line 513:
```suggestion
```

sdk/python/feast/registry.py
```diff
@@ -490,3 +495,72 @@ def _write_registry(self, registry_proto: RegistryProto):
...
+        try:
+            bucket = s3.Bucket(self._bucket)
```
tedhtchang, line 515:
```suggestion
            bucket = self.s3.Bucket(self._bucket)
```

sdk/python/feast/registry.py
```diff
@@ -490,3 +495,72 @@ def _write_registry(self, registry_proto: RegistryProto):
...
+            s3.meta.client.head_bucket(Bucket=bucket.name)
```
tedhtchang, line 516:
```suggestion
            self.s3.meta.client.head_bucket(Bucket=bucket.name)
```

sdk/python/feast/registry.py
```diff
@@ -490,3 +495,72 @@ def _write_registry(self, registry_proto: RegistryProto):
...
+        except FileNotFoundError:
+            registry_proto = RegistryProto()
+            registry_proto.registry_schema_version = REGISTRY_SCHEMA_VERSION
+        if updater:
+            registry_proto = updater(registry_proto)
+        self._write_registry(registry_proto)
+        return
+
+    def _write_registry(self, registry_proto: RegistryProto):
+        import boto3
```
tedhtchang, line 556:
```suggestion
```

sdk/python/feast/registry.py
```diff
@@ -490,3 +495,72 @@ def _write_registry(self, registry_proto: RegistryProto):
...
+
+        registry_proto.version_id = str(uuid.uuid4())
+        registry_proto.last_updated.FromDatetime(datetime.utcnow())
+        # we have already checked the bucket exists so no need to do it again
+        file_obj = TemporaryFile()
+        file_obj.write(registry_proto.SerializeToString())
+        file_obj.seek(0)
+        s3 = boto3.resource("s3", endpoint_url=os.environ.get("FEAST_S3_ENDPOINT_URL"))
```
tedhtchang, line 564:
```suggestion
```

sdk/python/feast/registry.py
```diff
@@ -490,3 +495,72 @@ def _write_registry(self, registry_proto: RegistryProto):
...
+        s3.put_object(Bucket=self._bucket, Body=file_obj, Key=self._key)
```
tedhtchang, line 565:
```suggestion
        self.s3.put_object(Bucket=self._bucket, Body=file_obj, Key=self._key)
```

sdk/python/feast/registry.py
tedhtchang, line 516:
```suggestion
            bucket = self.s3.Bucket(self._bucket)
            self.s3.meta.client.head_bucket(Bucket=bucket.name)
```

sdk/python/feast/infra/aws_dynamodb_provider.py
woop: wow @tedhtchang, you're on the money :)

sdk/python/feast/infra/aws_dynamodb_provider.py
```diff
@@ -0,0 +1,232 @@
...
+    _run_field_mapping,
+)
+from feast.protos.feast.types.EntityKey_pb2 import EntityKey as EntityKeyProto
+from feast.protos.feast.types.Value_pb2 import Value as ValueProto
+from feast.registry import Registry
+from feast.repo_config import DynamoDbOnlineStoreConfig, RepoConfig
+
+
+class AwsDynamodbProvider(Provider):
+    _wcu: int
```
woop, line 30: Can we just spell these out instead of abbreviating them?

sdk/python/feast/infra/aws_dynamodb_provider.py
```diff
@@ -0,0 +1,232 @@
...
+    _rcu: int
+
+    def __init__(self, config: RepoConfig):
+        assert isinstance(config.online_store, DynamoDbOnlineStoreConfig)
+        if config and config.online_store and config.online_store.rcu:
+            self._rcu = config.online_store.rcu
+        else:
+            self._rcu = 5
```
woop, line 38: isn't a default already set in `DynamoDbOnlineStoreConfig`?

sdk/python/feast/infra/aws_dynamodb_provider.py
```diff
@@ -0,0 +1,232 @@
...
+        if config and config.online_store and config.online_store.wcu:
+            self._wcu = config.online_store.wcu
+        else:
+            self._wcu = 5
+
+        if config and config.online_store and config.online_store.region_name:
+            self.region_name = config.online_store.region_name
+            os.environ["AWS_DEFAULT_REGION"] = self.region_name
+        else:
+            self.region_name = os.environ.get("AWS_DEFAULT_REGION", "us-west-2")
```
woop, line 49: Can we somehow pull out this configuration `us-west-2` from the code here?

sdk/python/feast/infra/aws_dynamo_provider.py
woop: Can we do that in a separate PR perhaps? I think we may need to do a code cleanup at some point for generalized functionality.

woop:
@leonid133 can you please rebase on the latest commit on master?

sdk/python/feast/infra/aws_dynamodb_provider.py
```diff
@@ -0,0 +1,232 @@
...
+                        {"AttributeName": "Project", "KeyType": "RANGE"},
+                    ],
+                    AttributeDefinitions=[
+                        {"AttributeName": "Row", "AttributeType": "S"},
+                        {"AttributeName": "Project", "AttributeType": "S"},
+                    ],
+                    ProvisionedThroughput={
+                        "ReadCapacityUnits": self._rcu,
+                        "WriteCapacityUnits": self._wcu,
+                    },
```
oavdeev, line 82: I'd remove these and just use pay-per-request mode here so that users have fewer configuration options to care about.

We could leave provisioned throughput mode as an advanced option, but my hunch is that even this is a stretch. I suspect it would very rarely make sense cost wise, given how bursty Feast write workloads are.

sdk/python/feast/registry.py
guykhazma: @leonid133 does:
```
s3.put_object(Bucket=self._bucket, Body=file_obj, Key=self._key)
```
work for you?
The code is using the `resource` API and not the `client` API so it fails with this error:
```
self.s3.put_object(Bucket=self._bucket, Body=file_obj, Key=self._key)
AttributeError: 's3.ServiceResource' object has no attribute 'put_object'
```

I think the syntax should be:
```
self.s3.Bucket(self._bucket).put_object(Body=file_obj, Key=self._key)
```

or else use the `client` API

sdk/python/feast/registry.py
```diff
@@ -537,3 +542,80 @@ def _write_registry(self, registry_proto: RegistryProto):
...
+        self.s3 = boto3.resource(
+            "s3", endpoint_url=os.environ.get("FEAST_S3_ENDPOINT_URL")
+        )
+        return
+
+    def get_registry_proto(self):
+        file_obj = TemporaryFile()
+        registry_proto = RegistryProto()
+        try:
+            import botocore
```
tsotnet, line 567: Let's do the try import of botocore in the __init__ together with boto3. Earlier we raise the exception on this, the better. Then here, you'll just be able to write `import botocore` without try/catch.

sdk/python/feast/registry.py
```diff
@@ -537,3 +542,80 @@ def _write_registry(self, registry_proto: RegistryProto):
...
+            return registry_proto
+        except botocore.exceptions.ClientError as e:
+            if e.response["Error"]["Code"] == "404":
+                raise FileNotFoundError(
+                    f'Registry not found at path "{self._uri.geturl()}". Have you run "feast apply"?'
+                )
+            else:
+                raise FileNotFoundError(
+                    f'Registry is not able to locate data under path "{self._uri.geturl()}" with [original error]: {e.response}'
+                )
```
tsotnet, line 598: Can you explain what's the difference between "Registry not found at path" and "Registry is not able to locate data under path"? Should we just merge these two and raise the 1st exception?

Also, let's do `raise FileNotFoundError(...) from e` so that the original exception is maintained with its own stack trace. Then you won't have to pass `[original error]` here.

sdk/python/feast/registry.py
```diff
@@ -537,3 +542,80 @@ def _write_registry(self, registry_proto: RegistryProto):
...
+    ):
+        try:
+            registry_proto = self.get_registry_proto()
+        except FileNotFoundError:
+            registry_proto = RegistryProto()
+            registry_proto.registry_schema_version = REGISTRY_SCHEMA_VERSION
+        if updater:
+            registry_proto = updater(registry_proto)
+        self._write_registry(registry_proto)
+        return
```
tsotnet, line 611:
```suggestion
```

sdk/python/feast/registry.py
```diff
@@ -537,3 +542,80 @@ def _write_registry(self, registry_proto: RegistryProto):
...
+
+    def _write_registry(self, registry_proto: RegistryProto):
+        registry_proto.version_id = str(uuid.uuid4())
+        registry_proto.last_updated.FromDatetime(datetime.utcnow())
+        # we have already checked the bucket exists so no need to do it again
+        file_obj = TemporaryFile()
+        file_obj.write(registry_proto.SerializeToString())
+        file_obj.seek(0)
+        self.s3.Bucket(self._bucket).put_object(Body=file_obj, Key=self._key)
+        return
```
tsotnet, line 621:
```suggestion
```

sdk/python/tests/test_feature_store.py
```diff
@@ -72,6 +72,24 @@ def feature_store_with_gcs_registry():
     )
 
 
+@pytest.fixture
+def feature_store_with_s3_registry():
```
tsotnet, line 76: Can you add test cases with this fixture? See how other fixtures are used in this file.

sdk/python/tests/test_feature_store.py
```diff
@@ -72,6 +72,24 @@ def feature_store_with_gcs_registry():
...
+    import boto3
+
+    s3 = boto3.resource("s3")
+    bucket_name = f"feast-registry-test-{int(time.time() * 1000)}"
```
tsotnet, line 80: Let's use `feast-integration-tests` bucket, which I have already created in Feast's AWS account for integration tests.

sdk/python/tests/test_feature_store.py
```diff
@@ -72,6 +72,24 @@ def feature_store_with_gcs_registry():
...
+    bucket = s3.Bucket(bucket_name)
+    s3.meta.client.head_bucket(Bucket=bucket.name)
+
+    return FeatureStore(
+        config=RepoConfig(
+            registry=f"s3://{bucket_name}/registry.db",
```
tsotnet, line 86: Let's change the path to `s3://{bucket_name}/registries/{int(time.time() * 1000)}/registry.db`

sdk/python/tests/test_cli_aws.py
```diff
@@ -0,0 +1,59 @@
...
+        repo_config.write_text(
+            dedent(
+                f"""
+            project: {project_id}
+            registry: {data_path / "registry.db"}
+            provider: aws
+            offline_store:
+                type: file
+            online_store:
+                type: dynamodb
```
tsotnet, line 36: Let's remove this. These are defaults.
```suggestion
```

sdk/python/feast/templates/aws/feature_store.yaml
```diff
@@ -0,0 +1,3 @@
+project: my_project
+registry: s3://data/registry.db
```
tsotnet, line 2: For now let's use a local registry in the template, since the bucket here is hardcoded and we haven't thought through the user experience on this.

sdk/python/feast/templates/aws/test.py
```diff
@@ -0,0 +1,41 @@
...
+    fs = FeatureStore(repo_path=".")
+
+    # Deploy the feature store to AWS
+    print("Deploying feature store to AWS...")
+    fs.apply([driver, driver_hourly_stats_view])
+
+    # Select features
+    feature_refs = ["driver_hourly_stats:conv_rate", "driver_hourly_stats:acc_rate"]
+
+    #     print()
```
tsotnet, line 23:
```suggestion
```

sdk/python/feast/templates/aws/test.py
```diff
@@ -0,0 +1,41 @@
...
+    print("Loading features into the online store...")
+    fs.materialize_incremental(end_date=datetime.now())
+
+    #     print()
```
tsotnet, line 27:
```suggestion
```

sdk/python/feast/templates/aws/test.py
```diff
@@ -0,0 +1,41 @@
...
+    print("Retrieving online features...")
+
+    #     Retrieve features from the online store (Dynamodb)
```
tsotnet, line 30:
```suggestion
    # Retrieve features from the online store (DynamoDB)
```

sdk/python/feast/templates/aws/bootstrap.py
```diff
@@ -0,0 +1,35 @@
...
+    end_date = datetime.now().replace(microsecond=0, second=0, minute=0)
+    start_date = end_date - timedelta(days=15)
+
+    driver_entities = [1001, 1002, 1003, 1004, 1005]
+    driver_df = create_driver_hourly_stats_df(driver_entities, start_date, end_date)
+
+    driver_stats_path = data_path / "driver_stats.parquet"
+    driver_df.to_parquet(path=str(driver_stats_path), allow_truncated_timestamps=True)
+
+    example_py_file = repo_path / "driver_example.py"
```
tsotnet, line 22: can you rename `driver_example.py` to `example.py`? Looks like this template is copied from the local one, which has `example.py` instead. I'd rather not diverge unnecessarily.

tsotnet (changes requested):
Can you also fix the image please? (can't leave the comment on binary files)
1. You will need to update the field names as you change them.
2. Also, in the top right box "values:" and "acc_rate:" are on the same line incorrectly.
3. The table name will need to be updated as well

tsotnet:
One more comment I forgot: could you also add the integration test for Dynamo? (a test that actually reads & writes to it). You can check out how Redis & Datastore tests work in https://github.com/feast-dev/feast/blob/master/sdk/python/tests/test_offline_online_store_consistency.py#L149

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -0,0 +1,186 @@
...
+    ) -> None:
+        online_config = config.online_store
+        assert isinstance(online_config, DynamoDbOnlineStoreConfig)
+        dynamodb = self._initialize_dynamodb(online_config)
+
+        table_instance = dynamodb.Table(table.name)
+        with table_instance.batch_writer() as batch:
+            for entity_key, features, timestamp, created_ts in data:
+                document_id = compute_datastore_entity_id(entity_key)  # TODO check id
+                # TODO compression encoding
```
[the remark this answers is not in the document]
blvp: It is about hashing of the field names in json, to reduce memory footprint of the RW queries since billing of DynamoDB is tight to a document size.

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -0,0 +1,186 @@
...
+        dynamodb = self._initialize_dynamodb(online_config)
+
+        result: List[Tuple[Optional[datetime], Optional[Dict[str, ValueProto]]]] = []
+        for entity_key in entity_keys:
+            table_instace = dynamodb.Table(table.name)
+            document_id = compute_datastore_entity_id(entity_key)  # TODO check id
+            response = table_instace.get_item(
+                Key={"Row": document_id, "Project": config.project}
+            )
+            value = response["Item"]
```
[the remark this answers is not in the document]
blvp: Should this raise an error instead for one entity key or just prune it? Since this method is for a batch of entity_id, what is the desired behavior for handling errors for each entity?

sdk/python/feast/registry.py
blvp: It could be possible to get 403 error or even 500 from AWS, I think it does make sense to re-raise exception in case  in the else branch

sdk/python/feast/registry.py
blvp: looks like linter is adding this, need to double check with linter rules.

sdk/python/feast/registry.py
blvp: the same could be applied to :582 line similarly.

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -0,0 +1,186 @@
...
+                batch.put_item(
+                    Item={
+                        "Row": document_id,  # PartitionKey
+                        "Project": config.project,  # SortKey
+                        "event_ts": str(utils.make_tzaware(timestamp)),
+                        "values": {
+                            k: v.SerializeToString()
+                            for k, v in features.items()  # Serialized Features
+                        },
+                    }
```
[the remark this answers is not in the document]
blvp: @woop Could you please leave some comments on the proposed design for "values" besides that I don't see any concern to not modify this implementation.

tsotnet:
Oh I also noticed that https://github.com/feast-dev/feast/blob/master/docs/specs/online_store_format.md will need to be updated. This is not as urgent as the rest of the code-related fixes though.

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -14,12 +14,11 @@
 from datetime import datetime
```
tsotnet, line 14: Can you change the rest of `table_name` to `table_instance` (where `table_name` is actually an instance of DynamoDB Table object)?

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -152,17 +143,15 @@ def online_read(
...
 
         result: List[Tuple[Optional[datetime], Optional[Dict[str, ValueProto]]]] = []
         for entity_key in entity_keys:
-            table_instace = dynamodb.Table(table.name)
-            document_id = compute_datastore_entity_id(entity_key)  # TODO check id
-            response = table_instace.get_item(
-                Key={"Row": document_id, "Project": config.project}
-            )
-            value = response["Item"]
+            table_instace = dynamodb.Table(f"{config.project}.{table.name}")
```
tsotnet, line 151:
```suggestion
            table_instance = dynamodb.Table(f"{config.project}.{table.name}")
```

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -152,17 +143,15 @@ def online_read(
...
+            document_id = compute_entity_id(entity_key)  # TODO check id
```
tsotnet, line 152:
```suggestion
            entity_id = compute_entity_id(entity_key)
```

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -152,17 +143,15 @@ def online_read(
...
+            response = table_instace.get_item(Key={"Row": document_id})
+            value = response.get("Item", None)
```
tsotnet, line 154: nit: `None` isn't necessary. The default value is `None` anyways.

sdk/python/feast/infra/online_stores/helpers.py
```diff
@@ -53,3 +54,13 @@ def _mmh3(key: str):
...
     key_hash = mmh3.hash(key, signed=False)
     return bytes.fromhex(struct.pack("<Q", key_hash).hex()[:8])
+
+
+def compute_entity_id(entity_key: EntityKeyProto) -> str:
+    """
+    Compute Datastore Entity id given Feast Entity Key.
+
+    Remember that Datastore Entity is a concept from the Datastore data model, that has nothing to
+    do with the Entity concept we have in Feast.
```
tsotnet, line 64: Remove Datastore from the docstring.
```suggestion
    Compute Entity id given Feast Entity Key for online stores.

    Remember that Entity here refers to `EntityKeyProto` which is used in some online stores to encode the keys.
    It has nothing to do with the Entity concept we have in Feast.
```

sdk/python/feast/registry.py
```diff
@@ -546,16 +546,17 @@ def _write_registry(self, registry_proto: RegistryProto):
...
             raise FeastExtrasDependencyImportError("aws", str(e))
-        self.s3 = boto3.resource(
+        self._uri = urlparse(uri)
+        self._bucket = self._uri.hostname
+        self._key = self._uri.path.lstrip("/")
+
+        self.s3_client = boto3.resource(
             "s3", endpoint_url=os.environ.get("FEAST_S3_ENDPOINT_URL")
         )
         return
```
tsotnet, line 562: nit: get rid of the return call
```suggestion
```

sdk/python/feast/registry.py
tsotnet: This comment still stands ^

sdk/python/feast/registry.py
tsotnet: The issue isn't that AWS could return other error code. It's that messages "Registry not found at path" and "Registry is not able to locate data under path" are basically telling the same thing. Maybe instead of "Registry is not able to locate data under path" we can say "Error while trying to locate Registry at path"?

tsotnet (changes requested):
Looks like there are a couple of unresolved conversations from the previous review. Some of them I commented again, some of them I may have missed. Could you double-check that previous conversations are closed? (e.g. on file `test_feature_store.py`). Thanks!

sdk/python/feast/infra/online_stores/dynamodb.py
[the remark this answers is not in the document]
woop: > Since features is already a proto object, let's just pass the binary encoded proto object to values

sgtm

sdk/python/feast/infra/online_stores/dynamodb.py
leonid133: done

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -34,25 +33,24 @@
...
     """Online store type selector"""
 
-    rcu: Optional[PositiveInt] = 5
-    """ Read capacity unit """
-
-    wcu: Optional[PositiveInt] = 5
-    """ Write capacity unit """
-
-    region_name: Optional[StrictStr] = None
+    region: Optional[StrictStr] = None
```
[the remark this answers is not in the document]
leonid133: which region should be used for integration test?

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -34,25 +33,24 @@
...
     """ AWS Region Name """
 
 
-class DynamoDbOnlineStore(OnlineStore):
-    def _initialize_dynamodb(self, online_config: DynamoDbOnlineStoreConfig):
-        return boto3.resource("dynamodb", region_name=online_config.region_name)
+class DynamoDBOnlineStore(OnlineStore):
+    """
+    OnlineStore is an object used for all interaction between Feast and the service used for offline storage of
+    features.
```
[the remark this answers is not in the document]
leonid133: done

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -34,25 +33,24 @@
...
+    """
+
+    def _initialize_dynamodb(self, online_config: DynamoDBOnlineStoreConfig):
+        return boto3.resource("dynamodb", region_name=online_config.region)
```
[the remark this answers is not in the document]
leonid133: Looks like it is already cached in session 
[boto3 L87](https://github.com/boto/boto3/blob/343e568d5cb16606b1bda7ea5ee87a3611d100aa/boto3/__init__.py#L87)
[boto3 L73](https://github.com/boto/boto3/blob/343e568d5cb16606b1bda7ea5ee87a3611d100aa/boto3/__init__.py#L73)

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -64,37 +62,33 @@ def update(
         partial: bool,
     ):
         online_config = config.online_store
-        assert isinstance(online_config, DynamoDbOnlineStoreConfig)
+        assert isinstance(online_config, DynamoDBOnlineStoreConfig)
         dynamodb = self._initialize_dynamodb(online_config)
 
+        client = boto3.client("dynamodb")
```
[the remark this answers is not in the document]
leonid133: ok, I'll move np, but it cached  in session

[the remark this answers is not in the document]
leonid133: done

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -124,25 +115,25 @@ def online_write_batch(
...
-        table_instance = dynamodb.Table(table.name)
+        table_instance = dynamodb.Table(f"{config.project}.{table.name}")
         with table_instance.batch_writer() as batch:
             for entity_key, features, timestamp, created_ts in data:
-                document_id = compute_datastore_entity_id(entity_key)  # TODO check id
-                # TODO compression encoding
+                document_id = compute_entity_id(entity_key)
                 batch.put_item(
                     Item={
                         "Row": document_id,  # PartitionKey
```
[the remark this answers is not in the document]
leonid133: done

sdk/python/feast/infra/online_stores/dynamodb.py
leonid133: done

sdk/python/feast/infra/online_stores/dynamodb.py
leonid133: done

sdk/python/feast/infra/online_stores/helpers.py
leonid133: done

sdk/python/feast/registry.py
leonid133: done

sdk/python/feast/registry.py
leonid133: still the same answer
```
feast/registry.py:551:13: F401 'botocore' imported but unused
feast/registry.py:570:16: F821 undefined name 'botocore'
feast/registry.py:585:16: F821 undefined name 'botocore'
make: *** [lint-python] Error 1
```

sdk/python/feast/registry.py
leonid133: done

sdk/python/setup.py
leonid133: done

sdk/python/tests/test_cli_aws.py
leonid133: done

sdk/python/tests/test_feature_store.py
leonid133: done

sdk/python/tests/test_feature_store.py
leonid133: done

sdk/python/tests/test_feature_store.py
leonid133: done

sdk/python/feast/infra/online_stores/dynamodb.py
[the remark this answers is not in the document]
tsotnet: Let's use us-west-2

sdk/python/feast/registry.py
tsotnet: I see, that makes sense then, forgot this comment.

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -0,0 +1,186 @@
...
+    """Online store type selector"""
+
+    rcu: Optional[PositiveInt] = 5
+    """ Read capacity unit """
+
+    wcu: Optional[PositiveInt] = 5
+    """ Write capacity unit """
+
+    region_name: Optional[StrictStr] = None
+    """ AWS Region Name """
```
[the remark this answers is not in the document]
leonid133: done

sdk/python/feast/infra/online_stores/dynamodb.py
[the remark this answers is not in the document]
leonid133: done

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -64,37 +62,33 @@ def update(
...
-                )
-                table.meta.client.get_waiter("table_exists").wait(
-                    TableName=table_name.name
+                    BillingMode="PAY_PER_REQUEST",
                 )
             except ClientError as ce:
                 print(ce)
                 if ce.response["Error"]["Code"] == "ResourceNotFoundException":
-                    table = dynamodb.Table(table_name.name)
+                    table = dynamodb.Table(f"{config.project}.{table_name.name}")
```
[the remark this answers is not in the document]
leonid133: done

sdk/python/feast/infra/online_stores/dynamodb.py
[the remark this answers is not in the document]
tsotnet: Nice, didn't know that!

sdk/python/feast/infra/online_stores/dynamodb.py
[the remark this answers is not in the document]
tsotnet: If there's no Item field, it just means that the entity with this id was not found in the online store. In other words it was never materialized. This isn't something we should raise an exception on, since it's expected. In feature_store.py this is handled by assigning the appropriate FieldStatus to the response (FieldStatus.NOT_FOUND): https://github.com/feast-dev/feast/blob/master/sdk/python/feast/feature_store.py#L571

tsotnet (approved):

woop:
Huge. Thanks @leonid133 @blvp @tsotnet !

PR #1488, opened 2021-04-21
Fix bug in allowing empty repositories to be applied to a GCS registry

woop:

We introduced a bug with #1477. We allow initialization of the registry during `apply`, even with empty repositories. However, support was only added for the local registry, not the GCS registry.

```release-note
Fix bug in allowing empty repositories to be applied to a GCS registry
```

### DIFF
--- a/sdk/python/feast/registry.py
+++ b/sdk/python/feast/registry.py
@@ -376,7 +376,7 @@ class RegistryStore(ABC):
 
     @abstractmethod
     def update_registry_proto(
-        self, updater: Optional[Callable[[RegistryProto], RegistryProto]]
+        self, updater: Optional[Callable[[RegistryProto], RegistryProto]] = None
     ):
         """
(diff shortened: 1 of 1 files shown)


jklegar (approved):

sdk/python/feast/registry.py
```diff
@@ -71,7 +69,7 @@ def __init__(self, registry_path: str, repo_path: Path, cache_ttl: timedelta):
 
     def _initialize_registry(self):
         """Explicitly forces the initialization of a registry"""
-        self._registry_store.update_registry_proto()
+        self._registry_store.update_registry_proto(updater=None)
```
tsotnet, line 72: nit: you don't need `updater=None` since you added the default

woop: Yea, this was a relic from my debugging. Good catch.

PR #1489, opened 2021-04-21
Update broken urls in contributing.md

tedhtchang:
Signed-off-by: ted chang <htchang@us.ibm.com>

These urls are broken


woop:
Thanks @tedhtchang, will update these shortly. It may be easier to do from our side since it spans branches.

PR #1498, opened 2021-04-23
Pass entities information to Provider

MattDelac:
Some Providers would need the information about the entities to keep and to delete.
Therefore this PR aims to pass this information to the Provider object.

Fixes #

Not really as even if a user created a custom Provider, they could not use it (yet) because of the hard coded logic happening at the `get_provider` function [see reference](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/provider.py#L127-L137))
```release-note
We now pass the information about the entities to the Provider object
```

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -217,12 +217,14 @@ class FeatureStore:
         if isinstance(objects, Entity) or isinstance(objects, FeatureView):
             objects = [objects]
         views_to_update = []
+        entities_to_update = []
         for ob in objects:
             if isinstance(ob, FeatureView):
                 self._registry.apply_feature_view(ob, project=self.project)
                 views_to_update.append(ob)
             elif isinstance(ob, Entity):
                 self._registry.apply_entity(ob, project=self.project)
+                entities_to_update.append(ob)
             else:
                 raise ValueError(
                     f"Unknown object type ({type(ob)}) provided as part of apply() call"
@@ -231,6 +233,8 @@ class FeatureStore:
             project=self.project,
             tables_to_delete=[],
             tables_to_keep=views_to_update,
+            entities_to_delete=[],
+            entities_to_keep=entities_to_update,
             partial=True,
         )
 
--- a/sdk/python/feast/infra/local.py
+++ b/sdk/python/feast/infra/local.py
@@ -8,6 +8,7 @@ import pandas as pd
 import pytz
 
 from feast import FeatureTable, utils
+from feast.entity import Entity
 from feast.feature_view import FeatureView
 from feast.infra.key_encoding_utils import serialize_entity_key
 from feast.infra.offline_stores.helpers import get_offline_store_from_sources
@@ -50,6 +51,8 @@ class LocalProvider(Provider):
         project: str,
         tables_to_delete: Sequence[Union[FeatureTable, FeatureView]],
         tables_to_keep: Sequence[Union[FeatureTable, FeatureView]],
+        entities_to_delete: Sequence[Entity],
+        entities_to_keep: Sequence[Entity],
         partial: bool,
     ):
         conn = self._get_conn()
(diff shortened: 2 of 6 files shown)


sdk/python/feast/repo_operations.py
```diff
@@ -111,10 +111,15 @@ def apply_total(repo_config: RepoConfig, repo_path: Path):
     all_to_keep.extend(repo.feature_tables)
     all_to_keep.extend(repo.feature_views)
 
+    entities_to_delete: List[Entity] = []
```
MattDelac, line 114: Unsure about this `repo_operation.py` file so some help in how to populate this `entities_to_delete` would be nice

sdk/python/feast/infra/local.py
```diff
@@ -60,9 +61,7 @@ def update_infra(
         for table in tables_to_delete:
             conn.execute(f"DROP TABLE IF EXISTS {_table_id(project, table)}")
 
-    def teardown_infra(
-        self, project: str, tables: Sequence[Union[FeatureTable, FeatureView]]
-    ) -> None:
+    def teardown_infra(self, *args, **kwargs,) -> None:
```
woop, line 64: Is there a reason to not have consistent arguments across methods for implementations of `Provider`?

MattDelac: Nor all arguments are necessary as it depends on the Provider itself (as here we don't use any of the Parameters).
For example on the "ShopifyProvider", we would not use the `project` parameter neither.

Therefore I thought that only using the arguments we need is a good practice and it makes the upstream Provider object more flexible (if we want to add other parameters in the future)

Happy to hard code everything. Don't have strong feelings about it

woop: One part that gives me pause: We'd lose our type checking if we go this route. If I (personally) were to implement a `Provider` I would want to know when the upstream project (`Feast`) has made a change to parameters that will be passed to me, and ideally not at runtime.

MattDelac: Fair enough 👍 

Will go that route then

MattDelac: Done

sdk/python/feast/feature_store.py
```diff
@@ -217,12 +217,14 @@ def apply(
...
             objects = [objects]
         views_to_update = []
+        entities_to_update = []
         for ob in objects:
             if isinstance(ob, FeatureView):
                 self._registry.apply_feature_view(ob, project=self.project)
                 views_to_update.append(ob)
             elif isinstance(ob, Entity):
-                self._registry.apply_entity(ob, project=self.project)
+                self._registry.apply_entity(ob, project=self.config.project)
```
woop, line 226: Perhaps
```suggestion
                self._registry.apply_entity(ob, project=self.project)
```

MattDelac: Glitch because of the rebase.

Fixed

sdk/python/feast/repo_operations.py
woop: So basically we just want to clean up any infra for entities that are removed from the feature repo. I think this raises another question: Would we ever want to just apply an operation on a single entity being added/removed, or is it always as part of a feature view or feature table? I don't really see the use case right now. So to me it may make sense to just pass a list of entities alongside "tables_to_delete" and "tables_to_keep" so that we have the necessary information to create/remove infrastructure. Instead of "entities_to_keep". Over time I would like to move entities into feature views/tables so that we dont have to pass it as a separate object. cc @jklegar @tsotnet

woop: So more tangibly: Maybe we should just rename these `entities_to_keep` parameters to `entities` and remove all `entities_to_delete` instances.

MattDelac: I see. That also makes sense to me 👌 

So we agree that the `entities` would be the ones we keep at any given point right ?

MattDelac: Changed

woop: Entities would still come and go, but we know which ones are coming and going based on the tables/views.

woop (approved):

PR #1499, opened 2021-04-23
Better logging for materialize command

jklegar:

**What this PR does / why we need it**: Improve logging for materialize and materialize_incremental

Fixes #

```release-note
NONE
```


sdk/python/feast/infra/gcp.py
```diff
@@ -101,11 +102,12 @@ def online_write_batch(
     ) -> None:
         client = self._initialize_client()
 
-        pool = ThreadPool(processes=40)
-        pool.map(
-            lambda b: _write_minibatch(client, project, table, b, progress),
-            _to_minibatches(data),
-        )
+        with tqdm(total=len(data)) as pbar:
```
woop, line 105: I think it may be better to have tqdm at a higher level so that its not optional to do logging. Would it make more sense in `materialize_incremental()`?

woop:
image

Nitpicks:

* do we need microseconds in the logs? 
* I prefer "to" to "until" and "from" instead of "starting at"
* I also think that its weird that we talk about the end date before the start date, instead of start->end together.
* I also think we can have fixed width bars instead of using the full console width.

sdk/python/feast/infra/gcp.py
jklegar: I don't see an easy way to bring it up to materialize_incremental since we don't know the number of rows of data at that level; however, `progress` is an argument in methods in the `Provider` class so that kinda makes it not optional for implementations of `Provider` to do some sort of progress logging, presumably by copying existing implementations and using `tqdm`

woop: I thought I replied here, but somehow my message didn't send. I think my question is just: How can we ensure consistency in the logging output if the implementations need to create their own tqdm object. Wouldn't the current implementation make it unscalable to have different configurations for tqdm? Even if the row count is deeper, maybe it makes sense to instantiate the tqdm object higher and pass it into the implementation.

tsotnet (approved):

PR #1501, opened 2021-04-23
Add support for third party providers

tsotnet:
**What this PR does / why we need it**: Currently there's no way to plug in custom providers in Feast CLI. With this PR, you can modify `provider` field in the `feature_store.yaml` to a Python class path. For example, if you define your provider `MyCustomProvider` under `my_package.providers`, you can specify `provider: my_package.providers.MyCustomProvider` and then normally run `feast apply` or any other CLI command.

Fixes #

```release-note
Add support for third party providers
```

### DIFF
--- a/sdk/python/feast/infra/provider.py
+++ b/sdk/python/feast/infra/provider.py
@@ -137,17 +137,18 @@ class Provider(abc.ABC):
 
 
 def get_provider(config: RepoConfig, repo_path: Path) -> Provider:
-    if config.provider == "gcp":
-        from feast.infra.gcp import GcpProvider
+    if "." not in config.provider:
+        if config.provider == "gcp":
+            from feast.infra.gcp import GcpProvider
 
-        return GcpProvider(config)
-    elif config.provider == "local":
-        from feast.infra.local import LocalProvider
+            return GcpProvider(config)
+        elif config.provider == "local":
+            from feast.infra.local import LocalProvider
 
-        return LocalProvider(config, repo_path)
+            return LocalProvider(config, repo_path)
+        else:
+            raise errors.FeastProviderNotImplementedError(config.provider)
     else:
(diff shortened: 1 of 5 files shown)


sdk/python/feast/infra/provider.py
```diff
@@ -144,7 +146,30 @@ def get_provider(config: RepoConfig, repo_path: Path) -> Provider:
 
         return LocalProvider(config, repo_path)
     else:
-        raise ValueError(config)
+        if "." not in config.provider:
```
woop, line 149: What if I provided the wrong name, like `azure`? should we have a `NotImplementedError` for not specifying a real "core" provider?

tsotnet: We can reserve specific names and raise NotImplementedError for now. Otherwise it'll just complain that we can't parse the provider

sdk/python/feast/infra/provider.py
```diff
@@ -144,7 +146,30 @@ def get_provider(config: RepoConfig, repo_path: Path) -> Provider:
...
+
+        # Try getting the provider class definition
+        try:
+            ProviderCls = getattr(module, class_name)
+        except AttributeError:
+            # This can only be one type of error, when class_name attribute does not exist in the module
+            # So we don't have to include the original exception here
+            raise errors.ProviderClassImportError(module_name, class_name) from None
+
+        return ProviderCls(config)
```
woop, line 172: We should probably pass in the `repo_path` for now, as with the `LocalProvider`. It's used for relative imports in some cases. It should definitely be cleaned up, but probably not as part of this PR.

sdk/python/feast/infra/provider.py
woop: Ok

woop (approved):

PR #1508, opened 2021-04-27
Upgrade grpcio version in python SDK

szalai1:
Feast pins the grpcio dependency version to 1.31.0, which makes hard to use with other packages that requires higher version. (In our case it's Dagster , that requires >=1.32.0 ) 
This PR changes the grpcio version from 1.31.0 to 1.32. (Happy to change it even further) 
Haven't opened an issue, but happy to do if you want. 
Fixes #

NONE

### DIFF
--- a/sdk/python/setup.py
+++ b/sdk/python/setup.py
@@ -47,7 +47,7 @@ REQUIRED = [
     "google-cloud-storage>=1.20.*",
     "google-cloud-core==1.4.*",
     "googleapis-common-protos==1.52.*",
-    "grpcio==1.31.0",
+    "grpcio==1.32.0",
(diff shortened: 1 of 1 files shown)


szalai1:
Question: This will cause a mismatch between the go grpcio version and the python one, shall I change that too? or just scratch my own itch.

woop:
Thanks @szalai1.

Should we perhaps use a more relaxed version (like what Dagster does with `grpcio>=1.32.0"`)?

I dont think it's necessary to update Go right now. I can't see a reason why they would be incompatible.

szalai1:
thanks @woop for the quick review. Changed the version requirement.

jklegar (approved):

PR #1516, opened 2021-04-28
Update Numpy and Pandas versions in Python SDK

JBernC:
Feast pins the range of numpy versions to <1.20 and the pandas version to ~=1.0.0, our internal software requires 1.20.x and 1.1.1 respectively. This PR increases the range of both dependencies by 1 minor version, "numpy<1.21.0" & "pandas>=1.0.0,<1.2.0". More than happy to increase these ranges if necessary, aware it's quite a granular change.

**Which issue(s) this PR fixes:**
None as of yet, can add if needed.
Fixes #

**Does this PR introduce a user-facing change?:**
None


woop:
Thanks @JBernC!

I'm not sure if there is a good reason to cap the pandas and numpy ranges. We may be able to support higher versions that what you've set, but we've not investigated whether there are compatibility issues.

JBernC:
Pandas compatibility errors seem to be being thrown will look into cause.

tsotnet:
@JBernC the change looks good to me. Can you resolve the conflict and update the PR to get it merge?

woop:
Actually this change isn't necessary any more is it? We've already bumped the versions on master.

JBernC:
Yep, pandas has been bumped so I can close this. Cheers all!

Issue #1496, opened 2021-04-22
Refactor Feast SDK test suite


woop:
Our current test suite has the following problems
* Lack of generalization across different sources, stores, providers. Ideally we would have a high level set of e2e tests that use the user API to test different combination of source, stores, providers and other infra config. An example of the problem here is https://github.com/feast-dev/feast/blob/master/sdk/python/tests/test_historical_retrieval.py#L313 where we have a bigquery test and a parquet test.
* Related to the above problem is the fact that we aren't reusing our setup/teardown for tests. Ideally the setup/teardown would be externalized and reusable across different kinds of tests (testing online retrieval, historical retrieval, materialization). Even better if we can just do the setup/teardown once, and then run a range of tests on the staged infrastructure.
* Ideally we would have a single "kitchen sink" use case that can be applied to all tests, removing the need for test authors to think about what kind of data they need to use
* We also need a place to add smaller tests, which are closer to unit tests. An example of this is testing whether entity dataframe inferencing works #1495. Its not scalable to keep adding conditional flags and rerun the whole test suite. Eventually our integration tests will take too long to complete.
* Lack of parallelism. We can run many of these tests async if they are well isolated.
* Smaller point: We need to test all the types that we support. Currently we're not testing fringe types with our integration tests.

PR #1520, opened 2021-04-28
Add Unix Timestamp value type

MattDelac:
As discussed in Slack, so far, we don't want to support datetime dtypes in general. Therefore, so far we would need to convert any datetime data types into a unix timestamp and store it as an INT in feast.

While converting in unix timestamp is not an issue, it makes it hard to differentiate the features that are int vs timestamp when looking at a FeatureView

Thus this PR aims to introduce a new datatype "Unix Timestamp" that beyond the scene would be an INT. For the end user, this will make it easier understand a FeatureView.
It also add a potential possibility to do things differently with Unix Timestamp features at retrieval time.

Fixes #

```release-note
Add UnixTimestamp valueType that is an Int64 under the hood
```


sdk/python/tensorflow_metadata/proto/v0/path_pb2.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
 # source: tensorflow_metadata/proto/v0/path.proto
-"""Generated protocol buffer code."""
```
woop, line 4: This is weird. Was this manually edited?

woop:
Hey @MattDelac, thanks for the contribution!

Overall it looks good to me. One thing that is a glaring omissions (not with your PR but with our code) is that we're not testing all the different types. Regardless, we can add that test a bit later. I've added a note to add a [test here](https://github.com/feast-dev/feast/issues/1496).

I think the only thing missing here is an update to our documentation. You may need to update the specs to document the new type: https://github.com/feast-dev/feast/tree/master/docs/specs

sdk/python/tensorflow_metadata/proto/v0/path_pb2.py
MattDelac: Don't think I edited anything manually for a pb2 file
I can start from a fresh master and see if that happens again after compiling protos ...

MattDelac:
> One thing that is a glaring omissions (not with your PR but with our code) is that we're not testing all the different types.

Yes that was my surprise that none of the tests had to be changed

> I've added a note to add a test here.

Sounds good. I let you fix it later 👌 

> I think the only thing missing here is an update to our documentation

Good catch ! Will look into the doc

MattDelac:
> I think the only thing missing here is an update to our documentation. You may need to update the specs to document the new type: https://github.com/feast-dev/feast/tree/master/docs/specs

Done. The docs are now up to date. Let me know if I missed another one.

sdk/python/feast/type_map.py
```diff
@@ -252,6 +252,18 @@ def _python_value_to_proto_value(feast_value_type, value) -> ProtoValue:
                 )
             )
 
+        if feast_value_type == ValueType.UNIX_TIMESTAMP_LIST:
```
woop, line 255: Should the SCALAR type also be handled below?

MattDelac: True.

Done

woop (approved):

PR #1523, opened 2021-04-29
Inferencing of Features in FeatureView and timestamp column of DataSource

### LINKED ISSUE #1500: Add schema inferencing to feature views
Please see [[RFC] FeatureView schema inference](https://docs.google.com/document/d/1MkWvexE4e5nYWcQLELFnJ5o9OlJDKC2rn_USHMDT9dg/edit#)

mavysavydav:
Test cases written and also tested by running with CLI. 

Milestone 1 of https://docs.google.com/document/d/1MkWvexE4e5nYWcQLELFnJ5o9OlJDKC2rn_USHMDT9dg/edit

Fixes #1500

```release-note
Feast will be able to infer the features to register if no features are included in the FeatureView definition, and also infer the event timestamp column of DataSources if that parameter is left out. This is milestone 1 of https://docs.google.com/document/d/1MkWvexE4e5nYWcQLELFnJ5o9OlJDKC2rn_USHMDT9dg/edit#heading=h.pqhio4s5uw2s
```

### DIFF
--- a/sdk/python/feast/data_source.py
+++ b/sdk/python/feast/data_source.py
@@ -14,7 +14,12 @@
 
 
 import enum
-from typing import Dict, Optional
+import re
+from typing import Dict, Iterable, Optional, Tuple
+
+from google.cloud import bigquery
+from pyarrow import DataType
+from pyarrow.parquet import ParquetFile
 
 from feast.data_format import FileFormat, StreamFormat
 from feast.protos.feast.core.DataSource_pb2 import DataSource as DataSourceProto
@@ -515,11 +520,42 @@ class DataSource:
         """
         raise NotImplementedError
 
+    def infer_event_timestamp_column(self, ts_column_type_regex_pattern):
+        ERROR_MSG_PREFIX = "Unable to infer DataSource event_timestamp_column"
+
+        if isinstance(self, FileSource) or isinstance(self, BigQuerySource):
+            event_timestamp_column, matched_flag = None, False
+            for col_name, col_datatype in self.get_table_column_names_and_types():
+                if re.match(ts_column_type_regex_pattern, col_datatype):
+                    if matched_flag:
+                        raise TypeError(
+                            f"""
+                            {ERROR_MSG_PREFIX} due to multiple possible columns satisfying
+                            the criteria.
+                            """
+                        )
+                    matched_flag = True
+                    event_timestamp_column = col_name
+            if matched_flag:
+                return event_timestamp_column
+            else:
+                raise TypeError(
+                    f"""
+                    {ERROR_MSG_PREFIX} due to an absence of columns that satisfy the criteria.
+                    """
+                )
+        else:
+            raise TypeError(
+                f"""
+                {ERROR_MSG_PREFIX} because this DataSource currently does not support this inference.
+                """
+            )
+
 
 class FileSource(DataSource):
     def __init__(
         self,
-        event_timestamp_column: str,
+        event_timestamp_column: Optional[str],
         file_url: Optional[str] = None,
         path: Optional[str] = None,
         file_format: FileFormat = None,
@@ -543,8 +579,10 @@ class FileSource(DataSource):
         Examples:
             >>> FileSource(path="/data/my_features.parquet", event_timestamp_column="datetime")
         """
+
         super().__init__(
-            event_timestamp_column,
+            event_timestamp_column
+            or self.infer_event_timestamp_column(r"timestamp\[\w\w\]"),
             created_timestamp_column,
             field_mapping,
             date_partition_column,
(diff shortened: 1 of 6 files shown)


woop:
Thanks @mavysavydav, we will have a look.

woop:
> Don't turn columns that start with or end with "__" (double underscore) into a Feature. (any objections?)

Seems reasonable.

mavysavydav:
About to push up new commit with test cases and some other minor fixes @jklegar. Feel free to wait on review until that commit arrives

mavysavydav:
Ok, tests have been added and ball is in ur court :D 🏀

sdk/python/tests/fixtures/data_source_fixtures.py
```diff
@@ -0,0 +1,65 @@
...
+            file_url=f.name,
+            event_timestamp_column=event_timestamp_column,
+        )
+        yield file_source
+
+
+def prep_bq_source(df, event_timestamp_column="") -> BigQuerySource:
+    client = bigquery.Client()
+    gcp_project = client.project
+    bigquery_dataset = "ds"
```
jklegar, line 49: Can we use a more descriptive name here? Maybe "feast_infer_features_test" or something, up to you

jklegar (commented):
Thank you David this is great! Left a few comments, most of them small - let me know if anything doesn't make sense

mavysavydav:
thx for review! i agree w ur suggestions and have made the changes (including adding a test for checking that bigQuerySource is properly inferencing based on `query` when table_ref is absent.

sdk/python/feast/data_source.py
```diff
@@ -728,21 +736,36 @@ def get_table_query_string(self) -> str:
...
+            """
+            table_schema = (
+                client.query(bq_columns_query).result().to_dataframe_iterable()
             )
+            for df in table_schema:
+                name_type_pairs.extend(
+                    list(zip(df["COLUMN_NAME"].to_list(), df["DATA_TYPE"].to_list()))
+                )
+        else:
+            bq_columns_query = f"SELECT * FROM {self.query} LIMIT 1"
```
jklegar, line 763: I think we may need parentheses around {self.query} here

sdk/python/tests/fixtures/data_source_fixtures.py
```diff
@@ -63,3 +67,13 @@ def prep_bq_source(df, event_timestamp_column="") -> BigQuerySource:
...
         table_ref=table_ref, event_timestamp_column=event_timestamp_column,
     )
+
+
+def simple_bq_source_using_query_arg(df, event_timestamp_column="") -> BigQuerySource:
+    bq_source_using_table_ref = simple_bq_source_using_table_ref_arg(
+        df, event_timestamp_column
+    )
+    return BigQuerySource(
+        query=bq_source_using_table_ref.table_ref,
```
jklegar, line 77: The query is supposed to be a select statement, so I think something like f"SELECT * FROM {bq_source_using_table_ref.table_ref}" would be better

sdk/python/tests/test_feature_store.py
```diff
@@ -203,27 +207,46 @@ def test_feature_view_inference_success(test_feature_store, dataframe_source):
...
         }
+        actual_bq_using_query_arg_source = {
+            (feature.name, feature.dtype) for feature in feature_view_3.features
+        }
         expected = {
             ("float_col", ValueType.DOUBLE),
             ("int64_col", ValueType.INT64),
             ("string_col", ValueType.STRING),
         }
+        expected2 = {  # parsing with the "query" param in bq sources uses different code path
```
jklegar, line 242: I think we want both code paths to end up producing the same types, see also comment on types above

sdk/python/feast/type_map.py
```diff
@@ -486,10 +486,12 @@ def pa_column_to_proto_column(
...
 def bq_to_feast_value_type(bq_type_as_str):
     type_map: Dict[ValueType, Union[str, Dict[str, Any]]] = {
-        "DATETIME": ValueType.STRING,  # Unsure if string is right
-        "TIMESTAMP": ValueType.STRING,  # Unsure if string is right
+        "DATETIME": ValueType.STRING,  # Update to ValueType.UNIX_TIMESTAMP once #1520 lands.
+        "TIMESTAMP": ValueType.STRING,  # Update to ValueType.UNIX_TIMESTAMP once #1520 lands.
+        "INTEGER": ValueType.INT32,
         "INT64": ValueType.INT64,
         "STRING": ValueType.STRING,
+        "FLOAT": ValueType.FLOAT,
```
jklegar, line 494: it's pretty weird on BigQuery's part that it's giving different types based on whether you query the INFORMATION_SCHEMA table or look at the field_type on schema_field; however, it seems INTEGER and INT64 are equivalent, and same for FLOAT and FLOAT64 - see https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#TableFieldSchema.FIELDS.type. 
So I think that here you can just have "INTEGER" map to the int64 value type and "FLOAT" map to the double value type, and this should also fix the mapping to different types in the test below

jklegar (commented):
This is great - a couple small followup comments and then should be good to go

jklegar (approved):

PR #1535, opened 2021-05-05
Add opt-out exception logging telemetry

jklegar:
**What this PR does / why we need it**: Telemetry will now log anonymized exceptions from Feast methods in addition to usage statistics

Fixes #

```release-note
Telemetry will now log anonymized exceptions from Feast methods in addition to usage statistics. To opt out, see https://docs.feast.dev/v/master/feast-on-kubernetes/advanced-1/telemetry
```

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -225,6 +225,8 @@ class FeatureStore:
 
         if isinstance(objects, Entity) or isinstance(objects, FeatureView):
             objects = [objects]
+        assert isinstance(objects, list)
+
         views_to_update = []
         entities_to_update = []
         for ob in objects:
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -34,7 +34,7 @@ from feast.protos.feast.serving.ServingService_pb2 import (
 from feast.protos.feast.types.EntityKey_pb2 import EntityKey as EntityKeyProto
 from feast.registry import Registry
 from feast.repo_config import RepoConfig, load_repo_config
-from feast.telemetry import Telemetry
+from feast.telemetry import Telemetry, public_method
 from feast.version import get_version
 
 
@@ -51,6 +51,7 @@ class FeatureStore:
     repo_path: Path
     _registry: Registry
 
+    @public_method
     def __init__(
         self, repo_path: Optional[str] = None, config: Optional[RepoConfig] = None,
     ):
@@ -73,6 +74,7 @@ class FeatureStore:
         )
         self._tele = Telemetry()
 
+    @public_method
     def version(self) -> str:
         """Returns the version of the current Feast SDK/CLI"""
 
@@ -86,6 +88,7 @@ class FeatureStore:
         # TODO: Bake self.repo_path into self.config so that we dont only have one interface to paths
         return get_provider(self.config, self.repo_path)
 
+    @public_method
     def refresh_registry(self):
         """Fetches and caches a copy of the feature registry in memory.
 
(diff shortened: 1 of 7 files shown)


sdk/python/feast/feature_store.py
```diff
@@ -34,7 +34,7 @@
 from feast.protos.feast.types.EntityKey_pb2 import EntityKey as EntityKeyProto
 from feast.registry import Registry
 from feast.repo_config import RepoConfig, load_repo_config
-from feast.telemetry import Telemetry
+from feast.telemetry import Telemetry, public_method
```
woop, line 37: Do we still need a `Telemetry` object if we are using `public_method`?

jklegar: It can almost but not entirely be rolled in - the problem is for `get_online_features` the `_telemetry_counter` needs a long-standing Telemetry() object either on the FeatureStore class or elsewhere. One way around it would be to have a global Telemetry() instance which I think we had at one point, wdyt?

woop: Yea, I think the [Global Object Pattern](https://python-patterns.guide/python/module-globals/) may make sense here.

woop: wdyt?

docs/feast-on-kubernetes/advanced-1/telemetry.md
```diff
@@ -2,9 +2,9 @@
 
 ## How telemetry is used
 
-The Feast maintainers use anonymous usage statistics to help shape the Feast roadmap. Several client methods are tracked, beginning in Feast 0.9. Users are assigned a UUID which is sent along with the name of the method, the Feast version, the OS \(using `sys.platform`\), and the current time. For more detailed information see [the source code](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/telemetry.py).
+The Feast maintainers use anonymous usage statistics and error tracking to help shape the Feast roadmap. Several client methods are tracked, beginning in Feast 0.9. Users are assigned a UUID which is sent along with the name of the method, the Feast version, the OS \(using `sys.platform`\), and the current time. For more detailed information see [the source code](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/telemetry.py).
```
woop, line 5: @jklegar can we pull the Telemetry docs out of `feast-on-kubernetes`? What about creating a `Telemetry` page under `Reference`?

sdk/python/feast/telemetry.py
```diff
@@ -30,29 +32,38 @@
...
+                    with open(telemetry_filepath, "r") as f:
+                        self._telemetry_id = f.read()
+                else:
+                    self._telemetry_id = str(uuid.uuid4())
+
+                    with open(telemetry_filepath, "w") as f:
+                        f.write(self._telemetry_id)
+                    print(
+                        "Feast is an open source project that collects anonymized error reporting and usage statistics. To opt out or learn"
+                        " more see https://docs.feast.dev/v/master/feast-on-kubernetes/advanced-1/telemetry"
```
woop, line 65: Same here, I think we should create a nicer URL. We also shouldnt point to master.

sdk/python/feast/telemetry.py
```diff
@@ -30,29 +32,38 @@
...
 class Telemetry:
     def __init__(self):
-        feast_home_dir = join(expanduser("~"), ".feast")
-        Path(feast_home_dir).mkdir(exist_ok=True)
-        telemetry_filepath = join(feast_home_dir, "telemetry")
-        self._telemetry_enabled = (
+        self._telemetry_enabled = False
+        self.check_env()
+
+    def check_env(self):
```
woop, line 38: `check` is not an appropriate name if we are actually writing and initializing Telemetry.

woop (approved):

PR #1538, opened 2021-05-06
Entity value_type inference for Feature Repo registration

### LINKED ISSUE #1500: Add schema inferencing to feature views

mavysavydav:
Signed-off-by: David Y Liu <davidyliuliu@gmail.com>

This change enables users to register Entities without needing to specify the value_type because we'll be able to infer it based on finding the entity column in the table based on the Entity name and using that type. Milestone 2 of [this RFC](https://docs.google.com/document/d/1MkWvexE4e5nYWcQLELFnJ5o9OlJDKC2rn_USHMDT9dg/edit#heading=h.pqhio4s5uw2s). 

Fixes #1500

```release-note
This change enables users to register Entities without needing to specify the value_type because we'll be able to infer it based on finding the entity column in the table based on the Entity name and using that type.  More details can be found in this doc as milestone 2 -- https://docs.google.com/document/d/1MkWvexE4e5nYWcQLELFnJ5o9OlJDKC2rn_USHMDT9dg/edit#heading=h.pqhio4s5uw2s. 
```


mavysavydav:
Sounds good, will take a look

sdk/python/tests/test_repo_operations.py
```diff
@@ -138,3 +149,32 @@ def test_feastignore_with_stars2():
                 (repo_root / "foo1/c.py").resolve(),
             ]
         )
+
+
+@pytest.mark.parametrize("dataframe_source", [lazy_fixture("simple_dataset_1")])
+@pytest.mark.parametrize("dataframe_source_2", [lazy_fixture("simple_dataset_2")])
```
woop, line 155: Is there a reason to use `parametrize` here?

```
import pytest


class Fruit:
    def __init__(self, name):
        self.name = name

    def __eq__(self, other):
        return self.name == other.name


@pytest.fixture
def my_fruit():
    return Fruit("apple")


@pytest.fixture
def fruit_basket(my_fruit):
    return [Fruit("banana"), my_fruit]


def test_my_fruit_in_basket(my_fruit, fruit_basket):
    assert my_fruit in fruit_basket
```

sdk/python/tests/test_repo_operations.py
```diff
@@ -5,8 +5,19 @@
 from typing import Optional
 
 import assertpy
+import pytest
+from fixtures.data_source_fixtures import simple_dataset_1  # noqa: F401
```
woop, line 9: is there a reason for the `noqa`?

sdk/python/feast/repo_operations.py
```diff
@@ -118,6 +118,13 @@ def apply_total(repo_config: RepoConfig, repo_path: Path):
     registry._initialize_registry()
     sys.dont_write_bytecode = True
     repo = parse_repo(repo_path)
+    repo = ParsedRepo(
+        feature_tables=repo.feature_tables,
+        entities=infer_entity_value_type_from_feature_views(
```
woop, line 123: I think we are missing inferencing when using the Python SDK https://github.com/feast-dev/feast/blob/master/sdk/python/feast/feature_store.py#L227

sdk/python/feast/repo_operations.py
```diff
@@ -325,3 +332,42 @@ def replace_str_in_file(file_path, match_str, sub_str):
...
+            if entity_name in incomplete_entities:
+                entity_col = list(
+                    filter(lambda tup: tup[0] == entity_name, col_names_and_types)
+                )
+                if len(entity_col) > 1:
+                    raise ValueError("More than one table column matches Entity name.")
+                if len(entity_col) == 0:
+                    raise ValueError("No column in the table matches Entity name.")
+                entity = incomplete_entities[entity_name]
+                inferrred_value_type = view.input.source_datatype_to_feast_value_type()(
```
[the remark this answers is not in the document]
mavysavydav: good catch

sdk/python/tests/test_repo_operations.py
mavysavydav: nope, reason is my ignorance on how to properly pass in fixtures w/o parametrize

sdk/python/tests/test_repo_operations.py
mavysavydav: the linter wants to group the imports from the same module but when i use parametrize, it wants the import to be a separate line. Since i no longer am going to use parametrize as per ur other comment, we won't have a noqa here, but it may be need again when importing fixtures in future with use with parametrize

mavysavydav:
Ready for another round of review. Sorry for the mess in this PR. Accidentally messed up my branch when syncing w master

woop:
> Ready for another round of review. Sorry for the mess in this PR. Accidentally messed up my branch when syncing w master

Thanks @mavysavydav

sdk/python/tests/test_repo_operations.py
woop: perhaps try https://stackoverflow.com/questions/47196947/py-test-deal-with-both-pylint-and-flake8-when-importing-features-from-a-module

mavysavydav:
btw I made the assumption that for featurestore.apply, we'd only want to infer the entities based on the feature_views passed in as part of that apply call. The reason I thought this was reasonable is that the alternative suggests a feature_view has been applied with an entity that hasn't been defined yet.  Though in the case of redefinitions of entities, inferring from previously applied FeatureViews seems like ok behavior but that won't be done with the current logic.  Happy to reconsider if ppl disagree.

sdk/python/tests/test_repo_operations.py
mavysavydav: ah got it

sdk/python/feast/inference_helpers.py
```diff
@@ -0,0 +1,50 @@
...
+    }
+    incomplete_entities_keys = incomplete_entities.keys()
+
+    for view in feature_views:
+        if not (incomplete_entities_keys & set(view.entities)):
+            continue
+        col_names_and_types = view.input.get_table_column_names_and_types()
+        for entity_name in view.entities:
+            if entity_name in incomplete_entities:
+                entity_col = list(
```
woop, line 24: This is called `entity_col` but its a list. Should it be `entity_columns`?

sdk/python/feast/inference_helpers.py
```diff
@@ -0,0 +1,50 @@
...
+                    filter(lambda tup: tup[0] == entity_name, col_names_and_types)
+                )
+                if len(entity_col) == 0:
+                    # Doesn't mention inference error because would also be an error without inferencing
+                    raise ValueError(
+                        f"""No column in the input source for the {view.name} FeatureView matches
+                        its Entity's name."""
```
woop, line 31: We can just use lower case `entity` and `feature view` going forward.

sdk/python/feast/inference_helpers.py
```diff
@@ -0,0 +1,50 @@
...
+                    )
+
+                entity = incomplete_entities[entity_name]
+                inferred_value_type = view.input.source_datatype_to_feast_value_type()(
+                    entity_col[0][1]
```
woop, line 36: It's not immediately clear what is being referenced here and being passed to `source_datatype_to_feast_value_type`.

sdk/python/feast/inference_helpers.py
```diff
@@ -0,0 +1,50 @@
+from typing import List
```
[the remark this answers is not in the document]
mavysavydav: was concerned about dependencies being introduced in modules that shouldn't depend on one another. entity.py would have a dependency on feature_view. wdyt?

sdk/python/feast/inference_helpers.py
mavysavydav: yep can rename

sdk/python/feast/inference_helpers.py
mavysavydav: renamed variable to make things more clear

sdk/python/feast/inference_helpers.py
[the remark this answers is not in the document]
woop: Ok, can we change the name to just `inference.py`? I just dislike the use of certain filler terms like `helper` / `util` etc.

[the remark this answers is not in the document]
mavysavydav: got it, sgtm

woop (approved):

PR #1552, opened 2021-05-10
Add offline_store config

tsotnet:
**What this PR does / why we need it**: This PR adds offline_store config to feature_store.yaml. Specifically, it makes it possible to override temporary entity dataframe bigquery dataset name.

Here is an example yaml file:
```
project: master_kitten
registry: data/registry.db
provider: gcp
offline_store:
    type: bigquery
    entity_dataset_name: foo
```

Fixes #

```release-note
Add offline_store config to feature_store.yaml
```


jklegar (commented):
My main concern here is that the local provider works with both local and BQ offline stores, as does the GCP provider, so having the offline_store type option might be confusing to users especially since it doesn't affect interaction with the offline store. Not sure if that's already been discussed, and not sure what the better alternative would be though. Thoughts?

woop:
> oes the GCP provider, so having the offline_store type option might be confusing to users especially since it doesn't affect interaction with the offline store. Not sure if that's already been discussed, and not sure what the better alternative would be though. Thoughts?

Ultimately it seems like we have to support decoupling storage (offline, online) from the provider if we want to have reuse, even if the provider still has to opt-in (or can override) the storage implementation. So at first glance, it seems natural to have the `offline_store` object to me. 

> since it doesn't affect interaction with the offline store

What is the "it" here, the "offline_store" object, or the "provider" option?

jparthasarthy:
+1 to Willem's question, I am not sure what the "it" here is. This seems like a fine change to our API, but I might be missing something here:

1. You choose a provider. Providers opt into supporting different offline stores
2. You choose an offline store with the type parameter. Right now, there is only one offline store, BQ. There should be one default offline store per provider.
3. Additional configs to the offline store go in additional keyword parameters.

@tsotnet, is there a default offline store type for each provider? I just want to check, but I don't think users should have to specify the `type` field in the happy path case. There is only one right now after all.

jklegar:
"it" here is the offline_store type - users can set the type to be bigquery and then use file sources or both file and bigquery sources, or vice versa

tsotnet:
> @tsotnet, is there a default offline store type for each provider? I just want to check, but I don't think users should have to specify the type field in the happy path case. There is only one right now after all.

@jay there is a default offline store type. As you'd expect FileOfflineStore for local provider and BigqueryOfflineStore for gcp provider.

> "it" here is the offline_store type - users can set the type to be bigquery and then use file sources or both file and bigquery sources, or vice versa

@jklegar There are 2 directions we can go from here:
1. During `feast apply` validate that all data sources are of the expected type.
2. Enable multiple offline stores to be used in one GCP project. In this case we'll need to make offline_store field a list instead of struct in feature_store.yaml. So for example:
```
project: master_kitten
registry: data/registry.db
provider: gcp
offline_store:
    - type: bigquery
      entity_dataset_name: foo
    - type: local
      imaginary_config: imaginary_value
```

I think in a long term we should support multiple offline stores (and even multiple online stores). But it's not as simple. Currently we partially support this: we can have different data sources in a project, but when getting a historical data all requested features need to be from the same offline store. I recommend a middle ground: enable defining multiple offline store configs, but the behavior of getting a historical data stays the same for now.

Thoughts?

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -94,7 +94,10 @@ def get_historical_features(
             )
 
             table_id = _upload_entity_df_into_bigquery(
-                config.project, entity_df, client
+                config.project,
+                getattr(config.offline_store, "entity_dataset_name", "feast"),
```
woop, line 98: Are you using getattr in order to set a default? The configuration object should be providing the defaults instead of placing configuration in the code. Can we use a property method on the configuration object instead?

sdk/python/feast/repo_config.py
```diff
@@ -37,6 +37,26 @@ class DatastoreOnlineStoreConfig(FeastBaseModel):
...
+    """ Offline store type selector"""
+
+
+class BigqueryOfflineStoreConfig(FeastBaseModel):
+    """ Offline store config for GCP Bigquery """
+
+    type: Literal["bigquery"] = "bigquery"
+    """ Offline store type selector"""
+
+    entity_dataset_name: Optional[StrictStr] = "feast"
```
woop, line 53: What if we just had a catchall `dataset` for the offline store? I can already see the need to have temp tables when doing large joins (to avoid OOM issues with BigQuery). Ideally we wouldnt define two datasets in the config. I think its natural to have a single dataset dedicated to Feast tables.

sdk/python/feast/repo_config.py
```diff
@@ -82,45 +105,84 @@ def _validate_online_store_config(cls, values):
...
+                raise ValidationError(f"Invalid online store type {online_store_type}")
+        except ValidationError as e:
+            raise ValidationError(
+                [ErrorWrapper(e, loc="online_store")], model=SqliteOnlineStoreConfig,
+            )
+
+        return values
+
+    @root_validator(pre=True)
+    def _validate_offline_store_config(cls, values):
```
woop, line 147: Damn I have created a mess with Pydantic. We'll need to figure out a clean way to fix this.

sdk/python/feast/infra/offline_stores/bigquery.py
tsotnet: This was necessary because right now you can have local provider defined in `feature_store.yaml` but still use BigQuery data source. Local provider will default to the file offline store config, which doesn't have `entity_dataset_name` property. This was a quick workaround that doesn't break any of our tests.

In a separate PR we can fix this by either allowing multiple offline store configs to be defined in the feature_store.yaml, or you can suggest something else. But I think we should merge this in first.

sdk/python/feast/repo_config.py
tsotnet: This `dataset` refers specifically to BigQuery's Dataset concept though. Other DWHes may have different concepts. Not sure how we'd generalize this. Instead I think each offline store implementation can define their own configuration that it can fully manage.

sdk/python/feast/infra/offline_stores/bigquery.py
tsotnet: After discussing this offline with Willem, we decided to enforce offline store during `feast apply`. This will restrict Feast repo to use a single offline store, and we can get rid of this ugly getattr.

While this is not an ideal solution, if a use case for multiple data sources appears in the future, we can think about removing this restriction then.

sdk/python/feast/repo_config.py
tsotnet: Cool, I'll rename this to `dataset`.

sdk/python/feast/infra/gcp.py
```diff
@@ -193,25 +194,16 @@ def materialize_single_feature_view(
...
 
-    @staticmethod
-    def _pull_query(query: str) -> pyarrow.Table:
-        client = bigquery.Client()
-        query_job = client.query(query)
-        return query_job.to_arrow()
-
-    @staticmethod
     def get_historical_features(
+        self,
```
woop, line 198: Just curious, is there a reason to convert this into a non-static method?

sdk/python/feast/repo_config.py
```diff
@@ -37,6 +37,26 @@ class DatastoreOnlineStoreConfig(FeastBaseModel):
...
 
+class FileOfflineStoreConfig(FeastBaseModel):
+    """ Offline store config for local (file-based) store """
+
+    type: Literal["file"] = "file"
+    """ Offline store type selector"""
+
+
+class BigqueryOfflineStoreConfig(FeastBaseModel):
+    """ Offline store config for GCP Bigquery """
```
woop, line 48: Typo: `BigQuery`

sdk/python/feast/repo_config.py
```diff
@@ -37,6 +37,26 @@ class DatastoreOnlineStoreConfig(FeastBaseModel):
...
 
 
+class FileOfflineStoreConfig(FeastBaseModel):
+    """ Offline store config for local (file-based) store """
+
+    type: Literal["file"] = "file"
+    """ Offline store type selector"""
+
+
+class BigqueryOfflineStoreConfig(FeastBaseModel):
```
woop, line 47: `BigqueryOfflineStoreConfig` -> `BigQueryOfflineStoreConfig` (inconsistent with how we have been typing it)

woop (approved):

PR #1554, opened 2021-05-12
Add release-patch script

jklegar:

**What this PR does / why we need it**: This script streamlines the release process to make a patch release much faster and simpler.

Fixes #

```release-note
NONE
```


infra/scripts/release-patch.sh
```diff
@@ -0,0 +1,259 @@
...
+    usage
+    exit 1
+fi
+
+echo "This script is mostly idempotent; check git status for temp files before restarting. It will always prompt you before making any non-local change."
+
+# Go to infra/scripts directory
+cd $(dirname "$0")
+# Login to GitHub CLI
+echo $GH_TOKEN | gh auth login --with-token
```
woop, line 46: I think contributors need to know which tools they should have installed.

infra/scripts/release-patch.sh
```diff
@@ -0,0 +1,259 @@
...
+}
+if [ -z $STARTING_COMMIT ]; then
+    read -p "No starting commit given. Skip this step (y) or exit (n)? " choice
+    case "$choice" in
+        y|Y ) echo "Skipping this step" ;;
+        n|N ) exit ;;
+        * ) exit ;;
+    esac ;
+else
+    read -p "Will cherry-pick starting from commit $STARTING_COMMIT. Continue (y) or skip this step (n)? " choice
```
woop, line 67: How do you cherry-pick only certain commits? Or how would the person running this script resolve issues with cherry-picks?

infra/scripts/release-patch.sh
```diff
@@ -0,0 +1,259 @@
...
+    git commit -m "Update CHANGELOG for Feast v$MAJOR.$MINOR.$PATCH"
+}
+update_changelog()
+{
+    echo "Running changelog generator (will take up to a few minutes)"
+    echo -e "# Changelog\n" > temp \
+    && docker run -it --rm ferrarimarco/github-changelog-generator \
+    --user feast-dev \
+    --project feast  \
+    --release-branch master  \
```
woop, line 88: What about patch commits where the release branch is `v0.10-branch` and the release is `v0.10.4` for example?

infra/scripts/release-patch.sh
jklegar: Yeah my thinking here is that a common case is going to be all commits from some starting commit to the most recent one, and if that's not the case you can just cherry-pick the commits you want yourself and then run the script without specifying a starting commit in which case it'll skip the cherry-picking step. (Also feels difficult for the script to make cherry-picking arbitrary commits easier than just doing it yourself.) Issues are resolved as you normally would for a cherry-pick and then you just rerun the script and skip the cherry-picking step.

infra/scripts/release-patch.sh
jklegar: I think for patch commits both ways will work. I just did it off master since it was easier to test without starting an actual release (since if the release-branch is v0.10-branch I'd have to cherry-pick and push the v0.10.4 commits onto that branch first before being able to generate a changelog). There's also the escape hatch of inspecting and modifying the generated changelog but if basing off master ends up too error-prone we should change it to the branch

woop: That all makes sense, but its not clear to me how you ensure that the right changes make it into the changelog. How does the changelog generator know only to include unreleased commits on the v0.10-branch?

jklegar: My understanding was that most of the time we'd be adding all unreleased commits into the next patch, in which case I think these would be the same thing but correct me if I'm wrong here. Since the script prompts you to inspect and modify the generated changelog, the worst case scenario is just having to modify the file before continuing with the script

woop: We should only be backporting patches to a branch, but right now we are trying to release functionality a bit faster so we are even backporting functionality like feature inferencing. So by right we will rarely release all unreleased commits to branches. Otherwise there really isnt a point in maintaining the branches, we can just use master for releasing all commits.

So my preference here would be to ask the user which branch they want to release from, but I dont feel super strongly. Its also fair to just ask them to edit the file for now, but its more error prone I think.

infra/scripts/release-patch.sh
```diff
@@ -0,0 +1,269 @@
...
+        y|Y ) push_cp ;;
+        n|N ) echo "Skipping this sub-step" ;;
+        * ) echo "Skipping this sub-step" ;;
+    esac ;
+fi
+
+create_release()
+{
+    echo "Creating GitHub release"
+    cat ../../CHANGELOG.md | sed -n '/## \[v0.10.4\]/,/## \[v0.10.3\]/p' | sed -n '/**Implemented enhancements/,$p' | sed '$d' > temp2 \
```
woop, line 245: Is there a reason for the hardcoding here?

infra/scripts/release-patch.sh
```diff
@@ -0,0 +1,269 @@
...
+else
+    read -p "The tag is not pushed. Continue (y) or skip this sub-step (n)? " choice
+    case "$choice" in
+        y|Y ) push_tag ;;
+        n|N ) echo "Skipping this sub-step" ;;
+        * ) echo "Skipping this sub-step" ;;
+    esac ;
+fi
+
+read -p "Now wait for the CI to pass. Continue (y) or exit and fix the problem (n)? " choice
```
woop, line 191: Ran into 
```
Now wait for the CI to pass. Continue (y) or exit and fix the problem (n)? y
./release-patch.sh: line 193: Moving on to the next step: command not found
```

woop:
@jklegar can you let me know when you are ready for review again?

infra/scripts/release-patch.sh
```diff
@@ -0,0 +1,279 @@
...
+        * ) echo "Skipping this sub-step" ;;
+    esac ;
+}
+echo "Step 1a: rebase commits"
+if git status | grep -q "is ahead of" ; then
+    read -p "Your local branch is ahead of its remote counterpart, indicating you may have already rebased. Skip this step (y) or run the rebase starting from commit $STARTING_COMMIT (n)? " choice
+    case "$choice" in
+        y|Y ) echo "Skipping this step" ;;
+        n|N ) rebase_from_master ;;
+        * ) rebase_from_master ;;
```
achals, line 84: nit, remove the `n|N` option?

infra/scripts/release-patch.sh
```diff
@@ -0,0 +1,279 @@
...
+    esac ;
+else
+    read -p "Will rebase starting from commit $STARTING_COMMIT. Continue (y) or skip this step (n)? " choice
+    case "$choice" in
+        y|Y ) rebase_from_master ;;
+        n|N ) echo "Skipping this step" ;;
+        * ) echo "Skipping this step" ;;
```
achals, line 91: Same question here

infra/scripts/release-patch.sh
```diff
@@ -0,0 +1,279 @@
...
+    esac ;
+fi
+
+commit_changelog()
+{
+    echo "Committing CHANGELOG.md"
+    git add ../../CHANGELOG.md
```
achals, line 98: Can we use `git rev-parse --show-toplevel` to get the root of the git repo, so that this `git add` doesn't have to be run from a specific directory?

infra/scripts/release-patch.sh
```diff
@@ -0,0 +1,279 @@
...
+    case "$choice" in
+        y|Y ) create_release ;;
+        n|N ) echo "Skipping this step" ;;
+        * ) echo "Skipping this step" ;;
+    esac ;
+fi
+
+echo "Do these final two steps on your own:"
+echo "Step 8: Update the Upgrade Guide"
+echo "Step 9: Update Feast Supported Versions"
```
achals, line 279: Can we add links for where to configure these from?

achals (commented):
looks mostly good, some nits. I haven't actually run the script though

woop:
@tsotnet so the changelog still has to be manually filtered down? Is there a way to automate that?

infra/scripts/release-patch.sh
tsotnet: You shouldn't be getting this after bunch of changes below. Closing this comment for now.

tsotnet:
> 
> 
> @tsotnet so the changelog still has to be manually filtered down? Is there a way to automate that?

What else do you recommend? The script has no idea which commits you want to cherry pick right? So you have to manually specify them

woop (approved):

PR #1555, opened 2021-05-12
Improve example description

theofpa:
Improve the get_online_features description in the hello-world example.

Fixes #1544


woop:
Thanks for this @theofpa. We're busy moving these examples out of the main repository since they are built for Feast 0.9. I'll be creating a new repository with the examples shortly.

PR #1559, opened 2021-05-14
Don't lose materialization interval tracking when re-applying feature views

jklegar:

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/registry.py
+++ b/sdk/python/feast/registry.py
@@ -186,7 +186,10 @@ class Registry:
                     and existing_feature_view_proto.spec.project == project
                 ):
                     # do not update if feature view has not changed; updating will erase tracked materialization intervals
-                    if FeatureView.from_proto(existing_feature_view_proto) == feature_view:
+                    if (
+                        FeatureView.from_proto(existing_feature_view_proto)
+                        == feature_view
+                    ):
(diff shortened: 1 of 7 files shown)


sdk/python/feast/registry.py
```diff
@@ -185,11 +185,60 @@ def updater(registry_proto: RegistryProto):
                     == feature_view_proto.spec.name
                     and existing_feature_view_proto.spec.project == project
                 ):
+                    # do not update if feature view has not changed; updating will erase tracked materialization intervals
```
woop, line 188: It's not immediately clear to me what is happening here. Mind explaining?

jklegar: Yeah, this function `apply_feature_view` does a "destructive update" - it erases tracked materialization intervals so as far as Feast is aware, no time interval of this feature view has been materialized yet. This is what we want if we changed the feature view, but if we didn't it'll cause unnecessary rematerializations. So this part just says don't do a destructive update if the feature view is exactly the same as what's already in the registry

achals (commented):
nits, but looks good otherwise.

achals (approved):
/lgtm

achals (approved):
/lgtm

Issue #1560, opened 2021-05-14
Incorrect exception when GCP dependencies aren't installed


woop:
When using the `gcp` provider within Feast that is installed using `pip install feast`, users will be faced with the following exception

https://github.com/feast-dev/feast/blob/77a79755561132fa68f17916f29e928522a9fa52/sdk/python/feast/registry.py#L439

Instead we should warn users to install `feast[gcp]` or provide the correct dependencies to install.

Issue #1485, opened 2021-04-21
feast apply and Google Datastore dependency


SaschaHeyer:
## Expected Behavior 
Running `feast apply` works without errors

## Current Behavior
Running `feast apply` runns into an error if client library is not installed https://cloud.google.com/datastore/docs/reference/libraries#client-libraries-install-python

```
cannot import name 'datastore' from 'google.cloud'
```
We should mention this dependency in the documentation. 

## Steps to reproduce

1. Run `feast apply` in an environment where the datastore client library is not installed

### Specifications

- Version: 0.10.0
- Platform: Mac
- Subsystem:

## Possible Solution
Install it with `pip install --upgrade google-cloud-datastore`

woop:
Good catch. Thanks @SaschaHeyer!

diogommartins:
Relates to #1560

PR #1561, opened 2021-05-14
Improve GCP exception handling

woop:

If no `gcloud` account is specified, then users today are not provided with a useful exception. Instead, the exception is hidden behind a failure due to an anonymous client not having access to their resources.

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/infra/gcp.py
+++ b/sdk/python/feast/infra/gcp.py
@@ -6,7 +6,6 @@ from typing import Any, Callable, Dict, Iterator, List, Optional, Sequence, Tupl
 import mmh3
 import pandas
 import pyarrow
-from google.auth.exceptions import DefaultCredentialsError
 from tqdm import tqdm
 
 from feast import FeatureTable, utils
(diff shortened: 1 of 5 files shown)


sdk/python/feast/infra/gcp.py
```diff
@@ -27,6 +26,14 @@
 from feast.registry import Registry
 from feast.repo_config import DatastoreOnlineStoreConfig, RepoConfig
 
+try:
+    from google.auth.exceptions import DefaultCredentialsError
+    from google.cloud import bigquery, datastore
```
jklegar, line 31: iirc we didn't have google cloud dependencies imported at the top because Feast will try to import them even when the user is not using gcp (e.g. they're just trying things locally) due to this file being imported somewhere else. If you uninstall the google cloud pip packages in your venv and try the quickstart you can confirm whether this happens

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -20,6 +19,15 @@
 from feast.registry import Registry
 from feast.repo_config import RepoConfig
 
+try:
+    from google.auth.exceptions import DefaultCredentialsError
+    from google.cloud import bigquery
```
jklegar, line 24: here as well, I think (though it may have been just one of gcp.py and bigquery.py where this happened)

sdk/python/feast/infra/gcp.py
woop: Yea, but I think its cleaner to conditionally import our providers and stores than their subdependencies. Let me test to validate it.

woop: I think only this [change](https://github.com/feast-dev/feast/pull/1561/commits/27040ec0bfa28485a5133f1d6180bb997db045a6) was necessary. Our tests are failing for other reasons

tedhtchang:
Fixes #1560 #1485

PR #1569, opened 2021-05-20
Don't create bigquery dataset if it already exists

tsotnet:

**What this PR does / why we need it**: In bigquery.py, first call `client.get_dataset` and only call `client.create_dataset` if it doesn't exist. This way users can manually create the dataset and only grant `bigquery.datasets.get` permission (not `bigquery.datasets.create`) to Feast.

Fixes #

```release-note
Require just `bigquery.datasets.get` permission (not `bigquery.datasets.create`) if the BigQuery dataset defined by `offline_store->dataset` is already created by the user.
```

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -208,8 +208,6 @@ def _upload_entity_df_into_bigquery(project, dataset_name, entity_df, client) ->
     dataset = bigquery.Dataset(f"{client.project}.{dataset_name}")
     dataset.location = "US"
 
-    client.get_dataset(dataset)
-
     try:
         client.get_dataset(dataset)
     except NotFound:
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -102,8 +102,15 @@ class BigQueryOfflineStore(OfflineStore):
 
             assert isinstance(config.offline_store, BigQueryOfflineStoreConfig)
 
+            # We should create a new dataset if the dataset name was not overridden in the config
+            should_create_dataset = "dataset" not in config.offline_store.__fields_set__
+
             table_id = _upload_entity_df_into_bigquery(
-                config.project, config.offline_store.dataset, entity_df, client,
+                config.project,
+                config.offline_store.dataset,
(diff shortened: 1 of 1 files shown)


sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -102,8 +102,15 @@ def get_historical_features(
 
             assert isinstance(config.offline_store, BigQueryOfflineStoreConfig)
 
+            # We should create a new dataset if the dataset name was not overridden in the config
+            should_create_dataset = "dataset" not in config.offline_store.__fields_set__
```
woop, line 106: I don't 100% agree with this approach. What if the dataset already exists but its not set in the config? I think we should just manually check whether the dataset exists before we run the create command.

tsotnet: That'll require an additional `bigquery.datasets.get` permission on the dataset in question

woop: So `get` permissions is too much?

woop: Also, its probably possible for us to get around the `bigquery.datasets.get` permission by trying an operation inside the dataset and seeing if it succeeds (and which exception is thrown).

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -206,9 +207,12 @@ def _upload_entity_df_into_bigquery(project, dataset_name, entity_df, client) ->
...
     dataset.location = "US"
-    client.create_dataset(
-        dataset, exists_ok=True
-    )  # TODO: Consider moving this to apply or BigQueryOfflineStore
+
+    try:
+        client.get_dataset(dataset)
+    except NotFound:
+        # Only create the dataset if it does not exist
+        client.create_dataset(dataset)
```
woop, line 215: Is there a reason we removed `exists_ok=True`? Wouldn't we introduce a race condition here?

tsotnet: I decided to remove it because the except block would only be executed if the dataset doesn't exist. I didn't think about race conditions. I'm adding this back.

woop (approved):

PR #1587, opened 2021-05-26
Update quickstart.md

gkorland:


woop:
@gkorland I think we should be targeting master here. We can backport changes that land on master for you.

woop:
I'll update this manually. This PR is targeting the wrong branch. It's just easier for us to use Gitbook for these changes. Thanks @gkorland!

PR #1593, opened 2021-05-28
Append nanoseconds to dataset name in test_historical_retrival to prevent tests stomping over each other

achals:

Becuase of the way we construct same dataset name in test_historical_retrival, when running a matrix test there's the possiblity of two different tests constructing the same dataset name, and hence stomping over each other.

This PR adds the python version to the name of the dataset - this should be sufficient since the test matrix tests different python versions. An alternative would be to use the github action run id, but that ties the test a little bit closer to github than seems necessary.

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/tests/test_historical_retrieval.py
+++ b/sdk/python/tests/test_historical_retrieval.py
@@ -336,8 +336,8 @@ def test_historical_features_from_bigquery_sources(
     ) = generate_entities(start_date, infer_event_timestamp_col)
 
     # bigquery_dataset = "test_hist_retrieval_static"
-    python_version = sys.version.split(" ")[0].replace(".", "_")
-    bigquery_dataset = f"test_hist_retrieval_{int(time.time())}_{python_version}"
+    version_string = f"{sys.version_info.major}_{sys.version_info.minor}_{sys.version_info.minor}"
+    bigquery_dataset = f"test_hist_retrieval_{int(time.time())}_{version_string}"
 
     with BigQueryDataSet(bigquery_dataset), TemporaryDirectory() as temp_dir:
         gcp_project = bigquery.Client().project
--- a/sdk/python/tests/test_historical_retrieval.py
+++ b/sdk/python/tests/test_historical_retrieval.py
@@ -1,6 +1,8 @@
 import os
 import random
 import string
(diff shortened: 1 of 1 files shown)


sdk/python/tests/test_historical_retrieval.py
```diff
@@ -335,7 +336,8 @@ def test_historical_features_from_bigquery_sources(
     ) = generate_entities(start_date, infer_event_timestamp_col)
 
     # bigquery_dataset = "test_hist_retrieval_static"
-    bigquery_dataset = f"test_hist_retrieval_{int(time.time())}"
```
woop, line 338: I think we may still run into this problem again if we are using seconds and we are running multiple tests. For example if we vary some other dimension (that isnt the python version).

What if we did something like

Problem
```
>>> time.time()
1622217984.4427726
>>> int(time.time())
1622217987
```
Solution?
```
>>> time.time_ns()
1622218044584774799
>>> int(time.time_ns())
1622218049461962906
```

achals: I think that's feasible - my preference would be to use github actions' job_id or run_id cause those are also guaranteed to be unique, because ns timestamps are probably never going to collide but there's still the possibility 🤷🏽 

Would you be okay with using those environment variables (with a default value of you are running locally)?

achals: yeah, fair, I was concerned about that as well - I'll just do `time_ns` along with a random number and it should be good enough to avoid the underlying issue

sdk/python/tests/test_historical_retrieval.py
```diff
@@ -335,7 +335,9 @@ def test_historical_features_from_bigquery_sources(
...
+    bigquery_dataset = (
+        f"test_hist_retrieval_{int(time.time_ns())}_{random.randint(0, 1000)}"
```
woop, line 339: One more minor nit, can we maybe use `randint(1000, 9999)` so that we have a fixed width dataset name?

achals: Sure

woop (approved):

Issue #1589, opened 2021-05-27
CLI `feast materialize` does not work with custom providers


MattDelac:
## Expected Behavior 
This should work as it works with `feast apply`

## Current Behavior
You get a `ModuleNotFoundError`
<img width="1135" alt="image" src="https://user-images.githubusercontent.com/18557047/119858530-8d73e080-bee2-11eb-83e3-cae866b5a8be.png">

## Steps to reproduce
I have my own custom provider
```
# feature_store.yaml
project: local
registry: tmp/registry.db
provider: pano.provider.LocalProvider
offline_store:
    type: bigquery
online_store:
    type: sqlite
    path: tmp/database.db
```
- `feast apply`  works as expected
- `feast materialize-incremental` does not because it cannot find my "pano" module

### Specifications

- Version: feast 0.10.5
- Platform: Mac
- Subsystem:

## Possible Solution

They both use [get_provider() from infra/provider.py](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/provider.py#L141)

The `apply` method in the CLI will do 
```
os.chdir(repo_path)
sys.path.append("")
...
infra_provider = get_provider(repo_config, repo_path)
```

The `materialize-incremental` method of the CLI will do
```
provider = self._get_provider()
# That will return `return get_provider(self.config, self.repo_path)`
```

So the difference is that the FeatureStore does not append the current directory in the sys path

woop:
The solution seems reasonable, if only for consistency sake. Although I am curious how you are structuring your repository now, and whether you think our current approach of importing the current working directory is appropriate.

Issue #1556, opened 2021-05-12
Ability to provide a custom `feature_store.yaml` during CLI operations


MattDelac:
**Is your feature request related to a problem? Please describe.**
We often want to run `feast apply` (or other CLI operations) on different GCP projects.

Therefore it would be nice if we could point to a specific "`feature_store.yaml`" when we use the CLI

**Describe the solution you'd like**
Something easy like `feast apply --conf feature_store_prod.yaml`. By default `--conf` would be `feature_store.yaml`

**Describe alternatives you've considered**
Copying a specific yaml to `feature_store.yaml` when we need to perform CLI operations on different environments

**Additional context**

woop:
Hey @MattDelac

Do you think #1509 (with separate repositories) would address this problem?

MattDelac:
> Hey @MattDelac
> 
> Do you think #1509 (with separate repositories) would address this problem?

I don't think so as you are staying in the same repository.

It's just that we would like the flexibility to have multiple configurations
```
conf/
conf/feature_store_adhoc.yaml
conf/feature_store_prod.yaml
conf/feature_store_local.yaml
```

For example, this would be useful to let the user confirm that their new FeatureView is properly applied by doing `feast apply --conf conf/feature_store_local.yaml`

woop:
@MattDelac Some options that I can imagine

**Option 1: One repo, one config"

This is what we have today. The idea is that the `feature_store.yaml` anchors a configuration repository. It tells us where to scan for all feature definitions within the repository (and what the root folder is). If you have multiple environments then the idea is that you will have something like
```
.
├── prod
│   ├── feature_store.yaml
│   └── my_feature_def.py
└── staging
    ├── feature_store.yaml
    └── my_feature_def.py
```
but I think its possible to do
```
.
├── common
│   └── my_feature_def.py
├── prod
│   ├── feature_store.yaml
│   └── my_feature_def.py
└── staging
    ├── feature_store.yaml
    └── my_feature_def.py
```

**Option 2: One repo, many configs"

Alternatively, we could make it possible to specify a remote configuration file. My main concern with that is that it could be unintuitive how it would function. Would we still consider it to be the root of a feature repo? 

When I see a command like `feast apply --conf conf/feature_store_local.yaml` then I dont think there is anything special about the `--conf` file. But in reality, that conf file location is important since we will use its location to scan for feature definition files.

shihgianlee:
[...]

PR #1594, opened 2021-05-28
Add current directory in sys path for CLI commands that might depend on custom providers

MattDelac:
Only `feast apply` works with custom providers. Other CLI commands that depend on the provider don't work with custom ones.
This is because the current directory is append to the SYS PATH only during `feast apply`

Fixes https://github.com/feast-dev/feast/issues/1589

```release-note
Append current directory in sys path for all CLI commands that depend on a provider
```

### DIFF
--- a/sdk/python/feast/repo_operations.py
+++ b/sdk/python/feast/repo_operations.py
@@ -112,7 +112,6 @@ def apply_total(repo_config: RepoConfig, repo_path: Path):
     from colorama import Fore, Style
 
     os.chdir(repo_path)
-    sys.path.append("")
     registry_config = repo_config.get_registry_config()
     project = repo_config.project
     if not_valid_name(project):
@@ -267,6 +266,7 @@ def registry_dump(repo_config: RepoConfig, repo_path: Path):
 
 
 def cli_check_repo(repo_path: Path):
+    sys.path.append(str(repo_path))
     config_path = repo_path / "feature_store.yaml"
     if not config_path.exists():
         print(


woop:
@MattDelac btw there is a fix for the flaky test over here https://github.com/feast-dev/feast/pull/1593

sdk/python/feast/repo_operations.py
```diff
@@ -112,7 +112,6 @@ def apply_total(repo_config: RepoConfig, repo_path: Path):
     from colorama import Fore, Style
 
     os.chdir(repo_path)
-    sys.path.append("")
```
woop, line 115: Is there a reason we're removing this one?

woop:
@MattDelac I think it may be possible to just use https://github.com/feast-dev/feast/pull/1594/files#diff-cc33a5e7afa90467a37c11f60b2a380bd86af97524300ca6484a0373e2f00cb6R70 so that we don't need to append in multiple places. Also, I think we need to support remote repos using the `-c` flag, right?

MattDelac:
> I think it may be possible to just use https://github.com/feast-dev/feast/pull/1594/files#diff-cc33a5e7afa90467a37c11f60b2a380bd86af97524300ca6484a0373e2f00cb6R70 so that we don't need to append in multiple places.

Yes I put it in `cli_check_repo()` that is called by all CLI commands


> Also, I think we need to support remote repos using the `-c` flag, right?

Sounds like out of topic and related to https://github.com/feast-dev/feast/issues/1556 no ? Unsure I understand your point here

MattDelac:
@woop This PR is ready for a final review

woop (approved):

PR #1601, opened 2021-05-31
Use CONCAT() instead of ROW_NUMBER()

MattDelac:
In order to calculate a unique ID for each row of the entity dataframe, we compute the following
```sql
SELECT ROW_NUMBER() OVER() AS row_number, edf.* FROM {{ left_table_query_string }} as edf
```

The problem is that BigQuery will need to send all the data to a single worker in order to properly calculate the row number of each row. For our use case, we end up with a OOM error
image

The solution is to calculate a **deterministic** funcyion that will act as a unique identifier.
Because the entity_dataframe should contain all entity keys, I use `CONCAT()` that will compute a deterministic string for a given input. This hash is computed in a distributed fashion as it only needs the datapoints of a given row.

```sql
WITH entity_dataframe AS (
    SELECT
        *,
        CONCAT(
            {% for entity_key in unique_entity_keys %}
                CAST({{entity_key}} AS STRING),
            {% endfor %}
            CAST({{entity_df_event_timestamp_col}} AS STRING)
        ) AS entity_row_unique_id
),
```

**Alternative**
First the result of the `CONCAT()` was passed to a Hash function ([`FARM_FINGERPRINT()`](https://cloud.google.com/bigquery/docs/reference/standard-sql/hash_functions)). The problem with a hash function is that we introduce the possibility of a collision between different keys resulting to the same Hash.

I tried [`GENERATE_UUID()`](https://cloud.google.com/bigquery/docs/reference/standard-sql/uuid_functions) that is non deterministic and the query got wrong because i suspect that it computed it multiple times (depending on how the SQL query gets optimized & parsed). So we ended up with all features being always Null

_TODO: Matt can look at how the query is interpreted and see if `GENERATE_UUID()` is called multiple times_

```release-note
Use CONCAT() instead of ROW_NUMBER() to generate the unique identifiers of `entity_dataframe`
```

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -307,6 +307,9 @@ def build_point_in_time_query(
         "max_timestamp": max_timestamp,
         "left_table_query_string": left_table_query_string,
         "entity_df_event_timestamp_col": entity_df_event_timestamp_col,
+        "unique_entity_keys": set(
+            [entity for fv in feature_view_query_contexts for entity in fv.entities]
+        ),
         "featureviews": [asdict(context) for context in feature_view_query_contexts],
     }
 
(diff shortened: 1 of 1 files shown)


MattDelac:
Base on a conversation with one of my coworker, using a hash function introduce a potential collision due to the randomness of it.

That makes me realize that I don't even need it as a `CONCAT(col1, col2, col3)` is also deterministic and will make each row unique.

Thus I am making changes in this direction

woop (approved):

PR #1602, opened 2021-05-31
Optimize historical retrieval with BigQuery offline store

MattDelac:
The current SQL template to perform historical retrieval of BigQuery datasets could be more efficient.

As a former data scientist, I often realized that big data engines (Spark, BigQuery, Presto, etc.) are often much more efficient with a series of GROUP BY rather than a Window function. Moreover, we should avoid ORDER BY operations as much as possible.

So this PR comes with a new approach to compute the point in time join. This is solely compose of JOINs and GROUP BYs

Here are the results of my benchmark:
**Context**
The api call is the following
```python
result = store.get_historical_features(
    entity_df="""
    SELECT user_id, TIMESTAMP '2021-01-01' AS observed_at
    FROM `my_dataset`
    LIMIT 100000000 -- 100M
    """,
    feature_refs=[
        "feature_view_A:feature_1",
        "feature_view_A:feature_2",
        "feature_view_B:feature_3",
        "feature_view_B:feature_4",
        "feature_view_B:feature_5",
        "feature_view_C:feature_6",
        "feature_view_D:feature_7",
    ]
)
```
And some idea of the scale of this historical retrieval
- `feature_view_A` contains ~5B rows and ~3.6B unique "user_id"
- `feature_view_B` contains ~5B rows and ~3.6B unique "user_id"
- `feature_view_C` contains ~1.7B rows and ~1.1B unique "user_id"
- `feature_view_D` contains ~42B rows and ~3.5B unique "user_id"

**Results**
On the original SQL template
- Elapsed time: 5 min 59 sec
- Slot time consumed: 5 days 5 hr
- Bytes shuffled: 5.89 TB
- Bytes spilled to disk: 0 B

With the SQL template of this PR
- Elapsed time: 3 min 21 sec (-44%)
- Slot time consumed: 2 days 12 hr (-52%)
- Bytes shuffled: 3.55 TB (-40%)
- Bytes spilled to disk: 0 B

So as we can see, the proposed SQL template consume half the resources that the one currently implemented when we asked for 100M "user_id".

Also, because this new SQL template is only composed of JOINs and GROUP BY, it should scale "indefinitely" except if the data is highly skewed (eg: a single "user_id" represents 20% of the dataset).

```release-note
Optimized how we perform historical retrieval with BigQuery offline store
```

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -394,6 +394,10 @@ def _get_bigquery_client():
 #   * Create temporary tables instead of keeping all tables in memory
 
 SINGLE_FEATURE_VIEW_POINT_IN_TIME_JOIN = """
+/*
+ Compute a deterministic hash for the `left_table_query_string` that will be used throughout
+ all the logic as the field to GROUP BY the data
+*/
 WITH entity_dataframe AS (
     SELECT
         *,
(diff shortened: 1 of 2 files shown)


sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -340,113 +341,136 @@ def _get_bigquery_client():
...
 SINGLE_FEATURE_VIEW_POINT_IN_TIME_JOIN = """
+/*
+ Compute a deterministic hash for the `left_table_query_string` that will be used throughout
+ all the logic as the field to GROUP BY the data
+*/
 WITH entity_dataframe AS (
-    SELECT ROW_NUMBER() OVER() AS row_number, edf.* FROM {{ left_table_query_string }} as edf
+    SELECT
+        *,
+        FARM_FINGERPRINT(CONCAT(
```
MattDelac, line 351: This comes from the following PR https://github.com/feast-dev/feast/pull/1601

woop:
This is great, thanks @MattDelac 

> **Results**
> On the original SQL template
> 
> * Elapsed time: 5 min 59 sec
> * Slot time consumed: 5 days 5 hr
> * Bytes shuffled: 5.89 TB
> * Bytes spilled to disk: 0 B
> 
> With the SQL template of this PR
> 
> * Elapsed time: 3 min 21 sec (-44%)
> * Slot time consumed: 2 days 12 hr (-52%)
> * Bytes shuffled: 3.55 TB (-40%)
> * Bytes spilled to disk: 0 B

These results are very promising. One thing that isn't clear from your test is whether all your entities occur on a single timestamp, or if a range of times are being used. The pathological case that I would want to test for is when there is a table with super granular events (like real-time features being updated every second) being joined onto a table with comparatively static data (like daily customer data).

Do you think it makes sense for us to 
* Make the BQ historical retrieval tests more representative (larger volume of data and larger differences in granularity of rows between views)
* Print timing as part of our existing BQ historical retrieval tests

At that point it would be trivial to assess whether an optimization in our BQ template has led to improvements.

MattDelac:
> One thing that isn't clear from your test is whether all your entities occur on a single timestamp, or if a range of times are being used

The tests with my data were on a single timestamp. But the integration tests are on a range of time I believe.

MattDelac:
> The pathological case that I would want to test for is when there is a table with super granular events (like real-time features being updated every second) being joined onto a table with comparatively static data (like daily customer data).

Yes I don't think while it should not be supported here ... 🤔 

Definitely, if you think that it's not covered by the integration tests, we should add those first.

woop:
> > The pathological case that I would want to test for is when there is a table with super granular events (like real-time features being updated every second) being joined onto a table with comparatively static data (like daily customer data).
> 
> Yes I don't think while it should not be supported here ... 
> 
> Definitely, if you think that it's not covered by the integration tests, we should add those first.

I think they are covered to some degree, but the problem is we don't use a very large amount of data, nor do we have proper benchmarking. So its more a matter of being able to see the performance difference between commits than a functional test.

MattDelac:
PR ready for a final round of reviews (and for approval 😉 )

woop (approved):

PR #1603, opened 2021-06-01
Update type conversion from pandas to timestamp to support various the timestamp types

achals:
Currently, Inferring features fails on columns of type timestamp[ms, tz=UTC] or other tz. This diff attempts to fix using a regex instead of enumerating all the different possibilities. 

An option is to use the pandas type instead of the string representation, but this is probably good enough.
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Timestamp.html

Fixes #

```release-note
Fix Inferring features columns of pandas timestamp types.
```

### DIFF
--- a/sdk/python/feast/type_map.py
+++ b/sdk/python/feast/type_map.py
@@ -14,6 +14,7 @@
 
 from datetime import datetime, timezone
 from typing import Any, Dict, List, Union
+import re
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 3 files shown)


sdk/python/feast/type_map.py
```diff
@@ -438,8 +439,14 @@ def pa_to_value_type(pa_type: object):
 
 
 def pa_to_feast_value_type(value: Union[pa.lib.ChunkedArray, str]) -> ValueType:
+    value_type = (
+        value.type.__str__() if isinstance(value, pa.lib.ChunkedArray) else value
```
woop, line 443: Would you mind explaining this line?

achals: This is moved from line 459 - basically this converts the type of a pandas column into a string representation, if the stringified version has not already been passed in.

woop (approved):

PR #1612, opened 2021-06-01
Allow telemetry configuration to fail gracefully

achals:

In some environments feast may be unable to writes files to the local filesystem because of it being readonly. Right now, this breaks the telemetry logging in a bad way. We should handle it gracefully.

Fixes #

```release-note
Allow telemetry configuration to fail gracefully
```

### DIFF
--- a/sdk/python/feast/telemetry.py
+++ b/sdk/python/feast/telemetry.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import logging
 import os
 import sys
 import uuid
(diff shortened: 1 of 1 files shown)


sdk/python/feast/telemetry.py
```diff
@@ -45,35 +47,38 @@ def check_env_and_configure(self):
...
+                        self._telemetry_id = str(uuid.uuid4())
+
+                        with open(telemetry_filepath, "w") as f:
+                            f.write(self._telemetry_id)
+                        print(
+                            "Feast is an open source project that collects anonymized error reporting and usage statistics. To opt out or learn"
+                            " more see https://docs.feast.dev/reference/telemetry"
+                        )
+                except Exception as e:
+                    _logger.debug("Unable to configure telemetry", e)
```
woop, line 73: Is logging disabled by default? Ideally we would not print logs relating to Telemetry to users

woop (approved):

PR #1614, opened 2021-06-02
Add entity column validations when getting historical features from bigquery

achals:

Right now, we don't validate that the entity_df contains the expected columns. This may result in strange errors downstream. But since we already have all the necessary information to validate that the needed columns exist, the validations can catch errors before we try to retrieve data from bigquery.

Fixes #

```release-note
Add entity column validations when getting historical features from bigquery
```


sdk/python/telemetry_tests/python-sdk-test-report.xml
```diff
@@ -0,0 +1 @@
+<?xml version="1.0" encoding="utf-8"?><testsuites><testsuite errors="0" failures="0" hostname="Achals-MBP" name="pytest" skipped="0" tests="6" time="154.414" timestamp="2021-06-01T17:24:58.989137"><testcase classname="telemetry_tests.test_telemetry" file="telemetry_tests/test_telemetry.py" line="32" name="test_telemetry_on_v09" time="18.383"></testcase><testcase classname="telemetry_tests.test_telemetry" file="telemetry_tests/test_telemetry.py" line="59" name="test_telemetry_off_v09" time="32.697"></testcase><testcase classname="telemetry_tests.test_telemetry" file="telemetry_tests/test_telemetry.py" line="88" name="test_telemetry_on" time="18.214"></testcase><testcase classname="telemetry_tests.test_telemetry" file="telemetry_tests/test_telemetry.py" line="120" name="test_telemetry_off" time="32.748"></testcase><testcase classname="telemetry_tests.test_telemetry" file="telemetry_tests/test_telemetry.py" line="153" name="test_exception_telemetry_on" time="18.851"></testcase><testcase classname="telemetry_tests.test_telemetry" file="telemetry_tests/test_telemetry.py" line="170" name="test_exception_telemetry_off" time="32.693"></testcase></testsuite></testsuites>
```
woop, line 1: I'm assuming this should be removed.

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -132,6 +141,53 @@ def get_historical_features(
...
+    join_keys: Set[str], entity_df_event_timestamp_col: str, entity_df: pandas.DataFrame
+):
+    entity_df_columns = set(entity_df.columns.values)
+    expected_columns = join_keys.copy()
+    expected_columns.add(entity_df_event_timestamp_col)
+
+    missing_keys = expected_columns - entity_df_columns
+
+    if len(missing_keys) != 0:
+        raise ValueError(
```
woop, line 154: @achals would love to get your take on whether we should be using a custom error type here (or ValueError)? I've accepted #1615, but now that I think about it we probably need tests that cover this failure case. We can always leave those tests as a `TODO` in the code until we have completed our test case refactor, but at that point we probably want different offline stores to raise identical exceptions.

achals: I think a specific validation error is better to throw since they can be caught and handled better potentially. 

And yeah, re: tests I'll fill in the gaps in the stuff I just merged.

achals:
@woop I've updated the test_historical_features in this PR (and will update the error and tests for stuff landed in #1615 in a separate PR).

woop (approved):

PR #1617, opened 2021-06-03
Use drop_duplicates() instead of groupby (about 1.5~2x faster)

rightx2:
`df.drop_duplicates()` is much faster than `groupby() + reset_index()`.

You can test it with the below codes (You can change the number of unique number for each column):
```
df = pd.DataFrame({
    "a": [np.random.randint(0, 100) for _ in range(1000)],
    "b": [np.random.randint(0, 100) for _ in range(1000)],
    "c": [np.random.randint(0, 100) for _ in range(1000)],
})

%%timeit
df.groupby(['a', 'b']).last().reset_index()

%%timeit
df.drop_duplicates(['a', 'b'], keep="last", ignore_index=True, inplace=True)
```

No issues related. It's sort of a little performance improvement

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/infra/offline_stores/file.py
+++ b/sdk/python/feast/infra/offline_stores/file.py
@@ -153,8 +153,9 @@ class FileOfflineStore(OfflineStore):
                     ]
 
                 df_to_join.sort_values(by=right_entity_key_sort_columns, inplace=True)
-                df_to_join = df_to_join.groupby(by=right_entity_key_columns).last()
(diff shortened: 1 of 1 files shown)


achals (approved):

achals:
Thanks for the PR @rightx2! I dont have land permissions yet but @woop should be able to get this merged in.

achals (approved):
Probably gotta approve again after being added to the OWNERS file

PR #1623, opened 2021-06-07
Add to_df() to convert get_online_feature response to pandas dataframe

### LINKED ISSUE #1613: Add `to_df()` to OnlineResponse
We need to provide a Pandas Dataframe based return type from `get_online_features()` for consistency with `get_historical_features()`.

tedhtchang:
Signed-off-by: ted chang <htchang@us.ibm.com>

Convert online response to Panda dataframe

Fixes #1613

```release-note
Users can convert online feature response to Panda dataframe
```

### DIFF
--- a/sdk/python/tests/test_online_retrieval.py
+++ b/sdk/python/tests/test_online_retrieval.py
@@ -2,7 +2,9 @@ import os
 import time
 from datetime import datetime
 
+import pandas
 import pytest
+from pandas.testing import assert_frame_equal
 
 from feast import FeatureStore, RepoConfig
 from feast.errors import FeatureViewNotFoundException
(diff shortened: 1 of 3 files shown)


woop:
Rerunning the tests now. Codecov has a bug that I tried to fix. You may need to rebase on master. https://github.com/feast-dev/feast/commit/3b22005bd7a81d93991a69da8373c44011e02a96

sdk/python/tests/test_online_retrieval.py
```diff
@@ -110,6 +112,29 @@ def test_online() -> None:
...
+                "driver": [1, 1],
+                "customer": [5, 5],
+                "driver_locations__lon": ["1.0", "1.0"],
+                "customer_profile__avg_orders_day": [1.0, 1.0],
+                "customer_profile__name": ["John", "John"],
+                "customer_driver_combined__trips": [7, 7],
+            }
+        )
+        # Ignore the order of columns
+        assert_frame_equal(result_df, expected_df, check_like=True)
```
woop, line 136: Can we try to use a few different entities? I want to make sure that we are testing for ordering as well (ideally the response rows would be in the same order as the request).

woop:
Thanks @tedhtchang, this is great!

sdk/python/tests/test_online_retrieval.py
tedhtchang: It seems OnlineResponse obj does not maintain the orders of the original request.  Perhaps I can pass the feature_refs and entity_rows into the the OnlineResponse and then sort the col and rows afterwards?

woop: > It seems OnlineResponse obj does not maintain the orders of the original request. Perhaps I can pass the feature_refs and entity_rows into the the OnlineResponse and then sort the col and rows afterwards?

Ideally we would push this responsibility down to the connectors and wouldn't sort all responses for them, otherwise we will always pay a performance penalty. What do you think about adding a test for consistency of ordering here and then we can create a new issue/PR to make sure ordering is upheld in responses from the storage connectors?

tedhtchang: Done. Please take a look if I understood correctly, thanks.

woop (approved):

woop:
Thanks @tedhtchang

sdk/python/tests/test_online_retrieval.py
```diff
@@ -234,3 +236,163 @@ def test_online() -> None:
...
+                1                   1.0                    0.1
+                2                   2.0                    0.2
+                3                   3.0                    0.3
+            """
+            driver_key = EntityKeyProto(
+                join_keys=["driver"], entity_values=[ValueProto(int64_val=d)]
+            )
+            provider.online_write_batch(
+                project=store.project,
+                table=driver_locations_fv,
```
achals, line 280: This API was updated recently, you may need to rebase and provide a `config` kwarg instead of `project`.

achals (changes requested):

achals (approved):
/lgtm

PR #1628, opened 2021-06-08
Introduce an OnlineStore interface

### LINKED ISSUE #1573: Decouple Online Store from Provider
Currently, all online store implementations are built into [Providers](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/provider.py#L25). There are two problems with this approach
1. It's not possible to use an online store other than the one configured for your provider
2. It's not possible to create an extension point for teams to provide new online store implementations without creating a new provider

achals:

Similar to how we decoupled the the OfflineStore for extensibility, this PR pulls out the OnlineStore. It also updates tests, and refactors the Sqlite implementation and GCP into their own components.

Fixes #1573 

```release-note
Introduce an OnlineStore interface.
```


sdk/python/feast/infra/online_stores/online_store.py
```diff
@@ -0,0 +1,52 @@
...
+
+class OnlineStore(ABC):
+    """
+    OnlineStore is an object used for all interaction between Feast and the service used for online storage of
+    features.
+    """
+
+    @staticmethod
+    @abstractmethod
+    def online_write_batch(
```
woop, line 34: Can we please add comments that describe these two abstract methods?

achals: Added

sdk/python/feast/infra/online_stores/online_store.py
```diff
@@ -0,0 +1,102 @@
...
+    ):
+        ...
+
+    @staticmethod
+    @abstractmethod
+    def teardown(
+        config: RepoConfig,
+        tables: Sequence[Union[FeatureTable, FeatureView]],
+        entities: Sequence[Entity],
+    ):
```
achals, line 101: @woop I added these methods here (and ripped out stuff from `gcp.py` and `local.py`). please take a look!

sdk/python/feast/infra/online_stores/datastore.py
```diff
@@ -0,0 +1,257 @@
...
+
+class DatastoreOnlineStore(OnlineStore):
+    """
+    OnlineStore is an object used for all interaction between Feast and the service used for offline storage of
+    features.
+    """
+
+    @classmethod
+    def setup(
+        cls,
```
woop, line 51: Just out of curiosity, is there a reason we went for @classmethods here as opposed to normal classes (and using self) or static methods?

achals: Not really any reason, I was thinking of making this regular instance methods too

sdk/python/tests/test_cli_redis.py
```diff
@@ -29,11 +29,11 @@ def test_basic() -> None:
                 f"""
         project: {project_id}
         registry: {data_path / "registry.db"}
-        provider: redis
+        provider: local
         offline_store:
             type: bigquery
         online_store:
-            redis_type: redis
+            type: redis
```
woop, line 36: wait, is this a bug? I believe there is a `type` and a `redis_type`, and they are different.

achals: This is needed cause redis provider assumed the type of the online store to be of redis type - but now there's no redis provider so we need to mention this type explicitly, and use the default `redis_type`, which is `redis`.

woop (approved):

PR #1629, opened 2021-06-08
Add AWS authentication using github actions

tsotnet:

**What this PR does / why we need it**: Adds authenticating in Feast's AWS account using Github actions. This will unlock integration tests using S3, Dynamo, Redshift.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -24,13 +24,21 @@ jobs:
           python-version: ${{ matrix.python-version }}
           architecture: x64
       - name: Set up Cloud SDK
-        uses: google-github-actions/setup-gcloud@master
+        uses: google-github-actions/setup-gcloud@v0
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/integration_tests.yml
```diff
@@ -24,13 +24,21 @@ jobs:
           python-version: ${{ matrix.python-version }}
           architecture: x64
       - name: Set up Cloud SDK
-        uses: google-github-actions/setup-gcloud@master
+        uses: google-github-actions/setup-gcloud@v0
```
woop, line 27: More context here?

tsotnet: Not sure we should be using a master branch here. All actions are versioned with tags, and v0 tag is updated on every new release of v0.x.y. You can see it here: https://github.com/google-github-actions/setup-gcloud/releases

woop: Ok, sgtm

PR #1634, opened 2021-06-10
Vt/to bigquery function

vtao2:

Fixes #

```release-note

```


mavysavydav:
Could we add a parameter for dry run in the method? QueryJobConfig has a dry_run option

MattDelac:
> Could we add a parameter for dry run in the method? QueryJobConfig has a dry_run option

Hi @mavysavydav ,

What would be the `dry-run` used for ? When would that be used with Feast CLI commands ?
This sounds like out of scope for this PR no ?

mavysavydav:
@MattDelac the dry run option could be used to estimate the cost of running that query, and it'd be great to also have the sql query printed when the dry run param is on. It wouldn't be for use with the cli.  For use in notebook when data scientist wants to make sure his/her join is as expected and will cost as expected. This will become even more relevant w the advanced join settings and retrieval time aggregate ops [(link)](https://docs.google.com/document/d/11aXWj9D_hG7-g1syO_0Yb48XkqJbLbg6kpGQjFlLzAI/edit#) that r still being discussed.

MattDelac:
> @MattDelac the dry run option could be used to estimate the cost of running that query

Yeah that's fine. Let's keep this option that is specific to BQ 👍 

> it'd be great to also have the sql query printed when the dry run param is on

For that one you can access the query by simply printing it
```python
historical_retrieval_job = store.get_historical_features(entity_df=sql_query, feature_refs=features)
print(historical_retrieval_job.query)
```

mavysavydav:
great thanks @MattDelac and @vtao2 !

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -206,15 +207,35 @@ def _infer_event_timestamp_from_dataframe(entity_df: pandas.DataFrame) -> str:
...
+        else:
+            while True:
+                query_job = self.client.get_job(bq_job.job_id)
+                if query_job.state in ["PENDING", "RUNNING"]:
+                    print(f"The job is still '{bq_job.state}'. Will wait for 30 seconds")
+                    time.sleep(30)
+                else:
+                    break
+
+        return path
```
MattDelac, line 237: The following would be cleaner

```python
        if dry_run is True:
            print("This query will process {} bytes.".format(bq_job.total_bytes_processed))
            return

        while True:
            query_job = self.client.get_job(bq_job.job_id)
            if query_job.state in ["PENDING", "RUNNING"]:
                print(f"The job is still '{bq_job.state}'. Will wait for 30 seconds")
                time.sleep(30)
            else:
                break

        return path
```

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -206,15 +207,35 @@ def _infer_event_timestamp_from_dataframe(entity_df: pandas.DataFrame) -> str:
...
         # TODO: Ideally only start this job when the user runs "get_historical_features", not when they run to_df()
         df = self.client.query(self.query).to_dataframe(create_bqstorage_client=True)
         return df
 
+    def to_bigquery(self, dry_run=False) -> str:
+        path = f"{self.client.project}.{self.config.offline_store.dataset}.tmp-{str(uuid.uuid4())}"
+        job_config = bigquery.QueryJobConfig(destination=path, dry_run=dry_run)
+        bq_job = self.client.query(self.query, job_config=job_config)
+
+        if dry_run is True:
```
vtao2, line 225: source: https://cloud.google.com/bigquery/docs/dry-run-queries

sdk/python/feast/infra/offline_stores/bigquery.py
vtao2: sounds good, changed

sdk/python/tests/test_historical_retrieval.py
```diff
@@ -441,6 +441,14 @@ def test_historical_features_from_bigquery_sources(
             ],
         )
 
+        bq_temp_table_path = job_from_sql.to_bigquery()
```
MattDelac, line 444: Can you as a test with `to_bigquery(dry_run=True)` please ?

vtao2: done

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -206,15 +207,35 @@ def _infer_event_timestamp_from_dataframe(entity_df: pandas.DataFrame) -> str:
...
         self.query = query
         self.client = client
+        self.config = config
 
     def to_df(self):
         # TODO: Ideally only start this job when the user runs "get_historical_features", not when they run to_df()
         df = self.client.query(self.query).to_dataframe(create_bqstorage_client=True)
         return df
 
+    def to_bigquery(self, dry_run=False) -> str:
```
woop, line 220: > def to_bigquery(self, dry_run=False) -> str:

are we then committing to the idea that execution for all job types only starts when the RetrievalJob object has its `to_X()` methods triggered? (which is different from what I was expecting here https://github.com/feast-dev/feast/pull/1634/files#diff-14f38fcbdc2240bb60b9f558d133b174f5b6059cfcd959eef8b9c49bb5be8838R216, but also a viable approach)

woop: Also, should the dry run configuration not be independent of the destination? I can imagine that `to_df()` and `to_bigquery()` could use the same underlying means of execution or at least configuration.

MattDelac: > are we then committing to the idea that execution for all job types only starts when the RetrievalJob object has its to_X() methods triggered

I feel that this is the best practice IMO

MattDelac: > Also, should the dry run configuration not be independent of the destination?

Unsure about its impact on 
```python
print("This query will process {} bytes.".format(bq_job.total_bytes_processed))
```

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -206,15 +207,35 @@ def _infer_event_timestamp_from_dataframe(entity_df: pandas.DataFrame) -> str:
...
+        path = f"{self.client.project}.{self.config.offline_store.dataset}.tmp-{str(uuid.uuid4())}"
+        job_config = bigquery.QueryJobConfig(destination=path, dry_run=dry_run)
+        bq_job = self.client.query(self.query, job_config=job_config)
+
+        if dry_run is True:
+            print("This query will process {} bytes.".format(bq_job.total_bytes_processed))
+            return
+
+        while True:
+            query_job = self.client.get_job(bq_job.job_id)
```
woop, line 230: I don't understand what this line of code does. What is the difference between the first job and the second job? Isn't it the same job?

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -206,15 +207,35 @@ def _infer_event_timestamp_from_dataframe(entity_df: pandas.DataFrame) -> str:
...
+            if query_job.state in ["PENDING", "RUNNING"]:
+                print(f"The job is still '{bq_job.state}'. Will wait for 30 seconds")
+                time.sleep(30)
```
woop, line 233: Is there anything preventing us from using the existing QueryJob functionality to wait for the job (like `result()` and `done()`)?

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -206,15 +207,35 @@ def _infer_event_timestamp_from_dataframe(entity_df: pandas.DataFrame) -> str:
...
         self.client = client
+        self.config = config
 
     def to_df(self):
         # TODO: Ideally only start this job when the user runs "get_historical_features", not when they run to_df()
         df = self.client.query(self.query).to_dataframe(create_bqstorage_client=True)
         return df
 
+    def to_bigquery(self, dry_run=False) -> str:
+        path = f"{self.client.project}.{self.config.offline_store.dataset}.tmp-{str(uuid.uuid4())}"
```
woop, line 221: should we perhaps use underscores and maybe use a more descriptive prefix for the table, like `training_`. Also, we can possibly add a date for convenience and ordering. 

`training_20210614_c7927c7f`

I would also trim the UUID to just 7 or 8 characters.

sdk/python/feast/infra/offline_stores/bigquery.py
vtao2: Yeah this is to bypass some strange behaviour -- without this, the job state just by using `result.state()` or `.done()` doesn't refresh properly. Encountered this both on Matt's and my end.

sdk/python/feast/infra/offline_stores/bigquery.py
vtao2: see above:)

sdk/python/feast/infra/offline_stores/bigquery.py
vtao2: done!

sdk/python/feast/infra/offline_stores/bigquery.py
MattDelac: > the job state just by using result.state() or .done() doesn't refresh properly

Indeed ! The state does not refresh automatically. So the best practice is to fetch the job through the API over and over ...
If you know a better way to achieve that, please let us know 🙏

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -206,15 +207,40 @@ def _infer_event_timestamp_from_dataframe(entity_df: pandas.DataFrame) -> str:
...
+        path = f"{self.client.project}.{self.config.offline_store.dataset}.training_{today}_{rand_id} "
+
+        job_config = bigquery.QueryJobConfig(destination=path, dry_run=dry_run)
+        bq_job = self.client.query(self.query, job_config=job_config)
+
+        if dry_run is True:
+            print(
+                "This query will process {} bytes.".format(bq_job.total_bytes_processed)
+            )
+            return path
```
MattDelac, line 232: Should we just return nothing when we asked for a dry_run ? I don't understand the utility of the `path` if the table never been created. What I am missing 🤔 ?

```python
if dry_run:
    return

...
return path
```

It just means that this function will output an Optional[str]
```python
def to_bigquery(self, dry_run=False) -> Optional[str]:
```

sdk/python/tests/test_historical_retrieval.py
```diff
@@ -441,6 +441,23 @@ def test_historical_features_from_bigquery_sources(
             ],
         )
 
+        # Just a dry run, should not create table
+        bq_dry_run = job_from_sql.to_bigquery(dry_run=True)
+        with pytest.raises(Exception):
```
MattDelac, line 446: Do you have access to a more accurate exception like `TableNotFound` ?

sdk/python/feast/infra/offline_stores/bigquery.py
vtao2: Good point :) changed :D

sdk/python/tests/test_historical_retrieval.py
vtao2: actually, since we won't return the path, can just test that nothing is returned when calling dry run :)

vtao2: changed it to reflect that

MattDelac (approved):
LGTM

sdk/python/feast/infra/offline_stores/bigquery.py
woop: Also, if we are going to write retry logic, should we just bite the bullet and use https://tenacity.readthedocs.io/en/latest/? I would rather make our code easier to read at the expense of dependencies, and I can see us needing retries in other places as well.

MattDelac: Tenacity looks nice. Thanks for sharing

For this use case, I would rather use BQ functions (like `done()`) as it's available. No need to add dependencies for now if that works

vtao2: > `done()`is supposed to work if you use reload=True

As per the docs you linked, reload=True is default so this would've been the case when we tried.

woop:
@vtao2 Just a small nitpick, but I noticed we're deleting our commits and force pushing the same commit message. This makes it impossible to see the history of changes and also leads to broken links, for example https://github.com/feast-dev/feast/pull/1634/files/6b6398530c69f221629b837225be2870fde75263..4cb2f010de767e8d1aef827c0446a0b968baded9

sdk/python/feast/infra/offline_stores/bigquery.py
MattDelac: To use the `done()` capabilities, we need to permissions too `bigquery.readsessions.create`

image

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -206,15 +208,41 @@ def _infer_event_timestamp_from_dataframe(entity_df: pandas.DataFrame) -> str:
...
 
     def to_df(self):
         # TODO: Ideally only start this job when the user runs "get_historical_features", not when they run to_df()
         df = self.client.query(self.query).to_dataframe(create_bqstorage_client=True)
         return df
 
+    def to_bigquery(self, dry_run=False) -> Optional[str]:
+        @retry(wait=wait_fixed(10), stop=stop_after_delay(1800), reraise=True)
+        def _block_until_done():
+            return bigquery.Client().get_job(bq_job.job_id).state in ["PENDING", "RUNNING"]
```
woop, line 224: Is there a reason to create a new client instead of using `self.client`?

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -206,15 +208,41 @@ def _infer_event_timestamp_from_dataframe(entity_df: pandas.DataFrame) -> str:
...
+
+        today = date.today().strftime("%Y%m%d")
+        rand_id = str(uuid.uuid4())[:7]
+        path = f"{self.client.project}.{self.config.offline_store.dataset}.training_{today}_{rand_id}"
```
woop, line 228: On second thought, some folks might use this `get_historical_features()` for batch scoring, so perhaps lets just make this `.historical_`?

MattDelac: yeah `historical` is a better prefix than `training` 💯 
Forgot to mention that ...

sdk/python/feast/infra/offline_stores/bigquery.py
MattDelac: No. Good catch

vtao2: woops thank you

sdk/python/setup.py
```diff
@@ -54,6 +54,7 @@
     "pydantic>=1.0.0",
     "PyYAML==5.3.*",
     "tabulate==0.8.*",
+    "tenacity==7.*",
```
woop, line 57: I think we can just make it `>=7.*`

vtao2: cool changed :)

woop (approved):

PR #1638, opened 2021-06-11
Don't load entire bigquery query results in memory

tsotnet:

**What this PR does / why we need it**: In `BigQueryOfflineStore.get_historical_features` Feast calls `entity_df_job.result()` which causes the entity data frame to be loaded in memory. We only need to wait for the job to be done and for the schema to be returned, while not changing the existing query (because the results of the query are being used for joining with offline feature data). This PR changes the call to `entity_df_job.result(max_results=0)`, which at the same time:
1. Executes the same query in BigQuery, so the table that gets joined does not change
2. Does not load any of the query rows in memory
3. Still allows the schema to be acquired from the resulting object

I checked all other places where we call `.result()` in the codebase, but none of them needed fixing due to couple of different reasons (happy to elaborate if anyone is curious).

Fixes #

```release-note
Fix BigQuery entity dataframe SQL query results being completely loaded in memory
```

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -92,7 +92,8 @@ class BigQueryOfflineStore(OfflineStore):
 
         if type(entity_df) is str:
             entity_df_job = client.query(entity_df)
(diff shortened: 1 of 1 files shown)


achals (approved):
/lgtm

PR #1641, opened 2021-06-11
Set default feature naming to not include feature view name. Add option to include feature view name in feature naming.

### LINKED ISSUE #1618: Provide option to strip feature view names from feature names
Users often have hardcode logic in model training pipelines based on specific feature names that come from their data sources. When Feast is introduced during productionization, we expect users to rename their feature names to the Feast convention `feature_view__feature`. This creates an unnecessary overhead for teams if they have many features.

Provide a way to strip the `feature_view__` prefix from feature names during retrieval and reuse feature names directly from sources.

before
```
training_df = store.get_historical_features(
    entity_df=entity_df, 
    feature_refs = [
        'driver_performance:conv_rate',
        'driver_performance:acc_rate',
        'driver_hourly_stats:avg_daily_trips'
    ],
)
for feature_name in training_df.columns:
    print(feature_name)
```
```
driver_performance__conv_rate
driver_performance__acc_rate
driver_hourly_stats__avg_daily_trips
```
after
```
training_df = store.get_historical_features(
    entity_df=entity_df, 
    feature_refs = [
        'driver_performance:conv_rate',
        'driver_performance:acc_rate',
        'driver_hourly_stats:avg_daily_trips'
    ],
    feature_names_only=True
)
for feature_name in training_df.columns:
    print(feature_name)
```
```
conv_rate
acc_rate
avg_daily_trips
```

Any collisions in feature names would result in an exception.

Mwad22:
Provides a way to strip the `feature_view__` prefix from feature names during retrieval and reuse feature names directly from sources. This is done by adding a `feature_names_only` arguments to the functions get_historical_features(), and get_online_features(). By default, we set this parameter to True, as that is what most users seem to prefer (per guidance given by @woop).

Fixes #1618 

Yes, the default behaviour of calling get_historical_features() and get_online_features() now return feature names that are stripped of the `feature_view__` prefix. This introduces the possibility of feature name collisions (in which case a descriptive error is thrown, informing users they can switch to the original behaviour by setting the `full_feature_names=True` in their retrieval function).

```release-note
Added option to include feature view name as prefix in feature naming convention.
Default naming convention is now of format "feature_name", with option to name
 features in the format "featureview__feature_name" by setting the 
**full_feature_names=True** argument in get_historical_features() and get_online_features()
```


.gitignore
```diff
@@ -187,4 +187,9 @@ dmypy.json
 # Protos
 sdk/python/docs/html
 sdk/python/feast/protos/
-sdk/go/protos/
\ No newline at end of file
+sdk/go/protos/
+
+# Local Repos
+innocent_redbird/
```
woop, line 193: You can use https://www.atlassian.com/git/tutorials/saving-changes/gitignore#personal-git-ignore-rules instead of modifying this gitignore

sdk/python/feast/feature_store.py
```diff
@@ -278,12 +285,12 @@ def get_historical_features(
             >>> fs = FeatureStore(config=RepoConfig(provider="gcp"))
             >>> retrieval_job = fs.get_historical_features(
             >>>     entity_df="SELECT event_timestamp, order_id, customer_id from gcp_project.my_ds.customer_orders",
-            >>>     feature_refs=["customer:age", "customer:avg_orders_1d", "customer:avg_orders_7d"]
-            >>> )
+            >>>     feature_refs=["customer:age", "customer:avg_orders_1d", "customer:avg_orders_7d"],
+            >>>     feature_names_only=True
```
woop, line 289: Just thinking out loud: If we are going to make the default `True`, would it make more sense to invert the condition? `full_feature_name=False`?

sdk/python/feast/feature_store.py
```diff
@@ -580,12 +601,27 @@ def _group_refs(
...
+            feature_collision_set.add(feat_name)
+        else:
+            feature_set.add(feat_name)
         if view_name not in view_index:
             raise FeatureViewNotFoundException(view_name)
         views_features[view_name].append(feat_name)
 
+    if feature_names_only and len(feature_collision_set) > 0:
+        err = ", ".join(x for x in feature_collision_set)
+        raise ValueError(
```
woop, line 621: Would it be possible to raise a new error type please? `FeatureNameCollisionError`? We can then test for that specific error type being raised, otherwise we might gobble up other ValueErrors.

woop:
Thanks for this @Mwad22!

sdk/python/feast/feature_store.py
mavysavydav: your suggestion makes sense. Seems to be more standard afaik

sdk/python/feast/feature_store.py
```diff
@@ -580,12 +601,27 @@ def _group_refs(
     # view name to feature names
     views_features = defaultdict(list)
 
+    # feature set
```
mavysavydav, line 604: i don't think this comment or the one just below is necessary

pyvenv.cfg
```diff
@@ -0,0 +1,8 @@
+home = /Library/Frameworks/Python.framework/Versions/3.8
```
mavysavydav, line 1: this file shouldn't be included in change set

sdk/python/feast/feature_store.py
```diff
@@ -535,13 +546,21 @@ def get_online_features(
 
                 if feature_data is None:
                     for feature_name in requested_features:
-                        feature_ref = f"{table.name}__{feature_name}"
+                        feature_ref = (
+                            f"{feature_name}"
```
mavysavydav, line 550: is f"..." wrapping needed for this var. same for line 560

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -295,8 +298,10 @@ def build_point_in_time_query(
     max_timestamp: datetime,
     left_table_query_string: str,
     entity_df_event_timestamp_col: str,
+    feature_names_only: bool = True,
 ):
     """Build point-in-time query between each feature view table and the entity dataframe"""
+
```
mavysavydav, line 304: remove accidentally added new line

sdk/python/feast/infra/provider.py
```diff
@@ -202,6 +212,13 @@ def _get_requested_feature_views_to_features_dict(
 
         if not found:
             raise ValueError(f"Could not find feature view from reference {ref}")
+
+    if feature_names_only and len(feature_collision_set) > 0:
+        err = ", ".join(x for x in feature_collision_set)
+        raise ValueError(
```
codyjlin, line 218: Same comment as above (https://github.com/feast-dev/feast/pull/1641/files#r650120378) for using a custom error type like `FeatureNameCollisionError` (Those can be set in [sdk/python/feast/errors.py](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/errors.py))

sdk/python/feast/feature_store.py
Mwad22: I agree with you both, will change the function signatures.

sdk/python/feast/feature_store.py
```diff
@@ -266,6 +273,10 @@ def get_historical_features(
                 SQL query. The query must be of a format supported by the configured offline store (e.g., BigQuery)
             feature_refs: A list of features that should be retrieved from the offline store. Feature references are of
                 the format "feature_view:feature", e.g., "customer_fv:daily_transactions".
+            full_feature_names: By default, this value is set to False. This strips the feature view prefixes from the data
```
mavysavydav, line 276: Should be - "This adds the feature view name to prefix the feature name."

mavysavydav: some of this still refers to the old feature_name_only parameter from the previous version

sdk/python/feast/feature_store.py
```diff
@@ -278,12 +289,12 @@ def get_historical_features(
             >>> fs = FeatureStore(config=RepoConfig(provider="gcp"))
             >>> retrieval_job = fs.get_historical_features(
             >>>     entity_df="SELECT event_timestamp, order_id, customer_id from gcp_project.my_ds.customer_orders",
-            >>>     feature_refs=["customer:age", "customer:avg_orders_1d", "customer:avg_orders_7d"]
-            >>> )
+            >>>     feature_refs=["customer:age", "customer:avg_orders_1d", "customer:avg_orders_7d"],
+            >>>     full_feature_names=False
```
mavysavydav, line 293: maybe this should be set as "True" since in the case of False, the user wouldn't have to specify it.

sdk/python/feast/infra/provider.py
```diff
@@ -179,15 +180,24 @@ def get_provider(config: RepoConfig, repo_path: Path) -> Provider:
 
 
 def _get_requested_feature_views_to_features_dict(
-    feature_refs: List[str], feature_views: List[FeatureView]
+    feature_refs: List[str], feature_views: List[FeatureView], full_feature_names: bool
 ) -> Dict[FeatureView, List[str]]:
-    """Create a dict of FeatureView -> List[Feature] for all requested features"""
+    """Create a dict of FeatureView -> List[Feature] for all requested features.
+    Features are prefixed by the feature view name, set value to True to obtain only the feature names."""
```
mavysavydav, line 186: this comment needs to be updated.

sdk/python/tests/test_historical_retrieval.py
```diff
@@ -314,6 +346,59 @@ def test_historical_features_from_parquet_sources(infer_event_timestamp_col):
             ).reset_index(drop=True),
         )
 
+        event_timestamp = (
```
mavysavydav, line 349: i don't think u need a copy of this. it's already defined above

sdk/python/tests/test_historical_retrieval.py
```diff
@@ -314,6 +346,59 @@ def test_historical_features_from_parquet_sources(infer_event_timestamp_col):
...
+            DEFAULT_ENTITY_DF_EVENT_TIMESTAMP_COL
+            if DEFAULT_ENTITY_DF_EVENT_TIMESTAMP_COL in orders_df.columns
+            else "e_ts"
+        )
+        expected_df_fno = get_expected_training_df(
```
mavysavydav, line 354: lets keep the ordering of these parts the same as the above. define job, then actual_df_fno = job.to_df(), then define the expected df, then asserts. Also actual_df_fno could be named more informatively like actual_df_simple_feature_names

mavysavydav (commented):
The tests in test_historical_features_from_parquet_sources and the bq one could be parametrized. If you look at the annotation at the top of those tests (pytest.mark.parametrize) u can see what's being done. With this parametrization, you don't have to copy all that code just with a different full_feature_names setting. Pytest will automatically run with both settings.

Let's also test collisions directly rather than putting it in the retrieval tests. We want to keep our tests focused, compact, and have a clear idea of what it's testing. You could test def _get_requested_feature_views_to_features_dict and  _group_refs instead where the collision checking logic actually lives.

sdk/python/tests/test_offline_online_store_consistency.py
```diff
@@ -196,7 +196,7 @@ def check_offline_and_online_features(
 ) -> None:
     # Check online store
     response_dict = fs.get_online_features(
-        [f"{fv.name}:value"], [{"driver": driver_id}]
+        [f"{fv.name}:value"], [{"driver": driver_id}], full_feature_names=True
```
mavysavydav, line 199: since this test isn't focused on testing the full_feature_names param, it should just use the default so the full_feature_names param doesn't need to be set in this test, and you can modify the expected feature names.

sdk/python/tests/test_online_retrieval.py
```diff
@@ -98,6 +98,7 @@ def test_online() -> None:
                 "customer_driver_combined:trips",
             ],
             entity_rows=[{"driver": 1, "customer": 5}, {"driver": 1, "customer": 5}],
+            full_feature_names=True,
```
mavysavydav, line 101: for tests that aren't testing full_feature_names, we should leave as default so no need to include this parameter in many of these cases and instead edit the expected output. Then directly test for full_feature_names behavior by having a test with it turned true

Mwad22:
> The tests in test_historical_features_from_parquet_sources and the bq one could be parametrized. If you look at the annotation at the top of those tests (pytest.mark.parametrize) u can see what's being done. With this parametrization, you don't have to copy all that code just with a different full_feature_names setting. Pytest will automatically run with both settings.
> 
> Let's also test collisions directly rather than putting it in the retrieval tests. We want to keep our tests focused, compact, and have a clear idea of what it's testing. You could test def _get_requested_feature_views_to_features_dict and _group_refs instead where the collision checking logic actually lives.

Parametrized the test code so that there's no redundant test code. Also, changed the collision-checking logic so that all checks are performed by _group_refs, and now I directly test the collision logic _group_refs  in separate tests.

sdk/python/tests/test_online_retrieval.py
Mwad22: Changed `full_feature_names` to default for all these tests and edited expected output. I do not test the full_feature_names feature here as it is tested thoroughly in the file `test_offline_online_store_consistency.py`.

docs/quickstart.md
```diff
@@ -234,7 +234,8 @@ feature_vector = store.get_online_features(
         'driver_hourly_stats:acc_rate',
         'driver_hourly_stats:avg_daily_trips'
     ],
-    entity_rows=[{"driver_id": 1001}]
+    entity_rows=[{"driver_id": 1001}],
+    full_feature_names=True
```
woop, line 238: The quickstart should be as simple as possible. Is there a reason to include this argument?

README.md
```diff
@@ -67,6 +67,7 @@ training_df = store.get_historical_features(
         'driver_hourly_stats:acc_rate',
         'driver_hourly_stats:avg_daily_trips'
     ],
+    full_feature_names=True
```
woop, line 70: Same as quickstart.

sdk/python/feast/errors.py
```diff
@@ -71,6 +71,13 @@ def __init__(self, offline_store_name: str, data_source_name: str):
         )
 
 
+class FeatureNameCollisionError(Exception):
+    def __init__(self, feature_name_collisions: str):
+        super().__init__(
+            f"The following feature name(s) have collisions: {feature_name_collisions}. Set 'full_feature_names' argument in the data retrieval function to True to use the full feature name which is prefixed by the feature view name."
```
woop, line 77: Perhaps `Set 'full_feature_names=True'`?

docs/quickstart.md
Mwad22: I was under the impression that it aims to show all the features in as simple a way as possible (so the ability to prefix feature names with their respective feature views by setting `full_feature_names=True`). But if the idea is to keep it as simple as possible I can remove the full_feature_names argument from quickstart and the README.

woop: The idea is to show the core functionality in as simple a way as possible, not all functionality.

README.md
```diff
@@ -75,11 +75,11 @@ print(training_df.head())
 # model = ml.fit(training_df)
 ```
 ```commandline
-      event_timestamp  driver_id  driver_hourly_stats__conv_rate  driver_hourly_stats__acc_rate
-  2021-04-12 08:12:10       1002                        0.497279                       0.357702
```
mavysavydav, line 79: how come these values are changed? were they out of date and different from what you get when you run these commands with the sample data?

Mwad22: I reran the code above to get the output in the new format (non-prefixed feature names) and these were the values I got. When I rerun the readme instructions from scratch in a new repository, I obtain a new set of values so I believe a different set is generated each time you create a feature repository using `feast init` (?).

I think the example is moreso to display the expected format of the output, although I'm not sure why the original example was missing a `avg_daily_trips` column in its output.

mavysavydav: cool got it

mavysavydav:
LGTM

README.md
```diff
@@ -75,11 +75,11 @@ print(training_df.head())
...
-  2021-04-12 10:59:42       1001                        0.979747                       0.008166
-  2021-04-12 15:01:12       1004                        0.151432                       0.551748
-  2021-04-12 16:40:26       1003                        0.951506                       0.753572
+            event_timestamp  driver_id  conv_rate  acc_rate  avg_daily_trips
+0 2021-04-12 08:12:10+00:00       1002   0.713465  0.597095              531
+1 2021-04-12 10:59:42+00:00       1001   0.072752  0.044344               11
```
woop, line 80: Are these row numbers new?

sdk/python/feast/feature_store.py
```diff
@@ -266,6 +273,9 @@ def get_historical_features(
                 SQL query. The query must be of a format supported by the configured offline store (e.g., BigQuery)
             feature_refs: A list of features that should be retrieved from the offline store. Feature references are of
                 the format "feature_view:feature", e.g., "customer_fv:daily_transactions".
+            full_feature_names: By default, this value is set to False. By setting the value to True, this adds the
```
woop, line 276: Can we please talk about defaults at the end of the description, not the start?

sdk/python/feast/feature_store.py
```diff
@@ -597,24 +621,40 @@ def _group_refs(
     # view name to feature names
     views_features = defaultdict(list)
 
+    feature_set = set()
+    feature_collision_set = set()
```
woop, line 625: Is there a reason why validation is mixed into `group_refs()` and not a standalone function?

sdk/python/tests/test_historical_retrieval.py
```diff
@@ -246,7 +271,12 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
 @pytest.mark.parametrize(
     "infer_event_timestamp_col", [False, True],
 )
-def test_historical_features_from_parquet_sources(infer_event_timestamp_col):
+@pytest.mark.parametrize(
+    "full_feature_names", [False, True],
```
woop, line 275: If we add this parametrize, then we double all end to end tests we are running on BigQuery, which could eventually amount to a lot of money. We are planning to refactor the tests to make them more modular, but for the time being is there a way to incorporate `full_feature_names` in such a way that it doesnt double our tests?

sdk/python/tests/test_offline_online_store_consistency.py
```diff
@@ -193,34 +193,51 @@ def check_offline_and_online_features(
...
+        if expected_value:
+            assert abs(df.to_dict()["value"][0] - expected_value) < 1e-6
+        else:
+            df = df.where(pd.notnull(df), None)
+            assert df.to_dict()["value"][0] is None
 
 
 def run_offline_online_store_consistency_test(
-    fs: FeatureStore, fv: FeatureView
+    fs: FeatureStore, fv: FeatureView, ffn: bool
```
woop, line 240: can we spell out the argument?

sdk/python/tests/test_historical_retrieval.py
mavysavydav: gotcha, @Mwad22 i think you could write a new test method that can be very similar to the stuff in this test. I think if you put the test code within this test method even without parametrization, it'll still run that new part you add multiple times because this is parametrized by other things. WDYT? definitely feel free to suggest any other ideas

Mwad22: So, I think it might be beneficial to still have these end to end tests that test the feature name, just to make sure that we are retrieving the correct whether the `full_feature_names` parameter is set to True or False. However, (as @mavysavydav suggested) I can place these tests in a separate test function so that those that are unrelated to (changes in) the `full_feature_names` are no longer parametrized by it. That should reduce the total number of tests. Please let me know what your thoughts are on this.

woop: > I can place these tests in a separate test function so that those that are unrelated to (changes in) the full_feature_names are no longer parametrized by it

I don't understand what this sentence means. What is "those"? and what is "it"?

Mwad22: Oh my bad, I'll clarify. So, 'those' is in reference to the tests that do not need/involve testing the `full_feature_names` functionality. And `it` refers to the full_feature_names parameter. What I meant was I can place all tests that test `full_feature_names` functionality in a separate test function. So tests that do not need to be run with `parametrize(
    "full_feature_names", [False, True])` no longer do so.

README.md
Mwad22: Yes, after running the code in the readme, this is the output

sdk/python/tests/test_historical_retrieval.py
woop: What bothers me is that this single method is becoming a monster. There are too many conditionals. I am open to the idea of having all these parameterizations and even loads of different combinations, but then the test itself should run really quickly.

What if we added a copy of this test that is very small. Like a `test_historical_features_from_bigquery_sources_minimal()`. We then try to ensure that it only uses like 100-1000 rows and executes all of the tests very quickly with all the combinations. I havent checked, but I believe the existing tests actually uses a lot more data. (Btw I only care about BQ. It's fine to let Parquet run over and over since its fast).

sdk/python/feast/feature_store.py
Mwad22: The logic in `_group_refs` lends itself well to checking for feature name collisions, but more importantly I think placing the check here is reasonable considering the description of the `_group_refs` function: """ Get list of feature views and corresponding feature names based on feature references""".

If the feature names are invalid (e.g. have a collision) this would be a fitting place to throw the error, I believe. Do let me know if you disagree.

sdk/python/tests/test_offline_online_store_consistency.py
Mwad22: Will do! I'm assuming you would like me to fully spell out the original `fs` and `fv` arguments also?

sdk/python/feast/feature_store.py
woop: `_group_refs` is simply not descriptive since its not meant for validation and doesn't hint at what you are doing. This can be seen in your tests when you call `_group_refs` but its not clear why you are calling it.

It seems like the only benefit to using this method is that your a looping over the names once.

Can we at least change the name to something like `_validate_and_group_feature_refs`?

sdk/python/tests/test_historical_retrieval.py
```diff
@@ -560,3 +608,79 @@ def test_historical_features_from_bigquery_sources(
...
+    customer_source = BigQuerySource(
+        table_ref=customer_table_id,
+        event_timestamp_column="datetime",
+        created_timestamp_column="",
+    )
+    customer_fv = create_customer_daily_profile_feature_view(customer_source)
+
+    # _group_refs is the function that checks for colliding feature names
+    with pytest.raises(FeatureNameCollisionError):
+        _group_refs(
```
woop, line 675: Do we need to write to BigQuery in order to test a function? This seems like a unit test.

sdk/python/tests/test_offline_online_store_consistency.py
woop: both `fs` and `fv` are indicative of the primitives we use. Most people will understand what they are. `ffn` was mysterious to me.

sdk/python/tests/test_historical_retrieval.py
Mwad22: So after doing a little digging, I believe the data being user is limited to 1000 rows by the hardcoded `order_count` in the `generate_entities` function [here](https://github.com/Mwad22/feast/blob/d59e0c3cebd49306dbe8b6006400117435acbae7/sdk/python/tests/test_historical_retrieval.py#L37). Is this sufficient or would a making the order count a variable parameter (and setting it to 100) be more ideal?

sdk/python/tests/test_historical_retrieval.py
Mwad22: I don't think I write to BigQuery here, but I did notice that I do write to a temporary directory in the test above the one you highlighted (using parquet sources) which I will change. Do let me know if I'm missing something.

sdk/python/tests/test_historical_retrieval.py
woop: I think it's probably fine in that case.

Mwad22: Cool, I added the `order_count` parameter to the `generate_entities()` so we are able to vary the number of rows generated. It's an optional parameter with the default value of `order_count=1000`, so that this change does not interfere with any of the current code.

sdk/python/tests/test_historical_retrieval.py
```diff
@@ -174,6 +177,10 @@ def get_expected_training_df(
                 f"driver_stats__{k}": driver_record.get(k, None)
                 for k in ("conv_rate", "avg_daily_trips")
             }
+            if full_feature_names
+            else {
+                k: driver_record.get(k, None) for k in ("conv_rate", "avg_daily_trips")
```
tsotnet, line 182: I think it'll be cleaner if you handle if/else only with k / f"driver_stats__{k}":
```python
formatted_k = f"driver_stats__{k}" if full_feature_names else k
order_record.update(
  {
      formatted_k: driver_record.get(k, None)
      for k in ("conv_rate", "avg_daily_trips")
  }
```

sdk/python/tests/test_historical_retrieval.py
```diff
@@ -184,6 +191,15 @@ def get_expected_training_df(
...
             }
+            if full_feature_names
+            else {
+                k: customer_record.get(k, None)
+                for k in (
+                    "current_balance",
+                    "avg_passenger_count",
+                    "lifetime_trip_count",
+                )
+            }
```
tsotnet, line 202: Similar comment here - let's just introduce `formatted_k` to deduplicate code.

sdk/python/tests/test_historical_retrieval.py
```diff
@@ -557,3 +605,74 @@ def test_historical_features_from_bigquery_sources(
...
+                "customer_profile:avg_passenger_count",
+                "customer_profile:lifetime_trip_count",
+                "customer_profile:avg_daily_trips",
+            ],
+            all_feature_views=[driver_fv, customer_fv],
+            full_feature_names=False,
+        )
+
+
+def test_feature_name_collision_on_historical_retrieval_from_bigquery_sources():
```
tsotnet, line 642: This logic isn't specific to data sources, so I wouldn't add this test case unnecessarily

tsotnet (changes requested):

sdk/python/feast/feature_store.py
```diff
@@ -594,8 +618,10 @@ def _entity_row_to_field_values(
     return result
 
 
-def _group_refs(
-    feature_refs: List[str], all_feature_views: List[FeatureView]
+def _validate_and_group_feature_refs(
```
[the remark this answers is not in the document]
woop: Previous discussion for reference https://github.com/feast-dev/feast/pull/1641#discussion_r655780532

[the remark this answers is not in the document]
tsotnet: @woop thanks for the reference.

@Mwad22
> The logic in _group_refs lends itself well to checking for feature name collisions, but more importantly I think placing the check here is reasonable considering the description of the _group_refs function: """ Get list of feature views and corresponding feature names based on feature references""".
> 
> If the feature names are invalid (e.g. have a collision) this would be a fitting place to throw the error, I believe. Do let me know if you disagree.

The purpose of `_group_refs` is to group features by FeatureView. The method works fine even if there are collisions. This is more obvious if you look at the modified lines: no line is removed and no existing variable inside the function touched. The new code only introduces new variables and adds an additional logic for handling collisions. If _group_refs intrinsically wouldn't work during collisions, I'd agree with you. But for these reasons, I think it still makes sense to split the methods into validation & grouping.

sdk/python/feast/feature_store.py
```diff
@@ -286,18 +296,18 @@ def get_historical_features(
...
-
         all_feature_views = self._registry.list_feature_views(project=self.project)
         try:
             feature_views = _get_requested_feature_views(
-                feature_refs, all_feature_views
+                feature_refs, all_feature_views, full_feature_names
             )
-        except FeatureViewNotFoundException as e:
+        except (FeatureNameCollisionError, FeatureViewNotFoundException) as e:
             sys.exit(e)
```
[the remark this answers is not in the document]
Mwad22: @tsotnet I see there is a separate issue for this [here](https://github.com/feast-dev/feast/issues/1666). I am currently finishing the suggested changes, would you like me to 'Remove sys.exit from exception handling' in this PR? or should I open a separate one and attach it to the new issue.

sdk/python/feast/feature_store.py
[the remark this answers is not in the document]
Mwad22: Sounds good,  I just committed the changes requested here and across the other comments. Thank you for the detailed explanations, they make a lot of sense and are appreciated on my end!

sdk/python/feast/feature_store.py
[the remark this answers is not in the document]
Mwad22: I decided to remove the try/except blocks and allow the exceptions to bubble up, per your suggestion. If we can merge this PR then that should resolve the other issue mentioned above too. Please let me know if you'd like me to do anything differently!

sdk/python/feast/errors.py
```diff
@@ -78,6 +78,16 @@ def __init__(self, offline_store_name: str, data_source_name: str):
         )
 
 
+class FeatureNameCollisionError(Exception):
```
woop, line 81: It's not clear to me why we are passing in refs here when we don't use the feature view name. What if we had something like 

```
def __init__(self, feature_views: List[str], feature_name: str):
```
That would be easier for me to understand as well. The current argument is cryptic.

sdk/python/feast/errors.py
```diff
@@ -78,6 +78,16 @@ def __init__(self, offline_store_name: str, data_source_name: str):
...
+    def __init__(self, feature_refs_collisions: List[str]):
+        feature_name_collisions = [ref.split(":")[1] for ref in feature_refs_collisions]
+        feature_names = ", ".join(x for x in feature_name_collisions)
+        super().__init__(
+            f"The following feature name(s) have collisions: {feature_names}. Set 'full_feature_names' "
```
woop, line 86: We should definitely talk about the feature view/table names that have collided.

woop, line 86: The option provided to the user is not the only way to solve the problem. Ensuring that the features have different names is another option.

sdk/python/feast/feature_store.py
```diff
@@ -599,7 +612,17 @@ def _entity_row_to_field_values(
     return result
 
 
-def _group_refs(
+def _validate_feature_refs(feature_refs: List[str], full_feature_names: bool = False):
+    feature_names = [ref.split(":")[1] for ref in feature_refs]
+    feature_name, count = Counter(feature_names).most_common(1)[0]
```
woop, line 617:
```suggestion
    feature_name, occurrences = Counter(feature_names).most_common(1)[0]
```

sdk/python/feast/feature_store.py
```diff
@@ -599,7 +612,17 @@ def _entity_row_to_field_values(
     return result
 
 
-def _group_refs(
+def _validate_feature_refs(feature_refs: List[str], full_feature_names: bool = False):
```
woop, line 615: Should we validate that the same feature name does not occur in the same feature view more than once?

sdk/python/feast/infra/provider.py
```diff
@@ -164,13 +165,16 @@ def get_provider(config: RepoConfig, repo_path: Path) -> Provider:
 def _get_requested_feature_views_to_features_dict(
     feature_refs: List[str], feature_views: List[FeatureView]
 ) -> Dict[FeatureView, List[str]]:
-    """Create a dict of FeatureView -> List[Feature] for all requested features"""
+    """Create a dict of FeatureView -> List[Feature] for all requested features.
+    Set full_feature_names to True to get feature names prefixed by its featureview."""
```
woop, line 169:
```suggestion
    Set full_feature_names to True to have feature names prefixed by their feature view name."""
```

sdk/python/feast/feature_store.py
Mwad22: Yes, that seems necessary. I just pushed the change, see comment above for example!

sdk/python/feast/errors.py
Mwad22: I have made the error message much more descriptive, hopefully that clears things up. For example, now a list of feature refs with 2 feature refs `driver_hourly_stats:conv_rate` would return the error message below. This generalizes across multiple feature views also.

```
The following feature name(s) have collisions: conv_rate. Set 'full_feature_names' argument in the data retrieval function to True to use the full feature name which is prefixed by the feature view name, or rename colliding features.
Collisions occur in the following feature view(s): driver_hourly_stats.
```

woop: Thanks @Mwad22, going to be nitpicky now.

I took a stab at the error message for collisions where prefixes are missing and the name is the same.
```
Duplicate features named conv_rate found.

To resolve this collision, either use the full feature name by setting 'full_feature_names=True', or ensure that the features in question have different names.
```
or when prefixes are already in use

```
Duplicate features named driver_conv_rate found.

To resolve this collision, please ensure that the features in question have different names.
```

Mwad22: So, for the case where feature names are not prefixed, we no longer want to mention the feature views where the collisions occur?

Mwad22: @woop I went ahead and pushed the change. Feel free to ignore my question above, as I think the error messages are descriptive enough in the cases you describe above.

woop:
I will let @tsotnet do the final approval.

woop:
@tsotnet @achals @felixwang9817 please note the label I am adding to this PR. We're introducing a breaking change. We probably don't want to backport this functionality. If we do want to backport it, then the default should be to maintain the prefixes. I'm assuming we don't, and that we should just ship 0.12 faster.

woop:
@Mwad22 mind fixing the DCO problem?

Mwad22:
> @Mwad22 mind fixing the DCO problem?

All fixed, and I believe everything else is passing. If there's nothing else you or @tsotnet would like changed, can we ship this?I only ask because the changes touch many files so the longer we wait, the greater number of merge conflicts we'll have to resolve.

sdk/python/feast/feature_view.py
```diff
@@ -182,6 +182,8 @@ def to_proto(self) -> FeatureViewProto:
             ttl_duration = Duration()
             ttl_duration.FromTimedelta(self.ttl)
 
+        print(f"Stream soruce: {self.stream_source}, {type(self.stream_source)}")
```
woop, line 185: Should this be removed?

woop:
> > @Mwad22 mind fixing the DCO problem?
> 
> All fixed, and I believe everything else is passing. If there's nothing else you or @tsotnet would like changed, can we ship this?I only ask because the changes touch many files so the longer we wait, the greater number of merge conflicts we'll have to resolve.

Unfortunately @tsotnet's power is offline today. I will give the PR one final scan.

sdk/python/feast/feature_view.py
achals: This has been removed in #1678

sdk/python/feast/infra/aws.py
```diff
@@ -129,6 +129,7 @@ def get_historical_features(
         entity_df: Union[pandas.DataFrame, str],
         registry: Registry,
         project: str,
+        full_feature_names: bool = False,
```
woop, line 132: Are defaults (`False`) necessary here?

sdk/python/feast/feature_view.py
Mwad22: Will remove and push, must have gotten added when I rebased.

sdk/python/feast/errors.py
```diff
@@ -88,6 +88,27 @@ def __init__(self, offline_store_name: str, data_source_name: str):
         )
 
 
+class FeatureNameCollisionError(Exception):
+    def __init__(self, feature_refs_collisions: List[str], full_feature_names):
```
achals, line 92: nit, type annotation for `full_feature_names`?

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -121,7 +122,7 @@ def get_historical_features(
 
         # Build a query context containing all information required to template the BigQuery SQL query
         query_context = get_feature_view_query_context(
-            feature_refs, feature_views, registry, project
+            feature_refs, feature_views, registry, project, full_feature_names
```
achals, line 125: nit, `full_feature_names= full_feature_names`

sdk/python/feast/infra/aws.py
Mwad22: I think just setting this default in `feature_store.py` may be sufficient. Will remove these default assignments from `aws.py`, `gcp.py`, `local.py` and `provider.py`.

sdk/python/tests/test_historical_retrieval.py
```diff
@@ -547,7 +588,26 @@ def test_historical_features_from_bigquery_sources(
...
             check_dtype=False,
         )
+
         table_from_df_entities = job_from_df.to_arrow()
         assert_frame_equal(
             actual_df_from_df_entities, table_from_df_entities.to_pandas()
         )
+
+
+@pytest.mark.integration
```
achals, line 598: Is there a reason this is an integration test?

sdk/python/tests/test_historical_retrieval.py
```diff
@@ -547,7 +588,26 @@ def test_historical_features_from_bigquery_sources(
...
+def test_feature_name_collision_on_historical_retrieval():
+
+    # _validate_feature_refs is the function that checks for colliding feature names
+    with pytest.raises(FeatureNameCollisionError):
```
achals, line 602: It'd be good we can assert the contents of the error message and that it points to the right column that has collisions.

achals (approved):
Thanks @Mwad22, this looks great to me. I had some super nitty things but happy to lgtm after they are addressed.

sdk/python/tests/test_historical_retrieval.py
Mwad22: Not anymore, will make that change.

sdk/python/tests/test_historical_retrieval.py
Mwad22: @achals Done! Thanks for the review, I added assertion for each of the two possible error messages alongside the other changes you requested. Hopefully it's good to go now.

Mwad22:
Note, the unit tests that fail above are failing because of the following error:
```
An error occurred (LimitExceededException) when calling the CreateTable operation: Subscriber limit exceeded: There is a limit of 256 tables per subscriber
```
They are not caused by the changes in this PR.

PR #1642, opened 2021-06-11
Don't use .result() in BigQueryOfflineStore, since it still leads to OOM

tsotnet:

**What this PR does / why we need it**: Follow up on #1638, which failed to fix the OOM issue.

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -24,6 +24,7 @@ try:
     from google.api_core.exceptions import NotFound
     from google.auth.exceptions import DefaultCredentialsError
     from google.cloud import bigquery
+    from google.cloud.bigquery import QueryJob
 
 except ImportError as e:
     from feast.errors import FeastExtrasDependencyImportError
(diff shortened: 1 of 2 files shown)


sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -91,18 +92,28 @@ def get_historical_features(
...
 
         if type(entity_df) is str:
-            entity_df_job = client.query(entity_df)
-            # Start the job and get the schema back. We don't need the actual rows.
-            entity_df_result = entity_df_job.result(max_results=0)
+            entity_df_job: QueryJob = client.query(entity_df)
+            # Note: Don't call entity_df_job.result() here, it will lead
+            # to OOM errors. Instead just wait until the job is done.
+            while not entity_df_job.done():
+                time.sleep(1)
```
woop, line 99: `done()` has a timeout and retry btw.

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -91,18 +92,28 @@ def get_historical_features(
...
 
+            entity_df_sql_table = f"`{entity_df_job.destination.project}.{entity_df_job.destination.dataset_id}.{entity_df_job.destination.table_id}`"
+
+            # Get a single row from this entity table to validate its schema
+            entity_df_limited_job: QueryJob = client.query(
```
woop, line 104: This seems like a lighter weight approach https://github.com/feast-dev/feast/pull/1495#discussion_r619486530

sdk/python/feast/infra/offline_stores/bigquery.py
tsotnet: Didn't realize that, but with the new approach won't be using done(), so I'm resolving this.

sdk/python/feast/infra/offline_stores/bigquery.py
tsotnet: I'm trying this one out now. The previous approach still didn't work for super large datasets. Looks like the anonymous tables were not created when the dataset is large. I'll try creating named table instead.

woop (approved):

PR #1627, opened 2021-06-08
Provide descriptive error on invalid table reference

codyjlin:
Signed-off-by: Cody Lin codyl@twitter.com

**What this PR does / why we need it**:
This PR creates a more useful error message when a user specifies an invalid table. It catches it before any other operations, like trying to infer the DataSource `event_timestamp_column`.

**Which issue(s) this PR fixes**:
Fixes #1596

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #1646, opened 2021-06-13
Schema Inferencing should happen at apply time

### LINKED ISSUE #1645: Clean up inferencing code
A few small refactors are necessary for our feature inferencing code.

All inferencing of columns from sources should only happen during `apply` time, not during the construction of objects. Otherwise these objects will execute eagerly during instantiation which prevents Feast from controlling the execution order and may even lead to duplicate executions.

One example is over here: https://github.com/feast-dev/feast/blob/f29cb8cca5d87b2f9a56fd02428e9a9f014a079f/sdk/python/feast/data_source.py#L676

Also, inferencing code should be more cleanly separated. Right now we have a DataSource being subclassed to implementations, which reference the DataSource base class, which contains hardcoded references to the same child classes, which then call methods on this child classes. The interdependencies are overly complex and cyclic. 
https://github.com/feast-dev/feast/commit/f55b51cf25efbf846e65115cfa23fe2c4a8d9edd#diff-fada7a84510bd3d3051cffc414ee51ed92eff4cdd4c2ac53a1b8b7c47286cab4R523

We can maybe just extract `_infer_event_timestamp_column` code into a function on its own.

Finally, we don't need to query tables in order to get schemas. We can simply use the `get_table()` method and inspect the schema instead of pulling data to the client https://github.com/feast-dev/feast/commit/f55b51cf25efbf846e65115cfa23fe2c4a8d9edd#diff-fada7a84510bd3d3051cffc414ee51ed92eff4cdd4c2ac53a1b8b7c47286cab4R756

mavysavydav:
https://github.com/feast-dev/feast/issues/1645

Fixes #1645

```release-note
None
```

### DIFF
--- a/sdk/python/feast/data_source.py
+++ b/sdk/python/feast/data_source.py
@@ -519,41 +519,7 @@ class DataSource:
         Converts an DataSourceProto object to its protobuf representation.
         """
         raise NotImplementedError
-
-    def _infer_event_timestamp_column(self, ts_column_type_regex_pattern):
-        ERROR_MSG_PREFIX = "Unable to infer DataSource event_timestamp_column"
-        USER_GUIDANCE = "Please specify event_timestamp_column explicitly."
-
-        if isinstance(self, FileSource) or isinstance(self, BigQuerySource):
-            event_timestamp_column, matched_flag = None, False
-            for col_name, col_datatype in self.get_table_column_names_and_types():
-                if re.match(ts_column_type_regex_pattern, col_datatype):
-                    if matched_flag:
-                        raise TypeError(
-                            f"""
-                            {ERROR_MSG_PREFIX} due to multiple possible columns satisfying
-                            the criteria. {USER_GUIDANCE}
-                            """
-                        )
-                    matched_flag = True
-                    event_timestamp_column = col_name
-            if matched_flag:
-                return event_timestamp_column
-            else:
-                raise TypeError(
-                    f"""
-                    {ERROR_MSG_PREFIX} due to an absence of columns that satisfy the criteria.
-                     {USER_GUIDANCE}
-                    """
-                )
-        else:
-            raise TypeError(
-                f"""
-                {ERROR_MSG_PREFIX} because this DataSource currently does not support this inference.
-                 {USER_GUIDANCE}
-                """
-            )
-
+        
 
 class FileSource(DataSource):
     def __init__(
@@ -598,7 +564,7 @@ class FileSource(DataSource):
         self._file_options = FileOptions(file_format=file_format, file_url=file_url)
 
         super().__init__(
-            event_timestamp_column or self._infer_event_timestamp_column(r"^timestamp"),
+            event_timestamp_column,
             created_timestamp_column,
             field_mapping,
             date_partition_column,
@@ -672,8 +638,7 @@ class BigQuerySource(DataSource):
         self._bigquery_options = BigQueryOptions(table_ref=table_ref, query=query)
 
         super().__init__(
-            event_timestamp_column
-            or self._infer_event_timestamp_column("TIMESTAMP|DATETIME"),
+            event_timestamp_column,
             created_timestamp_column,
             field_mapping,
             date_partition_column,
@@ -897,3 +862,4 @@ class KinesisSource(DataSource):
         data_source_proto.date_partition_column = self.date_partition_column
 
         return data_source_proto
+
--- a/sdk/python/feast/data_source.py
+++ b/sdk/python/feast/data_source.py
@@ -708,20 +708,12 @@ class BigQuerySource(DataSource):
         from google.cloud import bigquery
 
         client = bigquery.Client()
-        name_type_pairs = []
         if self.table_ref is not None:
-            project_id, dataset_id, table_id = self.table_ref.split(".")
-            bq_columns_query = f"""
-                SELECT COLUMN_NAME, DATA_TYPE FROM {project_id}.{dataset_id}.INFORMATION_SCHEMA.COLUMNS
-                WHERE TABLE_NAME = '{table_id}'
-            """
-            table_schema = (
-                client.query(bq_columns_query).result().to_dataframe_iterable()
-            )
-            for df in table_schema:
-                name_type_pairs.extend(
-                    list(zip(df["COLUMN_NAME"].to_list(), df["DATA_TYPE"].to_list()))
-                )
+            table_schema = client.get_table(self.table_ref).schema
+            if not isinstance(table_schema[0], bigquery.schema.SchemaField):
+                raise TypeError("Could not parse BigQuery table schema.")
+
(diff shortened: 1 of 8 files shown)


sdk/python/feast/data_source.py
```diff
@@ -598,7 +563,7 @@ def __init__(
         self._file_options = FileOptions(file_format=file_format, file_url=file_url)
 
         super().__init__(
-            event_timestamp_column or self._infer_event_timestamp_column(r"^timestamp"),
+            event_timestamp_column or "",
```
achals, line 566: can this just be `None` instead of an empty string?

sdk/python/feast/data_source.py
```diff
@@ -672,8 +637,7 @@ def __init__(
         self._bigquery_options = BigQueryOptions(table_ref=table_ref, query=query)
 
         super().__init__(
-            event_timestamp_column
-            or self._infer_event_timestamp_column("TIMESTAMP|DATETIME"),
+            event_timestamp_column or "",
```
achals, line 640: Same comment, any reason to use an empty string instead of `None`?

sdk/python/feast/inference.py
```diff
@@ -45,12 +48,72 @@ def infer_entity_value_type_from_feature_views(
...
+def infer_event_timestamp_column_for_data_sources(
+    data_sources: List[DataSource],
+) -> List[DataSource]:
+    ERROR_MSG_PREFIX = "Unable to infer DataSource event_timestamp_column"
+
+    for data_source in data_sources:
+        if (
+            data_source.event_timestamp_column is None
+            or data_source.event_timestamp_column == ""
+        ):
```
achals, line 72: Should we also verify that this timestamp column actually exists in the schema or not here?

sdk/python/feast/inference.py
```diff
@@ -45,12 +48,72 @@ def infer_entity_value_type_from_feature_views(
...
                     )
 
                 entity.value_type = inferred_value_type
 
     return entities
+
+
+def infer_event_timestamp_column_for_data_sources(
+    data_sources: List[DataSource],
+) -> List[DataSource]:
```
achals, line 65: Nit, is there any reason for this to return the same datasources? Since the method just updates them in-place. Also rename to `update_data_sources_with_event_timestamp_column` or something?

sdk/python/feast/feature_view.py
```diff
@@ -241,3 +219,35 @@ def most_recent_end_time(self) -> Optional[datetime]:
         if len(self.materialization_intervals) == 0:
             return None
         return max([interval[1] for interval in self.materialization_intervals])
+
+    def infer_features_from_input_source(self):
+        if not self.features:
```
achals, line 224: Same comment as the datasource, should we assert that the features here are actually present in the input schema? or is that already happening elsewhere in the code?

achals (commented):
Thanks for the PR @mavysavydav ! Left some comments about some cases that I wasn't sure are covered elsewhere in the codebase or not. Looks good otherwise!

sdk/python/feast/repo_operations.py
```diff
@@ -136,6 +139,12 @@ def apply_total(repo_config: RepoConfig, repo_path: Path):
         ),
         feature_views=repo.feature_views,
     )
+    infer_event_timestamp_column_for_data_sources(
+        [view.input for view in repo.feature_views]
+    )
+    for view in repo.feature_views:
+        view.infer_features_from_input_source()
+
```
codyjlin, line 147: Is there an ordering requirement that these be put up here? I think that in lines 162-166 [below](https://github.com/feast-dev/feast/pull/1646/files#diff-9922445cc092f7a2f8885c771cb4d15c4c21a84bcbc0513ba76009036ddbccf5R162-R166), we validate if the data sources are even supported (and in https://github.com/feast-dev/feast/pull/1627, whether the BQ tables even exist). I'm happy to change the order in #1627 later, but wanted to ask now about whether order matters.

mavysavydav: good call

sdk/python/feast/inference.py
mavysavydav: hmmm i think it should be done separately as a validate_data_source or something since there r other fields set in the datasource that may benefit from validation too. If it were the case that one passes a ts_column that doesn't exist and we'd want to infer it, then i think it would make more sense to have it here. But i think if it doesn't exist in the schema, we probably want to just throw exception rather than inferring it. It'd be nice if in the case that this method weren't called anymore, the only functionality lost is the inference, not any other needed checks.

sdk/python/feast/feature_view.py
mavysavydav: refer to reply to comment in datasource. I don't believe it's happening right now anywhere.

sdk/python/feast/inference.py
mavysavydav: sounds good

sdk/python/feast/data_source.py
mavysavydav: yep can delete the "or ""' part. Just need to change default value for event_timestamp_column of base class (DataSource) to None instead of "". Was hesitant to change base class param since didn't know if the use of "" was intentional.

mavysavydav: oh actually, it seems like the data source base class doesn't have event_timestamp_column as optional so its type is str instead of Option[str]. The subclasses constructors use Option[str] so when passing to the supers' constructor need to handle the type checking and `or ""` turns the type from a Option[str] to str. I'll add comments to make it clear this is for type checking.

mavysavydav:
Unsure why telemetry is failing

sdk/python/feast/data_source.py
achals: Could we make the `event_timestamp_column` optional in the base data source class? I think that would be reasonable if possible.

sdk/python/feast/inference.py
achals: Fair, I think we can make a different PR for this validation.

sdk/python/feast/feature_view.py
achals: Sounds good

achals:
Thanks for the patience @mavysavydav! I only have one suggestion, and I think this will be good to merge.

sdk/python/feast/data_source.py
mavysavydav: yep can do. Was thinking that not all subclasses of DataSource base class is guaranteed to have inference so maybe from an interface point of view, it's good to communicate that DataSource actually needs a event_timestamp_column.

mavysavydav: resolved on slack. Achal - "I think that the subclasses can add additional assertions as per their requirements, but the base class should be more broad". Will make the change

achals (approved):
/lgtm

Issue #1605, opened 2021-06-01
Remove hardcoded references to providers and stores


woop:
As per https://docs.google.com/document/d/1MD0aS2_hGzd1tJ7DNjE3NgEtcuekh3O06OeQ9aavylY/edit#, we need to ensure that there aren't any hardcoded references in our code to Provider or Store implementations.

achals:
@woop just to be clear you mean that we should be hardcoding to any specific instance of a store or a provider, right?

woop:
This card is done.

PR #1649, opened 2021-06-15
Refactor OnlineStoreConfig classes into owning modules

achals:
This PR:
- introduces a new pydantic base class for Config classes to extend
- moves the online store config classes into the corresponding modules
- adds logic to dynamically load the `OnlineStoreConfig` classes (and also the `OnlineStore` classes in a second pass).
- Keeps tests basically the same (and adding more tests for the dynamic classes)

Part 1 for #1605 

```release-note
Allow OnlineStore implementations to be loaded dynamically.
```

### DIFF
--- a/sdk/python/feast/errors.py
+++ b/sdk/python/feast/errors.py
@@ -42,15 +42,15 @@ class FeastProviderNotImplementedError(Exception):
         super().__init__(f"Provider '{provider_name}' is not implemented")
 
 
-class FeastProviderModuleImportError(Exception):
-    def __init__(self, module_name):
-        super().__init__(f"Could not import provider module '{module_name}'")
+class FeastModuleImportError(Exception):
+    def __init__(self, module_name, module_type="provider"):
+        super().__init__(f"Could not import {module_type} module '{module_name}'")
 
 
-class FeastProviderClassImportError(Exception):
-    def __init__(self, module_name, class_name):
+class FeastClassImportError(Exception):
+    def __init__(self, module_name, class_name, class_type="provider"):
         super().__init__(
-            f"Could not import provider '{class_name}' from module '{module_name}'"
+            f"Could not import {class_type} '{class_name}' from module '{module_name}'"
         )
 
 
@@ -68,6 +68,20 @@ class FeastOfflineStoreUnsupportedDataSource(Exception):
     def __init__(self, offline_store_name: str, data_source_name: str):
         super().__init__(
             f"Offline Store '{offline_store_name}' does not support data source '{data_source_name}'"
+        )
+
+
+class FeastOnlineStoreInvalidName(Exception):
+    def __init__(self, online_store_class_name: str):
+        super().__init__(
+            f"Online Store Class '{online_store_class_name}' should end with the string `OnlineStore`.'"
+        )
+
+
+class FeastOnlineStoreConfigInvalidName(Exception):
+    def __init__(self, online_store_config_class_name: str):
+        super().__init__(
+            f"Online Store Config Class '{online_store_config_class_name}' should end with the string `OnlineStoreConfig`.'"
         )
 
 
--- a/sdk/python/feast/infra/online_stores/helpers.py
+++ b/sdk/python/feast/infra/online_stores/helpers.py
@@ -1,31 +1,41 @@
+import importlib
 import struct
+from typing import Any
 
 import mmh3
 
+from feast import errors
 from feast.infra.online_stores.online_store import OnlineStore
 from feast.protos.feast.storage.Redis_pb2 import RedisKeyV2 as RedisKeyProto
 from feast.protos.feast.types.EntityKey_pb2 import EntityKey as EntityKeyProto
-from feast.repo_config import OnlineStoreConfig, RedisOnlineStoreConfig
 
 
-def get_online_store_from_config(
-    online_store_config: OnlineStoreConfig,
-) -> OnlineStore:
+def get_online_store_from_config(online_store_config: Any,) -> OnlineStore:
     """Get the offline store from offline store config"""
 
-    if online_store_config.__repr_name__() == "SqliteOnlineStoreConfig":
-        from feast.infra.online_stores.sqlite import SqliteOnlineStore
+    module_name = online_store_config.__module__
+    qualified_name = type(online_store_config).__name__
+    store_class_name = qualified_name.replace("Config", "")
+    try:
+        module = importlib.import_module(module_name)
+    except Exception as e:
+        # The original exception can be anything - either module not found,
+        # or any other kind of error happening during the module import time.
+        # So we should include the original error as well in the stack trace.
+        raise errors.FeastModuleImportError(
+            module_name, module_type="OnlineStore"
+        ) from e
 
-        return SqliteOnlineStore()
-    elif online_store_config.__repr_name__() == "DatastoreOnlineStoreConfig":
-        from feast.infra.online_stores.datastore import DatastoreOnlineStore
-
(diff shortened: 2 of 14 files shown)


sdk/python/feast/infra/online_stores/sqlite.py
```diff
@@ -65,6 +77,9 @@ def online_write_batch(
         ],
         progress: Optional[Callable[[int], Any]],
     ) -> None:
+
+        print(f"online_write_batch: {config}")
```
woop, line 81: Is this print supposed to be here?

sdk/python/feast/repo_config.py
```diff
@@ -1,80 +1,39 @@
...
-    """ (optional) Amount of threads to use when writing batches of feature rows into Datastore"""
-
-    write_batch_size: Optional[PositiveInt] = 50
-    """ (optional) Amount of feature rows per batch being written into Datastore"""
-
-
-class RedisType(str, Enum):
-    redis = "redis"
-    redis_cluster = "redis_cluster"
+        extra = "allow"
```
woop, line 28: Where are we using `extras`? Is this to support dynamic building of the whole config object?

sdk/python/tests/test_repo_config.py
```diff
@@ -99,7 +101,7 @@ def test_bad_type():
             path: 100500
         """
         ),
-        expect_error="__root__ -> online_store -> path\n  str type expected",
+        expect_error="path\n  str type expected",
```
woop, line 104: Does this mean our error messages don't show the full path to the nested field that is misconfigured?

sdk/python/tests/test_repo_config.py
```diff
@@ -24,8 +24,10 @@ def _test_config(config_text, expect_error: Optional[str]):
             error = e
 
         if expect_error is not None:
+            print(f"Error: {error}")
```
woop, line 27: Are these supposed to be left in?

sdk/python/tests/test_feature_store.py
```diff
@@ -63,7 +64,7 @@ def feature_store_with_gcs_registry():
     bucket.blob("registry.db")
 
     return FeatureStore(
-        config=RepoConfig(
+        config=create_repo_config(
```
woop, line 67: Is the idea that users would also be calling `create_repo_config`?

achals: Yes - RepoConfig can't know upfront what the online_store type will be so it's wrapped by `create_repo_config` to keep the layers separate.

sdk/python/tests/test_repo_config.py
achals: Nope all for my debugging

sdk/python/tests/test_repo_config.py
achals: I *think* i fixed this, let me check

sdk/python/feast/repo_config.py
achals: Yes, exactly. If we `forbid`, then anything that comes in from the store-specific Config object counts as "unexpected".

sdk/python/feast/infra/online_stores/sqlite.py
achals: Nope

sdk/python/tests/test_repo_config.py
achals: Yeah the original test still works so undid this change, thanks for pointing out

sdk/python/feast/repo_config.py
```diff
@@ -123,16 +82,23 @@ class RepoConfig(FeastBaseModel):
     """
 
     provider: StrictStr
-    """ str: local or gcp or redis """
+    """ str: local or gcp """
 
-    online_store: OnlineStoreConfig = SqliteOnlineStoreConfig()
+    online_store: Any
```
woop, line 87: So we can't use a common base class here?

achals: Yeah - there's no good way of expressing "Subclass of T". We could theoretically have:
```
    online_store: FeastConfigBaseModel
```

But since `FeastConfigBaseModel` has `extra='forbid'` that would break validation since each subclass will have fields no in `FeastConfigBaseModel`. You can have `extra='allow'` but than actually makes the validation of the individual config classes too loose.

woop (approved):

tsotnet (changes requested):
It seems like the combination of `get_online_config_from_type` and `get_online_store_from_config` provides the same functionality as `get_prodiver` (defined in `sdk/python/feast/infra/provider.py`) for online stores. Can you make a more generalized methods that unifiy custom class loading potentially for all components (provider, online store, offline store, etc.)?

Something like
```
def get_component(config, component_type, default_component_map, init_args=None, init_kwargs=None):
    ...
```

Where `config` is the configuration object (e.g. RepoConfig or OnlineStoreConfig), `component_type` is a string containing either one of the default component types, or import paths, `default_component_map` is a map of built-in component short names to class paths (like `ONLINE_CONFIG_CLASS_FOR_TYPE`), and `init_*` parameters are for constructing the component.

For providers, you could call
```
provider = get_component(repo_config, repo_config.provider, PROVIDER_CLASS_FOR_TYPE, (repo_config, repo_path))
````
While for online stores you could call
```
online_store = get_component(repo_config.online_store, repo_config.online_store.type, ONLINE_CONFIG_CLASS_FOR_TYPE)
```

File: sdk/python/feast/errors.py [hunk 7]
```diff
@@ -42,15 +42,15 @@ def __init__(self, provider_name):
         super().__init__(f"Provider '{provider_name}' is not implemented")
 
 
-class FeastProviderModuleImportError(Exception):
-    def __init__(self, module_name):
-        super().__init__(f"Could not import provider module '{module_name}'")
+class FeastModuleImportError(Exception):
+    def __init__(self, module_name, module_type="provider"):
```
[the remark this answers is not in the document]
achals: done in #1657

sdk/python/feast/errors.py
```diff
@@ -42,15 +42,15 @@ def __init__(self, provider_name):
...
+        super().__init__(f"Could not import {module_type} module '{module_name}'")
 
 
-class FeastProviderClassImportError(Exception):
-    def __init__(self, module_name, class_name):
+class FeastClassImportError(Exception):
+    def __init__(self, module_name, class_name, class_type="provider"):
```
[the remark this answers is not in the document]
achals: done in #1657

sdk/python/feast/infra/online_stores/helpers.py
```diff
@@ -1,65 +1,41 @@
...
-    OnlineStoreConfig,
-    RedisOnlineStoreConfig,
-    SqliteOnlineStoreConfig,
-)
 
 
-def get_online_store_from_config(
-    online_store_config: OnlineStoreConfig,
-) -> OnlineStore:
+def get_online_store_from_config(online_store_config: Any,) -> OnlineStore:
```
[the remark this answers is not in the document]
achals: It has to be `Any` unfortunately - can't be `FeastConfigBaseModel` since pydantic doesn't understand or handle "subclass of type" as a type annotation.

sdk/python/feast/repo_config.py
```diff
@@ -246,6 +205,38 @@ def __repr__(self) -> str:
         )
 
 
+def get_online_config_from_type(online_store_type: str):
```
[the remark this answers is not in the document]
achals: I like leaving it here since it's more config-related. But happy to change if you feel strongly.

PR #1653, opened 2021-06-18
Give option to duplicate a registry as json

### LINKED ISSUE #1643: Decouple Feature Registry from specific storage solution

We would like to have the feature registry stored in other places like BQ.  In terms of what we're trying to achieve, we want the features and feature views to be in an easy to explore format. We also would like want to be able to access this data performantly in the online context for retrieval of [feature sets (i.e. feature services)](https://docs.google.com/document/d/1jC0RJbyYLilXTOrLVBeR22PYLK5fe2JmQK1mKdZ-eno/edit#). Am open to accomplishing these goals by other means. 

Create abstract interface for Feature Registry storing that various storage solutions can implement. 

Different writers that can read the data from GCS to store in other storage option. 

**Additional context**

codyjlin:
This PR gives the option to duplicate a registry as a json file (the default used for fetch and apply still being the protobuf one). As per https://github.com/feast-dev/feast/issues/1643#issuecomment-862785511, this will enable one to design a view on BQ of the registry.

Fixes #1643

```release-note
* Adds a new format of the registry field in the feature_store.yaml configuration
* Gives option to specify path of registry instead of particular filename
* An optional publish_json field (defaults to false) allows a user to duplicate the registry as JSON.
```

The `feature_store.yaml` can now look like this:
```
registry:
  path: this/is/my/folder
  publish_json: true
```
The original syntax can now support both file and directory path:
```
registry: <filepath OR dirpath (filepath inferred by dot in filename)>
```


sdk/python/feast/registry.py
```diff
@@ -43,22 +45,33 @@ class Registry:
...
+        cache_ttl: timedelta,
+    ):
         """
         Create the Registry object.
 
         Args:
             repo_path: Path to the base of the Feast repository
             cache_ttl: The amount of time that cached registry state stays valid
-            registry_path: filepath or GCS URI that is the location of the object store registry,
+            registry_path: filepath or GCS URI that is the enclosing folder of the object store registry,
```
mavysavydav, line 61: capitalize 'f' in filepath

sdk/python/feast/registry.py
```diff
@@ -43,22 +45,33 @@ class Registry:
...
             or where it will be created if it does not exist yet.
         """
+        # registry_path = f"{registry_path}/registry.pb"
```
mavysavydav, line 64: forgot to remove

sdk/python/feast/templates/local/feature_store.yaml
```diff
@@ -1,5 +1,6 @@
 project: my_project
-registry: data/registry.db
+registry:
+    path: data
```
mavysavydav, line 3: maybe this should be "data/" to make it clear it's a dir. Unsure tho

sdk/python/tests/cli_utils.py
```diff
@@ -48,9 +48,10 @@ def local_repo(self, example_repo_py: str, offline_store: str):
             random.choice(string.ascii_lowercase + string.digits) for _ in range(10)
         )
 
-        with tempfile.TemporaryDirectory() as repo_dir_name, tempfile.TemporaryDirectory() as data_dir_name:
+        with tempfile.TemporaryDirectory() as repo_dir_name, tempfile.TemporaryDirectory() as registry_dir_name, tempfile.TemporaryDirectory() as data_dir_name:
```
mavysavydav, line 51: why make a separate registry folder?

mavysavydav:
Aside from those comments, LGTM

sdk/python/feast/repo_config.py
```diff
@@ -113,7 +116,7 @@ class RegistryConfig(FeastBaseModel):
 class RepoConfig(FeastBaseModel):
     """ Repo config. Typically loaded from `feature_store.yaml` """
 
-    registry: Union[StrictStr, RegistryConfig] = "data/registry.db"
+    registry: RegistryConfig = RegistryConfig()
```
woop, line 119: Unfortunately we can't have a backwards compatible change here. If we introduce this change then all Feast users will have broken deployments. We will need to test and support both the old and new approach (file path, dir path), at least for a few versions to allow gradual upgrades.

sdk/python/feast/templates/local/feature_store.yaml
codyjlin: Good point, I think it works either way in the code, but for the template and the docs it could be better to have trailing `/`

sdk/python/tests/cli_utils.py
codyjlin: One of the tests (I think it was `test_online_retrieval.py:test_online`) was messing with the previous registry name to test caching, but now that it's the enclosing folder, I decided to just keep the registry folder separate from the data folder (which holds the online store) to avoid the name change affecting the online store too.

sdk/python/feast/repo_config.py
codyjlin: Ah that makes a lot of sense, let me try to support both versions of the registry config using `Union`. Since I already updated the tests to model the new version, I'll just add back a simple test testing the old version. Does that sound good?

codyjlin: I used a `pytest.parametrize` [here in test_cli_local.py](https://github.com/feast-dev/feast/pull/1653/files#diff-ffdc23101f179930d4578bf59d2e2afe6ed94a73fa58368e243c1bee5b252ddaR17) to test both the old registry config and new. There shouldn't be any break changes now, but let me know if you think there should be more robust testing for the old version

sdk/python/feast/repo_config.py
```diff
@@ -43,9 +50,14 @@ class Config:
 class RegistryConfig(FeastBaseModel):
     """ Metadata Store Configuration. Configuration that relates to reading from and writing to the Feast registry."""
 
-    path: StrictStr
+    path: StrictStr = "data/"
     """ str: Path to metadata store. Can be a local path, or remote object storage path, e.g. a GCS URI """
 
+    publish_json: StrictBool = False
+
+    is_filepath: StrictBool = False
```
woop, line 58: Is there a reason to ask users to specify this configuration flag instead of handling the migration for them? We know if we are being provided a file location vs a directory, right?

sdk/python/feast/repo_operations.py
```diff
@@ -116,6 +116,9 @@ def apply_total(repo_config: RepoConfig, repo_path: Path):
 
     os.chdir(repo_path)
     registry_config = repo_config.get_registry_config()
+
+    print(type(registry_config))
```
woop, line 120: what is this for?

sdk/python/feast/repo_config.py
```diff
@@ -56,8 +68,10 @@ class RegistryConfig(FeastBaseModel):
 class RepoConfig(FeastBaseModel):
     """ Repo config. Typically loaded from `feature_store.yaml` """
 
-    registry: Union[StrictStr, RegistryConfig] = "data/registry.db"
-    """ str: Path to metadata store. Can be a local path, or remote object storage path, e.g. a GCS URI """
+    registry: Union[StrictStr, RegistryConfig] = RegistryConfig()
```
woop, line 71: Is there a reason to stop using `"data/registry.db"` as the default configuration path and create a whole object?

sdk/python/feast/templates/gcp/feature_store.yaml
```diff
@@ -1,3 +1,4 @@
 project: my_project
-registry: data/registry.db
+registry:
```
woop, line 2: It's not clear why we are making changes here. The default string based option is meant to be a concise way to specify a storage location.

sdk/python/tests/test_cli_local.py
```diff
@@ -1,18 +1,22 @@
...
 
 from feast.feature_store import FeatureStore
+from feast.protos.feast.core.Registry_pb2 import Registry as RegistryProto
 from tests.cli_utils import CliRunner
 from tests.online_read_write_test import basic_rw_test
 
 
 @pytest.mark.integration
-def test_workflow() -> None:
+@pytest.mark.parametrize("old_registry_config", [False, True])
```
woop, line 18: Instead of having a `True`, `False`, why not have the config as the parameter?

sdk/python/feast/repo_operations.py
codyjlin: ah oops, forgot to remove this when cleaning up. Thanks for catching

sdk/python/tests/test_cli_local.py
codyjlin: I didn't know if it was possible since it requires `data_path`, but if you think it's cleaner, I can use the `"...{}...".format(str)` notation and have it be it be pytest param. Also, I'm assuming you mean just the `registry` section of the config and not the whole config.

sdk/python/feast/templates/gcp/feature_store.yaml
codyjlin: That's a good point, I was thinking we were going to change the default to specify the metadata folder path instead of the filepath location. But maybe it could be a good design to allow `registry: <path>` accept both filepath or dirpath and just continue to support both syntaxes (single line for just specifying path and multiline for nested options).

sdk/python/feast/repo_config.py
codyjlin: Good point, let me change it to validate file vs directory

sdk/python/feast/repo_config.py
codyjlin: I was assuming we wanted to change the default to use the nested config for the additional options, but since we can just accommodate both moving forward, I'll change the default to the simple string

sdk/python/feast/repo_config.py
```diff
@@ -91,6 +102,10 @@ def get_registry_config(self):
         if isinstance(self.registry, str):
             return RegistryConfig(path=self.registry)
         else:
+            if self.registry.path and self.registry.path[-1] != "/":
+                raise Exception(
+                    f"Path {self.registry.path} is not a valid directory path; please end with '/'."
+                )
```
codyjlin, line 108: Note: going to enforce this because otherwise can't tell if it's a filepath or dirpath (can't distinguish between the old simple string 2 lines above `RegistryConfig(path=self.registry)`) - should be dirpath if using the larger config with option to publish json.

sdk/python/feast/registry.py
```diff
@@ -436,12 +448,24 @@ def update_registry_proto(
...
+        self, repo_path: Path, registry_path_string: str, publish_json: bool,
+    ):
         registry_path = Path(registry_path_string)
-        if registry_path.is_absolute():
-            self._filepath = registry_path
+
+        if not registry_path.is_absolute():
+            registry_path = repo_path.joinpath(registry_path)
+
+        if registry_path_string[-1] == "/":
```
codyjlin, line 459: Since file nor dir hasn't been created yet, there's no way to check if the `registry_path_string` passed in is intended to be dir or file. Here I infer that if it ends in `/` then it should be dir.

sdk/python/feast/feature_store.py
```diff
@@ -75,6 +75,7 @@ def __init__(
         self._registry = Registry(
             registry_path=registry_config.path,
             repo_path=self.repo_path,
+            publish_json=registry_config.publish_json,
```
woop, line 78: Sorry for giving you so much back and forth, but I actually think `export_json` might be more descriptive than publish. What do you think? For the config option, argument, as well as the file name (`registry_export.json`)?

I just want to make sure its clear that this is a read only file and not the real registry, and I think export is more descriptive than publish.

sdk/python/feast/registry.py
woop: What if we looked at the extension? we can assume if its `blahblah.something` then its a file, and everything else is a folder, like `blahblah` and `blahblah/` would be folders.

sdk/python/feast/registry.py
```diff
@@ -436,12 +448,24 @@ def update_registry_proto(
...
+            parent = registry_path
+            filename = "registry.pb"
         else:
-            self._filepath = repo_path.joinpath(registry_path)
+            parent = registry_path.parent
+            filename = registry_path.name
+
+        self._dirpath = parent
+        self._filepath = self._dirpath / filename
+        self.publish_json = publish_json
```
woop, line 468: is there a reason `publish_json` is not a private field?

sdk/python/feast/registry.py
```diff
@@ -523,10 +558,12 @@ def update_registry_proto(
             registry_proto.registry_schema_version = REGISTRY_SCHEMA_VERSION
         if updater:
             registry_proto = updater(registry_proto)
-        self._write_registry(registry_proto)
+        self._write_registry_blob(registry_proto)
+        if self.publish_json:
```
woop, line 562: There seems to be an inconsistency here. In the one case `_write_registry_json` is inside `_write_registry_blob` and in the other case outside.

sdk/python/feast/registry.py
codyjlin: I had thought about that too - the tradeoff is that if their folder names have a single dot (not sure if this is a large enough edge case), then there would be no way for us to infer their intention. In addition, do you know if all filepaths/filenames require an extension? Happy to make this change if you've thought about these things too!

sdk/python/feast/feature_store.py
codyjlin: No worries, I think `export_json` and `registry_export.json` are much better names too. Thanks for the idea!

sdk/python/feast/registry.py
woop: Filenames don't require an extension, but our default registry path is `data/registry.db`, so the majority of users will have an extension. I think it's the least bad approach tbh. I don't think there are folks creating folders with an extension. @tsotnet @achals any thoughts?

achals: I agree with @woop  - in the case that the registry file/dir has to be created, the presence or absence of an extension should be good enough for us to determine whether this should be a file or a dir.

sdk/python/feast/registry.py
codyjlin: oh wow, thanks for catching this - totally meant to put it in `update_registry_proto` for both cases. Inconsistency bugs me a ton too so not sure how I managed to do this.

I'll be putting both inside `update_registry_proto`

sdk/python/feast/registry.py
codyjlin: Sounds good, I'm cool with that.

sdk/python/feast/registry.py
```diff
@@ -577,10 +578,13 @@ def _write_registry_blob(self, registry_proto: RegistryProto):
...
         gs_bucket = self.gcs_client.get_bucket(self._bucket)
-        json_path = urlparse(f"{self._uri}/registry.json").path.lstrip("/")
+        json_path = urlparse(f"{self._uri_prefix}/registry_export.json").path.lstrip(
+            "/"
+        )
         blob = gs_bucket.blob(json_path)
         file_obj = TemporaryFile()
-        file_obj.write(bytes(MessageToJson(registry_proto), "utf-8"))
+        registry_dict = MessageToDict(registry_proto)
+        file_obj.write(bytes(json.dumps(registry_dict), "utf-8"))
```
codyjlin, line 587: This results in a [JSONL](https://jsonlines.org/examples/) format. Change was made due to [the finding](https://stackoverflow.com/questions/51390797/json-formatting-error-when-loading-into-google-big-query) that pretty JSON is not supported by BQ. Note that a local `_write_registry_json` would still output the normal JSON format with mult-line objects.

sdk/python/tests/test_feature_store.py
```diff
@@ -41,7 +41,7 @@ def feature_store_with_local_registry():
     fd, online_store_path = mkstemp()
     return FeatureStore(
         config=RepoConfig(
-            registry=registry_path,
+            registry=f"{registry_path}.pb",
```
codyjlin, line 44: https://github.com/feast-dev/feast/pull/1653#discussion_r659024618 We are the folks creating files without an extension 😄 
image

mavysavydav: lololol

codyjlin:
Closing per discussion in Slack [here](https://tectonfeast.slack.com/archives/C01NTDB88QK/p1624821196094700?thread_ts=1624641071.082700&cid=C01NTDB88QK).

Issue #1655, opened 2021-06-18
Ability to choose the BQ project in Feast


MattDelac:
**Is your feature request related to a problem? Please describe.**
After merging the [current commit](https://github.com/feast-dev/feast/commit/a3d7c8661039eb3abc1beb9cb48718a6ad24e3ef), we need to create datasets on the ServiceAccount that is currently using Feast. This cause permissions issues on our side

The issue it that the service account that using Feast might have permissions to write into another BQ account that is not its own.

**Describe the solution you'd like**
Thus we should be able to pass the `project` in which to write tables/datasets in the `BigQueryOfflineStoreConfig()`.
This could be an optional argument and get infer if not passed

**Describe alternatives you've considered**
Give permissions to service accounts that don't need it

**Additional context**

woop:
> The issue it that the service account that using Feast might have permissions to write into another BQ account that is not its own.

I'm a bit confused. How is having permissions a problem. Or are you saying that **do** want to write to a different project?

MattDelac:
> Or are you saying that do want to write to a different project?

Yes ! Having the ability to write to a project that is different from the current Service Account

woop:
I see, feel free to add the `project` field to the config. Sounds good to me.

PR #1656, opened 2021-06-18
Allow BigQuery project to be configured

MattDelac:
Add the possibility to specify a GCP project for the temporary datasets created.

For example, with the config file and a ServiceAccount coming from the project `ml-adhoc`
```yaml
[...]
offline_store:
    type: bigquery
    project: ml-production
```

```
>>> from google.cloud import bigquery
>>>from feast.feature_store import FeatureStore
>>> store = FeatureStore()

>>> store.config.offline_store.project
'ml-production'

>>> bigquery.Client().project
'ml-adhoc'
```

Which means that the ServiceAccount `ml-adhoc` will write into the BQ project `ml-production` (_names are just for examples_)

Fixes https://github.com/feast-dev/feast/issues/1655

NONE
```release-note
Possibility to specify a different GCP project than the ServiceAccount to save the BigQuery tables
```

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -94,8 +94,14 @@ class BigQueryOfflineStore(OfflineStore):
         expected_join_keys = _get_join_keys(project, feature_views, registry)
 
         assert isinstance(config.offline_store, BigQueryOfflineStoreConfig)
+        dataset_project = config.offline_store.project_id or client.project
+
         table = _upload_entity_df_into_bigquery(
-            client, config.project, config.offline_store.dataset, entity_df
+            client=client,
+            project=config.project,
+            dataset_name=config.offline_store.dataset,
+            dataset_project=dataset_project,
+            entity_df=entity_df,
         )
 
         entity_df_event_timestamp_col = _infer_event_timestamp_from_bigquery_query(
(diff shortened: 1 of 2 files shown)


sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -225,7 +231,8 @@ def _block_until_done():
 
         today = date.today().strftime("%Y%m%d")
         rand_id = str(uuid.uuid4())[:7]
-        path = f"{self.client.project}.{self.config.offline_store.dataset}.historical_{today}_{rand_id}"
+        dataset_project = self.config.offline_store.project or self.client.project
```
woop, line 234: Under what circumstances would `self.client.project` be different from `self.config.offline_store.project`

sdk/python/feast/repo_config.py
```diff
@@ -93,6 +93,9 @@ class BigQueryOfflineStoreConfig(FeastBaseModel):
     dataset: StrictStr = "feast"
     """ (optional) BigQuery Dataset name for temporary tables """
 
+    project: Optional[StrictStr] = None
+    """ (optional) BigQuery project name for temporary tables """
```
woop, line 97: But wouldn't this also be for permanent tables? `historical_` tables could be permanent right? I think we can just drop `for temporary tables` and make it `GCP project name used for the BigQuery offline store` `BigQuery dataset name used for the BigQuery offline store`

What do you think?

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -275,18 +282,21 @@ def _get_table_id_for_new_entity(
...
     project: str,
     dataset_name: str,
+    dataset_project: str,
     entity_df: Union[pandas.DataFrame, str],
 ) -> Table:
     """Uploads a Pandas entity dataframe into a BigQuery table and returns the resulting table"""
 
-    table_id = _get_table_id_for_new_entity(client, project, dataset_name)
+    table_id = _get_table_id_for_new_entity(
+        client, project, dataset_name, dataset_project
```
woop, line 298: would `gcp_project` be more accurate?

sdk/python/feast/repo_config.py
MattDelac: Yes you're right, those are permanent at the end. The user should choose how they want to structure their projects (with the retention policy they want etc.)

woop: Ok, great.

MattDelac: Done I rename the new field for `project_id` and updated the description

woop (approved):

PR #1657, opened 2021-06-21
Refactor OfflineStoreConfig classes into their owning modules

achals:

Follow up from #1649. This refactors the OfflineStoreConfig in the same exact way so that things are loaded dynamically instead of hardcoding in the `repo_config.py` module.
 
Part 2 for #1605

```release-note
Allow OfflineStore implementations to be loaded dynamically.
```

### DIFF
--- a/sdk/python/feast/errors.py
+++ b/sdk/python/feast/errors.py
@@ -85,10 +85,10 @@ class FeastOnlineStoreInvalidName(Exception):
         )
 
 
-class FeastOnlineStoreConfigInvalidName(Exception):
-    def __init__(self, online_store_config_class_name: str):
+class FeastStoreConfigInvalidName(Exception):
+    def __init__(self, online_store_config_class_name: str, store_type="Online"):
         super().__init__(
-            f"Online Store Config Class '{online_store_config_class_name}' should end with the string `OnlineStoreConfig`.'"
+            f"Online Store Config Class '{online_store_config_class_name}' should end with the string `{store_type}StoreConfig`.'"
         )
 
 
--- a/sdk/python/feast/errors.py
+++ b/sdk/python/feast/errors.py
@@ -85,11 +85,11 @@ class FeastOnlineStoreInvalidName(Exception):
         )
 
 
-class FeastStoreConfigInvalidName(Exception):
-    def __init__(self, store_config_class_name: str, store_type: str):
+class FeastClassInvalidName(Exception):
+    def __init__(self, class_name: str, class_type: str):
         super().__init__(
-            f"Config Class '{store_config_class_name}' "
-            f"should end with the string `{store_type}Config`.'"
+            f"Config Class '{class_name}' "
+            f"should end with the string `{class_type}`.'"
         )
 
 
--- a/sdk/python/feast/errors.py
+++ b/sdk/python/feast/errors.py
@@ -50,7 +50,7 @@ class FeastProviderNotImplementedError(Exception):
 
 
 class FeastModuleImportError(Exception):
-    def __init__(self, module_name, module_type="provider"):
+    def __init__(self, module_name: str, module_type: str):
         super().__init__(f"Could not import {module_type} module '{module_name}'")
 
 
--- a/sdk/python/feast/errors.py
+++ b/sdk/python/feast/errors.py
@@ -88,7 +88,8 @@ class FeastOnlineStoreInvalidName(Exception):
 class FeastStoreConfigInvalidName(Exception):
     def __init__(self, online_store_config_class_name: str, store_type="Online"):
         super().__init__(
-            f"Online Store Config Class '{online_store_config_class_name}' should end with the string `{store_type}StoreConfig`.'"
+            f"Online Store Config Class '{online_store_config_class_name}' "
+            f"should end with the string `{store_type}Config`.'"
         )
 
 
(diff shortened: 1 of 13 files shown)


sdk/python/feast/errors.py
```diff
@@ -85,10 +85,11 @@ def __init__(self, online_store_class_name: str):
         )
 
 
-class FeastOnlineStoreConfigInvalidName(Exception):
-    def __init__(self, online_store_config_class_name: str):
+class FeastStoreConfigInvalidName(Exception):
+    def __init__(self, online_store_config_class_name: str, store_type="Online"):
```
woop, line 89: Is there a reason to set a default here?

achals: Nope, going to clean this up

woop: Also, should `online_store_config_class_name` be online store agnostic?

sdk/python/feast/errors.py
```diff
@@ -85,10 +85,11 @@ def __init__(self, online_store_class_name: str):
...
         super().__init__(
-            f"Online Store Config Class '{online_store_config_class_name}' should end with the string `OnlineStoreConfig`.'"
+            f"Online Store Config Class '{online_store_config_class_name}' "
+            f"should end with the string `{store_type}Config`.'"
```
woop, line 92: It's not clear what `OnlineConfig` here means. Shouldn't this be the name of the Config class? (assuming the default value is used)

sdk/python/feast/errors.py
achals: done

sdk/python/feast/errors.py
achals: Sorry, I don't know what you're referring to here - the diff may have changed since the last time you looked at it

sdk/python/feast/infra/online_stores/sqlite.py
```diff
@@ -34,7 +34,9 @@
 class SqliteOnlineStoreConfig(FeastConfigBaseModel):
     """ Online store config for local (SQLite-based) store """
 
-    type: Literal["sqlite"] = "sqlite"
+    type: Literal[
+        "sqlite", "feast.infra.online_stores.sqlite.SqliteOnlineStore"
```
tsotnet, line 38: why is this change necessary?

sdk/python/feast/infra/provider.py
```diff
@@ -163,7 +163,7 @@ def get_provider(config: RepoConfig, repo_path: Path) -> Provider:
             # The original exception can be anything - either module not found,
             # or any other kind of error happening during the module import time.
             # So we should include the original error as well in the stack trace.
-            raise errors.FeastModuleImportError(module_name) from e
+            raise errors.FeastModuleImportError(module_name, "provider") from e
```
tsotnet, line 166: nit: capitalize provider

sdk/python/feast/repo_config.py
```diff
@@ -235,11 +207,33 @@ def get_online_config_from_type(online_store_type: str):
...
+
+def get_offline_config_from_type(offline_store_type: str):
+    if offline_store_type in OFFLINE_STORE_CLASS_FOR_TYPE:
+        offline_store_type = OFFLINE_STORE_CLASS_FOR_TYPE[offline_store_type]
+    else:
+        assert offline_store_type.endswith("OfflineStore")
+    module_name, offline_store_class_type = offline_store_type.rsplit(".", 1)
+    config_class_name = f"{offline_store_class_type}Config"
+
+    return get_config_class_from_type(module_name, config_class_name, "OfflineStore")
```
tsotnet, line 234: These 2 can also be deduplicated. For example, you could construct
```
STORE_CLASS_FOR_TYPE = {
    "OnlineStore": ONLINE_STORE_CLASS_FOR_TYPE,
    "OfflineStore": OFFLINE_STORE_CLASS_FOR_TYPE,
}

def get_config_class_from_type(store_type: str, store_class_type: str):
    if store_class_type in STORE_CLASS_FOR_TYPE[store_type]:
        store_class_type = STORE_CLASS_FOR_TYPE[store_type][store_class_type]
    else:
        assert store_class_type.endswith(store_type)
    module_name, store_class_type = store_class_type.rsplit(".", 1)
    config_class_name = f"{store_class_type}Config"

    # merge existing get_config_class_from_type function (line 188) here instead of splitting
    return get_config_class_from_type(module_name, config_class_name, store_type)
```

Ideally, we shouldn't have Providers handled in a different way, but it's a little bit different (e.g. no Config classes there), so I'll leave it up to you if you think we can deduplicate that in a reasonable way.

tsotnet (changes requested):

sdk/python/feast/infra/provider.py
achals: done

sdk/python/feast/repo_config.py
achals: I pulled in the provider to use a shared code path for loading.

I'm tempted to not introduce `STORE_CLASS_FOR_TYPE` cause the additional indirection/method doesn't seem that bad, and seems more legible. WDYT?

achals: Take a look at the latest version, it's more cleaned up

sdk/python/feast/infra/online_stores/sqlite.py
tsotnet: Gotcha, makes sense.

tsotnet (approved):

PR #1661, opened 2021-06-22
Provide the user with more options for setting the to_bigquery config

### LINKED ISSUE #1654: Provide ability to set more options when outputting joined data to bq table.

We need to be able to set expiries for the tables, project.dataset.tableName, and also option for create vs replace.

Ideally the user can optionally pass in a config object like QueryJobConfig. QueryJobConfig has many options such as `destination` which allows one to set the destination of table and name, but it doesn't seem like it can handle create/replace (it has create_disposition which offers some limited control) and expiry.

N/A. Willem and a few of us briefly discussed this and he played around with the idea of passing in and out sql queries. So one could modify the query and pass it back in.

codyjlin:

This PR gives the user more freedom to set configuration options for the `to_bigquery` call (on a BigQueryRetrievalJob following a `get_historical_features` call). For example, the user can specify a destination `project.dataset.tableName`, an expiration timestamp for the table, and also an option for replace/overwrite vs fail if already exists.

Fixes #1654

```release-note
* User should supply to_bigquery() with a bigquery.QueryJobConfig.
* User can also call to_sql() to get/modify the SQL directly (i.e. to set expiration_timestamp)
```

**References**
- [google.cloud.bigquery.job.QueryJobConfig](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.job.QueryJobConfig.html
)
- This PR builds on the design in https://github.com/feast-dev/feast/pull/1634 when to_bigquery() was first added

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -227,7 +227,17 @@ class BigQueryRetrievalJob(RetrievalJob):
     def to_sql(self):
         return self.query
 
-    def to_bigquery(self, job_config=None) -> Optional[str]:
+    def to_bigquery(self, job_config: bigquery.QueryJobConfig = None) -> Optional[str]:
+        """
+        Uploads the results of the BigQueryRetrievalJob directly to BigQuery
+
+        Args:
+            job_config: A bigquery.QueryJobConfig to specify options like destination table, dry run, etc.
+
+        Returns:
+            Returns either the destination table name, none if job_config.dry_run, or raises an exception if job fails
+        """
+
         @retry(wait=wait_fixed(10), stop=stop_after_delay(1800), reraise=True)
         def _block_until_done():
             return self.client.get_job(bq_job.job_id).state in ["PENDING", "RUNNING"]
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -1,6 +1,7 @@
 import time
+import uuid
 from dataclasses import asdict, dataclass
-from datetime import datetime, timedelta
+from datetime import date, datetime, timedelta
 from typing import List, Optional, Set, Union
 
 import pandas
@@ -231,18 +232,27 @@ class BigQueryRetrievalJob(RetrievalJob):
         def _block_until_done():
             return self.client.get_job(bq_job.job_id).state in ["PENDING", "RUNNING"]
 
+        if not job_config:
+            today = date.today().strftime("%Y%m%d")
+            rand_id = str(uuid.uuid4())[:7]
+            dataset_project = (
+                self.config.offline_store.project_id or self.client.project
+            )
+            path = f"{dataset_project}.{self.config.offline_store.dataset}.historical_{today}_{rand_id}"
+            job_config = bigquery.QueryJobConfig(destination=path)
+
         bq_job = self.client.query(self.query, job_config=job_config)
+
+        _block_until_done()
+
+        if bq_job.exception():
+            raise bq_job.exception()
 
         if job_config.dry_run:
             print(
                 "This query will process {} bytes.".format(bq_job.total_bytes_processed)
             )
             return None
-
-        _block_until_done()
-
-        if bq_job.exception():
-            raise bq_job.exception()
 
         print(f"Done writing to '{job_config.destination}'.")
         return str(job_config.destination)
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -1,7 +1,6 @@
 import time
-import uuid
 from dataclasses import asdict, dataclass
-from datetime import date, datetime, timedelta
+from datetime import datetime, timedelta
 from typing import List, Optional, Set, Union
 
 import pandas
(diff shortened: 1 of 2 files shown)


sdk/python/tests/test_historical_retrieval.py
```diff
@@ -454,13 +459,26 @@ def test_historical_features_from_bigquery_sources(
         actual_bq_temp_table = bigquery.Client().get_table(bq_temp_table_path)
         assert actual_bq_temp_table.table_id == bq_temp_table_path.split(".")[-1]
 
+        # Config to overwrite an existing table should succeed
+        job_config = bigquery.QueryJobConfig(
+            destination=path, write_disposition="WRITE_TRUNCATE"
```
codyjlin, line 464: Note: [write_disposition](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.job.QueryJobConfig.html#google.cloud.bigquery.job.QueryJobConfig.write_disposition) has the string options listed in the docs [here](https://cloud.google.com/bigquery/docs/reference/rest/v2/Job#JobConfigurationQuery.FIELDS.write_disposition).

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -224,19 +223,17 @@ def to_df(self):
         df = self.client.query(self.query).to_dataframe(create_bqstorage_client=True)
         return df
 
-    def to_bigquery(self, dry_run=False) -> Optional[str]:
+    def to_sql(self):
+        return self.query
```
codyjlin, line 227: Note: there's no option for expiration_timestamp in the [QueryJobConfig](https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.job.QueryJobConfig.html), so we should also give more granular control for the user to modify and run the SQL directly (like proposal 4 in https://github.com/feast-dev/feast/issues/1654#issuecomment-864218588). An example of this is in my fork [here](https://github.com/codyjlin/feast/blob/offline-to-bq/sdk/python/feast/infra/offline_stores/bigquery.py#L247-L256).

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -247,8 +244,8 @@ def _block_until_done():
         if bq_job.exception():
             raise bq_job.exception()
 
-        print(f"Done writing to '{path}'.")
-        return path
+        print(f"Done writing to '{job_config.destination}'.")
```
mavysavydav, line 247: job_config might be None

sdk/python/tests/test_historical_retrieval.py
```diff
@@ -439,12 +440,16 @@ def test_historical_features_from_bigquery_sources(
         )
 
         # Just a dry run, should not create table
-        bq_dry_run = job_from_sql.to_bigquery(dry_run=True)
+        job_config = bigquery.QueryJobConfig(dry_run=True)
```
mavysavydav, line 443: let's not write tests for these for the time being. It won't be worth the work and protection when tests are about to undergo refactoring. Also this is within a test that'll run 6 times due to its parametrizations

sdk/python/tests/test_historical_retrieval.py
mavysavydav: Same as comment above, and we don't need to test for individual settings in QueryJobConfig. We can assume they work

sdk/python/tests/test_historical_retrieval.py
mavysavydav: I also think these kind of small changes don't need test cases generally

sdk/python/feast/infra/offline_stores/bigquery.py
woop: I think we are missing docstrings here (and for the other methods). Can you please add it to your method? Also, should there be a return type?

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -224,31 +224,38 @@ def to_df(self):
         df = self.client.query(self.query).to_dataframe(create_bqstorage_client=True)
         return df
 
-    def to_bigquery(self, dry_run=False) -> Optional[str]:
+    def to_sql(self):
+        return self.query
+
+    def to_bigquery(self, job_config=None) -> Optional[str]:
```
woop, line 230: Can you please add a type here?

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -224,31 +224,51 @@ def to_df(self):
         df = self.client.query(self.query).to_dataframe(create_bqstorage_client=True)
         return df
 
-    def to_bigquery(self, dry_run=False) -> Optional[str]:
+    def to_sql(self) -> str:
+        """
+        Returns the SQL query that can be modified and run by the user to create the BQ table.
```
woop, line 229:
```suggestion
        Returns the SQL query that will be executed in BigQuery to build the historical feature table.
```

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -224,31 +224,51 @@ def to_df(self):
...
+        """
+        return self.query
+
+    def to_bigquery(self, job_config: bigquery.QueryJobConfig = None) -> Optional[str]:
+        """
+        Uploads the results of the BigQueryRetrievalJob directly to BigQuery.
```
woop, line 235:
```suggestion
        Triggers the execution of a historical feature retrieval query and exports the results to a BigQuery table.
```

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -224,31 +224,51 @@ def to_df(self):
...
+
+        Args:
+            job_config: A bigquery.QueryJobConfig to specify options like destination table, dry run, etc.
```
woop, line 238:
```suggestion
            job_config: An optional bigquery.QueryJobConfig to specify options like destination table, dry run, etc.
```

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -224,31 +224,51 @@ def to_df(self):
...
+
+        Returns:
+            Returns either the destination table name, none if job_config.dry_run, or raises an exception if job fails
```
woop, line 241:
```suggestion
            Returns the destination table name or returns None if job_config.dry_run is True.
```

woop (approved):

PR #1662, opened 2021-06-23
Add methods to offline/online stores to specify supported upstream sources

achals:

We need a way for Online and Offline stores to be explicit about what upstream sources are supported. This was added in #1552 but removed in #1649. This PR adds it back in.

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/repo_config.py
+++ b/sdk/python/feast/repo_config.py
@@ -3,11 +3,11 @@ from pathlib import Path
 from typing import Any
 
 import yaml
-from feast.data_source import DataSource
 from pydantic import BaseModel, StrictInt, StrictStr, ValidationError, root_validator
 from pydantic.error_wrappers import ErrorWrapper
 from pydantic.typing import Dict, Optional, Union
 
+from feast.data_source import DataSource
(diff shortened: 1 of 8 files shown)


sdk/python/feast/repo_config.py
```diff
@@ -32,13 +34,29 @@ class Config:
...
+    """ Feast Pydantic Configuration Class """
+
+    class Config:
+        arbitrary_types_allowed = True
+        extra = "forbid"
+
+    @abstractmethod
+    def supports_offline_store(self, offline_store: Any) -> bool:
+        ...
+
```
tsotnet, line 59: Why does online store need to define which offline stores to support? The online store and offline store do no interact with each other directly.

I'd say let providers define which online store & offline store they support.

achals: > Why does online store need to define which offline stores to support? The online store and offline store do no interact with each other directly.
> 
> I'd say let providers define which online store & offline store they support.

Spoke offline - I'm going to scrap this PR because there's no explicitly unsupported combination at the moment. We should wait until the need for this kind of API emerges, possibly after we have an AWS provider.

PR #1663, opened 2021-06-23
Add to_table() to RetrievalJob object

MattDelac:
We should add a notion of "OfflineJob" so that it will be easier to add new methods as well as accessible useful information.

For example, when you build a custom Provider, you might want to access the query coming from `pull_latest_from_table_or_query()` without executing it (as it will be executed differently)

Fixes #

```release-note
pull_latest_from_table_or_query() now returns an RetrievalJob object
Add method to_table() to RetrievalJob object
```

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -15,11 +15,7 @@ from feast import errors
 from feast.data_source import BigQuerySource, DataSource
 from feast.errors import FeastProviderLoginError
 from feast.feature_view import FeatureView
-from feast.infra.offline_stores.offline_store import (
-    OfflineJob,
-    OfflineStore,
-    RetrievalJob,
-)
+from feast.infra.offline_stores.offline_store import OfflineStore, RetrievalJob
 from feast.infra.provider import (
     DEFAULT_ENTITY_DF_EVENT_TIMESTAMP_COL,
     _get_requested_feature_views_to_features_dict,
@@ -63,7 +59,7 @@ class BigQueryOfflineStore(OfflineStore):
         created_timestamp_column: Optional[str],
         start_date: datetime,
         end_date: datetime,
-    ) -> OfflineJob:
+    ) -> RetrievalJob:
         assert isinstance(data_source, BigQuerySource)
         from_expression = data_source.get_table_query_string()
 
@@ -78,6 +74,7 @@ class BigQueryOfflineStore(OfflineStore):
         timestamp_desc_string = " DESC, ".join(timestamps) + " DESC"
         field_string = ", ".join(join_key_columns + feature_name_columns + timestamps)
 
+        client = _get_bigquery_client(project=config.offline_store.project_id)
         query = f"""
             SELECT {field_string}
             FROM (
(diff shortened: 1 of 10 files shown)


sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -228,6 +225,15 @@ def _infer_event_timestamp_from_dataframe(entity_df: pandas.DataFrame) -> str:
             )
 
 
+class BigQueryOfflineJob(OfflineJob):
+    def __init__(self, query: str, config: RepoConfig):
+        self.query = query
+        self.client = _get_bigquery_client(project=config.offline_store.project_id)
+
+    def to_table(self) -> pyarrow.Table:
+        return self.client.query(self.query).to_arrow()
```
woop, line 234: Is it intentional for `BigQueryOfflineJob.to_arrow()` to return a pyarrow table? I would have expected `to_table()` to create a BQ table. Would `to_arrow_table()` or `to_arrow()` be clearer?

woop:
I really like this idea @MattDelac!

The one jarring thing is that we now have two types of jobs that are completely isolated. `RetrievalJob` and `OfflineJob`. Is there any reason for these to be different?

I like the idea of having a single `OfflineJob` that is configured differently through methods like `materialize` and `get_historical_features()` and context objects, and then you automatically get all the `to_sql` and other base functionality. The wrapping command (`materialize`, `get_historical_features`) can then work with the OfflineJob differently, like in one case it can use `.to_table()` or whatever if it needs to load the latest values into the online store.

What do you think about that?

MattDelac:
> The one jarring thing is that we now have two types of jobs that are completely isolated. `RetrievalJob` and `OfflineJob`. Is there any reason for these to be different?

Hmmm good question 🤔 

I can see if those two objects could/should be the same

MattDelac:
@woop I changed this PR to add a new method to the `RetrievalJob` object instead of creating a new object

woop:
> @woop I changed this PR to add a new method to the `RetrievalJob` object instead of creating a new object

Are you ready for a review?

MattDelac:
> > @woop I changed this PR to add a new method to the `RetrievalJob` object instead of creating a new object
> 
> Are you ready for a review?

Let me check this linter issue since I rebased on master

MattDelac:
> > @woop I changed this PR to add a new method to the `RetrievalJob` object instead of creating a new object
> 
> Are you ready for a review?

It sounds good now 👍

achals (commented):
looks good to me. Can we exercise this code path in the existing integration tests?

MattDelac:
> Can we exercise this code path in the existing integration tests?

Not sure I understand what you mean here ...

achals:
> > Can we exercise this code path in the existing integration tests?
> 
> Not sure I understand what you mean here ...

Sorry, i meant can we call `to_table` during one of the integration tests?

woop:
Also lgtm

achals (approved):
/lgtm

woop:
@achals @MattDelac I know we merged this, but I just realized that this comment was unaddressed https://github.com/feast-dev/feast/pull/1663#discussion_r659032133

How do you feel about making the method `to_arrow()`?

MattDelac:
> @achals @MattDelac I know we merged this, but I just realized that this comment was unaddressed [#1663 (comment)](https://github.com/feast-dev/feast/pull/1663#discussion_r659032133)
> 
> How do you feel about making the method `to_arrow()`?
> 
> Example in the wild https://googleapis.dev/python/bigquerystorage/latest/bigquery_storage_v1/library.html?highlight=arrow#google.cloud.bigquery_storage_v1.reader.ReadRowsIterable.to_arrow

Ho sorry it's my fault. Will push a quick fix where I rename it `to_arrow()` 👍

PR #1664, opened 2021-06-23
Add streaming sources to the FeatureView API

### LINKED ISSUE #1564: StreamingFeatureView API for 0.10
Feast 0.10 does not support ingestion to the online store from streams like Kafka. This functionality was available in 0.9, and it's on the roadmap for Feast to support this with the FeatureView class before the 0.11 release.

Our proposed API is just to maintain the Feast existing Feast 0.9 API for defining ingestion from streams. This looks like:

```
driver_ft = FeatureView(
    name="driver_trips",
    entities=["driver_id"],
    features=[
      Feature("average_daily_rides", ValueType.FLOAT),
      Feature("rating", ValueType.FLOAT)
    ],
    max_age=Duration(seconds=3600),
    batch_source=BigQuerySource(
        table_ref="gcp_project:bq_dataset.bq_table",
        event_timestamp_column="datetime",
        created_timestamp_column="timestamp",
        field_mapping={
          "rating": "driver_rating"
        }
    ),
    stream_source=KafkaSource(
	  ...
    )
)
```

We also considered a decorator-based API:

```
kafka_source = KafkaSource(...) 

driver_ft = FeatureTable(
    name="driver_trips",
    entities=["driver_id"],
    features=[
      Feature("average_daily_rides", ValueType.FLOAT),
      Feature("rating", ValueType.FLOAT)
    ],
    max_age=Duration(seconds=3600),
		inputs={"orders": Input(source=kafka_source, window="7d")},
)
```

The main benefit of this decorator-based API is that it allows for the expression of more complex logic. We are designing the API for OnDemandFeatures in parallel (this discussion on OnDemandFeatures will happen on a public issue to be created soon.) A decorator-based approach could be nice for OnDemandFeatures. 

However, this decorator approach doesn't look like it adds value to streaming feature declarations and might be overkill, so the proposal is to stick with the Feast 0.9 API.

achals:
This diff only updates the API. It is currently up to the providers to actually use this information to spin up resources to consume events from the stream sources.

This PR introduces a field for streaming sources. It also introduces a field for batch sources, which is identical to the input sources.

Fixes #1564 

```release-note
Add streaming sources to the FeatureView API
```


woop:
Are we missing the `.proto` changes?

achals:
> Are we missing the `.proto` changes?

D'oh. Yes. Fixing

sdk/python/feast/feature_view.py
```diff
@@ -199,7 +225,13 @@ def from_proto(cls, feature_view_proto: FeatureViewProto):
                 and feature_view_proto.spec.ttl.nanos == 0
                 else feature_view_proto.spec.ttl
             ),
-            input=DataSource.from_proto(feature_view_proto.spec.input),
+            input=_input,
+            batch_source=_input,
+            stream_source=(
+                feature_view_proto.spec.stream_source
+                if feature_view_proto.spec.stream_source is None
```
woop, line 232: Isn't this always none?

sdk/python/feast/feature_view.py
```diff
@@ -83,7 +99,9 @@ def __init__(
             self.ttl = ttl
 
         self.online = online
-        self.input = input
+        self.input = _input
+        self.batch_source = _input
+        self.stream_source = stream_source
```
woop, line 104: Should we validate that a batch source exists for all stream sources?

protos/feast/core/FeatureView.proto
```diff
@@ -59,7 +59,8 @@ message FeatureViewSpec {
     google.protobuf.Duration ttl = 6;
 
     // Batch/Offline DataSource where this view can retrieve offline feature data.
-    DataSource input = 7;
+    DataSource batch_source = 7;
+    DataSource stream_source = 9;
```
woop, line 63: I think we need a comment here.

sdk/python/feast/feature_view.py
```diff
@@ -60,16 +69,23 @@ def __init__(
         name: str,
         entities: List[str],
         ttl: Optional[Union[Duration, timedelta]],
-        input: Union[BigQuerySource, FileSource],
+        input: Optional[Union[BigQuerySource, FileSource]],
+        batch_source: Optional[Union[BigQuerySource, FileSource]] = None,
```
tsotnet, line 73: What's the difference between input and batch_source?

achals: I want to deprecate `input` - `batch_source` is supposed to supercede `input`. The reason I'm adding a new field is that because we need to differentialte stream and batch sources. This also matches the API for FeatureTables.

protos/feast/core/FeatureView.proto
achals: yup

sdk/python/feast/feature_view.py
achals: yes

sdk/python/feast/feature_view.py
achals: oops

sdk/python/feast/feature_view.py
achals: This is already happening in line 79-80

sdk/python/feast/feature_view.py
```diff
@@ -50,7 +50,8 @@ class FeatureView:
     ttl: Optional[timedelta]
     online: bool
     input: DataSource
-
+    batch_source: Optional[DataSource] = None
```
woop, line 53: One more question. Is there a reason `batch_source` is Optional?

achals: Backwards compatibilty - users are currently using `input`. We should make `batch_source` mandatory  when we're going to make a version bump. wdyt?

woop (approved):

PR #1665, opened 2021-06-23
Infer min and max timestamps from entity_df to limit data read from BQ source

Mwad22:
Signed-off-by: Mwad22 <51929507+Mwad22@users.noreply.github.com>

Infers the minimum and maximum timestamps from the provided `entity_df` if possible (if entity_df is provided as a Pandas dataframe). Right now, too much data is being read since the range of time for the feature data doesn't account for the min and max timestamps for the base data (`entity_df`).

For instance, if the max timestamp on order ids is 5/1/2020, we want to avoid wasting time looking at data in the range 5/2/2020 to present day– we would not join feature data from this range. 

In short, this change will allow us to read less data from BigQuery sources.

No issue assigned to this at the moment, but was left as TODO item in `bigquery.py` [here](https://github.com/feast-dev/feast/blob/2013c04308314f0055eea469a3c92898d861365e/sdk/python/feast/infra/offline_stores/bigquery.py#L134)

```release-note
NONE

```


sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -496,6 +507,9 @@ def _get_bigquery_client():
             {{ feature }} as {{ featureview.name }}__{{ feature }}{% if loop.last %}{% else %}, {% endif %}
         {% endfor %}
     FROM {{ featureview.table_subquery }}
+    WHERE {{ featureview.event_timestamp_column }} <= '{{ max_timestamp }}' {% if featureview.ttl == 0 %}{% else %}
```
codyjlin, line 510: For better readability, could move `{% if featureview.ttl == 0 %}{% else %}` to its own line (to be able to match with the `{% endif %}` below).
```
    ...
    WHERE {{ featureview.event_timestamp_column }} <= '{{ max_timestamp }}'
    {% if featureview.ttl == 0 %}{% else %}
    AND {{ featureview.event_timestamp_column }} >= Timestamp_sub(TIMESTAMP '{{ min_timestamp }}', interval {{ featureview.ttl }} second)
    {% endif %}
    ...
```

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -131,12 +131,23 @@ def get_historical_features(
...
+        if isinstance(entity_df, pandas.DataFrame):
+            min_timestamp = datetime.fromisoformat(
+                min(entity_df[[DEFAULT_ENTITY_DF_EVENT_TIMESTAMP_COL]])
+            )
+            max_timestamp = datetime.fromisoformat(
+                max(entity_df[DEFAULT_ENTITY_DF_EVENT_TIMESTAMP_COL])
+            )
+        else:
+            min_timestamp = datetime.now() - timedelta(days=365)
+            max_timestamp = datetime.now() + timedelta(days=1)
```
codyjlin, line 144: In the case that `entity_df` is a `str` (a query that gets run in BQ), I think `_upload_entity_df_into_bigquery` a few lines [above](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/offline_stores/bigquery.py#L114-L120) will create a tem table that expires in 30 minutes. (see definition [here](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/offline_stores/bigquery.py#L335-L338))

I'm not sure how clean it'd be to try to extract mins and maxes from that table vs just having some kind of defaults for min and max timestamp in this case, but could be something to look into. Hopefully others have thoughts too.

woop:
Is anything preventing us from inferring the timestamps purely in SQL?

sdk/python/feast/infra/offline_stores/bigquery.py
Mwad22: Thanks Cody, in the latest change I infer and use the timestamps purely in SQL query (lines 464-469) so this is will also address the issue you point out!

Mwad22:
> Thanks for this @Mwadd22!
> 
> A bit of a nitpick, but is anything preventing us from inferring and using the timestamps purely in SQL? I suspect it would be significantly faster than scanning a dataframe.

Thanks for pointing that out @woop. I just committed some changes that infer and use the timestamps purely in SQL as requested (and removed the `min/max_timestamps` arguments as they are no longer needed).

mavysavydav:
LGTM

woop:
Looks much nicer. @MattDelac are you ok with this change?

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -468,6 +461,12 @@ def _get_bigquery_client():
     FROM {{ left_table_query_string }}
 ),
 
+timestamp_bounds AS (
+    SELECT
+        MAX({{entity_df_event_timestamp_col}}) AS max,
+        MIN({{entity_df_event_timestamp_col}}) AS min
```
tsotnet, line 467: can you call these columns something different from `min` and `max` to avoid confusion between columns and keywords?

MattDelac (commented):
**TLDR; I am fine with the current PR as the overhead of the MIN/MAX computations should be minimal (especially compare to the overall function). That being said, the more long term approach would be to fetch the min/max locally to avoid adding complexity to the SQL template**

Reasoning:
it sounds like `MAX()` and `MIN()` operations are going to be computed multiple times in BigQuery.

For example, I just looked at the "Execution details" of the following query in BQ and the MIN() MAX() operations are computed at least twice
```sql
WITH left_table AS (
    SELECT "user 1" AS col_agg, 1 AS col_value, TIMESTAMP '2021-01-01' AS col_timestamp
    UNION ALL
    SELECT "user 1" AS col_agg, 2 AS col_value, TIMESTAMP '2021-01-02' AS col_timestamp
    UNION ALL
    SELECT "user 2" AS col_agg, 3 AS col_value, TIMESTAMP '2021-01-03' AS col_timestamp
    UNION ALL
    SELECT "user 2" AS col_agg, 4 AS col_value, TIMESTAMP '2021-01-04' AS col_timestamp
    UNION ALL
    SELECT "user 2" AS col_agg, 5 AS col_value, TIMESTAMP '2021-01-05' AS col_timestamp
),

subquery_table AS (
    SELECT "user 1" AS col_agg, 6 AS col_value, TIMESTAMP '2021-01-01' AS col_timestamp
    UNION ALL
    SELECT "user 1" AS col_agg, 7 AS col_value, TIMESTAMP '2021-01-02' AS col_timestamp
    UNION ALL
    SELECT "user 2" AS col_agg, 8 AS col_value, TIMESTAMP '2021-01-03' AS col_timestamp
    UNION ALL
    SELECT "user 2" AS col_agg, 9 AS col_value, TIMESTAMP '2021-01-04' AS col_timestamp
    UNION ALL
    SELECT "user 2" AS col_agg, 10 AS col_value, TIMESTAMP '2021-01-05' AS col_timestamp
),

timestamp_bounds AS (
    SELECT
        MAX(col_timestamp) AS max_boundary,
        MIN(col_timestamp) AS min_boundary
    FROM left_table
),

subquery AS (
    SELECT *
    FROM left_table 
    WHERE left_table.col_timestamp <= (SELECT max_boundary FROM timestamp_bounds)
    AND left_table.col_timestamp >= TIMESTAMP_SUB((SELECT min_boundary FROM timestamp_bounds), INTERVAL 3000 second)
),

compute_something AS (
    SELECT
        col_agg,
        SUM(col_value) AS col_value_tot
    FROM subquery 
    GROUP BY 1
)

SELECT *
FROM subquery
LEFT JOIN compute_something USING (col_agg)
```

So it sounds like it would be more optimized and long term to just fetch the boundaries locally and avoid adding complexity to the current BQ template. Especially since it might be computed again and again because of the "for loop" happening on each FeatureView

```python
query = """
SELECT
  MIN(timestamp_col) AS min_timetamp,
  MAX(timestamp_col) AS max_timetamp
FROM left_table_query_string
"""
boundary_df = bigquery.query(query=query).to_pandas()
min_timestamp = boundary_df.loc[0, "min_timetamp"]
max_timestamp = boundary_df.loc[0, "max_timetamp"]
```

_Note: We could potentially get the min/max info of the pandas dataframe without performing another SQL query_

woop:
> Note: We could potentially get the min/max info of the pandas dataframe without performing another SQL query

They were originally scanning the local Pandas Dataframe for the min/max locally, but I asked them to move the logic into SQL for performance reasons. 

> So it sounds like it would be more optimized and long term to just fetch the boundaries locally and avoid adding complexity to the current BQ template.

In that case should we execute two queries? The first determines the min/max and the second is the main template.

woop:
> So it sounds like it would be more optimized and long term to just fetch the boundaries locally and avoid adding complexity to the current BQ template. Especially since it might be computed again and again because of the "for loop" happening on each FeatureView

I agree that we can further optimize the query by determining the min/max in another query.

> ```python
> query = """
> SELECT
>   MIN(timestamp_col) AS min_timetamp,
>   MAX(timestamp_col) AS max_timetamp
> FROM left_table_query_string
> """
> boundary_df = bigquery.query(query=query).to_pandas()
> min_timestamp = boundary_df.loc[0, "min_timetamp"]
> max_timestamp = boundary_df.loc[0, "max_timetamp"]
> ```

lgtm

> Note: We could potentially get the min/max info of the pandas dataframe without performing another SQL query

This PR originally scanned the local dataframe, but I'm not sure if that will hold up for large datasets. I'd want to move all compute into the DWH.

MattDelac:
> This PR originally scanned the local dataframe, but I'm not sure if that will hold up for large datasets. I'd want to move ~all~ as much compute into the DWH as possible.

Ho yes, it would only work when you pass a pandas dataframe as a left feature_view (instead of a query). So it does not solve all use cases.

achals (commented):
lgtm pending the column naming change

sdk/python/feast/infra/offline_stores/bigquery.py
Mwad22: Done, columns renamed and also moved the timestamp inference into a separate query to avoid recomputing and avoid adding complexity to the main retrieval query.

Mwad22:
Thank you all for your detailed reviews, definitely appreciated! If there are no other changes to be requested, I think this change should be good to go.

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -496,6 +523,10 @@ def _get_bigquery_client():
             {{ feature }} as {{ featureview.name }}__{{ feature }}{% if loop.last %}{% else %}, {% endif %}
         {% endfor %}
     FROM {{ featureview.table_subquery }}
+    WHERE {{ featureview.event_timestamp_column }} <= '{{max_timestamp}}'
+    {% if featureview.ttl == 0 %}{% else %}
+    AND {{ featureview.event_timestamp_column }} >= Timestamp_sub('{{min_timestamp}}', interval {{ featureview.ttl }} second)
```
woop, line 528: I don't feel super strongly, but subtracting the TTL outside of the query could make this query simpler than doing it within. Happy to ship as is though.

woop:
Overall it looks good to me. Just wanted to get your opinion on one last nitpick :)

Otherwise I am happy to ship it.

MattDelac (approved):
An integration test for `_get_entity_df_timestamp_bounds()` would be lovely 

Otherwise it LGTM

sdk/python/feast/infra/offline_stores/bigquery.py
Mwad22: So, I feel that might overcomplicate the logic. The way I see it, each feature view has its own TTL which would mean we would need to pass a min_timestamp per feature_view if we wanted to subtract the TTL outside the query. However, if you have a cleaner way of doing it in mind, I don't mind making the modification!

Mwad22:
@woop @MattDelac I've added my thoughts [here](https://github.com/feast-dev/feast/pull/1665#discussion_r664086198) on why I think subtracting the TTL in the query might be preferable. I am also open to adding an integration test for `_get_entity_df_timestamp_bounds()`, but am currently unsure where the best place to add this test would be? Should I create a new file `test_bigquery.py` or add it to one (or both) of the `test_historical_retrieval.py` or `test_online_retrieval.py`?Thanks!

woop:
Ok, from my side you can leave the query as is. 

I think `_get_entity_df_timestamp_bounds` should be a new test. You can use `test_historical_retrieval.py` and add a standalone function for BigQuery. We will be refactoring the tests over time, so for now it's just to ensure we don't forget about it.

sdk/python/feast/infra/offline_stores/bigquery.py
MattDelac: Ho I see what you mean. The current approach sounds fine to me to be honest

sdk/python/tests/test_historical_retrieval.py
```diff
@@ -557,3 +573,22 @@ def test_historical_features_from_bigquery_sources(
...
+@pytest.mark.integration
+def test_timestamp_bound_inference_from_entity_df_using_bigquery():
+    start_date = datetime.now().replace(microsecond=0, second=0, minute=0)
+    (_, _, _, entity_df, start_date) = generate_entities(
+        start_date, infer_event_timestamp_col=True
+    )
+
+    client = bigquery.Client()
+    table_id = "foo.table_id"
+    table = upload_entity_df_into_bigquery(client, entity_df, table_id)
```
woop, line 582: Any reason we can't reuse https://github.com/feast-dev/feast/blob/e8a361fc0a8994d846aeb3d8ae8e74db79003bc9/sdk/python/tests/test_historical_retrieval.py#L226:5?

Mwad22: Actually, I think a little refactoring and we can use it just fine. Just pushed that change!

Mwad22: @woop Unless there are other changes you would like me to make to the test, I think this should be good to ship? Again, happy to make last minute edits if you see any are necessary!

Mwad22: Update: Changed typo in assert statement (use `==` now instead of `is`) after seeing the integration test had failed below, now all integration tests are passing.

mavysavydav:
Looking great

woop (approved):

PR #1669, opened 2021-06-27
Add RedshiftDataSource

tsotnet:

**What this PR does / why we need it**: Adds RedshiftDataSource as a first step for supporting Redshift offline store.

Fixes #

```release-note
Add RedshiftDataSource
```


sdk/python/feast/data_source.py
```diff
@@ -875,3 +909,223 @@ def to_proto(self) -> DataSourceProto:
...
+    """
+    DataSource Redshift options used to source features from Redshift query
+    """
+
+    def __init__(self, table: Optional[str], query: Optional[str]):
+        self._table = table
+        self._query = query
+
+    @property
+    def query(self):
```
woop, line 924: Do we need a getter/setter if we aren't adding custom logic?

sdk/python/feast/data_source.py
```diff
@@ -875,3 +909,223 @@ def to_proto(self) -> DataSourceProto:
...
+                if len(table["ColumnList"]) == 0:
+                    raise DataSourceNotFoundException(self.table)
+
+                columns = table["ColumnList"]
+            else:
+                statement = client.execute_statement(
+                    ClusterIdentifier=config.offline_store.cluster_id,
+                    Database=config.offline_store.database,
+                    DbUser=config.offline_store.user,
+                    Sql=f"SELECT * FROM ({self.query}) LIMIT 1",
```
woop, line 1103: @tsotnet Out of scope for this PR, but what is your opinion on the utility of `query` vs `table`. We lose a lot of control by allowing `query`. I'm just wondering if it's worth the trouble long term, vs asking users to create their own views or adding more functionality to table references.

woop:
Looks good to me overall

sdk/python/feast/data_source.py
tsotnet: Yeah... We use the exact same approach for most other fields in DataSource/FileSource/BigQuerySource though. I think it would be good to clean this up separately. wdyt?

sdk/python/feast/data_source.py
tsotnet: I don't like the added complexity unless it's necessary. Would be good to understand how query is used. One common pattern I've seen is filtering. I don't think query should be used for complex things like grouping or aggregation, especially since we're using it like this ^ (limit 1) to get schema. If only filtering satisfies 99% of use cases, we could just add a filter DSL at most.

sdk/python/feast/data_source.py
woop: sgtm

sdk/python/feast/data_source.py
woop: > If only filtering satisfies 99% of use cases, we could just add a filter DSL at most.

Exactly.

sdk/python/feast/data_source.py
```diff
@@ -875,3 +909,223 @@ def to_proto(self) -> DataSourceProto:
...
+        event_timestamp_column: Optional[str] = "",
+        table: Optional[str] = None,
+        created_timestamp_column: Optional[str] = "",
+        field_mapping: Optional[Dict[str, str]] = None,
+        date_partition_column: Optional[str] = "",
+        query: Optional[str] = None,
+    ):
+        self._redshift_options = RedshiftOptions(table=table, query=query)
+
+        super().__init__(
```
achals, line 996: nit, this should usually be the first call in a subclass's `__init__` method

sdk/python/feast/data_source.py
achals: I personally also find supporting queries tricky because the behaviour is highly dependent on the underlying DWH. E.g. in some cases we can run the query lazily to evaluate schemas (like if we're using spark) but in some cases it's always executed eagerly which may be expensive.

tsotnet: @achals Yeah agree, I added this as an item to discuss this week.

achals (approved):
/lgtm

PR #1673, opened 2021-06-29
Allow strings for online/offline store instead of dicts

achals:

This allows the offline_store or online_store to be specified as a top level string instead of a dict.

Fixes #

```release-note
Allow strings for online/offline store instead of dicts
```


woop:
do we need to change any docs?

achals:
> do we need to change any docs?

Yep but I'm updating the how to in gitbook directly

tsotnet (approved):
/lgtm

PR #1675, opened 2021-06-30
BQ exception should be raised first before we check the timedout

MattDelac:

If the query failed before starting, its state will be `FAILURE`. So we should raise the BQ exceptions first otherwise we will end up seeing "BigQueryJobCancelled" errors even though it's not accurate

Fixes #

```release-note
Fix exception order when BigQuery failures occur
```

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -286,12 +286,12 @@ def block_until_done(client, bq_job):
     job_id = bq_job.job_id
     _wait_until_done(job_id=job_id)
 
+    if bq_job.exception():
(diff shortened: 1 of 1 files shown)


woop (approved):

woop:
@MattDelac, this is technically a bug. We should classify it as such.

/kind bug

MattDelac:
> @MattDelac, this is technically a bug, which requires a release note. I've added it.

Hooo ok 👌 

Where did you add the release note ? The description of this PR is unchanged

woop:
> > @MattDelac, this is technically a bug, which requires a release note. I've added it.
> 
> Hooo ok 
> 
> Where did you add the release note ? The description of this PR is unchanged

You're too fast.

PR #1676, opened 2021-06-30
Add FeatureService proto definition

### LINKED ISSUE #1572: Introduce Feature Services
You can view & comment [the following doc](https://docs.google.com/document/d/1jC0RJbyYLilXTOrLVBeR22PYLK5fe2JmQK1mKdZ-eno/edit#)

Feature Tables and Views allow for storage level grouping of features based on how they are produced. However, there is no concept of a retrieval-level grouping of features in Feast. Without a retrieval-level grouping of features that maps to models

- There is no way for users to track which features were used in the training of a model, or which features are being used for the serving of a specific model. 
- Retrieving features during training requires a complete list of features to be provided, and this complete list of features needs to be persisted and loaded for online serving. This persistence and loading is a low-level implementation that may be error prone (if not all features are references, or provided in the wrong order).
- There is no way to ensure that consumers won't have breaking changes when feature views and tables change since there is no awareness of features that are in use.

[See doc #design](https://github.com/feast-dev/feast/issues/1572)

[See doc #alternatives](https://github.com/feast-dev/feast/issues/1572)

**Additional context**

achals:

Fixes #1572 

```release-note
Introduce FeatureService to group Feature references
```


protos/feast/core/FeatureReference.proto
```diff
@@ -0,0 +1,18 @@
...
+import "feast/core/Feature.proto";
+
+
+// A reference to features in a feature view
+message FeatureReference {
+  // The feature view name
+  string feature_view_name = 1;
+
+  // The features of the feature view that are a part of the feature reference.
+  repeated FeatureSpecV2 feature_columns = 2;
```
woop, line 17: I am a bit confused. A feature ref is a reference to a single feature. Does this imply a reference to a feature view? like a feature view ref?

Also, should we dedupe it with https://github.com/feast-dev/feast/blob/bcf925500d3aaf8711aef7e2f00cf8fafd31b2ee/protos/feast/serving/ServingService.proto#L52?

protos/feast/core/FeatureReference.proto
```diff
@@ -0,0 +1,18 @@
...
+
+option go_package = "github.com/feast-dev/feast/sdk/go/protos/feast/core";
+option java_outer_classname = "FeatureReferenceProto";
+option java_package = "feast.proto.core";
+
+import "feast/core/Feature.proto";
+
+
+// A reference to features in a feature view
+message FeatureReference {
```
woop, line 12: Our existing definition of a feature reference is a reference to a single feature, not a collection. What about using `FeatureViewRef`? Or alternatively, is there a downside to using a full FeatureView object?

sdk/python/feast/feature_service.py
```diff
@@ -0,0 +1,82 @@
...
+class FeatureService:
+    name: str
+    features: List[FeatureReference]
+    created_timestamp: Optional[datetime] = None
+    last_updated_timestamp: Optional[datetime] = None
+
+    def __init__(
+        self,
+        name: str,
+        features: List[Union[FeatureTable, FeatureView, FeatureReference]],
```
woop, line 25: In which cases would users provide references and in which cases tables/views?

achals: Per the RFC, we want to support both of these cases:

```
from feast.feature_service import FeatureService, RawFeature
from my_repo.feature_views import shop_raw, customer_sales

    name="my_super_model_v1",
    features=[
        shop_raw,
        customer_sales[['average_order_value', 'max_order_value']],
    ],
)
```

protos/feast/core/FeatureReference.proto
achals: I'm okay using FeatureViewRef - I don't want to use a full FeatureView object because I think we should be able to support only using a subset of features from a feature view, so we need a layer of indirection. Does that sound reasonable?

protos/feast/core/FeatureReference.proto
achals: TIL about this message, but this is somewhat different from what we want in the FeatureViewReference. I think the structure of the new message is required, unless you disagree.

protos/feast/core/FeatureReference.proto
woop: sgtm

sdk/python/feast/feature_service.py
woop: Isn't it more complicated to support different types though? What if we just had references and not views/tables?

achals: They are supported only in the `__init__` method - we convert everything  to a reference, and that's what is stored in the FeatreService proto. I think that provides a good level of abstractions for users.

protos/feast/core/FeatureService.proto
```diff
@@ -0,0 +1,46 @@
...
+message FeatureServiceSpec {
+  // Name of the Feature Service. Must be unique. Not updated.
+  string name = 1;
+
+  // Name of Feast project that this Feature Service belongs to.
+  string project = 2;
+
+  // List names of entities to associate with the Features defined in this
+  // Feature Service. Not updatable.
+  repeated string entities = 3;
```
woop, line 28: Is there a reason why entities are defined here? Aren't they already within the feature views?

achals: Good point, I don't think this is needed here.

woop (changes requested):
If we fix `entities` then I think we're good.

woop (approved):

woop:
Btw @achals, if your PR has "Fixes #[0-9]+", then it will actually close the corresponding issue when the PR lands.

PR #1679, opened 2021-07-01
Remove default list from the FeatureView constructor

achals:

Mutable default arguments to methods are a well-known way to shoot yourself in the foot in python: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

In this particular case, if we define multiple FeatureViews that use feature inferencing, we'll end up having all of them refer to the *same* list of features, which would have been inferred for the first feature view we processed in `repo_operations.py`. 

As a result, the list of inferred features for all the *other* FeatureView would be the same as the first set, and totally wrong.

Fixes #

```release-note
Fix bug with mutable default argument in FeatureView constructor.
```

### DIFF
--- a/sdk/python/feast/feature_view.py
+++ b/sdk/python/feast/feature_view.py
@@ -65,14 +65,16 @@ class FeatureView:
         input: DataSource,
         batch_source: Optional[DataSource] = None,
         stream_source: Optional[DataSource] = None,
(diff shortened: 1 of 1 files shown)


tsotnet (approved):

woop:
Good catch!

PR #1680, opened 2021-07-02
Implement materialization for RedshiftOfflineStore & RedshiftRetrievalJob

tsotnet:

**What this PR does / why we need it**: This PR makes it possible to materialize data from Redshift to online stores. Also added half of test_offline_online_store_consistency.py for Redshift (for now we're only asserting online store after materialization from Redshift and don't check offline store by calling get_historical_features, since that functionality is not yet implemented). This required implementing unload/copy logic through S3. RedshiftRetrievalJob is also fully implemented (to_df, to_arrow, to_s3, to_redshift).

Lastly, needed to add `iam_role` to the redshift config. This role is necessary for Redshift to access S3. It'll need to be setup by the user (both the role creation, and association to Redshift).

Fixes #

```release-note
Adds materialization logic for RedshiftOfflineStore
```


sdk/python/feast/infra/offline_stores/redshift.py
```diff
@@ -46,7 +52,43 @@ def pull_latest_from_table_or_query(
...
+        assert isinstance(config.offline_store, RedshiftOfflineStoreConfig)
+
+        from_expression = data_source.get_table_query_string()
+
+        partition_by_join_key_string = ", ".join(join_key_columns)
+        if partition_by_join_key_string != "":
+            partition_by_join_key_string = (
+                "PARTITION BY " + partition_by_join_key_string
+            )
+        timestamps = [event_timestamp_column]
```
woop, line 65: perhaps
```suggestion
        timestamp_columns = [event_timestamp_column]
```

sdk/python/feast/infra/offline_stores/redshift.py
```diff
@@ -33,6 +36,9 @@ class RedshiftOfflineStoreConfig(FeastConfigBaseModel):
     s3_path: StrictStr
```
woop, line 36: perhaps
```suggestion
    s3_staging_location: StrictStr
```

sdk/python/feast/infra/offline_stores/redshift.py
```diff
@@ -58,3 +100,61 @@ def get_historical_features(
         project: str,
     ) -> RetrievalJob:
         pass
+
+
+class RedshiftRetrievalJob(RetrievalJob):
+    def __init__(self, query: str, redshift_client, s3_resource, config: RepoConfig):
```
woop, line 106: Can we please add type annotations and maybe a docstring to `__init__`? It's not clear to me what `s3_resource` is.

sdk/python/feast/infra/offline_stores/redshift.py
```diff
@@ -58,3 +100,61 @@ def get_historical_features(
...
+        self.query = query
+        self._redshift_client = redshift_client
+        self._s3_resource = s3_resource
+        self._config = config
+        self._s3_path = (
+            self._config.offline_store.s3_path + "/unload/" + str(uuid.uuid4())
+        )
+
+    def to_df(self) -> pd.DataFrame:
```
woop, line 115: We should definitely add docstrings to all user facing methods.

sdk/python/feast/infra/utils/aws_utils.py
```diff
@@ -0,0 +1,286 @@
...
+
+def get_bucket_and_key(s3_path: str) -> Tuple[str, str]:
+    """
+    Get the S3 bucket and key given the full path.
+
+    For example get_bucket_and_key("s3://foo/bar/test.file") returns ("foo", "bar/test.file")
+
+    If the s3_path doesn't start with "s3://", it throws ValueError.
+    """
+    assert s3_path.startswith("s3://")
```
woop, line 46: Can we please raise an exception instead of using `assert`? I am not sure if it's a good practice to use `assert` outside of tests since it can be disabled when optimizations are turned on.

sdk/python/feast/infra/utils/aws_utils.py
```diff
@@ -0,0 +1,286 @@
...
+    We use exponential backoff for checking the query state until it's not running. The backoff starts with
+    0.1 seconds and doubles exponentially until reaching 30 seconds, at which point the backoff is fixed.
+
+    :param redshift_data_client: Redshift Data API Service client
+    :param statement: The redshift statement to wait for (result of execute_redshift_statement)
+    """
+    desc = redshift_data_client.describe_statement(Id=statement["Id"])
+    if desc["Status"] in ("SUBMITTED", "STARTED", "PICKED"):
+        raise RedshiftStatementNotFinishedError  # Retry
+    if desc["Status"] != "FINISHED":
```
woop, line 97: How are we handling `ABORTED` and `FAILED` states?

sdk/python/feast/infra/utils/aws_utils.py
```diff
@@ -0,0 +1,286 @@
...
+    wait_for_redshift_statement(redshift_data_client, statement)
+    return statement["Id"]
+
+
+def get_redshift_statement_result(redshift_data_client, statement_id: str) -> dict:
+    """ Get the Redshift statement result """
+    return redshift_data_client.get_statement_result(Id=statement_id)
+
+
+def copy_df_to_redshift(
```
woop, line 128: I think it's a bit confusing to call this method `copy`. Should we use `upload_` to be consistent with BigQuery?

sdk/python/feast/infra/utils/aws_utils.py
```diff
@@ -0,0 +1,286 @@
...
+        raise RedshiftStatementNotFinishedError  # Retry
+    if desc["Status"] != "FINISHED":
+        raise RedshiftQueryError(desc)  # Don't retry. Raise exception.
+
+
+def execute_redshift_statement(
+    redshift_data_client, cluster_id: str, database: str, user: str, query: str
+) -> str:
+    """
+    Execute Redshift statement synchronously. Waits for the query to finish.
```
woop, line 105: We are using Google style docstrings throughout the Feast codebase btw.

sdk/python/feast/infra/utils/aws_utils.py
```diff
@@ -0,0 +1,286 @@
...
+    column_query_list = ", ".join(
+        [
+            f"{column_name} {column_type}"
+            for column_name, column_type in zip(column_names, column_types)
+        ]
+    )
+
+    # Create the temporary table with the desired schema
+    create_query = f"CREATE TABLE {table_name}({column_query_list})"
+    execute_redshift_statement(
```
woop, line 178: Would it make sense to create this table only after writing the parquet file to s3? Otherwise we're creating empty tables if failures occur directly after.

sdk/python/feast/infra/utils/aws_utils.py
```diff
@@ -0,0 +1,286 @@
...
+def delete_s3_directory(s3_resource, bucket: str, key: str):
+    """ Delete S3 directory recursively """
+    bucket_obj = s3_resource.Bucket(bucket)
+    if key != "" and not key.endswith("/"):
+        key = key + "/"
+    for obj in bucket_obj.objects.filter(Prefix=key):
+        obj.delete()
+
+
+def unload_redshift_query_to_s3(
```
woop, line 221: Nitpick, but this function seems like it's doing two separate things

1. Running a query
2. Exporting to S3

Is there a reason for it to be a single function?  I guess it just reads a bit weird. I'm not used to "unloading" a query.

sdk/python/feast/infra/offline_stores/redshift.py
tsotnet: to_df and to_arrow are already documented in the parent class (RetrievalJob). I didn't want to duplicate the docstrings.

sdk/python/feast/infra/utils/aws_utils.py
tsotnet: They will raise RedshiftQueryError exception as documented in the docstring above.

sdk/python/feast/infra/offline_stores/redshift.py
tsotnet: we can't add type annotations to boto3 client & resource objects, because of how boto3 generates these objects. I have type annotations for the rest of the objects.

sdk/python/feast/infra/utils/aws_utils.py
tsotnet: I didn't realize that! I'll change this, although we have couple of other places where we use assert too.

sdk/python/feast/infra/utils/aws_utils.py
tsotnet: Hmm not sure what you mean by this. Running the query by itself unloads the data to S3. I don't do anything besides running the query on my side. Can you elaborate?

sdk/python/feast/infra/utils/aws_utils.py
woop: I know we have type assertions for Mypy. Do we have others as well?

sdk/python/feast/infra/offline_stores/redshift.py
woop: I don't understand. So there is no class `s3_resource` associated class or type?

sdk/python/feast/infra/utils/aws_utils.py
woop: I think it may be a naming problem. This is technically "run a query and unload the results to S3", but we are calling it "unload query", which is confusing to me.

sdk/python/feast/type_map.py
```diff
@@ -462,51 +358,7 @@ def pa_to_feast_value_type(value: Union[pa.lib.ChunkedArray, str]) -> ValueType:
...
-
-    proto_column = []
-    for val in column:
-        timestamp = Timestamp()
-        timestamp.FromMicroseconds(micros=int(val.as_py().timestamp() * 1_000_000))
-        proto_column.append(timestamp)
-    return proto_column
-
-
-def pa_column_to_proto_column(
```
woop, line 480: So this function is not in use anywhere?

sdk/python/tests/test_offline_online_store_consistency.py
```diff
@@ -112,6 +114,79 @@ def prep_bq_fs_and_fv(
...
+            ),
+            offline_store=offline_store,
+        )
+        fs = FeatureStore(config=config)
+        fs.apply([fv, e])
+
+        yield fs, fv
+
+    # Clean up the uploaded Redshift table
+    aws_utils.execute_redshift_statement(
```
woop, line 181: Does this command execute even on failure?

sdk/python/feast/infra/offline_stores/redshift.py
tsotnet: Yes. Check out
https://stackoverflow.com/questions/52087307/adding-type-hinting-to-functions-that-return-boto3-objects
https://stackoverflow.com/questions/51118767/what-is-boto3-clients3-returning

sdk/python/feast/type_map.py
tsotnet: It was never used (check https://github.com/feast-dev/feast/search?q=pa_column_to_proto_column)

sdk/python/tests/test_offline_online_store_consistency.py
tsotnet: Yes that's how contextmanagers work (https://pymotw.com/2/contextlib/#from-generator-to-context-manager)

sdk/python/feast/infra/utils/aws_utils.py
tsotnet: Ah you're right mostly it's type assertions. There are couple of them which do actual checks though:
```
$ pygrep -r "assert " sdk/python/feast | grep -v isinstance | grep -v type
sdk/python/feast/infra/online_stores/sqlite.py:        assert (
sdk/python/feast/infra/local.py:        assert config is not None
sdk/python/feast/infra/utils/aws_utils.py:    assert s3_path.startswith("s3://")
sdk/python/feast/staging/storage_client.py:        assert remote_path_prefix is not None and remote_path_suffix is not None
sdk/python/feast/staging/storage_client.py:        assert remote_path.hostname is not None
sdk/python/feast/staging/storage_client.py:        assert remote_path.hostname is not None
sdk/python/feast/staging/storage_client.py:        assert uri.hostname == f"{self.account_name}.blob.core.windows.net"
sdk/python/feast/staging/storage_client.py:        assert uri.username
sdk/python/feast/feature_view.py:        assert _input is not None
sdk/python/feast/repo_config.py:        assert "provider" in values
sdk/python/feast/repo_config.py:        assert "provider" in values
```

sdk/python/feast/infra/utils/aws_utils.py
tsotnet: I disagree with "run a query and unload the results to S3". Running a query and unloading the results to S3 are not 2 separate events. Running the query is what unloads the data to S3. It's like saying "select query". In this case it's "unload query". What would you rename the function to?

sdk/python/feast/infra/utils/aws_utils.py
woop: Ok, well we do need to clean that up as well. But I think in some cases we have Optional[foo] being asserted into `foo` by assert statements, so it may also be for MyPy

sdk/python/feast/infra/utils/aws_utils.py
woop: > Running the query is what unloads the data to S3

Yes, because you are executing both the table creation query and table export with a `;` separator. I would have called it something like `execute_redshift_query_and_unload_to_s3`.

sdk/python/feast/infra/utils/aws_utils.py
tsotnet: Yeah that's true. I'll open a separate task for that.

woop (approved):

Issue #1694, opened 2021-07-07
Entities are never deleted from the Registry


achals:
## Expected Behavior 

When an entity is removed from the feature repo, it should be removed from the feature registry.

## Current Behavior

Entities are only added, never removed from the Registry. This is a storage leak of sorts.

## Steps to reproduce

- `feast init`
- In the new feature repo, `feast apply`
- Add a new entity in the feature repo. `feast apply`.
- Remove the new entity from the feature repo. `feast apply`.
- `feast registry-dump` will have deleted entity, even though no feature tables or views depend on it.

## Possible Solution

in `repo_operations.py#apply_total`, we need to keep track of the entities to delete, add an API in the feature store to delete entites, and call this new API.

amommendes:
Hey @achals, is this still open? Becaus, I can´t reproduce this in `Feast SDK Version: "feast 0.11.1.dev49+gbf557bcb"`.

dejavu284:
This issue is still relevant

franciscojavierarceo:
Oh we'll take a look at this! Thanks for reopening @tokoko !

meyudi:
Hello @dejavu284 
I added, applied and then deleted an `Entity` with
```
feast version
Feast SDK Version: "0.1.dev3709+ge5527ad"
```
followed by 
```
feast entities list
feast registry-dump
```

The deleted entity was not listed after deletion.


> This issue is still relevant
Could you share how to reproduce the issue? Perhaps it was fixed recently.

franciscojavierarceo:
@meyudi shat version of feast are you using? That looks like a pretty old one.

meyudi:
@franciscojavierarceo  I am using 
```
Feast SDK Version: "0.1.dev3709+ge5527ad"
```

franciscojavierarceo:
are you building that from source? or are you pip installing it?

PR #1697, opened 2021-07-08
Teardown infrastructure after integration tests

### LINKED ISSUE #1696: The FeatureStore class should expose a `teardown` method
## Expected Behavior 

A user should be able to use the FeatureStore class to teardown any infrastructure setup when `apply` was called. 

## Current Behavior

There's no user facing API to allow this. Users must call `feast teardown` as a CLI command.

## Possible Solution

The FeatureStore class should expose a `teardown` method. All CLI commands should delegate to FeatureStore methods.

achals:

The `test_offline_online_store_consistency.py` integration test didn't tear down any infrastructure after running. This led to an issue where we ran past the number of dynamo tables allowed for an account: https://github.com/feast-dev/feast/runs/3014392319?check_suite_focus=true

This PR changes the integration test to tear down the infrastructure set up by the fixtures. 

Fixes #1696. 

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -21,9 +21,10 @@ import pandas as pd
 from colorama import Fore, Style
 from tqdm import tqdm
 
-from feast import FeatureTable, utils
+from feast import utils
 from feast.entity import Entity
 from feast.errors import FeatureNameCollisionError, FeatureViewNotFoundException
+from feast.feature_table import FeatureTable
 from feast.feature_view import FeatureView
 from feast.inference import (
     update_data_sources_with_inferred_event_timestamp_col,
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -21,7 +21,7 @@ import pandas as pd
 from colorama import Fore, Style
 from tqdm import tqdm
 
-from feast import utils
+from feast import FeatureTable, utils
 from feast.entity import Entity
 from feast.errors import FeatureNameCollisionError, FeatureViewNotFoundException
 from feast.feature_view import FeatureView
@@ -253,6 +253,23 @@ class FeatureStore:
             entities_to_keep=entities_to_update,
             partial=True,
         )
+
+    @log_exceptions_and_usage
+    def teardown(self):
+        feature_defs: List[Union[FeatureView, FeatureTable]] = []
+        feature_views = self.list_feature_views()
+        feature_tables = self._registry.list_feature_tables(self.project)
+
+        feature_defs.extend(feature_views)
+        feature_defs.extend(feature_tables)
+
+        entities = self.list_entities()
+
+        self._get_provider().teardown_infra(self.project, feature_defs, entities)
+        for feature_view in feature_views:
+            self.delete_feature_view(feature_view.name)
+        for feature_table in feature_tables:
+            self._registry.delete_feature_table(feature_table.name, self.project)
 
     @log_exceptions_and_usage
     def get_historical_features(
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -257,16 +257,16 @@ class FeatureStore:
 
     @log_exceptions_and_usage
     def teardown(self):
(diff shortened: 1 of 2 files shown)


tsotnet:
@achals Should we add the teardown method to the FeatureStore class instead of calling a private method on it?

achals:
> @achals Should we add the teardown method to the FeatureStore class instead of calling a private method on it?

Definitely, that's what I wanted with #1696. I made this PR because I thought it may need to be landed sooner than that - if we're okay with me taking a few more hours I can fix that issue and update this PR to use the teardown method directly. What do you think?

woop:
> > @achals Should we add the teardown method to the FeatureStore class instead of calling a private method on it?
> 
> Definitely, that's what I wanted with #1696. I made this PR because I thought it may need to be landed sooner than that - if we're okay with me taking a few more hours I can fix that issue and update this PR to use the teardown method directly. What do you think?

sgtm

sdk/python/feast/feature_store.py
```diff
@@ -254,6 +254,23 @@ def apply(
...
+        feature_views = self.list_feature_views()
+        feature_tables = self._registry.list_feature_tables(self.project)
+
+        feature_defs.extend(feature_views)
+        feature_defs.extend(feature_tables)
+
+        entities = self.list_entities()
+
+        self._get_provider().teardown_infra(self.project, feature_defs, entities)
+        for feature_view in feature_views:
```
felixwang9817, line 270: Why are we deleting feature_views and feature_tables but not deleting entities?

achals: There's no support to delete the entities in the registry yet: #1694 . But it should exist, and be called from here.

Yes, these lines are intended to tear down the registry, but this behaviour is different from what the CLI does - you can check `teardown` in `repo_operations`, and it doesn't tear down the registry. That should be rationalized (i.e. `repo_operations` should call the feature store method, and the registry *should* be torn down.)

sdk/python/feast/feature_store.py
```diff
@@ -254,6 +255,23 @@ def apply(
             partial=True,
         )
 
+    @log_exceptions_and_usage
+    def teardown(self):
+        feature_defs: List[Union[FeatureView, FeatureTable]] = []
```
woop, line 260: Should we call this `tables`? That is what we're currently calling it elsewhere.

sdk/python/tests/test_offline_online_store_consistency.py
```diff
@@ -133,10 +135,13 @@ def prep_local_fs_and_fv() -> Iterator[Tuple[FeatureStore, FeatureView]]:
             join_key="driver_id",
             value_type=ValueType.INT32,
         )
+        project = f"test_local_correctness_{str(uuid.uuid4()).replace('-', '')}"
+        print(f"Using project: {project}")
```
woop, line 139: Should this print be here?

achals: Since we generate the project dynamicaly and this is for integration tests, I think it's actually useful for debugging. I can remove it if you feel strongly about removing it

woop: I dont feel strongly.

achals: In that case I'm going to leave it in here

sdk/python/tests/test_offline_online_store_consistency.py
```diff
@@ -147,6 +152,9 @@ def prep_local_fs_and_fv() -> Iterator[Tuple[FeatureStore, FeatureView]]:
 
             yield fs, fv
 
+            # This should be a call to `fs.teardown()` once https://github.com/feast-dev/feast/issues/1696 is resolved.
```
felixwang9817, line 155: Should this comment be deleted?

felixwang9817:
Had one small nit, otherwise lgtm

felixwang9817 (approved):
had one small nit, otherwise lgtm

felixwang9817 (approved):
/lgtm

PR #1698, opened 2021-07-08
Increase efficiency of Registry updates

felixwang9817:

**What this PR does / why we need it**: make Registry updates more efficient. Right now, Registry updates are done by looping through FeatureViews and Entities and applying them one at a time. Applying an update to the Registry forces the underlying RegistryStore to update the underlying RegistryProto and write it (either to disk, GCS, or S3, depending on the backend of choice). Thus, looped writes are highly inefficient.

Current design: the Registry current has a cache to improve reading speed. Writes always go through to the backend, but not through the cache. Thus, the cache is not always up-to-date with the ground truth in the backend. The user is granted a lot of optionality around the cache: they may choose not to use the cache, thus reading from backend directly, or they may force the cache to refresh. 

After this PR: the cache will be write-back, meaning writes always go to the cache, and only go the backend when flushed. Then, looping updates to the Registry will be fast since all updates will be performed in-memory, and flushed once at the end. Note that this change only affects writing to the cache. Reading from the cache is unchanged: the `refresh` and `cache_ttl` mechanisms are still available to readers.

Advantages: looping updates become fast. Arguably, makes the boundary between the Registry and RegistryStore classes more clear (right now the RegistryStore takes in an updater function which means it is changing the proto itself, whereas in the new version the Registry would be changing the proto, which makes more sense since the RegistryStore should only be storing the proto, not changing it). 

Disadvantages: changes the Registry interface (but not in a breaking fashion). More importantly, introduces more complexity around managing concurrent access to a Registry. Also, theoretically leaves a greater window for errors to occur (e.g. a bunch of operations could occur on the in-memory cache, then the process could crash without persisting those changes). 

Alternative solution: expose a new function such as apply_all_changes(feature_views: List[FeatureView], entities: List[Entity]) that applies all changes in memory and then persists it. This solution does not require changing the current usage of the cache, but changes the Registry interface with a method that seems somewhat hacky.

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/registry.py
+++ b/sdk/python/feast/registry.py
@@ -18,7 +18,7 @@ from abc import ABC, abstractmethod
 from datetime import datetime, timedelta
 from pathlib import Path
 from tempfile import TemporaryFile
-from typing import Callable, List, Optional
+from typing import List, Optional
 from urllib.parse import urlparse
 
 from feast.entity import Entity
@@ -41,6 +41,9 @@ class Registry:
     Registry: A registry allows for the management and persistence of feature definitions and related metadata.
     """
 
+    # The cached_registry_proto object is used for both reads and writes. In particular,
+    # all write operations refresh the cache and modify it in memory; the write must
+    # then be persisted to the underlying RegistryStore with a call to commit().
     cached_registry_proto: Optional[RegistryProto] = None
     cached_registry_proto_created: Optional[datetime] = None
     cached_registry_proto_ttl: timedelta
@@ -73,34 +76,39 @@ class Registry:
         return
 
     def _initialize_registry(self):
-        """Explicitly forces the initialization of a registry"""
-        self._registry_store.update_registry_proto(updater=None)
+        """Explicitly initializes the registry with an empty proto."""
+        registry_proto = RegistryProto()
+        registry_proto.registry_schema_version = REGISTRY_SCHEMA_VERSION
+        self._registry_store.update_registry_proto(registry_proto)
 
-    def apply_entity(self, entity: Entity, project: str):
+    def apply_entity(self, entity: Entity, project: str, commit: bool = True):
         """
         Registers a single entity with Feast
 
         Args:
             entity: Entity that will be registered
             project: Feast project that this entity belongs to
+            commit: Whether the change should be persisted immediately
         """
         entity.is_valid()
         entity_proto = entity.to_proto()
         entity_proto.spec.project = project
+        self._prepare_registry_for_changes()
+        assert self.cached_registry_proto
 
-        def updater(registry_proto: RegistryProto):
-            for idx, existing_entity_proto in enumerate(registry_proto.entities):
-                if (
-                    existing_entity_proto.spec.name == entity_proto.spec.name
-                    and existing_entity_proto.spec.project == project
-                ):
-                    del registry_proto.entities[idx]
-                    registry_proto.entities.append(entity_proto)
-                    return registry_proto
-            registry_proto.entities.append(entity_proto)
-            return registry_proto
+        for idx, existing_entity_proto in enumerate(
+            self.cached_registry_proto.entities
+        ):
+            if (
+                existing_entity_proto.spec.name == entity_proto.spec.name
+                and existing_entity_proto.spec.project == project
+            ):
+                del self.cached_registry_proto.entities[idx]
+                break
 
-        self._registry_store.update_registry_proto(updater)
+        self.cached_registry_proto.entities.append(entity_proto)
+        if commit:
+            self.commit()
         return
 
     def list_entities(self, project: str, allow_cache: bool = False) -> List[Entity]:
@@ -139,71 +147,73 @@ class Registry:
                 return Entity.from_proto(entity_proto)
         raise EntityNotFoundException(name, project=project)
 
-    def apply_feature_table(self, feature_table: FeatureTable, project: str):
+    def apply_feature_table(
+        self, feature_table: FeatureTable, project: str, commit: bool = True
+    ):
         """
         Registers a single feature table with Feast
 
         Args:
             feature_table: Feature table that will be registered
             project: Feast project that this feature table belongs to
+            commit: Whether the change should be persisted immediately
         """
         feature_table.is_valid()
         feature_table_proto = feature_table.to_proto()
         feature_table_proto.spec.project = project
+        self._prepare_registry_for_changes()
+        assert self.cached_registry_proto
 
-        def updater(registry_proto: RegistryProto):
-            for idx, existing_feature_table_proto in enumerate(
-                registry_proto.feature_tables
+        for idx, existing_feature_table_proto in enumerate(
+            self.cached_registry_proto.feature_tables
+        ):
+            if (
+                existing_feature_table_proto.spec.name == feature_table_proto.spec.name
+                and existing_feature_table_proto.spec.project == project
             ):
-                if (
-                    existing_feature_table_proto.spec.name
-                    == feature_table_proto.spec.name
-                    and existing_feature_table_proto.spec.project == project
-                ):
-                    del registry_proto.feature_tables[idx]
-                    registry_proto.feature_tables.append(feature_table_proto)
-                    return registry_proto
-            registry_proto.feature_tables.append(feature_table_proto)
-            return registry_proto
+                del self.cached_registry_proto.feature_tables[idx]
+                break
 
-        self._registry_store.update_registry_proto(updater)
+        self.cached_registry_proto.feature_tables.append(feature_table_proto)
+        if commit:
+            self.commit()
         return
 
-    def apply_feature_view(self, feature_view: FeatureView, project: str):
+    def apply_feature_view(
+        self, feature_view: FeatureView, project: str, commit: bool = True
+    ):
         """
         Registers a single feature view with Feast
 
         Args:
             feature_view: Feature view that will be registered
             project: Feast project that this feature view belongs to
+            commit: Whether the change should be persisted immediately
         """
         feature_view.is_valid()
         feature_view_proto = feature_view.to_proto()
         feature_view_proto.spec.project = project
+        self._prepare_registry_for_changes()
+        assert self.cached_registry_proto
 
-        def updater(registry_proto: RegistryProto):
-            for idx, existing_feature_view_proto in enumerate(
-                registry_proto.feature_views
+        for idx, existing_feature_view_proto in enumerate(
+            self.cached_registry_proto.feature_views
+        ):
+            if (
+                existing_feature_view_proto.spec.name == feature_view_proto.spec.name
+                and existing_feature_view_proto.spec.project == project
             ):
-                if (
-                    existing_feature_view_proto.spec.name
-                    == feature_view_proto.spec.name
-                    and existing_feature_view_proto.spec.project == project
-                ):
-                    # do not update if feature view has not changed; updating will erase tracked materialization intervals
-                    if (
-                        FeatureView.from_proto(existing_feature_view_proto)
-                        == feature_view
-                    ):
-                        return registry_proto
-                    else:
-                        del registry_proto.feature_views[idx]
-                        registry_proto.feature_views.append(feature_view_proto)
-                        return registry_proto
-            registry_proto.feature_views.append(feature_view_proto)
-            return registry_proto
+                # do not update if feature view has not changed; updating will erase tracked materialization intervals
+                if FeatureView.from_proto(existing_feature_view_proto) == feature_view:
+                    return
+                else:
+                    del self.cached_registry_proto.feature_views[idx]
+                    break
 
-        self._registry_store.update_registry_proto(updater)
+        self.cached_registry_proto.feature_views.append(feature_view_proto)
+        if commit:
+            self.commit()
+        return
 
     def apply_materialization(
         self,
@@ -211,6 +221,7 @@ class Registry:
         project: str,
         start_date: datetime,
         end_date: datetime,
+        commit: bool = True,
     ):
         """
         Updates materialization intervals tracked for a single feature view in Feast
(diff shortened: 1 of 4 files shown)


sdk/python/feast/registry.py
```diff
@@ -69,54 +67,62 @@ def __init__(self, registry_path: str, repo_path: Path, cache_ttl: timedelta):
...
+        self.cached_registry_proto = self._get_registry_proto()
         return
 
     def _initialize_registry(self):
-        """Explicitly forces the initialization of a registry"""
-        self._registry_store.update_registry_proto(updater=None)
+        """Explicitly initializes the registry with an empty proto."""
+        self.cached_registry_proto = RegistryProto()
+        self.cached_registry_proto.registry_schema_version = REGISTRY_SCHEMA_VERSION
+        self._registry_store.update_registry_proto(self.cached_registry_proto)
```
achals, line 78: Hmm does this mean that it will write an empty registry object as soon as it's initialized? I guess that's okay but I wonder if it can be deferred to a `flush` call.

sdk/python/feast/registry.py
```diff
@@ -69,54 +67,62 @@ def __init__(self, registry_path: str, repo_path: Path, cache_ttl: timedelta):
...
 
-    def apply_entity(self, entity: Entity, project: str):
+    def apply_entity(self, entity: Entity, project: str, flush: bool = True):
         """
         Registers a single entity with Feast
 
         Args:
             entity: Entity that will be registered
             project: Feast project that this entity belongs to
+            flush: True if the change should be persisted immediately
```
achals, line 87: Nit, this should be phrased more like `Whether the changes made to the registry should be immediately made to the registry or not.`

sdk/python/feast/registry.py
```diff
@@ -69,54 +67,62 @@ def __init__(self, registry_path: str, repo_path: Path, cache_ttl: timedelta):
...
-            registry_proto.entities.append(entity_proto)
-            return registry_proto
+        for idx, existing_entity_proto in enumerate(
+            self.cached_registry_proto.entities
+        ):
+            if (
+                existing_entity_proto.spec.name == entity_proto.spec.name
+                and existing_entity_proto.spec.project == project
+            ):
+                found_feature_view = True
```
achals, line 101: Sorry what does finding feature views have to do with the entity updating function?

achals (approved):
Looks good to me.

achals:
@felixwang9817 linter needs to be fixed, and I had some questsions/nits

sdk/python/feast/registry.py
felixwang9817: Good point, I think we can defer it to `flush`.

sdk/python/feast/registry.py
felixwang9817: Good point.

sdk/python/feast/registry.py
felixwang9817: Good point. I accidentally named all the booleans the same thing; I'll fix this.

sdk/python/feast/registry.py
```diff
@@ -356,57 +374,33 @@ def delete_feature_view(self, name: str, project: str):
...
             raise FeatureViewNotFoundException(name, project)
 
-        self._registry_store.update_registry_proto(updater)
+        self._registry_store.update_registry_proto(self.cached_registry_proto)
 
-    def refresh(self):
-        """Refreshes the state of the registry cache by fetching the registry state from the remote registry store."""
-        self._get_registry_proto(allow_cache=False)
+    def flush(self):
+        """Flushes the state of the registry cache to the reomte registry store."""
```
woop, line 392: Typo

sdk/python/feast/registry.py
```diff
@@ -35,26 +36,33 @@
...
-    cache_being_updated: bool = False
+    cached_registry_proto: RegistryProto
 
-    def __init__(self, registry_path: str, repo_path: Path, cache_ttl: timedelta):
+    def __init__(
+        self,
+        registry_path: str,
+        repo_path: Path,
+        cache_ttl: timedelta = None,
+        from_feast=False,
```
achals, line 54: nit, I think we want a better param than `from_feast`. Maybe `instantiated_internally`? Open to other ideas.

woop:
> By changing the Registry's cache to be write-back instead of write-through

I think it's important that this PR have explicit "documentation" on the changes that are being introduced to the registry. @felixwang9817 can you please write out the exact changes (and trade offs) that we are making here? What is the before/after?

sdk/python/feast/registry.py
felixwang9817: That's a good point. `instantiated_internally` is much more clear; I can't think of anything better, so I'll go with that.

felixwang9817:
> > By changing the Registry's cache to be write-back instead of write-through
> 
> I think it's important that this PR have explicit "documentation" on the changes that are being introduced to the registry. @felixwang9817 can you please write out the exact changes (and trade offs) that we are making here? What is the before/after?

@woop I added some documentation to the PR, let me know if you want any further clarification.

sdk/python/tests/test_registry.py
```diff
@@ -0,0 +1,347 @@
...
+)
+def test_apply_entity_success(test_registry):
+    entity = Entity(
+        name="driver_car_id",
+        description="Car driver id",
+        value_type=ValueType.STRING,
+        labels={"team": "matchmaking"},
+    )
+
+    project: StrictStr = "project"
```
achals, line 73: nit, `strictstr` is needed just for `pydantic` models. This can be `project = "project"`

sdk/python/tests/test_registry.py
```diff
@@ -0,0 +1,347 @@
...
+        and feature_views[0].features[0].name == "fs1_my_feature_1"
+        and feature_views[0].features[0].dtype == ValueType.INT64
+        and feature_views[0].features[1].name == "fs1_my_feature_2"
+        and feature_views[0].features[1].dtype == ValueType.STRING
+        and feature_views[0].features[2].name == "fs1_my_feature_3"
+        and feature_views[0].features[2].dtype == ValueType.STRING_LIST
+        and feature_views[0].features[3].name == "fs1_my_feature_4"
+        and feature_views[0].features[3].dtype == ValueType.BYTES_LIST
+        and feature_views[0].entities[0] == "fs1_my_entity_1"
+    )
```
achals, line 186: I'm worried that this is flaky because it depends on the ordering of things in the features views/features. Can we do something that doesn't rely on indexing into the lists?

sdk/python/feast/registry.py
```diff
@@ -69,54 +77,69 @@ def __init__(self, registry_path: str, repo_path: Path, cache_ttl: timedelta):
...
+                existing_entity_proto.spec.name == entity_proto.spec.name
+                and existing_entity_proto.spec.project == project
+            ):
+                found_entity = True
+                del self.cached_registry_proto.entities[idx]
+                self.cached_registry_proto.entities.append(entity_proto)
+                break
+
+        if not found_entity:
+            self.cached_registry_proto.entities.append(entity_proto)
```
achals, line 124: how about simplifying this to something like:

```
    for ...
        if blah:
                del self.cached_registry_proto.entities[idx]
                break


    self.cached_registry_proto.entities.append(entity_proto)
```

achals: Same for other things (views, tables, etc)

sdk/python/feast/registry.py
```diff
@@ -356,57 +391,42 @@ def delete_feature_view(self, name: str, project: str):
...
-        self._registry_store.update_registry_proto(updater)
+        self._registry_store.update_registry_proto(self.cached_registry_proto)
 
     def refresh(self):
-        """Refreshes the state of the registry cache by fetching the registry state from the remote registry store."""
-        self._get_registry_proto(allow_cache=False)
+        """
+        Deprecated.
+        """
+        pass
```
achals, line 415: Should this maybe call `_get_registry_proto` to get the latest value from the registry store? Otherwise I think we don't have a way of getting the latest value after creating a registry object.

sdk/python/feast/registry.py
```diff
@@ -35,26 +36,33 @@
...
+        Create the Registry object. It uses a write-back cache for speed, so the cached
+        registry proto is always up-to-date.
 
         Args:
             repo_path: Path to the base of the Feast repository
-            cache_ttl: The amount of time that cached registry state stays valid
+            cache_ttl: [Deprecated] Unnecessary since the cache is always up-to-date
             registry_path: filepath or GCS URI that is the location of the object store registry,
             or where it will be created if it does not exist yet.
+            instantiated_internally: Whether this Registry is being created and used by Feast
```
woop, line 65: `instantiated_internally` is a bit of an obscure name. What if we just support kwargs and then set a Feast specific key? otherwise we may be confusing users who don't know how to configure this argument.

sdk/python/feast/registry.py
woop: Yea. How can I forcefully get the latest registry?

woop:
> Signed-off-by: Felix Wang [wangfelix98@gmail.com](mailto:wangfelix98@gmail.com)
> 
> **What this PR does / why we need it**: make Registry updates more efficient. Right now, Registry updates are done by looping through FeatureViews and Entities and applying them one at a time. Applying an update to the Registry forces the underlying RegistryStore to update the underlying RegistryProto and write it (either to disk, GCS, or S3, depending on the backend of choice). Thus, looped writes are highly inefficient.
> 
> Current design: the Registry current has a cache to improve reading speed. Writes always go through to the backend, but not through the cache. Thus, the cache is not always up-to-date with the ground truth in the backend. The user is granted a lot of optionality around the cache: they may choose not to use the cache, thus reading from backend directly, or they may force the cache to refresh.
> 
> After this PR: the cache will be write-back, meaning writes always go to the cache, and only go the backend when flushed. Then, looping updates to the Registry will be fast since all updates will be performed in-memory, and flushed once at the end.
> 
> Advantages: looping updates become fast. Also, leads to some refactoring opportunities (e.g. removing refresh_registry from FeatureStore, removing cache_ttl argument and the associated logic for Registry). Arguably, makes the boundary between the Registry and RegistryStore classes more clear (right now the RegistryStore takes in an updater function which means it is changing the proto itself, whereas in the new version the Registry would be changing the proto, which makes more sense since the RegistryStore should only be storing the proto, not changing it).
> 
> Disadvantages: changes the Registry interface (but not in a breaking fashion). More importantly, introduces more complexity around managing concurrent access to a Registry. Also, theoretically leaves a greater window for errors to occur (e.g. a bunch of operations could occur on the in-memory cache, then the process could crash without persisting those changes).
> 
> Alternative solution: expose a new function such as apply_all_changes(feature_views: List[FeatureView], entities: List[Entity]) that applies all changes in memory and then persists it. This solution does not require changing the current usage of the cache, but changes the Registry interface with a method that seems somewhat hacky.
> 
> **Which issue(s) this PR fixes**:
> 
> Fixes #
> 
> **Does this PR introduce a user-facing change?**:
> 
> ```
> NONE
> ```

One part that isn't clear: If I am a reader only (like a model serving application that has no write access to the registry and doesn't plan to run `apply` at all), are there any regressions/changes to the way I do things? How do I update the registry? Is it being done at `get_online_features()` time (which is lazy) or `FeatureStore()` initialization? How do I keep the registry update based on TTL or through manual operations like `refresh()`? do any of these changes affect me?

sdk/python/feast/registry.py
felixwang9817: There is an internal method for getting the latest registry, `_get_registry_proto`. This gets called on initialization, so the Registry object will be populated with the latest saved registry. Since the cached registry proto is always up-to-date, I don't see why `refresh` would ever be needed?

sdk/python/feast/registry.py
felixwang9817: Good point; I agree that using kwargs should ensure that users don't get confused.

sdk/python/feast/registry.py
felixwang9817: Good point, I'll make this fix.

sdk/python/tests/test_registry.py
felixwang9817: Good point. I pulled this pattern from `test_feature_store.py`. One thought is that maybe this indexing is deliberate, in order to test that the features were in fact listed in the correct order?

achals: Hmmm. I am not sure if we guarantee any ordering in the responses (but happy to address that in a later diff)

felixwang9817:
> 
> One part that isn't clear: If I am a reader only (like a model serving application that has no write access to the registry and doesn't plan to run `apply` at all), are there any regressions/changes to the way I do things? How do I update the locally cached registry? Is it being done at `get_online_features()` time (which is lazy) or `FeatureStore()` initialization? How do I keep the in memory local registry updated based on TTL or through manual operations like `refresh()`? do any of these changes affect me?

@woop I revised my documentation, see above. tldr: there are no changes for a reader, only a writer.

sdk/python/feast/repo_operations.py
```diff
@@ -167,33 +160,43 @@ def apply_total(repo_config: RepoConfig, repo_path: Path):
         if registry_view.name not in repo_table_names:
             views_to_delete.append(registry_view)
 
+    sys.dont_write_bytecode = False
+    for entity in repo.entities:
+        registry.apply_entity(entity, project=project, commit=False)
+        click.echo(
+            f"Registered entity {Style.BRIGHT + Fore.GREEN}{entity.name}{Style.RESET_ALL}"
+        )
+
```
achals, line 169: Is there a reason why this was moved to after the feature views?

sdk/python/tests/test_registry.py
```diff
@@ -0,0 +1,347 @@
+# Copyright 2019 The Feast Authors
```
achals, line 1: nit, 2021

sdk/python/feast/repo_operations.py
felixwang9817: Yes, this is necessary because calling `list_feature_tables` and `list_feature_views` will forcibly refresh the cache, causing any uncommitted changes to be lost, so we either need to (a) immediately commit the `apply_entity` changes or (b) call `apply_entity` after `list_feature_tables` and `list_feature_views`.

achals: ok cool

PR #1699, opened 2021-07-09
Cancel BigQuery job if block_until_done call times out or is interrupted

codyjlin:

If a user decides to abort a `to_bigquery` run (like via a `ctrl-c`), the job should also be cancelled. The PR also fixes the retry logic of getting the bq_job state, and raises a better exception on timeout.

(Not large enough for issue)

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -10,7 +10,7 @@ from jinja2 import BaseLoader, Environment
 from pandas import Timestamp
 from pydantic import StrictStr
 from pydantic.typing import Literal
-from tenacity import retry, stop_after_delay, wait_fixed
+from tenacity import retry, retry_if_exception_type, stop_after_delay, wait_fixed
 
 from feast import errors
 from feast.data_source import BigQuerySource, DataSource
@@ -301,7 +301,12 @@ def block_until_done(client, bq_job):
         BigQueryJobCancelled exception on a KeyboardInterrupt.
     """
 
-    @retry(wait=wait_fixed(10), stop=stop_after_delay(1800), reraise=True)
+    @retry(
+        wait=wait_fixed(10),
+        stop=stop_after_delay(1800),
+        retry=retry_if_exception_type(BigQueryJobStillRunning),
+        reraise=True,
+    )
     def _wait_until_done(job_id):
         if client.get_job(job_id).state in ["PENDING", "RUNNING"]:
             raise BigQueryJobStillRunning(job_id=job_id)
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -287,8 +287,15 @@ class BigQueryRetrievalJob(RetrievalJob):
 
 
 def block_until_done(client, bq_job):
+    def _cancel_job(job_id):
+        client.cancel_job(job_id)
+        raise BigQueryJobCancelled(job_id=job_id)
+
     def _is_done(job_id):
-        return client.get_job(job_id).state in ["PENDING", "RUNNING"]
+        try:
+            return client.get_job(job_id).state in ["PENDING", "RUNNING"]
+        except KeyboardInterrupt:
+            _cancel_job(job_id)
 
     @retry(wait=wait_fixed(10), stop=stop_after_delay(1800), reraise=True)
     def _wait_until_done(job_id):
@@ -301,8 +308,7 @@ def block_until_done(client, bq_job):
         raise bq_job.exception()
 
     if not _is_done(job_id):
-        client.cancel_job(job_id)
-        raise BigQueryJobCancelled(job_id=job_id)
+        _cancel_job(job_id)
 
 
 @dataclass(frozen=True)
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -266,7 +266,7 @@ class BigQueryRetrievalJob(RetrievalJob):
         Args:
             job_config: An optional bigquery.QueryJobConfig to specify options like destination table, dry run, etc.
             timeout: An optional number of seconds for setting the time limit of the QueryJob.
-            retry_cadence: An optional number of seconds for setting how long the job should checked for completion. 
+            retry_cadence: An optional number of seconds for setting how long the job should checked for completion.
 
         Returns:
             Returns the destination table name or returns None if job_config.dry_run is True.
@@ -308,7 +308,7 @@ def block_until_done(
         client: A bigquery.client.Client to monitor the bq_job.
         bq_job: The bigquery.job.QueryJob that blocks until done runnning.
         timeout: An optional number of seconds for setting the time limit of the job.
-        retry_cadence: An optional number of seconds for setting how long the job should checked for completion. 
+        retry_cadence: An optional number of seconds for setting how long the job should checked for completion.
 
     Raises:
         BigQueryJobStillRunning exception if the function has blocked longer than 30 minutes.
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -360,7 +360,7 @@ def _upload_entity_df_into_bigquery(
 
     if type(entity_df) is str:
         job = client.query(f"CREATE TABLE {table_id} AS ({entity_df})")
-        job.result()
+        block_until_done(client, job)
     elif isinstance(entity_df, pandas.DataFrame):
         # Drop the index so that we dont have unnecessary columns
         entity_df.reset_index(drop=True, inplace=True)
@@ -370,7 +370,7 @@ def _upload_entity_df_into_bigquery(
         job = client.load_table_from_dataframe(
             entity_df, table_id, job_config=job_config
         )
-        job.result()
+        block_until_done(client, job)
     else:
         raise ValueError(
             f"The entity dataframe you have provided must be a Pandas DataFrame or BigQuery SQL query, "
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -10,7 +10,7 @@ from jinja2 import BaseLoader, Environment
 from pandas import Timestamp
 from pydantic import StrictStr
 from pydantic.typing import Literal
-from tenacity import retry, retry_if_exception_type, stop_after_delay, wait_fixed
+from tenacity import Retrying, retry_if_exception_type, stop_after_delay, wait_fixed
 
 from feast import errors
 from feast.data_source import BigQuerySource, DataSource
@@ -253,12 +253,20 @@ class BigQueryRetrievalJob(RetrievalJob):
         """
         return self.query
 
-    def to_bigquery(self, job_config: bigquery.QueryJobConfig = None) -> Optional[str]:
+    def to_bigquery(
+        self,
+        job_config: bigquery.QueryJobConfig = None,
+        timeout: int = 1800,
+        retry_cadence: int = 10,
+    ) -> Optional[str]:
         """
         Triggers the execution of a historical feature retrieval query and exports the results to a BigQuery table.
+        Runs for a maximum amount of time specified by the timeout parameter (defaulting to 30 minutes).
 
         Args:
             job_config: An optional bigquery.QueryJobConfig to specify options like destination table, dry run, etc.
+            timeout: An optional number of seconds for setting the time limit of the QueryJob.
+            retry_cadence: An optional number of seconds for setting how long the job should checked for completion. 
 
         Returns:
             Returns the destination table name or returns None if job_config.dry_run is True.
@@ -278,10 +286,7 @@ class BigQueryRetrievalJob(RetrievalJob):
             )
             return None
 
-        block_until_done(client=self.client, bq_job=bq_job)
-
-        if bq_job.exception():
-            raise bq_job.exception()
+        block_until_done(client=self.client, bq_job=bq_job, timeout=timeout)
 
         print(f"Done writing to '{job_config.destination}'.")
         return str(job_config.destination)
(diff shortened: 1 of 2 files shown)


mavysavydav:
if the notebook server/kernel crashes or is shutdown or is interrupted, that doesn't count as a keyboard interrupt right?

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -287,8 +287,15 @@ def to_arrow(self) -> pyarrow.Table:
...
 
 def block_until_done(client, bq_job):
+    def _cancel_job(job_id):
+        client.cancel_job(job_id)
+        raise BigQueryJobCancelled(job_id=job_id)
+
     def _is_done(job_id):
-        return client.get_job(job_id).state in ["PENDING", "RUNNING"]
+        try:
+            return client.get_job(job_id).state in ["DONE", "SUCCESS"]
```
codyjlin, line 296: @MattDelac Is this right? I'm not too familiar with the retry logic, but currently the job will be cancelled despite succeeding via the code a few lines below:

```python
    if not _is_done(job_id):
        _cancel_job(job_id)
```

codyjlin: Before, it seems like we might be giving back control to the user before the job is finished, which could be a problem if they are trying to access the results (i.e. to be stored in a dest. table) right?

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -287,8 +287,15 @@ def to_arrow(self) -> pyarrow.Table:
 
 
 def block_until_done(client, bq_job):
+    def _cancel_job(job_id):
```
mavysavydav, line 290: i don't think this needs to be in its own method. It's actually easier to see what's going on when these two lines are explicitly stated at the location where it's called.

MattDelac (approved):
Good catch about `KeyboardInterrupt` 🙌 
LGTM

MattDelac:
Why did you change everything ? I feel that the previous commit was much better ... 🤔

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -286,24 +286,24 @@ def to_arrow(self) -> pyarrow.Table:
         return self.client.query(self.query).to_arrow()
 
 
-def block_until_done(client, bq_job):
-    def _is_done(job_id):
-        return client.get_job(job_id).state in ["PENDING", "RUNNING"]
-
+def _block_until_done(client, bq_job):
```
MattDelac, line 289: Can we keep this function public ? Feast users could potentially use it rather than creating a similar version of it

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -286,24 +286,24 @@ def to_arrow(self) -> pyarrow.Table:
...
     @retry(wait=wait_fixed(10), stop=stop_after_delay(1800), reraise=True)
     def _wait_until_done(job_id):
-        return _is_done(job_id)
+        if client.get_job(job_id).state in ["PENDING", "RUNNING"]:
+            raise Exception
```
MattDelac, line 293: When reading the code, that makes no sense to raise an Exception everytime the state of the job is simply RUNNING ...

codyjlin: Before, the code inside the `@retry` annotation was not raising an `Exception`. Looking at the [tenacity retry docs](https://tenacity.readthedocs.io/en/latest/), it seems like this is the minimum to be take advantage of `retry`, is it not? Here I'm interpreting it as retrying every 10 seconds (up to max of 30 mins) until the job is anything but `PENDING` or `RUNNING` (i.e. `DONE`, `SUCCESS`, or `FAILURE` - see [docs](https://cloud.google.com/bigquery/docs/managing-jobs) for more).

sdk/python/feast/infra/offline_stores/bigquery.py
codyjlin: Oh interesting, did not consider that it was intentional to expose it to users. I was just thinking about what would happen if the user did exceed the 30 mins of the retry but the query was still running - I suppose this could be used in that case.

sdk/python/feast/infra/offline_stores/bigquery.py
codyjlin: If you have more expertise on the `@retry` logic, do please let me know!!

MattDelac: > In the old case, wouldn't it just pass first time around when there's no Exception raised?

Normally if the output of the function is a boolean, this should make it stop or continue
So I don't think that it was still retrying even when the job was done without exception. We've been using this function for the last couple of days and we never waited 30 min 🙂

MattDelac: Let me double check 🙂

MattDelac: Ok so something weird here but it sounds like we should use a [Custom callback](https://tenacity.readthedocs.io/en/latest/#custom-callbacks) where we pass our `_is_done()` function. For some reason, I thought that this was the current behaviour implemented 🤔 

For example, it works well with the following piece of code
```python
from tenacity import retry, wait_fixed, stop_after_delay, retry_if_result
from datetime import datetime

def get_state() -> int:
    number = int(datetime.now().timestamp())
    print(number)
    return number

def _is_done(number: int) -> bool:
    return (number % 10) != 0

@retry(wait=wait_fixed(1), stop=stop_after_delay(10), retry=retry_if_result(_is_done), reraise=True)
def test_retry():
    return get_state()
```

image

codyjlin: > Normally if the output of the function is a boolean, this should make it stop or continue

Oh interesting. Are there docs for this? I saw the part about custom callbacks [here](https://tenacity.readthedocs.io/en/latest/#custom-callbacks) where it seems like you have to specify `retry=retry_if_result(is_false)` so I assumed it's not a default behavior.

codyjlin: Just saw we commented around the same time haha. Let me implement the custom callback!

codyjlin: Actually, thinking more about it, do you think it would be cleaner? Doesn't seem like we really need the return value of `_is_done` (whether the state is in ["DONE", "SUCCESS"]) anymore in my implementation.

MattDelac: Yeah let me have a second thought 👍

MattDelac: * later today

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -287,23 +291,31 @@ def to_arrow(self) -> pyarrow.Table:
...
+        if client.get_job(job_id).state in ["PENDING", "RUNNING"]:
+            raise BigQueryJobStillRunning(job_id=job_id)
 
     job_id = bq_job.job_id
-    _wait_until_done(job_id=job_id)
+    try:
+        _wait_until_done(job_id=job_id)
+    except KeyboardInterrupt:
+        client.cancel_job(job_id)
+        raise BigQueryJobCancelled(job_id=job_id)
```
MattDelac, line 314: Should you also have a 

```python
except tenacity.RetryError as e:
   raise e
```
?

MattDelac: This is what happened when my 10 seconds timeout hit
image

codyjlin: Isn't it the same behavior with or without the second except clause?
```python
except tenacity.RetryError as e:
   raise e
```

I believe the try and except behavior (see [Python docs](https://docs.python.org/3/tutorial/errors.html#handling-exceptions)) will cause any other exception not named in the except clause to be raised as it would've:
> If an exception occurs which does not match the exception named in the except clause, it is passed on to outer try statements; if no handler is found, it is an unhandled exception and execution stops with a message as shown above.

MattDelac: Lol I am so stupid 🤦 
Sorry for giving you hard times with this PR, yesterday was wild.

Will have a final pass but it will still need to be approved by a Feast dev

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -287,23 +291,36 @@ def to_arrow(self) -> pyarrow.Table:
 
 
 def block_until_done(client, bq_job):
-    def _is_done(job_id):
-        return client.get_job(job_id).state in ["PENDING", "RUNNING"]
+    """
+    Waits a maximum of 30 minutes for bq_job to finish running.
+    Args:
+        client: A bigquery.client.Client to monitor the bq_job.
```
MattDelac, line 297: Thanks for the doc. Very useful.
[Optional] we could also add python types to the input of this function. As you want

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -287,23 +291,36 @@ def to_arrow(self) -> pyarrow.Table:
...
+    try:
+        _wait_until_done(job_id=job_id)
+    except KeyboardInterrupt:
+        client.cancel_job(job_id)
+        raise BigQueryJobCancelled(job_id=job_id)
 
     if bq_job.exception():
         raise bq_job.exception()
 
-    if not _is_done(job_id):
```
MattDelac, line 303: This piece of logic is still needed I think.

For example if the job timeout, an exception will be raised and the python script will end. But that won't cancel the BQ job.

Does that make sense ?

sdk/python/feast/infra/offline_stores/bigquery.py
codyjlin: No worries at all Matt! Thanks for all the time spent poking at this - just want to make sure things make sense.

sdk/python/feast/infra/offline_stores/bigquery.py
codyjlin: By timeout do you mean it took longer than the 30 minutes we allowed it to retry for? If yes, I wanted to make the behavior such that the job will still continue in case it is a job that requires longer than 30 minutes. I could perhaps make the error message more descriptive - currently it would be:
```python
BigQueryJobStillRunning ->  "The BigQuery job with ID '{job_id}' is still running."
```

Another option would be to cancel any job running for longer than 30 minutes, then maybe asking the user to manually retry running their query outside of the Feast api, like:
```python
client.query(<BigQueryRetrievalJob>.to_sql(), job_config=job_config)
```
Downside is they have to wait at least another 30 minutes, unless you think the first timeout on the Feast side might not actually mean that the job was taking that long to run but instead some problem with the BQ client API?

MattDelac: > Another option would be to cancel any job running for longer than 30 minutes, then maybe asking the user to manually retry running their query outside of the Feast api, like:

I was thinking that the solution would be to let the user pass a timeout if 30 min was not enough. I don't think that it's ideal to perform the query outside Feast.

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -287,23 +291,36 @@ def to_arrow(self) -> pyarrow.Table:
...
+        bq_job: The bigquery.job.QueryJob that blocks until done runnning.
+    Raises:
+        BigQueryJobStillRunning exception if the function has blocked longer than 30 minutes.
+        BigQueryJobCancelled exception on a KeyboardInterrupt.
+    """
 
-    @retry(wait=wait_fixed(10), stop=stop_after_delay(1800), reraise=True)
+    @retry(
+        wait=wait_fixed(10),
+        stop=stop_after_delay(1800),
```
achals, line 306: I'd love it if these were configurable by users (with sensible defaults)

sdk/python/feast/infra/offline_stores/bigquery.py
achals: I prefer adding type hints as well.

sdk/python/feast/infra/offline_stores/bigquery.py
codyjlin: by `pass a timeout` do you mean like a flag or argument to increase the time limit?

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -287,23 +291,36 @@ def to_arrow(self) -> pyarrow.Table:
...
     def _wait_until_done(job_id):
-        return _is_done(job_id)
+        if client.get_job(job_id).state in ["PENDING", "RUNNING"]:
+            raise BigQueryJobStillRunning(job_id=job_id)
 
     job_id = bq_job.job_id
-    _wait_until_done(job_id=job_id)
+    try:
+        _wait_until_done(job_id=job_id)
+    except KeyboardInterrupt:
```
achals, line 317: Just wondering, would it make more sense to have this in a `finally` block instead of a `except KeyboardInterrupt`? that may also handle sigterm/sigkill.

MattDelac: Good call. We could have a finally statement that send an asynchronous call to cancel the BQ job in case something goes wrong 👍

sdk/python/feast/infra/offline_stores/bigquery.py
MattDelac: Yes

Something like the following if it works (never tried it)

```python
def block_until_done(client, bq_job, timeout=1800):
...
     @retry(wait=wait_fixed(10), stop=stop_after_delay(timeout), reraise=True)
```

codyjlin: Sounds good! Looks like variables can't go in the decorator itself, so I'll attempt the second option [here](https://tenacity.readthedocs.io/en/latest/#changing-arguments-at-run-time):

sdk/python/feast/infra/offline_stores/bigquery.py
codyjlin: Added the `finally` clause to catch other needs for cancellation (including timeout and `KeyboardInterrupt`).

achals (approved):
/lgtm

MattDelac (approved):
This is great. Thanks for putting the effort 🙏

achals:
@codyjlin just need to fix lint, and then we should be god to go!

codyjlin:
> @codyjlin just need to fix lint, and then we should be god to go!

Ah sorry, should have ran lint locally first. Thanks @achals and @MattDelac for all the comments!

achals (approved):
/lgtm

Issue #1640, opened 2021-06-11
Materialization Errors for Float and Double Lists 


Agent007:
## Expected Behavior
1-hot encoded features should be materializeable to the online store without error.
## Current Behavior
Version 0.10 apparently introduced a regression where 1-hot encoded feature materialization throws errors.
## Steps to reproduce
```python
from datetime import datetime, timedelta
from feast import Entity, Feature, FeatureStore, FeatureView, FileSource, ValueType
import numpy as np
import pandas as pd
from pathlib import Path
df = pd.DataFrame(columns=['device_type', 'encoding', 'datetime'])
df['device_type'] = ['A', 'B']
df['encoding'] = [
    np.array([0.0, 1.0], dtype=np.float32),
    np.array([1.0, 0.0], dtype=np.float32)
]
df['datetime'] = pd.to_datetime([datetime.now(), datetime.now()])
device_encodings_path_str = 'data/device_encodings.parquet'
device_encodings_path = Path(device_encodings_path_str)
device_encodings_abs_path_str = str(device_encodings_path.absolute())
df.to_parquet(device_encodings_abs_path_str)
device_encodings = FileSource(path=device_encodings_abs_path_str,
                              event_timestamp_column='datetime')
device_type = Entity('device_type',
                     value_type=ValueType.STRING,
                     description='')
device_encoding_view = FeatureView(
    name="device_encoding",
    entities=["device_type"],
    features=[
        Feature(name="encoding", dtype=ValueType.FLOAT_LIST),
    ],
    ttl=timedelta(minutes=60),
    input=device_encodings,
    online=True,
)
fm = FeatureStore(repo_path=".")
fm.apply([device_type, device_encoding_view])
fm.materialize(start_date=datetime(2021, 1, 1), end_date=datetime.now())
```
The above code yields the below error output:
```shell
Materializing 2 feature views from 2021-01-01 00:00:00-08:00 to 2021-06-10 15:24:51-07:00 into the sqlite online store.
driver_hourly_stats:
100%|███████████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 2920.42it/s]
device_encoding:
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/anaconda3/envs/feast/lib/python3.8/site-packages/feast/telemetry.py", line 155, in exception_logging_wrapper
    result = func(*args, **kwargs)
[...]

PR #1700, opened 2021-07-10
Fix Python to Feast `list[int]` mapping

judahrand:
Corrects a case where if elements in list were not `numpy` types
type mapping would fail.

Fixes case where type mapping failed (in my case from Parquet FileSource)

Fixes #

```release-note

```

### DIFF
--- a/sdk/python/feast/type_map.py
+++ b/sdk/python/feast/type_map.py
@@ -235,7 +235,9 @@ def _python_value_to_proto_value(feast_value_type, value) -> ProtoValue:
             return ProtoValue(
                 int32_list_val=Int32List(
                     val=[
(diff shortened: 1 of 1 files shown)


judahrand:
Fixes an issue mentioned in #1640 but not the whole issue as this still does not deal with empty lists.

woop:
This is great @judahrand!

I'd love to merge this PR, but unfortunately it will probably be blocked by the lack of a test that ensures all our types are working. That isn't really your fault, it's just a technical debt in the project. We still need to write a test that runs through all the types.

adchia:
Wanted to follow up here. I believe this was already fixed in the main repo (presumably after this PR), so closing this. Thanks again though for adding the fix! We added some basic tests for types, but still a work in progress.

PR #1703, opened 2021-07-11
Handle case where`_LIST` type is empty

judahrand:
This PR deals with multiple issues around empty list features regarding type inference.

It is my opinion that these fixes should be somewhat of a stop gap and that really the way that types are dealt with should be overhauled in Feast.

Fixes #

```release-note
Fix issue which prevented empty lists being materialized into online stores or retrieved from historical stores.
```

### DIFF
--- a/sdk/python/feast/feature.py
+++ b/sdk/python/feast/feature.py
@@ -40,7 +40,7 @@ class Feature:
         if not isinstance(dtype, ValueType):
             raise ValueError("dtype is not a valid ValueType")
         if dtype is ValueType.UNKNOWN:
-            raise ValueError(f"dtype cannont be {dtype}")
+            raise ValueError(f"dtype cannot be {dtype}")
         self._dtype = dtype
         if labels is None:
             self._labels = dict()
--- a/sdk/python/feast/feature.py
+++ b/sdk/python/feast/feature.py
@@ -39,6 +39,8 @@ class Feature:
         self._name = name
         if not isinstance(dtype, ValueType):
             raise ValueError("dtype is not a valid ValueType")
+        if dtype is ValueType.UNKNOWN:
+            raise ValueError(f"dtype cannont be {dtype}")
         self._dtype = dtype
         if labels is None:
             self._labels = dict()
--- a/sdk/python/feast/online_response.py
+++ b/sdk/python/feast/online_response.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from collections import defaultdict
 from typing import Any, Dict, List, cast
 
 import pandas as pd
@@ -25,7 +26,7 @@ from feast.protos.feast.types.Value_pb2 import Value as Value
 from feast.type_map import (
     _python_value_to_proto_value,
     feast_value_type_to_python_type,
-    python_type_to_feast_value_type,
+    python_values_to_feast_value_type,
 )
 
 
@@ -94,25 +95,30 @@ def _infer_online_entity_rows(
     entity_rows_dicts = cast(List[Dict[str, Any]], entity_rows)
     entity_row_list = []
     entity_type_map = dict()
+    entity_value_map = defaultdict(list)
+
+    # Flatten keys-value dicts into lists for type inference
+    for entity in entity_rows_dicts:
+        for key, value in entity.items():
+            if not isinstance(value, Value):
+                entity_value_map[key].append(value)
+
+    # Loop over all entities to infer dtype first in case of empty lists or nulls
+    for key, values in entity_value_map.items():
+        entity_type_map[key] = python_values_to_feast_value_type(key, values)
 
     for entity in entity_rows_dicts:
         fields = {}
         for key, value in entity.items():
-            # Allow for feast.types.Value
+            if key not in entity_type_map:
+                raise ValueError(
+                    f"field {key} cannot have all null values for type inference."
+                )
+
             if isinstance(value, Value):
                 proto_value = value
             else:
-                # Infer the specific type for this row
-                current_dtype = python_type_to_feast_value_type(name=key, value=value)
-
-                if key not in entity_type_map:
-                    entity_type_map[key] = current_dtype
-                else:
-                    if current_dtype != entity_type_map[key]:
-                        raise TypeError(
-                            f"Input entity {key} has mixed types, {current_dtype} and {entity_type_map[key]}. That is not allowed. "
-                        )
-                proto_value = _python_value_to_proto_value(current_dtype, value)
+                proto_value = _python_value_to_proto_value(entity_type_map[key], value)
             fields[key] = proto_value
         entity_row_list.append(GetOnlineFeaturesRequestV2.EntityRow(fields=fields))
     return entity_row_list
--- a/sdk/python/feast/online_response.py
+++ b/sdk/python/feast/online_response.py
@@ -24,6 +24,7 @@ from feast.protos.feast.serving.ServingService_pb2 import (
 )
 from feast.protos.feast.types.Value_pb2 import Value as Value
 from feast.type_map import (
+    _proto_str_to_value_type,
     _python_value_to_proto_value,
     feast_value_type_to_python_type,
     python_values_to_feast_value_type,
@@ -100,12 +101,24 @@ def _infer_online_entity_rows(
     # Flatten keys-value dicts into lists for type inference
     for entity in entity_rows_dicts:
         for key, value in entity.items():
-            if not isinstance(value, Value):
+            if isinstance(value, Value):
+                entity_type_map[key] = _proto_str_to_value_type(
+                    str(value.WhichOneof("val"))
+                )
+            else:
                 entity_value_map[key].append(value)
 
     # Loop over all entities to infer dtype first in case of empty lists or nulls
     for key, values in entity_value_map.items():
-        entity_type_map[key] = python_values_to_feast_value_type(key, values)
+        inferred_type = python_values_to_feast_value_type(key, values)
+
+        # If any ProtoValues were present their types must match the inferred type
+        if key in entity_type_map and entity_type_map.get(key) != inferred_type:
+            raise TypeError(
+                f"Input entity {key} has mixed types, {entity_type_map.get(key)} and {inferred_type}. That is not allowed."
+            )
+
+        entity_type_map[key] = inferred_type
 
     for entity in entity_rows_dicts:
         fields = {}
--- a/sdk/python/feast/online_response.py
+++ b/sdk/python/feast/online_response.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from collections import defaultdict
-from typing import Any, Dict, List, cast
+from typing import Any, Dict, List, Optional, cast
 
 import pandas as pd
 
@@ -29,6 +29,7 @@ from feast.type_map import (
     feast_value_type_to_python_type,
     python_values_to_feast_value_type,
 )
+from feast.value_type import ValueType
 
 
 class OnlineResponse:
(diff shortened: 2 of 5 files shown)


judahrand:
Fixes an issue mentioned in #1640 but not the whole issue as this does not deal with int lists.

woop:
Thanks @judahrand, should we perhaps add a test to ensure that the change you've introduced works?

judahrand:
Hey @woop. Took me a while to get to this but it is easier now that the tests have been reworked a tad. Added the test you asked for. Worth revisiting?

judahrand:
Wanted to go back and get the integration tests to run with the code from `master` + my test to make sure this is actually still broken. Also to see if those tests that failed a second ago were just flakey.

Looks like I need approval to run them again...

judahrand:
> Wanted to go back and get the integration tests to run with the code from `master` + my test to make sure this is actually still broken. Also to see if those tests that failed a second ago were just flakey.
> 
> Looks like I need approval to run them again... @achals

I think you might have to unlabel and re-label `ok-to-test`?

judahrand:
Hmmm... Maybe I'm wrong... Any idea why the integration tests are skipping?

achals:
> Hmmm... Maybe I'm wrong... Any idea why the integration tests are skipping?

The tests are running - https://github.com/feast-dev/feast/actions/runs/1246998876
There's a lot of skipped actions you may see if you go to the `Checks` tab in the PR, but one of them is usually going to be running. Not sure why there's so many skipped ones but I've seen that before - it's something we'll try to fix.

judahrand:
> > Hmmm... Maybe I'm wrong... Any idea why the integration tests are skipping?
> 
> The tests are running - https://github.com/feast-dev/feast/actions/runs/1246998876
> There's a lot of skipped actions you may see if you go to the `Checks` tab in the PR, but one of them is usually going to be running. Not sure why there's so many skipped ones but I've seen that before - it's something we'll try to fix.

Ah, okay.

judahrand:
Okay, good. Tests do indeed fail with my new test but without fix.
https://github.com/feast-dev/feast/runs/3636658440?check_suite_focus=true

judahrand:
And passes after changes:
https://github.com/feast-dev/feast/runs/3637027115?check_suite_focus=true

judahrand:
So ready for some review 😁 @achals 

I'd add that my argument for this being right this that a `None`/`null` value is definitely different to an empty list.

judahrand:
Noticed #1846 and have added a test which I think should catch what @karlhigley is trying to fix.

judahrand:
I've determined that this issue also exists in `python_type_to_feast_value_type`. It is less clear how to deal with the issue here... 

https://github.com/feast-dev/feast/runs/3641985126?check_suite_focus=true

adchia:
Re `python_type_to_feast_value_type`:

What are the downsides to relaxing the null constraint on inference in this method? It seems relatively normal to have some feature values be null / empty e.g. a user history feature for a new user.

judahrand:
> Re `python_type_to_feast_value_type`:
> 
> What are the downsides to relaxing the null constraint on inference in this method? It seems relatively normal to have some feature values be null / empty e.g. a user history feature for a new user.

What would you suggest it should return? `ValueType.UNKNOWN`? I'm not sure what the downstream consequences of that might be?

adchia:
I think `ValueType.UNKNOWN` would probably throw an error since we need some type, and presumably it got here in many cases because there was no supplied type (e.g. "inference" code)

[120, null, 1, 3, 4], I'd expect that to still infer to an int list that we'd want to support. 

There's likely downstream effects of asserting that though, though I think the only real store we "support" right now is BQ and materialization crashes there (https://github.com/feast-dev/feast/issues/1839)

judahrand:
> I think `ValueType.UNKNOWN` would probably throw an error since we need some type, and presumably it got here in many cases because there was no supplied type (e.g. "inference" code)

Agreed.

> [120, null, 1, 3, 4], I'd expect that to still infer to an int list that we'd want to support.
> 
> There's likely downstream effects of asserting that though, though I think the only real store we "support" right now is BQ and materialization crashes there (#1839)

Yup, that's the approach I'm trying here: infer type from all the data provided rather than infer a type for each piece of data alone. This'll still fail in the case where all the data is empty lists (or all `null`) which isn't great in my opinion given that there is an underlying schema from the data source (at least for the DB sources).

Should we not be inferring types from the schema rather than the data really?

judahrand:
Tests pass!

sdk/python/feast/feature.py
```diff
@@ -39,6 +39,8 @@ def __init__(
         self._name = name
         if not isinstance(dtype, ValueType):
             raise ValueError("dtype is not a valid ValueType")
+        if dtype is ValueType.UNKNOWN:
+            raise ValueError(f"dtype cannont be {dtype}")
```
adchia, line 43: nit: typo

sdk/python/feast/online_response.py
```diff
@@ -94,25 +95,30 @@ def _infer_online_entity_rows(
...
+
+    # Loop over all entities to infer dtype first in case of empty lists or nulls
+    for key, values in entity_value_map.items():
+        entity_type_map[key] = python_values_to_feast_value_type(key, values)
 
     for entity in entity_rows_dicts:
         fields = {}
         for key, value in entity.items():
-            # Allow for feast.types.Value
+            if key not in entity_type_map:
```
adchia, line 113: this would break if the user added Value items in the entity rows right?

sdk/python/feast/type_map.py
```diff
@@ -180,6 +183,28 @@ def python_type_to_feast_value_type(
...
+    inferred_dtype = ValueType.UNKNOWN
+    for row in values:
+        current_dtype = python_type_to_feast_value_type(
+            name, value=row, recurse=recurse
+        )
+
+        if inferred_dtype is ValueType.UNKNOWN:
+            inferred_dtype = current_dtype
+        else:
+            if current_dtype != inferred_dtype:
```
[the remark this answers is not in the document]
adchia: Doesn't matter too much right now since you call it from the inference of entity rows, which shouldn't have null values, but for feature inference from datasources we'd probably want a different treatment.

sdk/python/tests/data/data_creator.py
```diff
@@ -57,8 +60,10 @@ def get_feature_values_for_dtype(dtype: str, is_list: bool) -> List:
         "bool": [True, None, False, True, False],
     }
     non_list_val = dtype_map[dtype]
-    # Duplicate the value once if this is a list
     if is_list:
+        if is_empty:
```
adchia, line 64: can we add a test case of completely empty and asserting the error?

sdk/python/tests/integration/registration/test_universal_types.py
```diff
@@ -193,53 +209,61 @@ def test_feature_get_online_features_types_match(online_types_test_fixtures):
...
     feature_dtype: str, historical_features_df: pd.DataFrame
 ):
     print("Asserting historical feature types")
     feature_dtype_to_expected_historical_feature_dtype = {
-        "int32": "int64",
-        "int64": "int64",
-        "float": "float64",
-        "string": {"string", "object"},
-        "bool": {"bool", "object"},
+        "int32": (pd.api.types.is_integer_dtype,),
```
adchia, line 258: this is slick

sdk/python/feast/type_map.py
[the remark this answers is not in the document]
judahrand: Huh, you're right that's an interesting edge case I hadn't considered and the behaviour here is a little odd actually because the order matters.

* `[None, None, None, 1]` would correctly be inferred to `ValueType.INT64_LIST`
* `[None, None, 1, None]` would end up being `ValueType.UNKNOWN`

Something to consider 😕 But to me this seems like a very odd edge case. I can't think of that many use cases for a mix of `null` and `not null` values inside the list.

Can you think of an example?

sdk/python/tests/data/data_creator.py
judahrand: The way the tests are set up that's a little tricky 😉 

Could add some unit tests to catch that?

sdk/python/tests/integration/registration/test_universal_types.py
```diff
@@ -43,14 +43,19 @@ def populate_test_configs(offline: bool):
...
+                            entity_type=entity_type,
+                            feature_dtype=feature_dtype,
+                            feature_is_list=feature_is_list,
+                            list_is_empty=list_is_empty,
+                            test_repo_config=test_repo_config,
+                        )
                     )
-                )
+                    # For non list features `list_is_empty` does nothing
+                    if feature_is_list is False:
```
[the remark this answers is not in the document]
judahrand: No this is correct. I still want one just not the second one from the loop. Was a bit of an ugly hack in hindsight but was the fewest lines of code 😛

sdk/python/feast/online_response.py
judahrand: I think I've addressed this now - let me know what you think.

sdk/python/feast/type_map.py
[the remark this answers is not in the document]
judahrand: I think I've addressed this now - let me know what you think.

sdk/python/tests/integration/registration/test_universal_types.py
```diff
@@ -293,6 +329,7 @@ def assert_expected_arrow_types(
             ) in [
                 f"struct<list: list<item: struct<item: {arrow_type}>> not null>",
                 f"struct<list: list<item: struct<item: {arrow_type}>>>",
+                "struct<list: list<item: null>>",
```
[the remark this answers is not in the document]
judahrand: This should go away when BQ lists are sorted out. They're currently quite broken anyway.

sdk/python/feast/online_response.py
judahrand: I think I've addressed this now - let me know what you think.

sdk/python/feast/feature.py
judahrand: Fixed

sdk/python/tests/integration/registration/test_universal_types.py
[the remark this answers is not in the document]
adchia: haha can you instead do this before adding the config? It almost looks like a bug, and should be functionally equivalent.

sdk/python/feast/online_response.py
adchia: Not sure if I'm missing something, but I think it's still an issue?

If the user passes in as entity rows something like

`[{"driver_id:" : Value}]`, then this would result in an empty entity_type_map and fail, when before it would have gone to the 

```
if isinstance(value, Value):
    proto_value = value
```
line

adchia: ah sorry was looking at the wrong commit diff.  this does seem like it'd fix it.

sdk/python/tests/integration/registration/test_universal_types.py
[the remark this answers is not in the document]
judahrand: No, because then no config will get added 😕. For `feature_is_list is False` we still need to add one config it just doesn't matter is `list_is_empty` is `True` or `False`.

sdk/python/feast/online_response.py
```diff
@@ -94,25 +96,42 @@ def _infer_online_entity_rows(
     entity_rows_dicts = cast(List[Dict[str, Any]], entity_rows)
     entity_row_list = []
     entity_type_map = dict()
+    entity_value_map = defaultdict(list)
+
+    # Flatten keys-value dicts into lists for type inference
+    for entity in entity_rows_dicts:
+        for key, value in entity.items():
+            if isinstance(value, Value):
+                entity_type_map[key] = _proto_str_to_value_type(
```
adchia, line 105: mind leaving a comment here on what's going on?

seems easy to miss that we use the last defined value as the type so if there are inconsistencies of values in entity rows it wouldn't be caught. But this is the old behavior too

sdk/python/feast/online_response.py
```diff
@@ -94,25 +96,42 @@ def _infer_online_entity_rows(
     entity_rows_dicts = cast(List[Dict[str, Any]], entity_rows)
     entity_row_list = []
     entity_type_map = dict()
+    entity_value_map = defaultdict(list)
```
adchia, line 99: might be worth renaming entity value since it actually doesn't include the Value class confusingly haha.

maybe something like entity_python_value_map?

sdk/python/tests/integration/registration/test_universal_types.py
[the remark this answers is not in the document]
judahrand: Could make it 
```
if feature_is_list is False and list_is_empty is True:
    continue
```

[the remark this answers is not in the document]
judahrand: Done.

sdk/python/tests/integration/registration/test_universal_types.py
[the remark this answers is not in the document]
judahrand: Done.

sdk/python/feast/online_response.py
judahrand: That feels like it's a different PR? I haven't actually changed any behaviour here and it wasn't tested before. Not saying it doesn't need a test just not sure this PR should keep growing and growing.

sdk/python/feast/online_response.py
judahrand: I just fixed the underlying issue 😉 (and added a comment).

sdk/python/feast/online_response.py
judahrand: Done.

sdk/python/feast/online_response.py
adchia: sg. we can leave this for later

sdk/python/tests/data/data_creator.py
adchia: we can push this for later if you leave a TODO so we don't forget.

mind renaming `is_empty` to `has_empty_list` to be more precise?

judahrand: Done.

adchia (approved):
/lgtm

PR #1712, opened 2021-07-14
Avoid skewed join between entity_df & feature views

MattDelac:

The problem is that if we ask for historical features coming from multiple entities that have a 1:many relationship between them, then we encounter skewed join

That's basically how the JOIN is performed with the current template
Imagine that driver_id=1 contains millions of rides
image

And that's what this PR is proposing
image

Let's have a look about the statistics of the SQL template on our use case (4 FeatureViews, 2 entities, _entity_dataframe_ containing 100M rows)

**SQL template currently in production**
I cancelled the query as it was still running after 25min
> 9:14 AM Query has been running for 25 min

**SQL template of this PR**
**Elapsed time** 2 min 34 sec
**Slot time consumed** 1 day 21 hr
**Bytes shuffled** 3.27 TB
**Bytes spilled to disk** 0 B

_Note: On our full entity_dataframe (3B rows) the current SQL template was still running after 45 min while the SQL template of this PR finished after 15min_

Fixes None

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -521,18 +521,29 @@ SINGLE_FEATURE_VIEW_POINT_IN_TIME_JOIN = """
  all the logic as the field to GROUP BY the data
 */
 WITH entity_dataframe AS (
-    SELECT
-        *,
-        CONCAT(
-            {% for entity_key in unique_entity_keys %}
-                CAST({{entity_key}} AS STRING),
-            {% endfor %}
-            CAST({{entity_df_event_timestamp_col}} AS STRING)
-        ) AS entity_row_unique_id
+    SELECT *,
+        {{entity_df_event_timestamp_col}} AS entity_timestamp,
+        {% for featureview in featureviews %}
+            CONCAT(
+                {% for entity in featureview.entities %}
+                    CAST({{entity}} AS STRING),
+                {% endfor %}
+                CAST({{entity_df_event_timestamp_col}} AS STRING)
+            ) AS {{featureview.name}}__entity_row_unique_id,
+        {% endfor %}
     FROM {{ left_table_query_string }}
 ),
 
 {% for featureview in featureviews %}
+
+{{ featureview.name }}__entity_dataframe AS (
+    SELECT
+        {{ featureview.entities | join(', ')}},
+        entity_timestamp,
+        {{featureview.name}}__entity_row_unique_id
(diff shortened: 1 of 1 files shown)


sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -521,19 +521,30 @@ def _get_bigquery_client(project: Optional[str] = None):
...
+                {% endfor %}
+                CAST({{entity_df_event_timestamp_col}} AS STRING)
+            ) AS {{featureview.name}}__entity_row_unique_id,
+        {% endfor %}
     FROM {{ left_table_query_string }}
 ),
 
 {% for featureview in featureviews %}
 
+{{ featureview.name }}__entity_dataframe AS (
```
MattDelac, line 539: Therefore if there are duplicates for a given entity on a given timestamp, this will be drop here to avoid skewed join

woop:
What isn't clear to me is what the before and after of this test is. What is the problem that we are seeing and how do we know we are solving it? I realize it has to do with one-to-many relationships. Can we add a test that uses a 1:many relationship and shows how this test actually fixes the response? Or could we just extend our existing historical retrieval to have one of these relationships?

MattDelac:
> What is the problem that we are seeing and how do we know we are solving it?

That's an optimization problem. So beside running some benchmark on my side and prove you that this new template is better to scale, I don't have an idea of a good unit test for it.

> Can we add a test that uses a 1:many relationship and shows how this test actually fixes the response? Or could we just extend our existing historical retrieval to have one of these relationships?

As I was saying, this is an optimization problem. We can extend the current test if we think that the coverage is not enough. A dedicated test for that does not seem like a good option IMO

woop:
> > What is the problem that we are seeing and how do we know we are solving it?
> 
> That's an optimization problem. So beside running some benchmark on my side and prove you that this new template is better to scale, I don't have an idea of a good unit test for it.
> 
> > Can we add a test that uses a 1:many relationship and shows how this test actually fixes the response? Or could we just extend our existing historical retrieval to have one of these relationships?
> 
> As I was saying, this is an optimization problem. We can extend the current test if we think that the coverage is not enough. A dedicated test for that does not seem like a good option IMO
> 
> Also I am going to spend time in benchmarking the 2 templates on our use case and will publish as many detail as possible in this PR

Thanks. As long as it's purely an optimization change then I don't see a need for a new test. Let me know when/if you feel comfortable merging after your analysis.

woop (approved):

PR #1713, opened 2021-07-15
Add custom data sources

achals:

After allowing for custom online stores and offline stores, we need to update the DataSource API so that implementors can create custom. DataSources to be used in their feature repo, and provide a mechanism through which these DataSources can be serialized/deserialized from the registry.

This is done by introducing a super simple `CustomSourceOptions` message that contains information about the class used for the source, and any serialized contents. Implementors are responsible for defining a `to_proto` and `from_proto` method and they can serialize the contents into this class as they need.

Fixes #

```release-note
Allow for custom Data Sources to be defined.
```


sdk/python/feast/data_source.py
```diff
@@ -453,77 +306,49 @@ def date_partition_column(self, date_partition_column):
...
     @staticmethod
-    def from_proto(data_source):
+    @abstractmethod
+    def from_proto(data_source: DataSourceProto):
         """
         Convert data source config in FeatureTable spec to a DataSource class object.
         """
 
+        if data_source.data_source_class_type:
+            print(f"Data Source class type: {data_source.data_source_class_type}")
```
woop, line 316: Should we start adopting Python logging instead of print?

sdk/python/feast/repo_config.py
```diff
@@ -195,6 +195,12 @@ def __repr__(self) -> str:
         )
 
 
+def get_data_source_class_from_type(offline_store_type: str):
```
woop, line 198: Why is this method in `repo_config.py`?

Also, what happens when a data source can be used in two offline stores?

sdk/python/feast/data_source.py
achals: Oops

sdk/python/feast/repo_config.py
achals: Can you give an example of this? Shouldn't the source be tied to the offline store?

woop: You could in theory have an S3/parquet source that is readable by Redshift, Athena, or a File offline store.

woop: In theory we could push this problem down the road and generalize later, but I can see it potentially happening.

achals: Hmmm. Actually it turns out that the way i've implemented this, it doesn't have a hard dependency on the offline store type - it only needs the fully qualified class name for the data source class which we stuff into the  proto, and use to import the class dynamically

woop (approved):

woop:
Just waiting for merge conflict to be resolved. Otherwise lgtm.

PR #1716, opened 2021-07-16
Append ns time and random integer to redshift test tables

achals:

As found earlier, integration tests fail because of race conditions which cause the tests to create/rely on tables with the same name. We started seeing this again with Redshift as well:

https://github.com/feast-dev/feast/runs/3089135542?check_suite_focus=true

Applying the same mitigation we used then, which was to use nanosecond precision, and add an additional random integer for high entropy.

Fixes #

```release-note
NONE
```


tsotnet (approved):
/lgtm

PR #1718, opened 2021-07-17
Registry teardown

felixwang9817:
**What this PR does / why we need it**: As of right now, registries do not get cleaned up during teardown. This is especially bad for GCS and S3 registries. This PR ensures that registries get cleaned up during teardown. 

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/registry.py
+++ b/sdk/python/feast/registry.py
@@ -21,8 +21,6 @@ from tempfile import TemporaryFile
 from typing import List, Optional
 from urllib.parse import urlparse
 
-from google.cloud.exceptions import NotFound
-
 from feast.entity import Entity
 from feast.errors import (
     EntityNotFoundException,
@@ -558,6 +556,8 @@ class GCSRegistryStore(RegistryStore):
         self._write_registry(registry_proto)
 
     def teardown(self):
+        from google.cloud.exceptions import NotFound
+
         gs_bucket = self.gcs_client.get_bucket(self._bucket)
         try:
             gs_bucket.delete_blob(self._blob)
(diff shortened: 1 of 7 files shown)


sdk/python/feast/registry.py
```diff
@@ -610,6 +641,10 @@ def update_registry_proto(self, registry_proto: RegistryProto):
         self._write_registry(registry_proto)
         return
 
+    def teardown(self):
+        self.s3_client.Object(self._bucket, self._key).delete()
+        return
```
achals, line 646: nit, returns aren't necessary for methods that don't return anything

achals (approved):

sdk/python/feast/registry.py
```diff
@@ -21,6 +21,8 @@
 from typing import List, Optional
 from urllib.parse import urlparse
 
+from google.cloud.exceptions import NotFound
```
woop, line 24: Should this dependency be eagerly loaded? What about teams that don't use GCP?

felixwang9817: Good catch - I'll place this import statement in the method.

sdk/python/feast/registry.py
```diff
@@ -399,6 +395,10 @@ def refresh(self):
         """Refreshes the state of the registry cache by fetching the registry state from the remote registry store."""
         self._get_registry_proto(allow_cache=False)
 
+    def teardown(self):
+        """Tears down all local and cloud resources."""
```
woop, line 399: Why does the registry tear down all local and cloud resources? Seems confusing. Wouldn't we only be removing all registry specific resources?

felixwang9817: Hm, `deployed` doesn't sound right to me - it sounds like there's a server involved. What about just `Tears down (removes) the registry.`?

woop: sgtm

woop: I just want to be clear that we aren't touching any real infrastructure. We are dealing with the registry here.

achals (approved):
/lgtm

PR #1719, opened 2021-07-17
MS SQL / Azure Synapse Analytics offline store

abrookins:

Adds an MS SQL-compatible offline store. This allows users to use MS SQL or Azure Synapse Analytics as the historical store.

**NOTE**: This PR needs more work, like tests and docs. I'm opening it now to get feedback on our approach. The approach was to copy the existing BigQuery offline store and adapt it to SQL Server / Azure Synapse Analytics.

Yes.

```release-note
Added MSSQL-compatible offline store.
```


protos/feast/core/DataSource.proto
```diff
@@ -34,6 +34,7 @@ message DataSource {
     STREAM_KAFKA = 3;
     STREAM_KINESIS = 4;
     BATCH_REDSHIFT = 5;
+    BATCH_SQLSERVER = 6;
```
woop, line 37: Should we make this `MSSQLSERVER` (and elsewhere)? Their product is a bit poorly named :)

protos/feast/core/DataSource.proto
```diff
@@ -72,6 +73,20 @@ message DataSource {
...
   }
 
+  // Defines options for DataSource that sources features from a MSSQL Query
+  message SqlServerOptions {
+    // SQLAlchemy-formatted string to SQL Server instance
+    string connection_str = 1;
+
+    // SQL query that returns a table containing feature data. Must contain an event_timestamp column, and respective
+    // entity columns
+    string query = 2;
```
woop, line 83: We are removing `query` as an option going forward.

sdk/python/feast/data_source.py
```diff
@@ -365,6 +367,100 @@ def to_proto(self) -> DataSourceProto.KinesisOptions:
         return kinesis_options_proto
 
 
+class SqlServerOptions:
```
woop, line 370: Data sources are now here https://github.com/feast-dev/feast/tree/master/sdk/python/feast/infra/offline_stores

woop:
@abrookins this is awesome. I'd love to figure out whether we can host this connector externally (perhaps as part of https://github.com/RedisLabs)? We can still pull in the code and run tests on it as part of our CI, but as a team we probably can't own this connector yet. The ideal case for us would be for the connector to start out as an external package that users can install and reference from their `feature_store.yaml` (we already have plugin support for external connectors). 

Once we see adoption (and stability) improving for the connector, we can look to make it part of the code base.

sdk/python/feast/data_source.py
woop: Our preference would be for you to try and implement the offline store and data source externally (separate Python package) and use custom data sources for configuration (as opposed to making changes to the core protos). Would that work? https://github.com/feast-dev/feast/pull/1713

sdk/python/feast/infra/offline_stores/mssql.py
```diff
@@ -0,0 +1,494 @@
...
+    query = template.render(template_context)
+    return query
+
+
+# TODO: Optimizations
+#   * Use NEWID() instead of ROW_NUMBER(), or join on entity columns directly
+#   * Precompute ROW_NUMBER() so that it doesn't have to be recomputed for every query on entity_dataframe
+#   * Create temporary tables instead of keeping all tables in memory
+
+SINGLE_FEATURE_VIEW_POINT_IN_TIME_JOIN = """
```
woop, line 351: By the way, this is a multiple feature view PIT join. The variable had improper naming previously.

abrookins:
> @abrookins this is awesome. I'd love to figure out whether we can host this connector externally (perhaps as part of https://github.com/RedisLabs)? We can still pull in the code and run tests on it as part of our CI, but as a team we probably can't own this connector yet. The ideal case for us would be for the connector to start out as an external package that users can install and reference from their `feature_store.yaml` (we already have plugin support for external connectors).
> 
> Once we see adoption (and stability) improving for the connector, we can look to make it part of the code base.

Thanks for the review, @woop! I'm finally coming back to this PR to iron out the details you brought up, and then I'll work on packaging it as an external connector.

PR #1723, opened 2021-07-19
Refactor data source classes to fix import issues

achals:

By keeping the Source classes in the same files as  the offline store implementations, we can't import the sources without the corresponding libraries installed. This is a terrible experience. This PR moves the classes to a different file, so they can be imported for just the declarative use cases and not require `boto` or `gcloud` stuff to be installed.

Fixes #

```release-note
NONE
```


woop:
Why do we need `__init__` everywhere?

achals:
> Why do we need `__init__` in sdk?

we don't, intellij added it. removing

sdk/python/usage_tests/test_usage.py
```diff
@@ -187,7 +187,7 @@ def test_exception_usage_off():
     assert rows.total_rows == 0
 
 
-@retry(wait=wait_exponential(multiplier=1, min=1, max=10), stop=stop_after_attempt(5))
+@retry(wait=wait_exponential(multiplier=1, min=1, max=10), stop=stop_after_attempt(7))
```
woop, line 190: Why was this change made?

achals: i've been noticing usage tests are starting to flake because the UUIDs aren't in BQ in the current time constraint. Should I split this out in a different PR/commit?

woop: so extending the execution time resolves the problem permanently? I just want to make sure that we're not glossing over actual flaky tests (where the test is broken) vs a test where the timing is just too tight.

achals: I don't know if the fix is permanent - we're relying on cloud functions logging showing up in cloud logs/bigquery but by gcps docs there's a delay and I couldn't find an sla for the delay. It's possible that this timeout is sufficient for now but not the future.

woop: ok

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -26,10 +25,10 @@
     DEFAULT_ENTITY_DF_EVENT_TIMESTAMP_COL,
     _get_requested_feature_views_to_features_dict,
 )
-from feast.protos.feast.core.DataSource_pb2 import DataSource as DataSourceProto
 from feast.registry import Registry
 from feast.repo_config import FeastConfigBaseModel, RepoConfig
-from feast.value_type import ValueType
+
+from .bigquery_source import BigQuerySource
```
woop, line 31: This import is suspect. Should it be `feast.bigquery_source`?

achals: No, it's a relative import - should work fine (tests are happy too)

woop: I can't remember where this has caused problems for me, but I distinctly remember always having to make all imports absolute. Somewhere I ran into problems with relative imports. I don't think there are any relative imports in the code base, except the __init__.py files.

I'm fine leaving it as is and seeing if a problem pops up though.

woop (approved):

PR #1729, opened 2021-07-22
Change internal references from input to batch_source

felixwang9817:

**What this PR does / why we need it**: FeatureViews have always required a batch data source as input; this argument is currently named `input`. We recently introduced stream sources [here](https://github.com/feast-dev/feast/pull/1664), so now we have three arguments: a required `input`, an optional `batch_source`, and an optional `stream_source`. We should remove `input` as an argument, make `batch_source` required, and leave `stream_source` optional; this is a breaking change, so we prepare for it by adding a deprecation warning. We also change internal references and docs to use `batch_source` instead of `input`, to point people in the right direction.

Fixes #

```release-note
There is now a deprecation warning in FeatureView.
```

### DIFF
--- a/sdk/python/feast/feature_view.py
+++ b/sdk/python/feast/feature_view.py
@@ -76,7 +76,7 @@ class FeatureView:
         warnings.warn(
             (
                 "The argument 'input' is being deprecated. Please use 'batch_source' "
-                "instead. Feast 0.11.3 and onwards will not support the argument 'input'."
+                "instead. Feast 0.12 and onwards will not support the argument 'input'."
             ),
             DeprecationWarning,
         )
--- a/sdk/python/feast/feature_view.py
+++ b/sdk/python/feast/feature_view.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import re
+import warnings
 from datetime import datetime, timedelta
 from typing import Dict, List, Optional, Tuple, Union
 
@@ -37,6 +38,8 @@ from feast.protos.feast.core.FeatureView_pb2 import (
 from feast.repo_config import RepoConfig
 from feast.usage import log_exceptions
 from feast.value_type import ValueType
+
+warnings.simplefilter("once", DeprecationWarning)
 
 
 class FeatureView:
@@ -70,6 +73,14 @@ class FeatureView:
         tags: Optional[Dict[str, str]] = None,
         online: bool = True,
     ):
+        warnings.warn(
+            (
+                "The argument 'input' is being deprecated. Please use 'batch_source' "
+                "instead. Feast 0.11.3 and onwards will not support the argument 'input'."
+            ),
+            DeprecationWarning,
+        )
+
         _input = input or batch_source
         assert _input is not None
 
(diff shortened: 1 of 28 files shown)


sdk/python/feast/feature_view.py
```diff
@@ -63,24 +62,23 @@ def __init__(
         name: str,
         entities: List[str],
         ttl: Optional[Union[Duration, timedelta]],
-        input: DataSource,
-        batch_source: Optional[DataSource] = None,
```
achals, line 67: This is an API breaking change - we probably want to raise deprecation warnings or something for a release before killing it since it will break existing users immediately on upgrading feast. wdyt?

achals (changes requested):

sdk/python/feast/feature_view.py
```diff
@@ -63,13 +66,21 @@ def __init__(
...
         batch_source: Optional[DataSource] = None,
         stream_source: Optional[DataSource] = None,
         features: List[Feature] = None,
         tags: Optional[Dict[str, str]] = None,
         online: bool = True,
     ):
+        warnings.warn(
+            (
+                "The argument 'input' is being deprecated. Please use 'batch_source' "
+                "instead. Feast 0.11.3 and onwards will not support the argument 'input'."
```
woop, line 79: Should we make this 0.12+? Our policy is to deprecate over minor versions. Plus, we are releasing 0.12 soon.

sdk/python/feast/feature_view.py
```diff
@@ -63,13 +66,21 @@ def __init__(
...
         batch_source: Optional[DataSource] = None,
         stream_source: Optional[DataSource] = None,
         features: List[Feature] = None,
         tags: Optional[Dict[str, str]] = None,
         online: bool = True,
     ):
+        warnings.warn(
+            (
+                "The argument 'input' is being deprecated. Please use 'batch_source' "
+                "instead. Feast 0.12 and onwards will not support the argument 'input'."
```
achals, line 79: This should be feast 0.13 (since 0.12 is the release that will introduce the deprecation warning).

felixwang9817: Ah, I understand now. I'll make the change.

achals (commented):
/lgtm

achals (approved):

sdk/python/feast/feature_view.py
```diff
@@ -63,13 +66,21 @@ def __init__(
...
         ttl: Optional[Union[Duration, timedelta]],
-        input: DataSource,
+        input: Optional[DataSource] = None,
         batch_source: Optional[DataSource] = None,
         stream_source: Optional[DataSource] = None,
         features: List[Feature] = None,
         tags: Optional[Dict[str, str]] = None,
         online: bool = True,
     ):
+        warnings.warn(
```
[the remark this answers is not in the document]
felixwang9817: @tedhtchang That's a good point, I think ideally we would have done that. However, given that (a) `input` was a required parameter before this change, so almost all users will continue to use `input`, and (b) releases are fairly frequent, I think we can leave this change as is.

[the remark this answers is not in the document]
felixwang9817: @tedhtchang btw, I managed to insert this change into a separate PR (#1746). Thanks for pointing this one out!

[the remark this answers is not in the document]
tedhtchang: @felixwang9817 Thanks. Looks good.

PR #1730, opened 2021-07-23
Clean up uploaded entities in Redshift offline store

tsotnet:

**What this PR does / why we need it**: Currently uploaded Redshift entities are not cleaned up. They're left as random tables in Redshift. With this PR, the entities will not be uploaded during get_historical_features, but when you call to_X (e.g. to_df) on the returned RetrievalJob. At the same time, we'll clean up the uploaded entity table at the end of to_X call.

Fixes #

```release-note
NONE
```


sdk/python/feast/infra/offline_stores/redshift.py
```diff
@@ -113,40 +114,53 @@ def get_historical_features(
         )
         s3_resource = aws_utils.get_s3_resource(config.offline_store.region)
 
-        table_name = offline_utils.get_temp_entity_table_name()
+        @contextlib.contextmanager
```
woop, line 117: Just to play devil's advocate: Is there anything preventing us from moving the logic within the query generator into the RetrievalJob object itself so that we don't have to define it here (and pass a callable into the RetrievalJob). Basically, could we just configure the RetrievalJob at this stage, and have this logic as a private method on the RetrievalJob somewhere?

tsotnet: I initially did that, before realizing that we use RedshiftRetrievalJob in cases where we don't want to upload entity_df (e.g. in pull_latest_from_table_or_query). And this seemed to be the simplest common interface for passing either string query or somehow being able to upload entity_df and cleanup inside RedshiftRetrievalJob

woop: Not sure I really follow how the string/upload distinction maps to this design. Wouldn't it be possible to pass in the necessary arguments to `RetrievalJob` from `get_historical_features`? I think 99% of the code would be the same, except we wouldn't need to pass in a function as an argument (and instead use a method). What do you think?

tsotnet: We could pass all necessary arguments to RedshiftRetrievalJob, but there are quite a few of them: `entity_df`, `project`, `feature_views`, `registry`, `feature_refs`, `full_feature_names`.

It doesn't make sense to pass these arguments when RedshiftRetrievalJob is also used in `pull_latest_from_table_or_query`. If you look at the usage of RedshiftRetrievalJob in that method, you'll see that we simply need to execute a SQL query and there's no need to pass any of these parameters.

Having `query` be string or string generator is a relatively simple interface in my opinion.

achals (approved):
/lgtm

PR #1731, opened 2021-07-23
AWS Template improvements (input prompt for configs, default to Redshift)

tsotnet:
AWS Template improvements (input prompt for configs, default to Redshift)

**What this PR does / why we need it**: Currently AWS template uses dynamodb online store and file offline store. We're now changing the offline store to Redshift. However, Redshift needs to be configured by the user to make things work (e.g. cluster_id, database name, etc.). We don't have a mechanism for that. This PR adds [python-inquirer](https://github.com/magmax/python-inquirer) library for user prompt during aws template bootstrap, which is used for configuring the feature_store.yaml. Here's the new flow in action:

image

Fixes #

```release-note
Allow configuring online & offline stores during AWS template, default to Redshift offline store
```


sdk/python/feast/templates/aws/bootstrap.py
```diff
@@ -1,3 +1,8 @@
+import inquirer
```
woop, line 1: Out of curiousity, is there anything `inquirer` gives us that `Click` doesn't?

woop (approved):

sdk/python/setup.py
```diff
@@ -57,6 +57,7 @@
     "tenacity>=7.*",
     "toml==0.10.*",
     "tqdm==4.*",
+    "inquirer==2.7.0",
```
achals, line 60: Any reason we're pinning to this version and not a range?

sdk/python/feast/templates/aws/bootstrap.py
achals: +1 to fewer deps if possible.

achals (approved):
/lgtm

PR #1734, opened 2021-07-23
Add the foundation of the universal feature repo and a test that uses it

achals:

This PR that implements some of the basis of the universal feature repo  as described in the the [RFC](https://docs.google.com/document/d/1QyaDFZ7yM1SclgFbsho1MjWZvjrkTmtERDpnr1jvgUs/edit#)

It adds a special decorator, `parametrize_e2e_test`, that makes it easy to tag tests to that should be executed against parameterized feature stores that are constructed by the testing framework. This means that users can add e2e tests and have them run against all supported combinations with a single decorator.

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/tests/data/data_creator.py
+++ b/sdk/python/tests/data/data_creator.py
@@ -1,0 +1,27 @@
+from datetime import datetime, timedelta
+
+import pandas as pd
+from pytz import timezone, utc
+
+
+def create_dataset() -> pd.DataFrame:
+    now = datetime.utcnow()
+    ts = pd.Timestamp(now).round("ms")
+    data = {
+        "id": [1, 2, 1, 3, 3],
+        "value": [0.1, None, 0.3, 4, 5],
+        "ts_1": [
+            ts - timedelta(hours=4),
+            ts,
+            ts - timedelta(hours=3),
+            # Use different time zones to test tz-naive -> tz-aware conversion
+            (ts - timedelta(hours=4))
+            .replace(tzinfo=utc)
+            .astimezone(tz=timezone("Europe/Berlin")),
+            (ts - timedelta(hours=1))
+            .replace(tzinfo=utc)
+            .astimezone(tz=timezone("US/Pacific")),
+        ],
+        "created_ts": [ts, ts, ts, ts, ts],
+    }
+    return pd.DataFrame.from_dict(data)
--- a/sdk/python/tests/integration/e2e/test_universal_e2e.py
+++ b/sdk/python/tests/integration/e2e/test_universal_e2e.py
@@ -1,0 +1,134 @@
+import math
+from datetime import datetime, timedelta
+from typing import Optional
+
+import pandas as pd
+import pytest
+from pytz import utc
+
+from feast import FeatureStore, FeatureView
+from tests.integration.feature_repos.universal.data_sources.bigquery import (
+    BigQueryDataSourceCreator,
+)
+from tests.integration.feature_repos.test_repo_configuration import construct_feature_store
+
+parameters = [
+    ("bigquery", BigQueryDataSourceCreator, "datastore"),
+]
+
+
+@pytest.mark.parametrize("params", parameters)
+@pytest.mark.skip(reason="Still working on this test")
+def test_e2e_consistency(params):
+    with construct_feature_store(params) as fs:
+        fv = fs.get_feature_view("test_correctness")
+        run_offline_online_store_consistency_test(fs, fv, True)
+
+
+def check_offline_and_online_features(
+    fs: FeatureStore,
+    fv: FeatureView,
+    driver_id: int,
+    event_timestamp: datetime,
+    expected_value: Optional[float],
+    full_feature_names: bool,
+    check_offline_store: bool = True,
+) -> None:
+    # Check online store
+    response_dict = fs.get_online_features(
+        [f"{fv.name}:value"],
+        [{"driver": driver_id}],
+        full_feature_names=full_feature_names,
+    ).to_dict()
+
+    if full_feature_names:
+        if expected_value:
+            assert abs(response_dict[f"{fv.name}__value"][0] - expected_value) < 1e-6
+        else:
+            assert response_dict[f"{fv.name}__value"][0] is None
+    else:
+        if expected_value:
+            assert abs(response_dict["value"][0] - expected_value) < 1e-6
+        else:
+            assert response_dict["value"][0] is None
+
+    # Check offline store
+    if check_offline_store:
+        df = fs.get_historical_features(
+            entity_df=pd.DataFrame.from_dict(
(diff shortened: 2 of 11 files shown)


sdk/python/tests/integration/feature_repos/universal/data_sources/bigquery.py
```diff
@@ -0,0 +1,50 @@
...
+        df: pd.DataFrame,
+        event_timestamp_column="ts",
+        created_timestamp_column="created_ts",
+        **kwargs,
+    ) -> DataSource:
+        gcp_project = self.client.project
+        bigquery_dataset = "test_ingestion"
+        dataset = bigquery.Dataset(f"{gcp_project}.{bigquery_dataset}")
+        self.client.create_dataset(dataset, exists_ok=True)
+        dataset.default_table_expiration_ms = (
```
woop, line 32: Would we need expiration if we are explicitly doing cleanup?

sdk/python/tests/integration/e2e/test_universal_e2e.py
```diff
@@ -0,0 +1,134 @@
...
+
+parameters = [
+    ("bigquery", BigQueryDataSourceCreator, "datastore"),
+]
+
+
+@pytest.mark.parametrize("params", parameters)
+@pytest.mark.skip(reason="Still working on this test")
+def test_e2e_consistency(params):
+    with construct_feature_store(params) as fs:
```
woop, line 23: I'm a bit lost here. It seems like

```
parameters = [
    ("bigquery", BigQueryDataSourceCreator, "datastore"),
]
```
is passed into 
```
construct_feature_store(params)
```
but `construct_feature_store` takes a `test_repo_config` as an input. 

Am I missing something :thinking:

sdk/python/tests/integration/feature_repos/universal/data_sources/bigquery.py
achals: We wouldn't

sdk/python/tests/integration/e2e/test_universal_e2e.py
achals: I updated this API to use the `TestRepoConfig` instead

sdk/python/tests/integration/feature_repos/test_repo_configuration.py
```diff
@@ -0,0 +1,79 @@
...
+            project=project,
+            provider=test_repo_config.provider,
+            offline_store=offline_store,
+            online_store=online_store,
+        )
+    fs = FeatureStore(config=config)
+    fv = correctness_feature_view(ds)
+    entity = driver
+    fs.apply([fv, entity])
+    yield fs
```
tsotnet, line 76: all of this probably needs to be inside with block, otherwise the temporary repo directory will be cleaned up

sdk/python/tests/integration/e2e/test_universal_e2e.py
```diff
@@ -0,0 +1,142 @@
...
+    C(),
+    C(
+        offline_store_creator="tests.integration.feature_repos.universal.data_sources.redshift.RedshiftDataSourceCreator",
+        online_store={"type": "dynamodb", "region": "us-west-2"},
+    ),
+    C(
+        offline_store_creator="tests.integration.feature_repos.universal.data_sources.bigquery.BigQueryDataSourceCreator",
+        online_store="datastore",
+    ),
+]
```
tsotnet, line 25: I think the idea behind universal repos is that you shouldn't have to come up with configs per test file. Can we make these common parameterizations part of the interface? (in a way that new offline store doesn't require adding the config in bunch of places)

tsotnet: Also for the e2e test we wanted to vary providers. Here all configs use local provider.

sdk/python/tests/data/data_creator.py
```diff
@@ -0,0 +1,27 @@
...
+            (ts - timedelta(hours=4))
+            .replace(tzinfo=utc)
+            .astimezone(tz=timezone("Europe/Berlin")),
+            (ts - timedelta(hours=1))
+            .replace(tzinfo=utc)
+            .astimezone(tz=timezone("US/Pacific")),
+        ],
+        "created_ts": [ts, ts, ts, ts, ts],
+    }
+    return pd.DataFrame.from_dict(data)
```
tsotnet, line 27: Maybe this is just a POC, but we'd like to have a bigger dataset here. Take a look at sdk/python/feast/driver_test_data.py, that file has some methods that test_historical_retrieval uses, and it generates much larger datasets.

sdk/python/tests/integration/feature_repos/universal/entities.py
```diff
@@ -0,0 +1,8 @@
+from feast import Entity, ValueType
+
+driver = Entity(
```
felixwang9817, line 3: Nit: this is just an `Entity` defined here, but in `universal/feature_views.py` we have a function `correctness_feature_view` that returns a `FeatureView`. We should standardize this.

sdk/python/tests/integration/feature_repos/test_repo_configuration.py
```diff
@@ -0,0 +1,79 @@
...
+            registry=str(Path(repo_dir_name) / "registry.db"),
+            project=project,
+            provider=test_repo_config.provider,
+            offline_store=offline_store,
+            online_store=online_store,
+        )
+    fs = FeatureStore(config=config)
+    fv = correctness_feature_view(ds)
+    entity = driver
+    fs.apply([fv, entity])
```
felixwang9817, line 75: Do we always want to have called `apply` here? For example, an e2e test should be calling `apply` on a brand new FeatureStore.

sdk/python/tests/integration/e2e/test_universal_e2e.py
felixwang9817: I agree that the parametrization should be part of the interface. I imagine that instead of `@pytest.mark.parametrize("config", configs)`, it would look like `@pytest.mark.parametrize("feature_store", generate_feature_stores(...))`. However, it's not obvious to me what the interface for `generate_feature_stores` should look like; one challenge is that different tests want to parametrize across different components of the FeatureStore (e2e across provider, registration across registry, etc.).

sdk/python/tests/integration/e2e/test_universal_e2e.py
```diff
@@ -0,0 +1,126 @@
...
+from typing import Optional
+
+import pandas as pd
+from pytz import utc
+
+from feast import FeatureStore, FeatureView
+from tests.integration.feature_repos.test_repo_configuration import parametrize_e2e_test
+
+
+# @pytest.mark.integration
```
felixwang9817, line 12: These should be removed.

felixwang9817 (approved):
/lgtm

PR #1735, opened 2021-07-23
Codyl/version update

codyjlin:
Similar to this PR in airflow's twitter-fork (https://github.com/twitter-forks/airflow/pull/29), we should be hardcoding our fork versions when we want to release to 3rdparty/python.


codyjlin:
Sorry, totally wrong upstream 🙄😬

PR #1739, opened 2021-07-27
Docstrings

### LINKED ISSUE #1610: Add docstrings to object constructors in Python reference API
In [https://rtd.feast.dev/en/latest/](https://rtd.feast.dev/en/latest/), we are missing documentation on what parameters do in the constructors for FeatureView, Entity, etc.

felixwang9817:
**What this PR does / why we need it**: This PR adds docstrings to public methods in public-facing classes (FeatureStore, FeatureView, Entity, etc.), and improves existing documentation. 

Fixes #1610 

```release-note
Added docstrings
```

### DIFF
--- a/sdk/python/feast/data_source.py
+++ b/sdk/python/feast/data_source.py
@@ -15,7 +15,7 @@
 
 import enum
 from abc import ABC, abstractmethod
-from typing import Callable, Dict, Iterable, Optional, Tuple
+from typing import Any, Callable, Dict, Iterable, Optional, Tuple
 
 from feast import type_map
 from feast.data_format import StreamFormat
@@ -220,20 +220,42 @@ class KinesisOptions:
 
 class DataSource(ABC):
     """
-    DataSource that can be used source features
+    DataSource that can be used to source features.
+
+    Args:
+        event_timestamp_column (optional): Event timestamp column used for point in time
+            joins of feature values.
+        created_timestamp_column (optional): Timestamp column indicating when the row
+            was created, used for deduplicating rows.
+        field_mapping (optional): A dictionary mapping of column names in this data
+            source to feature names in a feature table or view. Only used for feature
+            columns, not entity or timestamp columns.
+        date_partition_column (optional): Timestamp column used for partitioning.
     """
+
+    _event_timestamp_column: str
+    _created_timestamp_column: str
+    _field_mapping: Dict[str, str]
+    _date_partition_column: str
 
     def __init__(
         self,
-        event_timestamp_column: Optional[str] = "",
-        created_timestamp_column: Optional[str] = "",
+        event_timestamp_column: Optional[str] = None,
+        created_timestamp_column: Optional[str] = None,
         field_mapping: Optional[Dict[str, str]] = None,
-        date_partition_column: Optional[str] = "",
+        date_partition_column: Optional[str] = None,
     ):
-        self._event_timestamp_column = event_timestamp_column
-        self._created_timestamp_column = created_timestamp_column
+        """Inits the DataSource class."""
+        self._event_timestamp_column = (
+            event_timestamp_column if event_timestamp_column else ""
+        )
+        self._created_timestamp_column = (
+            created_timestamp_column if created_timestamp_column else ""
+        )
         self._field_mapping = field_mapping if field_mapping else {}
-        self._date_partition_column = date_partition_column
+        self._date_partition_column = (
+            date_partition_column if date_partition_column else ""
+        )
 
     def __eq__(self, other):
         if not isinstance(other, DataSource):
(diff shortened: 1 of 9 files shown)


sdk/python/feast/data_source.py
```diff
@@ -250,68 +272,73 @@ def __eq__(self, other):
...
     @abstractmethod
-    def from_proto(data_source: DataSourceProto):
-        """
-        Convert data source config in FeatureTable spec to a DataSource class object.
+    def from_proto(data_source: DataSourceProto) -> Any:
         """
+        Converts data source config in FeatureTable spec to a DataSource class object.
+
+        Args:
+            data_source: A protobuf representation of a DataSource.
```
achals, line 338: Don't you need a `returns` here?

sdk/python/feast/entity.py
```diff
@@ -30,8 +30,25 @@
...
+            within the collection. Used as a key for joining entities with their
+            associated features. If not specified, defaults to the name of the entity.
+        labels (optional): User-defined metadata in dictionary form.
     """
 
+    _name: str
+    _value_type: ValueType
+    _description: str
+    _join_key: str
+    _labels: MutableMapping[str, str]
```
achals, line 48: nit, `MutableMapping` is a proto concept. we should be using `dicts` in the python classes and only deal with MutableMapping (if at all) in the `to_proto` or `from_proto` methods.

I don't think we *need* to address that in this PR, but i think it should be fixed at some point.

sdk/python/feast/entity.py
```diff
@@ -30,8 +30,25 @@
...
+    _created_timestamp: Optional[Timestamp]
+    _last_updated_timestamp: Optional[Timestamp]
```
achals, line 50: Similar comment, we should be using `datetime` instead of `Timestamp`.

sdk/python/feast/feature_service.py
```diff
@@ -34,13 +41,10 @@ def __init__(
         tags: Optional[Dict[str, str]] = None,
     ):
         """
-        Create a new Feature Service object.
+        Inits the FeatureService class.
```
woop, line 44: `Inits` is not really descriptive. `Creates`?

achals (commented):
Looks good, mostly nits. Are doctests part of this PR or a different PR?

sdk/python/feast/feature_store.py
```diff
@@ -61,16 +62,22 @@ class FeatureStore:
     def __init__(
         self, repo_path: Optional[str] = None, config: Optional[RepoConfig] = None,
     ):
+        """
+        Inits the FeatureStore class.
```
woop, line 66: Same here.

sdk/python/feast/data_source.py
```diff
@@ -220,20 +220,42 @@ def to_proto(self) -> DataSourceProto.KinesisOptions:
...
+    _event_timestamp_column: str
+    _created_timestamp_column: str
+    _field_mapping: Dict[str, str]
+    _date_partition_column: str
+
     def __init__(
         self,
-        event_timestamp_column: Optional[str] = "",
-        created_timestamp_column: Optional[str] = "",
+        event_timestamp_column: Optional[str] = None,
```
woop, line 243: I'm curious about these type annotation changes. Are they necessary for the docs?

sdk/python/feast/feature_service.py
felixwang9817: That's fair, I'll leave this docstring as is (for context, I took this inits docstring from the Google style guide, but I agree it's pretty non-descriptive).

felixwang9817:
> Looks good, mostly nits. Are doctests part of this PR or a different PR?

Doctests are separate.

sdk/python/feast/data_source.py
felixwang9817: Good catch.

sdk/python/feast/entity.py
felixwang9817: Okay, I'll try to switch this to `dict`s.

sdk/python/feast/entity.py
felixwang9817: Got it.

sdk/python/feast/data_source.py
felixwang9817: These changes are not strictly necessary to enforce proper typing. However, with the default value set to `None`, I believe it's easier for a user to see that the line `self._event_timestamp_column = event_timestamp_column if event_timestamp_column else ""` will leave an input string untouched, and otherwise will set `self._event_timestamp_column` to the empty string. In general, if there is a parameter of type `Optional[X]` and a `None` value will lead to some inference, I think it's more clear to set the default value to `None`.

achals (approved):
/lgtm

PR #1741, opened 2021-07-28
Updating CLI apply to use FeatureStore & remove feature table logic

adchia:

Moves the feast CLI's apply method (`feast apply`) to use the SDK FeatureStore class more. This includes some cleanup that also removes obsolete logic around feature tables. Note that because the feast CLI's apply is a total apply whereas the SDK apply is partial, there is still leftover business logic (in particular for deleting entities / FVs that don't match the repo). 

```release-note 

action required: feature tables were unofficially supported via the CLI and now no longer are

```

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -92,6 +92,10 @@ class FeatureStore:
         return get_version()
 
     @property
+    def registry(self) -> Registry:
+        return self._registry
+
+    @property
     def project(self) -> str:
         """Gets the project of this feature store."""
         return self.config.project
--- a/sdk/python/feast/repo_operations.py
+++ b/sdk/python/feast/repo_operations.py
@@ -193,7 +193,7 @@ def apply_total(repo_config: RepoConfig, repo_path: Path, skip_source_validation
     apply_feature_services(registry, project, repo)
 
     infra_provider = get_provider(repo_config, repo_path)
-    for name in [table.name for table in views_to_keep]:
+    for name in [view.name for view in views_to_keep]:
         click.echo(
             f"Deploying infrastructure for {Style.BRIGHT + Fore.GREEN}{name}{Style.RESET_ALL}"
         )
--- a/sdk/python/feast/repo_operations.py
+++ b/sdk/python/feast/repo_operations.py
@@ -154,7 +154,6 @@ def apply_total(repo_config: RepoConfig, repo_path: Path, skip_source_validation
     sys.dont_write_bytecode = True
     repo = parse_repo(repo_path)
     _validate_feature_views(repo.feature_views)
-    data_sources = [t.batch_source for t in repo.feature_views]
 
     if not skip_source_validation:
         data_sources = [t.input for t in repo.feature_views]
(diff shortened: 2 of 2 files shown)


achals:
Looks good to me. I wonder if we should change the `apply` method on the feature store to be a total apply, or add a `total_apply` method, so that this logic can be used without relying on repo_operations.

sdk/python/feast/feature_store.py
```diff
@@ -91,6 +91,10 @@ def version(self) -> str:
         """Returns the version of the current Feast SDK/CLI."""
         return get_version()
 
+    @property
+    def registry(self) -> Registry:
```
felixwang9817, line 95: Nit: can we add a docstring?

sdk/python/feast/repo_operations.py
```diff
@@ -145,134 +141,98 @@ def apply_total(repo_config: RepoConfig, repo_path: Path, skip_source_validation
...
+    repo_entities_names = set([e.name for e in repo.entities])
+    for registry_entity in registry.list_entities(project=project):
+        if registry_entity.name not in repo_entities_names:
+            entities_to_delete.append(registry_entity)
+    return entities_to_delete, entities_to_keep
+
+
+def _tag_registry_views_for_keep_delete(project, registry, repo):
+    views_to_keep: List[FeatureView] = repo.feature_views
+    views_to_delete = []
```
felixwang9817, line 228: Nit: can we add a type?

sdk/python/feast/repo_operations.py
```diff
@@ -145,134 +141,98 @@ def apply_total(repo_config: RepoConfig, repo_path: Path, skip_source_validation
...
 
 
+def _tag_registry_entities_for_keep_delete(project, registry, repo):
+    entities_to_keep: List[Entity] = repo.entities
+    entities_to_delete: List[Entity] = []
+    repo_entities_names = set([e.name for e in repo.entities])
+    for registry_entity in registry.list_entities(project=project):
+        if registry_entity.name not in repo_entities_names:
+            entities_to_delete.append(registry_entity)
+    return entities_to_delete, entities_to_keep
```
felixwang9817, line 223: Nit: the function is named `_for_keep_delete`, but the lists returned are for deleting and keeping, respectively (i.e. the order is mixed up).

sdk/python/feast/repo_operations.py
```diff
@@ -145,134 +141,98 @@ def apply_total(repo_config: RepoConfig, repo_path: Path, skip_source_validation
...
-    for table in repo.feature_tables:
-        registry.apply_feature_table(table, project, commit=False)
-        click.echo(
-            f"Registered feature table {Style.BRIGHT + Fore.GREEN}{table.name}{Style.RESET_ALL}"
-        )
 
     # Delete views that should not exist
     for registry_view in views_to_delete:
-        registry.delete_feature_view(registry_view.name, project=project, commit=False)
+        store.delete_feature_view(registry_view.name, project=project, commit=False)
```
felixwang9817, line 175: Does this actually work? I thought that `FeatureStore.delete_feature_view` doesn't expose a `commit` parameter.

adchia: seemed to not complain, but you're right so removed.

adchia:
Moving to different PR since I added a dev branch to the main repo

PR #1744, opened 2021-07-29
Fix BQ historical retrieval with rows that got backfilled

MattDelac:
There is bug about finding the "latest" values when the rows have been backfilled (when _created_timestamp_ and _event_timestamp_ don't follow the same order).

The problem is that the `ANY_VALUE` strategy does not work here because it will take the first occurrence it sees. 

For example, the following example would fail
```markdown
| driver_id | value | event_timestamp | created_at |
|-----------|-------|-----------------|------------|
| 1         | 10    | today           | tomorrow   |
| 1         | 20    | tomorrow        | today      |
```

The JOIN happening later would consist of the entity_dataframe with the `__latest` CTE where created_at and event_timestamp got mixed
```markdown
| driver_id | value | event_timestamp | created_at |
|-----------|-------|-----------------|------------|
| 1         | 20    | tomorrow        | tomorrow   |
```

Because the JOIN operation happens on `USING(driver_id, event_timestamp, created_at)`, it cannot find a match and will return NULL for the desired feature.

The goal of this PR is to make this step work and deterministic by using a Window function. Then we preserve the information between event_timestamp and created_at on a given row

**Tophat**
```sh
# Logs of the example I ran (on billions of records)
Job started at 2021-08-05 11:07:28
Job started at 2021-08-05 11:42:00 # Thus it took 35 minutes
```

Metrics (compared to current SQL template)
**Elapsed time**: 24 min 49 sec (vs 26 min 35 sec)
**Slot time consumed**: 21 days 6 hr (vs 26 days 0 hr)
**Bytes shuffled**: 23.11 TB (vs 22.86 TB)
**Bytes spilled to disk**: 7.87 TB (10.09 TB)

I also confirm that it fixes the bug we had on a specific feature view

The new test properly catches the problem encountered with the backfill rows (as the feature of the second driver is Null instead of 40)
image

Fixes #

```release-note
[Bug] Fixed how we retrieve the latest value of a feature in the BigQuery SQL template
```

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -430,20 +430,21 @@ WITH entity_dataframe AS (
  Thus we only need to compute the latest timestamp of each feature.
 */
 {{ featureview.name }}__latest AS (
-    SELECT
-        {{featureview.name}}__entity_row_unique_id,
-        MAX(event_timestamp) AS event_timestamp
+    SELECT * EXCEPT(row_number)
+    FROM
+    (
+        SELECT *,
+            ROW_NUMBER() OVER(
+                PARTITION BY {{featureview.name}}__entity_row_unique_id
+                ORDER BY event_timestamp DESC{% if featureview.created_timestamp_column %},created_timestamp DESC{% endif %}
+            ) AS row_number,
+        FROM {{ featureview.name }}__base
         {% if featureview.created_timestamp_column %}
-            ,ANY_VALUE(created_timestamp) AS created_timestamp
+            INNER JOIN {{ featureview.name }}__dedup
+            USING ({{featureview.name}}__entity_row_unique_id, event_timestamp, created_timestamp)
         {% endif %}
-
-    FROM {{ featureview.name }}__base
-    {% if featureview.created_timestamp_column %}
-        INNER JOIN {{ featureview.name }}__dedup
-        USING ({{featureview.name}}__entity_row_unique_id, event_timestamp, created_timestamp)
-    {% endif %}
-
-    GROUP BY {{featureview.name}}__entity_row_unique_id
+    )
+    WHERE row_number = 1
 ),
 
 /*
--- a/sdk/python/tests/integration/offline_store/test_historical_retrieval.py
+++ b/sdk/python/tests/integration/offline_store/test_historical_retrieval.py
@@ -992,3 +992,137 @@ def test_feature_name_collision_on_historical_retrieval():
         "have different names."
     )
     assert str(error.value) == expected_error_message
+
+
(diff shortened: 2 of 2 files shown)


sdk/python/feast/driver_test_data.py
```diff
@@ -142,8 +143,23 @@ def create_driver_hourly_stats_df(drivers, start_date, end_date) -> pd.DataFrame
...
+            backfill_rows.assign(
+                created=(backfill_rows.iloc[0]["created"] - timedelta(days=1))
+            ),
+            backfill_rows.assign(
+                created=(backfill_rows.iloc[0]["created"] + timedelta(days=1))
+            ),
+        ]
+    )
+    df_all_drivers = pd.concat(
+        [df_all_drivers, duplicated_rows, backfill_rows], ignore_index=True
```
MattDelac, line 161: That does not change anything. Without the fix the tests were passing anyway

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -430,20 +430,16 @@ def _get_bigquery_client(project: Optional[str] = None):
...
-        MAX(event_timestamp) AS event_timestamp
-        {% if featureview.created_timestamp_column %}
-            ,ANY_VALUE(created_timestamp) AS created_timestamp
-        {% endif %}
-
+    SELECT *,
+        ROW_NUMBER() OVER(
+            PARTITION BY {{featureview.name}}__entity_row_unique_id
+            ORDER BY event_timestamp DESC{% if featureview.created_timestamp_column %},created_timestamp DESC{% endif %}
+        ) AS row_number,
```
woop, line 437: So is this table still the latest values? Wouldn't you have all the different rows with different row numbers?

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -453,7 +449,7 @@ def _get_bigquery_client(project: Optional[str] = None):
 {{ featureview.name }}__cleaned AS (
     SELECT base.*
     FROM {{ featureview.name }}__base as base
-    INNER JOIN {{ featureview.name }}__latest
+    INNER JOIN (SELECT * EXCEPT(row_number) FROM {{ featureview.name }}__latest WHERE row_number = 1)
```
woop, line 452: Any reason to do the `(SELECT * EXCEPT(row_number) FROM {{ featureview.name }}__latest WHERE row_number = 1)` here instead of doing `WHERE row_number = 1` when `__latest` is being created?

sdk/python/feast/infra/offline_stores/bigquery.py
MattDelac: Done

woop (approved):

PR #1752, opened 2021-07-30
Adding small note for project naming convention

codyjlin:

Unfortunately I had accidentally specified a project name with a dash. The error that popped up was from sqlite (following a `store.apply` call when trying to create tables in the online store.):
```
sqlite3.OperationalError: near “-”: syntax error
```
which actually took me some extra debugging to figure out what my issue was.

N/A. Too small for a PR, but it could be related to overall validation or more docs specifying naming limitations.

```release-note
NONE
```

### DIFF
--- a/docs/reference/feature-repository/feature-store-yaml.md
+++ b/docs/reference/feature-repository/feature-store-yaml.md
@@ -23,7 +23,7 @@ The following top-level configuration options exist in the `feature_store.yaml` 
 * **registry** — Configures the location of the feature registry.
 * **online\_store** — Configures the online store.
 * **offline\_store** — Configures the offline store.
-* **project** — Defines a namespace for the entire feature store. Can be used to isolate multiple deployments in a single installation of Feast.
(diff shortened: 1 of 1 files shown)


achals (commented):
Thanks for the PR @codyjlin. Do you think we should be adding some validation in code to catch this and raise a better exception?

mavysavydav:
LGTM

codyjlin:
> Thanks for the PR @codyjlin. Do you think we should be adding some validation in code to catch this and raise a better exception?

@achals I think so since the error message I did get was pretty unhelpful, but I'm not sure how frequently users would run into this error - might not be something to prioritize right now.

tedhtchang:
@codyjlin You may call the name validation function [here](https://github.com/feast-dev/feast/blob/a548c48927e6f6858d91a93cf356b43fe7c67aad/sdk/python/feast/repo_operations.py#L390) to validate the project name. Or you can create an issue and assign to me.

codyjlin:
Thanks @tedhtchang!! Just created https://github.com/feast-dev/feast/issues/1755.

achals (approved):
/lgtm

PR #1754, opened 2021-08-01
Fix `feast apply` bugs

tsotnet:

**What this PR does / why we need it**: This fixes 2 bugs during `feast apply`.
1. We always initialized the empty registry (instead of loading the existing registry). This means that if during 1st apply we deployed infra for FeatureView A, and in the 2nd apply we renamed FeatureView A -> B, `feast apply` would create the table for B, but wouldn't remove the table for A.
2. The `registry.commit()` happened before the infra update call on the provider. This means that if the infra update fails (for example bad permissions), the registry still got updated. Which means that during the next apply, Feast wouldn't do anything because it'll think that everything is up to date in the registry. I moved the commit at the very end of apply_total method, and updated the rest of the function accordingly.

Fixes #

```release-note
Fix `feast apply` bugs - incorrect initialization of registry and early registry commit
```

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -24,7 +24,11 @@ from tqdm import tqdm
 
 from feast import utils
 from feast.entity import Entity
-from feast.errors import FeatureNameCollisionError, FeatureViewNotFoundException
+from feast.errors import (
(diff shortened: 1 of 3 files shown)


achals (approved):
/lgtm

PR #1768, opened 2021-08-09
Adding initial type support related tests for BQ

adchia:
Adds tests that tests given input data of different types (for entities vs features), whether the resultant feature vectors from get_historical_features and get_online_features and inference match expectations.

As part of this, updating inference to respect entity join_key.

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/inference.py
+++ b/sdk/python/feast/inference.py
@@ -29,9 +29,11 @@ def update_entities_with_inferred_types_from_feature_views(
         col_names_and_types = view.batch_source.get_table_column_names_and_types(config)
         for entity_name in view.entities:
             if entity_name in incomplete_entities:
+                entity = incomplete_entities[entity_name]
+
                 # get entity information from information extracted from the view batch source
                 extracted_entity_name_type_pairs = list(
-                    filter(lambda tup: tup[0] == entity_name, col_names_and_types)
+                    filter(lambda tup: tup[0] == entity.join_key, col_names_and_types,)
                 )
                 if len(extracted_entity_name_type_pairs) == 0:
                     # Doesn't mention inference error because would also be an error without inferencing
@@ -40,7 +42,6 @@ def update_entities_with_inferred_types_from_feature_views(
                         its entity's name."""
                     )
 
-                entity = incomplete_entities[entity_name]
                 inferred_value_type = view.batch_source.source_datatype_to_feast_value_type()(
                     extracted_entity_name_type_pairs[0][1]
                 )
--- a/sdk/python/tests/integration/e2e/test_universal_e2e.py
+++ b/sdk/python/tests/integration/e2e/test_universal_e2e.py
@@ -78,7 +78,7 @@ def check_offline_and_online_features(
 def run_offline_online_store_consistency_test(
     fs: FeatureStore, fv: FeatureView
 ) -> None:
-    now = datetime.utcnow()
+    now = datetime.now()
 
     full_feature_names = True
     check_offline_store: bool = True
--- a/sdk/python/tests/integration/feature_repos/test_repo_configuration.py
+++ b/sdk/python/tests/integration/feature_repos/test_repo_configuration.py
@@ -10,6 +10,7 @@ import pytest
 
 from feast import FeatureStore, FeatureView, RepoConfig, driver_test_data, importer
 from feast.data_source import DataSource
+from feast.value_type import ValueType
 from tests.data.data_creator import create_dataset
 from tests.integration.feature_repos.universal.data_source_creator import (
     DataSourceCreator,
@@ -70,7 +71,6 @@ FULL_REPO_CONFIGS: List[TestRepoConfig] = [
     ),
 ]
 
-
 OFFLINE_STORES: List[str] = []
 ONLINE_STORES: List[str] = []
 PROVIDERS: List[str] = []
@@ -83,6 +83,9 @@ class Environment:
     feature_store: FeatureStore
     data_source: DataSource
     data_source_creator: DataSourceCreator
+    entity_type: ValueType
+    feature_dtype: str
+    feature_is_list: bool
 
     end_date = datetime.now().replace(microsecond=0, second=0, minute=0)
     start_date = end_date - timedelta(days=7)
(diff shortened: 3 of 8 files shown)


sdk/python/feast/inference.py
```diff
@@ -29,9 +29,17 @@ def update_entities_with_inferred_types_from_feature_views(
         col_names_and_types = view.batch_source.get_table_column_names_and_types(config)
         for entity_name in view.entities:
             if entity_name in incomplete_entities:
+                entity = incomplete_entities[entity_name]
+                entity_table_col_join_key = (
+                    entity.join_key if entity.join_key else entity.name
```
felixwang9817, line 34: Why not just `entity.join_key`? If `join_key` is not specified in the initialization of an Entity, it will be set to `name`.

sdk/python/feast/inference.py
```diff
@@ -29,9 +29,17 @@ def update_entities_with_inferred_types_from_feature_views(
         col_names_and_types = view.batch_source.get_table_column_names_and_types(config)
         for entity_name in view.entities:
             if entity_name in incomplete_entities:
+                entity = incomplete_entities[entity_name]
+                entity_table_col_join_key = (
```
felixwang9817, line 33: Nit: the name `entity_table_col_join_key` seems a bit verbose. What about just `entity_join_key`?

sdk/python/tests/integration/registration/test_types.py
```diff
@@ -0,0 +1,273 @@
...
+    }
+    non_list_val = dtype_map[dtype]
+    # Duplicate the value once if this is a list
+    if is_list:
+        return [[n, n] if n is not None else None for n in non_list_val]
+    else:
+        return non_list_val
+
+
+def create_dataset(
```
felixwang9817, line 54: I think all the dataset creation functions can go in `tests/data/data_creator.py`. At least, that's what the structure of the first universal feature repo PR (#1734) suggests.

sdk/python/feast/inference.py
adchia: Done

sdk/python/feast/inference.py
adchia: Good call. Changed

sdk/python/tests/integration/registration/test_types.py
adchia: Moved to universal repo. PTAL!

sdk/python/tests/integration/e2e/test_universal_e2e.py
```diff
@@ -78,7 +78,7 @@ def check_offline_and_online_features(
 def run_offline_online_store_consistency_test(
     fs: FeatureStore, fv: FeatureView
 ) -> None:
-    now = datetime.utcnow()
+    now = datetime.now()
```
achals, line 81: Why was this changed to `now` from `utcnow`?

sdk/python/tests/integration/feature_repos/test_repo_configuration.py
```diff
@@ -208,9 +214,14 @@ def construct_test_environment(
...
     :param test_repo_config: configuration
+    :param create_and_apply: whether to create and apply the repo config
+    :param materialize: whether to materialize features to online store
+    :param entity_type: the data type for the entity column (i.e. id)
+    :param feature_dtype: the data type for the feature column (i.e. value)
+    :param feature_is_list: whether the feature column (i.e. value) should be a list feature
     :return: A feature store built using the supplied configuration.
     """
-    df = create_dataset()
+    df = create_dataset(entity_type, feature_dtype, feature_is_list)
```
achals, line 224: I think this is cool for now, but just a note that I'm refactoring this path a bit so the dataset creation is more pluggable to allow for custom datasets a little more easily.

sdk/python/tests/integration/feature_repos/universal/entities.py
```diff
@@ -10,5 +10,10 @@ def driver():
     )
 
 
+def driver_no_value_type():
+    # Don't specify value type in entity to force inference
+    return Entity(name="driver", description="driver id", join_key="driver_id",)
+
+
```
achals, line 17: Do you want to use the existing driver entity but add some params to customize it? Like:
```
def driver(value=True):
    return Entity(
        name="driver",  # The name is derived from this argument, not object name.
        value_type=ValueType.INT64 if Value else ValueType.UNKNOWN,
        description="driver id",
        join_key="driver_id",
    )
```

sdk/python/tests/integration/materialization/test_offline_online_store_consistency.py
```diff
@@ -332,7 +332,7 @@ def check_offline_and_online_features(
 def run_offline_online_store_consistency_test(
     fs: FeatureStore, fv: FeatureView, full_feature_names: bool,
 ) -> None:
-    now = datetime.utcnow()
+    now = datetime.now()
```
achals, line 335: same q

achals (approved):
looks good with some nits.

sdk/python/tests/integration/e2e/test_universal_e2e.py
adchia: We have some places where we used utcnow and other places where we used now, which caused cases where we sometimes got no results when doing a historical retrieval

sdk/python/tests/integration/feature_repos/universal/entities.py
adchia: Done!

sdk/python/tests/integration/materialization/test_offline_online_store_consistency.py
adchia: same response. create_data_set was always using utcnow, but there are other tests that don't use utcnow and instead use now(). This unifies across our test suites

sdk/python/tests/integration/feature_repos/test_repo_configuration.py
adchia: Noted. Would you rather I wait to merge this in until after your refactor then?

achals (approved):
/lgtm

PR #1769, opened 2021-08-09
Add offline retrival integration tests using the universal repo

achals:

We wanna start parameterizing over all the possible combinations when reading feature values from the offline feature store. This PR introduces the decorator needed to set up the underlying tables as needed, updates the historical integration test to operate on an Environment object.

```
$ FEAST_TELEMETRY=False pytest --cov=./ --cov-report=xml --color=yes -s --integration sdk/python/tests/integration/offline_store/test_universal_historical_retrieval.py -n 8
======================================================================================= test session starts =======================================================================================
platform darwin -- Python 3.7.11, pytest-6.0.0, py-1.10.0, pluggy-0.13.1
rootdir: /Users/achal/tecton/feast/sdk/python
plugins: cov-2.12.1, xdist-2.3.0, ordering-0.6, mock-1.10.4, timeout-1.4.2, forked-1.3.0, lazy-fixture-0.6.3
gw0 [20] / gw1 [20] / gw2 [20] / gw3 [20] / gw4 [20] / gw5 [20] / gw6 [20] / gw7 [20]
....................

---------- coverage: platform darwin, python 3.7.11-final-0 ----------
Coverage XML written to file coverage.xml

================================================================================= 20 passed in 162.39s (0:02:42) ==================================================================================
```

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/tests/integration/feature_repos/test_repo_configuration.py
+++ b/sdk/python/tests/integration/feature_repos/test_repo_configuration.py
@@ -1,21 +1,25 @@
 import tempfile
 import uuid
 from contextlib import contextmanager
+from dataclasses import replace
+from datetime import timedelta, datetime
 from pathlib import Path
-from typing import Dict, List, Tuple, Union
+from typing import Dict, List, Tuple, Union, Optional
 
 import pytest
+import pandas as pd
 from attr import dataclass
 
-from feast import FeatureStore, RepoConfig, importer
+from feast import FeatureStore, RepoConfig, importer, driver_test_data, FeatureView
 from feast.data_source import DataSource
 from tests.data.data_creator import create_dataset
 from tests.integration.feature_repos.universal.data_source_creator import (
     DataSourceCreator,
 )
+from tests.integration.feature_repos.universal.entities import driver, customer
 
 
-@dataclass
+@dataclass(frozen=True, str=True)
 class TestRepoConfig:
     """
     This class should hold all possible parameters that may need to be varied by individual tests.
@@ -27,6 +31,7 @@ class TestRepoConfig:
     offline_store_creator: str = "tests.integration.feature_repos.universal.data_sources.file.FileDataSourceCreator"
 
     full_feature_names: bool = True
+    infer_event_timestamp_col: bool = True
 
 
 FULL_REPO_CONFIGS: List[TestRepoConfig] = [
@@ -49,10 +54,75 @@ ONLINE_STORES: List[str] = []
 PROVIDERS: List[str] = []
 
 
+@dataclass
+class Environment:
+    name: str
+    test_repo_config: TestRepoConfig
+    feature_store: FeatureStore
+    data_source: DataSource
+    data_source_creator: DataSourceCreator
+
+    customer_entities = list(range(1001, 1110))
+    driver_entities = list(range(5001, 5110))
+
+    end_date = datetime.now().replace(microsecond=0, second=0, minute=0)
+    start_date = end_date - timedelta(days=7)
+
+    def customer_fixtures(self) -> Tuple[pd.DataFrame, FeatureView]:
+        customer_df = driver_test_data.create_customer_daily_profile_df(
+            self.customer_entities, self.start_date, self.end_date
+        )
+        customer_table_id = self.data_source_creator.get_prefixed_table_name(self.name,
+                                                                             "customer_profile")
+        # self.data_source_creator.upload_df(customer_df, customer_table_id)
+
+    def driver_stats_fixtures(self) -> Tuple[pd.DataFrame, FeatureView]:
+        pass
+
+    def order_fixtures(self) -> Tuple[pd.DataFrame, FeatureView]:
+        pass
+
+    def orders_sql_fixtures(self) -> Optional[str]:
+        pass
+
+
+def vary_full_feature_names(configs: List[TestRepoConfig]) -> List[TestRepoConfig]:
+    new_configs = []
+    for c in configs:
+        true_c = replace(c, full_feature_names=True)
+        false_c = replace(c, full_feature_names=False)
+        new_configs.extend([true_c, false_c])
+    return new_configs
+
+
+def vary_infer_event_timestamp_col(configs: List[TestRepoConfig]) -> List[TestRepoConfig]:
+    new_configs = []
+    for c in configs:
+        true_c = replace(c, infer_event_timestamp_col=True)
+        false_c = replace(c, infer_event_timestamp_col=False)
+        new_configs.extend([true_c, false_c])
+    return new_configs
+
+
+def vary_providers_for_offline_stores(configs: List[TestRepoConfig]) -> List[TestRepoConfig]:
+    new_configs = []
+    for c in configs:
+        if 'FileDataSourceCreator' in c.offline_store_creator:
+            new_configs.append(c)
+        elif 'RedshiftDataSourceCreator' in c.offline_store_creator:
+            for p in ["local", "aws"]:
+                new_configs.append(replace(c, provider=p))
+        elif 'BigQueryDataSourceCreator' in c.offline_store_creator:
+            for p in ["local", "gcp", "gcp_custom_offline_config"]:
+                new_configs.append(replace(c, provider=p))
+    return new_configs
+
+
 @contextmanager
-def construct_feature_store(
+def construct_test_environment(
     test_repo_config: TestRepoConfig,
-) -> Tuple[FeatureStore, DataSource]:
+    create_and_apply: bool = False
+) -> Environment:
     """
     This method should take in the parameters from the test repo config and created a feature repo, apply it,
     and return the constructed feature store object to callers.
@@ -73,8 +143,8 @@ def construct_feature_store(
 
     offline_creator: DataSourceCreator = importer.get_class_from_type(
         module_name, config_class_name, "DataSourceCreator"
-    )()
-    ds = offline_creator.create_data_source(project, df)
+    )(project)
+    ds = offline_creator.create_data_sources(project, df)
     offline_store = offline_creator.create_offline_store_config()
     online_store = test_repo_config.online_store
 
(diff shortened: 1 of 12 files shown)


sdk/python/tests/integration/feature_repos/test_repo_configuration.py
```diff
@@ -1,24 +1,25 @@
...
 )
-from tests.integration.feature_repos.universal.entities import driver
-from tests.integration.feature_repos.universal.feature_views import (
-    correctness_feature_view,
-)
+from tests.integration.feature_repos.universal.entities import customer, driver
 
 
-@dataclass
+@dataclass(frozen=True, str=True)
```
adchia, line 22: what is str=True doing here? dont see it in documentation around dataclass?

sdk/python/tests/integration/feature_repos/universal/feature_views.py
```diff
@@ -4,11 +4,44 @@
...
         input=data_source,
     )
+
+
+def create_driver_hourly_stats_feature_view(source):
+    driver_stats_feature_view = FeatureView(
+        name="driver_stats",
+        entities=["driver"],
+        features=[
+            Feature(name="conv_rate", dtype=ValueType.FLOAT),
```
adchia, line 24: optional nit: we seem to reference these feature names + types across files. Could imagine a cleanup being having some field that's a map of feature_name -> Tuple[dtype, feature_ref(i.e. feature_view_name:feature_name), expected type] so we dont have to redefine the full set of features + their types.

sdk/python/tests/integration/offline_store/test_universal_historical_retrieval.py
```diff
@@ -0,0 +1,357 @@
...
+
+def stage_orders_bigquery(df, table_id):
+    client = bigquery.Client()
+    job_config = bigquery.LoadJobConfig()
+    df.reset_index(drop=True, inplace=True)
+    job = client.load_table_from_dataframe(df, table_id, job_config=job_config)
+    job.result()
+
+
+class BigQueryDataSet:
```
adchia, line 141: seems like a weird spot to have this? i.e. in what seems like an offline store agnostic test

sdk/python/tests/integration/feature_repos/test_repo_configuration.py
achals: This generates a human readable string for pytest summaries:

```
===================================================================================== short test summary info =====================================================================================
FAILED sdk/python/tests/integration/offline_store/test_universal_historical_retrieval.py::test_historical_features_from_bigquery_sources[TestRepoConfig(provider='local', online_store='sqlite', offline_store_creator='tests.integration.feature_repos.universal.data_sources.file.FileDataSourceCreator', full_feature_names=True, infer_event_timestamp_col=True)]
```

achals: It changed to `repr` cause I was using dataclasses from `attr` by mistake. Fixed now.

sdk/python/tests/integration/feature_repos/test_repo_configuration.py
```diff
@@ -52,8 +57,108 @@ class TestRepoConfig:
...
+
+    orders_df = driver_test_data.create_orders_df(
+        customers=customer_entities,
+        drivers=driver_entities,
+        start_date=before_start_date,
+        end_date=after_end_date,
+        order_count=1000,
+    )
+
+    def customer_fixtures(self) -> FeatureView:
```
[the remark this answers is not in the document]
achals: Yeah this started off producing more things but `x_feature_view` makes more sense.

sdk/python/tests/integration/feature_repos/test_repo_configuration.py
```diff
@@ -52,8 +57,108 @@ class TestRepoConfig:
...
+            self.driver_df,
+            event_timestamp_column="event_timestamp",
+            created_timestamp_column="created",
+        )
+        return create_driver_hourly_stats_feature_view(ds)
+
+    def order_fixtures(self) -> pd.DataFrame:
+        return self.orders_df
+
+    def orders_sql_fixtures(self) -> Optional[str]:
```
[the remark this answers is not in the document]
achals: yeah, I'll fill it out and uncomment some of the test cases (and also add comments on the environment methods)

sdk/python/tests/integration/feature_repos/test_repo_configuration.py
```diff
@@ -52,8 +57,108 @@ class TestRepoConfig:
...
+        pass
+
+
+def vary_full_feature_names(configs: List[TestRepoConfig]) -> List[TestRepoConfig]:
+    new_configs = []
+    for c in configs:
+        true_c = replace(c, full_feature_names=True)
+        false_c = replace(c, full_feature_names=False)
+        new_configs.extend([true_c, false_c])
```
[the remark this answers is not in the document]
achals: So, the configs argument *has* the params  but they are  set to a default value so that the config is still valid for a test to use. 

Also just to note this is only adding configs to the `new_configs` list which doesn't get the original value, so I was hoping that i could get away with it without having to worry about *what* the default value was exactly.

Bug, instead of doing this replace twice, would you prefer that I just do a `replace` with the non-default value? That will probably read better.

sdk/python/tests/integration/feature_repos/test_repo_configuration.py
```diff
@@ -89,21 +194,78 @@ def construct_feature_store(test_repo_config: TestRepoConfig) -> FeatureStore:
...
+                    ]
+                )
+                fs.apply(fvs + entities)
 
-        fs.teardown()
-        offline_creator.teardown(project)
+            yield environment
+        finally:
+            if create_and_apply:
+                offline_creator.teardown(project)
```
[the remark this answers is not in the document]
achals: Good point there's no harm.

sdk/python/tests/integration/feature_repos/universal/data_sources/bigquery.py
```diff
@@ -12,43 +11,64 @@
...
-        self.client.update_dataset(dataset, ["default_table_expiration_ms"])
 
         job_config = bigquery.LoadJobConfig()
-        table_ref = f"{gcp_project}.{bigquery_dataset}.{name}_{int(time.time_ns())}"
+        if self.gcp_project not in destination:
+            destination = f"{self.gcp_project}.{self.project_name}.{destination}"
+
         job = self.client.load_table_from_dataframe(
-            df, table_ref, job_config=job_config
+            df, destination, job_config=job_config
```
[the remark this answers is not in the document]
achals: I killed that method, they did the same thing.

sdk/python/tests/integration/offline_store/test_universal_historical_retrieval.py
```diff
@@ -0,0 +1,357 @@
...
+)
+from tests.integration.feature_repos.test_repo_configuration import (
+    Environment,
+    parametrize_offline_retrieval_test,
+)
+
+np.random.seed(0)
+
+
+# Converts the given column of the pandas records to UTC timestamps
```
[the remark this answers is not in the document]
achals: Yeah can rm

sdk/python/tests/integration/offline_store/test_universal_historical_retrieval.py
```diff
@@ -0,0 +1,357 @@
...
+def convert_timestamp_records_to_utc(records, column):
```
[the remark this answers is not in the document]
achals: Will add

sdk/python/tests/integration/feature_repos/universal/feature_views.py
achals: yeah there's probably some clean up we can do

sdk/python/tests/integration/offline_store/test_universal_historical_retrieval.py
achals: Yeah this is copy-pasta from the original test which had a bunch of fixtures in the same file. I'm going to kill em all

sdk/python/tests/integration/feature_repos/universal/data_sources/bigquery.py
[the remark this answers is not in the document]
adchia: the diff to me suggests you added the upload_df method?

the self.tables not being applied comment still applies regardless

sdk/python/tests/integration/feature_repos/test_repo_configuration.py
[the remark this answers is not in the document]
achals: I actually ended up leaving this as is - I think it's better to be comprehensive in the vary methods, and I think it reads okay - Lemme know if you feel strongly against

sdk/python/tests/integration/offline_store/test_universal_historical_retrieval.py
```diff
@@ -0,0 +1,309 @@
...
+        table_from_sql_entities = job_from_sql.to_arrow()
+        df_from_sql_entities = (
+            table_from_sql_entities.to_pandas()[expected_df.columns]
+            .sort_values(by=[event_timestamp, "order_id", "driver_id", "customer_id"])
+            .drop_duplicates()
+            .reset_index(drop=True)
+        )
+        assert_frame_equal(actual_df_from_sql_entities, df_from_sql_entities)
+
+    # timestamp_column = (
```
adchia, line 223: remove these?

sdk/python/tests/integration/offline_store/test_universal_historical_retrieval.py
```diff
@@ -0,0 +1,309 @@
...
+    #     features=[
+    #         "driver_stats:conv_rate",
+    #         "driver_stats:avg_daily_trips",
+    #         "customer_profile:current_balance",
+    #         "customer_profile:avg_passenger_count",
+    #         "customer_profile:lifetime_trip_count",
+    #     ],
+    # )
+
+    job_from_df = store.get_historical_features(
```
adchia, line 264: thoughts on splitting sql vs df into two separate tests with the same initial setup? would make it so one can distinguish from test failures whether their bug is in common logic or logic specific to sql vs df

sdk/python/tests/integration/offline_store/test_universal_historical_retrieval.py
adchia: or if you want to fix it, consider moving into separate test as I suggest below

adchia (approved):

sdk/python/tests/integration/offline_store/test_universal_historical_retrieval.py
achals: I think a single test is better - sql and df params are testing the same API roughly, and tests will run faster if there's fewer separate tests.

sdk/python/tests/integration/offline_store/test_universal_historical_retrieval.py
achals: I'm going to remove them because the behaviour isn't actually consistent across offline stores. File sources don't support sql queries at all.

PR #1784, opened 2021-08-17
Add Redis to the universal integration tests

achals:

Fixes #

```release-note
none
```

### DIFF
--- a/sdk/python/tests/integration/feature_repos/test_repo_configuration.py
+++ b/sdk/python/tests/integration/feature_repos/test_repo_configuration.py
@@ -158,6 +158,32 @@ def vary_infer_event_timestamp_col(
     return new_configs
 
 
+def vary_redis_online_store(configs: List[TestRepoConfig],) -> List[TestRepoConfig]:
+    new_configs = []
+    redis_config = {"type": "redis", "connection_string": "localhost:6379,db=0"}
+    for c in configs:
+        if (
+            "FileDataSourceCreator" in c.offline_store_creator
+            and "sqlite" in c.online_store
+        ):
+            new_configs.append(c)
+            new_configs.append(replace(c, online_store=redis_config))
+        elif (
+            "RedshiftDataSourceCreator" in c.offline_store_creator
+            and isinstance(c.online_store, Dict)
+            and c.online_store["type"] == "dynamodb"
+        ):
+            new_configs.append(c)
+            new_configs.append(replace(c, online_store=redis_config))
+        elif (
+            "BigQueryDataSourceCreator" in c.offline_store_creator
+            and c.online_store == "datastore"
+        ):
+            new_configs.append(c)
+            new_configs.append(replace(c, online_store=redis_config))
+    return new_configs
+
+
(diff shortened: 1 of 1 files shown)


sdk/python/tests/integration/feature_repos/test_repo_configuration.py
```diff
@@ -158,6 +158,32 @@ def vary_infer_event_timestamp_col(
...
+    redis_config = {"type": "redis", "connection_string": "localhost:6379,db=0"}
+    for c in configs:
+        if (
+            "FileDataSourceCreator" in c.offline_store_creator
+            and "sqlite" in c.online_store
+        ):
+            new_configs.append(c)
+            new_configs.append(replace(c, online_store=redis_config))
+        elif (
+            "RedshiftDataSourceCreator" in c.offline_store_creator
```
woop, line 172: I've not been following your implementation very closely, but what was the idea behind having redis configuration added based on conditional logic as opposed to having static configuration?

achals: There's a static list of base cases and methods that vary certain parameters. I opted for that to keep the static list small, the individual variations we're testing explicit, and flexible per integration test.

In this specific case I am fine adding it to the static list.

adchia: there is a benefit to having a static configuration somewhere so we can explicitly see somewhere these pairs of offline / online stores work well together, and anything else we don't really promise. 

in code, you're adding explicit whitelists anyways, so seems not too crazy to keep the static list the source of truth with offline / online store combos?

see my other comment above btw. i think in this case, you really are just adding a new redis config for each existing config anyways, so if its a convenience method to test every offline store with redis, seems like this could be a shorter method

sdk/python/tests/integration/feature_repos/test_repo_configuration.py
```diff
@@ -158,6 +158,32 @@ def vary_infer_event_timestamp_col(
     return new_configs
 
 
+def vary_redis_online_store(configs: List[TestRepoConfig],) -> List[TestRepoConfig]:
+    new_configs = []
+    redis_config = {"type": "redis", "connection_string": "localhost:6379,db=0"}
+    for c in configs:
+        if (
+            "FileDataSourceCreator" in c.offline_store_creator
```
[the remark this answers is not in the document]
adchia: also since this is the first method that processes the static list of configs, are these if statements even needed? you basically are taking every existing combo and adding redis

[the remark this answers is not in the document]
achals: Sure, I'll just make the redis config static and add to th list

sdk/python/tests/integration/feature_repos/test_repo_configuration.py
achals: Sg, I'll add this one to the static list itself

adchia (approved):

PR #1786, opened 2021-08-18
Fix get_online_features telemetry to only log every 10000 times

felixwang9817:

**What this PR does / why we need it**: Telemetry for `get_online_features` currently has a bug where telemetry is being logged during every call to `get_online_features`, instead of every 10000 calls. This PR fixes that bug. 

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/usage.py
+++ b/sdk/python/feast/usage.py
@@ -77,8 +77,8 @@ class Usage:
         self.check_env_and_configure()
         if self._usage_enabled and self.usage_id:
             if function_name == "get_online_features":
+                self._usage_counter["get_online_features"] += 1
                 if self._usage_counter["get_online_features"] % 10000 != 0:
-                    self._usage_counter["get_online_features"] += 1
                     return
 
             json = {
--- a/sdk/python/feast/usage.py
+++ b/sdk/python/feast/usage.py
@@ -78,9 +78,9 @@ class Usage:
         if self._usage_enabled and self.usage_id:
             if function_name == "get_online_features":
                 self._usage_counter["get_online_features"] += 1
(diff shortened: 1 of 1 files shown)


sdk/python/feast/usage.py
```diff
@@ -77,8 +77,8 @@ def log(self, function_name: str):
         self.check_env_and_configure()
         if self._usage_enabled and self.usage_id:
             if function_name == "get_online_features":
+                self._usage_counter["get_online_features"] += 1
                 if self._usage_counter["get_online_features"] % 10000 != 0:
-                    self._usage_counter["get_online_features"] += 1
```
achals, line 81: Nit, may be good to reset to 0 here to avoid the possibility of integer overflows. If things wrap around I guess that'd be fine too but we may as well take a little extra care

achals (approved):

sdk/python/feast/usage.py
woop: May also be better to use `% 10000 !=` some higher number, like `% 10000 != 1`. The process that loads the feature store may be serializing and deserializing the feature store object, or creating new ones using different workers. If the counter doesnt work then it may log constantly.

felixwang9817: @achals that makes sense.

felixwang9817: @woop I think the current implementation does what you want: the counter is incremented to 1 to start, so the first logging will happen on the 10000th call, not on the first call.

woop: Alright.

woop (approved):

PR #1790, opened 2021-08-18
Refactor Environment class and DataSourceCreator API, and use fixtures for datasets and data sources

### LINKED ISSUE #1797: Feature inferencing tests fail
## Expected Behavior 

Feature inferencing is enabled and works with the default universal feature repository.

## Current Behavior

When feature inferencing is enabled, many of the existing tests [fail](https://github.com/feast-dev/feast/runs/3395316140?check_suite_focus=true). It was necessary to [hardcode](https://github.com/feast-dev/feast/commit/459dae55aabb8d7eae89ae412cbf809e59c42941) feature schemas in the default repository in order to ensure that tests pass.

## Steps to reproduce

Remove the hardcoded feature columns here https://github.com/feast-dev/feast/commit/459dae55aabb8d7eae89ae412cbf809e59c42941 and rerun tests

achals:

This PR introduces a number of changes for speed and customizability:
- Removed `ds` and `data_source` from the Environment class, since they were used in an older version of the Environment.
- Added a DataSourceCache fixture to re-use staged data for speedups.
- Split out `construct_universal_test_environment` and `construct_test_environment` for both happy path integration tests and non-universal-repo integration tests.
- Removed the `get_prefixed_table_name` from the DataSourceCreator since it wasn't needed.

Additionally,
- Fixed schema inference and added test coverage on all sources.

Fixes #1797 

```release-note
none
```


sdk/python/tests/integration/feature_repos/test_repo_configuration.py
```diff
@@ -221,9 +244,13 @@ def construct_test_environment(
     offline_creator: DataSourceCreator = importer.get_class_from_type(
         module_name, config_class_name, "DataSourceCreator"
     )(project)
+
+    # This needs to be abstracted away for test_e2e_universal which uses a different dataset.
+    df = create_dataset()
     ds = offline_creator.create_data_sources(
-        project, df, field_mapping={"ts_1": "ts", "id": "driver_id"}
+        df, destination=project, field_mapping={"ts_1": "ts", "id": "driver_id"}
```
woop, line 251: Should this field mapping be hardcoded?

woop: Also, why would the destination be set to `project`? How can that be overriden? 

I think the key decision here is still, is `construct_test_environment` === `construct_universal_test_environment` in which case everything should be hardcoded but I as a user should be able to compose my own `custom_construct_test_environment`, or this function should be generalized. I would prefer the former.

sdk/python/tests/integration/feature_repos/universal/data_source_creator.py
```diff
@@ -28,5 +29,5 @@ def teardown(self):
         ...
 
     @abstractmethod
-    def get_prefixed_table_name(self, name: str, suffix: str) -> str:
+    def get_prefixed_table_name(self, table_name: str) -> str:
```
woop, line 32: Do we really need this method?

sdk/python/tests/integration/feature_repos/universal/data_sources/bigquery.py
```diff
@@ -42,14 +42,19 @@ def create_offline_store_config(self):
...
         df: pd.DataFrame,
+        destination: Optional[str] = None,
+        suffix: Optional[str] = None,
         event_timestamp_column="ts",
         created_timestamp_column="created_ts",
         field_mapping: Dict[str, str] = None,
         **kwargs,
     ) -> DataSource:
 
+        assert destination or suffix
```
woop, line 54: I really think these variables need names so that it's clear to users what a `destination` is and what `suffix` is meant to map to.

sdk/python/tests/integration/feature_repos/universal/data_sources/file.py
```diff
@@ -14,35 +14,46 @@
 
 
 class FileDataSourceCreator(DataSourceCreator):
-    f: Any
+    files: List[Any]
```
woop, line 17: Do we need to persist all of the files here? I don't understand how it's meant to be used. Aren't files created and a data source returned? Why do we need to keep it open? I can understand that the connection to the offline store or staging area should remain open, but not access to individual files.

sdk/python/tests/integration/feature_repos/test_repo_configuration.py
```diff
@@ -71,9 +72,63 @@ def ds_creator_path(cls: str):
...
+        drivers=entities["driver"],
+        start_date=end_time - timedelta(days=365),
+        end_date=end_time + timedelta(days=365),
+        order_count=1000,
+    )
+
+    return {"customer": customer_df, "driver": driver_df, "orders": orders_df}
+
+
+def construct_data_sources(
```
woop, line 99: should these functions maybe be prefixed with `universal_` since they aren't generalized?

sdk/python/feast/infra/utils/aws_utils.py
```diff
@@ -50,6 +55,11 @@ def get_bucket_and_key(s3_path: str) -> Tuple[str, str]:
     return bucket, key
 
 
+@retry(
```
woop, line 58: what's happening here?

achals: ignore this

sdk/python/tests/integration/feature_repos/test_repo_configuration.py
```diff
@@ -1,16 +1,17 @@
 import tempfile
```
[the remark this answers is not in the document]
woop: Yea I also had that comment. Agree.

sdk/python/tests/integration/feature_repos/test_repo_configuration.py
```diff
@@ -71,91 +73,140 @@ def ds_creator_path(cls: str):
...
+
+
+def setup_datasets(
+    environment: "Environment",
+    datasets_override: Optional[Dict[str, pd.DataFrame]] = None,
+) -> "Environment":
+    environment.datasets = (
+        datasets_override
+        if datasets_override
+        else construct_universal_datasets(
```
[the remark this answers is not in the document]
woop: > if you don't, it uses entities (which might have been overwritten previously) and then you could run into an error if you don't have customer or driver entities.

Yea, but in which case would you override the entities but not override the steps that follow (like `setup_datasets`)? Presumably you as the author would know that there are downstream effects.

I think the real question is what the value of `setup_datasets` is in the first place. It seems like we can get away with having `construct_universal_datasets` which seems useful, and otherwise just letting people operate on the environment object directly with custom code. This function isn't really doing much tbh.

> Could we almost do some sort of like builder approach? where basically nothing is assigned until a .build() method is called. Then we'd have only really 3 usage patterns

Getting some real AutoValue/Lombok vibes here :)

> Environment.builder().setEntities(...).setDatasets(...).setFeatureViews(...).build() -> force the user to explicitly define all the objects if they need something fully custom
How would this be different from a constructor like Environment(entities=.., datasets=...)?

> Environment.getDefault().toBuilder().setEntities(...).build() -> explicitly communicates that there are other parts of the environment that are "default" that may or may not be compatible with the entities

This seems very close to having
```
e = Environment(entities=.., datasets=...)
# My custom code
e.entities = do_my_custom_stuff()
e.setup()
```
To me the main value of the builder pattern in this context is that we can delay the actual execution/staging. I'm not sure if the getDefault/toBuilder logic is as important.

sdk/python/tests/integration/e2e/test_universal_e2e.py
```diff
@@ -3,27 +3,49 @@
...
+    "config", vary_infer_feature(FULL_REPO_CONFIGS), ids=lambda v: str(v)
+)
+def test_e2e_consistency(config, data_source_cache: DataSourceCache):
+    df = create_dataset()
+    test_suite_name = "test_e2e_consistency"
+    with construct_test_environment(test_suite_name, config) as test_environment:
+        fs = test_environment.feature_store
+        infer_features = test_environment.test_repo_config.infer_features
+        key = f"{test_suite_name}_{test_environment.test_repo_config.offline_store_creator}"
+        _, _, data_source, _ = data_source_cache.get_or_create(
```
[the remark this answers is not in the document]
woop: > Can you expand on that you think the fixture should contain? Should it just be the dataframes? Or also data sources?

It would only return environments.

It's possible to do something like

```
environment_configs = [
    SomeEnvConfigThing(provider="local"),
    SomeEnvConfigThing(provider="gcp"),
    SomeEnvConfigThing(provider="aws"),
]


@pytest.fixture(params=environment_configs, scope="module")
def environment(request):
    e = Environment(config=request.param)
    e = stage_data(e)
    e = apply(e)
    e = materialize(e)
    yield e
    e.teardown()
```

which could then be accessed like

```
def test_online_retrieval(environment):
    # do stuff
```

`test_online_retrieval` would be called once per environment. We could then also do things like

````
@pytest.mark.parametrize('some_other_flag', [True, False])
def test_online_retrieval(environment, some_other_flag):
    # do stuff
```

Which would run `test_online_retrieval` twice per env, with `some_other_flag` toggled on and off. But importantly the environment is only created once.

Another approach is to do something like

```
@pytest.fixture()
def bigquery_universal_datasource():
    from universal_repo import driver_df, driver_fv, customer_df, customer_fv

    dsc = BigQueryDataSourceCreator()
    dsc.create_data_source(driver_df, driver_fv)
    dsc.create_data_source(customer_df, customer_fv)
    yield dsc
    dsc.teardown()


@pytest.mark.parametrize(
    "provider,data_source,online_store,full_feature_names",
    [
        (
            "local",
            pytest.lazy_fixture("local_universal_datasource"),
            "sqlite",
            True
        ),
        (
            "local",
            pytest.lazy_fixture("local_universal_datasource"),
            "sqlite",
            False
        ),
        (
            "local",
            pytest.lazy_fixture("local_universal_datasource"),
            "redis",
            True
        ),
        (
            "local",
            pytest.lazy_fixture("local_universal_datasource"),
            "dymamodb",
            True
        ),
        (
            "local",
            pytest.lazy_fixture("local_universal_datasource"),
            "datastore",
            True
        ),
        (
            "gcp",
            pytest.lazy_fixture("bigquery_universal_datasource"),
            "sqlite",
            True
        ),
        (
            "aws",
            pytest.lazy_fixture("redshift_universal_datasource"),
            "sqlite",
            True
        )
    ],
)
def test_online_retrieval(provider, data_source, online_store, full_feature_names):
    with Environment(config(provider, data_source, online_store, full_feature_names) as e:
        # do stuff
```
The above API isn't functional, but the idea is that we'd use `lazy_fixtures` and scope the fixtures to session/module/function depending on if we want isolation or reuse across all tests. So we're essentially leveraging the caching that is built into Pytest since it won't redo the staging of data, and it's 100% clear what the test cases for a specific test are.

And if you as a test creator need to do anything off the beaten path then you can set up your own DataSourceCreator and feed it whatever datasets you want (and use a different repo than universal repo)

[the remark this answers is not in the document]
woop: > I think with the universal tests, there's many different tests that operate on what is basically the same data. So even if the test functions are different, there's advantages in sharing the data sources - for me it's equivalent to having the test data uploaded to redshift/bigquery outside of the tests and used directly with the assumptions that they exist. We would be sharing the same data, without needing different data sets for different tests.

100% agree with solving this problem, just want to keep things as simple as possible.

> There is an issue where both universal and non-universal sources are cached in the same container - which I can see value in keeping separate. Maybe I can create a different cache that's session scoped for universal tests, and function scoped for non-universal tests?

Well if we go with the cache approach, isn't it better to give control to the test writer? They can use keys to purposefully select or miss existing cache items, as long as the cache keys are named according to some standard.

sdk/python/tests/integration/feature_repos/test_repo_configuration.py
```diff
@@ -71,91 +73,140 @@ def ds_creator_path(cls: str):
...
+    return {
+        "customer": create_customer_daily_profile_feature_view(
+            data_sources["customer"]
+        ),
+        "driver": create_driver_hourly_stats_feature_view(data_sources["driver"]),
+    }
+
+
+def setup_entities(
+    environment: "Environment", entities_override: Optional[Dict[str, List[Any]]] = None
```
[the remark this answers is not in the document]
achals: Because it was a forward reference; but i've removed this func

sdk/python/tests/integration/feature_repos/universal/data_sources/bigquery.py
```diff
@@ -12,21 +13,27 @@
...
+    def get_dataset(self):
+        if not self.dataset:
+            self.dataset = bigquery.Dataset(self.dataset_id)
+            print(f"Creating dataset: {self.dataset_id}")
+            self.client.create_dataset(self.dataset, exists_ok=True)
+            self.dataset.default_table_expiration_ms = (
+                1000 * 60 * 60 * 24 * 14
+            )  # 2 weeks in milliseconds
+            self.client.update_dataset(self.dataset, ["default_table_expiration_ms"])
+        return self.dataset
```
adchia, line 35: nit: create_dataset instead? and doesn't seem like you use the returned self.dataset anywhere

adchia (commented):
/lgtm

sdk/python/feast/infra/provider.py
```diff
@@ -238,9 +238,11 @@ def _get_column_names(
             reverse_field_mapping[col] if col in reverse_field_mapping.keys() else col
             for col in feature_names
         ]
+    _feature_names = set(feature_names) - set(join_keys)
```
woop, line 241: Nitpick

I think it would be helpful if you
1. Added a comment above this code section to describe what you are doing
2. Perhaps have the `list(_feature_names)` outside of the return in order to encapsulate the logic a bit more cleanly

sdk/python/tests/integration/e2e/test_universal_e2e.py
```diff
@@ -3,27 +3,45 @@
...
-def test_e2e_consistency(test_environment: Environment):
-    fs, fv = (
-        test_environment.feature_store,
-        driver_feature_view(test_environment.data_source),
-    )
-    entity = driver()
-    fs.apply([fv, entity])
+@pytest.mark.integration
+@pytest.mark.parametrize(
+    "config", vary_infer_feature(FULL_REPO_CONFIGS), ids=lambda v: str(v)
```
woop, line 22: Question: Do we still need any of the `vary` methods or can they be replaced with `@pytest.mark.parametrize`? The reason I ask is because it seems like `vary_infer_feature` is doing the same thing as `@pytest.mark.parametrize` on a flag.

sdk/python/tests/integration/e2e/test_universal_e2e.py
```diff
@@ -3,27 +3,45 @@
...
+        fs = test_environment.feature_store
+        infer_features = test_environment.test_repo_config.infer_features
+        _, _, data_source, _ = (
+            None,
+            df,
+            test_environment.data_source_creator.create_data_source(
+                df, fs.project, field_mapping={"ts_1": "ts", "id": "driver_id"}
+            ),
+            test_environment.data_source_creator,
+        )
```
woop, line 37: It's a little bit unclear what is happening here. Why do we create a tuple if we can create the `data_source` directly?

sdk/python/tests/integration/feature_repos/repo_configuration.py
```diff
@@ -0,0 +1,242 @@
...
+@dataclass(frozen=True, repr=True)
+class IntegrationTestRepoConfig:
+    """
+    This class should hold all possible parameters that may need to be varied by individual tests.
+    """
+
+    provider: str = "local"
+    online_store: Union[str, Dict] = "sqlite"
+
+    offline_store_creator: str = "tests.integration.feature_repos.universal.data_sources.file.FileDataSourceCreator"
```
woop, line 31: Apologies if this has been mentioned before, but can you remind me why we use a string reference to the class instead of a code reference?

sdk/python/tests/integration/feature_repos/universal/data_source_creator.py
```diff
@@ -11,12 +11,26 @@ class DataSourceCreator(ABC):
...
         event_timestamp_column="ts",
         created_timestamp_column="created_ts",
         field_mapping: Dict[str, str] = None,
     ) -> DataSource:
+        """
+        Create a data source based on the dataframe. Implementing this method requires the underlying implementation to
+        persist the dataframe in offline store, using the destination string as a way to differentiate multiple
+        dataframes and data sources.
+
+        :param df: The dataframe to be used to create the data source.
```
woop, line 25: Nitpick: These are not Google style docstrings. Btw you can change your PyCharm to automatically generate the correct format.

sdk/python/tests/integration/feature_repos/universal/data_source_creator.py
```diff
@@ -11,12 +11,26 @@ class DataSourceCreator(ABC):
...
+        :param destination: The destination that this data frame is meant for. This str is used by the implementing
+            classes to isolate the multiple dataframes from each other.
```
woop, line 27: By the way `The destination that this data frame is meant for` isn't super descriptive.`, but from reading `This str is used by the implementing classes to isolate the multiple dataframes from each other.` it seems like `destination` is the dataframe name? Would that be more descriptive as a variable? Just want to make sure it's clear what the variable is used for by looking at the name

sdk/python/tests/integration/feature_repos/universal/data_sources/bigquery.py
achals: Yeah I can change

sdk/python/feast/infra/provider.py
achals: Sure done

sdk/python/tests/integration/e2e/test_universal_e2e.py
achals: I think they can be removed. Don't think we're even using them anywhere.

sdk/python/tests/integration/e2e/test_universal_e2e.py
achals: Clean up

sdk/python/tests/integration/feature_repos/repo_configuration.py
woop: Is there any downside to having that code dependency?

achals: Oops missed this. For tests, presumably not. However I can imagine we'll want this to be extensible to support the test suite for when users bring their own offline stores with a custom DataSourceCreator, so it may be needed then. Happy to do either.

woop: My personal preference is to YAGNI this baby until we need it, just because I feel like we might run into bugs with these strings.

sdk/python/feast/feature_view.py
```diff
@@ -325,6 +325,28 @@ def infer_features_from_batch_source(self, config: RepoConfig):
                 self.batch_source.created_timestamp_column,
             } | set(self.entities)
 
+            if (
```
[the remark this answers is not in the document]
achals: Those are already excluded in a section ablove this one.

sdk/python/tests/integration/feature_repos/repo_configuration.py
achals: Aight sg

woop (approved):

PR #1796, opened 2021-08-22
Add init to Provider contract

woop:

This PR adds an abstract `__init__` as part of the base class for Providers. Currently no such method exists which has led to custom providers and built in providers having different initialization contracts.
https://github.com/feast-dev/feast/blob/745a1b43d20c0169b675b1f28039854205fb8180/sdk/python/feast/infra/provider.py#L146
vs 
https://github.com/feast-dev/feast/blob/745a1b43d20c0169b675b1f28039854205fb8180/sdk/python/feast/infra/provider.py#L164

```release-note
All custom providers must now provide an initialization method and won't receive a repo_path argument
```

### DIFF
--- a/sdk/python/feast/infra/provider.py
+++ b/sdk/python/feast/infra/provider.py
@@ -21,6 +21,10 @@ from feast.type_map import python_value_to_proto_value
 
 class Provider(abc.ABC):
     @abc.abstractmethod
+    def __init__(self, config: RepoConfig):
+        ...
+
+    @abc.abstractmethod
     def update_infra(
         self,
         project: str,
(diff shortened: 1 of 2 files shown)


sdk/python/feast/infra/provider.py
```diff
@@ -20,6 +20,10 @@
 
 
 class Provider(abc.ABC):
+    @abc.abstractmethod
+    def __init__(self, config: RepoConfig):
```
achals, line 24: Won't we break existing third party providers?

woop: Yes.

achals: Should we also update the call site to not pass in repo_path when loading custom providers?

woop: Yea, good catch.

PR #1804, opened 2021-08-26
Enable entityless featureviews

codyjlin:

This PR executes on the [Entityless FeatureViews RFC](https://docs.google.com/document/d/1mOZ-a6Voq43JcsPu2BIaZEEa3FJ6RhyFxs2CZDbby-U/edit#), with minor adjustments to Implementation 1:
1. Registry
    1. Create `__entityless` entity on every apply
    2. Attach it to any entityless FVs when they are initialized
    3. Hide it from `registry_dump` (namely, `list_entities` and `list_featureviews`)
2. Offline
    1. For each of the offline store providers, update the joins to be able to join with no entities (only the timestamp col)
    2. (I think this implementation detail differs from the [RFC](https://docs.google.com/document/d/1mOZ-a6Voq43JcsPu2BIaZEEa3FJ6RhyFxs2CZDbby-U/edit#heading=h.9gaqqtox9jg6) - I think it is much simpler than having to create the extra `__entityless_id` dummy column to join on a constant value, and then having to hide it later when selecting columns)
3. Materialization/Online
    1. Add `__entityless_id` column as empty string upon materialize (namely, `pull_latest_from_table_or_query` in each offline store provider)
    2. If `get_online_features` requests any features from an entityless FV, `{"__entityless_id": ""}` will be a `join_key_row` to be converted to a `entity_row_proto` before look up in the online store.

Fixes ##1737

```release-note
The user can now specify FeatureViews with an empty `entities` list.
```


sdk/python/feast/feature_view.py
```diff
@@ -186,14 +190,11 @@ def is_valid(self):
         Validates the state of this feature view locally.
 
         Raises:
-            ValueError: The feature view does not have a name or does not have entities.
```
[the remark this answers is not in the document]
codyjlin: ah yeah good catch.

sdk/python/feast/feature_store.py
```diff
@@ -92,6 +98,10 @@ def __init__(
             repo_path=self.repo_path,
             cache_ttl=timedelta(seconds=registry_config.cache_ttl_seconds),
         )
+        entityless_entity = Entity(
+            name="__entityless", join_key="__entityless_id", value_type=ValueType.INT32,
```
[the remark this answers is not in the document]
codyjlin: thanks for the catch

sdk/python/feast/feature_store.py
```diff
@@ -92,6 +98,10 @@ def __init__(
...
+        )
+        self.apply(entityless_entity, commit=False)
```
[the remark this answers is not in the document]
codyjlin: yeah you're totally right. The test I just changed in the test_e2e_local actually caught this case.

sdk/python/feast/feature_store.py
```diff
@@ -159,14 +174,39 @@ def list_feature_services(self) -> List[FeatureService]:
...
+
+        Args:
+            allow_cache: Whether to allow returning entities from a cached registry.
+
+        Returns:
+            A list of feature views.
+        """
+        return self.__list_feature_views(allow_cache)
+
+    def __list_feature_views(
```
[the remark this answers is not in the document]
codyjlin: hmmm I'm actually not sure what the best practice is, but I'll switch it to single to stay consistent with other "private" methods in the class. (https://docs.python.org/3/tutorial/classes.html#private-variables)

sdk/python/tests/example_repos/example_feature_repo_2.py
```diff
@@ -24,3 +24,24 @@
...
+    path="%PARQUET_PATH_GLOBAL%",  # placeholder to be replaced by the test
+    event_timestamp_column="event_timestamp",
+    created_timestamp_column="created",
+)
+
+
+global_stats_feature_view = FeatureView(
+    name="global_daily_stats",
+    entities=[],
+    ttl=Duration(seconds=86400 * 1),
```
woop, line 39: please use timedelta.

sdk/python/tests/integration/e2e/test_universal_e2e.py
```diff
@@ -45,36 +53,62 @@ def check_offline_and_online_features(
             assert abs(response_dict[f"{fv.name}__value"][0] - expected_value) < 1e-6
         else:
             assert response_dict[f"{fv.name}__value"][0] is None
+        if expected_entityless_value:
```
woop, line 56: Is there any way in which we can avoid more and more conditional logic in our tests? It makes understanding the tests very hard.

codyjlin: Not sure what the best way to refactor into different tests would be, but I definitely agree tests keep having more conditional logic. Happy to hear @achals' thoughts though, I think the refactoring of the other 2 `test_universal_historical_retrieval.py` and `test_universal_online.py` are both great (and maybe it's the case that we don't need to make any changes to this file if the coverage there is good enough for the entityless case)

sdk/python/tests/integration/online_store/test_universal_online.py
```diff
@@ -127,6 +140,7 @@ def get_latest_feature_values_from_dataframes(driver_df, customer_df, entity_row
     latest_customer_row = customer_rows.loc[
         customer_rows["event_timestamp"].idxmax()
     ].to_dict()
+    latest_global_row = global_df.loc[global_df["event_timestamp"].idxmax()].to_dict()
 
     latest_customer_row.update(latest_driver_row)
-    return latest_customer_row
+    return {**latest_customer_row, **latest_driver_row, **latest_global_row}
```
felixwang9817, line 146: If you're merging the dictionaries like this, you can remove the line above.

sdk/python/tests/conftest.py
```diff
@@ -126,10 +126,16 @@ def universal_data_sources(environment):
 def e2e_data_sources(environment: Environment):
     df = create_dataset()
     data_source = environment.data_source_creator.create_data_source(
-        df, environment.feature_store.project, field_mapping={"ts_1": "ts"},
+        df, destination_name="data_source", field_mapping={"ts_1": "ts"},
```
felixwang9817, line 129: Why are we changing the destination name to `"data_source"`?

sdk/python/tests/integration/online_store/test_universal_online.py
codyjlin: my mistake, forgot to remove that line

sdk/python/tests/conftest.py
codyjlin: didn't want this name to conflict with the "entityless_data_source" below, wanted more consistent naming than just reusing the project name

sdk/python/feast/feature_store.py
```diff
@@ -226,7 +265,13 @@ def get_feature_view(self, name: str) -> FeatureView:
         Raises:
             FeatureViewNotFoundException: The feature view could not be found.
         """
-        return self._registry.get_feature_view(name, self.project)
+        return self._get_feature_view(name)
+
+    def _get_feature_view(self, name: str, hide_entityless: bool = True) -> FeatureView:
```
woop, line 270: why do we need a new method here?

sdk/python/feast/feature_store.py
```diff
@@ -161,14 +185,29 @@ def list_feature_services(self) -> List[FeatureService]:
...
+        Args:
+            allow_cache: Whether to allow returning entities from a cached registry.
+
         Returns:
             A list of feature views.
         """
-        return self._registry.list_feature_views(self.project)
+        return self._list_feature_views(allow_cache)
+
+    def _list_feature_views(
```
woop, line 200: is there a reason for having a new method here?

sdk/python/tests/integration/e2e/test_universal_e2e.py
```diff
@@ -8,34 +8,42 @@
...
+    global_feature_view,
+)
 
 
 @pytest.mark.integration
 @pytest.mark.parametrize("infer_features", [True, False])
 def test_e2e_consistency(environment, e2e_data_sources, infer_features):
     fs = environment.feature_store
-    df, data_source = e2e_data_sources
+    df, data_source, entityless_data_source = e2e_data_sources
```
woop, line 21: There are a few confusing things here
* Where is `df` used? I know you didn't introduce it, but it's probably worth figuring out what the intention is behind `e2e_data_sources`.
* It's not clear to me how `e2e_data_sources` is supposed to be different from the universal data sources. Is there anything preventing us from adding the entityless df/view/source as a part of the universal repo?

sdk/python/feast/feature_store.py
felixwang9817: This method is here because we want `FeatureStore`, not `Registry`, to handle all entityless logic. We also don't want the public method `list_feature_views` to expose `hide_entityless`, so we decided to add this private method.

codyjlin: I think we could expect users to use `FeatureStore.list_entities()` (or `list_featureviews()`) to examine the registry as well. So we want to hide it there as well.

sdk/python/tests/conftest.py
achals: Can you leave this one as is, and change the one down below to `{project}.entityless_data_source`? That will ensure that we're trying these data sources to the corresponding project, which is necessary to track down failures and such.

sdk/python/tests/integration/e2e/test_universal_e2e.py
achals: `df` isn't used in this case.

`e2e_data_sources` is a data source that is created with a handful of entries with specific timestamps, specifically for the purpose of validating the point in time correctness of the feature retrievals in the online and offline case. 

The universal data sources are generated randomly, and with an arbitrary number of records that can be specified by users. but there's no logic for time generating records with specific kinds of relative timestamps.

We should consolidate the two datasets for our internal tests, but I do imagine that we may still have the need for multiple kinds of datasets to check the correctness of specific features.

For this specific PR, do we need to add any of the time correctness checks for entityless feature views? If not, I'd leave this file untouched.

sdk/python/tests/conftest.py
achals: In fact, I think that we should leave this data set as is and add the entityless datasources to the universal data soures (line 111).

We should then update `test_historical_retreival` and `test_universal_online` to look up values from there.

The e2e datasets have specific timestamps to assert point in time correctness, and I don't know if that needs to be validated for entityless features, since we're always looking at the latest state of the features (but please correct me if I'm wrong).

sdk/python/feast/feature_view.py
```diff
@@ -40,6 +40,10 @@
 
 warnings.simplefilter("once", DeprecationWarning)
 
+ENTITYLESS_ENTITY_ID = "__entityless_id"
+ENTITYLESS_ENTITY_NAME = "__entityless"
+ENTITYLESS_ENTITY_VAL = ""
```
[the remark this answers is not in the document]
codyjlin: What would be your suggestion for a name? What about "dummy_entity"?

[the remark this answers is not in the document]
achals: I would personally prefer that, yes - and a docstring to maybe explain what's dummy about them.

sdk/python/tests/conftest.py
codyjlin: Yeah, I was debating whether `test_historical` and `test_online` would be enough coverage for this case - I think your explanation makes sense though. I'll remove changes to the universal_e2e

sdk/python/tests/integration/e2e/test_universal_e2e.py
codyjlin: Yeah that sounds good to me. I think `test_universal_historical` and `test_universal_online` will have enough coverage.

sdk/python/feast/driver_test_data.py
```diff
@@ -198,3 +198,41 @@ def create_customer_daily_profile_df(customers, start_date, end_date) -> pd.Data
     # TODO: Remove created timestamp in order to test whether its really optional
     df_all_customers["created"] = pd.to_datetime(pd.Timestamp.now(tz=None).round("ms"))
     return df_all_customers
+
+
+def create_global_daily_stats_df(start_date, end_date) -> pd.DataFrame:
```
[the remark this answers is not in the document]
codyjlin: Do you mean that instead of `create_global_daily_stats_df` creating its own dataset of columns `event_timestamp`, `num_rides`, `avg_ride_length`, and `created`, that I could use one of the others like `create_driver_hourly_stats_df` and just remove `driver_id`?

Wouldn’t that be confusing when it comes to using the same feature names? (and could break some tests that parametrize with `full_feature_names=False`). I think it could make sense to just have the separate, new entityless dataset to represent an example of entityless features.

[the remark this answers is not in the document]
woop: Fair enough. Lets stick to this approach.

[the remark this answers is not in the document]
achals: @codyjlin and I talked offline - fine to stick with this approach as well.

sdk/python/feast/feature_store.py
```diff
@@ -161,14 +185,29 @@ def list_feature_services(self) -> List[FeatureService]:
...
+        self, allow_cache: bool = False, hide_dummy_entity: bool = True
+    ) -> List[FeatureView]:
+        feature_views = []
+        for fv in self._registry.list_feature_views(
+            self.project, allow_cache=allow_cache
+        ):
+            if hide_dummy_entity and fv.entities[0] == DUMMY_ENTITY_NAME:
+                fv.entities = []
+            feature_views.append(fv)
```
achals, line 209: Is there the possibility that we return to `feature_views`, but this list gets re-applied using `FeatureStore.apply` and we lose the entity information?

sdk/python/usage_tests/test_usage.py
```diff
@@ -134,7 +134,7 @@ def test_exception_usage_off():
 @retry(wait=wait_exponential(multiplier=1, min=1, max=10), stop=stop_after_attempt(7))
 def ensure_bigquery_usage_id_with_retry(usage_id):
     rows = read_bigquery_usage_id(usage_id)
-    if rows.total_rows != 1:
+    if rows.total_rows == 0:
```
achals, line 137: Was this the issue with the telemetery jobs, double counting the usage ids?

achals (approved):
/lgtm

sdk/python/usage_tests/test_usage.py
codyjlin: yep exactly. I think it's due to an additional row from the auto apply of the dummy entity in the feature store initialization.

PR #1812, opened 2021-08-29
Registry store plugin

DvirDukhan:
Main feature: This PR allows adding third-party registry stores, to allow saving the registry in custom storage services other than s3/gcs/files. 
Additional changes:
1. Moved aws/gcp/local RegistryStore implementation to their respected files (together with their providers)
2. `Registry` class is now being initialized with `RegistryConfig`.

None

```release-note
This PR adds a new optional field under the registry configuration, called `registry_store_provider`.  This allows loading a third-party class, implementing `RegistryStore`.  The third-party class name must end with `RegisrtryStore`, for example `foo.registry_store.FooRegistryStore`.
```


achals:
I'm going to review this today.

sdk/python/feast/repo_config.py
```diff
@@ -52,6 +52,9 @@ class Config:
 class RegistryConfig(FeastBaseModel):
     """ Metadata Store Configuration. Configuration that relates to reading from and writing to the Feast registry."""
 
+    registry_store_provider: Optional[StrictStr]
```
woop, line 55: Reusing the name `provider` here is quite confusing since we have a `provider` concept already. Is it possible for us to infer the backend based on the URI?

DvirDukhan:
Rebased and addressed comments.

sdk/python/feast/registry.py
```diff
@@ -54,31 +51,77 @@ class Registry:
...
+                elif uri.scheme == "s3":
+                    from feast.infra.aws import S3RegistryStore
+
+                    self._registry_store = S3RegistryStore(registry_path)
+                elif uri.scheme == "file" or uri.scheme == "":
+                    from feast.infra.local import LocalRegistryStore
+
+                    self._registry_store = LocalRegistryStore(
+                        repo_path=repo_path, registry_path_string=registry_path
+                    )
```
achals, line 102: When would we hit this case? When users specified `registry_store_type: S3RegistryStore`, something like that?

I think that in that case, we probably want to have a dict of the "shortcut" types mapping to the fully qualified type (like in https://github.com/feast-dev/feast/blob/master/sdk/python/feast/repo_config.py#L22), and use the dynamic `importer.get_class_from_type` path.

achals (changes requested):
Sorry for the second back and forth @DvirDukhan , but I think that's the only other major concern I have. If we address that then i'm happy to lgtm

sdk/python/feast/registry_store.py
```diff
@@ -0,0 +1,39 @@
...
+
+        Args:
+            registry_proto: the new RegistryProto
+        """
+        pass
+
+    @abstractmethod
+    def teardown(self):
+        """
+        Tear down all resources.
```
woop, line 37: All resources or the registry?

sdk/python/feast/registry_store.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class RegistryStore(ABC):
+    """
+    RegistryStore: abstract base class implemented by specific backends (local file system, GCS)
+    containing lower level methods used by the Registry class that are backend-specific.
+    """
+
+    @abstractmethod
+    def get_registry_proto(self):
```
woop, line 13: Should the return type be a part of this method?

sdk/python/feast/registry_store.py
```diff
@@ -0,0 +1,39 @@
+from abc import ABC, abstractmethod
+
+from feast.protos.feast.core.Registry_pb2 import Registry as RegistryProto
+
+
+class RegistryStore(ABC):
+    """
+    RegistryStore: abstract base class implemented by specific backends (local file system, GCS)
```
woop, line 8: This comment can be made tighter. We already know the class is `RegistryStore`, and an abstract base class, so that doesn't need to be repeated.

What about

```
A registry store is a storage backend for the Feast registry.
```

sdk/python/feast/registry.py
DvirDukhan: @achals thanks for the review.
I did the modifications, however, since there is a single constructor call `self._registry_store = cls(registry_config, repo_path)` this enforces `S3RegistryStore` and `GCSRegistryStore` constructors to have the signature:
`def __init__(self, registry_config: RegistryConfig, repo_path: Path):`
This is a bit awkward IMO since `repo_path` is not being used in those classes, it is just for the `LocalRegistryStore`
Let me know what you think if there is a better approach to consolidate those constructors.

achals: I think it's better to have a single interface even if it means passing in an unused parameter in some of the implementations, so I'd be okay with that change.

DvirDukhan: thanks, @achals. The change is already in the latest modification on the PR.

achals (approved):
/lgtm

achals:
Thanks @DvirDukhan ! The change looks good to me.

PR #1814, opened 2021-08-30
Use get_multi instead of get for datastore reads

### LINKED ISSUE #1759: Datastore online request makes a call once for each entity
There are performance issues with Datastore when fetching features for multiple entities. An individual fetch request is made for each entity 

I would like to have a more efficient access to the Datastore for multiple entities. The python api for datastore offers a method `get_multi()` which allows to fetch multiple keys in one request.

---

**Additional context**

I have done some basic tests using the `get_multi` vs get and it seems that a improvement can be made by using the function `get_multi`

Here is part of code  I used for testing:

```
import numpy as np
import timeit
from google.cloud import datastore

def multi_online_read(num_keys=5):

    client = datastore.Client(
                    project=project_name, namespace=None,
                )
    
    feast_project = "feature_repo"
    key = client.key("Project", feast_project, "Table", table, "Row",  row)
    keys = []
    response = []
    for i in range(num_keys):
        keys.append(key)

    response = client.get_multi(keys)

    return response

def online_read(num_keys=5):
    client = datastore.Client(
                    project=project_name, namespace=None,
                )
    
    feast_project = "feature_repo"
    key = client.key("Project", feast_project, "Table", table, "Row", row)
    
    response = []
    for i in range(num_keys):
        response.append(client.get(key))


    return response

if __name__ == "__main__":

    results = timeit.repeat(online_read, repeat=10, number=1)
    print(results)
    print(np.mean(results))

    results = timeit.repeat(multi_online_read, repeat=10, number=1)
    print(results)
    print(np.mean(results))
```

achals:

get is a wrapper around get_multi, and get_multi reduces the network costs involved.

Fixes #1759 

```release-note
Use a batch lookup API for Google Datastore. Users should see lower latencies for `get_online_features` when using Datastore as the online store.
```

### DIFF
--- a/sdk/python/feast/infra/online_stores/datastore.py
+++ b/sdk/python/feast/infra/online_stores/datastore.py
@@ -30,6 +30,7 @@ from feast.repo_config import FeastConfigBaseModel, RepoConfig
 try:
     from google.auth.exceptions import DefaultCredentialsError
     from google.cloud import datastore
+    from google.cloud.datastore.client import Key
 except ImportError as e:
     from feast.errors import FeastExtrasDependencyImportError, FeastProviderLoginError
 
@@ -233,13 +234,18 @@ class DatastoreOnlineStore(OnlineStore):
 
         feast_project = config.project
 
+        keys: List[Key] = []
         result: List[Tuple[Optional[datetime], Optional[Dict[str, ValueProto]]]] = []
         for entity_key in entity_keys:
             document_id = compute_entity_id(entity_key)
             key = client.key(
                 "Project", feast_project, "Table", table.name, "Row", document_id
             )
-            value = client.get(key)
+            keys.append(key)
+
+        values = client.get_multi(keys)
+
+        for value in values:
             if value is not None:
                 res = {}
                 for feature_name, value_bin in value["values"].items():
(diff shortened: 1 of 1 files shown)


woop:
What is the performance difference between these two approaches? Have you run a benchmark?

achals:
> What is the performance difference between these two approaches? Have you run a benchmark?

retrieving 10 records, using pytest-benchmark

----------------------------------------------------------------------------------------- benchmark: 2 tests ----------------------------------------------------------------------------------------
Name (time in s)                              Min               Max              Mean            StdDev            Median               IQR            Outliers     OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_online_retrieval_datastore_batch      2.8716 (1.0)      3.2226 (1.0)      2.9943 (1.0)      0.0986 (1.0)      2.9681 (1.0)      0.1220 (1.0)           8;0  0.3340 (1.0)          30           5
test_online_retrieval_datastore_single     4.0147 (1.40)     4.6170 (1.43)     4.2247 (1.41)     0.1446 (1.47)     4.2366 (1.43)     0.1492 (1.22)          9;3  0.2367 (0.71)         30           5
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

felixwang9817 (approved):
/lgtm

PR #1824, opened 2021-09-01
Implementing initial on demand transforms for historical retrieval to_df

adchia:
Sets up a limited on demand transform for historical retrieval for to_df (not to_arrow yet)

Fixes #

```release-note
- Add support for row-level on-demand transformations of feature view features in get_offline_features.
- Requires offline store implementations to now also have full_feature_name and on_demand_feature_views properties
```

### DIFF
--- a/sdk/python/feast/infra/offline_stores/offline_store.py
+++ b/sdk/python/feast/infra/offline_stores/offline_store.py
@@ -20,6 +20,7 @@ import pyarrow
 
 from feast.data_source import DataSource
 from feast.feature_view import FeatureView
+from feast.on_demand_feature_view import OnDemandFeatureView
 from feast.registry import Registry
 from feast.repo_config import RepoConfig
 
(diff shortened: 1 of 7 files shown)


adchia:
> Now that #1803 has been merged in, can you update this diff to only have the changes for the historical retrieval API?

Done!

sdk/python/feast/infra/offline_stores/offline_store.py
```diff
@@ -20,18 +20,42 @@
...
 
+    @property
     @abstractmethod
+    def full_feature_names(self) -> bool:
+        pass
+
+    @property
+    @abstractmethod
+    def on_demand_feature_views(self) -> Optional[List[OnDemandFeatureView]]:
+        pass
```
achals, line 39: Why are we changing the API for the base class here and adding these fields? Are they necessary?

achals (commented):
It seems like the on demand feature views are not being consumed by the jobs anywhere - is that intentional?

sdk/python/feast/infra/offline_stores/offline_store.py
achals: NVM, I see why they're needed now. While I think this is okay for a first pass I want to see if we can get away from adding these additional fields on to the job. I think they are pretty inelegant for a user to have to implement themselves.

achals (approved):
/lgtm

PR #1827, opened 2021-09-02
Reducing wait interval for BQ integration tests

adchia:
From some quick logging, a hotspot when running integration tests lies in uploading entity_df for BQ related tests. This reduces the polling interval from 10 seconds to 0.1 seconds in test environments (and also reduce the prod environment interval to 1s since the main cost is catching a raised Exception, but there's nothing else happening on this thread)

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -207,7 +207,7 @@ def block_until_done(
     client: Client,
     bq_job: Union[bigquery.job.query.QueryJob, bigquery.job.load.LoadJob],
     timeout: int = 1800,
-    retry_cadence: int = 10,
+    retry_cadence: float = 1,
 ):
     """
     Waits for bq_job to finish running, up to a maximum amount of time specified by the timeout parameter (defaulting to 30 minutes).
@@ -226,7 +226,7 @@ def block_until_done(
     # For test environments, retry more aggressively
     is_test = os.getenv("IS_TEST", default="False") == "True"
     if is_test:
-        retry_cadence = 1
+        retry_cadence = 0.1
 
     def _wait_until_done(job_id):
         if client.get_job(job_id).state in ["PENDING", "RUNNING"]:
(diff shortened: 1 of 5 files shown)


sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -222,6 +223,11 @@ def block_until_done(
         BigQueryJobCancelled exception to signify when that the job has been cancelled (i.e. from timeout or KeyboardInterrupt).
     """
 
+    # For test environments, retry more aggressively
```
woop, line 226: From a completely idealistic perspective, why couldn't we keep tests and the product in sync? What is the downside to just making all retries more aggressive, especially if we are just polling the status.

sdk/python/tests/integration/feature_repos/repo_configuration.py
```diff
@@ -162,6 +163,7 @@ def construct_test_environment(
     test_repo_config: IntegrationTestRepoConfig,
     test_suite_name: str = "integration_test",
 ) -> Environment:
+    os.environ["IS_TEST"] = "TRUE"
```
woop, line 166: I think it's a better practice to set global env vars at the CI level, like here https://github.com/feast-dev/feast/blob/master/.github/workflows/pr_integration_tests.yml and here https://github.com/feast-dev/feast/blob/master/.github/workflows/integration_tests.yml, just so that we can have them in one place.

adchia: Done. Also added it to the Makefile so local tests also include it

sdk/python/feast/infra/offline_stores/bigquery.py
adchia: Not much of one I guess? There's a cost of raise exceptions, but we're blocked anyways until the retry is done. We also don't really have other tasks going on.

I reduced the prod one to 1s + the test one to 0.1. In the test data, we know that our entity df is usually just 1 or 2 entities, whereas we expect much larger dataframes in prod use, so _seems_ like we should have something different.

achals (approved):
/lgtm

PR #1835, opened 2021-09-03
Enforce case-insensitively unique feature view names

codyjlin:

CockroachDB's table names and the materialization jobs using dataflow need feature view names to be lowercase. (CRDB is an online store we're trying to use.) We can transform the names downstream / on our end, but we would like the collision check to happen at `feast apply` time.

It's such a small change that no issue created.

```release-note
FeatureView names must now be *case-insensitively* unique
```

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -982,12 +982,13 @@ def _print_materialization_log(
 
 
 def _validate_feature_views(feature_views: List[FeatureView]):
-    """ Verify feature views have unique names"""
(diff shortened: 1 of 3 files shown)


achals (approved):
I think this is fine, but is going to be a breaking change in case anyway was relying on this behavior. 

Any tests to update for this?

codyjlin:
> Any tests to update for this?

@achals Ah, didn't realize there were tests for this validation check. Updating the failed tests and modifying one to test the case-insensitive case.

achals (approved):
/lgtm

Issue #1839, opened 2021-09-05
Array/list feature materialization in BQ crashes in type conversion


adchia:
Currently, if you try to use BQ and materialize a feature that is a list (of numbers, strings, etc), Feast will crash because in BQ, the value type of the feature is a dictionary, such as 

`{'list': [{'item': 3}, {'item': 3}]}`
In materialize, we convert the latest values retrieval job to a pyarrow table and then converts to ValueProtos to write. This calls 

`python_type_to_feast_value_type`, which doesn't support dict types.

PR #1840, opened 2021-09-05
Enable the types test to run on all compatible environments

adchia:
Enables the types test to test feature types across compatible offline / online stores (but e.g. not using list features on Redshift since Redshift doesn't support those). This also doesn't fully enable testing materializing list features for BQ because of https://github.com/feast-dev/feast/issues/1839.

This also fixes some missing teardown functionality, which caused tests to fail because the same "project" across different environments that used the same offline store caused issues (e.g. inability to create a new dataset)

This surfaced a bug where we can't materialize list features in BQ. There's a J

Fixes #

```release-note
NONE
```


achals (approved):
/lgtm

PR #1846, opened 2021-09-08
Quick fixes for type map conversion issues related to empty lists

karlhigley:
These are the quick fixes I made in order to get list features working, as described in [this comment](https://github.com/feast-dev/feast/issues/1640#issuecomment-878437327) on #1640. I'm submitting this PR in order to share the code more than to get it merged, since I know there's a larger effort to improve testing of the type map beyond these fixes.


adchia:
Sorry actually seems like it works fine as is. Just tested it with https://gist.github.com/adchia/b242be8d066e15a7f26852400b1d2c59 and saw that this fix indeed fixes the empty list test case when you add it!

achals (approved):
/lgtm

judahrand:
Already got a PR open for what I think you're trying to fix #1703

judahrand:
I'm going through the loop of adding tests for this in the PR which I originally opened #1703.

karlhigley:
@judahrand Yeah, sounds good to me. Once your PR covers this case, we can close this one.

PR #1847, opened 2021-09-08
Add schema parameter to RedshiftSource

### LINKED ISSUE #1767: Add schema parameter to RedshiftSource to distinguish between database schemas
No possible way to specify schema via `table` parameter.
When you add schema parameter to table name e.g `schema_sample.feature_table` it throws exception when applying with message:
```
    raise DataSourceNotFoundException(self.table)
feast.errors.DataSourceNotFoundException: Unable to find table at 'schema_sample.feature_table'. Please check that table exists.
```

Add optional `schema` parameter to uniquely identify which schema to use to reference source table.

`query` parameter works just fine

felixwang9817:

**What this PR does / why we need it**: The RedshiftSource implementation currently does not support a `schema` parameter. This PR adds support for a `schema` parameter.

Fixes #1767 

```release-note
schema is now an optional parameter in RedshiftSource
```

### DIFF
--- a/sdk/python/feast/infra/offline_stores/redshift_source.py
+++ b/sdk/python/feast/infra/offline_stores/redshift_source.py
@@ -13,11 +13,27 @@ class RedshiftSource(DataSource):
         self,
         event_timestamp_column: Optional[str] = "",
         table: Optional[str] = None,
+        schema: Optional[str] = None,
         created_timestamp_column: Optional[str] = "",
         field_mapping: Optional[Dict[str, str]] = None,
         date_partition_column: Optional[str] = "",
         query: Optional[str] = None,
     ):
+        """
+        Creates a RedshiftSource object.
+
+        Args:
+            event_timestamp_column (optional): Event timestamp column used for point in
+                time joins of feature values.
+            table (optional): Redshift table where the features are stored.
+            schema (optional): Redshift schema in which the table is located.
+            created_timestamp_column (optional): Timestamp column indicating when the
+                row was created, used for deduplicating rows.
+            field_mapping (optional): A dictionary mapping of column names in this data
+                source to column names in a feature table or view.
+            date_partition_column (optional): Timestamp column used for partitioning.
+            query (optional): The query to be executed to obtain the features.
+        """
         super().__init__(
             event_timestamp_column,
             created_timestamp_column,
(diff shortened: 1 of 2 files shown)


sdk/python/feast/infra/offline_stores/redshift_source.py
```diff
@@ -13,25 +13,56 @@ def __init__(
...
         super().__init__(
             event_timestamp_column,
             created_timestamp_column,
             field_mapping,
             date_partition_column,
         )
 
-        self._redshift_options = RedshiftOptions(table=table, query=query)
+        # The default Redshift schema is named "public".
+        _schema = "public" if table and not schema else schema
```
woop, line 45: Is it appropriate for us to set the default or defer to Amazon for the schema? I'm not super familiar with how AWS deals with the lack of a schema.

woop:
My only real question is how we know that we have solved the problem without changing any tests. Do we consider this functionality to be far enough off the happy path that changing our tests or updating our tests isn't worth it?

sdk/python/feast/infra/offline_stores/redshift_source.py
felixwang9817: Redshift comes with a default schema named `public`. By default, operations occur in the `public` schema; for example, if you reference a table without specifying a schema, Redshift will look in the `public` schema.

I think it's better to be explicit here. It will make it easier for users to understand (especially users who aren't familiar with Redshift), and easier to debug.

felixwang9817:
> My only real question is how we know that we have solved the problem without changing any tests. Do we consider this functionality to be far enough off the happy path that changing our tests or updating our tests isn't worth it?

Yeah that's a fair question. Since this PR has added `"public"` as the default schema, the existing tests are now querying Redshift with both a table and a schema, instead of just relying on Redshift's default. However, we are not testing the case where a user might specify a different schema; I consider this far enough off the happy path that we shouldn't modify our tests.

sdk/python/feast/infra/offline_stores/redshift_source.py
woop: If Redshift uses public by default, what difference does setting the schema explicitly on our side make?

woop (approved):

PR #1850, opened 2021-09-10
Add support for multiple entities in Redshift

### LINKED ISSUE #1816: Redshift does not support multiple entities
## Expected Behavior 

Feast supports multiple entities with Redshift sources

## Current Behavior
When a feature requires more than 1 entities, Feast will throw error upon retrieving feature. It throws the follow error:
```
ERROR: function concat(character varying, character varying, character varying) does not exist\n  Hint: No function matches the given name and argument types. You may need to add explicit type casts.
```
The reason is, redshift CONCAT only support 2 arguments. you might want consider the concat operator or nested concat

felixwang9817:
**What this PR does / why we need it**: The current Redshift historical retrieval query does not support using multiple entities. This PR fixes that.

Fixes #1816 

```release-note
Users can now use multiple entities for historical retrieval on Redshift
```

### DIFF
--- a/sdk/python/feast/infra/online_stores/helpers.py
+++ b/sdk/python/feast/infra/online_stores/helpers.py
@@ -37,8 +37,7 @@ def get_online_store_from_config(online_store_config: Any,) -> OnlineStore:
 
 
 def _redis_key(project: str, entity_key: EntityKeyProto) -> bytes:
-    key: List[bytes] = [serialize_entity_key(entity_key)]
-    key.append(project.encode("utf-8"))
+    key: List[bytes] = [serialize_entity_key(entity_key), project.encode("utf-8")]
     return b"".join(key)
 
 
--- a/sdk/python/feast/infra/online_stores/helpers.py
+++ b/sdk/python/feast/infra/online_stores/helpers.py
@@ -1,13 +1,12 @@
 import importlib
 import struct
-from typing import Any
+from typing import Any, List
 
 import mmh3
 
 from feast import errors
 from feast.infra.key_encoding_utils import serialize_entity_key
 from feast.infra.online_stores.online_store import OnlineStore
-from feast.protos.feast.storage.Redis_pb2 import RedisKeyV2 as RedisKeyProto
 from feast.protos.feast.types.EntityKey_pb2 import EntityKey as EntityKeyProto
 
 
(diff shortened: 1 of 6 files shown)


protos/feast/storage/Redis.proto
```diff
@@ -1,34 +0,0 @@
...
-import "feast/types/Field.proto";
-import "feast/types/Value.proto";
-
-package feast.storage;
-
-option java_outer_classname = "RedisProto";
-option java_package = "feast.proto.storage";
-option go_package = "github.com/feast-dev/feast/sdk/go/protos/feast/storage";
-
-message RedisKeyV2 {
```
achals, line 28: Nervous that this will break things in the java server?

sdk/python/tests/integration/feature_repos/universal/feature_views.py
```diff
@@ -117,3 +117,16 @@ def create_global_stats_feature_view(source, infer_features: bool = False):
...
+    order_feature_view = FeatureView(
+        name="order",
+        entities=["driver", "customer_id"],
+        features=None
+        if infer_features
+        else [Feature(name="order_is_success", dtype=ValueType.INT32)],
+        batch_source=source,
+        ttl=timedelta(days=2),
+    )
+    return order_feature_view
```
achals, line 132: nit `return FeatureView(...)`

sdk/python/tests/integration/offline_store/test_universal_historical_retrieval.py
```diff
@@ -37,14 +37,21 @@ def find_asof_record(
     ts_key: str,
     ts_start: datetime,
     ts_end: datetime,
-    filter_key: str = "",
-    filter_value: Any = None,
+    filter_keys: List[str] = [],
+    filter_values: List[Any] = [],
```
achals, line 41: Oh, Don't have mutable default arguments, this is a very subtle footgun which will almost certainly break:

https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

You probably want to do 
```
    filter_keys: Optional[List[str]] = None,
    filter_values: Optional[List[Any]] = None):
        filter_keys = filter_keys or []
        filter_values = filter_values or []
```

achals (changes requested):

protos/feast/storage/Redis.proto
achals: https://github.com/feast-dev/feast-java/blob/master/storage/connectors/redis/src/main/java/feast/storage/connectors/redis/common/RedisKeyGenerator.java

sdk/python/feast/infra/online_stores/helpers.py
```diff
@@ -37,13 +36,10 @@ def get_online_store_from_config(online_store_config: Any,) -> OnlineStore:
...
-def _redis_key(project: str, entity_key: EntityKeyProto):
-    redis_key = RedisKeyProto(
-        project=project,
-        entity_names=entity_key.join_keys,
-        entity_values=entity_key.entity_values,
-    )
-    return redis_key.SerializeToString()
+def _redis_key(project: str, entity_key: EntityKeyProto) -> bytes:
+    key: List[bytes] = [serialize_entity_key(entity_key)]
+    key.append(project.encode("utf-8"))
```
achals, line 41:
```
    key: List[bytes] = [serialize_entity_key(entity_key), project.encode("utf-8")]
    return b"".join(key) 
```

achals (approved):
/lgtm

PR #1851, opened 2021-09-10
Support adding request data in on demand transforms

adchia:
Supports adding request data as a dependency in on demand transforms. This validates when if the entity_df or entity_rows don't have the required request data features.

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/data_source.py
+++ b/sdk/python/feast/data_source.py
@@ -573,8 +573,8 @@ class RequestDataSource(DataSource):
 
     def to_proto(self) -> DataSourceProto:
         schema_pb = {}
-        for key in self._schema.keys():
-            schema_pb[key] = self._schema.get(key).value
+        for key, value in self._schema.items():
+            schema_pb[key] = value
         options = DataSourceProto.RequestDataOptions(name=self._name, schema=schema_pb)
         data_source_proto = DataSourceProto(
             type=DataSourceProto.REQUEST_SOURCE, request_data_options=options
--- a/sdk/python/feast/data_source.py
+++ b/sdk/python/feast/data_source.py
@@ -523,10 +523,7 @@ class RequestDataSource(DataSource):
 
     @staticmethod
     def source_datatype_to_feast_value_type() -> Callable[[str], ValueType]:
-        pass
-
-    def get_table_query_string(self) -> str:
-        pass
+        raise NotImplementedError
 
     _name: str
     _schema: Dict[str, ValueType]
--- a/sdk/python/feast/data_source.py
+++ b/sdk/python/feast/data_source.py
@@ -512,6 +512,77 @@ class KafkaSource(DataSource):
         return type_map.redshift_to_feast_value_type
 
 
+class RequestDataSource(DataSource):
+    """
+    RequestDataSource that can be used to provide input features for on demand transforms
+
+    Args:
+        name: Name of the request data source
+        schema: Schema mapping from the input feature name to a ValueType
+    """
+
+    @staticmethod
+    def source_datatype_to_feast_value_type() -> Callable[[str], ValueType]:
+        pass
+
+    def get_table_query_string(self) -> str:
+        pass
+
+    _name: str
+    _schema: Dict[str, ValueType]
+
+    def __init__(
+        self, name: str, schema: Dict[str, ValueType],
+    ):
+        """Creates a RequestDataSource object."""
+        super().__init__()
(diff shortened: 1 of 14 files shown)


sdk/python/feast/data_source.py
```diff
@@ -512,6 +512,77 @@ def source_datatype_to_feast_value_type() -> Callable[[str], ValueType]:
...
+        name: Name of the request data source
+        schema: Schema mapping from the input feature name to a ValueType
+    """
+
+    @staticmethod
+    def source_datatype_to_feast_value_type() -> Callable[[str], ValueType]:
+        pass
+
+    def get_table_query_string(self) -> str:
+        pass
```
achals, line 529: This shuld probably have a return type of `Optional[str]` for correctness.

sdk/python/feast/feature_store.py
```diff
@@ -541,6 +544,23 @@ def get_historical_features(
...
+                odfv_inputs = odfv.inputs.values()
+                for odfv_input in odfv_inputs:
+                    if type(odfv_input) == RequestDataSource:
+                        request_data_source = cast(RequestDataSource, odfv_input)
+                        for feature_name in request_data_source.schema.keys():
+                            if feature_name not in entity_pd_df.columns:
+                                raise RequestDataNotFoundInEntityDfException(
+                                    feature_name=feature_name,
+                                    feature_view_name=odfv.name,
+                                )
```
achals, line 563: Can we add tests to assert this failure?

sdk/python/feast/feature_store.py
```diff
@@ -805,28 +825,61 @@ def get_online_features(
...
                     join_key_row[DUMMY_ENTITY_ID] = DUMMY_ENTITY_VAL
-            join_key_rows.append(join_key_row)
+            if len(join_key_row) > 0:
+                # May be empty if this entity row was request data
+                join_key_rows.append(join_key_row)
+
+        if len(needed_request_data_features) != len(request_data_features.keys()):
+            raise RequestDataNotFoundInEntityRowsException(
+                feature_names=needed_request_data_features
+            )
```
achals, line 858: Can we add a test that asserts this exception is raised correctly?

sdk/python/feast/feature_store.py
```diff
@@ -805,28 +825,61 @@ def get_online_features(
...
+        for feature_name, feature_values in request_data_features.items():
+            for row_idx, feature_value in enumerate(feature_values):
+                result_row = result_rows[row_idx]
+                result_row.fields[feature_name].CopyFrom(
+                    python_value_to_proto_value(feature_value)
+                )
+                result_row.statuses[
+                    feature_name
+                ] = GetOnlineFeaturesResponse.FieldStatus.PRESENT
+
```
achals, line 881: we should also assert these status are set correctly in tests.

achals (commented):
Nice! I think we should add some tests to assert proper handling of the error cases, but looks good otherwise.

sdk/python/feast/data_source.py
adchia: Just removed this since it's optional. Made the other method raise an exception

sdk/python/feast/feature_store.py
adchia: Done

sdk/python/feast/feature_store.py
adchia: Done

sdk/python/feast/feature_store.py
adchia: We don't seem to assert them now and it also doesn't seem to be exposed publically in the API?

sdk/python/feast/data_source.py
```diff
@@ -512,6 +512,74 @@ def source_datatype_to_feast_value_type() -> Callable[[str], ValueType]:
...
+    RequestDataSource that can be used to provide input features for on demand transforms
+
+    Args:
+        name: Name of the request data source
+        schema: Schema mapping from the input feature name to a ValueType
+    """
+
+    @staticmethod
+    def source_datatype_to_feast_value_type() -> Callable[[str], ValueType]:
+        raise NotImplementedError
```
achals, line 526: nit: I wonder if this method makes sense for this type of data source. We should probably refactor the `DataSource` interface to make it simpler and introduce a BatchDataSource somewhere in the type hierarchy.

achals (approved):
/lgtm

PR #1856, opened 2021-09-11
Add MVP support for on demand transforms for AWS to_s3 and to_redshift

adchia:
Add MVP support for on demand transforms for AWS to_s3 and to_redshift

Tested with a local repo that uploaded to s3 + to_redshift and inspected results to see on demand transform properly applied into a new column

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/infra/offline_stores/redshift.py
+++ b/sdk/python/feast/infra/offline_stores/redshift.py
@@ -264,6 +264,11 @@ class RedshiftRetrievalJob(RetrievalJob):
 
     def to_s3(self) -> str:
         """ Export dataset to S3 in Parquet format and return path """
+        if self.on_demand_feature_views is not None:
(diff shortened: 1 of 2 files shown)


sdk/python/feast/infra/offline_stores/redshift.py
```diff
@@ -264,6 +264,11 @@ def to_arrow(self) -> pa.Table:
 
     def to_s3(self) -> str:
```
achals, line 265: What does the `        with self._query_generator() as query:` do? Should the upload be in that block, or is it fine outside?

adchia: Should be ok since this is called within the self.to_df() call.

achals (approved):
/lgtm

PR #1864, opened 2021-09-15
Add flags file to include experimental flags and test/usage flags

adchia:
Adds a flags module + puts python feature server & on demand transform behind that flag (throwing errors if not enabled). This uses feature_store.yaml as the way for users to control features, but provides cli methods to enable/disable features. Features are off by default (but on for tests)

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/cli.py
+++ b/sdk/python/feast/cli.py
@@ -22,11 +22,7 @@ import pkg_resources
 import yaml
 
 from feast import flags, utils
-from feast.errors import (
-    ExperimentalFeatureNotEnabled,
-    FeastObjectNotFoundException,
-    FeastProviderLoginError,
-)
+from feast.errors import FeastObjectNotFoundException, FeastProviderLoginError
 from feast.feature_store import FeatureStore
 from feast.repo_config import load_repo_config
 from feast.repo_operations import (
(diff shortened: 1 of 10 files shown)


sdk/python/feast/flags.py
```diff
@@ -0,0 +1,39 @@
+import os
+
+FLAG_ALPHA_FEATURES_NAME = "ENABLE_ALPHA_FEATURES"
+FLAG_ON_DEMAND_TRANSFORM_NAME = "ENABLE_ON_DEMAND_TRANSFORMS"
+FLAG_PYTHON_FEATURE_SERVER_NAME = "ENABLE_PYTHON_FEATURE_SERVER"
```
woop, line 5: Do we need to have a per-functionality flag? What about having just one global flag?

sdk/python/feast/errors.py
```diff
@@ -239,3 +241,10 @@ def __init__(self, feature_view_name: str):
         super().__init__(
             f"The feature view name: {feature_view_name} refers to both an on-demand feature view and a feature view"
         )
+
+
+class ExperimentalFeatureNotEnabled(Exception):
+    def __init__(self, feature_flag_name: str):
+        super().__init__(
+            f"Trying to use experimental feature. Please ensure to set environment variables {FLAG_ALPHA_FEATURES_NAME} and {feature_flag_name} to true"
```
woop, line 249: What about 
```suggestion
            f"You are attempting to use an experimental feature. Please set the environment variables {FLAG_ALPHA_FEATURES_NAME} and {feature_flag_name} to true"
```

sdk/python/feast/flags.py
adchia: I think it could be nice to at least have a kill switch so users could use some but not all if there's some bad bug with one. Could make them enabled by default though?

woop: I'd definitely have them disabled by default. I am fine with the current approach. We can also consider using the `feature_store.yaml` for experimental config instead of env vars. I don't feel too strongly, but I think dealing with a bunch of env vars can get tiring beyond some point.

adchia: Moved to feature_store.yaml + added CLI commands to modify

sdk/python/feast/repo_config.py
```diff
@@ -86,6 +87,9 @@ class RepoConfig(FeastBaseModel):
     offline_store: Any
     """ OfflineStoreConfig: Offline store configuration (optional depending on provider) """
 
+    flags: Any
+    """ Flags: feature flags for experimental features (optional depending on provider) """
```
woop, line 91:
```suggestion
    """ Flags: Feature flags for experimental features (optional depending on provider) """
```

adchia: Done!

sdk/python/feast/cli.py
```diff
@@ -423,5 +423,100 @@ def serve_command(ctx: click.Context, port: int):
...
+@cli.group(name="alpha")
+def alpha_cmd():
+    """
+    Access alpha features
+    """
+    pass
+
+
+@alpha_cmd.command("list")
+def list_alpha_features():
```
achals, line 435: Nit, would be a good idea to show the status (enabled/disabled) of these flags in this command.

sdk/python/feast/errors.py
```diff
@@ -258,3 +258,10 @@ def __init__(self, feature_view_name: str):
         super().__init__(
             f"The feature view name: {feature_view_name} refers to both an on-demand feature view and a feature view"
         )
+
+
+class ExperimentalFeatureNotEnabled(Exception):
+    def __init__(self, feature_flag_name: str):
+        super().__init__(
+            f"You are attempting to use an experimental feature. Please run `feast alpha enable {feature_flag_name}`"
```
achals, line 266:
```suggestion
            f"You are attempting to use an experimental feature that is not enabled. Please run `feast alpha enable {feature_flag_name}`"
```

sdk/python/feast/flags.py
```diff
@@ -0,0 +1,10 @@
+FLAG_ALPHA_FEATURES_NAME = "enable_alpha_features"
+FLAG_ON_DEMAND_TRANSFORM_NAME = "enable_on_demand_transforms"
+FLAG_PYTHON_FEATURE_SERVER_NAME = "enable_python_feature_server"
```
achals, line 3: Super nit: can we remove the `enable_` prefix? If i understand correctly the CLI usage is something like this:
```
feast alpha enable enable_on_demand_transforms
```

which seems verbose

sdk/python/feast/cli.py
adchia: Done

sdk/python/feast/errors.py
adchia: Done

sdk/python/feast/flags.py
adchia: Done

achals (approved):
/lgtm

achals (approved):
/lgtm

PR #1865, opened 2021-09-15
Add feature server configuration for AWS lambda

felixwang9817:

**What this PR does / why we need it**: This PR adds the option to add a feature server configuration to `feature_store.yaml`. Right now, only AWS lambda is supported.

Fixes #

```release-note
The user can now define a feature server in `feature_store.yaml`.
```

### DIFF
--- a/sdk/python/feast/infra/feature_servers/aws_lambda/app.py
+++ b/sdk/python/feast/infra/feature_servers/aws_lambda/app.py
@@ -1,0 +1,23 @@
+from pydantic import StrictBool, StrictStr
+from pydantic.typing import Literal
+
+from feast.repo_config import FeastConfigBaseModel
+
+
+class AwsLambdaFeatureServerConfig(FeastConfigBaseModel):
+    """Feature server config for AWS Lambda."""
+
+    type: Literal["aws_lambda"] = "aws_lambda"
+    """Feature server type selector."""
+
+    enabled: StrictBool
+    """Whether the feature server should be launched."""
+
+    public: StrictBool
+    """Whether the endpoint should be publicly accessible."""
+
+    auth: StrictStr
+    """Authentication method for the endpoint."""
+
+    execution_role_name: StrictStr
+    """The execution role for the AWS Lambda function."""
--- a/sdk/python/feast/repo_config.py
+++ b/sdk/python/feast/repo_config.py
@@ -32,6 +32,10 @@ OFFLINE_STORE_CLASS_FOR_TYPE = {
     "redshift": "feast.infra.offline_stores.redshift.RedshiftOfflineStore",
 }
 
+FEATURE_SERVER_CONFIG_CLASS_FOR_TYPE = {
+    "aws_lambda": "feast.infra.feature_servers.aws_lambda.app.AwsLambdaFeatureServerConfig",
+}
+
 
 class FeastBaseModel(BaseModel):
     """ Feast Pydantic Configuration Class """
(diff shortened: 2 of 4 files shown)


sdk/python/feast/repo_config.py
```diff
@@ -244,6 +279,15 @@ def get_offline_config_from_type(offline_store_type: str):
     return get_class_from_type(module_name, config_class_name, config_class_name)
 
 
+def get_feature_server_config_from_type(feature_server_type: str):
+    # We do not support custom feature servers right now.
+    assert feature_server_type in FEATURE_SERVER_CONFIG_CLASS_FOR_TYPE
```
woop, line 284: Ideally we'd avoid `assert` usage since asserts can be disabled by an optimizer.

sdk/python/feast/infra/feature_servers/aws_lambda/app.py
```diff
@@ -0,0 +1,23 @@
...
+from feast.repo_config import FeastConfigBaseModel
+
+
+class AwsLambdaFeatureServerConfig(FeastConfigBaseModel):
+    """Feature server config for AWS Lambda."""
+
+    type: Literal["aws_lambda"] = "aws_lambda"
+    """Feature server type selector."""
+
+    enabled: StrictBool
```
tsotnet, line 13: Can we have a default value of False here?
```suggestion
    enabled: StrictBool = False
```

sdk/python/feast/infra/feature_servers/aws_lambda/app.py
```diff
@@ -0,0 +1,23 @@
...
+    """Whether the feature server should be launched."""
+
+    public: StrictBool
```
tsotnet, line 16: Let's set the default to True
```suggestion
    public: StrictBool = True
```

sdk/python/feast/infra/feature_servers/aws_lambda/app.py
```diff
@@ -0,0 +1,23 @@
...
+    """Whether the endpoint should be publicly accessible."""
+
+    auth: StrictStr
```
tsotnet, line 19: Let's set the default to `none` and change the type to be one of `none` and `api-key`
```suggestion
    auth: Literal["none", "api-key"] = "none"
```

sdk/python/feast/repo_config.py
felixwang9817: Got it, should we just raise a custom error defined in `errors.py` instead? cc @woop

sdk/python/feast/repo_config.py
```diff
@@ -190,6 +194,37 @@ def _validate_offline_store_config(cls, values):
...
+    def _validate_feature_server_config(cls, values):
+        # Having no feature server is the default.
+        if "feature_server" not in values:
+            return values
+
+        # Make sure that the provider configuration is set. We need it to set the defaults
+        assert "provider" in values
+
+        # Make sure that the type is not set, since we will set it based on the provider.
+        assert "type" not in values
```
tsotnet, line 207: should we be checking type in values["feature_server"] instead?

tsotnet: Also can you replace asserts with if/raise checks? (similar to @woop 's comment)

tsotnet (changes requested):

sdk/python/feast/repo_config.py
tsotnet: In my opinion, there's no need to create custom errors for super simple stuff. I'd recommend throwing `ValueError` on validations. In this particular case you can simply remove this line since `FEATURE_SERVER_CONFIG_CLASS_FOR_TYPE[feature_server_type]` will fail anyways.

sdk/python/feast/repo_config.py
```diff
@@ -190,6 +199,39 @@ def _validate_offline_store_config(cls, values):
 
         return values
 
+    @root_validator(pre=True)
+    def _validate_feature_server_config(cls, values):
+        # Having no feature server is the default.
+        if "feature_server" not in values:
+            return values
```
tsotnet, line 206: Instead of returning values here, can you create an empty `values["feature_server"]` dict? (e.g. see _validate_online_store_config how they do this). This way, it's easier to read config in other places. For example, we'll know that feature_server always exists (and enabled will be False by default), even if the entire section is missing from the yaml file

tsotnet (approved):
/lgtm

Issue #1762, opened 2021-08-06
Shadow entities mapped properly to real entity with entity aliasing


mavysavydav:
**Is your feature request related to a problem? Please describe.**

There may be multiple different entity keys on the entity source data side that should map onto the same entity from the feature data table during the join. For instance, "spammer_id" and "reporter_id" may want the "years_on_platform" feature which lives in a table keyed by "user_id". It is inconvenient to combine the "spammer_ids" and "reporter_ids" into a new additional "user_ids" column and then clean it up after the join. In this case, we can call "spammer" and "reporter" the shadow entities of "user". 

In another example, both "rider_id" and "driver_id" may want the "age" feature which lives in a table keyed by "user_id".  In this case, we can call "rider" and "driver" the shadow entities of "user". 

**Describe the solution you'd like**

Via aliasing on the FeatureServices level, the shadow entities can be mapped to the correct FeatureView entity. 

```python
abuse_feature_service = FeatureService(
    name="my_abuse_model_v1",
    features=[
        spammer_features,
        reporter_features,
        user_features.alias({'spammer_id', 'reporter_id'})
    ],
)
```

woop:
@mavysavydav happy to stab out a quick proposal for you here, but I am a bit confused by what is contained within each of these feature views and their source tables.

What is the union of join keys in this feature service? Why would aliasing `spammer_id` on `user_features` as `reporter_id` actually solve the join? Wouldn't spammer_features have a `spammer_id`?

If you could show me basic examples of what these tables look like using https://ozh.github.io/ascii-tables/ then it would really help a lot!

mavysavydav:
So what I meant by `user_features.alias({'spammer_id', 'reporter_id'})` is that user_features is the FV and it already has an entity key defined in the feature view definition (e.g user_id). And "spammer_id" and "reporter_id" would be aliased to whatever the entity key of the feature view is. In this case, since we're doing user_features.alias(..), spammer_id and reporter_id are both mapped to user_id. Let me know if ascii tables would still help or if these are clarified.

adchia:
the exactly API we probably need to rev on, but having this kind of functionality makes sense to me. 

Couple thoughts (think this is def worth writing an RFC for though):
1) Logically, this breaks down really to me as wanting the final feature vector as being:

entity_df + (spammer_features+spammer_user_features) + (reporter_features + reporter_user_features).

[...]

PR #1868, opened 2021-09-15
Implement shadow entities mapping functionality

codyjlin:

This PR implements the [Shadow Entities Mapping RFC](https://docs.google.com/document/d/1TsCwKf3nVXTAfL0f8i26jnCgHA3bRd4dKQ8QdM87vIA/edit#) - currently assuming we go with **Option 8b** (and the implementation idea at the bottom of the RFC). It will allow a user to override a FeatureView's join_key (like `"user_id"`) to be able to join with an `entity_df` source that might have other column names that are spelled differently or might be "subclasses" of the entity (like `"spammer_id"` and `"reporter_id"`). This functionality is to be made available on the offline side via `get_historical_features` and the online side via `get_online_features`.

**TODO**:
- testing to come
- will add changes to redshift.py once changes to bigquery.py are reasonable/approved, and/or testing is ready

Part of #1762
(Might want to add documentation after change is landed)

```release-note
User can now override the join keys of a FeatureView if the entity column name(s) are misspelled or
don't match the entity column name of the entity_df.
This can be done:
- by setting the FeatureView.join_key_map attribute
- at use-time FV in the FeatureService.features attribute via FeatureView.with_join_key_map(join_key_map)
```


sdk/python/feast/feature_store.py
```diff
@@ -1013,6 +1020,23 @@ def _entity_row_to_field_values(
     return result
 
 
+def _use_time_feature_views(all_feature_views, features):
```
woop, line 1023: Is `use_time` a term people use? It wasn't immediately clear to me what this meant. I was also imagining that the logic for manipulating the feature view would be on the feature view itself, not a separate method. 

Finally, you should add type annotations please.

woop:
Please remember to add documentation to this PR as well.

sdk/python/feast/feature_store.py
codyjlin: 1. Definitely open to alternatives of `use_time` - I agree that it's not clear, but I'm not sure how else to describe it or what to rename it to. Basically instead of grabbing all of the registered `FeatureView`s, we need to grab any changes that the user wants to specify at the time of using the FeatureView (like the `join_key_map` or a new name).

2. By "the logic for manipulating the feature view would be on the feature view itself", do you mean `.override_join_keys` or `.with_join_key_map` should manipulate the original object? I feel like we wouldn't want that behavior with this design given one might use it multiple times in different contexts.

3. Will do in next commit 👌

woop: > Definitely open to alternatives of use_time - I agree that it's not clear, but I'm not sure how else to describe it or what to rename it to. Basically instead of grabbing all of the registered FeatureViews, we need to grab any changes that the user wants to specify at the time of using the FeatureView (like the join_key_map or a new name).

I'd first want to see if we can avoid naming it at all. My instinct would have been to use a class like [FeatureViewProjection](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/feature_view_projection.py) which already supports filtering columns, and adding more options like renaming the feature view or overriding the join key. We may change it over time to be `FeatureViewQuery` or something. That would also prevent somebody from doing `fv.with_name("blah")` and then accidentally registering that FV, since we only anticipate the use of renaming to be for queries.

> By "the logic for manipulating the feature view would be on the feature view itself", do you mean .override_join_keys or .with_join_key_map should manipulate the original object? I feel like we wouldn't want that behavior with this design given one might use it multiple times in different contexts.

Not override the original object, but instead produce a new object. It's just that the logic itself seems like it is better to have on the feature view class itself, basically what @mavysavydav did with `with_name()`.

sdk/python/feast/feature_store.py
```diff
@@ -1013,6 +1020,23 @@ def _entity_row_to_field_values(
...
+    """Look for use-time FeatureView.join_key_maps and update all_feature_views"""
+    if isinstance(features, FeatureService):
+        feature_view_replacements = {
+            feature.name: feature
+            for feature in features.features
+            if isinstance(feature, FeatureView) and feature.join_key_map
```
codyjlin, line 1029: shoot, just realized these are all `FeatureViewProjection` objects cause it converts the features after initializing a FeatureService...

sdk/python/tests/integration/offline_store/test_universal_historical_retrieval.py
```diff
@@ -216,18 +251,30 @@ def test_historical_features(environment, universal_data_sources, full_feature_n
...
 
     feature_service = FeatureService(
-        "convrate_plus100",
+        name="convrate_plus100",
         features=[feature_views["driver"][["conv_rate"]], feature_views["driver_odfv"]],
     )
+    feature_service_shadow_entities = FeatureService(
+        name="shadow_entities",
+        features=[
+            location_fv.with_name("origin").with_join_key_map(
```
codyjlin, line 270: This (and every other usage of `FV.with_name()` is dependent on https://github.com/feast-dev/feast/pull/1872.

sdk/python/feast/feature_store.py
codyjlin: This is dependent on a PR by @mavysavydav to have FeatureService.features allow FeatureView as one of the types.

PR #1870, opened 2021-09-16
Read registry & config from env variables in AWS Lambda feature server

tsotnet:

**What this PR does / why we need it**: We want to read registry (binary) and config (feature_store.yaml) via base64-encoded environment variables. They'll be set during `feast apply` when deploying AWS Lambda resource.

Fixes #

```release-note
NONE
```


felixwang9817 (approved):
/lgtm

PR #1872, opened 2021-09-16
Added .with_name method in FeatureView class

### LINKED ISSUE #1867: Add ".with_name" method to FeatureView (Part of shadow entities mapping changes)
As part of the shadow entities mapping agreed design, we'll need this method. This method will prevent name collisions when the same FeatureView is joined more than once onto to the entity data. Here's the RFC - https://docs.google.com/document/d/1TsCwKf3nVXTAfL0f8i26jnCgHA3bRd4dKQ8QdM87vIA/edit#

Add a ".with_name" method to FeatureView class that'll return a copy of the FeatureView with the passed name instead.

We considered having a name be automatically produced in the shadow entities that has the shadow entity join_keys included in the name. The issue with this is that the name may become too long in the case of composite entities. We also considered just having a numerical counter in the name like fv_name_0_feature_name, fv_name_1_feature_name. 

**Additional context**

mavysavydav:
Signed-off-by: David Y Liu <davidyliuliu@gmail.com>

https://github.com/feast-dev/feast/issues/1867

Fixes #1867

```release-note
Added ".with_name" public method to FeatureView class as part of the shadow entities mapping implementation which will enable users to join a FeatureView to entity data more than once and therefore there may be feature name conflicts. For more context, see the shadow entities mapping RFC here - https://docs.google.com/document/d/1TsCwKf3nVXTAfL0f8i26jnCgHA3bRd4dKQ8QdM87vIA/edit#. 
```


sdk/python/feast/feature_view.py
```diff
@@ -199,6 +199,25 @@ def is_valid(self):
         if not self.entities:
             raise ValueError("Feature view has no entities.")
 
+    def with_name(self, name: str):
+        """
+        Produces a copy of this FeatureView with the passed name.
```
woop, line 204: Can we add `name` as an argument in the docstring?

woop:
Btw @mavysavydav your thoughts on https://github.com/feast-dev/feast/pull/1868#discussion_r710403410 would be appreciated since it may slightly change this PR.

codyjlin:
We may want to add another case to the `FeatureNameCollisionError` error message to reflect that they might also need to use `.with_name` if they're trying to use join on multiple featureviews with the same name (i.e. in the shadow entities case) since `full_feature_names=True` won't be enough.

https://github.com/feast-dev/feast/blob/e98d65a371a702ed19f7298dc3d26a5efd39dd20/sdk/python/feast/errors.py#L144-L157

sdk/python/feast/feature_view.py
mavysavydav: yep, will add documentation for the argument. We'll add an example in this docstring in the shadow entities mapping change cc: @codyjlin .

mavysavydav:
> We may want to add another case to the `FeatureNameCollisionError` error message to reflect that they might also need to use `.with_name` if they're trying to use join on multiple featureviews with the same name (i.e. in the shadow entities case) since `full_feature_names=True` won't be enough.
> 
> https://github.com/feast-dev/feast/blob/e98d65a371a702ed19f7298dc3d26a5efd39dd20/sdk/python/feast/errors.py#L144-L157

Cool, done

sdk/python/tests/integration/offline_store/test_historical_retrieval.py
```diff
@@ -399,13 +399,13 @@ def test_feature_name_collision_on_historical_retrieval():
             full_feature_names=False,
         )
 
-    expected_error_message = (
-        "Duplicate features named avg_daily_trips found.\n"
-        "To resolve this collision, either use the full feature name by setting "
-        "'full_feature_names=True', or ensure that the features in question have different names."
-    )
+        expected_error_message = (
```
mavysavydav, line 402: indentation was wrong here

sdk/python/tests/unit/test_feature_validation.py
```diff
@@ -20,13 +20,13 @@ def test_feature_name_collision_on_historical_retrieval():
             full_feature_names=False,
         )
 
-    expected_error_message = (
-        "Duplicate features named avg_daily_trips found.\n"
-        "To resolve this collision, either use the full feature name by setting "
-        "'full_feature_names=True', or ensure that the features in question have different names."
-    )
+        expected_error_message = (
```
mavysavydav, line 23: indentation was wrong here

PR #1873, opened 2021-09-16
Adding telemetry for on demand feature views and making existing usage calls async

adchia:
Adding telemetry for on demand feature views and making existing usage calls async

Tested this locally to ensure the events are logged correctly and that the get_online_features logs only trigger once every 10k requests.

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -56,11 +56,12 @@ from feast.protos.feast.types.EntityKey_pb2 import EntityKey as EntityKeyProto
 from feast.registry import Registry
 from feast.repo_config import RepoConfig, load_repo_config
 from feast.type_map import python_value_to_proto_value
-from feast.usage import log_exceptions, log_exceptions_and_usage
+from feast.usage import log_event, log_exceptions, log_exceptions_and_usage
 from feast.value_type import ValueType
 from feast.version import get_version
 
 warnings.simplefilter("once", DeprecationWarning)
+EVENT_APPLY_WITH_ODFV = "apply_with_on_demand_feature_views"
 
 
 class FeatureStore:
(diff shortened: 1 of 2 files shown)


sdk/python/feast/usage.py
```diff
@@ -89,14 +120,33 @@ def log(self, function_name: str):
...
+        self._usage_counter[event_name] = 2  # avoid overflow
+        return True
+
+    def log_event(self, event: UsageEvent):
+        self.check_env_and_configure()
+        if self._usage_enabled and self.usage_id:
+            event_name = str(event)
+            if event == UsageEvent.GET_ONLINE_FEATURES_WITH_ODFV:
+                if not self.should_log_for_get_online_features_event(event_name):
+                    return
```
achals, line 140: it collapse into `if event == ... and not self.should_log_for_get_online_features_event: return`

sdk/python/feast/feature_store.py
```diff
@@ -870,46 +878,16 @@ def get_online_features(
...
-                requested_features=requested_features,
+            self._populate_result_rows_from_feature_view(
+                entity_name_to_join_key_map,
+                full_feature_names,
+                provider,
+                requested_features,
+                result_rows,
+                table,
+                union_of_entity_keys,
             )
```
achals, line 890: Was this just a refactoring?

sdk/python/feast/usage.py
```diff
@@ -73,14 +90,28 @@ def usage_id(self) -> Optional[str]:
...
+        requests.post(USAGE_ENDPOINT, json=json)
+
+    async def _send_usage_request(self, json):
+        try:
+            task = asyncio.ensure_future(self._send_request(json))
+            if self._is_test:
+                await task
+        except Exception as e:
+            if self._is_test:
+                raise e
```
achals, line 103: Why are we doing this async?

sdk/python/feast/feature_store.py
adchia: Had to because our linter complains about complex methods

sdk/python/feast/usage.py
adchia: Done

sdk/python/feast/usage.py
adchia: Seemed like this was adding significant latency for the requests when they fire.

Verified that this was saving ~100-200ms for when the events are logged.

achals (approved):
/lgtm

PR #1874, opened 2021-09-16
Fix pytest_collection_modifyitems to select benchmark tests only

achals:

Fixes a bug. Additionally, remove test items instead of skipping them. This makes `--collect-only` more useful, since that flag doesn't filter out skipped flags. 

Fixes #

```release-note
none
```

### DIFF
--- a/sdk/python/tests/conftest.py
+++ b/sdk/python/tests/conftest.py
@@ -14,9 +14,11 @@
 import multiprocessing
 from datetime import datetime, timedelta
 from sys import platform
+from typing import List
 
 import pandas as pd
 import pytest
+from _pytest.nodes import Item
 
 from tests.data.data_creator import create_dataset
 from tests.integration.feature_repos.repo_configuration import (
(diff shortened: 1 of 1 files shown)


sdk/python/tests/conftest.py
```diff
@@ -52,18 +54,27 @@ def pytest_addoption(parser):
...
-        if "integration" in item.keywords and not should_run_integration:
-            item.add_marker(skip_integration)
-        if "benchmark" in item.keywords and not should_run_benchmark:
-            item.add_marker(skip_benchmark)
+
+    integration_tests = [t for t in items if "integration" in t.keywords]
+    if not should_run_integration:
+        for t in integration_tests:
+            items.remove(t)
+    else:
```
adchia, line 65: Seems like these are exclusive? 

If the user does should_run_integration and should_run_benchmark, it ends up only running benchmark tests.

I'd expect this more to be

if not should_run_integration, remove integration tests
if not should_run_benchmarks, remove benchmarks
done

achals: I changed the current behaviour to be exclusive.

if user specifies `--benchmark` and `--integration`, it runs tests marked with both. If the user specified only one of them, it runs tests with only those marks.

i prefer this behaviour since otherwise we run unit tests during all our integration tests - which seems like a duplication and a waste. It also is more straightforward to reason about, IMO.

achals: The reason why `integration` and `benchmark` tests are orthogonal is because it's possible for an integration test to need access to resources on aws/gcp, which requires the `integration` mark so that it can be set up correctly.

felixwang9817 (approved):
/lgtm

PR #1876, opened 2021-09-16
Refactor providers to remove duplicate implementations

achals:
Signed-off-by: Achal Shah <achals@gmail.com>

**What this PR does / why we need it**:

Currently, the AWS, GCP and Local providers have the exact same implementation - and the implementation only defers actual business logic of setting up infrastructure and materializing data to the offline/online store interfaces.

This is confusing for users, and is unnecessary code to maintain. Instead, this PR refactors the implementations into a single implementation, with a dummy subclass for the 3 existing providers for backwards compatibility.

**Which issue(s) this PR fixes**:

Fixes #

**Does this PR introduce a user-facing change?**:

```release-note
none
```

PR #1877, opened 2021-09-17
Upload docker image to ECR during feast apply

felixwang9817:

**What this PR does / why we need it**: This PR pulls the AWS Lambda feature server docker image and uploads it to AWS ECR during feast apply.

Fixes #

```release-note
During feast apply, a feature server docker image may be uploaded to AWS ECR
```

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -45,7 +45,6 @@ from feast.inference import (
     update_data_sources_with_inferred_event_timestamp_col,
     update_entities_with_inferred_types_from_feature_views,
 )
-from feast.infra.feature_servers.feature_server import FEATURE_SERVER_IMAGE_FOR_TYPE
 from feast.infra.provider import Provider, RetrievalJob, get_provider
 from feast.on_demand_feature_view import OnDemandFeatureView
 from feast.online_response import OnlineResponse, _infer_online_entity_rows
@@ -1025,59 +1024,6 @@ class FeatureStore:
             raise ExperimentalFeatureNotEnabled(flags.FLAG_PYTHON_FEATURE_SERVER_NAME)
 
         feature_server.start_server(self, port)
-
-    @log_exceptions_and_usage
-    def upload_docker_image(self) -> None:
-        """Upload the docker image for the feature server to the cloud."""
-
-        # TODO(felixwang9817): add error checking, logging, and avoid hardcoding the region
-        repository_name = "feast-python-server-test"
-        feature_server_type = (
-            self.config.feature_server.type if self.config.feature_server else None
-        )
-        if feature_server_type == "aws_lambda":
-            import base64
-
-            try:
-                import boto3
-                from botocore.exceptions import ClientError
-            except ImportError as e:
-                from feast.errors import FeastExtrasDependencyImportError
-
-                raise FeastExtrasDependencyImportError("aws", str(e))
-
-            try:
-                import docker
-            except ImportError as e:
-                from feast.errors import FeastExtrasDependencyImportError
-
-                raise FeastExtrasDependencyImportError("docker", str(e))
-
-            docker_client = docker.from_env()
-            image_name = FEATURE_SERVER_IMAGE_FOR_TYPE[feature_server_type]
-            docker_client.images.pull(image_name)
-
-            ecr_client = boto3.client("ecr", region_name="us-west-2")
-            try:
-                ecr_client.create_repository(repositoryName=repository_name)
-            except ClientError:
-                pass
-            auth_token = ecr_client.get_authorization_token()["authorizationData"][0][
-                "authorizationToken"
-            ]
-            username, password = base64.b64decode(auth_token).decode("utf-8").split(":")
-
-            sts_client = boto3.client("sts")
-            aws_account = sts_client.get_caller_identity()["Account"]
-            ecr_address = f"{aws_account}.dkr.ecr.us-west-2.amazonaws.com"
-            docker_client.login(
-                username=username, password=password, registry=ecr_address
-            )
-
-            # Pushing will likely take several minutes.
-            image = docker_client.images.get(image_name)
-            image.tag(f"{ecr_address}/{repository_name}:latest")
-            docker_client.api.push(f"{ecr_address}/{repository_name}:latest")
 
 
 def _entity_row_to_key(row: GetOnlineFeaturesRequestV2.EntityRow) -> EntityKeyProto:
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -1028,9 +1028,9 @@ class FeatureStore:
 
     @log_exceptions_and_usage
     def upload_docker_image(self) -> None:
-        """Upload the docker image for the feature consumption server to the cloud."""
+        """Upload the docker image for the feature server to the cloud."""
 
-        # TODO: add error checking and avoid hardcoding the region
+        # TODO(felixwang9817): add error checking, logging, and avoid hardcoding the region
         repository_name = "feast-python-server-test"
         feature_server_type = (
             self.config.feature_server.type if self.config.feature_server else None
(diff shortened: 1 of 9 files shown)


sdk/python/feast/feature_store.py
```diff
@@ -998,6 +999,48 @@ def serve(self, port: int) -> None:
 
         feature_server.start_server(self, port)
 
+    @log_exceptions_and_usage
+    def upload_docker_image(self) -> None:
+        """Upload the docker image for the feature consumption server to the cloud."""
```
woop, line 1004: I dont think `consumption server` is an official term.

sdk/python/feast/feature_store.py
```diff
@@ -998,6 +999,48 @@ def serve(self, port: int) -> None:
...
+
+        # TODO: add error checking and avoid hardcoding the region
+        repository_name = "feast-python-server-test"
+        feature_server_type = (
+            self.config.feature_server.type if self.config.feature_server else None
+        )
+        if feature_server_type == "aws_lambda":
+            import base64
+
+            import boto3
```
woop, line 1014: should we wrap this import in a try catch with an FeastImport exception?

sdk/python/feast/feature_store.py
```diff
@@ -998,6 +999,48 @@ def serve(self, port: int) -> None:
...
+            import docker
+            from botocore.exceptions import ClientError
+
+            docker_client = docker.from_env()
+            image_name = FEATURE_SERVER_IMAGE_FOR_TYPE[feature_server_type]
+            docker_client.images.pull(image_name)
```
woop, line 1020: I really think we should start to introduce a logging so that users can see what we are doing here.

sdk/python/feast/feature_store.py
```diff
@@ -998,6 +999,48 @@ def serve(self, port: int) -> None:
...
+            sts_client = boto3.client("sts")
+            aws_account = sts_client.get_caller_identity()["Account"]
+            ecr_address = f"{aws_account}.dkr.ecr.us-west-2.amazonaws.com"
+            docker_client.login(
+                username=username, password=password, registry=ecr_address
+            )
+
+            # Pushing will likely take several minutes.
+            image = docker_client.images.get(image_name)
+            image.tag(f"{ecr_address}/{repository_name}:latest")
```
woop, line 1041: `:latest` could cause issues. Any system that has already pulled the `latest` image may skip pulling an image again. Can we implement some kind of versioning?

sdk/python/feast/infra/feature_servers/feature_server.py
```diff
@@ -0,0 +1 @@
+FEATURE_SERVER_IMAGE_FOR_TYPE = {"aws_lambda": "feastdockerbot/feast-python-server"}
```
woop, line 1: Why is `feastdockerbot` used here instead of `feastdev`?

sdk/python/feast/feature_store.py
```diff
@@ -998,6 +999,48 @@ def serve(self, port: int) -> None:
 
         feature_server.start_server(self, port)
 
+    @log_exceptions_and_usage
+    def upload_docker_image(self) -> None:
```
tsotnet, line 1003: Why is this a top-level public method? I think this logic should live as part of update_infra in AwsProvider.

sdk/python/feast/feature_store.py
```diff
@@ -998,6 +999,48 @@ def serve(self, port: int) -> None:
...
+        """Upload the docker image for the feature consumption server to the cloud."""
+
+        # TODO: add error checking and avoid hardcoding the region
+        repository_name = "feast-python-server-test"
+        feature_server_type = (
+            self.config.feature_server.type if self.config.feature_server else None
+        )
```
tsotnet, line 1010: You won't need to determine `feature_server_type` once the logic is part of AwsProvider. We'll just need to check whether feature server is enabled or not.

sdk/python/feast/feature_store.py
```diff
@@ -998,6 +999,48 @@ def serve(self, port: int) -> None:
...
+        if feature_server_type == "aws_lambda":
+            import base64
+
+            import boto3
+            import docker
```
tsotnet, line 1015: You need to add this to setup.py. Also need to handle an error where docker binary isn't installed (I assume pip installing docker doesn't actually install the docker daemon)

sdk/python/feast/feature_store.py
```diff
@@ -998,6 +999,48 @@ def serve(self, port: int) -> None:
...
+            from botocore.exceptions import ClientError
+
+            docker_client = docker.from_env()
+            image_name = FEATURE_SERVER_IMAGE_FOR_TYPE[feature_server_type]
```
tsotnet, line 1019: Let's just define a constant for AWS (since the logic is being moved to AwsProvider). No need to create a new file, you can use `sdk/python/feast/constants.py`

sdk/python/feast/feature_store.py
```diff
@@ -998,6 +999,48 @@ def serve(self, port: int) -> None:
...
+            docker_client.images.pull(image_name)
+
+            ecr_client = boto3.client("ecr", region_name="us-west-2")
```
tsotnet, line 1022: Remove the region_name, it'll automatically be grabbed from the profile.

sdk/python/feast/feature_store.py
```diff
@@ -998,6 +999,48 @@ def serve(self, port: int) -> None:
...
+            try:
+                ecr_client.create_repository(repositoryName=repository_name)
+            except ClientError:
+                pass
```
tsotnet, line 1026: Why pass? We need to handle the error. You can check if the error is because it already exists, in which case you can pass, but if it's some other error make sure to raise the error.

sdk/python/feast/feature_store.py
```diff
@@ -998,6 +999,48 @@ def serve(self, port: int) -> None:
...
+            auth_token = ecr_client.get_authorization_token()["authorizationData"][0][
+                "authorizationToken"
+            ]
+            username, password = base64.b64decode(auth_token).decode("utf-8").split(":")
+
+            sts_client = boto3.client("sts")
+            aws_account = sts_client.get_caller_identity()["Account"]
+            ecr_address = f"{aws_account}.dkr.ecr.us-west-2.amazonaws.com"
```
tsotnet, line 1034: Doesn't ecr_client.create_repository return this information? If not, at least let's change us-west-2 to the actual region.

sdk/python/feast/feature_store.py
```diff
@@ -998,6 +999,48 @@ def serve(self, port: int) -> None:
...
+            docker_client.login(
+                username=username, password=password, registry=ecr_address
+            )
+
+            # Pushing will likely take several minutes.
+            image = docker_client.images.get(image_name)
```
tsotnet, line 1040: Let's not forget to add logging to these steps.

sdk/python/feast/feature_store.py
tsotnet: How about we use `feast.__version__` as a tag?

sdk/python/feast/repo_operations.py
```diff
@@ -257,6 +257,10 @@ def apply_total(repo_config: RepoConfig, repo_path: Path, skip_source_validation
     # Commit the update to the registry only after successful infra update
     registry.commit()
 
+    repo_config = store.config
+    if repo_config.feature_server and repo_config.feature_server.enabled:
+        store.upload_docker_image()
+
```
tsotnet, line 263: Let's move this to AwsProvider's update_infra.

tsotnet (changes requested):

sdk/python/feast/feature_store.py
felixwang9817: Since we unified all provider logic in the PassthroughProvider in #1876, I think I'll have to continue checking for `feature_server_type`.

sdk/python/feast/infra/aws.py
```diff
@@ -13,11 +17,66 @@
...
+            from feast.errors import DockerDaemonNotRunning
+
+            raise DockerDaemonNotRunning()
+
+        print(
+            f"Pulling remote image {Style.BRIGHT + Fore.GREEN}{AWS_LAMBDA_FEATURE_SERVER_IMAGE}{Style.RESET_ALL}:"
+        )
+        docker_client.images.pull(AWS_LAMBDA_FEATURE_SERVER_IMAGE)
+
+        repository_name = "feast-python-server-test"
```
tsotnet, line 50: remove test suffix

tsotnet (approved):
/lgtm

PR #1882, opened 2021-09-19
Use contextvars to maintain a call stack during the usage calls

achals:

Our usage handling doesn't handle calls to `usage` annotated events *within* the implementation of `get_online_features`. So, e.g., calling `get_feature_views` within `get_online_features` would always get called, which would result in bad performance. 

This fixes that by maintaining the call stack and making sure any call path that includes `get_online_features` is only sampled.

Fixes #

```release-note
none
```


adchia (approved):
/lgtm

PR #1886, opened 2021-09-20
Add `dill` to main dependencies

judahrand:
`dill` is imported by `sdk/python/feast/on_demand_feature_view.py` and
so is needed as a main dependency.

Fixes #

```release-note
Fix import error caused by missing dependency.
```


achals (approved):
/lgtm

PR #1889, opened 2021-09-20
Fix list feature format for BigQuery offline datasources. 

### LINKED ISSUE #1839: Array/list feature materialization in BQ crashes in type conversion

judahrand:
Should have been using this option from the start but the docs on it are very lacking. Took me an hour or two to find it. This should make BQ behave like the other data store with `pyarrow` lists instead of the weird `{list: {item: value}}`.

Some concerns about this being incompatible with already existing tables? Though I think really this is important as the transformation after the fact is slow and this it the 'right' fix.

Fixes #1839

```release-note
Fix list feature format for BigQuery offline datasources. 
```

### DIFF
--- a/sdk/python/tests/integration/registration/test_universal_types.py
+++ b/sdk/python/tests/integration/registration/test_universal_types.py
@@ -248,16 +248,10 @@ def assert_feature_list_types(
         "bool": "bool",
     }
     assert str(historical_features_df.dtypes["value"]) == "object"
-    if provider == "gcp":
-        assert (
-            feature_list_dtype_to_expected_historical_feature_list_dtype[feature_dtype]
-            in type(historical_features_df.value[0]["list"][0]["item"]).__name__
-        )
-    else:
-        assert (
-            feature_list_dtype_to_expected_historical_feature_list_dtype[feature_dtype]
-            in type(historical_features_df.value[0][0]).__name__
-        )
+    assert (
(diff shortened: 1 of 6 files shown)


judahrand:
Not quite there 😢

sdk/python/tests/integration/registration/test_universal_types.py
```diff
@@ -255,16 +248,10 @@ def assert_feature_list_types(
...
-        assert (
-            feature_list_dtype_to_expected_historical_feature_list_dtype[feature_dtype]
-            in type(historical_features_df.value[0]["list"][0]["item"]).__name__
-        )
-    else:
-        assert (
-            feature_list_dtype_to_expected_historical_feature_list_dtype[feature_dtype]
-            in type(historical_features_df.value[0][0]).__name__
-        )
+    assert (
```
adchia, line 251: think you'll need to revert this change that you did on the left

judahrand:
All tests passing.

adchia (approved):
/lgtm

PR #1896, opened 2021-09-22
Fix feast apply with imported modules

MattDelac:

Make `feast apply` works when we import object from other files.

This is useful when we want to create a FeatureService composed of FeatureViews that have been created in other files.

_Note: FeatureServices are the only objects that need to import other feast objects like FeatureViews._

Fixes https://github.com/feast-dev/feast/issues/1895

```release-note
[Bug] `feast apply` does not raise a duplicate error when Feast objects are imported in other files
```


MattDelac:
Sounds like it's impossible to really know if a module has been created within a specific file or if it has been imported.

Therefore I suggest that we only print a warning if some "Objects are duplicated". Our users can always add unit tests in their own codebase to enforce rules.
Thoughts ?

MattDelac:
you can look at my 2 commits. The first one pass in the tests but will actually bring a silent bug that is ignoring certain feature views (tested on my use case)

MattDelac:
Also everything I tried using the [`inspect` package](https://docs.python.org/3/library/inspect.html#retrieving-source-code) did not work

achals:
Rather than disabling this feature, #1905 updates the logic to handle imported objects.

PR #1899, opened 2021-09-22
Hide FeatureViewProjections from user interface & have FeatureViews carry FVProjections that carries the modified info of the FeatureView

### LINKED ISSUE #1907: Remove FeatureViewProjections & handle its responsibility with FeatureView and FeatureService
We'd like to enhance FeatureView with methods like ".with_name" (https://github.com/feast-dev/feast/issues/1867) and potentially ".with_entities" as part as this shadow entities mapping RFC -- https://docs.google.com/document/d/1TsCwKf3nVXTAfL0f8i26jnCgHA3bRd4dKQ8QdM87vIA/edit#heading=h.o7edalv7g2mz. 

The trouble right now is that at query time, even though the FV is one concept in the users mind when use get_historical_features or get_online_features, but it actually splits into a different class when the __getitem__ magic method is triggered by my_fv[[<my-feature-selections]] and it becomes a FVProjection. This means if users wanted to transform their FV, we'd have to support these methods on both the FV class and FVProjection class. 

FVProjection is only used for 1 second when FV[[..]] is passed in FeatureService. After that, everything goes back to feature_ref strings and FeatureViews. In a world in which we make FVProjection carry more data/logic to support the needed upgrades for shadow entities mapping, it would require a big refactor if we wanted to carry the needed mapping info in FVProjection and have all the downstream code accept FVProjection. Now I think even tho FVProjection may have value in the future depending on what additional feast product features we add in the future, rn the state of Feast is a lot better suited for having FV be the sole grouping object (rather than having FVProjection as well). I think once having FVProjection proves its value, we can make the changes to make the switch.

In terms of risk management, the approach of eliminating FVProjection wouldn’t require much code changes and if we were to change our mind later on, then we can do the big refactoring. This is as opposed to doing the big refactoring first to string FVProjections through everything and then if we change our minds, we have to do another big refactoring to get back to where we are now. So worse case scenarios are 1 big refactoring for the former whereas 2 big refactorings for latter.

We must take care to be very intentional about what new functionality is added to FV and FeatureService and the bounds of these additional responsibilities must be clear.

mavysavydav:
**DEPENDENT ON**  - https://github.com/feast-dev/feast/pull/1899

**What this PR does / why we need it**: https://github.com/feast-dev/feast/issues/1907

**Which issue(s) this PR fixes**: https://github.com/feast-dev/feast/issues/1907

Fixes #1907

```release-note
When users do my_feature_view[[feature_a, ...]], a FeatureView will be returned rather than a FeatureViewProjection. This doesn't really impact anything from a usage perspective and users don't need to make any adjustments.  Everything will work the same as before. 
```


sdk/python/feast/feature_store.py
```diff
@@ -1017,6 +1013,30 @@ def _augment_response_with_on_demand_transforms(
...
+        features: Optional[Union[List[str], FeatureService]],
+        allow_cache=False,
+        hide_dummy_entity: bool = True,
+    ) -> List[FeatureView]:
+
+        passed_in_feature_views = (
+            {view.name: view for view in features.feature_views}
+            if isinstance(features, FeatureService)
+            else {}
+        )
```
codyjlin, line 1027: FYI lint complains here:
```
feast/feature_store.py:1126: error: "str" has no attribute "features"
feast/feature_store.py:1127: error: "str" has no attribute "name"
```
even though we establish using `isinstance` that `features` is a `FeatureService` and that `features.feature_views` can thus only be of type `List[FeatureView]`.... not sure what the best way to resolve this is.

felixwang9817:
assigned to you @woop as part of shadow entities

sdk/python/feast/feature_store.py
mavysavydav: resolved, _group_feature_refs now only has to handle feature_ref strings as opposed to also FeatureServices

protos/feast/core/FeatureService.proto
```diff
@@ -25,13 +27,19 @@ message FeatureServiceSpec {
 
   // List of features that this feature service encapsulates.
   // Stored as a list of references to other features views and the features from those views.
-  repeated FeatureViewProjection features = 3;
```
achals, line 28: We should not reuse the id for this - it will cause existing feature services to be unreadable by new versions of feast.

We should remove this field and `reserve` it's id instead for safety.

protos/feast/core/FeatureService.proto
```diff
@@ -25,13 +27,19 @@ message FeatureServiceSpec {
...
+  repeated string features = 3;
+
+  repeated FeatureTable feature_tables = 4;
+
+  repeated FeatureView feature_views = 5;
+
+  repeated OnDemandFeatureView on_demand_feature_views = 6;
 
   // User defined metadata
-  map<string,string> tags = 4;
```
achals, line 31: Same here - ids should be immutable.

achals (changes requested):

protos/feast/core/FeatureService.proto
mavysavydav: gotcha sg

achals (approved):
/lgtm

protos/feast/core/FeatureService.proto
```diff
@@ -25,7 +30,13 @@ message FeatureServiceSpec {
 
   // List of features that this feature service encapsulates.
   // Stored as a list of references to other features views and the features from those views.
-  repeated FeatureViewProjection features = 3;
+  repeated string features = 6;
```
woop, line 33: so is there a convention here like `feature_table:feature`?

protos/feast/core/FeatureService.proto
```diff
@@ -25,7 +30,13 @@ message FeatureServiceSpec {
...
+
+  repeated FeatureTable feature_tables = 7;
```
woop, line 35: I don't understand what these new objects are supposed to contain. Don't we just need `features`?

sdk/python/feast/feature_store.py
```diff
@@ -1017,6 +1013,30 @@ def _augment_response_with_on_demand_transforms(
                     ] = GetOnlineFeaturesResponse.FieldStatus.PRESENT
         return OnlineResponse(GetOnlineFeaturesResponse(field_values=result_rows))
 
+    def _get_feature_views_to_use(
+        self,
+        features: Optional[Union[List[str], FeatureService]],
```
woop, line 1018: At which point do we pass a `FeatureService` to this method?

sdk/python/feast/feature_store.py
```diff
@@ -804,8 +800,8 @@ def get_online_features(
             >>> online_response_dict = online_response.to_dict()
         """
         _feature_refs = self._get_features(features, feature_refs)
-        all_feature_views = self._list_feature_views(
-            allow_cache=True, hide_dummy_entity=False
+        all_feature_views = self._get_feature_views_to_use(
```
woop, line 803: is this really `all feature views` if we are filtering the feature views?

sdk/python/feast/feature_store.py
```diff
@@ -1017,6 +1013,30 @@ def _augment_response_with_on_demand_transforms(
...
+        hide_dummy_entity: bool = True,
+    ) -> List[FeatureView]:
+
+        passed_in_feature_views = (
+            {view.name: view for view in features.feature_views}
+            if isinstance(features, FeatureService)
+            else {}
+        )
+
+        all_feature_views = [
```
woop, line 1029: This name `all_feature_views` seems inaccurate if we are filtering.

woop, line 1029: It's not immediately obvious to me what is happening here. We are querying the registry using `_list_feature_views` then filtering based on `passed_in_feature_views`, so why do we need `passed_in_feature_views` then?

protos/feast/core/FeatureService.proto
woop: Also, would it make more sense to use a feature reference instead of a string?

woop:
@mavysavydav btw you are introducing a breaking change to the API, so you should add a release note stating the change and add a compat breaking label

protos/feast/core/FeatureService.proto
mavysavydav: FeatureViews and OnDemandFeatureViews will carry the info of the modified fields such as modified name. I also added a field for FeatureTable for consistency.

sdk/python/feast/feature_store.py
mavysavydav: in both get_historical_features and get_online_features

sdk/python/feast/feature_store.py
mavysavydav: the logic here has changed in the new commits.

mavysavydav: the logic before may have been a bit confusing. The logic in the new commits should make it clear we're not actually filtering. All the FVs from the registry are present

protos/feast/core/FeatureService.proto
mavysavydav: yep that's the convention. I'll add a note on that expectation in the docstring. It should be a string instead of feature reference since we need to know which feature view the feature name is associated with.

woop: A feature reference contains a reference to its associated view/table https://github.com/feast-dev/feast/blob/master/protos/feast/serving/ServingService.proto#L52. I actually think the FeatureViewProjection object has a good structure for this use case, even if we remove it from the rest of the user facing API. Wouldn't it be safer and more convenient to split views/tables from feature names vs having to rely on string conventions?

woop:
What happened to setup.py?

protos/feast/core/FeatureService.proto
woop: Looking at this now, it doesn't look like the right approach. Wouldn't we run into consistency problems if we store copies of feature view objects with different configurations?

The original problem was 
> The trouble right now is that at query time, even though the FV is one concept in the users mind when use get_historical_features or get_online_features, but it actually splits into a different class when the getitem magic method is triggered by my_fv[[<my-feature-selections]] and it becomes a FVProjection. This means if users wanted to transform their FV, we'd have to support these methods on both the FV class and FVProjection class.

This is both a user facing API problem and a code duplication problem, but I am beginning to think that having an object like a FeatureViewProjection within the protos might be the right approach. Assuming the user could modify a FeatureView at their own whims, and assuming we could convert that final feature view into something like

```proto
message FeatureViewProjection {
  string feature_view_name = 1;
  repeated FeatureSpecV2 feature_columns = 2;
  string feature_view_name_alias = 3;
  something entity_mapping_etc = 4;
}
```

Wouldn't that both solve the consistency problem in the registry (normalized instead of denormalized), as well as the user facing API problem where we have to maintain FeatureViewProjections?

sdk/python/feast/feature_store.py
```diff
@@ -318,10 +318,11 @@ def _get_features(
...
         _feature_refs: List[str]
         if isinstance(_features, FeatureService):
-            # Get the latest value of the feature service, in case the object passed in has been updated underneath us.
-            _feature_refs = _get_feature_refs_from_feature_services(
-                self.get_feature_service(_features.name)
-            )
+            _feature_refs = self.get_feature_service(_features.name).features
+            if _features.features != _feature_refs:
+                raise ValueError(
+                    "FeatureService object that's passed in is inconsistent with version from Registry."
```
woop, line 324: I am not sure if raising an exception here is the right approach. We are getting the latest version back. The reason we query the registry is because it may have a newer version. If we go with this approach of raising an exception, we might run into a case where one person keeps updating the registry and another person keeps getting exceptions. 

I think it's sufficient to print a warning.

Also, we can be clearer than "FeatureService object that's passed in is inconsistent with version from Registry.". What about "A newer version of the FeatureService \"{name}"\ has been found.". What do you think?

sdk/python/feast/feature_store.py
```diff
@@ -1018,6 +1018,30 @@ def _augment_response_with_on_demand_transforms(
...
+        all_feature_views = {
+            fv.name: fv
+            for fv in self._list_feature_views(allow_cache, hide_dummy_entity)
+        }
+
+        if isinstance(features, FeatureService):
+            for fv in features.feature_views:
+                if fv.name not in all_feature_views:
+                    raise ValueError(
+                        f"{fv.name} used in the FeatureService is not in the registry."
```
woop, line 1037: We're trying to make our exceptions clearer for users. If I read an exception like this it isn't really clear to me what to do as a beginner. What about

Something like
`"The provided feature service {name} contains a reference to a feature view {fv_name} which doesn't exist. Please make sure that you have created the feature view {fv_name} and that you have registered it by running \"apply\""`

mavysavydav: sounds good

sdk/python/feast/feature_store.py
mavysavydav: sounds good. I'm doing this - 

```
                warnings.warn(
                    "The FeatureService object that has been passed in as an argument is"
                    "inconsistent with the version from Registry. Potentially a newer version"
                    "of the FeatureService has been applied to the registry.")
```

sdk/python/feast/feature_service.py
```diff
@@ -54,18 +52,20 @@ def __init__(
...
+            elif isinstance(feature_grouping, FeatureView) or isinstance(
+                feature_grouping, OnDemandFeatureView
             ):
-                self.features.append(FeatureViewProjection.from_definition(feature))
-            elif isinstance(feature, FeatureViewProjection):
-                self.features.append(feature)
+                self.feature_view_projections.append(feature_grouping.projection)
             else:
-                raise ValueError(f"Unexpected type: {type(feature)}")
+                raise ValueError(f"Unexpected type: {type(feature_grouping)}")
```
woop, line 67: Nitpick: Would it make sense to expand this exception to be more descriptive? I can imagine people running `apply` and seeing this exception and not knowing where the bad type is being used. What about `The feature service {fs_name} has been provided with an invalid type {type(feature_grouping)} as part of the \"features\" argument`?

sdk/python/feast/feature_service.py
```diff
@@ -104,17 +106,20 @@ def from_proto(feature_service_proto: FeatureServiceProto):
...
-            ],
+            features=[],
             tags=dict(feature_service_proto.spec.tags),
             description=(
                 feature_service_proto.spec.description
                 if feature_service_proto.spec.description != ""
                 else None
             ),
         )
+        fs.feature_view_projections.extend(
```
woop, line 117: I'm curious why you moved this code out of the constructor. Don't feel strongly about it though.

sdk/python/feast/feature_view.py
```diff
@@ -379,3 +388,23 @@ def infer_features_from_batch_source(self, config: RepoConfig):
                     "FeatureView",
                     f"Could not infer Features for the FeatureView named {self.name}.",
                 )
+
+    def set_projection(self, feature_view_projection: FeatureViewProjection) -> None:
+        """
+        Setter for the projection object held by this FeatureView. Performs checks to ensure
```
woop, line 394: Nitpick: These comments assume the user knows exactly what a projection is. I think it's important to define what a projection is.

sdk/python/feast/feature_view.py
```diff
@@ -379,3 +388,23 @@ def infer_features_from_batch_source(self, config: RepoConfig):
...
+        the projection is consistent with this FeatureView before doing the set.
+
+        Args:
+            feature_view_projection: The FeatureViewProjection object to set this FeatureView's
+                'projection' field to.
+        """
+        assert feature_view_projection.name == self.name
```
woop, line 401: We should not be using `assert` since it can be disabled inadvertently through optimization flags.

woop (approved):

PR #1900, opened 2021-09-22
Create & teardown Lambda & API Gateway resources for serverless feature server

tsotnet:
Create & teardown Lambda & API Gateway resources for serverless feature server

**What this PR does / why we need it**: Here we implement update_infra and teardown_infra methods of AwsProvider, creating and destroying lambda & API gateway resources.

Here's what `feast apply` looks like:
```
Deploying feature server...
  Creating AWS Lambda...
  Creating AWS API Gateway...
Registered entity driver_id
Registered feature view driver_hourly_stats
Deploying infrastructure for driver_hourly_stats
Deploying feature server...
```

And here's what `feast teardown` looks like:
```
Tearing down feature server...
  Tearing down AWS Lambda...
  Tearing down AWS API Gateway...
```

Currently `feast endpoint` isn't implemented, so we don't show the HTTP endpoint anywhere, although I checked it from AWS Console and sending some cURL requests worked as expected.

Fixes #

```release-note
NONE
```


sdk/python/feast/constants.py
```diff
@@ -17,4 +17,10 @@
 # Maximum interval(secs) to wait between retries for retry function
 MAX_WAIT_INTERVAL: str = "60"
 
-AWS_LAMBDA_FEATURE_SERVER_IMAGE = "feastdev/feature-server"
+AWS_LAMBDA_FEATURE_SERVER_IMAGE = "feastdev/feature-server:aws"
+
+# feature_store.yaml environment variable name for remote feature server
+SERVER_CONFIG_BASE64_ENV_NAME: str = "FEAST_CONFIG_BASE64"
```
achals, line 23: nit, can you have this variable name container `FEATURE_STORE_YAML`  so we know what it's referring to? `Config` by itself is a bit ambiguous.

sdk/python/feast/infra/aws.py
```diff
@@ -1,23 +1,165 @@
...
 from pathlib import Path
 from tempfile import TemporaryFile
+from typing import Sequence, Union
 from urllib.parse import urlparse
 
 from colorama import Fore, Style
 
 import feast
-from feast.constants import AWS_LAMBDA_FEATURE_SERVER_IMAGE
+from feast import Entity, FeatureTable, FeatureView, __version__
```
achals, line 13: nit, in internal code we should be using fully qualified import statements. i.e. 
```
```

instead of 
```
from feast import Entity
```

sdk/python/feast/infra/aws.py
```diff
@@ -1,23 +1,165 @@
...
+except ImportError as e:
+    from feast.errors import FeastExtrasDependencyImportError
+
+    raise FeastExtrasDependencyImportError("aws", str(e))
+
 
 class AwsProvider(PassthroughProvider):
-    def _upload_docker_image(self) -> None:
+    def _get_lambda_name(self):
+        return f"feast-python-server-{__version__.replace('+', '_').replace('.', '_')}"
```
achals, line 36: Shouldn't this lambda name have the project name in it somewhere? I think that would be useful for debugging.

sdk/python/feast/infra/aws.py
```diff
@@ -1,23 +1,165 @@
...
+                    Role=self.repo_config.feature_server.execution_role_name,
+                    Code={"ImageUri": image_uri},
+                    PackageType="Image",
+                    MemorySize=1769,
+                    Environment={
+                        "Variables": {
+                            SERVER_CONFIG_BASE64_ENV_NAME: config_base64,
+                            FEAST_USAGE: "False",
+                        }
+                    },
```
achals, line 84: Should these be configurable by users? Maybe not needed for this PR, but something to keep in mind.

sdk/python/feast/infra/aws.py
```diff
@@ -1,23 +1,165 @@
...
+            if not api:
+                # If the API Gateway doesn't exist, create it
+                print("  Creating AWS API Gateway...")
+                api = api_gateway_client.create_api(
+                    Name=resource_name,
+                    ProtocolType="HTTP",
+                    Target=function["FunctionArn"],
+                    RouteKey="POST /get-online-features",
+                )
+                assert api
```
achals, line 115: super nitty, we should raise modeled exceptions for these things instead of `AssertionErrors`.

sdk/python/feast/infra/aws.py
```diff
@@ -1,23 +1,165 @@
...
+
+            resource_name = self._get_lambda_name()
+            lambda_client = boto3.client("lambda")
+            api_gateway_client = boto3.client("apigatewayv2")
+            function = aws_utils.get_lambda_function(lambda_client, resource_name)
+
+            if function is None:
+                # If the Lambda function does not exist, create it.
+                print("  Creating AWS Lambda...")
+                lambda_client.create_function(
```
achals, line 73: AWS resources can be tagged: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda.html#Lambda.Client.create_function

we should add some common tags for ease of use. I recommend something like:
```
{
    "feast-owned": 'true",
    "project": project_id
}
```

Maybe also the python version for feast. Probably also true for the API gateway or any other resource we create and expect to manage..

sdk/python/feast/infra/utils/aws_utils.py
```diff
@@ -406,3 +406,89 @@ def unload_redshift_query_to_df(
...
+        lambda_client: AWS Lambda client.
+        function_name: Name of the AWS Lambda function.
+
+    Returns: Either a dictionary containing the get_function API response, or None if it doesn't exist.
+
+    """
+    try:
+        return lambda_client.get_function(FunctionName=function_name)["Configuration"]
+    except ClientError as ce:
+        # If the resource could not be found, return False.
```
achals, line 424:
```suggestion
        # If the resource could not be found, return None.
```

sdk/python/feast/infra/utils/aws_utils.py
```diff
@@ -406,3 +406,89 @@ def unload_redshift_query_to_df(
...
+        api_gateway_client: API Gateway V2 Client.
+        api_gateway_name: Name of the API Gateway function.
+
+    Returns: Either a dictionary containing the get_api response, or None if it doesn't exist
+
+    """
+    response = api_gateway_client.get_apis()
+    apis = response.get("Items", [])
+
+    while True:
```
achals, line 460: We should probably limit how often we page through this API response (something sufficiently high like 10 or something).

achals (commented):
I had some nits, but the PR looks mostly good. Are you planning on adding tests in this pass or is that going to be punted to a subsequent PR?

sdk/python/feast/infra/aws.py
felixwang9817: Yeah that's definitely a good point. I'll keep this in mind for the next PR.

felixwang9817:
@achals yes I plan on doing tests in a later PR (since tests are not necessary for a the serverless MVP).

achals:
> @achals yes I plan on doing tests in a later PR (since tests are not necessary for a the serverless MVP).

Cool. I had one more nit but looks good otherwise.

achals (approved):

achals (approved):
/lgtm

PR #1905, opened 2021-09-23
Fix issue with feature views being detected as duplicated when imported

### LINKED ISSUE #1895: FeatureView is considered "duplicated" during `feast apply` if its imported in another file
## Expected Behavior 
If I create a FeatureService in its own file, I need to import the desired FeatureViews. By doing so, `feat apply` is going to consider that `feature_view_a` is duplicated

## Current Behavior
If a FeatureView is imported in another file, `feast_apply` should not consider it as a duplicated FeatureView

Today, the current error is
```bash
ValueError: More than one feature view with name customer_raw_creation_info found.
Please ensure that all feature view names are unique.
It may be necessary to ignore certain files in your feature repository by using a .feastignore file.
```

## Steps to reproduce
1. Create a FeatureView name `feature_view_a` in its own file
2. Create a FeatureService in its own file and import `feature_view_a`
3. Run `feast apply`

### Specifications

- Version: 0.12.1
- Platform: MacOS bigSur
- Subsystem:

achals:

Fixes #1895 

```release-note
none
```

### DIFF
--- a/sdk/python/feast/entity.py
+++ b/sdk/python/feast/entity.py
@@ -19,6 +19,7 @@ import yaml
 from google.protobuf import json_format
 from google.protobuf.json_format import MessageToDict, MessageToJson
 
+from feast.importer import get_calling_file_name
 from feast.loaders import yaml as feast_yaml
 from feast.protos.feast.core.Entity_pb2 import Entity as EntityV2Proto
 from feast.protos.feast.core.Entity_pb2 import EntityMeta as EntityMetaProto
@@ -77,10 +78,7 @@ class Entity:
         self._created_timestamp: Optional[datetime] = None
         self._last_updated_timestamp: Optional[datetime] = None
 
-        stack = inspect.stack()
-        # Get two levels up from current, to ignore usage.py
-        previous_stack_frame = stack[2]
-        self.defined_in = previous_stack_frame.filename
+        self.defined_in = get_calling_file_name(inspect.stack())
 
     def __eq__(self, other):
         if not isinstance(other, Entity):
--- a/sdk/python/feast/entity.py
+++ b/sdk/python/feast/entity.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import inspect
 from datetime import datetime
 from typing import Dict, Optional
 
@@ -49,6 +49,8 @@ class Entity:
     _created_timestamp: Optional[datetime]
     _last_updated_timestamp: Optional[datetime]
 
+    defined_in: str
+
     @log_exceptions
     def __init__(
         self,
@@ -74,6 +76,11 @@ class Entity:
 
         self._created_timestamp: Optional[datetime] = None
         self._last_updated_timestamp: Optional[datetime] = None
+
+        stack = inspect.stack()
+        # Get two levels up from current, to ignore usage.py
+        previous_stack_frame = stack[2]
+        self.defined_in = previous_stack_frame.filename
 
     def __eq__(self, other):
         if not isinstance(other, Entity):
--- a/sdk/python/feast/feature_table.py
+++ b/sdk/python/feast/feature_table.py
@@ -22,6 +22,7 @@ from google.protobuf.timestamp_pb2 import Timestamp
 
 from feast.data_source import DataSource, KafkaSource, KinesisSource
 from feast.feature import Feature
+from feast.importer import get_calling_file_name
 from feast.loaders import yaml as feast_yaml
 from feast.protos.feast.core.FeatureTable_pb2 import FeatureTable as FeatureTableProto
 from feast.protos.feast.core.FeatureTable_pb2 import (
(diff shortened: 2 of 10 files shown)


sdk/python/feast/repo_operations.py
```diff
@@ -107,15 +107,20 @@ def parse_repo(repo_root: Path) -> ParsedRepo:
         for attr_name in dir(module):
             obj = getattr(module, attr_name)
             if isinstance(obj, FeatureTable):
-                res.feature_tables.append(obj)
+                if obj.defined_in is not None and obj.defined_in == module.__file__:
```
MattDelac, line 110: Can module.__file__ be None ?

sdk/python/feast/entity.py
```diff
@@ -75,6 +77,11 @@ def __init__(
         self._created_timestamp: Optional[datetime] = None
         self._last_updated_timestamp: Optional[datetime] = None
 
+        stack = inspect.stack()
+        # Get two levels up from current, to ignore usage.py
+        previous_stack_frame = stack[2]
+        self.defined_in = previous_stack_frame.filename
```
MattDelac, line 83: Should we abstract that in an helper function and also test it in a test file ?
Don't think we need to test each class but at least the helper function

What do you think ?

MattDelac (commented):
Nice !
I tried couple of the inspect functions like `getsourcefile()` in the other PR but could not make it work.
It makes sense to track where the objects are created 👍

sdk/python/feast/infra/online_stores/redis.py
```diff
@@ -188,6 +188,7 @@ def online_read(
 
         for entity_key in entity_keys:
             redis_key_bin = _redis_key(project, entity_key)
+            print(redis_key_bin)
```
woop, line 191: should this be removed?

sdk/python/feast/repo_operations.py
woop: we should maybe also raise exceptions in case defined_in is None

woop:
Just to play devil's advocate here. Tracking the file names does add complexity to our objects. Is there any reason we cant import all the objects and use built in functionality `__hash__()` and `__eq__` to ensure that they are the same if their names are the same, then we can dedupe/remove the additional ones?

sdk/python/feast/infra/online_stores/redis.py
achals: Yes

achals:
> Just to play devil's advocate here. Tracking the file names does add complexity to our objects. Is there any reason we cant import all the objects and use built in functionality `__hash__()` and `__eq__` to ensure that they are the same if their names are the same, then we can dedupe/remove the additional ones?

IMO updating `__hash__` and `__eq__` is more fragile since we may add/remove fields without updating those methods, and end up in problematic situations. I think the source file is a more robust solution and less likely to break with changes.

sdk/python/feast/entity.py
adchia: A helper class probably makes sense since there's some state tracking being done. wdyt @achals ?

MattDelac: > A helper class

yeah even better 👍

achals: Done

sdk/python/feast/feature_table.py
```diff
@@ -64,6 +66,11 @@ def __init__(
         self._created_timestamp: Optional[Timestamp] = None
         self._last_updated_timestamp: Optional[Timestamp] = None
 
+        stack = inspect.stack()
+        # Get two levels up from current, to ignore usage.py
+        previous_stack_frame = stack[2]
+        self.defined_in = previous_stack_frame.filename
```
MattDelac, line 72: Sounds like missing one `get_calling_file_name()` here

sdk/python/feast/infra/online_stores/redis.py
MattDelac, line 191: To be removed ?

sdk/python/feast/feature_table.py
achals: oh yeah missed this

sdk/python/feast/repo_operations.py
achals: Done. And no, module.__file__ cannot be None

adchia (approved):
/lgtm

MattDelac:
👋 
I just tried it on our repo and we shipped with a bug. The problem is that we save in `defined_in` one level too high.

Look at how my `feast apply` behaves
image

`module.__file__` is correct but `obj.defined_in` is not.
The entity "customer" is, indeed, defined in `pano/entities/customer.py`

woop:
Whether or not there is a bug here, I am not 100% convinced by the approach taken here. There are many reasons to support object equality, deduplication just being one of them. Reaching into the filesystem seems like it will be a lot more prone to errors since its OS/env dependent.

achals:
> Whether or not there is a bug here, I am not 100% convinced by the approach taken here. There are many reasons to support object equality, deduplication just being one of them. Reaching into the filesystem seems like it will be a lot more prone to errors since its OS/env dependent.

I'm coming around to this view point after this bug as well. It's pretty easy to get wrong. I'm going to undo this change and use equality based deduping.

PR #1911, opened 2021-09-28
Add final_output_feature_names in Query context to avoid SELECT * EXCEPT at the end

MattDelac:

Contrary to BigQuery, other offline stores like Trino & Redshit don't support `SELECT * EXCEPT(...)`
Therefore, we should inject the list of the columns we want at the end of an historical retrieval.

It also makes it easier to understand which columns are included. It's often hard to understand those SELECT * EXCEPT as nothing is explicit

Fixes #

```release-note
We now pass the list of the columns desired as the output of an historical retrieval in tje Jinja template
```


sdk/python/feast/infra/offline_stores/redshift.py
```diff
@@ -478,19 +479,23 @@ def _upload_entity_df_and_get_entity_schema(
 */
 {{ featureview.name }}__latest AS (
     SELECT
-        {{featureview.name}}__entity_row_unique_id,
```
MattDelac, line 481: This should have fail with the current unit tests ...

woop:
Strange that the tests didn't fail. Your proposed change here looks good to me.

MattDelac:
I don't understand why the tests fail for Redshift. 

The difference between the 2 files (BigQuery on the left vs Redshift on the right) are minimal
image
image

MattDelac:
> I don't understand why the tests fail for Redshift.
> 
> The difference between the 2 files (BigQuery on the left vs Redshift on the right) are minimal image image

Also the fact that there is a `CONCAT` missing in the redshift template is disturbing on my side. I will dig into the internals of Redshift to understand if this is implicit in some manner

sdk/python/feast/infra/offline_stores/redshift.py
```diff
@@ -201,8 +196,6 @@ def __init__(
             config: Feast repo config
             full_feature_names: Whether to add the feature view prefixes to the feature names
             on_demand_feature_views: A list of on demand transforms to apply at retrieval time
-            drop_columns: Optionally a list of columns to drop before unloading to S3.
-                          This is a convenient field, since "SELECT ... EXCEPT col" isn't supported in Redshift.
```
MattDelac, line 205: I will be honest with you but this concept of dropping columns is bad. If some SQL query engines do not support `SELECT * EXCEPT`, then let's not support it !

It's very weird to add this implementation outside of the SQL template. It makes things very different between BigQuery and other offline query engines

Thoughts about keeping it ?

MattDelac: I can be missing something, so feel free to change my mind 🙂

MattDelac:
Beside the formatting issue I am facing, this PR is ready for reviews

Thanks in advance 🙏

MattDelac:
Fixed the linter issue. PR is ready for final review 🙏

sdk/python/feast/infra/offline_stores/redshift.py
adchia: +1 to this change. Pretty sure it's just BQ that supports EXCEPT in this way

sdk/python/feast/infra/offline_stores/redshift.py
adchia: Think the tests you added might have been accidentally removed... Mind adding them back to the universal_historical_retrieval test?

sdk/python/feast/infra/offline_stores/redshift.py
adchia: it also is clearer to understand the query output

adchia (commented):
generally LG to me. pending we should revive the old test you wrote cc @achals

PR #1913, opened 2021-09-29
Use set when parsing repos to prevent duplicates

achals:

#1905 didn't solve the problem - so this is a simpler approach that is probably more robust.

Fixes #

```release-note
none
```

### DIFF
--- a/sdk/python/feast/entity.py
+++ b/sdk/python/feast/entity.py
@@ -11,7 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import inspect
 from datetime import datetime
 from typing import Dict, Optional
 
@@ -19,7 +18,6 @@ import yaml
 from google.protobuf import json_format
 from google.protobuf.json_format import MessageToDict, MessageToJson
 
-from feast.importer import get_calling_file_name
 from feast.loaders import yaml as feast_yaml
 from feast.protos.feast.core.Entity_pb2 import Entity as EntityV2Proto
 from feast.protos.feast.core.Entity_pb2 import EntityMeta as EntityMetaProto
@@ -50,8 +48,6 @@ class Entity:
     _created_timestamp: Optional[datetime]
     _last_updated_timestamp: Optional[datetime]
 
-    defined_in: str
-
     @log_exceptions
     def __init__(
         self,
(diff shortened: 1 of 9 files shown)


sdk/python/feast/entity.py
```diff
@@ -78,7 +74,8 @@ def __init__(
         self._created_timestamp: Optional[datetime] = None
         self._last_updated_timestamp: Optional[datetime] = None
 
-        self.defined_in = get_calling_file_name(inspect.stack())
+    def __hash__(self) -> int:
+        return hash(self.name)
```
woop, line 78: How would we detect and raise an exception when users provide two objects with the same name that are different?

achals: Having two entities with the same name is a bug though, isn't it?

achals: I don't think we support any FV/FT/entity/FS having the same name right now anyway

achals: Okay, I checked the code and we verify unique names for FVs but not other objects. I don't know what the behaviour would be for `FeatureStore.get_feature_view` if names can be duplicated. Plus, feature_refs would be busted (in the case of feature views and feature tables). wdyt?

achals: My earlier comments completely misunderstood your point -fixing using `id()`.

MattDelac (approved):
Looks good to me

Thanks for the test, this makes it safer

I can also confirm that based on the latest commit of this PR, it works on our side 🚀 
image

sdk/python/tests/integration/registration/test_cli_apply_duplicated_featureview_names.py
```diff
@@ -45,6 +45,49 @@ def test_cli_apply_duplicated_featureview_names() -> None:
         )
 
 
+def test_cli_apply_imported_featureview() -> None:
```
woop, line 48: Should we add a test that ensures that duplicate names with different contents raises an exception?

achals: There's already a test for such a case

achals: done

sdk/python/tests/integration/registration/test_cli_apply_duplicated_featureview_names.py
```diff
@@ -45,6 +45,105 @@ def test_cli_apply_duplicated_featureview_names() -> None:
...
+            "   batch_source=driver_hourly_stats,\n"
+            "   tags={'dummy': 'true'})\n"
+        )
+
+        rc, output = runner.run_with_output(["apply"], cwd=repo_path)
+
+        assert rc != 0
+        import pdb
+
+        pdb.set_trace()
```
MattDelac, line 141: Should that be removed ?

PR #1916, opened 2021-09-30
Delete tables

### LINKED ISSUE #1902: Clean up entity dataframes or temporary tables by default
Currently when a user runs `get_historical_features()` Feast will upload or create an entity dataframe in the offline store. Feast will then build a training dataset and return that to the user. Feast does not clean up the temporary tables, but often sets an expiry. This means that a data warehouse constantly has a set of temporary tables waiting to expire.

Delete entity dataframe tables after a training dataset has been created (as well as other temporary tables in the creation process).

felixwang9817:
**What this PR does / why we need it**: Currently, entity dataframes are uploaded to Bigquery and Redshift and are not always cleaned up properly. This PR ensures they are always cleaned up. 

Fixes #1902 

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -1,6 +1,7 @@
+import contextlib
 import uuid
 from datetime import date, datetime, timedelta
-from typing import Dict, List, Optional, Union
+from typing import Callable, ContextManager, Dict, Iterator, List, Optional, Union
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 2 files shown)


sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -122,38 +123,47 @@ def get_historical_features(
...
+                query_template=MULTIPLE_FEATURE_VIEW_POINT_IN_TIME_JOIN,
+                full_feature_names=full_feature_names,
+            )
+
+            try:
+                yield query
+            finally:
+                # Asynchronously clean up the uploaded Bigquery table, which will expire
+                # if cleanup fails
+                client.query(query=f"DROP TABLE {table_reference}")
```
achals, line 163: should probably be `DROP IF EXISTS`. Also, I think the bigquery client has a `delete_table` method too.

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -122,38 +123,47 @@ def get_historical_features(
...
+                left_table_query_string=table_reference,
+                entity_df_event_timestamp_col=entity_df_event_timestamp_col,
+                query_template=MULTIPLE_FEATURE_VIEW_POINT_IN_TIME_JOIN,
+                full_feature_names=full_feature_names,
+            )
+
+            try:
+                yield query
+            finally:
+                # Asynchronously clean up the uploaded Bigquery table, which will expire
```
felixwang9817, line 161: Note: for Redshift, we clean up synchronously. I decided to leave it asynchronous here while also preserving the 30-min expiration, instead of forcing a synchronous cleanup.

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -187,15 +206,16 @@ def on_demand_feature_views(self) -> Optional[List[OnDemandFeatureView]]:
...
+        with self._query_generator() as query:
+            df = self.client.query(query).to_dataframe(create_bqstorage_client=True)
+            return df
 
     def to_sql(self) -> str:
         """
         Returns the SQL query that will be executed in BigQuery to build the historical feature table.
         """
-        return self.query
+        with self._query_generator() as query:
```
felixwang9817, line 217: Note: technically the query generated here will not be the exact query used to build the historical feature table, since the randomly generated table name will be different. I couldn't think of a better way to handle `to_sql` given that we switched to a context manager - open to suggestions.

sdk/python/feast/infra/offline_stores/redshift.py
```diff
@@ -153,16 +153,17 @@ def query_generator() -> Iterator[str]:
...
+            try:
+                yield query
+            finally:
+                # Always clean up the uploaded Redshift table
+                aws_utils.execute_redshift_statement(
+                    redshift_client,
+                    config.offline_store.cluster_id,
+                    config.offline_store.database,
+                    config.offline_store.user,
+                    f"DROP TABLE {table_name}",
```
achals, line 165: This should be some form of `DELETE IF EXISTS` too, but I don't think Redshift SQL supports that

sdk/python/feast/infra/offline_stores/bigquery.py
achals: I think that's fine.

sdk/python/feast/infra/offline_stores/bigquery.py
achals: Hmmm in that case we run think we should either return a sql strinv with the table name a parameter that canbe filled the, or we should accept it as a param forthr to_sql method. As it stands a full SQL string that does is not expected to execute correctly seems like a bug

achals (approved):
/lgtm

PR #1919, opened 2021-10-01
Add a caching step to our github actions

achals:

Github has a way of caching dependencies on a per-job level, which can be used to speed up actions/builds by caching dependencies and builds. 

Our dependencies don't change that often - so we can cache the step of installing and setting up our python environment. 

Example:
- Before Caching: https://github.com/feast-dev/feast/runs/3773858745?check_suite_focus=true
- After Caching: https://github.com/feast-dev/feast/runs/3773917413?check_suite_focus=true

`Install Dependencies` went from 1m19s to 25s.

Fixes #

```release-note
none
```

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -29,6 +29,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Setup Python
+        id: setup-python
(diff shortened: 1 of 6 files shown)


felixwang9817 (approved):
/lgtm

PR #1920, opened 2021-10-01
Update makefile to use pip installed dependencies

loftiskg:
First-time contributor here.  I noticed that when I was running through the list of things in CONTRIBUTORS.md to set up my dev environment, I ran `make lint` and `make format` and a whole bunch of files were changes.  Turns out that the black being used in my local setup was not the same the version of the one installed with dev dependencies.  Since `black` is called directly as a CLI command rather than as a python module in the make command, it uses my system default rather than the one installed via `pip install -e 'sdk/python[ci]'`.  This would also apply to the other python linting tools utilized in the make file as well (ie mypy, isort, etc).  So to ensure that the correct versions of everything are used I changed the makefile to execute black and other python installed CLI tools as python modules.

Fixes #

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,26 +53,26 @@ install-python:
 	python -m pip install -e sdk/python -U --use-deprecated=legacy-resolver
 
 benchmark-python:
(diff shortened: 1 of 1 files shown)


achals (approved):
/lgtm 


Thanks for the fix @loftiskg !

PR #1923, opened 2021-10-05
Refactor serverless

felixwang9817:
**What this PR does / why we need it**: This PR cleans up some of the AWS serverless deployments code.

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/constants.py
+++ b/sdk/python/feast/constants.py
@@ -17,7 +17,8 @@
 # Maximum interval(secs) to wait between retries for retry function
 MAX_WAIT_INTERVAL: str = "60"
 
-AWS_LAMBDA_FEATURE_SERVER_IMAGE = "feastdev/feature-server:aws"
+# Dockerhub name for the AWS Lambda feature server docker image.
+AWS_LAMBDA_FEATURE_SERVER_IMAGE = "feastdev/feature-server:aws-v0.1"
 
 # feature_store.yaml environment variable name for remote feature server
 FEATURE_STORE_YAML_ENV_NAME: str = "FEATURE_STORE_YAML_BASE64"
--- a/sdk/python/feast/infra/aws.py
+++ b/sdk/python/feast/infra/aws.py
@@ -40,9 +40,6 @@ except ImportError as e:
 
 
 class AwsProvider(PassthroughProvider):
-    def _get_lambda_name(self, project: str):
-        return f"feast-python-server-{project}-{__version__.replace('+', '_').replace('.', '_')}"
-
     def update_infra(
         self,
         project: str,
@@ -65,7 +62,6 @@ class AwsProvider(PassthroughProvider):
             image_uri = self._upload_docker_image(project)
             print("Deploying feature server...")
 
-            assert self.repo_config.repo_path
             if not self.repo_config.repo_path:
                 raise RepoConfigPathDoesNotExist()
             with open(self.repo_config.repo_path / "feature_store.yaml", "rb") as f:
@@ -102,7 +98,7 @@ class AwsProvider(PassthroughProvider):
                 )
                 function = aws_utils.get_lambda_function(lambda_client, resource_name)
                 if not function:
-                    raise AwsLambdaDoesNotExist()
+                    raise AwsLambdaDoesNotExist(resource_name)
             else:
                 # If the feature_store.yaml has changed, need to update the environment variable.
                 env = function.get("Environment", {}).get("Variables", {})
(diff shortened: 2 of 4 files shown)


sdk/python/feast/constants.py
```diff
@@ -17,7 +17,8 @@
 # Maximum interval(secs) to wait between retries for retry function
 MAX_WAIT_INTERVAL: str = "60"
 
-AWS_LAMBDA_FEATURE_SERVER_IMAGE = "feastdev/feature-server:aws"
+# Dockerhub name for the AWS Lambda feature server docker image.
+AWS_LAMBDA_FEATURE_SERVER_IMAGE = "feastdev/feature-server:aws-v0.1"
```
woop, line 21: btw, where does this version come from? Just want to figure out how we will be bumping these versions. Don't want to randomize you, but normally these versions are assigned dynamically through CI. 

So basically when CI runs you push `feastdev/feature-server-aws:{GIT SHA}` and `feastdev/feature-server-aws:{SEMVER}` and if it's the latest stable image, then you also push `feastdev/feature-server-aws:latest`. You cant have `feastdev/feature-server:aws-latest`, there is no automatic selection of the latest image if the tag isn't exactly `latest`. 

Then above becomes
```suggestion
AWS_LAMBDA_FEATURE_SERVER_IMAGE = "feastdev/feature-server-aws"
```
which selects the latest stable image. Or you can pin it to the current version as well.

sdk/python/feast/infra/aws.py
```diff
@@ -38,11 +39,14 @@
 
     raise FeastExtrasDependencyImportError("aws", str(e))
 
+_logger = logging.getLogger(__name__)
+_logger.setLevel(logging.INFO)
```
woop, line 43: should log levels be set here? I think logging should be globally configured right?

sdk/python/feast/infra/aws.py
```diff
@@ -38,11 +39,14 @@
...
+_handler = logging.StreamHandler()
+_handler.setLevel(logging.INFO)
+_logger.addHandler(_handler)
```
achals, line 47: You shouldn't need to do any of the `setLevel` or `addHandler` business here - you only need to get a logger using `getLogger` and you can use it out of the box.

Any configuration should be happening in `feast/__init__.py`, or potentially using a logging file.

sdk/python/feast/constants.py
felixwang9817: Got it, synced with Achal about this.

sdk/python/feast/cli.py
```diff
@@ -57,8 +57,12 @@ def format_options(self, ctx: click.Context, formatter: click.HelpFormatter):
     "-c",
     help="Switch to a different feature repository directory before executing the given subcommand.",
 )
+@click.option(
+    "--log",
```
achals, line 61: This should be `--log-level` or something?

sdk/python/feast/cli.py
```diff
@@ -68,6 +72,11 @@ def cli(ctx: click.Context, chdir: str):
     """
     ctx.ensure_object(dict)
     ctx.obj["CHDIR"] = Path.cwd() if chdir is None else Path(chdir).absolute()
+    try:
+        if log is not None:
+            logging.basicConfig(level=getattr(logging, log.upper()))
```
achals, line 77: Can you check `feast/__init__.py`? We're calling some `basicConfig` there too which should be collapsed into this method.

sdk/python/feast/cli.py
```diff
@@ -68,6 +72,11 @@ def cli(ctx: click.Context, chdir: str):
     """
     ctx.ensure_object(dict)
     ctx.obj["CHDIR"] = Path.cwd() if chdir is None else Path(chdir).absolute()
+    try:
+        if log is not None:
```
achals, line 76: What if `log` is not specified? What's the default configuration?

sdk/python/feast/cli.py
```diff
@@ -68,6 +73,15 @@ def cli(ctx: click.Context, chdir: str):
     """
     ctx.ensure_object(dict)
     ctx.obj["CHDIR"] = Path.cwd() if chdir is None else Path(chdir).absolute()
+    try:
+        level = getattr(logging, log_level.upper())
+        logging.basicConfig(
```
adchia, line 78: does this override the info levels you set for other files?

also mind changing the PR description to be more specific to logging refactors for serverless?

felixwang9817: I removed the per-file configuration. Now all loggers will conform to the level set here. And yes, changed PR description.

PR #1924, opened 2021-10-05
cache provider in feature store instance

DvirDukhan:
This PR saves the provider instance, and as such the offline and the online store instances. This will reduce the amount of time spent in functions involving them, as re-instantiation of these objects is no longer necessary. 

None

```release-note
None
```

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -76,6 +76,7 @@ class FeatureStore:
     config: RepoConfig
     repo_path: Path
     _registry: Registry
+    _provider: Provider
 
     @log_exceptions
     def __init__(
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -100,6 +100,7 @@ class FeatureStore:
 
         registry_config = self.config.get_registry_config()
         self._registry = Registry(registry_config, repo_path=self.repo_path)
+        self._provider = get_provider(self.config, self.repo_path)
 
     @log_exceptions
     def version(self) -> str:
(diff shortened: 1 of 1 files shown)


achals (approved):
/lgtm

sdk/python/feast/feature_store.py
```diff
@@ -118,7 +119,7 @@ def project(self) -> str:
 
     def _get_provider(self) -> Provider:
         # TODO: Bake self.repo_path into self.config so that we dont only have one interface to paths
-        return get_provider(self.config, self.repo_path)
+        return self._provider
```
adchia, line 122: only nit here is, what if e.g. feature_store.yaml changes?

an alternate approach might be to mimic what we do for the registry. i.e. have a config that specifies the cache TTL (which can be infinite), after which it does a refresh

DvirDukhan: wouldn't it force you to create a new instance? or "refresh" it?

adchia: yeah it would recreate the provider after the ttl expires.

as is, this would force users to restart the feature server in order to see changes from the yaml config, which seems fine as a default behavior. But I can see a world where someone wants the original behavior eg for debugging

woop: I think this use case is different. We need caching for the registry for online retrieval. In the case of the provider I dont see a downside to having users create a new feature store instance when config changes, or is there a good use case?

adchia: i'm cool with this going in as is, but i see value in having a feature server running but wanting to mess around with parameters to get it working.

adchia: could basically be also a binary debug mode flag that continuously reinstantiates if true instead of a real cache

adchia (approved):
/lgtm

Issue #1863, opened 2021-09-15
direct ingestion into the Online Store


vas28r13:
*Problem*
We'd like to be able to ingest data into the Online store from streaming sources. We think that supporting direct ingestion into the Online store could be lightweight solution to support ingesting data from streaming sources in the newer Feast versions.

In Feast v0.12, materialization from the Offline store is the way to get data into the Online store.

We believe we can manage the streaming processes on our end but would love to have a way to ingest data into the Online Store directly.

*Use Case*
a model process/service will get a stream of data and would need to keep track of the new/updated features in the Online feature store

*Code example*
`feature_store_object.ingest_df("feature_view_name", dataframe_object)`

Process 1
```
    feature_store = _get_feature_store()
    for feature_data in _kafka_consume():
        # clean the data
        features_df = _clean_features(feature_data)

        # ingest the data directly into the online store
        feature_store.ingest_df("stream_ingestion", features_df)
```

Process 2 (job)
```
feature_store = _get_feature_store()

# materialization from Offline sources
feature_store.materialize_incremental(end_date=datetime.now(timezone.utc))
```

*Potential Issue*
Since there are now 2 ways to ingest the data into the online store (materialization and direct), `online_write_batch` in the online store implementation needs to have logic to look at the `event_timestamp` since event time could be delayed. For example, `Process 2` happens after `Process 1`, but the event timestamp in the features ingested by `Process 1` are actually more recent so the features in Process 2 should not overwrite what is in the Online store even though it happens after `Process 1`

Let us know what you think. We implemented this locally for our use case but wonder if this is something that could be useful in the official Feast version as well.

woop:
@vas28r13 Thanks for writing this up. Overall it makes sense to me. 

> We believe we can manage the streaming processes on our end but would love to have a way to ingest data into the Online Store directly.

How would you ensure that streaming processes/jobs stay in sync with feature views in the feature store? It seems like there are broadly two approaches
1. Provide a Push API and have an upstream system push events to the API. Write those events into the FS. Feast doesn't manage the upstream job
2. Feast launches ingestion jobs which consume from a topic and write into an online store (this write may go through a Push API or could be direct to the storage layer).

(1) adds less responsibility to Feast, but also makes it harder to keep jobs in sync. In (1) Feast would be able to launch new consuming jobs, take down old ones, or update the schema of events that the job processes.

[...]

PR #1939, opened 2021-10-14
direct data ingestion into Online store

vas28r13:

Issue docs: https://github.com/feast-dev/feast/issues/1863

Directly ingest data into the Online store.

Use Case
a model process/service will get a stream of data and would need to keep track of the new/updated features in the Online feature store

Code example
```python
feature_store_object.ingest_df("my_feature_view_name", dataframe_object)
```

**Process 1**
```python
# load feature store
feature_store = _get_feature_store()
for feature_data in _kafka_consume():
    # clean the data
    features_df = _clean_features(feature_data)

    # ingest the data directly into the online store
    feature_store.ingest_df("my_feature_view_name", features_df)
```

**Process 2 (materialization job)**
```python
feature_store = _get_feature_store()

# materialization from Offline sources
feature_store.materialize_incremental(end_date=datetime.now(timezone.utc))
```

### DIFF
--- a/sdk/python/tests/integration/feature_repos/repo_configuration.py
+++ b/sdk/python/tests/integration/feature_repos/repo_configuration.py
@@ -243,7 +243,6 @@ def construct_test_environment(
             offline_store=offline_store_config,
             online_store=online_store,
             repo_path=repo_dir_name,
-            flags={"direct_ingest_to_online_store": True, "alpha_features": True},
         )
         fs = FeatureStore(config=config)
         # We need to initialize the registry, because if nothing is applied in the test before tearing down


sdk/python/feast/feature_store.py
```diff
@@ -809,6 +809,26 @@ def tqdm_builder(length):
                 feature_view, self.project, start_date, end_date
             )
 
+    @log_exceptions_and_usage
+    def write_to_online_store(
```
benny-hal, line 813: @vas28r13 thx for adding this capability, we use this capability so we can write inference data in a real-time flow. However, the `get_feature_view` does not use the `allow_cache` option of the registry. This causes a lookup to refresh the registry for every call to `write_to_online_store`. This blocks the new capability from being used real-time. Would it make sense to add a PR that fixes that? thx...

adchia: would you mind making an issue so we can track this?

benny-hal: done

PR #1951, opened 2021-10-15
Add public docs for entity aliasing

### LINKED ISSUE #1762: Shadow entities mapped properly to real entity with entity aliasing

codyjlin:

This adds public documentation for entity aliasing functionality to the [concepts/feature-view](https://docs.feast.dev/getting-started/concepts/feature-view) page. (Also, this is my first time contributing the public docs - not sure what the process is / if it's ok to do it through a PR like this.)

Fixes #1762

```release-note
NONE
```

### DIFF
--- a/docs/getting-started/concepts/feature-view.md
+++ b/docs/getting-started/concepts/feature-view.md
@@ -53,6 +53,50 @@ global_stats_fv = FeatureView(
 {% endtab %}
 {% endtabs %}
 
+## Entity aliasing
+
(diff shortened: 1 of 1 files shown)


adchia (approved):
/lgtm

PR #1952, opened 2021-10-17
Add created timestamp for feature views

### LINKED ISSUE #1933: FeatureViews creation time
## Expected Behavior 
Feature views should have the creation time (i.e., `created_timestamp`) at the first `feast apply`

## Current Behavior
Features Views do not have creation time at feature view creation

## Steps to reproduce
```shell
feast init fs
cd fs
feast apply
feast registry-dump
{
  "spec": {
    "name": "driver_id",
    "valueType": "INT64",
    "description": "driver id",
    "joinKey": "driver_id"
  },
  "meta": {}
}
{
  "spec": {
    "name": "driver_hourly_stats",
    "entities": [
      "driver_id"
    ],
    "features": [
      {
        "name": "conv_rate",
        "valueType": "FLOAT"
      },
      {
        "name": "acc_rate",
        "valueType": "FLOAT"
      },
      {
        "name": "avg_daily_trips",
        "valueType": "INT64"
      }
    ],
    "ttl": "86400s",
    "batchSource": {
      "type": "BATCH_FILE",
      "eventTimestampColumn": "event_timestamp",
      "createdTimestampColumn": "created",
      "fileOptions": {
        "fileUrl": "/home/user/fs/data/driver_stats.parquet"
      },
      "dataSourceClassType": "feast.infra.offline_stores.file_source.FileSource"
    },
    "online": true
  },
  "meta": {}
```
> Note that `meta` is empty.

### Specifications

- Version: 0.12+
- Platform: Ubuntu 20.04.3 LTS

## Possible Solution
Check if the feature view already have a creation date timpstamp (`feature_view.created_timestamp`) and then add `datetime.now()` if necessary. Possibly in the [apply feature view](https://github.com/feast-dev/feast/blob/d191ee89b10decb5e35a6d2f8ce84d68567d9b2f/sdk/python/feast/registry.py#L292) method

amommendes:
Adds created timestamp for feature views

Fixes #1933 

```release-note
Added created timestamp to Feature Views
```


achals:
> Hey @achals , I think this is not good with a nested `if`. However `BaseFeatureView` does not have `created_timestamp` and since `OnDemandFeatureView` does not need `created_timestamp` does not make sense to add this in the base feature view... WDYT?

I think it's safer to add the `created_timestamp` to the `OnDemandFeatureView` for consistency, in the same way that it's defined for FeatureViews.

amommendes:
Hey @achals , there is an error on linting step because of this exception:
`feast/infra/gcp.py:25: error: Module 'google.cloud' has no attribute 'storage'`

Is there some problem with the version of this lib?

sdk/python/feast/base_feature_view.py
```diff
@@ -32,6 +33,7 @@ def __init__(self, name: str, features: List[Feature]):
         self._name = name
         self._features = features
         self._projection = FeatureViewProjection.from_definition(self)
+        self.created_timestamp: Optional[datetime] = None
```
achals, line 36: This will need be written into the proto object in the `to_proto` method, and read out if it's present in the `from_proto` method; otherwise we'll lose this information when the feature view is persisted in the registry object.

achals (changes requested):
Thanks @amommendes , I think we need one more change for this to work as expected

sdk/python/feast/base_feature_view.py
amommendes: Hey @achals, we already [read](https://github.com/feast-dev/feast/blob/2435777d87c19d4eee670a7ccf5e73e88e22ce9d/sdk/python/feast/feature_view.py#L379) from and [write](https://github.com/feast-dev/feast/blob/2435777d87c19d4eee670a7ccf5e73e88e22ce9d/sdk/python/feast/feature_view.py#L300) to proto, right? But I forget to add the read/write in the ODFV. Is this your point? BTW, WDYT think about "promoting" `created_timestamp` field to the `spec` instead inside `meta`

achals: Ah, you're right. Thanks for addressing this.

amommendes:
Hey @achals , some update on this PR? WDYT?

achals (approved):
/lgtm

Sorry, I missed this in my inbox. Looks good!

PR #1954, opened 2021-10-17
Add support for Dask during historical retrieval

### LINKED ISSUE #1953: Distribute Feast with Dask
I would like to add possibility to distribute Feast feature store with Dask library

I will extend Feast with the Dask library. Thanks to this the users will be able to distribute historical feature handling on the clusters (eg. VMs, k8s, yarn, local machine). This will help to handle huge datasets. 

Dask is the best solution because we stay in the python stack others like Spark will need additional setup and maintenance. 

**Additional context**
No

qooba:

Fixes #1953

```release-note
The PR will add additional method to the FileRetrievalJob  to_dask_df() which will return dask DataFrame instead od pandas DataFrame. Thanks to this the historical features retrieval can be distributed across the cluster. Additioanly I have added the use_dask argument to the materialization functions which will also distribute the materialization process across the cluster. 
```

### DIFF
--- a/sdk/python/feast/infra/offline_stores/file.py
+++ b/sdk/python/feast/infra/offline_stores/file.py
@@ -149,7 +149,7 @@ class FileOfflineStore(OfflineStore):
         )
 
         # Create lazy function that is only called from the RetrievalJob object
-        def evaluate_historical_retrieval(use_dask: bool = False):
+        def evaluate_historical_retrieval(use_dask=False):
 
             # Make sure all event timestamp fields are tz-aware. We default tz-naive fields to UTC
             entity_df[entity_df_event_timestamp_col] = entity_df[
@@ -407,7 +407,7 @@ class FileOfflineStore(OfflineStore):
         assert isinstance(data_source, FileSource)
 
         # Create lazy function that is only called from the RetrievalJob object
-        def evaluate_offline_job(use_dask: bool = False):
+        def evaluate_offline_job(use_dask=False):
 
             if use_dask:
                 storage_options = (
--- a/sdk/python/feast/infra/offline_stores/file.py
+++ b/sdk/python/feast/infra/offline_stores/file.py
@@ -275,15 +275,9 @@ class FileOfflineStore(OfflineStore):
                     feature_names.append(formatted_feature_name)
 
                     # Ensure that the source dataframe feature column includes the feature view name as a prefix
-                    if use_dask:
-                        df_to_join=df_to_join.rename(
-                            columns={feature: formatted_feature_name}
-                        )
-                    else:
-                        df_to_join.rename(
-                            columns={feature: formatted_feature_name}, inplace=True,
-                        )
-
+                    df_to_join=df_to_join.rename(
+                        columns={feature: formatted_feature_name}
+                    )
 
                 # Build a list of entity columns to join on (from the right table)
                 join_keys = []
@@ -306,10 +300,7 @@ class FileOfflineStore(OfflineStore):
                         created_timestamp_column
                     ]
 
-                if use_dask:
-                    df_to_join=df_to_join.sort_values(by=event_timestamp_column)
-                else:
-                    df_to_join.sort_values(by=right_entity_key_sort_columns, inplace=True)
+                df_to_join=df_to_join.sort_values(by=event_timestamp_column)
 
                 df_to_join.drop_duplicates(
                     right_entity_key_sort_columns,
@@ -395,21 +386,42 @@ class FileOfflineStore(OfflineStore):
         assert isinstance(data_source, FileSource)
 
         # Create lazy function that is only called from the RetrievalJob object
-        def evaluate_offline_job():
-            filesystem, path = FileSource.create_filesystem_and_path(
-                data_source.path, data_source.file_options.s3_endpoint_override
-            )
-            source_df = pd.read_parquet(path, filesystem=filesystem)
-            # Make sure all timestamp fields are tz-aware. We default tz-naive fields to UTC
-            source_df[event_timestamp_column] = source_df[event_timestamp_column].apply(
-                lambda x: x if x.tzinfo is not None else x.replace(tzinfo=pytz.utc)
-            )
-            if created_timestamp_column:
-                source_df[created_timestamp_column] = source_df[
-                    created_timestamp_column
-                ].apply(
+        def evaluate_offline_job(use_dask: bool = False):
+
+            if use_dask:
+                storage_options = {
+                    "client_kwargs": {
+                        "endpoint_url": data_source.file_options.s3_endpoint_override
+                    }
+                } if data_source.file_options.s3_endpoint_override else None
+
+                source_df = dd.read_parquet(path, storage_options=storage_options)
+
+                source_df[event_timestamp_column] = source_df[event_timestamp_column].apply(
+                    lambda x: x if x.tzinfo is not None else x.replace(tzinfo=pytz.utc), meta=(event_timestamp_column,'datetime64[ns, UTC]'
+                )
+                if created_timestamp_column:
+                    source_df[created_timestamp_column] = source_df[
+                        created_timestamp_column
+                    ].apply(
+                        lambda x: x if x.tzinfo is not None else x.replace(tzinfo=pytz.utc), meta=(event_timestamp_column,'datetime64[ns, UTC]'
+                    )
+
+            else:
+                filesystem, path = FileSource.create_filesystem_and_path(
+                    data_source.path, data_source.file_options.s3_endpoint_override
+                )
+                source_df = pd.read_parquet(path, filesystem=filesystem)
+                # Make sure all timestamp fields are tz-aware. We default tz-naive fields to UTC
+                source_df[event_timestamp_column] = source_df[event_timestamp_column].apply(
                     lambda x: x if x.tzinfo is not None else x.replace(tzinfo=pytz.utc)
                 )
+                if created_timestamp_column:
+                    source_df[created_timestamp_column] = source_df[
+                        created_timestamp_column
+                    ].apply(
+                        lambda x: x if x.tzinfo is not None else x.replace(tzinfo=pytz.utc)
+                    )
 
             source_columns = set(source_df.columns)
             if not set(join_key_columns).issubset(source_columns):
@@ -423,7 +435,8 @@ class FileOfflineStore(OfflineStore):
                 else [event_timestamp_column]
             )
 
-            source_df.sort_values(by=ts_columns, inplace=True)
+
+            source_df=source_df.sort_values(by=ts_columns)
 
             filtered_df = source_df[
                 (source_df[event_timestamp_column] >= start_date)
--- a/sdk/python/feast/infra/offline_stores/file.py
+++ b/sdk/python/feast/infra/offline_stores/file.py
@@ -1,8 +1,8 @@
 from datetime import datetime
 from typing import Callable, List, Optional, Tuple, Union
 
-import pandas as pd
 import dask.dataframe as dd
+import pandas as pd
 import pyarrow
 import pytz
 from pydantic.typing import Literal
@@ -22,8 +22,8 @@ from feast.infra.offline_stores.offline_utils import (
 )
 from feast.infra.provider import (
     _get_requested_feature_views_to_features_dict,
+    _run_dask_field_mapping,
     _run_field_mapping,
-    _run_dask_field_mapping
 )
 from feast.registry import Registry
 from feast.repo_config import FeastConfigBaseModel, RepoConfig
@@ -115,7 +115,9 @@ class FileOfflineStore(OfflineStore):
         project: str,
         full_feature_names: bool = False,
     ) -> RetrievalJob:
-        if not isinstance(entity_df, pd.DataFrame) and not isinstance(entity_df, dd.DataFrame):
+        if not isinstance(entity_df, pd.DataFrame) and not isinstance(
+            entity_df, dd.DataFrame
+        ):
             raise ValueError(
                 f"Please provide an entity_df of type {type(pd.DataFrame)} instead of type {type(entity_df)}"
             )
@@ -184,25 +186,33 @@ class FileOfflineStore(OfflineStore):
 
                 if use_dask:
                     merge_asof = dd.merge_asof
-                    storage_options = {
-                        "client_kwargs": {
-                            "endpoint_url": feature_view.batch_source.file_options.s3_endpoint_override
+                    storage_options = (
+                        {
+                            "client_kwargs": {
+                                "endpoint_url": feature_view.batch_source.file_options.s3_endpoint_override
+                            }
                         }
-                    } if feature_view.batch_source.file_options.s3_endpoint_override else None
+                        if feature_view.batch_source.file_options.s3_endpoint_override
+                        else None
+                    )
 
-                    df_to_join = dd.read_parquet(feature_view.batch_source.path, storage_options=storage_options)
+                    df_to_join = dd.read_parquet(
+                        feature_view.batch_source.path, storage_options=storage_options
+                    )
 
-                    # Rename entity columns by the join_key_map dictionary if it exists
-                    if feature_view.projection.join_key_map:
+                    if feature_view.batch_source.field_mapping is not None:
                         _run_dask_field_mapping(
-                            df_to_join, feature_view.projection.join_key_map
+                            df_to_join, feature_view.batch_source.field_mapping
                         )
 
                     # Make sure all timestamp fields are tz-aware. We default tz-naive fields to UTC
                     df_to_join[event_timestamp_column] = df_to_join[
                         event_timestamp_column
                     ].apply(
-                        lambda x: x if x.tzinfo is not None else x.replace(tzinfo=pytz.utc), meta=(event_timestamp_column,'datetime64[ns, UTC]')
+                        lambda x: x
+                        if x.tzinfo is not None
+                        else x.replace(tzinfo=pytz.utc),
+                        meta=(event_timestamp_column, "datetime64[ns, UTC]"),
                     )
                     if created_timestamp_column:
                         df_to_join[created_timestamp_column] = df_to_join[
@@ -211,8 +221,7 @@ class FileOfflineStore(OfflineStore):
                             lambda x: x
                             if x.tzinfo is not None
                             else x.replace(tzinfo=pytz.utc),
-                            meta=(event_timestamp_column,'datetime64[ns, UTC]')
-
+                            meta=(event_timestamp_column, "datetime64[ns, UTC]"),
                         )
 
                 else:
@@ -238,13 +247,13 @@ class FileOfflineStore(OfflineStore):
                     # https://pandas.pydata.org/pandas-docs/stable/user_guide/missing_data.html#values-considered-missing
                     df_to_join = table.to_pandas()
 
-
-
                     # Make sure all timestamp fields are tz-aware. We default tz-naive fields to UTC
                     df_to_join[event_timestamp_column] = df_to_join[
                         event_timestamp_column
                     ].apply(
-                        lambda x: x if x.tzinfo is not None else x.replace(tzinfo=pytz.utc)
+                        lambda x: x
+                        if x.tzinfo is not None
+                        else x.replace(tzinfo=pytz.utc)
                     )
                     if created_timestamp_column:
                         df_to_join[created_timestamp_column] = df_to_join[
@@ -274,7 +283,7 @@ class FileOfflineStore(OfflineStore):
                     feature_names.append(formatted_feature_name)
 
                     # Ensure that the source dataframe feature column includes the feature view name as a prefix
-                    df_to_join=df_to_join.rename(
+                    df_to_join = df_to_join.rename(
                         columns={feature: formatted_feature_name}
                     )
 
@@ -299,7 +308,7 @@ class FileOfflineStore(OfflineStore):
                         created_timestamp_column
                     ]
 
-                df_to_join=df_to_join.sort_values(by=event_timestamp_column)
+                df_to_join = df_to_join.sort_values(by=event_timestamp_column)
 
                 df_to_join.drop_duplicates(
                     right_entity_key_sort_columns,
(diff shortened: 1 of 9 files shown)


sdk/python/feast/infra/offline_stores/file.py
```diff
@@ -1,6 +1,7 @@
 from datetime import datetime
 from typing import Callable, List, Optional, Union
 
+import dask.dataframe as dd
```
woop, line 4: If Dask is now always a requirement, why is the dependency added into extras?

sdk/python/feast/infra/offline_stores/file.py
```diff
@@ -305,26 +404,29 @@ def evaluate_offline_job():
                 else [event_timestamp_column]
             )
 
-            source_df.sort_values(by=ts_columns, inplace=True)
+            # source_df = source_df.sort_values(by=ts_columns[0])
```
woop, line 407: what is happening here?

woop:
@qooba this PR makes sense directionally to me, but I would want to try this out locally.

sdk/python/feast/infra/offline_stores/file.py
qooba: Fixed - added to the main requirements.

sdk/python/feast/infra/offline_stores/file.py
qooba: I've commented this during profiling (I wanted to check how sorting affects the performance) and forgotten to uncomment.  Fixed - this sort is important because in the next step we ```drop_duplicates``` with ```keep="last"```. The dask limitation here is that we can sort only by one column in this case we have to sort by ```event_timestamp_column``` rather than ```[event_timestamp_column, created_timestamp_column]``` which is still correct.

qooba:
> @qooba this PR makes sense directionally to me, but I would want to try this out locally.
> 
> One more thing: Is there a reason we would not use Dask for most of our RetrievalJobs? For example if we export from BigQuery to GCS (Parquet) and then create a Dask dataframe from the resulting file(s)

@woop  - very good idea we can fetch historical features using db engine and then export the result to gcs (for BigQuery - https://cloud.google.com/bigquery/docs/exporting-data#bigquery_extract_table_compressed-python) or s3 (for Redshift - https://docs.aws.amazon.com/redshift/latest/dg/r_UNLOAD.html).  Thanks to this we will be able to use ```FileSource```, ```BigQuerySource``` and ```RedshiftSource``` simultaneously.

sdk/python/setup.py
```diff
@@ -63,6 +63,7 @@
     "tqdm==4.*",
     "fastapi>=0.68.0",
     "uvicorn[standard]>=0.14.0",
+    "dask==2021.9.1",
```
woop, line 66: I think hardcoding this dependency is too restrictive. Is there any reason to pin it?

judahrand:
> @qooba this PR makes sense directionally to me, but I would want to try this out locally.
> 
> One more thing: Is there a reason we would not use Dask for most of our RetrievalJobs? For example if we export from BigQuery to GCS (Parquet) and then create a Dask dataframe from the resulting file(s)

Is a Pandas-wrapping library something Feast wants to rely on permanently?

If not, I don't think that relying on Dask is a really good idea. It's a useful tool for analytics and data exploration but as the internals of Feast are not (by definition) user facing I don't see why we should limit ourselves with a tool which has all the same flaws as Pandas. Type consistency (and potentially complex logical data types) are really import for a Feature Store in my opinion - Dask does not help with any of this. It is just a more scalable version of Pandas... 

If the aim is to remove Pandas from the internals of Feast (which, I believe it should be) then why would we put Dask in place as an alternative?

https://wesmckinney.com/blog/apache-arrow-pandas-internals/

qooba:
> > @qooba this PR makes sense directionally to me, but I would want to try this out locally.
> > One more thing: Is there a reason we would not use Dask for most of our RetrievalJobs? For example if we export from BigQuery to GCS (Parquet) and then create a Dask dataframe from the resulting file(s)
> 
> Is a Pandas-wrapping library something Feast wants to rely on permanently?
> 
> If not, I don't think that relying on Dask is a really good idea. It's a useful tool for analytics and data exploration but as the internals of Feast are not (by definition) user facing I don't see why we should limit ourselves with a tool which has all the same flaws as Pandas. Type consistency (and potentially complex logical data types) are really import for a Feature Store in my opinion - Dask does not help with any of this. It is just a more scalable version of Pandas...
> 
> If the aim is to remove Pandas from the internals of Feast (which, I believe it should be) then why would we put Dask in place as an alternative?
> 
> https://wesmckinney.com/blog/apache-arrow-pandas-internals/

@judahrand  thanks for your comment. 
I don't see anything bad in using Pandas. 
First of all it is widely used and well known by the users in DataScience, ML world.
It is natively compatible with the most ML libraries.
Secondly ```Feast``` use ```SOLID``` and the historical fetch logic is abstracted with 
```RetrievalJob``` thus ```Feast``` won't be locked with ```Dask``` forever. 

The main aim of this PR is not removing Pandas. The ```Dask``` solves the Feast biggest limitation, handling big datasets from multiple separated parquets (which is must if we want to integrate with data lakes like DeltaLake, Iceberg or Hudi) and distributed training. 

Of course historical fetch is mostly based on pandas ```merge_asof``` method but do we really want 
to reinvent the wheel ?

Regarding complex types. The most ML libraries use flat data if you will introduce complex types this will require additional postprocessing on the user side also during inference which will affect the performance. In my opinion this should be done at the ETL level which is note ```Feast``` responsibility.

Currently I'm working on the benchmarks: ```Dask``` vs ```Modin on Ray``` vs ```Modin on Dask``` vs ```Dask on Ray``` vs ```Pandas``` to double check what will be the best for the users.

File: sdk/python/feast/infra/offline_stores/file.py [hunk 4]
```diff
@@ -215,44 +291,26 @@ def evaluate_historical_retrieval():
                         created_timestamp_column
                     ]
 
-                df_to_join.sort_values(by=right_entity_key_sort_columns, inplace=True)
-                df_to_join.drop_duplicates(
-                    right_entity_key_sort_columns,
-                    keep="last",
-                    ignore_index=True,
-                    inplace=True,
+                df_to_join = df_to_join.sort_values(by=event_timestamp_column)
```
woop, line 294: Wouldn't it be more consistent to run multiple sort operations one after the other? That would provide deterministic results.

woop:
> Is a Pandas-wrapping library something Feast wants to rely on permanently?
> I don't see why we should limit ourselves with a tool which has all the same flaws as Pandas. 

I'd be happy to explore alternatives. What do you see as a better option @judahrand? Remember we are trying to optimize for ease of use and local (notebook) driven development with File Sources. That already eliminates Spark/PySpark or any system that requires a cluster to be available, although Spark can be used as a non-default extension.

We are already telling users that they should not use FileSources for production use cases. A bigger question is whether we should support FileSources at all. They fundamentally require ETL based workflows (Dask, Spark, Pandas) where data is retrieved to some kind of computational environment. FileSources act very differently from all of our other offline stores and source combinations where we promote ELT flows and pushing down SQL queries into a DWH.

My opinion: I do think there is value in supporting FileSources purely from a local development standpoint. I don't think the engine has to be Dask/Pandas though, but I don't see tremendous harm in this PR. It seems like a net positive over Pandas, as long as the biggest downside is an additional dependency. In fact, Dask acts more like a DWH since it does support some form of SQL querying. That allows us to support SQL based batch transformations in the future (even though the point-in-time joins will still be written in a dataframe form).

> If the aim is to remove Pandas from the internals of Feast (which, I believe it should be) then why would we put Dask in place as an alternative?

I agree that Pandas and Dask should never be the only option that users have, and this PR does not force them to use it. Nothing prevents users from using Redshift or BigQuery or Snowflake and exporting dataframes to a bucket.

Would it make sense to discuss this PR at the next community call on 4 January?

woop:
@qooba I am getting the following exception when running your scripts as above

```
  File "test.py", line 82, in <module>
    training_df.head(training_df)
  File "/Users/willem/.pyenv/versions/3.8.12/lib/python3.8/site-packages/pandas/core/generic.py", line 5073, in head
    return self.iloc[:n]
  File "/Users/willem/.pyenv/versions/3.8.12/lib/python3.8/site-packages/pandas/core/indexing.py", line 931, in __getitem__
    return self._getitem_axis(maybe_callable, axis=axis)
  File "/Users/willem/.pyenv/versions/3.8.12/lib/python3.8/site-packages/pandas/core/indexing.py", line 1543, in _getitem_axis
    return self._get_slice_axis(key, axis=axis)
  File "/Users/willem/.pyenv/versions/3.8.12/lib/python3.8/site-packages/pandas/core/indexing.py", line 1578, in _get_slice_axis
    labels._validate_positional_slice(slice_obj)
  File "/Users/willem/.pyenv/versions/3.8.12/lib/python3.8/site-packages/pandas/core/indexes/base.py", line 3684, in _validate_positional_slice
    self._validate_indexer("positional", key.stop, "iloc")
  File "/Users/willem/.pyenv/versions/3.8.12/lib/python3.8/site-packages/pandas/core/indexes/base.py", line 5719, in _validate_indexer
    raise self._invalid_indexer(form, key)
TypeError: cannot do positional indexing on Int64Index with these indexers [                   event_timestamp  entity_id        f0        f1        f2        f3        f4        f5        f6        f7        f8        f9    y
12993487 2021-01-14 08:00:00+00:00     168512  0.619108  1.389970 -0.684378 -0.429007 -1.926855 -0.472471 -0.778973 -0.606108 -0.992472  0.022998 -1.0
12993488 2021-01-14 08:00:00+00:00     510208  2.027257  1.198917 -0.441613  1.623630 -2.128640  1.895450  0.070850  0.025715 -0.796271 -0.841684  1.0
12993489 2021-01-14 08:00:00+00:00     565596 -2.457811 -0.168425 -0.119889  0.342324  0.563727  0.070273 -1.817797 -0.193417 -0.865418 -1.355372  1.0
12993490 2021-01-14 08:00:00+00:00     512145 -0.283760  0.539562 -0.714456 -0.293446 -0.223264  1.716292  0.145620 -0.145531 -0.916247 -0.514406 -1.0
12993491 2021-01-14 08:00:00+00:00     548780  0.055321 -1.000210  0.648203  0.124067  0.006806  1.080359 -0.098729 -0.582816 -1.650742  0.025469 -1.0
...                            ...        ...       ...       ...       ...       ...       ...       ...       ...       ...       ...       ...  ...
13992981 2021-01-15 05:59:59+00:00     852357  0.295029  1.435992  0.219495  1.046250 -0.276504 -0.553237 -0.393574  0.499950  0.194490 -0.545239 -1.0
13992982 2021-01-15 05:59:59+00:00     481662  1.149298 -1.197205 -0.669387 -0.158838 -0.070396 -0.823982  0.764135 -2.336068  0.412822 -0.612314  1.0
13992983 2021-01-15 05:59:59+00:00     780582 -0.420958 -1.643214  0.422955  1.603177  0.331257 -1.871032  0.064401  1.193627 -0.265694  0.558288  1.0
13992984 2021-01-15 05:59:59+00:00     852424 -0.679193  1.372682  0.198214 -1.139055  0.636165  0.350482 -1.514897 -0.741290 -0.744529  1.221901 -1.0
13992985 2021-01-15 05:59:59+00:00     395455 -0.220917 -0.345322  1.394109  0.199252  0.656851  0.511934  2.087830 -2.539696 -1.881959 -0.660819  1.0

[999499 rows x 13 columns]] of type DataFrame

```

woop:
@qooba I've tried to follow the steps above, but for some reason the execution time on my local machine is still 110-120 seconds. I'm running the code on Apple Mac M1 Pro. 

Did you make any other changes to your code (or Dask configuration)?

qooba:
> @qooba I've tried to follow the steps above, but for some reason the execution time on my local machine is still 110-120 seconds. I'm running the code on Apple Mac M1 Pro. Using just Pandas (on the master branch) results in similar performance (around 110 seconds).
> 
> Did you make any other changes to your code (or Dask configuration)?

@woop to ensure environment consistency I recommend using Docker.
I have compared current ```feast release``` vs ```feast with dask``` vs ```feast with dask on ray```.
In each case I will base on the bare ```continuumio/miniconda3:4.10.3```
for simplicity you can run ([benchmark.sh](https://github.com/qooba/feast-benchmark/blob/main/docker/benchmark.sh))
```
git clone git@github.com:qooba/feast-benchmark.git
cd feast-benchmark/docker 
./benchmark.sh
```

You will get result
```
TEST MASTER
            event_timestamp  entity_id       f0        f1        f2        f3        f4        f5        f6        f7        f8        f9    y
0 2021-01-14 23:59:42+00:00        100  0.33756 -0.087184 -0.739495 -0.855755  1.423298 -0.146513  1.122917 -0.102386  0.528685 -1.795665 -1.0
--- 101.37947177886963 seconds ---
TEST DASK
             event_timestamp  entity_id       f0        f1        f2        f3        f4        f5        f6        f7        f8        f9    y
13 2021-01-14 14:19:54+00:00        100  0.33756 -0.087184 -0.739495 -0.855755  1.423298 -0.146513  1.122917 -0.102386  0.528685 -1.795665 -1.0
--- 3.7066633701324463 seconds ---
TEST DASK ON RAY
2021-12-29 01:00:26,372 INFO services.py:1338 -- View the Ray dashboard at http://127.0.0.1:8265
             event_timestamp  entity_id       f0        f1        f2        f3        f4        f5        f6        f7        f8        f9    y
13 2021-01-14 14:19:54+00:00        100  0.33756 -0.087184 -0.739495 -0.855755  1.423298 -0.146513  1.122917 -0.102386  0.528685 -1.795665 -1.0
--- 5.045600175857544 seconds ---
```

The feast with ```dask``` and ```dask on ray``` performs better than bare ```feast```.
For ```feast with dask``` no other configuration is required (unless you want to distribute calculation over the cluster).
For ```feast with dask on ray``` you have to initialize ray cluter and configure dask:
```
import dask
from ray.util.dask import ray_dask_get
ray.init()
dask.config.set(scheduler=ray_dask_get)
```

Environments for:
## 1. feast release

[Dockerfile.no_dask_test](https://github.com/qooba/feast-benchmark/blob/main/docker/Dockerfile.no_dask_test)
```
FROM continuumio/miniconda3:4.10.3
RUN apt update && apt install curl git ssh cmake ack g++ vim tmux -yq
RUN pip install scikit-learn feast
WORKDIR /app
# PREPARE AND APPLY FEAST
COPY feature_store.yaml .
COPY features.py .
RUN feast apply
# GENERATE INPUT DATA
COPY generate.py .
# RUN TEST
COPY run.py .
CMD /bin/bash
```

## 2. feast with dask

[Dockerfile.dask_test](https://github.com/qooba/feast-benchmark/blob/main/docker/Dockerfile.dask_test)
```
FROM continuumio/miniconda3:4.10.3
RUN apt update && apt install curl git ssh cmake ack g++ vim tmux -yq
RUN git clone -b feature/dask https://github.com/qooba/feast.git
RUN cd feast && make install-python-ci-dependencies
RUN cd feast/sdk/python && pip install -e .
RUN pip install scikit-learn
WORKDIR /app
# PREPARE AND APPLY FEAST
COPY feature_store.yaml .
COPY features.py .
RUN feast apply
# GENERATE INPUT DATA
COPY generate.py .
# RUN TEST
COPY run.py .
CMD /bin/bash
``` 

## 3. feast with dask on ray
 
Here in [run_ray.py](https://github.com/qooba/feast-benchmark/blob/main/docker/run_ray.py) we have to start ray cluster and setup dask:
```
import dask
from ray.util.dask import ray_dask_get
ray.init()
dask.config.set(scheduler=ray_dask_get)
```

[Dockerfile.ray_test](https://github.com/qooba/feast-benchmark/blob/main/docker/Dockerfile.ray_test)
```
FROM continuumio/miniconda3:4.10.3
RUN apt update && apt install curl git ssh cmake ack g++ vim tmux -yq
RUN git clone -b feature/dask https://github.com/qooba/feast.git
RUN cd feast && make install-python-ci-dependencies
RUN cd feast/sdk/python && pip install -e .
RUN pip install ray[default,data]==1.9.1 scikit-learn==1.0.2 dask[distributed]==2021.11.0
WORKDIR /app
# PREPARE AND APPLY FEAST
COPY feature_store.yaml .
COPY features.py .
RUN feast apply
# GENERATE INPUT DATA
COPY generate.py .
# RUN TEST
COPY run_ray.py .
CMD /bin/bash
```


where 
[generate.py](https://github.com/qooba/feast-benchmark/blob/main/docker/generate.py)
[run.py](https://github.com/qooba/feast-benchmark/blob/main/docker/run.py)
[run_ray.py](https://github.com/qooba/feast-benchmark/blob/main/docker/run_ray.py)

woop:
Tested on [n2-standard-8](https://cloud.google.com/compute/docs/general-purpose-machines#n2_machines) on GCP (8 cores)

Master: 124.7 seconds
Dask: 8.5 seconds
Ray: 10.7 seconds

Performance looks really good. No custom configuration necessary even.

sdk/python/setup.py
```diff
@@ -66,6 +66,7 @@
     "uvicorn[standard]>=0.14.0",
     "proto-plus<1.19.7",
     "tensorflow-metadata>=1.0.0,<2.0.0",
+    "dask>=2021.9.*",
```
woop, line 69: Should we perhaps relax this dependency more? Aren't we excluding a lot of dask versions by only having 2021.9+?

sdk/python/feast/infra/offline_stores/file.py
woop: @qooba we currently use `created_timestamp_column` in order to deduplicate rows with duplicate event timestamps. We use the latest `created_timestamp_column` in that case. This is often useful in append only logs, since it allows a user to just reingest new rows with the same `event_timestamp`, but with fixed feature values. 

If we don't use created timestamp for sorting then we lose the guarantee of only returning the latest event_timestamp value by created_timestamp. Wouldn't it make more sense to first sort by `created_timestamp` and then `event_timestamp` in order to have determinism in the output and also consistency with our other offline store implementations?

sdk/python/feast/infra/offline_stores/file.py
qooba: fixed

sdk/python/setup.py
qooba: fixed

sdk/python/feast/infra/offline_stores/file.py
qooba: fixed

woop: Thanks, still some failures. Will wait for tests to pass then review :)

woop: Hmmm, those failures may be unrelated to you. We are looking into it.

qooba: I fixed some related to my changes but still there is something with datetimes, anyway I will check :)

danilopeixoto:
@qooba This feature implementation is amazing. Thanks!

Will the implementation preserve the `ttl` documentation?

```
The amount of time this group of features lives. A ttl of 0 indicates that this group of features lives forever. Note that large ttl's or a ttl of 0 can result in extremely computationally intensive queries.
```

What was the reason to change `merge_asof` to `merge`?

qooba:
> @qooba would you mind rebasing and rerunning tests? Also, can you please sign your commits?

@woop  - rebased, signed and finally all tests pass :)

qooba:
> @qooba This feature implementation is amazing. Thanks!
> 

@danilopeixoto thanks for your comment :)

> Will the implementation preserve the `ttl` documentation?
> 
> ```
> The amount of time this group of features lives. A ttl of 0 indicates that this group of features lives forever. Note that large ttl's or a ttl of 0 can result in extremely computationally intensive queries.
> ```

Yes I have fixed this. If ttl is 0 than the whole dataset is considered.

> 
> What was the reason to change `merge_asof` to `merge`?
> 

The most important reason is performance. The `merge_asof` would require time utc normalization and sorting on full dataset. The more efficient approach is to filter by `join_keys` eg. requested list of `driver_id`  and than normalize, filter by ttl tolerance and drop duplicates.

> Should `sort_values` calls fail when time columns are NaN or NaT?
> 
> ```
> '<' not supported between instances of 'NoneType' and 'NoneType' sort_values.
> ```

according to [dask sort_values](https://docs.dask.org/en/latest/generated/dask.dataframe.DataFrame.sort_values.html) NaN should be handled (there were some bugs which I have fixed :) )

danilopeixoto:
> > @qooba This feature implementation is amazing. Thanks!
> 
> @danilopeixoto thanks for your comment :)
> 
> > Will the implementation preserve the `ttl` documentation?
> > ```
> > The amount of time this group of features lives. A ttl of 0 indicates that this group of features lives forever. Note that large ttl's or a ttl of 0 can result in extremely computationally intensive queries.
> > ```
> 
> Yes I have fixed this. If ttl is 0 than the whole dataset is considered.
> 
> > What was the reason to change `merge_asof` to `merge`?
> 
> The most important reason is performance. The `merge_asof` would require time utc normalization and sorting on full dataset. The more efficient approach is to filter by `join_keys` eg. requested list of `driver_id` and than normalize, filter by ttl tolerance and drop duplicates.
> 
> > Should `sort_values` calls fail when time columns are NaN or NaT?
> > ```
> > '<' not supported between instances of 'NoneType' and 'NoneType' sort_values.
> > ```
> 
> according to [dask sort_values](https://docs.dask.org/en/latest/generated/dask.dataframe.DataFrame.sort_values.html) NaN should be handled (there were some bugs which I have fixed :) )

Thanks @qooba! I'm testing this feature branch. Let me know if I can help with anything.

How should we handle the `timedelta(0) != 0` edge case?

@woop and @qooba, I believe this is a subject for another PR, but let me know if it makes sense:

I was trying to optimize the queries using `date_partition_column` as partition filters in `read_parquet()` but could be any column in `partition_columns: List[str]`.

Optimizing using `event_timestamp` by default did not work because Pyarrow did convert date partition columns to category data type and I could not use `<=, >=` operators to compare `str` and `datetime` types.

I end up using the `date_partition_column` and `in` filter with another column derivated from `event_timestamp` like `year`.

In fact, this could be supported out of the box:

- Set the `partition_columns` in the data source (same as `partition_on` from `to_parquet`)
- Query by `partition_columns` like `event_timestamp`
- Apply `in` filter for each partition column using conjunction

qooba:
> Thanks @qooba! I'm testing this feature branch. Let me know if I can help with anything.
> 
> How should we handle the `timedelta(0) != 0` edge case?

Fixed :) thanks for finding.
> 
> @woop and @qooba, I believe this is a subject for another PR, but let me know if it makes sense:

You are absolutely right. Partitioning will be very important in terms of optimization. But partitioning is strongly associated with datalake design ([https://docs.dask.org/en/latest/generated/dask.dataframe.read_parquet.html](https://docs.dask.org/en/latest/generated/dask.dataframe.read_parquet.html)).

> 
> I was trying to optimize the queries using `date_partition_column` as partition filters in `read_parquet()` but could be any column in `partition_columns: List[str]`.
> 
> Optimizing using `event_timestamp` by default did not work because Pyarrow did convert date partition columns to category data type and I could not use `<=, >=` operators to compare `str` and `datetime` types.
> 
> I end up using the `date_partition_column` and `<=, >=` filters with another column derivated from `event_timestamp` like `year`.
> 
> In fact, this could be supported out of the box:
> 
> * Set the `partition_columns` in the data source (same as `partition_on` from `to_parquet`)
> * Query by `partition_columns` like `event_timestamp`
> * Apply `<=, >=` filters for each partition column using conjunction

woop:
Thanks @qooba. This is a super valuable addition to Feast!

PR #1961, opened 2021-10-22
eventtime check before writing features, use pipelines, ttl

vas28r13:

Several parts here mostly for discussion then can break into smaller PRs:

1. check the event timestamp before writing the features
the use case here comes into play when there are multiple ways to ingest into the OnlineStore (i.e. materialization, direct/streaming ingestion) so timing could be different in different scenarios so we are checking the even timestamp to make sure only the latest features are written

2. using redis pipelines to limit the amount of network calls 
if there are a lot of entities to lookup in Redis then the number of network calls to Redis may be the bottleneck. Tested this with >1000 entities and it's slow without using pipelines.

3. (**removed from this PR**)  support expiring records in the Redis -- probably should be a separate PR for this but putting out there for discussion
entities should be able to be expired otherwise they remain in the store and the feature store continuously grows. 
In many of our use cases an entity has some natural timeline for being relevant so this could be part of the TTL in the OnlineStore.

4. (**removed from this PR**) ability to lookup all entities
use case here is to be able to support ranking models where we are constantly reranking a lot of entities. Works well if entities can be expired as well.

### DIFF
--- a/sdk/python/tests/integration/e2e/test_universal_e2e.py
+++ b/sdk/python/tests/integration/e2e/test_universal_e2e.py
@@ -16,6 +16,7 @@ from tests.integration.feature_repos.universal.feature_views import driver_featu
 @pytest.mark.parametrize("infer_features", [True, False])
 def test_e2e_consistency(environment, e2e_data_sources, infer_features):
     fs = environment.feature_store
+    fs.config.project = fs.config.project + str(infer_features)
     df, data_source = e2e_data_sources
     fv = driver_feature_view(data_source=data_source, infer_features=infer_features)
 
--- a/sdk/python/tests/utils/online_read_write_test.py
+++ b/sdk/python/tests/utils/online_read_write_test.py
@@ -67,16 +67,14 @@ def basic_rw_test(
         event_ts=time_1, created_ts=time_1, write=(1.1, "3.1"), expect_read=(1.1, "3.1")
     )
 
-    # Note: This behavior has changed for performance. We should test that older
-    # value can't overwrite over a newer value once we add the respective flag
+    # We should test that older value can't overwrite over a newer value once we add the respective flag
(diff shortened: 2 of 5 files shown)


sdk/python/tests/utils/online_read_write_test.py
```diff
@@ -89,18 +90,20 @@ def _driver_rw_test(event_ts, created_ts, write, expect_read):
...
-        created_ts=time_3 + timedelta(hours=1),
-        write=(96864, "I HAVE A NEWER created_ts SO I WIN"),
-        expect_read=(96864, "I HAVE A NEWER created_ts SO I WIN"),
-    )
+    # _driver_rw_test(
+    #     event_ts=time_1 + timedelta(hours=1),
+    #     created_ts=time_3 + timedelta(hours=1),
+    #     write=(96864, "I HAVE A NEWER created_ts SO I WIN"),
+    #     expect_read=(96864, "I HAVE A NEWER created_ts SO I WIN"),
+    # )
```
vas28r13, line 109: I don't believe `created_ts` tie breaker was a feature in the recent feast versions
I think the only reason it passed was because it works in sequential write order not because of tie breaker logic

sdk/python/tests/integration/e2e/test_universal_e2e.py
```diff
@@ -16,6 +16,7 @@
 @pytest.mark.parametrize("infer_features", [True, False])
 def test_e2e_consistency(environment, e2e_data_sources, infer_features):
     fs = environment.feature_store
+    fs.config.project = fs.config.project + str(infer_features)
```
vas28r13, line 19: different bins for different tests
otherwise the old data remains in the integration online store

adchia (approved):
/lgtm

PR #1985, opened 2021-10-29
Improve aws lambda deployment (logging, idempotency, etc)

tsotnet:

**What this PR does / why we need it**: This PR fixes or improves a couple of things
1. Docker push command was failing for me without any errors in Python, which took quite some time to debug. With this change I make it possible to print out docker command outputs. This is logged on the debug level, so you'll need to run it with `--log-level=debug`, for example `feast --log-level=debug apply`. Also, looks like log-level flag was not working correctly, so I also fixed that.
2. We shouldn't try docker image download/upload if the image is already in ECR. So I made a change that first checks this image in ECR and only runs the docker logic if necessary. This also means that from the 2nd `feast apply`, users won't need to have docker daemon running (which was required on each apply before).

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/infra/aws.py
+++ b/sdk/python/feast/infra/aws.py
@@ -30,6 +30,7 @@ from feast.feature_table import FeatureTable
 from feast.feature_view import FeatureView
 from feast.flags import FLAG_AWS_LAMBDA_FEATURE_SERVER_NAME
 from feast.flags_helper import enable_aws_lambda_feature_server
+from feast.infra.feature_servers.aws_lambda.config import AwsLambdaFeatureServerConfig
 from feast.infra.passthrough_provider import PassthroughProvider
 from feast.infra.utils import aws_utils
 from feast.protos.feast.core.Registry_pb2 import Registry as RegistryProto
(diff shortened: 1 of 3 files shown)


sdk/python/feast/infra/aws.py
```diff
@@ -244,52 +255,72 @@ def _upload_docker_image(self, project: str) -> str:
...
-    def _get_version_for_aws():
-        """Returns Feast version with certain characters replaced.
+def _get_lambda_name(project: str):
+    lambda_prefix = AWS_LAMBDA_FEATURE_SERVER_REPOSITORY
+    lambda_suffix = f"{project}-{_get_version_for_aws()}"
+    # AWS Lambda name can't have the length greater than 64 bytes.
+    # This usually occurs during integration tests or when feast is
+    # installed in editable mode (pip install -e), where feast version is long
+    if len(lambda_prefix) + len(lambda_suffix) >= 63:
+        lambda_suffix = base64.b64encode(lambda_suffix.encode()).decode()[:40]
```
felixwang9817, line 314: is 40 enough here? looks like the prefix is 36 characters?

felixwang9817 (approved):
looks great overall, just one quick comment

sdk/python/feast/infra/aws.py
adchia: isn't the prefix feast-python-server which is 19 characters? so 19+40 = 59 < 63.

adchia (approved):
will leave Felix to give the lgtm, but the code lgtm

sdk/python/feast/infra/aws.py
felixwang9817: oops it is

PR #1990, opened 2021-11-01
Fix duplicate update infra

### LINKED ISSUE #1795: Duplicate update_infra calls in `feast apply`
## Expected Behavior 

Running `feast apply` only calls `update_infra()` on a provider once. 

## Current Behavior

Running `feast apply` calls `update_infra()` on a provider twice.

## Steps to reproduce

There are two places where update infra is being called
* https://github.com/feast-dev/feast/blob/745a1b43d20c0169b675b1f28039854205fb8180/sdk/python/feast/repo_operations.py#L229
* https://github.com/feast-dev/feast/blob/745a1b43d20c0169b675b1f28039854205fb8180/sdk/python/feast/repo_operations.py#L188

## Possible Solution

Remove one of the update_infra calls

felixwang9817:
**What this PR does / why we need it**: This PR removes the duplicate `update_infra` call in `feast apply`. It also cleans up `apply_total`, by ensuring that entities and ODFVs are also being deleted.

Fixes #1795 

```release-note
feast apply only calls update_infra once
```

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -378,6 +378,7 @@ class FeatureStore:
                 ]
             ],
         ],
+        update_infra: bool = True,
         commit: bool = True,
     ):
         """Register objects to metadata store and update related infrastructure.
@@ -389,6 +390,7 @@ class FeatureStore:
 
         Args:
             objects: A single object, or a list of objects that should be registered with the Feature Store.
+            update_infra: If True, update online store infra to reflect the newly applied objects.
             commit: whether to commit changes to the registry
 
         Raises:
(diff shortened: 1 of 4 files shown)


sdk/python/feast/feature_store.py
```diff
@@ -378,6 +378,7 @@ def apply(
                 ]
             ],
         ],
+        update_infra: bool = True,
```
achals, line 381: Do we need to add a new flag here? Can't we just ensure that it's `FeatureStore` that's calling `update_infra` and it's not called by `repo_operations` instead? I'd really prefer to not introduce branching behaviour here unless absolutely necessary.

achals (changes requested):

sdk/python/feast/feature_store.py
```diff
@@ -421,11 +438,26 @@ def apply(
...
+        tables_to_update = [ob for ob in objects if isinstance(ob, FeatureTable)]
+
+        if len(entities_to_update) + len(views_to_update) + len(
+            request_views_to_update
+        ) + len(odfvs_to_update) + len(services_to_update) + len(
+            tables_to_update
+        ) != len(
+            objects
+        ):
+            raise ValueError("Unknown object type provided as part of apply() call")
```
achals, line 458: It would be a nicer error message here if we could point to which object was the unknown one.

achals (approved):
/lgtm

PR #1994, opened 2021-11-04
Fix bug where using some Pandas dtypes in the output of an ODFV fails

judahrand:
* When the DataFrame returned by an ODFV UDF uses dtypes like `pd.Int64Dtype()` type mapping fails.

Fixes #

```release-note

```


judahrand:
Can I have an `ok-to-test` label? I can't run the tests locally so this is the only way I can progress.

judahrand:
This looks good to go to me.

adchia (approved):
/lgtm

PR #1997, opened 2021-11-04
Moving Feast Java back into main repo under java/ package

adchia:
This is almost a verbatim copy of feast java logic back into feast under java/. A couple of key changes:
- the protos in datatypes have new symlinks to the protos/ base package
- the "mvn" command (in Makefile and infra scripts) refers to a pom file in the java/ subdirectory
- renaming of github workflows to be java_[workflow name]
- removing of makefile definitions that reference non-existent scripts

In followup PRs, we intend to remove:
- References to Spark ingestion
- Feast Core (including authentication, credentials)
- Out of date documentation (e.g. architecture diagrams)
- Feast Serving references to Feast Core (e.g. CoreSpecService)
- Feast Serving connectors that are not supported from Python (e.g. BigTable, Cassandra). These can be ported over from Feast Java when we re-introduce those in python (as plugins)

We also will consolidate GitHub actions into master vs release vs PR actions (though we need to verify the actions work first).

As far as releases will go. We will switch back to having one consistent version for Feast Java + Feast Python components. This will require we deprecate the old maven + DockerHub package names (e.g. on https://hub.docker.com/u/feastdev and https://mvnrepository.com/artifact/dev.feast) for Feast Java (since Feast Java is on 0.25.2 whereas the main repo is on 0.15) and update tutorials to reflect this.

Fixes #

```release-note
Re-introducing Feast Java into main repo
```


.github/workflows/java_mirror.yml
```diff
@@ -0,0 +1,25 @@
+name: mirror
```
pyalex, line 1: I think this part is obsolete. It used to mirror to internal repo in Gojek, but that repo is on 0.9

java/common-test/pom.xml
```diff
@@ -0,0 +1,172 @@
...
+            <artifactId>kafka</artifactId>
+            <version>1.15.1</version>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <version>5.6.2</version>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.kafka</groupId>
```
pyalex, line 127: it seems like we bringing back too much

pyalex:
Are we gonna support 0.9 registry? I'm confused by the goal of this resurrection

java/common-test/pom.xml
pyalex: Ingestion part was supposed to be deleted from feast-java long time ago, but it's possible that not everything was deleted

woop:
> Are we gonna support 0.9 registry? I'm confused by the goal of this resurrection. IMO, big part of it should die for good

As an extension to this question and our conversation yesterday. Would it make sense to list out the exact functionality that will not be supported (and is slated for deletion) after this PR is merged?

java/common-test/pom.xml
adchia: As discussed, will in a followup PR remove obsolete sections

adchia:
@pyalex @woop PTAL. Updated the description of the PR with TODOs. We can consolidate workflows in a followup PR

pyalex (approved):

adchia:
@pyalex will fix some tests very soon

PR #2002, opened 2021-11-05
ODFV UDFs should handle list types

### LINKED ISSUE #1995: Registering ODFV UDFs that operate on lists of numbers (e.g., cosine similarity of embeddings/vectors) throws an error
## Expected Behavior
Registering ODFV UDFs that operate on lists of numbers (e.g., cosine similarity of embeddings/vectors) should not 
throw errors.
## Current Behavior
Defining a ODFV UDF such as cosine similarity and then running `feast apply` will result in the following error:
```
def feast_value_type_to_pandas_type(value_type: ValueType) -> Any:
         value_type_to_pandas_type: Dict[ValueType, str] = {
         ValueType.FLOAT: "float",
         ValueType.INT32: "int",
         ValueType.INT64: "int",
         ValueType.STRING: "str",
         ValueType.DOUBLE: "float",
         ValueType.BYTES: "bytes",
         ValueType.BOOL: "bool",
         ValueType.UNIX_TIMESTAMP: "datetime",
         }
         if value_type in value_type_to_pandas_type:
         return value_type_to_pandas_type[value_type]
         raise TypeError(
         >           f"Casting to pandas type for type {value_type} failed. "
         f"Type {value_type} not found"
         )
         E       TypeError: Casting to pandas type for type ValueType.DOUBLE_LIST failed. Type ValueType.DOUBLE_LIST not found
```
## Steps to reproduce
Define an ODFV UDF for cosine similarity and try to register it:
```python
from feast import Entity, Feature, FeatureView, ValueType
from feast.data_source import RequestDataSource
from feast.infra.offline_stores.file_source import FileSource
from feast.on_demand_feature_view import on_demand_feature_view
from google.protobuf.duration_pb2 import Duration

import numpy as np
import pandas as pd


item = Entity(
    name="item_id", 
    value_type=ValueType.INT64, 
    description="item ID",
)

items_fv = FeatureView(
    name="items",
    entities=["item"],
    features=[
        Feature(name="embedding", dtype=ValueType.DOUBLE_LIST),
    ],
    batch_source=FileSource(
        path="YOUR_PATH",
        event_timestamp_column="event_timestamp",
        created_timestamp_column="created",
    ),
    online=True,
    ttl=Duration(),
    tags={},
)

similarity_req = RequestDataSource(
    name="similarity_input", 
    schema={
        "vector": ValueType.DOUBLE_LIST,
    },
)

@on_demand_feature_view(
    inputs={
        "items": items_fv,
        "similarity_req": similarity_req,
    },
    features=[
        Feature(name="cos", dtype=ValueType.DOUBLE),
    ],
)
def similarity(features_df: pd.DataFrame) -> pd.DataFrame:
    if features_df.size == 0:
        return pd.DataFrame({"cos": [0.0]})  # give hint to Feast about return type
    vectors_a = features_df["embedding"].apply(np.array)
    vectors_b = features_df["vector"].apply(np.array)
    dot_products = vectors_a.mul(vectors_b).apply(sum)
    norms_q = vectors_a.apply(np.linalg.norm)
    norms_doc = vectors_b.apply(np.linalg.norm)
    df = pd.DataFrame()
    df["cos"] = dot_products / (norms_q * norms_doc)
    return df
```
### Specifications

- Version: 0.14.0
- Platform: all
- Subsystem: Python SDK

## Possible Solution
Add the following 2 lines to `feast_value_type_to_pandas_type()` in `type_map.py`:
```python
ValueType.FLOAT_LIST: "object",
ValueType.DOUBLE_LIST: "object",
```

Agent007:
ODFV UDFs handling list types (e.g., embeddings/vectors) should be registered without error.

Signed-off-by: Jeff <jeffxl@apple.com>

Fixes a bug in which ODFV UDFs handling list types (e.g., embeddings/vectors) weren't able to be registered.

Fixes #1995 

```release-note
Supports list types in on demand feature view udfs.
```

### DIFF
--- a/sdk/python/feast/type_map.py
+++ b/sdk/python/feast/type_map.py
@@ -453,6 +453,9 @@ def pa_to_redshift_value_type(pa_type: pyarrow.DataType) -> str:
         # PyArrow decimal types (e.g. "decimal(38,37)") luckily directly map to the Redshift type.
         return pa_type_as_str
 
+    if pa_type_as_str.startswith("list"):
+        return "super"
+
     # We have to take into account how arrow types map to parquet types as well.
     # For example, null type maps to int32 in parquet, so we have to use int4 in Redshift.
     # Other mappings have also been adjusted accordingly.
--- a/sdk/python/feast/type_map.py
+++ b/sdk/python/feast/type_map.py
@@ -89,11 +89,9 @@ def feast_value_type_to_pandas_type(value_type: ValueType) -> Any:
         ValueType.BYTES: "bytes",
         ValueType.BOOL: "bool",
         ValueType.UNIX_TIMESTAMP: "datetime",
-        ValueType.DOUBLE_LIST: "object",
-        ValueType.FLOAT_LIST: "object",
-        ValueType.INT32_LIST: "object",
-        ValueType.INT64_LIST: "object",
     }
+    if value_type.name.endswith("_LIST"):
+        return "object"
     if value_type in value_type_to_pandas_type:
         return value_type_to_pandas_type[value_type]
     raise TypeError(
--- a/sdk/python/feast/type_map.py
+++ b/sdk/python/feast/type_map.py
@@ -89,6 +89,10 @@ def feast_value_type_to_pandas_type(value_type: ValueType) -> Any:
         ValueType.BYTES: "bytes",
         ValueType.BOOL: "bool",
         ValueType.UNIX_TIMESTAMP: "datetime",
+        ValueType.DOUBLE_LIST: "object",
+        ValueType.FLOAT_LIST: "object",
+        ValueType.INT32_LIST: "object",
+        ValueType.INT64_LIST: "object",
     }
     if value_type in value_type_to_pandas_type:
         return value_type_to_pandas_type[value_type]
--- a/sdk/python/tests/integration/registration/test_universal_odfv_feature_inference.py
+++ b/sdk/python/tests/integration/registration/test_universal_odfv_feature_inference.py
@@ -7,6 +7,8 @@ from tests.integration.feature_repos.universal.feature_views import (
     conv_rate_plus_100_feature_view,
     create_conv_rate_request_data_source,
     create_driver_hourly_stats_feature_view,
+    create_similarity_request_data_source,
+    similarity_feature_view,
 )
 
 
@@ -27,10 +29,21 @@ def test_infer_odfv_features(environment, universal_data_sources, infer_features
         infer_features=infer_features,
     )
 
-    feast_objects = [driver_hourly_stats, driver_odfv, driver(), customer()]
+    sim_odfv = similarity_feature_view(
+        {
+            "driver": driver_hourly_stats,
+            "input_request": create_similarity_request_data_source(),
+        },
+        infer_features=infer_features,
+    )
+
+    feast_objects = [driver_hourly_stats, driver_odfv, sim_odfv, driver(), customer()]
     store.apply(feast_objects)
     odfv = store.get_on_demand_feature_view("conv_rate_plus_100")
     assert len(odfv.features) == 3
+
+    odfv = store.get_on_demand_feature_view("similarity")
+    assert len(odfv.features) == 1
 
 
 @pytest.mark.integration
--- a/sdk/python/tests/integration/registration/test_universal_odfv_feature_inference.py
+++ b/sdk/python/tests/integration/registration/test_universal_odfv_feature_inference.py
@@ -43,7 +43,7 @@ def test_infer_odfv_features(environment, universal_data_sources, infer_features
     assert len(odfv.features) == 3
 
     odfv = store.get_on_demand_feature_view("similarity")
-    assert len(odfv.features) == 1
+    assert len(odfv.features) == 2
 
 
 @pytest.mark.integration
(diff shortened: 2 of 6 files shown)


sdk/python/feast/type_map.py
```diff
@@ -89,6 +89,10 @@ def feast_value_type_to_pandas_type(value_type: ValueType) -> Any:
         ValueType.BYTES: "bytes",
         ValueType.BOOL: "bool",
         ValueType.UNIX_TIMESTAMP: "datetime",
+        ValueType.DOUBLE_LIST: "object",
+        ValueType.FLOAT_LIST: "object",
+        ValueType.INT32_LIST: "object",
+        ValueType.INT64_LIST: "object",
```
judahrand, line 95: Might it be easier to have a:
```python
# All list types are `object` dtypes in Pandas
if "_LIST" in value_type.name:
    return 'object'
```
since any list type will be an object in Pandas (I also wonder if Feast can move away from handling Pandas type conversions and let PyArrow do it? I haven't given that much thought though.)

judahrand: For example you've missed:
* `ValueType.BOOL_LIST`
* `ValueType.BYTES_LIST`
* `ValueType.UNIX_TIMESTAMP_LIST`
* `ValueType.STRING_LIST`

Agent007: @judahrand Thank you for the excellent suggestion. I've updated this PR and I hope you don't mind that I used `value_type.name.endswith("_LIST")` instead.

adchia (approved):
/lgtm

sdk/python/feast/type_map.py
judahrand: Even better!

Agent007:
@adchia , I think your integration test environment's RedShift instance may need to be updated to recognize the [`SUPER` data type](https://docs.aws.amazon.com/redshift/latest/dg/r_SUPER_type.html). The RedShift-based integration tests are failing because the example test data containing array types aren't able to be saved in RedShift...

Agent007:
@adchia In addition to the aforementioned RedShift-related errors, I now see errors related to `tracing_span()` after rebasing the latest from the main branch. I thought `FEAST_USAGE` is `False` in the integration test environment, so I don't see how `tracing_span()` would raise its `RuntimeError` if `_is_enabled` is supposed to be `False`.
Also, FYI, in order for `make test-python-universal-local` pass, I added checks in `test_feature_get_online_features_types_match()` to see if `feature` is not `None` before making the assertions on the expected feature types. Did you intend to handle `None` features differently?

judahrand:
> @adchia , I think your integration test environment's RedShift instance may need to be updated to recognize the [`SUPER` data type](https://docs.aws.amazon.com/redshift/latest/dg/r_SUPER_type.html). The RedShift-based integration tests are failing because the example test data containing array types aren't able to be saved in RedShift...

Might it be simpler to change the tests used for verifying this PR to avoid this and open a separate PR to add a test for SUPER columns? Just conscious that this ODFV list feature applies to ALL offline + online stores so not sure it makes sense to block on Redshift.

Might get both merged faster too.

adchia:
yep agreed with @judahrand . The main thing preventing the merge is now https://github.com/feast-dev/feast/pull/2007 going in. We also had a fix for the tracing span bug that already win.

Agent007:
@judahrand Apologies for the delay. I'm juggling other tasks at work currently. Please feel free to add onto this PR branch if this is urgent for you.

sdk/python/tests/integration/registration/test_universal_types.py
```diff
@@ -216,12 +216,14 @@ def test_feature_get_online_features_types_match(online_types_test_fixtures):
...
-            for element in feature:
-                assert isinstance(element, expected_dtype)
+            if feature is not None:
+                assert isinstance(feature, list)
+                for element in feature:
+                    assert isinstance(element, expected_dtype)
     else:
         for feature in online_features["value"]:
-            assert isinstance(feature, expected_dtype)
+            if feature is not None:
```
adchia, line 225: you should be able to revert this now.

sdk/python/tests/integration/registration/test_universal_odfv_feature_inference.py
```diff
@@ -27,14 +31,26 @@ def test_infer_odfv_features(environment, universal_data_sources, infer_features
         infer_features=infer_features,
     )
 
-    feast_objects = [driver_hourly_stats, driver_odfv, driver(), customer()]
+    sim_odfv = similarity_feature_view(
```
adchia, line 34: is there a way to make a separate test that specifically tests the similarity odfv? so we still get redshift coverage for other functionality?

adchia: same with other tests

Agent007: @adchia Good point. Let me isolate it...

sdk/python/tests/integration/registration/test_universal_types.py
Agent007: @adchia Thanks. I've reverted.

sdk/python/tests/integration/registration/test_universal_odfv_feature_inference.py
Agent007: @achals @judahrand The unit test has been simplified. Now, the Java integration test is failing, and I'm not sure why. I haven't touched the Java code.

PR #2003, opened 2021-11-05
Always pass full and partial feature names to ODFV

judahrand:

I've run into a problem where when an ODFV depends on two FV which have colliding feature names the ODFV will always want the full feature names. However, it only gets them when `full_feature_names` is set when requesting the features.

This PR ensures that both full and partial feature names are always passed to ODFVs.

Fixes #

```release-note
Always pass full and partial feature names to ODFV.
```


adchia (approved):
/lgtm

judahrand:
This is one approach to solve this issue - the approach which will impact any current users the least. The alternative is to ONLY pass features with full names to ODFVs. This would mean that user's would always have to use such names in the ODFV.

I'm open to thoughts on which is the better approach.

Issue #1999, opened 2021-11-04
Returned feature vector when on demand feature views are specified output dependent features


adchia:
When specifying on demand feature views at retrieval time (e.g. get_X_features), the output feature vectors include e.g. request data or dependent feature vectors, even if users did not specify said features.

## Expected Behavior 
Non-specified dependent feature values are not returned in output

## Current Behavior
Non-specified dependent feature values are in output

### Specifications

- Version:
- Platform:
- Subsystem:

felixwang9817:
Resolved by #2004

adchia:
This isn't resolved yet for get_historical_features

tmihalac:
Hello, I am new to Feast can you please explain to in detail what is needed to fix this for get_historical_features ?

This is what Tornike Gurgenidze answered in Slack :
"Hi, most of the code you might be interested in is in RetrievalJob class (https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/offline_stores/offline_store.py#L61), look for to_df and to_arrow methods. When individual feature store implementations create subclasses of RetrievalJob, they normally override _to_df_internal and _to_arrow_internal methods only. to_df then calls _to_df_internal , applies final logic to calculate values for odfvs and returns the final output.

Also, I might be totally wrong about this, but in case of get_historical_features I don't think the bug is that extra unspecified feature values are being returned, but rather that if the user doesn't specify all the features that requested odfv features depend on, RetrievalJob will be unable to calculate feature views and the call will most likely fail.
[offline_store.py](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/offline_stores/offline_store.py)
class RetrievalJob(ABC):"

I need further explanation about it and also how to reproduce the bug to be able to test the fix.

Issue #1998, opened 2021-11-04
At retrieval time, users need to specify dependencies for on demand feature views 


adchia:
## Expected Behavior 
Users only need to specify the on demand feature view to get the values

## Current Behavior
Users need to specify dependent feature views in the get_X_features APIs

felixwang9817:
Resolved by #2004

adchia:
Also not resolved for get_historical_features

PR #2004, opened 2021-11-05
get_online_features on demand transform bug fixes + local integration test mode

adchia:
For get_online_features, removing request data (that isn't specified by a request feature view) and odfv dependencies (that aren't otherwise requested) from the online response. It also adds in dependent feature views when specified in odfvs, so users no longer need to specify dependencies in the get_online_features request.

This also adds a rule (make test-python-universal-local) that will run local only integration tests for convenience for devs.

Related issues: #1999  and #1998 

Fixes #

```release-note
Removes need to specify on demand feature view dependencies in get_online_features
```

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -1079,13 +1079,13 @@ class FeatureStore:
         return needed_request_data, needed_request_fv_features
 
     def _augment_response_with_on_demand_transforms(
-            self,
-            feature_refs: List[str],
-            requested_result_row_names: Set[str],
-            odfvs: List[OnDemandFeatureView],
-            full_feature_names: bool,
-            initial_response: OnlineResponse,
-            result_rows: List[GetOnlineFeaturesResponse.FieldValues],
+        self,
+        feature_refs: List[str],
+        requested_result_row_names: Set[str],
+        odfvs: List[OnDemandFeatureView],
+        full_feature_names: bool,
+        initial_response: OnlineResponse,
+        result_rows: List[GetOnlineFeaturesResponse.FieldValues],
     ) -> OnlineResponse:
         all_on_demand_feature_views = {view.name: view for view in odfvs}
         all_odfv_feature_names = all_on_demand_feature_views.keys()
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -1078,71 +1078,6 @@ class FeatureStore:
                 needed_request_fv_features.add(feature.name)
         return needed_request_data, needed_request_fv_features
 
-    def _augment_response_with_on_demand_transforms(
-        self,
-        feature_refs: List[str],
-        requested_result_row_names: Set[str],
-        odfvs: List[OnDemandFeatureView],
-        full_feature_names: bool,
-        initial_response: OnlineResponse,
-        result_rows: List[GetOnlineFeaturesResponse.FieldValues],
-    ) -> OnlineResponse:
-        all_on_demand_feature_views = {view.name: view for view in odfvs}
-        all_odfv_feature_names = all_on_demand_feature_views.keys()
-
-        if len(all_on_demand_feature_views) == 0:
-            return initial_response
-        initial_response_df = initial_response.to_df()
-
-        odfv_feature_refs = defaultdict(list)
-        for feature_ref in feature_refs:
-            view_name, feature_name = feature_ref.split(":")
-            if view_name in all_odfv_feature_names:
-                odfv_feature_refs[view_name].append(feature_name)
-
-        # Apply on demand transformations
-        odfv_result_names = set()
-        for odfv_name, _feature_refs in odfv_feature_refs.items():
-            odfv = all_on_demand_feature_views[odfv_name]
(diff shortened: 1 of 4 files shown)


sdk/python/feast/feature_store.py
```diff
@@ -1002,17 +991,78 @@ def get_online_features(
...
+    def _get_requested_result_fields_and_populate_odfv_dependencies(
+        self,
+        entity_name_to_join_key_map: Dict[str, str],
+        full_feature_names: bool,
+        grouped_odfv_refs: List[Tuple[OnDemandFeatureView, List[str]]],
+        provider: Provider,
+        request_data_features: Dict[str, List[Any]],
+        needed_request_fv_features: Set[str],
+        result_rows,
+        union_of_entity_keys,
```
judahrand, line 1026: /nit I had to go and find out what these were from elsewhere in the code as I wasn't sure iterating over them would be fast - which suggests to me that having the annotations here might be useful (even if they exist elsewhere).

sdk/python/feast/feature_store.py
```diff
@@ -1002,17 +991,78 @@ def get_online_features(
...
             _feature_refs,
-            all_on_demand_feature_views,
+            requested_result_row_names,
+            requested_on_demand_feature_views,
             full_feature_names,
             initial_response,
             result_rows,
         )
 
+    def _get_requested_result_fields_and_populate_odfv_dependencies(
```
judahrand, line 1017: `and` in a method name tends to be a bit of a code smell to me. I might be misunderstanding but it looks like this could be split into two separate methods?

* `_get_requested_result_fields`
* `_populate_odfv_dependencies`

judahrand: /nit Separate comment as this one is my own personal gripe - I'm not a fan of side effects which seem to be the way you're updating the `result_rows` relying on the fact that a Python list is mutable. 'Tis fine but I typically find that it's harder to work out what's going on compared to populating and returning a new list. There could be an argument for this way though - memory maybe?

sdk/python/feast/feature_store.py
```diff
@@ -1002,17 +991,78 @@ def get_online_features(
...
+        requested_result_row_names.update(needed_request_fv_features)
+        # Add data if odfv requests specific feature views as dependencies
+        if len(grouped_odfv_refs) > 0:
+            for odfv, _ in grouped_odfv_refs:
+                for fv in odfv.input_feature_views.values():
+                    table_join_keys = [
+                        entity_name_to_join_key_map[entity_name]
+                        for entity_name in fv.entities
+                    ]
+                    self._populate_result_rows_from_feature_view(
```
judahrand, line 1055: I suppose this method is the precedent for populating as a side effect.

sdk/python/tests/integration/online_store/test_universal_online.py
```diff
@@ -274,11 +277,19 @@ def test_online_retrieval(environment, universal_data_sources, full_feature_name
     )
     assert online_features is not None
 
+    # Test if the dependent conv_rate feature isn't present for odfv
```
felixwang9817, line 280: not really clear to me how this test is working - is the idea that if `conv_rate` isn't present and the ODFV logic fails, the response will be `None`? I don't see how that actually happens

sdk/python/feast/feature_store.py
adchia: Done thanks!

sdk/python/feast/feature_store.py
```diff
@@ -1002,17 +991,78 @@ def get_online_features(
...
 
+    def _get_requested_result_fields_and_populate_odfv_dependencies(
+        self,
+        entity_name_to_join_key_map: Dict[str, str],
+        full_feature_names: bool,
+        grouped_odfv_refs: List[Tuple[OnDemandFeatureView, List[str]]],
+        provider: Provider,
+        request_data_features: Dict[str, List[Any]],
+        needed_request_fv_features: Set[str],
+        result_rows,
```
felixwang9817, line 1025: nit: add type annotations

sdk/python/feast/feature_store.py
```diff
@@ -1002,17 +991,78 @@ def get_online_features(
...
+        union_of_entity_keys,
+    ) -> Set[str]:
+        # Get requested feature values so we can drop odfv dependencies that aren't requested
+        requested_result_row_names: Set[str] = set()
+        for row_idx in range(len(result_rows)):
```
felixwang9817, line 1030: just do `for result_row in result_rows`?

sdk/python/feast/feature_store.py
adchia: Fixed the first bit. Yeah I'm also not a fan of populating as a side effect. 

Though in this case, its actually not relying on the python list being mutable and more relying on the underlying proto builder, so in this case does seem slightly canonical.

sdk/python/tests/integration/online_store/test_universal_online.py
adchia: clarified a bit, but its that if conv_rate isn't present, the logic will succeed anyways because we know to pull the data based on the dependent feature view

felixwang9817 (approved):
/lgtm

PR #2018, opened 2021-11-11
Fix checked out branch for PR docker image build workflow

tsotnet:

**What this PR does / why we need it**: Looks like by default docker image build & push step checks out master branch, which is not what we want. Instead, we want to check out the PR branch and build the docker image from that codebase.

Fixes #

```release-note
NONE
```


felixwang9817 (approved):
/lgtm

Issue #2035, opened 2021-11-13
tensorflow_metadata protos are outdated


casassg:
## Expected Behavior 

When installing feast sdk in an environment with Tensorflow Model Analysis, TFMA will stop working due to the fact that feast overwrites the files on the tensorflow_metadata python package.

This generates issues like this: https://github.com/BACtaki/tfx-addons/runs/4196630120?check_suite_focus=true

## Current Behavior

Feast overwrites tf_metadata files on pip install.

## Steps to reproduce

Just open feast wheel and inspect the RECORD file you get:

```
tensorflow_metadata/__init__.py,sha256=47DEQpj8HBSa-_TImW-5JCeuQeRkm5NMpJWZG3hSuFU,0
tensorflow_metadata/__pycache__/__init__.cpython-37.pyc,,
tensorflow_metadata/proto/__init__.py,sha256=47DEQpj8HBSa-_TImW-5JCeuQeRkm5NMpJWZG3hSuFU,0
tensorflow_metadata/proto/__pycache__/__init__.cpython-37.pyc,,
tensorflow_metadata/proto/v0/__init__.py,sha256=47DEQpj8HBSa-_TImW-5JCeuQeRkm5NMpJWZG3hSuFU,0
tensorflow_metadata/proto/v0/__pycache__/__init__.cpython-37.pyc,,
tensorflow_metadata/proto/v0/__pycache__/path_pb2.cpython-37.pyc,,
tensorflow_metadata/proto/v0/__pycache__/schema_pb2.cpython-37.pyc,,
tensorflow_metadata/proto/v0/__pycache__/statistics_pb2.cpython-37.pyc,,
tensorflow_metadata/proto/v0/path_pb2.py,sha256=8f1IIPnwSr0TUeaBbLrrA0SOvVtWZmJuekClUK0fohg,2341
tensorflow_metadata/proto/v0/path_pb2.pyi,sha256=x7plHZ0kvkv5qT8mZTJk8yqlQHdNdyAQM52GrHDdsU4,1341
tensorflow_metadata/proto/v0/schema_pb2.py,sha256=mnnrq9bVYIP5Uz9ztYoIDknhJAiJN2o8NuAonroLUUk,107886
tensorflow_metadata/proto/v0/schema_pb2.pyi,sha256=We0EW9WdmjfXRkDgx_l_UHW3WyFnSu03tSxRJY2R1xw,41992
[...]

PR #2038, opened 2021-11-15
Remove tensorflow-metadata folders

casassg:
Instead of sourcing tensorflow-metadata files in the sdk, we add a dependency on tensorflow-metadata library.

Currently feast python adds tensorflow_metadata namespace. This causes pip to overwrite the tensorflow-metadata library if installed after. 

So instead of doing this, we can just add a dependency on tensorflow-metadata.

This PR is a proposal to fix the issue detected in #2035

```release-note
Switch to use tensorflow-metadata requirement instead of sourcing compiled tensorflow-metadata in the sdk.
```


woop:
Thanks @casassg!

sdk/python/setup.py
```diff
@@ -65,6 +65,7 @@
     "fastapi>=0.68.0",
     "uvicorn[standard]>=0.14.0",
     "proto-plus<1.19.7",
+    "tensorflow-metadata>=1.0.0<2.0.0",
```
achals, line 68: I think you'll need a comma here between `>=1.0.0` and `<2.0.0`, right?

casassg: you are correct, ty for seeing this. Totally slipped me.

casassg:
(rebased for signoff)

achals (approved):
/lgtm

casassg:
(is there a schedule for when this may be released? Mostly want to use this for tfx-addons project :D )

achals:
@casassg @adchia can comment but I believe we're planning on cutting a new release in a the coming couple of weeks.

casassg:
Sg! Will wait for confirmation. Would be nice otherwise if we can backport this to 0.15 as a patch release given that this is mostly a "bug" fix (I understand it's not, but it should not modify behaviour as it effectively should be the same)

adchia:
There's likely some patch release we could sneak in relatively soon since we do have several issues we want to resolve.

PR #2017, opened 2021-11-11
Extend "feast in production" page  with description of java feature server

pyalex:
Signed-off-by: pyalex <moskalenko.alexey@gmail.com>

**What this PR does / why we need it**:

This PR extends "Running Feast in production" page in documentation.

**Which issue(s) this PR fixes**:

Fixes #

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #2043, opened 2021-11-16
Support of GC and S3 storages for registry in Java Feature Server

pyalex:

Previously only local file could be used as registry in Java Feature Server. This PR adds support for both Google Storage and AWS S3 as possible sources for registry file. Feature Server can only read registry file and not update it. It also may periodically check (when this enabled) if registry file was updated and load new version into memory.

Fixes #

```release-note
New application property `feast.registry-refresh-interval` when set to value bigger than `0` will enable periodical check whether registry file was updated
```


java/serving/src/main/java/feast/serving/registry/RegistryRepository.java
```diff
@@ -16,27 +16,74 @@
...
+public class RegistryRepository {
+  private Registry registry;
+  private RegistryFile registryFile;
+
+  public RegistryRepository(RegistryFile registryFile, int refreshIntervalSecs) {
+    this.registryFile = registryFile;
+    this.registry = new Registry(this.registryFile.getContent());
+
+    if (refreshIntervalSecs > 0) {
+      setupPeriodicalRefresh(refreshIntervalSecs);
```
achals, line 39: In the AWS Lambda Feature server, every `feast apply` is responsible for tearing down existong lambdas and recreating them with the new registry file. Should we follow the same pattern in feast-serving? Just seems weird that the behavior for the two servings options is different in this respect.

pyalex: Although, I agree on consistency point, this is a little different case. Java Feature Server will be deployed in Kubernetes (see my proposal for deploying architecture in #2017 ) with Helm chart (at least for now). I also would argue that restarting pod on each registry change might be not be desired in production environment (not sure whether it's good thing for lambda as well, since it also doesn't start instantly). 
But the main problem that currently feast sdk isn't capable of managing complex infrastructure, even aws lambda is currently deployed not completely (still requires manually configuration), and for this case we would need to manage kubernetes object. Probably that will be solved with introduction of pulumi.
But I still think that this current simple solution deserves to exist. It could be short-term solution until `feast apply` is capable of managing complex infra or long-term for people who will manage infra manually.

achals: Okay, yeah I think that's fair. For more "productionized" services, the ability to run without restarts is valuable.

woop: Is it worth making a TODO comment here with our reasoning? I just want to make sure that we capture our decision.

achals (approved):

PR #2046, opened 2021-11-16
Fix bug causing OnDemandFeatureView.infer_features() to fail when the…

ArrichM:
Problem:
When running OnDemandFeatureView.infer_features() and there are features registered with labels, an RegistryInferenceFailure is raised. 

Currently, this will work as-is:

``` python
import feast
import pandas as pd

@feast.on_demand_feature_view.on_demand_feature_view(
    inputs={},
    features=[
        feast.Feature(name="not_labeled", dtype=feast.ValueType.DOUBLE)
    ],
)
def not_labeled_view(feature_df: pd.DataFrame) -> pd.DataFrame:
    df = pd.DataFrame(dtype="float", columns=["not_labeled"])
    return df

not_labeled_view.infer_features()
```

While this won't:

``` python
import feast
import pandas as pd

@feast.on_demand_feature_view.on_demand_feature_view(
    inputs={},
    features=[
        feast.Feature(name="not_labeled", dtype=feast.ValueType.DOUBLE, labels={"test": "label"})
    ],
)
def labeled_view(feature_df: pd.DataFrame) -> pd.DataFrame:
    df = pd.DataFrame(dtype="float", columns=["labeled"])
    return df

labeled_view.infer_features()
```

Suggested Fix:
~To allow features with labels in the OnDemandFeatureView definition, remove labels from the registered features when comparing infered features with registered ones.~

Do not include labels when checking equality of Features.

```
None
```

### DIFF
--- a/sdk/python/feast/feature.py
+++ b/sdk/python/feast/feature.py
@@ -48,11 +48,7 @@ class Feature:
             self._labels = labels
 
     def __eq__(self, other):
-        if (
-            self.name != other.name
-            or self.dtype != other.dtype
-            or self.labels != other.labels
-        ):
+        if self.name != other.name or self.dtype != other.dtype:
             return False
         return True
 
--- a/sdk/python/feast/on_demand_feature_view.py
+++ b/sdk/python/feast/on_demand_feature_view.py
@@ -215,7 +215,8 @@ class OnDemandFeatureView(BaseFeatureView):
 
         if self.features:
             missing_features = []
-            for specified_features in self.features:
+            delabeled_features = [Feature(name=f.name, dtype=f.dtype) for f in self.features]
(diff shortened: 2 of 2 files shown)


achals:
Thanks for the PR @ArrichM! I think your approach would work, but I'm wondering if a better fix would be to update https://github.com/feast-dev/feast/blob/master/sdk/python/feast/feature.py#L50 and remove the line 
```
            or self.labels != other.labels
```

Since that would fix these kinds of comparisons more generally. What do you think?

ArrichM:
Yes, that was also what I came up with originally, but then I was hesitant to introduce this change globally.

achals (approved):
/lgtm

PR #2047, opened 2021-11-16
Don't allow FeatureStore.apply with commit=False

nossrannug:
The `FeatureStore.apply` method will call `update_infra` for the provider without
offering a dry run, so invoking it without committing should not be
an option. It would make all the infra updates but not update the
registry causing them to be out of sync.

Added `commit=False` to `apply_feature_service` and `apply_feature_table` in the
`apply` method.

When calling `feast apply` and adding a feature table or feature service
the registry will get updated even if the update_infra call fails.
This PR makes the only registry commit happen after the call to update_infra
has completed successfully.

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -391,14 +391,13 @@ class FeatureStore:
             ]
         ] = [],
         partial: bool = True,
-        commit: bool = True,
     ):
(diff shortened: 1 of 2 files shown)


achals:
Thanks for identifying the bug and the fix @nossrannug!

achals (approved):
/lgtm

PR #2048, opened 2021-11-17
Properly exclude entities from feature inference

mavysavydav:
```release-note
Fix to correctly exclude entity columns from being considered features when using feature inference during registration.
```

### DIFF
--- a/sdk/python/feast/inference.py
+++ b/sdk/python/feast/inference.py
@@ -1,11 +1,18 @@
 import re
 from typing import List
 
-from feast import BigQuerySource, Entity, FileSource, RedshiftSource
+import pandas as pd
+
+from feast import BigQuerySource, Entity, Feature, FileSource, RedshiftSource, errors
 from feast.data_source import DataSource
 from feast.errors import RegistryInferenceFailure
 from feast.feature_view import FeatureView
+from feast.on_demand_feature_view import OnDemandFeatureView
 from feast.repo_config import RepoConfig
+from feast.type_map import (
+    feast_value_type_to_pandas_type,
+    python_type_to_feast_value_type,
+)
 from feast.value_type import ValueType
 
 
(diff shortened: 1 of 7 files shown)


sdk/python/feast/inference.py
```diff
@@ -118,3 +125,113 @@ def update_data_sources_with_inferred_event_timestamp_col(
...
+    """
+    entity_name_to_join_key_map = {entity.name: entity.join_key for entity in entities}
+
+    for fv in fvs:
+        if not fv.features:
+            columns_to_exclude = {
+                fv.batch_source.event_timestamp_column,
+                fv.batch_source.created_timestamp_column,
+            } | {
+                entity_name_to_join_key_map[entity_name] for entity_name in fv.entities
```
mavysavydav, line 146: this is the line that changed from the previous version of this method. Grabbing the join_keys to exclude rather than entity name. Everything else is mainly copy and paste.

sdk/python/feast/inference.py
```diff
@@ -118,3 +125,113 @@ def update_data_sources_with_inferred_event_timestamp_col(
...
+
+def update_odfvs_with_inferred_features(odfvs: List[OnDemandFeatureView]) -> None:
+    """
+    Infers the set of features associated to this feature view from the input source.
+
+    Raises:
+        RegistryInferenceFailure: The set of features could not be inferred.
+    """
+    for odfv in odfvs:
+        df = pd.DataFrame()
```
mavysavydav, line 202: this method is basically the same, just moved to this file from on_demand_feature_views.py for consistency, and adjusted it's signature to be consistent with other inference methods

codyjlin (approved):
Thanks for this fix David!! Much needed.

sdk/python/feast/inference.py
woop: @mavysavydav is there a reason for the move to `inference.py`? I am not against it, but it makes reviewing the PR harder. Would it make sense to at least split these two changes into different commits to make it easier on the reviewer?

mavysavydav: yea i agree that review becomes more difficult. I had to move this method as part of this fix b/c it needs the list of entity objects which isn't available in feature_view.py. I can split out the odfv method move into a separate PR since that's more independent if you'd like

mavysavydav: have taken out the odfv code move

PR #2057, opened 2021-11-17
Install redis extra in AWS Lambda feature server & add hiredis depend…

tsotnet:
…ency

1. Re-introduce installing redis extra in AWS lambda-based feature server. This makes it possible to use public Redis endpoints with lambda (e.g. Redis Enterprise).
2. Install hiredis library, which is a C-based client and allows the default redis client to perform better (https://github.com/redis/redis-py/blob/1a41cfd95a53a95b078084d8627be6b6fba3bb71/redis/utils.py#L4).

Fixes #

```release-note
Allow using Redis online store in AWS Lambda feature server
```


adchia (approved):
fyi @DvirDukhan 

/lgtm

PR #2058, opened 2021-11-18
Remove unused tensorflow-metadata proto files 

achals:

The tensorflow-metadata package was added in #2038, but the corresponding proto files were left behind so the python feels keep getting regenerated.  

Also this PR adds detailed error messages for test_univerisal_e2e failures which I suspect are flaky.

```release-note
none
```


achals:
Turns out we have a hard dependency on the tensorflow protos in CoreService.proto, so we can't get rid of the protos themselves without removing the dependency in CoreService (or getting rid of that service entirely).

I'm going to trash this PR.

PR #2060, opened 2021-11-18
Asynchronously refresh registry in transformation service

pyalex:

Currently cached registry could expire by preconfigured TTL and will be lazy reloaded on next request, which might affect performance of transformation service.
This PR proposes to run background thread that will periodically refresh registry instead.

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -141,8 +141,10 @@ class FeatureStore:
         downloaded synchronously, which may increase latencies if the triggering method is get_online_features()
         """
         registry_config = self.config.get_registry_config()
-        self._registry = Registry(registry_config, repo_path=self.repo_path)
-        self._registry.refresh()
+        registry = Registry(registry_config, repo_path=self.repo_path)
+        registry.refresh()
+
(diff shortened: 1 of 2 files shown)


sdk/python/feast/feature_store.py
```diff
@@ -141,8 +141,10 @@ def refresh_registry(self):
         downloaded synchronously, which may increase latencies if the triggering method is get_online_features()
         """
         registry_config = self.config.get_registry_config()
-        self._registry = Registry(registry_config, repo_path=self.repo_path)
-        self._registry.refresh()
+        registry = Registry(registry_config, repo_path=self.repo_path)
+        registry.refresh()
+
+        self._registry = registry
```
achals, line 147: What's the reasoning for this refactoring?

achals (approved):

sdk/python/feast/feature_store.py
pyalex: if there're multiple threads accessing `self._registry` it makes sense to fully construct replacement (actual proto is being loaded in `registry.refresh`) before making it available for everybody else

PR #2065, opened 2021-11-18
Raise import error when repo configs module cannot be imported

felixwang9817:

**What this PR does / why we need it**: Previously when the full repo configs module could not be imported, the error was suppressed and the integration tests continued with the default repo configs. This makes it difficult for users who are trying to extend the test suite to realize the issue is an import error.

Fixes #

```release-note
NONE
```


achals (approved):
/lgtm

MattDelac (approved):

PR #2066, opened 2021-11-18
Allow using cached registry when writing to the online store

### LINKED ISSUE #2051: Make direct data ingestion into online store faster
We use the direct data ingestion capability so we can write inference data in a real-time flow and then use it immediately. We need to be able it ingest in milliseconds. Currently, the online ingestion takes a couple of seconds since it does a lookup to refresh the registry for every call to `write_to_online_store`. This blocks the new capability from being used in real-time. 

It would be great to have an `allow_cache` argument added to `write_to_online_store` so that we can do the write without needing to refresh the registry. 

Currently, I moved the registry to Postgres so the refresh is quicker, but it is still not quick enough

achals:

Fixes #2051 

```release-note
Use cached registry to speed up direct writes to the online store
```

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -891,7 +891,7 @@ class FeatureStore:
         self,
         feature_view_name: str,
         df: pd.DataFrame,
-        allow_registry_cache: bool = False,
+        allow_registry_cache: bool = True,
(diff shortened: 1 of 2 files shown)


felixwang9817 (approved):
/lgtm

Issue #2071, opened 2021-11-19
Use memory efficiently in feature materialization


judahrand:
**Is your feature request related to a problem? Please describe.**
Currently, the materialization process loads all the data from the Offline Store to an Arrow table, then converts all the data to Protobuf, then writes all the data to the Online Store. This process requires holding the entire dataset in memory which is not practical.

**Describe the solution you'd like**
Instead of returning an Arrow table yield a series of RecordBatches from the OfflineStore and process each batch individually. 

**Describe alternatives you've considered**

**Additional context**

adchia:
Consolidating in the linked issue

PR #2073, opened 2021-11-19
Optimize memory usage during materialization

judahrand:
I have run memory usage into issues on initial materialization of a large FeatureView with approaching 10 million rows and 30 columns. Acquiring the Arrow table succeeds but the conversion to Protobuf seems to require duplicating the whole dataset in memory which is not practical.

This initial improvement batches the table returned by `to_arrow` into RecordBatches which improves the situation. Looking forward I think that there should be the option (or requirement?) for RetrievalJobs to yield these RecordBatches directly to further reduce memory requirements.

This PR improves the situation from #2071 but I do not believe is enough to close it.

Fixes #

```release-note
None
```

### DIFF
--- a/sdk/python/feast/infra/passthrough_provider.py
+++ b/sdk/python/feast/infra/passthrough_provider.py
@@ -23,6 +23,8 @@ from feast.registry import Registry
 from feast.repo_config import RepoConfig
 from feast.usage import RatioSampler, log_exceptions_and_usage, set_usage_attribute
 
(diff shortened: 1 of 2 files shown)


adchia (approved):
/lgtm

judahrand:
Failures don't look related to PR

adchia:
FYI @tsotnet looks like an issue with feature server docker images

PR #2077, opened 2021-11-19
Fix feature server docker image tag generation in pr integration tests

tsotnet:

**What this PR does / why we need it**: Sometimes is can happen that `build-docker-image` and `integration-test-python` workflows run on different git commit hashes. This is because the HEAD commit hash is a merge commit of master and current PR HEAD. If the master gets updated between these 2 workflows, the HEAD commit hash changes, causing integration tests to try to use a wrong feature server docker image tag.

Instead, we'll be setting the commit hash as a workflow output in `build-docker-image`, then pass it to `integration-test-python` as an environment variable and use that in integration tests. This will guarantee that the same tag is being used.

Fixes #

```release-note
NONE
```


adchia (approved):
/lgtm

sdk/python/feast/constants.py
```diff
@@ -37,3 +37,6 @@
 
 # Default FTS port
 DEFAULT_FEATURE_TRANSFORMATION_SERVER_PORT = 6569
+
+# Environment variable for feature server docker image tag
+DOCKER_IMAGE_TAG_ENV_NAME: str = "FEAST_SERVER_DOCKER_IMAGE_TAG"
```
woop, line 42: Shouldn't `DOCKER_IMAGE_TAG_ENV_NAME` be something "feature server" specific?

tsotnet: Why would we use different tags for different servers? Also, we could always rename this if the necessity arises. It won't break anything.

PR #2078, opened 2021-11-20
Add --host as an option for feast serve

nossrannug:

Today the value is hardcoded to 127.0.0.1. This will enable users to
specify the host that they want to use for feast serve.

```release-note
NONE
```


achals (approved):
/lgtm

PR #2090, opened 2021-11-25
Do not attempt to compute ODFVs when there are no ODFVs

### LINKED ISSUE #2072: Inconsistent logic with `on_demand_feature_views`
## Expected Behavior

We should be consistent across the repo about the default value of `on_demand_feature_views`. We also need to make our conditional statement more resilient as `on_demand_feature_views` is sometimes `None` and sometimes `[]`

## Current Behavior
The problem is that we often check if `on_demand_feature_views` is None but we sometimes set it's default value to an empty list

This has a huge impact as the `RetrievalJob.to_arrow()` method will use `self._to_arrow_internal()` or `self._to_df_internal()` depending on the on_demand_feature_views.

It makes it complicated to debug part of the codebase.

### Specifications

- Version:
- Platform:
- Subsystem:

## Possible Solution
Make the conditional statements more resilient
```python
def had_on_demand_feature_views(on_demand_feature_views) -> bool:
    if on_demand_feature_views is None:
        return False
    if isinstance(on_demand_feature_views, list):
        if len(on_demand_feature_views) == 0:
            return False
    return True

if had_on_demand_feature_views(self.on_demand_feature_views):
    return features_df
```
instead of
```python
if self.on_demand_feature_views is None:
    return features_df
```

felixwang9817:

**What this PR does / why we need it**: This PR ensures that we do not attempt to compute ODFVs in `RetrievalJobs` when there are no ODFVs.

Fixes #2072

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -107,11 +107,7 @@ class BigQueryOfflineStore(OfflineStore):
 
         # When materializing a single feature view, we don't need full feature names. On demand transforms aren't materialized
         return BigQueryRetrievalJob(
-            query=query,
-            client=client,
-            config=config,
-            full_feature_names=False,
-            on_demand_feature_views=None,
+            query=query, client=client, config=config, full_feature_names=False,
         )
 
     @staticmethod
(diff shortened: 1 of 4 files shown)


sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -200,7 +196,7 @@ def __init__(
         client: bigquery.Client,
         config: RepoConfig,
         full_feature_names: bool,
-        on_demand_feature_views: Optional[List[OnDemandFeatureView]],
+        on_demand_feature_views: List[OnDemandFeatureView] = [],
```
achals, line 199: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

This should continue to be `Optional` but you can set it to `[]` in the function body if it's `None`.

achals (changes requested):

achals (approved):

PR #2093, opened 2021-11-26
Use pip-tools to lock versions of dependent packages

### LINKED ISSUE #1993: Managing dependencies using tools like pipenv or poetry
Currently dependencies are managed in `sdk/python/setup.py` and installation in CI is done by `pip install -e "sdk/python"` via Makefile targets.
But users who use pip-compile and pipenv can suffer installing feast package (e.g see #1912) because the dependency check is loose.

Using pipenv. Installing dependencies in CI will be done by `pipenv install -s sdk/python`, which generates `Pipfile.lock`. After the lock file is generated, re-installing dependencies can be done by `pipenv sync`.

- pros
  - Dependencies are checked and version conflicts can be found when we execute `pipenv install`.
  - Using pipenv can be a good small start compared to poetry.
- cons
  - Locking might take long (but only when dependencies are update).
  - Commands in Makefile or GitHub Actions yaml files must be updated, e.g. `pytest` -> `pipenv run pytest`

Using pip-compile might work. Just I'm not familiar with pip-compile.

Using poetry. Installing dependencies in CI will be done by `poetry install` using `pyproject.toml`, which generates `poetry.lock`. After the lock file is generated, re-installing dependencies can be done by `poetry install`, this time using the lock file.

- pros
  - Dependencies are checked and version conflicts can be found when we execute `poetry install`.
  - poetry can not only check dependencies but also build and publish wheel packages.
  - poetry is a tool which aims to eliminate pipenv's pain. See https://github.com/python-poetry/poetry#what-about-pipenv.
- cons
  - poetry manages all settings in `pyproject.toml`. Using poetry can be a big change compared to pipenv.
  - Commands in Makefile or GitHub Actions yaml must be updated, e.g. `pytest` -> `poetry run pytest`

**Additional context**

#1912

Which do maintainers think is better? After the decision is made, I can give it a shot and create a pull request.

ysk24ok:

This PR introduces [pip-tools](https://pip-tools.readthedocs.io/en/latest/) to lock the version of dependent packages.
Other options are [Poetry](https://python-poetry.org/) and [Pipenv](https://pipenv.pypa.io/en/latest/).
Using Poetry, we have to replace `setup.py` with `pyproject.toml`, which leads to comparatively large diff. That's why I didn't choose Poetry.
At first I tried using pipenv but decided not to use it based on maintainers' feedback. pipenv has some drawbacks such as slow locking and we don't need its virtualenv management feature right now. 

Several changes to note are the followings.

- Adding `py-3.X-ci-requirements.txt` files. The result of `pip-compile` does not change between Linux and MacOS so the OS specifier is not included in the file name.
- Using `py-3.X-ci-requirements.txt` as a cache key instead of `setup.py`.

Fixes #1993 

```release-note
NONE
```


adchia:
More of a general question / comment. I haven't messed around personally with `pipenv`. The biggest problem I see right now is with dependency resolution in Feast.

Seems like generally, people have very mixed feelings about pipenv and that poetry is the strongly preferred alternative. `pipenv` _seems_ like overkill for that (and using `pip-tools` / `pip-compile` might be more useful for this). `poetry` I could be convinced. Do you have stronger thoughts on this?

ysk24ok:
> Seems like generally, people have very mixed feelings about pipenv and that poetry is the strongly preferred alternative.

Yes, `poetry` is a better solution these days. Time for dependency lock is less and it can handle packaging by `poetry build` and `poetry push` commands.
The reason I didn't choose `poetry` is that I have to replace `setup.py` with `pyproject.toml`. It's too much, at least for now, because what we want to solve is dependency conflict.
I chose `pipenv` because I'm familiar with it and I thought it would take less time to create this PR.

> pipenv seems like overkill for that (and using pip-tools / pip-compile might be more useful for this). poetry I could be convinced.

I also think `pip-tools` is a simple enough solution for dependency conflict. Replacing `pipenv` with `pip-tools` won't take that long.
Which do you prefer, `pip-tools` or `poetry`? After you decide, I'll work on replacing `pipenv` with it.

achals:
@ysk24ok Is the idea with this diff that we'd `pipenv` while developing feast so that dev environments use a consistent version of dependent packages? Or would this affect how end users installed feast, when they run `pip install feast`?

adchia:
my personal preference likely is to use pip tools to start

ysk24ok:
@achals 
Both.
Tools like `pipenv` try to lock dependencies based on `install_requires` and `extras_require` in `setup.py`, and users also try to install Feast by running `pip install feast` based on them. 
If the lock fails, it means dependency version specification in setup.py is wrong and developers have to fix it. It leads no failure around dependencies for users in installation.

@adchia 
Understood. I'll replace pipenv with pip-tools.

adchia:
Asking achal to take a look since he's had experience with pip-tools!

achals:
@ysk24ok The PR looks pretty good! Can you update the description so that it matches the new approach?

ysk24ok:
@achals Thank you. I've updated the descriptioin.

.github/workflows/linter.yml
```diff
@@ -27,9 +29,11 @@ jobs:
...
             /opt/hostedtoolcache/Python
             /Users/runner/hostedtoolcache/Python
-          key: ${{ runner.os }}-${{ steps.setup-python.outputs.python-version }}-pip-${{ hashFiles('**/setup.py') }}
+          key: ${{ runner.os }}-${{ steps.setup-python.outputs.python-version }}-pip-${{ hashFiles(format('**/py{0}-ci-requirements.txt', env.PYTHON)) }}
           restore-keys: |
             ${{ runner.os }}-${{ steps.setup-python.outputs.python-version }}-pip-
+      - name: Install pip-tools
+        run: pip install pip-tools
       - name: Install dependencies
         run: make install-python-ci-dependencies
```
achals, line 38: Any reason `pip-tools` can't be installed in the same step as `make install-python-ci-dependencies`? Applies to all the workflows that have been changed.

ysk24ok: No special reason actually. Is it better to change the step like this?

```yaml
- name: Install dependencies
  run: |
    pip install pip-tools
    make install-python-ci-dependencies
```

achals: No, I think that our `CI_REQUIRED` dependencies already has an entry for `pip-tools`, so I wonder if we even need to install `pip-tools` separately? `make install-python-ci-dependencies` should use the `py3.*-ci-requirements.txt` which already has an entry for `pip-tools`, so i think it may not be necessary.

ysk24ok: I see you point, but `pip-tools` must be installed before `make install-python` and `make install-python-ci-dependencies` are executed because `piptools sync` is executed in those Makefile targets.
`pip-tools` entry in requirements is not required any more, so I think it should be deleted.

The other approach is to use `pip install -r` instead of `piptools sync` in those targets. In this approach, installing `pip-tools` step can be deleted but we can't get the benefit of `piptools sync`. `pip install -r` executes only installation, but `piptools sync` installs/upgrades/uninstalls everything necessary to match the requirements.txt contents. See https://github.com/jazzband/pip-tools#example-usage-for-pip-sync for more info on `piptools sync`.

achals: Okay, I see your point - I'm fine with the way things are in the current version of your diff in that case. 

Final question, do you think we should pin the version of `pip-tools` we install in the github workflows?

ysk24ok: IMO, it's not required. I think newer the better.

achals (approved):
/lgtm

Issue #928, opened 2020-08-04
Transitive dependencies are too specific in install_requires


abfreiheit:
Python SDK client library is not compatible with the most popular DS installations in GCP.
I was using `!pip install feast` together with standard Jupyter notebook installations for TensorFlow 1.15.0 and 2.2.0.

In both cases the installations end up with the following errors:
```
Collecting feast
  Downloading feast-0.6.2-py3-none-any.whl (116 kB)
     |████████████████████████████████| 116 kB 7.1 MB/s eta 0:00:01
Collecting google-cloud-bigquery==1.18.*
  Downloading google_cloud_bigquery-1.18.1-py2.py3-none-any.whl (146 kB)
     |████████████████████████████████| 146 kB 21.2 MB/s eta 0:00:01
Requirement already satisfied: googleapis-common-protos==1.* in /opt/conda/lib/python3.7/site-packages (from feast) (1.51.0)
Requirement already satisfied: Click==7.* in /opt/conda/lib/python3.7/site-packages (from feast) (7.1.2)
Collecting google
  Downloading google-3.0.0-py2.py3-none-any.whl (45 kB)
     |████████████████████████████████| 45 kB 4.3 MB/s  eta 0:00:01
Collecting google-api-core==1.20.*
  Downloading google_api_core-1.20.1-py2.py3-none-any.whl (90 kB)
     |████████████████████████████████| 90 kB 9.2 MB/s  eta 0:00:01
Collecting kafka-python==1.*
  Downloading kafka_python-1.4.7-py2.py3-none-any.whl (266 kB)
     |████████████████████████████████| 266 kB 16.6 MB/s eta 0:00:01
Requirement already satisfied: tabulate==0.8.* in /opt/conda/lib/python3.7/site-packages (from feast) (0.8.7)
Requirement already satisfied: google-auth<2.0dev,>=1.14.0 in /opt/conda/lib/python3.7/site-packages (from feast) (1.19.2)
Collecting google-cloud-storage==1.20.*
  Downloading google_cloud_storage-1.20.0-py2.py3-none-any.whl (70 kB)
     |████████████████████████████████| 70 kB 10.7 MB/s eta 0:00:01
Collecting google-cloud-bigquery-storage==0.7.*
  Downloading google_cloud_bigquery_storage-0.7.0-py2.py3-none-any.whl (55 kB)
     |████████████████████████████████| 55 kB 5.3 MB/s  eta 0:00:01
[...]

PR #2094, opened 2021-11-29
Broaden google-cloud-core dependency

ptoman-pa:
Broadens dependency on google-cloud-core. The limitation to 1.4.* seems tighter than it needs to be. The version 1.7.2 (which is the latest before 2.0.0) passes unit tests and makes feast more interoperable as a library with other systems.

Relates to https://github.com/feast-dev/feast/issues/928

```release-note
NONE
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,14 +25,16 @@ pre-commit install --hook-type pre-commit --hook-type pre-push
 ```
 3. On push, the pre-commit hook will run. This runs `make format` and `make lint`.
 
-### Signing commits
-Use git signing to sign your commits. See 
+### Signing off commits
+Use git signoffs to sign your commits. See 
 https://docs.github.com/en/github/authenticating-to-github/managing-commit-signature-verification for details
 
-Then, you can sign commits with the `-s` flag:
+Then, you can sign off commits with the `-s` flag:
 ```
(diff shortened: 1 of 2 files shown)


ptoman-pa:
@felixwang9817 Thank you for the polite reminder! Signed.

felixwang9817:
hey @ptoman-pa, I don't think the signing worked properly. if you look at your [commit](https://github.com/feast-dev/feast/pull/2094/commits/2c36e3b8cdad8d49b587712e98517d3360327fd2), you'll see it doesn't have the "signed off by" note that other [commits](https://github.com/feast-dev/feast/pull/2099/commits/c4736dc944401bec113a358c51cb9bef109e78c3) have - could you try signing again with `git commit -s`? thanks!

ptoman-pa:
Done. Also added a note about the project wanting signoffs rather than GPG, but we can remove that or break it into a separate PR depending on your preferences for repo management.

felixwang9817 (approved):
/lgtm

felixwang9817:
thanks for adding that note to our docs @ptoman-pa! it looks like the first of two commits is signed off properly, but the second isn't - would you mind fixing that? apologies for all the back and forth!

felixwang9817 (approved):
/lgtm

PR #2095, opened 2021-11-30
Fix integration test that is unstable due to incorrect materialization boundaries

pyalex:

test_universal_e2e::test_e2e_consistency is flaky since test records' timestamps and materialization boundaries (start and end times) are generated independently. So materialization is inconsistent, most of the time it writes 3 rows, but in some moments it could materialize the whole test dataframe (5 rows) in one go, which is not expected behavior.

Fixes #

```release-note
NONE
```


achals (approved):
/lgtm

PR #2100, opened 2021-12-02
Only apply FVs which have changed

judahrand:

Currently, when `feast apply` is called all FeatureViews are applied whether or not they already exist in an unchanged state. This seems to wipe the `materialization_intervals` attribute which necessitates a full materialization. I have at least one FV which takes on the order of 4hrs to sync which causes issues.

This PR identifies FVs which already exists and does not apply them again.

Fixes #

```release-note
Only apply FeatureViews which have changed when calling `feast apply`.
```

### DIFF
--- a/sdk/python/feast/repo_operations.py
+++ b/sdk/python/feast/repo_operations.py
@@ -150,20 +150,22 @@ def apply_total(repo_config: RepoConfig, repo_path: Path, skip_source_validation
             data_source.validate(store.config)
 
     # For each object in the registry, determine whether it should be kept or deleted.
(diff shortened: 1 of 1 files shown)


judahrand:
AWS failure in the integration tests.

pyalex:
Hi @judahrand, thanks for your PR. Can you please add a test with this use case that is currently failing for you?

judahrand:
> Hi @judahrand, thanks for your PR. Can you please add a test with this use case that is currently failing for you?

Not easily, no. If you'd like to add a framework for testing the CLI I'll happily add a test case.

judahrand:
Huh, on further inspection this is not an issue. I'll try to work out what caused the problem I ran into.

PR #2101, opened 2021-12-02
Don't materialize FeatureViews where `online is False`

judahrand:

Currently, even if `online=False` is set on a FeatureView `feast materialize` will try to materialize the FeatureView.

This PR prevents this.

Fixes #

```release-note
Correct behavior of non-online FeatureViews.
```

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -758,9 +758,16 @@ class FeatureStore:
             feature_views_to_materialize = self._list_feature_views(
                 hide_dummy_entity=False
             )
+            feature_views_to_materialize = [
+                fv for fv in feature_views_to_materialize if fv.online
+            ]
         else:
             for name in feature_views:
                 feature_view = self._get_feature_view(name, hide_dummy_entity=False)
+                if not feature_view.online:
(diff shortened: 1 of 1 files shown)


sdk/python/feast/feature_store.py
```diff
@@ -758,9 +758,16 @@ def materialize_incremental(
             feature_views_to_materialize = self._list_feature_views(
                 hide_dummy_entity=False
             )
+            feature_views_to_materialize = [
+                fv for fv in feature_views_to_materialize if fv.online is True
```
pyalex, line 762: Hi @judahrand, I think `is True` and `is False` could be dropped here, since `online` property is always `boolean` you can simply use `if fv.online:` or `if not fv.online`

pyalex (changes requested):

sdk/python/feast/feature_store.py
judahrand: Made that change.

achals:
Thanks for the contribution @judahrand ! Good catch.

achals (approved):
/lgtm

pyalex (approved):

PR #2103, opened 2021-12-03
Pre compute the timestamp range for feature views

### LINKED ISSUE #2083: Does BigQuery template query take advantage of Partitions?
I'm unclear whether due to the dynamic nature of these `WHERE` clauses this query will take advantage of table partitions on the event timestamp? 

This could be solved by predetermining these min and max values and templating them in directly.

https://cloud.google.com/bigquery/docs/querying-partitioned-tables#pruning_limiting_partitions

https://github.com/feast-dev/feast/blob/b5d8a531c81fa11297e5c684f898646cb2ee2cd1/sdk/python/feast/infra/offline_stores/bigquery.py#L489-L491

judahrand:
This allows BigQuery to take advantage of Partitions when querying
historical features.

Fixes #2083 (doesn't necessarily completely fix but does fix in at least one case)

```release-note
None
```

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -385,7 +385,7 @@ def _upload_entity_df_and_get_entity_schema(
 ) -> Dict[str, np.dtype]:
     """Uploads a Pandas entity dataframe into a BigQuery table and returns the resulting table"""
 
-    if type(entity_df) is str:
+    if isinstance(entity_df, str):
         job = client.query(f"CREATE TABLE {table_name} AS ({entity_df})")
         block_until_done(client, job)
 
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -1,7 +1,16 @@
 import contextlib
 import uuid
 from datetime import date, datetime, timedelta
-from typing import Callable, ContextManager, Dict, Iterator, List, Optional, Union
+from typing import (
+    Callable,
+    ContextManager,
+    Dict,
+    Iterator,
+    List,
+    Optional,
+    Tuple,
+    Union,
+)
 
 import numpy as np
 import pandas as pd
@@ -156,9 +165,17 @@ class BigQueryOfflineStore(OfflineStore):
                 entity_schema, expected_join_keys, entity_df_event_timestamp_col
             )
 
+            entity_df_event_timestamp_range = _get_entity_df_event_timestamp_range(
+                entity_df, entity_df_event_timestamp_col, client, table_reference,
+            )
+
             # Build a query context containing all information required to template the BigQuery SQL query
             query_context = offline_utils.get_feature_view_query_context(
-                feature_refs, feature_views, registry, project,
+                feature_refs,
+                feature_views,
+                registry,
+                project,
+                entity_df_event_timestamp_range,
             )
 
             # Generate the BigQuery SQL query from the query context
@@ -394,6 +411,39 @@ def _upload_entity_df_and_get_entity_schema(
     return entity_schema
 
 
+def _get_entity_df_event_timestamp_range(
+    entity_df: Union[pd.DataFrame, str],
+    entity_df_event_timestamp_col: str,
+    client: Client,
+    table_name: str,
+) -> Tuple[datetime, datetime]:
+    if type(entity_df) is str:
+        job = client.query(
+            f"SELECT MIN({entity_df_event_timestamp_col}) AS min, MAX({entity_df_event_timestamp_col}) AS max FROM {table_name}"
+        )
+        res = next(job.result())
+        entity_df_event_timestamp_range = (
+            res.get("min"),
+            res.get("max"),
+        )
+    elif isinstance(entity_df, pd.DataFrame):
+        entity_df_event_timestamp = entity_df.loc[
+            :, entity_df_event_timestamp_col
+        ].infer_objects()
+        if pd.api.types.is_string_dtype(entity_df_event_timestamp):
+            entity_df_event_timestamp = pd.to_datetime(
+                entity_df_event_timestamp, utc=True
+            )
+        entity_df_event_timestamp_range = (
+            entity_df_event_timestamp.min(),
+            entity_df_event_timestamp.max(),
+        )
+    else:
+        raise InvalidEntityType(type(entity_df))
+
+    return entity_df_event_timestamp_range
+
+
 def _get_bigquery_client(project: Optional[str] = None, location: Optional[str] = None):
     try:
         client = bigquery.Client(project=project, location=location)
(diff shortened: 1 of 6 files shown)


achals:
Thanks @judahrand this looks great! Have you been able to very performance gains on bigquery as a result?

tsotnet:
Thanks for the PR @judahrand! Really curious, how are you checking that this is helping the partitioning? One suggestion would be maybe we can select MIN & MAX of entity_dataframe in the same SQL query using `WITH` statement. Something like:
```
WITH entity_dataframe_minmax AS (
    SELECT
        MAX(entity_timestamp) as max_timestamp,
        MIN(entity_timestamp) as min_timestamp
    FROM entity_dataframe
)
```

Would that achieve the same goal? Asking because the existing approach is adding more complexity to the offline store logic, which is completely fine if it's the easiest way to solve the partitioning issue, but I want to make sure this is the simplest way.

judahrand:
> Thanks @judahrand this looks great! Have you been able to very performance gains on bigquery as a result?

I haven't verified any speed gain but that also (in my opinion) isn't the main reason for doing this. The main reason is cost given that BigQuery charges according to the amount of data processed not the query time.

> Thanks for the PR @judahrand! Really curious, how are you checking that this is helping the partitioning? One suggestion would be maybe we can select MIN & MAX of entity_dataframe in the same SQL query using `WITH` statement. Something like:

To verify that the old query was not taking advantage of partitioning I copied the `{{ featureview.name }}__subquery` CTE query into the BigQuery editor window and added a CTE which just defined the min and max timestamps. The editor reported that 200Mb of data would be processed. I then replaced the `SELECT MIN...` and `SELECT MAX...` subqueries with hard coded strings of the same value and the editor reported 0Bytes of processed data. This verified that with the sub query (even though it was MIN and MAX) the partitions were ignored but with a pre computed string value they were not.

judahrand:
@MattDelac I've run a few tests and still don't think we can achieve use of the partitions without pre computing the bounds. Here is what I did:
```
WITH entity_dataframe AS (
    SELECT CAST(entity_timestamp AS TIMESTAMP) AS entity_timestamp FROM UNNEST(['2022-01-01', '2021-12-31']) AS entity_timestamp
)
SELECT
    user_id
FROM `feature_table`
WHERE event_timestamp <= (SELECT MAX(entity_timestamp) FROM entity_dataframe)
AND event_timestamp >= Timestamp_sub((SELECT MIN(entity_timestamp) FROM entity_dataframe), interval 100 second)
```
Results in 200Mb processed.

```
WITH entity_dataframe AS (
    SELECT CAST(entity_timestamp AS TIMESTAMP) AS entity_timestamp FROM UNNEST(['2022-01-01', '2021-12-31']) AS entity_timestamp
),
entity_timestamp_range AS (
    SELECT 
        MAX(entity_timestamp) AS max,
        MIN(entity_timestamp) as min,
    FROM entity_dataframe
)
SELECT
    user_id
FROM `feature_table`
WHERE event_timestamp <= (SELECT max FROM entity_timestamp_range)
AND event_timestamp >= Timestamp_sub((SELECT min FROM entity_timestamp_range), interval 100 second)
```
Results in 200Mb processed.

```
DECLARE entity_max_timestamp TIMESTAMP DEFAULT '2022-01-01';
DECLARE entity_min_timestamp TIMESTAMP DEFAULT '2021-12-31';
SELECT
    user_id
FROM `feature_table`
WHERE event_timestamp <= entity_max_timestamp
AND event_timestamp >= Timestamp_sub(entity_min_timestamp, interval 100 second)
```
Results in 0B processed.

This suggests to me that pre computing is necessary.

pyalex:
Hey @judahrand I fixed the issue with AWS policy, however, redshift tests are still failing. I presume we filter out too much. Can you please take a look?

sdk/python/feast/infra/offline_stores/redshift.py
```diff
@@ -428,9 +492,9 @@ def _upload_entity_df_and_get_entity_schema(
             {{ feature }} as {% if full_feature_names %}{{ featureview.name }}__{{feature}}{% else %}{{ feature }}{% endif %}{% if loop.last %}{% else %}, {% endif %}
         {% endfor %}
     FROM {{ featureview.table_subquery }}
-    WHERE {{ featureview.event_timestamp_column }} <= (SELECT MAX(entity_timestamp) FROM entity_dataframe)
+    WHERE {{ featureview.event_timestamp_column }} >= '{{ featureview.max_event_timestamp }}'
```
pyalex, line 495: Should be `<=`

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -394,6 +411,39 @@ def _upload_entity_df_and_get_entity_schema(
     return entity_schema
 
 
+def _get_entity_df_event_timestamp_range(
+    entity_df: Union[pd.DataFrame, str],
+    entity_df_event_timestamp_col: str,
+    client: Client,
+    table_name: str,
+) -> Tuple[datetime, datetime]:
+    if type(entity_df) is str:
```
pyalex, line 420: Is there a reason not to use `isinstance`?

sdk/python/feast/infra/offline_stores/offline_utils.py
```diff
@@ -130,6 +133,13 @@ def get_feature_view_query_context(
         event_timestamp_column = feature_view.input.event_timestamp_column
         created_timestamp_column = feature_view.input.created_timestamp_column
 
+        min_event_timestamp = None
+        if feature_view.ttl:
+            min_event_timestamp = (
+                entity_df_timestamp_range[0] - feature_view.ttl
+            ).isoformat()
```
pyalex, line 140: Another reason for test failing is that here we have timezone-aware timestamps, which are not very well handled by Redshift when passed into query. That probably has some explanation, like we not using timezone-aware types in Redshift tables, but I think keeping all timestamps in UTC is a much better solution.
So let's convert here to UTC.

pyalex: Correction, it's not all the time timezone-aware, but it is in some cases.

sdk/python/feast/infra/offline_stores/bigquery.py
judahrand: Good catch! Will change

sdk/python/feast/infra/offline_stores/redshift.py
judahrand: Whoops!

sdk/python/feast/infra/offline_stores/offline_utils.py
judahrand: Changed this!

judahrand:
> look?

Will do.

sdk/python/feast/infra/offline_stores/offline_utils.py
```diff
@@ -130,6 +133,25 @@ def get_feature_view_query_context(
         event_timestamp_column = feature_view.input.event_timestamp_column
         created_timestamp_column = feature_view.input.created_timestamp_column
 
+        min_event_timestamp = None
+        if feature_view.ttl:
+            min_event_timestamp = entity_df_timestamp_range[0] - feature_view.ttl
+            if min_event_timestamp.tzinfo is not None:
```
pyalex, line 139: this probably can be moved into some helper function (actually `_to_naive_utc` already exists)

judahrand:
I believe this is functional but it is so hard to tell will how flakey the AWS tests are 😢 Even just what seems like random connection errors.

pyalex (approved):

judahrand:
@pyalex Could we get this merged once the tests run. I keep having to merge in conflicts from `master`. Not sure why the docker build is failing?

judahrand:
@pyalex all good here

achals (approved):
/lgtm

PR #2108, opened 2021-12-06
Update release workflow to include new docker images

adchia:
There are new feature transformation server, lambda feature server, and java feature server docker images. This adds convenience commands in the Makefile + merges/updates our release workflow to build + push these images to GCR/dockerhub.

This also removes a workflow that builds a new FTS image + tags against the commit. We don't use this (the helm charts don't point to these either).

Fixes #

```release-note
NONE
```


Makefile
```diff
@@ -129,34 +129,50 @@ lint-go:
...
+		-f sdk/python/feast/infra/transformation_servers/Dockerfile .
 
-push-serving-docker:
-	docker push $(REGISTRY)/feast-serving:$(VERSION)
-
-build-core-docker:
-	docker build --build-arg VERSION=$(VERSION) -t $(REGISTRY)/feast-core:$(VERSION) -f infra/docker/core/Dockerfile .
+push-feature-server-java-docker:
+	docker push $(REGISTRY)/feature-server-java:$(VERSION)
+    docker push gcr.io/kf-feast/feature-server-java:$(VERSION)
```
pyalex, line 168: A little while ago we migrated to docker hub as a default container registry? Are we moving back to gcr.io?

pyalex (approved):

.github/workflows/master_only.yml
```diff
@@ -134,3 +134,43 @@ jobs:
         run: pytest --verbose --color=yes sdk/python/tests --integration --benchmark --benchmark-autosave --benchmark-save-data --durations=5
       - name: Upload Benchmark Artifact to S3
         run: aws s3 cp --recursive .benchmarks s3://feast-ci-pytest-benchmarks
+
+  build-publish-docker-images:
```
tsotnet, line 138: Why do we need to publish these images for every commit?

adchia: Removed as discussed offline. Keeping the build though for a continuous build

tsotnet (approved):

PR #2119, opened 2021-12-08
Change the feast serve endpoint to be sync rather than async.

nossrannug:

When an endpoint is async, everything happens on the event loop thread.
This is great for functionality that is IO heavy since the job can be
awaited and other requests processed in the meantime. But if there are
no awaits, then everything happens synchronously on one thread which
means that while we're waiting for a response from a DB or API call,
nothing else gets processed.
To address this I've removed the async keyword from the endpoint.
This makes FastAPI process each request on a thread in a thread pool.

Testing this with postgres as the online store, running everything locally,
and 1 worker, req/sec went from 200 to 800.

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/feature_server.py
+++ b/sdk/python/feast/feature_server.py
@@ -2,6 +2,7 @@ import click
 import uvicorn
 from fastapi import FastAPI, HTTPException, Request
 from fastapi.logger import logger
+from fastapi.params import Depends
 from google.protobuf.json_format import MessageToDict, Parse
 
 import feast
@@ -15,17 +16,21 @@ def get_app(store: "feast.FeatureStore"):
 
     app = FastAPI()
 
+    async def get_body(request: Request):
+        return await request.body()
+
     @app.post("/get-online-features")
-    async def get_online_features(request: Request):
+    def get_online_features(body=Depends(get_body)):
         try:
             # Validate and parse the request data into GetOnlineFeaturesRequest Protobuf object
-            body = await request.body()
             request_proto = GetOnlineFeaturesRequest()
             Parse(body, request_proto)
 
             # Initialize parameters for FeatureStore.get_online_features(...) call
             if request_proto.HasField("feature_service"):
-                features = store.get_feature_service(request_proto.feature_service)
+                features = store.get_feature_service(
+                    request_proto.feature_service, allow_cache=True
+                )
             else:
                 features = list(request_proto.features.val)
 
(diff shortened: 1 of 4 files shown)


achals (approved):
Thnaks @nossrannug! This looks very promising.

sdk/python/feast/feature_server.py
```diff
@@ -61,11 +66,13 @@ async def get_online_features(request: Request):
...
+    store: "feast.FeatureStore", host: str, port: int, no_access_log: bool
+):
     app = get_app(store)
     click.echo(
         "This is an "
         + click.style("experimental", fg="yellow", bold=True, underline=True)
         + " feature. It's intended for early testing and feedback, and could change without warnings in future releases."
     )
-    uvicorn.run(app, host=host, port=port)
+    uvicorn.run(app, host=host, port=port, access_log=(not no_access_log))
```
pyalex, line 78: Hey @nossrannug , thanks for your PR. But I'm confused, what's the benefit of moving to the synchronous interface if we'll continue to use uvicorn as our server and at the end of the day we'll have a single-threaded app.

pyalex: Did you use some other server in your tests?

nossrannug: Hey @pyalex,
We're not running a single-threaded app. FastAPI supports mixing async and sync methods(endpoints) as much as you want and it will take care of running it correctly. When a method is async it will be executed using the event loop, but when a method is synchronous it will be executed in a thread pool.

sync:
image

async:
image

nossrannug:
> I think we either should remove async frameworks (`fastapi` and `uvicorn`) from our code completely or use asynchronous dispatcher but run heavy parts in thread executors

FastAPI does what you're suggesting when the method isn't async.
https://github.com/tiangolo/fastapi/blob/0a87bc88b85445018d3ced9d07ac6a37f9e6d8f1/fastapi/routing.py#L151
If the method is async, it will be awaited. But if the method isn't async then it is run in a threadpool.

pyalex:
Thanks for this clarification @nossrannug . In that case, this approach makes sense. Can you please fix conflicts?

sdk/python/feast/registry.py
```diff
@@ -104,6 +104,7 @@ def __init__(self, registry_config: RegistryConfig, repo_path: Path):
             repo_path: Path to the base of the Feast repository
             or where it will be created if it does not exist yet.
         """
+        self.lock = Lock()
```
pyalex, line 107: Should we have a more descriptive name for this variable, like `refresh_lock`. And can you please add `_` as a prefix?

nossrannug:
> Thanks for this clarification @nossrannug . In that case, this approach makes sense. Can you please fix conflicts?

Resolved

sdk/python/feast/registry.py
```diff
@@ -733,31 +737,31 @@ def _get_registry_proto(self, allow_cache: bool = False) -> RegistryProto:
...
+                    datetime.now()
+                    > (
+                        self.cached_registry_proto_created
+                        + self.cached_registry_proto_ttl
+                    )
+                )
             )
-        )
 
-        if allow_cache and (not expired or self.cache_being_updated):
```
pyalex, line 747: What is the reason this code is being deleted? This is from a recent PR if I remember correctly

nossrannug: This has been moved to line 756 and the cache_being_updated isn't used since we're wrapping this with a lock.

PR #2122, opened 2021-12-08
Fix release workflow to pass the python version and docker build targets

adchia:
The new pip-tools way of building feast requires a python version to be passed as an argument to the makefile rule. In the release, we were not doing this.

Also the make-build-X-docker image rule had a typo that didn't include the -docker suffix.

Fixes #

```release-note
NONE
```


achals (approved):
/lgtm

PR #2126, opened 2021-12-09
<do not merge yet> Update type_map.py to have right BQ to Feast type mapping for float

mavysavydav:
```release-note
Update type_map.py to have right BQ to Feast type mapping for float.
```

### DIFF
--- a/sdk/python/feast/type_map.py
+++ b/sdk/python/feast/type_map.py
@@ -404,7 +404,7 @@ def bq_to_feast_value_type(bq_type_as_str: str) -> ValueType:
         "INTEGER": ValueType.INT64,
         "INT64": ValueType.INT64,
         "STRING": ValueType.STRING,
-        "FLOAT": ValueType.DOUBLE,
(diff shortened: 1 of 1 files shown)


judahrand:
Is this correct? Is BigQuery `FLOAT` a valid type? I don't see it documented either as a type or an alias in the docs: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#floating_point_types

Feast's `FLOAT` type mostly maps to `float32` and `DOUBLE` maps to `float64`. Given that I can't see anything in the docs which suggests that BigQuery supports single precision floating point numbers (`float32`) it seems like this mapping isn't correct?

felixwang9817:
cc @judahrand I found some documentation [here](https://cloud.google.com/bigquery/data-types) referring to legacy SQL data types; the docs indicate that `FLOAT` is in fact double-precision, so it seems like the current mapping is correct

given this, I'm going to close this PR @mavysavydav, feel free to re-open if it's still relevant!

PR #2132, opened 2021-12-10
Fix release workflow to use the new GCP action

adchia:
There's an outdated action that fails on running the publish helm charts. Switching the release workflow to be consistent with our other workflows, which uses the new action.

Fixes #

```release-note
NONE
```


felixwang9817 (approved):
/lgtm

PR #2138, opened 2021-12-13
Cache dynamodb client and resource in DynamoDB online store implement…

felixwang9817:
…ation

**What this PR does / why we need it**: This PR caches the DynamoDB client and resource in the current online store implementation. Previously, the client and resource were reconstructed on each call to the online store, which cost ~80ms.

Fixes #

```release-note
NONE
```


achals (approved):
/lgtm

PR #2139, opened 2021-12-14
Remove underscores from ECR docker versions

### LINKED ISSUE #2136: Release workflow does not work
## Expected Behavior 
When pushing a new release, the release.yml workflow should succeed

## Current Behavior

1. The build docker images don't succeed because the aws lambda feature server docker image is labeled as e.g. 0_16_1 (unlike others which are 0.16.1) and that fails. See https://github.com/feast-dev/feast/runs/4493547092?check_suite_focus=true for example:

`Error response from daemon: No such image: feastdev/feature-server-python-aws:0.16.1`

This comes specifically from the section that does "Only push to latest tag if tag is the highest semver version "

2. Since the first step fails, it also cancels the other builds since it's all part of the same matrix so those images don't get pushed. Right now, I updated the latest docker images manually with the makefile commands 

3. I'm also pretty sure the publish-helm-charts script doesn't work since it seems to create infinitely recursive directories when i run it locally.

achals:

The initial assumption about ECR not supporting periods isn't true, so we should be able simplify some of the release logic. Also, the README was updated without updating the template, which this PR fixes.

Fixes #2136 

```release-note
none
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -146,16 +146,12 @@ push-ci-docker:
 build-ci-docker:
 	docker build -t $(REGISTRY)/feast-ci:$(VERSION) -f infra/docker/ci/Dockerfile .
 
-# Note, we use underscores instead of periods in the version name since ECR doesn't support periods. We automatically
-# pull from Docker Hub and push to ECR.
 push-feature-server-python-aws-docker:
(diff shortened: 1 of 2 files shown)


felixwang9817 (approved):
/lgtm

PR #2140, opened 2021-12-14
Add DatastoreTable infra object

felixwang9817:

**What this PR does / why we need it**: This PR adds an `InfraObject` for Datastore tables, as part of the `feast plan` implementation.

Fixes #

```release-note
NONE
```

### DIFF
--- a/protos/feast/core/DatastoreTable.proto
+++ b/protos/feast/core/DatastoreTable.proto
@@ -18,32 +18,26 @@ syntax = "proto3";
 
 package feast.core;
 option java_package = "feast.proto.core";
-option java_outer_classname = "InfraObjectProto";
+option java_outer_classname = "DatastoreTableProto";
 option go_package = "github.com/feast-dev/feast/sdk/go/protos/feast/core";
 
-import "feast/core/DynamoDBTable.proto";
-import "feast/core/DatastoreTable.proto";
+// Represents a Datastore table
+message DatastoreTable {
+    // Feast project of the table
+    string project = 1;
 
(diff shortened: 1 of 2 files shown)


protos/feast/core/DatastoreTable.proto
```diff
@@ -0,0 +1,43 @@
...
+    oneof project_id {
+        bool project_id_null = 3;
+        string project_id_value = 4;
+    }
+
+    // Datastore namespace, which can be null
+    oneof namespace {
+        bool namespace_null = 5;
+        string namespace_value = 6;
+    }
```
achals, line 42: These should probably be `google.protobuf.StringValue`. That's a struct type, that can be null - so you don't have to implement the same wrapper here and can reuse `StringValue` instead.

sdk/python/feast/infra/online_stores/datastore.py
```diff
@@ -279,3 +271,101 @@ def _delete_all_values(client, key) -> None:
...
+
+        return DatastoreTable(
+            project=infra_object_proto.datastore_table.project,
+            name=infra_object_proto.datastore_table.name,
+            project_id=project_id_value,
+            namespace=namespace_value,
+        )
+
+    def update(self):
+        client = _initialize_client(self.project_id, self.namespace)
```
achals, line 357: any reason this can't be initialized in `__init__`?

achals (changes requested):

achals (approved):
/lgtm

PR #2142, opened 2021-12-14
Refactor tag methods to infer created, deleted, and kept repo objects 

achals:

This is a first PR to support the `feast plan` command, as it applies to FCOs. It also adds a `diff_betewen` method for Registry objects, which is not complete in its implementation at this point.

Fixes #

```release-note
none
```

### DIFF
--- a/sdk/python/feast/diff/FcoDiff.py
+++ b/sdk/python/feast/diff/FcoDiff.py
@@ -1,0 +1,165 @@
+from dataclasses import dataclass
+from enum import Enum
+from typing import Any, List, Set, Tuple, Union
+
+from feast.entity import Entity
+from feast.feature_service import FeatureService
+from feast.feature_table import FeatureTable
+from feast.feature_view import FeatureView
+from feast.on_demand_feature_view import OnDemandFeatureView
+from feast.request_feature_view import RequestFeatureView
+
+
+@dataclass
+class PropertyDiff:
+    property_name: str
+    val_existing: str
+    val_declared: str
+
+
+class TransitionType(Enum):
+    UNKNOWN = 0
+    CREATE = 1
+    DELETE = 2
+    UPDATE = 3
+    UNCHANGED = 4
+
+
+@dataclass
+class FcoDiff:
+    current_fco: Any
+    new_fco: Any
+    fco_property_diffs: List[PropertyDiff]
+    transition_type: TransitionType
+
+
+class RegistryDiff:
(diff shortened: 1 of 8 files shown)


sdk/python/feast/diff/FcoDiff.py
```diff
@@ -0,0 +1,109 @@
...
+    def __init__(self):
+        self.fco_diffs = []
+
+    def add_fco_diff(self, fco_diff: FcoDiff):
+        self.fco_diffs.append(fco_diff)
+
+
+def _tag_registry_entities_for_keep_delete(
+    existing_entities: Set[Entity], desired_entities: Set[Entity]
+) -> Tuple[Set[Entity], Set[Entity], Set[Entity]]:
```
felixwang9817, line 48: in fact, can all of these methods be combined into one? e.g. `_tag_registry_objects_for_keep_delete`? seems like the internal logic is the exact same

sdk/python/feast/registry.py
```diff
@@ -104,20 +114,60 @@ def __init__(self, registry_config: RegistryConfig, repo_path: Path):
...
+
+            self._registry_store = cls(registry_config, repo_path)
+            self.cached_registry_proto_ttl = timedelta(
+                seconds=registry_config.cache_ttl_seconds
+                if registry_config.cache_ttl_seconds is not None
+                else 0
+            )
+
+    @classmethod
+    def from_proto(cls, regsitry_proto: RegistryProto):
```
felixwang9817, line 134: nit: `regsitry` is mispelled

felixwang9817: q: why do we need a `from_proto` for the registry? I thought the registry contents would always be persisted by the registry store

to clarify, even when we want to compare two registries, we will be comparing protos; I don't think we'll need to reconstruct a registry from a proto?

felixwang9817 (changes requested):

sdk/python/feast/diff/FcoDiff.py
achals: Yeah it's the same. I tried to do it but it made pypy very unhappy, but I can clean it up a bit and rename.

sdk/python/feast/registry.py
achals: yeah fair, I can change this to operate on the protos themselves. I think I misunderstood earlier

felixwang9817 (approved):
/lgtm

PR #2147, opened 2021-12-15
Add a feast plan command, and have CLI output differentiates between created, deleted and unchanged objects

achals:

Follow up to #2142 ,  this method uses the info from the refactored tag methods to print out new output.

Screen Shot 2021-12-14 at 9 47 19 PM

```release-note
Add a `feast plan` command.
`feast plan` and `feast apply` CLI output now differentiates between created, deleted and unchanged objects.
```


judahrand:
Does this also distinguish Updated from Created?

achals:
> Does this also distinguish Updated from Created?

The plan is to do that in a separate PR - but yes, we should be able to infer and print out updated fields for objects.

This pr doesn't really grok in place updates to objects yet.

sdk/python/feast/feature_store.py
```diff
@@ -513,26 +519,128 @@ def apply(
...
+                    FeatureView,
+                    OnDemandFeatureView,
+                    RequestFeatureView,
+                    Entity,
+                    FeatureService,
+                ]
+            ]
+        ] = None,
+        partial: bool = True,
+    ):
```
felixwang9817, line 584: nit: return type should be diff

sdk/python/feast/feature_store.py
```diff
@@ -358,7 +360,7 @@ def _get_features(self, features: Union[List[str], FeatureService],) -> List[str
         return _feature_refs
 
     @log_exceptions_and_usage
-    def apply(
+    def plan(
```
felixwang9817, line 363: it's extremely unintuitive to me that `plan` already has a concept of objects to keep and delete - after all, the point of planning should be to determine which things to keep, add, and delete

what I had in mind with plan was to reconstruct a full RegistryProto from a ParsedRepo, and then compare that with the existing registry - thoughts on this approach vs. yours?

sdk/python/feast/feature_store.py
achals: yup

sdk/python/feast/feature_store.py
achals: Okay, I think that make sense. Lemme update.

felixwang9817 (approved):
/lgtm

sdk/python/feast/registry.py
```diff
@@ -128,32 +129,85 @@ def __init__(
                 else 0
             )
 
+    def clone(self) -> "Registry":
+        new_registry = Registry(None, None)
+        new_registry.cached_registry_proto_ttl = timedelta(seconds=0)
```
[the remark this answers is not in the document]
felixwang9817: @judahrand I'm actually not sure why, but we don't actually use this `clone` method anymore (I'll delete it in a future PR)

PR #2149, opened 2021-12-15
Implement rollback API for providers

### LINKED ISSUE #2116: Feast apply can leak resources and create inconsistent state
Consider feature store with AWS as a configured provider.
When we call `feast apply` under the hood `AwsProvider.update_infra` is being called:

```
def apply(...):
    ... preparation ...
    provider.update_infra(...)
    registry.commit()
```

Now, `update_infra` could be a multi-staged operation. For example, current AWS implementation consists of updating online store and then creating Lambda feature server:
```
def update_infra(...):
    online_store.update(...)
    deploy_lambda_server(...)
```

When an exception occurs in `deploy_lambda_server` online store is already created, however since `registry.commit()` is called only when all infra was successfully updated - these changes will be lost and resources from online store won't be cleaned up properly on `teardown`.

When number of different components that are managed by the provider will grow even further - the risk to leak resources and create inconsistent state will also increase. Thus, we need to add rollback mechanism, that will be triggered in case of exception.

## Expected Behavior 
Either all infrastructure is updated or none.

## Current Behavior
Infrastructure can be updated partially and the registry will not have records about resources that were actually created.

### Specifications

- Version: 0.15.1
- Platform:
- Subsystem:

## Possible Solution

Create rollback API.

judahrand:
Addresses the issue mentioned. This is very much a first pass solution so keen for any and all feedback.

Fixes #2116

```release-note
None
```

### DIFF
--- a/sdk/python/feast/infra/aws.py
+++ b/sdk/python/feast/infra/aws.py
@@ -63,47 +63,64 @@ class AwsProvider(PassthroughProvider):
         entities_to_keep: Sequence[Entity],
         partial: bool,
     ):
-        self.online_store.update(
-            config=self.repo_config,
-            tables_to_delete=tables_to_delete,
-            tables_to_keep=tables_to_keep,
-            entities_to_keep=entities_to_keep,
-            entities_to_delete=entities_to_delete,
-            partial=partial,
-        )
-
-        if self.repo_config.feature_server and self.repo_config.feature_server.enabled:
-            if not enable_aws_lambda_feature_server(self.repo_config):
-                raise ExperimentalFeatureNotEnabled(FLAG_AWS_LAMBDA_FEATURE_SERVER_NAME)
-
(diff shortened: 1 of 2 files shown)


judahrand:
The AWS failures make it a bit tricky to tell that this is working as intended. I'm not sure how to proceed on that front.

felixwang9817:
@judahrand thanks for this PR! we believe we've fixed most of the issues with AWS integration tests (we increased the size of our Redshift cluster for tests and found a bug [here](https://github.com/feast-dev/feast/pull/2154) that kept leaking Dynamo tables), so we're hoping the tests are fine going forward (we still expect occasional leakages until this [issue](https://github.com/feast-dev/feast/issues/2116) has been resolved, but they should be very manageable)

as for the rollback API, we're still considering whether it should be supported, but for now since tests should be fine, let's put a hold on this PR!

PR #2164, opened 2021-12-20
Improve online deserialization latency

judahrand:

For `_LIST` type features this implementation is up to 7x faster in my testing. 

For serializing and deserializing 10k rows of 3 columns `DOUBLE`, 3 columns `FLOAT_LIST`.

Old implementation:
image

New implementation:
image

Fixes #

```release-note
None
```

### DIFF
--- a/sdk/python/feast/type_map.py
+++ b/sdk/python/feast/type_map.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import base64
 import re
 from datetime import datetime
 from typing import Any, Dict, List, Optional, Set, Tuple, Type
@@ -20,7 +19,6 @@ from typing import Any, Dict, List, Optional, Set, Tuple, Type
 import numpy as np
 import pandas as pd
 import pyarrow
-from google.protobuf.json_format import MessageToDict
 from google.protobuf.pyext.cpp_message import GeneratedProtocolMessageType
 from google.protobuf.timestamp_pb2 import Timestamp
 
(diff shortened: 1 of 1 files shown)


judahrand:
@tsotnet Relevant for your benchmarking perhaps?

achals:
This is great and much cleaner to read @judahrand - I'm curious where the speed up is coming from, is it just because we're avoiding the excess branching and if statements?

judahrand:
> This is great and much cleaner to read @judahrand - I'm curious where the speed up is coming from, is it just because we're avoiding the excess branching and if statements?

I think it is the branching and ifs but also for list types the old implementation was looping over each element of the list and type casting it. This implementation avoids the need for that by not using `MessageToDict`. 

From what I can tell `MessageToDict` does some very odd things to types. Really it should probably be called `MessageToJSONDict`! Accessing the elements directly avoids the weirdness with types.

judahrand:
I'd also add that this is relying much more on letting Protobuf sort out the conversions which might mean more of this is being done in C++ extensions?

judahrand:
image

This is an interesting comparison between `master` and these changes for both with and without the C++ extensions.

judahrand:
Looks like most platform's wheels from `pip` do use the C++ extensions by default which is a relief!

achals (approved):
/lgtm

judahrand:
@achals Interestingly, there is the possibility of getting another small performance benefit by doing what is documented here: https://m.atthew.uk/posts/post.html

This is probably not worth the deploy nightmare, however.

achals:
> @achals Interestingly, there is the possibility of getting another small performance benefit by doing what is documented here: https://m.atthew.uk/posts/post.html
> 
> 
> 
> This is probably not worth the deploy nightmare, however.

Thanks for the detailed info! Yeah I'm inclined to opt for simplicity over speed in this particular case. I'm sure there's other avenues to speed things up that won't complicate our packaging and deploy story.

PR #2165, opened 2021-12-21
Improve serialization performance

judahrand:

Similar quick pass as the deserialize PR (#2164). A definite improvement but not as drastic as the other one. Looks like between 20-30% improvement. This is achieved by avoiding memory copies as much as possible by using Numpy and allowing Protobuf to handle that more sensibly than copying to convert to Python object then copying back again to create Protobuf Message.

Same benchmark as the deserialize PR:

On `master`:
image

With changes:
image

Fixes #

```release-note
None
```


achals (approved):
/lgtm

PR #2172, opened 2021-12-28
Convert python values into proto values in bulk

pyalex:

Packing python native values (or NumPy arrays) into proto objects (generated proto classes) takes roughly 80% of the time during materialization (not counting offline / online storage part). Ironically protobuf serialization itself takes around 10% of the time.

This PR refactors `python_values_to_proto_values` function to process data in bulk (essentially column by column), which improves performance by running all type checks only once (per batch / column). According to my analysis, this move will decrease packing time roughly by 50% without introducing significant changes (see first comment).

Fixes #

```release-note
none
```

### DIFF
--- a/sdk/python/feast/type_map.py
+++ b/sdk/python/feast/type_map.py
@@ -253,7 +253,7 @@ def _python_value_to_proto_value(
         List of Feast Value Proto
     """
     # ToDo: make a better sample for type checks (more than one element)
-    sample = next(filter(None, values), None)  # first not empty value
+    sample = next(filter(np.any, values), None)  # first not empty value
     if not sample:
         # all input values are None or empty lists
         return [ProtoValue()] * len(values)
@@ -316,7 +316,7 @@ def python_values_to_proto_values(
 ) -> List[ProtoValue]:
     value_type = feature_type
     if len(values) and feature_type == ValueType.UNKNOWN:
-        sample = next(filter(None, values), None)  # first not empty value
+        sample = next(filter(np.any, values), None)  # first not empty value
         if isinstance(sample, (list, np.ndarray)):
             value_type = (
                 feature_type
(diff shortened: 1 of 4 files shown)


pyalex:
My performance analysis was based on case proposed by @judahrand:
Feature view with 3 features of primitive type and 3 features of list type:
```
fv = FeatureView(
    name="fv",
    entities=["fake_entity"],
    features=[
        Feature(
            name=str(i),
            dtype=ValueType.DOUBLE,
        ) for i in range(3)
    ] + [
        Feature(
            name=str(i),
            dtype=ValueType.FLOAT_LIST,

        ) for i in range(5, 8)
    ],
    ttl=None,
    batch_source=FakeSource,
)
```

Dataset with 10 thousands rows was generated
```
ROWS = 10_000
data = {
    'id': np.random.random_integers(low=0, high=int(1e12), size=(ROWS,))
}
data.update(
    {
        str(i): np.random.random(size=(ROWS,)).astype('float64')
        for i in range(3)
    }
)
data.update(
    {
        str(i): [np.random.random(size=(10,)).astype('float32') for _ in range(ROWS)]
        for i in range(5, 8)
    }
)
```

`feast.infra.provider import _convert_arrow_to_proto` was benchmark target (on my local).

Results:
`master`: 570 ms (average)
`this PR`: 320 ms (average).

judahrand:
This looks great!

sdk/python/feast/type_map.py
```diff
@@ -238,50 +238,59 @@ def _type_err(item, dtype):
...
     Args:
         feast_value_type: The target value type
-        value: Value that will be converted
+        values: List of Values that will be converted
 
     Returns:
-        Feast Value Proto
+        List of Feast Value Proto
     """
+    # ToDo: make a better sample for type checks (more than one element)
```
adchia, line 255: Mind making an issue tracking this after the PR goes in?

adchia (commented):
/lgtm

adchia (approved):

PR #2173, opened 2021-12-29
ability to pull all entity rows for specific entities

vas28r13:

Add default ability to pull all entity_rows for the specified feature view(s) when requesting online features (only implemented for the Redis online store)

So if `entity_rows` are not provided and the feature flag is on then it will pull all relevant entity rows for the specified feature_view(s)

This can be useful for ranking problems where every entity needs to be scored and this way the client doesn't need to keep track of entity id's.
This can also be useful for testing and debugging.

~entity_rows=[{"driver_id": 1001}, {"driver_id": 1002}, {"driver_id": 1003}, {"driver_id": 1004}],~
```
from feast import FeatureStore, RepoConfig
fs = FeatureStore(config=RepoConfig(registry="feature_repo/data/registry.db", project="feature_repo", provider="local"))
online_response = fs.get_online_features(
    features=[
        "driver_hourly_stats:avg_daily_trips",
    ],
)
online_response_dict = online_response.to_dict()
# would return all rows 
```

### DIFF
--- a/sdk/python/feast/infra/key_encoding_utils.py
+++ b/sdk/python/feast/infra/key_encoding_utils.py
@@ -1,5 +1,5 @@
 import struct
-from typing import List, Tuple
+from typing import Any, ByteString, List, Tuple
 
 from feast.protos.feast.types.EntityKey_pb2 import EntityKey as EntityKeyProto
 from feast.protos.feast.types.Value_pb2 import Value as ValueProto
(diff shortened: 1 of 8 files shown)


sdk/python/feast/infra/key_encoding_utils.py
```diff
@@ -62,3 +62,27 @@ def serialize_entity_key(entity_key: EntityKeyProto) -> bytes:
         output.append(val_bytes)
 
     return b"".join(output)
+
+
+def deserialize_entity_values(entity_values: ByteString) -> List[Any]:
+    deserialized_entity_values = []
```
adchia, line 68: might be worth adding a comment that this is mostly partially reversing serialize_entity_key above, but assumes the input has already stripped out the entity keys themselves? Kind of hacky, but seems fine for now.

adchia (commented):
test?

sdk/python/feast/infra/key_encoding_utils.py
vas28r13: hey good point

vas28r13:
will add some tests for this

adchia:
As mentioned, it's not clear how canonical this should be.

e.g. there's good reason for https://github.com/feast-dev/feast/issues/1611 for historical retrieval. in the online usecase for batch retrieval, you don't really need an online store. Some other motivation might be needed to add this

achals:
@adchia @vas28r13 did we determine what we wanted to do with this PR?

adchia:
Closing since we probably want to think this through and now the original logic has diverged enough.

PR #2166, opened 2021-12-21
[Java feature server] Converge ServingService API to make Python and Java feature servers consistent

pyalex:
**What this PR does / why we need it**:

Currently, Java Feature Server and Python Feature Server provide two different client APIs. This PR aimed to fix that.

1. Changes to proto API: use `GetOnlineFeatureRequest` (that was created for and currently used by Python FS) as input in `GetOnlineFeatures` method. That also means that Java FS will support requests by **feature service**.
2. `GetOnlineFeatureResponseV2` is proposed as optimized version of `GetOnlineFeatureResponse` and was designed symmetrically to `GetOnlineFeatureRequest`. According to my evaluation serialization/deserialization of the new message is 3x times faster (on 100 rows). That difference increases with a growing number of rows and features. New message also optimized for compaction (message size), which will lead to faster network transmission.

What was updated:
1. Java Feature Server
2. Java SDK
3. Go SDK

What's not covered in this PR:
1. Python FS to return new response

Some clean up:
* "Feature Table" is replaced to "Feature View"
* Suffix "V<N>" is dropped when there are no other versions

**Which issue(s) this PR fixes**:

Fixes #

**Does this PR introduce a user-facing change?**:

```release-note
Java feature server supports retrieving features by feature service name.

ServingService proto API is refactored:
GetOnlineFeaturesResponse now has new argument `request_context` for passing input to ODFVs (previously this inputs were mixed into `entities`
GetOnlineFeaturesResponseV2 is refactored version of response optimized for fast serialization / deserialization and high compaction
```

PR #2174, opened 2021-12-29
optional include event ts in online feature set

vas28r13:

`include_event_ts` optional param to include the event timestamp in the online feature set

This can be useful if the event timestamp impacts logic on the service side of things. 
Can also help evaluate how old the feature vector is.

```
online_response = fs.get_online_features(
    features=[
        "driver_hourly_stats:avg_daily_trips",
    ],
    entity_rows=[{"driver_id": 1001}, {"driver_id": 1002}, {"driver_id": 1003}, {"driver_id": 1004}],
    include_event_ts=True,
)

df = online_response.to_df()
df["driver_hourly_stats__ts"]  # should return when the hourly stats were ingested per driver entity
```

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -69,6 +69,7 @@ from feast.protos.feast.serving.ServingService_pb2 import (
     GetOnlineFeaturesResponse,
 )
 from feast.protos.feast.types.EntityKey_pb2 import EntityKey as EntityKeyProto
+from feast.protos.feast.types.Value_pb2 import Value as ValueProto
 from feast.registry import Registry
 from feast.repo_config import RepoConfig, load_repo_config
 from feast.request_feature_view import RequestFeatureView
@@ -1005,6 +1006,7 @@ class FeatureStore:
         features: Union[List[str], FeatureService],
         entity_rows: List[Dict[str, Any]],
         full_feature_names: bool = False,
+        include_event_ts: bool = False,
     ) -> OnlineResponse:
         """
         Retrieves the latest online feature data.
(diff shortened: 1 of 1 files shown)


adchia:
You'll need to fix the linter btw. And also add a release note in the description. 

e.g. 

````
```release-note
Allows including event timestamps in online features response
```
````

**Comment by adchia** [comment]
also add to an existing test?

**Comment by woop** [comment]
Just a question here. The point of [FieldStatus](https://github.com/feast-dev/feast/blob/e435d927ecaa2037a6e96cbdf1584e4a15d17597/protos/feast/serving/ServingService.proto#L108) was to provide information about feature values (some of those fields relate to age).
* Is there a reason we'd return the timestamp as part of the feature vector and not a status object or some other metadata object? 
* How would a user replicate the timestamp response on the training side?
* What happens when you have multiple feature views in a single response?

**Comment by vas28r13** [comment]
@woop great questions.
* there is no reason this needs to be part of the fields in the `OnlineResponse`. Like you mentioned, it may be better to have it as separate metadata in the `OnlineReponse` object? I wanted to avoid changing the response object so it felt easiest to have it in the resulting df.
```
online_response = fs.get_online_features(
    features=[
    ...
    include_event_ts=True,
)

df = online_response.to_df()
df["driver_hourly_stats__ts"]  # should return when the hourly stats were ingested per driver entity
```

* I think this is generally not needed on the training side. I don't necessarily treat this as a feature for the model. There are 2 cases where we use the event timestamps in the ml service:
1. validate that the data is not stale at the time of prediction or track the age of the features at prediction time
2. there is some business logic on the client side based on the age of the features

adchia:
Looks like we started adding this in https://github.com/feast-dev/feast/pull/2166.

Would also be good to think through what the best way to use this as a user of get_online_features (to specifically post process) since the current response python is kind of clunky

adchia:
@tsotnet Assuming you're planning on releasing the new python format soon, so we can probably just not merge this in and migrate to the new response?

adchia:
Created a new issue to track this given the new response: https://github.com/feast-dev/feast/issues/2260

PR #2182, opened 2022-01-04
Speed up Datastore deletes by batch deletions with multithreading

ptoman-pa:
Speeds up teardown in Datastore by 3-4 orders of magnitude via Datastore's `delete_multi` plus multithreading with a work queue. (Previous approach was one-request-per-element single-threaded deletions.)

No issue

```release-note
NONE
```


sdk/python/feast/infra/online_stores/datastore.py
```diff
@@ -32,6 +35,9 @@
 from feast.protos.feast.types.Value_pb2 import Value as ValueProto
 from feast.repo_config import FeastConfigBaseModel, RepoConfig
 from feast.usage import log_exceptions_and_usage, tracing_span
+from feast.utils.generic_utils import AtomicCounter
```
achals, line 38: We've generally tried to avoid adding `utils` modules since it's a kitchen sink/code smell. If this `AtomicCounter` is only used in this datastore class, maybe we can just define it here? Or rename `feast.utils.generic_utils` to `feast.atomic` or something?

sdk/python/feast/infra/online_stores/datastore.py
```diff
@@ -262,15 +268,35 @@ def online_read(
...
+    BATCH_SIZE = 500  # Dec 2021: delete_multi has a max size of 500
+    NUM_THREADS = 3
+    deletion_queue = Queue()
+    status_info_counter = AtomicCounter()
+
+    def worker(shared_counter):
+        while True:
+            client.delete_multi(deletion_queue.get())
+            shared_counter.increment()
+            LOGGER.info(
```
achals, line 284: This should maybe be `.debug`?

achals (changes requested):

sdk/python/feast/infra/online_stores/datastore.py
ptoman-pa: I'm actually glad to hear this :) I wanted to use the patterns of the codebase. I'm going to move it into an innerclass just for the `_delete_all_values` since it's only used for that logging statement (or we could remove the logging), and it can always be migrated to increasing scope if needed.

sdk/python/feast/infra/online_stores/datastore.py
```diff
@@ -262,15 +267,46 @@ def online_read(
...
+        # for tracking how many deletions have already occurred; not used outside this method
+        def __init__(self):
+            self.value = 0
+            self.lock = Lock()
+
+        def increment(self):
+            with self.lock:
+                self.value += 1
+
+    BATCH_SIZE = 500  # Dec 2021: delete_multi has a max size of 500
```
achals, line 285: Nit is there a doc we can link to in case this changes in the future?

ptoman-pa: I just went looking and couldn't find an obvious place.  If we try to query with more than 500, the server responds:
```
google.api_core.exceptions.InvalidArgument: 400 cannot write more than 500 entities in a single call
```

achals: Maybe https://cloud.google.com/datastore/docs/concepts/limits?

achals (approved):
/lgtm

PR #2185, opened 2022-01-04
Avoid requesting features from OnlineStore twice

### LINKED ISSUE #2184: ODFV dependencies can be retrieved twice from the OnlineStore
## Expected Behavior
When calling `get_online_features` a given feature should only ever be retrieved from the OnlineStore once.

## Current Behavior
When a feature which is also a dependency of an ODFV which is requested is requested the feature is retrieved twice from the OnlineStore.

Once in [`_populate_result_rows_from_feature_view`](https://github.com/feast-dev/feast/blob/e435d927ecaa2037a6e96cbdf1584e4a15d17597/sdk/python/feast/feature_store.py#L1152) and again in [`_populate_odfv_dependencies`](https://github.com/feast-dev/feast/blob/e435d927ecaa2037a6e96cbdf1584e4a15d17597/sdk/python/feast/feature_store.py#L1165).

### Specifications

- Version:
- Platform:
- Subsystem:

## Possible Solution
Check if the feature has already been requested before retrieving it again.

judahrand:

Fixes #2184

```release-note
None
```

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -1233,12 +1233,16 @@ class FeatureStore:
         if len(grouped_odfv_refs) > 0:
             for odfv, _ in grouped_odfv_refs:
                 for fv in odfv.input_feature_views.values():
-                    feature_deps = [
+                    # Find the set of required Features which have not yet
+                    # been retrieved.
+                    not_yet_retrieved = {
                         feature.name
-                        for feature in fv.features
+                        for feature in fv.projection.features
                         if f"{fv.name}:{feature.name}" not in retrieved_feature_refs
-                    ]
-                    if feature_deps:
+                    }
+                    # If there are required Features which have not yet been retrieved
+                    # retrieve them.
+                    if not_yet_retrieved:
                         table_join_keys = [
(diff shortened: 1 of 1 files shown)


sdk/python/feast/feature_store.py
```diff
@@ -1223,19 +1233,25 @@ def _populate_odfv_dependencies(
         if len(grouped_odfv_refs) > 0:
             for odfv, _ in grouped_odfv_refs:
                 for fv in odfv.input_feature_views.values():
-                    table_join_keys = [
-                        entity_name_to_join_key_map[entity_name]
-                        for entity_name in fv.entities
+                    feature_deps = [
```
achals, line 1236: Trying to follow this code I'm confused what `feature_deps` means here and what this block of code does.. Any possibility to rename/refactor/add comments?

achals (changes requested):

sdk/python/feast/feature_store.py
judahrand: Had another pass at this and caught a case I hadn't considered + added some comments etc. More clear now?

achals (approved):
/lgtm

PR #2186, opened 2022-01-04
Use FeatureViewProjection instead of FeatureView in ODFV

judahrand:

Before this PR if an ODFV feature is requested then all Features from FeatureView dependencies must be retrieved. This is not performant for FeatureViews which are very wide but with few required features.

This change allows the following spelling to require only a subset of features:
```python
# Define a request data source which encodes features / information only 
# available at request time (e.g. part of the user initiated HTTP request)
input_request = RequestDataSource(
    name="vals_to_add",
    schema={
        "val_to_add": ValueType.INT64,
        "val_to_add_2": ValueType.INT64
    }
)

# Use the input data and feature view features to create new features
@on_demand_feature_view(
   inputs={
       'driver_hourly_stats': driver_hourly_stats_view[['conv_rate']],
       'vals_to_add': input_request
   },
   features=[
     Feature(name='conv_rate_plus_val1', dtype=ValueType.DOUBLE),
     Feature(name='conv_rate_plus_val2', dtype=ValueType.DOUBLE)
   ]
)
def transformed_conv_rate(features_df: pd.DataFrame) -> pd.DataFrame:
    df = pd.DataFrame()
    df['conv_rate_plus_val1'] = (features_df['conv_rate'] + features_df['val_to_add'])
    df['conv_rate_plus_val2'] = (features_df['conv_rate'] + features_df['val_to_add_2'])
    return df
```

This can dramatically reduce serving latency of ODFVs in some cases.

Fixes #

```release-note
- Use `FeatureViewProjections` to determine OnDemandFeatureView dependencies.
- Fix case where changes to OnDemandFeatureView would not be applied to FeatureStore.
- Reduce serving latency when features from both FeatureViews and OnDemandFeatureViews are requested.
- Improve performance of `get_online_features`.
```


protos/feast/core/OnDemandFeatureView.proto
```diff
@@ -41,9 +42,13 @@ message OnDemandFeatureViewSpec {
     // List of features specifications for each feature defined with this feature view.
     repeated FeatureSpecV2 features = 3;
 
-    // List of features specifications for each feature defined with this feature view.
+    // Map of inputs for this feature view.
     map<string, OnDemandInput> inputs = 4;
 
+    // Map of projections for the inputs of this feature view.
+    // Keys must also be present in `inputs`
+    map<string, FeatureViewProjection> projections = 6;
```
achals, line 50: Can't we add `FeatureViewProjection` in the `OnDemandInput` oneof? And remove the feature view if you so wish?

achals (changes requested):

protos/feast/core/OnDemandFeatureView.proto
judahrand: I ended up avoiding this retrieves the entities of the backing FeatureViews.  
https://github.com/feast-dev/feast/blob/7d4369f0b1ce1692bb4fab87c0f70d8e7cdc9fd3/sdk/python/feast/cli.py#L287

This can't be done if a FeatureView Projection is store instead of the FeatureView.

achals: Hmmm should we look up the entites expicitly via the feature view name, if it's just in that method?

achals: I'd be worried to have two fields in the proto which have this weird dependency between each other, and would prefer a single field - and it feels like we should be able to get away with it.

judahrand: 👍 I'll take another pass at doing that tomorrow.

judahrand: This ended up being quite a bit more involved.

The changes in this PR definitely reduce the number of calls to the OnlineStore in the simple case of Features being requested from 'an ODFV which depends on a projection of a FV' and from the unprojected version of that FV (1 call rather than 2 and only the required features).

However, when multiple different Projections are used in a FeatureService I'm not 100% we won't end up making multiple calls to the OnlineStore for each Projection + the projection used by the ODFV.

I'm also not 100% we catch the case where 2 ODFVs used different projections of the same FV. 

But I do think this approach of working out what we need before starting to request things from the OnlineStore is better than it was.

judahrand:
@achals I've tacked on some additional changes here which improve the performance of `get_online_features` quite significantly for large numbers of `entity_rows`. I'm seeing 2-3x speed ups for parts of the serialization of EntityKeyProtos and adding the `entity_rows` to the `results_rows`.

judahrand:
@achals All the tests other than the ones that are failing on `master` are now fixed. It looks like I've actually fixed a handful of bugs which required some alternations to the tests as you can see. This is because the `customer` entity is actually defined as `ValueType.STRING` but the tests treated it as if it were `ValueType.INT64`. 

Do you think you could give this another review and look at merging?

achals:
Nit, release notes are a bit unclear:

```
- Update ODFV in UDF code changes.
- Reduce number of calls to OnlineStore ODFVs are used.
```

judahrand:
> Nit, release notes are a bit unclear:
> 
> ```
> - Update ODFV in UDF code changes.
> - Reduce number of calls to OnlineStore ODFVs are used.
> ```

Reworded those two.

sdk/python/feast/on_demand_feature_view.py
```diff
@@ -132,6 +159,10 @@ def from_proto(cls, on_demand_feature_view_proto: OnDemandFeatureViewProto):
             if on_demand_input.WhichOneof("input") == "feature_view":
                 inputs[input_name] = FeatureView.from_proto(
                     on_demand_input.feature_view
+                ).projection
+            elif on_demand_input.WhichOneof("input") == "feature_view_projection":
+                inputs[input_name] = FeatureViewProjection.from_proto(
+                    on_demand_input.feature_view_projection
```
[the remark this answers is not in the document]
judahrand: Yes, I believe they should be updated the next time that `feast apply` is run. The diff is 'large' but makes sense. I'd also suggest maybe this isn't a big deal as the ODFVs are still labelled as Alpha (and I think will need some iteration as calling either `to_dict` or `to_df` for them isn't terribly performant).

sdk/python/feast/on_demand_feature_view.py
```diff
@@ -79,11 +78,37 @@ def proto_class(self) -> Type[OnDemandFeatureViewProto]:
...
+        if not self.input_request_data_sources == other.input_request_data_sources:
+            return False
+
+        if not self.udf.__code__.co_code == other.udf.__code__.co_code:
+            return False
+
+        return True
+
+    def __hash__(self):
+        return super().__hash__()
```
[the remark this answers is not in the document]
judahrand: By default only the `name` and `features` properties are checked. This seemed obviously wrong to me... You could change the inputs and UDF but keep the outputs and name the same. 

https://github.com/feast-dev/feast/blob/30f7bbad96651a19c698663582eafa43779d313e/sdk/python/feast/base_feature_view.py#L105-L117

Then is `__eq__` is overridden `__hash__` also has to be! Annoying but true.
https://docs.python.org/3/reference/datamodel.html#object.__hash__

sdk/python/feast/on_demand_feature_view.py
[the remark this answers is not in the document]
achals: Yeah fair I don't really want to block this change on the perceived change, but maybe it should be something we call out in the release notes (or the blog post about the release).

sdk/python/feast/on_demand_feature_view.py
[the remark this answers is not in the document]
achals: yeah I remember that `__hash__` has to be overridden too - okay cool makes sense.

PR #2187, opened 2022-01-04
Support multiple application properties files (incl from classpath)

pyalex:

After the recent deletion of Spring in Java feature server we have some functionality degradation:

* Multiple configuration files are not supported
* Default configuration (stored in classpath) cannot be loaded

Also `ApplicationProperties$Store` class cannot be deserialized (bug) due to the absense of the default constructor.

This PR addresses all those issues and returns back the possibility to override configurations.
Also helm chart and docker-compose were updated to new expected CMD format.

Fixes #

```release-note
None
```


achals (approved):
/lgtm

PR #2190, opened 2022-01-05
Add InfraDiff class for feast plan

felixwang9817:
**What this PR does / why we need it**: This PR adds the `InfraDiff` class and adds support for `Infra` objects to the `Registry`.

Fixes #

```release-note
NONE
```


sdk/python/feast/diff/utils.py
```diff
@@ -0,0 +1,17 @@
+from dataclasses import dataclass
+from enum import Enum
```
achals, line 2: nit, prefer a different name for this file than `utils.py`? It's a smell

achals (commented):
Can you update the commit message/PR title should be more descriptive

achals (approved):
Can you also print out what `feast plan` produces when infra is changed in the PR description?

felixwang9817:
> Can you also print out what `feast plan` produces when infra is changed in the PR description?

this PR doesn't actually change the output of `feast plan`

PR #2192, opened 2022-01-05
Refactor `_get_table_entity_keys` for small performance benefit and readability.

judahrand:

I've been looking for general latency improvements. For the example I'm testing against, which is ~9000 entity rows, 2 FVs, with one entity each and 1 ODFV with 2 entities, I'm seeing 30-60ms gain in cProfiler.

Fixes #

```release-note
None
```

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -15,7 +15,7 @@ import copy
 import itertools
 import os
 import warnings
-from collections import Counter, OrderedDict, defaultdict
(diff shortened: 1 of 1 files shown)


judahrand:
Question relating to this - is it safe to assume that the ordering in `entity_keys` is consistent? Eg. that `foo.join_keys` is always in the same order?

judahrand:
Another question, do you think it is worth trying to work the same exceptions back in?

judahrand:
Dealt with better by #2186.

PR #2197, opened 2022-01-06
Persisting results of historical retrieval

pyalex:

This PR introduces persisting results of historical retrieval. This feature is required by Data Quality Monitoring and described in the [DQM RFC](https://docs.google.com/document/d/110F72d4NTv80p35wDSONxhhPBqWRwbZXG4f9mNEMd98/edit#heading=h.9gaqqtox9jg6). The difference with RFC, though, is that this PR proposes to persist datasets in configured offline store (currently supported Redshift / BigQuery / File) instead of always writing it to cloud storage as parquet files.

Example of usage:
```
from feast.infra.offline_stores.bigquery_source import SavedDatasetBigQueryStorage

job = store.get_historical_features(...)

ds = store.create_saved_dataset(
   from_=job,
   name='my_persisted_ds',
   storage=SavedDatasetBigQueryStorage(table_ref='project.dataset.table_name'),
   tags={'key': 'value'}
)

ds = store.get_saved_dataset(name='my_persisted_ds')
ds.to_df()
```

`get_saved_dataset(name)` returns `SavedDataset` object which provides access to dataset metadata. Data can be retrieved by calling `.to_df()` or `.to_arrow()` on this object.

Fixes #

```release-note
Two new methods in FeatureStore class
create_saved_dataset - create dataset from historical retrieval job; persist it in offline store and write metadata to the registry
get_saved_dataset - retrieve metadata from registry and data from offline store
```

### DIFF
--- a/docs/getting-started/concepts/dataset.md
+++ b/docs/getting-started/concepts/dataset.md
@@ -1,0 +1,46 @@
+# Dataset
+
+Feast datasets allow for conveniently saving dataframes that include both features and entities to be subsequently used for data analysis and model training.
+[Data Quality Monitoring](https://docs.google.com/document/d/110F72d4NTv80p35wDSONxhhPBqWRwbZXG4f9mNEMd98) was the primary motivation for creating dataset concept.
+
+Dataset's metadata is stored in the Feast registry and raw data (features, entities, additional input keys and timestamp) is stored in the [offline store](../architecture-and-components/offline-store.md).
+
+Dataset can be created from:
+1. Results of historical retrieval
+2. [planned] Logging request (including input for [on demand transformation](../../reference/alpha-on-demand-feature-view.md)) and response during feature serving
+3. [planned] Logging features during writing to online store (from batch source or stream)
+
+
+### Creating Saved Dataset from Historical Retrieval
+
+To create a saved dataset from historical features for later retrieval or analysis, a user needs to call `get_historical_features` method first and then pass the returned retrieval job to `create_saved_dataset` method.
+`create_saved_dataset` will trigger provided retrieval job (by calling `.persist()` on it) to store the data using specified `storage`.
+Storage type must be the same as globally configured offline store (eg, it's impossible to persist data to Redshift with BigQuery source).
+`create_saved_dataset` will also create SavedDataset object with all related metadata and will write it to the registry.
+
+```python
+from feast import FeatureStore
+from feast.infra.offline_stores.bigquery_source import SavedDatasetBigQueryStorage
+
+store = FeatureStore()
+
+historical_job = store.get_historical_features(
+    features=["driver:avg_trip"],
+    entity_df=...,
+)
+
+dataset = store.create_saved_dataset(
+    from_=historical_job,
+    name='my_training_dataset',
+    storage=SavedDatasetBigQueryStorage(table_ref='<gcp-project>.<gcp-dataset>.my_training_dataset'),
+    tags={'author': 'oleksii'}
+)
+
+dataset.to_df()
+```
+
+Saved dataset can be later retrieved using `get_saved_dataset` method:
+```python
+dataset = store.get_saved_dataset('my_training_dataset')
+dataset.to_df()
+```
--- a/protos/feast/core/Registry.proto
+++ b/protos/feast/core/Registry.proto
@@ -28,6 +28,7 @@ import "feast/core/FeatureView.proto";
 import "feast/core/InfraObject.proto";
 import "feast/core/OnDemandFeatureView.proto";
 import "feast/core/RequestFeatureView.proto";
+import "feast/core/SavedDataset.proto";
 import "google/protobuf/timestamp.proto";
 
 message Registry {
@@ -37,6 +38,7 @@ message Registry {
     repeated OnDemandFeatureView on_demand_feature_views = 8;
     repeated RequestFeatureView request_feature_views = 9;
     repeated FeatureService feature_services = 7;
+    repeated SavedDataset saved_datasets = 11;
     Infra infra = 10;
 
     string registry_schema_version = 3; // to support migrations; incremented when schema is changed
--- a/protos/feast/core/SavedDataset.proto
+++ b/protos/feast/core/SavedDataset.proto
@@ -1,0 +1,76 @@
+//
+// Copyright 2021 The Feast Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+
+syntax = "proto3";
+
+package feast.core;
+option java_package = "feast.proto.core";
+option java_outer_classname = "SavedDatasetProto";
+option go_package = "github.com/feast-dev/feast/sdk/go/protos/feast/core";
+
+import "google/protobuf/timestamp.proto";
+import "feast/core/FeatureViewProjection.proto";
+import "feast/core/DataSource.proto";
+
+message SavedDatasetSpec {
+  // Name of the dataset. Must be unique since it's possible to overwrite dataset by name
+  string name = 1;
+
+  // Name of Feast project that this Dataset belongs to.
+  string project = 2;
+
+  // list of feature references with format "<view name>:<feature name>"
+  repeated string features = 3;
+
+  // entity columns + request columns from all feature views used during retrieval
+  repeated string join_keys = 4;
+
+  // Whether full feature names are used in stored data
+  bool full_feature_names = 5;
+
+  SavedDatasetStorage storage = 6;
+
+  // User defined metadata
+  map<string, string> tags = 7;
+}
+
+message SavedDatasetStorage {
+  oneof kind {
+    DataSource.FileOptions file_storage = 4;
+    DataSource.BigQueryOptions bigquery_storage = 5;
+    DataSource.RedshiftOptions redshift_storage = 6;
+  }
+}
+
+message SavedDatasetMeta {
+  // Time when this saved dataset is created
+  google.protobuf.Timestamp created_timestamp = 1;
+
+  // Time when this saved dataset is last updated
+  google.protobuf.Timestamp last_updated_timestamp = 2;
+
+  // Min timestamp in the dataset (needed for retrieval)
+  google.protobuf.Timestamp min_event_timestamp = 3;
+
+  // Max timestamp in the dataset (needed for retrieval)
+  google.protobuf.Timestamp max_event_timestamp = 4;
+}
+
+message SavedDataset {
+  SavedDatasetSpec spec = 1;
+  SavedDatasetMeta meta = 2;
+}
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -77,6 +77,7 @@ from feast.registry import Registry
 from feast.repo_config import RepoConfig, load_repo_config
 from feast.repo_contents import RepoContents
 from feast.request_feature_view import RequestFeatureView
+from feast.saved_dataset import SavedDataset, SavedDatasetStorage
 from feast.type_map import python_values_to_proto_values
 from feast.usage import log_exceptions, log_exceptions_and_usage, set_usage_attribute
 from feast.value_type import ValueType
@@ -763,6 +764,93 @@ class FeatureStore:
         )
 
         return job
+
+    @log_exceptions_and_usage
+    def create_saved_dataset(
+        self,
+        from_: RetrievalJob,
+        name: str,
+        storage: SavedDatasetStorage,
+        tags: Optional[Dict[str, str]] = None,
+    ) -> SavedDataset:
+        """
+            Execute provided retrieval job and persist its outcome in given storage.
+            Storage type (eg, BigQuery or Redshift) must be the same as globally configured offline store.
+            After data successfully persisted saved dataset object with dataset metadata is committed to the registry.
+            Name for the saved dataset should be unique within project, since it's possible to overwrite previously stored dataset
+            with the same name.
+
+            Returns:
+                SavedDataset object with attached RetrievalJob
+
+            Raises:
+                ValueError if given retrieval job doesn't have metadata
+        """
+        warnings.warn(
+            "Saving dataset is an experimental feature. "
+            "This API is unstable and it could and most probably will be changed in the future. "
+            "We do not guarantee that future changes will maintain backward compatibility.",
+            RuntimeWarning,
+        )
+
+        if not from_.metadata:
+            raise ValueError(
+                "RetrievalJob must contains metadata. "
+                "Use RetrievalJob produced by get_historical_features"
+            )
+
+        dataset = SavedDataset(
+            name=name,
+            features=from_.metadata.features,
+            join_keys=from_.metadata.keys,
+            full_feature_names=from_.full_feature_names,
+            storage=storage,
+            tags=tags,
+        )
+
+        dataset.min_event_timestamp = from_.metadata.min_event_timestamp
+        dataset.max_event_timestamp = from_.metadata.max_event_timestamp
+
+        from_.persist(storage)
+
+        self._registry.apply_saved_dataset(dataset, self.project, commit=True)
+
+        return dataset.with_retrieval_job(
+            self._get_provider().retrieve_saved_dataset(
+                config=self.config, dataset=dataset
+            )
+        )
+
+    @log_exceptions_and_usage
+    def get_saved_dataset(self, name: str) -> SavedDataset:
+        """
+        Find a saved dataset in the registry by provided name and
+        create a retrieval job to pull whole dataset from storage (offline store).
+
+        If dataset couldn't be found by provided name SavedDatasetNotFound exception will be raised.
+
+        Data will be retrieved from globally configured offline store.
+
+        Returns:
+            SavedDataset with RetrievalJob attached
+
+        Raises:
+            SavedDatasetNotFound
+        """
+        warnings.warn(
+            "Retrieving datasets is an experimental feature. "
+            "This API is unstable and it could and most probably will be changed in the future. "
+            "We do not guarantee that future changes will maintain backward compatibility.",
+            RuntimeWarning,
+        )
+
+        dataset = self._registry.get_saved_dataset(name, self.project)
+        provider = self._get_provider()
+
+        retrieval_job = provider.retrieve_saved_dataset(
+            config=self.config, dataset=dataset
+        )
+        return dataset.with_retrieval_job(retrieval_job)
 
     @log_exceptions_and_usage
     def materialize_incremental(
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -30,19 +30,24 @@ from feast.errors import (
 )
 from feast.feature_view import DUMMY_ENTITY_ID, DUMMY_ENTITY_VAL, FeatureView
 from feast.infra.offline_stores import offline_utils
-from feast.infra.offline_stores.offline_store import OfflineStore, RetrievalJob
+from feast.infra.offline_stores.offline_store import (
+    OfflineStore,
+    RetrievalJob,
+    RetrievalMetadata,
+)
 from feast.on_demand_feature_view import OnDemandFeatureView
 from feast.registry import Registry
 from feast.repo_config import FeastConfigBaseModel, RepoConfig
 
+from ...saved_dataset import SavedDatasetStorage
 from ...usage import log_exceptions_and_usage
-from .bigquery_source import BigQuerySource
+from .bigquery_source import BigQuerySource, SavedDatasetBigQueryStorage
 
 try:
     from google.api_core.exceptions import NotFound
     from google.auth.exceptions import DefaultCredentialsError
     from google.cloud import bigquery
-    from google.cloud.bigquery import Client
+    from google.cloud.bigquery import Client, Table
 
 except ImportError as e:
     from feast.errors import FeastExtrasDependencyImportError
@@ -121,6 +126,36 @@ class BigQueryOfflineStore(OfflineStore):
 
     @staticmethod
     @log_exceptions_and_usage(offline_store="bigquery")
+    def pull_all_from_table_or_query(
+        config: RepoConfig,
+        data_source: DataSource,
+        join_key_columns: List[str],
+        feature_name_columns: List[str],
+        event_timestamp_column: str,
+        start_date: datetime,
+        end_date: datetime,
+    ) -> RetrievalJob:
+        assert isinstance(data_source, BigQuerySource)
+        from_expression = data_source.get_table_query_string()
+
+        client = _get_bigquery_client(
+            project=config.offline_store.project_id,
+            location=config.offline_store.location,
+        )
+        field_string = ", ".join(
+            join_key_columns + feature_name_columns + [event_timestamp_column]
+        )
+        query = f"""
+            SELECT {field_string}
+            FROM {from_expression}
+            WHERE {event_timestamp_column} BETWEEN TIMESTAMP('{start_date}') AND TIMESTAMP('{end_date}')
+        """
+        return BigQueryRetrievalJob(
+            query=query, client=client, config=config, full_feature_names=False,
+        )
+
+    @staticmethod
+    @log_exceptions_and_usage(offline_store="bigquery")
     def get_historical_features(
         config: RepoConfig,
         feature_views: List[FeatureView],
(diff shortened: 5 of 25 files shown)


protos/feast/core/Registry.proto
```diff
@@ -37,6 +38,7 @@ message Registry {
     repeated OnDemandFeatureView on_demand_feature_views = 8;
     repeated RequestFeatureView request_feature_views = 9;
     repeated FeatureService feature_services = 7;
+    repeated SavedDataset saved_datasets = 11;
```
woop, line 41: Do you want comments in the PR or in the RFC? Is there anything preventing SavedDataset from just being `Dataset`? Or is the idea that this could eventually be generalized?

protos/feast/core/SavedDataset.proto
```diff
@@ -0,0 +1,73 @@
...
+option java_outer_classname = "SavedDatasetProto";
+option go_package = "github.com/feast-dev/feast/sdk/go/protos/feast/core";
+
+import "google/protobuf/timestamp.proto";
+import "feast/core/FeatureViewProjection.proto";
+import "feast/core/DataSource.proto";
+
+message SavedDatasetSpec {
+  // Name of the dataset. Must be unique since it's possible to overwrite dataset by name
+  string name = 1;
```
woop, line 31: Asking for a name now is fine, but I think it adds friction to the process (vs generating an id by default).

sdk/python/feast/feature_store.py
```diff
@@ -706,6 +709,11 @@ def get_historical_features(
             full_feature_names: A boolean that provides the option to add the feature view prefixes to the feature names,
                 changing them from the format "feature" to "feature_view__feature" (e.g., "daily_transactions" changes to
                 "customer_fv__daily_transactions"). By default, this value is set to False.
+            save_as: An optional configuration that enables persisting retrieved dataset in offline store.
```
woop, line 712: Isn't this API overlapping with [RetrievalJob.to_bigquery()](https://github.com/feast-dev/feast/blob/97dd41e0c0e1a9e8050f3931eb610f59789edeed/sdk/python/feast/infra/offline_stores/bigquery.py#L255) etc? I think it's a bit confusing to me to have two places to tell Feast to save a dataset.

pyalex: Yes, you can call `job.to_bigquery()` or `job.to_redshift()` and this is what saved dataset implementation does under the hood. The difference is that you will need to add additional DataSource in order to read those saved tables back to Feast. And this is what this PR does automatically via `store.get_saved_dataset`

protos/feast/core/Registry.proto
pyalex: I prefer comments here. I don't think it could be generalized, at least I don't see use cases right now. I think logged dataset should be a separate instance.

protos/feast/core/SavedDataset.proto
pyalex: We can generate some id. Just leaving user an option to load dataset by readable name.

pyalex: see `store.get_saved_dataset(name='...')`

sdk/python/feast/feature_store.py
adchia: it also is slightly unclear to me what the exact contract of RetrievalJob is. RetrievalJob otherwise is a deferred fetch that can be triggered on e.g. to_df() or to_arrow(). 

The save_as parameter seems to be ambiguous as to when it persists the data. (the implementation is it happens immediately). If we want the timing to be consistent, then it becomes very similar to the to_bigquery way of doing things.

pyalex: 1. How it works: you're right that the original query will be triggered eagerly and the dataset will be persisted right away after calling `get_historical_features`. The caller will receive a substitution of RetrievalJob that will wrap a much simpler query: `select * from persisted_table` and process (contract) of materializing this query (job) remains the same (will be triggered on `.to_df()` or `.to_arrow()`).
2. Why not postpone persisting until the user will call `to_df()`: current implementation doesn't cache query results, it means that if I'll call `.to_df()` 3 times - 3 jobs to BigQuery or Redshift would be created. In case we'll put persisting on that stage - we'll have to overwrite the destination table over and over, which I don't think is good contract. Ideally, we should write once and fail if the table already exists and not risk deleting data created by the user and not Feast.

protos/feast/core/SavedDataset.proto
```diff
@@ -0,0 +1,73 @@
...
+option java_package = "feast.proto.core";
+option java_outer_classname = "SavedDatasetProto";
+option go_package = "github.com/feast-dev/feast/sdk/go/protos/feast/core";
+
+import "google/protobuf/timestamp.proto";
+import "feast/core/FeatureViewProjection.proto";
+import "feast/core/DataSource.proto";
+
+message SavedDatasetSpec {
+  // Name of the dataset. Must be unique since it's possible to overwrite dataset by name
```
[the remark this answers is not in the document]
pyalex: Correct

protos/feast/core/SavedDataset.proto
```diff
@@ -0,0 +1,73 @@
...
+  string name = 1;
+
+  // Name of Feast project that this Dataset belongs to.
+  string project = 2;
+
+  // list of feature references with format "<view name>:<feature name>"
+  repeated string features = 3;
```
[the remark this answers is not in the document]
pyalex: 1. FeatureViewProjection seems to be overkill here because all mappings have already happened during historical retrieval. So this dataset is already a projection. What we need to read it is just a list of columns.
2. Although it seems like a reasonable idea to replace the list of features with a feature service name there's one concern: feature service can be updated and the dataset is kinda immutable. So we'll need to infer the schema for dataset columns, instead of using this property and thus we can omit it completely.

MattDelac:
Thanks @pyalex for putting that together !

> `get_saved_dataset(name)` returns `RetrievalJob` (similarly to `get_historical_features`), which can be materialized by calling `.to_df()` or `.to_arrow()`.

What is weird in this behaviour is that `store.get_historical_retrieval()` will sometimes return a `RetrievalJob` object and sometimes will make modifications **inplace**.

Should we have `SavedDatasetOptions()` supported at the historical retrieval methods instead (like `to_df()` ) ? What would be the pros/cons of doing so in your opinion ?

docs/getting-started/concepts/dataset.md
```diff
@@ -0,0 +1,42 @@
+# Dataset
+
+Feast Datasets allow for conveniently saving dataframes that include both features and entities to be subsequently used for data analysis and model training.
```
adchia, line 3: we should be consistent in casing with Dataset vs dataset here. probably just lower case throughout unless its code to be consistent?

docs/getting-started/concepts/dataset.md
```diff
@@ -0,0 +1,42 @@
...
+[Data Quality Monitoring](https://docs.google.com/document/d/110F72d4NTv80p35wDSONxhhPBqWRwbZXG4f9mNEMd98) was the primary motivation for creating Dataset concept.
+
+Dataset's metadata is stored in the Feast registry and dataframe with features, entities, additional input keys and timestamp is kept in the [offline store](../architecture-and-components/offline-store.md).
```
adchia, line 6: maybe instead: Dataset metadata is stored in the Feast registry and the raw data is stored in the offline store ?

docs/getting-started/concepts/dataset.md
```diff
@@ -0,0 +1,42 @@
...
+
+Dataset can be created from:
+1. Results of historical retrieval
+2. Logging request (including input for [on demand transformation](../../reference/alpha-on-demand-feature-view.md)) and response during feature serving
+3. Logging features during writing to online store (from batch source or stream)
```
adchia, line 11: this isn't supported yet right? Maybe we have a note that these are planned?

protos/feast/core/SavedDataset.proto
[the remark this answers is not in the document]
adchia: SGTM. Also, seems like we also might want to list the schema here too so it's easier to do type validation down the road? (and maybe even allow for string tags like feature views so users can add additional metadata for a dataset... like author of a dataset)

protos/feast/core/SavedDataset.proto
```diff
@@ -0,0 +1,73 @@
...
+message SavedDatasetStorage {
+  oneof kind {
+    DataSource.FileOptions file_storage = 4;
+    DataSource.BigQueryOptions bigquery_storage = 5;
+    DataSource.RedshiftOptions redshift_storage = 6;
+  }
+}
+
+message SavedDatasetMeta {
+  // Time where this Saved Dataset is created
```
adchia, line 57: Time where -> Time when for here and below. Also have consistent capitalization in comments? (e.g. always start with a capital)

sdk/python/feast/infra/offline_stores/offline_store.py
```diff
@@ -73,6 +87,24 @@ def to_arrow(self) -> pyarrow.Table:
             )
         return pyarrow.Table.from_pandas(features_df)
 
+    @abstractmethod
+    def persist(self, storage: SavedDatasetStorage) -> "RetrievalJob":
```
adchia, line 91: why do you need to do a forward reference type check? doesn't seem like there's a circular dependency?

pyalex: the idea is that persist returns new RetrievalJob, it could be modified self or completely new job (that would wrap new query). And the result of persist is what returned from `get_historical_features`

adchia:
> Thanks @pyalex for putting that together !
> 
> > `get_saved_dataset(name)` returns `RetrievalJob` (similarly to `get_historical_features`), which can be materialized by calling `.to_df()` or `.to_arrow()`.
> 
> What is weird in this behaviour is that `store.get_historical_retrieval()` will sometimes return a `RetrievalJob` object and sometimes will make modifications **inplace**.
> 
> Should we have `SavedDatasetOptions()` supported at the historical retrieval methods instead (like `to_df()` ) ? What would be the pros/cons of doing so in your opinion ?

I think as per our discussion, we might want to have get_historical_retrieval always do some in place modification so it becomes more consistent. Though there are certainly concerns around how long this dataset should persist. My guess would be that we by default store data in temporary tables unless users explicitly specify a SavedDataset.

pyalex:
Hi @MattDelac, thanks for your comment!

The main reason not to defer dataset persisting until RetrievalJob is used - is that besides storing dataframe we also create metadata object (see SavedDataset proto) and this object is being written to the registry. I don't think it's a good idea to modify the registry from RetrievalJob. That's why this PR proposes eager execution:
1. Original query is being triggered synchronously during `get_historical_features` (only if `save_as` parameter is provided)
2. Dataframe is stored to DWH
3. SavedDataset object is written to the registry

Does it make sense?

protos/feast/core/SavedDataset.proto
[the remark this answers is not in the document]
pyalex: Added tags to both SavedDataset proto & options

adchia:
@mavysavydav fyi, we'll move forward with not adding any breaking changes while we test DQM. We can discuss if there are more elegant solutions to handle `to_sql` better later on!

MattDelac:
> At the end of the day - this API (I mean with `save_as` parameter) is experimental and will most probably change. Also, the important thing that I want to stress once more is that the previous behavior (with `save_as` parameter not passed) remains unchanged. So if you won't use this functionality - you won't feel a difference

Sounds good then. I am fine with experimenting to gain context and experience about what should be built 👍 
I will try to review the PR. Can you make sure that if `save_as` is used, we show am ExperimentalWarning to our users ? We need to make it clear that this is not stable and that is can change anytime

pyalex:
@MattDelac , yes there's a warning https://github.com/feast-dev/feast/pull/2197/files#diff-ff4e2a65c8c29b7971edc77a61a7e970d410d77d9f173f72714d1bb49e2509a2R804

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -95,7 +100,9 @@ def pull_latest_from_table_or_query(
         if created_timestamp_column:
             timestamps.append(created_timestamp_column)
         timestamp_desc_string = " DESC, ".join(timestamps) + " DESC"
-        field_string = ", ".join(join_key_columns + feature_name_columns + timestamps)
+        field_string = ", ".join(
+            set(join_key_columns + feature_name_columns + timestamps)
```
adchia, line 104: is there a reason you made this a set? should probably call this out in the description of the PR too since it can cause a behavior change

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -310,6 +325,25 @@ def _execute_query(
         block_until_done(client=self.client, bq_job=bq_job, timeout=timeout)
         return bq_job
 
+    def persist(self, storage: SavedDatasetStorage) -> "BigQueryRetrievalJob":
```
adchia, line 328: Do we want this to be a publically facing method? Seems confusing to a user why there'd be

1. this persist method, which happens lazily. 
2. `to_bigquery`, which exists right next to this method and also saves.

I think the return type can be confusing too since .persist has indeed executed a query & persisted, but we return a `RetrievalJob`. I might as a user expect to need to call `RetrievalJob#to_bigquery` again, which actually will persist to a different location.

Two options I think could make sense:
- Have this be an internal only method by name so people don't try to call it
- Expose a separate way to instantiate a RetrievalJob (like RetrievalJob.from_saved_dataset which executes this wrapped query) so you can return it in `get_historical_features`

sdk/python/feast/infra/offline_stores/file.py
```diff
@@ -66,6 +74,35 @@ def _to_arrow_internal(self):
...
+
+        filesystem, path = FileSource.create_filesystem_and_path(
+            storage.file_options.file_url, storage.file_options.s3_endpoint_override,
+        )
+
+        if path.endswith(".parquet"):
+            pyarrow.parquet.write_table(
+                self._to_arrow_internal(), where=path, filesystem=filesystem
+            )
+        else:
```
adchia, line 88: probably makes sense to check it's not some other file format like .csv which we don't support

sdk/python/feast/infra/offline_stores/offline_store.py
adchia: makes sense. See my other comment for further thoughts.

sdk/python/feast/registry.py
```diff
@@ -695,6 +697,80 @@ def delete_entity(self, name: str, project: str, commit: bool = True):
 
         raise EntityNotFoundException(name, project)
 
+    def apply_saved_dataset(
```
adchia, line 700: not a blocking comment. More of an FYI

fyi @felixwang9817 since you are adding a new concept to the registry. Don't think `feast plan` could cover this since this isn't introduced at `feast apply` time. 

Also doesn't really make sense to treat this as a real FCO since the schema is a snapshot of how feature services / feature views resolve at a point in time. Doubt we'd be able to do this properly unless we have some form of stricter versioning.

adchia (commented):
How does a user "delete" a SavedDataset? At least remove references to it from the registry? As well as what happens if the underlying dataset is removed? (which isn't too crazy to imagine e.g. if user generated data must be tied to a TTL for GDPR reasons)

docs/getting-started/concepts/dataset.md
```diff
@@ -0,0 +1,43 @@
...
+        name='my_training_dataset',
+        storage=SavedDatasetBigQueryStorage(table_ref='<gcp-project>.<gcp-dataset>.my_training_dataset'),
+        tags={'author': 'oleksii'}
+    )
+)
+```
+
+Saved dataset can be later retrieved using `get_saved_dataset` method:
+```python
+retrieval_job = store.get_saved_dataset('my_training_dataset')
```
MattDelac, line 41: Is this when the saved dataset is computed ?

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -310,6 +325,25 @@ def _execute_query(
...
+        assert isinstance(storage, SavedDatasetBigQueryStorage)
```
MattDelac, line 329: Is the BigQuery offline store can only persist the SavedDataset into BigQuery, why making it possible to pass any `SavedDatasetStorage` and not infer it ?

MattDelac (commented):
My :2-cents:

sdk/python/feast/infra/offline_stores/bigquery.py
pyalex: `persist` doesn't happen lazily. It's generic equivalent of methods `to_bigquery` or `to_redshift` which exist only in specific implementation classes. `predict` is part of abstract API (`RetrievalJob`) and aimed to unify these methods. Under the hood, it's calling `to_bigquery` or `to_redshift`. 

The feature that is absent in previous methods is that `persist` returns RetrievalJob, which supposedly retrieves from a persisted table. 

Why it's public? As we discussed on call with Shopify - it could be useful if the user doesn't want `get_historical_features` to run query eagerly. In this case, user would not pass `save_as` parameter but call `persist` themselves:
```
job = fs.get_historical_feature(...)
persisted = job.persist(storage=SavedDatasetBigQueryStorage(...))
persisted.to_df()
```

Regarding cyclical calls, I think that we could deprecate specific methods like `.to_bigquery()` and then it should be less confusing.

> Expose a separate way to instantiate a RetrievalJob (like RetrievalJob.from_saved_dataset which executes this wrapped query) so you can return it in get_historical_features

This option makes sense to me. Please let me know what you think is better @adchia

sdk/python/feast/infra/offline_stores/bigquery.py
pyalex: I was trying to reuse `pull_latest_from_table_or_query` because it's already implemented in all offline stores including extensions. But those implementations also include deduplication, which shouldn't happen for retrieving a saved dataset.
Now I see that it probably wasn't a good idea. I will use a separate method.

pyalex:
> How does a user "delete" a SavedDataset? At least remove references to it from the registry? As well as what happens if the underlying dataset is removed? (which isn't too crazy to imagine e.g. if user generated data must be tied to a TTL for GDPR reasons)

As far as I know, only BigQuery supports TTL for tables, so we'll probably need to implement `FeatureStore` level method (`delete_saved_dataset`), which will include underlying storage cleanup. But that probably is not in scope of this PR.

docs/getting-started/concepts/dataset.md
```diff
@@ -0,0 +1,43 @@
...
+from feast.infra.offline_stores.bigquery_source import SavedDatasetBigQueryStorage
+
+store = FeatureStore()
+
+store.get_historical_features(
+    features=["driver:avg_trip"],
+    entity_df=...,
+    save_as=SavedDatasetOptions(
+        name='my_training_dataset',
+        storage=SavedDatasetBigQueryStorage(table_ref='<gcp-project>.<gcp-dataset>.my_training_dataset'),
```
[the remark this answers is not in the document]
pyalex: You won't be able to pass here different storage other than the one you have configured as a global offline store. There will be an exception if your features are being pulled from BigQuery and you will try to store them into Redshift.
This is just a generic API to support typed configuration for all kinds of stores. I couldn't find a better way (at least in Python type system is not so strong). If you know how to do it better please let me know.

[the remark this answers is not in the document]
pyalex: We can infer the type of storage, but each type will still require its own parameters (eg, table_ref for BQ, path for File) and we need those parameters provided by user anyway. So those classes (`SavedDatasetBigQueryStorage`, `SavedDatasetRedshiftStorage`, etc) are useful at least for parameters type checking.

docs/getting-started/concepts/dataset.md
pyalex: If by compute you mean materialized, then no. The saved dataset is computed eagerly on creation. For example, when you call `get_historical_features` with the `save_as` parameter saved dataset would be already materialized & stored to offline store & registry by the time this function returns.

sdk/python/feast/infra/offline_stores/bigquery.py
pyalex: As described above, that's merely for type checking: good for verification on development and could be helpful when using Feast in some kind of IDE.

docs/getting-started/concepts/dataset.md
```diff
@@ -0,0 +1,43 @@
...
+
+```python
+
+from feast import FeatureStore
+from feast.saved_dataset import SavedDatasetOptions
+from feast.infra.offline_stores.bigquery_source import SavedDatasetBigQueryStorage
+
+store = FeatureStore()
+
+store.get_historical_features(
```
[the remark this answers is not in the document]
pyalex: Yep, this could work. But what are benefits of this approach? 
1. Do you want to reuse `offline_job`? But in that case, `historical` query can be run several times, which is additional costs. Whereas in the current proposed version saved dataset plays the role of cache. Eg,
```
result = store.get_historical_features(..., save_as='...')
result.to_df()
result.to_df() # both times reads from persisted table, not running the original query
```
2. Will `create_saved_dataset` run `offline_job` eagerly & what is the purpose of `SavedDatasetJob`?

sdk/python/feast/infra/offline_stores/file.py
pyalex: In this `else` block we treat the destination as a directory, so it could be named even as `dir.csv` -> `dir.csv/` folder would be created.

sdk/python/feast/infra/offline_stores/bigquery.py
pyalex: If I understand correctly, you suggesting just plain dict:
```
fs.get_historical_features(..., save_as={'storage': {'table_ref': '...'}})
```
vs proposed API
```
fs.get_historical_features(..., save_as=SavedDatasetOpts(storage=SavedDatasetBigQueryStorage(table_ref=...))
```
just to free you from remembering those dict keys.

sdk/python/feast/infra/offline_stores/bigquery.py
pyalex: UPD. `persist` signature updated. Now it doesn't return `RetrievalJob` to avoid duplication of responsibilities. Read of saved dataset moved to Provider.read_saved_dataset, which uses OfflineSource.pull_all_from_table_or_query.

So now `.persist()` is just a universal replacement for `.to_bigquery()` and `.to_redshift()`.

sdk/python/feast/infra/offline_stores/file.py
pyalex: Added comment line

docs/getting-started/concepts/dataset.md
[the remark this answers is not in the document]
MattDelac: > But what are benefits of this approach?

Mostly consistency: `store.get_historical_features()` always return an `OfflineRetrievalJob` object

> Do you want to reuse offline_job?

I don't want to have strong assumptions on that. What I know is that the BQ Retrieval job has a `dry-run` option. So I would not be surprised if some users use it multiple times.

> query can be run several times, which is additional costs

That's on the user to decide if this additional cost is useful or not IMO

> Whereas in the current proposed version saved dataset plays the role of cache

I get that and this is great 👍 

> what is the purpose of SavedDatasetJob?

Now I understand it better, I think that it's not needed. Here is my updated example

```python
store = FeatureStore()
offline_job: OfflineRetrievalJob = store.get_historical_features(
    features=["driver:avg_trip"],
    entity_df=...,
)

retrieval_job: OfflineRetrievalJob = store.create_saved_dataset( # That operation executes the RetrievalJob
    name='my_training_dataset',
    storage=SavedDatasetBigQueryStorage(table_ref='<gcp-project>.<gcp-dataset>.my_training_dataset'),
    retrieval_job=offline_job,
    tags={'author': 'oleksii'},
)

retrieval_job.to_df()
```

As I said, I am thinking out-loud

sdk/python/feast/infra/offline_stores/bigquery.py
MattDelac: > If I understand correctly, you suggesting just plain dict

Or sending parameters right away. Something like
```python
get_historical_features(..., save_as=SavedDataset(table_ref="...")) # Example 1
get_historical_features(..., save_as=SavedDataset(table_ref="...", table_ttl=timedelta(days=30))) # Example 2
get_historical_features(..., save_as=SavedDataset(file_path="...")) # Example 3

# Under the hood
@dataclass
class SavedDataset:
    table_ref: Optional[str] = None
    table_ttl: Optional[timedelta] = None
    file_path: Optional[str] = None

def get_historical_features(..., save_as: SavedDatasetOpts):
    if isinstance(offline_store, BigQuery):
        return SavedDatasetBigQueryStorage(table_ref=save_as.table_ref, table_ttl=save_as.table_ttl)
```

Unsure if that makes it more confusing **or** easier to use for our users TBH 🤔

docs/getting-started/concepts/dataset.md
MattDelac: > For example, when you call get_historical_features with the save_as parameter saved dataset would be already materialized

That's where it could be confusing because `fs.get_historical_features()` does not materialize anything normally but it woudl do it when `save_as` is passed. Two completely different behaviour with the same method.

sdk/python/feast/infra/offline_stores/bigquery.py
pyalex: I think subclasses is a better approach:
1. Easier to extend (add new kind of storage)
2. You can have mandatory & optional parameters, whereas with a single class they all are optional.

Plus I believe this approach (more strict types) is used in various places in Feast, so keeping things consistent is good.

docs/getting-started/concepts/dataset.md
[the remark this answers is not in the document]
pyalex: API is updated.

sdk/python/feast/infra/offline_stores/bigquery.py
MattDelac: > I think subclasses is a better approach

Sounds good to me 👍

pyalex:
@adchia @MattDelac I have refactored some APIs. Please give it one more look.
1. Saved dataset creation moved to a dedicated method, which expects retrieval job and calls `.persist()` on it (among other things). See docs or PR description for details.
2. Both `create_saved_dataset` and `get_saved_dataset` return SavedDataset object instead of RetrievalJob (however the job is still inside). This is for better separation between these methods and for example `get_historical_features`, but also I need this for validation API (my next PR).
3. `.persist()` not returning anything anymore. It simplifies the code in general and now its only purpose is to be unifying replacement for `.to_bigquery()` and `.to_redshift()`, etc.

adchia (approved):
/lgtm

MattDelac:
> @adchia @MattDelac I have refactored some APIs. Please give it one more look.
> 
> 1. Saved dataset creation moved to a dedicated method, which expects retrieval job and calls `.persist()` on it (among other things). See docs or PR description for details.
> 2. Both `create_saved_dataset` and `get_saved_dataset` return SavedDataset object instead of RetrievalJob (however the job is still inside). This is for better separation between these methods and for example `get_historical_features`, but also I need this for validation API (my next PR).
> 3. `.persist()` not returning anything anymore. It simplifies the code in general and now its only purpose is to be unifying replacement for `.to_bigquery()` and `.to_redshift()`, etc.

This is awesome 🚀 🚀 🚀 🚀 

Thanks for putting the extra effort here 🙏

PR #2198, opened 2022-01-06
Fix grpc requirements in python

adchia:
Using the make install-python fails because `grpc_tools.protoc` can't be found.

This PR explicitly adds that dependency (consistent with guidelines on https://grpc.io/docs/languages/python/quickstart/)

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/requirements/py3.7-ci-requirements.txt
+++ b/sdk/python/requirements/py3.7-ci-requirements.txt
@@ -200,7 +200,6 @@ googleapis-common-protos==1.52.0
     #   tensorflow-metadata
 grpcio==1.42.0
     # via
-    #   feast (setup.py)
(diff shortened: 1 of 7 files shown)


adchia:
Not sure what we exactly want install-python to do. For a user to push commits, they'll need the ci dependencies too, so if this is just to simulate installing a version at HEAD, then we could remove "	cd sdk/python && python setup.py develop" from the install-python rule too

pyalex:
This requriment is already present her: https://github.com/feast-dev/feast/blob/master/sdk/python/setup.py#L234
The problem with protoc is that it needs to be installed a little earlier than everything else because it's being used during installation: https://github.com/feast-dev/feast/blob/master/sdk/python/setup.py#L191

PR #2199, opened 2022-01-06
Refactor all importer logic to belong in feast.importer

felixwang9817:

**What this PR does / why we need it**: This PR takes all importer logic and refactors it so that all importer logic belongs to a single method, `get_class_from_module`, in `feast.importer`.

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/importer.py
+++ b/sdk/python/feast/importer.py
@@ -1,5 +1,4 @@
 import importlib
-from typing import Optional
 
 from feast.errors import (
     FeastClassImportError,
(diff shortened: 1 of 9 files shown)


sdk/python/feast/importer.py
```diff
@@ -1,28 +1,47 @@
...
+        module_name: The name of the module.
+        class_name: The name of the class.
+        class_type: Optional suffix of the class.
+
+    Raises:
+        FeastClassInvalidName: If the class name does not end with the specified suffix.
+        FeastModuleImportError: If the module cannot be imported.
+        FeastClassImportError: If the class cannot be imported.
+    """
+    if class_type and not class_name.endswith(class_type):
```
pyalex, line 25: alternatively, we can make `class_type` something like base class expectations, eg we importing class FileStorage and expect that it's subtype of Storage.
Check base classes of imported type instead of naming could look like this
```
if not any(base_class.__name__ in class_type for base_class _class.__bases__):
    raise 
```

pyalex (approved):

PR #2200, opened 2022-01-07
Compare only specs in integration tests

felixwang9817:

**What this PR does / why we need it**: #1952 was recently merged and is currently breaking integration tests on master. Specifically, by setting `created_timestamp` for feature views, the `test_universal_cli` test breaks because the second `feast apply` changes FV metadata, causing the registry to be changed. This PR ensures that when registry contents are compared after the first and second calls to `feast plan`, only specs of the FCOs are compared, and metadata is excluded. This allows the test to pass.

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/registry.py
+++ b/sdk/python/feast/registry.py
@@ -708,7 +708,7 @@ class Registry:
         """Tears down (removes) the registry."""
         self._registry_store.teardown()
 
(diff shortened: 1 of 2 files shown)


achals (changes requested):
Can we update the test to only compare the specs instead? Would be easier

achals (approved):
/lgtm

PR #2202, opened 2022-01-07
Python FeatureServer optimization

### LINKED ISSUE #2201: Python FeatureServer inefficiently converts entity rows
Problem:
It seems silly for the Python FeatureServer to convert EntityValues to Python native types only to then immediately convert them back to Protobuf Values. This can only increase latency by requiring a bunch of memory assignment and copying.

https://github.com/feast-dev/feast/blob/14048fd535f982eed7a3ba3af7d93557335e325a/sdk/python/feast/feature_server.py#L44-L50

Solution:
One option would be to refactor `get_online_features()` to convert `entity_rows` to  a `Dict[str, List[Value]` and then call a new wrapper method, `_get_online_features()`, `entity_rows` as a `Dict[str, List[Value]`. This would allow the FeatureServer to directly call `_get_online_features()` without converting the EntityValues to Python native types. 

This solution would align nicely with the refactoring done in #2186.

judahrand:
Previously, the Python FeatureServer was doing pointless conversion of Protobuf entity values to Python native types only to convert them back again.

This PR is stacked on top of #2186 so only the last commit is relevant to the issue.

Fixes #2201

```release-note
None
```


achals:
@judahrand is this ready for review? Now that #2186 has been merged.

judahrand:
> @judahrand is this ready for review? Now that #2186 has been merged.

I'm struggling to work out why the FeatureServer tests are failing and don't have a good way to debug since the FeatureServer is only supported by the AWS test 🤔

sdk/python/feast/feature_store.py
```diff
@@ -1058,6 +1081,22 @@ def get_online_features(
...
+            entity_proto_values: Dict[str, List[Value]] = {
+                k: python_values_to_proto_values(
+                    v, entity_type_map.get(k, ValueType.UNKNOWN)
+                )
+                for k, v in entity_values.items()
+            }
+        else:
+            entity_proto_values = entity_values
+
+        num_rows = _validate_entity_values(entity_proto_values)
```
adchia, line 1099: From patching this PR in and doing some changes (https://github.com/feast-dev/feast/pull/2217) looks like the root cause is here. Basically the entity values are `RepeatedValues` (i.e. a list of entity ids for batch retrieval) and those don't work with len() which is used in `_validate_entity_values`

judahrand: I think I've handled this now - though I wonder whether the conversion to a `List` should be done in the FeatureServer on in the FeatureStore method?

sdk/python/feast/feature_store.py
```diff
@@ -1314,14 +1324,19 @@ def _set_table_entity_keys(
...
-                result_row.statuses[name] = FieldStatus.PRESENT
+        for feature_name, feature_values in data.items():
+
+            online_features_response.metadata.feature_names.val.append(feature_name)
+
+            for row_idx, proto_value in enumerate(feature_values):
+                result_row = online_features_response.results[row_idx]
+                result_row.values.append(proto_value)
+                result_row.statuses.append(FieldStatus.PRESENT)
+                result_row.event_timestamps.append(Timestamp())
```
adchia, line 1339: nit: maybe make the timestampat the beginning of the method + reuse?

sdk/python/feast/feature_store.py
adchia: Yeah especially since there's no other real reason to have them in a proto format. Probably ok for now though

sdk/python/feast/feature_store.py
```diff
@@ -1078,15 +1079,15 @@ def get_online_features(
...
+            entity_values=dict(columnar),
             full_feature_names=full_feature_names,
             native_entity_values=True,
         )
 
     def _get_online_features(
         self,
         features: Union[List[str], FeatureService],
-        entity_values: Dict[str, List[Any]],
+        entity_values: Dict[str, Union[Sequence[Any], Sequence[Value], RepeatedValue]],
```
adchia, line 1090: any reason why you have Sequence[Value] sinec you already have Any?

sdk/python/feast/feature_store.py
```diff
@@ -1078,15 +1079,15 @@ def get_online_features(
 
         return self._get_online_features(
             features=features,
-            entity_values=columnar,
+            entity_values=dict(columnar),
```
adchia, line 1082: is this intentional?

adchia (approved):

sdk/python/feast/feature_store.py
judahrand: It was deliberate as I thought it was more communicative of what is actually expected. My interpretation of our use of `Any` in many places is 'any native Python type' rather than actually any class. So in my mind `Sequence[Any]` is different to `Sequence[Value]`. And then `RepeatedValue` is another separate case. 

Flexible on that though.

sdk/python/feast/feature_store.py
judahrand: Yeah - two reasons:

1. I figured that passing it along as a `defaultdict` was just asking for an unhelpful error message
2. I think that `defaultdict` might not technically be a `Dict`? But might be wrong on that.

sdk/python/feast/feature_store.py
adchia: Fair enough. Mind updating the docstrings for the get_online_features method then to reflect this?

sdk/python/feast/feature_store.py
adchia: any reason why we wouldnt just use a normal dict from the getgo then?

adchia (approved):

sdk/python/feast/feature_store.py
judahrand: `get_online_features` hasn't really changed at all. I believe we'd still expect the user to pass native Python types. However, `get_online_features` now wraps `_get_online_features` which does expect to have to be able to handle Protobuf types. 

Do you still think the Docstring needs updating? Nothing has changed for the user.

sdk/python/feast/feature_store.py
judahrand: Laziness 😛

sdk/python/feast/feature_store.py
adchia: ah nvm. I had quickly skimmed and thought you also allowed users to technically pass values in the regular API. LG!

adchia (approved):
/lgtm

PR #2203, opened 2022-01-07
Add on demand feature views deletion

corentinmarek:
The deletion of on demand feature views is broken as it is not covered in `registry.delete_feature_view`. 
This is a small fix to cover deletion of odfv the same way it is done for standard or request feature views. 

```release-note
NONE
```


adchia (approved):
/lgtm

PR #2210, opened 2022-01-08
Updates to click==8.*

diogommartins:
Updates click from 7.x to 8.x - https://click.palletsprojects.com/en/8.0.x/upgrading/

```release-note
NONE
```

### DIFF
--- a/sdk/python/requirements/py3.7-ci-requirements.txt
+++ b/sdk/python/requirements/py3.7-ci-requirements.txt
@@ -85,7 +85,7 @@ charset-normalizer==2.0.8
     # via
     #   aiohttp
     #   requests
-click==7.1.2
+click==8.0.3
     # via
     #   black
     #   feast (setup.py)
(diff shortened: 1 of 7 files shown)


adchia:
Thanks for the PR! Mind signing the commit to pass the DCO check?

diogommartins:
> Thanks for the PR! Mind signing the commit to pass the DCO check?

Done it!

adchia:
will also need to update the pip-tools requirements files: https://github.com/feast-dev/feast/tree/master/sdk/python/requirements

diogommartins:
> will also need to update the pip-tools requirements files: https://github.com/feast-dev/feast/tree/master/sdk/python/requirements

will do

woop:
Is there any reason to only support a single major version?

diogommartins:
> Is there any reason to only support a single major version?

Since there isn't an interface change from 7.x and 8.x, I don't seed a reason not to support both. I fixed the version to 8.x to follow the previous design decision 🤷‍♂️

adchia:
this seems fine to me. we can keep this unless someone else feels strongly.

/lgtm

adchia (approved):

PR #2211, opened 2022-01-10
Modify feature_store.plan to produce an InfraDiff

felixwang9817:
**What this PR does / why we need it**: This PR modifies `feature_store.plan` to produce an `InfraDiff` object, which will be used to modify infrastructure as part of `feast plan`.

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/diff/FcoDiff.py
+++ b/sdk/python/feast/diff/FcoDiff.py
@@ -18,8 +18,8 @@ from feast.protos.feast.core.RequestFeatureView_pb2 import (
     RequestFeatureView as RequestFeatureViewProto,
 )
 
-U = TypeVar(
-    "U",
+FcoProto = TypeVar(
+    "FcoProto",
     EntityProto,
     FeatureViewProto,
     FeatureServiceProto,
@@ -30,11 +30,11 @@ U = TypeVar(
 
 
 @dataclass
-class FcoDiff(Generic[U]):
+class FcoDiff(Generic[FcoProto]):
     name: str
     fco_type: str
-    current_fco: U
-    new_fco: U
+    current_fco: FcoProto
+    new_fco: FcoProto
     fco_property_diffs: List[PropertyDiff]
     transition_type: TransitionType
 
@@ -50,12 +50,12 @@ class RegistryDiff:
         self.fco_diffs.append(fco_diff)
 
 
-T = TypeVar("T", Entity, BaseFeatureView, FeatureService)
+Fco = TypeVar("Fco", Entity, BaseFeatureView, FeatureService)
 
 
 def tag_objects_for_keep_delete_add(
-    existing_objs: Iterable[T], desired_objs: Iterable[T]
-) -> Tuple[Set[T], Set[T], Set[T]]:
+    existing_objs: Iterable[Fco], desired_objs: Iterable[Fco]
+) -> Tuple[Set[Fco], Set[Fco], Set[Fco]]:
     existing_obj_names = {e.name for e in existing_objs}
     desired_obj_names = {e.name for e in desired_objs}
 
@@ -67,8 +67,8 @@ def tag_objects_for_keep_delete_add(
 
 
 def tag_proto_objects_for_keep_delete_add(
-    existing_objs: Iterable[U], desired_objs: Iterable[U]
-) -> Tuple[Iterable[U], Iterable[U], Iterable[U]]:
+    existing_objs: Iterable[FcoProto], desired_objs: Iterable[FcoProto]
+) -> Tuple[Iterable[FcoProto], Iterable[FcoProto], Iterable[FcoProto]]:
     existing_obj_names = {e.spec.name for e in existing_objs}
     desired_obj_names = {e.spec.name for e in desired_objs}
 
@@ -82,7 +82,7 @@ def tag_proto_objects_for_keep_delete_add(
 FIELDS_TO_IGNORE = {"project"}
 
 
-def diff_between(current: U, new: U, object_type: str) -> FcoDiff:
+def diff_between(current: FcoProto, new: FcoProto, object_type: str) -> FcoDiff:
     assert current.DESCRIPTOR.full_name == new.DESCRIPTOR.full_name
     property_diffs = []
     transition: TransitionType = TransitionType.UNCHANGED
--- a/sdk/python/feast/diff/FcoDiff.py
+++ b/sdk/python/feast/diff/FcoDiff.py
@@ -1,20 +1,40 @@
 from dataclasses import dataclass
-from typing import Any, Iterable, List, Set, Tuple, TypeVar
+from typing import Generic, Iterable, List, Set, Tuple, TypeVar
 
 from feast.base_feature_view import BaseFeatureView
 from feast.diff.property_diff import PropertyDiff, TransitionType
 from feast.entity import Entity
 from feast.feature_service import FeatureService
 from feast.protos.feast.core.Entity_pb2 import Entity as EntityProto
+from feast.protos.feast.core.FeatureService_pb2 import (
+    FeatureService as FeatureServiceProto,
+)
+from feast.protos.feast.core.FeatureTable_pb2 import FeatureTable as FeatureTableProto
 from feast.protos.feast.core.FeatureView_pb2 import FeatureView as FeatureViewProto
+from feast.protos.feast.core.OnDemandFeatureView_pb2 import (
+    OnDemandFeatureView as OnDemandFeatureViewProto,
+)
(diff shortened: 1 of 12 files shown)


sdk/python/feast/diff/FcoDiff.py
```diff
@@ -1,20 +1,40 @@
...
+from feast.protos.feast.core.FeatureTable_pb2 import FeatureTable as FeatureTableProto
 from feast.protos.feast.core.FeatureView_pb2 import FeatureView as FeatureViewProto
+from feast.protos.feast.core.OnDemandFeatureView_pb2 import (
+    OnDemandFeatureView as OnDemandFeatureViewProto,
+)
+from feast.protos.feast.core.RequestFeatureView_pb2 import (
+    RequestFeatureView as RequestFeatureViewProto,
+)
+
+U = TypeVar(
```
adchia, line 21: maybe instead of "U", it could be something more descriptive like Fco?

sdk/python/feast/diff/infra_diff.py
```diff
@@ -17,13 +17,15 @@
 from feast.protos.feast.core.InfraObject_pb2 import Infra as InfraProto
 from feast.protos.feast.core.SqliteTable_pb2 import SqliteTable as SqliteTableProto
 
+U = TypeVar("U", DatastoreTableProto, DynamoDBTableProto, SqliteTableProto)
```
adchia, line 20: same comment here. maybe like InfraObject?

sdk/python/feast/diff/infra_diff.py
```diff
@@ -36,15 +38,29 @@ def __init__(self):
...
+        """Apply the infrastructure changes specified in this object."""
+        for infra_object_diff in self.infra_object_diffs:
+            if infra_object_diff.transition_type in [
+                TransitionType.DELETE,
+                TransitionType.UPDATE,
+            ]:
+                infra_object = InfraObject.from_proto(
+                    infra_object_diff.current_infra_object
+                )
+                infra_object.teardown()
```
adchia, line 50: why does an "UPDATE" cause a teardown?

sdk/python/feast/infra/online_stores/datastore.py
```diff
@@ -387,6 +387,19 @@ def from_infra_object_proto(infra_object_proto: InfraObjectProto) -> Any:
 
         return datastore_table
 
+    @staticmethod
+    def from_proto(datastore_table_proto: DatastoreTableProto) -> Any:
+        datastore_table = DatastoreTable(
+            project=datastore_table_proto.project, name=datastore_table_proto.name,
+        )
+
+        if datastore_table_proto.HasField("project_id"):
```
adchia, line 396: I think this will always be true since we're using proto3 & it auto populates these with the default values (https://developers.google.com/protocol-buffers/docs/reference/python-generated#singular-fields-proto3)

sdk/python/feast/infra/infra_object.py
```diff
@@ -49,15 +57,38 @@ def from_infra_object_proto(infra_object_proto: InfraObjectProto) -> Any:
...
+        """
+        Converts a protobuf representation of a subclass to an object of that subclass.
+
+        Args:
+            infra_object_proto: A protobuf representation of an InfraObject.
+
+        Raises:
+            FeastInvalidInfraObjectType: The type of InfraObject could not be identified.
+        """
+        if isinstance(infra_object_proto, DatastoreTableProto):
```
adchia, line 81: I see. Is this what you meant by not needing to change the interface of new online stores?

This seems ok for now. If the user uses feast plan and there's some plugin though, might be worth still outputting the rest of feast plan's output (i.e. FCOs) without throwing an error?

adchia: Thoughts on using multiple inheritance for this? e.g. https://docs.python.org/3/tutorial/classes.html#multiple-inheritance

sdk/python/feast/diff/infra_diff.py
felixwang9817: ah because an update transition is destructive: it needs to teardown the old infra and create the new one

as an example, consider changing the namespace of a datastore table: the table in the previous namespace should be deleted, and a new table in the new namespace should be created

an action item is to more clearly distinguish between destructive and non-destructive field changes - right now if any field is changed, it will be a destructive update

sdk/python/feast/infra/online_stores/datastore.py
felixwang9817: ah so the `project_id` field is actually a message field I believe?

`google.protobuf.StringValue project_id = 3;`

so I think this should work

sdk/python/feast/infra/infra_object.py
felixwang9817: that's fair - I think the exception should still be thrown here, and the higher level logic that executes feast plan and logs can just not act on this exception if it arises

sdk/python/feast/diff/FcoDiff.py
```diff
@@ -1,20 +1,40 @@
...
 
 from feast.base_feature_view import BaseFeatureView
 from feast.diff.property_diff import PropertyDiff, TransitionType
 from feast.entity import Entity
 from feast.feature_service import FeatureService
 from feast.protos.feast.core.Entity_pb2 import Entity as EntityProto
+from feast.protos.feast.core.FeatureService_pb2 import (
+    FeatureService as FeatureServiceProto,
+)
+from feast.protos.feast.core.FeatureTable_pb2 import FeatureTable as FeatureTableProto
```
adchia, line 12: remove FeatureTable

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -254,6 +254,12 @@ def from_infra_object_proto(infra_object_proto: InfraObjectProto) -> Any:
             region=infra_object_proto.dynamodb_table.region,
         )
 
+    @staticmethod
+    def from_proto(dynamodb_table_proto: DynamoDBTableProto) -> Any:
+        return DynamoDBTable(
+            name=dynamodb_table_proto.name, region=dynamodb_table_proto.region,
+        )
+
```
adchia, line 262: not sure if this is relevant, but I _think_ that dynamo would error if you try to update a feature view: https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_CreateTable.html

sdk/python/feast/infra/online_stores/datastore.py
adchia: Got it. Maybe add a comment that we're using StringValue? since proto3 by default makes string "".

sdk/python/feast/diff/infra_diff.py
adchia: got it that makes sense.

sdk/python/feast/infra/infra_object.py
```diff
@@ -49,15 +57,38 @@ def from_infra_object_proto(infra_object_proto: InfraObjectProto) -> Any:
...
             cls = _get_infra_object_class_from_type(
                 infra_object_proto.infra_object_class_type
             )
             return cls.from_infra_object_proto(infra_object_proto)
 
-        raise ValueError("Could not identify the type of the InfraObject.")
+        raise FeastInvalidInfraObjectType()
+
+    @staticmethod
+    def from_proto(infra_object_proto: Any) -> Any:
```
adchia, line 71: kind of weird that we have from_proto vs from_infra_object proto. also how does this work with CustomInfra?

sdk/python/feast/infra/online_stores/dynamodb.py
felixwang9817: yup am tracking this for future feast plan work

sdk/python/feast/infra/infra_object.py
felixwang9817: yup am tracking this for future feast plan work

sdk/python/feast/infra/provider.py
```diff
@@ -61,6 +63,18 @@ def update_infra(
...
+        self, config: RepoConfig, desired_registry_proto: RegistryProto
+    ) -> Infra:
+        """
+        Returns the Infra required to support the desired registry.
+
+        Args:
+            config: The RepoConfig for the current FeatureStore.
+            desired_registry_proto: The desired registry, in proto form.
+        """
+        return Infra()
```
adchia, line 76: does it make sense to print a message or something here? e.g. so a custom provider knows that feast plan won't cover their use case? 

Not sure how we intend the SDK version of plan to be used, so maybe it doesn't make sense to log though.

adchia (approved):
/lgtm

PR #2214, opened 2022-01-14
replace GetOnlineFeaturesResponse with GetOnlineFeaturesResponseV2 in…

tsotnet:
… Python

**What this PR does / why we need it**: In Java server we started using GetOnlineFeaturesResponseV2 instead of GetOnlineFeaturesResponse as a response proto type. This PR changes Python codebase to also use GetOnlineFeaturesResponseV2. We also get rid of the previous GetOnlineFeaturesResponse definition completely and rename GetOnlineFeaturesResponseV2 to GetOnlineFeaturesResponse. This is a backwards compatible change for the SDK. Namely, we aren't changing the results of `feature_store.get_online_features(...).{to_dict(),to_df()}` calls. However, the Python HTTP server's response will be different. This is an example of a response:
```
{
  "metadata": {
    "feature_names": [
      "driver_id",
      "conv_rate",
      "acc_rate",
      "avg_daily_trips",
      "conv_rate_plus_val1",
      "conv_rate_plus_val2"
    ]
  },
  "results": [
    {
      "values": [
        1001,
        0.7037263512611389,
        0.8724706768989563,
        308,
        1.703726351261139,
        10.703726351261139
      ],
      "statuses": [
        "PRESENT",
        "PRESENT",
        "PRESENT",
        "PRESENT",
        "PRESENT",
        "PRESENT"
      ],
      "event_timestamps": [
        "1970-01-01T00:00:00Z",
        "2021-12-31T23:00:00Z",
        "2021-12-31T23:00:00Z",
        "2021-12-31T23:00:00Z",
        "1970-01-01T00:00:00Z",
        "1970-01-01T00:00:00Z"
      ]
    },
    {
      "values": [
        1002,
        0.038169607520103455,
        0.48534533381462097,
        332,
        2.0381696075201035,
        20.038169607520103
      ],
      "statuses": [
        "PRESENT",
        "PRESENT",
        "PRESENT",
        "PRESENT",
        "PRESENT",
        "PRESENT"
      ],
      "event_timestamps": [
        "1970-01-01T00:00:00Z",
        "2021-12-31T23:00:00Z",
        "2021-12-31T23:00:00Z",
        "2021-12-31T23:00:00Z",
        "1970-01-01T00:00:00Z",
        "1970-01-01T00:00:00Z"
      ]
    },
    {
      "values": [
        1003,
        0.9665873050689697,
        0.7793770432472229,
        779,
        3.9665873050689697,
        30.96658730506897
      ],
      "statuses": [
        "PRESENT",
        "PRESENT",
        "PRESENT",
        "PRESENT",
        "PRESENT",
        "PRESENT"
      ],
      "event_timestamps": [
        "1970-01-01T00:00:00Z",
        "2021-12-31T23:00:00Z",
        "2021-12-31T23:00:00Z",
        "2021-12-31T23:00:00Z",
        "1970-01-01T00:00:00Z",
        "1970-01-01T00:00:00Z"
      ]
    },
    {
      "values": [
        0,
        null,
        null,
        null,
        null,
        null
      ],
      "statuses": [
        "PRESENT",
        "NOT_FOUND",
        "NOT_FOUND",
        "NOT_FOUND",
        "PRESENT",
        "PRESENT"
      ],
      "event_timestamps": [
        "1970-01-01T00:00:00Z",
        "1970-01-01T00:00:00Z",
        "1970-01-01T00:00:00Z",
        "1970-01-01T00:00:00Z",
        "1970-01-01T00:00:00Z",
        "1970-01-01T00:00:00Z"
      ]
    }
  ]
}
```

Fixes #

```release-note
Use a new response schema for online feature retrieval in Python
```


pyalex (approved):

sdk/python/feast/feature_store.py
```diff
@@ -1179,14 +1180,19 @@ def get_online_features(
...
+        )
         for key, values in join_key_proto_values.items():
-            for row_idx, result_row in enumerate(result_rows):
-                result_row.fields[key].CopyFrom(values[row_idx])
-                result_row.statuses[key] = FieldStatus.PRESENT
+            online_features_response.metadata.feature_names.val.append(key)
+            for row_idx, result_row in enumerate(online_features_response.results):
+                result_row.values.append(values[row_idx])
+                result_row.statuses.append(FieldStatus.PRESENT)
+                result_row.event_timestamps.append(Timestamp())
```
pyalex, line 1195: So python feature server won't provide real event timestamp? and correct status as well

sdk/python/feast/feature_store.py
```diff
@@ -1270,19 +1276,22 @@ def _set_table_entity_keys(
...
         request_data_features: Dict[str, List[Any]],
-        result_rows: List[GetOnlineFeaturesResponse.FieldValues],
     ):
         # Add more feature values to the existing result rows for the request data features
         for feature_name, feature_values in request_data_features.items():
             proto_values = python_values_to_proto_values(
                 feature_values, ValueType.UNKNOWN
             )
 
+            online_features_response.metadata.feature_names.val.append(feature_name)
```
pyalex, line 1288: Why do we put values from request back into response? It seems inconsistent with the fact that we don't put entities in response

sdk/python/feast/feature_store.py
pyalex: ah, this is entities. I got it.
But why we put entities in feature vectors?

tsotnet: Looks like this note got lost during the conflict resolution:
```
# Note: We temporarily store entities in the same field as features for backwards compatibility.
# Java server uses the same GetOnlineFeaturesResponse but does not return entities right now.
# We need to unify the logic between Python & Java on this.```
Basically Python SDK used to return entities as part of the online features response and I don't want to break that with this PR. Although we should have a followup conversation about how to properly resolve this.

sdk/python/feast/feature_store.py
tsotnet: Same reason as for entities

PR #2217, opened 2022-01-18
Adding a local feature server test

adchia:
We've run into issues debugging the lambda test, so introducing a local feature server that is instantiated on tests that have python_feature_server=True on a local provider.

This also logs the stack trace on the feature server. This was useful in debugging https://github.com/feast-dev/feast/pull/2202

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/feature_server.py
+++ b/sdk/python/feast/feature_server.py
@@ -1,3 +1,5 @@
+import traceback
+
 import click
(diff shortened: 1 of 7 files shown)


judahrand (approved):
This looks like it'll make server development way easier! Awesome. 

Test failures look like the class AWS failures.

PR #2221, opened 2022-01-18
Update to newer `redis-py`

judahrand:

As of `redis-py` 4.1 there is a cluster mode included which is based on an improved version of our old dependency. This change will remove a dependency and also allow us to use a newer version of the package.

Fixes #

```release-note
None
```


sdk/python/setup.py
```diff
@@ -77,7 +77,7 @@
 ]
 
 REDIS_REQUIRED = [
-    "redis-py-cluster==2.1.2",
+    "redis==4.1.*",
```
adchia, line 80: thoughts on this being too restrictive? ideally, we'd support both, but users might be using other versions of redis locally too.

pyalex:
Hey @judahrand, can you please update requirements/*.txt files as well, so tests would pass?

judahrand:
> Hey @judahrand, can you please update requirements/*.txt files as well, so tests would pass?

I'll do this but I was confused as I could not find the `*.in` files used to generate them with `pip-tools`. Am I missing them?

judahrand:
> Hey @judahrand, can you please update requirements/*.txt files as well, so tests would pass?

Ah! Got it `make lock-python-dependencies`.

pyalex (approved):

sdk/python/setup.py
adchia: i guess we can do this now and see if anybody complains

adchia: maybe at least do a redis >= 4.1.0?

sdk/python/setup.py
```diff
@@ -101,15 +101,15 @@
...
+    "pytest>=6.0.0",
     "pytest-cov",
     "pytest-xdist",
     "pytest-benchmark>=3.4.1",
     "pytest-lazy-fixture==0.6.3",
     "pytest-timeout==1.4.2",
     "pytest-ordering==0.6.*",
     "pytest-mock==1.10.4",
-    "Sphinx!=4.0.0",
+    "Sphinx!=4.0.0,<4.4.0",
```
adchia, line 112: was this Sphinx change intentional?

judahrand: Yes - without this restriction `pip-tools` was unable to find a valid set of dependencies for Python 3.7 🤷

adchia (approved):

adchia (approved):
/lgtm

PR #2224, opened 2022-01-19
Make online store nullable

### LINKED ISSUE #1809: Leaving "online_store" key out of "aws" provider in feature_store.yaml produces error
In `feature_store.yaml`, if your `provider` is set to `aws`, and you leave out the `online_store` key, it will produce the errors:

```
  File "pydantic/main.py", line 406, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for RepoConfig
__root__ -> online_store -> region
  field required (type=value_error.missing)

feast.repo_config.FeastConfigError: 1 validation error for RepoConfig
__root__ -> online_store -> region
  field required (type=value_error.missing)
at /Users/jimrohrer/Projects/feast-demo/demo/feature_store.yaml
```

Ideally `online_store` would be an optional key. In our case we don't need an online store currently.

For now I've just set the `online_store` to `sqlite`, which let me continue on.

mirayyuce:
**What this PR does / why we need it**: At "feast init" step if no provider is specified the current version initializes the online store with default type based on the provider. However, some users don't want to specify an online store. When the online store parameter is not set in feature_store.yaml with provider either gcp or aws, it will fail at apply().

Fixes #1809 

```release-note
NONE
```


pyalex:
Using Feast without an online store is rather an exception than a rule, simply because this is one of the core features of Feast. I also suspect the proposed change might introduce inconsistency for users who don't specify online store type and instead use inferred value.
That being said, I think the fix should be to enable setting `online_store` explicitly to `null` in `feature_store.yaml` instead of setting sqlite as default always.

sdk/python/feast/infra/online_stores/redis.py
```diff
@@ -61,7 +61,7 @@ class RedisOnlineStoreConfig(FeastConfigBaseModel):
     type: Literal["redis"] = "redis"
     """Online store type selector"""
 
-    redis_type: RedisType = RedisType.redis
+    redis_type: Optional[RedisType] = RedisType.redis
```
pyalex, line 64: any reason for this change?

sdk/python/feast/infra/offline_stores/redshift.py
```diff
@@ -38,22 +38,22 @@ class RedshiftOfflineStoreConfig(FeastConfigBaseModel):
     type: Literal["redshift"] = "redshift"
     """ Offline store type selector"""
 
-    cluster_id: StrictStr
+    cluster_id: StrictStr = "*****"
```
pyalex, line 41: what is this change for?

pyalex (changes requested):
Better solution would be to make `online_store` nullable.
PR also should be cleaned up: `.DS_Store` files should be removed, unrelated changes should be removed.

mirayyuce:
> @mirayyuce thanks for your contribution!
> 
> However, using Feast without an online store is rather an exception than a rule, simply because this is one of the core features of Feast. I also suspect the proposed change might introduce inconsistency for users who don't specify online store type and instead use inferred value. That being said, I think the fix should be to enable setting `online_store` explicitly to `null` in `feature_store.yaml` instead of setting sqlite as default always.

Thank you for the review @pyalex 

In the issue I see two users who don't specify an online store with provider aws having some errors. GCP also causes different errors if we don't set the online store. It searches for Datastore or Firebase in GCP project. These errors were caused by the fact that apply() updates the online store with the latest feature values whether or not there is an online store. If there isn't an online/offline store, repo_config is setting the store type based on the provider, however because of the required parameters for both online and offline store those two users had issues. 

I tried giving some dummy hardcoded values to aws online store parameters, but then it tried to connect and had credential errors. 

At first I set the online store to null, however this caused problems with the tests. In all tests I see that sqlite is assumed as the default online store if it is not specified. This change will be the quickest change with minimum effect. It will help users having errors when they don't set an online store.

sdk/python/feast/infra/online_stores/redis.py
mirayyuce: I was testing required/optional parameters. I meant to take it back before commit

pyalex:
1. Sqlite is not a really good default choice for AWS or GCP. Simply because you will need to store sqlite file on S3 or GS to make it accessible by all services / clients (this is not local mode anymore) and natively sqlite is not even supporting S3 or GS (there are some hacks to make it work, but I would avoid them if possible).

2. For most users (that are using online store) it's totally fine and expected that Feast is trying to setup datastore / dynamo during `apply`. And some of them might have `feature_store.yaml` looking like this:
```
provider: aws
online_store:
  region: us-west-2
```
The proposed change will break their setup.

3. That's why I propose to make `online_store` Optional and allow explicitly set it to `null`. Yes, this will require some changes to `apply` flow and maybe updating some tests. But this would be a more consistent solution that won't break existing users' configurations and actually would solve the original problem, which was that some users don't need an online store, which is not equivalent to "they want to have sqlite as default on AWS".

mirayyuce:
> 5\. That's why I propose to make `online_store` Optional and allow explicitly set it to `null`. Yes, this will require some changes to `apply` flow and maybe updating some tests. But this would be a more consistent solution that won't break existing users' configurations and actually would solve the original problem, which was that some users don't need an online store, which is not equivalent to "some users want to have sqlite as default on AWS/GCP".

Thank you very much for the detailed answer @pyalex. I understand it better now. Here are my current changes in the PR:

1. If a user doesn't want to have an online store, then they should set its type to "null" in feature_store.yaml. If they don't set it to anything or don't provide all config parameters, the flow will break as before.
2. Removed hard coded dummy values from redshift.py
3. Removed the Optional tag from redis.py

After these changes the unit tests passed successfully.

sdk/python/feast/feature_store.py
```diff
@@ -666,14 +666,16 @@ def apply(
...
-        self._get_provider().update_infra(
-            project=self.project,
-            tables_to_delete=views_to_delete if not partial else [],
-            tables_to_keep=views_to_update,
-            entities_to_delete=entities_to_delete if not partial else [],
-            entities_to_keep=entities_to_update,
-            partial=partial,
-        )
+        # Don't call update if there isn't an online store
+        if not isinstance(self.config.online_store, Dict):
```
pyalex, line 670: Provider is responsible for other things, not only online store (eg, AWS Provider deploys lambda function), so we shouldn't skip `update_infra`

sdk/python/feast/infra/online_stores/helpers.py
```diff
@@ -12,13 +13,16 @@
 from feast.protos.feast.types.EntityKey_pb2 import EntityKey as EntityKeyProto
 
 
-def get_online_store_from_config(online_store_config: Any) -> OnlineStore:
```
pyalex, line 15: I think it's better to keep this function unchanged.
Instead this function should be called with guard, eg
```
online_store = get_online_store_from_config(online_store_config) if online_store_config is not None else None
```

sdk/python/feast/infra/passthrough_provider.py
```diff
@@ -46,15 +46,17 @@ def update_infra(
...
-            config=self.repo_config,
-            tables_to_delete=tables_to_delete,
-            tables_to_keep=tables_to_keep,
-            entities_to_keep=entities_to_keep,
-            entities_to_delete=entities_to_delete,
-            partial=partial,
-        )
+        # Call update only if there is an online store
+        if self.online_store:
+            set_usage_attribute("provider", self.__class__.__name__)
```
pyalex, line 51: `set_usage_attribute` line should not be inside this `if` block

pyalex:
@mirayyuce please add a test that would create the desired configuration (with `online_store: null`)

sdk/python/feast/infra/online_stores/helpers.py
mirayyuce: I did the change, however I needed to remove the return type in `get_online_store_from_config`. There was a type mismatch in `passthrough_provider.py` initialization. This below line assumes `Optional[OnlineStore]` is the return type from `get_online_store_from_config` if I don't remove the return type. 

`self.online_store = (
            get_online_store_from_config(config.online_store)
            if not isinstance(config.online_store, Dict)
            else None
        )`

mirayyuce:
> @mirayyuce please add a test that would create the desired configuration (with `online_store: None`) and check that this configuration can be "applied".

Added some tests. All of them are passing successfully. On the other hand, I don't feel quite confident about testing it thoroughly and correctly. I'd like to hear your opinion.

sdk/python/tests/integration/feature_repos/repo_configuration.py
```diff
@@ -67,6 +67,11 @@
                 offline_store_creator=BigQueryDataSourceCreator,
                 online_store=REDIS_CONFIG,
             ),
+            IntegrationTestRepoConfig(
```
pyalex, line 70: This is a list of configurations for all tests. So this will add an additional option to parametrized tests including tests that work with an online store.
So probably it's not what we want. Adding tests to `test_cli` and `test_repo_config` should be enough

sdk/python/tests/integration/scaffolding/test_repo_config.py
```diff
@@ -34,6 +34,37 @@ def _test_config(config_text, expect_error: Optional[str]):
...
 
 
+def test_nullable_online_store_aws():
+    _test_config(
+        dedent(
+            """
+        project: foo
+        registry: "registry.db"
+        provider: aws
+        online_store:
```
pyalex, line 44: Aren't we agree on `online_store: null` format? why it's `type: null`?

sdk/python/tests/integration/registration/test_cli.py
```diff
@@ -128,6 +127,33 @@ def make_feature_store_yaml(project, test_repo_config, repo_dir_name: PosixPath)
     return yaml.safe_dump(config_dict)
 
 
+@pytest.mark.integration
+@pytest.mark.parametrize("test_repo_config", FULL_REPO_CONFIGS)
```
pyalex, line 131: so here, instead of all repo configs we should use only ones with `online_store: None`

sdk/python/feast/infra/online_stores/helpers.py
pyalex: Not sure I got it. Yes, `self.online_store` would be `Optional[OnlineStore]`. `get_online_store_from_config` signature can remain the same.

sdk/python/tests/integration/scaffolding/test_repo_config.py
mirayyuce: I misunderstood. Changed it to online_store: null

sdk/python/feast/infra/online_stores/helpers.py
mirayyuce: I see your point. changed it.

sdk/python/tests/integration/registration/test_cli.py
mirayyuce: I don't really understand how test_cli.py is testing the configs that I added. Adding 3 tests to test_repo_config.py increased total passed tests by 3. However the ones I added in test_cli didn't affect the total number of tests. Is there a doc that I read about the testing structure?

sdk/python/feast/infra/passthrough_provider.py
```diff
@@ -47,14 +52,17 @@ def update_infra(
...
+                config=self.repo_config,
+                tables_to_delete=tables_to_delete,
+                tables_to_keep=tables_to_keep,
+                entities_to_keep=entities_to_keep,
+                entities_to_delete=entities_to_delete,
+                partial=partial,
+            )
 
     def teardown_infra(
         self, project: str, tables: Sequence[FeatureView], entities: Sequence[Entity],
```
felixwang9817, line 68: don't we need to add the same `if self.online_store` check for `teardown_infra`?

if we add this check, we should be able to keep the type annotation in `get_online_store_from_config` as Oleksii suggested

sdk/python/feast/repo_config.py
```diff
@@ -176,6 +176,7 @@ def _validate_online_store_config(cls, values):
         try:
             online_config_class = get_online_config_from_type(online_store_type)
             online_config_class(**values["online_store"])
+
```
felixwang9817, line 179: can we remove the unnecessary new lines?

sdk/python/tests/integration/registration/test_cli.py
```diff
@@ -128,6 +141,58 @@ def make_feature_store_yaml(project, test_repo_config, repo_dir_name: PosixPath)
...
+                provider="gcp",
+                offline_store_creator=BigQueryDataSourceCreator,
+                online_store="null",
+            ),
+            IntegrationTestRepoConfig(
+                provider="aws",
+                offline_store_creator=RedshiftDataSourceCreator,
+                online_store="null",
+            ),
+            IntegrationTestRepoConfig(
```
felixwang9817, line 160: I think you can just add this `IntegrationTestRepoConfig` object into `NULLABLE_ONLINE_STORE_CONFIGS` to begin with, and remove the empty `IntegrationTestRepoConfig`, and then only extend with the `gcp` and `aws` configs

felixwang9817 (commented):
looks great overall!

sdk/python/tests/integration/registration/test_cli.py
felixwang9817: @mirayyuce here are our docs on tests: https://docs.feast.dev/how-to-guides/adding-or-reusing-tests

sdk/python/feast/infra/passthrough_provider.py
mirayyuce: I changed some methods, function signatures and returns. Could you please review? I made those changes based on the output of `make lint-python`

mirayyuce: Could I get one more review @felixwang9817

sdk/python/feast/infra/passthrough_provider.py
```diff
@@ -81,13 +90,14 @@ def online_read(
         table: FeatureView,
         entity_keys: List[EntityKeyProto],
         requested_features: List[str] = None,
-    ) -> List[Tuple[Optional[datetime], Optional[Dict[str, ValueProto]]]]:
+    ):
```
felixwang9817, line 93: can we maintain the return type? you can probably just return `[]` if there is no online store

felixwang9817 (approved):
@mirayyuce looks great overall! I think with a few more changes, this should be ready to merge in

sdk/python/tests/integration/registration/test_cli.py
```diff
@@ -128,6 +141,56 @@ def make_feature_store_yaml(project, test_repo_config, repo_dir_name: PosixPath)
...
 
+NULLABLE_ONLINE_STORE_CONFIGS: List[IntegrationTestRepoConfig] = [
+    IntegrationTestRepoConfig(
+        provider="local",
+        offline_store_creator=FileDataSourceCreator,
+        online_store=None,
+    ),
+]
+
+if os.getenv("FEAST_IS_LOCAL_TEST", "False") == "True":
```
woop, line 152: Can somebody explain the use case for `FEAST_IS_LOCAL_TEST` to me please?

felixwang9817: `FEAST_IS_LOCAL_TEST` means the test only uses the local provider, so it doesn't require cloud resources. We mainly use it in `repo_configuration.py`: if `FEAST_IS_LOCAL_TEST=True`, we don't use all the `IntegrationTestRepoConfigs`, only the one that uses the local provider.

felixwang9817 (approved):
/lgtm

felixwang9817:
@mirayyuce thanks for the contribution to Feast!!

mirayyuce:
> @mirayyuce thanks for the contribution to Feast!!

Thank you so much for the guidance!

PR #2227, opened 2022-01-20
Compare Python objects instead of proto objects

felixwang9817:

**What this PR does / why we need it**: This PR switches `Registry.diff_between` and other helper methods to compare registry objects as Python objects instead of proto objects.

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/diff/FcoDiff.py
+++ b/sdk/python/feast/diff/FcoDiff.py
@@ -17,22 +17,15 @@ from feast.protos.feast.core.RequestFeatureView_pb2 import (
     RequestFeatureView as RequestFeatureViewProto,
 )
 
-FcoProto = TypeVar(
-    "FcoProto",
-    EntityProto,
-    FeatureViewProto,
-    FeatureServiceProto,
-    OnDemandFeatureViewProto,
-    RequestFeatureViewProto,
-)
+Fco = TypeVar("Fco", Entity, BaseFeatureView, FeatureService)
 
 
 @dataclass
-class FcoDiff(Generic[FcoProto]):
+class FcoDiff(Generic[Fco]):
     name: str
     fco_type: str
-    current_fco: FcoProto
-    new_fco: FcoProto
+    current_fco: Fco
+    new_fco: Fco
     fco_property_diffs: List[PropertyDiff]
     transition_type: TransitionType
 
@@ -48,20 +41,28 @@ class RegistryDiff:
         self.fco_diffs.append(fco_diff)
 
 
-Fco = TypeVar("Fco", Entity, BaseFeatureView, FeatureService)
-
-
-def tag_objects_for_keep_delete_add(
+def tag_objects_for_keep_delete_update_add(
     existing_objs: Iterable[Fco], desired_objs: Iterable[Fco]
-) -> Tuple[Set[Fco], Set[Fco], Set[Fco]]:
+) -> Tuple[Set[Fco], Set[Fco], Set[Fco], Set[Fco]]:
     existing_obj_names = {e.name for e in existing_objs}
     desired_obj_names = {e.name for e in desired_objs}
 
     objs_to_add = {e for e in desired_objs if e.name not in existing_obj_names}
-    objs_to_keep = {e for e in desired_objs if e.name in existing_obj_names}
+    objs_to_update = {e for e in desired_objs if e.name in existing_obj_names}
+    objs_to_keep = {e for e in existing_objs if e.name in desired_obj_names}
     objs_to_delete = {e for e in existing_objs if e.name not in desired_obj_names}
 
-    return objs_to_keep, objs_to_delete, objs_to_add
+    return objs_to_keep, objs_to_delete, objs_to_update, objs_to_add
+
+
+FcoProto = TypeVar(
+    "FcoProto",
+    EntityProto,
+    FeatureViewProto,
+    FeatureServiceProto,
+    OnDemandFeatureViewProto,
+    RequestFeatureViewProto,
+)
 
 
 def tag_proto_objects_for_keep_delete_add(
@@ -80,23 +81,27 @@ def tag_proto_objects_for_keep_delete_add(
 FIELDS_TO_IGNORE = {"project"}
 
 
-def diff_between(current: FcoProto, new: FcoProto, object_type: str) -> FcoDiff:
-    assert current.DESCRIPTOR.full_name == new.DESCRIPTOR.full_name
+def diff_between(current: Fco, new: Fco, object_type: str) -> FcoDiff:
+    current_proto = current.to_proto()
+    new_proto = new.to_proto()
+    assert current_proto.DESCRIPTOR.full_name == new_proto.DESCRIPTOR.full_name
     property_diffs = []
     transition: TransitionType = TransitionType.UNCHANGED
-    if current.spec != new.spec:
-        for _field in current.spec.DESCRIPTOR.fields:
+    if current_proto.spec != new_proto.spec:
+        for _field in current_proto.spec.DESCRIPTOR.fields:
             if _field.name in FIELDS_TO_IGNORE:
                 continue
-            if getattr(current.spec, _field.name) != getattr(new.spec, _field.name):
+            if getattr(current_proto.spec, _field.name) != getattr(
+                new_proto.spec, _field.name
+            ):
                 transition = TransitionType.UPDATE
                 property_diffs.append(
                     PropertyDiff(
                         _field.name,
-                        getattr(current.spec, _field.name),
-                        getattr(new.spec, _field.name),
+                        getattr(current_proto.spec, _field.name),
+                        getattr(new_proto.spec, _field.name),
                     )
                 )
     return FcoDiff(
-        new.spec.name, object_type, current, new, property_diffs, transition,
+        new_proto.spec.name, object_type, current, new, property_diffs, transition,
     )
--- a/sdk/python/feast/diff/FcoDiff.py
+++ b/sdk/python/feast/diff/FcoDiff.py
@@ -16,8 +16,18 @@ from feast.protos.feast.core.OnDemandFeatureView_pb2 import (
 from feast.protos.feast.core.RequestFeatureView_pb2 import (
     RequestFeatureView as RequestFeatureViewProto,
 )
-from feast.registry import REGISTRY_OBJECT_TYPE_TO_STR, REGISTRY_OBJECT_TYPES, Registry
+from feast.registry import FeastObjectType, Registry
 from feast.repo_contents import RepoContents
+
+FEAST_OBJECT_TYPE_TO_STR = {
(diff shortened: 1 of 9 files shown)


sdk/python/feast/registry.py
```diff
@@ -143,44 +213,60 @@ def clone(self) -> "Registry":
...
+            request_feature_views=set(self.list_request_feature_views(project=project)),
+            on_demand_feature_views=set(
+                self.list_on_demand_feature_views(project=project)
+            ),
+            feature_services=set(self.list_feature_services(project=project)),
+        )
+
     # TODO(achals): This method needs to be filled out and used in the feast plan/apply methods.
     @staticmethod
     def diff_between(
```
adchia, line 235: this method seems to logically belong as a method within RegistryDiff or something. You no longer are passing in registry protos at least

sdk/python/feast/registry.py
```diff
@@ -65,9 +65,79 @@
...
+    "request_feature_views": "request feature view",
+    "feature_services": "feature service",
+}
+
+REGISTRY_OBJECT_TYPES = REGISTRY_OBJECT_TYPE_TO_STR.keys()
+
 logger = logging.getLogger(__name__)
 
 
+class RepoContents(NamedTuple):
```
adchia, line 81: consider having a separate class for this? This isn't super core to the registry function I think

Also, you have a typo in the docstrings (Feas -> Feast)

sdk/python/feast/registry.py
```diff
@@ -143,44 +213,60 @@ def clone(self) -> "Registry":
         new_registry._registry_store = NoopRegistryStore()
         return new_registry
 
+    def to_repo_contents(self, project: str) -> RepoContents:
```
adchia, line 216: Seems like a confusing concept to me. What is in the registry during `plan` _should_ be different from the repo itself, but we map both registry contents + the current repo python objects to repocontents?

not sure what naming is better for RepoContents... esp since we want to moev away from the term Fco

sdk/python/feast/registry.py
```diff
@@ -65,9 +65,79 @@
...
+REGISTRY_OBJECT_TYPES = REGISTRY_OBJECT_TYPE_TO_STR.keys()
+
 logger = logging.getLogger(__name__)
 
 
+class RepoContents(NamedTuple):
+    """
+    Represents the objects in a Feast feature repo.
+
+    Equivalently, represents the contents of a registry corresponding to a specific Feas project.
```
woop, line 85: typo

sdk/python/feast/diff/FcoDiff.py
```diff
@@ -48,20 +41,28 @@ def add_fco_diff(self, fco_diff: FcoDiff):
...
+    objs_to_update = {e for e in desired_objs if e.name in existing_obj_names}
+    objs_to_keep = {e for e in existing_objs if e.name in desired_obj_names}
     objs_to_delete = {e for e in existing_objs if e.name not in desired_obj_names}
 
-    return objs_to_keep, objs_to_delete, objs_to_add
+    return objs_to_keep, objs_to_delete, objs_to_update, objs_to_add
+
+
+FcoProto = TypeVar(
+    "FcoProto",
```
woop, line 59: Should we maybe just move to `object` instead of `FCO`, since `FCO` isn't a Feast term.

felixwang9817: am going to do this in a future PR, since it touches a lot of files and don't want to create a massive PR

sdk/python/feast/registry.py
felixwang9817: should be fixed since we are now comparing the current registry with the new repo contents

sdk/python/feast/registry.py
```diff
@@ -143,44 +154,84 @@ def clone(self) -> "Registry":
...
+                to_keep,
+                to_delete,
+                to_update,
+                to_add,
+            ) = tag_objects_for_keep_delete_update_add(
+                registry_object_type_to_objects[object_type],
+                getattr(desired_repo_contents, object_type),
             )
 
+            objs_to_keep[object_type] = to_keep
```
adchia, line 196: What's the purpose of having these mapped by object type? Seems like the caller of this immediately flattens it back into one big list. That also seems to introduce this need to look at REGISTRY_OBJECT_TYPES

sdk/python/feast/feature_store.py
```diff
@@ -149,6 +122,7 @@ def __init__(
 
         registry_config = self.config.get_registry_config()
         self._registry = Registry(registry_config, repo_path=self.repo_path)
+        self._registry._initialize_registry()
```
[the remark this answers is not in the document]
felixwang9817: not sure why originally separated - moved this to Registry constructor

sdk/python/feast/registry.py
felixwang9817: we map these by object type because `diff_between` uses `extract_objects_for_keep_delete_update_add`, and `diff_between` needs to know the keep/delete/update/add split by object type in order to compare objects of the same type

sdk/python/feast/registry.py
felixwang9817: yup this is fair - I moved this method to `FcoDiff.py`

sdk/python/feast/registry.py
```diff
@@ -143,44 +154,84 @@ def clone(self) -> "Registry":
         new_registry._registry_store = NoopRegistryStore()
         return new_registry
 
-    # TODO(achals): This method needs to be filled out and used in the feast plan/apply methods.
-    @staticmethod
-    def diff_between(
-        current_registry: RegistryProto, new_registry: RegistryProto
-    ) -> RegistryDiff:
-        diff = RegistryDiff()
+    def extract_objects_for_keep_delete_update_add(
```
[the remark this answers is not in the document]
felixwang9817: note that this is also called in `diff_between`, but still a fair point - it doesn't quite belong with the Registry, so I moved it to `FcoDiff.py`

sdk/python/feast/registry.py
```diff
@@ -65,6 +66,16 @@
     "": "LocalRegistryStore",
 }
 
+REGISTRY_OBJECT_TYPE_TO_STR = {
```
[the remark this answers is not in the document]
felixwang9817: we want to convert plural -> single, so removing underscores doesn't quite do it

sdk/python/feast/feature_store.py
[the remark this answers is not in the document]
felixwang9817: actually moved back to `repo_operations.py`, since we don't want to always initialize the `Registry` - there's a test in `test_online_retrieval.py` intended to ensure that if `registry.db` is deleted, an error is thrown:
```
# Force registry reload (should fail because file is missing)
        with pytest.raises(FileNotFoundError):
            fs_infinite_ttl.refresh_registry()
```

sdk/python/feast/diff/FcoDiff.py
```diff
@@ -80,23 +83,138 @@ def tag_proto_objects_for_keep_delete_add(
...
+    for object_type in REGISTRY_OBJECT_TYPES:
+        objects_to_keep = objs_to_keep[object_type]
+        objects_to_delete = objs_to_delete[object_type]
+        objects_to_update = objs_to_update[object_type]
+        objects_to_add = objs_to_add[object_type]
+
+        for e in objects_to_add:
+            diff.add_fco_diff(
+                FcoDiff(
+                    e.name,
```
adchia, line 193: nit: mind adding the named parameters? e.g. name=e.name. Will make this easier to read

sdk/python/feast/diff/FcoDiff.py
```diff
@@ -80,23 +83,138 @@ def tag_proto_objects_for_keep_delete_add(
...
+        desired_repo_contents: The desired repo state.
+    """
+    objs_to_keep = {}
+    objs_to_delete = {}
+    objs_to_update = {}
+    objs_to_add = {}
+
+    registry_object_type_to_objects: Dict[str, List[Any]]
+    registry_object_type_to_objects = {
+        "entities": registry.list_entities(project=current_project),
```
adchia, line 132: a less error prone way might be to use the direct class names themselves instead of "entities". Looking at this code, I'd be confused why the other fields in the Registry proto don't exist. 

One way to make this maybe a bit cleaner / less error prone could be to make an enum (https://docs.python.org/3/library/enum.html) so you can always reference the enum as key.

sdk/python/feast/registry.py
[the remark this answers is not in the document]
adchia: I see.. this is because we use them directly in feast plan output to the console.

Mind making a comment explaining why this mapping exists? Also, it may make more sense to store this map closer to where it's used (i.e. near some feast plan file or in what today is `repo_operations`)

adchia (approved):

sdk/python/feast/diff/FcoDiff.py
```diff
@@ -80,23 +93,158 @@ def tag_proto_objects_for_keep_delete_add(
...
+        current_project: The Feast project whose objects should be compared.
+        desired_repo_contents: The desired repo state.
+    """
+    objs_to_keep = {}
+    objs_to_delete = {}
+    objs_to_update = {}
+    objs_to_add = {}
+
+    registry_object_type_to_objects: Dict[FeastObjectType, List[Any]]
+    registry_object_type_to_objects = {
```
adchia, line 149: could maybe add a TODO here to clean up and push a lot of these into the enum itself (e.g. the enum could have a get_objects)

sdk/python/feast/diff/FcoDiff.py
```diff
@@ -16,23 +16,28 @@
...
 
-FcoProto = TypeVar(
-    "FcoProto",
-    EntityProto,
-    FeatureViewProto,
-    FeatureServiceProto,
-    OnDemandFeatureViewProto,
-    RequestFeatureViewProto,
-)
+FEAST_OBJECT_TYPE_TO_STR = {
```
adchia, line 22: non-blocking, but a) probably could have this in the enum or b) this map should live closer to the plan logic

sdk/python/feast/diff/FcoDiff.py
```diff
@@ -16,23 +16,28 @@
...
+    FeastObjectType.ENTITY: "entity",
+    FeastObjectType.FEATURE_VIEW: "feature view",
+    FeastObjectType.ON_DEMAND_FEATURE_VIEW: "on demand feature view",
+    FeastObjectType.REQUEST_FEATURE_VIEW: "request feature view",
+    FeastObjectType.FEATURE_SERVICE: "feature service",
+}
+
+FEAST_OBJECT_TYPES = FEAST_OBJECT_TYPE_TO_STR.keys()
```
adchia, line 30: nit: alternative more robust way probably is to do [e.value for e in FeastObjectType]

adchia (approved):
/lgtm

PR #2240, opened 2022-01-26
Delete entity key from Redis only when all attached feature views are gone

### LINKED ISSUE #2150: Redis can delete incorrect keys
I haven't tested this explicitly but from the data model + this function it looks to be as if two FeatureViews have the same Entities (but are different tables) and one view is deleted the online keys for the other view will also be deleted... This seems like a bit of a flaw either in the Redis data model or in the deletion mechanism. 

https://github.com/feast-dev/feast/blob/ec4165396f70ab20b42246b093f777dfcc9f5277/sdk/python/feast/infra/online_stores/redis.py#L75-L88

pyalex:

This PR adds the test for the use case described in #2150 as well as the fix for it.

Fixes #2150 

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/infra/online_stores/redis.py
+++ b/sdk/python/feast/infra/online_stores/redis.py
@@ -72,11 +72,11 @@ class RedisOnlineStoreConfig(FeastConfigBaseModel):
 class RedisOnlineStore(OnlineStore):
     _client: Optional[Union[Redis, RedisCluster]] = None
 
-    def delete_table_values(self, config: RepoConfig, table: FeatureView):
+    def delete_entity_values(self, config: RepoConfig, join_keys: List[str]):
         client = self._get_client(config.online_store)
         deleted_count = 0
         pipeline = client.pipeline()
-        prefix = _redis_key_prefix(table.entities)
+        prefix = _redis_key_prefix(join_keys)
 
         for _k in client.scan_iter(
             b"".join([prefix, b"*", config.project.encode("utf8")])
@@ -85,7 +85,7 @@ class RedisOnlineStore(OnlineStore):
             deleted_count += 1
         pipeline.execute()
 
-        logger.debug(f"Deleted {deleted_count} keys for {table.name}")
+        logger.debug(f"Deleted {deleted_count} rows for entity {', '.join(join_keys)}")
 
     @log_exceptions_and_usage(online_store="redis")
     def update(
@@ -98,10 +98,16 @@ class RedisOnlineStore(OnlineStore):
         partial: bool,
     ):
         """
-        We delete the keys in redis for tables/views being removed.
+        Look for join_keys (list of entities) that are not in use anymore
+        (usually this happens when the last feature view that was using specific compound key is deleted)
+        and remove all features attached to this "join_keys".
         """
-        for table in tables_to_delete:
-            self.delete_table_values(config, table)
+        join_keys_to_keep = set(tuple(table.entities) for table in tables_to_keep)
+
+        join_keys_to_delete = set(tuple(table.entities) for table in tables_to_delete)
+
+        for join_keys in join_keys_to_delete - join_keys_to_keep:
+            self.delete_entity_values(config, list(join_keys))
 
     def teardown(
         self,
@@ -112,8 +118,10 @@ class RedisOnlineStore(OnlineStore):
         """
         We delete the keys in redis for tables/views being removed.
         """
-        for table in tables:
-            self.delete_table_values(config, table)
+        join_keys_to_delete = set(tuple(table.entities) for table in tables)
+
+        for join_keys in join_keys_to_delete:
+            self.delete_entity_values(config, list(join_keys))
 
     @staticmethod
     def _parse_connection_string(connection_string: str):
--- a/sdk/python/tests/integration/online_store/test_universal_online.py
+++ b/sdk/python/tests/integration/online_store/test_universal_online.py
@@ -28,6 +28,7 @@ from tests.integration.feature_repos.universal.entities import (
 )
 from tests.integration.feature_repos.universal.feature_views import (
     create_driver_hourly_stats_feature_view,
+    driver_feature_view,
 )
 from tests.utils.data_source_utils import prep_file_source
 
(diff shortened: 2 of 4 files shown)


judahrand (commented):
What if, for example, you have two different FeatureViews both with the same join_key. Now you want to delete one of those FeatureViews but not the other.

In the Redis data model the features from both feature views are stored as keys under the same HASH. With these changes the extra features won't be deleted because the 'Entity' is still used.

I think there needs to be a method which is able to delete certain Features (from a FeatureView) when that FeatureView is changed or deleted. This method would have to use `HDEL`.

I guess this does fix #2150 but it introduces a new bug (which is admittedly less system breaking).

pyalex:
@judahrand 
Truth be told the current storage model used by Redis integration was not designed to delete features at all and generally speaking it's not possible (at least together with the current Provider API). Since we do not store feature names (only hashes) in Redis maps we need to know those names in advance. But what if you deleted some feature or renamed it. At the moment the feature view is deleted there's no guarantee that we can correctly determine all features (Redis map fields) that must be deleted. And Feast APIs are designed in such way that when the feature is renamed or deleted - it's not being clearly communicated to an online store implementation.

judahrand:
> Truth be told the current storage model used by Redis integration was not designed to delete features at all and generally speaking it's not possible (at least together with the current Provider API). 

Why not? The signature of the method you've renamed to `delete_entity_values` used to take a `FeatureView`. From that you can determine the hash key prefix `_redis_key_prefix(table.entities)`.

Then once you have that you can also determine the fields for that `FeatureView` by doing
```python
fields = [_mmh3(f"{table.name}:{feature.name}") for feature in table.features ]
```

Once you've done that you just need to loop over the HASHs which match your prefix and call HDEL for each field.

I was thinking that `delete_table_values` would just need to be changed to look like this:

```python
def delete_table_values(self, config: RepoConfig, table: FeatureView):
        client = self._get_client(config.online_store)
        deleted_count = 0
        pipeline = client.pipeline()
        prefix = _redis_key_prefix(table.entities)
        fields = [_mmh3(f"{table.name}:{feature.name}") for feature in table.features]

        for _k in client.scan_iter(
            b"".join([prefix, b"*", config.project.encode("utf8")])
        ):
            pipeline.hdel(_k, *fields)
            deleted_count += 1
        pipeline.execute()

        logger.debug(f"Deleted {deleted_count} keys for {table.name}")
```

This would at least delete the correct features when an entire FeatureView was deleted.


> Since we do not store feature names (only hashes) in Redis maps we need to know those names in advance.

> And Feast APIs are designed in such way that when the feature is renamed or deleted - it's not being clearly communicated to an online store implementation.

Yeah, this is true. And arguably is a bigger problem. Letting the OnlineStore just get less and less efficient/bloated etc should not be acceptable. 🤔

sdk/python/tests/integration/online_store/test_universal_online.py
```diff
@@ -503,6 +504,79 @@ def test_online_retrieval(environment, universal_data_sources, full_feature_name
...
+    fs.apply(
+        objects=[simple_driver_fv], objects_to_delete=[driver_stats_fv], partial=False
+    )
+
+    online_features = fs.get_online_features(
+        features=features, entity_rows=entity_rows
+    ).to_dict()
+    assert np.allclose(expected_values["value"], online_features["value"])
+
+    fs.apply(objects=[], objects_to_delete=[simple_driver_fv], partial=False)
```
felixwang9817, line 571: nit: since `partial=False`, there's no need to specify `objects_to_delete`, since only things in `objects` will be applied

sdk/python/feast/infra/online_stores/redis.py
```diff
@@ -98,10 +98,16 @@ def update(
         partial: bool,
     ):
         """
-        We delete the keys in redis for tables/views being removed.
+        Look for join_keys (list of entities) that are not in use anymore
```
felixwang9817, line 101: I agree that this approach will work, but I'm questioning whether we want to preserve this data model. What do you think about switching to a model where the prefix contains not only the `join_keys`, but also the FV name - this way we can just actually delete the correct maps from Redis, instead of waiting for all FVs sharing the same entities to be deleted from the feature store?

felixwang9817:
@pyalex 
> Truth be told the current storage model used by Redis integration was not designed to delete features at all and generally speaking it's not possible (at least together with the current Provider API). Since we do not store feature names (only hashes) in Redis maps we need to know those names in advance. But what if you deleted some feature or renamed it. At the moment the feature view is deleted there's no guarantee that we can correctly determine all features (Redis map fields) that must be deleted. And Feast APIs are designed in such way that when the feature is renamed or deleted - it's not being clearly communicated to an online store implementation.

do you think `feast plan` can resolve this? the goal here is that each online store should report exactly what tables should exist, and the registry contains info on what tables currently exist - then the diff can be applied. so even if you rename a FV, `feast plan` should be able to tear down the old table and create a new one

sdk/python/tests/integration/online_store/test_universal_online.py
pyalex: What I'm trying to achieve here is that this feature view will be passed to provider and then to online store as `tables_to_delete` parameter. Basically, it's equivalent to deleting feature view from the repository and running `apply_total`.

sdk/python/feast/infra/online_stores/redis.py
pyalex: I agree that current model is not ideal. There are obvious trade offs. For example, keeping all features together is better for reading all those features if they are from different feature views. We can discuss this in more details on design meeting.
This PR's only goal is to prevent deleting data, that wasn't supposed to be deleted.

pyalex:
@felixwang9817 yes, I hope that with new design and APIs objects like online store will have more control and will receive all information, like about renaming a view or renaming a feature, whereas right now options are very limited.

sdk/python/feast/infra/online_stores/redis.py
judahrand: > keeping all features together is better for reading all those features if they are from different feature views

I'd agree that this feels important. My gut says that online performance is much more important than code complexity in the offline/configuration paths.

There should be a way to have both I'd have thought - just needs some consideration.

felixwang9817 (approved):
/lgtm

PR #2242, opened 2022-01-27
Fix Redshift data creator

felixwang9817:
**What this PR does / why we need it**: This PR fixes the root cause of several recent failures with AWS integration tests on Github. The issue came from the `test_entity_inference_types_match` test in `test_universal_types.py`, and occurred since the data creator attempted to upload a table to Redshift with a table name of >127 characters. Since Redshift truncates table names to 127 characters, this ensured that subsequent attempts to reference the table would fail, causing the test to fail. Moreover, since PR tests include extra information in the table name, this bug was not reproducible through local testing.

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/errors.py
+++ b/sdk/python/feast/errors.py
@@ -243,6 +243,13 @@ class RedshiftQueryError(Exception):
         super().__init__(f"Redshift SQL Query failed to finish. Details: {details}")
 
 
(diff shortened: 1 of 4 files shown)


pyalex (approved):

judahrand:
Oh! That's a tricky little bug that is! 😮 Thanks for driving this @felixwang9817 !!

PR #2244, opened 2022-01-27
Return `UNIX_TIMESTAMP` as Python `datetime`

judahrand:

Title is pretty self explanatory. However, it should also be noted that before this PR the `unix_timestamp_val` and `unix_timestamp_list_val` attributes of `Value` were never set. So there was a Proto type which was, as far as I can tell, stored incorrectly.

Because this fixes a bug in how `UNIX_TIMESTAMP` types are represented in the OnlineStore this change could result in a mix of Python types being returned for `UNIX_TIMESTAMP` features in existing OnlineStores.

Fixes #

```release-note
Return `UNIX_TIMESTAMP` as Python `datetime`
```

### DIFF
--- a/sdk/python/feast/type_map.py
+++ b/sdk/python/feast/type_map.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import re
+from datetime import datetime, timezone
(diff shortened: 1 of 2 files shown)


felixwang9817 (approved):
/lgtm

PR #2250, opened 2022-01-27
Add backticks to left_table_query_string

dmille:
GCP Project IDs which include a hyphen followed immediately by a number are valid (such as `project-001`), however these need to be escaped with backticks when used in a BQ query as a table reference. This PR adds backticks around the table reference in the MULTIPLE_FEATURE_VIEW_POINT_IN_TIME_JOIN template so that BigQuery doesn't throw an error.

Fixes #

```release-note
NONE
```


adchia:
Thanks for the contribution! /ok-to-test

Just an FYI, you'll have to sign your commit to pass the DCO check

dmille:
> Thanks for the contribution! /ok-to-test
> 
> Just an FYI, you'll have to sign your commit to pass the DCO check

My apologies. I have fixed that.

adchia (approved):
/lgtm

PR #2256, opened 2022-01-29
Feast plan clean up

felixwang9817:
**What this PR does / why we need it**: This PR cleans up `feast plan`. Specifically:
1. `Fco` is renamed to `FeastObject`.
2. Unchanged objects are no longer logged.
first `feast apply`:
Screen Shot 2022-01-28 at 4 41 04 PM
then `feast plan`:
Screen Shot 2022-01-28 at 4 41 45 PM

3. Validation and inference are run during `feast plan`.

Fixes #

```release-note
feast plan output only includes changed objects
```

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -125,6 +125,7 @@ class FeatureStore:
 
         registry_config = self.config.get_registry_config()
         self._registry = Registry(registry_config, repo_path=self.repo_path)
+        self._registry._initialize_registry()
         self._provider = get_provider(self.config, self.repo_path)
 
     @log_exceptions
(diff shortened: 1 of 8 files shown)


sdk/python/feast/feature_store.py
```diff
@@ -504,43 +522,6 @@ def _apply_diffs(
...
-
-        # TODO(felixwang9817): move validation logic into _plan.
-        # Validate all feature views and make inferences.
-        self._validate_all_feature_views(
-            views_to_update, odfvs_to_update, request_views_to_update
-        )
-        self._make_inferences(entities_to_update, views_to_update, odfvs_to_update)
-
-        # Apply infra and registry changes.
         infra_diff.update()
```
adchia, line 525: non-blocking, but another cleanup would be to have InfraObject#update -> InfraObject#create or something since it only right now creates the object

adchia (approved):
/lgtm

PR #2263, opened 2022-02-02
Graduate Python feature server

felixwang9817:
**What this PR does / why we need it**: This PR graduates the Python feature server from alpha status by removing the alpha flag check and modifying the docs. It also deletes some useless doc files.

Fixes #

```release-note
NONE
```


adchia (commented):
Think you missed a couple spots. e.g. feature_store.yaml files that have `python_feature_server: true` (there's one in the java serving file + the helm charts at least)

felixwang9817:
@adchia good catch, I think I removed them all

pyalex:
Probably need to fix this as well:
https://github.com/feast-dev/feast/blob/master/java/serving/src/test/resources/docker-compose/feast10/feature_store.yaml

adchia (approved):
/lgtm

PR #2266, opened 2022-02-02
Set `created_timestamp` and `last_updated_timestamp` fields

### LINKED ISSUE #2267: `created_timestamp` and `last_updated_timestamp` are never set
## Expected Behavior 

`created_timestamp` and `last_updated_timestamp` should be set when things are updated in the Registry.

## Current Behavior

For:

- FeatureViews
- SavedDatasets
- Entites
- FeatureServices
 
this field is not currently set.

For OnDemandFeatureViews the `last_updated_timestamp` field does not even exist and the `created_timestamp` is not set.

### Specifications

- Version:
- Platform:
- Subsystem:

## Possible Solution
Set the fields 😉

judahrand:

Both the BaseFeatureView and OnDemandFeatureView were missing this field, whereas FeatureView had it.

It was also not set when applying a FeatureView.

Add the field where it is missing and actually set it when applying changes to the Registry.

Fixes #2267

```release-note
Fix bug where `last_updated_timestamp` was never set on FeatureViews.
```


sdk/python/feast/on_demand_feature_view.py
```diff
@@ -47,6 +48,8 @@ class OnDemandFeatureView(BaseFeatureView):
     input_feature_view_projections: Dict[str, FeatureViewProjection]
     input_request_data_sources: Dict[str, RequestDataSource]
     udf: MethodType
+    created_timestamp: Optional[datetime] = None
```
woop, line 51: The args above arent updated btw

woop:
On a different note, it's kind of weird how we have both `created_timestamp` for the original creation metadata of an object, as well as the timestamp field name when deduplicating source rows. I think we should maybe make the former just `created` and `last_updated` or something. Doesnt have to happen now though.

sdk/python/feast/on_demand_feature_view.py
judahrand: What do you mean? The other FeatureView classes don't take it in `__init__`. I was just making sure they all matched in terms of behaviour.

Arguably, this setting of `created_timestamp` to `None` should maybe be inherited behaviour.

woop: I mean the docstring. I also think this logic should be inherited.

judahrand: Still not sure what you mean 🤔 The Docstring shouldn't mention this should it as it isn't an arg?

judahrand: Turns out this logic was already performed by the parent class. The logic must've been left in accidentally.

felixwang9817 (approved):
/lgtm

PR #2273, opened 2022-02-03
Fix / update helm chart workflows to push the feast python server

adchia:
The existing script does not successfully run for a variety of reasons. Also, it does not package and push the python feature server. This fixes those (though requires we remove old versions of the helm chart for the java feature server due to weird versioning issues)

Fixes #

```release-note
NONE
```


felixwang9817 (approved):
/lgtm

PR #2275, opened 2022-02-03
Publish alternative python package with FEAST_USAGE=False by default

### LINKED ISSUE #2235: FEAST_USAGE should default to "False"
I'm currently upgrading feast for our organization, making quite a big jump from 0.8 to 0.14. During testing I discovered the addition of the Telemetry/Usage class, to send anonymized metrics/error reports the feast devs.

While it is mentioned in the logging and if you read the release log carefully, I'm surprised/annoyed the decision was made to make `FEAST_USAGE="True"` the default. For my organization, metrics reporting like this will result in a big NO from our security team, regardless of anonymization. While other organizations might not be so strict, I'm not so sure they would appreciate it either. 

A feature store is a complex system to deploy, and to make sure we disable usage I have to set up `FEAST_USAGE="True"` everywhere. While not impossible, it is really easy to miss one or two corner cases, and concerns such as these basically block us from using feast directly. 

I understand the need for metrics/feedback to steer future feast development work, but IMHO this cannot be the way to that goal.

For now, I will resort to forking feast inside our organization and changing the default to False myself.

Make sure the default `FEAST_USAGE="False"` is used

Mentioned above

pyalex:

There will be two python packages `feast` and `feast-no-telemetry`. The last one will have `FEAST_USAGE` set to False by default.
To minimize maintenance costs as much as possible I solved this issue on the last possible stage - GH workflow job.

Fixes #2235 

```release-note
NONE
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -144,6 +144,27 @@ jobs:
           python3 setup.py sdist bdist_wheel
           python3 -m twine upload --verbose dist/*
 
+  publish-python-sdk-no-telemetry:
+    runs-on: ubuntu-latest
+    env:
+      TWINE_USERNAME: __token__
+      TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+    container: python:3.7
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install pip-tools
+        run: pip install pip-tools
+      - name: Install dependencies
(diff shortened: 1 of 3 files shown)


felixwang9817 (approved):
/lgtm

svenharris:
Hey, I know I'm late to the party but I'm trying install this no telemetry version but I don't know how/where to find it on PyPi (I don't see any mention of it in the docs, nor can I find a version on PyPi)

PR #2276, opened 2022-02-03
Fix flaky tests (test_online_store_cleanup & test_feature_get_online_features_types_match)

pyalex:

Several issues addressed in this PR:
1. Test environment time range (start and end dates) was instantiated once on tests start. Which could be an issue when tests are started at the end of the hour and then continued to the next hour. So there will be difference between generated datasets boundaries and environment start and end timestamps.
2. In `test_online_store_cleanup` some online stores (like dynamo) can have eventual consistency of schema. So when table was deleted it might need some time before it can be recreated.
3. Bug with inferring feature type when all values are empty list.

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/tests/integration/registration/test_universal_types.py
+++ b/sdk/python/tests/integration/registration/test_universal_types.py
@@ -226,7 +226,9 @@ def test_feature_get_online_features_types_match(online_types_test_fixtures):
     driver_id_value = "1" if config.entity_type == ValueType.STRING else 1
     online_features = fs.get_online_features(
         features=features, entity_rows=[{"driver": driver_id_value}],
-    ).to_dict()
+    )
+    print(online_features.proto)  # will be in the logs when test fails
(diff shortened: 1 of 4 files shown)


sdk/python/tests/integration/registration/test_universal_types.py
```diff
@@ -221,7 +221,12 @@ def test_feature_get_online_features_types_match(online_types_test_fixtures):
     features = [fv.name + ":value"]
     entity = driver(value_type=config.entity_type)
     fs.apply([fv, entity])
-    fs.materialize(environment.start_date, environment.end_date)
+    fs.materialize(
+        environment.start_date,
+        environment.end_date
+        - timedelta(hours=1)  # throwing out last record to make sure
```
adchia, line 227: non blocking, but kind of confusing why this ensures we infer from all empty values

adchia (approved):
/lgtm

PR #2296, opened 2022-02-08
avoid using transactions on OSS Redis

DvirDukhan:
Signed-off-by: DvirDukhan <dvir@redis.com>

This PR avoids using Transactions in redis-py pipelines when on a single instance mode. 
redis-py pipelines are created by default as transactions https://github.com/redis/redis-py/blob/24cdd70761c7c275d5010f628f83d919bf35fc90/redis/client.py#L1020
However, on cluster mode, redis-py pipeline is built with `transaction=None` https://github.com/redis/redis-py/blob/045d5ed15305ccc44a0330e6f65f669998815598/redis/cluster.py#L698
from https://redis.io/topics/transactions :
```
All the commands in a transaction are serialized and executed sequentially.
It can never happen that a request issued by another client is served in the middle of the execution of a Redis transaction.
This guarantees that the commands are executed as a single isolated operation.
```
This could change can help increasing throughput when accessing a single Redis instance with multiple Feast clients.

```release-note
NONE
```

CC @chayim

### DIFF
--- a/sdk/python/feast/infra/online_stores/redis.py
+++ b/sdk/python/feast/infra/online_stores/redis.py
@@ -75,7 +75,7 @@ class RedisOnlineStore(OnlineStore):
     def delete_entity_values(self, config: RepoConfig, join_keys: List[str]):
         client = self._get_client(config.online_store)
         deleted_count = 0
-        pipeline = client.pipeline()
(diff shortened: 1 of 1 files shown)


adchia (approved):
/lgtm

PR #2300, opened 2022-02-10
Remove old flag warning with the python feature server

adchia:
We removed the need for a flag, but still have warnings that print when running `feast serve`. This removes those warnings.

Fixes #

```release-note
NONE
```


felixwang9817 (approved):
/lgtm

thanks for catching this!

PR #2304, opened 2022-02-11
podAnnotations Values in the feature-server chart

tpvasconcelos:

This PR adds the possibility of a user passing arbitrary `podAnnotations` Values to the feature-server chart.

Reference: https://tectonfeast.slack.com/archives/C01MSKCMB37/p1644600405847739


adchia (approved):
/lgtm

PR #2309, opened 2022-02-15
Don't require `snowflake` to always be installed

judahrand:

The import should be handled like in `aws_utils`

Fixes #

```release-note
None
```


adchia (approved):
/lgtm

PR #2316, opened 2022-02-17
feat: Add support for fetching features using multiple feature views with different entities in the Java feature server

### LINKED ISSUE #2302: Java feature server fails to retrieve features from multiple feature views
## Expected Behavior 
When querying features from multiple feature views, the server should be able to fetch the values of all features successfully.

## Current Behavior
The server responds with `NOT_FOUND` for all features if either of these conditions is true:

* The `features` section of the request contains at least two features from different feature views
* The `entities` section of the request contains at least one excess entity

## Steps to reproduce
* Define two feature views with different entities
* Materialize the data to Redis
* Start the Java feature server
* Send a query asking to fetch features from different feature views. Example:

```json
{
    "features": {
        "val": [
            "feature_view_1:feature_1",
            "feature_view_1:feature_2",
            "feature_view_2:feature_3"
        ]
    },
    "entities": {
        "entity_1_id": {
            "val": [
                {
                    "int32_val": 1
                },
                {
                    "int32_val": 2
                }
            ]
        },
        "entity_2_id": {
            "val": [
                {
                    "int32_val": 10
                },
                {
                    "int32_val": 11
                }
            ]
        }
    }
}
```

The server will respond with statuses `NOT_FOUND` for all features. If you were to remove `"feature_view_2:feature_3"` and `"entity_2_id"` the server should respond with actual values. 

### Specifications

- Version: 0.18.0
- Platform: Linux

## Possible Solution
I did some investigation and compared the functionality of the Python and Java feature server implementations. The Python server seems to split the requested features by feature view and then make separate requests for each of them. The Java server doesn't do that and that probably messes up the Redis key generation down the line.

I was able to implement this grouping by feature view locally and everything seems to be working now. If you feel like this is the correct approach, I would be happy to clean my code up and send in a PR!

vstrimaitis:
When fetching features from multiple features views using different entities, the feature server returns `NOT_FOUND` for all of the features instead of properly using the available information to get all feature values. This PR fixes this issue.

Fixes #2302

```release-note
Java feature server supports fetching from multiple feature views with different entities
```

### DIFF
--- a/java/serving/src/main/java/feast/serving/service/OnlineServingServiceV2.java
+++ b/java/serving/src/main/java/feast/serving/service/OnlineServingServiceV2.java
@@ -21,7 +21,9 @@ import com.google.common.collect.Maps;
 import com.google.protobuf.Duration;
 import com.google.protobuf.Timestamp;
 import feast.common.models.Feature;
+import feast.proto.core.EntityProto;
 import feast.proto.core.FeatureServiceProto;
+import feast.proto.core.FeatureViewProto.FeatureViewSpec;
 import feast.proto.serving.ServingAPIProto;
 import feast.proto.serving.ServingAPIProto.FeatureReferenceV2;
 import feast.proto.serving.ServingAPIProto.FieldStatus;
@@ -110,13 +112,29 @@ public class OnlineServingServiceV2 implements ServingServiceV2 {
       throw new RuntimeException("Requested features list must not be empty");
     }
 
+    List<EntityProto.Entity> entities = this.registryRepository.getEntities();
+    Map<String, List<String>> entityNamesPerFeatureView = retrievedFeatureReferences.stream()
+        .map(this.registryRepository::getFeatureViewSpec)
+        .distinct()
+        .collect(Collectors.toMap(
+            FeatureViewSpec::getName,
+            spec -> spec.getEntitiesList().stream().map(name ->
+                entities.stream()
+                    .filter(e -> e.getSpec().getName().equals(name))
+                    .findFirst()
+                    .get()).map(e -> e.getSpec().getJoinKey())
+                .collect(Collectors.toList())
+            )
+        );
+
     Span storageRetrievalSpan = tracer.buildSpan("storageRetrieval").start();
     if (storageRetrievalSpan != null) {
       storageRetrievalSpan.setTag("entities", entityRows.size());
       storageRetrievalSpan.setTag("features", retrievedFeatureReferences.size());
     }
     List<List<feast.storage.api.retriever.Feature>> features =
-        retriever.getOnlineFeatures(entityRows, retrievedFeatureReferences, entityNames);
+        retriever.getOnlineFeatures(entityRows, retrievedFeatureReferences, entityNames,
+            entityNamesPerFeatureView);
 
     if (storageRetrievalSpan != null) {
       storageRetrievalSpan.finish();
--- a/java/serving/src/main/java/feast/serving/service/OnlineServingServiceV2.java
+++ b/java/serving/src/main/java/feast/serving/service/OnlineServingServiceV2.java
@@ -21,7 +21,6 @@ import com.google.common.collect.Maps;
 import com.google.protobuf.Duration;
 import com.google.protobuf.Timestamp;
 import feast.common.models.Feature;
-import feast.proto.core.EntityProto;
 import feast.proto.core.FeatureServiceProto;
 import feast.proto.core.FeatureViewProto;
 import feast.proto.core.FeatureViewProto.FeatureViewSpec;
@@ -119,7 +118,7 @@ public class OnlineServingServiceV2 implements ServingServiceV2 {
             .distinct()
             .map(this.registryRepository::getFeatureViewSpecByName)
             .collect(Collectors.toList());
-    List<EntityProto.Entity> entities = this.registryRepository.getEntities();
+    Map<String, String> entityNameToJoinKey = this.registryRepository.getEntityJoinKeyLookup();
     Map<String, List<String>> entityNamesPerFeatureView =
         featureViewSpecs.stream()
             .collect(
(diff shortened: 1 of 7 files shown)


adchia (approved):
/lgtm

adchia:
mind rebasing against master? think the tests should be fixed

java/serving/src/main/java/feast/serving/service/OnlineServingServiceV2.java
```diff
@@ -110,13 +112,33 @@ public GetFeastServingInfoResponse getFeastServingInfo(
       throw new RuntimeException("Requested features list must not be empty");
     }
 
+    List<EntityProto.Entity> entities = this.registryRepository.getEntities();
+    Map<String, List<String>> entityNamesPerFeatureView =
+        retrievedFeatureReferences.stream()
+            .map(this.registryRepository::getFeatureViewSpec)
```
pyalex, line 118: so here we calling `getFeatureViewSpec` for each feature, which is not optimal, especially when there could be dozens features per single view. I think extracting feature view names -> deduplicating them (`distinct`) -> retrieving spec from registry would be much better.

java/serving/src/main/java/feast/serving/service/OnlineServingServiceV2.java
```diff
@@ -110,13 +112,33 @@ public GetFeastServingInfoResponse getFeastServingInfo(
...
+            .distinct()
+            .collect(
+                Collectors.toMap(
+                    FeatureViewSpec::getName,
+                    spec ->
+                        spec.getEntitiesList().stream()
+                            .map(
+                                name ->
+                                    entities.stream()
```
pyalex, line 127: Since it's latency sensitive part I would suggest to replace this lookup with Map<EntityName -> JoinKey> that will be provided by registry, for example

java/serving/src/main/java/feast/serving/service/OnlineServingServiceV2.java
```diff
@@ -110,13 +112,33 @@ public GetFeastServingInfoResponse getFeastServingInfo(
...
+
     Span storageRetrievalSpan = tracer.buildSpan("storageRetrieval").start();
     if (storageRetrievalSpan != null) {
       storageRetrievalSpan.setTag("entities", entityRows.size());
       storageRetrievalSpan.setTag("features", retrievedFeatureReferences.size());
     }
     List<List<feast.storage.api.retriever.Feature>> features =
-        retriever.getOnlineFeatures(entityRows, retrievedFeatureReferences, entityNames);
+        retriever.getOnlineFeatures(
+            entityRows, retrievedFeatureReferences, entityNames, entityNamesPerFeatureView);
```
pyalex, line 141: I think `entityNames` is now redundant when we have `entityNamesPerFeatureView`

java/storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/OnlineRetriever.java
```diff
@@ -55,22 +59,47 @@ public OnlineRetriever(
   public List<List<Feature>> getOnlineFeatures(
       List<Map<String, ValueProto.Value>> entityRows,
       List<ServingAPIProto.FeatureReferenceV2> featureReferences,
-      List<String> entityNames) {
+      List<String> entityNames,
+      Map<String, List<String>> entityNamesPerFeatureView) {
+
+    List<String> featureViewNames =
```
pyalex, line 65: Isn't 65-72 equivalent to `Set(entityNamesPerFeatureView.values())` ?

java/storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/OnlineRetriever.java
```diff
@@ -55,22 +59,47 @@ public OnlineRetriever(
...
+                entityRow ->
+                    uniqueEntityKeyCombinations.stream()
+                        .map(
+                            entitiesToRetain -> {
+                              Map<String, Value> res = new HashMap<>(entityRow);
+                              res.keySet().retainAll(entitiesToRetain);
+                              return res;
+                            })
+                        .collect(Collectors.toList()))
+            .map(es -> RedisKeyGenerator.buildRedisKeys(this.project, es))
```
pyalex, line 86: I suggest replace this to flatMap. So it will be flat `List<RedisProto.RedisKeyV2>` not list of lists. It seems that we know in advance how many sub keys for each row, our N rows will always transform to E * N keys, where E is amount of entities. So it will be easy to transform flat list of responses back into N element list

pyalex (changes requested):
@vstrimaitis thanks for PR.
overall looks good, but needs some clean-up and a little optimization

java/storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/OnlineRetriever.java
```diff
@@ -55,22 +59,47 @@ public OnlineRetriever(
...
+            .collect(Collectors.toList());
 
-    List<RedisProto.RedisKeyV2> redisKeys =
-        RedisKeyGenerator.buildRedisKeys(this.project, entityRows);
     return getFeaturesFromRedis(redisKeys, featureReferences);
```
pyalex, line 89: we also need to group featureReferences by entities. Otherwise we'll be retrieving all features for all keys and that will cause redis latency degradation

java/serving/src/main/java/feast/serving/service/OnlineServingServiceV2.java
vstrimaitis: Good catch!

java/storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/OnlineRetriever.java
vstrimaitis: You probably meant `.keySet`, not `.values`, but you're correct, thanks!

kevjumba:
Hey @vstrimaitis are you still working on this PR? The PR is very close barring a few small cleanup changes!

vstrimaitis:
> Hey @vstrimaitis are you still working on this PR? The PR is very close barring a few small cleanup changes!

I have most of the suggestions implemented, but I'm still missing the last few. I'm a bit preoccupied at the moment to finish this up, so if someone wants to help with that - feel free to do it! Otherwise, I'll try to get back to this whenever I have some spare time.

java/storage/connectors/redis/src/main/java/feast/storage/connectors/redis/retriever/OnlineRetriever.java
zhiyanshao: I was under the impression that `getOnlineFeatures` assumes `featureReferences` and `entities` are all from the same feature view. The grouping features by feature view is done at the service level. This way this logic can be moved to the service layer and we don't have to handle it at each connector level, especially if there is a need to add a customized connector.

PR #2321, opened 2022-02-18
Add owner field to FeatureService

felixwang9817:
**What this PR does / why we need it**: This PR adds the `owner` field to `FeatureService`. It also cleans up `FeatureService` and `FeatureViewProjection`.

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/feature_service.py
+++ b/sdk/python/feast/feature_service.py
@@ -19,21 +19,25 @@ from feast.usage import log_exceptions
 
 class FeatureService:
     """
-    A feature service is a logical grouping of features for retrieval (training or serving).
-    The features grouped by a feature service may come from any number of feature views.
+    A feature service defines a logical group of features from one or more feature views.
+    This group of features can be retrieved together during training or serving.
 
-    Args:
-        name: Unique name of the feature service.
-        features: A list of Features that are grouped as part of this FeatureService.
-            The list may contain Feature Views, Feature Tables, or a subset of either.
-        tags (optional): A dictionary of key-value pairs used for organizing Feature
-            Services.
+    Attributes:
+        name: The unique name of the feature service.
+        feature_view_projections: A list containing feature views and feature view
+            projections, representing the features in the feature service.
+        description: A human-readable description.
+        tags: A dictionary of key-value pairs to store arbitrary metadata.
+        owner: The owner of the feature service.
+        created_timestamp: The time when the feature service was created.
+        last_updated_timestamp: The time when the feature service was last updated.
     """
 
     _name: str
     _feature_view_projections: List[FeatureViewProjection]
+    _description: str
     _tags: Dict[str, str]
-    _description: Optional[str] = None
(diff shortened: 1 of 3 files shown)


sdk/python/feast/feature_service.py
```diff
@@ -19,21 +19,25 @@
...
+    A feature service defines a logical group of features from one or more feature views.
+    This group of features can be retrieved together during training or serving.
+
+    Attributes:
+        name: The unique name of the feature service.
+        feature_view_projections: A list containing feature views and feature view
+            projections, representing the features in the feature service.
+        description: A human-readable description.
+        tags: A dictionary of key-value pairs to store arbitrary metadata.
+        owner: The owner of the feature service.
```
woop, line 31: Should we give more details here? Should it be an email? Is this owner already registered elsewhere in the feature store? Can it be any string?

sdk/python/feast/feature_view_projection.py
```diff
@@ -10,15 +10,26 @@
 
 @dataclass
 class FeatureViewProjection:
+    """
+    A feature view projection represents a set of features from a single feature view.
```
woop, line 14: What about
```suggestion
    A feature view projection represents a selection of one more more features from a single feature view.
```

sdk/python/feast/feature_view_projection.py
```diff
@@ -10,15 +10,26 @@
...
+
+    Attributes:
+        name: The unique name of the feature view projection.
```
woop, line 17: Wait, but feature view projections aren't public facing objects right? So `name` is really the aliased name of the feature view, not the projection. Or am I missing something?

sdk/python/feast/feature_view_projection.py
```diff
@@ -10,15 +10,26 @@
...
+        name_alias: An optional alias for the name.
+        features: The list of features represented by the feature view projection.
+        join_key_map: A map to modify join key columns during retrieval of this feature
+            view projection.
+    """
+
     name: str
-    name_alias: Optional[str]
-    features: List[Feature]
+    name_alias: str = ""
```
woop, line 25: Is there a reason for this change?

sdk/python/feast/feature_view_projection.py
```diff
@@ -10,15 +10,26 @@
...
+    features: List[Feature] = []
```
woop, line 26: Doesn't this introduce a bug?

```class Service(object):
    data = []

s1 = Service()
s2 = Service()

s1.data.append(1)

print(s1.data)
[1]
print(s2.data)
[1]

s2.data.append(2)

print(s1.data)
[1, 2]
print(s2.data)
[1, 2]
```

sdk/python/feast/feature_service.py
felixwang9817: Sure, added a note about email.

woop (approved):

PR #2329, opened 2022-02-20
feat: Add owner field to Entity and rename labels to tags

felixwang9817:

**What this PR does / why we need it**: This PR adds the `owner` field to `Entity`. It also renames `labels` to `tags`, and cleans up some unused methods.

Fixes #

```release-note
Owner field added to Entity, and labels renamed to tags
```

### DIFF
--- a/sdk/python/feast/entity.py
+++ b/sdk/python/feast/entity.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import warnings
 from datetime import datetime
 from typing import Dict, Optional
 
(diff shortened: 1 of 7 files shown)


sdk/python/feast/entity.py
```diff
@@ -97,88 +95,68 @@ def __str__(self):
...
-        Gets the description of this entity.
-        """
-        return self._description
+    def value_type(self) -> ValueType:
+        return self._value_type
 
-    @description.setter
-    def description(self, description):
-        """
-        Sets the description of this entity.
```
woop, line 122: Is there a reason we are removing all comments?

sdk/python/feast/entity.py
```diff
@@ -97,88 +95,68 @@ def __str__(self):
...
-    def labels(self) -> Dict[str, str]:
-        """
-        Gets the labels of this entity.
-        """
-        return self._labels
+    def tags(self) -> Dict[str, str]:
+        return self._tags
 
-    @labels.setter
-    def labels(self, labels: Dict[str, str]):
```
woop, line 162: Isn't this a breaking change? I think we should keep backwards compat for at least one minor version.

sdk/python/feast/entity.py
felixwang9817: as discussed, we will remove all properties

sdk/python/feast/entity.py
felixwang9817: yeah good catch - just updated this

woop (approved):

PR #2336, opened 2022-02-24
feat: Metadata changes & making data sources top level objects to power Feast UI

adchia:
Key changes:
- Data sources are now registered in the registry at the top level and can be queried. They need to have names specified (defaulting to the table ref). Note that we move the request data source's name into the top level name instead. They also now have projects attached to them
- Removing unused fields in protos
- `feast registry-dump` becomes more official as a way of dumping data for the Web UI
- Saved datasets can optionally also reference a feature service if they were generated from a feature service.

Fixes #

```release-note
Data sources are now discoverable from the CLI and are attached to projects.
```

### DIFF
--- a/sdk/python/feast/data_source.py
+++ b/sdk/python/feast/data_source.py
@@ -139,6 +139,7 @@ class DataSource(ABC):
     DataSource that can be used to source features.
 
     Args:
+        name: Name of data source, which should be unique within a project
         event_timestamp_column (optional): Event timestamp column used for point in time
             joins of feature values.
         created_timestamp_column (optional): Timestamp column indicating when the row
@@ -149,6 +150,7 @@ class DataSource(ABC):
         date_partition_column (optional): Timestamp column used for partitioning.
     """
 
+    name: str
     event_timestamp_column: str
     created_timestamp_column: str
     field_mapping: Dict[str, str]
@@ -156,12 +158,14 @@ class DataSource(ABC):
 
     def __init__(
         self,
+        name: str,
         event_timestamp_column: Optional[str] = None,
         created_timestamp_column: Optional[str] = None,
         field_mapping: Optional[Dict[str, str]] = None,
         date_partition_column: Optional[str] = None,
     ):
         """Creates a DataSource object."""
+        self.name = name
         self.event_timestamp_column = (
             event_timestamp_column if event_timestamp_column else ""
         )
@@ -173,12 +177,16 @@ class DataSource(ABC):
             date_partition_column if date_partition_column else ""
         )
 
+    def __hash__(self):
+        return hash((id(self), self.name))
+
     def __eq__(self, other):
         if not isinstance(other, DataSource):
             raise TypeError("Comparisons should only involve DataSource class objects.")
 
         if (
-            self.event_timestamp_column != other.event_timestamp_column
+            self.name != other.name
+            or self.event_timestamp_column != other.event_timestamp_column
             or self.created_timestamp_column != other.created_timestamp_column
             or self.field_mapping != other.field_mapping
             or self.date_partition_column != other.date_partition_column
@@ -206,7 +214,9 @@ class DataSource(ABC):
             cls = get_data_source_class_from_type(data_source.data_source_class_type)
             return cls.from_proto(data_source)
 
-        if data_source.file_options.file_format and data_source.file_options.file_url:
+        if data_source.request_data_options and data_source.request_data_options.schema:
+            data_source_obj = RequestDataSource.from_proto(data_source)
+        elif data_source.file_options.file_format and data_source.file_options.file_url:
             from feast.infra.offline_stores.file_source import FileSource
 
             data_source_obj = FileSource.from_proto(data_source)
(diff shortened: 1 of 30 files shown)


sdk/python/feast/data_source.py
```diff
@@ -271,6 +279,13 @@ def __eq__(self, other):
 
         return True
 
+    @property
```
woop, line 282: btw @felixwang9817 has started to move us to `attributes` instead of having all these property getter/setters. May be worth sneaking in.

sdk/python/feast/diff/registry_diff.py
```diff
@@ -109,23 +114,28 @@ def diff_registry_objects(
     assert current_proto.DESCRIPTOR.full_name == new_proto.DESCRIPTOR.full_name
     property_diffs = []
     transition: TransitionType = TransitionType.UNCHANGED
-    if current_proto.spec != new_proto.spec:
-        for _field in current_proto.spec.DESCRIPTOR.fields:
+
+    if isinstance(current_proto, DataSourceProto):
```
woop, line 118: Mind explaining what is happening in this if statement? What are the two different conditions that we support?

sdk/python/feast/infra/offline_stores/file_source.py
```diff
@@ -40,38 +42,43 @@ def __init__(
...
+        if path is None:
+            if file_url is None:
+                raise ValueError(
+                    'No "path" argument provided. Please set "path" to the location of your file source.'
+                )
+            else:
+                from warnings import warn
+
+                warn(
+                    'Argument "file_url" is being deprecated. Please use the "path" argument.'
```
woop, line 61: Shouldnt we just kill this now?

woop:
Should we have tests for
1. Testing data source names being unique
2. Testing that a change to a data source shows up in plan/apply for feature views as well. Presumably we would indicate that the feature view is being changed
3. Testing name inference

sdk/python/feast/registry.py
```diff
@@ -511,9 +601,6 @@ def apply_materialization(
                 existing_feature_view = FeatureView.from_proto(
                     existing_feature_view_proto
                 )
-                existing_feature_view.materialization_intervals.append(
```
felixwang9817, line 514: why is this being deleted?

sdk/python/feast/registry.py
```diff
@@ -480,12 +548,34 @@ def get_on_demand_feature_view(
...
+                return DataSource.from_proto(data_source)
+        raise DataSourceObjectNotFoundException(name, project=project)
+
     def apply_materialization(
         self,
         feature_view: FeatureView,
         project: str,
         start_date: datetime,
         end_date: datetime,
+        earliest_latest_event_timestamp: Optional[Tuple[datetime, datetime]] = None,
```
felixwang9817, line 578: why is this being added?

sdk/python/tests/example_repos/example_feature_repo_1.py
```diff
@@ -16,7 +16,9 @@
 )
 
 customer_profile_source = BigQuerySource(
-    table_ref="feast-oss.public.customers", event_timestamp_column="event_timestamp",
+    name="customer_profile_source",
```
felixwang9817, line 19: maybe add names to all other example feature repos as well?

sdk/python/feast/registry.py
adchia: Accidental. Adding back in

sdk/python/feast/registry.py
adchia: Removed. An artifact that I missed from a larger PR

sdk/python/feast/diff/registry_diff.py
adchia: it's a difference in how we're serializing the proto. Every other FCO has a "spec" vs "meta" middle layer, but data source doesn't have that and has everything flattened.

sdk/python/feast/data_source.py
```diff
@@ -223,6 +223,7 @@ class DataSource(ABC):
     DataSource that can be used to source features.
 
     Args:
+        name: Name of data source
```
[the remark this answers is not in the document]
adchia: done

sdk/python/tests/example_repos/example_feature_repo_1.py
adchia: i wanted to have a mix of things with and without names too. But generally seems like most were using this one anyways?

adchia:
> Should we have tests for
> 
> 1. Testing data source names being unique
> 2. Testing that a change to a data source shows up in plan/apply for feature views as well. Presumably we would indicate that the feature view is being changed
> 3. Testing name inference

Added the first and 3rd test. The second one seems like it wouldn't necessarily be introduced by this PR (but still would be a good unit test to add), so leaving out of this PR

adchia:
@woop @felixwang9817 PTAL!

felixwang9817 (approved):
/lgtm

PR #2339, opened 2022-02-24
Go server

ctkhanhly:

This PR adds a go server as a subprocess in the feature store so that on get_online_features() call, features are served via the go subprocess instead of the python server. The go subprocess will be spun up on the first time get_online_features() is called from the python program and will start serving from there. Currently, feast serve is not supported by the go server. The hope is that go server will serve features faster than python server.

As for future work, we have a working code for feast serve and a python plugin connector and are working on some benchmarkings to add goroutines and replace some current functions with more optimized implementations to potentially speed up the go server, but I'm open to suggestions on but not limited to coding styles, bugs, and potential optimizations.

user can add go_feature_server flag in their feature_store.yaml configuration file to enable using go server on get_online_features()


Makefile
```diff
@@ -78,6 +78,9 @@ test-python-universal-local:
 test-python-universal:
 	FEAST_USAGE=False IS_TEST=True python -m pytest -n 8 --integration --universal sdk/python/tests
 
+test-python-go-server:
```
woop, line 81: Is the idea that this command would be run if you are developing Go code? (as the most comprehensive way to test Go retrieval)

Makefile
```diff
@@ -124,16 +127,17 @@ install-go-ci-dependencies:
...
-	cd ${ROOT_DIR}/sdk/go; go test ./...
+	go test ./...
 
 format-go:
-	cd ${ROOT_DIR}/sdk/go; gofmt -s -w *.go
+	gofmt -s -w go/**/*.go
 
 lint-go:
-	cd ${ROOT_DIR}/sdk/go; go vet
+	go vet ./go/feast ./go/server
```
woop, line 140: Adding a note so that we don't forget: Need to add lint-go to pre-commit hook

go/server/server.go
```diff
@@ -0,0 +1,22 @@
+package main
```
woop, line 1: We should probably move `go/server/server/go` to `go/cmd/server.go` as per https://github.com/golang-standards/project-layout/blob/master/cmd/README.md

go/feast/basefeatureview.go
```diff
@@ -0,0 +1,43 @@
+package feast
```
woop, line 1: I think it would make sense to move the files under `go/feast/` into `go/internal`. Eventually we want to have them under a `pkg` folder so that people can import it, but I think it's reasonable to start with all of it under `internal`

go/feast/registry.go
```diff
@@ -0,0 +1,277 @@
...
+		cachedOnDemandFeatureViews: make(map[string]map[string]*core.OnDemandFeatureView),
+		cachedRequestFeatureViews:  make(map[string]map[string]*core.RequestFeatureView),
+	}
+	loadEntitiesDone := make(chan struct{}, 1)
+	loadFeatureServicesDone := make(chan struct{}, 1)
+	loadFeatureViewsDone := make(chan struct{}, 1)
+	loadOnDemandFeatureViewsDone := make(chan struct{}, 1)
+	loadRequestFeatureViewsDone := make(chan struct{}, 1)
+	doneLoad := make(chan struct{}, 1)
+	go func() {
```
woop, line 49: What is the performance benefit of this go routine vs serial execution?

go/feast/repoconfig.go
```diff
@@ -0,0 +1,86 @@
...
+	FeatureServer map[string]interface{} `json:"feature_server"`
+	// Feature flags for experimental features
+	Flags map[string]interface{} `json:"flags"`
+}
+
+// NewRepoConfig reads file <repoPath>/feature_store.yaml
+// and converts the YAML format into RepoConfig struct.
+// It first uses repoPath to read feature_store.yaml if it exists,
+// however if it doesn't then it checks configJSON and tries parsing that.
+func NewRepoConfig(repoPath string, configJSON string) (*RepoConfig, error) {
```
woop, line 33: Would it make sense to have a `NewRepoConfigFromJson` and a wrapping `NewRepoConfigFromFile` (instead of the conditional arguments which is a bit less intuitive)

go/server/main.go
```diff
@@ -0,0 +1,92 @@
...
+	feastServerVersion   = "0.18.0"
+	defaultFeastHttpPort = "8081"
+)
+
+func main() {
+
+	repoPath := os.Getenv(flagFeastRepoPath)
+	repoConfig := os.Getenv(flagFeastRepoConfig)
+
+	if repoPath == "" && repoConfig == "" {
```
woop, line 30: We should also look for both being set (which shouldnt be the case)

Makefile
ctkhanhly: To your first question, yes, but I think we should add the configs for this test to the universal tests once ODFVs are supported for the go server? @felixwang9817 may be working on adding this command to CI?

go/server/main.go
```diff
@@ -0,0 +1,92 @@
...
+		log.Fatalln(fmt.Sprintf("One of %s of %s environment variables must be set", flagFeastRepoPath, flagFeastRepoConfig))
+		return
+	}
+	config, err := feast.NewRepoConfig(repoPath, repoConfig)
+	if err != nil {
+		log.Fatalln(err)
+		return
+	}
+	log.Println("Initializing feature store...")
```
woop, line 39: We should add a `TODO` to add a proper logging library like https://github.com/Sirupsen/logrus

go/feast/registry.go
ctkhanhly: I haven't done a benchmarking on this nor added this to my current benchmarking code since it doesn't affect get_online_features() performance directly. But the idea is we can load lists of feature views, entities, feature services, on-demand feature views, etc from the registry in parallel. This might be beneficial if we add a refresh function to reload the registry cache multiple times?

sdk/python/feast/go_server.py
```diff
@@ -0,0 +1,246 @@
...
+
+        self.grpc_server_started = False
+        self.pipe_closed = True
+
+
+def _get_unused_port() -> str:
+    sock = socket.socket()
+
+    # Binding to port 0 means the OS will choose an unused port
+    sock.bind(("", 0))
```
woop, line 239: We probably don't want to just randomly select a port. Otherwise we are adding non-determinism into the our users' environments. I think it would be more sane to start at a specific port value (something high) and fail if its used (which would indicate that multiple versions of this server are running). Or is there a use case for multiple Go servers?

sdk/python/feast/go_server.py
```diff
@@ -0,0 +1,246 @@
...
+            "FEAST_REPO_CONFIG": self._config.json(),
+            "FEAST_REPO_PATH": self._repo_path,
+            "FEAST_GRPC_PORT": self.unused_port,
+            **os.environ,
+        }
+        cwd = feast.__path__[0]
+
+        if "dev" in feast.__version__:
+            self.process = subprocess.Popen(
+                ["go", "run", "github.com/feast-dev/feast/go/server"],
```
woop, line 152: Seems like this part needs to be changed to load a static binary.

sdk/python/feast/go_server.py
```diff
@@ -0,0 +1,246 @@
...
+                    )
+                elif error_message.startswith(ValueError.__name__):
+                    parsed_error_message = error_message.split(": ")[1]
+                    raise ValueError(parsed_error_message)
+                else:
+                    raise
+
+        return OnlineResponse(response)
+
+    def _connect(self):
```
woop, line 137: A bit of a strange name, given that we are not just connecting but also spawning a process.

sdk/python/setup.py
```diff
@@ -165,6 +165,17 @@
 else:
     use_scm_version = None
 
+# Build go server for 3 targets: macos (intel), macos (m1), linux (64 bit)
+# First start by clearing the necessary directory
+binaries_path = (pathlib.Path(__file__) / "../feast/binaries").resolve()
```
woop, line 170: We should ensure that this code only runs during dev/test environments and not when users pip install feast

go/feast/registry.go
woop: It seems like a premature optimization to me

woop:
Just did a quick sweep. Will have a look again tomorrow.

go/feast/repoconfig.go
ctkhanhly: Makes sense, I will separate this init function!

Makefile
felixwang9817: any reason we're not using `gofmt` here?

sdk/python/setup.py
ctkhanhly: I've tried building the wheel using commands from release.yaml in the workflow and verified that binaries are shipped with the wheel

go/feast/registry.go
ctkhanhly: I will remove the goroutines

sdk/python/feast/go_server.py
ctkhanhly: What about _start_go_server?

sdk/python/feast/go_server.py
ctkhanhly: If there are multiple python processes running feast (in different shell sessions) all use go_server, then we need multiple go_server subprocesses. We could start something high, i.e: 54321, and add 1 if it's used until we find an unused port?

sdk/python/feast/go_server.py
woop: sgtm

sdk/python/feast/go_server.py
woop: sgtm

go/server/main.go
ctkhanhly: I think we want both even when repoConfig is in case user gives a relative registry path to repoPath?

go/feast/basefeatureview.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"github.com/feast-dev/feast/go/protos/feast/core"
+)
+
+type BaseFeatureView struct {
```
woop, line 9: I think I may have missed the addition of BaseFeatureView to the Python codebase, but is there any reason we can't have purely a FeatureView and NewFeatureView function?

go/feast/featurestore.go
```diff
@@ -0,0 +1,907 @@
...
+
+type entityKeyRow struct {
+	entityKey  *types.EntityKey
+	rowIndices []int
+}
+
+// NewFeatureStore constructs a feature store fat client using the
+// repo config (contents of feature_store.yaml converted to JSON map).
+func NewFeatureStore(config *RepoConfig) (*FeatureStore, error) {
+	onlineStore, err := getOnlineStore(config)
```
woop, line 28: Would it make sense to use `NewOnlineStore` here in order to stay consistent with the rest of the code base?

go/feast/connector.go
```diff
@@ -0,0 +1,19 @@
+package feast
+
+import (
+	"errors"
+	"fmt"
+)
+
+func getOnlineStore(config *RepoConfig) (OnlineStore, error) {
```
woop, line 8: Would it make more sense to have this function be placed in `onlinestore.go`?

go/feast/basefeatureview.go
```diff
@@ -0,0 +1,43 @@
...
+		features[index] = NewFeatureFromProto(featureSpecV2)
+	}
+	base.features = features
+	base.projection = NewFeatureViewProjectionFromDefinition(base)
+	return base
+}
+
+func (fv *BaseFeatureView) withProjection(projection *FeatureViewProjection) (*BaseFeatureView, error) {
+	if projection.name != fv.name {
+		return nil, errors.New(fmt.Sprintf("The projection for the %s FeatureView cannot be applied because it differs in name. "+
```
woop, line 28: Can we please uncapitalize this string and remove the ending punctuation? https://github.com/golang/go/wiki/CodeReviewComments#error-strings

go/feast/connector.go
```diff
@@ -0,0 +1,19 @@
...
+	onlineStoreType, ok := getOnlineStoreType(config.OnlineStore)
+	if !ok {
+		return nil, errors.New(fmt.Sprintf("could not get online store type from online store config: %+v", config.OnlineStore))
+	}
+	if onlineStoreType == "redis" {
+		onlineStore, err := NewRedisOnlineStore(config.Project, config.OnlineStore)
+		return onlineStore, err
+	} else {
+		return nil, errors.New("Only Redis is supported as an online store for now")
```
woop, line 17: Can we add contextual information here (like the actual store the user is trying to create)?

go/feast/featurestore.go
```diff
@@ -0,0 +1,907 @@
...
+
+/*
+	This function returns all feature references from GetOnlineFeaturesRequest.
+	If a list of feature references is passed, return it.
+	Otherwise, FeatureService was passed, parse this feature service
+	to get a list of FeatureViewProjection and return feature references
+	from this list
+*/
+
+func (fs *FeatureStore) getFeatures(parsedKind interface{}, allowCache bool) ([]string, error) {
```
woop, line 233: It seems a bit strange that we are parsing the kind and still continue to pass around an `interface{}`. Is it possible for us to use a proper type for the kind?

go/feast/featurestore.go
```diff
@@ -0,0 +1,907 @@
...
+
+	fvs := make(map[string]*FeatureView)
+	requestFvs := make(map[string]*RequestFeatureView)
+	odFvs := make(map[string]*OnDemandFeatureView)
+
+	featureViews, err := fs.listFeatureViews(allowCache, hideDummyEntity)
+	if err != nil {
+		// TODO (Ly): Review: return an empty list instead of
+		// error if no FeatureView is found in this project
+		featureViews = make([]*FeatureView, 0)
```
woop, line 276: So what response would the user see in this case? Would they understand what is happening?

go/feast/featurestore.go
```diff
@@ -0,0 +1,907 @@
...
+	}
+
+	if featureService, ok := parsedKind.(*FeatureService); ok {
+
+		fvsToUse := make([]*FeatureView, 0)
+		requestFvsToUse := make([]*RequestFeatureView, 0)
+		odFvsToUse := make([]*OnDemandFeatureView, 0)
+
+		for _, featureProjection := range featureService.projections {
+			// Create copies of FeatureView that may
```
woop, line 309: Nitpick: I think we can make these comments a little bit longer before breaking them. Is there some kind of length you are shooting for?

go/feast/featurestore.go
```diff
@@ -0,0 +1,907 @@
...
+	entityNameToJoinKeyMap := make(map[string]string)
+	var entityNames map[string]bool
+	var entityName string
+	var joinKeyMap map[string]string
+	var featureView *FeatureView
+
+	entities, err := fs.listEntities(true, false)
+
+	if err != nil {
+		entities = make([]*Entity, 0)
```
woop, line 373: What would happen here if the registry is an API and it's down. Don't we need to handle the error?

go/feast/featurestore.go
```diff
@@ -0,0 +1,907 @@
...
+
+func (fs *FeatureStore) validateEntityValues(joinKeyValues map[string]*types.RepeatedValue) (int, error) {
+	setOfRowLengths := make(map[int]bool)
+	var numRows int
+	for _, col := range joinKeyValues {
+		setOfRowLengths[len(col.Val)] = true
+		numRows = len(col.Val)
+	}
+	if len(setOfRowLengths) > 1 {
+		return 0, errors.New("ValueError: All entity rows must have the same columns.")
```
woop, line 417: Is ValueError still relevant here?

go/feast/featurestore.go
```diff
@@ -0,0 +1,907 @@
...
+			}
+		}
+		if len(collidedFeatureRefs) >= 1 {
+			collidedFeatureRefList := make([]string, len(collidedFeatureRefs))
+			index := 0
+			for featureName := range collidedFeatureRefs {
+				collidedFeatureRefList[index] = featureName
+				index += 1
+			}
+			return errors.New(fmt.Sprintf("FeatureNameCollisionError: %s; %t", strings.Join(collidedFeatureRefList, ", "), fullFeatureNames))
```
woop: Instead of using strings for error names, would it be better to use custom errors? I also think that just using an error name here is a bit cryptic. I would probably just start with an error message.

go/feast/featurestore.go
```diff
@@ -0,0 +1,907 @@
...
+		}
+	} else {
+		var featureName string
+		for _, featureRef := range featureRefs {
+
+			parsedFeatureName := strings.Split(featureRef, ":")
```
woop: Should we maybe pull `:` out into a constant that gets reused throughout the code base?

go/feast/featurestore.go
```diff
@@ -0,0 +1,907 @@
...
+			if _, ok := requestDataFeatures[feature]; !ok {
+				missingFeatures = append(missingFeatures, feature)
+			}
+		}
+		return errors.New(fmt.Sprintf("RequestDataNotFoundInEntityRowsException: %s", strings.Join(missingFeatures, ", ")))
+	}
+	return nil
+}
+
+func (fs *FeatureStore) checkOutsideMaxAge(featureTimestamp *timestamppb.Timestamp, currentTimestamp *timestamppb.Timestamp, ttl *durationpb.Duration) bool {
```
woop: should this be `checkOutsideTtl`?

go/feast/basefeatureview.go
ctkhanhly: I think Python codebase has BaseFeatureView and I'm based off of that. When ODFVs and Request FVs are supported, all of them can have projections and share function withProjection() so I thought this would be useful to share certain properties of these 3 types of FV

go/feast/featurestore.go
```diff
@@ -0,0 +1,907 @@
...
+
+				values[index] = &value
+
+				if _, ok := value.Val.(*types.Value_NullVal); ok {
+					values[index] = nil
+					status = serving.FieldStatus_NOT_FOUND
+				} else if fs.checkOutsideMaxAge(&eventTimeStamp, timestamppb.Now(), table.ttl) {
+					values[index] = &value
+					status = serving.FieldStatus_OUTSIDE_MAX_AGE
+				}
```
woop: Are we missing an else here?

protos/feast/third_party/grpc/connector/Connector.proto
```diff
@@ -0,0 +1,34 @@
+syntax = "proto3";
```
woop, line 1: I wonder if this connector interface shouldnt rather be within `serving`. Third party is normally for external protos that we pull in. This is still our proto.

sdk/python/feast/feature_store.py
```diff
@@ -1172,6 +1176,15 @@ def get_online_features(
                 except KeyError as e:
                     raise ValueError("All entity_rows must have the same keys.") from e
 
+        # If Go feature server is enabled, send request to it instead of going through a regular Python logic
+        if enable_go_feature_server(self.config):
+            # Lazily start the go server on the first request
+            if self._go_server is None:
+                self._go_server = GoServer(str(self.repo_path.absolute()), self.config)
```
woop, line 1183: Perhaps we can delay casting to `str` until the last moment?

sdk/python/feast/feature_store.py
```diff
@@ -1766,6 +1780,11 @@ def _get_feature_views_to_use(
 
         return views_to_use
 
+    def stop_go_server(self):
+        """Force the Go feature server to shut down."""
+        if enable_go_feature_server(self.config) and self._go_server is not None:
+            self._go_server.stop()
```
woop, line 1786: We should have an `else` here I think. Otherwise what would happen if the user ran this command. Maybe we should raise an exception?

sdk/python/feast/go_server.py
```diff
@@ -0,0 +1,246 @@
...
+            )
+
+        # Make sure the subprocess is terminated when the parent process dies
+        # Note: this doesn't handle cases where the parent process is abruptly
+        # killed (e.g. with SIGKILL)
+        atexit.register(lambda: self.stop())
+        self.start_grpc_server()
+        self.pipe_closed = False
+
+    def start_grpc_server(self):
```
woop, line 174: Is there a reason why `_connect()` is private but `start_grpc_server()` is public?

sdk/python/feast/go_server.py
```diff
@@ -0,0 +1,246 @@
...
+        # from constructor. We try connecting to the Go subprocess for 5 seconds (50
+        # times with 0.1 second intervals).
+        # TODO (Ly): Review: increase time.sleep(0.2) to 0.2 so wait 10 seconds
+        # since python connector plugin can take up to 8 seconds to spin up.
+        # 5 seconds is not enough sometimes to spin up grpc server itself
+        for i in range(50):
+            try:
+                self.client.GetFeastServingInfo(request=GetFeastServingInfoRequest())
+                break
+            except Exception as e:
```
woop, line 206: Should we use a more specific exception here?

sdk/python/tests/conftest.py
```diff
@@ -68,12 +72,19 @@ def pytest_addoption(parser):
...
+        default=False,
+        help="Run tests that use the go feature server",
+    )
 
 
 def pytest_collection_modifyitems(config, items: List[Item]):
     should_run_integration = config.getoption("--integration") is True
     should_run_benchmark = config.getoption("--benchmark") is True
     should_run_universal = config.getoption("--universal") is True
+    should_run_without_odfv = config.getoption("--goserver") is True
```
woop, line 87: should this be `should_run_go_server`?

sdk/python/setup.py
```diff
@@ -257,7 +322,13 @@ def run(self):
     ],
     entry_points={"console_scripts": ["feast=feast.cli:cli"]},
     use_scm_version=use_scm_version,
-    setup_requires=["setuptools_scm", "grpcio", "grpcio-tools==1.34.0", "mypy-protobuf==3.1.0", "sphinx!=4.0.0"],
+    setup_requires=[
```
woop, line 303: Why did so much formatting in this file change?

Makefile
```diff
@@ -31,7 +31,7 @@ lint: lint-python lint-java lint-go
 
 test: test-python test-java test-go
 
-protos: compile-protos-go compile-protos-python compile-protos-docs
+protos: compile-protos-docs
```
woop, line 34: What happened to these commands?

go/feast/basefeatureview.go
woop: Yea the question wasn't so much directed at you. I understand we want a parent. I guess the problem is that we don't have a nice name for the parent of FV and ODFV.

sdk/python/setup.py
```diff
@@ -166,61 +162,130 @@
 else:
     use_scm_version = None
 
+# Build go server for 3 targets: macos (intel), macos (m1), linux (64 bit)
+# First start by clearing the necessary directory
+binaries_path = (pathlib.Path(__file__) / "../feast/binaries").resolve()
```
woop: We already have the `repo_root` variable.

sdk/python/feast/go_server.py
```diff
@@ -0,0 +1,243 @@
...
+from feast.protos.feast.serving.ServingService_pb2 import (
+    GetFeastServingInfoRequest,
+    GetOnlineFeaturesRequest,
+)
+from feast.protos.feast.serving.ServingService_pb2_grpc import ServingServiceStub
+from feast.repo_config import RepoConfig
+from feast.type_map import python_values_to_proto_values
+
+
+class GoServer:
```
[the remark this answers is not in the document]
woop: Played around with running the subprocess in a background thread. Below worked for me. It automatically restarts the process and also allows the main process to kill the subprocess when it exits.

server.py
```python
from time import sleep


import random
import string


id = ''.join(random.choice(string.ascii_lowercase) for i in range(10))

count = 0
while True:
    try:
        print(f'Currently running server subprocess with name:{id} for count {count}')
        sleep(1)
        count += 1

        # # Kill the process every 5 iterations
        if count == 5:
            print('Server subprocess kills itself')
            exit(5)
    except KeyboardInterrupt as e:
        print("Cleaning up subprocess")
        exit(0)
```

main.py
```python
import atexit
import subprocess
import threading
import time
import signal


def background_process(stop_server):
    while True:
        # Start process
        proc = subprocess.Popen(['python3', 'server.py'])
        print()
        print(f"Started new server subprocess {proc.pid}")

        # Check to see if process is still running
        while proc.poll() is None and not stop_server():

            # TODO: Use a gRPC health check here to see if the server is up
            # Don't use readline since it will block!
            print(f"doing health check for process {proc.pid}")

            time.sleep(1)

        if stop_server():
            try:
                proc.wait(timeout=2)
            except subprocess.TimeoutExpired:
                pass
            finally:
                proc.kill()
                print('Cleaning up parent thread')
                break

def main():
    stop_server = False

    def stop_server_gracefully_atexit():
        nonlocal stop_server
        stop_server = True

    def stop_server_gracefully_on_signal(signal, frame):
        stop_server_gracefully_atexit()

    atexit.register(stop_server_gracefully_atexit)
    signal.signal(signal.SIGTERM, stop_server_gracefully_on_signal)
    signal.signal(signal.SIGINT, stop_server_gracefully_on_signal)

    t = threading.Thread(target=background_process, args=(lambda: stop_server,))

    t.start()
    try:
        t.join()
    except KeyboardInterrupt as e:
        pass

    print('Cleaning up main process')

main()
```

go/cmd/go_server/main.go
```diff
@@ -0,0 +1,106 @@
...
+	writer := bufio.NewWriter(os.Stdout)
+	serverCounts := 0
+	for {
+		text, _ := reader.ReadString('\n')
+		text = strings.Trim(text, "\n")
+		commands := strings.Split(text, " ")
+		if len(commands) == 0 {
+			log.Fatalln(errors.New("Invalid command. Should be [startGrpc] or [stop]"))
+			return
+		} else if commands[0] == "startGrpc" {
```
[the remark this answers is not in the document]
ctkhanhly: Originally, I wanted to have both startGrpc and startHttp commands so both can share a feature store. I can remove this for now and leave the discussion on the design of http server on feast serve when it's supported?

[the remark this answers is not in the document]
woop: Yip, sounds good.

sdk/python/feast/feature_store.py
woop: Well I can imagine that a user will only call this method if they think they are running a Go server. So I would probably raise an exception if the Go server is not enabled at all, and maybe leave the rest as is.

sdk/python/setup.py
```diff
@@ -168,59 +162,108 @@
...
-            '--grpc_python_out', self.this_package,
-            '--mypy_out', self.this_package] + proto_files)
+        subprocess.check_call(
+            self.go_protoc
+            + [
+                "-I",
+                self.proto_folder,
+                "--go_out",
+                self.go_folder,
+                "--grpc_python_out",
```
woop, line 219: This command `--grpc_python_out` makes my `pip install -e` fail. Do we need it?

.github/workflows/pr_integration_tests.yml
```diff
@@ -70,7 +70,7 @@ jobs:
           docker push $ECR_REGISTRY/$ECR_REPOSITORY:${{ steps.image-tag.outputs.DOCKER_IMAGE_TAG }}
     outputs:
       DOCKER_IMAGE_TAG: ${{ steps.image-tag.outputs.DOCKER_IMAGE_TAG }}
-  integration-test-python:
+  integration-test-python-and-go:
```
woop, line 73: By the way we also need to update `master_only.yml` with all the changes we have made here.

.github/workflows/pr_integration_tests.yml
```diff
@@ -170,3 +170,14 @@ jobs:
           env_vars: OS,PYTHON
           fail_ci_if_error: true
           verbose: true
+      - name: Test go
```
woop, line 173: My only concern here is that we are adding a step to an already slow-running test suite which will further slow it down. Would it make sense to run the Go server tests as part of the rest of the test suite so that it can be parallelized?

felixwang9817: this is far, I'll make sure the Go tests can get parallelized

go/feast/featurestore.go
ctkhanhly: I'm adding a RegistryStore like python sdk so it returns error (exception in calling GetOnlineFeatures from go server) if call to refresh/ initialize registry is not successful instead

sdk/python/feast/go_server.py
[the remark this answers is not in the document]
ctkhanhly: I'm fixing go server implementation by improving on this!

sdk/python/feast/feature_store.py
ctkhanhly: Can you clarify on what the last moment is?

woop: https://github.com/feast-dev/feast/pull/2339/files#diff-9ef7735ea7a5e6ea2b8ad263b7e3dd5c51ba67e29de9f3b2f34eacc063e0cebbR146

sdk/python/feast/go_server.py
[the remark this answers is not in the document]
woop: Nice!

Makefile
```diff
@@ -119,8 +119,10 @@ build-java-no-tests:
 # Go SDK
 
 install-go-ci-dependencies:
-	go get -u github.com/golang/protobuf/protoc-gen-go
-	go get -u golang.org/x/lint/golint
+	go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
```
woop, line 122: We can't let this dependency float. We should pin it to a specific version I think.

sdk/python/feast/go_server.py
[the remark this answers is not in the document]
ctkhanhly: I have implemented a background thread (as a blocking thread instead of polling) but this took a very long time for pytest and the test didn't work properly so I had 2 versions. I will experiment more on this...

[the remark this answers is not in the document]
woop: > as a blocking thread instead of polling

Blocking is ok, as long as it's still possible to send a shutdown command from the main process. The reason I didn't use something like `stdout.readline` was because it wouldn't receive a command from the main process to shut down.

.github/workflows/linter.yml
```diff
@@ -34,8 +34,10 @@ jobs:
             ${{ runner.os }}-${{ steps.setup-python.outputs.python-version }}-pip-
       - name: Install pip-tools
         run: pip install pip-tools
-      - name: Install dependencies
-        run: make install-python-ci-dependencies
+      - name: Install Python
+        run: |
+          make install-python-ci-dependencies
+          make install-python
```
achals, line 40: Why was this step necessary?

go/cmd/goserver/main.go
```diff
@@ -0,0 +1,85 @@
...
+	"fmt"
+	"github.com/feast-dev/feast/go/internal/config"
+	"github.com/feast-dev/feast/go/internal/feast"
+	"github.com/feast-dev/feast/go/protos/feast/serving"
+	"google.golang.org/grpc"
+	"log"
+	"net"
+	"os"
+)
+
```
achals, line 13: go fmt should reorder these for you I think

go/cmd/goserver/main.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	startGrpcServer(fs, grpcPort)
+}
+
+func startGrpcServer(fs *feast.FeatureStore, grpcPort string) {
+	server := servingServiceServer{
+		fs: fs,
+	}
+	log.Printf("Starting a gRPC server at port %s...", grpcPort)
+	lis, err := net.Listen("tcp", fmt.Sprintf(":%s", grpcPort))
```
achals, line 72: Is this server intended to be exposed to users directly, or only via the python API? If the latter, it should probably be using unix domain sockets which are often much faster. 

Ideally you should be able to support both depending on the actual scenario that the server is being spun up for.

https://stackoverflow.com/questions/54844882/grpc-server-in-python-with-unix-domain-socket

go/cmd/goserver/server.go
```diff
@@ -0,0 +1,23 @@
...
+func (s *servingServiceServer) GetFeastServingInfo(ctx context.Context, request *serving.GetFeastServingInfoRequest) (*serving.GetFeastServingInfoResponse, error) {
+	return &serving.GetFeastServingInfoResponse{
+		Version: feastServerVersion,
+	}, nil
+}
+
+func (s *servingServiceServer) GetOnlineFeatures(ctx context.Context, request *serving.GetOnlineFeaturesRequest) (*serving.GetOnlineFeaturesResponse, error) {
+	s.fs.SetOnlineStoreContext(ctx)
+	return s.fs.GetOnlineFeatures(request)
+}
```
achals, line 23: If I'm reading this correctly there's a race condition, right? Because concurrent requests to `GetOnlineFeatures` may set the `ctx` in the Online store? Why not just pass the ctx into the `fs.GetOnlineFeatures` method directly?

go/cmd/goserver/main.go
```diff
@@ -0,0 +1,85 @@
...
+	defaultFeastGrpcPort = "6566"
+	feastServerVersion   = "0.18.0"
+	defaultFeastHttpPort = "8081"
+)
+
+// TODO: Add a proper logging library such as https://github.com/Sirupsen/logrus
+
+func main() {
+
+	repoPath := os.Getenv(flagFeastRepoPath)
```
pyalex, line 27: There's very neat lib to do this w/o manual parsing for each variable
https://github.com/kelseyhightower/envconfig

go/cmd/goserver/server.go
pyalex: can we not do this set context at all and instead pass it everywhere as a parameter, as it being done in all go libraries.

go/internal/feast/featurestore.go
```diff
@@ -0,0 +1,885 @@
...
+
+	if entitylessCase {
+		dummyEntityColumn := &types.RepeatedValue{Val: make([]*types.Value, numRows)}
+		for index := 0; index < numRows; index++ {
+			dummyEntityColumn.Val[index] = &DUMMY_ENTITY
+		}
+		joinKeyValues[DUMMY_ENTITY_ID] = dummyEntityColumn
+	}
+
+	for table, requestedFeatures := range groupedRefs {
```
pyalex, line 202: Since we rewriting this anyway, I think it's a good chance to finally fix this behaviour. We should group features by entities, not by feature views (also name `table` here is even more misleading), and call `readFromOnlineStore` once per entity not once per feature view. Hence, let storage layer to decide what storage model should be used and leave room for optimization.

ctkhanhly: Hi Oleksii, so instead of querying a set of entity keys (serves as primary key) as a key in hash table (e.g: redis) which could uniquely identify a set of feature views and return those and only get the features from feature views we want, would your suggestion instead to query one entity at a time and retrieve all feature views that contains this entity key but not necessarily uniquely identifies a row in those feature views? If this is the case, we would have to change the implementation of materialization (currently subject to python sdk), which maybe out of scope for this PR? Do you have any suggestion on how we can initiate that change but in-scope for this PR?

pyalex: exactly

sdk/python/setup.py
felixwang9817: because I manually ran format and lint commands on it

achals (commented):
Looks mostly good for this iteration. Aside from the comments in line, we should avoid checking in the feature server binaries into git and have them in S3 or some place.

.gitignore
```diff
@@ -196,4 +197,7 @@ sdk/go/protos/
 **/registry.db
 **/*.aof
 **/*.rdb
-**/nodes.conf
\ No newline at end of file
+**/nodes.conf
+
+# Go subprocess binaries (built during feast pip package building)
+sdk/python/feast/binaries
```
achals, line 203: I think this will need a trailing slash.

go/internal/feast/featurestore.go
```diff
@@ -0,0 +1,885 @@
...
+		return errors.New(fmt.Sprintf("requestDataNotFoundInEntityRowsException: %s", strings.Join(missingFeatures, ", ")))
+	}
+	return nil
+}
+
+func (fs *FeatureStore) checkOutsideTtl(featureTimestamp *timestamppb.Timestamp, currentTimestamp *timestamppb.Timestamp, ttl *durationpb.Duration) bool {
+	return currentTimestamp.GetSeconds()-featureTimestamp.GetSeconds() > ttl.Seconds
+}
+
+func (fs *FeatureStore) populateResultRowsFromColumnar(response *serving.GetOnlineFeaturesResponse, data map[string]*types.RepeatedValue) {
```
pyalex, line 512: The name is misleading. `response` has columnar structure as well as `data`.

go/internal/feast/featurestore.go
```diff
@@ -0,0 +1,885 @@
...
+	timeStamp := timestamppb.Now()
+	for entityName, values := range data {
+		response.Metadata.FeatureNames.Val = append(response.Metadata.FeatureNames.Val, entityName)
```
pyalex, line 515: why it's called `entityName` if it's being added to features list?

go/internal/infra/local.go
```diff
@@ -0,0 +1,67 @@
...
+		return nil, err
+	}
+	return registry, nil
+}
+
+func (r *LocalRegistryStore) UpdateRegistryProto(rp *core.Registry) error {
+	return r.writeRegistry(rp)
+}
+
+func (r *LocalRegistryStore) Teardown() error {
```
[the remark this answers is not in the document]
ctkhanhly: I had this to comply with RegistryStore interface. I don't know if the plan is to migrate every feature from python sdk to go server, then I've made this interface in case we want to add more Registry Store (gcp, s3). @woop ?

go/internal/infra/local.go
```diff
@@ -0,0 +1,67 @@
+package infra
```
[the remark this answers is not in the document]
ctkhanhly: I wanted to mimic the infra package in python's implementation

go/internal/feast/featurestore.go
```diff
@@ -0,0 +1,885 @@
...
+			Statuses:        make([]serving.FieldStatus, 0),
+			EventTimestamps: make([]*timestamppb.Timestamp, 0),
+		}
+	}
+	// Merge joinKeyValues into requestDataFeatures
+	for entityName, vals := range joinKeyValues {
+		requestDataFeatures[entityName] = vals
+	}
+
+	fs.populateResultRowsFromColumnar(onlineFeatureResponse, requestDataFeatures)
```
pyalex, line 183: I think it's highly inefficient, what's being done here. I understand that it was moved here from Python, but we can fix it instead of just keeping inefficient implementation.

So, what's going on here: we copy features from request to response to make them accessible by ODFV transformations. And then at the end of this long method we delete them from response (because user actually didn't ask for these features - they are input parameters). Since these features are needed only by ODFV transformations I suggest we pass them separately  to transformations only.

go/internal/feast/featurestore.go
pyalex: I'm not sure what the reason behind keeping compatibility with python. Isn't this a complete replacement for python code?

pyalex: > populateEntityRowsFromColumnar
From the code above I see that this function is used only to add request features

go/internal/feast/featurestore.go
woop: @pyalex agreed, but can we add a TODO and use a follow up PR to fix it? The scope of this PR is already massive.

go/internal/infra/local.go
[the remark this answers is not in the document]
woop: We dont need teardown. We need the minimal amount of code to read from an online store.

go/internal/infra/local.go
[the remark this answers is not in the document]
woop: for what it's worth, we should ideally follow a flat structure for our go code. What is idiomatic in Python is not idiomatic for Go. Adding folders is often not the right approach.

woop:
> Aside from the comments in line, we should avoid checking in the feature server binaries into git and have them in S3 or some place.

Why do we need the built binaries persisted at all?

go/internal/feast/featurestore.go
ctkhanhly: The flow in python is: Copy all provided entities + entities in ODFVs + requested FV schema into onlineFeatureResponse -> convert this to Dataframe -> use this dataframe copy and add/ remove column as needed and call udf to get transformed columns -> append to onlineFeatureResponse requested transformed values. Depending on how we want to implement ODFV, I could see a new flow as:

Create a separate dataframe for odfvs: this could either be like python's implementation where we merge all entities + entities in ODFVs + requested FV schema into 1 dataframe and remove/add to this dataframe object as input to udf (essentially function augmentResponseWithOnDemandTransforms which I have skipped for this PR) and append to online response - and we never add features not requested in online response.
We can also create a new dataframe consisting of all required inputs for each ODFV, although I'm not sure if we will end up increasing time due to copy multiple data multiple times instead of reusing and fixing 1 dataframe.

There is also a dummy entity that gets added/deleted, we can optimize on this as well and since I'll be rewriting groupRefs, I can make a note on the new flow and rewrite?

go/internal/feast/featurestore.go
ctkhanhly: It actually combines requested entities + entities in ODFV + request features but I reuse the hash map. I wanted to make the code easy to read and understand on the first PR so I tried to mimic python sdk as much as possible - but I'm gonna rewrite more efficiently from now...

protos/feast/core/DataFormat.proto
```diff
@@ -18,7 +18,7 @@
 syntax = "proto3";
 package feast.core;
 
-option go_package = "github.com/feast-dev/feast/sdk/go/protos/feast/core";
+option go_package = "github.com/feast-dev/feast/go/protos/feast/core";
```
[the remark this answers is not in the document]
ctkhanhly: It does since currently go module is defined inside /feast not feast/go so we need a full path

go/internal/infra/local.go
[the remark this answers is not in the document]
ctkhanhly: this is a registry store and this teardown is used to remove registry.db or remove gcp, s3 resources. Should teardown be removed from RegistryStore interface?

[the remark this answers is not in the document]
achals: We should remove it from this interface since it's not expected to be called by anyone. If we need to we can add it later. 

The go code should basically only read from the the feature store or registry - not update or destroy it IMO.

go/internal/infra/local.go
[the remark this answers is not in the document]
achals: yeah, I'm okay with everything being in the `feast` package as well. We don't need the go code structure to exactly mirror the python structure.

.github/workflows/unit_tests.yml
```diff
@@ -8,6 +8,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [ 3.7, 3.8, 3.9 ]
+        go-version: [ 1.17.7 ]
```
woop, line 11: I dont think we need a matrix. Would we ever need to use more than one version?

felixwang9817: isn't there a possibility we would want to support e.g. the last 2 minor versions of Go? when 1.18 is released, would we not want to continue supporting 1.17?

woop: Do you mean for our build process or for developers? The only breaking change will be to 2.0 as far as I can tell.

sdk/python/setup.py
```diff
@@ -172,61 +172,177 @@
...
+                + proto_files,
+                env={
+                    "PATH": self.path_val
+                }
+            )
+        except CalledProcessError as e:
+            print(f"Stderr: {e.stderr}")
+            print(f"Stdout: {e.stdout}")
+
+    def _compile_go_feaure_server(self):
```
felixwang9817, line 309: misspelled

felixwang9817 (commented):
all nonblocking comments

also note that a random `goserver` binary file accidentally got in here

.github/workflows/linter.yml
```diff
@@ -15,6 +15,11 @@ jobs:
         with:
           python-version: "3.7"
           architecture: x64
+      - name: Setup Go
+        id: setup-go
+        uses: actions/setup-go@v2
+        with:
+          go-version: "1.17.0"
```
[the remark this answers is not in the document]
achals: We should be using 1.17.7 (latest)

go/cmd/goserver/main.go
```diff
@@ -0,0 +1,77 @@
+package main
+
+import (
+	"fmt"
```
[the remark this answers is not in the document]
achals: we should do whatever gofmt picks.

Makefile
```diff
@@ -120,20 +123,25 @@ build-java-no-tests:
 # Go SDK
 
 install-go-ci-dependencies:
-	go get -u github.com/golang/protobuf/protoc-gen-go
-	go get -u golang.org/x/lint/golint
+	go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.26.0
+	go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.1.0
+
+compile-protos-go: install-go-ci-dependencies
+	pip install -e "sdk/python[ci]"
```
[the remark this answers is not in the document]
achals: Oh it's super dumb. We rely on `grpc_tools.protoc` to compile things for go as well because we do it using `python setup.py build_go_protos`. But `grpc_tools.protoc` comes from `grpcio-tools` which has to be installed by installing something from `setup.py`. 

Ideally.. we have a target to do a minimal install of the libraries needed to build feast itself.

sdk/python/setup.py
achals: d'oh

Makefile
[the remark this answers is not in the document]
achals: One of my todos is to redo the Makefile so I'll leave it as is

felixwang9817 (approved):
/lgtm

go/cmd/goserver/main.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+const (
+	flagFeastRepoPath   = "FEAST_REPO_PATH"
+	flagFeastRepoConfig = "FEAST_REPO_CONFIG"
+	feastServerVersion  = "0.18.0"
+)
+
+type FeastEnvConfig struct {
+	RepoPath   string `envconfig:"FEAST_REPO_PATH"`
```
woop, line 20: Can we please add comments here explaining exactly what these env vars are used for. Also, is there a reason to favor env vars over CLI arguments? Just curious about the reasoning.

go/internal/feast/repoconfig.go
```diff
@@ -0,0 +1,98 @@
...
+	RepoPath string `json:"repo_path"`
+}
+
+type RegistryConfig struct {
+	RegistryStoreType string `json:"registry_store_type"`
+	Path              string `json:"path"`
+	CacheTtlSeconds   int64  `json:"cache_ttl_seconds" default:"600"`
+}
+
+// NewRepoConfig reads file <repoPath>/feature_store.yaml and converts the YAML format into RepoConfig struct.
```
woop, line 38: Is this comment out of date? Where is NewRepoConfig?

go/cmd/goserver/main.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	if feastEnvConfig.RepoPath == "" && feastEnvConfig.RepoConfig == "" {
+		log.Fatalln(fmt.Sprintf("One of %s of %s environment variables must be set", flagFeastRepoPath, flagFeastRepoConfig))
+	}
+	// TODO(Ly): Review: Should we return and error here if both repoPath and repoConfigJson are set and use the cwd for NewRepoConfigFromJson?
+
+	var repoConfig *feast.RepoConfig
+
+	if len(feastEnvConfig.RepoConfig) > 0 {
+		repoConfig, err = feast.NewRepoConfigFromJson(feastEnvConfig.RepoPath, feastEnvConfig.RepoConfig)
```
woop, line 43: So is feastEnvConfig.RepoConfig a json blob? Should it be renamed?

go/cmd/goserver/main.go
```diff
@@ -0,0 +1,77 @@
...
+		log.Fatal(err)
+	}
+	if feastEnvConfig.RepoPath == "" && feastEnvConfig.RepoConfig == "" {
+		log.Fatalln(fmt.Sprintf("One of %s of %s environment variables must be set", flagFeastRepoPath, flagFeastRepoConfig))
+	}
+	// TODO(Ly): Review: Should we return and error here if both repoPath and repoConfigJson are set and use the cwd for NewRepoConfigFromJson?
+
+	var repoConfig *feast.RepoConfig
+
+	if len(feastEnvConfig.RepoConfig) > 0 {
```
woop, line 42: Why do we compare RepoConfig len to 0 here, but above we use `== ""`? Should we use one approach for consistency?

go/cmd/goserver/server.go
```diff
@@ -0,0 +1,26 @@
+package main
```
woop, line 1: It's not clear to me why we call this folder `goserver`. Seems redundant. would `go/cmd/server/...` make more sense?

go/cmd/goserver/main.go
```diff
@@ -0,0 +1,77 @@
...
+			log.Fatalln(err)
+		}
+	} else {
+		repoConfig, err = feast.NewRepoConfigFromFile(feastEnvConfig.RepoPath)
+		if err != nil {
+			log.Fatalln(err)
+		}
+	}
+
+	log.Println("Initializing feature store...")
```
woop, line 54: Nitpick, but this ellipsis seems inconsistent with other logs where we just print a message.

go/cmd/goserver/server_test.go
```diff
@@ -0,0 +1,78 @@
...
+		server.Stop()
+	}
+
+	client := serving.NewServingServiceClient(conn)
+
+	return client, closer
+}
+
+func TestGetFeastServingInfo(t *testing.T) {
+	t.Skip("@todo(achals): feature_repo isn't checked in yet")
```
woop, line 61: Why is this skipped?

go/cmd/goserver/server_test.go
```diff
@@ -0,0 +1,78 @@
...
+	ctx := context.Background()
+	client, closer := getClient(ctx)
+	defer closer()
+	response, err := client.GetFeastServingInfo(ctx, &serving.GetFeastServingInfoRequest{})
+	assert.Nil(t, err)
+	assert.Equal(t, feastServerVersion, response.Version)
+}
+
+func TestGetOnlineFeatures(t *testing.T) {
+	t.Skip("@todo(achals): feature_repo isn't checked in yet")
```
woop, line 71: Same question here.

go/internal/feast/basefeatureview.go
```diff
@@ -0,0 +1,43 @@
...
+		features[index] = NewFeatureFromProto(featureSpecV2)
+	}
+	base.features = features
+	base.projection = NewFeatureViewProjectionFromDefinition(base)
+	return base
+}
+
+func (fv *BaseFeatureView) withProjection(projection *FeatureViewProjection) (*BaseFeatureView, error) {
+	if projection.name != fv.name {
+		return nil, errors.New(fmt.Sprintf("The projection for the %s FeatureView cannot be applied because it differs in name. "+
```
woop, line 28: I've made this comment before, but can we please sweep through the errors and ensure that they are lower case and without trailing punctuation?

woop, line 28: I've made this comment before, but can we please sweep through the errors and ensure that they are lower case and without trailing punctuation?

go/internal/feast/errors.go
```diff
@@ -0,0 +1,21 @@
...
+	featureRefCollisions []string
+	fullFeatureNames     bool
+}
+
+func NewFeatureNameCollisionError(featureRefCollisions []string, fullFeatureNames bool) FeatureNameCollisionError {
+	return FeatureNameCollisionError{featureRefCollisions, fullFeatureNames}
+}
+
+func (e FeatureNameCollisionError) Error() string {
+	return fmt.Sprintf("featureNameCollisionError: %s; %t", strings.Join(e.featureRefCollisions, ", "), e.fullFeatureNames)
```
woop, line 20: Should we use plain english for the error message instead of an error name?

go/internal/feast/errors.go
```diff
@@ -0,0 +1,21 @@
+package feast
+
+import (
+	"fmt"
+	"strings"
+)
+
+// TODO: Add more errors here especially once that have been created in python sdk
```
woop, line 8: It seems like this file is meant to duplicate what we did in Python. I don't really see the value yet. NewFeatureNameCollisionError is used in a single place. Can we move this error handling logic all into featurestore.go until we absolutely need to generalize across files?

go/internal/feast/errors.go
ctkhanhly: This is needed to pass one of the integration tests that requires FeatureNameCollision exception from python sdk. I hacked around by parsing error message to coordinate between 2 languages but would there be a better coordination method I should use instead?

go/cmd/goserver/main.go
felixwang9817: I don't have a good argument for env vars over CLI arguments, but will leave as env vars for now given that no one strongly objects

am also removing this struct since it doesn't serve much use

sdk/python/feast/go_server.py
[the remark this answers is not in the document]
felixwang9817: going to resolve this since Achal as successfully added `GoServerMonitorThread` and the lifecycle test is functioning properly

go/internal/feast/errors.go
woop: I just want to prevent polluting the Golang SDK with Python specific requirements. At the very least a translation layer would be useful here to go from a Go error message to this Python one, even if it's hardcoded. It may also be a good idea to eventually introduce error codes instead of having to parse strings.

Issue #1988, opened 2021-11-01
TTL for entities in the Online stores other Than Redis


vas28r13:
**Is your feature request related to a problem? Please describe.**
(for context: we are using the Redis online store)
At the moment, once an entity is written to the Online store, it's there until it's manually cleaned up.

In most of our business cases, the Entity will have a natural expiration time (when it's no longer relevant for the use case).

It would be nice to be able to expire the entity in the online store, with some kind of TTL setting.

**Describe the solution you'd like**
1. in our local setup we end up expiring the key bin in Redis (combination of entity + project)
```
def online_write_batch(...):
  redis_key_bin = _redis_key(project, entity_key)
  ...
  if table.ttl:
      redis_client_pipeline.expire(name=redis_key_bin, time=table.ttl)
```
the issue with this is that the `ttl` param in `feature_view/table` is overloaded (initially designed for a use case in the offline store)
Also, in this type of setup and use case, the TTL should be on the entity level not `feature_view` because the Redis key bin is based on entity (feature_view ttl is designed for something else also)

so the proposal:
add `online_store_ttl` param to the Entity object
example:
```
driver_entity = Entity(
 name="driver",
 description="Driver Identifier", 
 value_type=ValueType.INT64, 
 online_store_ttl=Duration(seconds=86400)
)
```

```
def online_write_batch(...):
  redis_key_bin = _redis_key(project, entity_key)
  ...
  if entity.online_store_ttl:
      redis_client_pipeline.expire(name=redis_key_bin, time=entity.online_store_ttl)
```

Note: implementation details may vary a bit based on the Online store type

felixwang9817:
Related to #1802

r-nagaraj:
Hi Team, 
I have also observed online database (Redis online store) is not sync with latest Materializing feature views.  In short Redis online store is not truncating and reloading with latest data.  Thanks

felixwang9817:
hey @r-nagaraj, could you elaborate on the exact problem you're seeing, and how we might reproduce it? thanks!

adchia:
See some initial discussion happening in https://docs.google.com/document/d/1K8k756lgsG-YoIiEyHMOOBEejQerurhfMIABlu8Cir4/edit

PR #2341, opened 2022-02-25
feat: Key ttl setting for redis online store

vas28r13:

TTL feature for Redis online store in order to expire entity keys after some time.

Some thoughts and discussions here:
https://docs.google.com/document/d/1K8k756lgsG-YoIiEyHMOOBEejQerurhfMIABlu8Cir4/edit#heading=h.gjdgxs

Issue:
https://github.com/feast-dev/feast/issues/1988

This implementation is explicitly for the Redis Online store. In the future, this may be better as a `FeatureView` or `Entity` ttl param but it would need some more development to be able to handle across different OnlineStore providers.

```release-note
Python Redis online store allows TTL expiration on entity keys
```

### DIFF
--- a/sdk/python/feast/infra/online_stores/redis.py
+++ b/sdk/python/feast/infra/online_stores/redis.py
@@ -68,6 +68,9 @@ class RedisOnlineStoreConfig(FeastConfigBaseModel):
     """Connection string containing the host, port, and configuration parameters for Redis
      format: host:port,parameter1,parameter2 eg. redis:6379,db=0 """
 
+    key_ttl_seconds: int = None
+    """(Optional) redis key bin ttl (in seconds) for expiring entities"""
+
 
 class RedisOnlineStore(OnlineStore):
     _client: Optional[Union[Redis, RedisCluster]] = None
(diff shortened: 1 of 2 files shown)


sdk/python/feast/infra/online_stores/redis.py
```diff
@@ -68,6 +68,9 @@ class RedisOnlineStoreConfig(FeastConfigBaseModel):
     """Connection string containing the host, port, and configuration parameters for Redis
      format: host:port,parameter1,parameter2 eg. redis:6379,db=0 """
 
+    key_ttl_seconds: int = None
```
adchia, line 71: nit: mark the int as optional too then? i.e. Optional[int]

vas28r13: ah yeah good point!

adchia:
If you rebase against master now I think the unit tests should pass now?

adchia (approved):
/lgtm

adchia (approved):
/lgtm

PR #2342, opened 2022-02-26
ci: Capping the dask dependency since Dask dropped Python 3.7 support

adchia:
See https://github.com/dask/dask/pull/8572 which dropped Python 3.7 support in Dask for the last 2 releases.

Fixes #

```release-note
NONE
```

### DIFF
--- a/sdk/python/setup.py
+++ b/sdk/python/setup.py
@@ -40,7 +40,7 @@ AUTHOR = "Feast"
 REQUIRES_PYTHON = ">=3.7.0"
 
 REQUIRED = [
-    "Click==8.*",
+    "Click>=7.*",
     "colorama>=0.3.9",
     "dill==0.3.*",
     "fastavro>=1.1.0",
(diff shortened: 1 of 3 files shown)


sdk/python/setup.py
```diff
@@ -40,7 +40,7 @@
 REQUIRES_PYTHON = ">=3.7.0"
 
 REQUIRED = [
-    "Click==8.*",
+    "Click>=7.*",
```
achals, line 43: Was this necessary for the same reason (python 3.7 support)?

adchia: ah no this was just a fix i wanted to sneak in from a previous PR because it seemed overly restrictive

achals: sg. stamping.

achals (approved):
/lgtm

sdk/python/setup.py
tpvasconcelos: @adchia @achals I think this was a typo as pip actually interprets this as `click==7.*`. See https://github.com/feast-dev/feast/pull/2450

PR #2344, opened 2022-02-27
fix: Method _should_use_plan only returns true for local sqlite provider

nossrannug:
Only the SQLite local provider supports _plan and _apply_diffs. Updated
the _should_use_plan method to only return true for that setup so that other local providers don't need to implement these methods right now.

Only the SQLite provider supports _plan and _apply_diffs today. Before this can be added to other providers there needs to be some discussion around the InfraObject and how the update and teardown is performed.

My main worry at the moment is that the InfraObject can be constructed from a proto and then the update method can be called. This requires all the connection information and credentials to be present in the infra object proto.

```release-note
NONE
```

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -392,8 +392,10 @@ class FeatureStore:
 
     def _should_use_plan(self):
         """Returns True if _plan and _apply_diffs should be used, False otherwise."""
-        # Currently only the local provider supports _plan and _apply_diffs.
-        return self.config.provider == "local"
+        # Currently only the local provider with sqlite online store supports _plan and _apply_diffs.
+        return self.config.provider == "local" and (
+            self.config.online_store and self.config.online_store.type == "sqlite"
+        )
 
     def _validate_all_feature_views(
         self,


felixwang9817 (approved):
/lgtm

felixwang9817:
oops missed a small bug: looks like you need to do a null check on `online_store` here:
```
self.config.online_store.type == "sqlite"
```
integration tests are failing

separately, we recently moved to [semantic release](https://semantic-release.gitbook.io/semantic-release/), so all PR titles now need to prefixed with e.g. "refactor" or "fix" depending on the type of PR - you can see some other PRs for an example, and the set of possible prefixes is documented [here](https://github.com/angular/angular/blob/master/CONTRIBUTING.md#type)

nossrannug:
> @nossrannug oops missed a small bug: looks like you need to do a null check on `online_store` here:
> 
> ```
> self.config.online_store.type == "sqlite"
> ```
> 
> integration tests are failing
> 
> separately, we recently moved to [semantic release](https://semantic-release.gitbook.io/semantic-release/), so all PR titles now need to prefixed with e.g. "refactor" or "fix" depending on the type of PR - you can see some other PRs for an example, and the set of possible prefixes is documented [here](https://github.com/angular/angular/blob/master/CONTRIBUTING.md#type)

Thanks @felixwang9817, I'm sorry I missed that. I've added the null check and updated the title.

felixwang9817 (approved):
/lgtm

PR #2347, opened 2022-02-28
fix: Rollback Redis-py to Redis-py-cluster

kevjumba:
We want to rollback redis-py cluster usage due to this bug. Also actually add redis py cluster as an online store for the integration tests to make sure it is a valid online store. 

Also fixes issues with redis cluster on the integration tests.
https://github.com/redis/redis-py/issues/2003

Fixes #

```release-note
NONE
```


.gitignore
```diff
@@ -107,6 +107,7 @@ coverage.xml
 *.cover
 .hypothesis/
 .pytest_cache/
+infra/scripts/*.conf
```
achals, line 110: Sorry, why is this being added?

sdk/python/tests/conftest.py
```diff
@@ -22,6 +22,9 @@
 import pandas as pd
 import pytest
 from _pytest.nodes import Item
+from sdk.python.tests.integration.feature_repos.repo_configuration import (
```
achals, line 25: Shouldn't this be `from tests.integration.feature_repos.repo_configuration`?

achals (changes requested):

.gitignore
kevjumba: My script for creating a cluster generates conf files as per redis configuration. ./create-cluster.sh clean should remove all of them but just in case people forget, I just added this line to make sure it's not tracked by git.

sdk/python/tests/conftest.py
kevjumba: Fixed.

kevjumba:
@felixwang9817 PTAL! The code coverage is failing for some odd reason but only for a subset of the tests => may be flaky github actions connection?

woop:
Why is this a `feat` and not a `bug`?

sdk/python/requirements/py3.7-ci-requirements.txt
```diff
@@ -1,8 +1,8 @@
 #
-# This file is autogenerated by pip-compile with python 3.7
+# This file is autogenerated by pip-compile with python 3.8
```
felixwang9817, line 2: this seems wrong, should be 3.7?

sdk/python/requirements/py3.9-ci-requirements.txt
```diff
@@ -1,8 +1,8 @@
 #
-# This file is autogenerated by pip-compile with python 3.9
+# This file is autogenerated by pip-compile with python 3.8
```
felixwang9817, line 2: this also seems wrong, should be 3.9?

felixwang9817 (commented):
just a few qq

sdk/python/setup.py
```diff
@@ -78,7 +78,8 @@
 ]
 
 REDIS_REQUIRED = [
-    "redis>=4.1.0",
+    "redis>=3.5.3",
```
achals, line 81: If we're downgrading `redis`, should we pin to a max allowed version?

kevjumba: I believe redis-py-cluster ensures that, should I still pin the version, this is autogenerated because redis-py-cluster requires something before redis-py 4.

sdk/python/requirements/py3.7-ci-requirements.txt
kevjumba: Right, I generated it wrong. Was wondering how to generate for each individual python version, then I realized I can just create another venv.

sdk/python/requirements/py3.9-ci-requirements.txt
kevjumba: Fixed regenerated with python3.9

sdk/python/setup.py
achals: I think we should still pin the version since we're directly relying on that pin in our logic.

woop:
> Why is this a `feat` and not a `fix`?

Just bumping this since I didn't get a response.

felixwang9817 (approved):
/lgtm

achals (approved):
/lgtm

PR #2350, opened 2022-03-01
fix: Update github workflow to prevent redis from overlapping ports.

kevjumba:

We want to point redis client at a port other than 6379 and point the redis cluster at 6379. Therefore I'm starting another redis client to make sure the two don't intersect. 
Redis single node clients are somehow connecting to the redis cluster node and the intersection is causing issues. Therefore, this PR starts up another redis node that we can connect to for the tests. 

Left the redis image in just in case we need to revert this change in the event of breaking changes.

Fixes #

```release-note
NONE
```


felixwang9817 (approved):
/lgtm

PR #2354, opened 2022-03-02
fix: Remove redis service to prevent more conflicts and add redis node to master_only

kevjumba:

There is still redis client of some sort and it may still be because of this random excess client so I'm going to remove it.
It is not used anymore anyway since the current build is stable in master using 12345 as the port. 

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/master_only.yml
+++ b/.github/workflows/master_only.yml
@@ -65,16 +65,6 @@ jobs:
     env:
       OS: ${{ matrix.os }}
       PYTHON: ${{ matrix.python-version }}
-    services:
-      redis:
-        image: redis
-        ports:
-          - 6379:6379
-        options: >-
-          --health-cmd "redis-cli ping"
-          --health-interval 10s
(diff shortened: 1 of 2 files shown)


felixwang9817 (approved):
/lgtm

.github/workflows/master_only.yml
```diff
@@ -120,6 +110,11 @@ jobs:
         run: pip install pip-tools
       - name: Install dependencies
         run: make install-python-ci-dependencies
+      - name: Start Redis
+        uses: supercharge/redis-github-action@1.4.0
+        with:
+          redis-version: ${{ matrix.redis-version }}
+          redis-port: 12345
```
achals, line 117: Are the tests configured to use this port for redis?

kevjumba: Yup, redis in master is using the ports but we didn't update mater_only.yml so we need to do that for integration tests in master to work.

PR #2355, opened 2022-03-03
feat: Event timestamps response

vas28r13:
```release-note
* ability to pull event timestamps from OnlineResponse
* fix small event timestamps bugs (proto reference bug + potential timezone bug)
```

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -1508,10 +1508,10 @@ class FeatureStore:
 
         # Each row is a set of features for a given entity key. We only need to convert
         # the data to Protobuf once.
-        row_ts_proto = Timestamp()
         null_value = Value()
         read_row_protos = []
         for read_row in read_rows:
+            row_ts_proto = Timestamp()
             row_ts, feature_data = read_row
             if row_ts is not None:
                 row_ts_proto.FromDatetime(row_ts)
--- a/sdk/python/feast/online_response.py
+++ b/sdk/python/feast/online_response.py
@@ -20,6 +20,8 @@ from feast.feature_view import DUMMY_ENTITY_ID
 from feast.protos.feast.serving.ServingService_pb2 import GetOnlineFeaturesResponse
 from feast.type_map import feast_value_type_to_python_type
 
+TIMESTAMP_POSTFIX: str = "__ts"
+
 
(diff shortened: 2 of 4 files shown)


sdk/python/feast/feature_store.py
```diff
@@ -1508,10 +1508,10 @@ def _read_from_online_store(
 
         # Each row is a set of features for a given entity key. We only need to convert
         # the data to Protobuf once.
-        row_ts_proto = Timestamp()
```
vas28r13, line 1511: since the same object is referenced in the list for each entity row then the last assignment is what gets set for all rows so we need a new proto object for each row

sdk/python/feast/online_response.py
```diff
@@ -66,3 +66,28 @@ def to_df(self) -> pd.DataFrame:
...
+    def event_timestamps_dict(self) -> Dict[str, List[int]]:
+        """
+        Converts GetOnlineFeaturesResponse feature event timestamps into a dictionary form.
+        Includes the entity id
+        """
+        response: Dict[str, List[int]] = {}
+
+        for result in self.proto.results:
+            for idx, feature_ref in enumerate(self.proto.metadata.feature_names.val):
+                # obtain the entity id
```
adchia, line 79: This seems to rely on an implementation detail (that isn't enforced by the proto schema itself) that we have entities first and then the feature values (in https://github.com/feast-dev/feast/blob/master/sdk/python/feast/feature_store.py#L1269-L1269)

On top of that, I think this becomes more broken when you have multiple join keys? Seems like ideally you'd actually somehow check the feature_ref and see if it's an entity.

Alternatively, I'd just output both timestamp + values for each. That would make this simpler and not make assumptions.

sdk/python/feast/online_response.py
```diff
@@ -44,9 +46,12 @@ def __init__(self, online_response_proto: GetOnlineFeaturesResponse):
                     del result.event_timestamps[idx]
                 break
 
-    def to_dict(self) -> Dict[str, Any]:
+    def to_dict(self, is_with_event_timestamps: bool = False) -> Dict[str, Any]:
```
adchia, line 49: maybe include_event_timestamps instead?

adchia (approved):
/lgtm

PR #2371, opened 2022-03-06
feat: Add support for DynamoDB online_read in batches

### LINKED ISSUE #2247: DynamoDB batch retrieval does slow sequential gets (online_read)
The current DynamoDB implementation does sequential gets (https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/online_stores/dynamodb.py#L163)

## Possible Solution
A better approach is to do some multi-get operation or at least run these queries in parallel and collect the results.

TremaMiguel:

Fixes #2247 

**Context**

Given a list of `entity_ids` set a `batch_size` parameter to determine the number of items to send in a `batch_get_item` request to DynamoDB.

DynamoDB can retrieve up to 16MB of data and the record size limit is 400kb, `batch_size` value could be at most 40.


TremaMiguel:
@felixwang9817 @vlin-lgtm I didn't found any tests for `DynamoDBOnlineStore` methods. I'm thinking about adding some using [moto](https://github.com/spulec/moto/blob/master/tests/test_dynamodb2/test_dynamodb.py#L2454).

woop:
@TremaMiguel thanks for this. Would you mind changing your PR title to some kind of explaination of what the PR does? Something like "Add support for DynamoDB reads and writes in batches". This will become a part of our changelog.

adchia:
Another thing is to not use git pull and instead use eg `git pull --rebase ` (you also should rebase changes against master again since there were test breakages that now are fixed)

as far as tests go, technically there is a sets of tests that call into dynamo (test_universal_online for example). Though I don't think there's a test that does a batch fetch. 

moto looks interesting though. might be a good idea to use that so we can let contributors run tests locally for aws. would love to see how that looks.

TremaMiguel:
@adchia added a test for `online_read` method parametrized to test 5, 50 or 100 samples at a time. 

To run these test AWS credential shoud be set 

```bash
export AWS_ACCESS_KEY_ID=your_access_key_id
export AWS_SECRET_ACCESS_KEY=your_secret_access_key
```

I think we can open an issue to improve tests for `DynamoDBOnlineStore`, especially `online_write_batch`.

TremaMiguel:
@adchia I've changed tests to run dynamodb behavior through `moto`, no need to set AWS credentials. 

Some tests are failing due to #2376

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -188,6 +188,7 @@ def online_read(
         table: FeatureView,
         entity_keys: List[EntityKeyProto],
         requested_features: Optional[List[str]] = None,
+        batch_size: int = 20,
```
vlin-lgtm, line 191: why do we even want to let caller pass this parameter? why not always use batch_size of 40?

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -199,26 +200,59 @@ def online_read(
             config: The RepoConfig for the current FeatureStore.
             table: Feast FeatureView.
             entity_keys: a list of entity keys that should be read from the FeatureStore.
+            batch_size: the number of items to send in a batch_get_item request to DynamoDB.
+                DynamoDB record size limit is 400kb and can retrieve 16MB per call, it is recommended
+                to set batch_size value less than 40 to avoid ``UnprocessedKeys`` and
+                ``ValidationException`` errors.
         """
+        if batch_size > 40:
```
vlin-lgtm, line 208: if we don't let user specify batch_size, then this is not needed.

vlin-lgtm (commented):
Can we hide `batch_size` from users? So something like this can work: 
```
batch_size = 40
entity_ids = ...
batches = []
for i in range(0, len(entity_ids), batch_size): 
  batches.append(entity_ids[i:i + batch_size])

for batch in batches:
  batch_entity_ids = ...
  ...
  ...
  ...
  response = dynamodb_resource.batch_get_item(RequestItems=batch_entity_ids)
  ...
```
So this always tries to send at most 40 entities at a time. It is better to maximize the batch size as much as possible so there is less network calls to DynamoDB.

sdk/python/feast/infra/online_stores/dynamodb.py
TremaMiguel: @vlin-lgtm setting `batch_size` at 40 might be a good idea to minimize overall latency but I'm more inclined to give that control to the user to specify the number of items to send in a `batch_get_item`.

TremaMiguel: @vlin-lgtm do you with how many `ReadCapacityUnits` is the DynamoDB table configured when created?

adchia: This does seem weird since it no longer follows the online store contract. 

Would it be sufficient to add this batch size to the DynamoDBOnlineStoreConfig (aka via feature_store.yaml)?

vlin-lgtm: > @vlin-lgtm do you with how many ReadCapacityUnits is the DynamoDB table configured when created?

I believe tables are created with "on-demand" RCU and WCU rather than with provisioned with some initial units. This can be changed by users at a later time.

vlin-lgtm: > I'm more inclined to give that control to the user to specify the number of items to send in a batch_get_item.

What are the trade-offs? If the worry is RCU, then sending this in different batch sizes hardly changes anything, it will most likely consume the same RCU. In fact, even without this PR, it would probably consume the same RCU today.

vlin-lgtm: FWIW, AWS' DynamoDB Java SDK hardcode a batch_size of 100 [here](https://github.com/willglynn/aws-sdk-for-java/blob/master/src/main/java/com/amazonaws/services/dynamodb/datamodeling/DynamoDBMapper.java#L878). Also, as you pointed out in [AWS' doc](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchGetItem.html) that 
> If you request more than 100 items, BatchGetItem returns a ValidationException with the message "Too many items requested for the BatchGetItem call."

So I think 40 makes sense because DDB's record size limit is 400KB.

TremaMiguel: @vlin-lgtm ok, will set the default `batch_size` to 40 without the user specifying this value.

sdk/python/feast/infra/online_stores/dynamodb.py
TremaMiguel: got it

sdk/python/feast/infra/online_stores/dynamodb.py
TremaMiguel: > Would it be sufficient to add this batch size to the DynamoDBOnlineStoreConfig (aka via feature_store.yaml)?

do you mean setting `batch_size` as a general configuration to interact with DynamoDB? Can the user modify these values?

adchia: Depends on what you mean by user. But yeah anybody who is setting up feast and calling get_online_features would have their own `feature_store.yaml`. Then you'd have a yaml file like

```yaml
project: my_project
registry: data/registry.db
provider: aws
online_store:
  type: dynamodb
  region: %AWS_REGION%
  batch_size: 40
```

which would populate the `DynamoDBOnlineStoreConfig` in this class

TremaMiguel:
@adchia could you provide guidance on how to run integration tests locally for AWS? It appears there's something missing in `CONTRIBUTING.md` 

> AWS
TODO(adchia): flesh out setting up AWS login (or create helper script)
Modify RedshiftDataSourceCreator to use your credentials

adchia:
> @adchia could you provide guidance on how to run integration tests locally for AWS? It appears there's something missing in `CONTRIBUTING.md`
> 
> > AWS
> > TODO(adchia): flesh out setting up AWS login (or create helper script)
> > Modify RedshiftDataSourceCreator to use your credentials

Hey!

So what I'd actually do is make a new repo_configuration that uses a local offline store + dynamo as the online store. That's configured at https://github.com/feast-dev/feast/blob/master/sdk/python/tests/integration/feature_repos/repo_configuration.py#L88-L88

You can basically just ignore all these and set 
`DEFAULT_FULL_REPO_CONFIGS = [IntegrationTestRepoConfig(online_store=DYNAMO_CONFIG)]`. Then as long as you're properly authenticated to be able to create / teardown new dynamo tables, then tests should work fine.

Normally to run everything, you'd run `make test-python-universal`, but in this case the only test failures are in test_universal_online, so you can just do:
```bash
FEAST_USAGE=False IS_TEST=True python -m pytest -n 8 --integration --universal sdk/python/tests/integration/online_store/
```

TremaMiguel:
@adchia some integrations tests expect results from DynamoDB query to have the same order as input `entity_keys`, I was reading that `BatchGetItem` [don't return results in order](https://stackoverflow.com/questions/56544496/dynamodb-batchget-get-results-in-same-order-as-provided-keys).

I added a `sort_response` parameter that defaults to `True` to sort DynamoDB results in the same order as the input `entity_keys`

Can you suggest something to improve the implementation of `_sort_dynamodb_response` method.

TremaMiguel:
@adchia [test_online_store_cleanup](https://github.com/feast-dev/feast/blob/master/sdk/python/tests/integration/online_store/test_universal_online.py) with the case `Redshift` as offline store and `DynamoDB` as online store. 

After doing some digging, there's a difference in the result of `fs.get_online_features` operation. 

1. The first one is to [check if features are available](https://github.com/feast-dev/feast/blob/master/sdk/python/tests/integration/online_store/test_universal_online.py#L684), returning `driver_id` and `value` as results values.

```python
metadata {
  feature_names {
    val: "driver_id"
    val: "value"
  }
}
results {
  values {
    int64_val: 5001
  }
  values {
    float_val: 0.03799890726804733
  }
  statuses: PRESENT
  statuses: PRESENT
  event_timestamps {
  }
  event_timestamps {
    seconds: 1647104400
  }
}
results {
  values {
    int64_val: 5002
  }
  values {
    float_val: 0.8108303546905518
  }
  statuses: PRESENT
  statuses: PRESENT
  event_timestamps {
  }
  event_timestamps {
    seconds: 1647104400
  }
}
```

2. Second case check if [features for other are available](https://github.com/feast-dev/feast/blob/master/sdk/python/tests/integration/online_store/test_universal_online.py#L693) after one `FeatureView` is deleted, results are similar as the previous one. 

3. This one is different, the third case [verify that features for other features are deleted](https://github.com/feast-dev/feast/blob/master/sdk/python/tests/integration/online_store/test_universal_online.py#L712), the result from `fs.get_online_features` is missing the values from the `"value"` feature

```python
metadata {
  feature_names {
    val: "driver_id"
    val: "value"
  }
}
results {
  values {
    int64_val: 5001
  }
  values {
  }
  statuses: PRESENT
  statuses: NOT_FOUND
  event_timestamps {
  }
  event_timestamps {
  }
}
results {
  values {
    int64_val: 5002
  }
  statuses: PRESENT
  event_timestamps {
  }
}
results {
  values {
    int64_val: 5003
  }
  statuses: PRESENT
  event_timestamps {
  }
}
```

do you have an idea why is this happening?

sdk/python/feast/online_response.py
```diff
@@ -57,7 +69,12 @@ def to_dict(self, include_event_timestamps: bool = False) -> Dict[str, Any]:
 
         for result in self.proto.results:
             for idx, feature_ref in enumerate(self.proto.metadata.feature_names.val):
-                native_type_value = feast_value_type_to_python_type(result.values[idx])
+                try:
+                    native_type_value = feast_value_type_to_python_type(result.values[idx])
+                except IndexError as e:
+                    logger.info(f'ERROR -- {e}')
+                    logger.info(f'ONLINE RESPONSE PROTO: {self.proto}')
```
woop, line 76: Shouldn't this be `DEBUB`? `INFO` could introduce a lot of latency.

TremaMiguel: @woop  sorry this has just the purpose to show the source of error from the integration test [test_online_store_cleanup](https://github.com/feast-dev/feast/blob/master/sdk/python/tests/integration/online_store/test_universal_online.py) for `AWS:Redshift:dynamodb`.


> @adchia [test_online_store_cleanup](https://github.com/feast-dev/feast/blob/master/sdk/python/tests/integration/online_store/test_universal_online.py) with the case `Redshift` as offline store and `DynamoDB` as online store fails.
> 
> After doing some digging, there's a difference in the result of `fs.get_online_features` operation.
> 


Any thoughts why the results from `fs.get_online_features` are different?

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -207,28 +208,59 @@ def online_read(
...
+        )
+        for i in range(iters):
+            start_index = min(i * batch_size, len_entity_ids)
+            end_index = min(i * batch_size + batch_size, len_entity_ids)
+
+            batch_entity_ids = {
+                table_instance.name: {
+                    "Keys": [
+                        {"entity_id": entity_id}
+                        for entity_id in entity_ids[start_index:end_index]
```
adchia, line 239: seems like this might be off by one? Would capture up to `batch_size - 1` in each iteration i think since slices are upper bound exclusive.

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -196,6 +196,7 @@ def online_read(
         table: FeatureView,
         entity_keys: List[EntityKeyProto],
         requested_features: Optional[List[str]] = None,
+        sort_response: bool = True,
```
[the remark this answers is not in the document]
TremaMiguel: agree

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -207,28 +208,59 @@ def online_read(
...
         result: List[Tuple[Optional[datetime], Optional[Dict[str, ValueProto]]]] = []
-        for entity_key in entity_keys:
-            table_instance = dynamodb_resource.Table(
-                _get_table_name(online_config, config, table)
-            )
-            entity_id = compute_entity_id(entity_key)
+        entity_ids = [compute_entity_id(entity_key) for entity_key in entity_keys]
+
+        len_entity_ids = len(entity_ids)
+        batch_size = 40
```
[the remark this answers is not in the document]
TremaMiguel: yes, it can be an attribute from `DynamoDBOnlineStore`,

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -207,28 +208,59 @@ def online_read(
...
+        # Iterate until the end_index is the value len_entity_ids
+        iters = (
```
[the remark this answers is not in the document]
TremaMiguel: makes sense in this case, in this case break if not more items to insert 

```python
while True:
      batch = list(itertools.islice(entity_ids_iter, batch_size))
      # No more items to insert
      if len(batch) == 0:
          break
```

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -207,28 +208,59 @@ def online_read(
...
+                response = dynamodb_resource.batch_get_item(
+                    RequestItems=batch_entity_ids
+                )
+
+            response = response.get("Responses")
+            table_responses = response.get(table_instance.name)
+
+            if table_responses:
+                if sort_response:
+                    table_responses = self._sort_dynamodb_response(
```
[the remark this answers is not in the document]
TremaMiguel: well, yes and no, we just need to sort the response from `get_online_features`, but I'll do it a separate issue as they might be many tests cases retrieving online features.

```python
online_features = fs.get_online_features(features=features, entity_rows=entity_rows)
# Sort this before calling assert
online_features = online_features.to_dict()
```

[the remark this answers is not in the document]
TremaMiguel: @adchia actually, [read_from_online_store](https://github.com/feast-dev/feast/blob/e17028dbb90cb88d9df9a9769abbd90c5c301563/sdk/python/feast/feature_store.py#L1563) states

> This method assumes that `provider.online_read` returns data for each
            combination of Entities in `entity_rows` in the same order as they
            are provided.

TremaMiguel:
@adchia just to follow up the conversations, is there something you'd like to add?

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -50,10 +51,13 @@ class DynamoDBOnlineStoreConfig(FeastConfigBaseModel):
...
     region: StrictStr
-    """ AWS Region Name """
+    """AWS Region Name"""
 
     table_name_template: StrictStr = "{project}.{table_name}"
-    """ DynamoDB table name template """
+    """DynamoDB table name template"""
+
+    sort_response: bool = True
+    """Wether or not to sort BatchGetItem response."""
```
woop, line 60: typo

TremaMiguel: thanks for reviewing this @woop , what do you mean with?

> remove the training full stop for consistency

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -50,10 +51,13 @@ class DynamoDBOnlineStoreConfig(FeastConfigBaseModel):
...
     region: StrictStr
-    """ AWS Region Name """
+    """AWS Region Name"""
 
     table_name_template: StrictStr = "{project}.{table_name}"
-    """ DynamoDB table name template """
+    """DynamoDB table name template"""
+
+    sort_response: bool = True
+    """Whether or not to sort BatchGetItem response."""
```
adchia, line 60: one final nit: i'd have the batch_size be a parameter here in teh online store config so that users can have some control. e.g. if they expect items to be very large, they may want to lower the batch size

adchia (approved):

sdk/python/feast/infra/online_stores/dynamodb.py
TremaMiguel: @adchia agree, user can specify the `batch_size` value defaulting to 40 in `DynamoDBOnlineStoreConfig`

```python
batch_size: int = 40
"""Number of items to retrieve in a DynamoDB BatchGetItem call."""
```

adchia (approved):
/lgtm

vlin-lgtm:
I'm late in the game, but I'd like to thank you @TremaMiguel very much for contributing this! This is going improve our feature serving latency by a great deal! ❤️

TremaMiguel:
Thanks @vlin-lgtm, I believe there's room for improvement related to unit testing the DynamoDb Online store module, we could use something like moto for this, as we've done with the [online_read](https://github.com/feast-dev/feast/blob/master/sdk/python/tests/unit/online_store/test_dynamodb_online_store.py#L43) method.

Having these unitary tests plus the local integration tests would definitively help make the code more robust and facilitate future changes.

PR #2379, opened 2022-03-07
fix: Making a name for data sources not a breaking change

adchia:
https://github.com/feast-dev/feast/pull/2336 introduced a change that would be breaking (forcing the first parameter of a DataSource to be a name). This pushes that to the end as an optional parameter and also outputs a warning for a future version where we require a name if users use a query in a data source.

Fixes #

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery_source.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery_source.py
@@ -3,7 +3,7 @@ from typing import Callable, Dict, Iterable, List, Optional, Tuple
 
 from feast import type_map
 from feast.data_source import DataSource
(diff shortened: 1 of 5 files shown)


achals (approved):
/lgtm

achals (approved):
/lgtm

roelschr:
Thanks for this @adchia, I was just finishing a new custom store when I found that newer versions of Feast are requiring data sources to have a name. Do you think this will be available for version 0.19.3 and onwards?

adchia:
@roelschr yes should be released soon! though i don't know if this will resolve issues for a custom offline store

roelschr:
> @roelschr yes should be released soon! though i don't know if this will resolve issues for a custom offline store

It does! In our case, it was only to make it easier for backward compatibility.  Thanks!

PR #2381, opened 2022-03-07
fix: Fixing Spark template to include source name

adchia:
The Spark bootstrap template doesn't include a name for data sources that use paths/queries, which is unsupported behavior for this new source

Fixes #


felixwang9817 (approved):
/lgtm

achals (approved):
/lgtm

PR #2385, opened 2022-03-08
ci: Have integration tests also run when a PR is LGTM'd

adchia:
Existing integration tests only run on approval or ok-to-test, when in practice often the last label is actually lgtm

Fixes #


felixwang9817 (approved):
/lgtm

PR #2386, opened 2022-03-08
fix: Properly close threadpool resources

ptoman-pa:
Properly closes threadpool resources.

Without this PR, large materializations produce `RuntimeError: can't start new thread`, because `threading.active_count()` infinitely increases toward system limit.  With properly closed resources, `threading.active_count()` stays low.

No open issue.


achals (approved):
/lgtm

PR #2389, opened 2022-03-09
fix: Added Redshift and Spark typecheck to data_source event_timestamp_col inference

shaibruhis:
`RedshiftSource` is a supported `data_source`, but the `update_data_sources_with_inferred_event_timestamp_col()` method omits it as a valid value in the assertion. This was causing the assertion to fail when using a `RedshiftSource`

Fixes #

### DIFF
--- a/sdk/python/feast/inference.py
+++ b/sdk/python/feast/inference.py
@@ -45,7 +45,7 @@ def update_entities_with_inferred_types_from_feature_views(
 
                 # get entity information from information extracted from the view batch source
                 extracted_entity_name_type_pairs = list(
-                    filter(lambda tup: tup[0] == entity.join_key, col_names_and_types,)
+                    filter(lambda tup: tup[0] == entity.join_key,col_names_and_types,)
                 )
                 if len(extracted_entity_name_type_pairs) == 0:
                     # Doesn't mention inference error because would also be an error without inferencing
@@ -54,8 +54,8 @@ def update_entities_with_inferred_types_from_feature_views(
                         its entity's name."""
                     )
 
-                inferred_value_type = view.batch_source.source_datatype_to_feast_value_type()(
-                    extracted_entity_name_type_pairs[0][1]
+                inferred_value_type = (
+                    view.batch_source.source_datatype_to_feast_value_type()(extracted_entity_name_type_pairs[0][1])
                 )
 
                 if (
@@ -111,6 +111,7 @@ def update_data_sources_with_inferred_event_timestamp_col(
             assert (
                 isinstance(data_source, FileSource)
                 or isinstance(data_source, BigQuerySource)
+                or isinstance(data_source, RedshiftSource)
                 or isinstance(data_source, SnowflakeSource)
             )
 
--- a/sdk/python/feast/inference.py
+++ b/sdk/python/feast/inference.py
@@ -45,7 +45,7 @@ def update_entities_with_inferred_types_from_feature_views(
 
                 # get entity information from information extracted from the view batch source
                 extracted_entity_name_type_pairs = list(
-                    filter(lambda tup: tup[0] == entity.join_key,col_names_and_types,)
+                    filter(lambda tup: tup[0] == entity.join_key, col_names_and_types,)
                 )
                 if len(extracted_entity_name_type_pairs) == 0:
                     # Doesn't mention inference error because would also be an error without inferencing
@@ -54,8 +54,8 @@ def update_entities_with_inferred_types_from_feature_views(
                         its entity's name."""
                     )
 
-                inferred_value_type = (
-                    view.batch_source.source_datatype_to_feast_value_type()(extracted_entity_name_type_pairs[0][1])
+                inferred_value_type = view.batch_source.source_datatype_to_feast_value_type()(
+                    extracted_entity_name_type_pairs[0][1]
                 )
 
                 if (
--- a/sdk/python/feast/inference.py
+++ b/sdk/python/feast/inference.py
@@ -113,6 +113,7 @@ def update_data_sources_with_inferred_event_timestamp_col(
                 or isinstance(data_source, BigQuerySource)
                 or isinstance(data_source, RedshiftSource)
                 or isinstance(data_source, SnowflakeSource)
+                or "SparkSource" == data_source.__class__.__name__
             )
 
             # loop through table columns to find singular match
--- a/sdk/python/tests/integration/registration/test_inference.py
+++ b/sdk/python/tests/integration/registration/test_inference.py
@@ -1,3 +1,4 @@
+from copy import deepcopy
 import pandas as pd
 import pytest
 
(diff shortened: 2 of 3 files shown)


adchia:
Thanks for catching and making a PR!

Mind updating to master? Also mind adding a test e.g. at https://github.com/feast-dev/feast/blob/master/sdk/python/tests/integration/registration/test_inference.py#L87-L87  ?

shaibruhis:
Hey Danny! Not sure what you mean by "updating to master", I set the base to master?

Also, the test you listed already checks the `RedshiftSource`. I think the right test here is to add a `RedshiftSource` in here: https://github.com/feast-dev/feast/blob/master/sdk/python/tests/integration/registration/test_inference.py#L119-L123. I saw there is a utils class `RedshiftDataSourceCreator` but curious if there is a test Redshift data source?

adchia:
> Hey Danny! Not sure what you mean by "updating to master", I set the base to master?
> 
> Also, the test you listed already checks the `RedshiftSource`. I think the right test here is to add a `RedshiftSource` in here: https://github.com/feast-dev/feast/blob/master/sdk/python/tests/integration/registration/test_inference.py#L119-L123. I saw there is a utils class `RedshiftDataSourceCreator` but curious if there is a test Redshift data source?

Ah sorry I was unclear. I meant pulling the latest changes from master.

With the test, I accidentally referenced a specific line there, but was referring to the overall file. There is a test Redshift data source available actually. You'll want to probably pass in the test fixtures we generate (e.g. https://github.com/feast-dev/feast/blob/master/sdk/python/tests/integration/feature_repos/repo_configuration.py#L222-L222 are test sources we generate) -> an example of where we pull this is in [test_historical_features](https://github.com/feast-dev/feast/blob/6fc0867f86cc67d4ee767d6a84763a51cecc2dcb/sdk/python/tests/integration/offline_store/test_universal_historical_retrieval.py#L281-L281) where it passes in the `universal_data_sources` test fixture. That gives you access to the data sources defined. 

Let me know if you have any questions!

sdk/python/feast/inference.py
```diff
@@ -111,6 +111,7 @@ def update_data_sources_with_inferred_event_timestamp_col(
             assert (
                 isinstance(data_source, FileSource)
                 or isinstance(data_source, BigQuerySource)
+                or isinstance(data_source, RedshiftSource)
```
adchia, line 114: would you mind also adding the ""SparkSource" == data_source.__class__.__name__" line here so our spark source passes this too? (and updating the PR title)

shaibruhis: is there a reason we explicitly check the class name rather than doing `isinstance(data_source, SparkSource)`?

adchia: yeah main reason is Spark is right now still a contrib plugin that requires an extra. If we use the explicit check, then that will break other users who don't install spark.

sdk/python/tests/integration/registration/test_inference.py
```diff
@@ -137,6 +139,16 @@ def test_update_data_sources_with_inferred_event_timestamp_col(simple_dataset_1)
                 [file_source], RepoConfig(provider="local", project="test")
             )
 
+    (_, _, data_sources) = universal_data_sources
```
shaibruhis, line 142: I looked through how universal_data_sources is created and the source type is defined by the environment -- https://github.com/feast-dev/feast/blob/master/sdk/python/tests/integration/feature_repos/repo_configuration.py#L416 but im not sure where/how to set that to redshift.

adchia: you shouldn't need to set to redshift. This will check across bigquery + snowflake + redshift as well.

adchia: could you split this into a separate test btw? as is it'll actually run the test before this for each repo configuration.

would also mark it as @pytest.mark.universal so it's run when we use make test-python-universal

sdk/python/feast/inference.py
```diff
@@ -45,7 +45,7 @@ def update_entities_with_inferred_types_from_feature_views(
 
                 # get entity information from information extracted from the view batch source
                 extracted_entity_name_type_pairs = list(
-                    filter(lambda tup: tup[0] == entity.join_key, col_names_and_types,)
+                    filter(lambda tup: tup[0] == entity.join_key,col_names_and_types,)
```
adchia, line 48: any reason why there are changes here? and the next section?

sdk/python/tests/integration/registration/test_inference.py
```diff
@@ -137,6 +139,18 @@ def test_update_data_sources_with_inferred_event_timestamp_col(simple_dataset_1)
                 [file_source], RepoConfig(provider="local", project="test")
             )
 
+@pytest.mark.universal
```
adchia, line 142: also @pytest.mark.integration?

adchia (approved):

sdk/python/feast/inference.py
shaibruhis: ahh my bad, it's my formatter

adchia:
@shaibruhis Did some quick debugging. Tried pushing to your branch but was denied. See 
https://github.com/adchia/feast/commit/56fb5b705aab4bb29b1d2a4c93fc156c07c10959

for some fixes to make the tests work! Main issues were that data_sources aren't iterable and the last array you generated wasn't properly duplicating the value and instead making one large concatenated string

shaibruhis:
ahh yeah good catch. surprised that @dataclasses dont have a `.keys()` and `.values()` method implicity but i like your commit! How do you want me to incorporate it? Should i cherry pick or can i just copy it?

adchia:
you can just copy the changes in!

sdk/python/tests/integration/registration/test_inference.py
```diff
@@ -138,6 +138,26 @@ def test_update_data_sources_with_inferred_event_timestamp_col(simple_dataset_1)
             )
 
 
+@pytest.mark.integration
+@pytest.mark.universal
+def test_update_data_sources_with_inferred_event_timestamp_col(universal_data_sources):
+    (_, _, data_sources) = universal_data_sources
+
+    # remove defined event_timestamp_column to allow for inference
+    for data_source in data_sources.values():
```
adchia, line 147: btw should we take data sources and make a deep copy? as is this will mutate the sources and impact other tests potentially

shaibruhis: hmm is it not a fixture? I thought it gets setup and torn down for every test?

Also, I'm looking at the failed linter and can't see what the issue is

adchia: they're session scoped though so should be reused across tests (at least that was the intention)

adchia: did you try running make format-python?

sdk/python/tests/integration/registration/test_inference.py
```diff
@@ -138,6 +139,27 @@ def test_update_data_sources_with_inferred_event_timestamp_col(simple_dataset_1)
             )
 
 
+@pytest.mark.integration
+@pytest.mark.universal
+def test_update_data_sources_with_inferred_event_timestamp_col(universal_data_sources):
+    (_, _, data_sources) = universal_data_sources
+
+    # remove defined event_timestamp_column to allow for inference
+    for data_source in deepcopy(data_sources).values():
```
adchia, line 148: you don't seem to use this copy below?

shaibruhis:
Hey @adchia, anything else i should do to make land?

adchia:
Looks like you need to pass the linter by sorting.

You can run this linter locally with `make lint-python`. You also should probably install the precommit hooks (https://github.com/feast-dev/feast/blob/master/CONTRIBUTING.md#pre-commit-hooks) so this is run automatically

adchia:
The linter looks now to be failing and your commit wasn't signed? 

If you give me push privileges to your branch btw happy to also help resolve quickly.

shaibruhis:
Fixed the lint issues! I realized I had the wrong linter set which was causing problems for me. Apologies for that

adchia (approved):
/lgtm

sdk/python/tests/integration/registration/test_inference.py
```diff
@@ -138,6 +140,29 @@ def test_update_data_sources_with_inferred_event_timestamp_col(simple_dataset_1)
...
+def test_update_data_sources_with_inferred_event_timestamp_col(universal_data_sources):
+    (_, _, data_sources) = universal_data_sources
+    data_sources_copy = deepcopy(data_sources)
+
+    # remove defined event_timestamp_column to allow for inference
+    for data_source in data_sources_copy.values():
+        data_source.event_timestamp_column = None
+
+    update_data_sources_with_inferred_event_timestamp_col(
+        data_sources_copy.values(),
```
adchia, line 154: not sure why your local linter isn't working, but what i see locally is this should be all in a single line:

```
data_sources_copy.values(), RepoConfig(provider="local", project="test"),
```

adchia (approved):
/lgtm

PR #2391, opened 2022-03-09
feat: Use a daemon thread to monitor the go feature server exclusively

achals:
Also reenable the tests for lifecycle management of the goserver

This fixes some left over things from #2339 

Fixes #


sdk/python/feast/go_server.py
```diff
@@ -249,96 +216,71 @@ def get_online_features(
...
     def __init__(
         self,
         name: str,
         shared_connection: GoServerConnection,
-        go_server_started: threading.Event,
+        go_server_first_started: threading.Event,
     ):
         threading.Thread.__init__(self)
-        self.name = name
+        self.setName(name)
```
felixwang9817, line 246: using `name` directly as a property is preferred as opposed to using `setName`: https://docs.python.org/3.7/library/threading.html#threading.Thread.setName

sdk/python/feast/go_server.py
```diff
@@ -249,96 +216,71 @@ def get_online_features(
...
         self._shared_connection = shared_connection
-        self._go_server_started = go_server_started
+        self._is_cancelled = False
+        self.setDaemon(True)
```
felixwang9817, line 249: similar to `name`, `daemon` should be used directly as a property: https://docs.python.org/3.7/library/threading.html#threading.Thread.daemon

sdk/python/go_build.py
```diff
@@ -1,37 +0,0 @@
...
-# See the License for the specific language governing permissions and
-# limitations under the License.
-import os
-import pathlib
-import shutil
-import subprocess
-
-# Build go server for 3 targets: macos (intel), macos (m1), linux (64 bit)
-# First start by clearing the necessary directory
-binaries_path = (pathlib.Path(__file__) / "../feast/binaries").resolve()
```
felixwang9817, line 21: should we delete this code right now? I suspect we'll need a modified version of this logic for the release process of building binaries

sdk/python/setup.py
```diff
@@ -196,7 +196,7 @@ def finalize_options(self):
 
     def _generate_python_protos(self, path: str):
         proto_files = glob.glob(os.path.join(self.proto_folder, path))
-
+        Path(self.python_folder).mkdir(exist_ok=True)
```
felixwang9817, line 199: any particular reason for adding this? `self.python_folder` is set to `feast/sdk/python/feast/protos`, which already exists as a directory in our git repo

sdk/python/tests/integration/feature_repos/repo_configuration.py
```diff
@@ -48,7 +48,7 @@
 DYNAMO_CONFIG = {"type": "dynamodb", "region": "us-west-2"}
 # Port 12345 will chosen as default for redis node configuration because Redis Cluster is started off of nodes
 # 6379 -> 6384. This causes conflicts in cli integration tests so we manually keep them separate.
-REDIS_CONFIG = {"type": "redis", "connection_string": "localhost:12345,db=0"}
+REDIS_CONFIG = {"type": "redis", "connection_string": "localhost:6379,db=0"}
```
felixwang9817, line 51: I'm not sure it's safe to do this: Kevin recently changed our integration tests (both on PRs and on master) to start Redis on port 12345 instead of 6379; if you want to change back to 6379, you should change the workflows in `pr_integration_tests.yml` and `master_only.yml` (I would recommend doing that, since Kevin's change was actually unnecessary)

sdk/python/tests/integration/online_store/test_universal_online.py
```diff
@@ -607,12 +608,12 @@ def test_online_retrieval(environment, universal_data_sources, full_feature_name
 
     entity_rows = [
         {
-            "driver": driver,
-            "customer_id": customer,
+            "driver": _driver,
```
felixwang9817, line 611: why is this change being made? what is the difference between driver and origin id here?

sdk/python/tests/integration/online_store/test_universal_online.py
```diff
@@ -923,17 +921,17 @@ def test_online_store_cleanup_with_go_server(go_environment, go_data_sources):
     )
 
     fs.apply([driver(), simple_driver_fv, driver_stats_fv])
+    fs._registry.commit()
```
felixwang9817, line 924: is this necessary? `apply` automatically commits at the end

sdk/python/tests/integration/online_store/test_universal_online.py
```diff
@@ -1012,29 +1004,30 @@ def test_go_server_life_cycle(go_cycle_environment, go_data_sources):
...
     # Check that background thread has terminated
-    for id, thread in threading._active.items():
-        assert thread.name != "GoServerBackgroundThread"
+    monitor_thread = None
+    for thread in threading.enumerate():
+        if thread.name == "GoServerMonitorThread":
+            monitor_thread = thread
+    assert monitor_thread
+    assert monitor_thread.is_alive()
+    assert monitor_thread.isDaemon()
```
felixwang9817, line 1030: same as previously, I believe you can just use `daemon` as a property

felixwang9817 (commented):
looks great overall!

sdk/python/tests/integration/feature_repos/repo_configuration.py
felixwang9817: btw this is probably the reason all the integration tests are failing

sdk/python/feast/go_server.py
achals: ok, fixing.

sdk/python/setup.py
achals: I had deleted the directory locally - i think you're right that this isn't needed.

sdk/python/tests/integration/online_store/test_universal_online.py
achals: `driver` was being shadowed by an imported function so I changed the loop variable name.

sdk/python/tests/integration/online_store/test_universal_online.py
achals: I saw race conditions in getting features from the go FS.. It could have been something else tho

sdk/python/tests/integration/online_store/test_universal_online.py
achals: yup

sdk/python/tests/integration/feature_repos/repo_configuration.py
achals: but changed it to 12345 and will fix in a different PR.

sdk/python/tests/integration/online_store/test_universal_online.py
achals: https://github.com/feast-dev/feast/runs/5484347065?check_suite_focus=true

Example of failure without this line 🤷🏽

sdk/python/go_build.py
felixwang9817: fine with deleting

sdk/python/tests/integration/online_store/test_universal_online.py
woop: Isn't this a major red flag? Do we expect users to do this?

sdk/python/feast/go_server.py
```diff
@@ -44,13 +42,22 @@
 from feast.repo_config import RepoConfig
 from feast.type_map import python_values_to_proto_values
 
+_logger = logging.getLogger(__name__)
```
woop, line 45: Thanks ❤️

sdk/python/tests/integration/online_store/test_universal_online.py
achals: This is a major blocker for usability. @felixwang9817 and I are debugging this over slack.

sdk/python/feast/go_server.py
```diff
@@ -249,96 +216,71 @@ def get_online_features(
...
                     continue
-                self._go_server_started.set()
-                while True:
+                else:
+                    _logger.info("Set go server first started event")
+                    self._go_server_started.set()
+                while not self._is_cancelled:
                     try:
                         # Making a blocking wait by setting timeout to a very long time so we don't waste cpu cycle
                         self._shared_connection.wait_for_process(3600)
```
woop, line 268: So if we send a cancel command, how does that command kill the process. Isn't the loop blocked here?

sdk/python/feast/go_server.py
```diff
@@ -249,96 +216,71 @@ def get_online_features(
...
                     except subprocess.TimeoutExpired:
                         pass
+
+                    _logger.info(
+                        "Process state: %s %s",
```
woop, line 273: As a user, how do I know what "process state" means? What process?

sdk/python/feast/go_server.py
```diff
@@ -249,96 +216,71 @@ def get_online_features(
...
-
-    def _get_id(self):
-        # returns id of the respective thread
-        if hasattr(self, "_thread_id"):
-            return self._thread_id
-        for id, thread in threading._active.items():
-            if thread is self:
-                return id
+    def stop(self):
+        _logger.info("Stopping GoServerMonitorThread")
```
woop, line 284: Is it clearer to use plain english, do our users know what `GoServerMonitorThread` is?

sdk/python/go_build.py
```diff
@@ -1,37 +0,0 @@
...
-binaries_path_abs = str(binaries_path.absolute())
-if binaries_path.exists():
-    shutil.rmtree(binaries_path_abs)
-os.mkdir(binaries_path_abs)
-# Then, iterate over target architectures and build executables
-for goos, goarch in (("darwin", "amd64"), ("darwin", "arm64"), ("linux", "amd64")):
```
woop, line 27: How does this deletion tie into the daemon thread? Should we begin making these changes in separate PRs instead of mixing them in?

sdk/python/tests/integration/online_store/test_universal_online.py
```diff
@@ -1052,6 +1045,30 @@ def test_go_server_life_cycle(go_cycle_environment, go_data_sources):
...
+    assert fs._go_server._shared_connection._process.poll() is None
+
+    # And we can still get feature values.
+    fs.get_online_features(features=features, entity_rows=entity_rows).to_dict()
+
+    fs.kill_go_server()
+    time.sleep(1)
+
+    # Ensure process is dead.
+    assert fs._go_server._shared_connection._process.poll() is not None
```
woop, line 1070: Should we also figure out if the daemon thread is dead?

sdk/python/feast/go_server.py
achals: No, the cancel command will kill the process and the loop with break out because `wait_for_process` will return when the process ends. The cancel command will have set `is_cancelled` so the loop will terminate.

sdk/python/feast/go_server.py
achals: removed this this was for my understanding of the process lifecycle.

sdk/python/feast/go_server.py
achals: updated the log, but made it debug logging too.

sdk/python/go_build.py
achals: Yeah I'll move this to a different PR.

sdk/python/tests/integration/online_store/test_universal_online.py
achals: yes

sdk/python/feast/go_server.py
```diff
@@ -81,7 +87,7 @@ def connect(self) -> bool:
         self._process = Popen([executable], cwd=cwd, env=env,)
 
         channel = grpc.insecure_channel(f"unix:{self.sock_file}")
-        self.client: ServingServiceStub = ServingServiceStub(channel)
+        self._client = ServingServiceStub(channel)
```
kevjumba, line 90: Why remove this type definition?

sdk/python/feast/go_server.py
```diff
@@ -249,96 +215,69 @@ def get_online_features(
...
-    def stop_go_server(self):
-        thread_id = self._get_id()
-        res = ctypes.pythonapi.PyThreadState_SetAsyncExc(
-            thread_id, ctypes.py_object(SystemExit)
-        )
-        # TODO: Review that kill process here but run also has to stop
-        if res > 1:
-            ctypes.pythonapi.PyThreadState_SetAsyncExc(thread_id, 0)
-        else:
-            self._shared_connection.kill_process()
```
kevjumba, line 336: Just curious, why was this the original implementation, feel like the current implementation to stop the thread is cleaner. Was there some type of problem with the new implementation that forced us do this messy asyncexc stuff and if so, how did we fix it?

sdk/python/feast/go_server.py
felixwang9817: it was moved to line 53 above

sdk/python/feast/go_server.py
achals: The first implementation was written during hackweek with a bunch of other moving pieces; not everything was optimal. post-hackweek we're trying to polish everything up.

sdk/python/tests/integration/online_store/test_universal_online.py
achals: Okay we've figured this out. The latest version of the code should be green.

The issue was that the feature store class would have left over state in the go server after running `teardown`.

felixwang9817 (approved):
/lgtm

PR #2393, opened 2022-03-09
fix: Fix Spark template to work correctly on feast init -t spark

adchia:
This includes not using SparkSessions to generate the test data (which generates a pickle exception as is) and also dynamically generating timestamps so feature view TTLs don't expire

Fixes #


kevjumba (approved):
/lgtm

PR #2394, opened 2022-03-10
feat: Add templating for dynamodb table name

dannyboland:
Sets the dynamodb online store's table names based on a template, which can be changed in config. 

Fixes #


achals (commented):
/ok-to-test

achals (approved):
/lgtm

PR #2395, opened 2022-03-10
ci: Reintroduce redis services at 6379

achals:

We moved this to a github action in #2350 and #2354 but this change wasn't needed. I'd like to move the tests back to using the standard redis port so that the tests can be run while developing locally without any changes.

Corresponding test changes to use this redis instance and remove the old one will be in a followup PR.

Fixes #

### DIFF
--- a/.github/workflows/master_only.yml
+++ b/.github/workflows/master_only.yml
@@ -66,6 +66,16 @@ jobs:
     env:
       OS: ${{ matrix.os }}
       PYTHON: ${{ matrix.python-version }}
+    services:
+      redis:
+        image: redis
+        ports:
+          - 6379:6379
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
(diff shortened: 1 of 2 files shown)


.github/workflows/master_only.yml
```diff
@@ -66,6 +66,16 @@ jobs:
     env:
       OS: ${{ matrix.os }}
       PYTHON: ${{ matrix.python-version }}
+    services:
```
felixwang9817, line 69: I think you missed a supercharge below, line 132

felixwang9817 (changes requested):
missed a supercharge

.github/workflows/master_only.yml
achals: I'm doing to remove this the superchange actions after this PR gets in.

felixwang9817 (approved):
/lgtm

PR #2403, opened 2022-03-14
feat: Add DQM Logging on GRPC Server with FileLogStorage for Testing

kevjumba:
DQM Logging GRPC endpoint that exposes API to add logs to channels.

Fixes #


woop:
@kevjumba are you expecting review comments on this PR? Seems like it's not in a reviewable state? Can you please mark it as WIP or put it in draft?

kevjumba:
> @kevjumba are you expecting review comments on this PR? Seems like it's not in a reviewable state? Can you please mark it as WIP or put it in draft?

Hm I wonder why the feast bot removed my work in progress label...

go/cmd/server/server.go
```diff
@@ -28,6 +30,53 @@ func (s *servingServiceServer) GetOnlineFeatures(ctx context.Context, request *s
...
+			}
+			featureValues[idx] = &types.Value{Val: featureValue.Val}
+		}
+		for idx, ts := range featureVector.EventTimestamps {
+			if featureVector.Statuses[idx] != serving.FieldStatus_PRESENT {
+				continue
+			}
+			eventTimestamps[idx] = &timestamppb.Timestamp{Seconds: ts.Seconds, Nanos: ts.Nanos}
+		}
+		// TODO(kevjumba): For filtering out and extracting entity names when the bug with join keys is fixed.
```
woop, line 57: You can't commit this code into master. Please remove it or at change this PR into a WIP or draft so that people don't spend time reviewing it.

go/cmd/server/logging/filelogstorage.go
```diff
@@ -0,0 +1,74 @@
...
+
+type FileLogStorage struct {
+	// Feast project name
+	project string
+	path    string
+}
+
+func NewFileOfflineStore(project string, offlineStoreConfig map[string]interface{}) (*FileLogStorage, error) {
+	store := FileLogStorage{project: project}
+	var abs_path string
```
pyalex, line 22: camelCase please

go/cmd/server/logging/filelogstorage.go
```diff
@@ -0,0 +1,74 @@
...
+
+	// create file if not exists
+	if os.IsNotExist(err) {
+		var file, err = os.Create(absPath)
+		if err != nil {
+			return nil, err
+		}
+		return file, nil
+	} else {
+		var file, err = os.OpenFile(absPath, os.O_RDWR, 0644)
```
pyalex, line 52: I think it's better to throw error here if file exists. Overwrite something or append can cause a flaky behaviour (especially in tests)

go/cmd/server/logging/filelogstorage.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/apache/arrow/go/v8/parquet/pqarrow"
+)
+
+type FileLogStorage struct {
+	// Feast project name
+	project string
+	path    string
+}
+
+func NewFileOfflineStore(project string, offlineStoreConfig map[string]interface{}) (*FileLogStorage, error) {
```
pyalex, line 20: can we pass offlineStoreConfig as some struct with predefined properties rather than map with interface?

go/cmd/server/logging/filelogstorage_test.go
```diff
@@ -0,0 +1,89 @@
...
+	assert.Nil(t, err)
+	err = fileStore.FlushToStorage(array.Table(table))
+	assert.Nil(t, err)
+	logPath, err := filepath.Abs(filepath.Join(".", "log.parquet"))
+	assert.Nil(t, err)
+	w, err := CreateOrOpenLogFile(logPath)
+	assert.Nil(t, err)
+	ctx := context.Background()
+
+	pf, err := file.NewParquetReader(w)
```
pyalex, line 37: `file.OpenParquetFile` does both file opening and create parquet reading
`CreateOrOpenLogFile` should be actually `CreateParquetWriter` function

go/cmd/server/logging/filelogstorage_test.go
```diff
@@ -0,0 +1,89 @@
...
+	defer tbl.Release()
+	expected_schema := map[string]arrow.DataType{
+		"driver_id": arrow.PrimitiveTypes.Int64,
+		"int32":     arrow.PrimitiveTypes.Int32,
+		"double":    arrow.PrimitiveTypes.Float64,
+		"int64":     arrow.PrimitiveTypes.Int64,
+		"float32":   arrow.PrimitiveTypes.Float32,
+	}
+
+	expected_columns := map[string]*types.RepeatedValue{
```
pyalex, line 55: ideally we should reuse here data generated above. Because in order to change any value I need to change it in several places (should be one)

go/internal/feast/model/basefeatureview.go
```diff
@@ -56,3 +57,11 @@ func (fv *BaseFeatureView) ProjectWithFeatures(featureNames []string) *FeatureVi
 		Features: features,
 	}
 }
+
+func CreateBaseFeatureView(name string, features []*Feature, projection *FeatureViewProjection) *BaseFeatureView {
```
pyalex, line 61: if this is for tests only, it should in test package

go/cmd/server/server.go
```diff
@@ -49,21 +52,87 @@ func (s *servingServiceServer) GetOnlineFeatures(ctx context.Context, request *s
...
 			Statuses:        vector.Statuses,
 			EventTimestamps: vector.Timestamps,
 		})
 	}
 
+	go generateLogs(s, entityValuesMap, featureNames, resp.Results, request.RequestContext)
 	return resp, nil
 }
+
+func generateLogs(s *servingServiceServer, entityMap map[string][]*prototypes.Value, featureNames []string, features []*serving.GetOnlineFeaturesResponse_FeatureVector, requestData map[string]*prototypes.RepeatedValue) error {
```
pyalex, line 81: Shouldn't this function be part of logging package?

go/cmd/server/server.go
```diff
@@ -49,21 +52,87 @@ func (s *servingServiceServer) GetOnlineFeatures(ctx context.Context, request *s
...
+		return nil
+	}
+
+	entityArr, err := s.fs.ListEntities(true)
+	if err != nil {
+		return err
+	}
+
+	joinKeys := make([]string, 0)
+	for _, entity := range entityArr {
```
pyalex, line 100: we shouldn't use all entities here, but only entities that required by specific feature service

go/cmd/server/logging/filelogstorage.go
pyalex: Also please add a docstring here. We should at least state that this implementation is for testing

go/cmd/server/server.go
```diff
@@ -49,21 +52,87 @@ func (s *servingServiceServer) GetOnlineFeatures(ctx context.Context, request *s
...
 	return resp, nil
 }
+
+func generateLogs(s *servingServiceServer, entityMap map[string][]*prototypes.Value, featureNames []string, features []*serving.GetOnlineFeaturesResponse_FeatureVector, requestData map[string]*prototypes.RepeatedValue) error {
+	// Add a log with the request context
+	if len(requestData) > 0 {
+		requestContextLog := logging.Log{
+			RequestContext: requestData,
+		}
+		s.loggingService.EmitLog(&requestContextLog)
```
pyalex, line 87: why request data is being emited separately from all the rest?

go/cmd/server/server.go
```diff
@@ -49,21 +52,87 @@ func (s *servingServiceServer) GetOnlineFeatures(ctx context.Context, request *s
...
+			featureValueLogRows[row_idx][idx-1] = features[idx].Values[row_idx]
+			featureStatusLogRows[row_idx][idx-1] = features[idx].Statuses[row_idx]
+			eventTimestampLogRows[row_idx][idx-1] = features[idx].EventTimestamps[row_idx]
+		}
+		entityRow := make([]*prototypes.Value, 0)
+		// ensure that the entity values are in the order that the schema defines which is the order that ListEntities returns the entities
+		for _, joinKey := range joinKeys {
+			entityRow = append(entityRow, entityMap[joinKey][row_idx])
+		}
+		newLog := logging.Log{
```
pyalex, line 126: if log is being emitted per each row, we need to have request id (that is should probably be generated somewhere here) in each log to understand which rows came from which request

go/cmd/server/server.go
```diff
@@ -49,21 +52,87 @@ func (s *servingServiceServer) GetOnlineFeatures(ctx context.Context, request *s
...
+	joinKeys := make([]string, 0)
+	for _, entity := range entityArr {
+		joinKeys = append(joinKeys, entity.JoinKey)
+	}
+
+	numFeatures := len(featureNames)
+	// Should be equivalent to how many entities there are(each feature row has (entity) number of features)
+	// acc_rate []
+	numRows := len(features[0].Values)
+	featureValueLogRows := make([][]*prototypes.Value, numRows)
```
pyalex, line 108: I don't get it, why we build these arrays if they are not used? In each loop we emit one row and then newer use it again.

go/cmd/server/server.go
pyalex: Ideally in this function we should already have schema generated from feature service and use it as source of truth for columns/fields names and order

go/cmd/server/logging/logging.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+type LoggingService struct {
+	memoryBuffer      *MemoryBuffer
+	logChannel        chan *Log
+	fs                *feast.FeatureStore
+	offlineLogStorage OfflineLogStorage
+}
+
+func NewLoggingService(fs *feast.FeatureStore, logChannelCapacity int, featureServiceName string, enableLogging bool) (*LoggingService, error) {
```
pyalex, line 42: `enabledLogging` seems very strange parameter in logging service constructor.
I understand what's the goal here, so I think a better approach would be to have `DummyLoggingService` (separate type) with separate constructor `NewDummyLoggingService` that would have method `EmitLog` that does nothing

go/cmd/server/server.go
kevjumba: Oh this was left over and I didnt' realize I needed to change the implementation. Fixed.

go/cmd/server/logging/logging.go
```diff
@@ -0,0 +1,306 @@
...
+	defer ticker.Stop()
+
+	for {
+		s.ProcessMemoryBuffer(ticker)
+	}
+}
+
+// Select that eitheringests new logs that are added to the logging channel, one at a time to add
+// to the in memory buffer or flushes all of them synchronously to the OfflineStorage on a time interval.
+func (s *LoggingService) ProcessMemoryBuffer(t *time.Ticker) {
```
pyalex, line 98: function name is slightly confusing

go/cmd/server/logging/logging.go
```diff
@@ -0,0 +1,306 @@
...
+	case new_log := <-s.logChannel:
+		log.Printf("Adding %s to memory.\n", new_log.FeatureValues)
+		s.memoryBuffer.logs = append(s.memoryBuffer.logs, new_log)
+	}
+}
+
+// Acquires the logging schema from the feature service, converts the memory buffer array of rows of logs and flushes
+// them to the offline storage.
+func (s *LoggingService) flushLogsToOfflineStorage(t time.Time) error {
+	log.Printf("Flushing buffer to offline storage with channel length: %d\n at time: "+t.String(), len(s.memoryBuffer.logs))
```
pyalex, line 111: `log.Printf("... "+t.String()` if format is used, let's use format for all parameters

go/cmd/server/logging/logging.go
```diff
@@ -0,0 +1,306 @@
...
+	columnNameToProtoValueArray := make(map[string][]*types.Value)
+	columnNameToStatus := make(map[string][]int32)
+	columnNameToTimestamp := make(map[string][]int64)
+	entityNameToEntityValues := make(map[string][]*types.Value)
+
+	for _, l := range memoryBuffer.logs {
+		// EntityTypes maps an entity name to the specific type and also which index in the entityValues array it is
+		// e.g if an Entity Key is {driver_id, customer_id}, then the driver_id entitytype would be dtype=int64, index=0.
+		// It's in the order of the entities as given by the schema.
+		for idx, entityName := range fcoSchema.Entities {
```
pyalex, line 157: on this point it's a `joinKey` not an `entityName`

go/cmd/server/logging/logging.go
```diff
@@ -0,0 +1,306 @@
...
+		if err != nil {
+			return nil, err
+		}
+		fields = append(fields, arrow.Field{
+			Name: featureName,
+			Type: arrowType,
+		})
+		columns = append(columns, arrowArray)
+	}
+	schema := arrow.NewSchema(
```
pyalex, line 222: we should build here schema from feature service specs. it would give us better guarantees that types are correct and expected

go/cmd/server/logging/logging.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	allFeatureTypes := make(map[string]types.ValueType_Enum)
+	features := make([]string, 0)
+	for _, featureProjection := range featureService.Projections {
+		// Create copies of FeatureView that may contains the same *FeatureView but
+		// each differentiated by a *FeatureViewProjection
+		featureViewName := featureProjection.Name
+		if fv, ok := fvs[featureViewName]; ok {
+			for _, f := range fv.Base.Features {
```
pyalex, line 269: we should use only projected features here

go/cmd/server/logging/logging.go
```diff
@@ -0,0 +1,306 @@
...
+				// add feature to map
+				features = append(features, f.Name)
+				allFeatureTypes[f.Name] = f.Dtype
+			}
+		} else if odfv, ok := odFvs[featureViewName]; ok {
+			for name, valueType := range odfv.GetRequestDataSchema() {
```
pyalex, line 275: on demand feature views dependencies (`RequestData`) should not be added to features. It's input, not output. I guess separate property requestData should be added to `Schema`.
On the other hand odfvs have output features and they should be added to list of features here.

go/cmd/server/logging/logging.go
```diff
@@ -0,0 +1,306 @@
...
+	FeaturesTypes map[string]types.ValueType_Enum
+}
+
+func GetSchemaFromFeatureService(featureService *model.FeatureService, entities []*model.Entity, featureViews []*model.FeatureView, onDemandFeatureViews []*model.OnDemandFeatureView) (*Schema, error) {
+	fvs := make(map[string]*model.FeatureView)
+	odFvs := make(map[string]*model.OnDemandFeatureView)
+
+	entityNames := make([]string, 0)
+	for _, entity := range entities {
+		entityNames = append(entityNames, entity.JoinKey)
```
pyalex, line 246: we need to add only join keys that are required by currect feature service not all entities that exist

go/cmd/server/logging/logging_test.go
```diff
@@ -0,0 +1,223 @@
...
+		FeatureStatuses: []serving.FieldStatus{serving.FieldStatus_PRESENT},
+		EventTimestamps: []*timestamppb.Timestamp{newTs, newTs},
+	}
+	err = loggingService.EmitLog(&newLog2)
+	// The channel times out and doesn't hang.
+	time.Sleep(20 * time.Millisecond)
+	assert.NotNil(t, err)
+}
+
+func TestSchemaTypeRetrieval(t *testing.T) {
```
pyalex, line 43: Taking into account comments above this test is not enough. We need better coverage of schema generation

go/cmd/server/logging/logging.go
kevjumba: Actually I'll remove this.

go/cmd/server/logging/logging.go
kevjumba: The specs are built in the loop, I use the type that is returned from GetSchema when looping through instead of the types that are in the arrow array.

go/cmd/server/server.go
kevjumba: Should this be emitted with each log? Isn't that the same issue as before which is that we only need request context once but it's duplicated for each log?

pyalex: request data is also split by rows, basically has the same structure as entity rows and features

go/cmd/server/logging/logging.go
pyalex: My bad. Missed that

go/cmd/server/logging/filelogstorage_test.go
kevjumba: Done

go/cmd/server/logging/logging.go
```diff
@@ -0,0 +1,306 @@
...
+		go loggingService.processLogs()
+	}
+	return loggingService, nil
+}
+
+func (s *LoggingService) EmitLog(l *Log) error {
+	select {
+	case s.logChannel <- l:
+		return nil
+	case <-time.After(20 * time.Millisecond):
```
pyalex, line 80: nitpick: would be good to parametrize this timeout

go/cmd/server/logging/logging.go
pyalex: also please use `joinKey` name

go/cmd/server/logging/logging.go
```diff
@@ -0,0 +1,306 @@
...
+	features := make([]string, 0)
+	for _, featureProjection := range featureService.Projections {
+		// Create copies of FeatureView that may contains the same *FeatureView but
+		// each differentiated by a *FeatureViewProjection
+		featureViewName := featureProjection.Name
+		if fv, ok := fvs[featureViewName]; ok {
+			for _, f := range fv.Base.Features {
+				// add feature to map
+				features = append(features, f.Name)
+				allFeatureTypes[f.Name] = f.Dtype
```
pyalex, line 272: feature name is not unique within feature service. it's unique only within its feature view. So would be good idea to use full feature name here `{view.name}__{feature.name}`

go/cmd/server/server_test.go
```diff
@@ -33,20 +42,40 @@ func getRepoPath(basePath string) string {
...
+
 	if err != nil {
 		panic(err)
 	}
 	fs, err := feast.NewFeatureStore(config, nil)
 	if err != nil {
 		panic(err)
 	}
-	serving.RegisterServingServiceServer(server, &servingServiceServer{fs: fs})
+	generateFeatureService(fs)
```
pyalex, line 71: not sure what this function does if it returns nothing

go/cmd/server/server_test.go
```diff
@@ -137,4 +166,139 @@ func TestGetOnlineFeaturesSqlite(t *testing.T) {
...
+
+	request := &serving.GetOnlineFeaturesRequest{
+		Kind: &serving.GetOnlineFeaturesRequest_FeatureService{
+			FeatureService: "test_service",
+		},
+		Entities: entities,
+	}
+	response, err := client.GetOnlineFeatures(ctx, request)
+	// Wait for logger to flush.
+	time.Sleep(1 * time.Second)
```
pyalex, line 198: `assert.Eventually` is better way to wait for smth

go/cmd/server/logging/logging_test.go
```diff
@@ -0,0 +1,223 @@
...
+	assert.Nil(t, err)
+	assert.Empty(t, loggingService.memoryBuffer.logs)
+	ts := timestamppb.New(time.Now())
+	newLog := Log{
+		FeatureStatuses: []serving.FieldStatus{serving.FieldStatus_PRESENT},
+		EventTimestamps: []*timestamppb.Timestamp{ts, ts},
+	}
+	loggingService.EmitLog(&newLog)
+	// Wait for memory buffer flush
+	time.Sleep(20 * time.Millisecond)
```
pyalex, line 30: please use `assert.Eventually`

go/cmd/server/logging/logging_test.go
```diff
@@ -0,0 +1,223 @@
...
+	tr := array.NewTableReader(table, -1)
+	expected_schema := map[string]arrow.DataType{
+		"driver_id": arrow.PrimitiveTypes.Int64,
+		"int32":     arrow.PrimitiveTypes.Int32,
+		"double":    arrow.PrimitiveTypes.Float64,
+		"int64":     arrow.PrimitiveTypes.Int64,
+		"float32":   arrow.PrimitiveTypes.Float32,
+	}
+
+	expected_columns := map[string]*types.RepeatedValue{
```
pyalex, line 71: please use arrow to proto conversion instead of hardcoding this

go/cmd/server/logging/logging_test.go
```diff
@@ -0,0 +1,223 @@
...
+		},
+		EventTimestamps: []*timestamppb.Timestamp{
+			ts, ts, ts, ts,
+		},
+	}
+
+	dummyTicker := time.NewTicker(10 * time.Second)
+	// stop the ticker so that the logs are not flushed to offline storage
+	dummyTicker.Stop()
+	loggingService.EmitLog(&log1)
```
pyalex, line 213: it's confusing that this method (with name `GenerateLogsAndConvertToArrowTable`) that generates some data and also emit some logs. Can you maybe split this function / generalize it better?

go/cmd/server/logging/filelogstorage.go
kevjumba: Done.

go/cmd/server/server_test.go
kevjumba: made function more descriptive.

go/cmd/server/logging/logging_test.go
kevjumba: Fixed.

go/cmd/server/server_test.go
kevjumba: Fixed.

go/cmd/server/logging/logging_test.go
kevjumba: Refactored..

go/cmd/server/logging/logging_test.go
kevjumba: This is the expected schema so I don't think it makes sense to use proto conversion. I refactored so that it exists elsewhere though.

go/cmd/server/logging/filelogstorage.go
kevjumba: Fixed.

go/cmd/server/logging/filelogstorage_test.go
kevjumba: Done.

go/internal/feast/model/basefeatureview.go
kevjumba: Done.

go/cmd/server/server.go
kevjumba: Done.

go/cmd/server/logging/logging.go
kevjumba: Fixed.

go/cmd/server/logging/logging.go
kevjumba: Fixed.

go/cmd/server/logging/logging.go
kevjumba: Fixed.

go/cmd/server/logging/logging.go
```diff
@@ -0,0 +1,359 @@
...
+			for name, valueType := range odfv.GetRequestDataSchema() {
+				features = append(features, name)
+				allFeatureTypes[name] = valueType
+			}
+		} else {
+			return nil, fmt.Errorf("no such feature view found in feature service %s", featureViewName)
+		}
+	}
+	schema := &Schema{
+		Entities:      joinKeys,
```
pyalex, line 282: I guess here we need only joinKeys that required by current feature service

go/cmd/server/logging/logging.go
kevjumba: I think that in itself may be redundant since we are essentially just copy pasting code for the dummy log service.
Also emit log is not the part that I disable. I disable the log processing part which I think makes sense even for log service.

go/cmd/server/logging/filelogstorage.go
kevjumba: Done.

go/cmd/server/logging/filelogstorage.go
kevjumba: Done.

go/cmd/server/logging/filelogstorage_test.go
kevjumba: Fixed.

go/cmd/server/logging/logging.go
kevjumba: Done.

go/cmd/server/logging/logging.go
kevjumba: For this, there no way of actually loading entities based on feature views or feature service. Should I add this functionality? I will have to update either featureservice to contain references to each entitiy or the entitys that reference the featureviews.

go/cmd/server/logging/logging.go
kevjumba: I'm not quite sure what this means?

go/cmd/server/server.go
kevjumba: Is it ok if I add in request data in a separate pr? Adding this will add a lot more complexity that will require me to update all of the tests again.

go/cmd/server/server.go
kevjumba: Addressing 1, refer to other comment about entities aren't tied to a specific feature service right now in our data model. 
Addressing 2, I think I implemented it this way because of the way the request is structured. A featureservice request contains a projection that contains the features in order as a slice. This is processed and returned in the a similar way via slices so as long as I refer to featureservice as the source of truth for the ordering of the field names, i feel like it makes more sense this way. Wdyt?

go/cmd/server/server.go
kevjumba: Done.

go/cmd/server/logging/logging.go
kevjumba: Done, but I also had to specify fullfeaturenames = true when querying since other wise the returned feature names don't actually have all of the information I need.

go/cmd/server/logging/logging.go
kevjumba: In the current data model, I don't think there is a direct connection between entities and feature service or entity names and featureviews.

go/cmd/server/logging/logging.go
kevjumba: Fixed.

go/cmd/server/logging/logging.go
pyalex:
```
joinKeysSet = make(map[string]interface{})

for _, projection := range featureService.Projections {
    fv := fvs[projection.Name]
    for _, entityName := range fv.Entities {
          entity := entities[entityName]
          if joinKeyAlias, ok := projection.JoinKeyMap[entity.JoinKey]; ok {
            joinKeysSet[joinKeyAlias] = nil
          } else {
            joinKeysSet[entity.JoinKey] = nil
          }
    }
}
```

go/cmd/server/logging/filelogstorage.go
```diff
@@ -0,0 +1,86 @@
...
+		return nil, errors.New("need path for file log storage")
+	}
+	if err != nil {
+		return nil, err
+	}
+	store.path = absPath
+	return &store, nil
+}
+
+func OpenLogFile(absPath string) (*os.File, error) {
```
pyalex, line 57: this method shouldn't be public probably

go/cmd/server/logging/logging.go
```diff
@@ -0,0 +1,384 @@
...
+const DEFAULT_LOG_INSERT_TIMEOUT = 20 * time.Millisecond
+
+type Log struct {
+	// Example: val{int64_val: 5017}, val{int64_val: 1003}
+	EntityValue []*types.Value
+
+	FeatureValues   []*types.Value
+	FeatureStatuses []serving.FieldStatus
+	EventTimestamps []*timestamppb.Timestamp
+	RequestContext  map[string]*types.RepeatedValue
```
pyalex, line 30: if it's just single row, the type should be same as features and entities - `types.Value`, not repeated value

go/cmd/server/logging/logging.go
```diff
@@ -0,0 +1,384 @@
...
+	defer ticker.Stop()
+
+	for {
+		s.HandleLogFlushing(ticker)
+	}
+}
+
+// Select that eitheringests new logs that are added to the logging channel, one at a time to add
+// to the in memory buffer or flushes all of them synchronously to the OfflineStorage on a time interval.
+func (s *LoggingService) HandleLogFlushing(t *time.Ticker) {
```
pyalex, line 108: This function does not only flushing but also pulls from logchannel and fills the buffer

go/cmd/server/logging/filelogstorage.go
kevjumba: Fixed.

go/cmd/server/logging/logging.go
kevjumba: Changed the name again.

go/cmd/server/logging/logging.go
kevjumba: Done.

go/cmd/server/logging/logging_test.go
```diff
@@ -0,0 +1,262 @@
...
+	assert.Nil(t, err)
+	assert.Empty(t, loggingService.memoryBuffer.logs)
+	ts := timestamppb.New(time.Now())
+	newLog := Log{
+		FeatureStatuses: []serving.FieldStatus{serving.FieldStatus_PRESENT},
+		EventTimestamps: []*timestamppb.Timestamp{ts, ts},
+	}
+	loggingService.EmitLog(&newLog)
+	// Wait for memory buffer flush
+	assert.Eventually(t, func() bool { return true }, 100*time.Millisecond, DEFAULT_LOG_INSERT_TIMEOUT)
```
pyalex, line 31: what is being checked here if it always true?
`assert.Eventually(t, func() bool { return true }`

kevjumba: Fixed.

go/cmd/server/logging/logging.go
kevjumba: Done.

go/cmd/server/logging/logging.go
```diff
@@ -0,0 +1,391 @@
...
+	fs                *feast.FeatureStore
+	offlineLogStorage OfflineLogStorage
+	logInsertTTl      time.Duration
+	logFlushInterval  time.Duration
+}
+
+func NewLoggingService(fs *feast.FeatureStore, logChannelCapacity int, featureServiceName string, enableLogProcessing bool) (*LoggingService, error) {
+	var featureService *model.FeatureService = nil
+	var err error
+	if fs != nil {
```
pyalex, line 51: shouldn't `fs` always be not nil?

go/cmd/server/logging/logging.go
```diff
@@ -0,0 +1,391 @@
...
+				features = append(features, GetFullFeatureName(featureViewName, f.Name))
+				allFeatureTypes[GetFullFeatureName(featureViewName, f.Name)] = f.Dtype
+			}
+		} else {
+			return nil, fmt.Errorf("no such feature view found in feature service %s", featureViewName)
+		}
+	}
+
+	// Only get entities in the current feature service.
+	for _, entity := range entities {
```
pyalex, line 305: why not to iterate over `joinKeysSet`?

go/cmd/server/logging/logging.go
```diff
@@ -0,0 +1,391 @@
...
+			for entityName := range fv.Entities {
+				entity := entityMap[entityName]
+				if joinKeyAlias, ok := featureProjection.JoinKeyMap[entity.JoinKey]; ok {
+					joinKeysSet[joinKeyAlias] = nil
+				} else {
+					joinKeysSet[entity.JoinKey] = nil
+				}
+			}
+		} else if odfv, ok := odFvs[featureViewName]; ok {
+			for _, f := range odfv.Base.Features {
```
pyalex, line 294: we need to use projection here and not all features (similarly to regular feature views)

go/cmd/server/logging/logging.go
```diff
@@ -0,0 +1,391 @@
...
+}
+
+type Schema struct {
+	Entities      []string
+	Features      []string
+	EntityTypes   map[string]types.ValueType_Enum
+	FeaturesTypes map[string]types.ValueType_Enum
+}
+
+func GetSchemaFromFeatureService(featureService *model.FeatureService, entities []*model.Entity, entityMap map[string]*model.Entity, featureViews []*model.FeatureView, onDemandFeatureViews []*model.OnDemandFeatureView) (*Schema, error) {
```
pyalex, line 257: why we need here both entities and entityMap?

go/cmd/server/logging/logging.go
```diff
@@ -0,0 +1,391 @@
...
+	fvs, err := s.fs.ListFeatureViews()
+	if err != nil {
+		return nil, nil, nil, nil, err
+	}
+	entities, err := s.fs.ListEntities(true)
+	if err != nil {
+		return nil, nil, nil, nil, err
+	}
+	entityMap := make(map[string]*model.Entity)
+	for _, entity := range entities {
```
pyalex, line 339: let's maybe move all maps creation here?

go/internal/feast/model/featureview.go
```diff
@@ -44,3 +44,11 @@ func (fs *FeatureView) NewFeatureViewFromBase(base *BaseFeatureView) *FeatureVie
 	}
 	return featureView
 }
+
+func CreateFeatureView(base *BaseFeatureView, ttl *durationpb.Duration, entities map[string]struct{}) *FeatureView {
```
pyalex: duplicate?

pyalex: please move this to tests as well

go/cmd/server/server_test.go
```diff
@@ -138,3 +170,149 @@ func TestGetOnlineFeaturesSqlite(t *testing.T) {
...
+	ts := timestamppb.New(time.Now())
+	featureServiceMeta := &core.FeatureServiceMeta{
+		CreatedTimestamp:     ts,
+		LastUpdatedTimestamp: ts,
+	}
+	featureService := &core.FeatureService{
+		Spec: featureServiceSpec,
+		Meta: featureServiceMeta,
+	}
+	registry := fs.Registry()
```
pyalex, line 315: so where this registry goes? is it used anywhere?

go/cmd/server/logging/logging.go
kevjumba: Actually already fixed.

go/cmd/server/logging/logging.go
kevjumba: I'll move it inside, It just makes parsing the entities a little easier.

go/internal/feast/model/featureview.go
kevjumba: Oh Already done.

go/cmd/server/server_test.go
kevjumba: It is used in feature Service right below, essentially, I cache the featureservice manually using the featureservice proto 
fs.Registry just returns a pointer and we need it to actually query the feature service from registry when the logging service queries for it.

go/cmd/server/logging/logging.go
kevjumba: In some tests it is nil because for example for logger tests, there isn't a need to instantiate a full feature service if I'm only testing individual components.

go/cmd/server/logging/logging.go
kevjumba: I don't think there are any other map creations that we need? Only this one, the other entityValueMap needs to be created in GetOnlineFeatures.

go/cmd/server/logging/logging_test.go
```diff
@@ -0,0 +1,404 @@
...
+	loggingService, err := NewLoggingService(nil, 1, "", false)
+	assert.Nil(t, err)
+	assert.Empty(t, loggingService.memoryBuffer.logs)
+	ts := timestamppb.New(time.Now())
+	newLog := Log{
+		FeatureStatuses: []serving.FieldStatus{serving.FieldStatus_PRESENT},
+		EventTimestamps: []*timestamppb.Timestamp{ts, ts},
+	}
+	loggingService.EmitLog(&newLog)
+	// Wait for memory buffer flush
```
pyalex, line 31: probably should delete comment as well

kevjumba: Fixed.

go/internal/feast/registry/registry.go
```diff
@@ -36,6 +37,17 @@ type Registry struct {
 	mu                             sync.Mutex
 }
 
+func (r *Registry) CacheFeatureService(featureService *core.FeatureService) {
```
pyalex, line 40: Please remove this function, since it's not used anymore

kevjumba: Fixed.

pyalex (approved):

PR #2405, opened 2022-03-14
feat: Make feast PEP 561 compliant

tpvasconcelos:
Feast is not PEP 561 compliant which means that mypy won't infer types from feast even though the library is fully typed. This can easily be fixed by including a `py.typed` marker file at the root of the package.

References:
- https://peps.python.org/pep-0561/
- https://mypy.readthedocs.io/en/stable/installed_packages.html#using-installed-packages-with-mypy-pep-561

Signed-off-by: Tomas Pereira de Vasconcelos <tomasvasconcelos1@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,7 @@ format-python:
 	cd ${ROOT_DIR}/sdk/python; python -m black --target-version py37 feast tests
 
 lint-python:
-	cd ${ROOT_DIR}/sdk/python; python -m mypy feast/ tests/
+	cd ${ROOT_DIR}/sdk/python; python -m mypy
 	cd ${ROOT_DIR}/sdk/python; python -m isort feast/ tests/ --check-only
 	cd ${ROOT_DIR}/sdk/python; python -m flake8 feast/ tests/
 	cd ${ROOT_DIR}/sdk/python; python -m black --check feast tests
--- a/sdk/python/setup.cfg
+++ b/sdk/python/setup.cfg
@@ -1,4 +1,5 @@
 [isort]
+src_paths = feast,tests
 multi_line_output=3
 include_trailing_comma=True
 force_grid_wrap=0
@@ -16,5 +17,7 @@ select = B,C,E,F,W,T4
 exclude = .git,__pycache__,docs/conf.py,dist,feast/protos
 
 [mypy]
-files=feast,test
+files=feast,tests
 ignore_missing_imports=true
+# ignore mypy error from generated protos modules
+exclude=feast/protos/*
--- a/sdk/python/setup.cfg
+++ b/sdk/python/setup.cfg
@@ -19,5 +19,3 @@ exclude = .git,__pycache__,docs/conf.py,dist,feast/protos
 [mypy]
 files=feast,tests
 ignore_missing_imports=true
-# ignore mypy error from generated protos modules
-exclude=feast/protos/*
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -92,7 +92,7 @@ format-python:
 	cd ${ROOT_DIR}/sdk/python; python -m black --target-version py37 feast tests
 
 lint-python:
-	cd ${ROOT_DIR}/sdk/python; python -m mypy feast/ tests/
+	cd ${ROOT_DIR}/sdk/python; python -m mypy
```
tpvasconcelos, line 95: this was redundant as it is already configured in setup.cfg

achals:
Thanks @tpvasconcelos! Can you add your sign-off to the commit and force push your branch? that should make the DCO check happy.

sdk/python/setup.cfg
```diff
@@ -16,5 +17,7 @@ select = B,C,E,F,W,T4
 exclude = .git,__pycache__,docs/conf.py,dist,feast/protos
 
 [mypy]
-files=feast,test
+files=feast,tests
 ignore_missing_imports=true
+# ignore mypy error from generated protos modules
+exclude=feast/protos/*
```
woop, line 23: Why is this suddenly necessary?

woop:
Thanks for the PR btw @tpvasconcelos!

sdk/python/setup.cfg
tpvasconcelos: @woop the generated protos directories are not recognized as valid python packages because they are missing the `__init__.py` files. As such, lines such as 
```python
from feast.protos.feast.serving import Connector_pb2 as feast_dot_serving_dot_Connector__pb2
```
are not valid since `Connector_pb2` (module) cannot be imported from `serving` (package). A better solution would be to add empty `__init__.py` files in this generated protos directory tree. 

I didn't spend time to take a look at it properly to see where this could be done or if there is a cleaner way to solve it.

WDYT?

tpvasconcelos: mypy error:
```
feast/protos/feast/serving/Connector_pb2_grpc.py:5: error: Module "feast.protos.feast.serving" has no attribute "Connector_pb2"
```

woop: We used to have `__init__.py` files there but for some reason it was removed after 0.19 was released.

achals: @tpvasconcelos I've created https://github.com/feast-dev/feast/pull/2410/files to generate these `__init__` files; you may need to rebase after that's merged in.

tpvasconcelos: @achals just pinging you on this ☝️

woop: @tpvasconcelos will follow up for you.

achals: @tpvasconcelos ugh sorry. I totally missed this. Thanks for flagging @woop as well. I'll get this fixed.

achals: Actually this shouldn't be needed, because `make install-python-ci-dependencies` does trigger the generation of python protos as expected. So I'm not sure where the error is coming from. I'm going to checkout your PR locally and see if I can reproduce the error.

achals: @tpvasconcelos I was able to reproduce this issue locally, looks like it was a combination of a missing `__init__.py` file when running py 3.7. I'm going to send out a PR to fix it.

achals: #2433

tpvasconcelos:
@woop @achals I think this is ready to merge. Can you take a look?

achals (approved):
/lgtm

achals:
> @woop @achals I think this is ready to merge. Can you take a look?

Just approved. Thanks for the contribution and your patience as we worked through some of these issues!

PR #2410, opened 2022-03-15
fix: Create __init__ files for the proto-generated python dirs

achals:

Python packages without the `__init__` file become "namespace packages". By default, mypy doesn't understand namespace packages when discovering types for typechecking. https://github.com/python/mypy/issues/1645.

We can add the `--namespace-packages` flag to solve this problem, but I'm not sure if this will require all consumers to also add this flag. We're not creating namespace packages to begin with so it seems better to just add the `__init__` files to simplify this setup. This PR updates the `setup.py` to create the `__init__` files.

The only difference is that we need to exclude `types` proto subdirectory because that clashes with an existing module in the python standard library. Apparently mypy confuses the two 🤷🏽 

Unblocks #2405

### DIFF
--- a/sdk/python/setup.py
+++ b/sdk/python/setup.py
@@ -210,11 +210,9 @@ class BuildPythonProtosCommand(Command):
                 self.python_folder,
             ]
             + proto_files,
-            )
+        )
 
     def run(self):
-        from pathlib import Path
-
         for sub_folder in self.sub_folders:
             self._generate_python_protos(f"feast/{sub_folder}/*.proto")
             # We need the __init__ files for each of the generated subdirs


woop:
> The only difference is that we need to exclude types proto subdirectory because that clashes with an existing module in the python standard library. Apparently mypy confuses the two

What if we just moved those proto files into the core folder?

achals:
> > The only difference is that we need to exclude types proto subdirectory because that clashes with an existing module in the python standard library. Apparently mypy confuses the two
> 
> What if we just moved those proto files into the core folder?

I misunderstood your original proposal. I think a better solution would be to rename the package to something like `primitives` or `feast_types`. I like having all the generated code together in a single directory, and wouldn't want to move just the `types` into a different location.

PR #2418, opened 2022-03-17
fix!: GetOnlineFeatureResponse in Python feature server should be consisten…

### LINKED ISSUE #2417: Java and Python implementations of feature servers have inconsistent responses
## Expected Behavior 
Both implementations return the same response.

## Current Behavior
Java implementation returns column oriented feature vectors in [GetOnlineFeatureResponse](https://github.com/feast-dev/feast/blob/master/protos/feast/serving/ServingService.proto#L96), which means that feature vector contains all values for single feature across all rows. And Python implementation returns feature vector with all features from single row.

### Specifications

- Version:
- Platform:
- Subsystem:

pyalex:
…t with Java implementation

FeatureVector in GetOnlineFeatureResponse should consists of only one feature (column). This is how new [GetOnlineFeatureResponse](https://github.com/feast-dev/feast/blob/master/protos/feast/serving/ServingService.proto#L96) proto message was intended to store the data, since values of the same type stored together (as a column) should be easier to operate & compact. Also this is how it's implemented in Java feature server. So this PR's goal is to restore consistency.

Fixes #2417

BREAKING CHANGE: This is a breaking change for those who deploys Python feature server as grpc service. However, Python SDK API remains the same.

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -1347,9 +1347,7 @@ class FeatureStore:
         )
 
         # Populate online features response proto with join keys and request data features
-        online_features_response = GetOnlineFeaturesResponse(
-            results=[GetOnlineFeaturesResponse.FeatureVector() for _ in range(num_rows)]
-        )
+        online_features_response = GetOnlineFeaturesResponse(results=[])
         self._populate_result_rows_from_columnar(
             online_features_response=online_features_response,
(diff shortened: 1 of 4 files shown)


tpvasconcelos:
@pyalex whats the status on this one? Is the plan to merge it soon? Is there a blocker?

adchia (approved):
/lgtm

PR #2420, opened 2022-03-18
feat: Allow using entity's join_key in get_online_features

pyalex:

Join Key of Entity should be used when passing `entity_rows` to `get_online_features` in Python SDK. Thus, request would be consistent with response, which already uses join keys and not entity names.
Usage of entity names is deprecated and will be disabled soon.

Fixes #

### DIFF
--- a/sdk/python/tests/integration/online_store/test_universal_online.py
+++ b/sdk/python/tests/integration/online_store/test_universal_online.py
@@ -72,7 +72,7 @@ def test_entity_ttl_online_store(local_redis_environment, redis_universal_data_s
             "driver_stats:acc_rate",
             "driver_stats:conv_rate",
         ],
-        entity_rows=[{"driver": 1}],
+        entity_rows=[{"driver_id": 1}],
     ).to_df()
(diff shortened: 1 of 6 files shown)


kevjumba (approved):
/lgtm

kevjumba:
Just leaving a todo here, that we also need to change this for the go server I believe? => The go integration tests should not pass here.

felixwang9817 (approved):
/lgtm

sdk/python/tests/integration/online_store/test_universal_online.py
```diff
@@ -715,6 +715,7 @@ def eventually_apply() -> Tuple[None, bool]:
     assert all(v is None for v in online_features["value"])
 
 
+@pytest.mark.skip
```
woop, line 718: Why are we skipping this test now?

sdk/python/tests/integration/online_store/test_universal_online.py
```diff
@@ -937,6 +939,7 @@ def eventually_apply() -> Tuple[None, bool]:
...
```
woop, line 942: Same, why is this test skipped?

PR #2422, opened 2022-03-18
fix: Change location of the proto-generated python code

achals:
The subdirectory only complicates structures because the proto generated imports are not correct (sine python 3 requires absolute imports, but the proto generated imports start with things like `feast.core` instead of `feast.proto.feast.core`.

Fixes #


sdk/python/feast/value_type.py
```diff
@@ -50,12 +61,26 @@ class ValueType(enum.Enum):
...
-ListType = Union[
-    Type[BoolList],
-    Type[BytesList],
-    Type[DoubleList],
-    Type[FloatList],
-    Type[Int32List],
-    Type[Int64List],
-    Type[StringList],
-]
+_value_type_proto_value_type_mapping = {
```
woop, line 64: Is there a reason for these changes?

sdk/python/pyproject.toml
```diff
@@ -2,22 +2,4 @@
 line-length = 88
 target-version = ['py37']
 include = '\.pyi?$'
-exclude = '''
```
woop, line 5: What happened to all these exclusions?

sdk/python/feast/diff/registry_diff.py
```diff
@@ -8,16 +8,14 @@
...
 from feast.feature_service import FeatureService
 from feast.feature_view import DUMMY_ENTITY_NAME
-from feast.protos.feast.core.DataSource_pb2 import DataSource as DataSourceProto
-from feast.protos.feast.core.Entity_pb2 import Entity as EntityProto
-from feast.protos.feast.core.FeatureService_pb2 import (
-    FeatureService as FeatureServiceProto,
-)
-from feast.protos.feast.core.FeatureView_pb2 import FeatureView as FeatureViewProto
-from feast.protos.feast.core.OnDemandFeatureView_pb2 import (
+from feast.proto_core.DataSource_pb2 import DataSource as DataSourceProto
```
woop, line 11: I am a little bit unclear on what problem you are trying to solve with this change? Didn't we have absolute imports before and after?

kevjumba:
@achals I think there is some merge conflicts that need to be resolved and also willem left some comments!

sdk/python/feast/diff/registry_diff.py
achals: While we did, there was a bug that I found later and fixed in #2437.

With this PR - I just think that the subdir structure is a bit stuttery - `feast.protos.feast.*`. It's a taste thing, not a major functionality thing.

achals:
Doesn't seem like this is causing a problem - and the value in minimal, so I'm going to close this PR for now.

PR #2423, opened 2022-03-21
feat: Add `/write-to-online-store` method to the python feature server

tpvasconcelos:
This PR adds a `/write-to-online-store` HTTP POST method to the python feature server.

**Example:** using curl to push fresh feature vectors to the online store:

```shell
curl -X POST "http://localhost:6566/write-to-online-store" -d '{
    "feature_view_name": "driver_hourly_stats",
    "df": {
            "driver_id": [1001],
            "event_timestamp": ["2022-05-13 10:59:42"],
            "created": ["2022-05-13 10:59:42"],
            "conv_rate": [1.0],
            "acc_rate": [1.0],
            "avg_daily_trips": [1000]
    }
  }'
```

Check the table below for the output difference of the `/get-online-features` before and after the operation above:

```shell
$ curl -X POST "http://localhost:6566/get-online-features" -d '{
    "features": [
      "driver_hourly_stats:avg_daily_trips"
    ],
    "entities": {
      "driver_id": [1001]
    }
  }' | jq
```

<table>
<tr>
    <th>Before:</th>
    <th>After:</th>
</tr>

<tr>
<td>

```json
{
  "metadata": {
    "feature_names": [
      "driver_id",
      "avg_daily_trips",
      "conv_rate",
      "acc_rate"
    ]
  },
  "results": [
    {
      "values": [
        1001,
        3,
        0.15759889781475067,
        0.5574429631233215
      ],
      "statuses": [...],
      "event_timestamps": [
        "1970-01-01T00:00:00Z",
        "2022-02-04T10:00:00Z",
        "2022-02-04T10:00:00Z",
        "2022-02-04T10:00:00Z"
      ]
    }
  ]
}
```

</td>
<td>

```json
{
  "metadata": {
    "feature_names": [
      "driver_id",
      "avg_daily_trips",
      "conv_rate",
      "acc_rate"
    ]
  },
  "results": [
    {
      "values": [
        1001,
        1000,
        1,
        1
      ],
      "statuses": [...],
      "event_timestamps": [
        "1970-01-01T00:00:00Z",
        "2022-05-13T10:59:42Z",
        "2022-05-13T10:59:42Z",
        "2022-05-13T10:59:42Z"
      ]
    }
  ]
}
```

</td>
</tr>

</table>

Signed-off-by: Tomas Pereira de Vasconcelos <tomasvasconcelos1@gmail.com> 
@tpvasconcelos

### DIFF
--- a/sdk/python/feast/feature_server.py
+++ b/sdk/python/feast/feature_server.py
@@ -1,14 +1,24 @@
+import json
 import traceback
+import warnings
 
+import pandas as pd
 import uvicorn
 from fastapi import FastAPI, HTTPException, Request
 from fastapi.logger import logger
 from fastapi.params import Depends
 from google.protobuf.json_format import MessageToDict, Parse
+from pydantic import BaseModel
 
 import feast
 from feast import proto_json
 from feast.protos.feast.serving.ServingService_pb2 import GetOnlineFeaturesRequest
+
+
+class WriteToFeatureStoreRequest(BaseModel):
+    feature_view_name: str
+    df: dict
+    allow_registry_cache: bool = True
 
 
 def get_app(store: "feast.FeatureStore"):
@@ -51,6 +61,28 @@ def get_app(store: "feast.FeatureStore"):
             # Convert the Protobuf object to JSON and return it
             return MessageToDict(  # type: ignore
                 response_proto, preserving_proto_field_name=True, float_precision=18
+            )
+        except Exception as e:
+            # Print the original exception on the server side
+            logger.exception(traceback.format_exc())
+            # Raise HTTPException to return the error message to the client
+            raise HTTPException(status_code=500, detail=str(e))
+
+    @app.post("/write-to-online-store")
+    def write_to_online_store(body=Depends(get_body)):
+        warnings.warn(
+            "write_to_online_store is an experimental feature. "
+            "This API is unstable and it could be changed in the future. "
+            "We do not guarantee that future changes will maintain backward compatibility.",
+            RuntimeWarning,
+        )
+        try:
+            request = WriteToFeatureStoreRequest(**json.loads(body))
+            df = pd.DataFrame(request.df)
+            store.write_to_online_store(
+                feature_view_name=request.feature_view_name,
+                df=df,
+                allow_registry_cache=request.allow_registry_cache,
             )
         except Exception as e:
             # Print the original exception on the server side


sdk/python/feast/feature_server.py
```diff
@@ -58,6 +67,22 @@ def get_online_features(body=Depends(get_body)):
             # Raise HTTPException to return the error message to the client
             raise HTTPException(status_code=500, detail=str(e))
 
+    @app.post("/write-to-online-store")
+    def write_to_online_store(body=Depends(get_body)):
+        try:
+            request = WriteToFeatureStoreRequest(**json.loads(body))
```
adchia, line 73: Think this is generally fine, though we'd probably also want to have some warning that this is experimental since we may want to expose a slightly different endpoint to allow for streaming transformations, or pushing directly to the offline store. 


An example warning might be e.g.

```python
warnings.warn(
    "write_to_online_store is an experimental feature. "
    "This API is unstable and it could be changed in the future. "
    "We do not guarantee that future changes will maintain backward compatibility.",
    RuntimeWarning,
)
```

tpvasconcelos: Done ✅

tpvasconcelos:
@adchia I used the same naming convention as in `store.write_to_online_store()` for the endpoint name and JSON fields. WDYT?

adchia:
> @adchia I used the same naming convention as in `store.write_to_online_store()` for the endpoint name and JSON fields. WDYT?

Hey! Can you actually use the `store.push()` method instead for this and rename accordingly? Reason why we wanted the name to be a little more general is we'll likely also want to support in the future pushing to the offline store too.

I'll add a warning on the existing write_to_online_store() method in favor of the push() method.

Also: you still have linter issues

tpvasconcelos:
> Hey! Can you actually use the `store.push()` method instead for this and rename accordingly? Reason why we wanted the name to be a little more general is we'll likely also want to support in the future pushing to the offline store too.

@adchia Will this method allow to **only** push to the online store? As I understand, the future goal of this method is to push to **both** the online and offline stores with a single call right? I would prefer still having an option to only push to the online store. If this is the case, should I just **add** the `/push` method instead of removing the `/write-to-online-store` one?

adchia:
I think that's fine for now as long as both of those methods have an alpha warning attached to the methods.

I think the .push method should probably allow flexibility in pushing to offline or online or both. Today it only pushes to the online store for feature views that depend on a `PushSource`.

Down the road, one example way of allowing pushing to offline stores might be having  a boolean flag that lets you configure pushing to the batch source. e.g. 

```python
click_event_source = PushSource(
    name="click_event_source",
    schema=[
       ...
    ],
    batch_source=BigQuerySource(
			table="my_project.my_Dataset.click_event_table"
		),
    push_to_batch=True # will now also create or append to the batch source.
)
```

And this way, you can have a stream based FeatureView depend on a `Union[PushSource, KafkaSource, KinesisSource]` etc.

adchia (approved):
/lgtm

PR #2415, opened 2022-03-17
fix: Don't prevent apply from running given duplicate empty names in data sources. Also fix repeated apply of Spark data source.

adchia:
**What this PR does / why we need it**:
This PR does several things:
- Prints out more warnings on Feast 0.21 needing users to specify unique names or tables for data sources.
- Converting exceptions to warnings if there are multiple duplicate data source names.
- Adding the `data_source_class_type` to the proto so repeated `feast apply` runs will work on Spark (and other contrib data sources)
- Disabling feast apply/plan output for data source changes until Feast 0.21
- Adding tests for this behavior
- Removing unnecessary logic in `DataSource#from_proto`. When we apply data sources (incl in feature views), we populate the class type, so it should always fall through in the first if condition.

**Which issue(s) this PR fixes**:

Fixes #2407 #2409

PR #2424, opened 2022-03-21
fix: Simplify DataSource.from_proto logic

### LINKED ISSUE #2409: Data source conversion from protobuf doesn't use type information
For some reason we are using field names https://github.com/feast-dev/feast/blob/b35e1e84720523cef70cba6d6306af8f193b469f/sdk/python/feast/data_source.py#L217 to infer data source types. We should be using the actual data source types in the protos.

Also, https://github.com/feast-dev/feast/blob/b35e1e84720523cef70cba6d6306af8f193b469f/sdk/python/feast/data_source.py#L213 doesnt exist

achals:

This simplifies the `DataSource.from_proto` logic 

Fixes #2409

### DIFF
--- a/sdk/python/feast/data_source.py
+++ b/sdk/python/feast/data_source.py
@@ -135,13 +135,13 @@ class KinesisOptions:
 
 
 _DATA_SOURCE_OPTIONS = {
-    "file_options": "feast.infra.offline_stores.file_source.FileSource",
-    "bigquery_options": "feast.infra.offline_stores.bigquery_source.BigQuerySource",
-    "redshift_options": "feast.infra.offline_stores.redshift_source.RedshiftSource",
-    "snowflake_options": "feast.infra.offline_stores.snowflake_source.SnowflakeSource",
-    "kafka_options": "feast.data_source.KafkaSource",
-    "kinesis_options": "feast.data_source.KinesisSource",
-    "request_data_options": "feast.data_source.RequestDataSource",
+    DataSourceProto.SourceType.BATCH_FILE: "feast.infra.offline_stores.file_source.FileSource",
+    DataSourceProto.SourceType.BATCH_BIGQUERY: "feast.infra.offline_stores.bigquery_source.BigQuerySource",
+    DataSourceProto.SourceType.BATCH_REDSHIFT: "feast.infra.offline_stores.redshift_source.RedshiftSource",
+    DataSourceProto.SourceType.BATCH_SNOWFLAKE: "feast.infra.offline_stores.snowflake_source.SnowflakeSource",
+    DataSourceProto.SourceType.STREAM_KAFKA: "feast.data_source.KafkaSource",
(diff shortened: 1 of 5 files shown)


adchia:
I think I also fix this in https://github.com/feast-dev/feast/pull/2415 ? in a different way

sdk/python/feast/data_source.py
```diff
@@ -210,48 +221,18 @@ def from_proto(data_source: DataSourceProto) -> Any:
...
-        elif data_source.snowflake_options.table or data_source.snowflake_options.query:
-            from feast.infra.offline_stores.snowflake_source import SnowflakeSource
-
-            data_source_obj = SnowflakeSource.from_proto(data_source)
-
-        elif (
-            data_source.kafka_options.bootstrap_servers
-            and data_source.kafka_options.topic
-            and data_source.kafka_options.message_format
+        option_type = data_source.WhichOneof("options")
```
woop, line 224: Wouldn't it be possible to use https://github.com/feast-dev/feast/blob/e7dd4b75ba0fbd86338aacf2ecd0cc8979dc803b/protos/feast/core/DataSource.proto#L56 to know the type of data source?

achals: we can use that too. Both are always set, so it's either or. I don't have a preference, do you feel strongly about using `SourceType`?

achals: Updated

sdk/python/feast/data_source.py
```diff
@@ -210,48 +222,20 @@ def from_proto(data_source: DataSourceProto) -> Any:
...
-            and data_source.kinesis_options.region
-            and data_source.kinesis_options.stream_name
-        ):
-            data_source_obj = KinesisSource.from_proto(data_source)
-        else:
             raise ValueError("Could not identify the source type being added.")
 
-        return data_source_obj
+        if data_source_type == DataSourceProto.SourceType.CUSTOM_SOURCE:
+            cls = get_data_source_class_from_type(data_source.data_source_class_type)
```
felixwang9817, line 234: from what I can tell, we're deprecating `data_source_class_type` for all DataSources other than Custom

if that's true, should we modify the logic in `feature_view.py` that sets `data_source_class_type` for all data sources? an even stronger change would be to push `data_source_class_type` into `CustomOptions`

felixwang9817: at the very least we should change the comment in `DataSource.proto` to indicate that this field is only necessary for custom sources, since the map for other sources is maintained in Python

achals: No - we were never setting this field for other types in the first place. That change is being introduced in #2415. As part of that PR, this code path may need to be updated. But that change is not backwards compatible, so I think this will be needed anyway.

felixwang9817 (approved):
/lgtm

PR #2426, opened 2022-03-21
feat: Graduate write_to_online_store out of experimental status

achals:

The online store write code path has been released, and in use by customers for a while, so we're comfortable promoting this feature to "production".

Additionally - this PR also changes the flag parsing logic to allow (and emit warnings) for unrecognized flags. If we didn't do this, graduated flags (which would be in customers' `feature_store.yaml`) would break `feast` commands.

Fixes #


felixwang9817 (approved):
/lgtm

PR #2427, opened 2022-03-21
refactor!: Remove deprecated parameters from Entity and FeatureView

felixwang9817:
**What this PR does / why we need it**: This PR removes the `labels` parameter from Entity and the `input` parameter from FeatureView, both of which were previously deprecated. It also deprecates the option to pass a Duration object as a ttl to a FeatureView.

Fixes #


woop:
This PR sparks joy 🚀 💥

woop:
So @felixwang9817 I think we should flag this change as a breaking change, even if gave users some time to migrate away from these fields. If we just make a semantic release commit with a breaking change, like `refactor!: Remove deprecated parameters...` then semantic release will bump this release to 1.0.0. What I think we should do first is modify the config for https://github.com/semantic-release/commit-analyzer so that breaking changes result in a minor release, then change the PR title here to something like `refactor!: Remove support for labels on entities and input on feature views` so that the changelog message is clear to users

felixwang9817:
> So @felixwang9817 I think we should flag this change as a breaking change, even if gave users some time to migrate away from these fields. If we just make a semantic release commit with a breaking change, like refactor!: Remove deprecated parameters... then semantic release will bump this release to 1.0.0. What I think we should do first is modify the config for https://github.com/semantic-release/commit-analyzer so that breaking changes result in a minor release, then change the PR title here to something like refactor!: Remove support for labels on entities and input on feature views so that the changelog message is clear to users

@woop sounds good - I just changed the prefix to `refactor!`. I'll change the commit analyzer in a follow-up PR.

felixwang9817:
btw @woop I need an approving review for this PR to be merged

woop (approved):

PR #2429, opened 2022-03-21
chore: Use Go-based implementation as embedded extension to Python SDK

pyalex:

Converts Go-based implementation of get online features functionality into library which will be embedded into Python code and called directly from Python SDK (w/o GRPC in the middle). This allows us to also use Python functions as callbacks to call some implementations (like online/offline store connectors) from Go feature server.

Fixes #


Makefile
```diff
@@ -77,13 +77,11 @@ test-python-universal:
 	FEAST_USAGE=False IS_TEST=True python -m pytest -n 8 --integration --universal sdk/python/tests
 
 test-python-go-server:
-	go build -o ${ROOT_DIR}/sdk/python/feast/binaries/server github.com/feast-dev/feast/go/cmd/server
+	go get golang.org/x/tools/cmd/goimports
```
woop, line 80: Should these `go get` statements be in a `install-go-ci-dependencies` command?

go/internal/feast/onlinestore.go
```diff
@@ -15,8 +15,8 @@ type FeatureData struct {
 }
 
 type OnlineStore interface {
-	// OnlineRead reads multiple features (specified in featureReferences) for multiple
-	// entity keys (specified in entityKeys) and returns an array of array of features,
+	// OnlineRead reads multiple Features (specified in featureReferences) for multiple
```
woop, line 18: Why are these being capitalized? Our convention thus far has been to use lower case unless we are referring to classes/objects.

pyalex: some autoreplace, will fix

go/internal/feast/featurestore.go
```diff
@@ -180,80 +176,98 @@ func (fs *FeatureStore) GetOnlineFeatures(ctx context.Context, request *serving.
...
 			}
 		}
 		return featureRefs, nil
 	} else {
-		return features.features, nil
+		return features.Features, nil
+	}
+}
+
+func (fs *FeatureStore) ExtractFeatureRefs(kind interface{}, fullFeatureNames bool) ([]string, error) {
```
woop, line 246: Is there a way that we can make `kind` here a type instead of `interface{}`. Haven't looked to closely at it though.

go/embedded/online_features.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (s *OnlineFeatureService) GetOnlineFeatures(
+	featureRefs []string,
+	featureServiceName string,
+	entities DataTable,
+	fullFeatureNames bool,
+	projectName string,
+	output DataTable) error {
+
+	entitiesRecord, err := readArrowRecord(entities)
```
woop, line 50: Would you mind dropping some comments in this function? I think it would be pretty informative for contributors to understand how it functions and how it will be used.

go/embedded/online_features.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		return err
+	}
+
+	var featureService *feast.FeatureService
+	if featureServiceName != "" {
+		featureService, err = s.fs.GetFeatureService(featureServiceName, projectName)
+	}
+
+	resp, err := s.fs.GetOnlineFeatures(
```
woop, line 67: I guess this has nothing to do with your PR, but do you think it may be more natural for these internal functions to be `fs.GetOnlinFeaturesByFeatureService()` and `fs.GetOnlineFeaturesByFeatureRefs()`?

go/internal/feast/featurestore.go
```diff
@@ -20,17 +26,24 @@ type FeatureStore struct {
...
-}
-
-// A Features struct specifies a list of features to be retrieved from the online store. These features
+// A Features struct specifies a list of Features to be retrieved from the online store. These Features
 // can be specified either as a list of string feature references or as a feature service. String
 // feature references must have format "feature_view:feature", e.g. "customer_fv:daily_transactions".
 type Features struct {
-	features       []string
-	featureService *FeatureService
+	Features       []string
```
woop, line 33: Should this be called `FeatureRefStrings` or something to be more explicit?

go/utils/typeconversion.go
```diff
@@ -0,0 +1,99 @@
...
+	"fmt"
+	"github.com/apache/arrow/go/arrow"
+	"github.com/apache/arrow/go/arrow/array"
+	"github.com/feast-dev/feast/go/protos/feast/types"
+)
+
+func ProtoTypeToArrowType(sample *types.Value) (arrow.DataType, error) {
+	switch sample.Val.(type) {
+	case *types.Value_BytesVal:
+		return arrow.FixedWidthTypes.Boolean, nil
```
woop, line 13: So arrow boolean maps to Feast bytes (as opposed to https://github.com/feast-dev/feast/blob/bb8274147888ba70ee50282653f2aec566672911/protos/feast/types/Value.proto#L34)

go/internal/feast/featurestore.go
```diff
@@ -20,17 +26,24 @@ type FeatureStore struct {
...
+}
+
+type FeatureVector struct {
+	Name       string
+	Values     array.Interface
+	Statuses   []serving.FieldStatus
+	Timestamps []*timestamppb.Timestamp
+}
+
+type featuresAndView struct {
```
kevjumba, line 44: I feel like this should be capitalized to follow convention for datatypes `FeatureAndView`

sdk/python/feast/embedded_go/online_features_service.py
```diff
@@ -0,0 +1,128 @@
...
+
+try:
+    from .lib.embedded import (
+        DataTable,
+        NewOnlineFeatureService,
+        OnlineFeatureServiceConfig,
+    )
+    from .lib.go import Slice_string
+except ImportError:
+    warnings.warn("Couldn't import embedded go library")
```
woop, line 23: Should this be an exception? What happens if the import fails? I see `Datatable` being used below which will cause an exception there.

go/internal/feast/featurestore.go
pyalex: The convention is to capitalize names that should be exposed. This is internal type

go/utils/typeconversion.go
```diff
@@ -0,0 +1,99 @@
...
+				&types.Value{Val: &types.Value_BoolVal{BoolVal: arr.(*array.Boolean).Value(idx)}})
+		}
+	default:
+		return nil, fmt.Errorf("unsupported arrow to proto conversion for type %s", arr.DataType())
+	}
+
+	return values, nil
+}
+
+func protoTypeToArrowType(sample *types.Value) (arrow.DataType, error) {
```
felixwang9817, line 83: is this a duplicate of `ProtoTypeToArrowType` above?

go/internal/feast/featurestore.go
felixwang9817: how about just `FeatureRefs`?

go/internal/feast/featurestore.go
```diff
@@ -20,17 +26,24 @@ type FeatureStore struct {
...
+	view     *FeatureView
+	features []string
 }
 
 type GroupedFeaturesPerEntitySet struct {
```
felixwang9817, line 49: nit: can we have a comment for this struct?

go/internal/feast/featurestore.go
```diff
@@ -39,14 +52,11 @@ type GroupedFeaturesPerEntitySet struct {
 	featureViewNames []string
 	// A list of requested featureName if fullFeatureNames = False or a list of featureViewNameAlias__featureName that share this
 	// entity set
-	featureResponseMeta []string
+	aliasedFeatureNames []string
```
felixwang9817, line 55: nit: line 54 above is a stray comment I believe

go/internal/feast/featurestore.go
```diff
@@ -20,17 +26,24 @@ type FeatureStore struct {
...
 // can be specified either as a list of string feature references or as a feature service. String
 // feature references must have format "feature_view:feature", e.g. "customer_fv:daily_transactions".
 type Features struct {
-	features       []string
-	featureService *FeatureService
+	Features       []string
+	FeatureService *FeatureService
+}
+
+type FeatureVector struct {
```
felixwang9817, line 37: nit: can we have comments for new structs describing their purpose?

go/internal/feast/featurestore.go
felixwang9817: this name is somewhat confusing at first to me - I don't feel strongly, but maybe something like `featureViewAndRefs` would be better?

go/internal/feast/featurestore.go
felixwang9817: where is `ExtractFeatureRefs` being used?

felixwang9817 (commented):
@pyalex some preliminary comments; will continue reviewing

sdk/python/tests/integration/online_store/test_universal_online.py
```diff
@@ -890,138 +883,6 @@ def test_online_retrieval_with_go_server(
     )
 
 
-@pytest.mark.skip
-@pytest.mark.integration
-@pytest.mark.goserver
-def test_online_store_cleanup_with_go_server(go_environment, go_data_sources):
```
kevjumba, line 896: Why do we have to delete these tests? They're the only tests currently working that tests the go feature server.

sdk/python/feast/embedded_go/online_features_service.py
```diff
@@ -0,0 +1,120 @@
...
+import pyarrow as pa
+from pyarrow.cffi import ffi
+
+from feast.errors import FeatureNameCollisionError
+from feast.feature_service import FeatureService
+from feast.online_response import OnlineResponse
+from feast.protos.feast.serving.ServingService_pb2 import GetOnlineFeaturesResponse
+from feast.protos.feast.types.Value_pb2 import RepeatedValue, Value
+from feast.repo_config import RepoConfig
+from tests.unit.test_proto_json import FeatureVector
```
kevjumba, line 12: I think this import should be from `feast.protos.feast.serving.ServingService_pb2.GetOnlineFeaturesResponse.FeatureVector`, not from the tests

go/utils/typeconversion.go
kevjumba: Why not use binary or fixed sized binary for bytes?

go/embedded/online_features.go
pyalex: Added some comments (see fun definition)

go/internal/feast/featurestore.go
pyalex: Added

go/internal/feast/featurestore.go
pyalex: added

go/internal/feast/featurestore.go
pyalex: fixed

sdk/python/tests/integration/online_store/test_universal_online.py
pyalex: What's the purpose of this tests? It doesn't check anything go server related. This is just duplicate of `test_online_store_cleanup`

pyalex: `test_go_server_life_cycle ` was testing process management. There's no process anymore

sdk/python/feast/embedded_go/online_features_service.py
pyalex: Great catch. Fixed

sdk/python/tests/integration/online_store/test_universal_online.py
kevjumba: Oh right I didn't actually check, I was under the impression that we only had 2 go online tests but we had 3, I was mainly worried about online_retrieval_with_go_server but that is still there!

go/utils/typeconversion.go
pyalex: Fixed

go/internal/feast/featurestore.go
pyalex: Updated

felixwang9817 (commented):
more comments, will continue looking

sdk/python/tests/integration/online_store/test_universal_online.py
```diff
@@ -867,12 +865,7 @@ def test_online_retrieval_with_go_server(
             )
 
     entity_rows = [
-        {
-            "driver_id": _driver,
-            "customer_id": _customer,
-            "origin_id": origin,
-            "destination_id": destination,
-        }
+        {"origin_id": origin, "destination_id": destination}
```
felixwang9817, line 868: why was this change made?

sdk/python/tests/conftest.py
```diff
@@ -94,8 +93,6 @@ def pytest_collection_modifyitems(config, items: List[Item]):
     should_run_integration = config.getoption("--integration") is True
     should_run_benchmark = config.getoption("--benchmark") is True
     should_run_universal = config.getoption("--universal") is True
-    should_run_goserver = config.getoption("--goserver") is True
```
felixwang9817, line 97: why are we removing the `goserver` mark? I feel like it will still be useful from a developer perspective to be able to run the specific tests that rely on the go server

if you feel strongly that you want to remove `goserver`, you should also remove `goserver` from Makefile

sdk/python/feast/feature_store.py
```diff
@@ -90,6 +90,10 @@
 warnings.simplefilter("once", DeprecationWarning)
 
 
+if TYPE_CHECKING:
+    from feast.embedded_go.online_features_service import EmbeddedOnlineFeatureServer
```
felixwang9817, line 94: just curious, why do we need to wrap this with `TYPE_CHECKING`?

go/internal/feast/featurestore_test.go
```diff
@@ -50,22 +49,191 @@ func TestGetOnlineFeaturesRedis(t *testing.T) {
 		},
 	}
 
-	featureViewNames := []string{"driver_hourly_stats:conv_rate",
+	featureNames := []string{"driver_hourly_stats:conv_rate",
```
achals, line 52: This test is currently skipped; should we unskip it if it's working locally now?

sdk/python/feast/embedded_go/online_features_service.py
```diff
@@ -0,0 +1,119 @@
...
+        out_ptr_array = int(ffi.cast("uintptr_t", out_c_array))
+
+        in_c_schema = ffi.new("struct ArrowSchema*")
+        in_ptr_schema = int(ffi.cast("uintptr_t", in_c_schema))
+
+        in_c_array = ffi.new("struct ArrowArray*")
+        in_ptr_array = int(ffi.cast("uintptr_t", in_c_array))
+
+        schema._export_to_c(in_ptr_schema)
+        batch._export_to_c(in_ptr_array)
```
achals, line 66: Some comments here would be nice

sdk/python/tests/integration/online_store/test_universal_online.py
pyalex: Entities `customer_id` and `driver_id` are not used in this request. And Go implementation explicitly returns error when some unexpected join keys were passed. In comparison Python implementation allow to pass any entities that exists in the registry, but also throw an error when entity name is not recognized.

sdk/python/tests/conftest.py
pyalex: Yeah, the idea was not to have go dedicated tests at all, and instead have dedicate test environment with flag. Currently the only blocker is support of ODFV on Go side. So I will rollback this for now, until tests are merged.

sdk/python/feast/feature_store.py
pyalex: This enables lazy importing of `EmbeddedOnlineFeatureServer`. In case if somebody would need to run feast locally (set up from GH, eg) and fix something small w/o need to build go library. Not very strong case, but still could be useful.

go/cmd/server/server.go
```diff
@@ -22,5 +25,55 @@ func (s *servingServiceServer) GetFeastServingInfo(ctx context.Context, request
...
+		request.GetEntities(),
+		request.GetFullFeatureNames())
+
+	resp := &serving.GetOnlineFeaturesResponse{
+		Results: make([]*serving.GetOnlineFeaturesResponse_FeatureVector, 0),
+		Metadata: &serving.GetOnlineFeaturesResponseMetadata{
+			FeatureNames: &serving.FeatureList{Val: make([]string, 0)},
+		},
+	}
+	for name, values := range request.Entities {
```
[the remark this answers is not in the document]
pyalex: This loop populate response with entities. Next loop populate with features.

go/internal/feast/featurestore_test.go
pyalex: This doesn't work locally yet, since requires registry, which apparently should be set up externally.

sdk/python/feast/embedded_go/online_features_service.py
pyalex: Added

go/embedded/online_features.go
```diff
@@ -0,0 +1,121 @@
...
+	repoConfig, err := feast.NewRepoConfigFromJSON(conf.RepoPath, conf.RepoConfig)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	fs, err := feast.NewFeatureStore(repoConfig)
+	if err != nil {
+		log.Fatalln(err)
+	}
+	defer fs.DestructOnlineStore()
```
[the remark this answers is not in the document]
pyalex: Removed

go/internal/feast/featurestore_test.go
```diff
@@ -50,22 +49,191 @@ func TestGetOnlineFeaturesRedis(t *testing.T) {
...
-	request := serving.GetOnlineFeaturesRequest{Kind: &featureListRequest, Entities: entities, FullFeatureNames: true}
 
 	fs, err := NewFeatureStore(&config)
 	assert.Nil(t, err)
 	ctx := context.Background()
-	response, err := fs.GetOnlineFeatures(ctx, &request)
+	response, err := fs.GetOnlineFeatures(ctx, featureNames, nil, entities, true)
 	assert.Nil(t, err)
-	assert.NotEmpty(t, response.Results)
+	assert.Len(t, response, 4)
```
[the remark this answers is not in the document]
pyalex: Added comment. It has columnar structure: 3 columns of features and 1 column with join key values

Makefile
```diff
@@ -123,21 +118,26 @@ build-java:
...
 
-# Go SDK
+# Go SDK & embedded
 
 install-go-proto-dependencies:
 	go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.26.0
 	go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.1.0
 
+install-go-ci-dependencies:
+	go get golang.org/x/tools/cmd/goimports
```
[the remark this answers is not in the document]
pyalex: Added ToDo

pyalex:
@achals @felixwang9817 Addressed your comments.

go/cmd/server/server.go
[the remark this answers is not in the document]
felixwang9817: got it - I'm fine with leaving this as is (no need for comment), I would just note that it was super confusing bc the proto definition is
```
message GetOnlineFeaturesResponseMetadata {
    FeatureList feature_names = 1;
}
```
which doesn't really make it seem like there's any entities involved

felixwang9817 (commented):
I'm comfortable with merging in this PR now; I'll let @achals have the final say

thanks @pyalex for addressing my comments!

go/cmd/server/server.go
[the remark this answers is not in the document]
pyalex: yeah, I understand. But this implementation just keeps compatibility with Python implementation that does the same as of this moment.

achals (commented):
looks mostly fine, but the biggest piece is the featurestore.go file which I don't think I have confidence that I've reviewed thoroughly. I'll take a pass over that tomorrow morning.

Makefile
```diff
@@ -123,21 +118,28 @@ build-java:
...
 	pip install grpcio-tools==1.34.0
 
 compile-protos-go: install-go-proto-dependencies install-protoc-dependencies
 	cd sdk/python && python setup.py build_go_protos
 
-compile-go-feature-server: compile-protos-go
-	go mod tidy
-	go build -o ${ROOT_DIR}/sdk/python/feast/binaries/server github.com/feast-dev/feast/go/cmd/server
+compile-go-lib: install-go-proto-dependencies install-go-ci-dependencies
+	python -m install pybindgen
```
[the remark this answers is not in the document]
pyalex: this is embarrassing) fixed

go/internal/feast/featurestore.go
```diff
@@ -622,30 +621,11 @@ func (fs *FeatureStore) transposeResponseIntoColumns(featureData2D [][]FeatureDa
...
-			builder := array.NewBuilder(arrowAllocator, fieldType)
-			err = utils.ProtoValuesToArrowArray(builder, protoValues)
-			if err != nil {
-				return nil, err
-			}
-
-			currentVector.Values = builder.NewArray()
-		} else {
-			currentVector.Values = array.NewNull(numRows)
+		arrowValues, err := utils.ProtoValuesToArrowArray(protoValues, arrowAllocator, numRows)
```
achals, line 624: nit, can we have a different module name instead of `utils`? Seems like it'll become a kitchen sink..

achals (approved):
/lgtm

go/internal/feast/featurestore.go
pyalex: renamed into `types` since it's type utils

PR #2432, opened 2022-03-22
fix: Typos in Feast UI error message

benjamintanweihao:

The error message in the Feast UI refers to `feature_store.json` while in the example (and most feature store repos I've seen so far) are using `feature_store.yaml` instead.

### DIFF
--- a/ui/src/pages/ProjectOverviewPage.tsx
+++ b/ui/src/pages/ProjectOverviewPage.tsx
@@ -51,7 +51,7 @@ const ProjectOverviewPage = () => {
                 body={
                   <p>
                     There was an error loading the Project Configurations.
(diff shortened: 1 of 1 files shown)


felixwang9817:
@benjamintanweihao thanks for the PR! in order for the DCO check to pass, you'll need to [sign your commits](https://github.com/feast-dev/feast/blob/master/CONTRIBUTING.md#signing-off-commits) and then force push to overwrite your current commit

achals (approved):
/lgtm

achals:
Thanks @benjamintanweihao ! If you can fix DCO then this should be good to land.

benjamintanweihao:
> Thanks @benjamintanweihao ! If you can fix DCO then this should be good to land.

Done! <3

PR #2433, opened 2022-03-22
fix: Ensure that __init__ files exist in proto dirs

achals:

This should unblock the issue when mypy is typechecking our code in python 3.7. https://github.com/feast-dev/feast/pull/2405#discussion_r831751241

Fixes #


felixwang9817 (approved):
/lgtm

PR #2434, opened 2022-03-22
ci: Pin setup-gcloud actions to v0 instead of master

felixwang9817:

**What this PR does / why we need it**: Tests on master were recently failing due to the `setup-gcloud` action being pinned to `master`. In the near future, `master` will be renamed to `main`; moreover, we should not be pinning to `master` anyways. This PR pins the `setup-gcloud` action to `v0` instead.

Fixes #


achals (approved):
/lgtm

PR #2435, opened 2022-03-22
fix: Fix registry proto

kevjumba:
Fixes the issue of registry proto caching not working at all. Added refreshes so calling any retrieval getter like list_entities actually looks into the cache to refresh before retrieving possible stale cache entities(quite a major bug that was left in the code). 

Fixes #

### DIFF
--- a/go/internal/feast/featurestore.go
+++ b/go/internal/feast/featurestore.go
@@ -93,7 +93,7 @@ func (fs *FeatureStore) GetOnlineFeatures(ctx context.Context, request *serving.
 		return nil, err
 	}
 
-	fvs, requestedFeatureViews, requestedRequestFeatureViews, requestedOnDemandFeatureViews, err := fs.getFeatureViewsToUse(features, true, false)
+	fvs, requestedFeatureViews, requestedRequestFeatureViews, requestedOnDemandFeatureViews, err := fs.getFeatureViewsToUse(features, false)
 
 	if len(requestedRequestFeatureViews)+len(requestedOnDemandFeatureViews) > 0 {
 		return nil, status.Errorf(codes.InvalidArgument, "on demand feature views are currently not supported")
@@ -217,7 +217,7 @@ func (fs *FeatureStore) parseFeatures(kind interface{}) (*Features, error) {
 		return &Features{features: featureList.Features.GetVal(), featureService: nil}, nil
 	}
 	if featureServiceRequest, ok := kind.(*serving.GetOnlineFeaturesRequest_FeatureService); ok {
-		featureService, err := fs.registry.getFeatureService(fs.config.Project, featureServiceRequest.FeatureService, true)
+		featureService, err := fs.registry.getFeatureService(fs.config.Project, featureServiceRequest.FeatureService)
 		if err != nil {
 			return nil, err
 		}
@@ -254,12 +254,12 @@ func (fs *FeatureStore) getFeatureRefs(features *Features) ([]string, error) {
 		retrieving all feature views. Similar argument to FeatureService applies.
 
 */
-func (fs *FeatureStore) getFeatureViewsToUse(features *Features, allowCache, hideDummyEntity bool) (map[string]*FeatureView, map[*FeatureView][]string, []*RequestFeatureView, []*OnDemandFeatureView, error) {
+func (fs *FeatureStore) getFeatureViewsToUse(features *Features, hideDummyEntity bool) (map[string]*FeatureView, map[*FeatureView][]string, []*RequestFeatureView, []*OnDemandFeatureView, error) {
 	fvs := make(map[string]*FeatureView)
 	requestFvs := make(map[string]*RequestFeatureView)
 	odFvs := make(map[string]*OnDemandFeatureView)
 
-	featureViews, err := fs.listFeatureViews(allowCache, hideDummyEntity)
+	featureViews, err := fs.listFeatureViews(hideDummyEntity)
 	if err != nil {
 		return nil, nil, nil, nil, err
 	}
@@ -267,7 +267,7 @@ func (fs *FeatureStore) getFeatureViewsToUse(features *Features, allowCache, hid
 		fvs[featureView.base.name] = featureView
 	}
 
-	requestFeatureViews, err := fs.registry.listRequestFeatureViews(fs.config.Project, allowCache)
+	requestFeatureViews, err := fs.registry.listRequestFeatureViews(fs.config.Project)
 	if err != nil {
 		return nil, nil, nil, nil, err
 	}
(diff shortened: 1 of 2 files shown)


go/internal/feast/registry.go
```diff
@@ -28,7 +29,7 @@ type Registry struct {
 	cachedFeatureViews         map[string]map[string]*core.FeatureView
 	cachedOnDemandFeatureViews map[string]map[string]*core.OnDemandFeatureView
 	cachedRequestFeatureViews  map[string]map[string]*core.RequestFeatureView
-
+	cachedRegistry             *core.Registry
```
felixwang9817, line 32: I thought the plan was to only keep the cached registry proto, just as in Python - what's the reason for keeping cachedEntities and other maps?

go/internal/feast/registry.go
```diff
@@ -228,89 +248,113 @@ func (r *Registry) listOnDemandFeatureViews(project string) []*OnDemandFeatureVi
...
 
-func (r *Registry) getEntity(project, entityName string) (*Entity, error) {
+func (r *Registry) getEntity(project, entityName string, allowCache bool) (*Entity, error) {
+	_, err := r.getRegistryProto(allowCache)
+	if err != nil {
+		return nil, err
+	}
 	if entities, ok := r.cachedEntities[project]; !ok {
-		return nil, errors.New(fmt.Sprintf("project %s not found in getEntity", project))
+		return nil, fmt.Errorf("project %s not found in getEntity", project)
```
felixwang9817, line 275: nit: if we keep this `cachedEntities` approach, this error message isn't super descriptive

maybe something like `no cached entities found for project %s`?

go/internal/feast/registry.go
```diff
@@ -228,89 +248,113 @@ func (r *Registry) listOnDemandFeatureViews(project string) []*OnDemandFeatureVi
...
-			return nil, errors.New(fmt.Sprintf("onDemandFeatureView %s not found inside project %s", onDemandFeatureViewName, project))
+			return nil, fmt.Errorf("onDemandFeatureView %s not found inside project %s", onDemandFeatureViewName, project)
 		} else {
 			return NewOnDemandFeatureViewFromProto(onDemandFeatureViewProto), nil
 		}
 	}
 }
 
-func (r *Registry) getRequestFeatureView(project, requestFeatureViewName string) (*RequestFeatureView, error) {
+func (r *Registry) getRequestFeatureView(project, requestFeatureViewName string, allowCache bool) (*RequestFeatureView, error) {
```
achals, line 333: I really don't like this `allowCache` bool that we pass around, even in the python implementation. For the go server can we always read the in-memory version `cachedRegistry`, and just support a different couple of methods - one to start a goroutine that refreshes on an interval, and another to force a refresh?

go/internal/feast/registry.go
kevjumba: It turns out we use the map in getEntity() and each of the other get Functionality. I think it makes sense now to generate the map before hand. 
Specifically, this makes it much faster to retrieve entities individually. Before we would compare every entity in the proto for either the project or the name but now we can just use the map instead.

go/internal/feast/registry.go
```diff
@@ -171,17 +179,21 @@ func (r *Registry) listEntities(project string) []*Entity {
...
 
-func (r *Registry) listFeatureViews(project string) []*FeatureView {
-	if featureViewProtos, ok := r.cachedFeatureViews[project]; !ok {
-		return []*FeatureView{}
+func (r *Registry) listFeatureViews(project string, allowCache bool) ([]*FeatureView, error) {
+	_, err := r.getRegistryProto(allowCache)
+	if err != nil {
+		return nil, err
+	}
+	if cachedFeatureViews, ok := r.cachedFeatureViews[project]; !ok {
```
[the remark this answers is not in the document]
kevjumba: Fixed.

go/internal/feast/registry.go
kevjumba: I think this design is probably better and removes latency from registry metadata retrieval and offloads the work of loading the featureviews to the actual cache refresh. 

Specifically, each time GetEntity() or listEntities() is called, it loops through every entity in the registry and compares the project name and entity name to retrieve the specified entities. I think it makes more sense to just keep it as is as a map so then retrieving them is just O(n) in the number of entities/project that matches the name.
What do you think @achals

go/internal/feast/registry.go
kevjumba: So what you're saying is just remove allowCache altogether?

achals: I'd want to yeah, I don't think the interface needs it

go/internal/feast/registry.go
achals: I think exploding out into individual maps is fine in that case. It probably makes access easier (but we should make sure that updates to the individual fields are atomic).

kevjumba: Yup, I locked the load functionality to make sure it's atomic.

go/internal/feast/registry.go
kevjumba: Ok I removed the allowcache interface and added a go routine. We already have a force refresh function and now we always read from the in memory cache.

go/internal/feast/featurestore.go
```diff
@@ -636,26 +649,32 @@ func (fs *FeatureStore) dropUnneededColumns(onlineFeaturesResponse *serving.GetO
 	}
 }
 
-func (fs *FeatureStore) listFeatureViews(allowCache, hideDummyEntity bool) []*FeatureView {
-	featureViews := fs.registry.listFeatureViews(fs.config.Project)
+func (fs *FeatureStore) listFeatureViews(hideDummyEntity bool) ([]*FeatureView, error) {
```
achals, line 652: How come we have a `FeatureStore.listFeatureViews` but no `FeatureStore.listRequestFeatureViews`?

go/internal/feast/registry.go
```diff
@@ -28,9 +30,10 @@ type Registry struct {
 	cachedFeatureViews         map[string]map[string]*core.FeatureView
 	cachedOnDemandFeatureViews map[string]map[string]*core.OnDemandFeatureView
 	cachedRequestFeatureViews  map[string]map[string]*core.RequestFeatureView
-
+	cachedRegistry             *core.Registry
 	cachedRegistryProtoCreated time.Time
```
achals, line 34: Maybe this should be `cachedRegistryProtoUpdated`? Since it's when the cached registry was last updated.

achals (approved):
small nits but looks fine

go/internal/feast/featurestore.go
kevjumba: Added.

go/internal/feast/registry.go
kevjumba: Done.

go/internal/feast/featurestore.go
```diff
@@ -654,14 +654,17 @@ func (fs *FeatureStore) listFeatureViews(hideDummyEntity bool) ([]*FeatureView,
...
 	return featureViews, nil
 }
 
+func (fs *FeatureStore) listRequestFeatureViews() ([]*RequestFeatureView, error) {
+	requestFeatureViews, err := fs.registry.listRequestFeatureViews(fs.config.Project)
+	if err != nil {
+		return requestFeatureViews, err
+	}
+	return requestFeatureViews, nil
+}
```
achals, line 666: can't this just be 
```
	return fs.registry.listRequestFeatureViews(fs.config.Project)
```

kevjumba: Ah originally I added the hideDummyEntity but that's irrelevant here.

achals (approved):
/lgtm

PR #2437, opened 2022-03-22
fix: Update Makefile to cd into python dir before running commands

achals:

Without the `cd`, the makefile commands generates the python/proto code with the wrong path since we rely on `os.getcwd()`.

Fixes #


felixwang9817 (approved):
/lgtm

PR #2440, opened 2022-03-22
feat: Add description, tags, owner fields to all feature view classes

felixwang9817:
**What this PR does / why we need it**: This PR adds the `description`, `tags`, and `owner` fields to all feature view classes: `FeatureView`, `OnDemandFeatureView`, and `RequestFeatureView`.

Fixes #

### DIFF
--- a/sdk/python/feast/feature_view.py
+++ b/sdk/python/feast/feature_view.py
@@ -50,28 +50,36 @@ DUMMY_ENTITY = Entity(
 
 class FeatureView(BaseFeatureView):
     """
-    A FeatureView defines a logical grouping of serveable features.
+    A FeatureView defines a logical group of features.
 
-    Args:
-        name: Name of the group of features.
-        entities: The entities to which this group of features is associated.
+    Attributes:
+        name: The unique name of the feature view.
+        entities: The list of entities with which this group of features is associated.
         ttl: The amount of time this group of features lives. A ttl of 0 indicates that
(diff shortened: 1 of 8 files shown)


sdk/python/feast/feature_view.py
```diff
@@ -50,28 +50,36 @@
...
+    Attributes:
+        name: The unique name of the feature view.
+        entities: The list of entities with which this group of features is associated.
         ttl: The amount of time this group of features lives. A ttl of 0 indicates that
             this group of features lives forever. Note that large ttl's or a ttl of 0
             can result in extremely computationally intensive queries.
         batch_source: The batch source of data where this group of features is stored.
         stream_source (optional): The stream source of data where this group of features
             is stored.
-        features (optional): The set of features defined as part of this FeatureView.
```
achals, line 64: Shouldn't we continue to mark optional params as such?

achals (approved):
Just a nit

sdk/python/feast/feature_view.py
felixwang9817: good catch - I removed this since `features` is an attribute, and it's always defined

if you want, I can add a docstring to the `__init__` functions, which would then mark `features` as an optional parameter

achals: I think that'd be good yeah

felixwang9817: @achals added docstrings

achals (approved):
/lgtm

PR #2442, opened 2022-03-23
chore: Rename inputs parameter to sources for on demand feature views

felixwang9817:
**What this PR does / why we need it**: This PR renames the `inputs` parameter to `sources` for odfvs.

Fixes #


felixwang9817:
there are some conflicts with Go, so I will hold this PR until #2429 is merged

sdk/python/feast/on_demand_feature_view.py
```diff
@@ -64,8 +70,13 @@ def __init__(
         self,
         name: str,
         features: List[Feature],
-        inputs: Dict[str, Union[FeatureView, FeatureViewProjection, RequestDataSource]],
         udf: MethodType,
+        inputs: Optional[
+            Dict[str, Union[FeatureView, FeatureViewProjection, RequestDataSource]]
+        ] = None,
```
achals, line 76: I think making this a `kwarg` from an `arg` is a backwards incompatibile change. e.g.
```
ODFV(
  "my name",
  [...],
  {"input1": some_feature_view}
)
```

will no longer be valid. Are you thinking that this is okay because it's experimental?

felixwang9817: reordering parameters so that this doesn't break

achals (approved):
/lgtm

PR #2443, opened 2022-03-23
chore: Remove unnecessary methods from FeatureView

felixwang9817:

**What this PR does / why we need it**: This PR removes unnecessary methods from `FeatureView`.

Fixes #

### DIFF
--- a/sdk/python/feast/base_feature_view.py
+++ b/sdk/python/feast/base_feature_view.py
@@ -33,8 +33,8 @@ class BaseFeatureView(ABC):
         tags: A dictionary of key-value pairs to store arbitrary metadata.
         owner: The owner of the base feature view, typically the email of the primary
             maintainer.
-        projection: The feature view projection to be applied to this base feature view
-            at retrieval time.
+        projection: The feature view projection storing modifications to be applied to
+            this base feature view at retrieval time.
         created_timestamp (optional): The time when the base feature view was created.
         last_updated_timestamp (optional): The time when the base feature view was last
             updated.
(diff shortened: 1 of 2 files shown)


sdk/python/feast/base_feature_view.py
```diff
@@ -152,15 +152,11 @@ def ensure_valid(self):
 
     def with_name(self, name: str):
         """
-        Renames this feature view by returning a copy of this feature view with an alias
-        set for the feature view name. This rename operation is only used as part of query
-        operations and will not modify the underlying FeatureView.
+        Returns a renamed copy of this base feature view. This renamed copy will only be
```
achals, line 155: Nit, 'should be' so that this comment can be used as guidance for future contributors

sdk/python/feast/feature_view.py
```diff
@@ -213,36 +213,16 @@ def ensure_valid(self):
     def proto_class(self) -> Type[FeatureViewProto]:
         return FeatureViewProto
 
-    def with_name(self, name: str):
```
achals, line 216: Are we sure this method isn't being used anywhere? I'd check with @mavysavydav in case they rely on it internally.

sdk/python/feast/feature_view.py
```diff
@@ -268,40 +248,6 @@ def with_join_key_map(self, join_key_map: Dict[str, str]):
 
         return cp
 
-    def with_projection(self, feature_view_projection: FeatureViewProjection):
```
achals, line 271: Same

sdk/python/feast/base_feature_view.py
felixwang9817: good catch, updated

sdk/python/feast/feature_view.py
felixwang9817: hm I think this method is inherited from `BaseFeatureView`, so should be fine? same with below method

achals (approved):
/lgtm

PR #2448, opened 2022-03-24
fix: Building of go extension fails

pyalex:

Calling `python setup.py build_go_lib` fails with error
```
distutils.errors.DistutilsGetoptError: invalid negative alias 'no-compile': option 'no-compile' not defined
make: *** [compile-go-lib] Error 1
```

Fixes #

### DIFF
--- a/sdk/python/setup.py
+++ b/sdk/python/setup.py
@@ -324,7 +324,7 @@ class BuildGoProtosCommand(Command):
             self._generate_go_protos(f"feast/{sub_folder}/*.proto")
 
 
-class BuildGoEmbeddedCommand(build_py):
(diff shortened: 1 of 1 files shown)


felixwang9817 (approved):
/lgtm

sdk/python/setup.py
```diff
@@ -324,7 +324,7 @@ def run(self):
             self._generate_go_protos(f"feast/{sub_folder}/*.proto")
 
 
-class BuildGoEmbeddedCommand(build_py):
+class BuildGoEmbeddedCommand(Command):
```
achals, line 327: I didn't catch this earlier, what's the difference between build_py and command?

PR #2452, opened 2022-03-25
chore: Add support for push sources in feature views

achals:

Part 2 after #2428 

Fixes #


sdk/python/feast/data_source.py
```diff
@@ -519,18 +519,27 @@ class PushSource(DataSource):
...
+    batch_source: DataSource
+    event_timestamp_column: str
 
     def __init__(
         self,
         name: str,
         schema: Dict[str, ValueType],
-        batch_source: Optional[DataSource] = None,
+        batch_source: DataSource,
+        event_timestamp_column="timestamp",
```
felixwang9817, line 530: nit: surprised lint didn't catch this, but I feel like it should be `event_timestamp_column = "timestamp"`?

also, can you add docstring for `__init__` here?

sdk/python/feast/feature_store.py
```diff
@@ -1144,6 +1144,27 @@ def tqdm_builder(length):
                 feature_view, self.project, start_date, end_date,
             )
 
+    @log_exceptions_and_usage
+    def push(self, push_source_name: str, df: pd.DataFrame):
+        data_sources = self.registry.list_data_sources(self.project, True)
```
felixwang9817, line 1149: just use `self.get_data_source`?

sdk/python/feast/feature_store.py
```diff
@@ -1144,6 +1144,27 @@ def tqdm_builder(length):
...
+        push_source = {
```
felixwang9817, line 1150: shouldn't we expect there to be only one PushSource since DS are unique by name?

sdk/python/feast/feature_store.py
```diff
@@ -1144,6 +1144,27 @@ def tqdm_builder(length):
...
+
+        all_fvs = self.registry.list_feature_views(self.project, True)
+        fvs_with_push_sources = {
+            fv for fv in all_fvs if fv.stream_source in push_source
+        }
+
+        for fv in fvs_with_push_sources:
+            entities = []
+            for entity_name in fv.entities:
+                entities.append(self._registry.get_entity(entity_name, self.project))
```
felixwang9817, line 1164: use `self.get_entity`

ideally also change the below code in `write_to_online_store` to use these FS methods instead of accessing registry directly

sdk/python/feast/feature_view.py
```diff
@@ -88,7 +88,7 @@ def __init__(
         name: str,
         entities: List[str],
         ttl: Union[Duration, timedelta],
-        batch_source: DataSource,
+        batch_source: Optional[DataSource] = None,
```
felixwang9817, line 91: nit: please change docstring

sdk/python/feast/feature_view.py
```diff
@@ -121,15 +121,28 @@ def __init__(
...
 
+        if stream_source is not None and isinstance(stream_source, PushSource):
+            assert stream_source.batch_source is not None
+            assert isinstance(stream_source.batch_source, DataSource)
+            self.batch_source = stream_source.batch_source
+        else:
+            assert batch_source is not None
+            self.batch_source = batch_source
+
+        if not self.batch_source:
```
felixwang9817, line 132: this error should be switched with the above assert I think

sdk/python/feast/repo_operations.py
```diff
@@ -112,6 +113,8 @@ def parse_repo(repo_root: Path) -> RepoContents:
                 res.data_sources.add(obj)
             if isinstance(obj, FeatureView):
                 res.feature_views.add(obj)
+                if isinstance(obj.stream_source, PushSource):
```
felixwang9817, line 116: maybe add a comment on how the batch source within a PushSource doesn't need to be defined on its own?

felixwang9817 (changes requested):
looks good overall - at some point we should add (a) docs for PushSource and (b) deprecation warning for `write_to_online_store`, happy for those to be in future PR - up to you

sdk/python/feast/feature_store.py
achals: done

sdk/python/feast/feature_view.py
achals: done

sdk/python/feast/repo_operations.py
achals: I'll add this to the docs, in addition to the docstring fro feature views.

sdk/python/tests/integration/online_store/test_push_online_retrieval.py
```diff
@@ -0,0 +1,44 @@
...
+    df_ingest = pd.DataFrame(data)
+
+    store.push("location_stats_push_source", df_ingest)
+
+    online_resp = store.get_online_features(
+        features=["pushable_location_stats:temperature"],
+        entity_rows=[{"location_id": 1}],
+    )
+    online_resp_dict = online_resp.to_dict()
+    print(online_resp_dict)
```
felixwang9817, line 42: remove print

sdk/python/tests/integration/online_store/test_push_online_retrieval.py
```diff
@@ -0,0 +1,44 @@
...
+from tests.integration.feature_repos.universal.entities import (
+    customer,
+    driver,
+    location,
+)
+
+
+@pytest.mark.integration
+@pytest.mark.universal
+def test_historical_features_with_missing_request_data(
```
felixwang9817, line 18: this test seems to be named incorrectly?

ideally I would also like to see a quick docstring of the test explaining what it does, see e.g. `test_online_store_cleanup` for what I mean

also does this need to be in a new file? or could we also squeeze it into `test_online_retrieval.py`

felixwang9817 (commented):
some more nits

sdk/python/feast/feature_store.py
```diff
@@ -1144,6 +1155,18 @@ def tqdm_builder(length):
                 feature_view, self.project, start_date, end_date,
             )
 
+    @log_exceptions_and_usage
+    def push(self, push_source_name: str, df: pd.DataFrame):
```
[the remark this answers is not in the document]
achals: good catch

sdk/python/tests/integration/online_store/test_push_online_retrieval.py
achals: yeah lol

sdk/python/tests/integration/online_store/test_push_online_retrieval.py
achals: renamed, and kept in a different file just so that it's easier to read. It doesn't add any overhead otherwise

felixwang9817 (approved):
/lgtm

PR #2456, opened 2022-03-25
fix: Add inlined data sources to the top level registry

### LINKED ISSUE #2453: Inline data sources not added to the registry fields
## Expected Behavior 
If I create a feature view with an inline data source, it should be added to the registry and discoverable via `feature_store.get_data_source`.

## Current Behavior
Only data sources defined as top level objects of the feature repo are stored and discoverable in the registry.

## Possible Solution
In registry.py, introspect all feature views and collect them to be added to the registry.

achals:

Fixes #2453


felixwang9817 (approved):
/lgtm

PR #2457, opened 2022-03-26
fix: Preserve ordering of features in _get_column_names

mavysavydav:
We need _get_column_names in provider.py to retain the order of feature list since our in-house code uses that method and relies on consistent ordering. In a recent commit, _get_column_names was modified and the order was no longer preserved. 

Fixes #


achals:
@mavysavydav can you update the pr description with why this change is needed?

mavysavydav:
@achals srry I should have made it clearer that this PR was in draft mode. Added the description and going to take it out of draft mode now. Ready for review

achals (approved):
/lgtm

PR #2459, opened 2022-03-26
chore: Build platform specific python package

pyalex:

Since we packaging python SDK with binary part (Go emebedded library) we need to publish package for each supported platform.

Fixes #

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -130,11 +130,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ 3.7, 3.8, 3.9 ]
+        python-version: [ 3.7, 3.8, 3.9, 3.10 ]
         os: [ ubuntu-latest, macOS-latest ]
+        compile-go: [ True ]
+        include:
+          - python-version: 3.7
+            os: ubuntu-latest
+            compile-go: False
     env:
       TWINE_USERNAME: __token__
       TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+      COMPILE_GO: ${{ matrix.compile-go }}
     steps:
       - uses: actions/checkout@v2
       - name: Setup Python
@@ -167,12 +173,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ 3.7, 3.8, 3.9 ]
+        python-version: [ 3.7, 3.8, 3.9, 3.10 ]
         os: [ ubuntu-latest, macOS-latest ]
+        compile-go: [ True ]
+        include:
+          - python-version: 3.7
+            os: ubuntu-latest
+            compile-go: False
     needs: get-version
     env:
       TWINE_USERNAME: __token__
       TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+      COMPILE_GO: ${{ matrix.compile-go }}
     steps:
       - uses: actions/checkout@v2
       - name: Setup Python
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -131,7 +131,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [ 3.7, 3.8, 3.9 ]
-        os: [ ubuntu-latest, macOS-latest, windows-latest ]
+        os: [ ubuntu-latest, macOS-latest ]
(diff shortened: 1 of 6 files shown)


go.sum
```diff
@@ -295,8 +295,8 @@ github.com/prometheus/procfs v0.0.0-20181005140218-185b4288413d/go.mod h1:c3At6R
 github.com/prometheus/procfs v0.0.0-20190117184657-bf6a532e95b1/go.mod h1:c3At6R/oaqEKCNdg8wHV1ftS6bRYblBhIjjI8uT2IGk=
 github.com/prometheus/procfs v0.0.2/go.mod h1:TjEm7ze935MbeOT/UhFTIMYKhuLP4wbCsTZCD3I8kEA=
 github.com/prometheus/procfs v0.0.8/go.mod h1:7Qr8sr6344vo1JqZ6HhLceV9o3AJ1Ff+GxbHq6oeK9A=
-github.com/pyalex/gopy v0.4.1-0.20220315220227-146bcd99e9f5 h1:+N63y8ub8/tUt4Z+U0WkoLpqX08biafsKfr1ZVe3Uus=
-github.com/pyalex/gopy v0.4.1-0.20220315220227-146bcd99e9f5/go.mod h1:ZO6vpitQ61NVoQP/2yOubPS6ET5pP3CAWCiMYn5eqCc=
+github.com/pyalex/gopy v0.4.1-0.20220328211859-c42fbc0e8c26 h1:HGVPQbIy3+ccH3rqX1K0ZE9mqDBubggTf2ZYaXTOaYE=
+github.com/pyalex/gopy v0.4.1-0.20220328211859-c42fbc0e8c26/go.mod h1:ZO6vpitQ61NVoQP/2yOubPS6ET5pP3CAWCiMYn5eqCc=
```
achals, line 299: Should we aim to get your changes upstreamed here? Or fork this repo into the feast-dev org?

pyalex: https://github.com/pyalex/gopy/commit/d705e6cd1d9bc95b5452fc60342315b40dd2c89b
This is the diff from gopy. Basically it uses not correct LD flags - same flags as used during building Python extension. Previously gopy used LD flags for compiling Python itself.

I'm happy to fork gopy in feast-dev for better visibility. @woop @adchia wdyt?

achals: If they flags were incorrect, would upstream the change, and eventually move to the main repo? And we fork into feast-dev to be unblocked?

pyalex: Will move gopy to feast-dev for now. Will prepare PRs to upstream a little later.

.github/workflows/publish.yml
```diff
@@ -126,37 +126,73 @@ jobs:
         run: ./infra/scripts/helm/push-helm-charts.sh $VERSION_WITHOUT_PREFIX
 
   publish-python-sdk:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [ 3.7, 3.8, 3.9 ]
```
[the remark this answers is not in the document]
pyalex: Correct. I can add 3.10 here but it means that we should also run integration tests on this version.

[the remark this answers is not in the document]
achals: Okay.. and is it possible for us to publish a catch-all version that disables the go feature server?

.github/workflows/publish.yml
```diff
@@ -126,37 +126,85 @@ jobs:
...
 
   publish-python-sdk:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [ 3.7, 3.8, 3.9, 3.10 ]
+        os: [ ubuntu-latest, macOS-latest ]
+        compile-go: [ True ]
```
achals, line 135: The alpha flag will still need to be enabled to use the go component, right?

pyalex: from what I see in the code - there's separate flag `go_feature_server` on the root level of `feature_store.yaml`. It doesn't require alpha flag to be set, though

.github/workflows/publish.yml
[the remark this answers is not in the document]
pyalex: Added no platform version

achals (approved):
/lgtm

PR #2467, opened 2022-03-30
fix: Make name a keyword arg for the Entity class

achals:

This change is backwards compatible, but produces a deprecation warning if the name is specified as a positional arg to the `__init__` method.

Fixes #

### DIFF
--- a/go/internal/feast/ondemandfeatureview.go
+++ b/go/internal/feast/ondemandfeatureview.go
@@ -6,7 +6,7 @@ import (
 )
 
 type OnDemandFeatureView struct {
-	base                        *BaseFeatureView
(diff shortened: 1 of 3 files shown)


kevjumba:
Resstarted the java tests

kevjumba:
@achals There seems to be a build issue with the java integration tests.

achals:
> @achals There seems to be a build issue with the java integration tests.

Ugh yeah. Same failure in https://github.com/feast-dev/feast/runs/5759198528?check_suite_focus=true.

achals:
> @achals There seems to be a build issue with the java integration tests.

Fixed by #2469

PR #2470, opened 2022-03-30
fix: Update file api

kevjumba:
Change file options param file_url to uri. Deprecates file_url immediately because file_url is not user facing and is only used in FileSource.
DOES NOT update the API for java(i.e java still uses fileurl).

Fixes #

### DIFF
--- a/protos/feast/core/DataSource.proto
+++ b/protos/feast/core/DataSource.proto
@@ -89,10 +89,18 @@ message DataSource {
     // s3://path/to/file for AWS S3 storage
     // gs://path/to/file for GCP GCS storage
     // file:///path/to/file for local storage
+    // Deprecated in Feast 0.23.0
     string file_url  = 2;
 
     // override AWS S3 storage endpoint with custom S3 endpoint
     string s3_endpoint_override  = 3;
+
+    // Target URL of file to retrieve and source features from.
+    // s3://path/to/file for AWS S3 storage
+    // gs://path/to/file for GCP GCS storage
+    // file:///path/to/file for local storage
+    // Deprecated in Feast 0.23.0
+    string uri = 4;
   }
 
   // Defines options for DataSource that sources features from a BigQuery Query
--- a/protos/feast/core/DataSource.proto
+++ b/protos/feast/core/DataSource.proto
@@ -89,18 +89,10 @@ message DataSource {
     // s3://path/to/file for AWS S3 storage
     // gs://path/to/file for GCP GCS storage
     // file:///path/to/file for local storage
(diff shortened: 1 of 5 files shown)


protos/feast/core/DataSource.proto
```diff
@@ -89,10 +89,18 @@ message DataSource {
     // s3://path/to/file for AWS S3 storage
     // gs://path/to/file for GCP GCS storage
     // file:///path/to/file for local storage
+    // Deprecated in Feast 0.23.0
     string file_url  = 2;
```
achals, line 92: You should change this in place; serialized protobufs only use the index number not the field name.

achals (changes requested):

protos/feast/core/DataSource.proto
kevjumba: Ah I didn't know that but that makes a lot of sense! Thanks for the suggestion that makes the update so much easier.

sdk/python/feast/infra/offline_stores/file_source.py
```diff
@@ -180,19 +179,20 @@ class FileOptions:
     def __init__(
         self,
         file_format: Optional[FileFormat],
-        file_url: Optional[str],
         s3_endpoint_override: Optional[str],
+        uri: Optional[str],
```
achals, line 183: Wanted to confirm that this is not user facing?

achals (commented):
just a question, but looks good otherwise.

sdk/python/feast/infra/offline_stores/file_source.py
kevjumba: Yup isn't called anymore but maybe @adchia can also confirm?

sdk/python/feast/infra/offline_stores/file_source.py
```diff
@@ -182,17 +183,31 @@ def __init__(
...
+                (
+                    "The option to pass a file_url parameter to FileOptions is being deprecated. "
+                    "Please pass the file url to the uri parameter instead. The parameter will be deprecated in Feast 0.23"
+                ),
+                DeprecationWarning,
+            )
         self._file_url = file_url
+        self._uri = uri
+        if not self._uri:
+            self._uri = self._file_url
```
achals, line 210: you can do `self._uri = uri or file_url` and probably not store `self._file_url` at all

kevjumba: Great catch! thanks!

achals (approved):
/lgtm

PR #2472, opened 2022-03-31
fix: Update feature view APIs to prefer keyword args

achals:

Kwargs are the preferred way to structure APIs for forwards/backwards compatibility reasons.

Fixes #

### DIFF
--- a/sdk/python/feast/feature_view.py
+++ b/sdk/python/feast/feature_view.py
@@ -87,9 +87,10 @@ class FeatureView(BaseFeatureView):
     @log_exceptions
     def __init__(
         self,
-        name: str,
-        entities: List[str],
-        ttl: Union[Duration, timedelta],
+        *args,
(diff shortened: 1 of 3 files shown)


sdk/python/feast/feature_view.py
```diff
@@ -121,6 +122,56 @@ def __init__(
...
+            warnings.warn(
+                (
+                    "feature view parameters should be specified as a keyword argument instead of a positional arg."
+                    "Feast 0.23+ will not support positional arguments to construct feature views"
+                ),
+                DeprecationWarning,
+            )
+            if len(args) > len(positional_attributes):
+                raise ValueError(
+                    f"Only {', '.join(positional_attributes)} are allowed positional args when defining "
```
kevjumba, line 142: nit: allowed as positional args

kevjumba (approved):

PR #2475, opened 2022-04-01
feat: Add Feast types in preparation for changing type system

felixwang9817:
**What this PR does / why we need it**: This PR adds the `FeastType` ABC, along with associated types, in preparation for changing the Feast type system. Specifically, this PR prepares for introducing a `Field` class which will replace the existing `Feature` class; the new `Field` class will allow user to specify types through `FeastType`s, instead of through the existing `ValueType` enum. Note that the user interface for specifying types will change, but the underlying proto representation will remain the same.

Fixes #

### DIFF
--- a/sdk/python/feast/types.py
+++ b/sdk/python/feast/types.py
@@ -1,0 +1,154 @@
+# Copyright 2022 The Feast Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from abc import ABC, abstractmethod
+
+from feast.protos.feast.types.Value_pb2 import ValueType as ValueTypeProto
+
+PRIMITIVE_FEAST_TYPES_TO_VALUE_TYPES = {
+    "Invalid": "INVALID",
+    "String": "STRING",
+    "Bytes": "BYTES",
+    "Bool": "BOOL",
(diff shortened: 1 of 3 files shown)


sdk/python/feast/types.py
```diff
@@ -0,0 +1,117 @@
...
+
+    INVALID = 0
+    BYTES = 1
+    STRING = 2
+    BOOL = 3
+    INT32 = 4
+    INT64 = 5
+    FLOAT32 = 6
+    FLOAT64 = 7
+    UNIX_TIMESTAMP = 8
```
achals, line 64: Does this numbering match the proto equivalents?

sdk/python/feast/types.py
```diff
@@ -0,0 +1,117 @@
...
+    "BOOL": "BOOL",
+    "INT32": "INT32",
+    "INT64": "INT64",
+    "FLOAT32": "FLOAT",
+    "FLOAT64": "DOUBLE",
+    "UNIX_TIMESTAMP": "UNIX_TIMESTAMP",
+}
+
+
+class FeastType(ABC):
```
achals, line 33: Should this be FeastComplexType or something

sdk/python/feast/types.py
```diff
@@ -0,0 +1,117 @@
...
+
+class Array(FeastType):
+    """
+    An Array represents a list of types.
+
+    Attributes:
+        base_type: The base type of the array.
+    """
+
+    base_type: Union[PrimitiveFeastType, FeastType]
```
achals, line 103: Does this mean we'd support Nested Arrays, for example?

sdk/python/feast/types.py
felixwang9817: switched to match

sdk/python/feast/types.py
felixwang9817: changed

sdk/python/feast/types.py
felixwang9817: eventually, yes - the user interface should be extensible enough to support nested types, but right now we only support the existing types (so no nested types)

achals (approved):
/lgtm

PR #2480, opened 2022-04-02
fix: Add spark to lambda dockerfile

achals:

There were two issues with the AWS Lambda feature server tests:
- The image would fail to start because of missing pyspark dependencies:
```
[ERROR] Runtime.ImportModuleError: Unable to import module 'app': No module named 'pyspark' Traceback (most recent call last):
```

This was temporarily fixed by adding pyspark to the image used for Lambda. Note that this should be removed as soon as possible since the image size doubled as a result of adding this dependency.

- After fixing the previous issue, we'd see this kind of error:
```
Exception: Failed to get online features from remote feature server ***'detail': 'Failed to parse entities field: from_json_object() takes 3 positional arguments but 4 were given.'
```
This was traced to an unpinned version of protobuf - `3.20.0` changed the API in the proto_json module, which we were patching in the feature server. We've temporarily constrained the upper bound of protobuf used but should follow up with an upgrade to the version of protobuf along with a fix to the patched version.

Fixes #

### DIFF
--- a/sdk/python/feast/proto_json.py
+++ b/sdk/python/feast/proto_json.py
@@ -15,6 +15,8 @@ ProtoMessage = Any
 JsonObject = Any
 
 
+# TODO: These methods need to be updated when bumping the version of protobuf.
+# https://github.com/feast-dev/feast/issues/2484
 def _patch_proto_json_encoding(
     proto_type: Type[ProtoMessage],
     to_json_object: Callable[[_Printer, ProtoMessage], JsonObject],
--- a/sdk/python/feast/proto_json.py
+++ b/sdk/python/feast/proto_json.py
@@ -140,7 +140,7 @@ def _patch_feast_repeated_value_json_encoding():
         return [printer._MessageToJsonObject(item) for item in message.val]
 
     def from_json_object(
-        parser: _Parser, value: JsonObject, message: ProtoMessage
+        parser: _Parser, value: JsonObject, message: ProtoMessage, *args
     ) -> None:
         array = value if isinstance(value, list) else value["val"]
         for item in array:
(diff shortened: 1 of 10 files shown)


sdk/python/feast/proto_json.py
```diff
@@ -140,10 +140,12 @@ def to_json_object(printer: _Printer, message: ProtoMessage) -> JsonObject:
         return [printer._MessageToJsonObject(item) for item in message.val]
 
     def from_json_object(
-        parser: _Parser, value: JsonObject, message: ProtoMessage
+        parser: _Parser, value: JsonObject, message: ProtoMessage,
     ) -> None:
+        print("going to convert message")
```
felixwang9817, line 145: nit: remove `print`

sdk/python/setup.py
```diff
@@ -52,7 +52,7 @@
     "mmh3",
     "pandas>=1.0.0",
     "pandavro==1.5.*",
-    "protobuf>=3.10",
+    "protobuf>=3.10,<3.20",
```
felixwang9817, line 55: nit: do we want to also update `py3.7-requirements.txt` and other files?

felixwang9817 (changes requested):

sdk/python/feast/proto_json.py
achals: good catch

sdk/python/setup.py
achals: oh yeah good point

achals: just read through what the `make lock-python-*dependencies` command does and existing requirements file - I don't think any changes are needed here.

felixwang9817 (approved):
/lgtm

PR #2486, opened 2022-04-04
fix: Timestamp update

kevjumba:
Updates data source api by replacing `event_timestamp_column` with `timestamp_field` param.

Fixes #


kevjumba:
I just want to run integration tests to find any major bugs that need to be fixed.

achals:
You may need to change https://github.com/feast-dev/feast/blob/master/java/serving/src/test/resources/docker-compose/feast10/definitions.py as well.

kevjumba:
> https://github.com/feast-dev/feast/blob/master/java/serving/src/test/resources/docker-compose/feast10/definitions.py

Hm I believe my pr does update the param.

sdk/python/feast/data_source.py
```diff
@@ -384,14 +396,15 @@ def __eq__(self, other):
...
+            event_timestamp_column=data_source.timestamp_field,
             field_mapping=dict(data_source.field_mapping),
             bootstrap_servers=data_source.kafka_options.bootstrap_servers,
             message_format=StreamFormat.from_proto(
                 data_source.kafka_options.message_format
             ),
             topic=data_source.kafka_options.topic,
-            event_timestamp_column=data_source.event_timestamp_column,
-            created_timestamp_column=data_source.created_timestamp_column,
+            created_timestamp_column=data_source.timestamp_field,
```
achals, line 406: is this right? Shouldn't this be `data_source.created_timestamp_column`?

kevjumba: Ah good catch, fixed.

kevjumba:
@achals Could you stamp this?

PR #2487, opened 2022-04-04
fix: Snowflake api update

kevjumba:
Updates snowflake source api to take `warehouse` parameter specified instead of needing to take it through `feature_store.yaml`

Fixes #


achals (approved):
/lgtm

PR #2489, opened 2022-04-05
chore: Add conversion from ValueTypeProto.Enum to new Feast type system

felixwang9817:

**What this PR does / why we need it**: #2475 added a new type system but did not include a method to convert from the existing type system enums (`ValueTypeProto.Enum`) to the new type system. This PR adds that conversion method.

Fixes #

### DIFF
--- a/sdk/python/feast/types.py
+++ b/sdk/python/feast/types.py
@@ -40,17 +40,21 @@ class ComplexFeastType(ABC):
         pass
 
     @abstractmethod
-    def to_int(self) -> int:
+    def to_value_type(self) -> ValueTypeProto.Enum:
         """
-        Converts a ComplexFeastType object to the appropriate int value corresponding to
-        the correct ValueTypeProto.Enum value.
+        Converts a ComplexFeastType object to the corresponding ValueTypeProto.Enum value.
         """
         raise NotImplementedError
+
+    def __eq__(self, other):
+        return self.to_value_type() == other.to_value_type()
 
 
 class PrimitiveFeastType(Enum):
     """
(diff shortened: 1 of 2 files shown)


sdk/python/feast/types.py
```diff
@@ -40,30 +40,37 @@ def __init__(self):
...
 
+    def __eq__(self, other):
+        return self.to_value_type() == other.to_value_type()
+
 
 class PrimitiveFeastType(Enum):
     """
     A PrimitiveFeastType represents a primitive type in Feast.
+
+    Note that these values must match the values in ValueTypeProto.Enum.
```
achals, line 57: Add a link to the file? Or path?

sdk/python/feast/types.py
```diff
@@ -110,8 +117,35 @@ def __init__(self, base_type: Union[PrimitiveFeastType, ComplexFeastType]):
...
+    Raises:
+        ValueError: The conversion could not be performed.
+    """
+    # Primitive types can be directly converted.
+    primitive_values = [primitive_type.value for primitive_type in PrimitiveFeastType]
+    if value_type in primitive_values:
+        return PrimitiveFeastType(value_type)
+
+    # Complex types must be constructed. Currently only arrays are supported. Note that
+    # enum values for arrays are precisely the enum value for the array's base type plus 10.
```
achals, line 145: Where does this 10 come from?

achals: Honestly a map is better here because this is sure to break if we add enums that aren't supported or something. We should have a map, and a test that iterates over all possible enum values and ensures that you get a valid feast type.

achals (approved):
/lgtm

PR #2493, opened 2022-04-05
ci: Use testcontainers to spin up redis and datastore instances 

achals:

Running integration tests for contributors has been challenging since they don't always have access to GCP or AWS. This makes development or contributions harder than they need to be.

Additionally, there's some infrastructure that's needed to run tests locally - such as redis. This shouldn't have to be installed by contributors explicitly if it's not necessary.

Fixes #

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -128,11 +128,14 @@ There are two sets of tests you can run:
 2. Full integration tests (requires cloud environment setups)
 
 #### Local integration tests
-To get local integration tests running, you'll need to have Redis setup:
+To get local integration tests running, you'll need to have Redis and Docker setup:
 
 Redis
 1. Install Redis: [Quickstart](https://redis.io/topics/quickstart)
 2. Run `redis-server`
+
+Docker:
+1. Install Docker: [Get Docker](https://docs.docker.com/get-docker/)
 
 Now run `make test-python-universal-local`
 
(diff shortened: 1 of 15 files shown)


sdk/python/tests/integration/feature_repos/universal/online_store/redis.py
```diff
@@ -0,0 +1,26 @@
...
+    OnlineStoreCreator,
+)
+
+
+class RedisOnlineStoreCreator(OnlineStoreCreator):
+    def __init__(self, project_name: str):
+        super().__init__(project_name)
+        self.container = DockerContainer("redis").with_exposed_ports("6379")
+
+    def create_online_store(self) -> typing.Dict[str, str]:
```
felixwang9817, line 16: nit: `from typing import Dict` instead of `typing.Dict`, and same elsewhere

sdk/python/tests/integration/feature_repos/universal/data_source_creator.py
```diff
@@ -9,6 +9,9 @@
 
 
 class DataSourceCreator(ABC):
+    def __init__(self, project_name: str):
```
felixwang9817, line 12: what's the reason for this being pulled out?

sdk/python/tests/integration/feature_repos/repo_configuration.py
```diff
@@ -341,9 +363,18 @@ def construct_test_environment(
...
+    if test_repo_config.online_store_creator:
+        online_creator = test_repo_config.online_store_creator(project)
+        online_store = (
+            test_repo_config.online_store
+        ) = online_creator.create_online_store()
+    else:
+        online_creator = None
+        online_store = test_repo_config.online_store
+
+    print(f"Online Store creator: {online_creator}, online store: {online_store}")
```
felixwang9817, line 377: nit: stray print

CONTRIBUTING.md
```diff
@@ -128,12 +128,15 @@ There are two sets of tests you can run:
 2. Full integration tests (requires cloud environment setups)
 
 #### Local integration tests
-To get local integration tests running, you'll need to have Redis setup:
+To get local integration tests running, you'll need to have Redis and Docker setup:
```
felixwang9817, line 131: I think this is actually outdated - the command run `make test-python-universal-local` no longer requires Redis or Docker

felixwang9817 (approved):
/lgtm

PR #2446, opened 2022-03-24
chore: Go Sql Online Store

kevjumba:
**What this PR does / why we need it**:
Implements a basic sql online store that we can use to perform integration testing within Go instead of requiring entry point to be through Python. (This makes it quite hard to do Go specific testing). Also contains integration tests for the sql online store and the getonlinefeatures grpc service to make sure the correct data is retrieved.

**Which issue(s) this PR fixes**:

Fixes #

PR #2494, opened 2022-04-05
feat: Support on demand feature views in go feature server

pyalex:

This PR enables using on demand feature transformations with go feature server. Now Go feature server has equal functionality with python implementation and passes all existing e2e tests.

Fixes #


sdk/python/tests/integration/feature_repos/repo_configuration.py
```diff
@@ -67,6 +67,7 @@
     # Local configurations
     IntegrationTestRepoConfig(),
     IntegrationTestRepoConfig(python_feature_server=True),
+    IntegrationTestRepoConfig(online_store=REDIS_CONFIG, go_feature_server=True),
```
felixwang9817, line 70: this should be included in the `DEFAULT_FULL_REPO_CONFIGS.extend(` part

the baseline `DEFAULT_FULL_REPO_CONFIGS` should only contain configurations that can be tested without launching infra; a repo config using Redis doesn't belong here

sdk/python/tests/integration/registration/test_universal_types.py
```diff
@@ -230,7 +230,7 @@ def test_feature_get_online_features_types_match(online_types_test_fixtures):
 
     driver_id_value = "1" if config.entity_type == ValueType.STRING else 1
     online_features = fs.get_online_features(
-        features=features, entity_rows=[{"driver": driver_id_value}],
+        features=features, entity_rows=[{"driver_id": driver_id_value}],
```
felixwang9817, line 233: since you're fixing this right now, you might as well also fix this in `test_s3_custom_endpoint.py` (line 48)

sdk/python/feast/feature_store.py
```diff
@@ -1297,8 +1299,8 @@ def get_online_features(
                 if isinstance(features, FeatureService)
                 else None,
                 entities=columnar,
+                request_data={},  # ToDo: add request data parameter to public API
```
felixwang9817, line 1302: nit: for consistency can you do `TODO`

ideally also add your username if you intend on doing this in a followup

sdk/python/feast/feature_store.py
```diff
@@ -1939,6 +1941,9 @@ def serve_transformations(self, port: int) -> None:
 
         transformation_server.start_server(self, port)
 
+    def _refresh_go_server(self):
```
felixwang9817, line 1944: this name is pretty misleading - I think your idea is that setting `self._go_server = None` will force the FS to reinitialize the go server on a new `get_online_features`, if so maybe add a comment to describe exactly what you're doing

felixwang9817 (commented):
just some initial comments; will review again later

sdk/python/feast/feature_store.py
pyalex: Renamed into `_teardown..`. Yeah, the idea is to kill current instance (if it exists) and it will be recreated on the next requests. Added comment about that

go/cmd/server/main.go
```diff
@@ -41,7 +45,7 @@ func main() {
 	}
 
 	log.Println("Initializing feature store...")
-	fs, err := feast.NewFeatureStore(repoConfig)
+	fs, err := feast.NewFeatureStore(repoConfig, dummyTransformCallback)
```
achals, line 48: If we wanted to start the standalone go feature server, how can instantiate the transformation callback in python and pass it in here?

achals (commented):
looks good

go.sum
```diff
@@ -32,6 +32,7 @@ github.com/Azure/go-autorest/logger v0.2.1/go.mod h1:T9E3cAhj2VqvPOtCYAvby9aBXkZ
 github.com/Azure/go-autorest/tracing v0.6.0/go.mod h1:+vhtPC754Xsa23ID7GlGsrdKBpUA79WCAKPPZVC2DeU=
 github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03qcyfWMU=
 github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802/go.mod h1:IVnqGOEym/WlBOVXweHU+Q+/VP0lqqI8lqeDx9IjBqo=
+github.com/JohnCGriffin/overflow v0.0.0-20211019200055-46fa312c352c/go.mod h1:X0CRv0ky0k6m906ixxpzmDRLvX58TFUKS2eePweuyxk=
```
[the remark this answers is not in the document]
pyalex: Seems like it's `github.com/apache/arrow/go/v7` dependency

go.sum
```diff
@@ -43,14 +44,17 @@ github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc/go.mod h1:LOuy
 github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751/go.mod h1:LOuyumcjzFXgccqObfd/Ljyb9UuFJ6TxHnclSeseNhc=
 github.com/alecthomas/units v0.0.0-20151022065526-2efee857e7cf/go.mod h1:ybxpYRFXyAe+OPACYpWeL0wqObRcbAqCMya13uyzqw0=
 github.com/alecthomas/units v0.0.0-20190717042225-c3de453c63f4/go.mod h1:ybxpYRFXyAe+OPACYpWeL0wqObRcbAqCMya13uyzqw0=
+github.com/andybalholm/brotli v1.0.3/go.mod h1:fO7iG3H7G2nSZ7m0zPUDn85XEX2GTukHGRSepvi9Eig=
 github.com/antihax/optional v1.0.0/go.mod h1:uupD/76wgC+ih3iEmQUL+0Ugr19nfwCT1kdvxnR2qWY=
+github.com/apache/arrow/go/arrow v0.0.0-20200730104253-651201b0f516 h1:byKBBF2CKWBjjA4J1ZL2JXttJULvWSl50LegTyRZ728=
```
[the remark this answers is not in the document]
pyalex: This is from this dependencies that was added in #2446
```
github.com/xitongsys/parquet-go v1.6.2
github.com/xitongsys/parquet-go-source v0.0.0-20220315005136-aec0fe3e777c
```

I will fix this in the following refactoring PR

go/embedded/online_features.go
```diff
@@ -25,26 +27,88 @@ type DataTable struct {
...
+			return nil, err
+		}
+		viewNames[viewName] = nil
+	}
+
+	entities, _ := s.fs.ListEntities(true)
+	entitiesByName := make(map[string]*feast.Entity)
+	for _, entity := range entities {
+		entitiesByName[entity.Name] = entity
+	}
```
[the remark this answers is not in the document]
pyalex: Good point. This will be addressed in the following refactoring PR

go/embedded/online_features.go
```diff
@@ -25,26 +27,88 @@ type DataTable struct {
...
+
+	for _, projection := range featureService.Projections {
+		view, err := s.fs.GetFeatureView(projection.Name, true)
+		if err != nil {
+			// skip on demand feature views
+			continue
+		}
+		for entityName, _ := range view.Entities {
+			entity := entitiesByName[entityName]
+			joinKeyTypes[entity.JoinKey] = int32(entity.ValueType.Number())
```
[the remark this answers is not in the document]
pyalex: `GetEntityTypesMapByFeatureService` is a function called from Python. Gopy doesn't understand that ValueType is an alias for int32, so packing and unpacking it will be additional headache, not sure that it will make it more readable.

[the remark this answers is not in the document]
pyalex: I'm just saying that using native types is easier in some cases.

go/cmd/server/main.go
pyalex: Good question. With current implementation Go code must be always wrapped in Python, it's an easy way to have Python interpreter in the same process. Go feature server can run in standalone mode, but w/o support of transformations and python-based store connectors.

go/cmd/server/main.go
```diff
@@ -17,6 +17,10 @@ const (
 	feastServerVersion  = "0.18.0"
 )
 
+func dummyTransformCallback(ODFVName string, inputArrPtr, inputSchemaPtr, outArrPtr, outSchemaPtr uintptr, fullFeatureNames bool) int {
```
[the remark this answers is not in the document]
pyalex: Basically, w/o Python as a wrapper on demand transformations are not supported. I replaced this dummy implementation with passing `nil` to the `FeatureStore` constructor.
I also added a doc about `TransformationCallback` signature (see `transformation.go`)

achals (approved):
/lgtm

PR #2498, opened 2022-04-06
fix: Remove spark from the AWS Lambda dockerfile

achals:

We discovered that this was a red herring when investigating earlier failures. It was an issue in 0.19.1 of Feast, but was fixed 0.19.2+.

Fixes #


felixwang9817 (approved):
/lgtm

PR #2500, opened 2022-04-06
feat: Add Field class

felixwang9817:

**What this PR does / why we need it**: This PR adds the `Field` class, which will replace the `Feature` class. The difference between `Field` and `Feature` is essentially cosmetic; `Field` will be stored under the hood using the existing protobuf structure for `Feature`.

Fixes #

### DIFF
--- a/sdk/python/feast/field.py
+++ b/sdk/python/feast/field.py
@@ -46,9 +46,9 @@ class Field:
         if self.name != other.name or self.dtype != other.dtype:
             return False
         return True
-    
+
     def __hash__(self):
-        return hash((id(self), self.name))
+        return hash((self.name, hash(self.dtype)))
 
     def __lt__(self, other):
         return self.name < other.name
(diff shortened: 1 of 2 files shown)


sdk/python/feast/field.py
```diff
@@ -0,0 +1,70 @@
...
+    Attributes:
+        name: The name of the field.
+        dtype: The type of the field, such as string or float.
+    """
+
+    name: str
+    dtype: FeastType
+
+    def __init__(
+        self, *, name: str, dtype: FeastType,
```
achals, line 32: Awww yisss

achals (approved):
/lgtm

sdk/python/feast/field.py
```diff
@@ -0,0 +1,70 @@
...
+            name: The name of the field.
+            dtype: The type of the field, such as string or float.
+        """
+        self.name = name
+        self.dtype = dtype
+
+    def __eq__(self, other):
+        if self.name != other.name or self.dtype != other.dtype:
+            return False
+        return True
```
achals, line 47: Do you need to implement `__hash__` here as well? Usually that and `__eq__` need to be implemented together.

PR #2503, opened 2022-04-06
fixed bug in passing config file params to snowflake python connector

sfc-gh-madkins:

Fixes an issue with how config params are pulled from a config file for authenticating with snowflake python connector

Fixes # None

### DIFF
--- a/sdk/python/feast/infra/utils/snowflake_utils.py
+++ b/sdk/python/feast/infra/utils/snowflake_utils.py
@@ -53,13 +53,20 @@ def get_snowflake_conn(config, autocommit=True) -> SnowflakeConnection:
     else:
         kwargs = {}
 
+    if "schema" in kwargs.keys():
+        kwargs["schema_"] = kwargs.pop("schema")
+
     kwargs.update((k, v) for k, v in config_dict.items() if v is not None)
     [
         kwargs.update({k: '"' + v + '"'})
         for k, v in kwargs.items()
         if k in ["role", "warehouse", "database", "schema_"]
     ]
-    kwargs["schema"] = kwargs.pop("schema_")
+
+    if "schema" in kwargs.keys():
+        kwargs["schema"] = kwargs.pop("schema_")
+    else:
+        kwargs["schema"] = '"PUBLIC"'
 
     try:
         conn = snowflake.connector.connect(
--- a/sdk/python/feast/infra/offline_stores/snowflake.py
+++ b/sdk/python/feast/infra/offline_stores/snowflake.py
@@ -82,7 +82,7 @@ class SnowflakeOfflineStoreConfig(FeastConfigBaseModel):
     database: Optional[str] = None
     """ Snowflake database name """
 
-    schema_: Optional[str] = Field("PUBLIC", alias="schema")
+    schema_: Optional[str] = Field(None, alias="schema")
     """ Snowflake schema name """
 
     class Config:


sdk/python/feast/infra/utils/snowflake_utils.py
```diff
@@ -53,13 +53,20 @@ def get_snowflake_conn(config, autocommit=True) -> SnowflakeConnection:
     else:
         kwargs = {}
 
+    if "schema" in kwargs.keys():
```
achals, line 56: nit, this should be `if "schema in kwargs` right?

sdk/python/feast/infra/utils/snowflake_utils.py
```diff
@@ -53,13 +53,20 @@ def get_snowflake_conn(config, autocommit=True) -> SnowflakeConnection:
...
         kwargs.update({k: '"' + v + '"'})
         for k, v in kwargs.items()
         if k in ["role", "warehouse", "database", "schema_"]
     ]
-    kwargs["schema"] = kwargs.pop("schema_")
+
+    if "schema" in kwargs.keys():
+        kwargs["schema"] = kwargs.pop("schema_")
+    else:
+        kwargs["schema"] = '"PUBLIC"'
```
achals, line 69: Any reason why this is happening here as well as on line 56?

sdk/python/feast/infra/utils/snowflake_utils.py
```diff
@@ -53,13 +53,20 @@ def get_snowflake_conn(config, autocommit=True) -> SnowflakeConnection:
     else:
         kwargs = {}
 
+    if "schema" in kwargs.keys():
+        kwargs["schema_"] = kwargs.pop("schema")
+
     kwargs.update((k, v) for k, v in config_dict.items() if v is not None)
     [
         kwargs.update({k: '"' + v + '"'})
```
achals, line 61: just noticed, you should be able to this here:
```
kwargs[k] = f'"{v}"' 
```

sdk/python/feast/infra/utils/snowflake_utils.py
sfc-gh-madkins: schema is a protected word so I need to do whacky stuff without forcing users to set schema_ as opposed to a more friendly schema

sdk/python/feast/infra/utils/snowflake_utils.py
sfc-gh-madkins: not quite following here

sfc-gh-madkins: there is specific keywords that need quotes around them, others dont require

achals: sorry, I just meant that instead of doing 
```
kwargs.update({k: '"' + v + '"'})
```

you can do the thing I suggested - you can still do it selectively for the keywords that need quotes, just that assignment style seems more pythonic than the `.update` style

sdk/python/feast/infra/utils/snowflake_utils.py
achals: Oops - i understand this snippet now, I misread it earlier. Makes sense.

sdk/python/feast/infra/utils/snowflake_utils.py
```diff
@@ -53,13 +53,20 @@ def get_snowflake_conn(config, autocommit=True) -> SnowflakeConnection:
...
         for k, v in kwargs.items()
         if k in ["role", "warehouse", "database", "schema_"]
     ]
-    kwargs["schema"] = kwargs.pop("schema_")
+
+    if "schema" in kwargs:
```
achals, line 66: should be `"schema_"` here right? Since that's what you're popping the next line?

achals (approved):
/lgtm

PR #2508, opened 2022-04-07
fix: Added private_key auth for Snowflake

### LINKED ISSUE #2482: Allow for connecting to Snowflake with a private key
When connecting to Snowflake with an account that requires MFA its does not work (or at least requires extra work to use MFA devices) to connect with just the username and password credentials. 

Instead of using username and password we should be able to use a private key and a private key passphrase to connect to Snowflake. Snowflake already supports this method of authentication. See [here](https://docs.snowflake.com/en/user-guide/python-connector-example.html#label-python-key-pair-authn-rotation). Feast should add this as an option to the feature_store.yaml as part of the Snowflake connection config.

When trying to use a private_key_path and passphrase this error was raised:
```
raise FeastConfigError(e, config_path)
feast.repo_config.FeastConfigError: 2 validation errors for RepoConfig
__root__ -> offline_store -> private_key_passphrase
  extra fields not permitted (type=value_error.extra)
__root__ -> offline_store -> private_key_path
  extra fields not permitted (type=value_error.extra)
```
It seems like fields related to a passphrase are not permitted [here](https://github.com/feast-dev/feast/blob/b95f4410ee91069ff84e81d2d5f3e9329edc8626/sdk/python/feast/infra/offline_stores/snowflake.py#L56)

cc @sfc-gh-madkins

sfc-gh-madkins:

Adds ability for key-pair auth for Snowflake

Fixes #2482

### DIFF
--- a/sdk/python/feast/infra/utils/snowflake_utils.py
+++ b/sdk/python/feast/infra/utils/snowflake_utils.py
@@ -4,7 +4,7 @@ import random
 import string
 from logging import getLogger
 from tempfile import TemporaryDirectory
-from typing import Dict, Iterator, List, Optional, Tuple, cast
+from typing import Dict, Iterator, List, Optional, Tuple, cast, Any
 
 import pandas as pd
 from cryptography.hazmat.backends import default_backend
@@ -50,10 +50,9 @@ def get_snowflake_conn(config, autocommit=True) -> SnowflakeConnection:
     # read config file
     config_reader = configparser.ConfigParser()
     config_reader.read([config_dict["config_path"]])
+    kwargs: Dict[str, Any] = {}
     if config_reader.has_section(config_header):
         kwargs = dict(config_reader[config_header])
-    else:
-        kwargs = {}
 
     if "schema" in kwargs:
         kwargs["schema_"] = kwargs.pop("schema")
--- a/sdk/python/feast/infra/utils/snowflake_utils.py
+++ b/sdk/python/feast/infra/utils/snowflake_utils.py
@@ -70,6 +70,7 @@ def get_snowflake_conn(config, autocommit=True) -> SnowflakeConnection:
         kwargs["schema"] = '"PUBLIC"'
 
     #https://docs.snowflake.com/en/user-guide/python-connector-example.html#using-key-pair-authentication-key-pair-rotation
+    #https://docs.snowflake.com/en/user-guide/key-pair-auth.html#configuring-key-pair-authentication
     if "private_key" in kwargs:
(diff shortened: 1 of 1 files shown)


sdk/python/feast/infra/utils/snowflake_utils.py
```diff
@@ -288,3 +295,21 @@ def chunk_helper(lst: pd.DataFrame, n: int) -> Iterator[Tuple[int, pd.DataFrame]
...
+            key.read(),
+            password=private_key_passphrase.encode(),
+            backend=default_backend(),
+        )
+
+    pkb = p_key.private_bytes(
+        encoding=serialization.Encoding.DER,
+        format=serialization.PrivateFormat.PKCS8,
+        encryption_algorithm=serialization.NoEncryption(),
+    )
```
achals, line 313: Is this the only supported format for the key pair?

sfc-gh-madkins: Yes -- I have add the links to the docs  as comments.

achals:
@sfc-gh-madkins thanks, looks good but linter seems to be failing because of some type mismatches.

sfc-gh-madkins:
@adchia do you have any ideas on how to fix this lint error:

feast/infra/utils/snowflake_utils.py:75: error: Incompatible types in assignment (expression has type "bytes", target has type "str")

here is the code:

    if "private_key" in kwargs:
        kwargs["private_key"] = parse_private_key_path(
            kwargs["private_key"], kwargs["private_key_passphrase"]
        )
        
parse_private_key_path() returns "bytes"
kwargs["private_key"] is a str that I am looking to overwrite with bytes

PS. Changing the name to create a new entry in the dict doesnt solve the problem either

achals:
> @adchia do you have any ideas on how to fix this lint error:
> 
> feast/infra/utils/snowflake_utils.py:75: error: Incompatible types in assignment (expression has type "bytes", target has type "str")
> 
> here is the code:
> 
> ```
> if "private_key" in kwargs:
>     kwargs["private_key"] = parse_private_key_path(
>         kwargs["private_key"], kwargs["private_key_passphrase"]
>     )
> ```
> 
> parse_private_key_path() returns "bytes" kwargs["private_key"] is a str that I am looking to overwrite with bytes
> 
> PS. Changing the name to create a new entry in the dict doesnt solve the problem either

I pushed a couple of fixes here to solve the typing problem! LMK if you have concerns, otherwise we'll get this merged in on Monday.

sfc-gh-madkins:
Lgtm

PR #2510, opened 2022-04-07
chore: Update Trino data source creator to use testcontainers

achals:

This should help with testing #2517 

Fixes #

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -14,7 +14,7 @@ jobs:
       PYTHON: ${{ matrix.python-version }}
     steps:
       - uses: actions/checkout@v2
-      - uses: docker-practice/actions-setup-docker@master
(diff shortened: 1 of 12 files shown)


achals:
cc @olivierlabreche @chhabrakadabra

.github/workflows/unit_tests.yml
```diff
@@ -14,6 +14,7 @@ jobs:
       PYTHON: ${{ matrix.python-version }}
     steps:
       - uses: actions/checkout@v2
+      - uses: docker-practice/actions-setup-docker@master
```
felixwang9817, line 17: this makes me slightly nervous since one time the `gcloud` action broke since they switched from `master` to `main` - should we maybe pin to a specific version?

felixwang9817 (commented):
one small concern

felixwang9817 (approved):
/lgtm

PR #2511, opened 2022-04-08
chore: Switch new type system to use ValueType instead of ValueTypeProto.Enum

felixwang9817:

**What this PR does / why we need it**: #2489 added a conversion method from the new `FeastType`s to to `ValueTypeProto.Enum`. This conversion method is slightly awkward since it relies on a proto class, `ValueTypeProto.Enum`, instead of `ValueType`, the corresponding Python enum. This PR switches the conversion method to rely on `ValueType` instead of `ValueTypeProto.Enum`.

Fixes #


achals (approved):
/lgtm

PR #2513, opened 2022-04-09
chore: Split single package into sub package in Go feature server

pyalex:

This is purely refactoring PR. No functionality was added. But some tests were.

1. Code is cleaner, when not all functions are methods of `FeatureStore` and stored in single file
2. Easier to write unit tests
3. Easier to debug (since when `cgo` is involved, like in `transformation` package, which dependes on `arrow/cdata`, some more advanced debuggers can't be used)

Fixes #


go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/feast-dev/feast
 go 1.17
 
 require (
-	github.com/apache/arrow/go/v7 v7.0.0
+	github.com/apache/arrow/go/v8 v8.0.0-20220408212425-58fe60f59289
```
pyalex, line 6: In v7 there's bug with reading parquet. File stored in Python can't be read

go/cmd/server/server_test.go
```diff
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"github.com/feast-dev/feast/go/internal/feast/registry"
```
achals, line 5: nit, goimports should reorder this

achals (approved):
/lgtm

PR #2521, opened 2022-04-11
fix: Fix entity row to use join key instead of name

felixwang9817:

**What this PR does / why we need it**: Recently the `sdk/python/tests/benchmarks/test_benchmark_universal_online_retrieval.py::test_online_retrieval` test has been failing on master with the following error: `RuntimeError: key driver_id is missing in provided entity rows`. This PR fixes the bug.

Fixes #


achals (approved):
/lgtm

PR #2523, opened 2022-04-11
fix: Support passing batch source to streaming sources for backfills

achals:

Similar to the PushSource, we want to is to fold in batch source as a field in streaming sources, since they are used for semantically for "backfill". 

Fixes #

### DIFF
--- a/sdk/python/feast/data_source.py
+++ b/sdk/python/feast/data_source.py
@@ -360,6 +360,7 @@ class KafkaSource(DataSource):
         tags: Optional[Dict[str, str]] = None,
         owner: Optional[str] = "",
         timestamp_field: Optional[str] = "",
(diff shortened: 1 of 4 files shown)


sdk/python/feast/data_source.py
```diff
@@ -634,6 +644,7 @@ class PushSource(DataSource):
 
     def __init__(
         self,
+        *,
         name: str,
         schema: Dict[str, ValueType],
         batch_source: DataSource,
```
kevjumba, line 650: SInce we're adding the * for keyword args, should we change these to keyword args too then aka Optional?

achals: They become keyword args, but are not optional (also this is a safe change since Push sources haven't been released yet)

PR #2527, opened 2022-04-12
chore: Fix ODFV decorator

felixwang9817:
**What this PR does / why we need it**: #2514 switched from using `Feature` to `Field` for declaring types of features in `FeatureView`s and `OnDemandFeatureView`s. It left a `TODO` to make the same change for the ODFV decorator; this PR completes that TODO. It also ensures that a previous change, #2442, does not break users using the ODFV decorator. It also enforces the usage of kwargs for the ODFV decorator.

Fixes #


achals (approved):
/lgtm

PR #2534, opened 2022-04-13
chore: Fix typing

felixwang9817:
**What this PR does / why we need it**: This PR removes `Array` and `String` and other Feast types as top-level objects, so that users must do `from feast.types import Array, String` instead of `feast import Array, String`.

Fixes #


achals (approved):
/lgtm

PR #2540, opened 2022-04-13
docs: Add docs for Go feature server

felixwang9817:

**What this PR does / why we need it**: This PR adds docs for the Go feature server.

Fixes #


docs/SUMMARY.md
```diff
@@ -82,6 +82,7 @@
   * [.feastignore](reference/feature-repository/feast-ignore.md)
 * [Feature servers](reference/feature-servers/README.md)
   * [Local feature server](reference/feature-servers/local-feature-server.md)
+  * [Go feature server](reference/feature-servers/go-feature-server.md)
```
achals, line 85: Everything else looks good, but the only nit I have is that Go feature server is a bit misleading since it's not just Go, and there's no standalone go feature server that can be deployed.

I'd probably called this Go-based feature retrieval and update the docs and flag. What do you think?

docs/reference/feature-servers/go-feature-server.md
```diff
@@ -0,0 +1,32 @@
...
+
+Currently, the Go feature server only supports online serving and does not have an offline component including APIs to create feast feature repositories or apply configuration to the registry to facilitate online materialization. It also does not expose its own dedicated cli to perform feast actions. Currently, the Go feature server is only meant to expose an online serving API that can be called through the python SDK to facilitate faster online feature retrieval.
+
+The Go feature server currently only supports Redis and Sqlite as online stores; support for other online stores will be added soon. Initial benchmarks indicate that the Go feature server is significantly faster than the Python feature server. We plan to release a more comprehensive set of benchmarks. For more details, see the [RFC](https://docs.google.com/document/d/1Lgqv6eWYFJgQ7LA_jNeTh8NzOPhqI9kGTeyESRpNHnE).
+
+## Usage
+
+To enable the Go feature server, set `go_feature_server: True` in your `feature_store.yaml`.
```
adchia, line 13: We also should have installation instructions?

e.g. COMPILE_GO=True and also having the Go toolchain installed.

adchia, line 13: We also should have installation instructions?

e.g. COMPILE_GO=True and also having the Go toolchain installed.

achals: good point. We create a wheel with the library embedded on supported platforms (mac and linux, python 3.7-3.10). So no action should be needed but cc @pyalex to correct me if I'm wrong.

pyalex: @achals this is correct. Nothing changes from user perspective. Pip will automatically pull platform specific wheel

docs/SUMMARY.md
kevjumba: I left it in the `feature servers` folder since I think out of all folders it's still most relevant to Feature Servers.

docs/reference/feature-servers/go-feature-retrieval.md
```diff
@@ -0,0 +1,36 @@
...
+
+Currently, this component only supports online serving and does not have an offline component including APIs to create feast feature repositories or apply configuration to the registry to facilitate online materialization. It also does not expose its own dedicated cli to perform feast actions. Furthermore, this component is only meant to expose an online serving API that can be called through the python SDK to facilitate faster online feature retrieval.
+
+The Go Feature Retrieval component currently only supports Redis and Sqlite as online stores; support for other online stores will be added soon. Initial benchmarks indicate that it is significantly faster than the Python feature server for online feature retrieval. We plan to release a more comprehensive set of benchmarks. For more details, see the [RFC](https://docs.google.com/document/d/1Lgqv6eWYFJgQ7LA_jNeTh8NzOPhqI9kGTeyESRpNHnE).
+
+## Installation
+
+To install, run `export COMPILE_GO=true` before you run `pip install feast`.
```
achals, line 13: This is not true - if you're using macOS or linux, x86, python 3.7-3.10, you'll get the version of feast installed with the go library anyway.

achals (changes requested):

docs/reference/feature-servers/go-feature-retrieval.md
kevjumba: Hm, in setup.py the build commands for compiling go and also building go are only enabled if we are using COMPILE_GO=true. I don't think we run the compilation without the flag.

docs/reference/feature-servers/go-feature-server.md
kevjumba: Oh I see, what about the COMPILE_GO flag, we need to export that in order to compile the feature server correct? @achals @pyalex

docs/reference/feature-servers/go-feature-retrieval.md
kevjumba: Confirmed it's precompiled and fixed.

docs/reference/feature-servers/go-feature-retrieval.md
```diff
@@ -10,7 +10,9 @@ The Go Feature Retrieval component currently only supports Redis and Sqlite as o
 
 ## Installation
 
-To install, run `export COMPILE_GO=true` before you run `pip install feast`.
+As long as you are running macOS or linux x86 with python version 3.7-3.10, the go component comes pre-compiled when you run install feast.
```
achals, line 13:
```suggestion
As long as you are running macOS or linux, on x86, with python version 3.7-3.10, the go component comes pre-compiled when you install feast.
```

kevjumba: Fixed.

achals (approved):
/lgtm

achals (approved):
/lgtm

PR #2509, opened 2022-04-07
fix: Schema update

kevjumba:
**What this PR does / why we need it**:
Update schema in RequestDataSource.
**Which issue(s) this PR fixes**:

Fixes #

PR #2541, opened 2022-04-13
fix: Fix definitions.py with new definition

kevjumba:

Fixes PR #2509, since the java docker container pulls feast from latest master. We can only make the RequestSource change after merging into master.

Fixes #


achals (approved):
/lgtm

PR #2542, opened 2022-04-14
docs: Update DynamoDB BatchGetItem permissions

TremaMiguel:
Signed-off-by: Miguel Trejo <armando.trejo.marrufo@gmail.com>

`DynamoDBOnlineStore` retrieve items calling the `BatchGetItem` operation, documentation is updated for this.


achals (commented):
Do we only use batchget now, and never get? Or are there instance where we still use get?

adchia (approved):
/lgtm

Issue #2537, opened 2022-04-13
ValueError: Please install the 'db-dtypes' package to use this function.


chhabrakadabra:
## Expected Behavior 

If I install feast with the `[gcp]` extra, I should be able to use `BigQueryRetrievalJob.to_bigquery()` without problems.

## Current Behavior

When feast is installed with the `[gcp]` extra, it installs `google-cloud-bigquery` as per [this line](https://github.com/feast-dev/feast/blob/master/sdk/python/setup.py#L74). It does not use the [pandas extra](https://github.com/googleapis/python-bigquery/blob/main/setup.py#L54), which means nothing triggers the installation of the `db-dtypes` package. 

When `bigquery.py` uses `to_dataframe()` on the result of a BQ query as [done here](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/offline_stores/bigquery.py#L471), it triggers a [verification of imports available](https://github.com/googleapis/python-bigquery/blob/main/google/cloud/bigquery/table.py#L1925), which causes the exception [here](https://github.com/googleapis/python-bigquery/blob/7897b5558dc45f6438415f7fe6f3e68daad4a31d/google/cloud/bigquery/_pandas_helpers.py#L987).

## Steps to reproduce

Install `feast[gcp]` and use `BigQueryRetrievalJob.to_bigquery()`.

### Specifications

- Version: `feast[gcp]==0.16.1`
- Platform: OSX
- Subsystem: Not sure what this is referring to

## Possible Solution

Use `google-cloud-bigquery[pandas]` for the `[gcp]` extra.

felixwang9817:
Hey @chhabrakadabra, thanks for reporting this issue. What version of `google-cloud-bigquery` do you have? I suspect the issue is that `google-cloud-bigquery` recently released v3.0.0 which had the `db-dtypes` breaking change. See this [release](https://github.com/googleapis/python-bigquery/releases/tag/v3.0.0) and this [migration guide](https://github.com/googleapis/python-bigquery/pull/1027/files?short_path=f411a50#diff-f411a5009131db5a6c7242b49ca78488f2d274ef9e5e5f4de9262a42108a5386) for more details. Can you try reverting to `google-cloud-bigquery<3.0.0` and seeing if this error still persists? I'm on `2.34.1` and it works fine for me.

I'll keep this issue open to track an upgrade to v3.0.0.

felixwang9817:
[...]

Issue #2532, opened 2022-04-13
Feast BigQuery offline store gets wrong argument types for `event_timestamp`


KarolisKont:
## Expected Behavior 
Retrieving training data from an offline store that is BigQuery by using `get_historical_features` method from Python SDK.

## Current Behavior
Getting this error when running `get_historical_features`:

```sh
Traceback (most recent call last):
  File "get_offline_data.py", line 22, in <module>
    "item_user_stats_v1:is_high_risk_seller",
  File "/usr/local/lib/python3.7/site-packages/feast/infra/offline_stores/offline_store.py", line 77, in to_df
    features_df = self._to_df_internal()
  File "/usr/local/lib/python3.7/site-packages/feast/infra/offline_stores/bigquery.py", line 290, in _to_df_internal
    df = self._execute_query(query).to_dataframe(create_bqstorage_client=True)
  File "/usr/local/lib/python3.7/site-packages/feast/usage.py", line 280, in wrapper
    raise exc.with_traceback(traceback)
  File "/usr/local/lib/python3.7/site-packages/feast/usage.py", line 269, in wrapper
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/feast/infra/offline_stores/bigquery.py", line 357, in _execute_query
    block_until_done(client=self.client, bq_job=bq_job, timeout=timeout)
  File "/usr/local/lib/python3.7/site-packages/feast/infra/offline_stores/bigquery.py", line 415, in block_until_done
    raise bq_job.exception()
google.api_core.exceptions.BadRequest: 400 No matching signature for operator <= for argument types: TIMESTAMP, DATETIME. Supported signature: ANY <= ANY at [78:13]
``` 
The error happens on a query where it compares event timestamps (last line):
 ```SQL
 item_user_stats_v1__base AS (
    SELECT
        subquery.*,
        entity_dataframe.entity_timestamp,
        entity_dataframe.item_user_stats_v1__entity_row_unique_id
    FROM item_user_stats_v1__subquery AS subquery
    INNER JOIN item_user_stats_v1__entity_dataframe AS entity_dataframe
    ON TRUE
        AND subquery.event_timestamp <= entity_dataframe.entity_timestamp
```
 
## Steps to reproduce
[...]

PR #2554, opened 2022-04-15
fix: Update RedisCluster to use redis-py official implementation

### LINKED ISSUE #2552: Switch from `redis-py-cluster` to `redis-py`
As [documented](https://github.com/Grokzen/redis-py-cluster#redis-py-cluster-eol), `redis-py-cluster` has reached EOL. It is now being merged into [`redis-py`](https://github.com/redis/redis-py#cluster-mode). We previously tried to switch from `redis-py-cluster` to `redis-py` but ran into various issues; see #2328. The upstream [bug](https://github.com/redis/redis-py/issues/2003) has since been fixed, so we can now switch from `redis-py-cluster` to `redis-py`.

kevjumba:
Update redis cluster implementation and remove redis cluster dependency. Also update crypto for mac m1 to 3.4.8(this is the newest version that patches some fixes in order to be compatible with openssl 1.1.1(the older version has been giving people with mac m1 trouble in installation).

Fixes #2552


sdk/python/requirements/py3.10-ci-requirements.txt
```diff
@@ -208,22 +204,24 @@ google-auth==1.35.0
     #   google-cloud-storage
 google-auth-httplib2==0.1.0
     # via google-api-python-client
-google-auth-oauthlib==0.5.0
+google-auth-oauthlib==0.5.1
     # via gcsfs
-google-cloud-bigquery==2.34.1
-    # via feast (setup.py)
-google-cloud-bigquery-storage==2.12.0
+google-cloud-bigquery==3.0.1
```
achals, line 209: We should pin this below 3.0.0. I think the 3+ is causing some issues, cc #2537 or #2532

achals (approved):
Single change

sdk/python/requirements/py3.10-ci-requirements.txt
kevjumba: Sure, let me do that right now.

achals (approved):
/lgtm

PR #2555, opened 2022-04-15
fix: Build platform specific python packages with ci-build-wheel

pyalex:

Currently we cannot upload platform specific wheels generated by `wheel`. It produces not-portable wheels (see [PEP-600](https://peps.python.org/pep-0600/)), which are not accepted anymore by PyPi.

This PR reworks our packaging process.

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -145,15 +145,15 @@ install-go-ci-dependencies:
 	go get github.com/go-python/gopy
 	go install golang.org/x/tools/cmd/goimports
 	go install github.com/go-python/gopy
+	python -m pip install pybindgen==0.22.0
 
 install-protoc-dependencies:
-	pip install grpcio-tools==1.44.0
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -145,15 +145,15 @@ install-go-ci-dependencies:
 	go get github.com/go-python/gopy
 	go install golang.org/x/tools/cmd/goimports
 	go install github.com/go-python/gopy
+	python -m pip install pybindgen==0.22.0
 
 install-protoc-dependencies:
-	pip install grpcio-tools==1.34.0
+	pip install grpcio-tools==1.44.0 mypy-protobuf
```
achals, line 151: need to pin mypy-protobuf?

achals (approved):
only one question

Makefile
pyalex: pinned

achals (approved):
/lgtm

PR #2556, opened 2022-04-15
fix: Fix `__hash__` methods

### LINKED ISSUE #2506: __hash__ methods broken
## Expected Behavior

The `__hash__` methods should not be implemented like this:
```
def __hash__(self) -> int:
    return hash((id(self), self.name))
```

Objects with the `__hash__` method implemented in such a way are not being deduplicated correctly in e.g. sets and dicts.

### Specifications

- Version:
- Platform:
- Subsystem:

felixwang9817:
**What this PR does / why we need it**: This PR fixes the broken `__hash__` methods in all Feast objects.

Fixes #2506

### DIFF
--- a/sdk/python/feast/saved_dataset.py
+++ b/sdk/python/feast/saved_dataset.py
@@ -92,17 +92,23 @@ class SavedDataset:
         return str(MessageToJson(self.to_proto()))
 
     def __hash__(self):
-        return hash((id(self), self.name))
+        return hash((self.name))
 
     def __eq__(self, other):
         if not isinstance(other, SavedDataset):
             raise TypeError(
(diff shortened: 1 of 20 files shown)


achals (approved):
/lgtm

achals (approved):
/lgtm

sdk/python/feast/saved_dataset.py
```diff
@@ -92,17 +92,23 @@ def __str__(self):
         return str(MessageToJson(self.to_proto()))
 
     def __hash__(self):
-        return hash((id(self), self.name))
+        return hash((self.name))
```
judahrand, line 95: Unnecessary brackets?

PR #2559, opened 2022-04-16
feat: Create stream and batch feature view abstractions

achals:

in prep for upcoming streaming work

Fixes #

### DIFF
--- a/sdk/python/feast/batch_feature_view.py
+++ b/sdk/python/feast/batch_feature_view.py
@@ -1,11 +1,10 @@
 from datetime import timedelta
 from typing import Dict, List, Optional, Union
 
-from google.protobuf.duration_pb2 import Duration
-
-from feast import Entity, Feature, Field
 from feast.data_source import DataSource
+from feast.entity import Entity
 from feast.feature_view import FeatureView
+from feast.field import Field
 from feast.protos.feast.core.DataSource_pb2 import DataSource as DataSourceProto
 
 SUPPORTED_BATCH_SOURCES = {
@@ -24,8 +23,7 @@ class BatchFeatureView(FeatureView):
         *,
         name: Optional[str] = None,
         entities: Optional[Union[List[Entity], List[str]]] = None,
-        ttl: Optional[Union[Duration, timedelta]] = None,
-        features: Optional[List[Feature]] = None,
+        ttl: Optional[timedelta] = None,
         tags: Optional[Dict[str, str]] = None,
         online: bool = True,
         description: str = "",
@@ -51,7 +49,6 @@ class BatchFeatureView(FeatureView):
             ttl=ttl,
             batch_source=None,
             stream_source=None,
-            features=features,
             tags=tags,
             online=online,
             description=description,
--- a/sdk/python/feast/batch_feature_view.py
+++ b/sdk/python/feast/batch_feature_view.py
@@ -1,0 +1,61 @@
+from datetime import timedelta
+from typing import Dict, List, Optional, Union
+
+from google.protobuf.duration_pb2 import Duration
+
+from feast import Entity, Feature, Field
+from feast.data_source import DataSource
+from feast.feature_view import FeatureView
+from feast.protos.feast.core.DataSource_pb2 import DataSource as DataSourceProto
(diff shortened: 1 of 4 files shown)


sdk/python/feast/batch_feature_view.py
```diff
@@ -0,0 +1,61 @@
+from datetime import timedelta
+from typing import Dict, List, Optional, Union
+
+from google.protobuf.duration_pb2 import Duration
+
+from feast import Entity, Feature, Field
```
felixwang9817, line 6: nit: elsewhere we tend to do `from feast.entity import Entity` etc.

we should try to stay consistent

sdk/python/feast/batch_feature_view.py
```diff
@@ -0,0 +1,61 @@
...
+    "SnowflakeSource",
+    "SparkSource",
+    "TrinoSource",
+}
+
+
+class BatchFeatureView(FeatureView):
+    def __init__(
+        self,
+        *,
```
felixwang9817, line 24: nice

sdk/python/feast/batch_feature_view.py
```diff
@@ -0,0 +1,61 @@
...
+        name: Optional[str] = None,
+        entities: Optional[Union[List[Entity], List[str]]] = None,
+        ttl: Optional[Union[Duration, timedelta]] = None,
+        features: Optional[List[Feature]] = None,
```
felixwang9817, line 28: is BFV meant to be backwards compatible with FV? if not, we should just eliminate this now and switch to `schema`

and make all the changes to other parameters, e.g. `ttl` should not support a `Duration` object

sdk/python/feast/stream_feature_view.py
```diff
@@ -0,0 +1,57 @@
+from datetime import timedelta
+from typing import Dict, List, Optional, Union
+
+from google.protobuf.duration_pb2 import Duration
+
+from feast import Entity, Feature, Field
```
felixwang9817, line 6: all the comments from BFV apply here

sdk/python/tests/unit/test_feature_views.py
```diff
@@ -0,0 +1,70 @@
+from datetime import timedelta
+
+import pytest
+
+from feast import KafkaSource
```
felixwang9817, line 5: same comment

sdk/python/tests/unit/test_feature_views.py
```diff
@@ -0,0 +1,70 @@
...
+from feast.batch_feature_view import BatchFeatureView
+from feast.data_format import AvroFormat
+from feast.infra.offline_stores.file_source import FileSource
+from feast.stream_feature_view import StreamFeatureView
+
+
+def test_create_batch_feature_view():
+    batch_source = FileSource("some path")
```
felixwang9817, line 13: nit: use kwarg?

felixwang9817 (commented):
some nits

sdk/python/feast/batch_feature_view.py
```diff
@@ -0,0 +1,61 @@
+from datetime import timedelta
+from typing import Dict, List, Optional, Union
+
+from google.protobuf.duration_pb2 import Duration
```
woop, line 4: Do we still need to support Duration? Isn't it a good time to deprecate it?

sdk/python/feast/batch_feature_view.py
achals: good catch

sdk/python/feast/batch_feature_view.py
achals: good catch. The new classes don't need to support the old API.

sdk/python/tests/unit/test_feature_views.py
achals: lol yea

sdk/python/feast/batch_feature_view.py
achals: yes it is. Removed

felixwang9817 (approved):
/lgtm

PR #2561, opened 2022-04-18
fix: Fix push sources and add docs / tests pushing via the python feature server

adchia:
This also adds a `allow_registry_cache` parameter to the feature server /push endpoint and updates local-feature-server->python-feature-server in documentation

It also removes schema / timestamp_field & supports schema inference from the underlying batch source.

Fixes #


achals (approved):

felixwang9817 (approved):
/lgtm

PR #2568, opened 2022-04-18
feat: Support entity fields in feature view `schema` parameter by dropping them

felixwang9817:

**What this PR does / why we need it**: This PR adds support for entity fields in the feature view `schema` parameter. As of now, these fields are dropped during inference. A followup PR will deprecate `value_type` in entities, at which point entity fields will actually matter for inference.

Fixes #


achals (approved):
/lgtm

PR #2572, opened 2022-04-19
feat: Add Dockerfile for Feast UI

jenoOvchi:
To run Feast UI locally in Docker.


adchia:
Mind signing your commits?

jenoOvchi:
> Mind signing your commits?

Yes. Do I need to recreate pull request for this?

adchia:
nope you can resign in this PR and force push

benjamintanweihao:
@jenoOvchi Could the base image be tagged too with the version number?

jenoOvchi:
> nope you can resign in this PR and force push

Done

jenoOvchi:
> @jenoOvchi Could the base image be tagged too with the version number?

Done

adchia:
What's the intention behind this btw? Since for local development, yarn start is still definitely the fastest path.

If you want to be deploying this as a service, you'll have to also be regularly syncing with the registry (as is).

Additionally, presumably if you want to use it you also would want to run e.g. minify, etc. e.g. something like

```Dockerfile
FROM node:17.9.0-slim

WORKDIR /app
# add `/app/node_modules/.bin` to $PATH
ENV PATH /app/node_modules/.bin:$PATH

# Install UI dependencies
COPY ui/package.json .
RUN npm install

# Copy over app
COPY ui/tsconfig.json .
COPY ui/public ./public
COPY ui/src ./src

# Build for production.
RUN npm run build --production

# Serve the UI from a port
RUN npm install -g serve
EXPOSE 3000
CMD ["serve", "-s", "build"]

```

benjamintanweihao:
@adchia I think this would be useful when deploying the Feast UI as a service. I assume that you can configure Feast UI to point to a remote repository.

jenoOvchi:
> @adchia I think this would be useful when deploying the Feast UI as a service. I assume that you can configure Feast UI to point to a remote repository.

Yes, I think so too. Now we ingegrates Feast in our platform, as Feature Registry use an S3 Bucket and as workaround make a sync procedure from S3 to Feast UI files directory. But in the future it will be greate to have an option to configure on the UI side connection to the remote repository and use it as a service.

adchia:
hey sorry just checking in here. looks like commits aren't signed anymore? Otherwise, happy to merge

achals:
Hi @jenoOvchi, would you mind signing your commits?

jenoOvchi:
> Hi @jenoOvchi, would you mind signing your commits?

Done.

jenoOvchi:
> What's the intention behind this btw? Since for local development, yarn start is still definitely the fastest path.
> 
> If you want to be deploying this as a service, you'll have to also be regularly syncing with the registry (as is).
> 
> Additionally, presumably if you want to use it you also would want to run e.g. minify, etc. e.g. something like
> 
> ```dockerfile
> FROM node:17.9.0-slim
> 
> WORKDIR /app
> # add `/app/node_modules/.bin` to $PATH
> ENV PATH /app/node_modules/.bin:$PATH
> 
> # Install UI dependencies
> COPY ui/package.json .
> RUN npm install
> 
> # Copy over app
> COPY ui/tsconfig.json .
> COPY ui/public ./public
> COPY ui/src ./src
> 
> # Build for production.
> RUN npm run build --production
> 
> # Serve the UI from a port
> RUN npm install -g serve
> EXPOSE 3000
> CMD ["serve", "-s", "build"]
> ```

If it is actual i can change Dockerfile in this PR to something like that.

adchia:
Looks like you need to sign commits too.

If you want to deploy as a service, might be better to have this Dockerfile instead call into the `feast ui` command instead then that I just added?

https://docs.feast.dev/v/master/reference/alpha-web-ui#feast-cli

Having this, in combination with a `feature_store.yaml` that points at a remote registry which is being updated should have a UI that continuously updates against that remote registry

PR #2573, opened 2022-04-19
fix: Fix DynamoDB fetches when there are entities that are not found

adchia:
Before this, having an unknown entity (e.g. such as is commonly the case in the feast-benchmarks repo) in a batch would result in an exception because 

https://github.com/feast-dev/feast/blob/master/sdk/python/feast/feature_store.py#L1789-L1789 

expects that the number of response rows is equal to the number of requested entities. Dynamo will only return values for entities that are found, so there could be a mismatch and that causes protobuf to complain because we mix None values and Value values.

This also reverts the change to allow users to not sort feature values. Feast expects the results to be in the original order to return the correct values in a GetOnlineFeatures response.

Fixes #

### DIFF
--- a/sdk/python/feast/infra/online_stores/online_store.py
+++ b/sdk/python/feast/infra/online_stores/online_store.py
@@ -76,9 +76,9 @@ class OnlineStore(ABC):
             entity_keys: a list of entity keys that should be read from the FeatureStore.
             requested_features: (Optional) A subset of the features that should be read from the FeatureStore.
         Returns:
-            Data is returned as a list, one item per entity key. Each item in the list is a tuple
-            of event_ts for the row, and the feature data as a dict from feature names to values.
-            Values are returned as Value proto message.
+            Data is returned as a list, one item per entity key in the original order as the entity_keys argument.
+            Each item in the list is a tuple of event_ts for the row, and the feature data as a dict from feature names
+            to values. Values are returned as Value proto message.
         """
         ...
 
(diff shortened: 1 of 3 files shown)


sdk/python/tests/unit/infra/online_store/test_dynamodb_online_store.py
```diff
@@ -175,6 +177,34 @@ def test_online_read(repo_config, n_samples):
...
+    features = list(features)
+    features.append(None)
+    dynamodb_store = DynamoDBOnlineStore()
+    returned_items = dynamodb_store.online_read(
+        config=repo_config,
+        table=MockFeatureView(name=f"{TABLE_NAME}_{n_samples}"),
+        entity_keys=entity_keys,
+    )
+    assert len(returned_items) == len(entity_keys)
+    assert [item[1] for item in returned_items] == list(features)
```
achals, line 205: Wanna assert that the nonsense key has none values?

adchia: Done

sdk/python/feast/infra/online_stores/online_store.py
```diff
@@ -76,9 +76,9 @@ def online_read(
             entity_keys: a list of entity keys that should be read from the FeatureStore.
             requested_features: (Optional) A subset of the features that should be read from the FeatureStore.
         Returns:
-            Data is returned as a list, one item per entity key. Each item in the list is a tuple
-            of event_ts for the row, and the feature data as a dict from feature names to values.
-            Values are returned as Value proto message.
+            Data is returned as a list, one item per entity key in the original order as the entity_keys argument.
+            Each item in the list is a tuple of event_ts for the row, and the feature data as a dict from feature names
+            to values. Values are returned as Value proto message.
```
achals, line 81: This should probably be highlighted - also smells like a fragile interface. But that discussion is out of scope for this PR.

achals (approved):
/lgtm

PR #2574, opened 2022-04-19
feat: Write logged features to an offline store (Python API)

pyalex:

This PR is part of Data Quality Monitoring. In order to validate served features we need first to log them and store in an offline store. This PR introduces "write logs" API in offline store and provider interfaces. As well as adds implementations for core offline stores: File, BigQuery, Redhift, Snowflake.

Fixes #


sdk/python/feast/infra/passthrough_provider.py
```diff
@@ -214,3 +217,51 @@ def retrieve_saved_dataset(
...
+            data=logs,
+            source=FeatureServiceLoggingSource(feature_service, config.project),
+            logging_config=feature_service.logging_config,
+            registry=registry,
+        )
+
+    def retrieve_feature_service_logs(
+        self,
+        feature_service: FeatureService,
+        from_: datetime,
```
adchia, line 243: might be better to use like start_date and end_date

sdk/python/feast/infra/offline_stores/bigquery_source.py
```diff
@@ -291,3 +295,28 @@ def to_proto(self) -> SavedDatasetStorageProto:
 
     def to_data_source(self) -> DataSource:
         return BigQuerySource(table=self.bigquery_options.table)
+
+
+class BigqueryLoggingDestination(LoggingDestination):
+    _proto_attr_name = "bigquery_destination"
+
+    table_ref: str
```
adchia, line 303: should probably be consistent with our other names (this would be table instead of table_ref)

adchia: similar for other DWH

sdk/python/feast/feature_logging.py
```diff
@@ -0,0 +1,151 @@
...
+    LoggingConfig as LoggingConfigProto,
+)
+from feast.types import from_value_type
+
+if TYPE_CHECKING:
+    from feast import FeatureService
+    from feast.registry import Registry
+
+
+class LoggingSource:
```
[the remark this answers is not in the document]
pyalex: Not exactly.
1. User doesn't instantiate logging source, this is being created mostly internally depending on where logs are coming from.
2. Currently the only available logging source is a feature server. Other example could be materialization job or streaming job.
3. We do not append to feature sources. Instead logs are being written to `LoggingDestination`. This one is defined by user as part of LoggingConfig in feature service (see changes to proto).

sdk/python/feast/infra/offline_stores/bigquery_source.py
```diff
@@ -291,3 +295,28 @@ def to_proto(self) -> SavedDatasetStorageProto:
...
+
+    def __init__(self, table_ref):
```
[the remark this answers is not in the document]
pyalex: Not sure I got the idea. See explanation above, this is not really a data source, rather data sink.

protos/feast/core/FeatureService.proto
```diff
@@ -46,3 +50,41 @@ message FeatureServiceMeta {
...
+  message RedshiftDestination {
+    string table_ref = 1;
+  }
+
+  message SnowflakeDestination {
+    // Snowflake table name
+    string table = 1;
+
+    // Snowflake schema name
+    string schema = 2;
```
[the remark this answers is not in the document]
pyalex: removed

sdk/python/feast/feature_logging.py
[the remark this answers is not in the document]
pyalex: I added some doc string.

protos/feast/core/FeatureService.proto
```diff
@@ -46,3 +50,34 @@ message FeatureServiceMeta {
...
+message LoggingConfig {
+  float sample_rate = 1;
+  google.protobuf.Duration partition_interval = 2;
+
+  oneof destination {
+    FileDestination file_destination = 3;
+    BigQueryDestination bigquery_destination = 4;
+    RedshiftDestination redshift_destination = 5;
+    SnowflakeDestination snowflake_destination = 6;
+  }
```
tsotnet, line 64: Is there a goal of making feature logging extensible? How will third-party offline stores enable feature logging?

Suggestion: we could have destination be a map<string, string> (essentially a flag JSON). We're losing strong typing, but other offline stores can also use it.

sdk/python/feast/feature_logging.py
```diff
@@ -0,0 +1,167 @@
...
+                        entity.join_key, entity.join_key
+                    )
+                    fields[join_key] = FEAST_TYPE_TO_ARROW_TYPE[
+                        from_value_type(entity.value_type)
+                    ]
+
+        # system columns
+        fields["request_id"] = pa.string()
+        fields["log_timestamp"] = pa.timestamp("us", tz=UTC)
+        fields["log_date"] = pa.date32()
```
tsotnet, line 97: Do we have a list of "system columns" blacklisted somewhere when we define features? How do you avoid column collisions?

sdk/python/feast/infra/offline_stores/redshift_source.py
```diff
@@ -372,3 +376,28 @@ def to_proto(self) -> SavedDatasetStorageProto:
...
+
+    def to_proto(self) -> LoggingConfigProto:
+        return LoggingConfigProto(
+            redshift_destination=LoggingConfigProto.RedshiftDestination(
+                table_ref=self.table
+            )
+        )
+
+    def to_data_source(self) -> DataSource:
+        return RedshiftSource(table=self.table)
```
tsotnet, line 403: I think you also need to pass `schema` (optional which defaults to `public` in RedshiftSource). Confusing name, but schemas as like namespaces / databases. If you don't specify schema, we generally use `public.table_name`, but you can override schema and say get `my_schema.table_name` which is different from `public.table_name`.

The idea is that user may want to use non-public Redshift schema for writing Feast logs.

pyalex: There was comment from @adchia suggesting that we should rather use schema from store configuration. See https://github.com/feast-dev/feast/pull/2574#discussion_r854413831

sdk/python/feast/feature_logging.py
pyalex: Good point. We use full feature names here, so collisions with feature names is unlikely. However, there might be conflicts with entities or request data. So I will add prefixes to this names.

protos/feast/core/FeatureService.proto
pyalex: That's done similarly to [data source proto](https://github.com/feast-dev/feast/blob/master/protos/feast/core/DataSource.proto), so the plan is to add CustomDestination with map or bytes later.

sdk/python/feast/feature_logging.py
pyalex: Added prefixes to system fields and put them into constants.

pyalex: I cannot find any list of forbidden names for entities (although we already have some system names, like `__dummy_entity`), so it's probably make sense to create an issue and address it in subsequent PR.

protos/feast/core/FeatureService.proto
pyalex: Added custom destination proto message.

pyalex: Added custom destination proto message.

pyalex:
@adchia @tsotnet thanks for review. I addressed your comments.

sdk/python/feast/infra/offline_stores/snowflake_source.py
```diff
@@ -382,3 +386,28 @@ def to_proto(self) -> SavedDatasetStorageProto:
...
     def to_data_source(self) -> DataSource:
         return SnowflakeSource(table=self.snowflake_options.table)
+
+
+class SnowflakeLoggingDestination(LoggingDestination):
+    table: str
+
+    _proto_kind = "snowflake_destination"
+
+    def __init__(self, table: str):
```
achals, line 396: same comment

sdk/python/feast/infra/utils/aws_utils.py
```diff
@@ -231,35 +221,92 @@ def upload_df_to_redshift(
...
+    schema: Optional[pyarrow.Schema] = None,
+    fail_if_exists: bool = True,
+):
+    """Uploads an Arrow Table to Redshift to a new or existing table.
+
+    Here's how the upload process works:
+        1. PyArrow Table is serialized into a Parquet format on local disk
+        2. The Parquet file is uploaded to S3
+        3. The S3 file is uploaded to Redshift as a new table through COPY command
+        4. The local disk & s3 paths are cleaned up
```
achals, line 256: never mind my earlier question, just saw this comment.

sdk/python/tests/integration/feature_repos/universal/data_source_creator.py
```diff
@@ -51,6 +52,10 @@ def create_offline_store_config(self) -> FeastConfigBaseModel:
     def create_saved_dataset_destination(self) -> SavedDatasetStorage:
         ...
 
+    @abstractmethod
+    def create_logged_features_destination(self) -> LoggingDestination:
+        ...
```
achals, line 57: I would create a dummy no-op implementation rather than create an `@abstractmethod` otherwise plugin tests will break when they get this change.

sdk/python/tests/integration/offline_store/test_feature_logging.py
```diff
@@ -0,0 +1,124 @@
...
+        feature_service=feature_service, project=store.project
+    ).get_schema(store._registry)
+
+    store.write_logged_features(
+        source=feature_service, logs=pa.Table.from_pandas(first_batch, schema=schema),
+    )
+
+    store.write_logged_features(
+        source=feature_service, logs=pa.Table.from_pandas(second_batch, schema=schema),
+    )
```
achals, line 67: I didn't see this call happening anywhere else in the code - is my understanding correct that this would need to be added to `get_online_features` before we would log anything?

sdk/python/feast/feature_logging.py
```diff
@@ -0,0 +1,173 @@
...
+
+    @abc.abstractmethod
+    def get_schema(self, registry: "Registry") -> pa.Schema:
+        """ Generate schema for logs destination. """
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def get_partition_column(self, registry: "Registry") -> str:
+        """ Return partition column that must exist in generated schema. """
+        raise NotImplementedError
```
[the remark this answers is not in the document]
pyalex: Agreed, it should be part of logging config, since most of offline store do not even support partitioning explicitly.

sdk/python/tests/integration/offline_store/test_feature_logging.py
pyalex: Integration would be added in following PRs. Yes, this function will be called from `ge_online_features`.

achals (approved):

PR #2588, opened 2022-04-21
fix: Feature with timestamp type is incorrectly interpreted by Go FS

pyalex:
Timestamp type stored in online store should be parsed with resolution "seconds".

Fixes #


adchia (approved):
/lgtm

achals:
> Timestamp type stored in online store should be parsed with resolution "seconds".

Is this true for all online stores? Is it something we need to document for users implementing their own online store?

PR #2589, opened 2022-04-21
chore: Extra dependencies in setup.py for Go <-> Python interop

pyalex:

In order for Python to fully interop with Go we might need to install some extra python packages. This PR adds new dependency group `go` to `setup.py`.

Fixes #

### DIFF
--- a/sdk/python/setup.py
+++ b/sdk/python/setup.py
@@ -114,7 +114,7 @@ GE_REQUIRED = [
 ]
 
 GO_REQUIRED = [
-    "cffi==1.15.0",
+    "cffi==1.15.*",
 ]
 
 CI_REQUIRED = (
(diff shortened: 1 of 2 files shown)


sdk/python/setup.py
```diff
@@ -113,6 +113,10 @@
     "great_expectations>=0.14.0,<0.15.0"
 ]
 
+GO_REQUIRED = [
+    "cffi==1.15.0",
```
achals, line 117: Is pinning in, or should we leave the patch version/minor version floating?

achals (commented):
Can you add some context in the PR description?

sdk/python/setup.py
pyalex: Sure, keeping patch version floating

achals:
Also, do we need to change any of our packaging scripts?

adchia (approved):
/lgtm

PR #2593, opened 2022-04-21
fix: Use timestamp type when converting unixtimestamp feature type to arrow

pyalex:

To be compatible with conversion of arrow -> various offline store formats we should use timestamp and not time. 

Fixes #


tsotnet (approved):
/lgtm

Issue #2595, opened 2022-04-21
Snowflake has connection issues


esadler-hbo:
## Expected Behavior
I try to connect to snowflake using a private key
## Current Behavior
I get an error from the snowflake python connection object saying that I need to set the warehouse.
## Steps to reproduce
I ran the demo
### Specifications

- Version:
- Platform:
- Subsystem:

## Possible Solution
I removed this double quoting and it worked

```python
# line 61 in snowflake_utils.py
    for k, v in kwargs.items():
        if k in ["role", "warehouse", "database", "schema_"]:
            kwargs[k] = f'"{v}"'
```

Happy to replicate through the testing setup and make a PR

sfc-gh-madkins:
@esadler-hbo are you setting creds in a config file?

esadler-hbo:
@sfc-gh-madkins yeah I am. I posted more details in #2601.

PR #2601, opened 2022-04-22
fix: Remove extra quoting in snowflake python connection

esadler-hbo:
Signed-off-by: Evan <evan.sadler@warnermedia.com>

Possible extra quotes. Need to verify by running integration tests.

https://github.com/feast-dev/feast/issues/2595

Fixes #

### DIFF
--- a/sdk/python/feast/infra/utils/snowflake_utils.py
+++ b/sdk/python/feast/infra/utils/snowflake_utils.py
@@ -59,10 +59,6 @@ def get_snowflake_conn(config, autocommit=True) -> SnowflakeConnection:
 
     kwargs.update((k, v) for k, v in config_dict.items() if v is not None)
 
-    for k, v in kwargs.items():
-        if k in ["role", "warehouse", "database", "schema_"]:
-            kwargs[k] = f'"{v}"'
-
     if "schema_" in kwargs:
         kwargs["schema"] = kwargs.pop("schema_")
     else:


esadler-hbo:
/assign @woop

I noticed an issue for me, but obviously the quotes could be important for somethings.  I was hoping to throw up a PR that would run the python integration test suite. 

Also let me know if you have ideas on how to test this more officially than editing the repo locally. I'm sure it is possible, but I am new to this repo!

achals:
@sfc-gh-madkins If I remember correctly you added this quite recently, could you take a look?

@esadler-hbo do you mind adding a test or something that reproduces the issue in #2595 without this patch?

sfc-gh-madkins:
would be great to understand what your feature_store.yaml / config file looks like?

esadler-hbo:
> would be great to understand what your feature_store.yaml / config file looks like?

I could totally be doing this wrong, but this is what I had. I did not use quotes for any of the filled in values. 
```yaml
# feature_store.yaml
project: firm_finch
registry: registry.db
provider: local
offline_store:
    type: snowflake.offline
    account: 
    user: 
    password: 
    role:
    warehouse:
    database:
    schema:
    config_path: "local path/config.ini"
```
The config file
```ini
[connections.feast_offline_store]
private_key_passphrase = phrase
private_key = path_to_key
```

esadler-hbo:
> @sfc-gh-madkins If I remember correctly you added this quite recently, could you take a look?
> 
> @esadler-hbo do you mind adding a test or something that reproduces the issue in #2595 without this patch?

Happy to add a test once @sfc-gh-madkins lets me know if my config setup makes sense. I just need to get setup with a test environment and then getting some direction to where and how be much appreciated.

sfc-gh-madkins:
my guess is you are not respecting case sensitivity -- most likely your warehouse is ALL CAPS in Snowflake but you are putting it in lowercase in your config file

Currently feast is using double quoted identifier syntax -- see https://docs.snowflake.com/en/sql-reference/identifiers-syntax.html

i appreciate that this is not documented well in the feast docs

esadler-hbo:
> my guess is you are not respecting case sensitivity -- most likely your warehouse is ALL CAPS in Snowflake but you are putting it in lowercase in your config file
> 
> Currently feast is using double quoted identifier syntax -- see https://docs.snowflake.com/en/sql-reference/identifiers-syntax.html
> 
> i appreciate that this is not documented well in the feast docs

Well I reset my environment and I couldn't reproduce the issue. Sorry to bother you all!

PR #2602, opened 2022-04-22
fix: Fix java server after odfv update

kevjumba:

Java server is run against latest master version so I can only update definitions after odfv parameter changes are merged in.

Fixes #


achals (approved):
So this is still working in a backwards compatible way right?

kevjumba:
> backwards

yup

PR #2609, opened 2022-04-25
fix: Use C-type long long (8 bytes) type format to pack int64 values

### LINKED ISSUE #2345: INT64 uses "C long" type in struct.pack which is 4 bytes.
## Expected Behavior 

INT64 should be packed as "C long long" to have 64 bits memory capacity. 

## Current Behavior

INT64 is packed as "C long" which is 32 bits

## Steps to reproduce

Materialization of entities with IDs larger than supported range of "C long" raises exception.

### Specifications

The bug is in the _serialize_val function in key_encoding_utils.
```
def _serialize_val(value_type, v: ValueProto) -> Tuple[bytes, int]:
    if value_type == "string_val":
        return v.string_val.encode("utf8"), ValueType.STRING
    elif value_type == "bytes_val":
        return v.bytes_val, ValueType.BYTES
    elif value_type == "int32_val":
        return struct.pack("<i", v.int32_val), ValueType.INT32
    elif value_type == "int64_val":
        return struct.pack("<q", v.int64_val), ValueType.INT64
    else:
        raise ValueError(f"Value type not supported for Firestore: {v}")
```

- Version:
- feast: 0.18.1
- Platform:
- MacOS 12.0.1
- Subsystem:
- python: 3.8.1

## Possible Solution

changing `struct.pack("<l", v.int64_val), ValueType.INT64` to `struct.pack("<q", v.int64_val), ValueType.INT64` fixes this.

roelschr:
This PR fixes a bug related to the materialization of INT64 entity keys, by packing int64 values as C-type "long long". It also adds a new test that tests the serialization behavior with an INT64 entity key value bigger than int32 max.

Fixes #2345.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -108,12 +108,15 @@ jobs:
           node-version: '17.x'
           registry-url: 'https://registry.npmjs.org'
       - name: Build and install dependencies
+        # There's a `git restore` in here because `make install-go-ci-dependencies` is actually messing up go.mod & go.sum.
         run: |
           pip install -U pip setuptools wheel twine
           make install-protoc-dependencies
           make install-go-proto-dependencies
           make install-go-ci-dependencies
           make build-ui
+          git status
+          git restore go.mod go.sum
       - name: Build
         run: |
           python3 setup.py sdist
--- a/sdk/python/feast/infra/key_encoding_utils.py
+++ b/sdk/python/feast/infra/key_encoding_utils.py
@@ -14,7 +14,7 @@ def _serialize_val(value_type, v: ValueProto) -> Tuple[bytes, int]:
     elif value_type == "int32_val":
         return struct.pack("<i", v.int32_val), ValueType.INT32
     elif value_type == "int64_val":
-        return struct.pack("<l", v.int64_val), ValueType.INT64
+        return struct.pack("<q", v.int64_val), ValueType.INT64
     else:
         raise ValueError(f"Value type not supported for Firestore: {v}")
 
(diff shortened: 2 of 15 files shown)


roelschr:
Converted to draft to work on a backward compatible solution.

achals:
@roelschr friendly ping, are you working on this? If not, I'd be more than happy to invest some time here to introduce this compatibility layer.

.github/workflows/build_wheels.yml
```diff
@@ -108,12 +108,15 @@ jobs:
           node-version: '17.x'
           registry-url: 'https://registry.npmjs.org'
       - name: Build and install dependencies
+        # There's a `git restore` in here because `make install-go-ci-dependencies` is actually messing up go.mod & go.sum.
```
felixwang9817, line 111: remove this stuff

sdk/python/feast/infra/key_encoding_utils.py
```diff
@@ -6,15 +6,17 @@
...
+) -> Tuple[bytes, int]:
     if value_type == "string_val":
         return v.string_val.encode("utf8"), ValueType.STRING
     elif value_type == "bytes_val":
         return v.bytes_val, ValueType.BYTES
     elif value_type == "int32_val":
         return struct.pack("<i", v.int32_val), ValueType.INT32
     elif value_type == "int64_val":
-        return struct.pack("<l", v.int64_val), ValueType.INT64
+        return struct.pack("<q", v.int64_val), ValueType.INT64
```
felixwang9817, line 19: not sure I understand the compatibility approach you're taking here - can you explain?

what do the various values of `entity_key_serialization_version` represent? and there needs to be some fork in the behavior here depending on `entity_key_serialization_version` right?

I was imagining we should do something like the following:
* add `entity_key_serialization_version` to the protos, as you've done, where 0 will represent the current (broken) serialization method and 1 will represent the new (correct) serialization method
* ensure that existing feature view protos do not magically get `entity_key_serialization_version` switched from 0 (the default value since the field doesn't exist) to 1, which requires special logic in the registry (as I think you've done)
* default all new feature views to use serialization method 1

PR #2610, opened 2022-04-25
fix: Update field api to add tag parameter corresponding to labels in Feature.

### LINKED ISSUE #2607: Keep labels in Field api
I found that new api 'Field' will take place of 'Feature' in 0.21+ feast. but `Field` only have 'name' and 'dtype' parameters. The parameter 'labels' is disappeared. 
In my use case 'labels' is very import.  'labels' stores the default value, descriptions,and other things. for example

```python
comic_feature_view = FeatureView(
    name="comic_featureV1",
    entities=["item_id"],
    ttl=Duration(seconds=86400 * 1),
    features=[
        Feature(name="channel_id", dtype=ValueType.INT32, labels={"default": "14", "desc":"channel"}),
        Feature(name="keyword_weight", dtype=ValueType.FLOAT, labels={"default": "0.0", "desc":"keyword's weight"}),
        Feature(name="comic_vectorv1", dtype=ValueType.FLOAT, labels={"default": ";".join(["0.0" for i in range(32)]), "desc":"deepwalk vector","faiss_index":"/data/faiss_index/comic_featureV1__comic_vectorv1.index"}),
        Feature(name="comic_vectorv2", dtype=ValueType.FLOAT, labels={"default": ";".join(["0.0" for i in range(32)]), "desc":"word2vec vector","faiss_index":"/data/faiss_index/comic_featureV1__comic_vectorv2.index"}),
        Feature(name="gender", dtype=ValueType.INT32, labels={"default": "0", "desc":" 0-femal 1-male"}),
        Feature(name="pub_time", dtype=ValueType.STRING, labels={"default": "1970-01-01 00:00:00", "desc":"comic's publish time"}),
        Feature(name="update_time", dtype=ValueType.STRING, labels={"default": "1970-01-01 00:00:00", "desc":"comic's update time"}),
        Feature(name="view_cnt", dtype=ValueType.INT64, labels={"default": "0", "desc":"comic's hot score"}),
        Feature(name="collect_cnt", dtype=ValueType.INT64, labels={"default": "0", "desc":"collect count"}),
        Feature(name="source_id", dtype=ValueType.INT32, labels={"default": "0", "desc":"comic is from(0-unknown，1-japen，2-usa，3- other)"}),
```

So please keep the parameter 'labels' in Field api

kevjumba:

Fixes #2607

### DIFF
--- a/protos/feast/types/Field.proto
+++ b/protos/feast/types/Field.proto
@@ -29,7 +29,7 @@ message Field {
   string name = 1;
 
   // Value type of the feature. Not updatable.
-  feast.types.ValueType.Enum value = 2;
+  feast.types.ValueType.Enum value_type = 2;
 
   // User defined metadata
   map<string,string> tags = 3;
--- a/protos/feast/types/Field.proto
+++ b/protos/feast/types/Field.proto
@@ -25,7 +25,10 @@ option java_outer_classname = "FieldProto";
 option go_package = "github.com/feast-dev/feast/go/protos/feast/types";
 
 message Field {
+  // Name of the feature. Not updatable.
   string name = 1;
+
+  // Value type of the feature. Not updatable.
   feast.types.ValueType.Enum value = 2;
 
   // User defined metadata
--- a/protos/feast/types/Field.proto
+++ b/protos/feast/types/Field.proto
@@ -27,4 +27,7 @@ option go_package = "github.com/feast-dev/feast/go/protos/feast/types";
 message Field {
   string name = 1;
   feast.types.ValueType.Enum value = 2;
+
+  // User defined metadata
+  map<string,string> tags = 3;
 }
--- a/protos/feast/types/Field.proto
+++ b/protos/feast/types/Field.proto
@@ -25,12 +25,6 @@ option java_outer_classname = "FieldProto";
 option go_package = "github.com/feast-dev/feast/go/protos/feast/types";
 
 message Field {
-  // Name of the feature. Not updatable.
   string name = 1;
-
-  // Value type of the feature. Not updatable.
-  feast.types.ValueType.Enum value_type = 2;
-
-  // User defined metadata
-  map<string,string> tags = 3;
+  feast.types.ValueType.Enum value = 2;
 }
--- a/sdk/python/feast/field.py
+++ b/sdk/python/feast/field.py
@@ -16,9 +16,8 @@ from typing import Dict, Optional
 
 from feast.feature import Feature
 from feast.protos.feast.core.Feature_pb2 import FeatureSpecV2 as FieldProto
-from feast.types import FeastType, from_value_type
+from feast.types import FeastType, PrimitiveFeastType, from_value_type
 from feast.value_type import ValueType
-from feast.types import PrimitiveFeastType
 
 
 class Field:
--- a/sdk/python/feast/field.py
+++ b/sdk/python/feast/field.py
@@ -15,8 +15,8 @@
 from typing import Dict, Optional
 
 from feast.feature import Feature
-from feast.protos.feast.core.Feature_pb2 import FeatureSpecV2 as FieldProto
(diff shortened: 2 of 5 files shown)


achals (approved):
/lgtm

sdk/python/feast/field.py
```diff
@@ -25,31 +27,43 @@ class Field:
...
 
         Args:
             name: The name of the field.
             dtype: The type of the field, such as string or float.
+            tags (optional): User-defined metadata in dictionary form.
         """
-        self.name = name
-        self.dtype = dtype
+        self.name = name or ""
+        self.dtype = dtype or PrimitiveFeastType.INVALID
```
felixwang9817, line 53: can we directly import `Invalid` instead of using `PrimitiveFeastType`?

sdk/python/feast/field.py
```diff
@@ -25,31 +27,43 @@ class Field:
...
 
     name: str
     dtype: FeastType
+    tags: Dict[str, str]
 
     def __init__(
-        self, *, name: str, dtype: FeastType,
+        self,
+        *,
+        name: Optional[str] = None,
```
felixwang9817, line 40: can we leave this as required (as well as `dtype`) but make `tags` optional?

sdk/python/feast/field.py
```diff
@@ -25,31 +27,43 @@ class Field:
...
+            self.name != other.name
+            or self.dtype != other.dtype
+            or self.tags != other.tags
+        ):
             return False
         return True
 
     def __hash__(self):
-        return hash((self.name, hash(self.dtype)))
+        return hash((self.name, hash(self.dtype), hash(self.tags)))
```
felixwang9817, line 66: I don't think you can hash a `Dict`? it's fine to just leave this out

sdk/python/feast/field.py
```diff
@@ -58,12 +72,12 @@ def __repr__(self):
...
 
     def __str__(self):
-        return f"Field(name={self.name}, dtype={self.dtype})"
+        return f"Field(name={self.name}, dtype={self.dtype}, tags={self.tags})"
 
     def to_proto(self) -> FieldProto:
         """Converts a Field object to its protobuf representation."""
         value_type = self.dtype.to_value_type()
-        return FieldProto(name=self.name, value_type=value_type.value)
+        return FieldProto(name=self.name, value_type=value_type.value, labels=self.tags)
```
felixwang9817, line 80: this is fine, although we should probably rename `labels` to `tags` in `FeatureSpecV2` (which has been imported here as `FieldProto`)

protos/feast/types/Field.proto
```diff
@@ -27,4 +27,7 @@ option go_package = "github.com/feast-dev/feast/go/protos/feast/types";
 message Field {
   string name = 1;
   feast.types.ValueType.Enum value = 2;
+
+  // User defined metadata
+  map<string,string> tags = 3;
```
felixwang9817, line 32: I don't think this change is necessary rn; we've been using `FeatureSpecV2` as the underlying proto structure

that being said, if you want we can migrate from using `FeatureSpecV2` to using `Field`

sdk/python/feast/field.py
kevjumba: Done.

sdk/python/feast/field.py
kevjumba: Sure.

sdk/python/feast/field.py
kevjumba: Removed.

protos/feast/types/Field.proto
kevjumba: I will migrate, I think that makes more sense.

sdk/python/feast/field.py
kevjumba: I updated it to use Field.proto instead.

felixwang9817 (approved):
/lgtm

PR #2611, opened 2022-04-26
chore: Deprecate value type

felixwang9817:
**What this PR does / why we need it**: This PR deprecates the `value_type` parameter for entities. Users are asked to specify entity types by adding a `Field` to the `schema` of a feature view.

Fixes #


go/internal/feast/model/featureview.go
```diff
@@ -13,12 +13,13 @@ const (
...
-var DUMMY_ENTITY types.Value = types.Value{Val: &types.Value_StringVal{StringVal: DUMMY_ENTITY_VAL}}
+var DUMMY_ENTITY_VALUE types.Value = types.Value{Val: &types.Value_StringVal{StringVal: DUMMY_ENTITY_VAL}}
 
 type FeatureView struct {
-	Base     *BaseFeatureView
-	Ttl      *durationpb.Duration
-	Entities []string
+	Base          *BaseFeatureView
+	Ttl           *durationpb.Duration
+	Entities      []string
```
pyalex, line 21: do I understand correctly that `Entities` consists of entity names and `EntityColumns` stores join keys?

go/internal/feast/model/featureview.go
```diff
@@ -30,23 +31,37 @@ func NewFeatureViewFromProto(proto *core.FeatureView) *FeatureView {
...
 		Ttl:      &ttl,
-		Entities: fs.Entities,
+		Entities: fv.Entities,
 	}
 	return featureView
 }
 
-func (fs *FeatureView) HasEntity(lookup string) bool {
-	for _, entityName := range fs.Entities {
+func (fv *FeatureView) HasEntity(lookup string) bool {
```
pyalex, line 51: maybe rename lookup to EntityName?

go/internal/feast/model/featureview.go
```diff
@@ -30,23 +31,37 @@ func NewFeatureViewFromProto(proto *core.FeatureView) *FeatureView {
...
+	for _, entityName := range fv.Entities {
 		if entityName == lookup {
 			return true
 		}
 	}
 	return false
 }
+
+func (fv *FeatureView) GetEntityType(lookup string) types.ValueType_Enum {
```
pyalex, line 60: and here to JoinKey?

go/internal/feast/model/featureview.go
```diff
@@ -13,12 +13,13 @@ const (
...
+var DUMMY_ENTITY_VALUE types.Value = types.Value{Val: &types.Value_StringVal{StringVal: DUMMY_ENTITY_VAL}}
 
 type FeatureView struct {
-	Base     *BaseFeatureView
-	Ttl      *durationpb.Duration
-	Entities []string
+	Base          *BaseFeatureView
+	Ttl           *durationpb.Duration
+	Entities      []string
+	EntityColumns []*Feature
```
pyalex, line 22: should we rename `Feature` to `Field` similarly to Python code?

pyalex:
General question: how do we resolve conflicts now if two entities with the same defined in different feature views have different types? How will grouping by entities work (during online serving)?

sdk/python/feast/entity.py
```diff
@@ -210,7 +229,6 @@ def to_proto(self) -> EntityProto:
 
         spec = EntitySpecProto(
             name=self.name,
-            value_type=self.value_type.value,
```
pyalex, line 213: it looks like we're removin value type completely and not deprecating it. should we write it to registry for a while?

sdk/python/feast/feature_logging.py
```diff
@@ -87,9 +86,14 @@ def get_schema(self, registry: "Registry") -> pa.Schema:
                     join_key = projection.join_key_map.get(
                         entity.join_key, entity.join_key
                     )
-                    fields[join_key] = FEAST_TYPE_TO_ARROW_TYPE[
-                        from_value_type(entity.value_type)
-                    ]
+                    entity_columns = list(
```
pyalex, line 89: can't we just iterate over entity columns in this for loop? seems like it has all information that we need for the schema generation.

go/internal/feast/model/featureview.go
felixwang9817: > do I understand correctly that Entities consists of entity names and EntityColumns stores join keys?

yes, so I'll change `Entities` -> `EntityNames`

go/embedded/online_features.go
```diff
@@ -117,7 +117,7 @@ func (s *OnlineFeatureService) GetEntityTypesMapByFeatureService(featureServiceN
 		}
 		for _, entityName := range view.Entities {
```
pyalex, line 118: we can replace this loop to `range view.EntityColumns` since it already has join key and type, right?

go/internal/feast/model/featureview.go
felixwang9817: renamed to `name` since `entityName` would conflict with the variable name

go/internal/feast/model/featureview.go
felixwang9817: good catch, renamed

go/internal/feast/model/featureview.go
felixwang9817: yup

sdk/python/feast/entity.py
felixwang9817: good catch

pyalex (approved):

sdk/python/feast/entity.py
```diff
@@ -104,8 +104,23 @@ def __init__(
         if not self.name:
             raise ValueError("Name needs to be specified")
 
-        self.value_type = value_type
+        if value_type:
+            warnings.warn(
+                (
+                    "The `value_type` parameter is being deprecated. Instead, the type of an entity "
+                    "should be specified as a Field in the schema of a feature view. Feast 0.22 and "
```
achals, line 111: 0.23? since we may be releasing a new minor version soon.

sdk/python/feast/feature_store.py
```diff
@@ -687,6 +683,9 @@ def apply(
 
         data_sources_to_update = list(data_sources_set_to_update)
 
+        # Handle all entityless feature views by using DUMMY_ENTITY as a placeholder entity.
+        entities_to_update.append(DUMMY_ENTITY)
```
achals, line 687: Any reason this moved up?

PR #2630, opened 2022-05-02
chore: Update python packaging to use build to produce sdists/wheels

achals:

Fixes #


setup.py
```diff
@@ -57,7 +57,6 @@
     "Jinja2>=2.0.0",
     "jsonschema",
     "mmh3",
-    "numpy<1.22",  # 1.22 drops support for python 3.7.
```
kevjumba, line 60: Why remove this?

achals: oh hm I think this was by mistake

setup.py
```diff
@@ -246,17 +247,19 @@ def run(self):
             # so that they are regular packages, and don't need the `--namespace-packages` flags
             # when being typechecked using mypy. BUT, we need to exclude `types` because that clashes
             # with an existing module in the python standard library.
-            if sub_folder == "types":
-                continue
+
+            # if sub_folder == "types":
```
kevjumba, line 251: Is this still needed, if not probably should remove.

kevjumba (approved):

PR #2638, opened 2022-05-04
chore: Update new contrib plugins to include usage

adchia:

Fixes #


sdk/python/feast/infra/online_stores/contrib/postgres.py
```diff
@@ -33,6 +34,7 @@ def _get_conn(self, config: RepoConfig):
             self._conn = _get_conn(config.online_store)
         return self._conn
 
+    @log_exceptions_and_usage(online_store="postgres")
```
achals, line 37: should this be `postgres-online` or something? Or will we be able to infer from the methods?

adchia: we can infer because it's a different key (under online store)

achals (approved):
/lgtm

PR #2640, opened 2022-05-05
feat: Validating logged features via Python SDK

pyalex:

This PR adds `FeatureStore.validate_logged_features` method that utilizes previously created infrastructure to load logged features and validate them against reference (saved) dataset via GE profiler. 

Fixes #


achals (approved):
/lgtm

PR #2645, opened 2022-05-05
chore: Use latest Python code in Java Integration Tests

pyalex:

Mount root repo directory in docker compose to install latest Python code. This will allow Python and Java code to be in sync during integration tests.

Fixes #


achals (approved):
/lgtm

PR #2647, opened 2022-05-06
fix: Pin dependencies to nearest major version

### LINKED ISSUE #2557: Pin dependencies to major version ranges
We have recently had a bunch of issues due to dependencies not being limited appropriately. For example, having `google-cloud-bigquery>=2.28.1` led to issues when `google-cloud-bigquery` released breaking changes in `v3.0.0`: see #2537 for the issue and #2554 which included the fix. Similarly, #2484 occurred since our `protobuf` dependency was not limited.

I think we should limit dependencies to the next major version. For example, if we currently use version N of a package, we should also limit it to v<(N+1). This way we are not exposed to breaking changes in all our upstream dependencies, while also maintaining a reasonable amount of flexibility for users. If a version N+1 is released and users want us to support it, they can let us know and we can add support; limiting to v<(N+1) just ensures that we aren't being broken all the time.

kevjumba:
Pins dependencies to major version range to prevent breaking changes with newest version.

Fixes #2557


achals (approved):
/lgtm

sdk/python/requirements/py3.10-requirements.txt
```diff
@@ -32,28 +32,28 @@ dask==2022.1.1
...
+fastapi==0.76.0
     # via feast (setup.py)
-fastavro==1.4.10
+fastavro==1.4.11
     # via
     #   feast (setup.py)
     #   pandavro
 fsspec==2022.3.0
     # via dask
-google-api-core==2.7.2
```
achals, line 43: should we be concerned that we went down a whole major version here? Is it even still supported?

sdk/python/requirements/py3.10-requirements.txt
```diff
@@ -32,28 +32,28 @@ dask==2022.1.1
...
+google-api-core==1.31.5
     # via feast (setup.py)
-google-auth==2.6.5
```
achals, line 45: Same q

setup.py
```diff
@@ -47,46 +47,46 @@
...
+    "colorama>=0.3.9,<1",
     "dill==0.3.*",
-    "fastavro>=1.1.0",
-    "google-api-core>=1.23.0",
-    "googleapis-common-protos==1.52.*",
-    "grpcio>=1.34.0",
-    "grpcio-reflection>=1.34.0",
-    "Jinja2>=2.0.0",
+    "fastavro>=1.1.0,<2",
+    "google-api-core>=1.23.0,<2",
```
achals, line 53: We should probably do 
```suggestion
    "google-api-core>=2,<3",
```

setup.py
```diff
@@ -47,46 +47,46 @@
...
+    "googleapis-common-protos==1.52.*,<2",
```
achals, line 54: Same here

setup.py
```diff
@@ -47,46 +47,46 @@
...
-    "fastapi>=0.68.0",
-    "uvicorn[standard]>=0.14.0",
+    "fastapi>=0.68.0,<1",
+    "uvicorn[standard]>=0.14.0,<1",
     "proto-plus<1.19.7",
     "tensorflow-metadata>=1.0.0,<2.0.0",
     "dask>=2021.*,<2022.02.0",
 ]
 
 GCP_REQUIRED = [
```
achals, line 79: All of these should probably have the same version ranges

setup.py
kevjumba: This is already newest version (1.52.*)

sdk/python/requirements/py3.10-requirements.txt
kevjumba: Hm this was specific to 3.10. Every other version has a major version of 1.

sdk/python/requirements/py3.10-requirements.txt
kevjumba: Similar

setup.py
kevjumba: Yup I checked the version against 3.8 and there weren't any major version bumps. I think it's only 3.10 that has a version bump. Do you have a suggestion for what I should do. I can bump google-api-core restrictions up to 3?

achals: Sure, that's fine by me.

setup.py
achals: ok

PR #2649, opened 2022-05-06
fix: Feature service to templates

kevjumba:

Fixes #

### DIFF
--- a/docs/getting-started/quickstart.md
+++ b/docs/getting-started/quickstart.md
@@ -367,7 +367,7 @@ pprint(feature_vector)
 
 ## Step 7: Using a featureservice to fetch online features instead.
 
-You can also use the feature service to manage multiple features and be a central source of truth for a series of feature views. The feature store can also be used to fetch either online or historical features using the same api below. More information can be found [here](https://docs.feast.dev/getting-started/concepts/feature-service).
+You can also use feature services to manage multiple features, and decouple feature view definitions and the features needed by end applications. The feature store can also be used to fetch either online or historical features using the same api below. More information can be found [here](https://docs.feast.dev/getting-started/concepts/feature-service).
 
 {% tabs %}
 {% tab title="Python" %}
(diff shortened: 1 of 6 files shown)


docs/getting-started/quickstart.md
```diff
@@ -355,6 +365,41 @@ pprint(feature_vector)
 {% endtab %}
 {% endtabs %}
 
+## Step 7: Using a featureservice to fetch online features instead.
+
+You can also use the feature service to manage multiple features and be a central source of truth for a series of feature views. The feature store can also be used to fetch either online or historical features using the same api below. More information can be found [here](https://docs.feast.dev/getting-started/concepts/feature-service).
```
achals, line 370:
```suggestion
You can also use feature services to manage multiple features, and decouple feature view definitions and the features needed by end applications. The feature store can also be used to fetch either online or historical features using the same api below. More information can be found [here](https://docs.feast.dev/getting-started/concepts/feature-service).
```

docs/getting-started/quickstart.md
```diff
@@ -355,6 +365,41 @@ pprint(feature_vector)
...
+```
+
+{% tabs %}
+{% tab title="Output" %}
+```bash
+{
+ 'acc_rate': [0.5732735991477966, 0.7828438878059387],
+ 'avg_daily_trips': [33, 984],
+ 'conv_rate': [0.15498852729797363, 0.6263588070869446],
+ 'driver_id': [1004, 1005]
```
achals, line 396: Do we expect these exact values? If not, we should caveat that the numbers here in the example may be different from what users see.

docs/getting-started/quickstart.md
kevjumba: Done.

docs/getting-started/quickstart.md
kevjumba: Done.

PR #2653, opened 2022-05-09
fix: Infer features for feature services when they depend on feature views without schemas

adchia:
If a feature service depends on a feature view that doesn't have a schema (i.e. relying on feature inference), the feature service does not inherit the inferred schema.

This results in a particularly weird error for FileSources suggesting that pandas dataframes do not have a "persist()" method (like Dask dataframes do)

Fixes #


felixwang9817 (approved):
/lgtm

PR #2659, opened 2022-05-10
chore: Add feature repo for version 0.19

felixwang9817:
**What this PR does / why we need it**: This PR adds a feature repo that uses the SDK from version 0.19 and uses it in a test. The feature repo does not use a `FeatureService` or data sources other than a `FileSource`.

Fixes #


achals (approved):
/lgtm

Issue #2353, opened 2022-03-02
Feast Web UI


adchia:
**Is your feature request related to a problem? Please describe.**
Several things are harder than they should be today:
- Tag searching / filtering
- Discovery of feature views / data sources / models
- Understanding lineage of models are difficult.
- Understanding what features are available / materialized to the online store
- Managing features across projects

**Describe the solution you'd like**
A Feast Web UI that enables discovery of features, data sources, and models across projects. 

- [ ] Tests checking that changes to the feast registry protos don't break the web UI (i.e. an integration test that starts up the Feast UI)
- [x] a feast cli command like `feast ui` that scans subdirectories for `feature_store.yaml` and calls `feast registry-dump` and pipes it to a known location for the web UI to reference. It also should produce the `project-list.json `that we use for the landing page. This should also start the local web ui server.
- [x] Improved search bar experience that does type-ahead suggestions on tag names
- [ ] release process that automatically publishes the npm package.

After initial launch
- [ ] A project level description config (e.g. in `feature_store.yaml` which populates the landing page of a given project in the UI and is output in the "registry" dump)
- [ ] Working across all types
- [ ] Links to e.g. feast-dev documentation (e.g a ? that shows details on hover)
- [ ] Flagging system for different UI components
- [x] Custom plugin system
- [x] Having a UI that dynamically updates + re-syncs against the registry file
- [ ] Migration towards perhaps https://github.com/protobufjs/protobuf.js/ instead of zod to parse data
- [ ] There's a "statistics" page in feature views that needs data to be populated. It's hidden by default behind a flag, but once we start pulling sample data and statistics in generating the json dump, this can be re-enabled
- [ ] Cleaned up UI for the data sets that gives information about the declared expectations
- [ ] UI for the data sets that highlights validation errors in a saved dataset

**Describe alternatives you've considered**

**Additional context**

woop:
>  There's a "statistics" page in feature views that needs data to be populated. It's hidden by default behind a flag, but once we start pulling sample data and statistics in generating the json dump, this can be re-enabled

Have you spent time thinking about the initial stats that you want to trend here, and whether we can start producing those as part of DQM efforts?

adchia:
Yeah cc @pyalex and @kevjumba 

Simple statistics I think will go a long way. e.g. for numeric features min / max / % missing / # unique values etc. For DQM, probably also want a way to compare two data sets side by side. 

[...]

PR #2672, opened 2022-05-11
fix: Asynchronously refresh registry for the feast ui command

adchia:
Also allows adding host + port names in the `feast ui` command, removes some unnecessary files, reverts an accidental change to the public demo in ui/, and bumps version numbers

https://github.com/feast-dev/feast/issues/2353

Fixes #

### DIFF
--- a/sdk/python/feast/ui/package.json
+++ b/sdk/python/feast/ui/package.json
@@ -6,7 +6,7 @@
     "@elastic/datemath": "^5.0.3",
     "@elastic/eui": "^57.0.0",
     "@emotion/react": "^11.9.0",
-    "@feast-dev/feast-ui": "^0.20.2",
+    "@feast-dev/feast-ui": "^0.20.3",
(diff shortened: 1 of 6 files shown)


felixwang9817 (approved):
/lgtm

sdk/python/feast/ui/package.json
```diff
@@ -6,7 +6,7 @@
     "@elastic/datemath": "^5.0.3",
     "@elastic/eui": "^57.0.0",
     "@emotion/react": "^11.9.0",
-    "@feast-dev/feast-ui": "^0.20.2",
+    "@feast-dev/feast-ui": "^0.20.3",
```
achals, line 9: Should this be bumped by the release process?

adchia: it's kind of wonky right now and I have to fix it. NPM Publishes require an OTP so I'll have to fix the workflow

PR #2676, opened 2022-05-12
fix: Remove incorrect call to logging.basicConfig

### LINKED ISSUE #2664: basicConfig is called at the module level
## Expected Behavior 

```
import feast
logging.basicConfig(level=level, format=FORMAT)
logging.error("msg")
```

should print logging message according to `FORMAT`

## Current Behavior

It uses the format defined in `feast` at the module level.

## Steps to reproduce

Same as in "Expected Behavior"

### Specifications

- Version: 0.18.1
- Platform: Linux
- Subsystem: -

## Possible Solution

I see that `basicConfig` is called here: https://github.com/feast-dev/feast/blob/c9eda79c7b1169ef05a481a96f07960c014e88b9/sdk/python/feast/cli.py#L84 so it is possible that simply removing this call here is enough: https://github.com/feast-dev/feast/blob/0ca62970dd6bc33c00bd5d8b828752814d480588/sdk/python/feast/__init__.py#L30

If there are any other entry points that need to set up logging, they should call the function, but the call in `__init__.py` must be removed.

achals:
This configures logging for all applications may just be interested in importing the feast package, which is incorrect. We should only configure logging at the application level, so things like the CLI, or or webservers.

Fixes #2664

### DIFF
--- a/sdk/python/feast/__init__.py
+++ b/sdk/python/feast/__init__.py
@@ -1,5 +1,3 @@
-import logging
-
 from pkg_resources import DistributionNotFound, get_distribution
 
 from feast.infra.offline_stores.bigquery_source import BigQuerySource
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -48,9 +48,6 @@ feast apply
 ### 4. Explore your data in the web UI (experimental)
 
 ![Web UI](ui/sample.png)
```
adchia, line 50: Revert?

adchia (approved):

README.md
achals: Not sure how that snuck in there

adchia (approved):
/lgtm

PR #2677, opened 2022-05-12
fix: Forcing ODFV udfs to be __main__ module and fixing false positive duplicate data source warning

adchia:
- dill only can properly serialize a function if the module is __main__.
    - Otherwise, when you deserialize, it looks for the presence of the exact module the ODFV was defined in (e.g. so if you don’t have the feature repo handy, using odfvs will fail)
    - Forcing the udf to be __main__ resolves this issue. This should otherwise be fine since we expect the udf to be able to execute in arbitrary environments.
- Fixing duplicate data source conflict between push source and batch source which occurs if you depend on both a push source and a batch source. 

Fixes #


achals (commented):
/lgtm

PR #2682, opened 2022-05-13
fix: Fix case where on demand feature view tab is broken if no custom tabs are passed. 

adchia:
Also allow projectListPromise to be passed without a tabs registry

Continuing with #2353 

Fixes #


achals (approved):
/lgtm

PR #2686, opened 2022-05-13
fix: Use ParquetDataset for Schema Inference

### LINKED ISSUE #2685: Parquet Schema Inference only supports File, not directory
When using a FileSource that is in Parquet format, if the source happens to be a directory of partitioned Parquet files, the following lines throw an error:

https://github.com/feast-dev/feast/blob/01d3568168bb9febb9fbda4988283b3886c32a31/sdk/python/feast/infra/offline_stores/file_source.py#L182-L184

`OSError: Expected file path, but /home/ubuntu/project/data/driver_stats_partitioned is a directory`

How to replicate:

1. Start with a demo feast project (`feast init`)
2. Create a partitioned Parquet Dataset.  Use the following to create a dataset with only a single timestamp for inference
```
import pyarrow.parquet as pq
df = pq.read_table("./data/driver_stats.parquet")
df = df.drop(["created"])
pq.write_to_dataset(df, "./data/driver_stats_partitioned")
```
3. Update the file source in `example.py` to look like this:
```
driver_hourly_stats = FileSource(
    path="/home/ubuntu/cado-feast/feature_store/exciting_sunbeam/data/driver_stats_partitioned2",
)
```

4. Run `feast apply`
For now, I've been able to fix by updating the above lines to:
```
schema = ParquetDataset(
    path if filesystem is None else filesystem.open_input_file(path)
).schema.to_arrow_schema()
```

dvanbrug:
Updates File Schema inference in offline stores to use ParquetDataset instead of ParquetFile to do schema inference.  This supports both single files and directories of partitioned parquet datasets.

Fixes #2685

### DIFF
--- a/sdk/python/feast/infra/offline_stores/file_source.py
+++ b/sdk/python/feast/infra/offline_stores/file_source.py
@@ -3,7 +3,7 @@ from typing import Callable, Dict, Iterable, List, Optional, Tuple
 
 from pyarrow._fs import FileSystem
 from pyarrow._s3fs import S3FileSystem
-from pyarrow.parquet import ParquetFile
+from pyarrow.parquet import ParquetDataset
 
 from feast import type_map
 from feast.data_format import FileFormat, ParquetFormat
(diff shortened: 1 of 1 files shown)


adchia (approved):
/lgtm

adchia:
Thanks! This is great!

PR #2692, opened 2022-05-14
fix: Fix issue when user specifies a port for feast ui 

adchia:
Changes the JS to look for a fixed projects-list.json (aka the default), and now `feast ui` will write to that json file so we can configure the host + port appropriately.

Before, this was hard-coded so users could not actually control the host / port.

Fixes #

### DIFF
--- a/sdk/python/feast/ui/public/projects-list.json
+++ b/sdk/python/feast/ui/public/projects-list.json
@@ -1,0 +1,11 @@
+{
+    "projects": [
+      {
+        "name": "Project",
+        "description": "Test project",
+        "id": "project_id",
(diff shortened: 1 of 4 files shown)


sdk/python/feast/ui/public/projects-list.json
```diff
@@ -0,0 +1,11 @@
+{
```
achals, line 1: Is this file needed at all, if the `feast ui` command is generating it on the fly?

adchia: I changed this because to generate it on the fly, I need to hard-code a host + port in that initial React component initialization, which wouldn't work properly

achals (approved):
/lgtm

PR #2701, opened 2022-05-16
fix: Add S3FS dependency (which Dask depends on for S3 files)

adchia:
See https://docs.dask.org/en/stable/install.html for details. This is relevant if you use a FileSource that points to an S3 bucket.

This made it necessary to cap some versions of packages like boto + gcsfs because of conflicting downstream package dependencies with botocore versions.

Fixes #


achals (approved):
Will you have to regenerate all the requirements files?

adchia:
Done

achals (approved):
/lgtm

PR #2709, opened 2022-05-16
fix: Correct feature statuses during feature logging test

pyalex:

Infer correct feature status in feature logging e2e test.

Fixes #

### DIFF
--- a/sdk/python/tests/integration/feature_repos/repo_configuration.py
+++ b/sdk/python/tests/integration/feature_repos/repo_configuration.py
@@ -349,7 +349,7 @@ class Environment:
             worker_id_num = int(parsed_worker_id[0])
         else:
             worker_id_num = 0
-        return 16000 + 100 * worker_id_num + self.id
+        return g6000 + 100 * worker_id_num + self.id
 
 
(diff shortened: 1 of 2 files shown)


sdk/python/tests/integration/feature_repos/repo_configuration.py
```diff
@@ -349,7 +349,7 @@ def get_local_server_port(self) -> int:
             worker_id_num = int(parsed_worker_id[0])
         else:
             worker_id_num = 0
-        return 6000 + 100 * worker_id_num + self.id
+        return 16000 + 100 * worker_id_num + self.id
```
achals, line 352: Should we try to get an available port from the OS directly instead? This seems inherently flaky.

pyalex: Yeah, I thought about that. The problem is that we're running tests in multiprocessing mode and, in addition, this port determination (this function) happens separately from actual port acquiring. So if we will just select some free port here - it doesn't guarantee that it will be actually free on the moment of starting actual server.

felixwang9817 (approved):
/lgtm

PR #2715, opened 2022-05-19
chore: Fix RTD and update from make build-sphinx

adchia:

Fix broken Python documentation, which occurred when setup.py got moved. (See https://readthedocs.org/projects/feast/builds/)

Also adds Postgres Registry Store + moves RequestSource / PushSource as subpages under data source

Also fixes `make html` to use the current environment so it can generate a preview of the html docs properly (respecting pyenv)

Fixes #


felixwang9817 (approved):
/lgtm

PR #2718, opened 2022-05-19
feat: CLI interface for validation of logged features

pyalex:

Validation of logged features can be called via CLI interface:
```
feast validate --feature-service <service_name> --reference <reference_name> <start_ts> <end_ts>
```

This PR also includes storing / retrieving of validation reference to / from the registry.

Fixes #

### DIFF
--- a/protos/feast/core/FeatureService.proto
+++ b/protos/feast/core/FeatureService.proto
@@ -54,7 +54,6 @@ message FeatureServiceMeta {
 
 message LoggingConfig {
   float sample_rate = 1;
-  google.protobuf.Duration partition_interval = 2;
 
   oneof destination {
     FileDestination file_destination = 3;
--- a/protos/feast/core/Registry.proto
+++ b/protos/feast/core/Registry.proto
@@ -30,6 +30,7 @@ import "feast/core/OnDemandFeatureView.proto";
 import "feast/core/RequestFeatureView.proto";
 import "feast/core/DataSource.proto";
 import "feast/core/SavedDataset.proto";
+import "feast/core/ValidationProfile.proto";
 import "google/protobuf/timestamp.proto";
 
 // Next id: 13
@@ -42,6 +43,7 @@ message Registry {
     repeated RequestFeatureView request_feature_views = 9;
     repeated FeatureService feature_services = 7;
     repeated SavedDataset saved_datasets = 11;
+    repeated ValidationReference validation_references = 13;
     Infra infra = 10;
 
     string registry_schema_version = 3; // to support migrations; incremented when schema is changed
--- a/protos/feast/core/Registry.proto
+++ b/protos/feast/core/Registry.proto
@@ -33,7 +33,7 @@ import "feast/core/SavedDataset.proto";
 import "feast/core/ValidationProfile.proto";
 import "google/protobuf/timestamp.proto";
 
-// Next id: 13
+// Next id: 14
 message Registry {
     repeated Entity entities = 1;
     repeated FeatureTable feature_tables = 2;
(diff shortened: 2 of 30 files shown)


protos/feast/core/FeatureService.proto
```diff
@@ -54,7 +54,6 @@ message FeatureServiceMeta {
 
 message LoggingConfig {
   float sample_rate = 1;
-  google.protobuf.Duration partition_interval = 2;
```
achals, line 57: Do we need to reserve #2 or does it it not matter since it's never been used by anyone?

protos/feast/core/Registry.proto
```diff
@@ -30,6 +30,7 @@ import "feast/core/OnDemandFeatureView.proto";
 import "feast/core/RequestFeatureView.proto";
 import "feast/core/DataSource.proto";
 import "feast/core/SavedDataset.proto";
+import "feast/core/ValidationProfile.proto";
 import "google/protobuf/timestamp.proto";
 
 // Next id: 13
```
achals, line 36: Update 13 -> 14

sdk/python/feast/dqm/profilers/ge_profiler.py
```diff
@@ -140,9 +141,12 @@ def analyze_dataset(self, df: pd.DataFrame) -> Profile:
         return GEProfile(expectation_suite=self.user_defined_profiler(dataset))
 
     def to_proto(self):
+        # keep only the code and drop context for now
+        # ToDo (pyalex): include some context, but not all (dill tries to pull too much)
```
achals, line 145: Curious what this means?

sdk/python/feast/saved_dataset.py
```diff
@@ -203,21 +206,106 @@ def to_arrow(self) -> pyarrow.Table:
...
+    name: str
+    dataset_name: str
     profiler: Profiler
 
-    def __init__(self, dataset: SavedDataset, profiler: Profiler):
-        self.dataset = dataset
+    _profile: Optional[Profile] = None
+    _dataset: Optional[SavedDataset] = None
+
+    def __init__(self, name: str, dataset_name: str, profiler: Profiler):
```
achals, line 226: We should add description and tags here, since this is effectively a new top level object in feast?

setup.py
```diff
@@ -64,6 +64,7 @@
     "proto-plus<1.19.7",
     "pyarrow>=4,<7",
     "pydantic>=1,<2",
+    "pygments==2.12.0",
```
achals, line 67: What do we need this for? Should it be unpinned?

protos/feast/core/FeatureService.proto
pyalex: It's never been used.

sdk/python/feast/dqm/profilers/ge_profiler.py
pyalex: So right now for easier deserialization I drop the whole context of user defined function and keep only its code. The opposite of that (what dill does by default) is include all globals and refer the original module, which won't be available at the runtime. But maybe in future somebody would want to use at least function clojure, so we might want to support that.

setup.py
pyalex: this is for coloring the output in CLI

protos/feast/core/Registry.proto
pyalex: Fixed

achals (approved):
/lgtm

setup.py
pyalex: More specifically it highlights the code. Since we printing out some JSON (when validation failed), it makes it more readable.

PR #2722, opened 2022-05-20
chore: Add a .gitpod.yml file 

achals:
Trying out https://www.gitpod.io/ since it has allows a web-based development flow that easy for users to on board on to.

Fixes #


.gitpod.yml
```diff
@@ -0,0 +1,31 @@
+# This configuration file was automatically generated by Gitpod.
```
kevjumba, line 1: should we remove this since we changed the gitpod file.

kevjumba (approved):

.gitpod.yml
achals: lol yea

achals: updated

PR #2735, opened 2022-05-26
fix: Fixing Spark min / max entity df event timestamps range return order

levpick:
… max entity-DF event timestamps in the Spark offline store.

Signed-off-by: Lev Pickovsky <lev.pickovsky@ironsrc.com>

This PR is meant to fix the returned order of elements when calculating the timestamp range of an entity-DF in the _get_entity_df_event_timestamp_range method of the Spark offline store class in case the entity-DF is provided as a string. This method returns a tuple with 2 elements - the min and max timestamps encountered in the entity-DF. Currently, in case the entity-DF is a string, it returns the max as the first element and the min as the second, but the code that uses these values later on seems to expect the min element to be the first one (and it can also be observed that the min is the first one to be returned when the provided entity-DF is a Pandas DF). The issue was discovered following failing real-life tests of the Spark offline store which this fix seems to have resolved.


adchia:
thanks for the contribution!

Would you mind adding a test for this?

kevjumba:
@levpick  are you still working on this?

adchia (approved):

PR #2742, opened 2022-05-27
chore: Collection of small improvements for feature validation

pyalex:

1. Support null-type column in arrow to proto conversion
2. Set value to NaN for not found features when preparing dataset for validation
3. Add more useful information to validation result

Fixes #


achals (approved):
/lgtm

PR #2749, opened 2022-05-31
chore: Install dependencies into a virtualenv for gitpod

achals:

When using prebuilds (which we are), only things in the /workspace/ directory are saved. But, by default gitpod's python environment is system-wide and installs into /home,  which means that all installed dependencies are lost when the workspace actually starts.

Fixes #


pyalex (approved):

adchia (approved):
/lgtm

PR #2753, opened 2022-06-02
Made URL to registry relative, fixes #2731

### LINKED ISSUE #2731: Unable to access data in Feast UI when deployed to remote instance
## Expected Behavior 
Should be able to view registry data when launching UI with `feast ui` on remote instances (like EC2).

## Current Behavior
I’ve tried setting the host to `0.0.0.0`  and the static assets get loaded and can accessed via the public IP. But the requests to the registry ( http://0.0.0.0:8888/registry) fails, so no data shows up.

I've also tried setting the host to the private IP, but the request to `/registry` times out.

## Steps to reproduce
Run `feast ui --host <instance private ip>` in EC2 instance.

### Specifications

- Version:`0.21.2`
- Platform: EC2
- Subsystem:

## Possible Solution
Potential CORS issue that needs to be fixed?

creativedutchmen:
Instead of having the URL to the registry be absolute (default to 0.0.0.0:8888), use a relative path. This means there's no SSL issues when putting the UI behind a SSL proxy, nor will it try to load it from localhost. This change allows the UI to be deployed to a central location.

Fixes #2731

### DIFF
--- a/sdk/python/feast/ui_server.py
+++ b/sdk/python/feast/ui_server.py
@@ -62,7 +62,7 @@ def get_app(
                     "name": "Project",
                     "description": "Test project",
                     "id": project_id,
-                    "registryPath": f"http://{host}:{port}/registry",
+                    "registryPath": "/registry",
                 }
             ]
         }


adchia (approved):
/lgtm

PR #2733, opened 2022-05-25
feat: Add feast repo-upgrade for automated repo upgrades

achals:
Signed-off-by: Achal Shah <achals@gmail.com>

**What this PR does / why we need it**:

This PR introduces a new `feast` command, `feast repo-upgrade`. The intention is for feast to be able to upgrade existing feature repos to the newest API that feast supports, as far as possible.

**Which issue(s) this PR fixes**:

First pass of solving #2629

PR #2754, opened 2022-06-02
fix: Fix bugs in applying stream feature view and retrieving online features 

kevjumba:
Previous pr failed to apply stream feature views correctly and did not initialize an online store correctly with a stream feature view defined. This pr fixes those issues.

Fixes #


protos/feast/core/StreamFeatureView.proto
```diff
@@ -95,4 +96,7 @@ message StreamFeatureViewMeta {
 
     // Time where this Feature View is last updated
     google.protobuf.Timestamp last_updated_timestamp = 2;
+
+    // List of pairs (start_time, end_time) for which this feature view has been materialized.
+    repeated MaterializationInterval materialization_intervals = 3;
```
felixwang9817, line 101: hm do we really need this? I don't think we're planning on materializing SFVs

sdk/python/feast/repo_operations.py
```diff
@@ -133,6 +138,17 @@ def parse_repo(repo_root: Path) -> RepoContents:
                     # Don't add if the push source's batch source is a duplicate of an existing batch source
                     if push_source_dep not in data_sources_set:
                         res.data_sources.append(push_source_dep)
+            elif isinstance(obj, StreamFeatureView):
```
felixwang9817, line 141: I think you also want to add a `and not any((obj is sfv) for sfv in res.stream_feature_views)` check here?

the reason this check exists above is if to handle the case where you define a FV in one file and import it in another file - if you don't deduplicate by `is` equality, the FV will get applied twice

felixwang9817 (commented):
few small points

protos/feast/core/StreamFeatureView.proto
felixwang9817: actually am fine with this for now, ideally will remove it long term

sdk/python/feast/repo_operations.py
kevjumba: fixed.

protos/feast/core/StreamFeatureView.proto
kevjumba: addressed. We don't materialize but since streamfeatureview inherits featureview, will need this param in proto.

felixwang9817 (approved):
/lgtm

protos/feast/core/StreamFeatureView.proto
achals: Technically you can still run batch materialization for a stream feature view right?

sdk/python/feast/feature_store.py
```diff
@@ -1321,9 +1354,14 @@ def write_to_online_store(
...
-            feature_view_name, allow_registry_cache=allow_registry_cache
-        )
+        try:
+            feature_view = self.get_stream_feature_view(
+                feature_view_name, allow_registry_cache=allow_registry_cache
+            )
+        except FeatureViewNotFoundException:
+            feature_view = self.get_feature_view(
+                feature_view_name, allow_registry_cache=allow_registry_cache
+            )
```
achals, line 1364: IMO we should try to build on top of #2733 to migrate these kinds of stray feature views to the right location.

sdk/python/feast/registry.py
```diff
@@ -717,6 +722,30 @@ def get_feature_view(
...
+            name: Name of stream feature view
+            project: Feast project that this stream feature view belongs to
+            allow_cache: Allow returning feature view from the cached registry
+
+        Returns:
+            Returns either the specified feature view, or raises an exception if
+            none is found
+        """
+        registry_proto = self._get_registry_proto(allow_cache=allow_cache)
+        for feature_view_proto in registry_proto.stream_feature_views:
```
achals, line 741: Is there a reason we moved this into a different field in the `Registry` proto? Could we have reused the existing `FeatureView` proto message?

kevjumba: I don't think so, the proto for stream feature view and feature view are different.

sdk/python/feast/feature_store.py
kevjumba: Hm I see, I'll add a todo for this.

PR #2756, opened 2022-06-03
chore: Create a registry base class

achals:

This PR introduces a base class for the methods that the Registry should support, and leaving the other methods in the current Registry class as implementation details.

Fixes #

### DIFF
--- a/sdk/python/feast/registry.py
+++ b/sdk/python/feast/registry.py
@@ -14,6 +14,7 @@
 import abc
 import json
 import logging
+from abc import abstractmethod
 from collections import defaultdict
 from datetime import datetime, timedelta
 from enum import Enum
@@ -148,6 +149,7 @@ def get_registry_store_class_from_scheme(registry_path: str):
 
 class BaseRegistry(abc.ABC):
     # Entity operations
+    @abstractmethod
     def apply_entity(self, entity: Entity, project: str, commit: bool = True):
         """
         Registers a single entity with Feast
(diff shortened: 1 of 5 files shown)


sdk/python/feast/registry.py
```diff
@@ -145,7 +146,413 @@ def get_registry_store_class_from_scheme(registry_path: str):
         return get_registry_store_class_from_type(registry_store_type)
 
 
-class Registry:
+class BaseRegistry(abc.ABC):
+    # Entity operations
+    def apply_entity(self, entity: Entity, project: str, commit: bool = True):
```
felixwang9817, line 151: did you want to make these `@abstractmethods`?

sdk/python/feast/registry.py
```diff
@@ -145,7 +146,413 @@ def get_registry_store_class_from_scheme(registry_path: str):
...
+    ) -> List[RequestFeatureView]:
+        """
+        Retrieve a list of request feature views from the registry
+
+        Args:
+            allow_cache: Allow returning feature views from the cached registry
+            project: Filter feature views based on project name
+
+        Returns:
+            List of feature views
```
kevjumba, line 420: request feature view

kevjumba (approved):
/approve

sdk/python/feast/registry.py
achals: updated, was copy pasta from the Registry class.

PR #2765, opened 2022-06-07
feat: Update stream fcos to have watermark and sliding interval

kevjumba:

Fixes #

### DIFF
--- a/protos/feast/core/Aggregation.proto
+++ b/protos/feast/core/Aggregation.proto
@@ -11,5 +11,5 @@ message Aggregation {
     string column = 1;
     string function = 2;
     google.protobuf.Duration time_window = 3;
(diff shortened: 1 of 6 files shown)


protos/feast/core/Aggregation.proto
```diff
@@ -11,4 +11,5 @@ message Aggregation {
     string column = 1;
     string function = 2;
     google.protobuf.Duration time_window = 3;
+    google.protobuf.Duration sliding_window = 4;
```
felixwang9817, line 14: isn't it "slide / sliding interval" instead of "sliding window"?

felixwang9817 (commented):
lgtm just one nit

felixwang9817 (approved):
/lgtm

PR #2766, opened 2022-06-07
Change online UI schema field to be optional

### LINKED ISSUE #2719: UI fails to run correctly on project with offline Featureviews
## Expected Behavior 
When FeatureViews have `online=False`, running `feast ui` should start the UI and load the project.
 
## Current Behavior
The UI fails to load with the following error:
```
Error Loading Project Configs
There was an error loading the Project Configurations. Please check that feature_store.yaml file is available and well-formed.
```

This looks like it is caused by the registry dump not containing the `online` key for the feature views.

## Steps to reproduce
1. Create a default repo using `feast init`
2. In `example.py` change `online=True` to `online=False`.
3. Run `feast ui`

### Specifications

- Version: 0.21.2
- Platform: macOS
- Subsystem:

## Possible Solution
To fix this, the registry should always contain the `online` key for FeatureViews even if they are set to False.

Perhaps this is related to this quirk of protobufs: https://github.com/protocolbuffers/protobuf/issues/1772 which would mean that the `online` key doesn't come through in the call to `to_proto` here:

https://github.com/feast-dev/feast/blob/f4eed30c7a01b8e68cb6b5ed4f8ed70cdef86462/sdk/python/feast/registry.py#L914

dvanbrug:
Change the `online` field in feastFeatureViews schema to be optional.
This addresses a problem with the UI erroring out when trying to present
offline FeatureViews since protobuf would not pass along the `online`
tag.

Fixes #2719

### DIFF
--- a/ui/src/parsers/feastFeatureViews.ts
+++ b/ui/src/parsers/feastFeatureViews.ts
@@ -39,7 +39,7 @@ const FeastFeatureViewSchema = z.object({
     features: z.array(FeastFeatureColumnSchema),
     ttl: z.string().transform((val) => parseInt(val)),
     batchSource: FeastBatchSourceSchema,
-    online: z.boolean(),
+    online: z.boolean().optional(),
     owner: z.string().optional(),
     tags: z.record(z.string()).optional(),
   }),


achals (approved):

adchia (approved):
/lgtm

PR #2768, opened 2022-06-07
fix: Bugfixes for how registry is loaded

achals:

Fixes #

### DIFF
--- a/sdk/python/feast/registry.py
+++ b/sdk/python/feast/registry.py
@@ -663,6 +663,75 @@ class BaseRegistry(abc.ABC):
     def refresh(self):
         """Refreshes the state of the registry cache by fetching the registry state from the remote registry store."""
 
+    @staticmethod
+    def _message_to_sorted_dict(message: Message) -> Dict[str, Any]:
+        return json.loads(MessageToJson(message, sort_keys=True))
+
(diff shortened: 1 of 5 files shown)


sdk/python/feast/registry.py
```diff
@@ -689,6 +758,16 @@ def get_user_metadata(
     cached_registry_proto_created: Optional[datetime] = None
     cached_registry_proto_ttl: timedelta
 
+    def __new__(
```
felixwang9817, line 761: it's not clear to me why this is necessary - can you explain in the PR description what the bug is / how it's being fixed?

also are there any tests that would catch the bugs?

felixwang9817 (commented):
just one nit

sdk/python/feast/registry.py
achals: https://docs.python.org/3/reference/datamodel.html#object.__new__

Basically, `__new__` is called before `__init__`, and can be used to *change* the object that's being created by a call to a class constructor. So in our case, we are calling `Registry()`, but we want to return a `SqlRegistry` based on the config information. This implementation achieves that.

I can add a test for this specifically.

felixwang9817 (approved):
/lgtm

felixwang9817 (approved):
/lgtm

PR #2782, opened 2022-06-10
fix: Hydrate infra object in the sql registry proto() method 

achals:

Fixes #

### DIFF
--- a/sdk/python/feast/infra/registry_stores/sql.py
+++ b/sdk/python/feast/infra/registry_stores/sql.py
@@ -584,7 +584,6 @@ class SqlRegistry(BaseRegistry):
     def proto(self) -> RegistryProto:
         r = RegistryProto()
         project = ""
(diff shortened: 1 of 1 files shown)


sdk/python/feast/infra/registry_stores/sql.py
```diff
@@ -575,7 +584,7 @@ def get_user_metadata(
     def proto(self) -> RegistryProto:
         r = RegistryProto()
         project = ""
-        # TODO(achal): Support Infra object, and last_updated_timestamp.
+        # TODO(achal): Support last_updated_timestamp.
```
felixwang9817, line 587: is this comment still necessary? looks like you're supporting `last_updated_timestamp` below?

also should it be `last_updated_timestamp` instead of `last_update_timestamp` below?

felixwang9817 (approved):
/lgtm

PR #2792, opened 2022-06-14
chore: Added docs section on env vars in yaml

creativedutchmen:
Because it was unclear to me you could do this, and there was another question in Slack, I've added a section to the docs to explain you can do this:

```yaml
project: my_project
registry: data/registry.db
provider: local
online_store:
    type: redis
    connection_string: ${REDIS_CONNECTION_STRING}
```


creativedutchmen:
I just noticed a previous commit (that I thought was already merged) has found its way into this PR. Please let me know if I should create a new PR with only the latest commit. Thanks.

achals:
Thanks for the contribution @creativedutchmen ! I think the previous commit is okay to pull in since there's no merge conflicts.

achals (approved):
/lgtm

achals:
Actually @creativedutchmen you'll need to signoff your commit - you should be able to follow the instructions at https://github.com/feast-dev/feast/pull/2792/checks?check_run_id=6883873559

creativedutchmen:
I just learned it's possible to set a default like `${ENV_VAR:"default"}` - will update to add that.

achals (approved):
/lgtm

mkhalaf2:
> creativedutchmen

I've tried to set default - but it doesn't work anymore - to my understanding load_repo_config uses os.path.expandvars and the latter does not support default value.

PR #2798, opened 2022-06-15
feat: Enable stream feature view materialization

felixwang9817:
**What this PR does / why we need it**: This PR enables materialization for stream feature views.

Fixes #


sdk/python/feast/inference.py
```diff
@@ -99,6 +99,10 @@ def update_feature_views_with_inferred_features_and_entities(
     other columns except designated timestamp columns are considered to be feature columns. If
     the feature view already has features, feature inference is skipped.
 
+    Note that this inference logic currently does not take transformations into account. For
```
kevjumba, line 102: Probably clarify transformations and aggregations.

sdk/python/tests/integration/online_store/test_universal_online.py
```diff
@@ -490,7 +490,7 @@ def test_online_retrieval(
     )
 
     entity_sample = datasets.orders_df.sample(10)[
-        ["customer_id", "driver_id", "order_id", "event_timestamp"]
+        ["customer_id", "driver_id", "order_id", "origin_id", "event_timestamp"]
```
kevjumba, line 493: is this relevant to this pr or is it just a datasource that we forgot to include in the test.

kevjumba (approved):
/approve

sdk/python/tests/integration/online_store/test_universal_online.py
felixwang9817: this is relevant since we now want to retrieve features from the pushable_location_stats SFV, which requires a new join key (specifically a location_id), so when we sample, we also need to pull a location_id (in this case the column has been renamed to origin_id, so we pull that column)

sdk/python/feast/inference.py
felixwang9817: wdym?

kevjumba: Like clarify that stream feature views doesn't take into account udf or aggregations(transformations is a little ambiguous)

sdk/python/tests/integration/online_store/test_universal_online.py
kevjumba: Ah sounds good.

sdk/python/feast/inference.py
felixwang9817: addressed

kevjumba (commented):
/lgtm

PR #2801, opened 2022-06-16
docs: Add docs for SQL Registry

achals:

Fixes #


felixwang9817 (approved):
/lgtm

docs/tutorials/using-scalable-registry.md
```diff
@@ -0,0 +1,36 @@
...
+offline_store: file
+registry:
+    registry_type: sql
+    path: postgresql://postgres:mysecretpassword@127.0.0.1:55001/feast
+```
+
+Specifically, the registry_type needs to be set to sql in the registry config block. On doing so, the path should refer to the [Database URL](https://docs.sqlalchemy.org/en/14/core/engines.html#database-urls) for the database to be used, as expected by SQLAlchemy. No other additional commands are currently needed to configure this registry.
+
+There are some things to note about how the SQL registry works:
+- Once instantiated, the Registry ensures the tables needed to store data exist, and creates them if they do now.
```
mirayyuce, line 31: Do you mean the tables are created by Feast, if they do not exist? If so, there is a typo in the last word

achals: good catch, fixed

mirayyuce (approved):

PR #2807, opened 2022-06-16
feat: Enable stream feature view materialization

felixwang9817:
**What this PR does / why we need it**: This PR does the same as #2798, but rewrites the SFV online retrieval logic in a separate test to avoid flakiness.

Fixes #


kevjumba (commented):
/lgtm

PR #2808, opened 2022-06-17
feat: Add file write_to_offline_store functionality

kevjumba:

The interface for write_to_offline_store is changing(i.e the dataframe passed in is an arrow table instead of a pandas dataframe. The reason for this is because is we have to convet the pandas dataframe to pyarrow table for write anyways but some functions such as `_run_field_mapping` use pyarrow table so it makes more sense to not rewrite code and just convert the pyarrow table earlier. 

Fixes #


sdk/python/feast/infra/passthrough_provider.py
```diff
@@ -104,10 +104,14 @@ def offline_write_batch(
         self,
         config: RepoConfig,
         table: FeatureView,
-        data: pd.DataFrame,
+        data: pa.Table,
         progress: Optional[Callable[[int], Any]],
     ) -> None:
         set_usage_attribute("provider", self.__class__.__name__)
+
+        if "created" not in data.column_names:
```
felixwang9817, line 112: why is this necessary? I think most feature views don't have a created ts column, and it also might not be named `created`

sdk/python/tests/integration/offline_store/test_offline_push.py
```diff
@@ -0,0 +1,226 @@
+import random
```
felixwang9817, line 1: slight preference for naming this file `test_offline_write.py` as this behavior is independent of the push API

sdk/python/tests/integration/offline_store/test_offline_push.py
```diff
@@ -0,0 +1,226 @@
...
+import pytest
+
+from feast import FeatureView, Field
+from feast.types import Float32, Int32
+from tests.integration.feature_repos.universal.entities import driver
+
+
+@pytest.mark.integration
+@pytest.mark.universal_online_stores
+def test_writing_incorrect_order_fails(environment, universal_data_sources):
```
felixwang9817, line 15: can you explain (and maybe add this explanation in a docstring for this test) why exactly we expect it to fail? the name suggests an incorrect order, but from inspecting the underlying data source it looks like the df might not just be out of order, but it's also missing some columns?

sdk/python/tests/integration/offline_store/test_offline_push.py
```diff
@@ -0,0 +1,226 @@
...
+    )
+    with pytest.raises(ValueError):
+        store.write_to_offline_store(
+            driver_stats.name, expected_df, allow_registry_cache=False
+        )
+
+
+@pytest.mark.integration
+@pytest.mark.universal_online_stores
+def test_writing_incorrect_schema_fails(environment, universal_data_sources):
```
felixwang9817, line 63: is this a copy of the above test?

sdk/python/tests/integration/offline_store/test_offline_push.py
```diff
@@ -0,0 +1,226 @@
...
+    df = store.get_historical_features(
+        entity_df=entity_df,
+        features=["driver_stats:conv_rate", "driver_stats:avg_daily_trips"],
+        full_feature_names=False,
+    ).to_df()
+
+    assert df["conv_rate"].isnull().all()
+    assert df["avg_daily_trips"].isnull().all()
+
+    first_df = pd.DataFrame.from_dict(
```
felixwang9817, line 146: might leave a comment here indicating that these columns are arranged in exactly the same order as the underlying DS (whose schema can be found in `driver_test_data.py`)

sdk/python/tests/integration/offline_store/test_offline_push.py
```diff
@@ -0,0 +1,226 @@
...
+    )
+    with pytest.raises(ValueError):
+        store.write_to_offline_store(
+            driver_stats.name, expected_df, allow_registry_cache=False
+        )
+
+
+@pytest.mark.integration
+@pytest.mark.universal_online_stores
+def test_writing_consecutively_to_offline_store(environment, universal_data_sources):
```
felixwang9817, line 111: nice this is a great test

sdk/python/feast/infra/offline_stores/offline_store.py
```diff
@@ -275,7 +275,7 @@ def write_logged_features(
     def offline_write_batch(
         config: RepoConfig,
         table: FeatureView,
```
achals, line 277: nit this shouldn't be called table it should be `feature_view` or something.

sdk/python/feast/infra/provider.py
```diff
@@ -127,7 +127,7 @@ def ingest_df(
         pass
 
     def ingest_df_to_offline_store(
-        self, feature_view: FeatureView, df: pd.DataFrame,
+        self, feature_view: FeatureView, df: pyarrow.Table,
```
achals, line 130: why is this interface changing? can you add that to the PR description?

kevjumba: Sure left some comments

sdk/python/feast/infra/offline_stores/offline_store.py
kevjumba: i was adhering to how we defined in online write batch but ill change it.

PR #2810, opened 2022-06-17
ci: Stop running tests for python 3.7

achals:

Python 3.7  is approaching EoL, and isn't used by a lot of feast users. It's also not a great experience on M1 Macs per @kevjumba. Let's remove it to speed up CI time. We still support 3.7 as the minimum python version, and don't prevent users installing feast on py 3.7, just that we should run tests on newer versions.

Fixes #


felixwang9817 (approved):
/lgtm

Issue #2803, opened 2022-06-16
No support for null UnixTimestamp


aapope:
## Expected Behavior 

This script should return online features, with one null in `last_purchased_date`
```python
from feast import Entity, FeatureView
from feast.infra.offline_stores.file_source import FileSource
from feast.repo_config import RegistryConfig, RepoConfig
from feast import FeatureStore
from datetime import datetime
from feast.types import Int32, UnixTimestamp
from feast import Field

import pandas as pd


# create dataset
pd.DataFrame([
    {"user_id": 1, "event_timestamp": datetime(2022, 5, 1), "created": datetime(2022, 5, 1), "purchases": 3, "last_purchase_date": datetime(2022, 4, 23, 13, 4, 1)},
    {"user_id": 2, "event_timestamp": datetime(2022, 5, 2), "created": datetime(2022, 5, 2), "purchases": 1, "last_purchase_date": datetime(2022, 2, 1, 11, 4, 1)},
    {"user_id": 3, "event_timestamp": datetime(2022, 5, 2), "created": datetime(2022, 5, 2), "purchases": 0, "last_purchase_date": None},    
]).to_parquet('user_stats.parquet')


user = Entity(name="user_id", description="user id")

user_stats_view = FeatureView(
    name="user_stats",
    entities=[user],
    source=FileSource(
        path="user_stats.parquet",
        timestamp_field="event_timestamp",
        created_timestamp_column="created",
    ),
    schema=[
        Field(name="purchases", dtype=Int32),
        Field(name="last_purchase_date", dtype=UnixTimestamp),
    ]
)

online_store_path = 'online_store.db'
registry_path = 'registry.db'

repo = RepoConfig(
    registry="registry.db",
    project='feature_store',
    provider="local",
    offline_store="file",
    use_ssl=True, 
    is_secure=True,
    validate=True,
)

fs = FeatureStore(config=repo)

fs.apply([user, user_stats_view])

fs.materialize_incremental(end_date=datetime.utcnow())


entity_rows = [{"user_id": i} for i in range(1, 4)]


feature_df = fs.get_online_features(
    features=[
        "user_stats:purchases",
        "user_stats:last_purchase_date",
    ],
    entity_rows=entity_rows
).to_df()
print(feature_df)
```

## Current Behavior

```
Materializing 1 feature views to 2022-06-16 20:30:41-06:00 into the sqlite online store.

Since the ttl is 0 for feature view user_stats, the start date will be set to 1 year before the current time.
[...]

PR #2814, opened 2022-06-17
fix: Conversion of null timestamp from proto to python

aapope:
Signed-off-by: Andrew Pope <apope@nursefly.com>

Currently, the python to proto function converts a null timestamp to the value -9223372036854775808, which then fails when converting back to python type. This fixes the conversion of null UnixTimestamp values from proto to python by checking for that value and converting it to null.

Fixes # https://github.com/feast-dev/feast/issues/2803


achals:
Thanks for the PR @aapope! I think PR title needs updating but otherwise looks good.

achals (approved):
/lgtm

PR #2819, opened 2022-06-21
feat: Push to Redshift batch source offline store directly

kevjumba:

Fixes #


sdk/python/tests/integration/offline_store/test_offline_write.py
```diff
@@ -107,7 +109,8 @@ def test_writing_incorrect_schema_fails(environment, universal_data_sources):
 
 
 @pytest.mark.integration
-@pytest.mark.universal_online_stores
+@pytest.mark.universal_offline_stores(only=["file", "redshift"])
+@pytest.mark.universal_online_stores(only=["sqlite"])
```
felixwang9817, line 113: hm in this case you probably don't need to specify `universal_online_stores` then, since the default will just be `sqlite`?

sdk/python/feast/feature_store.py
```diff
@@ -1423,6 +1423,8 @@ def _write_to_offline_store(
             feature_view = self.get_feature_view(
                 feature_view_name, allow_registry_cache=allow_registry_cache
             )
+        df.reset_index(drop=True)
```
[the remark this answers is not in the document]
kevjumba: It just drops the index columns so that the dataframe is easier to convert and doesn't have random collisions.

sdk/python/tests/integration/offline_store/test_offline_write.py
kevjumba: the integration tests default to redis

felixwang9817 (approved):
/lgtm

PR #2820, opened 2022-06-21
feat: Updated Trino class to support custom Connection object

zyxue:

It updates Trino class to support more customized authorized Connection object. E.g. I have a conn that's authorized with [http-headers](https://github.com/trinodb/trino-python-client/blob/771eec36b658c536d8f5f6bb1710fca6b766c390/trino/dbapi.py#L102).

Just put here for discussion, an alternative way of supporting `conn` without removing `_get_curosr(self)` method can be like

```diff
diff --git a/sdk/python/feast/infra/offline_stores/contrib/trino_offline_store/trino_queries.py b/sdk/python/feast/infra/offline_stores/contrib/trino_offline_store/trino_queries.py
index 1d4b5881..159a4e83 100644
--- a/sdk/python/feast/infra/offline_stores/contrib/trino_offline_store/trino_queries.py
+++ b/sdk/python/feast/infra/offline_stores/contrib/trino_offline_store/trino_queries.py
@@ -11,7 +11,7 @@ import numpy as np
 import pandas as pd
 import pyarrow as pa
 import trino
-from trino.dbapi import Cursor
+from trino.dbapi import Connection, Cursor
 from trino.exceptions import TrinoQueryError
 
 from feast.infra.offline_stores.contrib.trino_offline_store.trino_type_map import (
@@ -36,6 +36,7 @@ class Trino:
         catalog: Optional[str] = None,
         auth: Optional[Any] = None,
         http_scheme: Optional[str] = None,
+        conn: Optional[Connection] = None,
     ):
         self.host = host or os.getenv("TRINO_HOST")
         self.port = port or os.getenv("TRINO_PORT")
@@ -43,6 +44,7 @@ class Trino:
         self.catalog = catalog or os.getenv("TRINO_CATALOG")
         self.auth = auth or os.getenv("TRINO_AUTH")
         self.http_scheme = http_scheme or os.getenv("TRINO_HTTP_SCHEME")
+        self._conn = conn
         self._cursor: Optional[Cursor] = None
 
         if self.host is None:
@@ -56,15 +58,17 @@ class Trino:
 
     def _get_cursor(self) -> Cursor:
         if self._cursor is None:
-            self._cursor = trino.dbapi.connect(
-                host=self.host,
-                port=self.port,
-                user=self.user,
-                catalog=self.catalog,
-                auth=self.auth,
-                http_scheme=self.http_scheme,
-            ).cursor()
-
+            if self._conn is None:
+                self._cursor = trino.dbapi.connect(
+                    host=self.host,
+                    port=self.port,
+                    user=self.user,
+                    catalog=self.catalog,
+                    auth=self.auth,
+                    http_scheme=self.http_scheme,
+                ).cursor()
+            else:
+                self._cursor = self._conn.cursor()
         return self._cursor
 
     def create_query(self, query_text: str) -> Query:
```

Fixes # NA

### DIFF
--- a/sdk/python/feast/infra/offline_stores/contrib/trino_offline_store/trino_queries.py
+++ b/sdk/python/feast/infra/offline_stores/contrib/trino_offline_store/trino_queries.py
@@ -5,13 +5,14 @@ import os
 import signal
 from dataclasses import dataclass
 from enum import Enum
+from optparse import Option
 from typing import Any, Dict, List, Optional
 
 import numpy as np
 import pandas as pd
 import pyarrow as pa
 import trino
-from trino.dbapi import Cursor
+from trino.dbapi import Cursor, Connection
 from trino.exceptions import TrinoQueryError
 
 from feast.infra.offline_stores.contrib.trino_offline_store.trino_type_map import (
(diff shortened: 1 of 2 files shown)


zyxue:
hmmm.. this still doesn't quite solve the problem as it's unclear how user would control how to call the Trino class

zyxue:
> @zyxue have you seen how snowflake does this? I think it has some logic to get a dictionary of arguments from the feature_store.yaml and pass it into the method that creates the connection object.
> 
> https://github.com/feast-dev/feast/blob/3dd779f4038a757bcf0b935626605ee1e7f04bf6/sdk/python/feast/infra/utils/snowflake_utils.py#L46

Thank you for comment, @achals . the thing is that the `Connection` object in trino https://github.com/trinodb/trino-python-client/blob/master/trino/dbapi.py#L93 supports many args, and not all from can be specified in yaml, e.g. `auth` is supposed to be a class object. Will the snowflake/`feature_store.yaml`'s way still apply?

achals:
@zyxue Good point. Creating custom objects would be possible (by passing in nested arguments in the `feature_store.yaml`, and configuring how the nested objects (like auth) would be created in some kind of `get_trino_conn` method that parsed this config information. 

An alternative approach would requiring refactoring how the offline store is instantated to allow for some custom logic in the creation of the offline store connection.. Which will require some more thought and probably an RFC.

PR #2835, opened 2022-06-22
feat: Support retrieval from multiple feature views with different join keys

yongheng:

Currently Java Feature Server doesn't support retrieval from multiple feature views with different join keys. For each gPRC request, `OnlineServingServiceV2` calls `OnlineRetriever` once and only once. In this call the former sends all join keys in the original request to the latter, and the latter simply sorts and concatenates all join keys to make a Redis key.

This PR supports retrieval from multiple feature views with different join keys. For each gPRC request, it groups feature references by join keys and for each group it makes a call to `OnlineRetriever`.

Fixes #

### DIFF
--- a/java/serving/src/main/java/feast/serving/service/OnlineServingServiceV2.java
+++ b/java/serving/src/main/java/feast/serving/service/OnlineServingServiceV2.java
@@ -37,7 +37,6 @@ import feast.storage.api.retriever.OnlineRetrieverV2;
 import io.opentracing.Span;
 import io.opentracing.Tracer;
 import java.util.*;
-import java.util.function.Function;
 import java.util.stream.Collectors;
 import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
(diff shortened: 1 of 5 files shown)


java/serving/src/main/java/feast/serving/service/OnlineServingServiceV2.java
```diff
@@ -255,6 +246,84 @@ private List<Map<String, ValueProto.Value>> getEntityRows(
...
+          new ArrayList<>(featureReferences.size());
+      for (int j = 0; j < featureReferences.size(); j++) {
+        featuresPerEntity.add(null);
+      }
+      features.add(featuresPerEntity);
+    }
+
+    // Group feature references by join keys.
+    Map<String, List<FeatureReferenceV2>> groupNameToFeatureReferencesMap =
+        featureReferences.stream()
```
pyalex, line 282: To speed up this part we may want to extract distinct feature views from all feature references. And then group feature views instead.

yongheng: IIUC grouping by join keys results in the same or less groups (therefore same or more efficient) than grouping by feature view. The is because multiple feature views can have the same join keys. In [L286](https://github.com/feast-dev/feast/pull/2835/files#diff-ff604cedff2831dc1db08b017aab8dce07a91e20be23dd26787e468da8a1e845R286), `this.registryRepository.getEntitiesList(featureReference)` internally gets feature view spec first, then gets entity names of the feature view spec, then we find join keys for the entity names.

Actually, I grouped by feature view in the beginning. Then I changed to group by join keys in the [second commit](https://github.com/feast-dev/feast/pull/2835/commits/2c9f702756a4313218d78ea51007e753dae12170) of this PR, as an optimization.

yongheng:
> Hi @yongheng , can you please add integration test for this flow?

@pyalex Can you show me where are the current integration tests? Then I can add to there.

kevjumba:
> > Hi @yongheng , can you please add integration test for this flow?
> 
> @pyalex Can you show me where are the current integration tests? Then I can add to there.

Hey @yongheng the integration tests should be any tests that have a tag for @pytest.mark.integration. For this particular test, just take a look at test_feature_views.py

achals:
@yongheng You can find the java integration tests here: https://github.com/feast-dev/feast/tree/master/java/serving/src/test/java/feast/serving/it

yongheng:
@achals @pyalex Integration test has been added and CI is green. Please take a look.

PR #2839, opened 2022-06-22
fix: Prefer installing gopy from feast's fork as opposed to upstream

achals:

In `go.mod`, we're replacing `go-python/gopy` with `feast-dev/gopy`, but for whatever reason we don't do the same thing in `setup.py` or `Makefile`. This results in us using a different version of gopy from the one that will actually work (i.e. our fork). 

This was made worse by the face the the upstream repo landed a bunch of commits and the new version of that project totally borked our build process.

I *think* this should fix the release pipeline (https://github.com/feast-dev/feast/runs/6993658741?check_suite_focus=true) as well as the flaky test (#2830)

Fixes #

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,8 @@ module github.com/feast-dev/feast
 
 go 1.17
 
+replace github.com/go-python/gopy v0.4.0 => github.com/feast-dev/gopy v0.4.1-0.20220429180328-4257ac71a4d0
+
 require (
 	github.com/apache/arrow/go/v8 v8.0.0-20220408212425-58fe60f59289
 	github.com/ghodss/yaml v1.0.0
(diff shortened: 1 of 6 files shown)


kevjumba (approved):
/lgtm

sdk/python/tests/utils/cli_utils.py
```diff
@@ -76,6 +76,9 @@ def local_repo(self, example_repo_py: str, offline_store: str):
                 path: {data_path / "online_store.db"}
             offline_store:
                 type: {offline_store}
+            flags:
```
felixwang9817, line 79: any particular reason for this change? I think we usually accomplish this with `IS_TEST=True` right?

setup.py
```diff
@@ -423,8 +424,8 @@ def build_extension(self, ext: Extension):
         )
 
         destination = os.path.dirname(os.path.abspath(self.get_ext_fullpath(ext.name)))
-        subprocess.check_call(["go", "install", "golang.org/x/tools/cmd/goimports"])
-        subprocess.check_call(["go", "install", "github.com/go-python/gopy"])
+        subprocess.check_call(["go", "mod", "tidy"], env={"PATH": bin_path, **go_env})
+        # subprocess.check_call(["go", "install", "github.com/go-python/gopy@v0.4.0"])
```
kevjumba, line 428: Is this needed?

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/feast-dev/feast
 
 go 1.17
 
+replace github.com/go-python/gopy v0.4.0 => github.com/feast-dev/gopy v0.4.1-0.20220429180328-4257ac71a4d0
```
adchia, line 5: was the main change to pin it at v0.40.0? was the re-ordering of “replace” needed in go.mod?

setup.py
adchia, line 428: remove this line?

go.mod
achals: @pyalex is right, the line change has no effect.

achals: Pinning was the first main change. The other main change was in setup.py to pass in the go env to go commands. Without that, the go subprocess would behave unexpectedly since the go config would be different since the process environment isn't inherited by default.

setup.py
achals: Done

kevjumba (approved):
/lgtm

judahrand:
@achals Is this the long term solution or are we planning to move back to upstream gopy? Is there some fundamental reason to use a fork?

achals:
@judahrand I would love to upstream these changes and not maintain a fork. @pyalex do you think this is feasible?

PR #2815, opened 2022-06-17
fix: Enable faulthandler and disable flaky tests

achals:
This is going to only be enabled if the embedded go fs component is instantiated.

Signed-off-by: Achal Shah <achals@gmail.com>

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

PR #2841, opened 2022-06-23
fix: Renable tests after fixing gopy installation issue that was blocking tests.

kevjumba:

Undoes part of #2815 and does so because of #2839. 

Fixes #


achals:
We should be using `pytest.mark.skip` instead of commenting out tests.

Also, can you update the PR title and description to provide more context? And link to the previous PR?

PR #2842, opened 2022-06-23
fix: Explicitly translate errors when instantiating the go fs 

achals:
Also convert bool params when calling the transformation callback, since typeguard doesn't like type-unsafe funny business.

Fixes #

### DIFF
--- a/sdk/python/tests/integration/online_store/test_universal_online.py
+++ b/sdk/python/tests/integration/online_store/test_universal_online.py
@@ -443,7 +443,7 @@ def test_online_retrieval_with_event_timestamps(
 
 @pytest.mark.integration
 @pytest.mark.universal_online_stores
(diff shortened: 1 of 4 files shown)


sdk/python/tests/integration/online_store/test_universal_online.py
```diff
@@ -443,7 +443,7 @@ def test_online_retrieval_with_event_timestamps(
 
 @pytest.mark.integration
 @pytest.mark.universal_online_stores
-# @pytest.mark.goserver Disabling because the go fs tests are flaking in CI. TODO(achals): uncomment after fixed.
+@pytest.mark.goserver
```
felixwang9817, line 446: not sure if your changes here only fix some of the tests, but there are a few more tests that have this comment

kevjumba (approved):
/lgtm

PR #2851, opened 2022-06-23
feat: Add custom JSON table tab w/ formatting

kindalime:
Hello! This is a small commit that adds a custom tab to the custom tab folder. The tab takes in JSON data from a specified JSON file and displays it in a table that lists object names on the left and their values on the right. The values are formatted to use EUI's <EuiCode> tag with an added LineHeight CSS property (through React) for easy viewing.

This tab has been currently attached to the FeatureView pages, but is not active for any of the other pages; it is my hope that this can serve as a sort of template for formatted data tables.


adchia (approved):
/lgtm

PR #2856, opened 2022-06-24
chore: Upgrade version of arrow/go

achals:

We were using a pre-release version but v8 of arrow/go came out in May

Fixes #


felixwang9817 (approved):
/lgtm

PR #2858, opened 2022-06-24
ci: Disable Go for Python 3.10 release

felixwang9817:

Fixes #


achals (commented):
Should we try to move this to cibuildwheel at the same time? Or do we expect this to "just work" after disabling go?

kevjumba:
> Should we try to move this to cibuildwheel at the same time? Or do we expect this to "just work" after disabling go?

@achals from our testing I think it should just work, since Felix’s build on his fork worked when gopy was disabled

achals (approved):
/lgtm

PR #2877, opened 2022-06-28
fix: Stop running go mod tidy in setup.py

achals:

This fixes a problem similar to https://github.com/pypa/cibuildwheel/issues/189#issuecomment-549933912.

Running `go mod tidy` was updating `go.mod` and `go.sum`, which was causeing cibuildwheel to produce dirty builds. This comamnd isn't really even needed, so nixing it.

Fixes #


felixwang9817 (approved):
/lgtm

PR #2883, opened 2022-06-29
ci: Update go.mod and go.sum but git restore before building

achals:

Fixes #


.github/workflows/build_wheels.yml
```diff
@@ -76,6 +76,7 @@ jobs:
             make install-go-proto-dependencies
             make install-go-ci-dependencies
             git status
+            git restore go.*
```
kevjumba, line 79: can you add a comment so ppl know why we're doing this.

kevjumba (approved):
/lgtm

PR #2893, opened 2022-06-29
chore: Verification workflow in build wheels that ensures the wheels built are not development and the correct versions

kevjumba:
Hit a bunch of issues in 0.22.0 release, one of which is the fact that we did not catch that the build caused dirty git state that caused the wheels that were built to be dev wheels. This check in `verify-python-wheel` fixes that.

Fixes #


.github/workflows/build_wheels.yml
```diff
@@ -190,6 +197,22 @@ jobs:
...
         run: brew install coreutils
+      # Validate that the feast version installed is not development and is the correct version of the tag we ran it off of.
+      - name: Validate Feast Version
+        run: |
+          VERSION_REGEX='[0-9]+\.[0-9]+\.[0-9]+'
+          OUTPUT_REGEX='^Feast SDK Version: "$VERSION_REGEX"$'
+          VERSION_OUTPUT=$(feast version)
+          VERSION=$(echo $VERSION_OUTPUT | grep -oE "$VERSION_REGEX")
+          OUTPUT=$(echo $VERSION_OUTPUT | grep -E "$REGEX")
+          if [ "VERSION_WITHOUT_PREFIX" = "erification_workflow" ]; then
```
achals, line 208: When would this be the case?

achals (approved):
/ok-to-test
/lgtm

PR #2894, opened 2022-06-29
fix: Fix grpc and update protobuf

### LINKED ISSUE #2484: protobuf pinned to <3.20.0
We pinned the version of protobuf to `protobuf>=3.10,<3.20` because we patch some functionality when converting json <> proto in the AWS Lambda feature server (https://github.com/feast-dev/feast/blob/master/sdk/python/feast/proto_json.py#L18).

protobuf 3.20.0 changes the API we patch - so the version bump needs to happen along with an update to the patched methods.

#2480 has more context.

### LINKED ISSUE #2882: Feast 0.21.3 wheels do not build with Python 3.10 on MacOS M1X
Screen Shot 2022-06-29 at 10 27 22 AM
## Current Behavior
The wheels do not build because grpc are built with universal wheels when they are not suitable for m1. The problem is most likely due to this bug: https://github.com/grpc/grpc/issues/28387

## Steps to reproduce
Create a new virtual environment with python 3.10.3 and install feast on MacOS M1.

### Specifications

- Version: 0.21.3
- Platform: Mac M1X OSX 12.2.1
- Subsystem: Python 3.10.3

## Possible Solution

The solution should be to update the grpc versions to latest to force m1's to build from source.

kevjumba:

Fixes #2882
Fixes #2484


.github/workflows/build_wheels.yml
```diff
@@ -1,7 +1,11 @@
 name: build_wheels
 
 # Call this workflow from other workflows in the repository by specifying "uses: ./.github/workflows/build_wheels.yml"
-on: [workflow_dispatch, workflow_call]
+on:
```
achals, line 4: This should be a different PR

.github/workflows/build_wheels.yml
```diff
@@ -134,7 +138,7 @@ jobs:
 
   verify-python-wheels:
     runs-on: ${{ matrix.os }}
-    needs: [build-python-wheel, build-source-distribution]
+    needs: [build-python-wheel, build-source-distribution, get-version]
```
achals, line 141: Why is this change needed?

kevjumba: we need to give the job access the release version which means we must add it to needs

.github/workflows/build_wheels.yml
```diff
@@ -76,10 +80,10 @@ jobs:
             make install-protoc-dependencies
             make install-go-proto-dependencies
             make install-go-ci-dependencies
-            git status
-            git restore go.mod go.sum
+          # git status
```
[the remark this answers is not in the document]
kevjumba: Oh Let me change that.

sdk/python/requirements/py3.7-requirements.txt
```diff
@@ -0,0 +1,202 @@
+#
```
achals, line 1: Why are we adding this back?

sdk/python/requirements/py3.7-ci-requirements.txt
```diff
@@ -0,0 +1,806 @@
+#
```
achals, line 1: Why are we adding this back?

setup.py
```diff
@@ -52,15 +52,15 @@
...
+    "grpcio>=1.47.0,<2",
+    "grpcio-reflection>=1.47.0,<2",
     "Jinja2>=2,<4",
     "jsonschema",
     "mmh3",
     "numpy<1.22,<2",
     "pandas>=1,<2",
     "pandavro==1.5.*",
-    "protobuf>=3.10,<3.20",
+    "protobuf>=3.10,<4",
```
achals, line 63:
```suggestion
    "protobuf>3.20,<4",
```

sdk/python/requirements/py3.7-requirements.txt
kevjumba: ah lol, I didnt realize we had already removed 3.7 requirements.

setup.py
kevjumba: fixed,

sdk/python/requirements/py3.7-ci-requirements.txt
kevjumba: same as above

PR #2899, opened 2022-07-01
chore: Pytest scoping

felixwang9817:
**What this PR does / why we need it**: This PR changes the scope of several pytest fixtures, including `environment` to function-scoped instead of session-scoped. This will ensure that each test environment is separate. It also adds a test to confirm that each test environment is separate.

Fixes #

### DIFF
--- a/sdk/python/tests/unit/test_registry_conflict.py
+++ b/sdk/python/tests/unit/test_registry_conflict.py
@@ -1,0 +1,22 @@
+import pytest
+
+from feast.entity import Entity
+
+@pytest.mark.integration
+def test_apply_first_entity(environment):
+    entity = Entity(name="first")
+    fs = environment.feature_store
+    fs.apply([entity])
+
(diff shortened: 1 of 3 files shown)


kevjumba:
look sgood btw

sdk/python/tests/unit/test_registry_conflict.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+
+from feast.entity import Entity
+
+
+@pytest.mark.integration
+def test_apply_first_entity(environment):
```
achals, line 7: If the Scopes have been changed i don't see a huge point to these tests - even if there's a race condition in the future it seems like the repro for that will need it's own test. Wdyt?

felixwang9817: yeah this is fair; I'll remove this in some future PR when I clean up test suite

PR #2903, opened 2022-07-02
fix: Fix snowflake offline testing

sfc-gh-madkins:

**What this PR does / why we need it**: Bug fix

Fixes #


achals (approved):
/ok-to-test
/lgtm

Oops

adchia:
A quick followup nit btw. In the future, please make sure the commit also has the "fix: " prefix since that's what the release processes uses!

adchia:
(this also maybe shouldn't be a fix since that registers as a user facing "bug fix" in the changelog)

PR #2904, opened 2022-07-02
fix: Change the feature store plan method to public modifier

### LINKED ISSUE #2895: FeatureStore._plan should be a public method

https://github.com/feast-dev/feast/blob/master/sdk/python/feast/feature_store.py#L639

Ideally, all FeatureStore methods that back a CLI operation should have a public facing programatic API. In this case, `feast plan` users `FeatureStore._plan` -> which signifies that it should be treated as a private method.

The change would be as simple as refactoring it to `FeatureStore.plan`.

breno-costa:
Signed-off-by: Breno Costa <brenocosta0901@gmail.com>

According to issue #2895 , the private method `FeatureStore._plan` should be refactored to the public method `FeatureStore.plan` because methods that back a CLI operation should have a public facing programatic API.

Fixes #2895

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -511,8 +511,8 @@ class FeatureStore:
         return _feature_refs
 
     def _should_use_plan(self):
-        """Returns True if _plan and _apply_diffs should be used, False otherwise."""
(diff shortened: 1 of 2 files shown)


achals (commented):
/ok-to-test

achals (approved):

PR #2916, opened 2022-07-06
feat: Add to_remote_storage method to RetrievalJob

achals:
Signed-off-by: Achal Shah <achals@gmail.com>

**What this PR does / why we need it**:

Follow up to #2901 , as defined in [RFC-037](https://docs.google.com/document/d/1J7XdwwgQ9dY_uoV9zkRVGQjK9Sy43WISEW6D5V9qzGo/edit#)

**Which issue(s) this PR fixes**:

Fixes #

PR #2901, opened 2022-07-01
feat: Add interfaces for batch materialization engine

achals:
Signed-off-by: Achal Shah <achals@gmail.com>

**What this PR does / why we need it**:

This is the first pass implementation of the proposal in [RFC-037](https://docs.google.com/document/d/1J7XdwwgQ9dY_uoV9zkRVGQjK9Sy43WISEW6D5V9qzGo/edit#). It's purely interfaces and refactoring without any new logic. The engine is also currently hard coded to the Local engine, which is the same materialization logic as in use today. In a subequent PR, this will be made pluggable.

**Which issue(s) this PR fixes**:

Fixes #

PR #2918, opened 2022-07-06
chore: Implement to_remote_storage for supported offline stores

achals:

Following up to #2916 and #2901 by implementing the to_remote_storage methods for Redshift, Snowflake and BigQuery. 

The implementation utilizes the native export/unload/copy into functionality present in each of the individual DWHs. There's more configuration possible (such as max file size) but I'm not going to add it into the config just yet.

In fact, a related thought I had is that it would be good to have a way to pass in configuration information (somehow) without needing to make code changes into the config classes.

Fixes #


sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -478,6 +490,39 @@ def persist(self, storage: SavedDatasetStorage):
     def metadata(self) -> Optional[RetrievalMetadata]:
         return self._metadata
 
+    def supports_remote_storage_export(self) -> bool:
+        return True
+
+    def to_remote_storage(self) -> List[str]:
+        if self.config.offline_store.gcs_staging_location or not self._gcs_path:
```
felixwang9817, line 497: shouldn't this check just be `if not self._gcs_path`, since that is initialized in `__init__` using `self.config.offline_store.gcs_staging_locatio`

sdk/python/feast/infra/offline_stores/snowflake.py
```diff
@@ -90,6 +91,10 @@ class SnowflakeOfflineStoreConfig(FeastConfigBaseModel):
     schema_: Optional[str] = Field(None, alias="schema")
     """ Snowflake schema name """
 
+    storage_integration_name: Optional[str] = None
```
felixwang9817, line 94: docstring?

sdk/python/feast/infra/offline_stores/snowflake.py
```diff
@@ -453,6 +463,41 @@ def persist(self, storage: SavedDatasetStorage):
     def metadata(self) -> Optional[RetrievalMetadata]:
         return self._metadata
 
+    def supports_remote_storage_export(self) -> bool:
```
felixwang9817, line 466: this is inconsistent with BQ - does this method represent "whether this offline store supports remote storage export" (in which case it should just be `True`) or "whether remote storage export will work for this instance" (in which case it should be this kind of check)

sdk/python/tests/integration/offline_store/test_universal_historical_retrieval.py
```diff
@@ -370,6 +370,9 @@ def test_historical_features(environment, universal_data_sources, full_feature_n
         full_feature_names=full_feature_names,
     )
 
+    if job_from_df.supports_remote_storage_export:
```
felixwang9817, line 373: should we be doing a more thorough test of this behavior? e.g. after unloading, perhaps retrieving and combining to ensure that the data is the same?

not sure how much info we get from these two lines

sdk/python/feast/infra/offline_stores/snowflake.py
achals: yea

sdk/python/feast/infra/offline_stores/snowflake.py
achals: I think this should be the latter. Updated.

sdk/python/tests/integration/offline_store/test_universal_historical_retrieval.py
achals: I'm gonna do that in a separate PR to unblock release

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -478,6 +490,43 @@ def persist(self, storage: SavedDatasetStorage):
     def metadata(self) -> Optional[RetrievalMetadata]:
         return self._metadata
 
+    def supports_remote_storage_export(self) -> bool:
+        return self._gcs_path
```
felixwang9817, line 494: does this actually return a `bool`? should this be `return self._gcs_path != None` or somethign?

felixwang9817 (approved):
/lgtm

felixwang9817 (approved):
/lgtm

PR #2924, opened 2022-07-08
chore: Update Feast UI dependency to reflect recent changes

adchia:

I pushed a new NPM release for the UI model to fix recent issues. This updates Feast to use that new npm module (though strictly speaking, this isn't necessary because "^" includes newer patch releases

Fixes #


achals (approved):
/lgtm

PR #2927, opened 2022-07-08
ci: Fixing local integration tests, defaulting to test containers

adchia:
This changes the local integration test flow to use the test containers. 
- It also extends the timeout for test containers because I was seeing flakiness locally.
- It also checks for the docker daemon in the commands and fails if it's not running
- It makes the local integration tests also run on PRs. This does not setup any cloud credentials and should work (emulates the dev workflow).
 

Fixes #

### DIFF
--- a/.github/workflows/pr_local_integration_tests.yml
+++ b/.github/workflows/pr_local_integration_tests.yml
@@ -1,4 +1,5 @@
 name: pr-local-integration-tests
+# This runs local tests with containerized stubs of online stores. This is the main dev workflow
 
 on:
   pull_request_target:
@@ -13,7 +14,6 @@ jobs:
     if:
       (github.event.action == 'labeled' && (github.event.label.name == 'approved' || github.event.label.name == 'lgtm' || github.event.label.name == 'ok-to-test')) ||
       (github.event.action != 'labeled' && (contains(github.event.pull_request.labels.*.name, 'ok-to-test') || contains(github.event.pull_request.labels.*.name, 'approved') || contains(github.event.pull_request.labels.*.name, 'lgtm')))
-    needs: build-docker-image
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
@@ -60,7 +60,4 @@ jobs:
         run: make install-python-ci-dependencies
       - name: Test local integration tests
         if: ${{ always() }}  # this will guarantee that step won't be canceled and resources won't leak
-        env:
-          FEAST_USAGE: "False"
-          IS_TEST: "True"
         run: make test-python-integration-local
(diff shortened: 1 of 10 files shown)


.github/workflows/pr_local_integration_tests.yml
```diff
@@ -0,0 +1,66 @@
...
+      - name: Install pip-tools
+        run: pip install pip-tools
+      - name: Install dependencies
+        run: make install-python-ci-dependencies
+      - name: Test local integration tests
+        if: ${{ always() }}  # this will guarantee that step won't be canceled and resources won't leak
+        env:
+          FEAST_USAGE: "False"
+          IS_TEST: "True"
+        run: make test-python-integration-local
```
achals, line 66: Dumb q but how's this different from the existing `pr_integration_tests.yml` workflow? I think it's worth clarifying as a comment or in the PR notes.

CONTRIBUTING.md
```diff
@@ -166,7 +168,15 @@ To test across clouds, on top of setting up Redis, you also need GCP / AWS / Sno
...
 
 **Snowflake**
-- See https://signup.snowflake.com/
+1. See https://signup.snowflake.com/ to setup a trial. 
+2. Then to run successfully, you'll need some environment variables setup:
+```sh
+export SNOWFLAKE_CI_DEPLOYMENT='[snowflake_deployment]'                            
+export SNOWFLAKE_CI_USER='[your user]'
+export SNOWFLAKE_CI_PASSWORD='[your pw]'
+export SNOWFLAKE_CI_ROLE='SYSADMIN'
```
achals, line 177: This is not necessary `SYSADMIN`, depends on a contributors' specific setup.

sdk/python/tests/integration/feature_repos/repo_configuration.py
```diff
@@ -141,6 +142,7 @@
         }
 
 
+# Replace online store contaniers
 if os.getenv("FEAST_LOCAL_ONLINE_CONTAINER", "False").lower() == "true":
```
achals, line 146: We should document these flags somewhere.. yeesh

.github/workflows/pr_local_integration_tests.yml
adchia: I think it's more just that we should ensure this runs & passes & isn't flaky since I expect people to use this for local dev.

sdk/python/tests/integration/feature_repos/repo_configuration.py
adchia: maybe i guess. I'd prefer people just use the Makefile commands

kevjumba (approved):

PR #2928, opened 2022-07-10
chore: Widen dependencies

### LINKED ISSUE #2906: Widen dependencies

Feast often needs to be installed alongside other libraries. Currently Feast pins many of its dependencies to the minor version instead of using a broader range. This means that if other libraries also pin to narrow version ranges, dependencies just can't be resolved.

Change `setup.py` to use wider ranges for various dependencies. I'm not sure how that changes your testing strategy. Do you have to do some amount of matrix testing for important dependencies?

Not sure.

**Additional context**

N/A

chhabrakadabra:

Widening the dependency ranges will allow Feast to be easily installable alongside other packages. Incidentally, the dependencies are widened upwards (the upper-bound on the ranges is increased), which will lead to lots of library upgrades. I might split this into multiple PRs if the upgrades cause too much chaos in CI.

Fixes #2906 

## Additional notes

- Most of the files changed here are related to black formatting. That's because unfortunately, I had to upgrade `black` because the older version of black interacted poorly with the latest version of click (see [this](https://github.com/psf/black/issues/2976) for more details).
- The `sdk/python/feast/embedded_go/lib` directory has been excluded from `black`. The same thing had already been done for isort, flake8 and mypy, but not for black. It makes sense to exclude autogenerated code from linting and the newest version of black was trying to reformat autogenerated/non-committed code and failing.


setup.py
```diff
@@ -127,9 +126,9 @@
 CI_REQUIRED = (
     [
         "build",
-        "cryptography==35.0",
+        "cryptography>=35.0,<36",
         "flake8",
-        "black==19.10b0",
+        "black>=22.6.0,<23",
```
achals, line 131: Is this what caused all the formatting changes?

achals (approved):
Generally looks good just glossed over the formatting changes

setup.py
chhabrakadabra: Yeah. I added this in the "Additional notes" section in the PR description. Essentially, the older version of black is not compatible with the newer version of click.

PR #2936, opened 2022-07-13
chore: Fixes and Readme for python<>go interface

achals:

Fixes #

### DIFF
--- a/go/README.md
+++ b/go/README.md
@@ -1,18 +1,18 @@
-This dir contains the go logic that's executed by the `EmbeddedOnlineFeatureServer` from python.
+This directory contains the Go logic that's executed by the `EmbeddedOnlineFeatureServer` from Python.
 
 ## Building and Linking
-[gopy](https://github.com/go-python/gopy) generates (and compiles) a CPython extension module from a go package. That's what we're using here, as visible in [setup.py](../setup.py)
+[gopy](https://github.com/go-python/gopy) generates (and compiles) a CPython extension module from a Go package. That's what we're using here, as visible in [setup.py](../setup.py).
 
-Under the hood, gopy invoked `go build`, and then templates `cgo` stubs for the go module that exposes the public functions from the Go module as C functions.
-For our project, this stuff can be found at `sdk/python/feast/embedded_go/lib/embedded.go` & `sdk/python/feast/embedded_go/lib/embedded_go.h` after running `python setup.py develop`
+Under the hood, gopy invokes `go build`, and then templates `cgo` stubs for the Go module that exposes the public functions from the Go module as C functions.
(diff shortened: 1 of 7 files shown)


go/README.md
```diff
@@ -0,0 +1,110 @@
...
+    entitiesProto,
+    requestDataProto,
+    fullFeatureNames)
+
+entitiesRecord.Release()
+requestDataRecords.Release()
+
+```
+
+Additionally, we need to pass in a pair of pointers into which are set in `GetOnlineFeatures()` so that the resultant feature values can be passed back to Python (via the C layer) using zero-copy semantics.
```
kevjumba, line 66: maybe rephrase to, "Additionally, we need to pass in a pair of pointers to `GetOnlineFeatures()` that are populated by Go and the resultant populated feature values pointers(in the C layer) are passed back to Python via zero-copy semantics.

kevjumba (commented):
@achals I think in `online_features_service.py` right after `resp = record_batch_to_online_response(record_batch)` we should also explicilty release the returned recordbatch. I'm not quite sure if returning from getOnlineFeatures actually decrements the reference counter but better safe than sorry?

achals:
> @achals I think in `online_features_service.py` right after `resp = record_batch_to_online_response(record_batch)` we should also explicilty release the returned recordbatch. I'm not quite sure if returning from getOnlineFeatures actually decrements the reference counter but better safe than sorry?

I think we can explicitly `del` it

go/internal/feast/featurestore.go
```diff
@@ -113,7 +113,7 @@ func (fs *FeatureStore) GetOnlineFeatures(
 	}
 
 	result := make([]*onlineserving.FeatureVector, 0)
-	arrowMemory := memory.NewGoAllocator()
+	arrowMemory := memory.NewCgoArrowAllocator()
```
achals, line 116: We should fix this these lint issues

go/internal/feast/server/logging/memorybuffer_test.go
```diff
@@ -159,7 +159,7 @@ func TestSerializeToArrowTable(t *testing.T) {
 	expectedRecord := builder.NewRecord()
 	assert.Nil(t, err)
 	for colIdx := 0; colIdx < int(record.NumCols()); colIdx++ {
-		assert.Equal(t, expectedRecord.Column(colIdx), record.Column(colIdx), "Columns with idx %d are not equal", colIdx)
+		assert.Equal(t, true, array.Equal(expectedRecord.Column(colIdx), record.Column(colIdx)), "Columns with idx %d are not equal", colIdx)
```
achals, line 162: https://pkg.go.dev/github.com/stretchr/testify/assert#True instead of `assert.Equal`

felixwang9817: fixed

go/internal/feast/featurestore.go
felixwang9817: I think #2940 should resolve this?

kevjumba (approved):
/lgtm

Issue #2942, opened 2022-07-14
`_serialize_val()` treats int64 as a long instead of long long


vtao2:
## Expected Behavior 
In [`key_encoding_utils.py`, when serializing an int64](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/key_encoding_utils.py#L17), we should be able to serialize `int64` numbers (up to 2^64 - 1). 

## Current Behavior
Currently, we get:

`struct.error: 'l' format requires -2147483648 <= number <= 2147483647`

but `l` is for long, which is 4 bytes. `int64` should be 8 bytes, thus `q`, representing long long.
Source: https://docs.python.org/3.10/library/struct.html#format-characters

## Steps to reproduce
Call `_serialize_val()` on a number > 2147483647.

### Specifications

- Version: 0.22
- Platform: 
- Subsystem:

## Possible Solution
`return struct.pack("<l", v.int64_val), ValueType.INT64` --> `return struct.pack("<q", v.int64_val), ValueType.INT64`

achals:
@vtao2 you've stumbled onto a long standing bug (#1908, #2345) 🙁

Good news is it's on my plate to fix this in the coming sprint. I should have an update soon.

PR #2944, opened 2022-07-15
fix: Version entity serialization mechanism and fix issue with int64 vals

### LINKED ISSUE #1908: FeatureStore.get_online_features is broken with int64 vals
## Expected Behavior 
I should be able to use an int64 value as the join key value for an entity, when the value type is `value_type=ValueType.INT64`.

## Current Behavior
Only values -2147483648 <= number <= 2147483647 can be used as the value. This is equivalent to -2^31 <= number <= 2^31 - 1

## Steps to reproduce
Specify a value greater than 2^31 - 1 as a the join key value when looking up values from the feature store.

```
>>> fs.get_online_features(features=['driver_hourly_stats:conv_rate'],  entity_rows=[{'driver_id': 2147483648}]).to_df()
  File "<stdin>", line 1, in <module>
  File "/Users/achal/tecton/feast/sdk/python/feast/usage.py", line 202, in exception_logging_wrapper
    result = func(*args, **kwargs)
  File "/Users/achal/tecton/feast/sdk/python/feast/feature_store.py", line 897, in get_online_features
    union_of_entity_keys,
  File "/Users/achal/tecton/feast/sdk/python/feast/feature_store.py", line 924, in _populate_result_rows_from_feature_view
    requested_features=requested_features,
  File "/Users/achal/tecton/feast/sdk/python/feast/infra/passthrough_provider.py", line 81, in online_read
    result = self.online_store.online_read(config, table, entity_keys)
  File "/Users/achal/tecton/feast/sdk/python/feast/infra/online_stores/redis.py", line 190, in online_read
    redis_key_bin = _redis_key(project, entity_key)
  File "/Users/achal/tecton/feast/sdk/python/feast/infra/online_stores/helpers.py", line 40, in _redis_key
    key: List[bytes] = [serialize_entity_key(entity_key), project.encode("utf-8")]
  File "/Users/achal/tecton/feast/sdk/python/feast/infra/key_encoding_utils.py", line 41, in serialize_entity_key
    val_bytes, value_type = _serialize_val(v.WhichOneof("val"), v)
  File "/Users/achal/tecton/feast/sdk/python/feast/infra/key_encoding_utils.py", line 17, in _serialize_val
    return struct.pack("<l", v.int64_val), ValueType.INT64
struct.error: 'l' format requires -2147483648 <= number <= 2147483647
```

## Possible Solution

The trivial fix is to make [this line](https://github.com/feast-dev/feast/blob/dcae1606f53028ce5413567fb8b66f92cfef0f8e/sdk/python/feast/infra/key_encoding_utils.py#L17) use `"q"` as the format string, which refers to a `signed long long` value which is 8 bytes in size.

Changing this behavior is a backwards incompatible change, since it would change the bytes encoded for an entity key which takes on int64 values. This means that materialized records oh that value type would would be unreadable.

The "correct" way of fixing this would require us to version the encoding scheme in the feature store.

achals:
This PR should supercedes #2609. The PR is kinda WIP (specifically, fixing java and go deserialization and adding more tests), but mostly ready for review.

Fixes #1908, #2345, #2942


protos/feast/core/FeatureView.proto
```diff
@@ -75,6 +75,9 @@ message FeatureViewSpec {
 
     // Whether these features should be served online or not
     bool online = 8;
+
+    // Needed for backwards compatible behaviour when fixing
```
kevjumba, line 79: I think you need to update the new available id

sdk/python/feast/base_feature_view.py
```diff
@@ -49,6 +49,8 @@ class BaseFeatureView(ABC):
     created_timestamp: Optional[datetime]
     last_updated_timestamp: Optional[datetime]
 
+    _entity_key_serialization_version: Optional[int]
```
kevjumba, line 52: I'm assuming there would just be 2 versions, 1 b4 and 1 after. Maybe we can add a comment that details what values this can be?

sdk/python/feast/feature_view.py
```diff
@@ -521,6 +524,11 @@ def from_proto(cls, feature_view_proto: FeatureViewProto):
                 )
             )
 
+        if feature_view_proto.spec.entity_key_serialization_version <= 1:
+            feature_view.set_entity_key_serialization_version(1)
```
kevjumba, line 528: Probably add some type of explanation here to differentiate what these versions mean.

sdk/python/feast/infra/key_encoding_utils.py
```diff
@@ -6,15 +6,22 @@
...
         return v.bytes_val, ValueType.BYTES
     elif value_type == "int32_val":
         return struct.pack("<i", v.int32_val), ValueType.INT32
     elif value_type == "int64_val":
-        return struct.pack("<l", v.int64_val), ValueType.INT64
+        if (
+            entity_key_serialization_version >= 0
+            and entity_key_serialization_version <= 1
+        ):
+            return struct.pack("<l", v.int64_val), ValueType.INT64
```
kevjumba, line 23: Maybe raise a warning here?

sdk/python/tests/unit/infra/test_key_encoding_utils.py
```diff
@@ -0,0 +1,30 @@
...
+    )
+    # True int64, but should also be fine.
+    serialize_entity_key(
+        EntityKeyProto(
+            join_keys=["user"], entity_values=[ValueProto(int64_val=int(2 ** 31))]
+        ),
+        entity_key_serialization_version=2,
+    )
+
+    # Old serilization scheme, should fail.
```
kevjumba, line 24: typo `serialization`

sdk/python/tests/unit/infra/test_key_encoding_utils.py
```diff
@@ -0,0 +1,30 @@
...
+        entity_key_serialization_version=2,
+    )
+    # True int64, but should also be fine.
+    serialize_entity_key(
+        EntityKeyProto(
+            join_keys=["user"], entity_values=[ValueProto(int64_val=int(2 ** 31))]
+        ),
+        entity_key_serialization_version=2,
+    )
+
```
kevjumba, line 23: We should probably assert that the serialization worked by trying to deserializing(make sure struct.pack works)

protos/feast/core/FeatureView.proto
kevjumba: Also below

go/internal/feast/onlinestore/redisonlinestore.go
```diff
@@ -355,9 +361,15 @@ func serializeValue(value interface{}) (*[]byte, types.ValueType_Enum, error) {
 		return &valueBuffer, types.ValueType_INT32, nil
 	case *types.Value_Int64Val:
 		// TODO (woop): We unfortunately have to use 32 bit here for backward compatibility :(
```
kevjumba, line 363: delete comment

sdk/python/feast/infra/key_encoding_utils.py
kevjumba: same comment, probably raise a warning that this is encoding int64 to 4 bytes

sdk/python/feast/repo_config.py
```diff
@@ -139,6 +139,8 @@ class RepoConfig(FeastBaseModel):
 
     go_feature_retrieval: Optional[bool] = False
 
+    entity_key_serialization_version: StrictInt = 0
```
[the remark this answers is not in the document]
achals: I added docs - I'm keeping it an Int, but yea there should only be 2 values.

go/internal/feast/onlinestore/redisonlinestore.go
achals: Updated

sdk/python/feast/feature_view.py
achals: Added docs in the repo_config

protos/feast/core/FeatureView.proto
achals: removed this change

sdk/python/tests/unit/infra/test_key_encoding_utils.py
achals: fixed

sdk/python/tests/unit/infra/test_key_encoding_utils.py
achals: the error would manifest on serializing too (it'd raise a valueerror)

kevjumba (approved):
will let @felixwang9817  take a pass and stamp

felixwang9817 (approved):
/lgtm

PR #2948, opened 2022-07-16
feat: Add Snowflake materialization engine

### LINKED ISSUE #2633: Snowflake type behavior is strange
## Current Behavior

In `test_universal_online::test_online_retrieval`, `orders_df` has the following `dtypes`:
```
order_id             int64
driver_id            int64
customer_id          int64
origin_id            int64
destination_id       int64
order_is_success     int32
event_timestamp     object
dtype: object
```

When type inference occurs, `get_table_column_names_and_types` returns `[('order_id', 'int8'), ('driver_id', 'int16'), ('customer_id', 'int16'), ('origin_id', 'int8'), ('destination_id', 'int8'), ('order_is_success', 'int8'), ('event_timestamp', 'datetime64[ns]')]`.

### Specifications

- Version:
- Platform:
- Subsystem:

sfc-gh-madkins:

Allows Snowflake to be a materialization option when Snowflake is used as an offline store

Fixes #2633

### DIFF
--- a/sdk/python/feast/infra/materialization/snowflake_engine.py
+++ b/sdk/python/feast/infra/materialization/snowflake_engine.py
@@ -1,0 +1,492 @@
+import os
+import shutil
+from dataclasses import dataclass
+from datetime import datetime
+from pathlib import Path
+from typing import Callable, List, Literal, Optional, Sequence, Union
+
+import click
+import pandas as pd
+from colorama import Fore, Style
+from pydantic import Field, StrictStr
+from tqdm import tqdm
+
+import feast
+from feast.batch_feature_view import BatchFeatureView
+from feast.entity import Entity
+from feast.feature_view import FeatureView
+from feast.infra.materialization.batch_materialization_engine import (
+    BatchMaterializationEngine,
+    MaterializationJob,
+    MaterializationJobStatus,
+    MaterializationTask,
+)
+from feast.infra.offline_stores.offline_store import OfflineStore
+from feast.infra.online_stores.online_store import OnlineStore
+from feast.infra.registry.base_registry import BaseRegistry
+from feast.infra.utils.snowflake.snowflake_utils import (
+    _run_snowflake_field_mapping,
+    assert_snowflake_feature_names,
+    get_snowflake_conn,
+    get_snowflake_materialization_config,
+    package_snowpark_zip,
+)
+from feast.protos.feast.types.EntityKey_pb2 import EntityKey as EntityKeyProto
+from feast.protos.feast.types.Value_pb2 import Value as ValueProto
+from feast.repo_config import FeastConfigBaseModel, RepoConfig
+from feast.stream_feature_view import StreamFeatureView
+from feast.type_map import _convert_value_name_to_snowflake_udf
+from feast.utils import _coerce_datetime, _get_column_names
+
+
+class SnowflakeMaterializationEngineConfig(FeastConfigBaseModel):
+    """Batch Materialization Engine config for Snowflake Snowpark Python UDFs"""
+
+    type: Literal["snowflake.engine"] = "snowflake.engine"
+    """ Type selector"""
+
+    config_path: Optional[str] = (
+        Path(os.environ["HOME"]) / ".snowsql/config"
+    ).__str__()
+    """ Snowflake config path -- absolute path required (Cant use ~)"""
+
+    account: Optional[str] = None
+    """ Snowflake deployment identifier -- drop .snowflakecomputing.com"""
+
+    user: Optional[str] = None
+    """ Snowflake user name """
+
+    password: Optional[str] = None
+    """ Snowflake password """
+
+    role: Optional[str] = None
+    """ Snowflake role name"""
+
+    warehouse: Optional[str] = None
+    """ Snowflake warehouse name """
+
+    authenticator: Optional[str] = None
+    """ Snowflake authenticator name """
+
+    database: StrictStr
+    """ Snowflake database name """
+
+    schema_: Optional[str] = Field("PUBLIC", alias="schema")
+    """ Snowflake schema name """
+
+    class Config:
+        allow_population_by_field_name = True
+
+
+@dataclass
+class SnowflakeMaterializationJob(MaterializationJob):
+    def __init__(
+        self,
+        job_id: str,
+        status: MaterializationJobStatus,
+        error: Optional[BaseException] = None,
+    ) -> None:
+        super().__init__()
+        self._job_id: str = job_id
+        self._status: MaterializationJobStatus = status
+        self._error: Optional[BaseException] = error
+
+    def status(self) -> MaterializationJobStatus:
+        return self._status
+
+    def error(self) -> Optional[BaseException]:
+        return self._error
+
+    def should_be_retried(self) -> bool:
+        return False
+
+    def job_id(self) -> str:
+        return self._job_id
+
+    def url(self) -> Optional[str]:
+        return None
+
+
+class SnowflakeMaterializationEngine(BatchMaterializationEngine):
+    def update(
+        self,
+        project: str,
+        views_to_delete: Sequence[
+            Union[BatchFeatureView, StreamFeatureView, FeatureView]
+        ],
+        views_to_keep: Sequence[
+            Union[BatchFeatureView, StreamFeatureView, FeatureView]
+        ],
+        entities_to_delete: Sequence[Entity],
+        entities_to_keep: Sequence[Entity],
+    ):
+        click.echo(
+            f"Deploying materialization functions for {Style.BRIGHT + Fore.GREEN}{project}{Style.RESET_ALL}"
+        )
+        click.echo()
+
+        conn_config = get_snowflake_materialization_config(self.repo_config)
+
+        stage_context = f'"{conn_config.database}"."{conn_config.schema_}"'
+        stage_path = f'{stage_context}."feast_{project}"'
+        with get_snowflake_conn(conn_config) as conn:
+            cur = conn.cursor()
+
+            # if the stage already exists,
+            # assumes that the materialization functions have been deployed
+            cur.execute(f"SHOW STAGES IN {stage_context}")
+            stage_list = pd.DataFrame(
+                cur.fetchall(), columns=[column.name for column in cur.description]
+            )
+            if f"feast_{project}" in stage_list["name"].tolist():
+                click.echo(
+                    f"Materialization functions for {Style.BRIGHT + Fore.GREEN}{project}{Style.RESET_ALL} already exists"
+                )
+                click.echo()
+                return None
+
+            cur.execute(f"CREATE STAGE {stage_path}")
+
+            copy_path, zip_path = package_snowpark_zip(project)
+            cur.execute(f"PUT file://{zip_path} @{stage_path}")
+            shutil.rmtree(copy_path)
+
+            # Execute snowflake python udf creation functions
+            sql_function_file = f"{os.path.dirname(feast.__file__)}/infra/utils/snowflake/snowpark/snowflake_python_udfs_creation.sql"
+            with open(sql_function_file, "r") as file:
+                sqlFile = file.read()
+
+                sqlCommands = sqlFile.split(";")
+                for command in sqlCommands:
+                    command = command.replace("STAGE_HOLDER", f"{stage_path}")
+                    command = command.replace("PROJECT_NAME", f"{project}")
+                    cur.execute(command)
+
+        return None
+
+    def teardown_infra(
+        self,
+        project: str,
+        fvs: Sequence[Union[BatchFeatureView, StreamFeatureView, FeatureView]],
+        entities: Sequence[Entity],
+    ):
+        conn_config = get_snowflake_materialization_config(self.repo_config)
+
+        stage_path = (
+            f'"{conn_config.database}"."{conn_config.schema_}"."feast_{project}"'
+        )
+        with get_snowflake_conn(conn_config) as conn:
+            cur = conn.cursor()
+
+            cur.execute(f"DROP STAGE IF EXISTS {stage_path}")
+
+            # Execute snowflake python udf deletion functions
+            sql_function_file = f"{os.path.dirname(feast.__file__)}/infra/utils/snowflake/snowpark/snowflake_python_udfs_deletion.sql"
+            with open(sql_function_file, "r") as file:
+                sqlFile = file.read()
+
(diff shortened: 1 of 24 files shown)


sdk/python/feast/infra/materialization/snowflake/snowflake_engine.py
```diff
@@ -0,0 +1,243 @@
...
+        *,
+        repo_config: RepoConfig,
+        offline_store: OfflineStore,
+        online_store: OnlineStore,
+        **kwargs,
+    ):
+        assert (
+            repo_config.offline_store.type == "snowflake.offline"
+        ), "To use SnowflakeMaterializationEngine, you must use Snowflake as an offline store."
+        assert (
```
sfc-gh-madkins, line 101: @adchia @achals right now snowpark is only available for 3.8, and soon 3.9. How should I think about skipping testing for 3.7,3.9

sdk/python/feast/infra/materialization/snowflake/snowflake_engine.py
```diff
@@ -0,0 +1,243 @@
...
+                join_key_columns=join_key_columns,
+                feature_name_columns=feature_name_columns,
+                timestamp_field=timestamp_field,
+                created_timestamp_column=created_timestamp_column,
+                start_date=start_date,
+                end_date=end_date,
+            )
+
+            table = offline_job.to_arrow()
+            fv_latest_values_sql = offline_job.to_sql()
```
sfc-gh-madkins, line 165: @adchia @achals I get a lint error as offline_job doesnt have a to_sql() function

sdk/python/feast/infra/materialization/snowflake/snowflake_engine.py
```diff
@@ -0,0 +1,243 @@
...
+            fv_timestamp = feature_view.batch_source.timestamp_field
+
+            if feature_view.batch_source.created_timestamp_column:
+                fv_created = feature_view.batch_source.created_timestamp_column
+            else:
+                fv_created = None
+
+            online_table = f"""{self.repo_config.online_store.database}"."{self.repo_config.online_store.schema_}"."{self.repo_config.project}_{feature_view.name}"""
+
+            with Session.builder.configs(
```
sfc-gh-madkins, line 193: @achals @adchia There actually might be a code path that doesnt need snowpark, I will just need to take these inputs and build a dynamic sql query -- which might be the most efficient anyways

sdk/python/feast/infra/materialization/snowflake/snowflake_engine.py
```diff
@@ -0,0 +1,243 @@
...
+                    fv_latest_values_sql,
+                    online_table,
+                    fv_columns,
+                    join_key_to_value_type,
+                    fv_timestamp,
+                    fv_created,
+                )
+
+            """
+            if self.repo_config.online_store.type == "snowflake.online":
```
sfc-gh-madkins, line 215: @adchia @achals for right now I am going to assume that your data is in snowflake / offline store ... and therefore will be 2 code paths ... one to write it to your online store of choice, or one specific to snowflake as online store.

@adchia I guess in theory one could use snowflake ONLY as the materialization engine, but that is probably a pretty expensive endeavor for them if there data doesnt even live in snowflake.

sdk/python/feast/type_map.py
```diff
@@ -548,9 +547,11 @@ def snowflake_python_type_to_feast_value_type(
     return type_map[snowflake_python_type_as_str.lower()]
 
 
-def pa_to_redshift_value_type(pa_type: pyarrow.DataType) -> str:
+def pa_to_redshift_value_type(pa_type: "pyarrow.DataType") -> str:
```
sfc-gh-madkins, line 550: @achals @adchia one blocker for this endeavor is this pyarrow library in type_map.py. I want to reuse the code here (and not have to maintain a separate file) in my snowflake udfs but snowflake doesnt support the pyarrow package as of today.

Given pyarrow is only required for certain infra configs, I moved the import into the function and made the type hint a string hint -- which I am not sure how feast feels about ... looking for advice here

sdk/python/feast/templates/snowflake/driver_repo.py
```diff
@@ -51,14 +50,6 @@
     # amount of historical scanning required for historical feature values
     # during retrieval
     ttl=timedelta(weeks=52),
-    # The list of features defined below act as a schema to both define features
```
sfc-gh-madkins, line 54: Just looking to leverage the infer function

sdk/python/feast/infra/utils/snowflake_udfs.py
```diff
@@ -0,0 +1,54 @@
+import pandas
+from _snowflake import vectorized
+
+from feast.protos.feast.types.EntityKey_pb2 import EntityKey as EntityKeyProto
```
sfc-gh-madkins, line 4: @adchia @achals 

How do I get this file to ignore the __init__.py folder in feast main?

This file is the main file where the python udf handler is set, but obviously with wanting to reuse code already written, it looks to import the entire dependecies of the feast project. If I delete the __init__.py file, it works!

Maybe for right now on deployment, I just delete the file before it gets uploaded to snowflake.

sdk/python/feast/infra/utils/snowflake_udfs.py
```diff
@@ -0,0 +1,239 @@
...
+  RETURNS BINARY
+  LANGUAGE PYTHON
+  RUNTIME_VERSION = '3.8'
+  PACKAGES = ('protobuf', 'pandas')
+  HANDLER = 'feast.infra.utils.snowflake_udfs.feast_entity_key_proto_to_string'
+  IMPORTS = ('@feast_stage/feast.zip')
+"""
+# converts 1 to n many entity keys to a single binary for lookups
+@vectorized(input=pandas.DataFrame)
+def feast_entity_key_proto_to_string(df):
```
sfc-gh-madkins, line 190: SNowflake passes an array of variants, which might have another datatype? How do this arrive in the python UDF? And does it matter? Does python_values_to_proto_values create the right proto as long as it knows ValueType?

https://docs.snowflake.com/en/sql-reference/data-types-semistructured.html#array

sdk/python/feast/type_map.py
achals: I think moving the import inside the func and using a stringified type hint is fine.

If you want, you can also do a 
```
if typing.TYPE_CHECKING:
    import pyarrow
```

in the imports section.

sdk/python/feast/infra/utils/snowflake_udfs.py
achals: I'm not quite sure I understand the problem here - how is this file being packaged and used by snowflake? And is there a specific import that's causing other dependencies to come in transitively?

sdk/python/feast/infra/utils/snowflake_udfs.py
```diff
@@ -0,0 +1,239 @@
...
+        dict(zip(df[1].index, df[1].values)), orient="index", columns=df[0].values[0]
+    )
+
+    proto_values_by_column = {
+        column: python_values_to_proto_values(df[column].to_numpy(), value_type)
+        for column, value_type in list(join_keys.items())
+    }
+
+    serialized_entity_keys = [
+        serialize_entity_key(
```
achals, line 168: Highly recommend setting `entity_key_serialization_version=2` in this method.

sdk/python/feast/infra/utils/snowflake_udfs.py
achals: cc @adchia

achals (commented):
Overall looks reasonable to me

sdk/python/feast/infra/materialization/snowflake_engine.py
```diff
@@ -0,0 +1,341 @@
...
+            return SnowflakeMaterializationJob(
+                job_id=job_id, status=MaterializationJobStatus.SUCCEEDED
+            )
+        except BaseException as e:
+            return SnowflakeMaterializationJob(
+                job_id=job_id, status=MaterializationJobStatus.ERROR, error=e
+            )
+
+
+def _convert_value_name_to_snowflake_udf(value_name: str) -> str:
```
kevjumba, line 320: should go in feast type map

kevjumba (commented):
Just need some documentation and a few nits.

sdk/python/feast/infra/materialization/snowflake_engine.py
```diff
@@ -0,0 +1,341 @@
...
+                start_date=start_date,
+                end_date=end_date,
+            )
+
+            fv_latest_values_sql = (
+                offline_job.to_sql()
+            )  # do we want to run this first & miss udf overhead?
+
+            """
+            if feature_view.batch_source.field_mapping is not None:
```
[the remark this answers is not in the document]
sfc-gh-madkins: @kevjumba I need to write the function to handle field_mappings, its just on the to do list

sdk/python/feast/type_map.py
sfc-gh-madkins: Snowflake should be able to accept the origin code in 2 weeks without having to do anything wonky with pyarrow

sdk/python/feast/infra/utils/snowflake/snowpark/snowflake_udfs.py
```diff
@@ -0,0 +1,222 @@
...
+    _convert_value_type_str_to_value_type,
+    python_values_to_proto_values,
+)
+from feast.value_type import ValueType
+
+#    UNKNOWN = 0
+#    NULL = 19
+
+# nullable vs non-nullable?
+# date/timestamps with tz?
```
kevjumba, line 17: remove

kevjumba (approved):
I think there's still a few comments you haven't addressed.

kevjumba:
@sfc-gh-madkins  could you fix dco?

sdk/python/feast/infra/materialization/snowflake_engine.py
```diff
@@ -0,0 +1,459 @@
...
+        stage_path = (
+            f'"{conn_config.database}"."{conn_config.schema_}"."feast_{project_name}"'
+        )
+        with get_snowflake_conn(conn_config) as conn:
+            cur = conn.cursor()
+
+            cur.execute(f"CREATE STAGE IF NOT EXISTS {stage_path}")
+
+            cur.execute(f"PUT file://{zip_path} @{stage_path} OVERWRITE=TRUE")
+            shutil.rmtree(copy_path)
```
adchia, line 167: probably not a huge issue if this logic doesn't happen frequently, but as is any existing materialization jobs would fail after you remove right?

sdk/python/feast/infra/materialization/snowflake_engine.py
```diff
@@ -0,0 +1,459 @@
...
+                FROM (
+                  {fv_latest_mapped_values_sql}
+                )
+            """
+
+            feature_names = [feature.name for feature in feature_view.features]
+            feature_names_str = '", "'.join(feature_names)
+
+            conn_config = snowflake_materialization_config(self.repo_config)
+            if self.repo_config.online_store.type == "snowflake.online":
```
adchia, line 355: maybe factor out the logic for snowflake online store into a separate method? and likewise for the non-online store logic into another?

sdk/python/feast/infra/materialization/snowflake_engine.py
sfc-gh-madkins: copy_path is just the path to the locally create zip file. That zip file has already been loaded into snowflake, so I am just cleaning up here. 

Maybe the one thing here that could happen is during the PUT, the overwrite happens at the same exact time someone does a materialization job and it wouldnt be able to find the code, not likely but possible

sdk/python/feast/infra/materialization/snowflake_engine.py
```diff
@@ -0,0 +1,459 @@
...
+        views_to_delete: Sequence[
+            Union[BatchFeatureView, StreamFeatureView, FeatureView]
+        ],
+        views_to_keep: Sequence[
+            Union[BatchFeatureView, StreamFeatureView, FeatureView]
+        ],
+        entities_to_delete: Sequence[Entity],
+        entities_to_keep: Sequence[Entity],
+    ):
+        click.echo(
```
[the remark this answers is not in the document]
sfc-gh-madkins: its dependent on version + project -- each project should have its own unique infra.
The major bummer here is it takes like 5s for it to realize each function already exists x10

I can write a check to see if merely the zip file is in the stage, and if it is skip the function creation/checks

sdk/python/feast/infra/materialization/snowflake_engine.py
sfc-gh-madkins: See comment above .. by checking to see if the file already exists, we can skip this after the first run

sdk/python/feast/infra/offline_stores/offline_store.py
```diff
@@ -140,6 +140,12 @@ def to_arrow(
 
         return pyarrow.Table.from_pandas(features_df)
 
+    def to_sql(self):
```
achals, line 143: type hint?

sdk/python/feast/infra/online_stores/snowflake.py
```diff
@@ -80,7 +83,7 @@ def online_write_batch(
                     "entity_feature_key",
                     "entity_key",
                     "feature_name",
-                    "value",
+                    f"{config.project}_value",
```
achals, line 86: Is the idea that multiple projects wold share a single table for the online store?

sfc-gh-madkins: @achals this can actually go back to value -- in general, if wanting to use snowflake as the entire offline,engine,online I need to introduce a protected keyword list ... basically you cant have a feature with the same column name in offline and online table when doing a direct UPSERT

I still need to add this

achals: But if using snowflake as offline+online store, there would still be two different schemas or tables, right? which should allow the same value name? 

(Partly I'm worried about the weird name, and partly that this change may break existing users of snowflake as the online store)

sfc-gh-madkins: @achals reverted to old schema

achals (approved):
/lgtm

achals:
Tests are red because of a couple of tests that aren't being skipped correctly (since pr-local-integration-tests is triggered on pull_request_target). 

Going to force merge this in.

achals (approved):
/lgtm

PR #2951, opened 2022-07-19
chore: Remove UI from quickstart colab 

adchia:
since Colab only supports Python 3.7 (and Feast deprecated Python 3.7). This means Colab isn't installing the latest versions of Feast (which fix issues in the UI).

Fixes #


kevjumba (approved):
/LGTM

PR #2970, opened 2022-07-25
fix: Remove hard-coded integration test setup for AWS & GCP

kevjumba:

Fixes #

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -213,9 +213,10 @@ Then run `make test-python-integration`. Note that for Snowflake / GCP / AWS, th
 
 #### Running specific provider tests or running your test against specific online or offline stores
 
-1. If you don't need to have your test run against all of the providers(`gcp`, `aws`, and `snowflake`) or don't need to run against all of the online stores, you can tag your test with the ones you need by adding the pytest markers(`@pytest.mark.universal_online_stores` or `@pytest.mark.universal_online_stores` with the `only` parameter). The `only` parameter selects specific offline providers and online stores that your test will test against. Example:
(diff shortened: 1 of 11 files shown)


CONTRIBUTING.md
```diff
@@ -180,12 +211,28 @@ export SNOWFLAKE_CI_WAREHOUSE='[your warehouse]'
...
+```python
+# Only parametrizes this test with the sqlite online store
+@pytest.mark.universal_online_stores(only=["sqlite"])
+def test_feature_get_online_features_types_match():
+```
+
+2. You can also filter tests to run by using pytest's cli filtering. Instead of using the make commands to test Feast, you can filter tests by name with the `-k` parameter. The parametrized integration tests are all uniquely identified by their provider and online store so the `-k` option can select only the tests that you need to run. For example, to run only Redshift related tests, you can use the following command:
+
+```sh
+FEAST_USAGE=False IS_TEST=True python -m pytest -n 8 --integration -k Redshift sdk/python/tests
```
achals, line 227: I wonder if you can just disable these in `conftest.py`? That simplifies the command that users have to run?

kevjumba: fixed

sdk/python/tests/conftest.py
```diff
@@ -24,13 +25,15 @@
 import pytest
 from _pytest.nodes import Item
 
-from feast import FeatureStore
-from feast.wait import wait_retry_backoff
-from tests.data.data_creator import create_basic_driver_dataset
-from tests.integration.feature_repos.integration_test_repo_config import (
+os.environ["FEAST_USAGE"] = "False"
+os.environ["IS_TEST"] = "True"
```
achals, line 29: I think we should wrap this wrap these somehow so that we set them when running tests and restore the old values once we're done running tests. Pytest should have a way of making this possible

achals (approved):
/lgtm

PR #2971, opened 2022-07-26
fix: Fix file offline store logic for feature views without ttl

### LINKED ISSUE #2960: Inconsistent behavior between File and Snowflake offline stores
## Expected Behavior 

With equivalent definitions and the same data in each offline store, `get_historical_features` should return the same results regardless of backing store.

## Current Behavior

With the same underlying data, equivalent definitions and given the same entity data frame, File and Snowflake offline stores return different results.

```
$ python clear.py  query-file
      event_timestamp node
0 2015-10-21 02:00:00  foo
1 2015-10-21 03:00:00  foo
result:
            event_timestamp node  price  temp_f
0 2015-10-21 02:00:00+00:00  foo      6      67
1 2015-10-21 03:00:00+00:00  foo      6      67
$ python clear.py  query-snow
      event_timestamp node
0 2015-10-21 02:00:00  foo
1 2015-10-21 03:00:00  foo
result:
      event_timestamp node  price  temp_f
0 2015-10-21 02:00:00  foo      5      50
1 2015-10-21 03:00:00  foo      6      67
```

## Steps to reproduce

 * feature_store.yaml  : https://gist.github.com/cburroughs/17b343445757367044a9ed56917a769a
 * features.py : https://gist.github.com/cburroughs/f1825026e0ee19335bd3a34da965eb99
 * little cli script : https://gist.github.com/cburroughs/d6b79c31e9b1564e3a06e06e273e8b57

```
# download files; adjust credentials and db names
$ feast apply
$ python clear.py write-dummy-data-snowflake
$ python clear.py query-snow
# Switch which section is commented out in yaml
$ python clear.py write-dummy-data-file
$ python clear.py query-file
```

### Specifications

- Version: Feast SDK Version: "feast 0.22.1"
- Platform  x86_64 on Python 3.9.12
- Subsystem: Linux 5.4.188

## Possible Solution

I presume it is incorrect that these are returning different results, but if this is undefined behavior where both are in some sense legitament, which behavior in Feast is undefined should be more explicit in the docs.

felixwang9817:

Fixes #2960

### DIFF
--- a/sdk/python/feast/infra/offline_stores/file.py
+++ b/sdk/python/feast/infra/offline_stores/file.py
@@ -636,6 +636,14 @@ def _filter_ttl(
         ]
 
         df_to_join = df_to_join.persist()
+    else:
+        df_to_join = df_to_join[
+            # do not drop entity rows if one of the sources returns NaNs
+            df_to_join[timestamp_field].isna()
+            | (df_to_join[timestamp_field] <= df_to_join[entity_df_event_timestamp_col])
+        ]
+
+        df_to_join = df_to_join.persist()
 
     return df_to_join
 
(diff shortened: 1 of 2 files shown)


kevjumba (approved):
/lgtm

PR #2974, opened 2022-07-27
feat: Initial Bytewax materialization engine

whoahbot:

This PR adds a Bytewax batch materialization engine.

Fixes #

### DIFF
--- a/docs/reference/batch-materialization/bytewax.md
+++ b/docs/reference/batch-materialization/bytewax.md
@@ -1,0 +1,59 @@
+# Bytewax
+
+## Description
+
+The [Bytewax](https://bytewax.io) batch materialization engine provides an execution
+engine for batch materializing operations (`materialize` and `materialize-incremental`).
+
+### Guide
+
+In order to use the Bytewax materialization engine, you will need a [Kubernetes](https://kubernetes.io/) cluster running version 1.22.10 or greater.
+
+#### Kubernetes Authentication
+
+The Bytewax materialization engine loads authentication and cluster information from the [kubeconfig file](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/). By default, kubectl looks for a file named `config` in the `$HOME/.kube directory`. You can specify other kubeconfig files by setting the `KUBECONFIG` environment variable.
+
+#### Resource Authentication
+
+Bytewax jobs can be configured to access [Kubernetes secrets](https://kubernetes.io/docs/concepts/configuration/secret/) as environment variables to access online and offline stores during job runs.
+
+To configure secrets, first create them using `kubectl`:
+
+``` shell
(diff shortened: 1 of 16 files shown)


sdk/python/feast/infra/materialization/contrib/bytewax/bytewax_materialization_engine.py
```diff
@@ -0,0 +1,397 @@
...
+                                        "name": f"feast-{job_id}",
+                                    },
+                                ],
+                            }
+                        ],
+                        "containers": [
+                            {
+                                "command": ["sh", "-c", "sh ./entrypoint.sh"],
+                                "env": job_env,
+                                "image": "bytewax/bytewax-feast:latest",
```
achals, line 349: I'm guessing that this is https://hub.docker.com/r/bytewax/bytewax-feast - can we add some docs on what dependencies this container comes with and how it's built?

Additionally, I think this should be configurable (so that feast users can potentially supply a customer container with custom online store implementation), with this image being the default.

docs/reference/batch-materialization/bytewax.md
```diff
@@ -0,0 +1,59 @@
...
+``` yaml
+batch_engine:
+  type: bytewax
+  namespace: bytewax
+  pods: 3
+```
+
+The `namespace` configuration directive specifies which Kubernetes [namespace](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/) jobs, services and configuration maps will be created in.
+
+The `pods` configuration directive is used to configure the number of Bytewax pods that will be started for each FeatureView that is being materialized.
```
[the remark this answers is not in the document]
whoahbot: Good suggestion. I will make that change.

sdk/python/feast/infra/materialization/contrib/bytewax/bytewax_materialization_engine.py
whoahbot: I think making it configurable is a good idea.

My thought was to create a repo that would be used to build the `bytewax/bytewax-feast` image and push it to Dockerhub. People that want to make changes can fork that repo and build whichever images they would like to use. What do you think?

I haven't taken that step quite yet as I was building the Feast SDK from the source files that included the changes in this PR.

achals: I think having a repo that builds the image makes sense. Having it as part of the main repo may be good too, since that keeps everything in the same place. Doing it in a subsequent PR is fine by my

whoahbot: That sounds good. I'll add it in a subsequent PR, since it will depend on being able to install a release version of Feast that includes this code.

achals:
@whoahbot looking at this PR now, but do you mind signing your commits? Also you may need a rebase!

achals (approved):
/lgtm

Issue #2996, opened 2022-08-02
Feast UI breaks with savedDatasets


eso-exs:
I'm encountering an issue with Feast UI which stops working after I create a Dataset in Snowflake. This is the console error that I am seeing:
```
query.js:356 ZodError: [
  {
    "code": "invalid_type",
    "expected": "object",
    "received": "undefined",
    "path": [
      "savedDatasets",
      0,
      "spec",
      "storage",
      "fileStorage"
    ],
    "message": "Required"
  },
  {
    "code": "invalid_type",
    "expected": "object",
    "received": "undefined",
    "path": [
      "savedDatasets",
      0,
      "spec",
      "featureService"
    ],
    "message": "Required"
  }
]
    at yk (index.mjs:449:23)
    at n.value (index.mjs:561:16)
    at n.value (index.mjs:541:29)
    at useLoadRegistry.ts:35:47
```

AlexEijssen:
Hi, I'm experiencing the same issue when testing my implementation of Postgres. If you look at the zod parser it expects the values you're missing here: https://github.com/feast-dev/feast/blob/master/ui/src/parsers/feastSavedDataset.ts
I tested it locally making those values optional and it works, but it will take me some time to submit my PR for this issue.

PR #2956, opened 2022-07-20
fix: Add dummy alias to pull_all_from_table_or_query

ikrizanic:
Signed-off-by: Ivan Krizanic <ikrizanic75@gmail.com>

**What this PR does / why we need it**:
Fixes missing alias in subquery when using `pull_all_from_table_or_query()` method.
**Which issue(s) this PR fixes**:

Fixes #2954

Issue #2954, opened 2022-07-20
Missing alias in `PostgreSQLOfflineStore.pull_all_from_table_or_query()`


ikrizanic:
## Expected Behavior
Call `pull_all_from_table_or_query()` method and retrieve wanted data from offline store.

## Current Behavior
SQL error `psycopg2.errors.SyntaxError: subquery in FROM must have an alias`

## Steps to reproduce
Call `pull_all_from_table_or_query()` on `PostgreSQLOfflineStore`, where `PostgreSQLSource` is initialized with `query="SELECT * FROM table_name"`
### Specifications

- Version: 0.22.1
- Platform: x86_64
- Subsystem: Linux 5.18

## Possible Solution

Adding alias in query located in `PostgreSQLOfflineStore.pull_all_from_table_or_query()`, for example adding `AS dummy_alias`. This was tested on my use-case and it works as expected, and if it looks ok, I could open a PR with this change.

```
query = f"""
            SELECT {field_string}
            FROM {from_expression} AS dummy_alias
            WHERE "{timestamp_field}" BETWEEN '{start_date}'::timestamptz AND '{end_date}'::timestamptz
        """
```

PR #3002, opened 2022-08-03
feat: Adding saved dataset capabilities for Postgres

### LINKED ISSUE #2995: Add Postgres as a storage for creating saved datasets.
Currently, creating a dataset with Postgres as a storage is not supported.

The option to use Postgres as a storage when creating a dataset.

**Additional context**
I'm close to creating a pull request for this issue myself.

AlexEijssen:
Signed-off-by: alex.eijssen <alex.eijssen@energyessentials.nl>

This PR adds the possibility to save datasets in Postgres and register them in Feast. I also tried to move the Postgres code more in the direction of Redshift and Bigquery.

Fixes #2995 

**DISCLAIMER**:
If you register a dataset the UI breaks. This appears to be independent of Postgres, as issue #2996 describes the same issue I have, but then for Snowflake. I suspect the issue is in feast/ui/src/parsers/feastSavedDataset.ts that is too stringent/limited. And I'm working on a fix (wrapping up a PR) that ensures registering a dataset doesn't break the UI.

### DIFF
--- a/sdk/python/feast/infra/registry_stores/sql.py
+++ b/sdk/python/feast/infra/registry_stores/sql.py
@@ -688,15 +688,16 @@ class SqlRegistry(BaseRegistry):
         self,
         table: Table,
         project: str,
-        id_field_name,
-        obj,
-        proto_field_name,
-        name=None,
+        id_field_name: str,
+        obj: Any,
+        proto_field_name: str,
+        name: Optional[str] = None,
     ):
         self._maybe_init_project_metadata(project)
 
         name = name or obj.name if hasattr(obj, "name") else None
         assert name, f"name needs to be provided for {obj}"
+
         with self.engine.connect() as conn:
             update_datetime = datetime.utcnow()
             update_time = int(update_datetime.timestamp())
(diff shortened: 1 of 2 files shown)


sdk/python/feast/infra/offline_stores/contrib/postgres_offline_store/postgres.py
```diff
@@ -214,7 +223,7 @@ def pull_all_from_table_or_query(
 
         query = f"""
             SELECT {field_string}
-            FROM {from_expression} AS paftoq_alias
```
adchia, line 217: i think someone added this in because it caused a crash without this? https://github.com/feast-dev/feast/pull/2956

adchia (commented):
I'm assuming that this fixes at least one test in https://github.com/feast-dev/feast/blob/master/Makefile#L112

Mind removing the excluded tests and running to make sure the new tests pass?

sdk/python/feast/infra/offline_stores/contrib/postgres_offline_store/postgres.py
AlexEijssen: I had a similar issue. I thought it was caused by the fact data_source.get_table_query_string() always puts "()" around the from_expression. Even when from_expression is just a table name. I decided to adjust get_table_query_string() inspired by redshift_source.py. It's rather unfortunate that the context of #2954 isn't clear. because I had the exact same error. But I'm unsure if his fix works for me and vice versa. I think adding AS paftoq_alias and combining it with my version of get_table_query_string() would be the safest option.

adchia: yeah im not sure. But what you said as the safe option sounds good to me

AlexEijssen:
I'm not sure if my latest commit is what you meant by: "Mind removing the excluded tests and running to make sure the new tests pass?"

adchia:
> I'm not sure if my latest commit is what you meant by: "Mind removing the excluded tests and running to make sure the new tests pass?"

I sent a PR in your repo to help with it, but had some trouble getting the normal `get_historical_features` tests to pass

adchia:
Are you on the Slack? Might be easier to discuss this syncrhonously in slack.feast.dev

adchia:
the postgres tests aren't being run continuously (mostly because we don't have anybody who is willing to "own" the component) so you'd actually have to manually run the male command to run the tests

PR #3005, opened 2022-08-03
chore: Remove old way of flagging features as experimental

adchia:
For a while now, we've been using warnings as the default way to tell users a feature is experimental. This formally deprecates needing `feature_store.yaml` flags to enable ODFV and lambda feature servers.

Fixes #


sdk/python/docs/source/feast.rst
```diff
@@ -169,14 +169,6 @@ feast.field module
    :undoc-members:
    :show-inheritance:
 
-feast.flags module
```
felixwang9817, line 172: does `sdk/python/docs/index.rst` also need to be updated? not sure which of these two is the source of truth

docs/reference/feature-servers/alpha-aws-lambda-feature-server.md
```diff
@@ -41,12 +34,21 @@ If enabled, the feature server will be deployed during `feast apply`. After it i
 
 Feast requires the following permissions in order to deploy and teardown AWS Lambda feature server:
 
+<<<<<<< HEAD:docs/reference/feature-servers/alpha-aws-lambda-feature-server.md
```
felixwang9817, line 37: rebase issues here

adchia: done

sdk/python/docs/source/feast.rst
adchia: nope. They are just different. index.rst gives the top level structure and this isn't referenced there.

This change was generated by running `make build-sphinx`

felixwang9817 (approved):
/lgtm

PR #3008, opened 2022-08-03
chore: Delete unused code in entity

felixwang9817:

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ benchmark-python-local:
 	FEAST_USAGE=False IS_TEST=True FEAST_IS_LOCAL_TEST=True python -m pytest --integration --benchmark  --benchmark-autosave --benchmark-save-data sdk/python/tests
 
 test-python:
-	FEAST_USAGE=False IS_TEST=True python -m pytest -n 8 sdk/python/tests
+	FEAST_USAGE=False IS_TEST=True python -m pytest -n 8 sdk/python/tests -k "not test_sql_registry"
 
 test-python-integration:
 	FEAST_USAGE=False IS_TEST=True python -m pytest -n 8 --integration sdk/python/tests
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,7 @@ benchmark-python:
 benchmark-python-local:
 	FEAST_USAGE=False IS_TEST=True FEAST_IS_LOCAL_TEST=True python -m pytest --integration --benchmark  --benchmark-autosave --benchmark-save-data sdk/python/tests
 
+# Runs all unit tests. Omits test_sql_registry since it requires Docker containers.
 test-python:
 	FEAST_USAGE=False IS_TEST=True python -m pytest -n 8 sdk/python/tests -k "not test_sql_registry"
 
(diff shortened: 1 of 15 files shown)


Makefile
```diff
@@ -63,7 +63,7 @@ benchmark-python-local:
 	FEAST_USAGE=False IS_TEST=True FEAST_IS_LOCAL_TEST=True python -m pytest --integration --benchmark  --benchmark-autosave --benchmark-save-data sdk/python/tests
 
 test-python:
-	FEAST_USAGE=False IS_TEST=True python -m pytest -n 8 sdk/python/tests
+	FEAST_USAGE=False IS_TEST=True python -m pytest -n 8 sdk/python/tests -k "not test_sql_registry"
```
adchia, line 66: why did you exclude this test? Can you add a comment?

sdk/python/feast/entity.py
```diff
@@ -33,7 +32,10 @@ class Entity:
 
     Attributes:
         name: The unique name of the entity.
-        value_type (deprecated): The type of the entity, such as string or float.
+        value_type: The type of the entity, such as string or float.
```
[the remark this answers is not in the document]
felixwang9817: `value_type` is still an attribute of an entity; it's just not set during initialization, but instead inferred during type inference

sdk/python/feast/entity.py
```diff
@@ -33,7 +32,10 @@ class Entity:
...
+        join_keys: A list of properties that uniquely identifies different entities within the
+            collection. This currently only supports a list of size one, but is intended to
+            eventually support multiple join keys.
         join_key: A property that uniquely identifies different entities within the
```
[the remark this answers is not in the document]
felixwang9817: we haven't removed the usage of `join_key` throughout the codebase, so we still want to be able to do `entity.join_key`, which is why `join_key` is left as an attribute

I clarified the TODO below

Makefile
felixwang9817: added comment

felixwang9817: actually, just marked the `test_sql_registry` tests as integration tests since they require external services

sdk/python/tests/integration/registration/test_sql_registry.py
```diff
@@ -112,6 +112,7 @@ def mysql_registry():
     sys.platform == "darwin" and "GITHUB_REF" in os.environ,
     reason="does not run on mac github actions",
 )
+@pytest.mark.integration
```
achals, line 115: These should be able to run as unit tests - any reason they can't?

achals (commented):
lg other than a nit

Makefile
felixwang9817: actually after talking with Achal, we decided to make it a unit test; will do in a separate PR

sdk/python/tests/integration/registration/test_sql_registry.py
felixwang9817: after discussing, we agreed to make these unit tests; will do in separate PR

achals (approved):
/lgtm

PR #3015, opened 2022-08-04
ci: Automate UI release using an NPM automation token. Add feature-server helm chart to release

adchia:
It now automatically bumps the UI package version, only allows the SDK to release after an npm publish succeeds, and uses an NPM automation token to bypass the 2FA error we used to get with GitHub actions.

It also adds in the feast-feature-server helm chart to the release process (I manually released the helm chart to 0.23.0) so it gets automatically published in the future.

Fixes #


infra/scripts/helm/validate-helm-chart-versions.sh
```diff
@@ -3,7 +3,7 @@
 set -e
 
 # Amount of file locations that need to be bumped in unison when versions increment
-UNIQUE_VERSIONS_COUNT=18
+UNIQUE_VERSIONS_COUNT=20
```
felixwang9817, line 6: why is this incrementing by 2 given that we're adding 3 new files to `files_to_bump.txt`?

adchia: this is just for helm charts, so that'd be two files

achals (approved):
/lgtm

PR #3016, opened 2022-08-04
chore!: Delete unused code in feature view

felixwang9817:

Fixes #

BREAKING CHANGE: Deprecate feature view parameters such as `batch_source` and `stream_source` and deprecate the option to use positional arguments


sdk/python/feast/on_demand_feature_view.py
```diff
@@ -94,136 +92,29 @@ def __init__(  # noqa: C901
...
-                _udf = args[3]
-
-        if not _name:
-            raise ValueError(
-                "The name of the on demand feature view must be specified."
-            )
-
-        if not _sources:
-            raise ValueError("The `sources` parameter must be specified.")
-
```
achals, line 215: 💯

achals (commented):
Is this still a WIP PR? Looks pretty good to me.

sdk/python/feast/feature_view.py
```diff
@@ -61,10 +60,10 @@ class FeatureView(BaseFeatureView):
         ttl: The amount of time this group of features lives. A ttl of 0 indicates that
             this group of features lives forever. Note that large ttl's or a ttl of 0
             can result in extremely computationally intensive queries.
-        batch_source (optional): The batch source of data where this group of features
+        batch_source: The batch source of data where this group of features
```
adchia, line 63: huh did you do this in reverse? the docs suggest we want source instead of batch_source vs stream_source

adchia: oh these are attributes. Maybe remove the two lines on "deprecated in favor of source"

felixwang9817: @adchia the `__init__` takes in `source` as a parameter; it then sets `batch_source` and `stream_source` as attributes

adchia (approved):
just a nit

sdk/python/feast/feature_view.py
felixwang9817: good catch

adchia: yep. just remove the two lines here saying "This is deprecated in favor of `source`" ?

adchia (approved):
/lgtm

PR #3027, opened 2022-08-05
fix: Add a test for applying feature views

felixwang9817:

Fixes #


adchia:
any reason this is added? seems like it's covered with https://github.com/feast-dev/feast/blob/master/sdk/python/tests/unit/local_feast_tests/test_local_feature_store.py but the naming of the test is probably clearer

felixwang9817:
closing as this test already exists, as mentioned above

PR #3028, opened 2022-08-06
fix: Fixing Web UI, which fails for the SQL registry

adchia:
This is caused by the SQL registry not populating the created timestamp for feature services.

Also cleans some unused imports

Fixes #


achals (approved):
/ok-to-test
/lgtm

PR #3029, opened 2022-08-06
chore: Suppress alpha warnings in test. Fix entity serialization in test

adchia:
The `make test-python` and `make test-python-integration-local` commands were littered with warnings. This suppresses that to make for a better testing experience.

**Note:** test_lambda fails when switching to entity_key_serialization_version=2 so this temporarily keeps it at 1 for that test

### DIFF
--- a/sdk/python/feast/stream_feature_view.py
+++ b/sdk/python/feast/stream_feature_view.py
@@ -8,7 +8,7 @@ from typing import Dict, List, Optional, Tuple, Union
 import dill
 from typeguard import typechecked
 
-from feast import utils
+from feast import flags_helper, utils
 from feast.aggregation import Aggregation
 from feast.data_source import DataSource
 from feast.entity import Entity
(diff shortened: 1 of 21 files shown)


kevjumba (approved):
/lgtm

sdk/python/tests/utils/feature_records.py
```diff
@@ -6,7 +6,7 @@
 from pandas.testing import assert_frame_equal as pd_assert_frame_equal
 from pytz import utc
 
-from feast import FeatureStore, utils
+from feast import FeatureService, FeatureStore, utils
```
achals, line 9: ideally these imports should come from `feast.feature_service`, `feast.feature_store`, etc.

Also,  `utils` :(

sdk/python/tests/integration/materialization/test_lambda.py
```diff
@@ -33,7 +33,10 @@ def test_lambda_materialization_consistency():
         },
         registry_location=RegistryLocation.S3,
     )
-    lambda_environment = construct_test_environment(lambda_config, None)
+    # TODO(adchia): figure out why entity_key_serialization_version 2 breaks with this test
+    lambda_environment = construct_test_environment(
+        lambda_config, None, entity_key_serialization_version=1
+    )
```
achals, line 39: This is probably becase we need to build and push a new docker image (that uses the new version number). The docker image isn't automatically updated via CI at the moment.

sdk/python/feast/stream_feature_view.py
```diff
@@ -8,7 +8,7 @@
 import dill
 from typeguard import typechecked
 
-from feast import utils
+from feast import flags_helper, utils
```
achals, line 11: same import comment.

achals (approved):

PR #3031, opened 2022-08-07
fix: Fix Feast Java inconsistency with int64 serialization vs python

adchia:
Verified the fix locally works now (previously didn't fetch features from Redis in Java when v2 was enabled)

This also updates the java-demo repo to not use BatchFeatureView (which has a bug where it doesn't apply) and also removes RequestFeatureView (which creates a bug)

Fixes #


achals (approved):
/lgtm

PR #3032, opened 2022-08-07
chore: Refactoring Feast Serving to have fewer packages

adchia:
This consolidates more of the packages, removes the packages that weren't being published / depended on by Feast serving. 

It also updates the Java tests to use the entity key serialization v2 mechanism.

It also renames the files of the "sdk" package to match its name (which is named feast-serving-client on Maven)

The resulting Java repo setup (described in CONTRIBUTING.md) only has:
- feast-serving (which has the feature server service)
- feast-serving-client (which has a Java client to interface with the feature service)
- coverage
- datatypes

A followup PR will refactor out logic to create a separate fat client

Fixes #


java/serving/src/main/java/feast/serving/config/ServingServiceConfigV2.java
```diff
@@ -38,14 +39,14 @@ public ServingServiceV2 registryBasedServingServiceV2(
     final ServingServiceV2 servingService;
     final ApplicationProperties.Store store = applicationProperties.getFeast().getActiveStore();
 
-    OnlineRetrieverV2 retrieverV2;
```
achals, line 41: rename var?

java/coverage/pom.xml
```diff
@@ -30,7 +30,7 @@
     <groupId>dev.feast</groupId>
     <artifactId>feast-parent</artifactId>
     <version>${revision}</version>
-    <relativePath>../..</relativePath>
+    <relativePath>..</relativePath>
```
achals, line 33: What does this module do exactly?

achals (commented):
Maybe worth updating the README for the java/ subdir for how the new code is laid out

java/coverage/pom.xml
adchia: It generates code coverage reports (JaCoCo) which presumably can be used in CI and in IDEs

java/serving/src/main/java/feast/serving/config/ServingServiceConfigV2.java
adchia: done

adchia:
@achals PTAL!

java/CONTRIBUTING.md
```diff
@@ -2,25 +2,48 @@
...
 
 > Don't see the Feast component that you want to contribute to here?  
 > Check out the [Development Guide](https://docs.feast.dev/v/master/project/development-guide)
 > to learn how Feast components are distributed over multiple repositories.
 
-#### Common Setup
+### Repository structure
+There are four key top level packages:
+- `serving`: Feast Serving (a gRPC service to serve features)
+- `serving-client`: Feast Serving Client (a thin Java client to communicate with Feast serving via gRPC )
```
achals, line 18: If you want to refactor this to be usable as a fat client, should the `serving` component be split into something like `retriever` which just talks to the online store and `serving` which depends on `retriever` and exposes a gRPC service?

achals:
> @achals PTAL!

lgtm but only one question - happy to merge after

java/CONTRIBUTING.md
adchia: yes. I wanted to do that as well, but seemed to make this PR too big. It would be essentially abstracting out most of what lies under ServingServiceV2 into a `FeatureStore` class

achals (approved):
/lgtm
/ok-to-test

PR #3036, opened 2022-08-08
fix: Parse inline data sources

### LINKED ISSUE #2991: data sources displayed different depending on "inline" definition
## Expected Behavior 
Many of the docs define data sources "inline" from the feature view.  For example <https://docs.feast.dev/getting-started/concepts/feature-view>
```
    #...
    source=BigQuerySource(
        table="feast-oss.demo_data.driver_activity"
    )
)
```

I would expect the above example to work the same with the `BigQuerySource` defined as is, or if it did `foo = BigQuerySource()` and `foo` was passed in.

## Current Behavior
```
node = feast.Entity(name='node', join_keys=['node'])

node_temp_file_fv = feast.FeatureView(
    name='node_temperature_fv',
    entities=[node],
    schema=[
        feast.Field(name='node', dtype=feast.types.String),
        feast.Field(name='temp_f', dtype=feast.types.Int64),
    ],
    online=False,
    source=feast.FileSource(
        name='temperature_source',
        path='./data/temperature_source.parquet',
        timestamp_field='timestamp',
    )
)
```

```
$ feast apply
Created entity node
Created feature view node_temperature_fv

Created sqlite table demo_node_temperature_fv
```

```
 $ feast data-sources list
NAME    CLASS
```

Change the above definition:
```
node_temperature_file_source = feast.FileSource(
    name='temperature_source',
    path='./data/temperature_source.parquet',
    timestamp_field='timestamp',
)

node_temp_file_fv = feast.FeatureView(
    name='node_temperature_fv',
    entities=[node],
    schema=[
        feast.Field(name='node', dtype=feast.types.String),
        feast.Field(name='temp_f', dtype=feast.types.Int64),
    ],
    online=False,
    source=node_temperature_file_source
)
```

```
$ feast apply
No changes to registry
No changes to infrastructure
```

But despite "no changes" being made, the source is now listed:
```
$ feast data-sources list
NAME                CLASS
temperature_source  <class 'feast.infra.offline_stores.file_source.FileSource'>
```

### Specifications

- Version: Feast SDK Version: "feast 0.22.2"
- Platform: x86_64 on Python 3.9.12
- Subsystem: Linux 5.4.188

felixwang9817:

Fixes #2991


sdk/python/feast/stream_feature_view.py
```diff
@@ -119,7 +119,7 @@ def __init__(
         super().__init__(
             name=name,
             entities=entities,
-            ttl=ttl or timedelta(days=0),
```
achals, line 122: may be better to have these changes in a different pr but won't block you

sdk/python/feast/repo_operations.py
```diff
@@ -113,44 +113,64 @@ def parse_repo(repo_root: Path) -> RepoContents:
...
-                    if batch_source_dep and batch_source_dep not in data_sources_set:
-                        res.data_sources.append(batch_source_dep)
+
+                # Handle batch sources defined with feature views.
+                batch_source = obj.batch_source
+                if not any((batch_source is ds) for ds in res.data_sources):
+                    res.data_sources.append(batch_source)
+
+                # Handle stream sources defined with feature views.
+                if obj.stream_source:
```
achals, line 170: At this point this should definitely be non-`None` right? Since it's a SFV? Or can you create a SFV without a `stream_source` somehow?

sdk/python/feast/repo_operations.py
```diff
@@ -113,44 +113,64 @@ def parse_repo(repo_root: Path) -> RepoContents:
...
+                    stream_source = obj.stream_source
+                    if not any((stream_source is ds) for ds in res.data_sources):
```
achals, line 172:
```suggestion
                    if stream_source not in res.data_sources:
```

is there any reason this wouldn't work?

sdk/python/feast/repo_operations.py
```diff
@@ -113,44 +113,64 @@ def parse_repo(repo_root: Path) -> RepoContents:
...
-                ):
-                    batch_source_dep = obj.stream_source.batch_source
-                    # Don't add if the push source's batch source is a duplicate of an existing batch source
-                    if batch_source_dep and batch_source_dep not in data_sources_set:
-                        res.data_sources.append(batch_source_dep)
+
+                # Handle batch sources defined with feature views.
+                batch_source = obj.batch_source
+                if not any((batch_source is ds) for ds in res.data_sources):
+                    res.data_sources.append(batch_source)
```
achals, line 167:
```suggestion
                if batch_source not in res.data_sources:
                    res.data_sources.append(batch_source)
```

May be cleaner? Or cleaner still to continue using a set and just add blindly and let the set dedupe.

sdk/python/feast/stream_feature_view.py
felixwang9817: yup will make separate PR

sdk/python/feast/repo_operations.py
felixwang9817: good catch

sdk/python/feast/repo_operations.py
felixwang9817: yeah so the reason I want to filter by `is` instead of `==` is precisely to avoid deduping

the only kind of duplication we should allow is directly importing an object from another file in the feature repo

sdk/python/feast/repo_operations.py
felixwang9817: same reason as above

felixwang9817:
also note that the test errors here will be fixed once #3037 lands

sdk/python/feast/repo_operations.py
felixwang9817: you might e.g. define a file source in a different `.py` file with just the path `data.parquet` and then re-define it inline for a feature view with the same path

this would result in two unique FileSource objects that are `==`, which I think we should prevent

adchia:
nit: have more descriptive PR title since this goes into the changelog

felixwang9817:
@adchia fixed!

adchia (approved):
/lgtm

PR #3044, opened 2022-08-09
feat: Feast AWS Athena offline store (again)

toping4445:

It enables Feast users to use **S3+AWS Athena** as an offline store.

1. All unit tests passed (except for tests in test_sql_registry.py.) 
The tests above failed in my environment because of MySQLdb package-M1 compatibility issues.
2. some integration tests passed. (passed:109, failed:19, ignored:16)
I didn't implement some methods related to feature write. and tests with fixed S3 bucket name and related to GCP failed.
However, all tests in test_universal_historical_retrieval.py and test_univeral_types.py related to feature extraction have passed.

Which issue(s) this PR fixes:

Fixes #

Does this PR introduce a user-facing change?:

AWS Users can choose between Redshift and S3+Athena for an offline store.


adchia:
I noticed there were a bunch of weird merge artifacts so I fixed them. Also removed Athena from the integration tests that run as part of CI since this is a contrib plugin.

can you sign your commits? None of them seem signed. 

Also: I have no way of reproducing the test results because I don't have Athena setup. It's ok for this PR, but maybe leave a comment in the Makefile to specify that it needs the user to have their own custom athena setup?

sdk/python/feast/infra/offline_stores/contrib/athena_offline_store/athena_source.py
```diff
@@ -0,0 +1,350 @@
...
+        )
+
+        if table is None and query is None:
+            raise ValueError('No "table" argument provided.')
+        _name = name
+        if not _name:
+            if table:
+                _name = table
+            else:
+                warnings.warn(
```
adchia, line 72: can you remove this? It's past Feast 0.21 now haha so should enforce there is a name

adchia (approved):

sdk/python/feast/infra/offline_stores/contrib/athena_offline_store/athena_source.py
toping4445: I changed AthenaSource to have a name!

adchia:
it’d be good to have some documentation, but this is good for an initial cut. Thanks!

toping4445:
Is there a documentation guide? If you share it, I will refer to it and fill it out when I have time.

PR #3053, opened 2022-08-09
ci: Fix docker image java workflow to read from the right maven cache

adchia:

Fixes #


felixwang9817 (approved):
/lgtm

adchia:
Force merging since the workflow will still fail since it uses the old one

PR #3059, opened 2022-08-10
ci: Auto update yarn.lock as part of release

adchia:
There was an issue now that the release publishes new NPM packages for the UI as part of the flow. Roughly it's:
1. Semantic release bumps file versions (including ui package version) and creates a tagged commit. 
2. When building the wheel, it needs to compile the Feast UI to bundle into the release. This updates the `yarn.lock` file because now there's a new published NPM version of the Feast UI.
3. Because now the client is dirty with local changes, the built wheel is a dev wheel instead of a clean wheel (see https://github.com/pypa/setuptools_scm/#default-versioning-scheme).

This change instead now will publish the NPM release before running semantic release:
- Find the new release with a dry run of the semantic release to get the old / new release versions (Semantic Release does not support this yet, as per https://github.com/semantic-release/semantic-release/issues/753)
- Temporarily bump file versions using above versions
- Publish Web UI

Then, Semantic release will additionally run `make build-ui` and check in the updated yarn.lock file before tagging the release

Fixes #

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -133,7 +133,7 @@ jobs:
   publish-python-sdk:
     if: github.repository == 'feast-dev/feast'
     runs-on: ubuntu-latest
-    needs: [build_wheels, publish-web-ui-npm]
+    needs: [build_wheels]
     steps:
       - uses: actions/download-artifact@v2
         with:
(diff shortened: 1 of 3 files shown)


kevjumba (approved):
/lgtm

PR #3068, opened 2022-08-11
chore: Move registry files to `infra/registry`

felixwang9817:
**What this PR does / why we need it**: The `sdk/python/feast/` directory is extremely flat, with ~50 Python files. In addition, the registry files are somewhat disorganized right now. This PR (a) moves all the registry files to a new submodule, `feast/infra/registry`, (b) renames `LocalRegistryStore` to `FileRegistryStore`, (c) separates the base registry into its own file, and (d) modifies the RTD docs structure to match these changes.

Fixes #


achals:
At a high level this change seems reasonable to me. I'll wait for tests to be green, and I'm assuming there's no real changes other than refactoring and moving things around.

achals (approved):
lg, suggest adding back a `registry.py` to maintain backwards compatibility

sdk/python/feast/registry.py
```diff
@@ -0,0 +1,10 @@
+import warnings
+
+from feast.infra.registry.registry import BaseRegistry, Registry
+
+warnings.warn(
```
achals, line 5:
```
warnings.simplefilter("once")
```

?

achals (approved):
/lgtm

sdk/python/feast/registry.py
```diff
@@ -1,1871 +1,12 @@
...
-from pathlib import Path
-from threading import Lock
-from typing import Any, Dict, List, Optional
-from urllib.parse import urlparse
+import warnings
 
-from google.protobuf.internal.containers import RepeatedCompositeFieldContainer
-from google.protobuf.json_format import MessageToJson
-from proto import Message
+from feast.infra.registry.registry import BaseRegistry, Registry  # noqa
```
adchia, line 3: might be able to avoid this by using https://github.com/feast-dev/feast/blob/master/sdk/python/feast/__init__.py#L23-L23 ?

felixwang9817: hmm @adchia not sure that works? the new `registry.py` containing `BaseRegistry` is on a different level

also the `noqa` thing is to avoid errors since `BaseRegistry` and `Registry` are imported and not used

adchia: not quite sure what you mean? BaseRegistry and Registry are in different files, yes, but you can make feast alias them into feast.Registry or feast.BaseRegistry

we do this already for the other offline/online stores which are similarly on a different level (e.g. you can do

```python
from .infra.registry.base_registry import BaseRegistry
from .infra.registry.registry import Registry
```

and add to the __all__ variable
) and then remove this file

felixwang9817: after discussion, decision was to just delete `registry.py` since there's no good reason for a user to be doing `from feast.registry import Registry` right now

PR #3071, opened 2022-08-11
fix: Fix Spark offline store type conversion to arrow

niklasvm:

Fixes some integration tests related to the spark offline store. Specifically when empty list data is converted from a spark data frame to arrow. This PR fixes 5 failing tests.

The current implementation converts from spark df --> pandas --> arrow.
The new implementation writes the data temporarily to parquet and then loads it with arrow.

None


adchia (approved):
/lgtm

PR #3074, opened 2022-08-11
fix: Fix incorrect on demand feature view diffing and improve Java tests

adchia:
There was an issue with diffing on demand feature views casting the wrong proto to OnDemandFeatureViewSpec. This was not caught for two reasons:
- There is no Python unit test testing that diff logic (or an integration test testing changing an existing odfv)
- The Java tests also started breaking, but in a silent way. This was caused by the tests relying on a checked in registry, which does not change between PRs. This PR now sets it up so that that registry is dynamically generated by Maven during tests. 

I haven't fully tested that the GitHub actions workflows work yet (because it requires we now also install Python on the workflows and the GitHub workflows aren't easy to test).

Fixes #


felixwang9817 (approved):
/lgtm

felixwang9817 (approved):
/lgtm

PR #3077, opened 2022-08-12
feat: Allow passing repo config path via flag

achals:

Fixes #

### DIFF
--- a/docs/tutorials/structuring-feature-repos.md
+++ b/docs/tutorials/structuring-feature-repos.md
@@ -1,0 +1,3 @@
+# How to Structure Feature Repositories
+
+## Overview
--- a/docs/tutorials/structuring-feature-repos.md
+++ b/docs/tutorials/structuring-feature-repos.md
@@ -1,3 +1,0 @@
-# How to Structure Feature Repositories
-
-## Overview
--- a/sdk/python/feast/cli.py
+++ b/sdk/python/feast/cli.py
@@ -498,7 +498,7 @@ def plan_command(ctx: click.Context, skip_source_validation: bool):
     repo = ctx.obj["CHDIR"]
     fs_yaml_file = ctx.obj["FS_YAML_FILE"]
     cli_check_repo(repo, fs_yaml_file)
-    repo_config = load_repo_config( repo, fs_yaml_file)
+    repo_config = load_repo_config(repo, fs_yaml_file)
     try:
         plan(repo_config, repo, skip_source_validation)
     except FeastProviderLoginError as e:
@@ -520,7 +520,7 @@ def apply_total_command(ctx: click.Context, skip_source_validation: bool):
     fs_yaml_file = ctx.obj["FS_YAML_FILE"]
     cli_check_repo(repo, fs_yaml_file)
 
-    repo_config = load_repo_config( repo, fs_yaml_file)
+    repo_config = load_repo_config(repo, fs_yaml_file)
     try:
         apply_total(repo_config, repo, skip_source_validation)
     except FeastProviderLoginError as e:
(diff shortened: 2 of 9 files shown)


docs/tutorials/structuring-feature-repos.md
```diff
@@ -0,0 +1,3 @@
+# How to Structure Feature Repositories
```
felixwang9817, line 1: belongs in another PR?

sdk/python/feast/repo_operations.py
```diff
@@ -325,13 +325,12 @@ def registry_dump(repo_config: RepoConfig, repo_path: Path) -> str:
     return json.dumps(registry_dict, indent=2, sort_keys=True)
 
 
-def cli_check_repo(repo_path: Path):
+def cli_check_repo(repo_path: Path, fs_yaml_file: Path):
```
felixwang9817, line 328: nitty but it's slightly confusing to me that `cli_check_repo` and `load_repo_config` have the args switched in order; is it possible to use the same order?

sdk/python/feast/cli.py
```diff
@@ -72,8 +72,17 @@ def format_options(self, ctx: click.Context, formatter: click.HelpFormatter):
     default="info",
     help="The logging level. One of DEBUG, INFO, WARNING, ERROR, and CRITICAL (case-insensitive).",
 )
+@click.option(
```
felixwang9817, line 75: can you add a test for this option? maybe in one of the CLI tests?

sdk/python/feast/feature_store.py
```diff
@@ -114,30 +114,39 @@ class FeatureStore:
...
         """
         Creates a FeatureStore object.
 
         Raises:
             ValueError: If both or neither of repo_path and config are specified.
         """
-        if repo_path is not None and config is not None:
-            raise ValueError("You cannot specify both repo_path and config.")
-        if config is not None:
+        if fs_yaml_file is not None and config is not None:
```
felixwang9817, line 132: hmm shouldn't this continue to check `repo_path`? as well as the logic below? since we still need exactly one of `repo_path` and `config`, and if `repo_path` is specified, there's an option for `fs_yaml_file` to be specified as well

felixwang9817 (commented):
looks good overall!

sdk/python/feast/repo_operations.py
achals: sure makes sense

sdk/python/feast/feature_store.py
achals: I think we actually need repo path regardless, and one of config or fs_yaml_path.

... I may change the FS constructor cause I feel like we should be able to only pass in in-memory objects to construct it. Dealing with the file system etc should not be something the FS has to care about (ideally)

felixwang9817 (approved):
/lgtm

adchia:
@achals can you also add docstrings for this?

cburroughs:
Does this resolve #1556?

achals:
> Does this resolve #1556?

Yes!

sdk/python/feast/feature_store.py
```diff
@@ -1569,7 +1581,7 @@ def _get_online_features(
         }
 
         # If the embedded Go code is enabled, send request to it instead of going through regular Python logic.
-        if self.config.go_feature_retrieval:
+        if self.config.go_feature_retrieval and self._go_server:
```
idahoakl, line 1584: @achals What was the reason for adding `and self._go_server` to this condition expression?  This seems to be the cause of why the go server will not start and trying to understand if there was another issue that was being solved for.

PR #3080, opened 2022-08-13
ci: Updating GE profiler test case with an actual unrealistic expecation

franciscojavierarceo:

It looks like the original test case for the profiler did not actually trigger a false expecation. I'm still debugging but this is the first fix. It looks like when an expecation is not met the profiler actually omits the test case, I'll likely fix that in a separate PR but will investigate.

Fixes a bug in the `profiler_with_unrealistic_expectations` test.

Fixes #

### DIFF
--- a/sdk/python/tests/integration/e2e/test_validation.py
+++ b/sdk/python/tests/integration/e2e/test_validation.py
@@ -4,6 +4,7 @@ import shutil
 import pandas as pd
 import pyarrow as pa
 import pytest
+import json
 from great_expectations.core import ExpectationSuite
 from great_expectations.dataset import PandasDataset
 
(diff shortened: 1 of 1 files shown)


adchia (approved):
/lgtm

adchia:
@franciscojavierarceo looks like there's a test failure!

franciscojavierarceo:
yeah i'm going to move this back to draft!

kevjumba:
@franciscojavierarceo Hey there's a merge conflict!

franciscojavierarceo:
i fixed this in a separate PR!

franciscojavierarceo:
Will close it out!

PR #3088, opened 2022-08-15
fix: Prevent overwriting existing file during `persist`

### LINKED ISSUE #3069: Error in FileOfflineStore.get_historical_features.<locals>.evaluate_historical_retrieval()
## Expected Behavior 

`.to_df()` should return a dataframe.

## Current Behavior

When trying to execute `.to_df()` after executing `.to_df(validation_reference=<some-validation-reference/>)` the `self.evaluation_function().compute()` call in the `_to_df_internal()` method inside the `FileRetrievalJob` class fails. 

While the Data Quality monitoring feature the is implemented within `validation_reference` is still in alpha, `.to_df()` should not be an issue.

## Steps to reproduce
I've provided a minimally reproducible example in [this notebook](https://github.com/franciscojavierarceo/feast/blob/1dc68381afe34c2a7ee283214cac8e2447e96de6/examples/data-quality-monitoring/dqm-minimal-failure.ipynb)
### Specifications

- Version: 0.23
- Platform: Python3.8
- Subsystem:

## Possible Solution
Following the stack trace it appears that there's an issue with the created date.

felixwang9817:

Fixes #3069

### DIFF
--- a/sdk/python/tests/integration/feature_repos/universal/data_sources/file.py
+++ b/sdk/python/tests/integration/feature_repos/universal/data_sources/file.py
@@ -59,17 +59,7 @@ class FileDataSourceCreator(DataSourceCreator):
             field_mapping=field_mapping or {"ts_1": "ts"},
         )
 
-    def create_saved_dataset_destination(
-        self, data_source: Optional[DataSource] = None
-    ) -> SavedDatasetFileStorage:
-        if data_source:
-            assert isinstance(data_source, FileSource)
-            return SavedDatasetFileStorage(
-                path=data_source.path,
-                file_format=ParquetFormat(),
-                s3_endpoint_override=None,
-            )
-
+    def create_saved_dataset_destination(self) -> SavedDatasetFileStorage:
         d = tempfile.mkdtemp(prefix=self.project_name)
         self.dirs.append(d)
         return SavedDatasetFileStorage(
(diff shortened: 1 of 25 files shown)


sdk/python/tests/integration/feature_repos/universal/data_sources/file.py
```diff
@@ -59,7 +59,17 @@ def create_data_source(
...
+    def create_saved_dataset_destination(
+        self, data_source: Optional[DataSource] = None
+    ) -> SavedDatasetFileStorage:
+        if data_source:
+            assert isinstance(data_source, FileSource)
+            return SavedDatasetFileStorage(
+                path=data_source.path,
+                file_format=ParquetFormat(),
+                s3_endpoint_override=None,
+            )
```
achals, line 71: Is this the main reason we're changing the interface for `create_saved_dataset_destination`? Can you explain what purpose the passed in data source serves?

felixwang9817: yeah so I want to copy an existing destination - however, each saved dataset object has a unique way of specifying destination (e.g. path vs. table vs. schema, etc.) so there's no super clean way to modify the `create_saved_dataset_destination` interface to allow for specifying a destination

the easiest thing to do is to just specify a data source, whose destination can be copied to a saved dataset object

achals: tbh i really dislike the idea of passing in a `source` as the `destination` :/ What else do you think we can pass in instead?

felixwang9817: switched to a different method, lmk what you think @achals

achals (approved):
/lgtm

PR #3089, opened 2022-08-16
fix: Fix feature service inference logic

### LINKED ISSUE #3075: can't retrieve specific field from feature service
## Expected Behavior 

```
from driver_ratings_feature_view import driver_ratings_fv
from driver_trips_feature_view import driver_stats_fv

driver_stats_fs = FeatureService(
    name="driver_activity",
    features=[driver_stats_fv, driver_ratings_fv[["lifetime_rating"]]]
)
```
If I define my feature service as above, 
it should return **all** fields of driver_stats_fv and **lifetime_rating** field from driver_ratings_fv. 

## Current Behavior
I get **all** fields defined in feature views even though I specify a certain field. 
Could you please check if this is the right behavior of feature service? 

### Specifications

- Version: feast 0.23.1
- Platform: Python3.8
- Subsystem:

felixwang9817:

Fixes #3075


sdk/python/feast/feature_service.py
```diff
@@ -86,29 +86,48 @@ def __init__(
...
             if isinstance(feature_grouping, BaseFeatureView):
-                # For feature services that depend on an unspecified feature view, apply inferred schema
+                projection = feature_grouping.projection
+
                 if fvs_to_update and feature_grouping.name in fvs_to_update:
-                    if feature_grouping.projection.desired_features:
-                        desired_features = set(
-                            feature_grouping.projection.desired_features
-                        )
+                    # There are three situations to be handled. First, the projection specifies
```
adchia, line 101: Might be easier to read if you have the cases within the blocks themselves, instead of all in one block before the if starts

sdk/python/feast/feature_service.py
```diff
@@ -123,8 +123,15 @@ def infer_features(self, fvs_to_update: Optional[Dict[str, FeatureView]] = None)
                             if f.name in desired_features:
                                 projection.features.append(f)
                     elif not projection.desired_features and projection.features:
-                        # Second cass, so nothing needs to be done.
-                        pass
+                        # Second cass, so nothing needs to be done. In case something went wrong
```
achals, line 126:
```suggestion
                        # Second case, so nothing needs to be done. In case something went wrong
```

sdk/python/feast/feature_service.py
```diff
@@ -86,29 +86,48 @@ def __init__(
...
+
                         # We need to set the features for the projection at this point so we ensure we're starting with
                         # an empty list.
-                        feature_grouping.projection.features = []
+                        projection.features = []
                         for f in fvs_to_update[feature_grouping.name].features:
                             if f.name in desired_features:
-                                feature_grouping.projection.features.append(f)
+                                projection.features.append(f)
+                    elif not projection.desired_features and projection.features:
```
adchia, line 125: would be cleaner imo if you had early returns instead of the elif and else.. and maybe some example definitions

e.g. 
```python
if projection.desired_features:
  # The projection wants to reference inferred features. Validate they exist
   # Example: FeatureService(features=[[fv_with_no_schema["feature]])
   ...
   return

if projection.features:
  # The projection only references features from a FV's known schema (not inferred). 
  # Example 1: FeatureService(features=[fv_with_schema[["feature"]])
  # Example 2: FeatureService(features=[fv_with_schema])
  return

# The projection wants all features from a FV that has an inferred schema
# Example: FeatureService(features=[fv_with_no_schema])
...
```

adchia, line 125: wondering if we can exit earlier than this. i.e in line 98 instead

sdk/python/feast/feature_service.py
```diff
@@ -85,32 +86,67 @@ def __init__(
...
+                            feature_view_name=feature_grouping.name,
+                            feature_service_name=self.name,
+                        )
+
+                    continue
+
+                if projection.features:
+                    # The projection has already selected features from a feature view with a
+                    # known schema, so no action needs to be taken.
+                    # Example: FeatureService(features=[fv[["existing_feature"]]]), where
```
adchia, line 136: This also applies if the user is selecting from a fv with a known schema, but doesn't select features right?

sdk/python/feast/feature_service.py
```diff
@@ -85,32 +86,67 @@ def __init__(
...
+                        projection.features = []
                         for f in fvs_to_update[feature_grouping.name].features:
                             if f.name in desired_features:
-                                feature_grouping.projection.features.append(f)
+                                projection.features.append(f)
                     else:
-                        feature_grouping.projection.features = fvs_to_update[
-                            feature_grouping.name
-                        ].features
+                        raise FeatureViewMissingDuringFeatureServiceInference(
```
adchia, line 126: probably should also throw this error in FeatureView#__getitem__ ?

adchia (approved):
/lgtm

PR #3113, opened 2022-08-19
feat: Refactor feature server helm charts to allow passing feature_store.yaml in environment variables

adchia:
The old python feature server helm chart relied on users to build their own Docker image to bundle in a `feature_store.yaml` file. This change makes it so users instead pass a base64 encoded version of this as an environment variable (similar to the AWS lambda feature server approach) to hopefully reduce time to production

- This adds a tutorial on how to run a Python feature server on K8s, hitting Redis
- This also adds documentation for how to query the feature server in the concepts page for feature retrieval

Fixes #


kevjumba (approved):
/lgtm

kevjumba:
verified that the tutorial works on mac m1!

PR #3115, opened 2022-08-20
fix: Make infer dummy entity join key idempotent

### LINKED ISSUE #3114: Snowflake materialization fails on feature views with dummy entity
## Expected Behavior 

Same as feature views with entities

## Current Behavior

Materialization query errors, query has duplicate column names caused by a dummy entity having multiple dummy join keys

### Specifications

- Version:
- Platform:
- Subsystem:

## Possible Solution

even if infer features is ran more than once, only update the dummy entity join key once

sfc-gh-madkins:

Updates a dummy entity join key only once even if `update_feature_views_with_inferred_features_and_entities` is called multiple times in `feast.apply()`

Fixes #3114

### DIFF
--- a/sdk/python/feast/inference.py
+++ b/sdk/python/feast/inference.py
@@ -156,7 +156,11 @@ def update_feature_views_with_inferred_features_and_entities(
                 )
 
         # Infer a dummy entity column for entityless feature views.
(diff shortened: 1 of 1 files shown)


kevjumba (approved):
/lgtm

PR #3119, opened 2022-08-22
chore: Support consistent reads in dynamodb

achals:

consistent reads will also help with the lambda integration tests, since the flakes are likley because of eventually consistent reads after materialization.

Fixes #


kevjumba (approved):
/lgtm

PR #3123, opened 2022-08-23
chore: Build Python protos during sphinx build

felixwang9817:

Fixes #

### DIFF
--- a/sdk/python/docs/conf.py
+++ b/sdk/python/docs/conf.py
@@ -25,6 +25,29 @@ sys.path.insert(0, os.path.abspath("../../feast"))
 sys.path.insert(0, os.path.abspath("../.."))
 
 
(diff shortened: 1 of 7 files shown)


sdk/python/docs/conf.py
```diff
@@ -25,6 +25,29 @@
...
+from pathlib import Path
+
+# cwd will be feast/sdk/python/docs/source
+cwd = Path(os.getcwd())
+
+# Change to feast/
+os.chdir(cwd.parent.parent.parent.parent)
+
+# Compile Python protos
+result = subprocess.run(["make", "compile-protos-python"], capture_output=True)
```
achals, line 45: This assumes make is installed in the rtd environment

felixwang9817: updated

PR #3138, opened 2022-08-25
fix: Fix release workflow to release 0.24.0

kevjumba:

This PR fixes a couple of issues. The smoke test doesn't work anymore because of `feast init` change where `feature_repo` is now located inside the repo. 

Also removes the ui build temporarily since 0.24.0 is already released. Will add back in future pr. 

Fixes #


felixwang9817 (approved):
/lgtm

PR #3151, opened 2022-08-29
fix: Add getter/setter for RestServer

yongheng:

The error below occurs when we start Java Server:

```
Caused by: com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "rest" (class feast.serving.service.config.ApplicationProperties), not marked as ignorable (2 known properties: "feast", "grpc"])
```

This PR fixes it.

Fixes #

### DIFF
--- a/java/serving/src/main/java/feast/serving/service/config/ApplicationProperties.java
+++ b/java/serving/src/main/java/feast/serving/service/config/ApplicationProperties.java
@@ -56,6 +56,14 @@ public class ApplicationProperties {
     this.grpc = grpc;
   }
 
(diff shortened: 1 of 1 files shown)


adchia:
I think the RestServer doesn't actually do anything? We should probably just remove all references to it instead

yongheng:
> I think the RestServer doesn't actually do anything? We should probably just remove all references to it instead

I didn't try to understand what the RestServer does too much and just wanted to make the error go away. Please feel free the close this PR if you decide to remove all references to it. Thanks!

PR #3154, opened 2022-08-30
fix: Handle complex Spark data types in SparkSource

rebstastic:
`get_table_column_names_and_types` of `SparkSource` returns non-string data type for complex Spark types.

For example, if a feature view contains a feature of type `ArrayType(DoubleType(), False)`, current implementation would return a dictionary such as
```
{'type': 'array', 'elementType': 'double', 'containsNull': False}
```
With this fix, the returning data type would be `array<double>`.

This PR ensures that all Spark types are serialized as strings.

Fixes #


achals (approved):
/lgtm

PR #3155, opened 2022-08-30
chore: Remove rest settings from java app properties

achals:
The rest setting was not being used since the java server moved to being gRPC only. Specifying it is only likely to cause warnings/errors and confusion.

Should be a better solution instead of #3151 

Fixes #


felixwang9817 (approved):
/lgtm

PR #3172, opened 2022-09-03
fix: Fix push API to respect feature view's already inferred entity types

### LINKED ISSUE #3171: Entity data type (int32) do not match PushSource data type (int64) when pushing to online store
### Specifications

- Version:
- Platform:
- Subsystem:

adchia:
This issue showed up with pushing to the online store using Snowflake as an offline store. The bug being:

1001 in Snowflake is an Int64. In pandas, it is an Int32. Feast stores inferred entity types in the feature view, not the entity, and so this was causing the Snowflake quickstart's `push` call to not know that 1001 should be an int64, so it never coerces it. This resulted in it looking like `push` calls never impacted the online store

This also updates the roadmap

Fixes #3171

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -356,14 +356,10 @@ lint-go: compile-protos-go compile-go-lib
 
 # Docker
 
-build-docker: build-ci-docker build-feature-server-python-docker build-feature-server-python-aws-docker build-feature-transformation-server-docker build-feature-server-java-docker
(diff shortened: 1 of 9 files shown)


felixwang9817 (commented):
good catch! should we add a test for this?

sfc-gh-madkins:
Verified on snowflake template

kevjumba (approved):

PR #3208, opened 2022-09-13
chore: Replace zod by protobufjs

jerive:

Fixes #


jerive:
Two issues pending

1. [Conversion from JSON fails for google.protobuf.Timestamp types](https://github.com/protobufjs/protobuf.js/issues/893) (kind of expected, string != [{seconds, nanos}](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/timestamp.proto)) => the UI ends up needing a protobuf binary instead of the JSON string that is exposed today by the backend (unless we find a way to do automatic conversion, which I have not found yet how to do ; we could do regex search/replace but it does not feel like the right way :slightly_smiling_face: )
2. I have not found a proto equivalent for featureViewSummaryStatistics and I am still figuring out what to do. Certainly you know better than I do

Except for these two issues, the UI works like a charm with the new protobufjs parsers

adchia:
Looks like there's still some outstanding issues in showing enum values (e.g. data source enum types)

ui/.npmrc
```diff
@@ -1 +1 @@
-//registry.npmjs.org/:_authToken=${NPM_TOKEN}
\ No newline at end of file
+# //registry.npmjs.org/:_authToken=${NPM_TOKEN}
```
adchia, line 1: woops this was me, but probably revert

sdk/python/feast/ui_server.py
```diff
@@ -1,3 +1,4 @@
+import io
```
adchia, line 1: can remove this now too

adchia (approved):
/lgtm

adchia (approved):
/lgtm

PR #3214, opened 2022-09-15
fix: Return 422 on bad push source name

### LINKED ISSUE #3210: Push api call to a push source that doesn't exist returns 200
## Expected Behavior 
When we `POST` to a push source that doesn't exist like so
```
curl -X POST "http://localhost:6566/push" -d '{
    "push_source_name": "push_source_name_that_does_not_exist",
    "df": {
            "some_entity": [1001],
            "event_timestamp": ["2022-05-13 10:59:42"],
            "created": ["2022-05-13 10:59:42"],
            ...
    },
    "to": "online_and_offline",
  }' | jq
```
I'd expect an error. It is a valid format to the correct endpoint, but is an entity that doesn't exist. A [422 unprocessable entity](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) error would probably make sense here.

## Current Behavior
Method `FeatureStore().push` receives a [`push_source_name`](https://github.com/feast-dev/feast/blob/7c50ab510633c11646b6ff04853f3f26942ad646/sdk/python/feast/feature_store.py#L1418) indicating which feature views are impacted by the new data being pushed. The set of feature views is filtered down [here](https://github.com/feast-dev/feast/blob/7c50ab510633c11646b6ff04853f3f26942ad646/sdk/python/feast/feature_store.py#L1437-L1445). If a push source name is provided that doesn't actually exist then the resulting set is empty, the loop is skipped and the function is a no-op. This successful return means that the [`/push` call in feature_server.py](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/feature_server.py#L95-L100) never throws to return an error code.

## Steps to reproduce
* start a local feature server
* make a push call to a push source that does not exist

### Specifications

- Version:
- Platform:
- Subsystem:

## Possible Solution
* Raise a custom exception when no push source is found with the provided name
* Catch that exception in the feature server
* Return 422 in that case

If this is considered reasonable, I'm happy to make the contribution myself.

robhowley:
Signed-off-by: Rob Howley <rhowley@seatgeek.com>

* added a new exception for push source not found
* in the feature store, raise the `PushSourceNotFoundException` when a push source name is provided that doesn't exist
* in feature server /push endpoint, return 422 in the case of a push source not found

Fixes #3210

### DIFF
--- a/docs/reference/online-stores/snowflake.md
+++ b/docs/reference/online-stores/snowflake.md
@@ -35,7 +35,7 @@ online_store:
 
 ## Tags KWARGs Actions:
 
-"ONLINE_PATH": Adding the "ONLINE_PATH" key to a FeatureView tags parameter allows you to choose the online table path for the online serving table (ex. "{database}"."{schema}").
+"snowflake-online-store/online_path": Adding the "snowflake-online-store/online_path" key to a FeatureView tags parameter allows you to choose the online table path for the online serving table (ex. "{database}"."{schema}").
 
 {% code title="example_config.py" %}
 ```python
--- a/sdk/python/feast/errors.py
+++ b/sdk/python/feast/errors.py
@@ -384,3 +384,8 @@ class InvalidFeaturesParameterType(Exception):
         super().__init__(
             f"Invalid `features` parameter type {type(features)}. Expected one of List[str] and FeatureService."
         )
+
+
+class PushSourceNotFoundException(Exception):
+    def __init__(self, push_source_name: str):
+        super().__init__(f"Unable to find push source '{push_source_name}'.")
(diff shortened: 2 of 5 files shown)


adchia (approved):
/lgtm

robhowley:
im a little confused. i signed the commits and DCO check agreed. not sure what block is referring to when it says, "Merging is blocked. The base branch requires all commits to be signed."

adchia:
Yeah that's just a red herring. Your pushed changes removed the "lgtm" tag so it couldn't merge

adchia (approved):
/lgtm

robhowley:
converted to draft while fix integration test

robhowley:
hey @adchia, i finally got around to fixing the integration test. looks like everything passed. thanks for your patience.

adchia (approved):
/lgtm

PR #3219, opened 2022-09-15
feat: Add ability to give boto extra args for registry config

mzwiessele:
Give the ability to add extra arguments to the boto3 put_object method when using the S3 registry.
We use this feature to enable `ServerSideEncryption=AES256` if required by our buckets.
For example: `feature_store.yaml`:
```
...
registry:
    path: s3://my-bucket/registry.db
    boto_extra_args:
        ServerSideEncryption: AES256
...
```

N/A


felixwang9817 (approved):
/lgtm

felixwang9817:
thanks for the PR @mzwiessele!

PR #3231, opened 2022-09-17
feat: Add possibility to define feature_store.yaml path with env variable

arcabucero:
Signed-off-by: arcabucero <arcabucero@gmail.com>

Fixes #

### DIFF
--- a/sdk/python/feast/cli.py
+++ b/sdk/python/feast/cli.py
@@ -99,7 +99,7 @@ def cli(
     ctx.obj["FS_YAML_FILE"] = (
         Path(feature_store_yaml).absolute()
         if feature_store_yaml
-        else ctx.obj["CHDIR"] / "feature_store.yaml"
+        else utils.get_default_yaml_file_path(ctx.obj["CHDIR"])
     )
     try:
         level = getattr(logging, log_level.upper())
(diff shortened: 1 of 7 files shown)


sdk/python/feast/cli.py
```diff
@@ -99,7 +99,7 @@ def cli(
     ctx.obj["FS_YAML_FILE"] = (
         Path(feature_store_yaml).absolute()
         if feature_store_yaml
-        else ctx.obj["CHDIR"] / "feature_store.yaml"
+        else utils.get_default_yaml_file_path(ctx.obj["CHDIR"])
```
felixwang9817, line 102: nice! I think you're missing a few spots in the codebase where you could also introduce this change

if you search for '"feature_store.yaml"' you should fine all of them, e.g. there's one in `infra/aws.py`

felixwang9817: also do you mind adding a test? I would take a look at `test_cli.py`, specifically you should be able to modify `test_3rd_party_registry_store_with_fs_yaml_override` to test your change pretty easily!

felixwang9817 (commented):
thanks for the contribution @arcabucero, looks solid overall!

sdk/python/feast/cli.py
arcabucero: @felixwang9817, thank you for your suggestions,
I've also introduced changes in aws.py and lambda_engine.py and added one test into test.cli.py

felixwang9817 (approved):
/lgtm

PR #3232, opened 2022-09-19
fix: Fix handling of TTL in Go server

wdhorton:
According to the FeatureView docstring, 'A ttl of 0 indicates that this group of features lives forever.' However, the Go feature server doesn't currently respect this, returning OUTSIDE_MAX_AGE for FeatureViews that have a TTL of 0. This fixes that issue by always returning false for checkOutsideTtl if the TTL is 0.


adchia (approved):
/lgtm

achals (approved):
/lgtm

Issue #3235, opened 2022-09-20
Test loading S3 parquet folder in dataset format


mzwiessele:
Test that a parquet dataset saved in dataset format can be loaded and accessed using `get_historical_features`.

Steps:

- upload a parquet dataset to the test s3 folder
- point a source to the parquet dataset s3 folder (**not the file**)
- load the source features using `get_historical_features`

> Ideally, there'd be a test for this running, but you'd be blocked on making a file within the Feast integration tests S3 bucket. Mind making an issue for that separately? @mzwiessele

_Originally posted by @adchia in https://github.com/feast-dev/feast/issues/3217#issuecomment-1249562980_

felixwang9817:
I think there are probably two ways to do this:
* there is an `S3FileDataSourceCreator` in `sdk/python/tests/integration/universal/feature_repos/universal/data_sources/file.py`, which uses `minio` - it hasn't been used in a while, but it could be updated and used
* instead of using `minio`, we could also just create a dataset and upload data directly to an `S3` bucket - we could just use the `feast-integration-tests` S3 bucket, which is used in `universal/data_sources/redshift.py`

PR #3236, opened 2022-09-20
fix: Dask zero division error if parquet dataset has only one partition

mzwiessele:
Signed-off-by: Max Zwiessele <ibinbei@gmail.com>

When loading data in parquet dataset format it can happen that the dataset only has one partition in the `event_timestamp` column. If that is the case, dask will fail to process the dataset, erroring with a `ZeroDividionError` similar to 

https://github.com/feast-dev/feast/blob/769c31869eb8d9bb693f8a2876cc68b8cdd16521/sdk/python/feast/infra/offline_stores/file.py#L327

This PR adds a `try-catch` block to gracefully circumvent the error and process the data in only one partition.

N/A

### DIFF
--- a/sdk/python/feast/infra/offline_stores/file.py
+++ b/sdk/python/feast/infra/offline_stores/file.py
@@ -662,14 +662,33 @@ def _drop_duplicates(
     created_timestamp_column: str,
     entity_df_event_timestamp_col: str,
 ) -> dd.DataFrame:
-    if created_timestamp_column:
-        df_to_join = df_to_join.sort_values(
-            by=created_timestamp_column, na_position="first"
-        )
+    column_order = df_to_join.columns
+
+    # try-catch block is added to deal with this issue https://github.com/dask/dask/issues/8939.
+    # TODO(kevjumba): remove try catch when fix is merged upstream in Dask.
+    try:
+        if created_timestamp_column:
+            df_to_join = df_to_join.sort_values(
+                by=created_timestamp_column, na_position="first"
+            )
+            df_to_join = df_to_join.persist()
+
+        df_to_join = df_to_join.sort_values(by=timestamp_field, na_position="first")
         df_to_join = df_to_join.persist()
 
-    df_to_join = df_to_join.sort_values(by=timestamp_field, na_position="first")
-    df_to_join = df_to_join.persist()
+    except ZeroDivisionError:
+        # Use 1 partition to get around case where everything in timestamp column is the same so the partition algorithm doesn't
+        # try to divide by zero.
+        if created_timestamp_column:
+            df_to_join = df_to_join[column_order].sort_values(
+                by=created_timestamp_column, na_position="first", npartitions=1
+            )
+            df_to_join = df_to_join.persist()
+
+        df_to_join = df_to_join[column_order].sort_values(
+            by=timestamp_field, na_position="first", npartitions=1
+        )
+        df_to_join = df_to_join.persist()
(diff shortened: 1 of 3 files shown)


sdk/python/feast/infra/offline_stores/file.py
```diff
@@ -662,14 +662,33 @@ def _drop_duplicates(
     created_timestamp_column: str,
     entity_df_event_timestamp_col: str,
 ) -> dd.DataFrame:
-    if created_timestamp_column:
-        df_to_join = df_to_join.sort_values(
-            by=created_timestamp_column, na_position="first"
-        )
+    column_order = df_to_join.columns
```
mzwiessele, line 665: Needed as during `ZeroDivisionError` the column order changes.

achals (approved):
Could you add a test for this case?

mzwiessele:
> Could you add a test for this case?
Yes, I need help though. We need a test folder with a parquet dataset in the test S3 bucket. That parquet dataset must have only one partition in the `event_timestamp` column. The reason is simply that this code correction fixes exactly that edge case.

mzwiessele:
Before merging this: Is it possible to update the Dask version feast relies on? Or in other words, why is the version restricted like this?
```
"dask>=2021.*,<2022.02.0"
```

mzwiessele:
@achals I'll need help with the tests. Happy to do the ground work. Please point me to the right testing suite (in the unit tests) for loading a local parquet file from a FileSource.

mzwiessele:
@achals I've added the test, please take a look and tell me if this does the trick!

felixwang9817:
hey @mzwiessele thanks for the PR! a few responses to your comments above:

> Yes, I need help though. We need a test folder with a parquet dataset in the test S3 bucket. That parquet dataset must have only one partition in the event_timestamp column. The reason is simply that this code correction fixes exactly that edge case.

unless I'm misunderstanding, there's no need for the parquet dataset to be in an S3 bucket, right? the only thing that matters is that the Parquet dataset doesn't have `npartitions` set

if that's correct, I think the best way to write a test would be to do it locally - you can check out our unit tests in `sdk/python/feast/infra/tests/unit`, all of which run locally. I would try to imitate `test_e2e_local` in `test_e2e_local.py` - it's a good example of setting up a dataset locally, constructing a feature repo + `FeatureStore` object, and then running some tests, in your case I think you would just want to check that `get_historical_features` works on a Parquet dataset with `npartitions` unset

> Before merging this: Is it possible to update the Dask version feast relies on? Or in other words, why is the version restricted like this?

I forget exactly why we restricted that version; I'll go back and check, and if there's no strong reason I'm happy to bump up the upper bound restriction (although I think that can happen in a follow up PR)

felixwang9817:
@mzwiessele also left some additional comments for you on #3235!

mzwiessele:
@felixwang9817 I've added the dataset source to the test suite as suggested at #3235 to test loading of parquet datasets. 

Do you know if any of the created datasets feature only one partition in the event_timestamp column?

mzwiessele:
kind ping :)

achals:
> @felixwang9817 @achals kind ping :)

The DCO check still seems to be failing! Good to merge as soon as that's fixed. https://github.com/feast-dev/feast/pull/3236/checks?check_run_id=8744237672

mzwiessele:
@achals fixed :)

achals (approved):
/lgtm

PR #3223, opened 2022-09-15
fix: Feature-server image is missing mysql dependency for mysql registry

ptran32:
Signed-off-by: 11425176+ptran32@users.noreply.github.com <11425176+ptran32@users.noreply.github.com>

Author: 11425176+ptran32@users.noreply.github.com <11425176+ptran32@users.noreply.github.com>

What this PR does / why we need it:
This PR is need to fix feast-feature-server (using helm chart) that can't connect to mysql as registry. (see issue below for more details)

* Add mysqlclient in order to use mysql as registry
* Add it to requirements.txt and also moved pip install feast in it
Which issue(s) this PR fixes:

Fixes [#3218](https://github.com/feast-dev/feast/issues/3218)

PR #3239, opened 2022-09-21
fix: Build dockerfile correctly

felixwang9817:

**What this PR does / why we need it**: #3223 introduced this `requirements.txt`. However, `docker build` copies files from the root directory, so the correct command would be `RUN pip install -r sdk/python/feast/infra/feature_servers/multicloud/requirements.txt` instead of `RUN pip install -r requirements.txt`. Instead of running this command, I decided to just remove the `requirements.txt` file and install the dependencies directly.

This bug is the reason why the Docker images were not built correctly during the latest 0.25.0 release.

Fixes #

### DIFF
--- a/sdk/python/feast/infra/feature_servers/multicloud/Dockerfile
+++ b/sdk/python/feast/infra/feature_servers/multicloud/Dockerfile
@@ -8,9 +8,8 @@ RUN apt update && \
         build-essential
 
 RUN pip install pip --upgrade
-COPY . .
+RUN pip install "feast[aws,gcp,snowflake,redis,go,mysql]"
 
-RUN pip install -r requirements.txt
 RUN apt update
 RUN apt install -y -V ca-certificates lsb-release wget
 RUN wget https://apache.jfrog.io/artifactory/arrow/$(lsb_release --id --short | tr 'A-Z' 'a-z')/apache-arrow-apt-source-latest-$(lsb_release --codename --short).deb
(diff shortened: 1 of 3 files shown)


sdk/python/feast/infra/feature_servers/multicloud/Dockerfile
```diff
@@ -8,9 +8,8 @@ RUN apt update && \
         build-essential
 
 RUN pip install pip --upgrade
-COPY . .
+RUN pip install "feast[aws,gcp,snowflake,redis,go,mysql]"
```
achals, line 11: Hmmm won't this install the library from pypi, as opposed to a locally built wheel? Can that cause inconsistencies?

felixwang9817: fixed by restoring `COPY . .`

sdk/python/feast/infra/feature_servers/multicloud/Dockerfile
```diff
@@ -9,8 +9,8 @@ RUN apt update && \
 
 RUN pip install pip --upgrade
 COPY . .
+RUN pip install ".feast[aws,gcp,snowflake,redis,go,mysql]"
```
achals, line 12: Do you need the period as well as `feast`?

felixwang9817: good catch

achals (approved):
/lgtm

PR #3241, opened 2022-09-21
fix: Fix numpy version specification

acmarco:

The install script for version 0.22 creates conflicts with other packages over the version of numpy. It looks like the version specification is wrong as it is unusual to see redundant constraints like what is there now (i.e: <1.22 _and_ also <2). 

Fixes #

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ REQUIRED = [
     "Jinja2>=2,<4",
     "jsonschema",
     "mmh3",
-    "numpy<1.22,<2",
+    "numpy>1.22,<2",
     "pandas>=1,<2",
     "pandavro==1.5.*",
     "protobuf>3.20,<4",


achals:
🤦🏽  Thanks for the catch @acmarco . Do you mind signing your commit?

acmarco:
> 🤦🏽 Thanks for the catch @acmarco . Do you mind signing your commit?

My commit is signed (i.e: "This commit was signed with the committer’s verified signature. ") ... but maybe you mean another type of signature. Sorry if I am misunderstanding.

felixwang9817:
hey @acmarco, thanks for the PR! here's what we mean by [signing your commits](https://docs.feast.dev/project/development-guide#signing-off-commits)

also, is there a particular reason you're using Feast 0.22 instead of Feast 0.24?

acmarco:
> hey @acmarco, thanks for the PR! here's what we mean by [signing your commits](https://docs.feast.dev/project/development-guide#signing-off-commits)
> 
> also, is there a particular reason you're using Feast 0.22 instead of Feast 0.24?

Ah! OK. Sorry for the confusion. signed off.

The reason we are stuck on 0.22 is not Feast related, but that we install feast in same python env as other packages that require older versions of numpy and pandas. Upgrading other packages or re-jigging our envs would unblock upgrading Feast, but is a bit complicated right now. Definitely on our roadmap to upgrade Feast and the other packages we use.

felixwang9817:
@acmarco got it, that makes sense. I'll approve this PR and cut a patch release (0.22.5) once it's in!

felixwang9817 (approved):
/lgtm

acmarco:
> I'll approve this PR and cut a patch release (0.22.5) once it's in!

thanks for taking the time to address this PR. I was just wondering when the 0.22.5 patch might become available. thanks!

adchia:
Let me cherrypick this right now and release!

adchia:
Actually this PR doesn't make sense for specifically 0.22. We dropped Python 3.7 support in 0.23 (numpy 1.22 also drops this), so this results in a failed release.

acmarco:
> Actually this PR doesn't make sense for specifically 0.22. We dropped Python 3.7 support in 0.23 (numpy 1.22 also drops this), so this results in a failed release.

Sorry, i don't follow, but I just want to confirm that a it will _not_ be possible to publish a 0.22.5 patch release with the changes from this PR. If not, is there any easy way to remedy that on my end? If there is no way, I'll consider this matter closed.

thanks

adchia:
Yeah it will not be feasible. It's feasible to cherry-pick this into any minor release afterwards (i.e. 0.23+)

PR #3245, opened 2022-09-23
fix: Updated ui_dir server call to make the directory if missing

franciscojavierarceo:
This updates the `feast ui` command to create the `ui/build/` folder if it does not exist. Without this, the `feast ui` will fail.

Fixes #

### DIFF
--- a/sdk/python/feast/ui_server.py
+++ b/sdk/python/feast/ui_server.py
@@ -1,3 +1,4 @@
+import os
 import json
 import threading
 from typing import Callable, Optional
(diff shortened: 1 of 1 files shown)


adchia:
have you tested this locally? 

if there's no directory that means the built ui resources don't exist so it wouldn't be able to mount the html / css / js etc files.

if there is one, then there should be bundled resources. i think then this shouldn't be needed. if you're missing the directory it must be some broken build of feast or something

adchia:
I locally verified with Feast 0.25 on a fresh install that the `feast ui` command works? Closing this, but feel free to re-open if you have more details!

PR #3253, opened 2022-09-27
feat: Adding billing_project_id in BigQueryOfflineStoreConfig

vmallya-123:
Signed-off-by: “Varun <varun.mallya@tech.jago.com>

One is able to add big query billing project, this can enable to run Bigquery job in a separate project which is not the project which has the BQ feature datasets. Useful in cases when certain projects have reservations (slots) for Bigquery jobs to run.

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -19,7 +19,7 @@ import numpy as np
 import pandas as pd
 import pyarrow
 import pyarrow.parquet
-from pydantic import StrictStr
+from pydantic import StrictStr, validator
 from pydantic.typing import Literal
 from tenacity import Retrying, retry_if_exception_type, stop_after_delay, wait_fixed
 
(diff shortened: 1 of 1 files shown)


adchia:
were you able to verify this works? also what does the regular project_id do when billing_project_id is set?

vmallya-123:
@adchia yes I tried it and it works as expected. The project_id is the GCP project which has the feature view datasets, this will be the project_id which will be used by the query template etc (as it is currently being used). billing_project id is the GCP project where the big query job would actually run.

adchia:
mind fixing the lint errors and the title to conform to semantic release convention?

vmallya-123:
Thanks for fixing the lint error
Is the title better now?  Following the instructions given here

`Ensure the title of the PR matches semantic release conventions (e.g. start with feat: or fix: or ci: or chore: or docs:). Keep in mind that any PR with feat: or fix: will directly make it into the change log of a release, so make sure they are understandable!`
Please let me know if there is anything else I can do

adchia (approved):
/lgtm

PR #3254, opened 2022-09-27
fix: Use configured user in env var instead of "user" for Trino

ammarar:
Signed-off-by: ammarar <ammar.alrashed@gmail.com>

**What this PR does / why we need it**: To use the configured user specified in env var. The current way is incorrect since it shadows whatever the user provides.

Fixes #

### DIFF
--- a/sdk/python/feast/infra/offline_stores/contrib/trino_offline_store/trino.py
+++ b/sdk/python/feast/infra/offline_stores/contrib/trino_offline_store/trino.py
@@ -157,7 +157,7 @@ class TrinoOfflineStore(OfflineStore):
         created_timestamp_column: Optional[str],
         start_date: datetime,
         end_date: datetime,
-        user: str = "user",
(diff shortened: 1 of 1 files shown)


adchia (commented):
This doesn't seem to use a given environment variable? Also does it make sense to have this instead in TrinoOfflineStoreConfig?

ammarar:
@adchia the environment variable will be evaluated later in [here](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/offline_stores/contrib/trino_offline_store/trino_queries.py#L44). My actual preference in general is maybe to have multiple level of evaluation of these configurations. It can be in `TrinoOfflineStoreConfig`, if it doesn't find it, it can look into an env var. if it doesn't find it, it can use default. However, I haven't seen that pattern in the code, thats why I did it this way.

adchia (approved):
/lgtm

PR #3255, opened 2022-09-27
fix: Configuration to stop coercion of tz for entity_df

ammarar:
Signed-off-by: ammarar <ammar.alrashed@gmail.com>

**What this PR does / why we need it**: Currently, for `get_historical_features` coerces `entity_df` datetime columns to have timezone if the columns don't have timezone. The problem this coercion causes for Trino offline specifically is that if the `entity_df` passed does not have timezone, the Trino provider converts it to `timestamp(3) with time zone`.  However, it should be `timestamp(3)` instead.

So this fix is basically to give option to the user to use this functionality or not.

Fixes #


adchia (approved):
/lgtm

PR #3258, opened 2022-09-28
feat: Add tag description in Field in the Feast UI

benjamintanweihao:
Signed-off-by: Benjamin Tan <benjamintanweihao@gmail.com>

We have added descriptions via tags in Fields. While the Feast UI shows descriptions for Feature Views, this is not so for Fields.


adchia (approved):
/lgtm

PR #3274, opened 2022-10-06
fix: Fix UI build

jerive:
[Zod to protobuf ](https://github.com/feast-dev/feast/pull/3208) passed but UI build was failing. This fixes it.

Fixes #3273


kevjumba:
@jerive Is this different than danny's changes?

jerive:
We did it in parallel and in the end Danny won :)
I will close this one.

PR #3289, opened 2022-10-13
Update README.md

petersutter5:
When you run 'feast init my_feature_repo', it makes a 'my_feature_repo' directory. However the 'feature_store.yaml' file is within another directory called 'feature_repo'. So the feast apply command gives an error if you are simply in the 'my_feature_store' directory.

Fixes #

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ pip install feast
 ### 2. Create a feature repository
 ```commandline
 feast init my_feature_repo
-cd my_feature_repo
+cd my_feature_repo/feature_repo
 ```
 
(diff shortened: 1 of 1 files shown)


petersutter5:
When you run 'feast init my_feature_repo', it makes a 'my_feature_repo' directory. However the 'feature_store.yaml' file is within another directory called 'feature_repo'. So the feast apply command gives an error if you are simply in the 'my_feature_store' directory.

Which issue(s) this PR fixes:
This fixes a minor issue in the README with the install instructions

Fixes #

PR #3323, opened 2022-10-30
chore: Provide related exceptions for empty data frame from entity sql execution in spark sql query execution.

amithadiraju1694:
Executing spark sql statements from template data results in failed execution,giving un-related errors while root cause is empty data frame. These changes gives user better context around the error and provides guidance on further step.

1. Added spark utils to check for empty spark data frames and successful spark sql query executions ( with at least 1 row ).
2. Added checks in spark.py at appropriate locations

Signed-off-by: amithadiraju1694 <amith.adiraju@gmail.com>

Label: `kind/housekeeping` and `kind/bug`

This PR provides meaningful errors to developers when data frame returned is empty when executing spark sql statements, hence providing better guidance on how to solve the underlying issue.

Partially Fixes: #3324


amithadiraju1694:
@DvirDukhan  Any update on this would be highly appreciated :)

.gitignore
```diff
@@ -224,3 +224,4 @@ ui/.vercel
 
 # Go subprocess binaries (built during feast pip package building)
 sdk/python/feast/binaries/
+=4.0.30
```
kevjumba, line 227: What is this for?

sdk/python/feast/infra/utils/spark_utils.py
```diff
@@ -0,0 +1,24 @@
+from pyspark.sql import DataFrame
+
+from feast.errors import EntitySQLEmptyResults
+
+
+def _assert_empty_sparkDF(df: DataFrame) -> bool:
+
+    """Asserts if a pyspark data frame is empty or not."""
+
+    if df.rdd.isEmpty():
```
kevjumba, line 10: do you need to check if the df is null?

.gitignore
amithadiraju1694: Not sure why this strange file popped up in my folders, was installing a package in a different window of vscode of this version, and this file started to show up here. 

This file is unrelated and not required for this PR, and hence in gitignore.

sdk/python/feast/infra/utils/spark_utils.py
amithadiraju1694: in my POV, it's better practice to check if resulting data frame from executing sql query is empty, without this empty data frame check , user wouldn't know if their sql statement resulted in any data.

This check is optimal, since it's not performing any action, per se.

Users will see an un-related error, much further in execution ( as this entity sql returns NoneType, down stream code breaks with un-related errors as I highlighted in issue: #3324 )

amithadiraju1694:
@kevjumba  Any changes you'd suggest for this ?

sdk/python/feast/infra/utils/spark_utils.py
kevjumba: ah I see that makes sense

.gitignore
achals: I think you have a bad file in your local workspace; can you delete it from there instead of adding it to the project's .gitignore?

sdk/python/.gitignore
```diff
@@ -116,3 +116,4 @@ dmypy.json
 
 .vscode/*
 playground
+=4.0.30
```
achals, line 119: same comment

sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
```diff
@@ -449,7 +449,13 @@ def _get_entity_df_event_timestamp_range(
         # If the entity_df is a string (SQL query), determine range
         # from table
         df = spark_session.sql(entity_df).select(entity_df_event_timestamp_col)
+
+        # Checks if resulting data frame has any data, if not
+        # raises a meaningful error
+        spark_utils._assert_query_success(df, entity_df)
```
achals, line 455: Not sure if these these utility methods should be in a different file - they're pretty small. Can't they just be defined in this same file?

sdk/python/.gitignore
amithadiraju1694: makes sense, will do.

sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
amithadiraju1694: > Not sure if these these utility methods should be in a different file - they're pretty small. Can't they just be defined in this same file?

Just following the existing project structure for this. Didn't want to overload `spark.py` ; may be, as feast-spark matures from contrib, this structure may be useful  ?

achals: No i think we don't need a new file just for this, we can add it later if the need to organize code differently arises

amithadiraju1694:
@achals Made changes you suggested and rebased my branch as well. Please review.

achals (approved):

achals (commented):
/lgtm

felixwang9817 (approved):
/lgtm

PR #2887, opened 2022-06-29
fix: Change numpy version on setup.py and upgrade it to resolve dependabot warning

breno-costa:
**What this PR does / why we need it**:
Dependabot cannot update numpy to a non-vulnerable version because the numpy version is pinned on setup.py, and it looks like numpy version was wrongly pinned by a recent change introduced on PR https://github.com/feast-dev/feast/pull/2647

**Which issue(s) this PR fixes**:

Fixes #

PR #3343, opened 2022-11-09
chore: Remove upper bound on dask dependency

oliverholworthy:

_Motivation_: Being able to use feast in an application with more recent versions of dask in a Python 3.8+ environment without having to work around conflicts due to versions specified in package.

This PR removes the upper bound on dask requirement added in #2342 This upper bound was added because newer versions of dask dropped support for Python 3.7. 

However, **the upper bound isn't necesesary for Python 3.7 compatibility (unless a version of dask was released that had the wrong `python_requires` specified.)**
  - pip install dask (or "dask>=2021.*") in a Python 3.7 environment installs dask==2022.2.0 which was the last version that specfied comaptitibility with Python 3.7 (with `python_requires` in the pacakge config)
  

_Additional context_

It seems that in practice feast has dropped support for Python 3.7:
- CI tests are no longer run for 3.7 https://github.com/feast-dev/feast/pull/2810
  - So any incompatibility introduced for 3.7 would be unlikely to be caught
  - Example: the latest version of feast specifies ["numpy>=1.22,<3", "pandas>=1.4.3,<2",](https://github.com/feast-dev/feast/blob/v0.26.0/setup.py#L60-L61). Both of these require Python 3.8. Which means a pip install feast==0.26.0 won't work in a Python 3.7 environment. numpy version was updated most recently in https://github.com/feast-dev/feast/pull/2887

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,7 +74,7 @@ REQUIRED = [
     "typeguard",
     "fastapi>=0.68.0,<1",
     "uvicorn[standard]>=0.14.0,<1",
-    "dask>=2021.*,<2022.02.0",
+    "dask>=2021.*",
     "bowler",  # Needed for automatic repo upgrades
 ]
 


beauchbum:
Just bumped into this conflict today when trying to add feast as a dependency in our model-training repo. Thanks for opening a PR to address this.

achals (approved):

PR #3356, opened 2022-11-21
feat: Cassandra online store, concurrent fetching for multiple entities

hemidactylus:
This changes the retrieval of features from the Cassandra online store by leveraging the
Cassandra driver's native concurrency capabilities.

When there are several entities to be retrieved, instead of a sequential read one-by-one, entity after entity,
the reads are executed concurrently, with the driver ensuring the results are kept in the correct order and the call
returns when all results are available.
This, as measured in realistic environments, implies a speedup of 2-3x for retrieval of 20 to 100 entities at once.

Using the Cassandra driver's `execute_concurrent_with_args` function requires a new parameter controlling the maximum amount of concurrency to use (somewhat bounded by the number of vCPUs at hand): for transparency, this is exposed in the feature store configuration yaml as a new parameter, which is documented and correctly handled by the guided procedure of `feast init -t cassandra`.


adchia (approved):
/lgtm

PR #3358, opened 2022-11-21
feat: Get Snowflake Query Output As Pyspark Dataframe (#2504)

### LINKED ISSUE #2504: Get_historical_features() Does Not Have Option To Return Distributed Dataframe Like A Spark DF
When doing get_historical_features().to_df on a large training dataset in databricks I am hitting memory full errors. Since to_df is returning the data as a pandas dataframe it is not able to use the full capacity of the databricks cluster and distribute it across the different nodes the way a spark dataframe might.

amithadiraju1694:
1. Added feature to offline_store-> snowflake.py to return results of snowflake query as pyspark data frame.This helps spark-based users to distribute data, which often doesn't fit in driver nodes through pandas output.

2. Also added relevant error class, to notify user on missing spark session , particular to this usecase.

Signed-off-by: amithadiraju1694 <amith.adiraju@gmail.com>

This adds feature to `SnowflakeRetrievalJob` to return result of snowflake query execution as a pyspark data frame. 

Fixes #2504

### DIFF
--- a/sdk/python/feast/infra/offline_stores/snowflake.py
+++ b/sdk/python/feast/infra/offline_stores/snowflake.py
@@ -1,7 +1,9 @@
 import contextlib
 import os
 import uuid
+import warnings
 from datetime import datetime
+from functools import reduce
 from pathlib import Path
 from typing import (
     Any,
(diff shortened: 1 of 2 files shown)


sdk/python/feast/infra/offline_stores/snowflake.py
```diff
@@ -447,6 +459,51 @@ def to_sql(self) -> str:
         with self._query_generator() as query:
             return query
 
+    def to_spark_df(
+        self, spark_session: Optional[SparkSession] = None
+    ) -> pyspark_DataFrame:
```
sfc-gh-madkins, line 464: @amithadiraju1694 can you use just DataFrame here? Not the alias

sfc-gh-madkins:
@amithadiraju1694 can you run make lint-python

adchia (approved):
/lgtm

PR #3363, opened 2022-11-28
feat: Add Snowflake Registry

### LINKED ISSUE #3086: Add Snowflake as a registry

**Additional context**

sfc-gh-madkins:

Fixes #3086


sfc-gh-madkins:
@adchia any idea why this check wont kick off?

adchia (commented):
Is there a reason you aren't using the regular SQL registry? It should work fine since SQLAlchemy works with Snowflake right?

sfc-gh-madkins:
> Is there a reason you aren't using the regular SQL registry? It should work fine since SQLAlchemy works with Snowflake right?

@adchia

1.) we have a lot of different and evolving authentication methods and didnt want to put any dependencies on 3rd party packages
2.) I forsee a point in time where a Snowflake Provider is built that adds additional tables, columns to the registry for its operations. A good example of this is determining whether to skip a materialization if the tables last commit time hasnt changed. That Snowflake table commit time would need to be stored in a new column in the FV registry table for example

sdk/python/feast/infra/registry/snowflake.py
```diff
@@ -0,0 +1,1083 @@
...
+        self.cached_registry_proto = self.proto()
+        self.cached_registry_proto_created = datetime.utcnow()
+        self._refresh_lock = Lock()
+        self.cached_registry_proto_ttl = timedelta(
+            seconds=registry_config.cache_ttl_seconds
+            if registry_config.cache_ttl_seconds is not None
+            else 0
+        )
+
+    def refresh(self):
```
sfc-gh-rdervay, line 148: This is where my issue was for some reason if i gave it an arbitrary second argument feast apply succeeded

sfc-gh-madkins: @sfc-gh-rdervay Im not quite understanding here -- is this about the TTL?

sdk/python/feast/feature_store.py
```diff
@@ -80,6 +80,7 @@
 from feast.infra.provider import Provider, RetrievalJob, get_provider
 from feast.infra.registry.base_registry import BaseRegistry
 from feast.infra.registry.registry import Registry
+from feast.infra.registry.snowflake import SnowflakeRegistry
```
achals, line 83: I'd move this import to line 172 right above `self._registry = SnowflakeRegistry(...)` so that it's not necessary for users to have installed the `snowflake` extra.

sdk/python/feast/repo_config.py
```diff
@@ -175,6 +180,12 @@ class RepoConfig(FeastBaseModel):
     def __init__(self, **data: Any):
         super().__init__(**data)
 
+        if "registry" in data and "type" in data["registry"]:
+            if data["registry"]["type"] == "snowflake.registry":
+                self.registry = get_registry_config_from_type(data["registry"]["type"])(
+                    **data["registry"]
+                )
```
achals, line 187: Sorry I'm not sure I understand this - does this mean that the repo config object is also going to have a reference to the registry, but only in case of snowflake? What about other types? And is there a reason why this reference needs to exist here specifically? Seems like it isn't being used anywhere?

adchia: I refactored this so they're all consistent

PR #3367, opened 2022-11-30
feat: Cassandra online store, concurrency in bulk write operations

hemidactylus:
This implements the counterpart of #3356 but for writing. Using the native concurrency offered by Cassandra drivers allows for much faster write operations to the online store, which is crucial especially in the Materialize phase.

On a realistic setup (EC2 instance running the materialization on a DB in the same region), speedups of a factor about 12x are achieved.

Similarly to the read-optimization mentioned above, a new `write_concurrency` parameter is introduced (with defaults and full backward-compatibility) to control the level of concurrency should it ever be needed (the defaults should be fine in all cases, anyway).

In order to preserve the behaviour of the callbacks to `progress` during writes to the online store, which makes the progress bar behave correctly, in the function `online_write_batch` an ad-hoc iterator is built (see `unroll_insertion_tuples`) which, while it unfolds the whole set of rows to write, takes the care of invoking `progress` once per entity (each entity in general entails multiple rows to the DB table).

The documentation and the guided `feast init -t cassandra` procedure are also updated to reflect this.

### DIFF
--- a/docs/reference/online-stores/cassandra.md
+++ b/docs/reference/online-stores/cassandra.md
@@ -33,6 +33,7 @@ online_store:
         local_dc: 'datacenter1'                                             # optional
         load_balancing_policy: 'TokenAwarePolicy(DCAwareRoundRobinPolicy)'  # optional
     read_concurrency: 100                                                   # optional
+    write_concurrency: 100                                                  # optional
 ```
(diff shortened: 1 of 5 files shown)


adchia (approved):
/lgtm

PR #3368, opened 2022-11-30
fix: Update bytewax materialization

whoahbot:

This PR addresses an issue with Bytewax materialization, updates to a more recent version of Bytewax, and removes the creation of services in k8s when running a batch materialization job.


adchia (approved):
/lgtm

PR #3371, opened 2022-12-02
fix: Fix memory leak from usage.py not properly cleaning up call stack

adchia:

Fixes #


kevjumba (approved):
/lgtm

PR #3379, opened 2022-12-05
fix: Fix workflow to contain env vars

kevjumba:

Github release action failed bc SEMVER version environment variables don't exist.
Before
```
publish-web-ui-npm:
    if: github.repository == 'feast-dev/feast'
    needs: [validate_version_bumps]
    runs-on: ubuntu-latest
    env:
      # This publish is working using an NPM automation token to bypass 2FA
      NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
      CURRENT_VERSION: ${{ needs.get_dry_release_versions.outputs.current_version }}
      NEXT_VERSION: ${{ needs.get_dry_release_versions.outputs.next_version }}
```

`get_dry_release_versions` is not imported even though `validate_version_bumps` depends on that workflow. It's possible that importing outputs that a workflow depends on doesn't import the underlying output vars so I just direct imported it so that we can see the outputs for sure.

Fixes #

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -87,7 +87,7 @@ jobs:
 
   publish-web-ui-npm:
     if: github.repository == 'feast-dev/feast'
-    needs: validate_version_bumps
+    needs: [validate_version_bumps, get_dry_release_versions]
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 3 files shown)


achals (approved):
/lgtm

PR #3386, opened 2022-12-11
feat: Support postgres DSNs with alternate scheme

khrisrichardson:
Signed-off-by: Khris Richardson <khris.richardson@gmail.com>

First of all my apologies for the lengthy commit, and I understand if this is not something you wish to take on in its entirety, and am okay with refactoring or maintaining a fork to retain this functionality. A play in six acts:

ACT I

It started with a simple attempt to connect feast to a `cockroachdb` registry and offline store. While the offline store worked as-is, the registry was a different story, since `sqlalchemy` failed to detect the database version without also updating the DSN scheme from `postgresql` to `cockroachdb` in order to speak that dialect.

It was in light of the aforementioned that `sdk/python/feast/infra/utils/postgres/connection_utils.py` was updated to support `dsn` as one of the keyword arguments, since it seems that `scheme` is not one of the fields supported by the `psycopg2.connect` method.

ACT II

Updating the `_get_conn` method in turn required updating the `PostgresRegistryConfig` class in `sdk/python/feast/infra/registry/contrib/postgres/postgres_registry_store.py`. In doing so, I made the `password` attribute optional, since it's not technically required, nor is it encouraged when instead one could use private key authentication. I also set the default `port` and `db_schema` attributes consistent with the `PostgreSQLConfig` class in `sdk/python/feast/infra/utils/postgres/postgres_config.py`, wherein I also made the `password` attribute optional.

ACT III

At this point I made a decision to update the tests with a `cockroachdb` flavor, while I could have simply used `postgresql+psycopg2` as an alternative scheme. If `cockroachdb` isn't an integration you care to assume for whatever reason, I'd be happy to simplify this PR accordingly.

Since the main impetus behind this PR was to allow one to update the DSN for the registry to be compliant with `sqlalchemy`, tests were added in `sdk/python/tests/unit/test_sql_registry.py` for `cockroachdb`. A new `pytest` fixture was added for `cockroachdb`  in the likeness of `postgresql`, using public key authentication instead of password authentication, and the test fixture was added to all the tests. As can be observed on line 144, the DSN is set to `path=f"cockroachdb://{COCKROACH_USER}@127.0.0.1:{container_port}/{COCKROACH_DATABASE}?{params}"`, thus exercising the updates to the aforementioned `infra` modules.

ACT IV

In light of the updates to the registry, the `docs/tutorials/using-scalable-registry.md` document was updated to reflect how to make this change work in one's environment, whether from a standalone or containerized perspective. Namely this entails installing the appropriate supporting `sqlalchemy` module. In keeping with the other changes, I made this update specific to `cockroachdb`, but can make it more generic if so desired. There were not similar updates made to the `postgres` offline and online store markdown documents, given that the `sqlalchemy` requirement only impacted the registry.

ACT V

While the `_get_conn` method is invoked by the `postgres` offline and online stores in the `sdk/python/feast/infra/offline_stores/contrib/postgres_offline_store/postgres.py` and `sdk/python/feast/infra/online_stores/contrib/postgres.py` files respectively without requiring any explicit changes, test coverage should still be added in the event that one opts to use an alternate scheme. Consequently to test the DSN updates, a new `cockroachdb` test fixture and creator class were added to `sdk/python/feast/infra/offline_stores/contrib/postgres_offline_store/tests/data_source.py`. In doing so I modified the `field_mapping` attribute in the `DataSource` to reflect its optional nature (i,e. - it's default was set to `None`) . I also changed the signature of `create_online_store` to reflect that its output type was `FeastConfigBaseModel`, not `Dict[str, str]`.

Doing so had the after effect that the `FeastConfigBaseModel` `pydantic` class in `sdk/python/feast/repo_config.py` needed to allow extra fields. I figured it would be better to opt for accuracy than to skirt around the type safety checks and `pydantic` schema validation by using the incorrect output type. I acknowledge that there are other such test data sources for other stores that similarly output `Dict[str, str]` instead of `FeastConfigBaseModel`, but opted to leave them as-is, and suggest addressing them in a follow-up PR.

ACT VI

An alternative `postgres_repo_configuration` was added for offline store testing with `sdk/python/feast/infra/offline_stores/contrib/postgres_repo_configuration_cockroachdb.py`, and this addition was reflected in the `Makefile`, where I also addressed a handful of whitespace inconsistencies.

Similarly an alternative `postgres_repo_configuration` was added for online store testing with `sdk/python/feast/infra/online_stores/contrib/postgres_repo_configuration_cockroachdb.py`, and the `Makefile` was updated in kind.

I acknowledge that the names of these files break from the naming conventions and am open to suggestions for rectifying this.

As can be observed on lines 203 and 219 of `sdk/python/feast/infra/offline_stores/contrib/postgres_offline_store/tests/data_source.py`, the `scheme` is set to `cockroachdb` to exercise the updates to the `infra` modules, while technically the `postgresql` scheme works as-is for both the offline and online stores.

The update to the `README.md` was a result of the pre-commit hook. `¯\_(ツ)_/¯`

Fixes #

### DIFF
--- a/.github/fork_workflows/fork_pr_integration_tests_gcp.yml
+++ b/.github/fork_workflows/fork_pr_integration_tests_gcp.yml
@@ -44,7 +44,7 @@ jobs:
         with:
           go-version: 1.18.0
       - name: Set up gcloud SDK
-        uses: google-github-actions/setup-gcloud@v0
+        uses: google-github-actions/setup-gcloud@v1
         with:
           project_id: ${{ secrets.GCP_PROJECT_ID }}
           service_account_key: ${{ secrets.GCP_SA_KEY }}
--- a/.github/workflows/java_master_only.yml
+++ b/.github/workflows/java_master_only.yml
@@ -21,9 +21,9 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: 'true'
-      - uses: google-github-actions/setup-gcloud@v0
+      - uses: google-github-actions/setup-gcloud@v1
         with:
-          version: '290.0.1'
+          version: '411.0.0'
           export_default_credentials: true
           project_id: ${{ secrets.GCP_PROJECT_ID }}
           service_account_key: ${{ secrets.GCP_SA_KEY }}
@@ -150,7 +150,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-it-maven-
       - name: Set up gcloud SDK
-        uses: google-github-actions/setup-gcloud@v0
+        uses: google-github-actions/setup-gcloud@v1
         with:
           project_id: ${{ secrets.GCP_PROJECT_ID }}
           service_account_key: ${{ secrets.GCP_SA_KEY }}
--- a/.github/workflows/java_pr.yml
+++ b/.github/workflows/java_pr.yml
@@ -72,9 +72,9 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: 'true'
-      - uses: google-github-actions/setup-gcloud@v0
+      - uses: google-github-actions/setup-gcloud@v1
         with:
-          version: '290.0.1'
+          version: '411.0.0'
           export_default_credentials: true
           project_id: ${{ secrets.GCP_PROJECT_ID }}
           service_account_key: ${{ secrets.GCP_SA_KEY }}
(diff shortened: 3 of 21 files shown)


khrisrichardson:
Hello. I revisited this with fresh eyes and believe I may have kickstarted this PR under the false premise that since the registry needed the scheme updated to satisfy `sqlalchemy`, then so too did the offline/online stores. It appears to work if all one does is creates a custom image with the needed Python module(s), update the registry DSN and leave the offline/online stores well enough alone.

In other words I think this PR can be reduced to a footnote in the registry's markdown.

felixwang9817:
> In other words I think this PR can be reduced to a footnote in the registry's markdown.

got it! if this PR is no longer necessary, do you want to close it and open a new PR with the footnote you'd like to add?

Issue #3383, opened 2022-12-08
`pyspark` dependency breaks Snowflake offline store


beauchbum:
[This commit](https://github.com/feast-dev/feast/pull/3358/commits/14c082ef5c039f2b7fa261a485b66e820c56568f) introduced a `pyspark` dependency for the Snowflake offline store that is not reflected in [setup.py](https://github.com/feast-dev/feast/blob/1d3c111c4adb28b7063f45e92af0ad3aef296ee3/setup.py#L98-L104). This causes feast actions to fail due to:

```
File "/usr/local/lib/python3.9/site-packages/feast/infra/offline_stores/snowflake.py", line 26, in <module>
  from pyspark.sql import DataFrame, SparkSession
ModuleNotFoundError: No module named 'pyspark'`.
```

## Steps to reproduce
- Install `feast = {extras = ["snowflake"], version = "^0.27.0"}`
- Set up a Feature Repository with a Snowflake offline store
- Run `feast plan`

### Specifications

- Version: 0.27.0
- Platform:
- Subsystem:

## Possible Solution
Not sure - would prefer not needing `pyspark` for Snowflake offline store to work, but if that's the case then add to `setup.py`?

achals:
cc @adchia @sfc-gh-madkins  - should this dep on pyspark be optional (by moving the import inside the `to_spark_df` method, or should it be added as a proper dep to the `snowflake` extra?

adchia:
yeah should be a try import probably

beauchbum:
Thanks for weighing in on this. Any ideas on when this might be worked on (and eventually released)?

PR #3397, opened 2022-12-15
fix: Update import logic to remove `pyspark` dependency from Snowflake Offline Store

beauchbum:
* Moves the pyspark import inside of the relevant method so that the Snowflake offline store does not have a necessary dependency on pyspark
* Uses typing.TYPE_CHECKING to do the imports while type checking, but not in runtime

**Resulting behavior**
* The Snowflake Offline Store can be used without `pyspark` installed
  * However, when running an environment without `pyspark`, then calling `to_spark_df` raises a `FeastExtrasDependencyImportError`

Fixes [this issue](https://github.com/feast-dev/feast/issues/3383)

### DIFF
--- a/sdk/python/feast/infra/offline_stores/snowflake.py
+++ b/sdk/python/feast/infra/offline_stores/snowflake.py
@@ -1,5 +1,3 @@
-from __future__ import annotations
-
 import contextlib
 import os
 import uuid
@@ -461,7 +459,7 @@ class SnowflakeRetrievalJob(RetrievalJob):
         with self._query_generator() as query:
             return query
 
-    def to_spark_df(self, spark_session: SparkSession) -> DataFrame:
+    def to_spark_df(self, spark_session: "SparkSession") -> "DataFrame":
         """
         Method to convert snowflake query results to pyspark data frame.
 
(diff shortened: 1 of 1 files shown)


beauchbum:
This removes the method if `pyspark` is not installed. I tried to do the import inside of the method definition so that the method would still be available (and then raise the extras exception), but that doesn't work due to the type hinting. Then I tried using [Python's `typing.TYPE_CHECKING` to condition the import](https://stackoverflow.com/a/61394121), but that fails the unittests.

beauchbum:
Went back and forth on this a lot to get things to work. Dealing with an optional import (`pyspark`) that needs to pass mypy type checking was difficult. I found a solution from [this post](https://stackoverflow.com/a/61394121).

sdk/python/feast/infra/offline_stores/snowflake.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
beauchbum, line 1: More info on how this works to enforce postponed type evaluation

https://adamj.eu/tech/2021/05/15/python-type-hints-future-annotations/

achals (approved):
/lgtm

sfc-gh-madkins (approved):

PR #3401, opened 2022-12-19
fix: Assertion condition when value is 0

### LINKED ISSUE #3400: sampled scalar value verification failure for 0 value if float is defined
## Expected Behavior 
Regardless of whether it is 0 or 0.0, conversion for int and float types must be performed.

## Current Behavior
It is defined as a float type, but if 0 is set other than 0.0, an [assert](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/type_map.py#L405) fail occurs.

## Steps to reproduce
If 0 is set in the BigQuery table that owns the float type, it may be recognized as an int when materializing using the feast sdk.

### Specifications

- Version: 0.27.1
- Platform: regardless
- Subsystem: regardless

## Possible Solution
If the value is set to 0, the assertion is modified so that it is not related to int or float.

kysersozelee:
When materializing a feature defined as a float type, if int is 0, materialization fails.

Fixes #3400

### DIFF
--- a/sdk/python/feast/type_map.py
+++ b/sdk/python/feast/type_map.py
@@ -403,7 +403,7 @@ def _python_value_to_proto_value(
         ) = PYTHON_SCALAR_VALUE_TYPE_TO_PROTO_VALUE[feast_value_type]
         if valid_scalar_types:
             if sample == 0 or sample == 0.0:
-                # In the case of a value of 0, type validation cannot be performed with only one int or float type.
+                # Numpy convert 0 to int. However, in the feature view definition, the type of column may be a float.
                 assert type(sample) in [np.int64, np.float64, float, int]
             else:
                 assert type(sample) in valid_scalar_types
--- a/sdk/python/feast/type_map.py
+++ b/sdk/python/feast/type_map.py
@@ -403,6 +403,7 @@ def _python_value_to_proto_value(
         ) = PYTHON_SCALAR_VALUE_TYPE_TO_PROTO_VALUE[feast_value_type]
         if valid_scalar_types:
             if sample == 0 or sample == 0.0:
+                # In the case of a value of 0, type validation cannot be performed with only one int or float type.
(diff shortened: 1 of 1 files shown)


sdk/python/feast/type_map.py
```diff
@@ -402,7 +402,10 @@ def _python_value_to_proto_value(
             valid_scalar_types,
         ) = PYTHON_SCALAR_VALUE_TYPE_TO_PROTO_VALUE[feast_value_type]
         if valid_scalar_types:
-            assert type(sample) in valid_scalar_types
+            if sample == 0 or sample == 0.0:
```
felixwang9817, line 405: nit: would you mind adding a comment indicating why this special case is necessary?

felixwang9817 (approved):

felixwang9817:
hey @kysersozelee thanks for the PR! I left a small comment, and you'll have to [sign](https://docs.feast.dev/project/development-guide#signing-off-commits) your commit for the DCO check to pass - otherwise lgtm!

felixwang9817 (approved):
/lgtm

sdk/python/feast/type_map.py
felixwang9817: @kysersozelee hm I think this comment is still not super clear to me - it would be great if you could add _why_ type validation cannot be performed with only one type (if I understand correctly from #3400, the issue is because you might e.g. have an int value in a float column)

kysersozelee: @felixwang9817 

Left the two comment. Let me know if it is not inappropriate!
```python
# Numpy convert 0 to int. However, in the feature view definition, the type of column may be a float.
# So, if value is 0, type validation must pass if scalar_types are either int or float.
```

felixwang9817 (approved):
/lgtm

sdk/python/feast/type_map.py
felixwang9817: thanks @kysersozelee!

PR #3403, opened 2022-12-19
chore: Feature view as property instead to remove duplicate source of truth

gbmarc1:
Signed-off-by: gbmarc1 <marcantoine.belanger@shopify.com>

This PR replaces the `schema` attribute in the `FeatureView` object by a property. The `schema` and `entity_columns` + `features` would duplicate the source of truth internally. 
This PR exposes the `schema` as a property that is built from the `entity_columns` and `features` attributes.

Furthermore, there was a problem in the `from_proto` method were the `schema` attribute would not be assigned. This PR also avoid that gymnastic.

### DIFF
--- a/sdk/python/tests/unit/infra/test_inference_unit_tests.py
+++ b/sdk/python/tests/unit/infra/test_inference_unit_tests.py
@@ -244,8 +244,8 @@ def test_feature_view_inference_on_entity_value_types():
         ),
     )
 
-    # The schema is only used as a parameter, as is therefore not updated during inference.
-    assert len(feature_view_1.schema) == 1
+    # The schema must be entity and features
+    assert len(feature_view_1.schema) == 2
 
     # Since there is already a feature specified, additional features are not inferred.
     assert len(feature_view_1.features) == 1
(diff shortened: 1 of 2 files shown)


sdk/python/tests/unit/infra/test_inference_unit_tests.py
```diff
@@ -244,8 +244,8 @@ def test_feature_view_inference_on_entity_value_types():
         ),
     )
 
-    # The schema is only used as a parameter, as is therefore not updated during inference.
-    assert len(feature_view_1.schema) == 1
```
adchia, line 248: this seems like it'd be then potentially a backwards incompatible change?

gbmarc1: Yes I agree that the schema attribute will change behaviour. However, the comment
`# The schema is only used as a parameter, as is therefore not updated during inference.`
suggests that is should more or less have been a private attribute and that its state can be inconsistent.

I don't see a use case where someone would want to rely on the inconsistent state.

This PR tries to make this public attribute at least consistent.

felixwang9817 (approved):
/lgtm

sdk/python/tests/unit/infra/test_inference_unit_tests.py
hao-affirm: apparently this caused incompatible issue.
we use `feature_view.schema` to get feature schema, but now 
```
    def schema(self) -> List[Field]:
        return self.entity_columns + self.features
```
has duplicates.
a temp fix for us is
```
    def schema(self) -> List[Field]:
        return list(set(self.entity_columns + self.features))
```
but im not sure it's ideal or not
@gbmarc1 @adchia

felixwang9817: @hao-affirm good catch, let me put in a fix right now

PR #3406, opened 2022-12-20
fix: Batch Snowflake materialization queries to obey Snowpark 100 fea…

### LINKED ISSUE #3380: Snowflake Materialization can only handle 128 features at a time
### Specifications

- Version:
- Platform:
- Subsystem:

sfc-gh-madkins:
…ture limit

Fixes #3380

### DIFF
--- a/sdk/python/feast/infra/materialization/snowflake_engine.py
+++ b/sdk/python/feast/infra/materialization/snowflake_engine.py
@@ -276,32 +276,65 @@ class SnowflakeMaterializationEngine(BatchMaterializationEngine):
 
             fv_latest_values_sql = offline_job.to_sql()
 
+            if feature_view.entity_columns:
+                join_keys = [entity.name for entity in feature_view.entity_columns]
+                unique_entities = '"' + '", "'.join(join_keys) + '"'
+
+                query = f"""
+                    SELECT
+                        COUNT(DISTINCT {unique_entities})
+                    FROM
+                        {feature_view.batch_source.get_table_query_string()}
+                """
+
+                with GetSnowflakeConnection(self.repo_config.offline_store) as conn:
+                    entities_to_write = conn.cursor().execute(query).fetchall()[0][0]
+            else:
+                entities_to_write = (
+                    1  # entityless feature view has a placeholder entity
+                )
+
             if feature_view.batch_source.field_mapping is not None:
                 fv_latest_mapped_values_sql = _run_snowflake_field_mapping(
                     fv_latest_values_sql, feature_view.batch_source.field_mapping
                 )
 
-            fv_to_proto_sql = self.generate_snowflake_materialization_query(
-                self.repo_config,
-                fv_latest_mapped_values_sql,
-                feature_view,
-                project,
-            )
+            features_full_list = feature_view.features
+            feature_batches = [
+                features_full_list[i : i + 100]
+                for i in range(0, len(features_full_list), 100)
+            ]
 
             if self.repo_config.online_store.type == "snowflake.online":
-                self.materialize_to_snowflake_online_store(
-                    self.repo_config,
-                    fv_to_proto_sql,
-                    feature_view,
-                    project,
-                )
+                rows_to_write = entities_to_write * len(features_full_list)
             else:
-                self.materialize_to_external_online_store(
-                    self.repo_config,
-                    fv_to_proto_sql,
-                    feature_view,
-                    tqdm_builder,
-                )
+                rows_to_write = entities_to_write * len(feature_batches)
+
+            with tqdm_builder(rows_to_write) as pbar:
+                for i, feature_batch in enumerate(feature_batches):
+                    fv_to_proto_sql = self.generate_snowflake_materialization_query(
+                        self.repo_config,
+                        fv_latest_mapped_values_sql,
+                        feature_view,
+                        feature_batch,
+                        project,
+                    )
+
+                    if self.repo_config.online_store.type == "snowflake.online":
+                        self.materialize_to_snowflake_online_store(
+                            self.repo_config,
+                            fv_to_proto_sql,
+                            feature_view,
+                            project,
+                        )
+                        pbar.update(entities_to_write * len(feature_batch))
+                    else:
+                        self.materialize_to_external_online_store(
+                            self.repo_config,
+                            fv_to_proto_sql,
+                            feature_view,
+                            pbar,
+                        )
 
             return SnowflakeMaterializationJob(
                 job_id=job_id, status=MaterializationJobStatus.SUCCEEDED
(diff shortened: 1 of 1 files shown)


sdk/python/feast/infra/materialization/snowflake_engine.py
```diff
@@ -276,27 +276,35 @@ def _materialize_one(
...
-
-            if self.repo_config.online_store.type == "snowflake.online":
-                self.materialize_to_snowflake_online_store(
+            features_full_list = feature_view.features
+            feature_batches = [
+                features_full_list[i : i + 100]
+                for i in range(0, len(features_full_list), 100)
+            ]
+            for feature_batch in feature_batches:
+
```
adchia, line 285: nit: remove extra line

sdk/python/feast/infra/materialization/snowflake_engine.py
```diff
@@ -276,27 +276,35 @@ def _materialize_one(
...
-                self.repo_config,
-                fv_latest_mapped_values_sql,
-                feature_view,
-                project,
-            )
-
-            if self.repo_config.online_store.type == "snowflake.online":
-                self.materialize_to_snowflake_online_store(
+            features_full_list = feature_view.features
+            feature_batches = [
```
adchia, line 280: more idiomatic is probably to use itertools.islice

sdk/python/feast/infra/materialization/snowflake_engine.py
```diff
@@ -276,27 +276,35 @@ def _materialize_one(
...
+
+                if self.repo_config.online_store.type == "snowflake.online":
+                    self.materialize_to_snowflake_online_store(
+                        self.repo_config,
+                        fv_to_proto_sql,
+                        feature_view,
+                        project,
+                    )
+                else:
+                    self.materialize_to_external_online_store(
```
adchia, line 302: you might need to change how tqdm works here since it keeps on resetting as is for each feature batch, when you really probably want one big tqdm that increases partially for each batch of features

sfc-gh-madkins:
Ok -- this one should be good to push through

sfc-gh-madkins:
@adchia good to merge I think

sdk/python/feast/infra/materialization/snowflake_engine.py
```diff
@@ -271,32 +271,63 @@ def _materialize_one(
...
-                feature_view,
-                project,
-            )
+            features_full_list = feature_view.features
+            feature_batches = [
+                features_full_list[i : i + 100]
+                for i in range(0, len(features_full_list), 100)
+            ]
+
+            with tqdm_builder(rows_to_write) as pbar:
```
adchia, line 307: probably is rows to write * num batches

sdk/python/feast/infra/materialization/snowflake_engine.py
```diff
@@ -271,32 +271,63 @@ def _materialize_one(
...
 
+            # Determine payload size if writing to an external online store
+            if self.repo_config.online_store.type != "snowflake.online":
+                if feature_view.entity_columns:
+                    join_keys = [entity.name for entity in feature_view.entity_columns]
+                    unique_entities = '"' + '", "'.join(join_keys) + '"'
+
+                    query = f"""
+                        SELECT
+                            {unique_entities}
```
adchia, line 282: can you just do a count distinct in the query instead of getting a length of the individual rows

sdk/python/feast/infra/materialization/snowflake_engine.py
```diff
@@ -271,32 +271,68 @@ def _materialize_one(
...
+                    query = f"""
+                    SELECT
+                        COUNT(*)
+                    FROM
+                    (
+                        SELECT
+                            {unique_entities}
+                        FROM
+                            {feature_view.batch_source.get_table_query_string()}
+                        QUALIFY ROW_NUMBER() OVER (PARTITION BY {unique_entities} ORDER BY {unique_entities}) = 1
```
adchia, line 289: why do you need a window function here? Cant you do like APPROX_COUNT_DISTINCT(unique_entities) or something?

sdk/python/feast/infra/materialization/snowflake_engine.py
```diff
@@ -271,32 +271,68 @@ def _materialize_one(
...
+                        self.repo_config,
+                        fv_to_proto_sql,
+                        feature_view,
+                        project,
+                    )
+                else:
+                    click.echo(
+                        f"Materializing batch {Style.BRIGHT + Fore.GREEN}#{i+1}{Style.RESET_ALL} of {Style.BRIGHT + Fore.GREEN}{len(feature_batch)}{Style.RESET_ALL} features"
+                    )
+                    with tqdm_builder(rows_to_write) as pbar:
```
adchia, line 329: think here you need rows_to_write * len(feature_batches)

sfc-gh-madkins: @adchia I thought so to ... but apparently we aren't counting rows to write =  entities * features, we count just entities. Look at a simple "feast init" example.

sdk/python/feast/infra/materialization/snowflake_engine.py
sfc-gh-madkins: updated

sfc-gh-madkins: ideally we do DISTINCT not APPROx

sdk/python/feast/infra/materialization/snowflake_engine.py
adchia: hmm in the old world that was correct before because we didn't split it up and there are only num_entities rows in a normal feast materialization. 

In this case, you're calling into the pbar more than that though since you also are doing batches. so i think you'd hit 100% after the first batch of 100 features, and show nothing on the second batch onwards?

sfc-gh-madkins: @adchia it should reset every loop of 100, no? the way you are suggesting would give the feel of one bar that shows full completion but the final number wouldnt make any sense

sfc-gh-madkins: @adchia updated to one bar ...merge it

sdk/python/feast/infra/materialization/snowflake_engine.py
```diff
@@ -271,32 +271,65 @@ def _materialize_one(
...
+                    query = f"""
+                        SELECT
+                            COUNT(DISTINCT {unique_entities})
+                        FROM
+                            {feature_view.batch_source.get_table_query_string()}
+                    """
+
+                    with get_snowflake_conn(self.repo_config.offline_store) as conn:
+                        rows_to_write = conn.cursor().execute(query).fetchall()[0][0]
+            else:
```
adchia, line 289: add a tab here? seems wrong

sfc-gh-madkins:
@adchia ok take a look at this one ... added pbar for snowflake.online as well

PR #3411, opened 2022-12-23
updating the batch field so that if you want return the created date …

franciscojavierarceo:
…of a model you can just add it in the get_online_features feature argument

Fixes #


sdk/python/feast/infra/offline_stores/file.py
```diff
@@ -599,6 +599,11 @@ def _normalize_timestamp(
         created_timestamp_column_type = df_to_join_types[created_timestamp_column]
 
     if not hasattr(timestamp_field_type, "tz") or timestamp_field_type.tz != pytz.UTC:
+        # if you are querying for the event timestamp field, we have to deduplicate
```
adchia, line 602: not fully sure i understand what these duplicate columns are coming from? Can you leave more comments on why?

franciscojavierarceo: yeah i'll update the PR but basically querying for `created` in `get_online_features` duplicates the `created_timestamp_column`. So I could go deeper and try to fix the duplication but I needed a quick fix on my side and it works correctly for us.

franciscojavierarceo: @adchia updated the comment 👍

adchia (approved):
/lgtm

franciscojavierarceo:
sorry @adchia looks like I had to redo the sign off will need another stamp

PR #3419, opened 2022-12-28
feat: Add better spark support for snowflake offline store

### LINKED ISSUE #3364: Support for Snowflake connector with Spark

For folks on data bricks or other spark-based platforms, getting large data as a pandas df may not be an option .... It's possible to get a `spark df` using `Snowflake Connector for Python` but it's orders of magnitude slower compared to `Snowflake connector for Spark`

Add a function to `SnowflakeRetrievalJob` which supports executing & retrieving snowflake query through `Snowflake Connector for Spark`

Tried getting snowflake query results as `arrow_batches` -> converting to pandas on each batch -> to spark data frame -> concatenate all spark dos to one. This is orders go marinated slower than reading same data frame from `Snowflake Connector for Spark`.

cc: @sfc-gh-madkins

sfc-gh-madkins:

Add spark output to snowflake

Fixes #3364

### DIFF
--- a/docs/reference/offline-stores/overview.md
+++ b/docs/reference/offline-stores/overview.md
@@ -47,10 +47,10 @@ Below is a matrix indicating which `RetrievalJob`s support what functionality.
 | export to dataframe                                   | yes | yes | yes | yes | yes | yes | yes |
 | export to arrow table                                 | yes | yes | yes | yes | yes | yes | yes |
 | export to arrow batches                               | no  | no  | no  | yes | no  | no  | no  |
-| export to SQL                                         | no  | yes | no  | yes | yes | no  | yes |
+| export to SQL                                         | no  | yes | yes  | yes | yes | no  | yes |
 | export to data lake (S3, GCS, etc.)                   | no  | no  | yes | no  | yes | no  | no  |
 | export to data warehouse                              | no  | yes | yes | yes | yes | no  | no  |
-| export as Spark dataframe                             | no  | no  | no  | no  | no  | yes | no  |
+| export as Spark dataframe                             | no  | no  | yes  | no  | no  | yes | no  |
 | local execution of Python-based on-demand transforms  | yes | yes | yes | yes | yes | no  | yes |
 | remote execution of Python-based on-demand transforms | no  | no  | no  | no  | no  | no  | no  |
 | persist results in the offline store                  | yes | yes | yes | yes | yes | yes | no  |
--- a/docs/reference/offline-stores/snowflake.md
+++ b/docs/reference/offline-stores/snowflake.md
@@ -7,7 +7,7 @@ The [Snowflake](https://trial.snowflake.com) offline store provides support for 
 * Entity dataframes can be provided as a SQL query or can be provided as a Pandas dataframe. A Pandas dataframes will be uploaded to Snowflake as a temporary table in order to complete join operations.
 
 ## Getting started
-In order to use this offline store, you'll need to run `pip install 'feast[snowflake]'`. 
+In order to use this offline store, you'll need to run `pip install 'feast[snowflake]'`.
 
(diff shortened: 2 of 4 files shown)


sfc-gh-madkins:
@amithadiraju1694 try this out ... you will need to have a pyspark environment with the snowflake spark connector installed.

You will need to pass in the spark session plus a dict of snowflake login params ... see the function comments

sfc-gh-madkins:
@amithadiraju1694 you probably already tried this code is my guess ... the reason you were getting that error is because the initial input dataframe is scoped to a different connection and spark cant find it

amithadiraju1694:
> @amithadiraju1694 you probably already tried this code is my guess ... the reason you were getting that error is because the initial input dataframe is scoped to a different connection and spark cant find it

Thanks for this @sfc-gh-madkins. Tried different variation of this along with your original solution, but the execution never stops ( data bricks keeps saying "Running Command" ).

My final attempt looked like this in `snowflake.py`

`
       def to_pyspark_df(self, spark_session: SparkSession, sfparam: dict) -> DataFrame:
        """
        Method to convert snowflake query results to pyspark data frame.

        Args:
            spark_session: spark Session variable of current environment.

        Returns:
            spark_df: A pyspark dataframe.
        """

        if isinstance(spark_session, SparkSession):
            table_name = "feast_spark_" + uuid.uuid4().hex
            self.to_snowflake(table_name = table_name)
            query = f'SELECT * FROM "{table_name}"'

            spark_df = spark_session.read.format( "net.snowflake.spark.snowflake"  ).options(**sfparam).option("query", query).option("autopushdown" , "on").load()

            query = f'DROP TABLE "{table_name}"'
            execute_snowflake_statement(self.snowflake_conn, query)

            return spark_df` 


I tried the original solution as well, which's giving me the same result. I'm wondering if from original solution `snowflake.py -> line no 486 to 500` should be run inside `with query` scope or outside of it. If inside, I'm confused on why it needs to be run inside that scope ?

sfc-gh-madkins:
Are you sure you have the snowflake spark connector installed? Due to see
the query being issued to the snowflake side? I was able to test this with
success locally. Temporary tables are scoped to a specific snowflake
session.

amithadiraju1694:
> Are you sure you have the snowflake spark connector installed? Do you see the query being issued to the snowflake side? I was able to test this with success locally. Temporary tables are scoped to a specific snowflake session.
> […](#)

I was running my code on databricks, so `snowflake connector for spark` must be installed already ... Since we're giving "auto pushdown: on" query should be executing on snowflake side, though I'm not able to figure out why it's running forever. Will try again.

amithadiraju1694:
> Are you sure you have the snowflake spark connector installed? Do you see the query being issued to the snowflake side? I was able to test this with success locally. Temporary tables are scoped to a specific snowflake session.
> […](#)

I debugged and found that, the program halts at `self.to_snowflake(table_name)` ; temporary table with given table name isn't created at all in given database.schema for some reason;  not sure if this is cuz of access issues, I'm using a dev instance crews and should have required accesses.

sfc-gh-madkins:
Is there an error on the snowflake side?

amithadiraju1694:
> Is there an error on the snowflake side?
> […](#)

in `to_snowflake` method, `temporary` argument was set to `false` by default, changing that to true solved the unresponsiveness of the query. But now, I see `SQL compilation error` , my 'DB.SCHEMA."table_name"' is not found or not authorized. I faced a similar issue before for which I made a quick fix, but even the quick fix isn't working now ( my schema is not public contains underscore in its name ).

sfc-gh-madkins:
Can you try this using the default snowflake project? feast init -t
snowflake The temporary argument should be set to false, as you dont want
to create a temporary table. sf_params should match the offline store
params.

sfc-gh-madkins:
@amithadiraju1694 is this new PR going to break your existing code?

sfc-gh-madkins:
@adchia this might cause a breaking change for a single user, but he has been unresponsive

PR #3421, opened 2022-12-29
feat: Add template for Github Codespaces

sfc-gh-madkins:

Make it easier for people to develop feast

Fixes #

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,16 @@
+{
+  "name": "feast-devcontainer",
+  "image": "mcr.microsoft.com/devcontainers/base:ubuntu",
+  "features": {
+    "ghcr.io/devcontainers/features/docker-in-docker:2": {
+      "version": "latest"
+    },
+    "ghcr.io/devcontainers/features/python:1": {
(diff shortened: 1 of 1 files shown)


sfc-gh-madkins:
@adchia this is actually a very slow way to get someone started with github codespaces, we should just publish a docker image that can be pulled in to reduce startup times.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,16 @@
...
+  "features": {
+    "ghcr.io/devcontainers/features/docker-in-docker:2": {
+      "version": "latest"
+    },
+    "ghcr.io/devcontainers/features/python:1": {
+      "version": "3.9"
+    },
+    "ghcr.io/meaningful-ooo/devcontainer-features/homebrew:2": {
+      "version": "latest"
+    }
```
achals, line 13: What does this layer do?

achals (approved):

.devcontainer/devcontainer.json
sfc-gh-madkins: It installs brew, so I can easily install mysql :)

PR #3422, opened 2022-12-30
fix: Remove snowflake source warehouse tech debt

### LINKED ISSUE #2981: Snowflake source warehouse field is unused
## Expected Behavior

Id expect if I defined a warehouse source that warehouse would be used for inference functions attached to the source. 

## Current Behavior

Source uses offline store credentials

### Specifications

- Version:
- Platform:
- Subsystem:

## Possible Solution

I think we remove warehouse from snowflake source as source and offline store are currently intertwined

you can either define a query or a database.schema.table path

sfc-gh-madkins:

Fixes #2981

### DIFF
--- a/protos/feast/core/DataSource.proto
+++ b/protos/feast/core/DataSource.proto
@@ -197,6 +197,8 @@ message DataSource {
 
   // Defines options for DataSource that sources features from a Snowflake Query
   message SnowflakeOptions {
+    reserved 5; // Snowflake warehouse name
+
     // Snowflake table name
     string table = 1;
 
@@ -209,9 +211,6 @@ message DataSource {
 
     // Snowflake schema name
     string database = 4;
-
-    // Snowflake warehouse name
-    string warehouse = 5;
   }
 
   // Defines options for DataSource that sources features from a spark table/query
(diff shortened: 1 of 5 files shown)


protos/feast/core/DataSource.proto
```diff
@@ -209,9 +209,6 @@ message DataSource {
 
     // Snowflake schema name
     string database = 4;
-
-    // Snowflake warehouse name
-    string warehouse = 5;
```
achals, line 214: should be added as a `reserved` index up top. https://developers.google.com/protocol-buffers/docs/proto3#reserved

sdk/python/feast/infra/offline_stores/snowflake_source.py
```diff
@@ -26,7 +26,6 @@ def __init__(
         name: Optional[str] = None,
         timestamp_field: Optional[str] = "",
         database: Optional[str] = None,
-        warehouse: Optional[str] = None,
```
achals, line 29: This should be kept in, and just raise a deprecation warning when specified. Otherwise it'll break users pretty bad

sfc-gh-madkins: Moving to draft -- will add deprecation flag

sfc-gh-madkins: added back

protos/feast/core/DataSource.proto
sfc-gh-madkins: done

sfc-gh-madkins:
@adchia @achals can we push this through?

adchia (approved):
/lgtm

PR #3423, opened 2022-12-30
fix: Enforce fully qualified table names in Snowflake

### LINKED ISSUE #3286: reconstitute (Snowflake) DataSources without "duplicate" database declarations

I'm investigating improvements to the DataHub Feast Integration <https://datahubproject.io/docs/generated/ingestion/sources/feast/>.  To complete a data lineage graph, the ingestion plugin needs to be able to associate a feature view with the "fully qualified" name of the table.  That is, something like `${db}.${schema}.${table}`.

Right now this is only works by manually duplicating the database name from `feature_store.yaml` into the source definition, which feels well... duplicative. Otherwise the properties of the class are not fully reconstituted from the registry.

(From the `init` demo)

```
driver_stats_source = SnowflakeSource(
    # The Snowflake table where features can be found
    database=yaml.safe_load(open("feature_store.yaml"))["offline_store"]["database"],
    table=f"{project_name}_feast_driver_hourly_stats",
    # The event timestamp is used for point-in-time joins and for ensuring only
    # features within the TTL are returned
    # The (optional) created timestamp is used to ensure there are no duplicate
    # feature rows in the offline store or when building training datasets
    created_timestamp_column="created",
)
```

Results in 
```
$ feast data-sources describe notable_hyena_feast_driver_hourly_stats
type: BATCH_SNOWFLAKE
timestampField: event_timestamp
createdTimestampColumn: created
snowflakeOptions:
  table: notable_hyena_feast_driver_hourly_stats
  schema: PUBLIC
  database: EXPERIMENTS_CHRIS
name: notable_hyena_feast_driver_hourly_stats
```

While 
```
driver_stats_source = SnowflakeSource(
    # The Snowflake table where features can be found
    #database=yaml.safe_load(open("feature_store.yaml"))["offline_store"]["database"],
    table=f"{project_name}_feast_driver_hourly_stats",
    # The event timestamp is used for point-in-time joins and for ensuring only
    # features within the TTL are returned
    # The (optional) created timestamp is used to ensure there are no duplicate
    # feature rows in the offline store or when building training datasets
    created_timestamp_column="created",
)
```

only gives
```
$ feast data-sources describe notable_hyena_feast_driver_hourly_stats
type: BATCH_SNOWFLAKE
timestampField: event_timestamp
createdTimestampColumn: created
snowflakeOptions:
  table: notable_hyena_feast_driver_hourly_stats
name: notable_hyena_feast_driver_hourly_stats
```

Note that even the `init` example just loads from the yaml file.

Since `get_historical_features` works in either case, I presume there must already be a code path that handles this.  I would like referencing the `database` and `schema` properties of a `SnowflakeSource` to always have the same values as when `get_historical_features` is called.

The DataHub ingestion plugin could read `feature_store.yaml` and smoosh strings together.  That would work for this particular case, but I suspect it would be brittle in the long run.

sfc-gh-madkins:

Enforces fully qualified snowflake source names

Fixes #3286

### DIFF
--- a/sdk/python/feast/infra/offline_stores/snowflake.py
+++ b/sdk/python/feast/infra/offline_stores/snowflake.py
@@ -129,10 +129,6 @@ class SnowflakeOfflineStore(OfflineStore):
         assert isinstance(config.offline_store, SnowflakeOfflineStoreConfig)
         assert isinstance(data_source, SnowflakeSource)
 
-        from_expression = data_source.get_table_query_string()
-        if not data_source.database and data_source.table:
-            from_expression = f'"{config.offline_store.database}"."{config.offline_store.schema_}".{from_expression}'
-
         if join_key_columns:
             partition_by_join_key_string = '"' + '", "'.join(join_key_columns) + '"'
             partition_by_join_key_string = (
@@ -167,7 +163,7 @@ class SnowflakeOfflineStore(OfflineStore):
             FROM (
                 SELECT {field_string},
                 ROW_NUMBER() OVER({partition_by_join_key_string} ORDER BY {timestamp_desc_string}) AS "_feast_row"
-                FROM {from_expression}
+                FROM {data_source.get_table_query_string()}
                 WHERE "{timestamp_field}" BETWEEN TIMESTAMP '{start_date}' AND TIMESTAMP '{end_date}'
             )
             WHERE "_feast_row" = 1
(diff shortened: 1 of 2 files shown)


sfc-gh-madkins:
@cburroughs let me know if this fixes your data lineage issue?

sfc-gh-madkins:
@cburroughs if you look at the failed unit test, you will see that there are other offline store options that only require a table name as well, so my guess is this issue isnt snowflake specific.

The reason you dont need fully qualified paths is due to the fact that some people set their offline store config to be in the same qualified path as where the table lives, and so this just works.

sfc-gh-madkins:
@cburroughs maybe one last comment here, it looks like the datahub integration relies on a .yaml file -- this is often not used in prod environments

cburroughs:
Thanks!  I took a look at the PR.  I think it's fine that the snowflake accepts input that isn't fully qualified (and it would be a big hassle to work around requiring it to be fully qualified).  My original trip-up was that when I ask a Snowflake Source "hey what database are you using?" I expect to get the database it would use for a query whether that is explicit or inferred.

PR #3425, opened 2022-12-31
feat: Adds description attribute for features/fields

### LINKED ISSUE #3277: `Feature.proto` should have a description field

I would like to attach a human-readable description to each field in a feature view. Since data scientists shop for individual features and not feature-views, it's useful to have documentation associated with each feature individually (rather than for the entire feature view). Persisting it here will allow us to surface them in various UIs.

Similar to `Entity.proto`, we should add a description field to features and store them in the registry.

For now, we're just using a `description` tag on each feature. This works, but it's clunky. It also means that in our UI, we have to remove certain tags before showing the raw tags to the users.

**Additional context**

amommendes:
Signed-off-by: Amom Mendes <amommendes@hotmail.com>

Adds description attribute for features/fields

Fixes #3277

### DIFF
--- a/protos/feast/core/Feature.proto
+++ b/protos/feast/core/Feature.proto
@@ -33,4 +33,8 @@ message FeatureSpecV2 {
 
     // Tags for user defined metadata on a feature
     map<string,string> tags = 3;
+
+    // Description of the feature.
+
+    string description = 4;
 }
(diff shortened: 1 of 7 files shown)


achals (approved):
/lgtm
/ok-to-test

PR #3429, opened 2023-01-03
fix: Improve BQ point-in-time joining scalability

### LINKED ISSUE #3426: BigQueryOfflineStore facing problems with complex feature views
## Expected Behavior 
I expect BigQueryOffline store would be able to have freedom in joining multiple feature views at large scale, including table-based feature views and SQL query-based feature views

## Current Behavior
When I conducted point-in-time joins across like 30 feature views, BQ client thrown this error

``` bash
400 Resources exceeded during query execution: Not enough resources for query planning - too many subqueries or query is too complex.
```

## Steps to reproduce
1. Prepare 2 table-based feature views
2. Prepare 2 query-based feature views (with pivot statement on 10 values)
3. Conduct point-in-time joining on all feature views

### Specifications

- Version: 0.27.1
- Platform: Linux

## Possible Solution
Convert BQ template from using `WITH` to using `TEMP TABLE`

sudohainguyen:
Signed-off-by: Hai Nguyen <quanghai.ng1512@gmail.com>

**What this PR does / why we need it**: point-in-time joining in BQ offline store can be more scalable

Fixes #3426

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -488,10 +488,24 @@ class BigQueryRetrievalJob(RetrievalJob):
             return str(job_config.destination)
 
         with self._query_generator() as query:
-            self._execute_query(query, job_config, timeout)
+            dest = job_config.destination
+            # because setting destination for scripts is not valid
+            # remove destination attribute if provided
+            job_config.destination = None
+            bq_job = self._execute_query(query, job_config, timeout)
 
-            print(f"Done writing to '{job_config.destination}'.")
-            return str(job_config.destination)
+            if not job_config.dry_run:
+                config = bq_job.to_api_repr()["configuration"]
+                # get temp table created by BQ
+                tmp_dest = config["query"]["destinationTable"]
+                temp_dest_table = f"{tmp_dest['projectId']}.{tmp_dest['datasetId']}.{tmp_dest['tableId']}"
+
+                # persist temp table
+                sql = f"CREATE TABLE {dest} AS SELECT * FROM {temp_dest_table}"
+                self._execute_query(sql, timeout=timeout)
+
+            print(f"Done writing to '{dest}'.")
+            return str(dest)
 
     def _to_arrow_internal(self) -> pyarrow.Table:
         with self._query_generator() as query:
@@ -777,7 +791,7 @@ MULTIPLE_FEATURE_VIEW_POINT_IN_TIME_JOIN = """
  Compute a deterministic hash for the `left_table_query_string` that will be used throughout
  all the logic as the field to GROUP BY the data
 */
-WITH entity_dataframe AS (
+CREATE TEMP TABLE entity_dataframe AS (
     SELECT *,
         {{entity_df_event_timestamp_col}} AS entity_timestamp
         {% for featureview in featureviews %}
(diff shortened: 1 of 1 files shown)


sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -917,7 +931,7 @@ def arrow_schema_to_bq_schema(arrow_schema: pyarrow.Schema) -> List[SchemaField]
             ,created_timestamp
         {% endif %}
     )
-){% if loop.last %}{% else %}, {% endif %}
```
adchia, line 920: haven't taken a close look, but seems like you removed this line even though there's a loop. Was that intended?

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -488,10 +488,24 @@ def to_bigquery(
             return str(job_config.destination)
 
         with self._query_generator() as query:
-            self._execute_query(query, job_config, timeout)
+            dest = job_config.destination
+            # because setting destination for scripts is not valid
```
adchia, line 492: was this just not working before? Seems like it was using the job_config.destination before?

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -488,10 +488,24 @@ def to_bigquery(
...
-            print(f"Done writing to '{job_config.destination}'.")
-            return str(job_config.destination)
+            if not job_config.dry_run:
+                config = bq_job.to_api_repr()["configuration"]
+                # get temp table created by BQ
+                tmp_dest = config["query"]["destinationTable"]
+                temp_dest_table = f"{tmp_dest['projectId']}.{tmp_dest['datasetId']}.{tmp_dest['tableId']}"
+
+                # persist temp table
+                sql = f"CREATE TABLE {dest} AS SELECT * FROM {temp_dest_table}"
```
adchia, line 504: similarly, confused why there is a new need to do this when it seemed like maybe it worked before?

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -777,7 +791,7 @@ def arrow_schema_to_bq_schema(arrow_schema: pyarrow.Schema) -> List[SchemaField]
  Compute a deterministic hash for the `left_table_query_string` that will be used throughout
  all the logic as the field to GROUP BY the data
 */
-WITH entity_dataframe AS (
+CREATE TEMP TABLE entity_dataframe AS (
```
adchia, line 794: is the main change here to create temp tables for everything? (and everything else got indented)?

sudohainguyen: yes, create temp table for entity dataframe and every joining feature views.

sdk/python/feast/infra/offline_stores/bigquery.py
sudohainguyen: when we run multiple scripts separated by semicolon, BigQuery just doesn't accept passing destination explicitly, will throw error
`google.api_core.exceptions.BadRequest: 400 configuration.query.destinationTable cannot be set for scripts`

that's why I add some workaround here

sudohainguyen: simply retrieve BQ-generated temporary table after scripts being executed (should look like this `<project_id>._08dc845331ibb01824ea0f3c3e079e93b3a076.anon933abd7cb6950c3c488af61fe94ad1713b6820cd1115fddd0c09cde32ec12823` )

then we persist it as our naming convention with prefix `historical_`

sdk/python/feast/infra/offline_stores/bigquery.py
sudohainguyen: and the temp tables only exist within the current execution session

chhabrakadabra: Are you able to confirm what the scope of the session is? Does each invocation of `self._execute_query` create a new session? At a cursory glance, I see that we're not explicitly passing `create_session=True` in the `QueryJobConfig` instance. I think it would be very helpful to have a confirmation that there isn't going to be any cross contamination between concurrent historical fetches.

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -488,10 +488,24 @@ def to_bigquery(
             return str(job_config.destination)
 
         with self._query_generator() as query:
-            self._execute_query(query, job_config, timeout)
+            dest = job_config.destination
```
chhabrakadabra, line 491: Conceptually, the PR seems reasonable (once any doubts about temp table cross-contamination are put to rest). But in terms of code readability, I can't help but notice that if this method is called without `job_config`, we first create it containing only the destination at the start of the method, then we remove the destination here. Perhaps this could use a bit of a cleanup where the job_config management is all in one spot?

sdk/python/feast/infra/offline_stores/bigquery.py
sudohainguyen: `Does each invocation of self._execute_query create a new session?` yes it does

`I see that we're not explicitly passing create_session=True in the QueryJobConfig instance.` we only need to declare this config when we need to reuse the session in a different code scope. In this case, we gather all of feature views in a collection of scripts to execute once. I can confirm after the scripts are executed, temp tables do not exist anywhere else.

sdk/python/feast/infra/offline_stores/bigquery.py
sudohainguyen: strongly agree! but in this PR I'm trying to not break any other flows. Perhaps some users would prefer passing some of their own job configurations to `to_bigquery` method, that's why I only deal with the `destination` attribute.

sudohainguyen: I believe we shouldn't allow users to pass the destination, there would be some changes needed but let's deal with them in a separate PR

sudohainguyen:
FYI: spark offline store also following the similar concept
https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py#L547

sudohainguyen:
> 

hmm, did you check your data? In case your feature view ttl is 1, and there were no records of that entity in 2022-05-26 but 25th, the results are expected to be the same.

FYI: temp table only be reused when the query is exactly the same.

RadionBik:
the problem has gone when I downgraded feast to `0.29.0`

TTL = 0.

so here is the data-snippet I spot the problem on:
image

`related_deactivated_at` is the `timestamp_field`.

here is the test code:
```
@pytest.mark.parametrize(
    'entity_name,entities,model_name,results,event_timestamps',
    [
        (
            'alias',
            ['Pfi9FE1Qbcb022z1D317', 'Pfi9FE1Qbcb022z1D317'],
            ModelName.onboarding_model,
            [
                {
                    fields.bad_full_name_matches.name: 1,
                    fields.bad_mobile_matches.name: 0,
                    fields.bad_address_matches.name: 0,
                    fields.bad_nob_matches.name: 2,
                },
                {
                    fields.bad_full_name_matches.name: 1,
                    fields.bad_mobile_matches.name: 1,
                    fields.bad_address_matches.name: 1,
                    fields.bad_nob_matches.name: 3,
                },
            ],
            [datetime(2022, 5, 25, 0, 0, 0), datetime(2022, 5, 26, 0, 0, 0)],
        ),
    ]
)
def test_model_offline_features(entity_name, entities, model_name, results, event_timestamps, feature_store_client):
    offline_result = feature_store_client.get_historical_features(
        entity_name=entity_name,
        entities=entities,
        model_name=model_name,
        event_timestamps=event_timestamps
    )
    offline_dicts = offline_result.to_dict(orient='records')
    for offline_dict, result in zip(offline_dicts, results):
        offline_dict.pop(entity_name)
        assert offline_dict == result
```

it passed from the first attempt, but when I ran it the second time -> it failed, with all entries for the `datetime(2022, 5, 25, 0, 0, 0)` timestamp being equal to the second test-case (i.e. 1,1,1,3). That's why I assumed that temp tables were not properly re-created.

I am not sure where the problem lies, but it occurs only in feast `0.30.2`, and I think it is related to this merged PR, because I didn't find others that could have changed historical feature querying. 

Do you have any ideas how to fix the problem?

sudohainguyen:
I see, let me check real quick

sudohainguyen:
I conducted retrieval with 5 entities and 2 different timestamps (2023/03/24 and 2023/03/25) in the same notebook session and they are not identical :(

PR #3437, opened 2023-01-07
feat: Add better spark support for snowflake offline store

sfc-gh-madkins:

Fixes #


sfc-gh-madkins:
@amithadiraju1694 wait. This code will not work.

sfc-gh-madkins:
@amithadiraju1694 your best chance here is to use the code that I already wrote and see if you can get the right access to your snowflake and databricks accounts to diagnose the issue

PR #3441, opened 2023-01-11
ci: Bump sphinx dependency

felixwang9817:

**What this PR does / why we need it**: Bump Sphinx dependency as it was causing issues during CI.

Additionally adds the `py` dependency as the latest `pytest` version (7.2.1, released today) contains a change that breaks the `pytest-benchmark` dependency. See https://github.com/pytest-dev/pytest/issues/10420 for more details.

Fixes #


adchia (approved):
/lgtm

PR #3448, opened 2023-01-15
chore: Force entity inference without modifying `fv.schema`

felixwang9817:
**What this PR does / why we need it**: #3403 is currently blocked on being able to run an entity inference test without overwriting the `fv.schema` attribute. This PR unblocks #3403.

Fixes #


adchia (approved):
/lgtm

Issue #3451, opened 2023-01-19
0.28.0 Python SDK: Assertion failure on Boolean types in type_map.py #3443


beauchbum:
## Expected Behavior 
In Snowflake
`select FEAST_FEAST_FEATURES_SNOWFLAKE_BOOLEAN_TO_BOOL_PROTO(FALSE);`
--> `3800`

## Current Behavior
`select FEAST_FEAST_FEATURES_SNOWFLAKE_BOOLEAN_TO_BOOL_PROTO(FALSE);`

```
Traceback (most recent call last):
  File "/home/udf/6137508531934/feast.zip/feast/infra/utils/snowflake/snowpark/snowflake_udfs.py", line 150, in feast_snowflake_boolean_to_bool_boolean_proto
    python_values_to_proto_values(df[0].to_numpy(), ValueType.BOOL),
  File "/home/udf/6137508531934/feast.zip/feast/type_map.py", line 454, in python_values_to_proto_values
    return _python_value_to_proto_value(value_type, values)
  File "/home/udf/6137508531934/feast.zip/feast/type_map.py", line 408, in _python_value_to_proto_value
    assert type(sample) in [np.int64, int, np.float64, float]
AssertionError
 in function FEAST_FEAST_FEATURES_SNOWFLAKE_BOOLEAN_TO_BOOL_PROTO with handler feast.infra.utils.snowflake.snowpark.snowflake_udfs.feast_snowflake_boolean_to_bool_boolean_proto
```

## Steps to reproduce
- `feast==0.28.0`
- Batch materialization job where Boolean values (must include `FALSE`)

### Specifications

- Version: 0.28.0
- Platform: mac, linux
- Subsystem: ?

## Possible Solution
As pointed out in [this issue](https://github.com/feast-dev/feast/issues/3443), I believe this bug was introduced in 0.28.0 such that:

Boolean value evaluates to
https://github.com/feast-dev/feast/blob/98a24a34c4274464f43c49b33b2b0baa88221cbd/sdk/python/feast/type_map.py#L399-L403
https://github.com/feast-dev/feast/blob/98a24a34c4274464f43c49b33b2b0baa88221cbd/sdk/python/feast/type_map.py#L303

Which causes it to enter this conditional
https://github.com/feast-dev/feast/blob/98a24a34c4274464f43c49b33b2b0baa88221cbd/sdk/python/feast/type_map.py#L404-L410

Where it can fail if `sample == False` because `False == 0` --> `True`

PR #3452, opened 2023-01-19
fix: Add check for bool type in addition to sample

beauchbum:
Signed-off-by: Ryan Beauchamp <ryan.beauchamp@ezcater.com>

The current implementation throws an exception when `sample == False` as it enters this conditional:
```python
# This block is entered as `False == 0` evaluates to `True`
if sample == 0 or sample == 0.0:

  # However, this fails because `type(False) => bool`
  assert type(sample) in [np.int64, int, np.float64, float]
```

This PR adds an additional check to bypass this condition when dealing with `bool`. 

Fixes https://github.com/feast-dev/feast/issues/3451


adchia (approved):
/lgtm

PR #3460, opened 2023-01-20
fix: Ensure no duplicates in `fv.schema`

### LINKED ISSUE #3459: schema has duplicate fields
## Expected Behavior 
feature_view.schema has no duplicate fields

## Current Behavior
feature_view.schema has no duplicate fields

### Specifications

- Version:
- Platform:
- Subsystem:

## Possible Solution

in feature_view:
```
    def schema(self) -> List[Field]:
        return list(set(self.entity_columns + self.features))
```

felixwang9817:

Fixes #3459

### DIFF
--- a/sdk/python/feast/feature_view.py
+++ b/sdk/python/feast/feature_view.py
@@ -259,7 +259,7 @@ class FeatureView(BaseFeatureView):
 
     @property
     def schema(self) -> List[Field]:
-        return self.entity_columns + self.features
+        return list(set(self.entity_columns + self.features))
 
     def ensure_valid(self):
         """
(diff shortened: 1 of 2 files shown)


sdk/python/feast/feature_view.py
```diff
@@ -259,7 +259,7 @@ def join_keys(self) -> List[str]:
 
     @property
     def schema(self) -> List[Field]:
-        return self.entity_columns + self.features
+        return list(set(self.entity_columns + self.features))
```
achals, line 262: Does Field implement `__eq__` correctly?

achals (approved):

sdk/python/feast/feature_view.py
felixwang9817: just updated it to check all attributes of `Field` properly, so yes I think so

achals (approved):
/lgtm

Issue #3458, opened 2023-01-20
move google-api-core/googleapis-common-protos depdencies to extras


cburroughs:
**Is your feature request related to a problem? Please describe.**
Feast currently requires
```
    "google-api-core>=1.23.0,<3",
    "googleapis-common-protos>=1.52.*,<2",
```

As core dependencies.  I'd expect google cloud apis to not be a requirement to use (for example) AWS and my naive grepping of the code didn't turn up any obvious counter examples.  From the `git blame` chain it looks like there were not yet any precedents for extras when these were added: https://github.com/feast-dev/feast/blame/0da112b363b6e67a95ead3d95c209be301db56a5/sdk/python/setup.py

**Describe the solution you'd like**

I presume they could move to `GCP_REQUIRED` or another section.

**Additional context**
I noticed this because where adding feast as a dependency to an existing project has caused pip's resolve to take 10s of minutes. Based on the the plethora of `INFO: pip is looking at multiple versions of google-api-core to determine which version is compatible with other requirements. This could take a while.`  I suspect these are particularly tricky dependencies for pip satisfy.

PR #3463, opened 2023-01-24
chore: Move google-api dependencies to GCP extras

cburroughs:
These dependencies were added before any extras existed in setup.py. Feast users would generally expect that using AWS or Azure does not require Google Cloud apis.

Fixex #3458

Signed-off-by: Chris Burroughs <chris.burroughs@gmail.com>


achals (approved):
/lgtm

adchia (approved):
/lgtm

PR #3466, opened 2023-01-27
fix: Fix delete sfv twice issue

hao-affirm:
Signed-off-by: hao-affirm <104030690+hao-affirm@users.noreply.github.com>

StreamFeatureView is subclass of FeatureView, so when update a StreamFeatureView, it will be deleted twice as it is in `views_to_delete` and `sfv_to_delete`

Fixes #


achals (approved):
/ok-to-test
/lgtm

PR #3471, opened 2023-01-30
ci: Fix setuptools issue with wildcard

adchia:
Converts dependencies to use ~= or explicit versions since wildcard versions don't seem to be working with setuptools

One issue that came up was mypy-protobuf had to be pinned to an older version to overcome some lint issues around stubs not successfully installing

Fixes #


gbmarc1 (approved):
LGTM

achals:
Ugh that's annoying. did this change with specific versions of pip or something else?

PR #3473, opened 2023-01-31
chore: Add tests to bool test type

gbmarc1:
- We need test to avoid regressions
- Also, add support for int in `_python_value_to_proto_value` for bool values. 

@adchia Make sure that it indeed make sense to add the int support! Not sure about this one.


achals (approved):
/ok-to-test
/lgtm

PR #3479, opened 2023-02-03
fix: Spark kafka processor sorting

shaurya-nwse:
Signed-off-by: shaurya.rawat <shaurya.rawat@new-work.se>

Minor fix for the sorting in `sdk/python/feast/infra/contrib/spark_kafka_processor.py` to make sure latest values are picked up for each entity.


felixwang9817 (approved):
/lgtm

Issue #3430, opened 2023-01-04
SQL Registry does not populate cache correctly


cbarcroft:
## Expected Behavior 
Using SQL registry, should be able to retrieve online features for an existing feature view using the Python SDK.

## Current Behavior
FeatureStore.get_online_features() currently returns a FeatureviewNotFoundException, despite the feature view existing as evidenced by the output of FeatureStore.list_feature_views().

## Steps to reproduce
1.  Configure feature store using an SQL registry
2. Create a feature view
3. Materialize data to online store for the feature view
4. Attempt to retrieve the online features using FeatureStore.get_online_features()

### Specifications

- Version: 0.27.1
- Platform: macOS Monterey 12.4
- Python 3.9.16
- SQL Registry, using postgres on an AWS AuroraDB instance

## Possible Solution
This error seems to be specific to the SQL registry, as I have followed the exact same steps using a File registry and there is no issue.  I can also circumvent the error by disabling the cache for the SQL registry, so it seems localized to the SQL registry cache itself.  The feature views are successfully retrieved, but are rejected by [this line](https://github.com/feast-dev/feast/blob/81c3483699a9e9ac3f6057dbf3e45ee748ccc8e8/sdk/python/feast/infra/registry/proto_registry_utils.py#L144) because their spec does not have a project.  

It appears that normally, the project is attached to the proto spec right before committing registry changes ([like so](https://github.com/feast-dev/feast/blob/81c3483699a9e9ac3f6057dbf3e45ee748ccc8e8/sdk/python/feast/infra/registry/registry.py#L394), but the SQL registry does not do this and instead relies on each class's to_proto() output, which does not contain the project.  This issue appears to extend to other types of objects as well (entities, etc.)

cbarcroft:
Confirmed issue still exists in 0.28.0

EXPEbdodla:
We have the same issue and my debugging leads to the same solution as mentioned above.

cbarcroft:
Incase it's helpful for anyone, I've been using the following fix in my project for a couple of versions now (works in at least 0.27.1 through 0.29.0).

Replace [this line](https://github.com/feast-dev/feast/blob/d3b7c060bdc2a4d2b19dbd4e0dd8f35e76589d5f/sdk/python/feast/infra/registry/sql.py#L819) with the following:

```
                    protos = []
                    for obj in objs:
                        proto = obj.to_proto()
                        if hasattr(proto, "spec"):
                            proto.spec.project = project
[...]

PR #3482, opened 2023-02-05
fix: Fix SQL Registry cache miss

### LINKED ISSUE #3457: Problems to retrieve online data (get_online_features with Redis) on v.0.28.0: it says the feature view does not exist
Contour definitions:
- A FeatureStore object named "store" is properly instatiated
- A Feature Service object named "feature_service" is properly initialized

## Expected Behavior 
By using the method "get_online_features()", we expect to retrieve data defined in feature_service.
The example uses data from open breast_cancer database, and the method looks like:
```
online_response = store.get_online_features(
    features=feature_service,
    entity_rows=[{"patient_id": 0}],
)
```
which should return (for example) : {'patient_id': [0], 'my_feature': [184.60000610351562]}

## Current Behavior
This works as expected with versions 0.27 and below. By changing to feast version 0.28, we started to see an error
```
feast.errors.FeatureViewNotFoundException: Feature view d does not exist in project feature_repo
```
Important: the same feature_service is used to retrieve data from the offline store, and this process specifically works as expected.
In other words: "get_historical_features" works as expected, while "get_online_features" not, for the same features as input.
We already tried to use names of Feature Views and Features instead of the Feature Service, but without success.

## Steps to reproduce
- Using PostgreSQL for Offline Store and Registry, hosted on Azure
- Using Azure Cache for Redis for the Online Store
- Try to reproduce the same environment and then check if the aforementioned code works on Feast v0.28

### Specifications

- Version: 0.28
- Platform: Windows NT x64 10.0.17763
- Subsystem: Python 3.9.4 x64

adchia:
There were issues reading from the SQL registry cache because the self.proto() method didn't populate the project in each of the object protos.

This also adds a test for reading from the SQL cache

Fixes #3457 #3430

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -166,9 +166,9 @@ class FeatureStore:
 
         registry_config = self.config.get_registry_config()
         if registry_config.registry_type == "sql":
-            self._registry = SqlRegistry(registry_config, None)
+            self._registry = SqlRegistry(registry_config, self.config.project, None)
         else:
-            r = Registry(registry_config, repo_path=self.repo_path)
+            r = Registry(self.config.project, registry_config, repo_path=self.repo_path)
             r._initialize_registry(self.config.project)
             self._registry = r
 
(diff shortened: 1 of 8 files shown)


felixwang9817 (approved):
/lgtm

PR #3498, opened 2023-02-15
Update quickstart.md

Harpreet-Randhawa:
Adding an instruction if people have python3 installed and don't realise how to use pip with Python3

Fixes #

### DIFF
--- a/docs/getting-started/quickstart.md
+++ b/docs/getting-started/quickstart.md
@@ -45,6 +45,15 @@ pip install feast
 {% endtab %}
 {% endtabs %}
 
+Using Python3,
+{% tabs %}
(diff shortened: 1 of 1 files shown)


achals:
I'm not sure this is the right instruction, since we *only* support python3.7+ and don't support python 2 at all.. Are the current instructions incorrect or do they produce an error?

Harpreet-Randhawa:
Hello Achal

I tried installing feast with pip but my system doesnt recognise pip. I am
able to install it with pip3 install feast command but now the system
doesn't recognise feast command. So I am unable to proceed. Possibly
something with my config file or path variable. I am still trying to figure
it out.

Harpreet

PR #3504, opened 2023-02-21
fix: Support param timeout when persisting

sudohainguyen:
**What this PR does / why we need it**: to support declaring timeout when persisting historical retrieval job

Fixes #

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -456,7 +456,7 @@ class BigQueryRetrievalJob(RetrievalJob):
     def to_bigquery(
         self,
         job_config: Optional[bigquery.QueryJobConfig] = None,
-        timeout: Optional[int] = 1800,
+        timeout: Optional[int] = None,
         retry_cadence: Optional[int] = 10,
     ) -> str:
         """
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -456,8 +456,8 @@ class BigQueryRetrievalJob(RetrievalJob):
     def to_bigquery(
         self,
         job_config: Optional[bigquery.QueryJobConfig] = None,
-        timeout: int = 1800,
-        retry_cadence: int = 10,
+        timeout: Optional[int] = 1800,
+        retry_cadence: Optional[int] = 10,
     ) -> str:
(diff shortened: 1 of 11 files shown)


gbmarc1:
Hello @sudohainguyen !
I am working on a similar [PR](https://github.com/feast-dev/feast/pull/3505) but for `to_df` and `to_arrow` methods.
- I think it would be good to add unit-tests to those methods.
- What about `Optional[int]` with default None if you don't want a timeout.

sudohainguyen:
> Hello @sudohainguyen ! I am working on a similar [PR](https://github.com/feast-dev/feast/pull/3505) but for `to_df` and `to_arrow` methods.
> 
> * I think it would be good to add unit-tests to those methods.

sure let me see

sudohainguyen:
> * What about `Optional[int]` with default None if you don't want a timeout.
@gbmarc1 there should always be a timeout I believe

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -454,8 +454,8 @@ def to_sql(self) -> str:
     def to_bigquery(
         self,
         job_config: Optional[bigquery.QueryJobConfig] = None,
-        timeout: int = 1800,
-        retry_cadence: int = 10,
+        timeout: Optional[int] = 1800,
+        retry_cadence: Optional[int] = 10,
```
achals, line 458: I think this will be stomped over by #3505

sdk/python/feast/infra/offline_stores/offline_store.py
```diff
@@ -179,7 +179,12 @@ def on_demand_feature_views(self) -> List[OnDemandFeatureView]:
         pass
 
     @abstractmethod
-    def persist(self, storage: SavedDatasetStorage, allow_overwrite: bool = False):
+    def persist(
+        self,
+        storage: SavedDatasetStorage,
+        allow_overwrite: Optional[bool] = False,
+        timeout: Optional[int] = 1800,
```
achals, line 186: I'd prefer the default value to be `None` - implementations that support a timeout can handle it explicitly

sudohainguyen:
just rebased

sdk/python/feast/infra/offline_stores/bigquery.py
sudohainguyen: set default to `None`: done

PR #3505, opened 2023-02-22
feat: Adding query timeout to `to_df` and `to_arrow` retrieval methods

gbmarc1:
- Introduces a `timeout` argument for `RetrievalJob` class `to_df` and `to_arrow` methods. In the bigquery implementation, the timeout was hardcoded to 1800s using those 2 functions. We want to make it configurable.
- Also, adds some base unit tests for all retrieval classes.


sudohainguyen:
I think we shouldn't bump libs versions in this PR 😉

achals (approved):
/ok-to-test
/lgtm

PR #3514, opened 2023-03-02
fix: UI working behind base url

### LINKED ISSUE #3481: feast ui does not work on proxy subpath
## Expected Behavior 

Feast UI should work when it is served behind a proxy, on a subpath e.g. `/feast-ui`

## Current Behavior

Parts of the feast UI works behind a subpath, but not entirely (nothing is displayed, just the feast logo with a "404" text - refer to screenshot). No requests in the network tab of the web browser are hitting 404.

image

## Steps to reproduce

Serve feast UI as you would e.g. `feature_store.serve_ui()`, optionally passing in the `root_path` parameter (it does not help).

Set up an nginx pod with the following configuration (i.e. the nginx pod should have `/etc/nginx/conf.d/default.conf` with the following contents - `dummy_project` is the project name, and `http://feast-ui-service:8080` is where the feast UI can be accessed from your nginx pod / container):

```
server {
  listen 80 default_server;

  location = /feast-ui/ {
      rewrite (.*) /feast-ui/p/dummy_project permanent;
  }

  location /feast-ui/ {
      proxy_pass http://feast-ui-service:8080/;
  }

  location / {
      proxy_pass http://feast-ui-service:8080/;
  }
}
```

This configuration works on localhost when nginx can listen on the root path `/`. However, note that the URL after all the redirects is wrong (it does not have the prefix).

- The first block is required to force a redirect to the `/p/{project_name}`. Without this, the page will display 404 as above.
- The second block is required to strip away `/feast-ui` so the UI app does not receive that path that it is not aware of
- The third block is a trick to make this setup work in a local environment, because the app itself will redirect the user back to `/p/dummy_project` (without the prefix), which we then proxy into the feast UI app. However, in an actual environment, this setup does not work, because when the url does not contain the `/feast-ui` prefix, the ingress will not route it to the nginx pod, so the nginx pod cannot proxy the connection to the right place.

Ideally, if the feast ui app is capable of being served on a subpath, only the second `location` block should be required in the nginx configuration. The first and third `location` blocks are workarounds.

### Specifications

- Version: 0.29.0

## Possible Solution

The app should redirect to relative and not absolute paths

sudohainguyen:
**What this PR does / why we need it**: resolve base URL in react code

Fixes #3481

### DIFF
--- a/sdk/python/feast/ui_server.py
+++ b/sdk/python/feast/ui_server.py
@@ -15,7 +15,7 @@ def get_app(
     store: "feast.FeatureStore",
     project_id: str,
     registry_ttl_secs: int,
-    root_path: str = ""
+    root_path: str = "",
 ):
     app = FastAPI()
 
--- a/sdk/python/feast/ui_server.py
+++ b/sdk/python/feast/ui_server.py
@@ -13,11 +13,9 @@ import feast
 
 def get_app(
     store: "feast.FeatureStore",
-    get_registry_dump: Callable,
     project_id: str,
     registry_ttl_secs: int,
-    host: str,
-    port: int,
+    root_path: str = ""
 ):
     app = FastAPI()
 
@@ -62,7 +60,7 @@ def get_app(
                     "name": "Project",
                     "description": "Test project",
                     "id": project_id,
-                    "registryPath": "/registry",
+                    "registryPath": f"{root_path}/registry",
                 }
             ]
         }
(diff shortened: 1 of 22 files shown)


adchia (approved):
/lgtm

shumway743:
We had a similar requirement to run feast ui behind a proxy and were able to make it work simply by adding the basename attribute to the BrowserRouter tag in FeastUI.tsx and setting homepage in package.json (granted this was an older release ~0.19.0). It seems a bit cumbersome and error prone to update each url as is done here. It would be nice if basename was exposed as a prop on the FeastUI component and then forwarded to the router to avoid having to do a custom build for a single attribute.

sudohainguyen:
Cool, would you like to make some changes here?

shumway743:
I created a draft PR [here](https://github.com/feast-dev/feast/pull/3541) showing how this can be done with the basename prop. Let me know your thoughts. Biggest benefit is probably maintainability.

PR #3519, opened 2023-03-03
chore: Update python dependencies and use the latest pip

achals:

The pinned version of pip seems like it's incompatible with the version of pip-tools that we're using.

Fixes #


felixwang9817 (approved):
/lgtm

PR #3520, opened 2023-03-03
chore: Stop compiling embedded go components

achals:

This basically doesn't work and I doubt it ever has. At this point it makes our build process more complicated and is dead code 

Fixes #


felixwang9817 (approved):
/lgtm

PR #3524, opened 2023-03-08
fix: Feature view `entities` from_proto type

gbmarc1:
fix: Converts `feature_view_proto.spec.entities` from `google.protobuf.internal.containers.RepeatedScalarFieldContainer` to `list`

`google.protobuf.internal.containers.RepeatedScalarFieldContainer` is not serializable and still a proto not a list.

image


achals:
@gbmarc1 you need to sign off your commits!

PR #3527, opened 2023-03-10
build: Looser upper bound on pyarrow dependency

gbmarc1:
Pyarrow==7.0.0 and 8.0.0 + homebrew python@3.10 do not like each [other](https://github.com/Homebrew/homebrew-core/issues/125329) on MacOS ventura.

Unbounding pyarrow to greater than 8 solves the issue.


vtao2:
Amazing

achals (approved):
/ok-to-test
/Lgtm

PR #3542, opened 2023-03-17
fix: Pin typeguard to 2.13.3 which is what we are currently using.

kevjumba:

Newest typeguard release has breaking changes. This pins typeguard to newest compatible release.

Fixes #


adchia (approved):
/lgtm

Issue #3352, opened 2022-11-20
the go server is deprecated?


laiyongtao:
## Current Behavior
https://github.com/feast-dev/feast/blob/ef0ed083d7f31a48b9548c5774eba54a4d647bce/sdk/python/feast/feature_store.py#L2275
### Specifications

- Version:
- Platform:
- Subsystem:

PR #3552, opened 2023-03-22
fix: Init go server triggering

phil-park:

Unlike the [documentation](https://docs.feast.dev/reference/feature-servers/go-feature-server)
go_feature_serving: Go server will not run with only True flag.

The existing code assumed the situation where _lazy_init_go_server() is executed and self._go_server is not None, but with the usual feast serve --go

Fixes (https://github.com/feast-dev/feast/issues/3352)


achals:
I'm going to close this because I'm actually going to do a more major clean up here.

PR #3555, opened 2023-03-24
ci: Use build --wheel instead of cibuildwheel

achals:

Since we've deleting all the go code, we do not need platform specific builds anymore. We should exploit this fact by building universal images. Anyway, cibuildwheel barfs when trying to build platform specific builds anyway.

Fixes #


adchia (approved):
/lgtm

PR #3562, opened 2023-03-24
fix: Use pyarrow in a way that works across versions

achals:

Fixes #


kevjumba (approved):
/lgtm

Issue #3568, opened 2023-03-27
SQL Registry writes to public PostGres Schema


zerafachris:
**Is your feature request related to a problem? Please describe.**
Hi, I am looking at using an SQL registry for my Feast implementation. I have modified the `config.yaml` as per [](https://docs.feast.dev/tutorials/using-scalable-registry)  for PostGres. However, this is creating the registry inside of the public schema. <img width="287" alt="image" src="https://user-images.githubusercontent.com/22851907/228049699-754f7ee9-07a9-4e59-9379-953c1471527c.png">. 

I would like that the registry gets created inside of **feast_registry** . My connection string is `postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/${POSTGRES_DATABASE} `.

**Describe the solution you'd like**
I would like that the registry gets created inside of **feast_registry**. Is it possible to add **db_schema** inside of the config.yaml and this is used in the creation of the MetaData (https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/registry/sql.py#L189)

**Describe your attempts**
I have tried setting `?options=-csearch_path%3Dfeast_registry` as part of the connection string, to try and force pycog2 to use **feast_registry**  but no luck. I am getting
`sqlalchemy.exc.ProgrammingError: (psycopg2.errors.InvalidSchemaName) no schema has been selected to create in`

**Additional context**

zerafachris:
Kindly have a look at this PR https://github.com/feast-dev/feast/pull/3575

PR #3575, opened 2023-03-28
feat: Custom schema for SQL Registry

zerafachris:
Current SQL Registry uses public schema. This provides a **registry_schema** inside of config_yaml that will use _registry_schema_ for all registry related tables.

https://github.com/feast-dev/feast/issues/3568

Fixes #[3568](https://github.com/feast-dev/feast/issues/3568)


adamschmidt:
Have you considered using the `db_schema` property that's in the [pgsql connection config properties](https://github.com/feast-dev/feast/blob/2c04ec175f9155c906f90502bffe1bd5a5619ddb/sdk/python/feast/infra/utils/postgres/postgres_config.py#L12)? This same configuration class is used for the online store, so this should work too.

Adding a specific schema name seems redundant if you use this.

zerafachris:
Hi @adamschmidt , the feature handles the metadata object inside of PostGres. I can use **_db_schema_**  instead of **_registry_schema_** no worries, however the class model is not being used.  All objects that are supplied in the config are parsed independently.

zerafachris:
@adamschmidt bump re above ☝️

adchia:
Assigning to @achals since he's more familiar with this code!

zerafachris:
Conflicts have been resolved. Can we get a review

zerafachris:
Hi @achals , do you think we can get a peer review? I do not wish to stay resyncing with master branch and resolve conflicts again.

jeremyary:
Hi @zerafachris, with new maintainers in place we're making best efforts to try and resolve long-standing conversations such as what you've encountered here. With that in mind, we'd still have a few things to work through prior to merging your contribution. As a start, our DCO check indicates [here](https://github.com/feast-dev/feast/pull/3575/checks?check_run_id=15427097214) some fixes that are needed. If you're still interested in persuing the PR, please take a look at this & other running checks. Thanks!

PR #3578, opened 2023-03-29
ci: Add 'rockset' as feast pip install extra

danielin917:
Allows users to install Rockset specific packages with pip

Fixes #


achals (commented):
You prob need to regenerate the requirements files

adchia (approved):

PR #3584, opened 2023-04-05
fix: Fix Snowflake template

sfc-gh-madkins:

Fixes snowflake template so feature store instantiation happens after feast apply

Fixes #


adchia (approved):
/lgtm

PR #3585, opened 2023-04-05
feat: Relax aws extras requirements

cburroughs:
boto3 1.20.23 was released in late 2021. There have been many releases since that downstream projects would reasonably want to use.

s3fs is a particularly thorny dependency because each version of s3fs depends on a very narrow range of aiobotocore versions.  s3fs (as opposed to pyarrow._s3fs) does not appear to be used directly for AWS related code.  If users want to avail themself of pandas<-->s3 integration with s3fs they can still install it directly.


achals (approved):
/ok-to-test

Can you regen the requirements file as well?

achals:
> 

Yes, that's the one

PR #3588, opened 2023-04-05
fix: Make snowflake to remote tables temporary

### LINKED ISSUE #3545: to_remote_storage() resulting in undeleted temporary tables in Snowflake
## Expected Behavior 
When calling get_historical_features.to_remote_storage(), any temporary tables created in Snowflake are deleted after the Snowflake session ends.
## Current Behavior
When calling get_historical_features.to_remote_storage(), the temporary tables created during the join process are not deleted after the Snowflake session ends. These tables are set to a retention time of 1 day, but they are not deleted and still exist after 24 hours.

I tested this with `to_df()` and the above described issue does not occur. I also tried explicitly ending the session to make sure that wasn't the issue, but even after confirming the session was ended, the issue still persists.

## Steps to reproduce
1. For the FeatureStore object, set the RepoConfig offline store config to specify the following:
```json
{
     "blob_export_location": <s3_staging_url>,
     "storage_integration_name": <storage_integration>,
     "role": <stage_role>,
     "schema_": <stage_schema>,
}
```
2. Call `get_historical_features(entity_df=entity, features=features, full_feature_names=True).to_remote_storage()`
3. Check snowflake stage tables and look for tables created at the time of running that start with `temporary_`

### Specifications

- Version: 0.30.0
- Platform: Ubuntu
- Subsystem:

## Possible Solution
No possible solution known at the time of reporting

sfc-gh-madkins:

Fixes #3545

### DIFF
--- a/sdk/python/feast/infra/offline_stores/snowflake.py
+++ b/sdk/python/feast/infra/offline_stores/snowflake.py
@@ -556,7 +556,7 @@ class SnowflakeRetrievalJob(RetrievalJob):
             )
 
         table = f"temporary_{uuid.uuid4().hex}"
-        self.to_snowflake(table)
+        self.to_snowflake(table, temporary=True)
 
         query = f"""
             COPY INTO '{self.export_path}/{table}' FROM "{self.config.offline_store.database}"."{self.config.offline_store.schema_}"."{table}"\n


adchia (approved):
/lgtm

PR #3593, opened 2023-04-08
fix: Support param timeout when persisting

sudohainguyen:
**What this PR does / why we need it**: to support declaring timeout when persisting historical retrieval job

Fixes #

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -456,8 +456,8 @@ class BigQueryRetrievalJob(RetrievalJob):
     def to_bigquery(
         self,
         job_config: Optional[bigquery.QueryJobConfig] = None,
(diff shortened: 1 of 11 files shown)


sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -456,8 +456,8 @@ def to_sql(self) -> str:
     def to_bigquery(
         self,
         job_config: Optional[bigquery.QueryJobConfig] = None,
-        timeout: int = 1800,
-        retry_cadence: int = 10,
+        timeout: Optional[int] = None,
```
adchia, line 459: can you keep the default value? otherwise, this might become a breaking change

sudohainguyen: done

adchia (approved):
/lgtm

PR #3595, opened 2023-04-12
feat: Add AWS Redshift Serverless support

cbarcroft:
This adds support for optionally using a serverless workgroup rather than a provisioned cluster when utilizing the AWS Redshift offline store.  Ultimately it is just a change in arguments given to boto3 functions (WorkgroupName vs ClusterIdentifier & DbUser), so this adds a configuration field, some validation, and passes the new field through to relevant functions, before ultimately deciding which keyword args to pass into boto3.

N/A


cbarcroft:
Trying to follow the development guide, but two problems;  I don't seem to be able to add labels, and I did not see a field to add a release note?

setup.py
```diff
@@ -89,7 +91,7 @@
     "hiredis>=2.0.0,<3",
 ]
 
-AWS_REQUIRED = ["boto3>=1.17.0,<=1.20.23", "docker>=5.0.2", "s3fs>=0.4.0,<=2022.01.0"]
+AWS_REQUIRED = ["boto3>=1.24.11,<=1.24.95", "docker>=5.0.2", "s3fs>=0.4.0,<=2022.01.0"]
```
adchia, line 94: there's a separate PR on #3585 . After that goes in, want to rebase? You'll also have to run `make lock-python-ci-dependencies PYTHON=x` and `make lock-python-dependencies PYTHON=x`

adchia: The PR has landed now!

adchia (approved):
/lgtm

cbarcroft:
Sorry, looks like there was a casing error on one of the Redshift args

Issue #3591, opened 2023-04-07
feat: feast teardown command brings all projects down regardless of which feature_repo you are in in shared registry


jmeisele:
## Expected Behavior 
`feast teardown` command in the feature_repo will remove resources _only_ tied to the project listed in the corresponding `feature_store.yaml`

## Current Behavior
`feast teardown` command in any feature repo referencing the same registry removes all projects resources under that registry

## Steps to reproduce
You can use the same GItHub repo or spread across multiple repos.
1. Create feature repo 1
- `feast init -t snowflake snowflake_feast`
- Fill out generated snowflake cred prompts in CLI
- cd snowflake_feast/feature_repo
- `feast apply`

2. Create feature repo 2
- `feast init -t snowflake concrete_opossum`
- Fill out generated snowflake cred prompts in CLI
- cd concrete_opossum/feature_repo
- `feast apply`

You'll see both projects listed in registry metadata (as expected).

![image](https://user-images.githubusercontent.com/29043118/230666005-fe9d4522-504c-4809-ba1e-bad4e27697ef.png)

3. Perform `feast teardown` in either feature_repo directory we generated earlier

### Specifications

- Version: 30.2
- Platform: Mac | Linux
- Subsystem:

## Possible Solution
`feast teardown-project` command only removes resources tied to the project listed in the feature_repo/feature_store.yaml it is run from and leaves the other projects infra alone in the registry.

jmeisele:
I re-read the docs to see if there's something I'm missing using the SQL based registry (Snowflake). Bullet 2 references tearing down a project (desired behavior I'm after). I didn't see a CLI command to `teardown_project` or anything similar. The `teardown` [command](https://github.com/feast-dev/feast/blob/2c04ec175f9155c906f90502bffe1bd5a5619ddb/sdk/python/feast/feature_store.py#L1002-L1013) does indeed remove the registry.

![image](https://user-images.githubusercontent.com/29043118/230913938-2804950c-6b8a-418f-b521-816f200b77c3.png)

wjhhuizi:
I don't think the multi-project implementation for the FEAST working really well in the registry at this moment... there need some code works for that part...

sfc-gh-madkins:
@jmeisele can you retag this as an improvement? given this is currently not a bug

jmeisele:
> @jmeisele can you retag this as an improvement? given this is currently not a bug

Tried to change the label from GitHub and I cannot ?

[...]

PR #3598, opened 2023-04-13
feat: Add teardown-project command

jmeisele:

This PR adds an additional CLI command `teardown-project`. The existing `teardown` command brings _ALL_ infrastructure down declared in the feature registry, including the registry itself. There are many instances when projects are using a shared registry for discoverability on the UI and they need the ability to bring down the project level infrastructure without interfering with each other. 

Which issue(s) this PR fixes:

Fixes https://github.com/feast-dev/feast/issues/3591

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -1013,6 +1013,18 @@ class FeatureStore:
         self._registry.teardown()
 
     @log_exceptions_and_usage
+    def teardown_project(self):
+        """Tears down all local and cloud resources for the feature store project defined."""
+        tables: List[FeatureView] = []
+        feature_views = self.list_feature_views()
+
+        tables.extend(feature_views)
+
+        entities = self.list_entities()
+
+        self._get_provider().teardown_infra(self.project, tables, entities)
+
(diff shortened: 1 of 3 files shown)


sfc-gh-madkins:
@jmeisele dont we need to add logic that deletes all the registry data associated with the specific project as well?

sdk/python/feast/feature_store.py
```diff
@@ -1012,6 +1012,18 @@ def teardown(self):
         self._get_provider().teardown_infra(self.project, tables, entities)
         self._registry.teardown()
 
+    @log_exceptions_and_usage
+    def teardown_project(self):
```
adchia, line 1016: this _seems_ wrong since the registry file or sql registry itself still holds the references to the project. Probably better is to have each registry store also support tearing down references specific to a project?

jmeisele: 💯 it's not as simple as first expected since the registries don't have parity in their behavior. I changed the PR to draft to chew on it a big longer and think through how each type of registry would solve this pattern. Thanks for the feedback @adchia 🙌🏻

PR #3614, opened 2023-05-01
fix: Fix timestamp consistency in push api

### LINKED ISSUE #3613: Time unit mismatch issue when calling Bigquery Push API
## Expected Behavior 

When data is ingested with the push api, the same time series data must be added to the online store and the offline store (bigquery).

## Current Behavior

When calling the Push API with the BigQuery Offline Store, if the timestamp field exists, the pyarrow timestamp unit in the code defaults to us (microsecond).

However, when bigquery load_table is executed, the timestamp column operates in ns (nano second), so the unit is inconsistent.

Because there is a time difference of 1000 times, bigquery actually inputs 1/1000 times the time, so if you put the current time, the old time of the 1970s is entered.

## Steps to reproduce

After creating Bigquery Push Source with timestamp field, call Push API (w/ONLINE_AND_OFFLINE).

### Specifications

- Version:
- Platform:
- Subsystem: BigQuery

## Possible Solution

When converting to pyarrow schema, the time unit is input according to the offline store.

phil-park:

When calling the Push API with the BigQuery Offline Store, if the timestamp field exists, the pyarrow timestamp unit in the code defaults to us (microsecond).

However, when bigquery load_table is executed, the timestamp column operates in ns (nano second), so the unit is inconsistent.

Because there is a time difference of 1000 times, bigquery actually inputs 1/1000 times the time, so if you put the current time, the old time of the 1970s is entered.

Fixes #3613

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -363,7 +363,7 @@ class BigQueryOfflineStore(OfflineStore):
         assert isinstance(feature_view.batch_source, BigQuerySource)
 
         pa_schema, column_names = offline_utils.get_pyarrow_schema_from_batch_source(
-            config, feature_view.batch_source
+            config, feature_view.batch_source, timestamp_unit="ns"
         )
         if column_names != table.column_names:
             raise ValueError(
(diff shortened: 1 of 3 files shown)


adchia (approved):
/lgtm

phil-park:
@adchia
Sorry, I just pushed another fix to my repository and the lgtm tag was removed. I rolled back now, but please approve again.

adchia (approved):
/lgtm

PR #3626, opened 2023-05-11
feat: Replace Identity Comparison with Equality Comparison in `parse_repo`

### LINKED ISSUE #3312: feast duplicates sourses
I use feast with local storage. 
my project structure:
```
data/
data/features.parquet
feature_repo/
feature_repo/entities.py
feature_repo/sources.py
feature_repo/views.py
feature_repo/feature_store.yaml
...
```

When I start "feast -c feature_repo plan" it throws an feast.errors.DataSourceRepeatNamesException: "Multiple data sources share the same case-insensitive name my_awesome_source." 

I store source's definition in the separate file `feature_repo/sources.py` and import them from that file in file `feature_repo/views.py`. For example:
```
from feature_repo import sources, entities

my_awesome_view= FeatureView(
    name="awesome_view",
    entities=[entities.id],
    ttl=dt.timedelta(days=3),
    schema=[
        Field(...),
    ],
    online=True,
    source=sources.my_awesome_source
    tags={},
)
```
Feast adds this source twice. Firstly it parses file 'sourses.py' and adds from there and secondly from 'views.py'

## Expected Behavior 
Don't add the same source twice

## Current Behavior
feast add source twice 

## Steps to reproduce
Reconstruct project structure defined before (sourses and views definitions should be in the repo folder, but different files), then start "feast plan".

### Specifications

- Version: 0.26.0
- Platform: mac m1, macOS Monterey 12.6 
- Subsystem:

## Possible Solution
The problem is in function `parse_repo` (https://github.com/feast-dev/feast/blob/master/sdk/python/feast/repo_operations.py#L99). This function parses primarily 'sources.py' and gets 'my_awesome_source' from there. Then the function parses 'views.py' and gets 'my_awesome_source' as an attribute of 'my_awesome_view'. That wat it adds the same source twice, and parse function sees them as different objects, because the source from views file is imported.
My guess is that the check ```elif isinstance(obj, BatchFeatureView) and not any(``` (https://github.com/feast-dev/feast/blob/master/sdk/python/feast/repo_operations.py#L167) is wrong. ``` batch_source == ds```  works better. 
Or maybe you need special comparison function.

srausser:
This Pull Request updates the parse_repo function in `repo_operations.py`, transitioning from the identity operator `is` to the equality operator `==` for object comparisons.

Previously, the `parse_repo` function utilized the `is` operator to compare Feast objects such as data sources, feature views, and entities. The `is` operator compares if two variables point to the exact same object in memory. However, this approach can be overly restrictive.

When Feast objects are imported across different parts of the repository definition files, they may not share the same memory address, even though they represent the same logical object. This discrepancy led to erroneous distinctions between objects that should have been recognized as identical.

To address this, I've replaced the `is` operator with `==`. The `==` operator evaluates whether two variables are equal in value, independent of their memory addresses. This adjustment enhances the flexibility and accuracy of object comparison within `parse_repo`, permitting objects imported across the repository to be correctly recognized as equal when their values match.

Fixes #3312

### DIFF
--- a/sdk/python/feast/repo_operations.py
+++ b/sdk/python/feast/repo_operations.py
@@ -122,7 +122,7 @@ def parse_repo(repo_root: Path) -> RepoContents:
             obj = getattr(module, attr_name)
 
             if isinstance(obj, DataSource) and not any(
-                (obj is ds) for ds in res.data_sources
+                (obj == ds) for ds in res.data_sources
             ):
                 res.data_sources.append(obj)
 
@@ -135,12 +135,12 @@ def parse_repo(repo_root: Path) -> RepoContents:
                     batch_source = obj.batch_source
 
                     if batch_source and not any(
-                        (batch_source is ds) for ds in res.data_sources
+                        (batch_source == ds) for ds in res.data_sources
                     ):
                         res.data_sources.append(batch_source)
             if (
                 isinstance(obj, FeatureView)
-                and not any((obj is fv) for fv in res.feature_views)
+                and not any((obj == fv) for fv in res.feature_views)
                 and not isinstance(obj, StreamFeatureView)
                 and not isinstance(obj, BatchFeatureView)
             ):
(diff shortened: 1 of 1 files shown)


srausser:
closing since i can't seem to replicate the errors im seeing in my production repository

PR #3628, opened 2023-05-16
fix: Fix scan datasize to 0 for inference schema

### LINKED ISSUE #3433: Schema inference of BQ OfflineStore costly
## Expected Behavior 
When making schema inference, I expect there would be an extra minor cost when scanning tables.

## Current Behavior
BQ OfflineStore made a full scan on the entire table, although we are using `limit` statement. According to GCP documentation:
`Applying a LIMIT clause to a SELECT * query does not affect the amount of data read. You are billed for reading all bytes in the entire table, and the query counts against your free tier quota.`

https://cloud.google.com/bigquery/docs/best-practices-costs

## Steps to reproduce
1. Prepare a large feature table
2. feast apply as a feature view

### Specifications

- Version: 0.28.0
- Platform: Linux
- Subsystem:

## Possible Solution
We can add a filter by `timestamp_field` by modifying this [line](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/offline_stores/bigquery_source.py#L183) even though there is no data, the schema could be inferred eventually

phil-park:

For bigquery schema inference, query the table (or query) data with LIMIT 1 and obtain a schema object.

LIMIT 1 actually returns one record, but the entire table is full scanned and costs are incurred.

For simple schema inference, it makes sense to get the schema with LIMIT 0 without scan bytes and billing.

Fixes #3433

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery_source.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery_source.py
@@ -181,8 +181,8 @@ class BigQuerySource(DataSource):
                 raise TypeError("Could not parse BigQuery table schema.")
         else:
             bq_columns_query = f"SELECT * FROM ({self.query}) LIMIT 0"
-            queryRes = client.query(bq_columns_query).result()
-            schema = queryRes.schema
+            query_res = client.query(bq_columns_query).result()
+            schema = query_res.schema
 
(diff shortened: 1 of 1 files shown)


adchia (approved):
/lgtm

phil-park:
@adchia I forgot to sign the code, so I added it and pushed it again. Please check again.

Issue #3611, opened 2023-04-28
SQL registry disconnections


jiewpeng:
## Expected Behavior 

SQL registry works.

## Current Behavior

When a SQL registry is used, there is a chance that we may hit a `psycopg2.OperationalError: SSL connection has been closed unexpectedly` issue. This can happen due to the connection being closed for being idle for too long.

## Steps to reproduce

Use a SQL registry and set a connection idle timeout at the DB. Create a feature store object to connect to this registry, and wait for the timeout (i.e. don't query the feature store). Then try to query the feature store.

### Specifications

- Version: 0.25.0 (but should not matter)
- Platform: Linux

## Possible Solution

When creating the SQL registry, use `self.engine = create_engine(..., pool_pre_ping=True)`. See https://docs.sqlalchemy.org/en/14/core/pooling.html#pool-disconnects

tokoko:
@jeremyary looks like this was fixed and accidentally left open.

blopezpi:
@jeremyary @tokoko We encountered the same issue with Feast version 0.35.0. It seems the issue was resolved for using PostgreSQL as the online store, but when utilizing PostgreSQL as a registry, we cannot configure a connection pool. Instead, we've adjusted the Keepalive idle parameters to a larger value or used a proxy like `RDS Proxy` or `PgBouncer` to mitigate this error. This error occurs only when using the Python API Feature Server, and we need to adjust these values or attempt a retry to retrieve the features. Could we open this issue again? Thank you.

tokoko:
@blopezpi you're right, sorry about that. Have you tried the solution that's suggested in the issue? You can pass `{'pool_pre_ping': True}` as `sqlalchemy_config_kwargs` if you're using sql registry. Something like this if you're configuring from `feature_store.yaml`:
```
registry:
    registry_type: sql
    path: postgresql://postgres:mysecretpassword@127.0.0.1:55001/feast
    sqlalchemy_config_kwargs:
        pool_pre_ring: true
```

PR #3633, opened 2023-05-18
fix: Implements connection pool for postgres online store

Natan16:
When using a postgres online store, there's a chance the single connection times-out due to inactivity, so we get: 
    InterfaceError
    connection already closed 
The mainstream solution for that is implementing a connection pool with psycopg in order to guanratee the connections to the database are managed correctly.

https://github.com/feast-dev/feast/issues/3611

Fixes #

### DIFF
--- a/sdk/python/feast/infra/online_stores/contrib/postgres.py
+++ b/sdk/python/feast/infra/online_stores/contrib/postgres.py
@@ -1,9 +1,9 @@
+import contextlib
 import logging
 from collections import defaultdict
 from datetime import datetime
 from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple
 
-import contextlib
 import psycopg2
 import pytz
 from psycopg2 import sql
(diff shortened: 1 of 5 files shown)


sdk/python/feast/infra/online_stores/contrib/postgres.py
```diff
@@ -13,26 +14,28 @@
 from feast.feature_view import FeatureView
 from feast.infra.key_encoding_utils import serialize_entity_key
 from feast.infra.online_stores.online_store import OnlineStore
-from feast.infra.utils.postgres.connection_utils import _get_conn
+from feast.infra.utils.postgres.connection_utils import _get_connection_pool
```
avelino, line 17: We are forcing `pool` connection **by default**, we should leave it configurable by the user (he should choose the connection time that will be used), force `pool` connection in small project will be *"expensive"* (unnecessary resource usage).

**My suggestion** would be to let the user **choose** the connection type, **not `pool`** by *default* and the projects that need pool define it as pool

sdk/python/feast/infra/utils/postgres/postgres_config.py
```diff
@@ -6,6 +6,8 @@
 
 
 class PostgreSQLConfig(FeastConfigBaseModel):
+    min_conn: int = 1
+    max_conn: int = 10
```
avelino, line 10: create a connection definition variable that will be used from `_get_conn()`

**e.g:**
```py
type_conn: str = "normal" # or "pool"
```

avelino (changes requested):

avelino:
> Ok, I added back the possibility to use a single connection, now with the parameter `keepalives_idle` in case the database is configured to drop idle connections. 0 uses the system default.

@Natan16 great, I think it is important to write a unit test to validate the behavior that was implemented

@adchia are we going in the best direction for the project?

felixwang9817 (approved):
/lgtm

PR #3634, opened 2023-05-21
fix: Handle unknown postgres source types gracefully

msnidal:

Right now when trying to read a postgres table, if there are any unknown types *even if you're using a query on top of the table with only valid types* it will throw an exception. Technically you can avoid this by creating a view on top of your table but this can be a hassle when the custom query is serving this purpose anyways imo

With this fix, the only change is to surface unknown types as "unknown", which your code handles as-is with a warning and ignoring the column, allowing for usage of these tables without needing to support the arbitrary types themselves.

### DIFF
--- a/sdk/python/feast/type_map.py
+++ b/sdk/python/feast/type_map.py
@@ -873,13 +873,26 @@ def feast_value_type_to_pa(
 
 
 def pg_type_code_to_pg_type(code: int) -> str:
-    return {
+    """ Map the postgres type code a Feast type string
+
+    Rather than raise an exception on an unknown type, we return the
+    string representation of the type code. This way rather than raising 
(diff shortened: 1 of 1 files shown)


felixwang9817:
@msnidal thanks for the PR! would you mind [signing](https://docs.feast.dev/project/development-guide#signing-off-commits) your commit to pass the DCO check? also could you rebase on master? we had some dependencies issues that were breaking unit tests that were just recently fixed

adchia (approved):
/lgtm

msnidal:
Hey, sorry its been ages but I just got back to this, my rebase got a bit messy and had to sign hence the force pushes (no changes to the commit itself) but I managed to get up to date with main, let me know if this looks alright!

PR #3636, opened 2023-05-23
feat: Add gunicorn for serve with multiprocess

phil-park:
Required for better performance of serve api application.
Using only one uvicorn process per pod in a typical kubernetes environment is inefficient.
gunicorn is recommended to manage stable multi-process in fastapi. Therefore, add the function to use processes as many as the number of workers received as arguments. 

In addition, when using LoadBalancer, options related to keep alive, which must be handled with care, are added to the gunicorn argument.

Fixes #

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -70,6 +70,7 @@ REQUIRED = [
     "typeguard==2.13.3",
     "fastapi>=0.68.0,<1",
     "uvicorn[standard]>=0.14.0,<1",
+    "gunicorn",
     "dask>=2021.1.0",
     "bowler",  # Needed for automatic repo upgrades
     # FastAPI does not correctly pull starlette dependency on httpx see thread(https://github.com/tiangolo/fastapi/issues/5656).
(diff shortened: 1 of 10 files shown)


setup.py
```diff
@@ -70,6 +70,7 @@
     "typeguard==2.13.3",
     "fastapi>=0.68.0,<1",
     "uvicorn[standard]>=0.14.0,<1",
+    "gunicorn",
```
adchia, line 73: you probably need to regenerate some requirements with this with:
- `make lock-python-ci-dependencies PYTHON={3.8,3.9,3.10}`
- `make lock-python-dependencies PYTHON={3.8,3.9,3.10}`

phil-park:
@adchia Thanks for the advice. It worked, but in make lock-python-ci-dependencies, the cryptography version conflicted, so only gunicorn was added.
Is there any other way?

In addition, when using typing-extensions==4.6.0, it was confirmed that there was a bug in import_module in importer.import_class in the provider, so typing-extention was manually modified to 4.5.0.

setup.py
phil-park: @adchia Thanks for the advice.
When I ran the suggested command, the versions of other libraries were also changed, but only gunicorn was added first.

phil-park:
Along with the version up of other libraries, I found requirements.txt that I created incorrectly. Rolled back and only added gunicorn.

felixwang9817:
hey @phil-park sorry for the delay here - this PR looks ready to be merged, would you mind rebasing on master first? we fixed some dependencies issue recently that were causing all unit tests to fail

phil-park:
@felixwang9817 Thank you. I did a rebase.
However, a resource limit exceeded error occurs in a part unrelated to the code I committed.
Could you please check?

PR #3640, opened 2023-05-27
fix: Entityless fv breaks with `KeyError: __dummy` applying feature_store.plan() on python

wfoschiera:
using python SDK when try to execute `feature_store.plan(...)` using feature_views with and without entities, feast breaks looking for __dummy keys.

There is no open issue for this bug.

### DIFF
--- a/sdk/python/feast/inference.py
+++ b/sdk/python/feast/inference.py
@@ -1,5 +1,5 @@
 import re
-from typing import List, Set, Union
+from typing import List, Optional, Set, Union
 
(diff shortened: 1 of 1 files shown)


felixwang9817:
@wfoschiera thanks for the PR! would you mind rebasing on top of master? we had a dependencies issue that was causing unit tests to fail, and was just recently fixed

also could you run the linter with `make lint-python`?

wfoschiera:
@felixwang9817 let me know If you need any additional information.

adchia (approved):
/lgtm

PR #3645, opened 2023-06-06
feat: Add possibility to save dataset as table, when spark config has remote warehouse info

### LINKED ISSUE #3644: Saved_dataset for spark offline store can be accessed only within the scope of the spark session, where it was created.
I would like to have a possibility to use the data of the registered saved_dataset in a different spark session. Now I have only a name in the feast registry without the data if I create a new spark session. 

Part of persist function:

```
"""
Run the retrieval and persist the results in the same offline store used for read.
Please note the persisting is done only within the scope of the spark session.
"""
assert isinstance(storage, SavedDatasetSparkStorage)
table_name = storage.spark_options.table
if not table_name:
    raise ValueError("Cannot persist, table_name is not defined")
self.to_spark_df().createOrReplaceTempView(table_name)
```

Add possibility to save dataset as a table, for example when Spark Session config is included info about remote storage (hive, s3 path, etc)

Add an optional parameter for SparkOptions, which allows to save dataset as a table in any spark session configurations.

**Additional context**

nadejdaSuraeva:

This PR adds a possibility to use saved_dataset as a table for different spark session.  The existing solution saves the dataset’s name in registry, but the data exists only within the scope of the spark session. 

Fixes #3644

### DIFF
--- a/sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
+++ b/sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
@@ -352,13 +352,36 @@ class SparkRetrievalJob(RetrievalJob):
     ):
         """
         Run the retrieval and persist the results in the same offline store used for read.
-        Please note the persisting is done only within the scope of the spark session.
+        Please note the persisting is done only within the scope of the spark session for local warehouse directory.
         """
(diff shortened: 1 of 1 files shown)


felixwang9817 (approved):
/lgtm

PR #3646, opened 2023-06-06
feat: Use string as a substitute for unregistered types during schema inference

phil-park:

For column types that exist in bigquery but don't currently exist in Feast's typemap (e.g. DATE), most of them can be used as strings.
Currently, attempting schema inference will result in an error saying that there is no type.
In order to use schema inference more actively, it seems reasonable to infer these columns as string type.

Fixes #


felixwang9817 (approved):
/lgtm

PR #3667, opened 2023-07-03
Feat: Pydantic Model Compatibility

expediamatt:
[All tests and lint checks are passing.]

This PR introduces Pydantic model conversions for Entity, DataSource, and FeatureView, so that they may be transmitted over the wire using FastAPI. This allows a Feast project to be broken up into a local SDK and a remote registry service.

Only RequestSource and SparkSource are supported for now. We plan to add support for FeatureService models in the near future.


expediamatt:
Wrong branch, my apologies

PR #3680, opened 2023-07-11
fix: Optimize bytes processed when retrieving entity df schema to 0

sudohainguyen:
**What this PR does / why we need it**: Reduce bytes processed to 0 when retrieving entity df schema from BigQuery sql

Fixes #


adchia (approved):
/lgtm

Issue #3648, opened 2023-06-09
AmazonS3ClientBuilder assumed the Node role instead of ServiceAccount role 


harmeet-singh-discovery:
## Expected Behavior 
When Java server runs on AWS EKS pod, I get the following error:
`Caused by: com.amazonaws.services.s3.model.AmazonS3Exception: Access Denied (Service: Amazon S3; Status Code: 403; Error Code: AccessDenied;`

After debugging, I found that the S3Client that interacts with S3 assumed the Node role, instead of the ServiceAccount Role. 

## Current Behavior
The  S3Client that interacts with S3 assumed the Node role, instead of the ServiceAccount Role. 

### Specifications

- Version:
- Platform:
- Subsystem:

## Possible Solution
I made the following change and deployed the java service on AWS EKS. I was able to run the server without encountering above mentioned error.

```
@Provides
 public AmazonS3 awsStorage(ApplicationProperties applicationProperties) {
     AmazonS3ClientBuilder builder=AmazonS3ClientBuilder.standard().withRegion(applicationProperties.getFeast().getAwsRegion());  
    String roleArn = applicationProperties.getFeast().getRoleRegion();
    String webIdentityTokenFile = applicationProperties.getFeast().getToken();

    if (roleArn != null && webIdentityTokenFile != null) {
        WebIdentityTokenCredentialsProvider credentialsProvider = WebIdentityTokenCredentialsProvider.builder()
                .roleArn(roleArn)
                .webIdentityTokenFile(webIdentityTokenFile)
                .build();

        builder.withCredentials(credentialsProvider);
    }
    return builder.build();
  }
```

PR #3696, opened 2023-07-23
fix: Add aws-sts dependency in java sdk so that S3 client acquires IRSA role

harmeet-singh-discovery:
To use S3 based registry, Java server needs access to S3 bucket and objects. Currently, the default S3 client uses default chain to acquire credentials. At present, S3 client acquire the role associated with EKS Node, instead of IRSA role. 

This change is to include `aws-java-sdk-sts` dependency so that default credential chain uses web identity and assume IRSA Role.

Fixes https://github.com/feast-dev/feast/issues/3648

Reference: 
https://github.com/aws/aws-sdk-java/issues/2136

### DIFF
--- a/java/serving/pom.xml
+++ b/java/serving/pom.xml
@@ -244,6 +244,12 @@
     </dependency>
 
     <dependency>
+      <groupId>com.amazonaws</groupId>
+      <artifactId>aws-java-sdk-sts</artifactId>
+      <version>1.12.476</version>
+    </dependency>
+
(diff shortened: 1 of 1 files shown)


felixwang9817 (approved):
/lgtm

harmeet-singh-discovery:
@felixwang9817 
How do i merge this PR?

felixwang9817:
@harmeet-singh-discovery merged

Issue #3597, opened 2023-04-13
Online store latency (redis) is growing over time


RadionBik:
## Expected Behavior 
The online-store latency remains the same and doesn't grow over the time. We assume that amount of data in redis is not growing.

## Current Behavior
<img width="984" alt="image" src="https://user-images.githubusercontent.com/19979628/231738987-e79f848b-b7d7-4e57-a80b-26af120f1bc8.png">

The median latency growth over the past 30 days is on the chart above. 

## Steps to reproduce

We use a custom aiohttp python service to relay requests to the online-store, where we invoke the native feast client. The service is completely stateless, and I don't expect it to be the source of the problem.

here is our feast config:
```
config = RepoConfig(
    project='feast',
    registry='gs://our-bucket/feast/registry_file_3.db',
    provider='gcp',
    online_store=RedisOnlineStoreConfig(
        connection_string=get_redis_connection_string(),
        redis_type='redis',
        key_ttl_seconds=REDIS_TTL_SECONDS,
    ),
    offline_store=BigQueryOfflineStoreConfig(
        dataset='feature_store',
        project_id='project-id',
    ),
    entity_key_serialization_version=2,
)
```
as you see, we use the default registry cache TTL (=600).

### Specifications

- Version: 0.28.0
- Platform: amd64
- Subsystem: debian 10

## Possible Solution

We noticed that changing the path to file-based registry (i.e. effectively re-creating it) eliminates the latency growth and it back to normal (today's chart):
<img width="988" alt="image" src="https://user-images.githubusercontent.com/19979628/231740198-2f981445-c133-49da-89f4-9cabcccd2d36.png">

Therefore, a solution might be related to fixing the registry caching mechanism.

Let me know if further details are needed!

RadionBik:
I have noticed that timestamps of incremental materialization are stored in the registry and are sent to client as well. We run incr. mat. every 15 min, so over month it yields to 30 * 24 * 4 = 2880 timestamps per view, which might explain the gradual increase of the response time. Not sure if it is the reason, but decided to share the info here.

adchia:
Thanks for filing! do you mean that you changed from the file based registry to have a TTL=0 and it was ok?

RadionBik:
I have never adjusted registry cache TTL, it has always been set to the default value.

nturusin:
Hi guys. Any news on this? This issue forces us to update the registry file in production every couple of weeks to reset the latency..

nturusin:
[...]

PR #3702, opened 2023-07-28
feat: Apply cache to load proto registry for performance

### LINKED ISSUE #3649: performance issues of getting online features related to parsing protobuf data

The profiler of one application I am working on shows the `from_proto` method of the `FeatureView` class takes up 80% of the execution time. The result of cProfile is shown below.
```
    ncalls  tottime  percall  cumtime  percall filename:lineno(function)
         1    0.003    0.003   78.931   78.931 profiler.py:53(main)
       100    0.018    0.000   78.849    0.788 get_features.py:122(get_xxxx_features)
  1100/400    0.020    0.000   77.077    0.193 usage.py:274(wrapper)
 444200/9800    1.713    0.000   72.206    0.007 __init__.py:1030(wrapper)
       200    0.078    0.000   68.913    0.345 feature_store.py:1527(get_online_features)
       200    0.087    0.000   68.835    0.344 feature_store.py:1590(_get_online_features)
      3500    5.634    0.002   62.442    0.018 feature_view.py:369(from_proto)
       200    0.005    0.000   59.362    0.297 feature_store.py:2149(_get_feature_views_to_use)
       200    0.002    0.000   58.501    0.293 feature_store.py:281(_list_feature_views)
       200    0.001    0.000   58.499    0.292 registry.py:523(list_feature_views)
       200    0.016    0.000   58.495    0.292 proto_registry_utils.py:150(list_feature_views)
```
There are 3 feature views accessed by `get_xxxx_features`, however 100 calls of `get_xxxx_features` lead to 3500 calls of `from_proto`. There are 17 feature views in the feature store of this application and 3 of them are used by `get_xxxx_features`.

Environment: `continuumio/miniconda3:4.11.0` (linux/amd64) base image, `python==3.9.7`, `feast=0.31.1`, `protobuf==4.23.2`

Instead of caching the protobuf blob of `FeatureView`, cache the `FeatureView` python object in memory.

modify the `get_online_features` [method](https://github.com/feast-dev/feast/blob/v0.31.1/sdk/python/feast/feature_store.py#L1528) to
```python
def get_online_features(
    self,
    features: Union[List[str], List[FeatureView], FeatureService],
    entity_rows: List[Dict[str, Any]],
    full_feature_names: bool = False,
):
``` 
so that the application developer has the option to cache the `FeatureView` objects and use them to get features directly (by passing the `get_feature_views_to_use` [step](https://github.com/feast-dev/feast/blob/v0.31.1/sdk/python/feast/feature_store.py#L1606-L1613) )

phil-park:
If you use File Registry, the registry file becomes heavy because materialize records are accumulated while a long period.
Even if the registry cache is enabled, performance degraded because each online feature lookup traverses the registry object and searches the meta.

As long as the registry_proto object does not change, you must avoid multiple lookups through the function-level cache.

Fixes #3649 #3597

### DIFF
--- a/sdk/python/feast/infra/registry/proto_registry_utils.py
+++ b/sdk/python/feast/infra/registry/proto_registry_utils.py
@@ -1,4 +1,5 @@
 import uuid
+from functools import wraps
 from typing import List, Optional
 
 from feast import usage
@@ -21,6 +22,25 @@ from feast.protos.feast.core.Registry_pb2 import Registry as RegistryProto
 from feast.request_feature_view import RequestFeatureView
 from feast.saved_dataset import SavedDataset, ValidationReference
 from feast.stream_feature_view import StreamFeatureView
+
+
+def registry_proto_cache(func):
+    cache_key = None
+    cache_value = None
+
+    @wraps(func)
+    def wrapper(registry_proto: RegistryProto, project: str):
+        nonlocal cache_key, cache_value
+
+        key = tuple([id(registry_proto), registry_proto.version_id, project])
+
+        if key == cache_key:
+            return cache_value
+        else:
+            cache_value = func(registry_proto, project)
+            return cache_value
+
+    return wrapper
 
 
 def init_project_metadata(cached_registry_proto: RegistryProto, project: str):
@@ -137,6 +157,7 @@ def get_validation_reference(
     raise ValidationReferenceNotFound(name, project=project)
 
 
+@registry_proto_cache
 def list_feature_services(
     registry_proto: RegistryProto, project: str
 ) -> List[FeatureService]:
@@ -147,6 +168,7 @@ def list_feature_services(
     return feature_services
 
 
+@registry_proto_cache
 def list_feature_views(
     registry_proto: RegistryProto, project: str
 ) -> List[FeatureView]:
@@ -157,6 +179,7 @@ def list_feature_views(
     return feature_views
 
 
+@registry_proto_cache
 def list_request_feature_views(
     registry_proto: RegistryProto, project: str
 ) -> List[RequestFeatureView]:
@@ -169,6 +192,7 @@ def list_request_feature_views(
     return feature_views
 
 
+@registry_proto_cache
 def list_stream_feature_views(
     registry_proto: RegistryProto, project: str
 ) -> List[StreamFeatureView]:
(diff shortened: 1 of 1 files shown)


sdk/python/feast/infra/registry/proto_registry_utils.py
```diff
@@ -23,6 +24,25 @@
...
+    @wraps(func)
+    def wrapper(registry_proto: RegistryProto, project: str):
+        nonlocal cache_key, cache_value
+
+        key = tuple([id(registry_proto), registry_proto.version_id, project])
+
+        if key == cache_key:
+            return cache_value
+        else:
+            cache_value = func(registry_proto, project)
```
achals, line 40: Don't you need to update `cache_key` here as well?

sdk/python/feast/infra/registry/proto_registry_utils.py
```diff
@@ -23,6 +24,25 @@
 from feast.stream_feature_view import StreamFeatureView
 
 
+def registry_proto_cache(func):
+    cache_key = None
+    cache_value = None
```
achals, line 29: I'm not sure I understand this.. If you use the same annotation for multiple methods then the cache value needs have different shapes, such as `List[FeatureService]` or `List[FeatureView]` right? Is that going to work as expected here? Can we add tests?

sdk/python/feast/infra/registry/proto_registry_utils.py
phil-park: @achals I'm sorry, you're right. I added commit https://github.com/feast-dev/feast/pull/3702/commits/b8a373fa6cd48e0347322ed3685346c7ddb35f56

sdk/python/feast/infra/registry/proto_registry_utils.py
phil-park: The context of the wrapper decorator is created whenever different argument functions (e.g list_feature_views, list_feature_services..) are called. 
Therefore, when calling list_feature_views and when calling list_feature_services, different results are returned.
The operation of this cache is similar to Python's functools.lru_cache, but since the registryProto object is not hashable, it is implemented to hash using the object's id.

achals:
Okay, can you also sign off your commits? https://github.com/feast-dev/feast/pull/3702/checks?check_run_id=16298395815

phil-park:
I signed off the commit. Thank you for your review.

phil-park:
@achals Does this PR need further review?

felixwang9817 (approved):
/lgtm

PR #3734, opened 2023-08-15
fix: Fix python unit tests

### LINKED ISSUE #3731: Python unit tests failing - latest sql image uses newer version that is not caught in regex
## Expected Behavior 
Python unit tests pass (in test_sql_registry.py)

## Current Behavior
Tests currently fail (both locally and in cicd such as [here](https://github.com/feast-dev/feast/actions/runs/5866990613/job/15910343225?pr=3730)), as the mysql:latest image defined [here](https://github.com/feast-dev/feast/blob/7bc7c47b4507310850474290131c03fb6d480834/sdk/python/tests/unit/test_sql_registry.py#L83) now uses mysql Version 8.1.0, not 8.0.* as expected [here](https://github.com/feast-dev/feast/blob/7bc7c47b4507310850474290131c03fb6d480834/sdk/python/tests/unit/test_sql_registry.py#L94)

## Steps to reproduce
Run one of the failing tests, such as `pytest sdk/python/tests/unit/test_sql_registry.py::test_registry_cache[mysql_registry] -vvs` locally

## Possible Solution
A solution is to simply change this [line](https://github.com/feast-dev/feast/blob/7bc7c47b4507310850474290131c03fb6d480834/sdk/python/tests/unit/test_sql_registry.py#L94) from `log_string_to_wait_for = "/usr/sbin/mysqld: ready for connections. Version: '8.0.*'  socket: '/var/run/mysqld/mysqld.sock'  port: 3306"` to `log_string_to_wait_for = "/usr/sbin/mysqld: ready for connections. Version: '(\d+(\.\d+){1,2})'  socket: '/var/run/mysqld/mysqld.sock'  port: 3306"`. I have tested this change locally, and tests then pass.

Changing the version to regex is more flexible, as if we change it just to `'8.1.*'` then we will run into this issues again. The regex will match any number of digits with either 1 or two dots separating them. This is what we want for version numbers.

crispin-ki:
See #3731

Fixes #3731

Have confirmed locally that python unit tests all now pass with `make test-python`

### DIFF
--- a/sdk/python/tests/unit/test_sql_registry.py
+++ b/sdk/python/tests/unit/test_sql_registry.py
@@ -91,7 +91,7 @@ def mysql_registry():
     container.start()
 
     # The log string uses '8.0.*' since the version might be changed as new Docker images are pushed.
-    log_string_to_wait_for = "/usr/sbin/mysqld: ready for connections. Version: '(\d+(\.\d+){1,2})'  socket: '/var/run/mysqld/mysqld.sock'  port: 3306"
+    log_string_to_wait_for = "/usr/sbin/mysqld: ready for connections. Version: '(\d+(\.\d+){1,2})'  socket: '/var/run/mysqld/mysqld.sock'  port: 3306"  # noqa: W605
     waited = wait_for_logs(
         container=container,
         predicate=log_string_to_wait_for,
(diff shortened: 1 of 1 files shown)


adchia (approved):
Thanks!

/lgtm

PR #3735, opened 2023-08-15
fix: Update py3.8 ci requirements for cython 3.0 release

### LINKED ISSUE #3732: Bump versions for packages that now won't install due to Cython 3.0 release
## Expected Behavior 
When following the development setup instructions as per [here](https://github.com/feast-dev/feast/issues/2105), in step 5, you need to run `make install-python-ci-dependencies PYTHON=3.8`. You would expect this to install the dependencies with no issues.

## Current Behavior
When following the development setup instructions as per [here](https://github.com/feast-dev/feast/issues/2105), in step 5, you need to run `make install-python-ci-dependencies PYTHON=3.8`. However, since Cython 3.0 has been recently released, a number of packages now do not install successfully.

## Steps to reproduce
Just follow the instructions as per the guide linked above, until you get to the line in step 5 `make install-python-ci-dependencies PYTHON=3.8`, where most packages will install fine apart from fastavro, pyyaml and pymssql.

### Specifications
M2 Mac, python 3.8.12, using venv for virtual environment.

## Possible Solution
We can just change the versions of the problem packages in the sdk/python/requirements/py3.8-ci-requirements.txt file. I tried this locally, and everything installs fine, also, units tests and local integration tests still pass as expected (all version updates are patch or minor).

The problem packages are:

[fastavro==1.7.4](https://github.com/feast-dev/feast/blob/7bc7c47b4507310850474290131c03fb6d480834/sdk/python/requirements/py3.8-ci-requirements.txt#L213)
change from 1.7.4 to 1.8.1 due to Cython 3.0 release: https://github.com/fastavro/fastavro/blob/b63523ffbb8175a0b678f90e5e1840fb0f14d530/ChangeLog#L5

[pymssql==2.2.7](https://github.com/feast-dev/feast/blob/7bc7c47b4507310850474290131c03fb6d480834/sdk/python/requirements/py3.8-ci-requirements.txt#L716)
change from 2.2.7 to 2.2.8, due to due to Cython 3.0 release
see https://github.com/pymssql/pymssql/pull/827
which also references https://github.com/pymssql/pymssql/pull/803 so use instead pymssql==2.2.8

[pyyaml==6.0
](https://github.com/feast-dev/feast/blob/7bc7c47b4507310850474290131c03fb6d480834/sdk/python/requirements/py3.8-ci-requirements.txt#L782)has to be >6.0 due to Cython 3.0 release see: https://github.com/OctoPrint/OctoPrint/issues/4854 so use pyyaml==6.0.1

Likely we should also do this for the same file for py3.9 and py3.10 (both ci reqs and just reqs files), as they contain the same versions for those 3 packages.

crispin-ki:
See #3732

Fixes #3732

### DIFF
--- a/sdk/python/requirements/py3.10-ci-requirements.txt
+++ b/sdk/python/requirements/py3.10-ci-requirements.txt
@@ -206,7 +206,7 @@ executing==1.2.0
     # via stack-data
 fastapi==0.95.2
     # via feast (setup.py)
-fastavro==1.7.4
+fastavro==1.8.1
     # via
     #   feast (setup.py)
     #   pandavro
@@ -702,7 +702,7 @@ pyjwt[crypto]==2.7.0
     #   adal
     #   msal
     #   snowflake-connector-python
-pymssql==2.2.7
+pymssql==2.2.8
     # via feast (setup.py)
 pymysql==1.0.3
     # via feast (setup.py)
@@ -767,7 +767,7 @@ pytz==2023.3
     #   pandas
     #   snowflake-connector-python
     #   trino
-pyyaml==6.0
+pyyaml==6.0.1
     # via
     #   dask
     #   feast (setup.py)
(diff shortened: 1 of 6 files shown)


achals (approved):
.ok-to-test
/lgtm

crispin-ki:
Thanks @achals I think there was a minor typo in your last comment with .ok-to-test having a `.` instead of a `/`

PR #3647, opened 2023-06-08
docs: Add Dragonfly as an online-store option to Feast Readme and docs

yirenlu92:
**What this PR does / why we need it**:
Adds Dragonfly (drop in Redis replacement as an online-store option)

**Which issue(s) this PR fixes**:

Fixes #

PR #3744, opened 2023-08-24
docs: Add Dragonfly as online store alternative

Niennienzz:

Add documentation about Dragonfly as online store alternatives.
Dragonfly can use the already existing Redis connector.

Would like to replace https://github.com/feast-dev/feast/pull/3647.
No manual editing of README, followed `adding-support-for-a-new-online-store.md`.


adchia:
Already merged a different PR with this

PR #3758, opened 2023-09-08
feat: Add support for arrays in snowflake

### LINKED ISSUE #2280: Snowflake support does not include list features
## Expected Behavior 
Snowflake offline store works with list features

## Current Behavior
Right now Snowflake produces errors when using list features. This isn't caught by tests because of this skip with the aws provider: https://github.com/feast-dev/feast/blob/master/sdk/python/tests/integration/registration/test_universal_types.py#L39

JohnLemmonMedely:
Snowflake doesn't support type-checked arrays but support for arrays can still be added if developers can guarantee the type consistency themselves through the external data pipelines.

Fixes #2280


JohnLemmonMedely:
@sfc-gh-madkins Here's my draft PR (tests and documentation still pending). I've confirmed it unblocks my specific use-case to support a List[String] and materializing it to a Redis Online store but I'll need to setup the tests to confirm the other paths in the code work.

I'd appreciate a sanity check on some of the choices I had to make before I invest more time in this.

1. To cast the data correctly from pandas I just apply a `json.loads(x)` on the columns for any array type.
2. I only added support for Arrays to limit scope--but I think the same setup can work for objects.
3. In the function `snowflake_type_to_feast_value_type` I force array types to cast to Feast's String type. It looks like this function is used when inferring column types but that can't be done perfectly in this direction. So casting to a String seemed like the safest option. 
4. Do you know any place in the code I'm missing changes?

JohnLemmonMedely:
Closing in favor of https://github.com/feast-dev/feast/pull/3769

PR #3766, opened 2023-09-14
fix: UI project cannot handle fallback routes

### LINKED ISSUE #3765: Feast ui cannot parse url path
## Expected Behavior 

One of example cases:
When user navigate localhost:8888/p/order_count_project/feature-view/user_3_and_7_days_order_count should see related feature-view page

## Current Behavior

One of example cases:
When user navigate localhost:8888/p/order_count_project/feature-view/user_3_and_7_days_order_count see "Internal Server Error"

## Steps to reproduce

install feast 0.34.1
run feast ui
navigate homepage localhost:8888
navigate any page (entities or feature-view or data sources doesn't matter)
you will see the page you clicked at browser search bar like http://localhost:8888/p/order_count_project/data-source 
then refresh or copy url open in new tab
you will see internal server error

### Specifications

- Version: 0.34.1
- Platform: macos
- Subsystem: 

## Possible Solution

ui_server.py file updated recently. commit changes resource finder library and then it returns PosixPath. 
We should convert to str and add little "/" to "@app.api_route("/p/{path_name:path}", methods=["GET"])" function

snowron:
to serve ui project successfully 

Fixes #3765

### DIFF
--- a/sdk/python/feast/ui_server.py
+++ b/sdk/python/feast/ui_server.py
@@ -77,7 +77,7 @@ def get_app(
     # For all other paths (such as paths that would otherwise be handled by react router), pass to React
     @app.api_route("/p/{path_name:path}", methods=["GET"])
     def catch_all():
-        filename = ui_dir + "index.html"
+        filename = str(ui_dir) + "/index.html"
 
         with open(filename) as f:
             content = f.read()
--- a/sdk/python/feast/ui_server.py
+++ b/sdk/python/feast/ui_server.py
@@ -77,7 +77,7 @@ def get_app(
     # For all other paths (such as paths that would otherwise be handled by react router), pass to React
     @app.api_route("/p/{path_name:path}", methods=["GET"])
     def catch_all():
-        filename = str(ui_dir) + "/index.html"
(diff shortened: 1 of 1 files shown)


sdk/python/feast/ui_server.py
```diff
@@ -77,7 +77,7 @@ def read_registry():
     # For all other paths (such as paths that would otherwise be handled by react router), pass to React
     @app.api_route("/p/{path_name:path}", methods=["GET"])
     def catch_all():
-        filename = ui_dir + "index.html"
+        filename = str(ui_dir) + "/index.html"
```
adchia, line 80: never really noticed, but the existing use of `ui_dir` _seems_ wrong given it's on that try resource block. Any ideas on how to change this?

snowron: dear danny, I couldn't get you that point. existing ui_dir usage in whole file is working as expected without that line.

adchia (approved):

snowron:
@adchia hey i cant merge some problems about signing issue and now i cant merge again there is a job which is failed but it is not related with pr. can you help

Issue #3759, opened 2023-09-11
Write to postgres offline store


ndemashov:
Are there any planes to implement functionality to write data to postgres offline store?
**Describe the solution you'd like**
Implement `offline_write_batch` function for postgres as it has been done for spark for example

xtian-jampp:
+1

PR #3767, opened 2023-09-14
feat: Postgres offline_write_batch

ndemashov:
Implement offline_write_batch for postgresql offline store.

Fixes #
#3759

### DIFF
--- a/sdk/python/feast/infra/offline_stores/contrib/postgres_offline_store/postgres.py
+++ b/sdk/python/feast/infra/offline_stores/contrib/postgres_offline_store/postgres.py
@@ -1,4 +1,5 @@
 import contextlib
+import sys
 from dataclasses import asdict
 from datetime import datetime
 from typing import (
@@ -18,7 +19,11 @@ import numpy as np
 import pandas as pd
 import pyarrow
 import pyarrow as pa
-from adbc_driver_postgresql.dbapi import connect
+
+try:
+    from adbc_driver_postgresql.dbapi import connect
+except ImportError:
+    pass
 from jinja2 import BaseLoader, Environment
 from psycopg2 import sql
 from pydantic.typing import Literal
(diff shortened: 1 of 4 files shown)


ndemashov:
Please skip tests for python3.8. adbc-driver-postgresql package support only python>=3.9.

ndemashov:
@adchia can you review plese?

sdk/python/feast/infra/offline_stores/contrib/postgres_offline_store/postgres.py
```diff
@@ -16,7 +17,13 @@
 
 import numpy as np
 import pandas as pd
+import pyarrow
 import pyarrow as pa
+
+try:
+    from adbc_driver_postgresql.dbapi import connect
```
HaoXuAI, line 24: Maybe can move to offline_write_batch since it's only used there?

setup.py
```diff
@@ -114,6 +114,8 @@
 
 POSTGRES_REQUIRED = [
     "psycopg2-binary>=2.8.3,<3",
+    "adbc-driver-manager==0.6.0",
```
HaoXuAI, line 117: Not every one using Postgres as store needs this. I wonder if there is another way to handle the dependency

tokoko: It'll probably make sense to rewrite other methods to use adbc as well in the future, so i don't think it's worth it to split up extras now.

HaoXuAI: do all the offline data sources support the adbc now?

tokoko: No, just postges and sqlite in the [docs](https://arrow.apache.org/adbc/current/python/index.html). I thought about this some more and using adbc in some methods of the same class and `psycopg2` in others is probably not the best solution anyway. While the feature would be useful, since the author is no longer active here, I think we should close this one. @jeremyary

PR #3769, opened 2023-09-18
feat: Add support for arrays in snowflake

JohnLemmonMedely:
Snowflake doesn't support type-checked arrays but this adds support for arrays where developers have to guarantee the type consistency themselves through the external data pipelines.

This implementation adds support for these data types when stored as an array column in Snowflake:
- Array(String)
- Array(Bytes)
- Array(Int32)
- Array(Int64)
- Array(UnixTimestamp)
- Array(Float64)
- Array(Float32)
- Array(Bool)

This *does not work* with Feast's Feature Inference or the push/spark online store. The schema for these columns has to be defined in Feast because Snowflake itself isn't storing the data type of these variant columns for us to know how to translate them.

Which issue(s) this PR fixes:

Fixes https://github.com/feast-dev/feast/issues/2280

### DIFF
--- a/sdk/python/feast/infra/offline_stores/snowflake.py
+++ b/sdk/python/feast/infra/offline_stores/snowflake.py
@@ -1,4 +1,5 @@
 import contextlib
+import json
 import os
 import uuid
 import warnings
@@ -51,6 +52,17 @@ from feast.infra.utils.snowflake.snowflake_utils import (
 )
 from feast.repo_config import FeastConfigBaseModel, RepoConfig
 from feast.saved_dataset import SavedDatasetStorage
+from feast.types import (
+    Array,
+    Bool,
+    Bytes,
+    Float32,
+    Float64,
+    Int32,
+    Int64,
+    String,
+    UnixTimestamp,
+)
 from feast.usage import log_exceptions_and_usage
 
 try:
@@ -320,6 +332,7 @@ class SnowflakeOfflineStore(OfflineStore):
             on_demand_feature_views=OnDemandFeatureView.get_requested_odfvs(
                 feature_refs, project, registry
             ),
+            feature_views=feature_views,
             metadata=RetrievalMetadata(
                 features=feature_refs,
                 keys=list(entity_schema.keys() - {entity_df_event_timestamp_col}),
@@ -398,9 +411,12 @@ class SnowflakeRetrievalJob(RetrievalJob):
         config: RepoConfig,
         full_feature_names: bool,
         on_demand_feature_views: Optional[List[OnDemandFeatureView]] = None,
+        feature_views: Optional[List[FeatureView]] = None,
         metadata: Optional[RetrievalMetadata] = None,
     ):
 
+        if feature_views is None:
+            feature_views = []
         if not isinstance(query, str):
             self._query_generator = query
         else:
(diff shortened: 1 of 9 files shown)


sdk/python/feast/infra/offline_stores/snowflake.py
```diff
@@ -436,6 +453,20 @@ def _to_df_internal(self, timeout: Optional[int] = None) -> pd.DataFrame:
...
+                    Array(String),
+                    Array(Bytes),
+                    Array(Int32),
+                    Array(Int64),
+                    Array(UnixTimestamp),
+                    Array(Float64),
+                    Array(Float32),
+                    Array(Bool),
+                ]:
+                    df[feature.name] = [json.loads(x) for x in df[feature.name]]
```
JohnLemmonMedely, line 468: These arrays come in as a json string so this converts them to a list.

sdk/python/tests/integration/materialization/test_snowflake.py
```diff
@@ -84,15 +88,32 @@ def test_snowflake_materialization_consistency_internal():
...
+        ("float", Array(Float64)),
+        ("bool", Array(Bool)),
+        ("datetime", Array(UnixTimestamp)),
+    ],
+)
+@pytest.mark.parametrize("feature_is_empty_list", [False])
+@pytest.mark.parametrize("online_store", [SNOWFLAKE_ONLINE_CONFIG, "sqlite"])
 @pytest.mark.integration
-def test_snowflake_materialization_consistency_external():
+def test_snowflake_materialization_consistency_internal_with_lists(
```
JohnLemmonMedely, line 106: This test is similar to the one above but I had to verify the output myself instead of the shared function because I'm using Array types.

It's a lot of individual tests but if needed I can probably run a single test with multiple features to test each data type. It's just a lot more custom code for the setup.

sdk/python/tests/integration/materialization/test_snowflake.py
```diff
@@ -105,23 +126,62 @@ def test_snowflake_materialization_consistency_external():
...
 
+    schema = [
+        Field(name="driver_id", dtype=Int32),
+        Field(name="value", dtype=feast_dtype),
+    ]
     driver_stats_fv = FeatureView(
         name="driver_hourly_stats",
         entities=[driver],
         ttl=timedelta(weeks=52),
+        schema=schema,
```
JohnLemmonMedely, line 137: These arrays only work with specified schemas (we can't infer what the type is from a plain `Array` in Snowflake).

JohnLemmonMedely (commented):
I'd still need to update the documentation but I'd like to make sure that at least this approach is fine. And again I'm choosing to not support inferred schemas or the push online store because they both require being able to infer the schema from the Snowflake schema.

JohnLemmonMedely:
@sfc-gh-madkins Please let me if you're still the right person to review and if you can enable the full tests in Github actions.

sfc-gh-madkins:
@JohnLemmonMedely Hey John -- thanks for the work on this. I approved this to run, however I am not the right person to review anymore as I have switched focused to the Snowflake Native feature store we announced.

JohnLemmonMedely:
Thanks @sfc-gh-madkins. Can you add @feast-dev/maintainers as a reviewer? I know things are in flux with Tecton pulling out so I'm not sure who to loop in specifically.

Also I don't think the integration tests are running and I don't see the labels on this PR. Is there something else to be done?

sfc-gh-madkins:
good question on reviewer side. I approved so they should now be running

tmihalac:
lgtm

etirelli (approved):

PR #3789, opened 2023-10-09
fix: Redundant feature materialization and premature incremental materialization timestamp updates

### LINKED ISSUE #3786: Bytewax materializer materializes entire dataset on every pod
## Expected Behavior 

Each pod should materialize a separate batch of data

## Current Behavior

Every pod redundantly materializes the entire dataset, causing a massive overspend on processing time and database writes

## Steps to reproduce

Observed with snowflake offline store and dynamo db online store.  Materialize a feature with enough records for the snowflake s3 integration to output multiple files.  View the logs of each pod and add up the total records materialized.  Observe that every pod has materialized the entire dataset

### Specifications

- Version: 0.31
- Platform: fedora linux
- Subsystem: bytewax batch_engine

## Possible Solution

Use JOB_COMPLETION_INDEX in each pods dataflow to select a single file to process

### LINKED ISSUE #3787: Bytewax materializer sets incremental materialization date before materialization completes
## Expected Behavior 

There should be an option to match the LocalMaterializer behavior, which waits for materialization to finish before updating the incremental materialization timestamp for a feature view

## Current Behavior

The incremental materialization timestamp is updated as soon as the materialization job is successfully created, even though that job may not successfully complete.  This means that some updates will never be materialized to the online store.

## Steps to reproduce

Set the batch_engine image to an image that immediately exits with exit code 1.  Run an incremental materialization with end date set to current time.  Observe that the pods and job fail.  Run an incremental materialization again.  Observe that the start date is now the end date of the last run.

### Specifications

- Version: 0.31
- Platform: fedora linux
- Subsystem: bytewax batch_engine

## Possible Solution

Optionally allow for a synchronous materialization in bytewax, which will wait for a feature view materialization to finish before returning success and moving on to the next feature view and throwing an exception if the job fails.  This will match the behavior of the LocalMaterializer

### LINKED ISSUE #3788: Bytewax materialization can run infinitely
## Expected Behavior 

Bytewax materialization should run all pods once successfully and then set job status as success

## Current Behavior

In the event that a node crashes, successful pod records can be lost and the job will rerun all of those lost pods.  If these node crashes occur often enough, this can result in a job continuously rerunning successful pods and never completing.

## Steps to reproduce

Run a materialization job against a multi-node kubernetes cluster.  Terminate one of the nodes, observe that pods are lost and rerun

### Specifications

- Version: 0.31
- Platform: fedora linux
- Subsystem: bytewax batch_engine

## Possible Solution

For safety, the job should have a configurable activeDeadlineSeconds.  The larger job should also be able to be split into smaller batches to mitigate the effect a node crash can have

james-crabtree-sp:

This PR fixes 2 issues encountered while attempting to run very large materializations (Several billion rows as part of an initial run of online materialization).

1.  Every pod created materializes every file in the dataset.  This is redundant and costs a lot of time and money.  To fix this, I updated the dataflow code to select only the single file corresponding to the running pod's JOB_COMPLETION_INDEX as input.
2. The last materialized date used for incremental materialization was being updated before materialization had even run.  This meant that if materialization had failed for some reason, many updates could be ignored, because the next run would continue from beyond the end date of the failed run.  I solved this by implementing a synchronous mode for bytewax materialization that will wait for materialization of a feature to complete successfully before returning and updating the last materialized date
3. Bytewax jobs can run infinitely if node crashes occur often enough that the job has to continuously reprocess the same pods after they are lost.  I mitigated this by adding a configurable activeDeadlineSeconds and by adding an option to synchronous materialization to allow these jobs to be split into smaller batches so fewer pods have to be rerun in the event of a node failure

Fixes #3786 
Fixes #3787 
Fixes #3788

### DIFF
--- a/sdk/python/feast/infra/materialization/contrib/bytewax/bytewax_materialization_engine.py
+++ b/sdk/python/feast/infra/materialization/contrib/bytewax/bytewax_materialization_engine.py
@@ -293,7 +293,7 @@ class BytewaxMaterializationEngine(BatchMaterializationEngine):
             },
             {
                 "name": "BYTEWAX_REPLICAS",
-                "value": f"{pods}",
+                "value": "1",
             },
             {
                 "name": "BYTEWAX_KEEP_CONTAINER_ALIVE",
@@ -349,7 +349,7 @@ class BytewaxMaterializationEngine(BatchMaterializationEngine):
                                 "env": [
                                     {
                                         "name": "BYTEWAX_REPLICAS",
-                                        "value": f"{pods}",
+                                        "value": "1",
                                     }
                                 ],
                                 "image": "busybox",
--- a/sdk/python/feast/infra/materialization/contrib/bytewax/bytewax_materialization_engine.py
+++ b/sdk/python/feast/infra/materialization/contrib/bytewax/bytewax_materialization_engine.py
@@ -216,7 +216,10 @@ class BytewaxMaterializationEngine(BatchMaterializationEngine):
                     total_pods,
                 )
                 offset += batch_size
-                if offset >= total_pods:
+                if (
+                    offset >= total_pods
+                    or job.status() == MaterializationJobStatus.ERROR
+                ):
                     break
         else:
             job_id = str(uuid.uuid4())
--- a/sdk/python/feast/infra/materialization/contrib/bytewax/bytewax_materialization_engine.py
+++ b/sdk/python/feast/infra/materialization/contrib/bytewax/bytewax_materialization_engine.py
@@ -248,9 +248,6 @@ class BytewaxMaterializationEngine(BatchMaterializationEngine):
                 f"(of {total_pods}) complete with status {job.status()}"
             )
         except BaseException as e:
-            if self.batch_engine_config.print_pod_logs_on_failure:
-                self._print_pod_logs(job.job_id(), feature_view, batch_start)
-
             logger.info(f"Deleting job {job.job_id()}")
             try:
                 self.batch_v1.delete_namespaced_job(job.job_id(), self.namespace)
@@ -267,6 +264,12 @@ class BytewaxMaterializationEngine(BatchMaterializationEngine):
                 logger.warning(
                     f"Could not delete configmap due to API Error: {ae.body}"
                 )
+
+            if (
+                job.status() == MaterializationJobStatus.ERROR
+                and self.batch_engine_config.print_pod_logs_on_failure
+            ):
+                self._print_pod_logs(job.job_id(), feature_view, batch_start)
 
         return job
 
--- a/sdk/python/feast/infra/materialization/contrib/bytewax/bytewax_materialization_engine.py
+++ b/sdk/python/feast/infra/materialization/contrib/bytewax/bytewax_materialization_engine.py
@@ -71,7 +71,7 @@ class BytewaxMaterializationEngineConfig(FeastConfigBaseModel):
     """ (optional) additional labels to append to kubernetes objects """
 
     max_parallelism: int = 10
-    """ (optional) Maximum number of pods allowed to run in parallel per job"""
+    """ (optional) Maximum number of pods allowed to run in parallel"""
 
     synchronous: bool = False
     """ (optional) If true, wait for materialization for one feature to complete before moving to the next """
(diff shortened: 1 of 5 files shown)


sudohainguyen:
great, I have been working on the similar PR, but you did it first and way better 💯

sdk/python/feast/infra/materialization/contrib/bytewax/bytewax_materialization_engine.py
```diff
@@ -170,8 +191,87 @@ def _materialize_one(
...
+        try:
+            while job.status() in (
+                MaterializationJobStatus.WAITING,
+                MaterializationJobStatus.RUNNING,
+            ):
+                logger.info(
+                    f"{feature_view.name} materialization for pods {batch_start}-{batch_end} "
+                    f"(of {total_pods}) running..."
+                )
+                sleep(30)
```
sudohainguyen, line 234: do you think we should use tenacity to handle while loop instead of doing sleep?

sudohainguyen:
the logic for job_batch_size looks quite confusing and doesn't apply if `synchronous` is false right?
this config param should be consumed by job's parallelism value, to ensure maximum pods to be active at the same time

sdk/python/feast/infra/materialization/contrib/bytewax/bytewax_materialization_engine.py
james-crabtree-sp: In this case, I dont think tenacity would be the right framework, because we don't want to ignore exceptions.

sdk/python/feast/infra/materialization/contrib/bytewax/bytewax_materialization_engine.py
```diff
@@ -65,7 +71,22 @@ class BytewaxMaterializationEngineConfig(FeastConfigBaseModel):
...
+    synchronous: bool = False
+    """ (optional) If true, wait for materialization for one feature to complete before moving to the next """
+
+    retry_limit: int = 2
+    """ (optional) Maximum number of times to retry a materialization worker pod"""
+
+    active_deadline_seconds: int = 86400
+    """ (optional) Maximum amount of time a materialization job is allowed to run"""
+
+    job_batch_size: int = 100
```
sudohainguyen, line 85: can we reuse `max_parallelism`? seems similar purpose

james-crabtree-sp:
> @james-crabtree-sp Thanks again for putting this PR together. I think in order to merge, you'll need to sign your commits: https://docs.feast.dev/project/development-guide#signing-off-commits

I think they are all signed off already.  I see that the DCO task that checks this completed successfully https://github.com/feast-dev/feast/pull/3789/checks?check_run_id=17541354092

sudohainguyen:
hey @james-crabtree-sp  this is a great PR, you need any hands to make it merged? I see there's a lot of commits in the past unverified

if you don't mind I can create another PR to mimic this one

sdk/python/feast/infra/materialization/contrib/bytewax/bytewax_materialization_engine.py
sudohainguyen: or maybe we should try reusing the function `wait_retry_backoff` in `sdk/python/feast/wait.py`.
It would be a bit cleaner

james-crabtree-sp:
@whoahbot @sudohainguyen Ahh, I understand now.  Commits are now verified

sudohainguyen:
> @whoahbot @sudohainguyen Ahh, I understand now. Commits are now verified

exactly 🤣 looks good now but you need to run code format though

sudohainguyen:
@james-crabtree-sp I've been adopting your changes on my side

so far do we need to wait the job to be done, like forever without timeout?

james-crabtree-sp:
> @james-crabtree-sp I've been adopting your changes on my side
> 
> so far do we need to wait the job to be done, like forever without timeout?

Well, this PR adds the configurable active_deadline_seconds setting, which can be used to impose a timeout on the kubernetes jobs.  Does that answer your question?

sudohainguyen:
> > @james-crabtree-sp I've been adopting your changes on my side
> > so far do we need to wait the job to be done, like forever without timeout?
> 
> Well, this PR adds the configurable active_deadline_seconds setting, which can be used to impose a timeout on the kubernetes jobs. Does that answer your question?

cool clear

PR #3790, opened 2023-10-11
fix: Resolve hbase hotspot issue when materializing

sudohainguyen:
**What this PR does / why we need it**: This PR resolves hotspot issue when materializing records to HBase online store.
Previously, row keys are generated with simple serialization, which led to unbalanced row keys distribution across HBase regions. So I resolve this by applying the similar mechanism being used in BigTable online store because BigTable and HBase work the same way under the hood.

Fixes #


sudohainguyen:
hi @achals , who is currently in charge of HBase module to review my PR? 🤔

achals (approved):
/ok-to-test
/lgtm

sudohainguyen:
> /ok-to-test /lgtm

seems not working 😢

PR #3797, opened 2023-10-16
fix: Make generated temp table name escaped

BearDimonR:
We need to encapsulate the table path as the "tmp" table path can contain symbols that BigQuery can't process without encapsulation.

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -515,7 +515,7 @@ class BigQueryRetrievalJob(RetrievalJob):
                 temp_dest_table = f"{tmp_dest['projectId']}.{tmp_dest['datasetId']}.{tmp_dest['tableId']}"
 
                 # persist temp table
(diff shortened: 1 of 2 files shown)


BearDimonR:
I have this issue with every script due to a specific project name (contains "-": _production-abcd123_). The solution is to add encapsulation to the reference, as I proposed. Tested by saving the result to GCS, and it worked fine.
image

sudohainguyen:
> I have this issue with every script due to a specific project name (contains "-": _production-abcd123_). The solution is to add encapsulation to the reference, as I proposed. Tested by saving the result to GCS, and it worked fine. image

hmm that makes sense, I agree there should be encapsulation to make sure things go well

sudohainguyen:
hey @BearDimonR, DCO step failed, please sign-off all the commits
https://docs.feast.dev/project/development-guide#signing-off-commits

BearDimonR:
@sudohainguyen it is done

sudohainguyen (approved):
/lgtm

PR #3819, opened 2023-10-25
fix: Update SparkSource to have proper comparable that inspects SparkOptions

thechopkins:
This updates the `SparkSource` data type to be properly comparable. The previous form defaults to using the `__eq__` implementation from the `DataSource` class. This misses updates to additional properties on the `SparkSource` class (i.e. anything that feeds into the `SparkOptions` class). 

This PR resolves this and adds tests to confirm the updated functionality. This means that `FeatureStore.plan()` and `FeatureStore.apply()` will properly pick up changes to `SparkSources` correctly.

### DIFF
--- a/sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark_source.py
+++ b/sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark_source.py
@@ -185,6 +185,18 @@ class SparkSource(DataSource):
 
         return f"`{tmp_table_name}`"
 
+    # Note: Python requires redefining hash in child classes that override __eq__
+    def __hash__(self):
+        return super().__hash__()
+
+    def __eq__(self, other):
+        if not isinstance(other, SparkSource):
+            raise TypeError("Comparisons should only involve SparkSource class objects.")
+        return (
+            super().__eq__(other)
+            and self.spark_options == other.spark_options
+        )
+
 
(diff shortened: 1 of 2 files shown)


sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark_source.py
```diff
@@ -185,6 +185,18 @@ def get_table_query_string(self) -> str:
 
         return f"`{tmp_table_name}`"
 
+    # Note: Python requires redefining hash in child classes that override __eq__
+    def __hash__(self):
+
+        return super().__hash__()
```
rebstastic, line 191: Should this also include `self.spark_options`?

thechopkins: It doesn't necessarily need to as the two pieces of logic are used for different things. Take a look at the [redshift_source.py example](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/offline_stores/redshift_source.py#L116-L131) for how it is implemented elsewhere.

franciscojavierarceo:
looks like linter is failing here. @thechopkins can you update?

jeremyary:
closing out as this was addressed via the mentioned PR [4028](https://github.com/feast-dev/feast/pull/4028). Thanks!

PR #3821, opened 2023-10-26
fix: Bytewax engine create configmap from object

sudohainguyen:
**What this PR does / why we need it**: Instead of reading the content of config yaml file to create configmap for bytewax job, we parse directly from repo_config object

Fixes #


achals (approved):

whoahbot (approved):

sudohainguyen:
@whoahbot  looks like the failed tests don't come from this change :<

sudohainguyen:
@whoahbot hey this needs some actions as well

PR #3826, opened 2023-11-04
feat: Optimize bytewax pod resource with zero-copy

### LINKED ISSUE #3825: Bytewax pods consume lots of resource
## Expected Behavior 
I conducted benchmark on a feature table with 50m rows x 10 cols, and expect I can efficiently materialize records into online store. In Bytewax mechanism, latest records are extracted to staging location as parquet files, in my case each file contains ~140k rows. 
In the efficient way, bytewax pods should process the file with as less memory footprint as possible

## Current Behavior
Currently every bytewax pods pull the entire parquet file into memory before writing to online store, which causes huge memory footprint, ~3GB of memory.

## Steps to reproduce
Conduct materialization with bytewax engine

### Specifications

- Version: master

## Possible Solution
Apply zero-copy mechanism from pyarrow to stream the parquet files and process on-the-fly before pushing to online store

sudohainguyen:
**What this PR does / why we need it**: To optimize bytewax pods resource usage, for the case in #3825, each pods only consume 300MB of memory.

Fixes #3825

### DIFF
--- a/sdk/python/feast/infra/materialization/contrib/bytewax/bytewax_materialization_engine.py
+++ b/sdk/python/feast/infra/materialization/contrib/bytewax/bytewax_materialization_engine.py
@@ -82,12 +82,6 @@ class BytewaxMaterializationEngineConfig(FeastConfigBaseModel):
     mini_batch_size: int = 1000
     """ (optional) Number of rows to process per write operation (default 1000)"""
 
-    bytewax_replicas: int = 5
-    """ (optional) Number of process to spawn in each pods to handle a file in parallel"""
-
-    bytewax_worker_per_process: int = 1
-    """ (optional) Number of threads as worker per bytewax process"""
-
     active_deadline_seconds: int = 86400
     """ (optional) Maximum amount of time a materialization job is allowed to run"""
 
@@ -356,7 +350,7 @@ class BytewaxMaterializationEngine(BatchMaterializationEngine):
             {"name": "BYTEWAX_WORKDIR", "value": "/bytewax"},
             {
                 "name": "BYTEWAX_WORKERS_PER_PROCESS",
-                "value": f"{self.batch_engine_config.bytewax_worker_per_process}",
+                "value": "1",
             },
             {
                 "name": "BYTEWAX_POD_NAME",
@@ -369,7 +363,7 @@ class BytewaxMaterializationEngine(BatchMaterializationEngine):
             },
             {
                 "name": "BYTEWAX_REPLICAS",
-                "value": f"{self.batch_engine_config.bytewax_replicas}",
+                "value": "1",
             },
             {
                 "name": "BYTEWAX_KEEP_CONTAINER_ALIVE",
--- a/sdk/python/feast/infra/materialization/contrib/bytewax/bytewax_materialization_engine.py
+++ b/sdk/python/feast/infra/materialization/contrib/bytewax/bytewax_materialization_engine.py
@@ -82,6 +82,12 @@ class BytewaxMaterializationEngineConfig(FeastConfigBaseModel):
     mini_batch_size: int = 1000
     """ (optional) Number of rows to process per write operation (default 1000)"""
 
+    bytewax_replicas: int = 5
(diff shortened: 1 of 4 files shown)


sdk/python/feast/infra/materialization/contrib/bytewax/bytewax_materialization_engine.py
```diff
@@ -82,6 +82,12 @@ class BytewaxMaterializationEngineConfig(FeastConfigBaseModel):
     mini_batch_size: int = 1000
     """ (optional) Number of rows to process per write operation (default 1000)"""
 
+    bytewax_replicas: int = 5
+    """ (optional) Number of process to spawn in each pods to handle a file in parallel"""
+
+    bytewax_worker_per_process: int = 1
```
whoahbot, line 88: I don't believe that this setting, and the setting of the number of replicas will currently be respected. 

I'd be happy to help with getting this fixed, but perhaps that should be the subject of a separate PR.

sudohainguyen: ah yeah you're right, looking forward to working on a separate PR

sudohainguyen: > I don't believe that this setting, and the setting of the number of replicas will currently be respected.
> 
> I'd be happy to help with getting this fixed, but perhaps that should be the subject of a separate PR.

could you point out exactly where the setting is not respected? seems like the env of bytewax pods consume it correctly

whoahbot: This is where we call the bytewax entrypoint for the dataflow:

https://github.com/feast-dev/feast/blob/9583ed6b4ae8d3b97934bf0c80ecb236ed1e2895/sdk/python/feast/infra/materialization/contrib/bytewax/bytewax_materialization_dataflow.py#L90

which ends up calling:

https://github.com/bytewax/bytewax/blob/v0.15.1/src/execution/mod.rs#L709-L717

Since we aren't using any of the environment variables at this point, we should only ever see one Python process with one worker per pod. 

Happy to chat further about enabling multiple workers. Feel free to reach out on slack if you would prefer to do it there.

sudohainguyen: Great, would love to chat more, but by the way we're using bytewax 0.17.1 as of now

sudohainguyen: by the way, in case the REPLICAS setting doesn't work, we still have at least 1 process and 1 worker per pods.

and the implementation of this PR still reduce memory footprint compares to master's one. should we process this first? @achals

achals: @sudohainguyen  I'd prefer to update the pr so only the changes we expect to actually make a difference are merged (specifically the arrow changes) - we can bump the dependency on bytewax and make any necessary changes for replicas in a separate pr

sudohainguyen: agree, then I will narrow down the PR, to make bytewax settings unexposed from batch engine config

sudohainguyen: done @achals

achals (approved):

achals:
@sudohainguyen are the current tests sufficient to ensure this test is sitting as expected? Can you test some of this manually?

sudohainguyen:
Yep I did, and already applied to my production, that's how I can measure memory usage reduced.

andrewshrout:
Bumping this out of interest. Hoping it's merged soon.

whoahbot (approved):

mjurkus:
An amazing improvement! Used this in our workflows. Peak usage was at 512GB before the update. 

image

PR #3840, opened 2023-11-17
feat: Added option for global tables for Dynamodb online 

danielhstahl:

The current dynamodb online store does not support [global tables](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/V2globaltables_HowItWorks.html).  This PR adds an optional attribute to the dynamodb config.  If a `global_table_region` is specified, `feast apply` will create a replica dynamodb table in that region.


tmihalac:
Please fix linter and DCO test failing

tmihalac (changes requested):

PR #3851, opened 2023-12-01
fix: Verify the existence of Registry tables in snowflake before calling CREATE sql command. Allow read-only user to call feast apply.

### LINKED ISSUE #3844: allow read-only feature while using snowflake as (sql) registry host.
Based on the community user's  (Zvonimir Cikojevic) feedback, we want to create a read-only feature for snowflake user who only can do READ to the registry on snowflake. 

Hi guys! I hope I’m in the right channel.
I have Feast deployed on Snowflake and I’d like to limit the write access to the feature store.
The idea was to have a Snowflake role that only has Read privileges to the feature store schema (where the feast registry and our batch sources are deployed).
However, when assuming this role and running
store = feast.FeatureStore()
I get unauthorized error since the first SQL query Feast runs is CREATE TABLE IF NOT EXISTS DB.SCHEMA.DATA_SOURCES…
Does this mean that anyone who wishes to query the feature store has to have RW access?
Can you help me understand the reasoning behind this or did I miss something?
Thanks!

Yes, Feast tries to create registry Snowflake tables if they don't exist. I've created the registry when I first ran feast apply. This block of code always executed https://github.com/feast-dev/feast/blob/9df2224283e04760116b61bed3c8bfa7f17cbf7e/sdk/python/feast/infra/registry/snowflake.py#L128
when I run store = FeatureStore() .
In my case, this is unnecessary since I've (as a feature store admin) already created the registry.
I'd like to enable users to just have the ability to query the feature store - and not give them the privileges to write to the feature store tables (if they're not developers/admins).

(TBD)

(TBD)

**Additional context**
(TBD)

shuchu:
For snowflake users who only have "USAGE" and "SELECT" privileges, we want to allow them able to run the "feast apply" command and do READ-only access existing Registry tables in snowflake cloud.

Please be aware that the Feast Registry is designed to allow every user can modify the records of registry tables under their own "project" name. This PR didn't change the original design. It only changes the previous implementation logic of snowflake.py which always calls "CREATE" sql command while initializing the SnowflakeRegistry(), which will be called while run statement:

store = FeatureStore()

This PR is tested with snowflake cloud privilege for a Role with setting as below:

for feast Registry database : SELECT - FUTURE TABLE, SELECT - FUTURE VIEW, USAGE
for feast Registry 11 tables: SELECT
for feast schema (default name PUBLIC): SELECT - FUTURE TABLE, SELECT - FUTURE VIEW, USAGE
for warehouse: USAGE

The test will expect the above specific role can call "feast apply" and does not see errors like:
snowflake.connector.errors.ProgrammingError: 003001 (42501): SQL access control error: Insufficient privileges to operate on *****
during the init() function call of SnowflakeRegistry().

Which issue(s) this PR fixes:
Fixes #3844

### DIFF
--- a/sdk/python/feast/infra/registry/snowflake.py
+++ b/sdk/python/feast/infra/registry/snowflake.py
@@ -124,15 +124,19 @@ class SnowflakeRegistry(BaseRegistry):
             f'"{self.registry_config.database}"."{self.registry_config.schema_}"'
         )
 
-        with GetSnowflakeConnection(self.registry_config) as conn:
-            sql_function_file = f"{os.path.dirname(feast.__file__)}/infra/utils/snowflake/registry/snowflake_table_creation.sql"
-            with open(sql_function_file, "r") as file:
-                sqlFile = file.read()
+        if not self._verify_registry_database():
+            # Verify the existing resitry database schema from snowflake. If any table names and column types is wrong, run table recreation SQL.
+            with GetSnowflakeConnection(self.registry_config) as conn:
+                sql_function_file = f"{os.path.dirname(feast.__file__)}/infra/utils/snowflake/registry/snowflake_table_creation.sql"
+                with open(sql_function_file, "r") as file:
+                    sqlFile = file.read()
 
-                sqlCommands = sqlFile.split(";")
-                for command in sqlCommands:
-                    query = command.replace("REGISTRY_PATH", f"{self.registry_path}")
-                    execute_snowflake_statement(conn, query)
+                    sqlCommands = sqlFile.split(";")
+                    for command in sqlCommands:
+                        query = command.replace(
+                            "REGISTRY_PATH", f"{self.registry_path}"
+                        )
+                        execute_snowflake_statement(conn, query)
 
         self.cached_registry_proto = self.proto()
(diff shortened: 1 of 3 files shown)


tmihalac:
Please add tests to make sure the fix works as intended

tmihalac:
Can you please add more details on how to test the fix ?

shuchu:
I did few manual tests with my free Snowflake account while I was creating this PR.  To test it, we need to configure the Snowflake as I mentioned in the PR and check the result of "feast apply" on the tester's terminal. 

It will be great if someone has a Snowflake account and test it.

etirelli (approved):

shuchu:
This PR can not fix the feature that allows READ-only users to access the DB. There is an optimization in the get_historical_features() that will create a temporary table in the DB for the point-of-time correct Join. (https://www.hopsworks.ai/dictionary/point-in-time-correct-joins#:~:text=A%20point%2Din%2Dtime%20correct,a%20specific%20point%20in%20time.)  For example, in line 815-822 of bigquery.py (https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/offline_stores/bigquery.py), there is a short description about this. The temporary table is deleted as in Line 298 after the usage.

PR #3857, opened 2023-12-05
feat: Added delete_table to redis online store

### LINKED ISSUE #3856: Add delete_table to redis online store
Even if the feature view is no longer used, the data is not deleted until there are no feature views left in the join key.

Memory in redis is a very expensive resource.

When deleting feature_view, a function is added so that the corresponding key can be found and deleted in the redis hash.

phil-park:

Even if the feature view is no longer used, the data is not deleted until there are no feature views left in the join key.

Memory in redis is a very expensive resource.

When deleting feature_view, a function is added so that the corresponding key can be found and deleted in the redis hash.

Fixes #3856

### DIFF
--- a/sdk/python/feast/infra/online_stores/redis.py
+++ b/sdk/python/feast/infra/online_stores/redis.py
@@ -133,7 +133,7 @@ class RedisOnlineStore(OnlineStore):
                 if len(_tables) == 1:
                     pipe.delete(_k)
                 else:
-                    pipe.hdel(_k, redis_hash_keys)
+                    pipe.hdel(_k, *redis_hash_keys)
                 deleted_count += 1
             pipe.execute()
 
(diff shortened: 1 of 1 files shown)


redhatHameed:
@etirelli The code looks good to me, but I would like another pair of eyes to take a look at it.

sudohainguyen:
great idea but is it not a good practice to conduct scanning on production redis database?

etirelli:
@phil-park thank you for your PR. Unfortunately we have an integration test failure, can you please take a look?

@sudohainguyen while scanning Redis tables is [not a best practice](https://redis.com/blog/7-redis-worst-practices/), the existing code already [scans the db](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/online_stores/redis.py#L100) . Since a change to this will probably require considerable changes, I would like to do so under a different ticket/PR.

phil-park:
@etirelli There were some issues with my commit and I resolved those issues.Please check. thank you!

jeremyary (approved):
Making maintenance rounds, approving on basis of previous checks/lgtm's, should be good to go here. Thanks for the contrib!

PR #3874, opened 2023-12-29
feat: Add python bytes to array type conversion support proto

### LINKED ISSUE #3871: support for redshift super -> feast array types.
Redshift's `super` column type allows for JSON. Currently the best approach with feast appears to be using either `ValueType.Bytes` with a `super` column or `ValueType.String` with a `varchar` and serialize the data ahead of time. Neither are optimal since the online store will then receive either bytes or a string instead of an array.

Would like to be able to define a column as `Array(base_type=[appropriate primitive)` for `Field` backed by a `super` column. When we `materialize` the appropriate type is loaded into and eventually returned from the online stoer at request time.

Working with json strings or bytes and having the client be in charge of deserializing.

Thanks

### LINKED ISSUE #3884: can't /push (inferred) integers to a float feature
## Expected Behavior 
Attempts to `/push` the following data 
```
{
  "df": {
    "feature_one": [123, 456]
  }
}
```
to the online store where the feature view field is `Field("feature_one", dtype=types.Float64)` should succeed.

## Current Behavior
This fails with the following error.

image

## Possible Solution
The `PYTHON_SCALAR_VALUE_TYPE_TO_PROTO_VALUE` map defined [here](https://github.com/feast-dev/feast/blob/e436f776ccf291809c802c469073adf7f1540d69/sdk/python/feast/type_map.py#L299-L300) states that value types of `FLOAT` have to be inferred as either `np.float` or `float`. Add `int` and `np.int_` to the list and it works as expected.

robhowley:

Support for complex types in offline stores occasionally comes in the form of `bytes`, e.g. redshift's `super`. It may also just be desirable to store data as bytes offline. That shouldn't, however, preclude the use of strongly typed online retrieval. So, allow for `bytes -> ValueType.[something]_List` conversion while constructing the `proto` in `_python_value_to_proto_value`

Also fixes a small bug identified in #3884 

Fixes #3871 
Fixes #3884


redhatHameed (approved):
/lgtm

sdk/python/feast/type_map.py
```diff
@@ -297,7 +298,7 @@ def _type_err(item, dtype):
         None,
     ),
     ValueType.FLOAT: ("float_val", lambda x: float(x), None),
-    ValueType.DOUBLE: ("double_val", lambda x: x, {float, np.float64}),
+    ValueType.DOUBLE: ("double_val", lambda x: x, {float, np.float64, int, np.int_}),
```
robhowley, line 301: see issue #3884

etirelli (approved):

robhowley:
hey @etirelli thanks for merging this! 

curious what the intended release cycle is going to be under the (mostly) new maintainers group. v35 came out 3 days ago. seems like it used to be about monthly, so now that feast is back in action should we expect roughly the same?

PR #3898, opened 2024-01-22
fix: Trino as an OfflineStore Access Denied when BasicAuthenticaion

Woo-Dong:

When Feast with OfflineStore as Trino with basic Authentication, password is converted with `SecretStr` in Pydantic type.
Then error occurred with 401('Access Denied: Invalid credentials').
It can be solved when I replace `SecretStr` into `StrictStr` where `.feast.infra.offline_stores.contrib.trino_offline_store.trino.py` line 43.

* Before
image

*  After 
image

And `feature_store.yaml` format is like below
```yaml
project: deepfm_repo
# By default, the registry is a file (but can be turned into a more scalable SQL-backed registry)
registry: data/registry.db
# The provider primarily specifies default offline / online stores & storing the registry in a given cloud
provider: local

offline_store:
  type: trino
  host: localhost
  port: 443
  catalog: lakehouse
  connector:
    type: hive
    file_format: parquet
  user: aidt-devops
  source: feast-trino-offline-store
  http-scheme: https
  ssl-verify: true
  # enables authentication in Trino connections, pick the one you need
  # if you don't need authentication, you can safely remove the whole auth block
  auth:
    # Basic Auth
    type: basic
    config:
      username: aidt-devops
      password: devops123@
entity_key_serialization_version: 2
```

### DIFF
--- a/sdk/python/feast/infra/offline_stores/contrib/trino_offline_store/trino.py
+++ b/sdk/python/feast/infra/offline_stores/contrib/trino_offline_store/trino.py
@@ -40,7 +40,7 @@ from feast.usage import log_exceptions_and_usage
 
 class BasicAuthModel(FeastConfigBaseModel):
     username: StrictStr
-    password: StrictStr  # original: SecretStr
+    password: StrictStr
 
 
 class KerberosAuthModel(FeastConfigBaseModel):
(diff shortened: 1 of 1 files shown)


HaoXuAI (approved):
Nice. LGTM!

PR #3905, opened 2024-01-25
fix: Create index only if not exists during MySQL online store update

gee-senbong:
The primary rationale behind these changes is to prevent duplicate key errors that occur when attempting to create an index that already exists. By checking for the existence of an index before creating it, the PR ensures calling `feast apply` more than once won't cause an error.

**Which issue(s) this PR fixes**
This PR addresses a potential issue where the MySQL online store could encounter duplicated key errors when the Feast SDK tries to create an index that already exists. 

Fixes https://github.com/feast-dev/feast/issues/3906

### DIFF
--- a/sdk/python/feast/infra/online_stores/contrib/mysql_online_store/mysql.py
+++ b/sdk/python/feast/infra/online_stores/contrib/mysql_online_store/mysql.py
@@ -178,8 +178,11 @@ class MySQLOnlineStore(OnlineStore):
 
         # We don't create any special state for the entities in this implementation.
         for table in tables_to_keep:
+
+            table_name = _table_id(project, table)
+            index_name = f"{table_name}_ek"
             cur.execute(
-                f"""CREATE TABLE IF NOT EXISTS {_table_id(project, table)} (entity_key VARCHAR(512),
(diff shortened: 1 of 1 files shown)


sdk/python/feast/infra/online_stores/contrib/mysql_online_store/mysql.py
```diff
@@ -178,18 +178,28 @@ def update(
...
+            index_exists = cur.execute(
+                f"""
+                SELECT 1 FROM information_schema.statistics
+                WHERE table_schema = DATABASE() AND table_name = '{table_name}' AND index_name = '{index_name}'
+                """
             )
+            if not index_exists:
+                cur.execute(
+                    f"ALTER TABLE {table_name} ADD INDEX {index_name} (entity_key);"
+                )
```
gee-senbong, line 202: As MySQL does not support `ADD INDEX IF NOT EXISTS`, a separate query is created to handle this.

redhatHameed:
LGTM
@etirelli could take a look for this PR thanks

etirelli (approved):

Issue #3893, opened 2024-01-18
Bytewax materialization engine fails when loading feature_store.yaml


gterziysky:
## Expected Behavior 

Loading the `feature_store.yaml` file from within a Bytewax pod should work.

## Current Behavior

`yaml.safe_load()` raises an error while trying to reconstruct the object below:

- pathlib.PosixPath

The error occurs while running materialization using Bytewax at the point where the `feature_store.yaml` is loaded. The code where this happens is in [sdk/python/feast/infra/materialization/contrib/bytewax/dataflow.py](https://github.com/feast-dev/feast/blob/ac6529c380bc623a15054350c9936b69617b4bee/sdk/python/feast/infra/materialization/contrib/bytewax/dataflow.py#L14). Below is an excerpt:

```python
# ...
 with open("/var/feast/feature_store.yaml") as f:
        feast_config = yaml.safe_load(f) # <---- yaml.safe_load() fails
# ...
```

The exact message is as below:

```python
Defaulted container "process" out of: process, init-hostfile (init)
Feast is an open source project that collects anonymized error reporting and usage statistics. To opt out or learn more see https://docs.feast.dev/reference/usage
Traceback (most recent call last):
  File "/bytewax/dataflow.py", line 15, in <module>
    feast_config = yaml.safe_load(f)
  File "/usr/local/lib/python3.9/site-packages/yaml/__init__.py", line 125, in safe_load
    return load(stream, SafeLoader)
  File "/usr/local/lib/python3.9/site-packages/yaml/__init__.py", line 81, in load
    return loader.get_single_data()
  File "/usr/local/lib/python3.9/site-packages/yaml/constructor.py", line 51, in get_single_data
    return self.construct_document(node)
  File "/usr/local/lib/python3.9/site-packages/yaml/constructor.py", line 60, in construct_document
    for dummy in generator:
  File "/usr/local/lib/python3.9/site-packages/yaml/constructor.py", line 413, in construct_yaml_map
    value = self.construct_mapping(node)
  File "/usr/local/lib/python3.9/site-packages/yaml/constructor.py", line 218, in construct_mapping
    return super().construct_mapping(node, deep=deep)
  File "/usr/local/lib/python3.9/site-packages/yaml/constructor.py", line 143, in construct_mapping
    value = self.construct_object(value_node, deep=deep)
  File "/usr/local/lib/python3.9/site-packages/yaml/constructor.py", line 100, in construct_object
    data = constructor(self, node)
[...]

PR #3912, opened 2024-01-29
fix: Bytewax materialization engine fails when loading feature_store.yaml

marti-jorda-roca:
PR ensures that the Bytewax materialization engine works.

**Which issue(s) this PR fixes**
yaml.safe_load() raises an error while trying to reconstruct the object below:

    pathlib.PosixPath

The error occurs while running materialization using Bytewax at the point where the feature_store.yaml is loaded. The code where this happens is in [sdk/python/feast/infra/materialization/contrib/bytewax/dataflow.py](https://github.com/feast-dev/feast/blob/ac6529c380bc623a15054350c9936b69617b4bee/sdk/python/feast/infra/materialization/contrib/bytewax/dataflow.py#L14). 

Fixes https://github.com/feast-dev/feast/issues/3893

### DIFF
--- a/sdk/python/feast/infra/materialization/contrib/bytewax/Dockerfile
+++ b/sdk/python/feast/infra/materialization/contrib/bytewax/Dockerfile
@@ -25,5 +25,5 @@ COPY README.md README.md
 # git dir to infer the version of feast we're installing.
 # https://github.com/pypa/setuptools_scm#usage-from-docker
 # I think it also assumes that this dockerfile is being built from the root of the directory.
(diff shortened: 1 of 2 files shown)


redhatHameed:
Looks good to me.

@etirelli can you take a look further on this. Thanks

etirelli (approved):

gterziysky:
Glad to see that the solution I suggested in https://github.com/feast-dev/feast/issues/3893 was adopted :)

PR #3922, opened 2024-01-31
chore: Loosen fsspec requirements to allow recent releases

cburroughs:
(I'm not sure the project has a super consistent pattern for when to specify a maximum version, but was going for the smallest possible change.)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -158,7 +158,6 @@ CI_REQUIRED = (
         "moto<5",
         "mypy>=1.4.1",
         "avro==1.10.0",
-        "fsspec<2023.10.0",
         "urllib3>=1.25.4,<3",
         "psutil==5.9.0",
         "py>=1.11.0",  # https://github.com/pytest-dev/pytest/issues/10420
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -155,7 +155,7 @@
         "grpcio-testing>=1.56.2,<2",
         "minio==7.1.0",
         "mock==2.0.0",
-        "moto",
+        "moto<5",
```
cburroughs, line 158: 5 changed around the api and current tests do not pass without modification.

sudohainguyen:
the last time we updated fsspec (from my commits), I got stuck when compiling by pyarrow deps.
But now we upgraded pyarrow successfully and It's fine to upgrade fsspec 🙌

setup.py
tokoko: I also stumbled into this, any idea why CI isn't failing as well?

cburroughs: I *suspect* that some transitive dependency was previously holding it back.

sudohainguyen: Moto v5 just released last week though, that's why CI succeeded

cburroughs:
What does this PR need to move forward?

sudohainguyen:
Just wanted to do integration tests since this issue is related to dependencies.

So closing this now? 🤔

cburroughs:
cc @franciscojavierarceo per dev call

franciscojavierarceo (approved):
lgtm

PR #3924, opened 2024-02-01
feat: Add gRPC Registry Server

### LINKED ISSUE #3901: Implement Python Registry Server
Different registry implementations are python-based that makes it hard to interact with registry objects from non-python applications. For example, Java-based feature server is still unable to use sql-based registry despite sql-based registry being recommended for production use. File-based registries are supported but necessitated reimplementation of feature registry logic in Java. 

Implement python-based registry server that will wrap the chosen registry calls into as a grpc server similar to how python-based feature server wraps calls to online store. The solution would simplify the development of non-python applications that can directly integrate with feature server rather than having to reimplement registry logic in the language of choice. 

Registry server might introduce some operational overhead, but will still be lightweight enough. Performance benefits offered by ease of use from apps like java feature server will most likely offset this operational overhead.

tokoko:
Implements a read-only python grpc server that wraps calls to the registry configured in feature store. Adds a cli command to start registry server from feast cli. Primary motivation behind this feature is to enable non-python feast applications (java feature server) to easily communicate with feast registry.

Fixes #3901


sudohainguyen:
do you mind adding a quick guide how to define registry in `feature_store.yaml` within the PR description? so that we can update documentation accordingly in another PR

tokoko:
Thanks for asking that. I'm trying to split the changes into smaller PRs, so the short answer right now is that you can't. This PR enables users to call `feast serve_registry` with a normal feature store configuration that will start the registry server on a specified port. I'm thinking of adding two different clients in subsequent PRs, a python client that will extend BaseRegistry and a java client for java feature server.

I'm not sure yet how feature_store.yaml should look like, I'm interested what others think about that. I think the cleanest approach would be to let the users use feature_store.yaml as a way to configure their clients. So, in cases when they need to access the registry directly (including the container from where `feast serve_registry` will be run), the config file will look something like this:

```
project: <your project name>
...
registry:
    registry_type: sql
    path: postgresql://postgres:mysecretpassword@127.0.0.1:55001/feast
    cache_ttl_seconds: 60
```

For applications where clients (java or python) connect to registry server, it might look something like this (this is still unimplemented!!!):

```
project: <your project name>
...
registry:
    registry_type: server
    url: grpc://127.0.0.1:8888
```

The biggest downside obviously is that you will need to have two different configurations depending on how you want to configure your client, but imho that's more of a feature than a bug. For example if a feast admin wants to force users to exclusively use registry server rather than connect to underlying implementation, they would be able to only share server information and hide implementation details. The same approach could in future be extrapolated to online store configuration, if you would like your `store.get_online_features` call to query feature server rather than an underlying online store, you would configure your store something like this:

```
project: <your project name>
...
online_store:
  type: server
  url: http://127.0.0.1:80 // or grpc://127.0.0.1:80
```

I definitely want to know what you guys think before delving into the implementations too much.

sudohainguyen:
@tokoko got your ideas, and I believe this is a pretty great one, which reduces headaches connecting components

sdk/python/feast/cli.py
```diff
@@ -753,6 +756,22 @@ def serve_transformations_command(ctx: click.Context, port: int):
...
+    type=click.INT,
+    default=DEFAULT_REGISTRY_SERVER_PORT,
+    help="Specify a port for the server",
+)
+@click.pass_context
+def serve_registry_command(ctx: click.Context, port: int):
+    """Start a registry server locally on a given port."""
+    store = create_feature_store(ctx)
+
+    store.serve_registry(port)
```
sudohainguyen, line 772: I expect there should be an option to broadcast the TTL cache to downstream function as well

tokoko: Do you mean cache_ttl_seconds config? That's not part of BaseRegistry interface and one would be able to configure it in feature_store.yaml anyway, right?

sudohainguyen: grpc registry server should have a cache itself to prevent concurrent repeated calls to the server

I believe we should add as an option, it's user's decision to keep it or not:

- default ttl may be considered as a small number (less than or equal 60 secs) 
- to disable cache, set it to -1

tokoko: The cache should be an option, I agree, but all registry implementations do have client-side (In this case the registry server itself acts as a "client") caches already that will accomplish exactly what you are looking for, no? If we add a separate cache for a registry server, you would end up having two identical chained caches right next to each other in the same python process. What would be an added benefit? Plus, it will be technically possible to specify different ttls for them and that's bound to get confusing.

sudohainguyen: yeah got it, think we should avoid confusion as you mentioned 🙌 
everything else looks good to me

sudohainguyen (approved):
lgtm

HaoXuAI (approved):
Realy great work. Nothing to comment on.

PR #3928, opened 2024-02-04
chore: Support pandas 2

sudohainguyen:
**What this PR does / why we need it**:
Relax pandas upper bound to support v2.0

**Which issue(s) this PR fixes**: 

Fixes #3709 

**Additional changes**:

Set upper bound for moto due to its recent breaking changes in [5.0.0](https://github.com/getmoto/moto/blob/master/CHANGELOG.md)

PR #3929, opened 2024-02-05
ci: Extend python base version for test cases

sudohainguyen:
Following up with [NEP](https://numpy.org/neps/nep-0029-deprecation_policy.html) and many libraries are dropping py38 support
We need to migrate our CI integration tests to py39+ version

**Additional information:**

Related to #3928, if we want to run integration tests with pandas v2, we must run on py3.9+,
So that we need to extend our integration tests workflow matrix to run on py310 additionally

Fixes #

### DIFF
--- a/.github/workflows/pr_integration_tests.yml
+++ b/.github/workflows/pr_integration_tests.yml
@@ -86,7 +86,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ "3.8" ]
+        python-version: [ "3.8", "3.10" ]
         os: [ ubuntu-latest ]
     env:
       OS: ${{ matrix.os }}
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -12,8 +12,6 @@ jobs:
         exclude:
           - os: macOS-latest
             python-version: "3.9"
-          - os: macOS-latest
-            python-version: "3.10"
     env:
       OS: ${{ matrix.os }}
       PYTHON: ${{ matrix.python-version }}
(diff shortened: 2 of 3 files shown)


tokoko:
Running integration tests only for 3.10 sounds a bit too risky unless we explicitly drop support for lower versions. I think all tests should include the lowest supported python version (3.8 currently, but if we in fact drop it then 3.9)

sudohainguyen:
> Running integration tests only for 3.10 sounds a bit too risky unless we explicitly drop support for lower versions. I think all tests should include the lowest supported python version (3.8 currently, but if we in fact drop it then 3.9)

actually we still do have full integration tests across 3 python versions on master only (after merged), however I agree that we should extend, not replace for now

sudohainguyen:
updated

.github/workflows/pr_integration_tests.yml
```diff
@@ -86,7 +86,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ "3.8" ]
+        python-version: [ "3.8", "3.10" ]
```
franciscojavierarceo, line 89: this purely adds 3.10 support

.github/workflows/unit_tests.yml
```diff
@@ -12,8 +12,6 @@ jobs:
         exclude:
           - os: macOS-latest
             python-version: "3.9"
-          - os: macOS-latest
-            python-version: "3.10"
```
franciscojavierarceo, line 16: did you mean to drop 3.10 here?

sudohainguyen: Drop from exclude list, which means we will conduct py310 unit test on macos

HaoXuAI (approved):
lgtm!

.github/workflows/unit_tests.yml
franciscojavierarceo: oh i missed the `exclude` key. Nice.

PR #3936, opened 2024-02-06
fix: Rewrite Spark materialization engine to use mapInPandas

tokoko:
Spark materialization engine currently converts data on partitions to a list of dicts before trying to write it to online store. This is terribly inefficient and all but guarantees OOM errors on executors for sufficiently large datasets. This PR changes the implementation to use `mapInPandas` and sidesteps the conversion to a list of dicts entirely.

P.S. There's unfortunately no equivalent `applyInPandas` method in pyspark, so the implementation simulates it by using `mapInPandas` and returning dummy results back to the driver.

P.P.S. Starting from 3.3.0 pyspark also has `mapInArrow`, which would simplify the operation even more, but would mean that feast would no longer support pyspark versions older than 3.3.0. I decided not to use it for now.


HaoXuAI (approved):
LGTM.
Will there be any type conversion changes?

tokoko:
Unfortunately we aren't testing data types extensively, so that's really hard to say. There will definitely be some conversion changes between pyspark<->pandas translation, because current implementation goes through python dicts and discards numeric types smaller than int for example. Having said that, it's unlikely to affect the final feast data types as those aren't part of feast type system anyway.

HaoXuAI:
@tokoko can you please resolve the conflicts and I'll add the `ok-to-test`, nice work!

tokoko:
@HaoXuAI All fixed, thanks.

HaoXuAI:
Curious if you are able to test the performance somehow?

tokoko:
Not really, this one seemed too obvious to do that. I plan to later try out conditionally using `mapInArrow` (only if running with pyspark>=3.3) instead of `mapInPandas`. I will try to benchmark it then.

rehevkor5:
+1 it also seems like this should be using real Spark sink connectors rather than ad-hoc procedural side effects.

tokoko:
@rehevkor5 true, but that's hard to do if you want to support arbitrary online store implementations. Maybe we should keep this as the default way to materialize and somehow introduce a way to override the logic for specific online stores. Not sure where that  sort of code belongs to, though... online stores or materilization engines?

Issue #3940, opened 2024-02-07
Add abstract caching registry


tokoko:
**Is your feature request related to a problem? Please describe.**
Several registry implementations employ client-side caching of registry proto, for example `sql` and `snowflake`. The code for caching is virtually identical and repeated.

**Describe the solution you'd like**
Add an abstract layer on top of base registry that implements caching logic and lets subclasses handle the logic only for the cases when there's a cache miss.

PR #3941, opened 2024-02-07
feat: Add python client for remote registry server

tokoko:
Adds another registry that wraps grpc calls to remote registry server. The registry is called `remote` (best i managed to come up with) and can be configured from feature_store.yaml like this:

```
registry:
    registry_type: remote
    path: localhost:6570
```
Both registry server and `remote` registry are still read-only for now. I also decided to skip implementing cache for the time being. I plan to add caching once #3940 gets resolved.

### DIFF
--- a/sdk/python/feast/infra/registry/registry.py
+++ b/sdk/python/feast/infra/registry/registry.py
@@ -178,6 +178,10 @@ class Registry(BaseRegistry):
             from feast.infra.registry.snowflake import SnowflakeRegistry
 
             return SnowflakeRegistry(registry_config, project, repo_path)
+        elif registry_config and registry_config.registry_type == "remote":
+            from feast.infra.registry.remote import RemoteRegistry
+
+            return RemoteRegistry(registry_config, project, repo_path)
         else:
             return super(Registry, cls).__new__(cls)
 
--- a/sdk/python/feast/infra/registry/remote.py
+++ b/sdk/python/feast/infra/registry/remote.py
@@ -1,0 +1,375 @@
+import logging
+import grpc
+from datetime import datetime
+from pathlib import Path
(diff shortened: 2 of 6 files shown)


sdk/python/feast/infra/registry/remote.py
```diff
@@ -0,0 +1,369 @@
...
+        self,
+        registry_config: Union[RegistryConfig, RemoteRegistryConfig],
+        project: str,
+        repo_path: Optional[Path],
+    ):
+        self.channel = grpc.insecure_channel(registry_config.path)
+        self.stub = RegistryServer_pb2_grpc.RegistryServerStub(self.channel)
+
+    def apply_entity(self, entity: Entity, project: str, commit: bool = True):
+        raise Exception("Unimplemented: remote registry is read-only")
```
sudohainguyen, line 46: consider adding new exception class in `errors.py`?

tokoko: fixed it

HaoXuAI (approved):
lgtm, cool feature!

sdk/python/feast/infra/registry/registry.py
```diff
@@ -178,6 +178,10 @@ def __new__(
             from feast.infra.registry.snowflake import SnowflakeRegistry
 
             return SnowflakeRegistry(registry_config, project, repo_path)
+        elif registry_config and registry_config.registry_type == "remote":
```
HaoXuAI, line 181: from this point I felt like the `Registry` is more like a common class instead of a file registry?

tokoko: yeah, confused me a lot as well. It probably used to be the common one, but then another base class was added and now it's a bit jumbled.

PR #3944, opened 2024-02-09
docs: Add ScyllaDB as online store alternative (fixed DCO)

zseta:

Add documentation about ScyllaDB and ScyllaDB Cloud as online store alternatives. ScyllaDB can use the already existing Cassandra connector.


HaoXuAI (approved):
LGTM!

PR #3949, opened 2024-02-12
feat: Decouple transformation types from ODFVs

### LINKED ISSUE #3945: Substrait-based on demand feature views
On demand feature views as implemented right now are very limited. The only way to specify odfvs is through a python function that takes in pandas Dataframe as input and outputs another pandas Dataframe. This leads to problems for both offline and online interfaces:

- Even the most scalable offline stores are forced to collect the whole dataset as a single pandas Dataframe to apply odfv function. There's no way for offline stores to incorporate computation in their engines.
- udfs in odfvs are inherently bound to pandas and python runtime. Non-python feature servers are stuck with the problem of figuring out how to run this functions if necessary. Java feature server for example has a separate python transformation service only for this reason, but that's obviously a subpar solution as the whole point of a java feature server was to avoid python runtime in feature serving in the first place.

Allow constructing odfvs as substrait plans. [Substrait](https://substrait.io/) is a protobuf-based serialization format for relational algebra operations. It is meant to be used as a cross-language and cross-engine format for sharing logical or physical execution plans. It has a number of producers (tools that can generate substrait) and consumers (engines that can run substrait) in different languages.

- Different offline stores will be able to inspect and incorporate substrait plans in their transformations. Even if that's impossible the default implementation inside feast to apply these functions will avoid pandas.
- Most importantly, non-python feature servers like a java feature server will be able to apply the functions without a separate python component. Apache Arrow java implementation comes with java bindings to Acero query engine that can consume substrait plans. (https://arrow.apache.org/docs/java/substrait.html#executing-queries-using-substrait-plans)

The example code in my PoC implementation looks something like this:

```
def generate_substrait():
    import ibis
    from ibis_substrait.compiler.core import SubstraitCompiler

    compiler = SubstraitCompiler()

    t = ibis.table([("conv_rate", "float"), ("acc_rate", "float")], "t")

    expr = t.select((t['conv_rate'] + t['acc_rate']).name('conv_rate_plus_acc_substrait'))

    return compiler.compile(expr).SerializeToString()

substrait_odfv = OnDemandFeatureView(
    name='substrait_view',
    sources=[driver_stats_fv],
    schema=[
        Field(name="conv_rate_plus_acc_substrait", dtype=Float64)
    ],
    substrait_plan=generate_substrait()
)
```
Substait plan object that feast accepts is `bytes` and introduces no external dependency. I'm using `ibis` and `ibis-substrait` to generate the plan. Right now that's the most practical way to generate substrait plan in python with DataFrame-like API, but this could have been any other substrait producer.

An obvious alternative to substrait is sql-based odfvs, but using SQL has a number of important downsides:

1. The presence of different sql dialects means that, it will be especially hard to ensure that sql-based feature functions will behave the same way across different offline store and online store implementations.
2. The user is implicitly bound to their offline store and online store of choice, because the dialect used in sql strings has to match offline store engine.

Having said that, it probably makes sense to support both substrait-based and sql-based odfvs, because at the moment it might be easier for sql-based logic to be incorporated inside offline store engines.

tokoko:
This PR decouples transformation logic and default engine from on demand feature views. It's a first step to eventually resolve - #3945. It introduces a new class `OnDemandPandasTransformation` and refactors pandas-related code from `OnDemandFeatureView`. This is to be followed by other types of transformations like `OnDemandSqlTransformation` and `OnDemandSubstraitTransformation`.

PR technically breaks `OnDemandFeatureView` "public" api, as the following code will no longer work:
```
OnDemandFeatureView(
        name="my-on-demand-feature-view",
        sources=sources,
        schema=[
            Field(name="output1", dtype=Float32),
            Field(name="output2", dtype=Float32),
        ],
        udf=udf1,
        udf_string="udf1 source code",
    )
```

Instead now you would do:
```
OnDemandFeatureView(
        name="my-on-demand-feature-view",
        sources=sources,
        schema=[
            Field(name="output1", dtype=Float32),
            Field(name="output2", dtype=Float32),
        ],
        transformation=OnDemandPandasTransformation(
            udf=udf1, udf_string="udf1 source code"
        ),
    )
```

Having said that, I think that's fine as except for internal tests probably noone creates on demand feature views like this. They instead would use a decorator like shown in [docs](https://docs.feast.dev/reference/alpha-on-demand-feature-view) which remains unchanged. Plus, on demand feature views are considered an alpha feature.

### DIFF
--- a/protos/feast/core/OnDemandFeatureView.proto
+++ b/protos/feast/core/OnDemandFeatureView.proto
@@ -48,7 +48,9 @@ message OnDemandFeatureViewSpec {
     // Map of sources for this feature view.
     map<string, OnDemandSource> sources = 4;
 
-    UserDefinedFunction user_defined_function = 5;
+    oneof transformation {
+        UserDefinedFunction user_defined_function = 5;
+    }
 
     // Description of the on demand feature view.
     string description = 6;
--- a/sdk/python/feast/on_demand_feature_view.py
+++ b/sdk/python/feast/on_demand_feature_view.py
@@ -2,7 +2,6 @@ import copy
 import functools
 import warnings
 from datetime import datetime
-from types import FunctionType
 from typing import Any, Dict, List, Optional, Type, Union
 
 import dill
@@ -16,6 +15,7 @@ from feast.errors import RegistryInferenceFailure, SpecifiedFeaturesNotPresentEr
 from feast.feature_view import FeatureView
 from feast.feature_view_projection import FeatureViewProjection
 from feast.field import Field, from_value_type
+from feast.on_demand_pandas_transformation import OnDemandPandasTransformation
 from feast.protos.feast.core.OnDemandFeatureView_pb2 import (
     OnDemandFeatureView as OnDemandFeatureViewProto,
 )
@@ -23,9 +23,6 @@ from feast.protos.feast.core.OnDemandFeatureView_pb2 import (
     OnDemandFeatureViewMeta,
     OnDemandFeatureViewSpec,
     OnDemandSource,
-)
-from feast.protos.feast.core.OnDemandFeatureView_pb2 import (
-    UserDefinedFunction as UserDefinedFunctionProto,
 )
 from feast.type_map import (
     feast_value_type_to_pandas_type,
@@ -51,8 +48,7 @@ class OnDemandFeatureView(BaseFeatureView):
             sources with type FeatureViewProjection.
         source_request_sources: A map from input source names to the actual input
             sources with type RequestSource.
-        udf: The user defined transformation function, which must take pandas dataframes
-            as inputs.
+        transformation: The user defined transformation.
         description: A human-readable description.
         tags: A dictionary of key-value pairs to store arbitrary metadata.
         owner: The owner of the on demand feature view, typically the email of the primary
@@ -63,8 +59,7 @@ class OnDemandFeatureView(BaseFeatureView):
     features: List[Field]
     source_feature_view_projections: Dict[str, FeatureViewProjection]
     source_request_sources: Dict[str, RequestSource]
-    udf: FunctionType
-    udf_string: str
+    transformation: Union[OnDemandPandasTransformation]
     description: str
     tags: Dict[str, str]
     owner: str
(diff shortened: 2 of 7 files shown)


protos/feast/core/OnDemandFeatureView.proto
```diff
@@ -48,7 +48,9 @@ message OnDemandFeatureViewSpec {
     // Map of sources for this feature view.
     map<string, OnDemandSource> sources = 4;
 
-    UserDefinedFunction user_defined_function = 5;
```
HaoXuAI, line 51: This won't be backwards compatible right?

sdk/python/feast/on_demand_feature_view.py
```diff
@@ -51,8 +48,7 @@ class OnDemandFeatureView(BaseFeatureView):
             sources with type FeatureViewProjection.
         source_request_sources: A map from input source names to the actual input
             sources with type RequestSource.
-        udf: The user defined transformation function, which must take pandas dataframes
```
HaoXuAI, line 54: can we still keep the `udf` and add `deprecated` to make it still backward compatible? I like the new field `transformation` but just in case some teams still use `udf` in the code base

tokoko: Sure, that's a good idea

tokoko: Done

HaoXuAI: Cool. nit, maybe keep the `udf` in the comment as well.

tokoko: It's in the comment for the init method. Do you mean the comment that lists the attributes? it's no longer an attribute, so probably shouldn't be there. It's being converted to transformation in init method and discarded.

HaoXuAI: ah that make sense

HaoXuAI (approved):
LGTM

tokoko:
@sudohainguyen this is no longer breaking. I added `udf` and `udf_string` parameters back in per Hao's suggestion.

tokoko:
@sudohainguyen @HaoXuAI any idea why integration tests haven't run for this?

HaoXuAI:
> @sudohainguyen @HaoXuAI any idea why integration tests haven't run for this?

I don't know what's going on but i re-run the job hope that works.

tokoko:
@HaoXuAI Can we merge this before other PRs get in the way? 😄

HaoXuAI:
> @HaoXuAI Can we merge this before other PRs get in the way? 😄

Just merged. great job!

franciscojavierarceo:
oh @HaoXuAI @tokoko i have all sorts of opinion here sad i missed this

franciscojavierarceo:
I was just looking to merge the python native transformations we made

PR #3952, opened 2024-02-16
fix: Revert mypy config

bushwhackr:
revert mypy scoped config

Changes:
+ Lint fixes

Fixes #

### DIFF
--- a/sdk/python/feast/infra/contrib/spark_kafka_processor.py
+++ b/sdk/python/feast/infra/contrib/spark_kafka_processor.py
@@ -5,7 +5,6 @@ import pandas as pd
 from pyspark.sql import DataFrame, SparkSession
 from pyspark.sql.avro.functions import from_avro
 from pyspark.sql.functions import col, from_json
-from pyspark.sql.streaming import StreamingQuery
 
 from feast.data_format import AvroFormat, JsonFormat
 from feast.data_source import KafkaSource, PushMode
@@ -68,13 +67,10 @@ class SparkKafkaProcessor(StreamProcessor):
         # data_source type has been checked to be an instance of KafkaSource.
         self.data_source: KafkaSource = self.data_source  # type: ignore
 
-    def ingest_stream_feature_view(
-        self, to: PushMode = PushMode.ONLINE
-    ) -> StreamingQuery:
+    def ingest_stream_feature_view(self, to: PushMode = PushMode.ONLINE) -> None:
         ingested_stream_df = self._ingest_stream_data()
         transformed_df = self._construct_transformation_plan(ingested_stream_df)
-        online_store_query = self._write_stream_data(transformed_df, to)
-        return online_store_query
+        self._write_stream_data(transformed_df, to)
 
     # In the line 64 of __init__(), the "data_source" is assigned a stream_source (and has to be KafkaSource as in line 40).
     @no_type_check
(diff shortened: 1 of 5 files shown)


sdk/python/feast/infra/offline_stores/snowflake.py
```diff
@@ -470,18 +470,9 @@ def _to_df_internal(self, timeout: Optional[int] = None) -> pd.DataFrame:
...
-            return pa_table
-        else:
-            empty_result = execute_snowflake_statement(
-                self.snowflake_conn, self.to_sql()
-            )
-
-            return pyarrow.Table.from_pandas(
-                pd.DataFrame(columns=[md.name for md in empty_result.description])
-            )
+        return pa_table
```
bushwhackr, line 475: simply, refer to 
```
    def fetch_arrow_all(self, force_return_table: bool = False) -> Table | None:
        """
        Args:
            force_return_table: Set to True so that when the query returns zero rows,
                an empty pyarrow table will be returned with schema using the highest bit length for each column.
                Default value is False in which case None is returned in case of zero rows.
        """
 ```

**Comment by bushwhackr** [comment]
@sudohainguyen Hi Harry, just reverting the mypy changes.

File: sdk/python/feast/infra/contrib/spark_kafka_processor.py [hunk 2]
```diff
@@ -68,13 +67,10 @@ def __init__(
         # data_source type has been checked to be an instance of KafkaSource.
         self.data_source: KafkaSource = self.data_source  # type: ignore
 
-    def ingest_stream_feature_view(
-        self, to: PushMode = PushMode.ONLINE
-    ) -> StreamingQuery:
+    def ingest_stream_feature_view(self, to: PushMode = PushMode.ONLINE) -> None:
```
sudohainguyen, line 70: Why should we modify the function behaviour 🤔 ?

bushwhackr: The base class does not return anything, I'm just following its signature.

sudohainguyen: I see, clear

sdk/python/feast/infra/offline_stores/snowflake.py
sudohainguyen: think we should leave it for another PR, since this is a feature-level change

sdk/python/feast/infra/contrib/spark_kafka_processor.py
bushwhackr: I reverted this change because I don't know enough about the codebase. Looking at the pyspark.StreamingQuery class, some users might be actively using the return type, for example, to stop the query.

bushwhackr: I changed the signature to return Optional[Any] instead to fix the linting issue.

sdk/python/feast/infra/offline_stores/snowflake.py
bushwhackr: reverted

sudohainguyen (approved):
lgtm

bushwhackr:
is anything blocking this PR?

shuchu:
@bushwhackr  I met a lint error: 

feast/infra/offline_stores/snowflake.py:473: error: Unexpected keyword argument "force_return_table" for "fetch_arrow_all" of "SnowflakeCursor"  [call-arg]

my dev env:
python 3.8, 
snowflake-connector-python    3.5.0

bushwhackr:
[Refer to this commit](https://github.com/snowflakedb/snowflake-connector-python/pull/1801/files)

Maybe we have to bump the minimum version of snowflake-connector python

sudohainguyen:
upgrade snowflake connector to 3.7.1 should work @shuchu

bushwhackr:
@shuchu just checked the minimum version of snowflake-connector-python that supports that parameter, 3.7.0. Is there any concerns for supporting a version of snowflake-connector-python < 3.6.0? Else I can create a PR that bumps the minimum.

The alternative is to change the method signature of `_to_arrow_internal`. Because `snowflake-connector-python` might return `None` for empty tables.

bushwhackr:
Just wondering if you are going to increase the minimum version of snowflake to 3.7.0. Kindly refer to this PR: https://github.com/snowflakedb/snowflake-connector-python/issues/1800. 
We could essentially remove the whole None check and just return the result from `fetch_arrow_all` to simplify things.

shuchu:
No worries, let me update my snowflake connector version. :)

PR #3961, opened 2024-02-21
fix: Fix for materializing entityless feature views in Snowflake

JohnLemmonMedely:
**What this PR does / why we need it**: When using an entityless featureview with a Snowflake offline store it fails to materialize correctly.

Fixes # 3960

### DIFF
--- a/sdk/python/feast/infra/materialization/snowflake_engine.py
+++ b/sdk/python/feast/infra/materialization/snowflake_engine.py
@@ -14,7 +14,7 @@ from tqdm import tqdm
 import feast
 from feast.batch_feature_view import BatchFeatureView
 from feast.entity import Entity
-from feast.feature_view import FeatureView
+from feast.feature_view import DUMMY_ENTITY_ID, FeatureView
 from feast.infra.materialization.batch_materialization_engine import (
     BatchMaterializationEngine,
     MaterializationJob,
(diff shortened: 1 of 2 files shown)


sdk/python/feast/infra/materialization/snowflake_engine.py
```diff
@@ -274,7 +274,11 @@ def _materialize_one(
 
             fv_latest_values_sql = offline_job.to_sql()
 
-            if feature_view.entity_columns:
+            if (
+                feature_view.entity_columns[0].name == DUMMY_ENTITY_ID
+            ):  # entityless Feature View's placeholder entity
+                entities_to_write = 1
+            else:
```
JohnLemmonMedely, line 281: Entityless feature views do have an entity column, it's just that it's a dummy value.

franciscojavierarceo:
@JohnLemmonMedely can you provide more context on this one? Do we support entityless feature views outside of this? Or is this a bit of a hack?

JohnLemmonMedely:
> @JohnLemmonMedely can you provide more context on this one? Do we support entityless feature views outside of this? Or is this a bit of a hack?

Yes, I hit this bug when setting up a feature view [like this](https://docs.feast.dev/getting-started/concepts/feature-view#feature-views-without-entities). It looks like the documentation calls this "Feature views without entities" but the code refers to it as "entityless feature views".

In this case the original code was checking for an entityless feature view by checking if `entity_columns` is truthy. I'm not sure if this was once a valid check or not but today in Feast entityless feature views are given a dummy entity_column instead.

JohnLemmonMedely:
@franciscojavierarceo Does my above comment answer your question? Anything else I need to do before merging this in?

franciscojavierarceo (approved):
this lgtm. your pr title lint failed

JohnLemmonMedely:
> this lgtm. your pr title lint failed

Thanks @franciscojavierarceo, I think I fixed the PR title earlier so the check is passing now.

PR #3964, opened 2024-02-21
fix: Bugfix for grabbing historical data from Snowflake with array type features.

JohnLemmonMedely:

A bug appears when the following happens.

1. Fetch from a Snowflake-backed FeatureView.
2. The FeatureView has an array-type column.
3. One of the fetched entities does not have a record in the Snowflake table.

This add a null check and includes the documentation around Snowflake now supporting array types. I forgot to add it in my previous PR.

Fixes # 3963

### DIFF
--- a/docs/reference/data-sources/overview.md
+++ b/docs/reference/data-sources/overview.md
@@ -19,13 +19,13 @@ Details for each specific data source can be found [here](README.md).
 Below is a matrix indicating which data sources support which types.
 
 | | File | BigQuery | Snowflake | Redshift | Postgres | Spark | Trino |
-| :-------------------------------- | :-- | :-- | :-- | :-- | :-- | :-- | :-- |
-| `bytes`     | yes | yes | yes | yes | yes | yes | yes |
-| `string`    | yes | yes | yes | yes | yes | yes | yes |
-| `int32`     | yes | yes | yes | yes | yes | yes | yes |
-| `int64`     | yes | yes | yes | yes | yes | yes | yes |
-| `float32`   | yes | yes | yes | yes | yes | yes | yes |
-| `float64`   | yes | yes | yes | yes | yes | yes | yes |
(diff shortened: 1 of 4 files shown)


docs/reference/data-sources/overview.md
```diff
@@ -19,13 +19,13 @@ Details for each specific data source can be found [here](README.md).
...
+| :-------------------------------- | :-- | :-- |:----------| :-- | :-- | :-- | :-- |
+| `bytes`     | yes | yes | yes       | yes | yes | yes | yes |
+| `string`    | yes | yes | yes       | yes | yes | yes | yes |
+| `int32`     | yes | yes | yes       | yes | yes | yes | yes |
+| `int64`     | yes | yes | yes       | yes | yes | yes | yes |
+| `float32`   | yes | yes | yes       | yes | yes | yes | yes |
+| `float64`   | yes | yes | yes       | yes | yes | yes | yes |
+| `bool`      | yes | yes | yes       | yes | yes | yes | yes |
+| `timestamp` | yes | yes | yes       | yes | yes | yes | yes |
+| array types | yes | yes | yes       | no  | yes | yes | no  |
```
JohnLemmonMedely, line 31: I marked this as yes to keep it simple. But the full truth is that it does not work when inferring features. That note is included in the other edited document.

sdk/python/feast/infra/offline_stores/snowflake.py
```diff
@@ -463,7 +463,9 @@ def _to_df_internal(self, timeout: Optional[int] = None) -> pd.DataFrame:
                     Array(Float32),
                     Array(Bool),
                 ]:
-                    df[feature.name] = [json.loads(x) for x in df[feature.name]]
+                    df[feature.name] = [
+                        json.loads(x) if x else None for x in df[feature.name]
```
franciscojavierarceo, line 467: nice

franciscojavierarceo (approved):

PR #3969, opened 2024-02-24
feat: Add Substrait-based ODFV transformation

### LINKED ISSUE #3945: Substrait-based on demand feature views

tokoko:
This PR introduces a new ODFV transformation type, based on [substrait](https://substrait.io/). The rationale behind the feature is described in an issue #3945. PR also adds a simple way for users to express substrait transformations using ibis. `ibis` and `ibis-substrait` librarie dependencies are added under an extra called `ibis` for users willing to use this feature. 

The way to express substrait-based ODFV is pretty similar to the existing pandas-based version, the only difference being that the user-defined function must have a return type of `ibis.expr.types.relations.Table` instead of `pandas.DataFrame`.

```
from ibis.expr.types import Table
@on_demand_feature_view(
    sources=[driver_stats_fv[["conv_rate", "acc_rate"]]],
    schema=[Field(name="conv_rate_plus_acc_substrait", dtype=Float64)],
)
def substrait_view(inputs: Table) -> Table:
    return inputs.select(
        (inputs["conv_rate"] + inputs["acc_rate"]).name("conv_rate_plus_acc_substrait")
    )
```

Fixes #3945


sdk/python/tests/unit/test_on_demand_substrait_transformation.py
```diff
@@ -0,0 +1,112 @@
...
+            df["conv_rate_plus_acc"] = inputs["conv_rate"] + inputs["acc_rate"]
+            return df
+
+        from ibis.expr.types import Table
+
+        @on_demand_feature_view(
+            sources=[driver_stats_fv[["conv_rate", "acc_rate"]]],
+            schema=[Field(name="conv_rate_plus_acc_substrait", dtype=Float64)],
+        )
+        def substrait_view(inputs: Table) -> Table:
```
HaoXuAI, line 75: Something can try is creat a FeastTable or FeastDataframe class that wrap pandas df and other table etc.

tokoko: You mean like an abstraction on top of both? But that would necessitate whole API for transformations around it and in the end we would essentially be recreating ibis.

HaoXuAI: Maybe just a type class that can be accepting different apis. Anyway I'm not sure how to implement it yet. I think it looks good with your implementation

HaoXuAI (approved):

HaoXuAI:
I think this is a great feature that should include in the documentation somewhere

tokoko:
Thanks, I'll follow up shortly with a PR for doc changes.

sudohainguyen:
hey this is a cool feature, would love to see the RFC or something similar to have a deeper understanding about your idea

tokoko:
@sudohainguyen hey, I shared this [document](https://docs.google.com/document/d/1NEyp7J-cQTqFOX7haLHS7aa5VXapWnTwA_sb7dq1GOQ) a couple months back on slack. It's a first draft for a RFC, somewhat incomplete though. I should probably request access to RFC google folder and turn this into a real RFC document.

PR #3980, opened 2024-03-02
fix: remove not use input parameter in spark source

ElliotNguyen68:
**What this PR does / why we need it**: Remove a not used anywhere parameter when init SparkSource 

Fixes #


HaoXuAI (approved):
LGTM

Issue #3979, opened 2024-03-01
Abstract ibis-based Offline Store


tokoko:
**Is your feature request related to a problem? Please describe.**
- Each feast offline store has to more or less reimplement the same data transformation logic (mainly point-in-time join) in their respective technologies. It's hard to ensure feature parity.
- It's next to impossible to properly incorporate ODFV logic into offline store operations.

**Describe the solution you'd like**
- Implement abstract offline store with ibis that other feature stores can extend and customize. Ibis is a lightweight python library that offers a universal dataframe API with pluggable query engine backends. As of now it can interface with 15+ engines with varying level of API coverage, including BigQuery, Snowflake, Pyspark, PosgreSQL, Trino and others.
- ibis-based offline stores should be able to turn Substrait-based on demand feature view transformations into ibis expressions and execute them right alongside core offline store logic in one go.

PR #3981, opened 2024-03-02
feat: Add duckdb offline store

tokoko:
This PR adds duckdb as a contrib offline store to feast. Introduces another extra named `duckdb` which is not part of a `ci` extra as we don't run ci tests for contrib components. The changes can be tested by running `test-python-universal-duckdb` command from Makefile. The primary reason why this need to be a contrib component as of now is that it's only targeting Python 3.9+. While possible to backport it to Python 3.8 as well, `ibis` on which the codebase depends on has dropped 3.8 support a long time ago and I thought it wasn't worth wasting time on 3.8 compatibility when we are slated to drop the support pretty soon anyway.

This PR supersedes #3822 and is a first step to eventually address - #3979. While this PR targets duckdb as an offline store, the main motivation behind it is to use `duckdb` implementation as a testing ground and eventually start porting other offline stores to ibis as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -186,6 +186,19 @@ test-python-universal-athena:
 			not test_snowflake" \
 	sdk/python/tests
 
+test-python-universal-duckdb:
+	PYTHONPATH='.' \
+	FULL_REPO_CONFIGS_MODULE=sdk.python.feast.infra.offline_stores.contrib.duckdb_repo_configuration \
+	FEAST_USAGE=False IS_TEST=True \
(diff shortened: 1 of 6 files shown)


jeremyary:
Syntactically & inclusion-wise, this lgtm. Ibis license is Apache v2.0, project is active to the tune of multiple commits a day. Would be great to have one more set of eyes for the technical concept of this add to complement, so will await another reviewer.

HaoXuAI (approved):
LGTM!

PR #3986, opened 2024-03-05
fix: Update setup-python from v3 to v4 to avoid cache size limit

HaoXuAI:

Fixes #

### DIFF
--- a/.github/workflows/lint_pr.yml
+++ b/.github/workflows/lint_pr.yml
@@ -7,12 +7,13 @@ on:
       - edited
       - synchronize
 
+permissions:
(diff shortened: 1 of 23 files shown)


jeremyary:
apologies for the title bump! I noticed possible [skips](https://github.com/feast-dev/feast/pull/3987) with the title lint job, so I'm just making sure we're all good on PR's while that's worked out.

HaoXuAI:
> apologies for the title bump! I noticed possible [skips](https://github.com/feast-dev/feast/pull/3987) with the title lint job, so I'm just making sure we're all good on PR's while that's worked out.

no worries, thanks for the update :)

PR #3987, opened 2024-03-05
fix: Swap security label check on the PR title validation job to explicit permissions instead

jeremyary:
After recently adding security label checks to all `pull_request_target` actions that were missing it, I noticed title validation checkjs started being [skipped](https://github.com/feast-dev/feast/actions/runs/8162568306/job/22313824999?pr=3986) due to the conditional paired with this trigger type. This PR takes the check logic back to the simple repo verify, but adds explicit read-only permissions matching the [recommended](https://github.com/amannn/action-semantic-pull-request#installation) install configuration.

### DIFF
--- a/.github/workflows/lint_pr.yml
+++ b/.github/workflows/lint_pr.yml
@@ -7,12 +7,13 @@ on:
       - edited
       - synchronize
 
+permissions:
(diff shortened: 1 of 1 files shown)


HaoXuAI (approved):
LGTM!

PR #3988, opened 2024-03-06
feat: Add Entity df in format of a Spark Dataframe instead of just pd.DataFrame or string for SparkOfflineStore

ElliotNguyen68:
**What this PR does / why we need it**: For handling bigdata, using sparkofflinestore, it is not efficient to create entity df in pandas dataframe type then pass to get_historical_features function, this pr will add ability to input a Spark DataFrame entitydf 

Fixes #

### DIFF
--- a/sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
+++ b/sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
@@ -125,7 +125,7 @@ class SparkOfflineStore(OfflineStore):
         config: RepoConfig,
         feature_views: List[FeatureView],
         feature_refs: List[str],
-        entity_df: Union[pandas.DataFrame, str],
+        entity_df: Union[pandas.DataFrame, str, pyspark.sql.DataFrame],
         registry: Registry,
         project: str,
         full_feature_names: bool = False,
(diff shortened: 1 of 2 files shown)


sudohainguyen:
@ElliotNguyen68 do you mind adding some unit tests to cover the new incoming code?

ElliotNguyen68:
@sudohainguyen  yes I will add it

ElliotNguyen68:
@sudohainguyen, could you help me where should I put the unittest into ?

sudohainguyen:
oh we didn't have a unit test for spark offline store yet
think we can handle it on another PR

ElliotNguyen68:
OK @sudohainguyen , so for this PR we are all good ?

sudohainguyen:
Yep all good

ElliotNguyen68:
@sudohainguyen  so can you merge my commit to the repo ?

jeremyary (approved):

HaoXuAI:
hey @sudohainguyen good feature 👍  can you also please update the README here: https://github.com/feast-dev/feast/blob/43b2c287705c2a3e882517524229f155c9ce0a01/docs/reference/offline-stores/spark.md?plain=1?

tokoko:
While this is probably a worthwhile change, I'd point out that this is breaking the method signature from `OfflineStore` abstract class (I know, it's python, who cares). The same about `get_historical_features` method of `FeatureStore` class. For a client to reach offline store, it has to first call that method on `FeatureStore` which is expecting either a pandas dataframe or a string, but fortunately is not actively typechecking. BTW, another way to accomplish the same end result would be to call `createOrReplaceTempView` on the dataframe and pass the name as a string to `entity_df` parameter.

PR #3990, opened 2024-03-06
feat: Read file recursively on sparksource

ElliotNguyen68:
**What this PR does / why we need it**: When do the feature engineering(feature1) with parquet files, we want by cadence return a 'file' with a partition as file name, example /path/to/feature1/day1, /path/to/feature1/day2, so when declare path for sparksource, we want to just need to declare to /path/to/feature1 and under the hood, there will have another team or job to take care of the day1, day2,...,dayn part(those will be partitioned by cadence).

Fixes #

### DIFF
--- a/sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark_source.py
+++ b/sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark_source.py
@@ -8,7 +8,7 @@ from pyspark.sql import SparkSession
 
 from feast import flags_helper
 from feast.data_source import DataSource
-from feast.errors import DataSourceNoNameException
+from feast.errors import DataSourceNoNameException, DataSourceNotFoundException
 from feast.infra.offline_stores.offline_utils import get_temp_entity_table_name
 from feast.protos.feast.core.DataSource_pb2 import DataSource as DataSourceProto
 from feast.protos.feast.core.SavedDataset_pb2 import (
@@ -183,6 +183,7 @@ class SparkSource(DataSource):
             logger.exception(
                 "Spark read of file source failed.\n" + traceback.format_exc()
             )
+            raise DataSourceNotFoundException(path=self.path)
         tmp_table_name = get_temp_entity_table_name()
         df.createOrReplaceTempView(tmp_table_name)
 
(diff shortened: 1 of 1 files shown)


ElliotNguyen68:
> great one! @ElliotNguyen68

ElliotNguyen68:
cc @sudohainguyen  please give @jeremyary  the answer, I also really want to hear from you :v

tokoko:
@ElliotNguyen68 I may be wrong but I think with a "table" that has the following folder structure, this option won't be backwards-compatible. `partition_day` column will no longer be part of the resulting dataframe. (?) If that's the case, maybe we can somehow allow users to specify desired options during `feast apply`, but we should not enable this option for all data sources. 
```
/path/to/feature1/partition_day=day1/dataset1.parquet
/path/to/feature1/partition_day=day2/dataset2.parquet
```

Also, I understand what you're trying to accomplish here, but wouldn't using some sort of a higher-level table format like delta or iceberg be a much better solution? They have features that make the need to rewrite the whole partitions go away and so much more.

ElliotNguyen68:
thanks @tokoko, I think it would be great if you could give me some references for this case.

sudohainguyen:
> does this count as a review approval? :) or just interested in the content?

😄 
Genuinely not an official approval though, just a compliment

ElliotNguyen68:
Hi Guys, I think I can solve my problem using @tokoko   s  approach, so when write feature file, I just need to at feature1/partition=day1 and everything will be kep the same. So I think we can close this PR

sudohainguyen:
@ElliotNguyen68 it is called hive partitioning layout FYI, which is a standardized approach, instead of handling string.

I think @tokoko meant to decouple code logic from table format, so that adding recursive reads will break that philosophy

ElliotNguyen68:
> @ElliotNguyen68 it is called hive partitioning layout FYI, which is a standardized approach, instead of handling string.
> 
> I think @tokoko meant to decouple code logic from table format, so that adding recursive reads will break that philosophy

Yes I see, so we can now close this PR, thanks for your support @tokoko , @sudohainguyen

jeremyary:
@ElliotNguyen68 thanks to you for contributing & engaging us in conversation, glad we found resolution for you!

PR #3993, opened 2024-03-07
feat: New PR - Updating dependencies so that feast can be run on 3.11. 

lokeshrangineni:

Fixes #

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ NAME = "feast"
 DESCRIPTION = "Python SDK for Feast"
 URL = "https://github.com/feast-dev/feast"
 AUTHOR = "Feast"
-REQUIRES_PYTHON = ">=3.8.0"
+REQUIRES_PYTHON = ">=3.9.0"
 
 REQUIRED = [
     "click>=7.0.0,<9.0.0",
@@ -195,7 +195,6 @@ CI_REQUIRED = (
         "types-setuptools",
         "types-tabulate",
         "virtualenv<20.24.2",
-        "pandas>=1.4.3,<2; python_version < '3.9'",
     ]
     + GCP_REQUIRED
     + REDIS_REQUIRED
--- a/.github/fork_workflows/fork_pr_integration_tests_aws.yml
+++ b/.github/fork_workflows/fork_pr_integration_tests_aws.yml
@@ -67,7 +67,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ "3.8" ]
+        python-version: [ "3.9" ]
         os: [ ubuntu-latest ]
     env:
       OS: ${{ matrix.os }}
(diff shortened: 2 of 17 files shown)


setup.py
```diff
@@ -194,7 +194,7 @@
         "types-setuptools",
         "types-tabulate",
         "virtualenv<20.24.2",
-        "pandas>=1.4.3,<2; python_version < '3.9'",
+        "pandas>=1.4.3,<2",
```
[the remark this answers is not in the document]
lokeshrangineni: Sorry @tokoko, I am a beginner to python so forgive my understanding. I have updated the PR again with the updated feedback. Please review it again.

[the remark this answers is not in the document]
tokoko: looks good, thanks

sdk/python/feast/infra/materialization/contrib/bytewax/bytewax_materialization_dataflow.py
whoahbot, line 1: Thanks for putting this PR together.

There are a number of API changes between Bytewax 0.15.1 and 0.18.2 that would need to be addressed.

I took at stab at rewriting this dataflow using the 0.18.2 API here: [gist](https://gist.github.com/whoahbot/29255324f925447cc9b3aa6db3a4e9ef), but I may need some help testing these changes as I don't currently have a working Feast materialization environment.

One other thing to consider is that we would need to coordinate building a new docker image that will be used for materialization. The version referenced in the [documentation](https://github.com/lokeshrangineni/feast/blob/master/docs/reference/batch-materialization/bytewax.md?plain=1#L57) is unfortunately using the `latest` tag, so existing materialization that use that tag would break when publishing a new version of the image. I'm happy to coordinate building and pushing that image to our docker hub account when we get things sorted.

tokoko: @whoahbot thanks for pointing that out. How can we test it better? I saw that the existing integration test depends on eks, is there any specific reason for that? Can we rewrite it to work on minikube or kind for example?

whoahbot: IIRC the test depends on S3, Redshift and DynamoDB, and uses AWS permissions from an EKS cluster to access those resources.

jeremyary: First off, thanks for jumping back in with us @whoahbot! Tackling all the breaking changes we'll need to step through is sure to be easier with your help. @tokoko was just discussing over in Slack with us the possibility of splitting the PR into 2 efforts - one for dropping 3.8 and another for adding 3.11 so that we can give the latter proper testing. If things start shuffling without discussion here, that's probably why.

I'm sure you stay covered up with work, but if you find a chance to swing through Slack or catch a community call, would be great to say hello and introduce some new faces.

lokeshrangineni: After the conversation to split up this PR in to two separate tasks for better manageability - I have created the PR to drop the Python 3.8 support - https://github.com/feast-dev/feast/pull/4010

PR #3996, opened 2024-03-07
feat: Enable Arrow-based columnar data transfers 

ElliotNguyen68:
**What this PR does / why we need it**: When to_df for Spark Retrieval Job, we will need to transfer from spark dataframe to pandas dataframe, to be better handle this transfer process, I add a config to spark session, refer to https://docs.databricks.com/en/pandas/pyspark-pandas-conversion.html

Fixes #

### DIFF
--- a/docs/reference/offline-stores/spark.md
+++ b/docs/reference/offline-stores/spark.md
@@ -30,6 +30,8 @@ offline_store:
         spark.sql.catalogImplementation: "hive"
         spark.sql.parser.quotedRegexColumnNames: "true"
         spark.sql.session.timeZone: "UTC"
+        spark.sql.execution.arrow.fallback.enabled: "true"
+        spark.sql.execution.arrow.pyspark.enabled: "true"
 online_store:
     path: data/online_store.db
 ```
--- a/sdk/python/feast/templates/spark/feature_repo/feature_store.yaml
+++ b/sdk/python/feast/templates/spark/feature_repo/feature_store.yaml
@@ -12,6 +12,8 @@ offline_store:
     spark.sql.catalogImplementation: "hive"
     spark.sql.parser.quotedRegexColumnNames: "true"
     spark.sql.session.timeZone: "UTC"
+    spark.sql.execution.arrow.fallback.enabled: "true"
+    spark.sql.execution.arrow.pyspark.enabled: "true"
 online_store:
   path: data/online_store.db
 entity_key_serialization_version: 2


ElliotNguyen68:
Hi @sudohainguyen  could you reivew the pr ?

sudohainguyen:
can we eliminate type hint changes and leave it for another PR?
it's better one thing focused only 😄

sudohainguyen:
another comment, have you tested the performance before and after enabling arrow transfer? how was it? 
genuinely wondering

ElliotNguyen68:
> can we eliminate type hint changes and leave it for another PR? it's better one thing focused only 😄

got it

ElliotNguyen68:
> another comment, have you tested the performance before and after enabling arrow transfer? how was it? genuinely wondering

yes I tested it using databrick (3 worker nodes), the date contains over 6 milions rows, with the config enable, this take about 11sec, when not even over 6 minutes  still cannot get the result

ElliotNguyen68:
@sudohainguyen  detail the tests 
image
image
image

sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
```diff
@@ -338,6 +338,11 @@ def to_spark_df(self) -> pyspark.sql.DataFrame:
 
     def _to_df_internal(self, timeout: Optional[int] = None) -> pd.DataFrame:
         """Return dataset as Pandas DataFrame synchronously"""
+        spark_session = get_spark_session_or_start_new_with_repoconfig(
+            self._config.offline_store
+        )
+        spark_session.conf.set("spark.sql.execution.arrow.fallback.enabled", "true")
```
HaoXuAI, line 344: Maybe we can explicitly claim this in the document, the spar offline store is leveraging arrow to process data. And it's probably better to have a config for it to enable or not as wished by the user.

ElliotNguyen68: So how  can we achieve this @HaoXuAI ?

ElliotNguyen68: Hi @HaoXuAI , I check on snowflake onffline store source code, snowflake.py and can see that the code owner also use spark.conf.set('spark.sql.execution.arrow.pyspark.enabled', 'true') within the codebase, without from a config, so I think we can also follow this, because this is a good thing for feast performance, (1 more thinkg to notice is that our requirement for feast is pyspark version >=3.0.0, and the default value for this config is false for spark session (https://spark.apache.org/docs/latest/configuration.html#:~:text=3.0%2C%20please%20set%20%27-,spark.sql.execution.arrow.pyspark.enabled,-%27.)), what do you think  ?

sudohainguyen: I suggest turning it into a offline store configuration

ElliotNguyen68: So you mean we will let the user to config this when they init FeatureStore object using repo config or feature_store.yml , isnt it ?

tokoko: @sudohainguyen It's already an offline store configuration in a way though, isn't it? You can specify spark conf values from offline store config

ElliotNguyen68: Yes I understand, but I think not very much feast user will know about this config, while this config is helpful in boosting feast performance

tokoko: @ElliotNguyen68 Yes, I agree. But I still don't like the idea of having two methods of setting the same config. We will have to handle cases when user sets both, for example and stuff like that. Maybe we should just document this option better on [spark](https://docs.feast.dev/reference/offline-stores/spark) page. Let's include it in the example and also add a sentence about why it might be useful.

ElliotNguyen68: Ok so how to include it in the document 🙂?

HaoXuAI: https://github.com/feast-dev/feast/blob/master/docs/reference/offline-stores/spark.md. this is the spark document code, you can add it here.
Also looks like this is the place to initialize the spark config: https://github.com/feast-dev/feast/blob/a0f7472f200300f3a45aa404922dd67bb4ad237f/sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py#L455, can you instead add it here?

tokoko: @ElliotNguyen68 https://github.com/feast-dev/feast/tree/master/sdk/python/feast/templates/spark can you add it to spark template as well? Makes sense that people might use it when bootstrapping a new project.

ElliotNguyen68: Hi @HaoXuAI , @tokoko , I change those changes to md and template file already, do you guys think that is ok for now ?

HaoXuAI (approved):
LGTM!

PR #3997, opened 2024-03-07
fix: Extend SQL registry config with a sqlalchemy_config_kwargs key

galen-ft:

The PR introduces an optional `sqlalchemy_config_kwargs` key to the [SQL Registry](https://docs.feast.dev/getting-started/concepts/registry#sql-registry) config. The new key allows passing in custom `**kwargs` to [sqlalchemy.create_engine](https://github.com/feast-dev/feast/blob/fa8492dfe7f38ab493a8d35a412ec9334a0ff6b9/sdk/python/feast/infra/registry/sql.py#L202).

One such kwarg is [pool_pre_ping: true](https://docs.sqlalchemy.org/en/20/core/engines.html#sqlalchemy.create_engine) which in turn helps avoid errors due to closed or timed out connections. Such errors happen in particular when running the Python Feature Server as an API.

Example errors this prevents:
```python
psycopg2.OperationalError: connection to server at "<SERVER_DNS>" (<IP_ADDRESS>), port 5432 failed: Connection timed out
	Is the server running on that host and accepting TCP/IP connections?
```

```python
sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
```

If `sqlalchemy_config_kwargs` is not provided, by default the `echo=False` option is applied just as in the original code [sdk/python/feast/infra/registry/sql.py#L202](https://github.com/feast-dev/feast/blob/fa8492dfe7f38ab493a8d35a412ec9334a0ff6b9/sdk/python/feast/infra/registry/sql.py#L202).

The following files are updated:
* docs/getting-started/concepts/registry.md - to show how the new key can be used
* docs/tutorials/using-scalable-registry.md - same as above
* sdk/python/feast/infra/registry/sql.py - required
* sdk/python/feast/repo_config.py - needed to silence a linter error that enforces strict OOP (i.e. a base class must define property which is overriden in a derived class)
*  sdk/python/tests/unit/test_sql_registry.py - added to the PostgreSQL and MySQL unit tests

All unit tests pass.

Fixes #

### DIFF
--- a/docs/getting-started/concepts/registry.md
+++ b/docs/getting-started/concepts/registry.md
@@ -57,6 +57,8 @@ registry:
     registry_type: sql
     path: postgresql://postgres:mysecretpassword@127.0.0.1:55001/feast
     cache_ttl_seconds: 60
+    sqlalchemy_config_kwargs:
+        pool_pre_ping: true
 ```
 
 This supports any SQLAlchemy compatible database as a backend. The exact schema can be seen in [sql.py](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/registry/sql.py)
--- a/docs/getting-started/concepts/registry.md
+++ b/docs/getting-started/concepts/registry.md
@@ -58,6 +58,7 @@ registry:
     path: postgresql://postgres:mysecretpassword@127.0.0.1:55001/feast
     cache_ttl_seconds: 60
     sqlalchemy_config_kwargs:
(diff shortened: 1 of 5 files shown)


tokoko (approved):
LGTM

HaoXuAI (approved):

PR #3999, opened 2024-03-08
fix: Use CopyFrom() instead of __deepycopy__() for creating a copy of protobuf object.

### LINKED ISSUE #3921: crash (segfault?) on protobuf>=4.24.0
## Expected Behavior 

Feast should not crash.

## Current Behavior

As commented in 028cc20a28118bd31deca8965782d5ad25f74300, Feast crashes with  protobuf>=4.24.0. The most recent version of protobuf that Feast can use is 4.23 which stopped being supported in August of 2023 <https://protobuf.dev/support/version-support/>

## Steps to reproduce

I see this on `feast apply`, the original report is unclear.

shuchu:
This PR fix the segmentation fault error while upgrading the "protobuf" version to above 4.24.0.

Based on my debug, the root cause is this line of code under the "plan()" function of "FeatureStore" class. (around line 748 of https://github.com/feast-dev/feast/blob/master/sdk/python/feast/feature_store.py )
`current_infra_proto = self._registry.proto().infra.__deepcopy__()`

The above "self._registry.proto().infra" could return an Empty object and will create a segmentation fault. 

My current solution is to create an empty object first and use the "CopyFrom()"  method from protobuf.

My test env:
python 3.8
protobuf 4.25.3

Fixes #3921

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -82,7 +82,7 @@ from feast.infra.registry.registry import Registry
 from feast.infra.registry.sql import SqlRegistry
 from feast.on_demand_feature_view import OnDemandFeatureView
 from feast.online_response import OnlineResponse
-from feast.protos.feast.core.Registry_pb2 import Registry as RegistryProto
+from feast.protos.feast.core.InfraObject_pb2 import Infra as InfraProto
 from feast.protos.feast.serving.ServingService_pb2 import (
     FieldStatus,
     GetOnlineFeaturesResponse,
@@ -746,7 +746,7 @@ class FeatureStore:
         # Compute the desired difference between the current infra, as stored in the registry,
         # and the desired infra.
         self._registry.refresh(project=self.project)
-        current_infra_proto = RegistryProto().infra
+        current_infra_proto = InfraProto()
         current_infra_proto.CopyFrom(self._registry.proto().infra)
         desired_registry_proto = desired_repo_contents.to_registry_proto()
         new_infra = self._provider.plan_infra(self.config, desired_registry_proto)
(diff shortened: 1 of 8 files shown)


sdk/python/feast/feature_store.py
```diff
@@ -745,7 +746,8 @@ def plan(
         # Compute the desired difference between the current infra, as stored in the registry,
         # and the desired infra.
         self._registry.refresh(project=self.project)
-        current_infra_proto = self._registry.proto().infra.__deepcopy__()
+        current_infra_proto = RegistryProto().infra
```
tokoko, line 749: You can use Infra proto object directly instead of getting it from RegistryProto

tokoko:
@shuchu I think you need to also update version lock files when making changes to `setup.py` by running `make lock-python-ci-dependencies` and `make lock-python-dependencies` for all versions (don't bother with 3.8, it's not tested anymore)

sdk/python/feast/feature_store.py
franciscojavierarceo: Will review this to check. Need to check some behaviors of doing that.

tokoko:
@shuchu just making sure, do you plan to update requirements files, right? I'm waiting for that to review.

shuchu:
Yes, I will update the requirements files.

sdk/python/feast/feature_store.py
franciscojavierarceo: so you have to be careful when you call `self._registry.proto()` you're referencing the `FeatureStore` class and you want to make sure that the initialization was done so it could be fetched from the database. 

All of the initialization happens within the `FeatureStore.__init__()` method. 

If this still behaves as expected using that, then I'm fine with this but it doesn't seem like it would but I only briefly looked at this. 

I knew to check this because my team found that this implementation of the registry has some bottlenecks I intend on fixing.

franciscojavierarceo: CC @tokoko @shuchu

tokoko: Looks to me it should be fine in this scenario. There's even a `self._registry.refresh()` call before. It would have failed then if called on an uninitialized `FeatureStore` object.

tokoko (approved):
LGTM

PR #4004, opened 2024-03-09
chore: Clean up dependencies in setup.py

### LINKED ISSUE #4000: Update the behavior of install-python-ci-dependencies
## Expected Behavior 

There's a very confusing and probably unnecessary python setup.py develop call in install-python-ci-dependencies make command that's causing this error. It should just be a call to only build protos instead

## Current Behavior

install-python-ci-dependencies:
	python -m piptools sync sdk/python/requirements/py$(PYTHON)-ci-requirements.txt
	COMPILE_GO=true python setup.py develop

### Specifications

- Version:
- Platform:
- Subsystem:

## Possible Solution

install-python-ci-dependencies:
	python -m piptools sync sdk/python/requirements/py$(PYTHON)-ci-requirements.txt
	pip install --no-deps -e .
	python setup.py build_python_protos --inplace

tokoko:

- removed unused dependency on `avro`
- removed unused dependency on `proto-plus` (it's still pulled in through transitive dependency on gcp libraries)
- moved `httpx` to `ci` extra as it's only used by `fastapi` testing utilities
- `docs` and `dev` extras are same as `ci`, there are no longer any reasons at this point for them to be different, problematic mysql dependency was fixed a while ago

Also fixes #4000

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,8 @@ build: protos build-java build-docker
 
 install-python-ci-dependencies:
 	python -m piptools sync sdk/python/requirements/py$(PYTHON)-ci-requirements.txt
-	COMPILE_GO=true python setup.py develop
+	pip install --no-deps -e .
+	python setup.py build_python_protos --inplace
 
 lock-python-ci-dependencies:
 	python -m piptools compile -U --extra ci --output-file sdk/python/requirements/py$(PYTHON)-ci-requirements.txt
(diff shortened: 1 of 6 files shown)


franciscojavierarceo:
@tokoko looks like you need to resolve some conflicts here

tokoko:
I'm waiting for Lokesh's PR dropping 3.8 to merge first

tokoko:
@franciscojavierarceo this is good to go now

jeremyary (approved):

PR #4014, opened 2024-03-15
fix: Azure blob storage support in Java feature server (#2319)

### LINKED ISSUE #2319: Azure blob storage support in Java feature server
## Expected Behavior 
Feast Java feature servers should be able to read from a registry in Azure blob storage.

## Current Behavior
Azure users can use the python feature server through https://github.com/Azure/feast-azure/tree/main/cluster/sdk or use an old Spark centric approach for feature serving. We should enable Java serving support as well

## Possible Solution
https://github.com/feast-dev/feast/blob/master/java/serving/src/main/java/feast/serving/config/RegistryConfig.java#L48 has AWS/GCP equivalents. We should implement the interface for Azure.

szaboferee:
 This PR is implementing the feature requested in issue: Azure blob storage support in Java feature server #2319

Changes:
- Add azure blob storage support in java feature server
- Fix S3 integration test to work without a real AWS account (to make sure it works before opening the PR)
- Add GCS mock to integration tests to be able to run them without a real google cloud account (to make sure it works before opening the PR)
- Adding dependency management in maven for libraries with older incompatible versions as transitive dependencies (0.35.0 is actually broken because of transitive dependency version conflicts)

Fixes #2319 

**How it was tested**
- All  integration tests are green
- A docker image has been built from my fork and it is running with an actual azure infrastructure

### DIFF
--- a/java/CONTRIBUTING.md
+++ b/java/CONTRIBUTING.md
@@ -50,7 +50,7 @@ Automatically format the code to conform the style guide by:
 
 ```sh
 # formats all code in the feast-java repository
-mvn spotless:apply
+make format-java
 ```
 
 > If you're using IntelliJ, you can import these [code style settings](https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml)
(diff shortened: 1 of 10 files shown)


tokoko (approved):
LGTM. sorry I had a hard time setting up dev environment. docker in docker is a nightmare 😄

franciscojavierarceo (approved):
i haven't set up the Java feature server so i'll have to test this later but i'm going to approve since it sounds like it's currently broken. thank you for doing this @szaboferee !

PR #4018, opened 2024-03-17
feat: Updating protos to separate transformation

franciscojavierarceo:

This PR updates the `OnDemandFeatureView` and `StreamingFeatureView` protobuf objectss into separate `FeatureTransformation` protobuf objects explicitly. This is a much cleaner way for machine learning engineers to understand the transformations. 

This PR will do the following:
1. Create the new `FeatureTransformation.proto` file and object
2. Deprecate `UserDefinedFunction` in the `OnDemandFeatureView.proto` and `StreamFeatureView.proto` files
3. Store the new `FeatureTransformationProto` in the `OnDemandFeatureViewSpec` and `StreamFeatureViewSpecProto` objects so that when writes are executed the new protos are stored alongside the old protos
4. Read from both the old `UserDefinedFunction` proto and the new `FeatureTransformation` proto and handle the case when  `UserDefinedFunction` is present and  `FeatureTransformation` is missing
5. Add warnings that the `UserDefinedFunction` will be deprecated
6. Update the UI to use the newly structured object

We will make this release with dual writes and the "coalesce-like" read so that the behavior is backwards compatible and also enables a safe roll out for users. 

We will deprecate the `UserDefinedFunction` before releasing to Feast `1.0.0`.

Fixes #

### DIFF
--- a/protos/feast/core/OnDemandFeatureView.proto
+++ b/protos/feast/core/OnDemandFeatureView.proto
@@ -50,8 +50,8 @@ message OnDemandFeatureViewSpec {
     map<string, OnDemandSource> sources = 4;
 
     oneof transformation {
-        UserDefinedFunction user_defined_function = 5;
-        OnDemandSubstraitTransformation on_demand_substrait_transformation = 9;
+        UserDefinedFunction user_defined_function = 5 [deprecated = true];
+        OnDemandSubstraitTransformation on_demand_substrait_transformation = 9 [deprecated = true];
     }
     // Oneof with {user_defined_function, on_demand_substrait_transformation}
     FeatureTransformationV2 feature_transformation = 10;
--- a/protos/feast/core/OnDemandFeatureView.proto
+++ b/protos/feast/core/OnDemandFeatureView.proto
@@ -27,6 +27,7 @@ import "feast/core/FeatureView.proto";
 import "feast/core/FeatureViewProjection.proto";
 import "feast/core/Feature.proto";
 import "feast/core/DataSource.proto";
+import "feast/core/Transformation.proto";
 
 message OnDemandFeatureView {
     // User-specified specifications of this feature view.
@@ -48,10 +49,8 @@ message OnDemandFeatureViewSpec {
     // Map of sources for this feature view.
     map<string, OnDemandSource> sources = 4;
 
-    oneof transformation {
-        UserDefinedFunction user_defined_function = 5;
-        OnDemandSubstraitTransformation on_demand_substrait_transformation = 9;
-    }
+    // Oneof with {user_defined_function, on_demand_substrait_transformation}
+    FeatureTransformation transformation = 5;
 
     // Description of the on demand feature view.
     string description = 6;
@@ -61,6 +60,7 @@ message OnDemandFeatureViewSpec {
 
     // Owner of the on demand feature view.
     string owner = 8;
+    string mode = 9;
 }
 
 message OnDemandFeatureViewMeta {
@@ -77,20 +77,4 @@ message OnDemandSource {
         FeatureViewProjection feature_view_projection = 3;
         DataSource request_data_source = 2;
     }
-}
-
-// Serialized representation of python function.
-message UserDefinedFunction {
-    // The function name
-    string name = 1;
-
-    // The python-syntax function body (serialized by dill)
-    bytes body = 2;
-
-    // The string representation of the udf
-    string body_text = 3;
-}
-
-message OnDemandSubstraitTransformation {
-    bytes substrait_plan = 1;
 }
--- a/protos/feast/core/OnDemandFeatureView.proto
+++ b/protos/feast/core/OnDemandFeatureView.proto
@@ -49,8 +49,12 @@ message OnDemandFeatureViewSpec {
     // Map of sources for this feature view.
     map<string, OnDemandSource> sources = 4;
 
+    oneof transformation {
+        UserDefinedFunction user_defined_function = 5;
+        OnDemandSubstraitTransformation on_demand_substrait_transformation = 9;
+    }
     // Oneof with {user_defined_function, on_demand_substrait_transformation}
-    FeatureTransformation transformation = 5;
+    FeatureTransformationV2 feature_transformation = 10;
 
     // Description of the on demand feature view.
     string description = 6;
@@ -60,7 +64,7 @@ message OnDemandFeatureViewSpec {
 
     // Owner of the on demand feature view.
     string owner = 8;
-    string mode = 9;
+    string mode = 11;
 }
 
 message OnDemandFeatureViewMeta {
@@ -77,4 +81,24 @@ message OnDemandSource {
         FeatureViewProjection feature_view_projection = 3;
         DataSource request_data_source = 2;
     }
+}
+
+// Serialized representation of python function.
+message UserDefinedFunction {
+    option deprecated = true;
+
+    // The function name
+    string name = 1;
+
+    // The python-syntax function body (serialized by dill)
+    bytes body = 2;
+
+    // The string representation of the udf
+    string body_text = 3;
+}
+
+message OnDemandSubstraitTransformation {
+    option deprecated = true;
+
+    bytes substrait_plan = 1;
 }
--- a/protos/feast/core/Transformation.proto
+++ b/protos/feast/core/Transformation.proto
@@ -19,10 +19,12 @@ message UserDefinedFunction {
     string body_text = 3;
 }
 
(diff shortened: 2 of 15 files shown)


tokoko:
@franciscojavierarceo I like the split, but this will break the existing proto layout for everyone, won't it? factoring out oneof in a separate message type, especially when there are already fields in `OnDemandFeatureView` with higher numerical positions won't be backwards-compatible. @HaoXuAI What do you think?

franciscojavierarceo:
@tokoko yes, this is a breaking change. 

The refactor is fine it's really bumping the enum value of `user_defined_function` to 1. I think it's worth the long term clarity, otherwise we'd have to start counting from 5 and have a weird comment. 

I think it's worth it but I can change the counter if you feel strongly. We can add to the release notes that this is a breaking change.

franciscojavierarceo:
Eh, it's not worth the inconvenience to users. I made the protos backwards compatible.

tokoko:
I didn't just mean the numbering. I think introducing `FeatureTransformation`-typed field simply can't be made in a backwards-compatible way no matter the numbering. if previously encoded proto message kept `UserDefinedFunction` field in position number 5, now it's part of another field in position number 17. We won't read previously saved protos correctly. Having said that, I don't have strong feelings about it, we can make a breaking change, it just might be too much of a headache for people already using it.

tokoko:
Just to understand this better, do we have a good practical justification for the breaking change though? For example, any idea why we would want to have a separate `FeatureTransformation` message, any additional fields we may want to add there in the future except for ? Or is the primary goal to reuse that message type in `StreamFeatureView`?

franciscojavierarceo:
> Just to understand this better, do we have a good practical justification for the breaking change though? 
To decouple Transformation and Serving explicitly and share that code across Stream and On Demand

>For example, any idea why we would want to have a separate FeatureTransformation message, any additional fields we may want to add there in the future except oneof? 
This PR makes an explicit `FeatureTransformation` message where `transformation` is a child object and we could add additional metadata to it in the future (e.g., the commit sha). 

>Or is the primary goal to reuse that message type in StreamFeatureView?
This is a primary goal that's a consequences of the goal of decoupling.

protos/feast/core/OnDemandFeatureView.proto
```diff
@@ -61,6 +60,7 @@ message OnDemandFeatureViewSpec {
 
     // Owner of the on demand feature view.
     string owner = 8;
+    string mode = 9;
```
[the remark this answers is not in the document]
tokoko: I remember that, but I'm a little confused. If you care about proto encoding compatibility, this change will break your protos anyway, with or without `mode` field. If all you care about is user-facing API (in other words `mode` parameter in `on_demand_feature_view` decorator) we can do that without a redundant `mode` field in the proto. `mode` parameter in `on_demand_feature_view` would simply determine which one of `oneof` fields fill be populated. Maybe I'm missing something...

tokoko:
@franciscojavierarceo thanks, `StreamFeatureView` argument sounds solid to me. P.S. Interesting if we can do this somehow without breaking `StreamFeatureView` proto encoding too?

protos/feast/core/OnDemandFeatureView.proto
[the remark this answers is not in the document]
franciscojavierarceo: That's a fair point but it does make API interface similar to stream feature view, which I prefer.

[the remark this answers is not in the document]
tokoko: I'm not disputing the `mode` field in the API. I agree we should do that as part of native python PR. All I'm saying is we don't need a `mode` field in proto for that to be possible.

franciscojavierarceo:
@etirelli I'll update the protos to make them backwards compatible and deprecate them.

protos/feast/core/OnDemandFeatureView.proto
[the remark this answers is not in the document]
HaoXuAI: i gues @tokoko means to handle `mode` in Python only. either way it works. Explicitly adding it in the proto does help understanding the Feature View schema.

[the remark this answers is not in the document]
franciscojavierarceo: >Explicitly adding it in the proto does help understanding the Feature View schema.
+1

HaoXuAI (approved):
LGTM.
I know it was a painful process. but it's great we kicked off the implementation of transformation. 👍

tokoko:
@franciscojavierarceo thanks, looks good to me as well at first glance, but pretty complicated to actually make sure that it will be backwards-compatible, so please give me a couple days to go through. P.S. I don't know how feasible it is right now but maybe we can also come up with a way to check proto compatibility against the latest version of feast on pypi.

protos/feast/core/Transformation.proto
```diff
@@ -0,0 +1,33 @@
...
+    string body_text = 3;
+}
+
+// A feature transformation executed as a user-defined function
+message FeatureTransformationV2 {
+  // Note this Transformation starts at 5 for backwards compatibility
+  oneof transformation {
+    UserDefinedFunctionV2 user_defined_function = 5;
+    OnDemandSubstraitTransformationV2 on_demand_substrait_transformation = 6;
+  }
```
tokoko, line 28: I think it's a good time to rethink the naming here. My first suggestion was to rename the field (not message type) to `on_demand_pandas_transformation` instead of `user_defined_function`. But on second thought, since we are also aiming to reuse this in `StreamFeatureViews`, I think protos should no longer be called `OnDemand..`. What do you think? I'm thinking of something like this:

```
UserDefinedFunctionV2 pandas_transformation = 1;
SubstraitTransformationV2 substrait_transformation = 2;

Comment (franciscojavierarceo) [reply protos/feast/core/Transformation.proto:28]: I was planning on doing that in a follow up PR to not add too much complexity here.

**Comment by franciscojavierarceo** [comment]
@tokoko I need to actually test the backwards compatibility so it's not quite done yet. I need to adjust the registry diff and add tests but most of the behavior is here. I'll post in the channel once ready again. Thanks for keeping an eye on it @tokoko @HaoXuAI!

**Comment by tokoko** [comment]
@franciscojavierarceo sorry about that, looks like we jumped the gun.

File: protos/feast/core/OnDemandFeatureView.proto [hunk 3]
```diff
@@ -49,9 +50,11 @@ message OnDemandFeatureViewSpec {
     map<string, OnDemandSource> sources = 4;
 
     oneof transformation {
```
tokoko, line 52: This hasn't made it to an official release yet, so can't we remove `transformation` oneof here? just leave `UserDefinedFunction user_defined_function = 5 [deprecated = true]`. It would also enable you to call the new field `transformation` which sounds less awkward (`FeatureTransformationV2 transformation = 10;`) and cut down on overall code size.

franciscojavierarceo: I'm going to do that in a separate PR as a follow up. I'll probably end up with 4 by the end. I want to leave this one as is though.

franciscojavierarceo: It's better to have separate clean PRs than one massive one imo.

franciscojavierarceo: @tokoko lmk if you're good with this. I have the next PR coming soon where I'll rename things and the subsequent will do the PythonTransformation type.

tokoko: @franciscojavierarceo yeah, sounds good to me. Let's just not make a new release until we get all these ironed out.

tokoko (approved):
Let's merge this and worry about naming in the next one. I also wonder if we could add some sort of a test for compatibility... generate protos with latest released version of feast on pypi (maybe in docker) and test if that still works. Easier said than done, though... 😄

franciscojavierarceo:
Since the proto enums didn't change it's backwards compatible, I also had a unit test do this explicitly to confirm the behavior.

sdk/python/tests/unit/test_on_demand_feature_view.py
```diff
@@ -129,3 +129,69 @@ def test_hash():
...
         udf2, "udf2 source code"
     )
+    assert (
+        on_demand_feature_view_5.feature_transformation
+        == on_demand_feature_view_5.transformation
+    )
+
+
+@pytest.mark.filterwarnings("ignore:udf and udf_string parameters are deprecated")
+def test_from_proto_backwards_compatable_udf():
```
franciscojavierarceo, line 139: @tokoko see here

Issue #3972, opened 2024-02-27
support API access to offline/online stores without Feast abstractions


jeremyary:
**Is your feature request related to a problem? Please describe.**
Feast should allow for use of just the APIs that talk to online/offline stores without having to use all the Feast constructs like data sources and feature views in order to support use cases that utilize other non-Feast tooling for data frame population.

e.g. User has pipelines that handle getting the data into data frames already and just wants an optimized way to write and read from storage without writing the connections from scratch.

**Describe the solution you'd like**
API access to offline/online without Feast abstractions

sudohainguyen:
🤔 sorry I can not capture the idea, but every storage backends have their own native client, don't they?
For example BigQuery, BigTable, Redis. HBase,...

shuchu:
I guess they want to use Feast as a feature management tool only.

jeremyary:
I should have done a better job here indicating this issue [came up in Slack](https://feastopensource.slack.com/archives/C01NTDB88QK/p1709051779419099) based on feedback solicited from social media (Reddit, specifically)

PR #4023, opened 2024-03-21
feat: Rewrite ibis point-in-time-join w/o feast abstractions

tokoko:
Ibis point-in-time join is refactored out as a separate function with no dependency on feast. The goal is to be able to unit test core logic without setting up feast objects for each test. Also introduces a brute-force pure-python implementation of point-in-time join to be used in unit tests.


HaoXuAI (approved):
LGTM!

tokoko:
@HaoXuAI thanks
@jeremyary btw, I don't know if exposing something like this is a good idea, but this can probably lead to a solution to #3972

tokoko:
@HaoXuAI Hey, can we merge here? contains requirements changes so likely to hit conflicts again 😄 thanks

PR #4025, opened 2024-03-21
chore: Deprecate PostgreSQLRegistryStore

### LINKED ISSUE #4020: Deprecate PostgreSQLRegistryStore
Right now we have 2 ways to use postgres as a registry backend. The first is with scalable `SqlRegistry` that uses `sqlalchemy`, another is an older option of using `PostgreSQLRegistryStore` which keeps the whole proto in a single table. Since we are [recommending](https://docs.feast.dev/tutorials/using-scalable-registry) the scalable registry anyway, we should deprecate `PostgreSQLRegistryStore` and remove it soon after. Or maybe remove it directly? It's under contribs as of now.

tokoko:
Fixes #4020

### DIFF
--- a/sdk/python/feast/infra/registry/contrib/postgres/postgres_registry_store.py
+++ b/sdk/python/feast/infra/registry/contrib/postgres/postgres_registry_store.py
@@ -1,3 +1,4 @@
+import warnings
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


HaoXuAI (approved):
Lgtm

PR #4029, opened 2024-03-21
feat: Add local tests for s3 registry using minio

tokoko:
S3 registry is currently impossible to test locally, as all tests depend on the presence of an aws account credentials. This PR adds local S3 registry tests using minio.


HaoXuAI:
I use to use localstack(https://www.localstack.cloud/) for local testing, what's the difference with the Minio?

tokoko:
@HaoXuAI Minio is s3-compatible object storage, not just for testing though, it's actually used out there in the wild. localstack is obviously a lot more broad, but only for testing.

franciscojavierarceo:
this lgtm @HaoXuAI any hesitation here?

HaoXuAI (approved):
LGTM

PR #4035, opened 2024-03-23
chore: Run tests as make commands in workflows

tokoko:
As some workflows are marked as `pull_request_target`, it's hard to change the pytest command to run tests in a single PR. This PR changes workflows to run tests as make commands.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,24 +64,20 @@ benchmark-python-local:
 	FEAST_USAGE=False IS_TEST=True FEAST_IS_LOCAL_TEST=True python -m pytest --integration --benchmark  --benchmark-autosave --benchmark-save-data sdk/python/tests
 
 test-python:
-	FEAST_USAGE=False \
-	IS_TEST=True \
-	python -m pytest -n 8 sdk/python/tests \
+	python -m pytest -n 8 --color=yes sdk/python/tests
 
 test-python-integration:
-	FEAST_USAGE=False IS_TEST=True python -m pytest -n 8 --integration sdk/python/tests
+	python -m pytest -n 8 --integration --color=yes --durations=5 --timeout=1200 --timeout_method=thread sdk/python/tests
 
 test-python-integration-local:
 	@(docker info > /dev/null 2>&1 && \
-		FEAST_USAGE=False \
-		IS_TEST=True \
 		FEAST_IS_LOCAL_TEST=True \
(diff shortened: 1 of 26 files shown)


Makefile
```diff
@@ -64,24 +64,20 @@ benchmark-python-local:
...
-	python -m pytest -n 8 sdk/python/tests \
+	python -m pytest -n 8 --color=yes sdk/python/tests
 
 test-python-integration:
-	FEAST_USAGE=False IS_TEST=True python -m pytest -n 8 --integration sdk/python/tests
+	python -m pytest -n 8 --integration --color=yes --durations=5 --timeout=1200 --timeout_method=thread sdk/python/tests
 
 test-python-integration-local:
 	@(docker info > /dev/null 2>&1 && \
-		FEAST_USAGE=False \
```
sudohainguyen, line 76: I believe we should keep `FEAST_USAGE` and `IS_TEST`

Makefile
```diff
@@ -64,24 +64,20 @@ benchmark-python-local:
 	FEAST_USAGE=False IS_TEST=True FEAST_IS_LOCAL_TEST=True python -m pytest --integration --benchmark  --benchmark-autosave --benchmark-save-data sdk/python/tests
 
 test-python:
```
sudohainguyen, line 66: consider renaming it to `test-python-unit` ?

sudohainguyen (changes requested):

Makefile
tokoko: I found that there's being set in [conftest](https://github.com/feast-dev/feast/blob/master/sdk/python/tests/conftest.py#L27) regardless

Makefile
tokoko: good idea. done

Makefile
tokoko: fixed, thanks

sudohainguyen (approved):
lgtm

tokoko:
@franciscojavierarceo fyi, had to make some changes to `stream_feature_view`, typecheck was complaining about `body_text` of type string being passed to `feature_transformation` parameter. Not sure how this was missed in ci previously though, something about bumping library versions, i suppose.

jeremyary:
This one's been soaking for a few days for feedback now. LGTM, I support pulling commands up and out of the action via leveraging make. Thanks!

PR #4038, opened 2024-03-24
feat: Rename OnDemandTransformations to Transformations

franciscojavierarceo:

This PR renames:
- Re-introoduces the `user_defined_function` proto object and deprecates it 
- renames 
    - `OnDemandPandasTransformation` -> `PandasTransformation`
    - `OnDemandSubstraitTransformation` -> `SubstraitTransformation`
- Reverts/blends some of the changes from https://github.com/feast-dev/feast/pull/3969 and https://github.com/feast-dev/feast/pull/3949
- Incorporates https://github.com/feast-dev/feast/pull/4018

This PR is a follow up to https://github.com/feast-dev/feast/pull/4018

Fixes #


sdk/python/feast/infra/registry/base_registry.py
```diff
@@ -662,10 +663,16 @@ def to_dict(self, project: str) -> Dict[str, List[Any]]:
             key=lambda on_demand_feature_view: on_demand_feature_view.name,
         ):
             odfv_dict = self._message_to_sorted_dict(on_demand_feature_view.to_proto())
-
-            odfv_dict["spec"]["userDefinedFunction"][
+            # We are logging a warning because the registry object may be read from a proto that is not updated
```
tokoko, line 666: This `to_dict` method really annoys me, always complicates proto changes 😄 Do we actually need it anywhere? Let me look through the project to see if it's being used. Why can't we just convert the whole proto message to json and be done with it?

franciscojavierarceo: For services that are python backed, they will likely use it. We do at Affirm.

protos/feast/core/Transformation.proto
```diff
@@ -21,13 +21,12 @@ message UserDefinedFunctionV2 {
 
 // A feature transformation executed as a user-defined function
 message FeatureTransformationV2 {
-  // Note this Transformation starts at 5 for backwards compatibility
   oneof transformation {
     UserDefinedFunctionV2 user_defined_function = 1;
```
tokoko, line 25: I know this isn't ready, but let me suggest some names. What if we call this `PythonTransformation` instead of `UserDefinedFunctionV2`. We could reuse that message type both for `pandas_transformation` and upcoming `python_transformation` fields and V2 in the naming (I think) will no longer be necessary. wdyt?

franciscojavierarceo: I tend to like stating V2 so that people understand it's a replacement for the `deprecated` proto. Are you thinking of making `PythonTransformation` an enum as well with `Pandas` and `Python` as elements? Feel free to suggest what you're thinking to make it a little more concret if you want. 

My guess is something like

```
message FeatureTransformationV2 {
    oneof PythonTransformation {
        NativePython native_python = 1;
        Pandas pandas = 2;
    }
    SubstraitTransformationV2 substrait_transformation = 3;
}
```

Or something else?

tokoko: No, that would leave the possibility of having both python and substrait fields set, so probably not the best approach. I was thinking more like this (I'll omit V2s here just for brevity).

```
message FeatureTransformation {
    oneof transformation {
        PythonTransformation pandas_transformation  = 1;
        SubstraitTransformation substrait_transformation = 2;
        PythonTransformation python_transformation  = 3;
    }
}
```
note that `pandas_transformation` and `python_transformation` fields share the message type but that's just incidental because it just so happens that they need same type of information. If in the future we see that that's no longer the case, we could introduce `PandasTransformation` message as well and the first field of transformation will become `PandasTransformation pandas_transformation  = 1;`

franciscojavierarceo: I like the UDF structure as it's a common industry pattern/convention especially for [Spark](https://docs.databricks.com/en/udf/index.html).

@HaoXuAI any thoughts?

tokoko: I'm not specifically against UDFs, but the way I like to think about it all these options are sort of udfs anyway, so calling the message just UDF without any quilifier seems redundant, if it was called `PythonUserDefinedFunction` then it would be okay. I guess what I'm saying is I'm equally okay with the trio of (`PythonTransformation`, `SubstraitTransformation`, `PandasTransformation`) and with that of (`PythonUDF`, `SubstraitUDF` and `PandasUDF`).

sdk/python/feast/on_demand_feature_view.py
```diff
@@ -68,8 +68,8 @@ class OnDemandFeatureView(BaseFeatureView):
     features: List[Field]
     source_feature_view_projections: Dict[str, FeatureViewProjection]
     source_request_sources: Dict[str, RequestSource]
-    transformation: Union[OnDemandPandasTransformation]
-    feature_transformation: Union[OnDemandPandasTransformation]
+    transformation: Union[PandasTransformation]
+    feature_transformation: Union[PandasTransformation]
```
tokoko, line 72: Sorry for repeating myself, just making sure... Do you intend to remove either `transformation` or `feature_transformation` parameters in another PR?

franciscojavierarceo: oh yes let me remove them

HaoXuAI (approved):
LGTM

HaoXuAI:
also can update the document now :)

protos/feast/core/Transformation.proto
franciscojavierarceo: @jeremyary @etirelli any opinions here? I am in favor of `user_defined_function` and the code for this PR is ready otherwise.

sdk/python/feast/diff/registry_diff.py
```diff
@@ -146,21 +146,27 @@ def diff_registry_objects(
                 continue
             elif getattr(current_spec, _field.name) != getattr(new_spec, _field.name):
                 # TODO: Delete "transformation" after we've safely deprecated it from the proto
-                if _field.name in ["transformation", "feature_transformation"]:
+                if _field.name == "feature_transformation":
                     warnings.warn(
                         "transformation will be deprecated in the future please use feature_transformation instead.",
```
tokoko, line 151: missed this warning here

sdk/python/feast/on_demand_feature_view.py
tokoko: Can you change the type to `Union[PandasTransformation, SubstraitTransfromation]`? I think I missed it in my original PR.

franciscojavierarceo: Done

sdk/python/feast/diff/registry_diff.py
franciscojavierarceo: Done

protos/feast/core/Transformation.proto
franciscojavierarceo: Lazy consensus will win here. I'm going to merge as is since everything's covered now.

PR #4041, opened 2024-03-26
feat: Isolate input-dependent calculations in `get_online_features`

tokoko:
This is for now purely a refactoring commit. This PR isolates the sort of calculations that are dependent only on feature service in `get_online_features` from the ones that are dependent on the entity values passed. The idea is to introduce a caching mechanism for these in the future and minimize the number of operations run during individual `get_online_features` calls.


franciscojavierarceo (approved):
lgtm. good idea too.

PR #4043, opened 2024-03-26
feat: Lin with ruff

### LINKED ISSUE #3956: feat: faster linting with ruff
Since we're adopting flake8 and isort which are conventional formatters, however apparently they usually take some time to finish the job.

Adopting ruff would improve significantly lint check due to its rust-backed mechanism.
Moreover, ruff offers unified settings and simplifies packages installation, all we need to focus is ruff configuration.
So that we could remove fragmented config files in our repo

Stick with current settings

dmartinol:
Replace the current linting and formatting stack with `ruff>=0.3.3` (latest version).

Collateral changes:
- Update a few source code files to adapt to the known differences between `isort` and `ruff`, see [here](https://docs.astral.sh/ruff/faq/#how-does-ruffs-import-sorting-compare-to-isort).
- Remove `sdk/python/setup.cfg` because not supported by `ruff`, and replaced by the equivalent `pyproject.toml`.
  - Move here also the configuration for `mypy` .
- Remove lint category `B' as it seems to be never raised by `flake8` (https://flake8.pycqa.org/en/latest/user/error-codes.html)
  - Instead, this category raises numerous errors with `ruff` ([flake8-bugbear](https://docs.astral.sh/ruff/rules/#flake8-bugbear-b))
- Remove lint category `C`, because the only error raised by that rule was `C901` ([here](https://flake8.pycqa.org/en/latest/user/error-codes.html), which was also added to the `ignore` list.
  - Vice versa, `ruff` has other errors in the category `C4` (see [here](https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4) that would be raised instead.
  - Note about code complexity: `flake8` and `ruff` compute different complexities for our code base, so it would be difficult to map the `max-complexity` number to a value that matches the new linting tool.
- Added ignore rule for `E721`, a new error code introduced by `flake8>=6.1.0`, the reference version for `ruff`
  -  `ruff` raises a few violations in our codebase for this error, which could be fixed in a dedicated PR
  - Final note: the list of files and lines of code affected by this error is different in `flake8 version 7.0.0` and `ruff`, meaning that `ruff` re-implemented this rule with a different behavior
- Remove lint category `T4` because never raised by `flake8` and unknown to `ruff` (`Unknown rule selector: T4`).
- Remove ignore rule for `W503` because never raised by `flake8` and unknown to `ruff` (`Unknown rule selector: W503`).
- Add rule `I` representing the `isort` category (see [here](https://docs.astral.sh/ruff/rules/#isort-i))
- Update requirements files using `pip-compile`.

Fixes #3956


sudohainguyen:
`ruff check` is not enough isn't it?
there's still `ruff format --check`

dmartinol:
> `ruff check` is not enough isn't it?

You mean instead of `python -m ruff check feast tests`? 
1- I just wanted to preserve the existing convention used in the Makefile to run modules with `python -m`
2- If agreed, we can change to `ruff check feast tests` (`ruff check` alone scans also undesired folders like docs)

> there's still ruff format --check
Isn't it better to separate the concerns and use `check` for linting and `format` only for formatting, as in `python -m ruff format feast/ tests/`?

Note: I see that the latter command tries to format more files than the `ruff check --fix`, so I think I need to investigate a bit more the formatting options that were previously implemented by `black --target-version py38 feast tests`

dmartinol:
Adding ruff formatting to replace black and the 52 files updated due to this change.
All changes seems due to the [Known Deviations from Black](https://docs.astral.sh/ruff/formatter/black/).

Makefile
```diff
@@ -317,17 +317,12 @@ test-python-universal:
...
+	cd ${ROOT_DIR}/sdk/python; python -m ruff check --fix feast/ tests/
+	cd ${ROOT_DIR}/sdk/python; python -m ruff format feast/ tests/
 
 lint-python:
-	cd ${ROOT_DIR}/sdk/python; python -m mypy feast
-	cd ${ROOT_DIR}/sdk/python; python -m isort feast/ tests/ --check-only
-	cd ${ROOT_DIR}/sdk/python; python -m flake8 feast/ tests/
-	cd ${ROOT_DIR}/sdk/python; python -m black --check feast tests
+	# cd ${ROOT_DIR}/sdk/python; python -m mypy feast
+	cd ${ROOT_DIR}/sdk/python; ruff check feast/ tests/
```
sudohainguyen, line 325: I meant this line
ruff has 2 steps to fully conduct lint check, `ruff check` and `ruff format --check`
I know it's a bit confusing but they are planning to unify these two commands

sdk/python/feast/feature_store.py
```diff
@@ -950,7 +951,9 @@ def apply(
                     validation_references.name, project=self.project, commit=False
                 )
 
-        tables_to_delete: List[FeatureView] = views_to_delete + sfvs_to_delete if not partial else []  # type: ignore
+        tables_to_delete: List[FeatureView] = (
+            views_to_delete + sfvs_to_delete if not partial else []
+        )  # type: ignore
```
dmartinol, line 956: This is change created by `ruff format` raises a type check error with `mypy`: moving the annotation one line above to fix the issue

Makefile
dmartinol: I don't think they behave the same, e.g. the error codes defined in the `select` key of the `tool.ruff.lint` configuration table are not taken into account by `ruff format --check`.

For instance, if I create `feast/my.py` as:
```python
import itertools
import copy
```
Then `ruff check` detects the error:
```console
$ python -m ruff check feast/my.py 
warning: One or more modules are part of multiple import sections, including: `feast`
feast/my.py:1:1: I001 [*] Import block is un-sorted or un-formatted
feast/my.py:1:8: F401 [*] `itertools` imported but unused
feast/my.py:2:8: F401 [*] `copy` imported but unused
```
but `ruff format --check` doesn't:
```console
$ ruff format --check feast/my.py 
warning: One or more modules are part of multiple import sections, including: `feast`
1 file already formatted
```

sudohainguyen: yes they do different jobs, `ruff check` will take care of isort stuff while `ruff format --check` can do code format similar to `black`
since I see we're replacing black in this PR so we need to handle `ruff format` as well

dmartinol: > since I see we're replacing black in this PR so we need to handle `ruff format` as well

can you look at the latest changes? it should be there in the `format-python` target, IIUC you point

shuchu:
by any chance, will you update the Document regarding the code style & lint? or that will be a separate PR. https://docs.feast.dev/project/development-guide#code-style--linting

dmartinol:
> by any chance, will you update the Document regarding the code style & lint? or that will be a separate PR. https://docs.feast.dev/project/development-guide#code-style--linting

Yes it's [here](https://github.com/feast-dev/feast/blob/14525b08dcc91d68adcb119c78961b156d53df4d/docs/project/development-guide.md?plain=1#L169)

I have some doubts about the sentence "Conforms to Black code style" because even if it's implemented by `ruff format`, the [ruff formatter](https://docs.astral.sh/ruff/formatter/) page states: 
> The Ruff formatter is an extremely fast Python code formatter designed as a drop-in replacement for [Black](https://pypi.org/project/black/), available as part of the ruff CLI via ruff format.

jeremyary:
@sudohainguyen @shuchu @tokoko any further thoughts to share on the topic here? Just hoping to see if there's more for us to tackle next week. Thanks!

Makefile
sudohainguyen: yeah sorry didn't declare clearly, I meant [this](https://arc.net/l/quote/ywotfszk)

running `ruff format --check` will help checking format without touching the code, I believe we should add this to `lint-python` as well

dmartinol: let me add it then, I misunderstood

pyproject.toml
```diff
@@ -5,27 +5,24 @@ build-backend = "setuptools.build_meta"
...
-    | pb2.py
-    | \.pyi
-    | protos
-    | sdk/python/feast/embedded_go/lib
-  )/
-)
-'''
+target-version = "py39"
+include = ["*.py", "*.pyi"]
+[tool.ruff.format]
```
sudohainguyen, line 12: (optional) add space to separate sessions

pyproject.toml
```diff
@@ -5,27 +5,24 @@ build-backend = "setuptools.build_meta"
...
-    | build
-    | dist
-    | pb2.py
-    | \.pyi
-    | protos
-    | sdk/python/feast/embedded_go/lib
-  )/
-)
-'''
+target-version = "py39"
```
sudohainguyen, line 10: (optional) should we target py310 in prior?

setup.cfg
```diff
@@ -1,25 +1,2 @@
...
-max-complexity = 20
-select = B,C,E,F,W,T4
-exclude = .git,__pycache__,docs/conf.py,dist,feast/protos,feast/embedded_go/lib
-
-[mypy]
-files=feast,tests
-ignore_missing_imports=true
-exclude=feast/embedded_go/lib
-
 [bdist_wheel]
```
sudohainguyen, line 1: (optional) can we somehow move this config to pyproject.toml? so that we can eliminate the `setup.cfg` file

pyproject.toml
tokoko: What would be the difference? doc seems to indicate that this should be a minimum version, which is py39 now

dmartinol: > (optional) should we target py310 in prior?

I can do it: I've just tried and no further changes were identified. I'm just worried about having any deprecated syntax or features that were instead allowed in py39. Is it really worth anticipating now?

sudohainguyen: yeah it's fine, can keep now

setup.cfg
dmartinol: I don't feel confident to change this one, even if I agree with your request to minimize the configuration files.
The `wheels` docs only mention the `setup.cfg` as the configuration option, so I'm not sure which of the different alternatives Google offers to trust and how to validate them later.
Can we have a dedicated ticket instead?

dmartinol: @sudohainguyen rather, I would move the formatting configuration from `/pyproject.tom`l to `/sdk/python/pyproject.toml`, that seems its natural location, looking at the `Makefile`. WDYT?

sudohainguyen: I prefer keeping as where it is now. I believe in the future we will eventually deprecate `setup.py` to use `pyproject.toml` for publishing, according to [PEP 621](https://peps.python.org/pep-0621/)
So I don't wanna mess up that flow 😄 

back to `setup.cfg`, let's leave it there for the further PR

sudohainguyen (approved):
ltgm

tokoko (approved):
lgtm

PR #4044, opened 2024-03-27
fix: Remove proto-plus imports

tokoko:
#4004 removed `proto-plus` from required dependencies but missed some dangling imports that were still part of the codebase.


alex-vinnik-sp (approved):
Looking forward to have it merged. Thanks for the quick turnaround.

jeremyary (approved):

PR #4045, opened 2024-03-27
feat: Adding support for Native Python feature transformations for ODFVs

franciscojavierarceo:

This pull request adds support for native Python transformation. Historically, On Demand Feature Views only supported `pandas`. We introduce support for `python` in this Pull Request.

The resulting behavior should look something like this:

```
from feast import Field, RequestSource
from feast.types import Float64, Int64
import pandas as pd

# Define a request data source which encodes features / information only
# available at request time (e.g. part of the user initiated HTTP request)
input_request = RequestSource(
    name="vals_to_add",
    schema=[
        Field(name='val_to_add', dtype=Int64),
        Field(name='val_to_add_2', dtype=Int64)
    ]
)

# Use the input data and feature view features to create new features
@on_demand_feature_view(
   sources=[
       driver_hourly_stats_view,
       input_request
   ],
   schema=[
     Field(name='conv_rate_plus_val1', dtype=Float64),
     Field(name='conv_rate_plus_val2', dtype=Float64)
   ],
   mode="python"
)
def transformed_conv_rate(features: Dict[str, Any]) -> pd.DataFrame:
    output = {}
    output['conv_rate_plus_val1'] = 2 + features['conv_rate'] + features['val_to_add']
    output['conv_rate_plus_val2'] = 2 + features['conv_rate'] + features['val_to_add_2']
    return output
```
Note this has nice performance gains when serving particularly for single rows.

Fixes #


sdk/python/feast/feature_store.py
```diff
@@ -2096,26 +2096,53 @@ def _augment_response_with_on_demand_transforms(
...
+                    initial_response_dict = initial_response.to_dict()
+                transformed_features_dict: Dict[
+                    str, List[Any]
+                ] = odfv.get_transformed_features(
+                    initial_response_dict,
+                    full_feature_names,
                 )
-                for feature in selected_subset
-            ]
+            elif odfv.mode == "pandas":
```
tokoko, line 2115: This should be `in {"pandas", "substrait"}`. Their `transform` methods have identical signatures.

sdk/python/feast/transformation/python_transformation.py
```diff
@@ -0,0 +1,64 @@
...
+        Creates an PythonTransformation object.
+        Args:
+            udf: The user defined transformation function, which must take pandas
+                dataframes as inputs.
+            udf_string: The source code version of the udf (for diffing and displaying in Web UI)
+        """
+        self.udf = udf
+        self.udf_string = udf_string
+
+    def transform(self, input_dict: Dict) -> Dict:
```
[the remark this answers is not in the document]
franciscojavierarceo: return should be `Dict[str, List[Any]]`

sdk/python/feast/online_response.py
```diff
@@ -77,3 +77,46 @@ def to_df(self, include_event_timestamps: bool = False) -> pd.DataFrame:
         """
 
         return pd.DataFrame(self.to_dict(include_event_timestamps))
+
+
+class OnlineResponseRow(OnlineResponse):
```
tokoko, line 82: Is this used anywhere? I couldn't find any usage.

sdk/python/feast/feature_store.py
```diff
@@ -2096,26 +2096,55 @@ def _augment_response_with_on_demand_transforms(
...
-            ]
+            elif odfv.mode in {"pandas", "substrait"}:
+                if initial_response_df is None:
+                    initial_response_df = initial_response.to_df()
+                print("about to happen\n", initial_response_df)
+                transformed_features_df: pd.DataFrame = odfv.get_transformed_features(
+                    initial_response_df,
+                    full_feature_names,
+                )
+                print("it did not happen")
```
tokoko, line 2123: missed some debug print statements here

sdk/python/feast/online_response.py
franciscojavierarceo: Oh nice catch, I accidentally copied this over. Will delete.

sdk/python/feast/feature_store.py
franciscojavierarceo: Nice catch thank you.

sdk/python/feast/feature_store.py
```diff
@@ -2096,26 +2096,53 @@ def _augment_response_with_on_demand_transforms(
...
+            transformed_columns = (
+                transformed_features.columns
+                if isinstance(transformed_features, pd.DataFrame)
+                else transformed_features
+            )
+            selected_subset = [f for f in transformed_columns if f in _feature_refs]
+
+            proto_values = []
+            for selected_feature in selected_subset:
+                if odfv.mode in ["python", "pandas"]:
```
tokoko, line 2141: this kinda looks like it's breaking substrait, but I guess it's fine. seems like online path isn't tested for substrait in ci, I'll add tests and a fix (if necessary) for this later.

sdk/python/feast/transformation/python_transformation.py
```diff
@@ -0,0 +1,65 @@
...
+        if not isinstance(input_dict, Dict):
+            raise TypeError(
+                f"input_dict should be type Dict[str, List[Any]] but got {type(input_dict).__name__}"
+            )
+        # Ensuring that the inputs are included as well
+        output_dict = {**input_dict, **self.udf.__call__(input_dict)}
+        if not isinstance(output_dict, Dict):
```
tokoko, line 30: makes more sense to type check udf.__call__ output instead of `output_dict`

tokoko (approved):
LGTM

sdk/python/tests/unit/test_on_demand_feature_view.py
```diff
@@ -128,6 +143,75 @@ def test_hash():
...
+    with pytest.raises(TypeError):
+        # This should fail
+        on_demand_feature_view_python_native_err.feature_transformation.transform(
+            {
+                "feature1": 0,
+                "feature2": 1,
+            }
+        )
+
+    assert on_demand_feature_view_python_native.get_transformed_features(
```
franciscojavierarceo, line 207: @HaoXuAI see here as example of usage.

HaoXuAI (approved):
LGTM！

sdk/python/tests/unit/test_on_demand_feature_view.py
```diff
@@ -32,10 +38,18 @@ def udf1(features_df: pd.DataFrame) -> pd.DataFrame:
 def udf2(features_df: pd.DataFrame) -> pd.DataFrame:
     df = pd.DataFrame()
     df["output1"] = features_df["feature1"] + 100
-    df["output2"] = features_df["feature2"] + 100
+    df["output2"] = features_df["feature2"] + 101
     return df
 
 
+def python_native_udf(features_dict: Dict[str, List[Any]]) -> Dict[str, Any]:
```
tokoko, line 45: @franciscojavierarceo sorry for the late find. Can you look at this test examples? The signatures are correct, but actual transformation seems to assume that `features_dict["feature1"]` is an int, but it should be a list of ints, right?

franciscojavierarceo: whoops let me update the test case tomorrow but it should be a singleton, the type hint is wrong. I adjusted this late and missed this one.

tokoko: ah, got it. If it should be a singleton, then also double-check what is actually passed from odfv class as well. I may be wrong, but I think it's passing a list.

franciscojavierarceo: I'll add another test when using `get_online_features` for an ODFV to confirm it more explicitly. I do have a test below confirming the correct behavior for `self.feature_transformation.transform()` returning the expected result.

tokoko: I'm planning to add an explicit `get_online_features` test for `substrait` as well. Just testing `transform` is clearly not enough.

franciscojavierarceo: here's the fix for the type: https://github.com/feast-dev/feast/pull/4054

PR #4046, opened 2024-03-27
feat: Feature/upgrade 3.11 and remove bytewax

lokeshrangineni:

Fixes #


lokeshrangineni:
I am looking at the PR again to modify the the logic without actually removing the bytewax engine. If this approach works then probably i will update the PR otherwise we can go ahead with this approach.

.github/workflows/unit_tests.yml
```diff
@@ -7,8 +7,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ "3.9", "3.10" ]
-        os: [ ubuntu-latest, macos-12 ]
+        python-version: [ "3.9", "3.10", "3.11"]
+        os: [ ubuntu-latest, macOS-12 ]
```
jeremyary, line 11: SUPER nitpicky of me, but casing here is fine in lowercase which is how GitHub references them too

sdk/python/requirements/py3.9-requirements.txt
```diff
@@ -144,10 +144,12 @@ six==1.16.0
     # via python-dateutil
 sniffio==1.3.1
     # via anyio
-sqlalchemy[mypy]==2.0.29
+sqlalchemy[mypy]==1.4.52
```
jeremyary, line 147: not saying it's wrong, more curious, can you please share what's causing the downgrade of sqlalchemy?

setup.py
```diff
@@ -65,15 +65,15 @@
     "fastapi>=0.68.0",
     "uvicorn[standard]>=0.14.0,<1",
     "gunicorn; platform_system != 'Windows'",
-    "dask[dataframe]>=2021.1.0",
+    "dask[dataframe]>=2024.4.2",
```
jeremyary, line 68: I feel like I saw an answer in slack somewhere that I can't find at the moment, but this is a pretty big version jump, was Dask having issues with new Python version?

setup.py
```diff
@@ -65,15 +65,15 @@
...
     "bowler",  # Needed for automatic repo upgrades
 ]
 
 GCP_REQUIRED = [
     "google-api-core>=1.23.0,<3",
     "googleapis-common-protos>=1.52.0,<2",
     "google-cloud-bigquery[pandas]>=2,<3.13.0",
-    "google-cloud-bigquery-storage >= 2.0.0,<3",
+    "google-cloud-bigquery-storage >= 2.23.0,<3",
```
jeremyary, line 76: Did we encounter an issue w/BigQuery dependency too?

setup.py
lokeshrangineni: We had a issue with Dask with new python version. Build is failing so had to upgrade but i have tried the immediate version so let me give a shot with the version released in 2022.

sdk/python/requirements/py3.9-requirements.txt
lokeshrangineni: I have just re-ran the command to regenerate this file but not sure why this was modified. I will pay attention to it.

setup.py
jeremyary: you're good! I was just looking for reason - since we need to upgrade we might as well catch up

jeremyary:
work continuing [here](https://github.com/feast-dev/feast/pull/4159) due to conflict complexity

PR #4050, opened 2024-03-28
chore: Upgrade dask dependency

tokoko:
This PR updates `dask` dependency and adds `dataframe` extra to dask. dask warnings should go away at last.


sdk/python/feast/infra/offline_stores/contrib/mssql_offline_store/mssql.py
```diff
@@ -423,7 +423,7 @@ def _upload_entity_df_into_sqlserver_and_get_entity_schema(
 
     elif isinstance(entity_df, pandas.DataFrame):
         # Drop the index so that we don't have unnecessary columns
-        engine.execute(_df_to_create_table_sql(entity_df, table_id))
+        engine.execute(_df_to_create_table_sql(entity_df, table_id))  # type: ignore
```
HaoXuAI, line 426: Why suddenly adding the type ignore?

tokoko: not exactly sure tbh, but probably something related to this #4066. We broke mssql with sqlalchemy upgrade, it passed the ci, because we are not testing contribs in the ci. in other words, it's already broken with or without the type check.

franciscojavierarceo (approved):

PR #4053, opened 2024-03-29
fix: Remove or empty string statement in to_proto() definition

### LINKED ISSUE #4030: Bug: UserWarning when passing `FeatureService` object to both `apply()` and `get_online_features()` - Caused by difference in `FeatureViewProjection.from_proto()` and `FeatureViewProjection.from_definition()`
## Context

If a `FeatureService` object is created and is being passed to both the `apply()` and the `get_online_features()` method, the following user warning is thrown: 

```
UserWarning: The FeatureService object that has been passed in as an argument is inconsistent with the version from the registry. Potentially a newer version of the FeatureService has been applied to the registry.
```

This is caused by a bug in the creation/retrieval of `feature_view_projections`, which is an attribute of `FeatureService`. An empty string is set to the `name_alias` value of `FeatureViewProjection` when calling [`from_proto`](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/feature_view_projection.py#L52-L64). However, when creating the `FeatureViewProjection` by creating a `FeatureService` object (using the default value) `name_alias` is set to `None`, because it is created with [`FeatureViewProjection.from_definition()`](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/feature_view_projection.py#L66-L73).

The warning is raised [here](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/feature_store.py#L532-L541), because the `feature_service_from_registry` has the empty string value for the underlying `FeatureViewProjection.name_alias` object, while the for `_features` the underlying `FeatureViewProjection.name_alias` is valued `None`.

## Expected Behavior

I would expect that if a `FeatureViewProjection` encapsulated within a `FeatureService` has a `None` value for the `name_alias` attribute when being stored in the feature store, that it would also load with the same `None` value when calling `FeatureViewProjection.from_proto()`.

Then the comparison should also not fail, and the warning should not be thrown.

## Current Behavior

The `FeatureViewProjection` is loaded with an empty string for the value `name_alias` when `FeatureViewProjection.from_proto()` is called. 

This causes the. comparison to fail, and the warning to be thrown. 

## Steps to reproduce

docker-compose.yml
```
---
version: "3"
services:
  db:
    restart: always
    image: postgres:16-alpine
    container_name: feast_db
    ports:
      - "5432:5432"
    environment:
      - POSTGRES_DB=feature_store
      - POSTGRES_USER=postgres
      - POSTGRES_PASSWORD=test
```

feature_store.yaml
```
project: project_name
provider: local
registry:
    registry_type: sql
    path: postgresql://postgres:test@0.0.0.0:5432/feature_store
```

Python script
```

import pandas as pd
from feast import (
    Entity,
    FeatureService,
    FeatureStore,
    FeatureView,
    Field,
    SnowflakeSource,
    ValueType,
)

feature_store = FeatureStore()

# Apply
entity = Entity(
    name="entity",
    join_keys=["ID"],
    value_type=ValueType.STRING,
)

source = SnowflakeSource(
    name="source_snowflake",
    timestamp_field="EVENT_TIMESTAMP",
    schema="TEMP",
    table="FEAST_FEATURES",
)

feature_view = FeatureView(
    name="feature_view__v1",
    entities=[entity],
    ttl=timedelta(days=0),
    schema=[
        Field(name="FEATURE", dtype=Float32),
    ],
    online=True,
    source=source,
)

    name="feature_service__v1",
    features=[feature_view],
)

feature_store.apply(
    [
        entity,
        source,
        feature_view,
        feature_service,
    ]
)

# Inference
entity_rows = [{"ID": "ID1"}, {"ID": "ID2"}]
entity_df = pd.DataFrame(entity_rows)
entity_df["event_timestamp"] = pd.to_datetime("now", utc=True)

online_features = feature_store.get_online_features(
    entity_rows=entity_rows,
    features=feature_service,
).to_dict()

print(online_features)
```

### Specifications

Using postgres registry. 

- Version: 0.36.0
- Platform: MacOS - M1
- Subsystem: Sonoma 14.1.1

## Possible Solution

I have little to no experience with proto, so I'm not sure whether it would be possible to load the None value directly from the proto definition.

One potential solution would be to check whether an empty string is loaded for this field when loading from proto, and then set it to None in the `from_proto` function.

job-almekinders:
Fixes #4030

### DIFF
--- a/sdk/python/feast/feature_view_projection.py
+++ b/sdk/python/feast/feature_view_projection.py
@@ -41,7 +41,7 @@ class FeatureViewProjection:
     def to_proto(self) -> FeatureViewProjectionProto:
         feature_reference_proto = FeatureViewProjectionProto(
             feature_view_name=self.name,
-            feature_view_name_alias=self.name_alias or "",
+            feature_view_name_alias=self.name_alias or None,
             join_key_map=self.join_key_map,
         )
         for feature in self.features:


jeremyary:
Thanks @job-almekinders, I'll enable the rest of the checks. Here's the current lint error copied out for convenience:
```
feast/feature_view_projection.py:44: error: Argument "feature_view_name_alias" to "FeatureViewProjection" has incompatible type "Optional[str]"; expected "str"  [arg-type]
```

sdk/python/feast/feature_view_projection.py
```diff
@@ -41,7 +41,7 @@ def name_to_use(self):
     def to_proto(self) -> FeatureViewProjectionProto:
         feature_reference_proto = FeatureViewProjectionProto(
             feature_view_name=self.name,
-            feature_view_name_alias=self.name_alias or "",
+            feature_view_name_alias=self.name_alias or None,
```
tokoko, line 44: I guess we have our answer now 😄 this was probably due to mypy checks and also because protos don't really distinguish between null and empty string anyway. Maybe try changing `from_proto` method instead to populate the object with `None` even if `name_alias` is set to `""`, empty string for a name doesn't make sense anyway.

franciscojavierarceo:
Looks like the linter is failing here.

tokoko:
@franciscojavierarceo this is for the same issue as the one that you just merged. you can close this.

PR #4059, opened 2024-03-30
feat: Pull duckdb from contribs, add to CI

tokoko:
This PR pulls duckdb and ibis from contrib folder and adds duckdb to default integration tests. All tests are now passing.


tokoko:
@HaoXuAI can you take a look here when you find the time? it's mostly just moving stuff around and a minor fix for a single test (namely ibis now uses `created_timestamp_field` to drop older records from feature table if records with duplicate event timestamps are present in the source). thanks a lot

HaoXuAI:
@tokoko Just get back. looks good to me 👍

PR #4060, opened 2024-03-31
feat: Feast/IKV online store contrib plugin integration

pushkarmoi:
We add a new online store (IKV - https://www.inlined.io/) to feast by following the guide here:
docs.feast.dev/how-to-guides/customizing-feast/adding-support-for-a-new-online-store


franciscojavierarceo (approved):
🚀🚀🚀

setup.py
```diff
@@ -209,6 +213,7 @@
     + CASSANDRA_REQUIRED
     + AZURE_REQUIRED
     + ROCKSET_REQUIRED
+    + IKV_REQUIRED
```
tokoko, line 216: This is not necessary, since we are not adding tests to run on ci at this point

pushkarmoi: thanks for flagging, removed

pushkarmoi:
Closed this and opened https://github.com/feast-dev/feast/pull/4068 to work-around DCO issue

Issue #3965, opened 2024-02-22
Mlops architecture to reconcile feast and vector search in the serving layer


boumelhaa:
This is not an issue. We are currently working on developing a scalable architecture for our ranking system using Feast. As a backend, we are utilizing GCP for the offline store and Redis for the online store in AWS, which is in close proximity to our serving environment.

Feast effectively abstracts the feature vectors for classical models or batch inference. However, the complexity arises when we integrate embeddings into our recommendation system.

While Feast serves well for training the embeddings model and encoding the embeddings in offline batches, the challenge lies in serving these embeddings. My question pertains to how a vector search solution fits into our architecture. Where should the embeddings reside, and do we need to register them initially?

In essence, considering our two-tower ranking model where the first tower's embeddings are encoded offline and the second tower's embeddings are encoded upon request, followed by a search against the pre-encoded embeddings, how can we structure this using Feast?

HaoXuAI:
This is a feature we are planning to add to feast. On a in-mature thought we can add an API that can both index and retrieve the embedding. probably can use Faiss as a stab. Will initiate a RFC for the work.

boumelhaa:
Yeah, Faiss seems like a promising candidate, although the in-memory aspect can pose scalability and re-indexing challenges. 
Integrating it with Elasticsearch would enhance its appeal, allowing for the registration of embeddings and direct materialization into ES or any other search API, offering an alternative to traditional online store technologies.

HaoXuAI:
@boumelhaa for the online store, are you using the elasticsearch?

tokoko:
Don't really know much about the subject, but feels like people might want to use different technologies for normal feature lookup and vector search. Is it the right time to also start thinking about (better) supporting configuration of multiple online stores in the same feast project?

HaoXuAI:
> Don't really know much about the subject, but feels like people might want to use different technologies for normal feature lookup and vector search. Is it the right time to also start thinking about (better) supporting configuration of multiple online stores in the same feast project?

Not sure how that would help the search use case but yeah definitely a good feature to have.

franciscojavierarceo:
This is a great topic and something I was actually quite excited about supporting. Glad to see @HaoXuAI already on it! 🚀

boumelhaa:
> @boumelhaa for the online store with serach functionality, are you using the elasticsearch?

[...]

PR #4061, opened 2024-03-31
feat: Enable Vector database and retrieve_online_documents API

HaoXuAI:
RFC: https://docs.google.com/document/d/18IWzLEA9i2lDWnbfbwXnMCg3StlqaLVI-uRpQjr_Vos/edit#heading=h.9gaqqtox9jg6

Fixes #

https://github.com/feast-dev/feast/issues/3965


franciscojavierarceo (changes requested):
Overall this looks great, I left some comments in the doc and think the only thing that's worth doing is us aligning on some of the naming conventions to be more similar to the industry and to what Feast is already doing.

Great work and can't wait to get this out! 🚀

sdk/python/feast/feature_store.py
```diff
@@ -1690,6 +1690,67 @@ def _get_online_features(
...
 
+    @log_exceptions_and_usage
+    def retrieve_online_documents(
+        self,
+        feature: str,
+        query: Union[str, List[float]],
+        top_k: int,
+    ) -> OnlineResponse:
+        """
+        Retrieves the top k closest document features.
```
franciscojavierarceo, line 1701:
```suggestion
        Retrieves the top k closest document features. Note, embeddings are a subset of features.
```

sdk/python/feast/feature_store.py
```diff
@@ -1690,6 +1690,67 @@ def _get_online_features(
...
+
+        Args:
+            feature: The list of document features that should be retrieved from the online document store. These features can be
+                specified either as a list of string document feature references or as a feature service. String feature
+                references must have format "feature_view:feature", e.g, "document_fv:document_embedding_feature".
```
franciscojavierarceo, line 1706:
```suggestion
                references must have format "feature_view:feature", e.g, "document_fv:document_embeddings".
```

sdk/python/feast/feature_store.py
```diff
@@ -1690,6 +1690,67 @@ def _get_online_features(
         )
         return OnlineResponse(online_features_response)
 
+    @log_exceptions_and_usage
+    def retrieve_online_documents(
```
franciscojavierarceo, line 1694: There's probably something to be said about having a configurable distance metric to let the user choose which way to get the `top_k`

sdk/python/feast/infra/online_stores/contrib/postgres.py
```diff
@@ -251,6 +260,49 @@ def teardown(
...
+            requested_feature: The requested feature as the column to search
+            embedding: The query embedding to search for
+            top_k: The number of items to return
+        Returns:
+            List of tuples containing the event timestamp and the document feature
+
+        """
+
+        # Convert the embedding to a string to be used in postgres vector search
+        query_embedding_str = f"'[{','.join(str(el) for el in embedding)}]'"
```
franciscojavierarceo, line 285: Is this the best serialization we can do? This feels pretty brittle but I get it.

sdk/python/feast/infra/passthrough_provider.py
```diff
@@ -47,6 +47,7 @@ def __init__(self, config: RepoConfig):
         self.repo_config = config
         self._offline_store = None
         self._online_store = None
+        self._document_store = None
```
tokoko, line 50: This is no longer necessary

sdk/python/feast/infra/online_stores/contrib/postgres.py
```diff
@@ -251,6 +260,49 @@ def teardown(
...
             raise
 
+    def retrieve_online_documents(
+        self,
+        config: RepoConfig,
+        table: FeatureView,
+        requested_feature: str,
+        embedding: List[float],
+        top_k: int,
+    ) -> List[Tuple[Optional[datetime], Optional[Dict[str, ValueProto]]]]:
```
tokoko, line 270: Do we need `Optional[Dict[str, ValueProto]]`? If only a single feature can be searched, wouldn't `Optional[ValueProto]` be sufficient?

sdk/python/tests/integration/online_store/test_universal_online.py
```diff
@@ -785,3 +785,26 @@ def assert_feature_service_entity_mapping_correctness(
                 entity_rows=entity_rows,
                 full_feature_names=full_feature_names,
             )
+
+
+@pytest.mark.integration
+@pytest.mark.universal_online_stores(only=["postgres"])
+def test_retrieve_online_documents(
```
franciscojavierarceo, line 792: Will you be outputting the cosine similarity as well? That would be useful possibly for debugging. Would be good to be able to test that the engine computes it...maybe not doable though.

HaoXuAI: Should be possible. Somehow just the integration test doesn't startup the Postgres container. And I'm debugging it.

franciscojavierarceo (approved):
lgtm

HaoXuAI:
> lgtm

thanks!

HaoXuAI:
@franciscojavierarceo @tokoko fyi this is the result of the API. there are still some TODOs I've added into the code directly. Will add to the document as well. Also feels like having a layer to abstract the `vector` online store might help to control the input and output flow.  Anyway we have a working solution now.

Screenshot 2024-04-14 at 3 24 34 PM
Screenshot 2024-04-14 at 3 24 30 PM

sdk/python/feast/infra/passthrough_provider.py
boumelhaa: Thanks for addressing the issue. One question, though: what if we want to continue using Redis or any other online store for usual features, and use PG vector solely for embedding and search? Do we have the option to use the online store and the document store in the feature_store.yaml, both?

franciscojavierarceo: That could get complicated but agreed it'd be good to add. I could imagine a Redis + another DB layer would be an obvious one.

boumelhaa: Thank you again, guys, for these amazing features. Yes, having multiple online stores will make it easier to use the right database layer for the appropriate use case! 🙌

Makefile
```diff
@@ -200,7 +200,7 @@ test-python-universal-postgres-offline:
 test-python-universal-postgres-online:
 	PYTHONPATH='.' \
 		FULL_REPO_CONFIGS_MODULE=sdk.python.feast.infra.online_stores.contrib.postgres_repo_configuration \
-		PYTEST_PLUGINS=sdk.python.feast.infra.offline_stores.contrib.postgres_offline_store.tests \
```
franciscojavierarceo, line 203: Looking at this now, was this the right choice?

PR #4064, opened 2024-04-02
fix: Substrait ODFVs for online

tokoko:
Fixes `substrait` odfvs for online processing and adds necessary unit tests.

### DIFF
--- a/sdk/python/tests/unit/test_substrait_transformation.py
+++ b/sdk/python/tests/unit/test_substrait_transformation.py
@@ -84,6 +84,8 @@ def test_ibis_pandas_parity():
             [driver, driver_stats_source, driver_stats_fv, substrait_view, pandas_view]
         )
 
+        store.materialize(start_date=datetime(2000, 4, 12, 10, 59, 42), end_date=datetime(3000, 4, 12, 10, 59, 42))
+
         entity_df = pd.DataFrame.from_dict(
             {
                 # entity's join key -> entity values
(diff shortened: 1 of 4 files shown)


sdk/python/tests/unit/test_substrait_transformation.py
```diff
@@ -84,6 +84,11 @@ def substrait_view(inputs: Table) -> Table:
             [driver, driver_stats_source, driver_stats_fv, substrait_view, pandas_view]
         )
 
+        store.materialize(
+            start_date=datetime(2000, 4, 12, 10, 59, 42),
+            end_date=datetime(3000, 4, 12, 10, 59, 42),
```
franciscojavierarceo, line 89: probably better to do make these relate dates using datetime.now()

franciscojavierarceo (commented):
very nice one nit just in case this code stays around for a thousand years 😂

sdk/python/tests/unit/test_substrait_transformation.py
tokoko: 😄 actually that's a good point. thanks. changed all the dates relative to `start_date` and `end_date` variables.

franciscojavierarceo: nice, looks like java integration tests failed though

tokoko:
Looks like it was a flaky test, some connection error setting up JDK. forced a rerun and all tests are passing now.

franciscojavierarceo (approved):
nice

PR #4069, opened 2024-04-03
chore: Adding more tests for On Demand Feature Views

franciscojavierarceo:

This PR adds explicit tests for `get_online_features` for the on demand feature views for Python and Pandas transformations. It also fixes a bug when serializing the transformation into a protobuf object.

Fixes bug in serializing Python Native transformation object into Pandas Transformation.

# Fixes #


sdk/python/tests/data/driver_hourly_stats.parquet
tokoko, line 1: Is this file necessary? looks like it should be auto-generated during tests in a temp folder

franciscojavierarceo: oh good catch I didn't mean to check this in

tokoko: binary files in a tests folder are scary these days lol 😄

sdk/python/tests/unit/test_on_demand_python_transformation.py
```diff
@@ -0,0 +1,139 @@
...
+                    inputs["conv_rate"] + inputs["acc_rate"]
+                )
+                return df
+
+            @on_demand_feature_view(
+                sources=[driver_stats_fv[["conv_rate", "acc_rate"]]],
+                schema=[Field(name="conv_rate_plus_acc_python", dtype=Float64)],
+                mode="python",
+            )
+            def python_view(inputs: Dict[str, Any]) -> Dict[str, Any]:
```
tokoko, line 84: I'm still a little confused about the required signature here. Are these functions supposed to accept a dict of lists (looks like that in this test) and apply the udf for all entities at once? I thought from the previous PR that the goal was to have a udf that would be applied to individual entities...

tokoko: Can you also alter the tests so that more than one entity is passed? this will probably fail in such a case as only first entity is processed. If we are sticking with this signature, udf should look something like this:
```
return {
                'conv_rate_plus_acc_python': [
                    conv_rate + acc_rate
                    for conv_rate, acc_rate in zip(inputs['conv_rate'], inputs['acc_rate'])
                ]
            }

franciscojavierarceo: If you look at `_infer_features_dict` you'll see it expects a dict of lists. I added an explicit test that shows this will result in a type failure when running the apply operations. We can add singleton execution as a follow up but this is sufficient to highlight the currently supported behavior and then we can cut a release.

tokoko: @franciscojavierarceo got it, good... that's probably more efficient anyway. no rush, but in that case it will probably be a good idea to change type annotations for relevant functions to `Dict[str, List[Any]]`.

franciscojavierarceo: I actually originally had that setup but I received a ton of type failures from that which is why I did it this way. 

Let me address both of those as folllowups. I want to merge this and cut a release.

tokoko (approved):
LGTM

jeremyary (approved):

sdk/python/tests/unit/test_on_demand_python_transformation.py
franciscojavierarceo: I made an issue here https://github.com/feast-dev/feast/issues/4075, will close it later.

PR #4071, opened 2024-04-03
fix: Fix type mapping spark

ElliotNguyen68:
**What this PR does / why we need it**: Currently when inference datatype from a spark source, feast not able to infer column of type decimal, this pr aim to turn decimal type column in spark to double feast data type

Fixes #


sudohainguyen:
I prefer explicit type name instead of using regex 🤔 
how many we need to do?

ElliotNguyen68:
> I prefer explicit type name instead of using regex 🤔 how many we need to do?

https://spark.apache.org/docs/3.1.3/api/python/reference/api/pyspark.sql.types.DecimalType.html, I think the number is so many

sdk/python/feast/type_map.py
```diff
@@ -774,6 +776,10 @@ def spark_to_feast_value_type(spark_type_as_str: str) -> ValueType:
         "array<boolean>": ValueType.BOOL_LIST,
         "array<timestamp>": ValueType.UNIX_TIMESTAMP_LIST,
     }
+    decimal_regex_pattern = r"^decimal\([0-9]{1,2},[0-9]{1,2}\)$"
```
tokoko, line 779: startswith("decimal") is probably more appropriate for now if we don't care about the precision.

ElliotNguyen68:
@tokoko , already change

tokoko:
@ElliotNguyen68 Can you add some handling for array of decimals as well while we are at it? just to keep things symmetrical

ElliotNguyen68:
> @ElliotNguyen68 Can you add some handling for array of decimals as well while we are at it? just to keep things symmetrical

Hi @tokoko , i dont get what you mean ?

tokoko:
@ElliotNguyen68 `type_map` contains mapping for spark data types and appropriate array types (`int` and `array<int>` for example). This change would only add support for `decimal` and leave `array<decimal>` unsupported. Maybe add another step for `spark_type_as_str.startswith("array<decimal")`?

tokoko (approved):
lgtm

ElliotNguyen68:
@HaoXuAI  hope you can merge the pr

sudohainguyen (approved):
lgtm

PR #4085, opened 2024-04-09
fix: Change numpy version <1.25 dependency to <2 in setup.py

### LINKED ISSUE #4084: Remove numpy <1.25 dependency in setup.py
In setup.py, I can see that the dependency for pandas has already been updated from 
"pandas>=1.4.3,<2" (which is still in the current PyPI version) to "pandas>=1.4.3,<3", but numpy hasn't, which will break the installation if I am using e.g. pandas 2.2.1, that requires numpy (>=1.26.0,<2)

## Problem
"numpy>=1.22,<1.25"

## Solution
"numpy>=1.22,<2"

## Steps to reproduce
poetry add git+https://github.com/feast-dev/feast.git

softwareentrepreneer:

change numpy version to <2 to avoid installation error

Fixes #4084

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ REQUIRED = [
     "Jinja2>=2,<4",
     "jsonschema",
     "mmh3",
-    "numpy>=1.22,<1.25",
+    "numpy>=1.22,<2",
     "pandas>=1.4.3,<3",
     # Higher than 4.23.4 seems to cause a seg fault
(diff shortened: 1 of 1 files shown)


lokeshrangineni:
LGTM

lokeshrangineni (approved):

lokeshrangineni (approved):

PR #4087, opened 2024-04-09
feat: Kubernetes materialization engine written based on bytewax

sudohainguyen:

As our efforts to prune bytewax dependencies to support python 3.11, in this PR I've rewritten bytewax engine as a new one called kubernetes engine.
Most of the code is unchanged, genuinely modify image entrypoint to process paths from remote storage

Relates to #4046 

# Additional notes
Docs will be provided in a separate PR and also bytewax removal

### DIFF
--- a/sdk/python/feast/infra/materialization/kubernetes/Dockerfile
+++ b/sdk/python/feast/infra/materialization/kubernetes/Dockerfile
@@ -5,7 +5,7 @@ RUN apt-get update && \
 
 WORKDIR /app
 
-COPY sdk/python/feast/infra/materialization/contrib/kuberentes/main.py /app
+COPY sdk/python/feast/infra/materialization/kuberentes/main.py /app
 
 # Copy necessary parts of the Feast codebase
 COPY sdk/python sdk/python
--- a/sdk/python/feast/infra/materialization/kubernetes/Dockerfile
+++ b/sdk/python/feast/infra/materialization/kubernetes/Dockerfile
@@ -1,0 +1,22 @@
+FROM python:3.9-slim-bullseye AS build
+
+RUN apt-get update && \
+    apt-get install --no-install-suggests --no-install-recommends --yes git
+
+WORKDIR /app
+
+COPY sdk/python/feast/infra/materialization/contrib/kuberentes/main.py /app
+
+# Copy necessary parts of the Feast codebase
+COPY sdk/python sdk/python
+COPY protos protos
+COPY go go
+COPY setup.py setup.py
+COPY pyproject.toml pyproject.toml
+COPY README.md README.md
+
+# We need this mount thingy because setuptools_scm needs access to the
+# git dir to infer the version of feast we're installing.
+# https://github.com/pypa/setuptools_scm#usage-from-docker
+# I think it also assumes that this dockerfile is being built from the root of the directory.
+RUN --mount=source=.git,target=.git,type=bind pip3 install --no-cache-dir '.[aws,gcp,k8s,snowflake,postgres]'
--- a/sdk/python/feast/infra/materialization/kubernetes/k8s_materialization_engine.py
+++ b/sdk/python/feast/infra/materialization/kubernetes/k8s_materialization_engine.py
@@ -1,0 +1,421 @@
+import logging
+import uuid
+from datetime import datetime
+from time import sleep
+from typing import Callable, List, Literal, Sequence, Union
+
+import yaml
+from kubernetes import client, utils
+from kubernetes import config as k8s_config
+from kubernetes.client.exceptions import ApiException
+from kubernetes.utils import FailToCreateError
+from pydantic import StrictStr
+from tqdm import tqdm
+
+from feast import FeatureView, RepoConfig
+from feast.batch_feature_view import BatchFeatureView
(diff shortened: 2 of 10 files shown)


sdk/python/feast/repo_config.py
```diff
@@ -47,6 +47,7 @@
     "snowflake.engine": "feast.infra.materialization.snowflake_engine.SnowflakeMaterializationEngine",
     "lambda": "feast.infra.materialization.aws_lambda.lambda_engine.LambdaMaterializationEngine",
     "bytewax": "feast.infra.materialization.contrib.bytewax.bytewax_materialization_engine.BytewaxMaterializationEngine",
+    "k8s.engine": "feast.infra.materialization.kubernetes.kubernetes_materialization_engine.KubernetesMaterializationEngine",
```
tokoko, line 50: Can we call this simply `k8s`? please.. 😄

sdk/python/feast/infra/materialization/kubernetes/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM python:3.9-slim-bullseye AS build
+
+RUN apt-get update && \
+    apt-get install --no-install-suggests --no-install-recommends --yes git
+
+WORKDIR /app
+
+COPY sdk/python/feast/infra/materialization/contrib/kuberentes/main.py /app
```
tokoko, line 8: the source path is wrong, i think

sudohainguyen: thanks, updated

sdk/python/feast/repo_config.py
sudohainguyen: haha been considering this too, just followed the spark.engine right below

tokoko:
Thanks, this looks really cool, especially as a transition path for people who use `bytewax` engine today. @alex-vinnik-sp

I'm all for merging this, but I would still like to note that I was actually talking with @lokeshrangineni about this the other day and I'm not sure it will be a good idea to keep this engine around if/when we will have good alternatives based on spark, dask or something similar, for example a spark engine that's not limited to working with just spark offline store and also has an adequate performance. The maintenance burden of a homegrown simple distributed engine will be too much then. I may be wrong of course... maybe people will still prefer to use this just to avoid spark overhead.

sudohainguyen:
@tokoko yep spark overhead sometimes frustrates user cause I was in that situation before 😃 
I believe users need an engine somewhere between local and spark.

And IIRC spark engine only available to spark offline store 🤔

tokoko:
@sudohainguyen only from a functionality perspective, you're probably right, materialization isn't so complicated after all. But if you consider that users will also want to have better monitoring, resilience and so on, using some other off-the-shelf engine makes more sense imho.

yup, spark only works with spark offline store, but there's nothing stopping us from adding a first step in spark engine as well similar to the one in here that exports offline store output to some storage first and only then spark would kick in to parallelize reading and writing.

tokoko (approved):
LGTM

lokeshrangineni:
Thank you for submitting this @sudohainguyen. This will unblock the [PR](https://github.com/feast-dev/feast/pull/4046) for now. I do agree with all the feedback on this PR.

lokeshrangineni:
@sudohainguyen - We have an [existing PR](https://github.com/feast-dev/feast/pull/4046) to remove bytewax dependencies and upgrading to 3.11 here. Just making sure you are aware of it so that we can avoid duplicate of work. Once this PR merged then we can merge the other PR.

sdk/python/feast/infra/materialization/kubernetes/k8s_materialization_engine.py
```diff
@@ -0,0 +1,421 @@
...
+                "parallelism": min(pods, self.batch_engine_config.max_parallelism),
+                "activeDeadlineSeconds": self.batch_engine_config.active_deadline_seconds,
+                "completionMode": "Indexed",
+                "template": {
+                    "metadata": {
+                        "annotations": self.batch_engine_config.annotations,
+                        "labels": {**pod_labels, **self.batch_engine_config.labels},
+                    },
+                    "spec": {
+                        "restartPolicy": "Never",
```
HaoXuAI, line 377: `"restartPolicy": "Never",` This can be also a config i think

sdk/python/feast/infra/materialization/kubernetes/k8s_materialization_engine.py
```diff
@@ -0,0 +1,421 @@
...
+        securityContextCapabilities = None
+        if self.batch_engine_config.include_security_context_capabilities:
+            securityContextCapabilities = {
+                "add": ["NET_BIND_SERVICE"],
+                "drop": ["ALL"],
+            }
+
+        job_labels = {"feast-materializer": "job"}
+        pod_labels = {"feast-materializer": "pod"}
+        job_definition = {
```
HaoXuAI, line 356: Can take this out as a template file, also make it a config option for the user to override the template

HaoXuAI (approved):
@sudohainguyen LGTM. left comments and I think can serve as future work.

I like this feature. it feels both powerful and lightweight. 👍

PR #4100, opened 2024-04-15
feat: Add tags to dynamodb config

nanohanno:
Add the option to configure and set tags on the AWS DynamoDB table resource created for an online store. Many organizations use tags to manage their resources which makes it convenient to directly add tags in the configuration file of Feast so that they are added upon creation of the tables.

# Fixes


franciscojavierarceo (approved):
lgtm

nanohanno:
The integration test failed due to an aws iam permission error. It seems that even when the `Tags` attribute in the CreateTable call is an empty array a call to the TagResource API is made which needs more permissions.

I think this is not only an issue of the integration test but in general it would break existing setups. The permission to add tags should only be required if tags are actually configured in the config file for DynamoDB.

I made a change that should avoid such an issue and add the `Tags` attribute only when configured in the file.

franciscojavierarceo (approved):

PR #4102, opened 2024-04-16
feat: Incorporate substrait ODFVs into ibis-based offline store queries

### LINKED ISSUE #3979: Abstract ibis-based Offline Store

tokoko:
This PR changes the offline flow for ODFV execution in ibis-based offline stores (currently only duckdb). Instead of collecting raw offline store output into a pyarrow table and applying a substrait transformation with acero, ibis-based offline stores now get ibis functions and directly extend ibis logical plan, meaning ODFVs will be executed by the offline store engine itself.

The feature necessitated a couple of underlying changes in `substrait` ODFVs:
1. `substrait` ODVS now also store ibis python functions in protos to directly apply these functions in `transform_ibis`. This can probably be avoided in the future if there's a good substrait to ibis compiler in place, but currently is the simplest solution. Note: there are other problems here as well, substrait plans need to know precisely what input columns they will consume, this is problematic in the offline flow as there's no way to know beforehand which features the user will request in `get_historical_features` call. To correctly apply ODFVs, one would need to apply each ODFV to a relevant subset of columns and then join all tables back together (kind of like how it's done in the online flow), which is a horrible waste of resources. 
2. Unlike pandas transformations, ibis functions must now return all the columns that are passed as an input to the function (in addition to on demand features) as opposed to returning only the on demand features that they generate. This is to avoid an extra join back to the main table after ibis functions are applied to the output. This way multiple ODFVs can easily be chained, each mutating the offline store output to append the relevant columns. 

With this PR, all of the planned ibis/substrait features will now be in place in python sdk. The next steps will be to enable native substrait ODFV execution in non-python sdks as well.

Fixes #3979

### DIFF
--- a/sdk/python/feast/transformation/substrait_transformation.py
+++ b/sdk/python/feast/transformation/substrait_transformation.py
@@ -1,5 +1,7 @@
+from types import FunctionType
 from typing import Any, Dict, List
 
+import dill
 import pandas as pd
 import pyarrow
 import pyarrow.substrait as substrait  # type: ignore # noqa
@@ -16,14 +18,16 @@ from feast.type_map import (
 
 
 class SubstraitTransformation:
-    def __init__(self, substrait_plan: bytes):
+    def __init__(self, substrait_plan: bytes, ibis_function: FunctionType):
         """
         Creates an SubstraitTransformation object.
 
         Args:
             substrait_plan: The user-provided substrait plan.
+            ibis_function: The user-provided ibis function.
         """
         self.substrait_plan = substrait_plan
+        self.ibis_function = ibis_function
 
     def transform(self, df: pd.DataFrame) -> pd.DataFrame:
         def table_provider(names, schema: pyarrow.Schema):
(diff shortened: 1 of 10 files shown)


sdk/python/feast/transformation/substrait_transformation.py
```diff
@@ -16,14 +18,16 @@
 
 
 class SubstraitTransformation:
-    def __init__(self, substrait_plan: bytes):
+    def __init__(self, substrait_plan: bytes, ibis_function: FunctionType):
```
HaoXuAI, line 21: should this param be optional?

tokoko: I think it should be fine this way. If we find some way to make it totally independent from ibis, then we can make it optional...

PR #4109, opened 2024-04-17
chore: Drop dependency on non-standard importlib libraries

tokoko:
Explicit dependency on non-standard importlib libraries (`importlib-resources` and `importlib_metadata`) is no longer required after dropping 3.8. This PR switches to use alternatives from standard library.


franciscojavierarceo (approved):
lgtm

PR #3953, opened 2024-02-16
fix: Using version args to install the correct feast version

bushwhackr:
**What this PR does / why we need it**:
feature_servers/multicloud does not use `master` sources to install feast in image. It uses pypi's `feast` latest version.
Even though the github publish action has a `needs` on `publish-python-sdk` before running the publish step, pypi cache might not be updated causing an older feast sdk version to be installed.

Refer to #3915 where,
```shell
$ docker run -it --rm feastdev/feature-server:0.35.0 pip show feast | grep Version
Version: 0.34.1
```

**Which issue(s) this PR fixes**:

Fixes #3915

PR #4112, opened 2024-04-17
revert: Reverts "fix: Using version args to install the correct feast version"

jeremyary:
Reverts feast-dev/feast#3953

### DIFF
--- a/sdk/python/feast/infra/feature_servers/multicloud/Dockerfile
+++ b/sdk/python/feast/infra/feature_servers/multicloud/Dockerfile
@@ -1,9 +1,5 @@
 FROM python:3.9
 
-# Input the feast version to install
-# This requires feast package to be available in pypi before building this image
-ARG VERSION
-
 RUN apt update && \
         apt install -y \
         jq \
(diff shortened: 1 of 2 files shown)


franciscojavierarceo (approved):

bushwhackr:
Wouldn't reverting this means that publish workflow `build-wheels/build-docker-image` would test the latest version published in pypi, in this case `0.36.0` and not the release `0.37.0` version? I don't see what is the point of the docker build test then.

I feel that the root problem is that the multicloud Dockerfile pulls from pypi instead of using the `master` branch artifacts which the other Dockerfiles do.

jeremyary:
I feel that the root problem is that the multicloud Dockerfile pulls from pypi instead of using the master branch artifacts which the other Dockerfiles do.

This is correct, however, we'll need to address with a different (or deeper) approach as the fix that was put in place has broken our release process by attempting to force resolution of the PyPi latest/new version without ensuring that the dependency functions correctly. We will be opening an issue shortly to chase the original multicloud image issue after we recover our release process.

HaoXuAI:
> Wouldn't reverting this means that publish workflow `build-wheels/build-docker-image` would test the latest version published in pypi, in this case `0.36.0` and not the release `0.37.0` version? I don't see what is the point of the docker build test then.
> 
> 
> 
> I feel that the root problem is that the multicloud Dockerfile pulls from pypi instead of using the `master` branch artifacts which the other Dockerfiles do.

That's right. Can you point to the other docker file that has the code using master artifacts? I can update to make it work.

In the meantime we need to cut a release which this revert needs

bushwhackr:
Understandable

PR #4114, opened 2024-04-18
chore: Use pixi to lock python dependencies in a single command

tokoko:
Adds `lock-python-dependencies-all` command in Makefile, that uses [pixi](https://prefix.dev/) (basically a faster alternative to conda) to regenerate all requirements files in a single command. I haven't changed the way dependencies are compiled yet, pixi is only used to provide a python runtime of a different version.


sudohainguyen:
how does it compare to poetry? 🤔

tokoko:
@sudohainguyen Well.. generally, poetry does a lot more, pixi is just a faster conda with better interface. If we are talking about environment management part, pixi and conda are more universal package managers, so they can manage python runtime itself as well, poetry manages only pypi packages with a python runtime that needs to be in-place.

sudohainguyen:
I see, just wondering that is there a success example of other opensource projects using pixi? 

another point, it would be great if we can integrate `uv` for packages resolver as well

tokoko:
Not sure about pixi adoption tbh, haven't really come across anywhere else, but I think the use case in this PR is not critical enough to warrant that. uv probably has better adoption, i think airflow is using it already. pixi does use `uv` under the hood for python packages btw and the projects are sort of intertwined generally.

Did you have time to look at this #4086? That PR only changes install step without touching resolver. I tried resolving locally as well, it was a lot faster, it just generated requirements files that were slightly different from current ones in formatting so I put it off for a little bit. Definitely worth making the change though.

jeremyary (approved):

PR #4117, opened 2024-04-19
fix: Pass native input values to `get_online_features` from feature server

### LINKED ISSUE #4110: feast serve convert number to INT64
## Expected Behavior 
when we use the get-online-features endpoint the request body numbers converted to Int64 type and while from python file calling get_online_feature function convert number to Int32 which leads to inconsistency between both use 
if the entity value type is Int32 then feature servering with online feast serve command will not work expected behavior is to convert the value to Int32 while it is being converted to Int64
## Current Behavior
when using feast serve number are converted to Int64 type 
## Steps to reproduce
create a entity with a column with type Int32 create a feature view with the entity apply the changes using feast apply materialize the data to online store start feature server using feast serve and call endpoint /get-online-feature to retrive feature ... expected response is the feature retrived  while it show none due to type issue 
### Specifications

- Version: 0.36
- Platform: 
- Subsystem:

## Possible Solution
before converting the type to Int64 get the repo and convert the type to the entity data type defined 
other solution is just write in doc to use Int64 and Float64 type while defining entity

tokoko:
Feature server invokes `get_online_features` a little differently from sdk. It converts native python values to proto values with a simple proto `Parse` and as a result doesn't take into account the exact data types of the entities involved. This PR removes this preprocessing step from feature server, which now passes native python values to `get_online_features` that already has a logic in place to better handle type conversion.

Fixes #4110

### DIFF
--- a/sdk/python/feast/feature_server.py
+++ b/sdk/python/feast/feature_server.py
@@ -100,7 +100,7 @@ def get_app(
             ).proto
 
             # Convert the Protobuf object to JSON and return it
-            return MessageToDict(  # type: ignore
+            return MessageToDict(
                 response_proto, preserving_proto_field_name=True, float_precision=18
             )
         except Exception as e:
--- a/sdk/python/feast/feature_server.py
+++ b/sdk/python/feast/feature_server.py
@@ -10,7 +10,7 @@ from dateutil import parser
 from fastapi import FastAPI, HTTPException, Request, Response, status
 from fastapi.logger import logger
 from fastapi.params import Depends
-from google.protobuf.json_format import MessageToDict, Parse
+from google.protobuf.json_format import MessageToDict
 from pydantic import BaseModel
 
 import feast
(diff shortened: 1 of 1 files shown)


sdk/python/feast/feature_server.py
```diff
@@ -83,30 +82,21 @@ def shutdown_event():
     @app.post("/get-online-features")
     def get_online_features(body=Depends(get_body)):
         try:
-            # Validate and parse the request data into GetOnlineFeaturesRequest Protobuf object
-            request_proto = GetOnlineFeaturesRequest()
-            Parse(body, request_proto)
-
+            body = json.loads(body)
```
franciscojavierarceo, line 85: I'm curious why they serialized this into Protobuf....it was probably on purpose so I'm not exactly sure changes it to Python objects is ideal but if tests pass I suppose it's hard to argue.

jeremyary (approved):

PR #4123, opened 2024-04-20
feat: Add delta format to `FileSource`, add support for it in ibis/duckdb

tokoko:
This PR adds delta format to `FileSource` (only parquet was supported before), also implements logic for handling delta sources in ibis offline store. It also adds integration tests for duckdb offline store with delta sources, dukdb tests are effectively run twice, once for parquet and then for delta.

### DIFF
--- a/protos/feast/core/DataFormat.proto
+++ b/protos/feast/core/DataFormat.proto
@@ -27,8 +27,12 @@ message FileFormat {
   // Defines options for the Parquet data format
   message ParquetFormat {}
 
+  // Defines options for delta data format
+  message DeltaFormat {}
+
   oneof format {
     ParquetFormat parquet_format = 1;
+    DeltaFormat delta_format = 2;
   }
 }
 
--- a/sdk/python/feast/data_format.py
+++ b/sdk/python/feast/data_format.py
@@ -43,6 +43,8 @@ class FileFormat(ABC):
         fmt = proto.WhichOneof("format")
         if fmt == "parquet_format":
             return ParquetFormat()
+        elif fmt == "delta_format":
+            return DeltaFormat()
         if fmt is None:
             return None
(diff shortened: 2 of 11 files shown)


HaoXuAI (commented):
Nice. 
I was wondering how we can use delta lake. One question, how is it possible to use the delta format with the spark offline store or spark materialization?

tokoko:
SparkSource already supports it afaik, SparkSource can also be used to basically just describe the source as a table name without actually caring about the format (I think). while that is probably fine for spark offline store, one downside is that it's tied to just a single implementation.

What I'm hoping to achieve with FileSource is that we can have a single source type that can be read by multiple offline stores (dask, duckdb, spark and others) so you can define a set of sources in your feature repository that won't lock you to use a specific offline store only.

HaoXuAI (approved):

HaoXuAI:
> SparkSource already supports it afaik, SparkSource can also be used to basically just describe the source as a table name without actually caring about the format (I think). while that is probably fine for spark offline store, one downside is that it's tied to just a single implementation.
> 
> 
> 
> What I'm hoping to achieve with FileSource is that we can have a single source type that can be read by multiple offline stores (dask, duckdb, spark and others) so you can define a set of sources in your feature repository that won't lock you to use a specific offline store only.

Make sense. I guess for the pyspark, in order to use delta format, it needs to install a delta spark plugin?

HaoXuAI:
LGTM!
Would be great to update the document as well

tokoko:
> Make sense. I guess for the pyspark, in order to use delta format, it needs to install a delta spark plugin?

Yeah, delta data source needs to be configured beforehand in some way.

ion-elgreco:
@tokoko nice PR! 

I kind of had put this on a back burner : P and focused my energy more on delta-rs.

tokoko:
thanks, this wouldn't be possible without delta-rs.

PR #4125, opened 2024-04-21
fix: Get container host addresses from testcontainers (java)

tokoko:
This PR makes similar changes to java integration tests as #3946 did to python ones. Changes are necessary for the tests to be runnable from inside a container.

- Resolves container address by calling `getHost()` instead of hardcoding `localhost`. The user will usually have to also set `TESTCONTAINERS_HOST_OVERRIDE` env variable to `host.docker.internal`
- `DockerComposeContainer` by default works by first running another container with `docker-compose` inside and mounting build context and running `docker-compose up` from there. This doesn't really work if you're already inside a container as volume mounts and `docker-in-docker` don't really get along well. This PR introduces an environment variable `FEAST_TESTCONTAINERS_LOCAL_COMPOSE` which if set changes the default behavior to running docker-compose locally (`docker-compose` needs to be in PATH).

### DIFF
--- a/java/serving/src/test/java/feast/serving/it/ServingRedisAzureRegistryIT.java
+++ b/java/serving/src/test/java/feast/serving/it/ServingRedisAzureRegistryIT.java
@@ -50,8 +50,10 @@ public class ServingRedisAzureRegistryIT extends ServingBaseTests {
     return new BlobServiceClientBuilder()
         .endpoint(
             String.format(
-                "http://localhost:%d/%s",
-                azureBlobMock.getMappedPort(BLOB_STORAGE_PORT), TEST_ACCOUNT_NAME))
+                "http://%s:%d/%s",
+                azureBlobMock.getHost(),
+                azureBlobMock.getMappedPort(BLOB_STORAGE_PORT),
+                TEST_ACCOUNT_NAME))
         .credential(CREDENTIAL)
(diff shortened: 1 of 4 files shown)


java/serving/src/test/java/feast/serving/it/ServingRedisAzureRegistryIT.java
```diff
@@ -50,8 +50,10 @@ private static BlobServiceClient createClient() {
     return new BlobServiceClientBuilder()
         .endpoint(
             String.format(
-                "http://localhost:%d/%s",
-                azureBlobMock.getMappedPort(BLOB_STORAGE_PORT), TEST_ACCOUNT_NAME))
+                "http://%s:%d/%s",
```
franciscojavierarceo, line 53: nit: f-strings are a little nicer in many ways.

tokoko: 😄 this is java, i don't think there's a better option

franciscojavierarceo: 🤦‍♂️ this is what I get for reviewing PRs before coffee

franciscojavierarceo (approved):

PR #4126, opened 2024-04-21
chore: Move feast install to docker build in java it tests

tokoko:
In a docker-compose setup used in java integration tests, `pip install .` is part of an entrypoint.sh file, which effectively means that it's run once for each test (there are 6 tests as of now). This PR changes compose setup and moves feast pip install to a docker build step. This should noticeably speed up java integration tests.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -7,10 +7,7 @@
     },
     "ghcr.io/devcontainers/features/python:1": {
       "version": "3.9"
-    },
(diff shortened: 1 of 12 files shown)


tokoko:
@lokeshrangineni @jeremyary seems like `macos-latest` label switched to using arm, had to pin it to `macos-12` everywhere.

tokoko:
got to green, finally. managed to cut down java integration tests runtime from ~10m to ~7m. hoped for more, but still something :)

jeremyary (approved):

PR #4134, opened 2024-04-23
fix: Specify version instead of latest macOS runner version for unit test actions

jeremyary:
Attempt pinning a version for MacOS action runner in unit tests as I suspect macos-latest (resolving to 14) is causing issues for dependency libs such as seen [here](https://github.com/feast-dev/feast/actions/runs/8789249143/job/24119321935).


jeremyary:
closing PR, ran as a secondary test to validate the runner image update has caused us issues. Will take Tornike's #4134 with the fix across the board for runners instead.

Issue #4135, opened 2024-04-23
Multicloud Dockerfile uses latest published version of Feast instead of version under testing


jeremyary:
## Expected Behavior 
Dockerfiles within GitHub Actions should utilize the latest/pending version of library instead of using latest/previous. [This PR](https://github.com/feast-dev/feast/pull/3953) describes the original issue:

> feature_servers/multicloud does not use master sources to install feast in image. It uses pypi's feast latest version.
Even though the github publish action has a needs on publish-python-sdk before running the publish step, pypi cache might not be updated causing an older feast sdk version to be installed.

[This PR](https://github.com/feast-dev/feast/pull/4112) reverted the change - our builds began breaking due to a race condition wherein the Dockerfile expected the pending version of library to be available based on `needs` conditional associated to the Action, however, the action failed as either the sequential dependency isn't respected or something else is leading the Dockerfile to poll external repositories for the library version and failing in cases where the pending code version isn't yet published.
 
## Current Behavior
Dockerfile is using latest published version as opposed to pending version - see original [PR](https://github.com/feast-dev/feast/pull/3953) for more.

### Specifications

- Version:
- Platform:
- Subsystem:

PR #4138, opened 2024-04-23
fix: Bad feature_servers/multicloud Dockerfile

bushwhackr:

Multicloud dockerfile is not consistent with the rest of the infra/feature_servers in that it pulls pypi feast package. Whilst others uses the repository code to build feast.

This PR seeks to standardise it so that we are not reliant on the pypi feast package.

**NOTE:** I am not a user of multicloud feast and do not understand its use case. The Dockerfile I created is based on my observation on what the incumbent was doing. 

Which I understood as the following:
1. Install feast into docker image
2. Install runtime dependencies like libarrow-dev

I would need clarification on what is the purpose of this Docker image.

#4135, Related: #3953 

# Fixes


bushwhackr:
cc @jeremyary

jeremyary:
Thanks! We'll take a look. As far as use case goes (as far as I understand), it's a stretch to imagine an end user utilizing every cloud provider we enable, but this checks the box for ensuring whatever variant of subsets a user plugs in are good to go together.

bushwhackr:
any blockers?

franciscojavierarceo:
i believe this is used during our deployment to verify the build

shuchu:
I believe the original Dockerfile  (with pypi installation) still useful while the users want to test official Feast releases.

Makefile
```diff
@@ -395,7 +395,7 @@ build-feature-server-java-docker:
 build-feature-server-dev:
 	docker buildx build --build-arg VERSION=dev \
 		-t feastdev/feature-server:dev \
-		-f sdk/python/feast/infra/feature_servers/multicloud/Dockerfile.dev --load .
```
franciscojavierarceo, line 398: was this used at all for deploying any images? Do you want to ask achal or any of the old maintainers?

bushwhackr: The only difference in the previous Dockerfile vs Dockerfile.dev was the `COPY . .` instruction.

The old Dockerfile installs only the pypi available feast package, whereas the old Dockerfile.dev installs the pypi + copies the current codebase into the image.

With the changes I made to the Dockerfile, it does not rely on pypi for the feast package and we would only use the current codebase to install feast. Essentially the same as what old Dockerfile.dev was doing.

shuchu:
> I believe the original Dockerfile (with pypi installation) still useful while the users want to test official Feast releases.

ah, I know the problem. In the document, development part. The last step is: 
`docker build -t docker-whale -f ./sdk/python/feast/infra/feature_servers/multicloud/Dockerfile .`

As a result, it will call the pypi version....

Issue #4128, opened 2024-04-23
mysql test case(test_sql_registry.py) on mac environment is failing due to the container is not starting correctly.


lokeshrangineni:
## Expected Behavior 
test_sql_registry.py should pass on mac environment.

## Current Behavior
test_sql_registry.py should be able to run on the local mac environment without any issues. Currently these are skipped in integration tests as well. 

Here is the error seeing on mac local environment

`===================================================================================================== short test summary info =====================================================================================================
ERROR sdk/python/tests/unit/test_sql_registry.py::test_apply_entity_success[mysql_registry] - TimeoutError: container did not emit logs satisfying predicate in 60.000 seconds
ERROR sdk/python/tests/unit/test_sql_registry.py::test_registry_cache[mysql_registry] - TimeoutError: container did not emit logs satisfying predicate in 60.000 seconds
ERROR sdk/python/tests/unit/test_sql_registry.py::test_apply_on_demand_feature_view_success[mysql_registry] - TimeoutError: container did not emit logs satisfying predicate in 60.000 seconds
ERROR sdk/python/tests/unit/test_sql_registry.py::test_apply_feature_view_success[mysql_registry] - TimeoutError: container did not emit logs satisfying predicate in 60.000 seconds
ERROR sdk/python/tests/unit/test_sql_registry.py::test_modify_feature_views_success[mysql_registry] - TimeoutError: container did not emit logs satisfying predicate in 60.000 seconds
ERROR sdk/python/tests/unit/test_sql_registry.py::test_update_infra[mysql_registry] - TimeoutError: container did not emit logs satisfying predicate in 60.000 seconds
ERROR sdk/python/tests/unit/test_sql_registry.py::test_apply_data_source[mysql_registry] - TimeoutError: container did not emit logs satisfying predicate in 60.000 seconds
===================================================================================== 249 passed, 831 warnings, 7 errors in 100.34s (0:01:40) =====================================================================================
make: *** [test-python-unit] Error 1`

## Steps to reproduce
Just run the tests on mac environment(intel) - run command `make test-python-unit` on the project root directory.
### Specifications

- Version:
- Platform:
- Subsystem:

lokeshrangineni:
Created the PR to fix the issue on mac local environment - https://github.com/feast-dev/feast/pull/4140

PR #4140, opened 2024-04-23
fix: Changed the code the way mysql container is initialized. 

lokeshrangineni:
changed the code the way mysql container is initialized. Trying to fix the issue - https://github.com/feast-dev/feast/issues/4128

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -7,10 +7,7 @@
     },
     "ghcr.io/devcontainers/features/python:1": {
       "version": "3.9"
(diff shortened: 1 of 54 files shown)


tokoko:
Are skip markers still necessary?

lokeshrangineni:
> Are skip markers still necessary?

Yes, at the moment. Docker is not installed by default in the macos due to licensing issues but colima is available. I need to spend some more time to figure out initializing and make it work. I will handle that in separate PR.

jeremyary (approved):

PR #4143, opened 2024-04-24
feat: Make arrow primary interchange for online ODFV execution

tokoko:
This PR is an online counterpart of #4083. online response object no longer needs to be converted to pandas for substrait transformations to be executed. Also cleans up some leftover unused code from `on_demand_feature_view.py`.

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -2138,7 +2138,7 @@ class FeatureStore:
                 )
 
         initial_response = OnlineResponse(online_features_response)
-        initial_response_df: Optional[pd.DataFrame] = None
+        initial_response_arrow: Optional[pa.Table] = None
         initial_response_dict: Optional[Dict[str, List[Any]]] = None
 
         # Apply on demand transformations and augment the result rows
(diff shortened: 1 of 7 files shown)


HaoXuAI (approved):
lgtm!

HaoXuAI:
What if we store arrow format in the online/offline store? Is it going to be better than the current proto serialized value?

tokoko:
@HaoXuAI that did occur to me, I compared the sizes for some sample results (very non-scientific). We would lose considerably on size (and retrieval latency as a result, I'm assuming), arrow has significantly more overhead, which is a big problem when you start using it on a per-row basis. Additionally it is intended to have O(1) retrieval for particular row index, which means that proto encoding tricks (like varints) go out the windows.

As for the offline store, most of the same reasons apply there as well, retrieval latency is not as important, but formats liek parquet take significantly less storage size still. but i guess we could support arrow in `FileSource` as an additional format if that makes sense for anyone.

HaoXuAI:
> @HaoXuAI that did occur to me, I compared the sizes for some sample results (very non-scientific). We would lose considerably on size (and retrieval latency as a result, I'm assuming), arrow has significantly more overhead, which is a big problem when you start using it on a per-row basis. Additionally it is intended to have O(1) retrieval for particular row index, which means that proto encoding tricks (like varints) go out the window.
> 
> As for the offline store, most of the same reasons apply there as well, retrieval latency is not as important, but formats like parquet take significantly less storage size still. but i guess we could support arrow in `FileSource` as an additional format if that makes sense for anyone.

That's good to know. thanks, but I guess for retrieval (get_online_feature) is just a `where feature = value`, so it doesn't matter using protobuf or arrow format?

HaoXuAI:
https://github.com/adriangb/pgpq Something like this. but looks like it still needs to encode into postgresl field which doesn't give benefits anymore.

Issue #4144, opened 2024-04-24
Add a Kubernetes Operator for the Feast Feature Server, to the infra directory


tchughesiv:
**Is your feature request related to a problem? Please describe.**
We should leverage the power of [kubernetes controllers](https://kubernetes.io/docs/concepts/extend-kubernetes/operator/) to manage the deployment state of the feast-feature-server. This is more operationally sound, as it can “repair” issues with deployments in a cluster. It also simplifies the management of multiple deployments within a cluster and opens the door to integrations with other operator-based projects.

**Describe the solution you'd like**
Wrap the existing [feast-feature-server helm chart](https://github.com/feast-dev/feast/tree/master/infra/charts/feast-feature-server) into a simple [helm-based operator](https://sdk.operatorframework.io/docs/building-operators/helm/quickstart/).

**Describe alternatives you've considered**
I considered creating a go-based operator instead. This would add significant complexity to the development and maintenance of feast. Given that feast and its users already utilize helm, this seemed unnecessary.

**Additional context**
This new helm-based operator will leverage a few static kustomize manifests to deploy the operator, but it will utilize the helm chart to deploy the feature server(s) themselves. Users will not have to install helm, the operator will handle everything. Those familiar with the existing helm chart will be able to easily port their settings if they so choose.

PR #4145, opened 2024-04-24
feat: Add a Kubernetes Operator for the Feast Feature Server

tchughesiv:

This PR wraps the existing [feast-feature-server helm chart](/feast-dev/feast/tree/master/infra/charts/feast-feature-server) into a simple [helm-based operator](https://sdk.operatorframework.io/docs/building-operators/helm/quickstart/).

We should leverage the power of [kubernetes controllers](https://kubernetes.io/docs/concepts/extend-kubernetes/operator/) to manage the deployment state of the feast-feature-server. This is more operationally sound, as it can “repair” issues with deployments in a cluster. It also simplifies the management of multiple deployments within a cluster and opens the door to integrations with other operator-based projects.

This will require the creation of a new image repo @ `docker.io/feastdev/feast-operator`. Builds are automated, same as the existing feature server images.

While this solution will leverage a few kustomize manifests to deploy the operator itself, these manifests will require little to no maintenance. They should not change often (if at all) as they are not impacted by changes to the helm chart, nor do they impact how the feature server is deployed.

By using a symlink to the feast-feature-server helm directory, future operator images will always have the latest chart version enhancements.

Below is a sample FeastFeatureServer [Custom Resource](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/). As you can see, the fields in the `FeastFeatureServer.spec` directly match existing [helm chart parameters](/feast-dev/feast/blob/master/infra/charts/feast-feature-server/values.yaml) -
```yaml
apiVersion: charts.feast.dev/v1alpha1
kind: FeastFeatureServer
metadata:
  name: feastfeatureserver-sample
spec:
  # Default values copied from helm-charts/feast-feature-server/values.yaml
  affinity: {}
  # base64 encoding of `sdk/python/feast/templates/local/feature_repo/feature_store.yaml`
  feature_store_yaml_base64: "cHJvamVjdDogbXlfcHJvamVjdAojIEJ5IGRlZmF1bHQsIHRoZSByZWdpc3RyeSBpcyBhIGZpbGUgKGJ1dCBjYW4gYmUgdHVybmVkIGludG8gYSBtb3JlIHNjYWxhYmxlIFNRTC1iYWNrZWQgcmVnaXN0cnkpCnJlZ2lzdHJ5OiBkYXRhL3JlZ2lzdHJ5LmRiCiMgVGhlIHByb3ZpZGVyIHByaW1hcmlseSBzcGVjaWZpZXMgZGVmYXVsdCBvZmZsaW5lIC8gb25saW5lIHN0b3JlcyAmIHN0b3JpbmcgdGhlIHJlZ2lzdHJ5IGluIGEgZ2l2ZW4gY2xvdWQKcHJvdmlkZXI6IGxvY2FsCm9ubGluZV9zdG9yZToKICAgIHR5cGU6IHNxbGl0ZQogICAgcGF0aDogZGF0YS9vbmxpbmVfc3RvcmUuZGIKZW50aXR5X2tleV9zZXJpYWxpemF0aW9uX3ZlcnNpb246IDIK"
  fullnameOverride: ""
  image: {}
  imagePullSecrets: []
  livenessProbe:
    initialDelaySeconds: 30
    periodSeconds: 30
  nameOverride: ""
  nodeSelector: {}
  podAnnotations: {}
  podSecurityContext: {}
  readinessProbe:
    initialDelaySeconds: 20
    periodSeconds: 10
  replicaCount: 1
  resources: {}
  securityContext: {}
  service:
    port: 80
    type: ClusterIP
  tolerations: []
```

NOTE: This operator could easily be expanded in the future to also handle the release of the [feast-release helm chart](/infra/charts/feast)

Fixes https://github.com/feast-dev/feast/issues/4144


infra/feast-operator/Makefile
```diff
@@ -0,0 +1,227 @@
+# VERSION defines the project version for the bundle.
+# Update this value when you upgrade the version of your project.
+# To re-generate a bundle for another specific version without changing the standard setup, you can:
+# - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
+# - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
+VERSION ?= 0.37.1
```
franciscojavierarceo, line 6: this should probably be picked up from somewhere...

tchughesiv: i added it to the existing release version scripting -
https://github.com/feast-dev/feast/blob/816d8f1c9ca796a689c341973c1d52678d419b69/infra/scripts/release/files_to_bump.txt#L13

and it is picked up during builds from the root Makefile - 
https://github.com/feast-dev/feast/blob/816d8f1c9ca796a689c341973c1d52678d419b69/Makefile#L407

let me know though if i should script something separately.

infra/feast-operator/README.md
```diff
@@ -0,0 +1,11 @@
+# Feast Feature Server Helm-based Operator
+
+Leverages [operator-sdk](https://github.com/operator-framework/operator-sdk) and the [feast-feature-server helm chart](/infra/charts/feast-feature-server).
+
+### To run against an OCP cluster -
```
jeremyary, line 5: small request, can we swap OCP to K8S here since we're upstream?

tchughesiv: done

jeremyary: thanks!

infra/feast-operator/README.md
```diff
@@ -0,0 +1,11 @@
+# Feast Feature Server Helm-based Operator
+
+Leverages [operator-sdk](https://github.com/operator-framework/operator-sdk) and the [feast-feature-server helm chart](/infra/charts/feast-feature-server).
+
+### To run against an K8S cluster -
+```bash
+$ make deploy
+
```
franciscojavierarceo, line 8: Could we add getting started docs for this using Kind?

infra/feast-operator/Makefile
```diff
@@ -0,0 +1,227 @@
...
+
+.PHONY: kustomize
+KUSTOMIZE = $(shell pwd)/bin/kustomize
+kustomize: ## Download kustomize locally if necessary.
+ifeq (,$(wildcard $(KUSTOMIZE)))
+ifeq (,$(shell which kustomize 2>/dev/null))
+	@{ \
+	set -e ;\
+	mkdir -p $(dir $(KUSTOMIZE)) ;\
+	curl -sSLo - https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize/v5.2.1/kustomize_v5.2.1_$(OS)_$(ARCH).tar.gz | \
```
dmartinol, line 134: shouldn't we extract some constants for the kustomize and helm-operator versions?

infra/feast-operator/config/samples/charts_v1alpha1_feastfeatureserver.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: charts.feast.dev/v1alpha1
+kind: FeastFeatureServer
+metadata:
+  name: feastfeatureserver-sample
+spec:
+  # Default values copied from helm-charts/feast-feature-server/values.yaml
+  affinity: {}
+  # base64 encoding of `sdk/python/feast/templates/local/feature_repo/feature_store.yaml`
+  feature_store_yaml_base64: "cHJvamVjdDogbXlfcHJvamVjdAojIEJ5IGRlZmF1bHQsIHRoZSByZWdpc3RyeSBpcyBhIGZpbGUgKGJ1dCBjYW4gYmUgdHVybmVkIGludG8gYSBtb3JlIHNjYWxhYmxlIFNRTC1iYWNrZWQgcmVnaXN0cnkpCnJlZ2lzdHJ5OiBkYXRhL3JlZ2lzdHJ5LmRiCiMgVGhlIHByb3ZpZGVyIHByaW1hcmlseSBzcGVjaWZpZXMgZGVmYXVsdCBvZmZsaW5lIC8gb25saW5lIHN0b3JlcyAmIHN0b3JpbmcgdGhlIHJlZ2lzdHJ5IGluIGEgZ2l2ZW4gY2xvdWQKcHJvdmlkZXI6IGxvY2FsCm9ubGluZV9zdG9yZToKICAgIHR5cGU6IHNxbGl0ZQogICAgcGF0aDogZGF0YS9vbmxpbmVfc3RvcmUuZGIKZW50aXR5X2tleV9zZXJpYWxpemF0aW9uX3ZlcnNpb246IDIK"
```
dmartinol, line 9: shouldn't we document somewhere what's the actual feature_store.yaml that this is representing? (and maybe in the future we can also add a more complex sample with integrated psql DB deployment)

tchughesiv: it's currently documented in the comment above field... that said, i can certainly add something additional to the getting started docs i'll be adding.

infra/feast-operator/Makefile
tchughesiv: good call

infra/feast-operator/config/samples/charts_v1alpha1_feastfeatureserver.yaml
tchughesiv: fwiw, to keep maintenance low... i'd like to lean heavily on existing feast docs and samples. there's no need/reason for us to duplicate anything. for example - 
https://github.com/feast-dev/feast/blob/master/sdk/python/feast/templates/postgres/feature_repo/feature_store.yaml
https://docs.feast.dev/project/development-guide#feast-python-feature-server-helm-chart

infra/feast-operator/Makefile
tchughesiv: done

infra/feast-operator/config/samples/charts_v1alpha1_feastfeatureserver.yaml
tchughesiv: i've added some additional docs, including how one might pass in their own `feature_store.yaml` during a deployment.

jeremyary:
@tokoko @shuchu @HaoXuAI @sudohainguyen hey all, may I kindly ask for your considerations of the effort here? We'd love to hear any feedback or questions/convo around technical and/or objective aspects from some of our heavier contributors, so if you're able, we'd really appreciate it!

infra/feast-operator/README.md
franciscojavierarceo: I meant this documentation: https://kind.sigs.k8s.io/docs/user/quick-start/

Probably a lot of users may not be familiar with K8s and so a quick intro with the minimum commands to be able to run it locally is useful.

tchughesiv: we aren't using the `kind` tool here so i worry those docs would confuse the user.

tchughesiv: we aren't using the `kind` tool here so i worry those docs would confuse the user

tchughesiv: maybe you're refering to the [CRD kind](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/)?

franciscojavierarceo: Kind is another tool for running local kubernetes clusters (like minikube). Kind is the recommended for Kubeflow so I figured it'd be a good thing for us to add some docs on it but that's fine.

franciscojavierarceo: Here's an example: https://github.com/kubeflow/training-operator/blob/master/docs/development/developer_guide.md

tokoko:
> This is more operationally sound, as it can “repair” issues with deployments in a cluster. It also simplifies the management of multiple deployments within a cluster and opens the door to integrations with other operator-based projects.

I'll be completely honest, this is my first time hearing of a helm-based operator and I'm not sure I understand the point. 😄What's the difference with normal helm deployment in any practical sense other than the fact that parameters are used to populate a CRD instead of helm values file? Previously I thought (maybe just my limited experience) that controllers came in handy when you needed some very custom behavior in cluster or maybe even orchestrating stuff outside the cluster. So my questions right now are 1) Is something like that even possible with helm-based operators? and 2) do we have any use case for it (actual or expected)? java feature server deployment doesn't look that unusual to me. I'm probably unaware of some other benefits, so some more context there will be appreciated...

HaoXuAI (approved):
Nice. I personally think the feast deployment process is not devops friendly. Hope this feature streamlines the process better

tokoko:
Another slightly unrelated point to note here. I think feast has bit of an identity crisis when it comes to deployments like this. For example we have [lambda feature server](https://docs.feast.dev/reference/feature-servers/alpha-aws-lambda-feature-server) which has it's own registry object and is actually deployed when user runs `feast apply` (not a fan of this myself). At the same time we have ways to start servers either with cli or deploy with helm. I think we should try to make all this less confusing. Personally I think deploying anything anywhere with a `feast apply` needs to be killed off.

franciscojavierarceo:
> Another slightly unrelated point to note here. I think feast has bit of an identity crisis when it comes to deployments like this. For example we have [lambda feature server](https://docs.feast.dev/reference/feature-servers/alpha-aws-lambda-feature-server) which has it's own registry object and is actually deployed when user runs `feast apply` (not a fan of this myself). At the same time we have ways to start servers either with cli or deploy with helm. I think we should try to make all this less confusing. Personally I think deploying anything anywhere with a `feast apply` needs to be killed off.

I think we need to be able to allow users to pick and choose based on tradeoffs. Being able to deploy feature changes via `feast apply` is a choice that has been very successful for some and not within the risk tolerance of others. So I don't think we should be overly prescriptive on that.

tokoko:
@franciscojavierarceo sorry, let me clarify. I wasn't referring to feature changes, I meant deployments like aws lambda feature server.

tchughesiv:
@tokoko let me know if you have additional Qs

> > This is more operationally sound, as it can “repair” issues with deployments in a cluster. It also simplifies the management of multiple deployments within a cluster and opens the door to integrations with other operator-based projects.
> 
> I'll be completely honest, this is my first time hearing of a helm-based operator and I'm not sure I understand the point. 😄What's the difference with normal helm deployment in any practical sense other than the fact that parameters are used to populate a CRD instead of helm values file?

- The Operator’s controller will automatically recover missing objects in the cluster, and fix misconfigured ones. For example, if a feature server’s k8s `Service` or `Deployment` were accidentally deleted, the operator will recreate it. In critical environments, this is invaluable.
- Changes made to a `FeastFeatureServer` CR in-cluster will automatically be reflected in the feature server deployment. Nothing else has to be done.
- Once an admin installs the operator to a cluster, all users can take advantage using native k8s tools to deploy their own feature servers, with the use of the new `FeastFeatureServer` CR. No helm knowledge or installation is required. This will be identical to how a user might create a `Deployment` object in their own namespace.

> Previously I thought (maybe just my limited experience) that controllers came in handy when you needed some very custom behavior in cluster or maybe even orchestrating stuff outside the cluster. So my questions right now are 1) Is something like that even possible with helm-based operators?

- Helm-based operators are limited in that respect. We would have to use a Go-based operator instead. But, as I stated in the related issue (https://github.com/feast-dev/feast/issues/4144), a go-based operator would add significant complexity to the development and maintenance of feast. Given that feast and its users already utilize, and prefer, helm… this seemed unnecessary. My intention with this approach was to continue to funnel future deployment contributions to the helm charts… rather than other deployment methods. Any future changes made to the helm charts will be automatically reflected in this helm-based operator.
- That said, do we think the community would prefer a Go-based operator instead? They can be a lot to take on and maintain, but i have significant experience with them.

> and 2) do we have any use case for it (actual or expected)?

- IMO, aside from the controller benefits stated above, it's a great way to potentially consolidate all of the helm charts into a single, optional deployment mechanism. In the future, should the community so choose, all helm chart installs could be handled by this same feast-operator. Each would have its own CRD. e.g. `FeastJavaFeatureServer` or `FeastTransformationService`, etc.
- Also, should the community choose to, in the future this feast-operator could be added to https://operatorhub.io/ for greater visibility to potential k8s users… or to OpenShift’s internal OperatorHub, which is then made available to all OpenShift & OKD cluster users.
- More immediately, we’d be able to easily integrate the feast-feature-server into the [Open Data Hub (ODH) Operator](https://github.com/opendatahub-io/opendatahub-operator).

> java feature server deployment doesn't look that unusual to me. I'm probably unaware of some other benefits, so some more context there will be appreciated...

sudohainguyen:
totally support this, since there's a lot of room to consolidate deployment for feast
I would prefer going with the helm-based operator first due to its simplification 🙌

tokoko:
@tchughesiv thanks for the lengthy response. I do see the benefits and since people will also be able to use normal helm charts as well, I don't see why we shouldn't do this. as for the go-based approach, I agree with Harry, right now helm is enough. If a suitable use case arises for a go-based one, we can revisit this later.

tchughesiv:
thanks @tokoko @sudohainguyen
if you get a minute, there's a related PR https://github.com/feast-dev/feast/pull/4161 that affects the full functionality of this operator. very simple helm fix.

PR #4147, opened 2024-04-25
fix: Change checkout action back to v3 from v5 which isn't released yet

jeremyary:

Fixes newly broken action in automation by updating back to correct checkout action version

# Fixes

### DIFF
--- a/.github/workflows/master_only.yml
+++ b/.github/workflows/master_only.yml
@@ -81,7 +81,7 @@ jobs:
           --health-timeout 5s
           --health-retries 5
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v3
       - name: Setup Python
         id: setup-python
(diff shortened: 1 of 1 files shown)


franciscojavierarceo (approved):

.github/workflows/master_only.yml
```diff
@@ -81,7 +81,7 @@ jobs:
           --health-timeout 5s
           --health-retries 5
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v3
```
franciscojavierarceo, line 84: v4?

https://github.com/actions/checkout?tab=readme-ov-file#checkout-v4

jeremyary: We have a couple dozen of em, probably need to get them all at once. Can we fix tests with this PR and I'll open another to update the whole lot from 3 --> 4?

PR #4148, opened 2024-04-25
chore: Update checkout action from v3 to v4 in all workflows

jeremyary:
After recently fixing a typo issue, we noticed the checkout action is ready for a version bump, so applying across the board.


franciscojavierarceo (approved):
nice!

PR #4152, opened 2024-04-25
chore: Bump macOS runners to macos-13

tokoko:
macos runners are obviously slower than linux ones in the ci. We are using `macos-12` which is (3CPU/14RAM) while `linux-latest` is (4CPU/16RAM). This PR bumps macos runner version to 13, which should be a similar image with [4 cores](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories).


HaoXuAI (approved):
LGTM

PR #4156, opened 2024-04-28
feat: Adding support for dictionary writes to online store 

franciscojavierarceo:

This PR enables support for passing a Python `Dict` object to write to the online store. This is a first step along the path to optimizing feature writes to the online store.

Note, this won't speed up anything and, in fact, introduces minor latency to serialize the dict into a DataFrame. In a subsequent PR I will directly use the dictionary object as it propagates through the Feast backend layer to write rather than any of the existing pandas code. When I do this next step, I will measure the performance benefits to highlight the latency gains.

N/A

# Fixes
N/A


sdk/python/feast/feature_store.py
```diff
@@ -1406,7 +1407,8 @@ def push(
     def write_to_online_store(
         self,
         feature_view_name: str,
-        df: pd.DataFrame,
+        df: Optional[pd.DataFrame] = None,
+        dict: Optional[Dict] = None,
```
shuchu, line 1411: "dict" is a preserved key word of Python. Maybe a better variable name?

sdk/python/feast/feature_store.py
```diff
@@ -1406,7 +1407,8 @@ def push(
     def write_to_online_store(
         self,
         feature_view_name: str,
-        df: pd.DataFrame,
+        df: Optional[pd.DataFrame] = None,
+        input_dict: Optional[Dict] = None,
```
tokoko, line 1411: Wouldn't this make more sense as a single param with a Union type?

franciscojavierarceo: We can do a staggered release for this and get feedback from folks but I wouldn't want to make any breaking changes.

sudohainguyen:
hey quick question, how will the change optimize feature writes? 🤔  
as I can see this is about parsing an input dictionary to a dataframe, would you mind walking me through the mechanism?

HaoXuAI:
> hey quick question, how will the change optimize feature writes? 🤔 as I can see this is about parsing an input dictionary to a dataframe, would you mind walking me through the mechanism?

I think this is the same question for me :)

sdk/python/feast/feature_store.py
tokoko: yeah, `df` name is really awkward. maybe call the new one something other than `input_dict` so that it can become the universal one after we deprecate and remove `df`? just `inputs` for example.

franciscojavierarceo:
> > hey quick question, how will the change optimize feature writes? 🤔 as I can see this is about parsing an input dictionary to a dataframe, would you mind walking me through the mechanism?

It won't yet. First I want to introduce this at this layer and add the tests. Then I want to measure the incremental performance of skipping pandas all together using `cProfiler`.  So this is step 1 that just introduces the argument to the user.

sdk/python/feast/feature_store.py
franciscojavierarceo: Yeah I like that. I'll update.

tokoko:
@franciscojavierarceo btw, what is `Dict` here exactly? Is this always `Dict[str, np.array]` or there are other options? P.S. If you're gonna spend some time performance testing this, I think adding pyarrow table in the mix would be a good idea.

franciscojavierarceo:
> @franciscojavierarceo btw, what is `Dict` here exactly? Is this always `Dict[str, np.array]` or there are other options? P.S. If you're gonna spend some time performance testing this, I think adding pyarrow table in the mix would be a good idea.

I haven't spent any time using pyarrow but definitely open to it.

sdk/python/feast/feature_store.py
```diff
@@ -1406,7 +1407,8 @@ def push(
     def write_to_online_store(
         self,
         feature_view_name: str,
-        df: pd.DataFrame,
+        df: Optional[pd.DataFrame] = None,
+        inputs: Optional[Union[Dict[str, List[Any]], pd.DataFrame]] = None,
```
franciscojavierarceo, line 1411: @tokoko updated

jeremyary (approved):

PR #4159, opened 2024-04-29
feat: Upgrading python version to 3.11, adding support for 3.11 as well.

lokeshrangineni:

# Fixes


franciscojavierarceo (approved):
lgtm

lokeshrangineni:
> Maybe we can also change `integration-test-python-local` runtime to 3.11? Running integration tests twice sounds like too much of a headache, but at least we can try to cover two different versions with normal and local it tests.

I have incorporated your change.

PR #4164, opened 2024-04-30
fix: Correct nightly install command, move all installs to uv

### LINKED ISSUE #4163: Nightly CI runs failing on install-python-ci-dependencies / test-python-integration
## Current Behavior
nightly-ci runs are erroring out, first run with failure can be seen [here](https://github.com/feast-dev/feast/actions/runs/8813278172/job/24190831412#step:19:38).

image

tokoko:
Fixes #4163


jeremyary:
yeeeey throttling... I'll give your failed jobs another bump once current pipelines are caught up

tokoko:
@jeremyary lets' try, but I doubt it will help, three integration tests pipelines side-by-side will surely fail again, I think. That was probably the reason why we only had one before.

jeremyary (approved):

PR #3925, opened 2024-02-02
fix: Remove unnecessary dependency on mysqlclient

tokoko:
**What this PR does / why we need it**:
feast mysql extra declares dependency on `mysqlclient`, but that's actually incorrect because `mysqlclient` is only used in sql registry tests, mysql online store implementation uses `pymysql`. this PR changes the relevant tests to use pymysql as well and removes `mysqlclient` from the project.

P.S. This might break someone's build if they relied on a transitive dependency provided by feast, but I don't think that's enough of a reason to keep an unnecessary extra dependency.

**Which issue(s) this PR fixes**:
Fixes #3916

PR #4166, opened 2024-05-01
chore: Change dev workflow to use uv (and pixi) both for lock and install

tokoko:
This PR makes a number of changes to the make commands and dev guide:

- docs now recommend using `uv` for setting up local development. I changed env install step (step 9) to run `make install-python-ci-dependencies-uv`, which differs from the current version in 2 ways: 1) It's using uv instead of pip 2) It's installing package versions from the lock file rather than resolving versions from pypi. This will make local env more predictable and comparable to ci environment which uses the same make command. (Let me know if you disagree with this...)
- Changed all `lock-python` commands to use uv instead of pip-tools. dev guide now recommends the usage of `lock-python-dependencies-all` command to update lock files. This command relies on uv and pixi to update lock files for all python versions. (if you want to avoid pixi dependency, you can still run individual commands for each python env) Note that **it's important that we all switch to using uv** for lock file updates as lock files generated aren't 100% identical to pip-tools generates files in terms of formatting. (see diffs in this PR for an example, the biggest difference is that uv doesn't add `feast (setup.py)` comment in the files) If some people still keep using pip-tools and others switch to uv for locking, we will generate unnecessary formatting diffs in lock files.


docs/project/development-guide.md
```diff
@@ -123,43 +123,43 @@ Note that this means if you are midway through working through a PR and rebase,
...
+```
+6. (Optional): Install Node & Yarn. Then run the following to build Feast UI artifacts for use in `feast ui`
 ```
 make build-ui
 ```
-8. Install mysql (needed for ci dependencies)
+7. (Optional) install pixi
+pixi is necessary to run step 8 for all python versions at once.
 ```sh
-brew install mysql
```
franciscojavierarceo, line 148: We removed this?

docs/project/development-guide.md
```diff
@@ -123,43 +123,43 @@ Note that this means if you are midway through working through a PR and rebase,
...
-9. Install development dependencies for Feast Python SDK / CLI
+8. (Optional): Recompile python lock files
+If you make changes to requirements or simply want to update python lock files to reflect latest versioons.
 ```sh
-pip install -e ".[dev]"
-```
-
-This will allow the installed feast version to automatically reflect changes to your local development version of Feast without needing to reinstall everytime you make code changes.
-
-10. Compile the protubufs
```
franciscojavierarceo, line 157: Do we no longer need the compile the protos? if someone makes a Message change we will, no?

docs/project/development-guide.md
```diff
@@ -123,43 +123,43 @@ Note that this means if you are midway through working through a PR and rebase,
...
+make lock-python-dependencies-all
```
franciscojavierarceo, line 155: nice

docs/project/development-guide.md
tokoko: We do, it's just part of the `make install-python-ci-dependencies-uv` already

docs/project/development-guide.md
tokoko: This was necessary for `mysqlclient` which was removed in #3925. `pymysql` doesn't require any native dependencies.

jeremyary (approved):

PR #4170, opened 2024-05-02
feat: Enable other distance metrics for Vector DB and Update docs

franciscojavierarceo:

- Adds 3 other distance metrics configurable by the `get_online_documents` call based on [PGVector's documentation](https://github.com/pgvector/pgvector?tab=readme-ov-file#querying)
- Also updated docs to outlines more explicit steps on how to get started

# Fixes

### DIFF
--- a/sdk/python/feast/infra/online_stores/contrib/postgres.py
+++ b/sdk/python/feast/infra/online_stores/contrib/postgres.py
@@ -22,10 +22,10 @@ from feast.repo_config import RepoConfig
 from feast.usage import log_exceptions_and_usage
 
 SUPPORTED_DISTANCE_METRICS_DICT = {
-    "L2": "<->",
-    "inner_product": "<#>",
     "cosine": "<=>",
     "L1": "<+>",
+    "L2": "<->",
+    "inner_product": "<#>",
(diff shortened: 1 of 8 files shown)


sdk/python/feast/infra/online_stores/contrib/postgres.py
```diff
@@ -21,6 +21,13 @@
 from feast.repo_config import RepoConfig
 from feast.usage import log_exceptions_and_usage
 
+SUPPORTED_DISTANCE_METRICS_DICT = {
```
HaoXuAI, line 24: that's cool!

sdk/python/feast/infra/provider.py
```diff
@@ -303,6 +303,7 @@ def retrieve_online_documents(
         requested_feature: str,
         query: List[float],
         top_k: int,
+        distance_metric: str = "euclidean",
```
HaoXuAI, line 306: this might be "L2"?

franciscojavierarceo: Nice catch

HaoXuAI (approved):

PR #4177, opened 2024-05-06
fix: Make sure schema is used when generating `from_expression` for Snowflake

TomSteenbergen:

This PR ensures that the `from_expression` used in the Snowflake offline store is generated correctly when the datasource already specifies the schema name. 

Currently, the schema name of the Snowflake data source is disregarded when generating the `from_expression`. So if the data source already specifies a schema, an invalid `from_expression` is generated where both the schema of the offline store config as well as the schema on the data source are used to construct the `from_expression`. 

# Fixes


franciscojavierarceo (approved):
thank you!

TomSteenbergen:
@franciscojavierarceo The integration tests are failing unfortunately with the following error:

```
botocore.exceptions.ClientError: An error occurred (AccessDenied) when calling the PutObject operation: Access Denied
```

I'm happy to rebase the branch to see if that fixes it, but seems like the AWS credentials might not be passed correctly to the build?

tokoko:
@TomSteenbergen should be fixed now, can you rebase?

PR #4183, opened 2024-05-07
fix: Update master-only benchmark bucket name due to credential update

jeremyary:

Update bucket name due to global resource name collisions with old AWS acct. in order to fix master-only test runs failing as seen [here](https://github.com/feast-dev/feast/actions/runs/8989121663/job/24693009473#step:17:29).


tokoko (approved):
+1 for creative naming 😄

franciscojavierarceo (approved):
👍

PR #4188, opened 2024-05-09
feat: Elasticsearch vector database

HaoXuAI:
- Add ElasticSearch in the online store as another vector database.
- Adjust the `retrieve_online_document` API

# Fixes

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -1550,6 +1550,54 @@ class FeatureStore:
             native_entity_values=True,
         )
 
+    @log_exceptions_and_usage
+    async def get_online_features_async(
+        self,
+        features: Union[List[str], FeatureService],
+        entity_rows: List[Dict[str, Any]],
+        full_feature_names: bool = False,
+    ) -> OnlineResponse:
+        """
+        [Alpha] Retrieves the latest online feature data asynchronously.
+
+        Note: This method will download the full feature registry the first time it is run. If you are using a
+        remote registry like GCS or S3 then that may take a few seconds. The registry remains cached up to a TTL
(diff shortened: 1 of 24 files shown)


sdk/python/feast/infra/online_stores/contrib/elastichsearch.py
```diff
@@ -0,0 +1,221 @@
...
+
+from feast import Entity, FeatureView, RepoConfig
+from feast.infra.key_encoding_utils import get_list_val_str, serialize_entity_key
+from feast.infra.online_stores.online_store import OnlineStore
+from feast.protos.feast.types.EntityKey_pb2 import EntityKey as EntityKeyProto
+from feast.protos.feast.types.Value_pb2 import Value as ValueProto
+from feast.repo_config import FeastConfigBaseModel
+
+
+class ElasticsearchOnlineStoreConfig(FeastConfigBaseModel):
```
franciscojavierarceo, line 17: nice

sdk/python/feast/feature_store.py
```diff
@@ -1886,7 +1886,7 @@ def retrieve_online_documents(
         feature: str,
         query: Union[str, List[float]],
         top_k: int,
-        distance_metric: str,
+        distance_metric: Optional[str] = None,
```
HaoXuAI, line 1889: have to make it optional as `elasticsearch` doesn't allow specifying the metric in the online API. Instead, it has to update the index.

franciscojavierarceo (approved):
Nice 🚀🚀🚀

sdk/python/feast/feature_store.py
franciscojavierarceo: Makes sense can we add some explicit tests for this to highlight the behavior? Also this is awesome.

Issue #4133, opened 2024-04-23
Feature View's ttl is misleading


breno-costa:
**Is your feature request related to a problem? Please describe.**
The field name `ttl` used in `FeatureView` is misleading. 
Time to Live (TTL) is a well known term that defines a mechanism to automatically expire database records. In Redis, for example, the command `EXPIRE` defines a time to live for a given key. 

When I used that FeatureView.ttl configuration for the first time, I expected to set some expiration time (or TTL) for keys written into online stores. The documentation states something like this [here](https://github.com/breno-costa/feast/blob/master/sdk/python/feast/feature_view.py#L63-L65).

> The amount of time this group of features lives. A ttl of 0 indicates that this group of features lives forever. 

However FeatureView's TTL is not used in that way. It's used to define `start_date` to extract data from offline store during  materialization step [here](https://github.com/breno-costa/feast/blob/master/sdk/python/feast/feature_store.py#L1246). The actual key ttl configuration used to expire entities is `key_ttl_seconds` at online store level [here](https://github.com/breno-costa/feast/blob/master/sdk/python/feast/infra/online_stores/redis.py#L77).

I had to explain this sometimes to different people because of this misunderstanding. I think this is also kind of mentioned [here](https://github.com/feast-dev/feast/issues/3435) and [here](https://github.com/feast-dev/feast/issues/3596).

**Describe the solution you'd like**
My suggestion is renaming this field to something like `materialization_lookback_interval`, `lookback_interval` or `lookback_delta` since this is the actual meaning of this configuration as far as I understand. Also update docs accordingly.

franciscojavierarceo:
I need to spend more time thinking about this but I do agree the `ttl` at the `FeatureView` level is misleading as I had this exact experience in my last role and it caused me some headaches. I think renaming it to follow industry conventions would be good.

tokoko:
I'm not a big fan of a `lookback` name tbh. I think it only makes sense if online store is the only component that we're focusing on. If we set aside online stores for a minute, during `get_historical_features` call the way offline store handles the rows sounds compatible to the meaning of `ttl` to me. When a client askს for a feature value for a specific point in time, only the values that haven't expired at that time are considered. The fact that it's doing a "lookback" to do that is an implementation detail.

[...]

PR #4189, opened 2024-05-09
feat: Set optional full-scan for deletion

phil-park:

In this link (https://github.com/feast-dev/feast/pull/3857), I added function to perform a full-scan and delete data to save Redis memory. However, full-scan in Redis does not work beautifully, takes too long at production scale, and often fails in CI that performs feast-apply. So, I think it would be a good idea to add the 'full_scan_for_deletion' option in redis so that the user can select and delete it.

# Fixes

### DIFF
--- a/sdk/python/feast/infra/online_stores/redis.py
+++ b/sdk/python/feast/infra/online_stores/redis.py
@@ -77,7 +77,7 @@ class RedisOnlineStoreConfig(FeastConfigBaseModel):
     key_ttl_seconds: Optional[int] = None
     """(Optional) redis key bin ttl (in seconds) for expiring entities"""
 
-    full_scan_for_deletion: Optional[bool] = False
+    full_scan_for_deletion: Optional[bool] = True
     """(Optional) whether to scan for deletion of features"""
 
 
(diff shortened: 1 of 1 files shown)


tokoko:
I think this might be valuable in the interim, although you might want to chime in the discussion here as well #4133. The implicit consensus from there is to add a `feast cleanup` command that will delete all redundant keys from online store (keys that belong to deleted tables and also keys containing feature values that have expired according to ttl).

sdk/python/feast/infra/online_stores/redis.py
```diff
@@ -160,9 +163,13 @@ def update(
             entities_to_keep: Entities to keep
             partial: Whether to do a partial update
         """
+        online_store_config = config.online_store
+
+        assert isinstance(online_store_config, RedisOnlineStoreConfig)
 
-        for table in tables_to_delete:
-            self.delete_table(config, table)
+        if online_store_config.full_scan_for_deletion:
```
breno-costa, line 170: the old implementation deletes tables by default.
should `full_scan_for_deletion`'s default value be `True`?

phil-park: As you said, I set default to True for compatibility with older implementations.
If compatibility is not enforced, I think False would be better.

phil-park: @breno-costa Changed to default True to ensure backward compatibility

tokoko (approved):
lgtm

franciscojavierarceo (approved):

PR #4201, opened 2024-05-14
fix: Integration tests for async sdk method

breno-costa:
Follow up on https://github.com/feast-dev/feast/pull/4172 After checking integration tests, I understand this one is relevant to the async method. Let me know if you understand more integration tests are needed.

# Fixes

### DIFF
--- a/sdk/python/tests/integration/online_store/test_universal_online.py
+++ b/sdk/python/tests/integration/online_store/test_universal_online.py
@@ -13,6 +13,7 @@ import pytest
 import requests
 from botocore.exceptions import BotoCoreError
 
+from feast import FeatureStore
 from feast.entity import Entity
 from feast.errors import FeatureNameCollisionError
 from feast.feature_service import FeatureService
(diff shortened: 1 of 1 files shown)


tokoko:
I wonder it there's a way to do this without introducing another test. We could eventually use `@pytest.mark.parametrize("use_async_client", [True, False], ids=lambda v: str(v))` on existing tests, but not sure how would we be able to right now limit async tests to only redis (without also limiting sync ones).

breno-costa:
> I wonder it there's a way to do this without introducing another test. We could eventually use `@pytest.mark.parametrize("use_async_client", [True, False], ids=lambda v: str(v))` on existing tests, but not sure how would we be able to right now limit async tests to only redis (without also limiting sync ones).

I'm not sure whether it's possible to configure tests in this way. I refactored a little bit to reuse same setup and assertions for both sync and async sdk methods. This may help. wdyt?

tokoko:
thanks, this looks a lot better.

tokoko (approved):
lgtm

tokoko:
@breno-costa Can you take a look here? you can ignore `test_historical_features` fails, but others seem legit.

breno-costa:
> @breno-costa Can you take a look here? you can ignore `test_historical_features` fails, but others seem legit.

Fixed the issue. 
Would it be possible to run integration tests locally?

tokoko:
@breno-costa thanks. There's a `test-python-integration-local` command in Makefile, It's used to run integration tests except for the ones that depend on AWS/GCP accounts.

franciscojavierarceo (approved):

Issue #4204, opened 2024-05-15
Add optional params to repo config for Snowflake private key authentication


ArijeetC:
**Is your feature request related to a problem? Please describe.**
When using private key authentication for Snowflake, the private key path and passphrase params must be defined in a config file and then the config file path must be added in the `feature_store.yaml` file (https://github.com/feast-dev/feast/issues/2482#issuecomment-1303882805).

Instead of creating a separate config INI file just for these 2 params, they could be simply added in the `feature_store.yaml` file along with the other params.

**Describe the solution you'd like**
Adding optional params `private_key` and `private_key_passphrase` to the Snowflake config classes (`SnowflakeRegistryConfig`, `SnowflakeOnlineStoreConfig`, `SnowflakeOfflineStoreConfig`, `SnowflakeMaterializationEngineConfig`) so that they can be used in `feature_store.yaml` file directly.

**Describe alternatives you've considered**
None

PR #4205, opened 2024-05-15
feat: Add optional private key params to Snowflake config

ArijeetC:

Adding optional params `private_key` and `private_key_passphrase` to the Snowflake config classes so that they can be used in `feature_store.yaml` file directly.

Fixes https://github.com/feast-dev/feast/issues/4204

### DIFF
--- a/sdk/python/feast/infra/materialization/snowflake_engine.py
+++ b/sdk/python/feast/infra/materialization/snowflake_engine.py
@@ -67,6 +67,12 @@ class SnowflakeMaterializationEngineConfig(FeastConfigBaseModel):
     authenticator: Optional[str] = None
     """ Snowflake authenticator name """
 
+    private_key: Optional[str] = None
+    """ Snowflake private key file path"""
+
+    private_key_passphrase: Optional[str] = None
+    """ Snowflake private key file passphrase"""
+
     database: StrictStr
     """ Snowflake database name """
 
--- a/sdk/python/feast/infra/offline_stores/snowflake.py
+++ b/sdk/python/feast/infra/offline_stores/snowflake.py
@@ -105,6 +105,12 @@ class SnowflakeOfflineStoreConfig(FeastConfigBaseModel):
     authenticator: Optional[str] = None
     """ Snowflake authenticator name """
 
+    private_key: Optional[str] = None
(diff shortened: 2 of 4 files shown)


franciscojavierarceo:
@ArijeetC can you fix the DCO? You can do it in the UI.

ArijeetC:
@franciscojavierarceo fixed it

franciscojavierarceo (approved):
Lgtm

franciscojavierarceo:
Looks like we have some errors. I think Jeremy is working on this.

tokoko:
Shouldn't these keys be used anywhere? What's the point of adding them?

ArijeetC:
> What's the point of adding them?

@tokoko because it is convenient to have all config defined at a single place (`feature_store.yaml`). Also since these params are optional, I don't think they will cause any problems.

tokoko:
I got that part but the PR is just adding new params, not using them in any way. snowflake connection will be acquired just like before because of [this](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/utils/snowflake/snowflake_utils.py#L72), no?

ArijeetC:
> I got that part but the PR is just adding new params, not using them in any way. snowflake connection will be acquired just like before because of [this](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/utils/snowflake/snowflake_utils.py#L72), no?

Oh, got your question now.

These new params will already be present inside `config_dict` [here](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/utils/snowflake/snowflake_utils.py#L68) if they have been defined in the `feature_store.yaml` and they will be added to `kwargs` [here](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/utils/snowflake/snowflake_utils.py#L77), that's why no other code change is required.

The `feature_store.yaml` entry I used for testing:
```
offline_store:
    type: snowflake.offline
    user: <...>
    database: <...>
    account: <...>
    private_key: <...>
    private_key_passphrase: <...>
```

tokoko:
@ArijeetC okay, now it's clear. btw, you will have to rebase on master to get rid of it failures.

ArijeetC:
@tokoko done

PR #4206, opened 2024-05-16
chore: Skip test_historical_features_main for Snowflake

tokoko:
Temporarily skips failing `test_historical_features_main` test (had to change the name to correctly apply a pytest expression) for snowflake.


jeremyary:
Thanks Tornike! Added note for others since this has mostly played out in slack - it's going to take me a little time to resolve the Snowflake issue with DevOps/AWS admins, so this should unblock PR's/nightly while we work out it out.

jeremyary (approved):

PR #4214, opened 2024-05-21
chore: Remove usage tracking

tokoko:
This PR removes usage tracking code entirely from feast codebase and tests. We already made it as an opt-in feature previously, which probably no one is going to opt in for :)


sdk/python/feast/batch_feature_view.py
```diff
@@ -60,7 +60,7 @@ def __init__(
         *,
         name: str,
         source: DataSource,
-        entities: Optional[Union[List[Entity], List[str]]] = None,
```
franciscojavierarceo, line 63: This seems to be u related to usage fyi

tokoko: I know, after I removed decorators, a lot of linter rules started to fail and I had to fix a bunch of them (~20). Probably something with mypy and decorators not getting along, not sure why it wasn't an issue before.

franciscojavierarceo: Nice, can you add that to the description and PR notes?

franciscojavierarceo (approved):
Nice!

PR #4217, opened 2024-05-21
feat: Added deadline to gRPC Java client

acevedosharp:
Added configurable deadlines for the Feast Java gRPC client. We need it because clients could otherwise [get stuck forever](https://grpc.io/docs/guides/deadlines/) waiting for the feature server to respond, and we should make it simpler for users of this library to implement retries and timeouts; especially when it's already supported in gRPC by default.

N/A

# Fixes
N/A

### DIFF
--- a/java/serving-client/src/main/java/dev/feast/FeastClient.java
+++ b/java/serving-client/src/main/java/dev/feast/FeastClient.java
@@ -26,6 +26,7 @@ import feast.proto.serving.ServingServiceGrpc;
 import feast.proto.serving.ServingServiceGrpc.ServingServiceBlockingStub;
 import feast.proto.types.ValueProto;
 import io.grpc.CallCredentials;
+import io.grpc.Deadline;
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
 import io.grpc.netty.shaded.io.grpc.netty.GrpcSslContexts;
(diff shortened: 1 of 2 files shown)


java/serving-client/src/test/java/dev/feast/FeastClientTest.java
```diff
@@ -86,7 +87,7 @@ public void setup() throws Exception {
     ManagedChannel channel =
         this.grpcRule.register(
             InProcessChannelBuilder.forName(serverName).directExecutor().build());
-    this.client = new FeastClient(channel, Optional.empty());
+    this.client = new FeastClient(channel, Optional.empty(), Deadline.after(200, TimeUnit.MILLISECONDS));
```
franciscojavierarceo, line 90: nit: can you make this a field defined at the top of the class?

franciscojavierarceo (changes requested):

java/serving-client/src/main/java/dev/feast/FeastClient.java
```diff
@@ -57,9 +58,9 @@ public class FeastClient implements AutoCloseable {
    * @param port port number of Feast serving GRPC server
    * @return {@link FeastClient}
    */
-  public static FeastClient create(String host, int port) {
+  public static FeastClient create(String host, int port, Deadline deadline) {
```
tokoko, line 61: Probably a good idea to make this backwards-compatible and provide some default value here as well.

java/serving-client/src/test/java/dev/feast/FeastClientTest.java
acevedosharp: ✅

java/serving-client/src/main/java/dev/feast/FeastClient.java
acevedosharp: ✅

franciscojavierarceo (approved):

PR #4127, opened 2024-04-22
chore: Incorporate release 0.37.1 to master

jeremyary:
### Bug Fixes

* Pgvector patch ([#4108](https://github.com/feast-dev/feast/issues/4108)) ([1a1f0b1](https://github.com/feast-dev/feast/commit/1a1f0b1c56aa2ac00b1e1aa1e21cc200ea659334))

### Reverts

* Reverts "fix: Using version args to install the correct feast version" ([#4112](https://github.com/feast-dev/feast/issues/4112)) ([d5ded69](https://github.com/feast-dev/feast/commit/d5ded69dea9af3a363feaa948cd3d2dcf10fb80c)), closes [#3953](https://github.com/feast-dev/feast/issues/3953)

# What this PR does / why we need it:

# Which issue(s) this PR fixes:

# Fixes

PR #4221, opened 2024-05-23
chore: Adjusting the file version validation for minor releases

franciscojavierarceo:

This is changing the validation to ignore patch version changes. So the CI should trigger correctly for increments from `0.37.1` -> `0.38.0`.

```bash
python infra/scripts/release/bump_file_versions.py 0.37.0 0.38.0
Updated 16 files with new version 0.38.0
```

Previously this failed as `0.37.1` wasn't recognized as a tag.

# Fixes

### DIFF
--- a/infra/scripts/release/bump_file_versions.py
+++ b/infra/scripts/release/bump_file_versions.py
@@ -1,5 +1,6 @@
 # This script will bump the versions found in files (charts, pom.xml) during the Feast release process.
 
+import re
 import pathlib
 import sys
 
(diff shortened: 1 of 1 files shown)


sudohainguyen (approved):
lgtm 😄

tokoko (commented):
Despite validation, this is still not going to work, as line 48 looks for the passed `current_version` to replace in files. I think much easier solution is to simply revert version bumps.

franciscojavierarceo:
> Despite validation, this is still not going to work, as line 48 looks for the passed `current_version` to replace in files. I think much easier solution is to simply revert version bumps.

The version bumps are automatically generated through this code. Are you suggesting we do that for every patch release @tokoko?

tokoko:
@franciscojavierarceo No, version bumps to 0.37.1 was automatic, but not on master, it was done on `v-0.37-branch`. We manually introduced those version bumps to master with a later PR (#4127) which was a mistake. I don't mind if we scrap validation from the scripts either, just pointing out that it's unnecessary if we simply revert some of the changes from that PR (version bumps specifically).

franciscojavierarceo:
> We manually introduced those version bumps to master with a later PR (#4127) which was a mistake. 

Got it! I missed this, my bad. Lack of sleep 🥲 .

franciscojavierarceo:
@tokoko https://github.com/feast-dev/feast/pull/4227

Issue #4121, opened 2024-04-19
Add `remote-http` online store


tokoko:
**Is your feature request related to a problem? Please describe.**
Users can deploy http and grpc feature servers that wrap `get_online_features` calls, but currently there is no way to interact with feature servers from python sdk.

**Describe the solution you'd like**
Add online store types `remote-http` and `remote-grpc` (or maybe just call it a `remote` and differentiate with a url?) that wrap network calls to the respective feature servers. This follows the same pattern as [remote](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/registry/remote.py) registry implementation that wraps calls to the registry server.

lokeshrangineni:
I am currently working on this and will get back in case if i find any difficulty.

lokeshrangineni:
@franciscojavierarceo FYI

franciscojavierarceo:
Some things we'll have to do:

1. Handle purely `feature_views` i.e., `features="feature_view"`
2. Handle feature references; i.e., `["feature_view:feature_1","feature_view:feature_2",...]` etc.
3. Handle [feature services ](https://docs.feast.dev/getting-started/concepts/feature-retrieval#feature-services)

We'll have to serialize the data on the server to match what's in the feature registry from the `FeatureView` itself and fail loudly when it doesn't match, probably `400`.

tokoko:
@franciscojavierarceo I think you closed this by accident.

> * Handle purely `feature_views` i.e., `features="feature_view"`

Is this something we support? The signature for `get_online_features` is only feature service or a list of features -> `features: Union[List[str], FeatureService]`

franciscojavierarceo:
Whoops, nevermind that was based on outdated documentation I think.

franciscojavierarceo:
@tokoko I'm confused about why we would do this? I think adding a client that makes the behavior intuitive makes sense but that's basically in the [Python feature server](https://github.com/feast-dev/feast/blob/ee51fbfe3f40e637084c5ecf51f7463b840d58b4/sdk/python/feast/feature_server.py#L24).

tokoko:
Well, it's more of convenience thing, i think. They can still use the feature server directly of course, but what this additional layer would enable is writing a piece of code using `store.get_online_features` and then configuring how you want to resolve features with `feature_store.yaml`. You could change the backend from the actual online store to http feature server or a grpc one without changing anything in the code itself.

lokeshrangineni:
[...]

PR #4226, opened 2024-05-24
feat: Feature/adding remote online store

lokeshrangineni:
Adds the new online store of type "remote" which will interact with remote feature server. Able to read and write the features only at the moment. 

https://github.com/feast-dev/feast/issues/4121

### DIFF
--- a/docs/reference/online-stores/remote.md
+++ b/docs/reference/online-stores/remote.md
@@ -1,0 +1,21 @@
+# Remote online store
+
+## Description
+
+This remote online store will let you interact with remote feature server. At this moment this only supports the read operation. You can use this online store and able retrieve online features `store.get_online_features`  from remote feature server. 
+
+## Examples
+
+The registry is pointing to registry of remote feature store. If it is not accessible then should be configured to use remote registry.
+
+{% code title="feature_store.yaml" %}
+```yaml
+project: my-local-project
+  registry: /remote/data/registry.db
+  provider: local
+  online_store:
+    path: http://localhost:6566
+    type: remote
+  entity_key_serialization_version: 2
+```
+{% endcode %}
--- a/sdk/python/feast/infra/online_stores/remote.py
+++ b/sdk/python/feast/infra/online_stores/remote.py
@@ -1,0 +1,167 @@
+# Copyright 2021 The Feast Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import json
+import logging
+from datetime import datetime
+from typing import Any, Callable, Dict, List, Literal, Optional, Sequence, Tuple
+
+import requests
+from pydantic import StrictStr
+
+from feast import Entity, FeatureView, RepoConfig
+from feast.infra.online_stores.online_store import OnlineStore
+from feast.protos.feast.types.EntityKey_pb2 import EntityKey as EntityKeyProto
+from feast.protos.feast.types.Value_pb2 import Value as ValueProto
+from feast.repo_config import FeastConfigBaseModel
+from feast.type_map import python_values_to_proto_values
+from feast.value_type import ValueType
+
+logger = logging.getLogger(__name__)
+
+
+class RemoteOnlineStoreConfig(FeastConfigBaseModel):
+    """Remote Online store config for remote online store"""
+
+    type: Literal["remote"] = "remote"
+    """Online store type selector"""
+
+    path: StrictStr = "http://localhost:6566"
+    """ str: Path to metadata store.
+    If type is 'remote', then this is a URL for registry server """
+
+
+class RemoteOnlineStore(OnlineStore):
+    """
+    remote online store implementation wrapper to communicate with feast online server.
+    """
+
+    def online_write_batch(
+        self,
+        config: RepoConfig,
+        table: FeatureView,
+        data: List[
+            Tuple[EntityKeyProto, Dict[str, ValueProto], datetime, Optional[datetime]]
+        ],
+        progress: Optional[Callable[[int], Any]],
+    ) -> None:
+        raise NotImplementedError
+
+    def online_read(
+        self,
+        config: RepoConfig,
+        table: FeatureView,
+        entity_keys: List[EntityKeyProto],
+        requested_features: Optional[List[str]] = None,
+    ) -> List[Tuple[Optional[datetime], Optional[Dict[str, ValueProto]]]]:
+        assert isinstance(config.online_store, RemoteOnlineStoreConfig)
+        config.online_store.__class__ = RemoteOnlineStoreConfig
+
+        req_body = self._construct_online_read_api_json_request(
+            entity_keys, table, requested_features
+        )
+        response = requests.post(
+            f"{config.online_store.path}/get-online-features", data=req_body
+        )
+        if response.status_code == 200:
+            logger.debug("Able to retrieve the online features from feature server.")
+            response_json = json.loads(response.text)
+            event_ts = self._get_event_ts(response_json)
+            # Iterating over results and converting the API results in column format to row format.
+            result_tuples: List[
+                Tuple[Optional[datetime], Optional[Dict[str, ValueProto]]]
+            ] = []
+            for feature_value_index in range(len(entity_keys)):
+                feature_values_dict: Dict[str, ValueProto] = dict()
+                for index, feature_name in enumerate(
+                    response_json["metadata"]["feature_names"]
+                ):
+                    if (
+                        requested_features is not None
+                        and feature_name in requested_features
+                    ):
+                        if (
+                            response_json["results"][index]["statuses"][
+                                feature_value_index
+                            ]
+                            == "PRESENT"
+                        ):
+                            message = python_values_to_proto_values(
+                                [
+                                    response_json["results"][index]["values"][
+                                        feature_value_index
+                                    ]
+                                ],
+                                ValueType.UNKNOWN,
+                            )
+                            feature_values_dict[feature_name] = message[0]
+                        else:
+                            feature_values_dict[feature_name] = ValueProto()
+                result_tuples.append((event_ts, feature_values_dict))
+            return result_tuples
+        else:
+            error_msg = f"Unable to retrieve the online store data using feature server API. Error_code={response.status_code}, error_message={response.reason}"
+            logger.error(error_msg)
+            raise RuntimeError(error_msg)
+
+    def _construct_online_read_api_json_request(
+        self,
+        entity_keys: List[EntityKeyProto],
+        table: FeatureView,
+        requested_features: Optional[List[str]] = None,
+    ) -> str:
+        api_requested_features = []
+        if requested_features is not None:
+            for requested_feature in requested_features:
+                api_requested_features.append(f"{table.name}:{requested_feature}")
+
+        entity_values = []
+        entity_key = ""
+        for row in entity_keys:
+            entity_key = row.join_keys[0]
+            entity_values.append(
+                getattr(row.entity_values[0], row.entity_values[0].WhichOneof("val"))
+            )
+
+        req_body = json.dumps(
+            {
+                "features": api_requested_features,
+                "entities": {entity_key: entity_values},
+            }
+        )
+        return req_body
+
+    def _get_event_ts(self, response_json) -> datetime:
+        event_ts = ""
+        if len(response_json["results"]) > 1:
+            event_ts = response_json["results"][1]["event_timestamps"][0]
+        return datetime.fromisoformat(event_ts.replace("Z", "+00:00"))
+
+    def update(
+        self,
+        config: RepoConfig,
+        tables_to_delete: Sequence[FeatureView],
+        tables_to_keep: Sequence[FeatureView],
+        entities_to_delete: Sequence[Entity],
+        entities_to_keep: Sequence[Entity],
+        partial: bool,
+    ):
+        pass
+
+    def teardown(
+        self,
+        config: RepoConfig,
+        tables: Sequence[FeatureView],
+        entities: Sequence[Entity],
+    ):
+        pass
--- a/sdk/python/tests/conftest.py
+++ b/sdk/python/tests/conftest.py
@@ -32,8 +32,8 @@ from tests.data.data_creator import (  # noqa: E402
     create_basic_driver_dataset,
     create_document_dataset,
 )
-from tests.integration.feature_repos.integration_test_repo_config import (  # noqa: E402
-    IntegrationTestRepoConfig,
+from tests.integration.feature_repos.integration_test_repo_config import (
+    IntegrationTestRepoConfig,  # noqa: E402
 )
 from tests.integration.feature_repos.repo_configuration import (  # noqa: E402
     AVAILABLE_OFFLINE_STORES,
@@ -45,8 +45,8 @@ from tests.integration.feature_repos.repo_configuration import (  # noqa: E402
     construct_universal_feature_views,
     construct_universal_test_data,
 )
-from tests.integration.feature_repos.universal.data_sources.file import (  # noqa: E402
-    FileDataSourceCreator,
+from tests.integration.feature_repos.universal.data_sources.file import (
+    FileDataSourceCreator,  # noqa: E402
 )
 from tests.integration.feature_repos.universal.entities import (  # noqa: E402
     customer,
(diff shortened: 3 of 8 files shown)


lokeshrangineni:
cc: @tokoko @dmartinol

sdk/python/tests/integration/feature_repos/universal/online_store/remote.py
```diff
@@ -0,0 +1,89 @@
...
+from tests.integration.feature_repos.universal.online_store_creator import (
+    OnlineStoreCreator,
+)
+
+class RemoteOnlineStoreCreator(OnlineStoreCreator):
+    _FEAST_SERVER_TIMEOUT_SEC:int = 30
+    def __init__(self, project_name: str, **kwargs):
+        super().__init__(project_name)
+        
+        for key, value in kwargs.items():
```
tokoko, line 18: Not a deal-breaker, but we should probably change `OnlineStoreCreator` similar to `DataSourceCreator`, meaning add a `setup` method that's called with a registry config.

sdk/python/tests/integration/feature_repos/repo_configuration.py
```diff
@@ -153,6 +160,7 @@
     AVAILABLE_ONLINE_STORES["datastore"] = ("datastore", None)
     AVAILABLE_ONLINE_STORES["snowflake"] = (SNOWFLAKE_CONFIG, None)
     AVAILABLE_ONLINE_STORES["bigtable"] = (BIGTABLE_CONFIG, None)
+    AVAILABLE_ONLINE_STORES["remote"] = (REMOTE_ONLINE_CONFIG, None)
```
tokoko, line 163: Is this necessary? `remote` should already be in the list from above

sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -0,0 +1,180 @@
...
+                "push_source_name": table.name,
+                "df": json_df,
+                "to": "online"
+            }
+        
+        return req_body
+
+    def _check_if_feature_requested(self, feature_name, requested_features):
+        for requested_feature in requested_features:
+            if feature_name in requested_feature:
```
tokoko, line 150: Isn't this a substring check? Shouldn't a simple equality check be sufficient? Or maybe there's an edge case because of `full_feature_names`?

sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -0,0 +1,180 @@
...
+        headers = {'Content-type': 'application/json', 'Accept': 'text/plain'}     
+        response = requests.post(f'{config.online_store.path}/push', data=json.dumps(req_body), headers=headers)
+        if response.status_code == 200:
+            logger.debug("Able to write the online features from the feature server.")
+        else:
+            error_message = f"Unable to write the data to online store using feature server API. Error_code={response.status_code}, error_message={response.reason}"
+            logger.error(error_message)
+            raise RuntimeError(error_message)
+
+    def online_read(
```
tokoko, line 72: I'm realizing we will probably have the same problem with `FeatureService`s here as in offline server discussed [here](https://github.com/RHEcosystemAppEng/feast/pull/3#issuecomment-2127894618). cc @dmartinol 

This is also probably a very chatty protocol with a list of features always transmitted on the wire even if feature service was requested.

sdk/python/feast/infra/online_stores/remote.py
lokeshrangineni: Yes, that is correct. Just to incorporate when full_feature_names is used. Please let me know if there is any better way.

sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -0,0 +1,181 @@
...
+from feast.protos.feast.types.Value_pb2 import Value as ValueProto
+from feast.repo_config import FeastConfigBaseModel
+from feast.type_map import python_values_to_proto_values
+from feast.value_type import ValueType
+
+logger = logging.getLogger(__name__)
+
+
+class RemoteOnlineStoreConfig(FeastConfigBaseModel):
+    """Online store config for Redis store"""
```
dmartinol, line 34: Redis?

sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -0,0 +1,181 @@
...
+    def online_write_batch(
+        self,
+        config: RepoConfig,
+        table: FeatureView,
+        data: List[
+            Tuple[EntityKeyProto, Dict[str, ValueProto], datetime, Optional[datetime]]
+        ],
+        progress: Optional[Callable[[int], Any]],
+    ) -> None:
+        pass
```
dmartinol, line 58: Is this a `pass` or `NotImplementedError`?

sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -0,0 +1,181 @@
...
+        if response.status_code == 200:
+            logger.debug("Able to retrieve the online features from feature server.")
+            response_json = json.loads(response.text)
+            event_ts = self._get_event_ts(response_json)
+            # Iterating over results and converting the API results in column format to row format.
+            feature_value_index = 0
+            result_tuples: List[
+                Tuple[Optional[datetime], Optional[Dict[str, ValueProto]]]
+            ] = []
+            while feature_value_index < len(entity_keys):
```
dmartinol, line 85: `for feature_value_index in range(len(entity_keys))` ?

sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -0,0 +1,181 @@
...
+                feature_values_dict = dict()
```
dmartinol, line 86: shouldn't you also specify the type?

sdk/python/feast/infra/online_stores/remote.py
dmartinol: this returns True if we feature name is a substring of the requested feature, it can return false positives IMO.

sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -0,0 +1,181 @@
...
+        for requested_feature in requested_features:
+            if feature_name in requested_feature:
+                return True
+        return False
+
+    def _get_event_ts(self, response_json) -> datetime:
+        event_ts = ""
+        for index, result in enumerate(response_json["results"]):
+            if index == 1:
+                event_ts = result["event_timestamps"][0]
```
dmartinol, line 159: isn't this equivalent to the following?
```py
# maybe check len of first array
event_ts = response_json["results"][1]["event_timestamp"][0]
```

sdk/python/tests/integration/feature_repos/repo_configuration.py
```diff
@@ -509,6 +498,13 @@ def construct_test_environment(
             cache_ttl_seconds=1,
         )
 
+    if test_repo_config.online_store_creator:
```
dmartinol, line 501: why did you move it down?

sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -0,0 +1,181 @@
...
+            ] = []
+            while feature_value_index < len(entity_keys):
+                feature_values_dict = dict()
+                for index, feature_name in enumerate(
+                    response_json["metadata"]["feature_names"]
+                ):
+                    if self._check_if_feature_requested(
+                        feature_name, requested_features
+                    ):
+                        is_present = (
```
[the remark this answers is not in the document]
dmartinol: and you probably don't even need the `is_present` variable, but 🤷‍♂️

sdk/python/feast/infra/online_stores/remote.py
lokeshrangineni: just trying to avoid the array index exception, i can check the length of the array before instead of iterating over it.

sdk/python/feast/infra/online_stores/remote.py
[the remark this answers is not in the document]
lokeshrangineni: this logic is evolved and i didn't realize that it is not needed anymore. i have removed it. thanks for pointing it out.

sdk/python/tests/integration/feature_repos/repo_configuration.py
lokeshrangineni: it is not needed to move now, initially made some changes to incorporate that logic into the creator framework at that time there was some dependency but later that plan was dropped in my case.

sdk/python/feast/infra/online_stores/remote.py
lokeshrangineni: i don't want to through the hard stop as this life cycle method was used in some of the providers so used pass instead of throwing exception. please let me know in case if you think it should be other way.

dmartinol: @tokoko what’s the preferred approach to ensure application consistency?

docs/reference/online-stores/remote.md
```diff
@@ -0,0 +1,21 @@
...
+
+The registry is pointing to registry of remote feature store. If it is not accessible then should be configured to use remote registry.
+
+{% code title="feature_store.yaml" %}
+```yaml
+project: my-local-project
+  registry: /remote/data/registry.db
+  provider: local
+  online_store:
+    path: http://localhost:6566
```
dmartinol, line 17: shouldn't we have host and port fields instead? this is how we configured the remote offline:
```
offline_store:
    type: remote
    host: localhost
    port: 8815
```

lokeshrangineni: I have noticed this for offline_Store. We are already having path field for online store so reusing the same instead of introducing new properties. Remote registry is also doing the same. Also original recommendation by @tokoko but happy to change the code to keep it consistent across. 

https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/registry/remote.py#L30

sdk/python/feast/infra/online_stores/remote.py
lokeshrangineni: updated the code after performing some tests, it makes sense to use ==.

sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -0,0 +1,174 @@
...
+            result_tuples: List[
+                Tuple[Optional[datetime], Optional[Dict[str, ValueProto]]]
+            ] = []
+            for feature_value_index in range(len(entity_keys)):
+                feature_values_dict: Dict[str, ValueProto] = dict()
+                for index, feature_name in enumerate(
+                    response_json["metadata"]["feature_names"]
+                ):
+                    if (
+                        requested_features is not None
```
dmartinol, line 90: what if requested_features is None, instead? should we return all features or no features? (in which case, why did we call the remote API? 🤔 )

sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -0,0 +1,174 @@
...
+
+        req_body = json.dumps(
+            {
+                "features": api_requested_features,
+                "entities": {entity_key: entity_values},
+            }
+        )
+        return req_body
+
+    def _check_if_feature_requested(self, feature_name, requested_features):
```
dmartinol, line 145: Is this `_check_if_feature_requested` function still in use?

docs/reference/online-stores/remote.md
tokoko: I also simply avoided introducing new properties. Sounds more convenient as a single url to me, you always have some space to play with protocols and so on...

sdk/python/feast/infra/online_stores/remote.py
tokoko: None means all features

sdk/python/tests/integration/online_store/test_remote_online_store.py
```diff
@@ -0,0 +1,222 @@
+import os
+import tempfile
+from datetime import datetime
+from multiprocessing import Process
```
dmartinol, line 4: should we use Subprocess as it was recently done [here](https://github.com/feast-dev/feast/pull/4262/commits/dec05c91adf9ed7415961ad6bc90ee91b10227b3#diff-93e66e2dc557553d09ce1e0fc77b45003e1edcc8592979886970d5a878f731b2R398)?

sdk/python/feast/infra/online_stores/remote.py
dmartinol: so we should probably change to
```py
if requested_features is None or feature_name in requested_features:
```

tokoko: That feels like something that should have been part of integration tests already 😆

lokeshrangineni: it is already part of the integration test case, just didn't realize until exploring more details about full_feature_names flag.

sdk/python/tests/conftest.py
```diff
@@ -173,7 +173,15 @@ def simple_dataset_2() -> pd.DataFrame:
 
 def start_test_local_server(repo_path: str, port: int):
     fs = FeatureStore(repo_path)
-    fs.serve("localhost", port, no_access_log=True)
+    fs.serve(
```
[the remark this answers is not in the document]
lokeshrangineni: Yes, those are not defined as optional for feature_store.serve() method.

sdk/python/feast/infra/online_stores/remote.py
lokeshrangineni: removed it.

sdk/python/tests/integration/online_store/test_remote_online_store.py
lokeshrangineni: Implemented it. please take a look.

sdk/python/tests/conftest.py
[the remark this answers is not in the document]
tokoko: Other parameters have [default values](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/feature_store.py#L2514) set. You don't need to provide them, only `host` and `port` are mandatory.

sdk/python/tests/integration/online_store/test_remote_online_store.py
dmartinol: Lgtm, but please @redhatHameed can you verify it is consistent with the offline implementation?

sdk/python/tests/conftest.py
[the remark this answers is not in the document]
lokeshrangineni: @tokoko - I see that feature_store is modified [recently](https://github.com/feast-dev/feast/commit/fa744380ad48ee394a05b2f600db5fb0a99c03aa#diff-ff4e2a65c8c29b7971edc77a61a7e970d410d77d9f173f72714d1bb49e2509a2R2451) after i started working on this change. The default values have been added to original API so that I can revert this change.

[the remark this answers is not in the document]
lokeshrangineni: reverted this change, please take a look.

sdk/python/tests/integration/online_store/test_remote_online_store.py
redhatHameed: Yup looks lgtm that implementation

Makefile
```diff
@@ -377,6 +377,9 @@ kill-trino-locally:
 install-protoc-dependencies:
 	pip install --ignore-installed protobuf==4.24.0 "grpcio-tools>=1.56.2,<2" mypy-protobuf==3.1.0
 
+install-feast-ci-locally:
+	pip install -e ".[ci]"
+
```
redhatHameed, line 382: does we required this 
How it differentiate with ```pip install -e ".[dev]"```

redhatHameed (commented):
lgtm minor comments

Makefile
lokeshrangineni: just it is a additional target to install dev specific libraries, i had to install this on local to test and resolve one of the issue. 
https://github.com/feast-dev/feast/blob/master/setup.py#L158-L224

tokoko: @lokeshrangineni It's not mandatory of course, but you can also use `make install-python-ci-dependencies-uv` per dev [guide](https://github.com/feast-dev/feast/blob/master/docs/project/development-guide.md#feast-python-sdk--cli). Just a lot better experience imho once you set some stuff up. It also uses pinned requirements files instead of installing latest versions which means your environment is similar to where github runs the tests.

tokoko: P.S. `dev` and `ci` extras are identical now. Just decided to leave both of them not to mess with anyone's workflow

sdk/python/feast/infra/online_stores/remote.py
tokoko: Sorry, missed this. +1 for `NotImplementedError`. `pass` in this context (without the method marked as abstract) just returns `None` and may send you on very elaborate debugging journeys.

lokeshrangineni: updated the PR with `NotImplementedError `

Makefile
lokeshrangineni: if both are doing the same and `make install-python-ci-dependencies-uv` is recommended then better to remove this target. Updated the PR.

sdk/python/feast/infra/online_stores/remote.py
lokeshrangineni: It is throwing run time error so had to revert the change for update and teardown methods as it is mandatory. Please review the updated code.

tokoko (approved):
lgtm

lokeshrangineni:
After [approved](https://github.com/feast-dev/feast/pull/4226#pullrequestreview-2117133596) by @tokoko, just rebased the branch with feast-dev:master branch.  No changes made to PR after the approval.

franciscojavierarceo (approved):

PR #4231, opened 2024-05-24
feat: Make RegistryServer writable

### LINKED ISSUE #4216: Make RegistryServer writable
[RegistryServer](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/registry_server.py) is currently read-only. It needs to be writable as well to fit into the security model #4198.

tokoko:
- This PR adds apply/delete endpoints to the grpc registry server, also makes necessary changes in `remote` registry that call those endpoints.

Fixes #4216

### DIFF
--- a/sdk/python/tests/integration/registration/test_universal_registry.py
+++ b/sdk/python/tests/integration/registration/test_universal_registry.py
@@ -18,6 +18,7 @@ from datetime import timedelta
 from tempfile import mkstemp
 from unittest import mock
 
+import grpc_testing
 import pandas as pd
 import pytest
 from pytest_lazyfixture import lazy_fixture
@@ -36,8 +37,11 @@ from feast.field import Field
 from feast.infra.infra_object import Infra
 from feast.infra.online_stores.sqlite import SqliteTable
 from feast.infra.registry.registry import Registry
+from feast.infra.registry.remote import RemoteRegistry, RemoteRegistryConfig
 from feast.infra.registry.sql import SqlRegistry
 from feast.on_demand_feature_view import on_demand_feature_view
+from feast.protos.feast.registry import RegistryServer_pb2, RegistryServer_pb2_grpc
+from feast.registry_server import RegistryServer
 from feast.repo_config import RegistryConfig
 from feast.stream_feature_view import Aggregation, StreamFeatureView
 from feast.types import Array, Bytes, Float32, Int32, Int64, String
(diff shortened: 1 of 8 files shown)


sdk/python/tests/integration/registration/test_feature_store.py
```diff
@@ -11,68 +11,21 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import os
```
franciscojavierarceo, line 14: There's a lot being deleted here. Can you update the PR notes to explain why? I'm not fully familiar with the usage of all of these tests so I'd like to make sure I understand the implications of removing all of these tests (along with the other files).

sdk/python/tests/integration/registration/test_universal_registry.py
```diff
@@ -187,19 +191,70 @@ def sqlite_registry():
...
+
+# sql_fixtures = [
+#     "pg_registry",
+#     "mysql_registry",
+#     "sqlite_registry",
+# ]
+
+
 @pytest.mark.integration
-@pytest.mark.parametrize(
```
franciscojavierarceo, line 191: Same here

sdk/python/tests/unit/infra/registry/test_remote.py
```diff
@@ -1,69 +0,0 @@
-import assertpy
```
franciscojavierarceo, line 1: Same here

sdk/python/tests/unit/test_registry_server.py
```diff
@@ -1,60 +0,0 @@
-import assertpy
```
franciscojavierarceo, line 1: Same here

franciscojavierarceo (changes requested):
Can you add more details to the PR notes for this? Looks like we're removing a lot of tests and I'd like to make sure I understand the consequences of doing so as I'm not familiar with the usage of those tests.

tokoko:
Updated the PR description, let me know if it needs more clarity.

franciscojavierarceo:
Thanks for updating the PR description. 

Can you separate the removal of the tests into another PR? It'd be much cleaner in case we need to revert it.

tokoko:
@franciscojavierarceo cut down on the scope of the PR, reverted all test removals.

franciscojavierarceo (approved):
thank you!

PR #4232, opened 2024-05-27
chore: Registry tests cleanup

tokoko:
- tests requiring container resources in `test_universal_registry.py` are pinned to a single worker using `xdist_group` to cut down on setup times as session-scoped fixtures aren't reused across workers.
- `test_universal_cli.py` was unnecessarily marked with `universal_offline_stores`, changed to run only once with `file` offline store.
- `test_apply_feature_view_integration` and `test_apply_entity_integration` tests in `test_feature_store.py` are removed as there are basically identical to `test_apply_feature_view_success` and `test_apply_entity_success` tests in `test_universal_registry.py`.
- `test_remote.py` was removed as it was testing for only a small subset of functionality compared to `test_universal_registry.py`. `mock_remote_registry` fixture from `test_remote.py` was instead moved to `test_universal_registry.py`. This was impossible before because universal registry tests depended on `apply`/`delete` operations as well.
- `test_registry_server.py` is also effectively superseded by `test_universal_registry.py` as it's testing for a much wider range of functionality from a higher-level interface (`remote` registry instead of direct grpc server calls).

P.S. In my dev environment the changes cut the overall runtime of `test-python-integration-local` from ~7m down to ~4m.


franciscojavierarceo (approved):

franciscojavierarceo:
Very nice!

PR #4233, opened 2024-05-27
chore: Remove PostgreSQLRegistryStore

tokoko:
`PostgreSQLRegistryStore` was deprecated a couple of versions back (#4020). This PR removes it from feast.


shuchu (approved):
LGTM

PR #4237, opened 2024-05-29
fix: Make Java gRPC client use timeouts as expected

acevedosharp:
In my previous [PR](https://github.com/feast-dev/feast/pull/4217) I added deadlines to the gRPC stub used in Java `FeastClient`, however deadlines in gRPC don't behave like timeouts and in that PR are only set once when the client is constructed, causing requests after construction to fail with deadline exceeded even though the time they took was was within the deadline the used thought they had set up; here's an example of the issue:

```java
// Create the client, deadline starts counting from the moment client is instantiated
FeastClient client = new FeastClient(channel, Optional.empty(), Optional.of(Deadline.after(200, TimeUnit.MILLISECONDS)));
Thread.sleep(200);
client.getOnlineFeatures(...); // throws a StatusRuntimeException: DEADLINE_EXCEEDED
```
The `getOnlineFeatures(...)` call would exceed the deadline even if it took 1ms.

The official gRPC examples show the correct usage for adding timeouts is to create a copy stub with a new deadline setup for every new request as shown [here](https://github.com/grpc/grpc-java/blob/5c6b80881deb188586346f2cbbf2f35dd7f5f6f6/examples/src/main/java/io/grpc/examples/deadline/DeadlineClient.java#L56C31-L56C48), which is what is included in this PR.

N/A

# Fixes
N/A

### DIFF
--- a/java/serving-client/src/main/java/dev/feast/FeastClient.java
+++ b/java/serving-client/src/main/java/dev/feast/FeastClient.java
@@ -26,7 +26,6 @@ import feast.proto.serving.ServingServiceGrpc;
 import feast.proto.serving.ServingServiceGrpc.ServingServiceBlockingStub;
 import feast.proto.types.ValueProto;
 import io.grpc.CallCredentials;
-import io.grpc.Deadline;
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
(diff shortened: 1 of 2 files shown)


franciscojavierarceo (approved):
lgtm

PR #4243, opened 2024-05-31
docs: Add limitation infomation to the document of Snowflak offline st…

### LINKED ISSUE #4242: Update the Snowflake document to mention the SQL query string limitation.
Based on the feedback from the Feast community:

====
Hey team! I’m having a problem when running a materialization in Snowflake. [Here](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/materialization/snowflake_engine.py#L271C95-L271C117), this is executed:

```
query = f"""SELECT SYSTEM$LAST_CHANGE_COMMIT_TIME('{feature_view.batch_source.get_table_query_string()}') AS last_commit_change_time"""
```

The problem is that if feature_view.batch_source.get_table_query_string() returns something with ' , that single quotes breaks when running in snowflake.
If I’m using a query like:
```
SELECT
    some_column
FROM
    some_table
WHERE
    other_column = 'value'
That 'value' fails in Snowflake because of the extra quotes.
```
====

That 'value' fails in Snowflake because of the extra quotes.

For snowflake, it’s workable by quoting strings with $$ :
```
SELECT
    some_column
FROM
    some_table
WHERE
    other_column = $$value$$
```
I faced this some times, where I was able to apply a feature definition, but then the materialization failed due to an error on one of the queries used to generate the materialization

As snowflake requires strings to be between $$ or '.  
reference: https://docs.snowflake.com/en/sql-reference/data-types-text#single-quoted-string-constants

- Version: 0.38.0
- Platform:
- Subsystem:

shuchu:
Add SQL query string limitation information to Snowflake offline storage.

# Fixes
#4242


sudohainguyen (approved):
lgtm

tokoko:
Not a big snowflake user here.. Is there a workaround we can suggest?

shuchu:
I believe there always have a solution. It's just that we don't have a testing ENV from Snowflake. The 30-day free Snowflake account requires a company domain based email address and I already consumed mine.

shuchu:
> @shuchu No, I meant we should also mention in the docs that they should use dollar signs syntax instead that you're provided in the ticket (`$$value$$`).

Information added. please help me review it again.

HaoXuAI (approved):

tokoko:
@shuchu Can you rebase? offending flaky tests have already been disabled in another PR.

tokoko (approved):
LGTM

Issue #4245, opened 2024-05-31
Add `--tags` filtering capability to the `feast feature-views list` cli command


tchughesiv:
**Is your feature request related to a problem? Please describe.**
We should be able to list feature-views by tags, same as the UI already supports.

**Describe the solution you'd like**
```bash
$ feast feature-views list --tags team:driver_performance              
NAME                       ENTITIES    TYPE
driver_hourly_stats        {'driver'}  FeatureView
driver_hourly_stats_fresh  {'driver'}  FeatureView
```
```bash
$ feast feature-views list
NAME                       ENTITIES    TYPE
driver_hourly_stats        {'driver'}  FeatureView
driver_hourly_stats_fresh  {'driver'}  FeatureView
transformed_conv_rate_fresh  {'driver'}  OnDemandFeatureView
transformed_conv_rate        {'driver'}  OnDemandFeatureView
```

PR #4203, opened 2024-05-15
fix: Improve the code related to on-demand-featureview.

shuchu:
# What this PR does / why we need it:
The SonarCloud (company which provides SonarCube) suggests several improvements about the code quality. 

Also, this PR is about to improve the code quality and the first step of improving running efficiency of "on-demand feature view" feature.

# Which issue(s) this PR fixes:

# Fixes #4202

PR #4246, opened 2024-05-31
feat: Add the ability to list objects by `tags`

tchughesiv:
In the feast registry & cli, add the ability to list objects by `tags`... same as the UI already supports.
Here are the object types with this new `--tags` capability -

- data-sources (`DataSource`)
- entities (`Entity`)
- feature-services (`FeatureService`)
- feature-views (`FeatureView`)
- on-demand-feature-views (`OnDemandFeatureView`)
- `StreamFeatureView`

Fixes: https://github.com/feast-dev/feast/issues/4245

e.g. 
```bash
$ feast feature-views list --help
Usage: feast feature-views list [OPTIONS]

  List all feature views

Options:
  --tags TEXT  Filter by tags (e.g. --tags 'key:value' --tags 'key:value,
               key:value, ...'). Items return when ALL tags match.
  --help       Show this message and exit.
```
```bash
$ feast feature-views list --tags team:driver_performance              
NAME                       ENTITIES    TYPE
driver_hourly_stats        {'driver'}  FeatureView
driver_hourly_stats_fresh  {'driver'}  FeatureView
```
```bash
$ feast feature-views list
NAME                       ENTITIES    TYPE
driver_hourly_stats        {'driver'}  FeatureView
driver_hourly_stats_fresh  {'driver'}  FeatureView
transformed_conv_rate_fresh  {'driver'}  OnDemandFeatureView
transformed_conv_rate        {'driver'}  OnDemandFeatureView
```

### DIFF
--- a/sdk/python/feast/cli.py
+++ b/sdk/python/feast/cli.py
@@ -47,6 +47,12 @@ from feast.repo_operations import (
 from feast.utils import maybe_local_tz
 
 _logger = logging.getLogger(__name__)
+tagsOption = click.option(
+    "--tags",
+    help="Filter by tags (e.g. --tags 'key:value' --tags 'key:value, key:value, ...'). Items return when ALL tags match.",
+    default=[""],
+    multiple=True,
+)
 
 
 class NoOptionDefaultFormat(click.Command):
@@ -226,14 +232,16 @@ def data_source_describe(ctx: click.Context, name: str):
 
 
 @data_sources_cmd.command(name="list")
+@tagsOption
 @click.pass_context
-def data_source_list(ctx: click.Context):
+def data_source_list(ctx: click.Context, tags: list[str]):
     """
     List all data sources
     """
     store = create_feature_store(ctx)
     table = []
-    for datasource in store.list_data_sources():
+    tags_filter = utils.tags_list_to_dict(tags)
+    for datasource in store.list_data_sources(tags=tags_filter):
         table.append([datasource.name, datasource.__class__])
 
     from tabulate import tabulate
@@ -272,14 +280,16 @@ def entity_describe(ctx: click.Context, name: str):
 
 
 @entities_cmd.command(name="list")
+@tagsOption
 @click.pass_context
-def entity_list(ctx: click.Context):
+def entity_list(ctx: click.Context, tags: list[str]):
     """
     List all entities
     """
     store = create_feature_store(ctx)
     table = []
-    for entity in store.list_entities():
+    tags_filter = utils.tags_list_to_dict(tags)
+    for entity in store.list_entities(tags=tags_filter):
         table.append([entity.name, entity.description, entity.value_type])
 
     from tabulate import tabulate
@@ -320,14 +330,16 @@ def feature_service_describe(ctx: click.Context, name: str):
 
 
 @feature_services_cmd.command(name="list")
+@tagsOption
 @click.pass_context
-def feature_service_list(ctx: click.Context):
+def feature_service_list(ctx: click.Context, tags: list[str]):
     """
     List all feature services
     """
     store = create_feature_store(ctx)
     feature_services = []
-    for feature_service in store.list_feature_services():
+    tags_filter = utils.tags_list_to_dict(tags)
+    for feature_service in store.list_feature_services(tags=tags_filter):
         feature_names = []
         for projection in feature_service.feature_view_projections:
             feature_names.extend(
@@ -371,16 +383,18 @@ def feature_view_describe(ctx: click.Context, name: str):
 
 
 @feature_views_cmd.command(name="list")
+@tagsOption
 @click.pass_context
-def feature_view_list(ctx: click.Context):
+def feature_view_list(ctx: click.Context, tags: list[str]):
     """
     List all feature views
     """
     store = create_feature_store(ctx)
     table = []
+    tags_filter = utils.tags_list_to_dict(tags)
     for feature_view in [
-        *store.list_feature_views(),
-        *store.list_on_demand_feature_views(),
+        *store.list_batch_feature_views(tags=tags_filter),
+        *store.list_on_demand_feature_views(tags=tags_filter),
     ]:
         entities = set()
         if isinstance(feature_view, FeatureView):
@@ -434,14 +448,16 @@ def on_demand_feature_view_describe(ctx: click.Context, name: str):
 
 
 @on_demand_feature_views_cmd.command(name="list")
+@tagsOption
 @click.pass_context
-def on_demand_feature_view_list(ctx: click.Context):
+def on_demand_feature_view_list(ctx: click.Context, tags: list[str]):
     """
     [Experimental] List all on demand feature views
     """
     store = create_feature_store(ctx)
     table = []
-    for on_demand_feature_view in store.list_on_demand_feature_views():
+    tags_filter = utils.tags_list_to_dict(tags)
+    for on_demand_feature_view in store.list_on_demand_feature_views(tags=tags_filter):
         table.append([on_demand_feature_view.name])
 
     from tabulate import tabulate
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -204,23 +204,29 @@ class FeatureStore:
 
         self._registry = registry
 
-    def list_entities(self, allow_cache: bool = False) -> List[Entity]:
+    def list_entities(
+        self, allow_cache: bool = False, tags: Optional[dict[str, str]] = None
+    ) -> List[Entity]:
         """
         Retrieves the list of entities from the registry.
 
         Args:
             allow_cache: Whether to allow returning entities from a cached registry.
+            tags: Filter by tags.
 
         Returns:
             A list of entities.
         """
-        return self._list_entities(allow_cache)
+        return self._list_entities(allow_cache, tags=tags)
 
     def _list_entities(
-        self, allow_cache: bool = False, hide_dummy_entity: bool = True
+        self,
+        allow_cache: bool = False,
+        hide_dummy_entity: bool = True,
+        tags: Optional[dict[str, str]] = None,
     ) -> List[Entity]:
         all_entities = self._registry.list_entities(
-            self.project, allow_cache=allow_cache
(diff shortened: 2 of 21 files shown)


sdk/python/feast/feature_store.py
```diff
@@ -247,23 +247,28 @@ def list_feature_services(self) -> List[FeatureService]:
...
             A list of feature views.
         """
-        return self._list_feature_views(allow_cache)
+        return self._list_feature_views(allow_cache=allow_cache, tags=tags)
 
     def _list_feature_views(
         self,
         allow_cache: bool = False,
         hide_dummy_entity: bool = True,
+        tags: str = "",
```
lokeshrangineni, line 269: This comment is not specific to your change. WDYT of moving this method to BaseRegistry by adding some life cycle method to super class so that all the implementation of registries will invoke it. The goal is to keep feature store as simple delegator and keep it clean.

tchughesiv: i don't really have an opinion on it. if done though, i'd imagine it would have to be in a separate PR.

lokeshrangineni: I agree.

tokoko: That might be even more important when security model comes into picture (I don't think we have considered `list` permissions, though cc @dmartinol). Putting it in `BaseRegistry` means we can potentially have a way to allow giving permissions to `list` based on tags.

tokoko:
We should probably follow up (maybe in other PRs) with similar changes for other feast objects that have tags (entities, data sources). Seems odd to have this only for feature views.

tchughesiv:
@tokoko i didn't realize that about the other objects. maybe its better to go ahead and add to this PR>?

tokoko:
Yeah, changes to others in this PR would be great. All the objects in [BaseRegistry](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/registry/base_registry.py) are worth checking out, but I don't think all of them have a `list` command in cli, so might be a moot point for some for now. You can simply go ahead and implement it for all `*_list` methods in `cli.py` (feature view, data source, entity, feature service, on demand feature view).

sdk/python/feast/feature_store.py
franciscojavierarceo: +1

dmartinol: +1 @tokoko we can discuss this in a different thread, but I assume that the LIST action is just equivalent to the READ one, applied to all the listed resources.
In short, if you can LIST objects [o1,...,oN], then you must also have permission to READ from any of them. If not, this means that the list API is returning resources that you're not authorized to READ.

sdk/python/feast/feature_store.py
```diff
@@ -275,7 +280,8 @@ def _list_feature_views(
             ):
                 fv.entities = []
                 fv.entity_columns = []
-            feature_views.append(fv)
+            if all(fv.tags.get(key, None) == val for key, val in tagDict.items()):
```
dmartinol, line 283: @tokoko later we can move this logic in a global function like:
```py
def has_all_tags(resource_tags: Dict[str, str], requested_tags: Dict[str, str]) -> bool:
...
```
in the `utils` module? I see at least 2 copies of the same logic in this PR and others will follow when we check for authorized permissions (against all the managed objects, not just `FeatureView`s)

tokoko: We won't need a separate `list` permission that way, right? If a user has a `read` permission on every resulting object, that will imply a list permission as well...

sdk/python/feast/feature_store.py
tokoko: @dmartinol yeah, that was what I was getting at as well... Maybe if we extend registry API in the future to allow just listing the names of the objects w/o the objects themselves, then we might reconsider it, but right now that makes sense.

dmartinol:
> We should probably follow up (maybe in other PRs) with similar changes for other feast objects that have tags (entities, data sources). Seems odd to have this only for feature views.

@tokoko, I would also add that, since the direction is to expose authorized access through Feast services, we should include the same filtering by tags for all List APIs in RegistryServer. What do you think?

sdk/python/feast/utils.py
```diff
@@ -256,3 +256,13 @@ def _convert_arrow_to_proto(
         created_timestamps = [None] * table.num_rows
 
     return list(zip(entity_keys, features, event_timestamps, created_timestamps))
+
+
+def tags_str_to_dict(tags: str):
```
franciscojavierarceo, line 261:
```suggestion
def tags_str_to_dict(tags: str) -> Dict[str, str]:
```

franciscojavierarceo: It'd be good to have an explicit test for this.

sdk/python/feast/feature_store.py
tchughesiv: this has been moved to a global function in `utils`

sdk/python/feast/feature_store.py
tchughesiv: @tokoko @lokeshrangineni the `tags` filtering logic has been moved to the `BaseRegistry` and module.

sdk/python/feast/utils.py
tchughesiv: done

tchughesiv:
@tokoko @dmartinol i've added filtering by tags to all feast objects that have the `tags` field.

tokoko:
@tchughesiv Looks good to me, but better place for tests is in [test_universal_registry.py](https://github.com/feast-dev/feast/blob/master/sdk/python/tests/integration/registration/test_universal_registry.py). That's the one that's run for all registry implementations.

dmartinol:
👍  @tchughesiv shouldn't we also update the documentation for the CLI options?
BTW: unless you think it's obvious, I would also explain that all tags must match to satisfy the filtering condition (ATM there's no way to select resources matching any of the given tags)

tokoko:
@dmartinol I also had to check in the code that all tags had to match lol. Maybe it should have been, but wasn't obvious to me.

sdk/python/feast/cli.py
```diff
@@ -47,6 +47,11 @@
 from feast.utils import maybe_local_tz
 
 _logger = logging.getLogger(__name__)
+tagsOption = click.option(
+    "--tags",
+    help="Filter by tags (e.g. 'key:value, key:value, ...')",
+    multiple=True,
```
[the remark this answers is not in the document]
tchughesiv: good catch! my original intention was to also support the passing of multiple `--tags` options, but i was hitting the following issue as described in the docs you linked and i forgot to remove that setting.
_When passing a default with `multiple=True`, the default value must be a list or tuple, otherwise it will be interpreted as a list of single characters._

Those docs have gotten me past that hurdle and we can now support the passing of multiple tags AND/OR a singular comma-separated list. If both are used, the dictionary is updated accordingly. For example - 
```bash
$ feast feature-views list --tags "team : driver_performance, other:tag" --tags missingtag: --tags "other:two" --tags other:3 
```
would become the following dict -
```python
{'team': 'driver_performance', 'other': '3', 'missingtag': ''}
```
i like this approach because it affords flexibility to the user and we have to massage the string inputs anyway, so there's no harm in supporting a comma-separated list as well.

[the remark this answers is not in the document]
tchughesiv: thoughts?

sdk/python/feast/cli.py
```diff
@@ -226,14 +232,16 @@ def data_source_describe(ctx: click.Context, name: str):
 
 
 @data_sources_cmd.command(name="list")
+@tagsOption
 @click.pass_context
-def data_source_list(ctx: click.Context):
+def data_source_list(ctx: click.Context, tags: list[str]):
```
dmartinol, line 237: isn't this `List` instead of `list`? 🤔

sdk/python/feast/utils.py
```diff
@@ -256,3 +256,34 @@ def _convert_arrow_to_proto(
...
+        return None
+    tags_dict: dict[str, str] = {}
+    for tags_str in tags_list:
+        tags_dict.update(tags_str_to_dict(tags_str))
+    return tags_dict
+
+
+def tags_str_to_dict(tags: str = "") -> dict[str, str]:
+    tags_list = (
+        str(tags).strip().strip("()").replace('"', "").replace("'", "").split(",")
```
dmartinol, line 282: is `str(tags)` needed if `tags` is already an `str`?

sdk/python/feast/cli.py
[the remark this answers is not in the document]
dmartinol: fine for me, @tokoko ?

sdk/python/feast/cli.py
tchughesiv: The upper case `List` has been apparently been deprecated since 3.9, the linter rightly encouraged me to use the lower case `list`.
https://docs.python.org/3/library/typing.html#deprecated-aliases

sdk/python/feast/utils.py
tchughesiv: not anymore! :) . this was oddly necessary when the input was -
```python
def tags_str_to_dict(tags: Optional[str] = None)
```
given the recent changes i've made around supporting the multiple `--tags` cli option, this can go away. thanks

sdk/python/feast/cli.py
dmartinol: Right, [PEP 585](https://peps.python.org/pep-0585/) introduced built-in types in Python 3.9, probably we can do some code cleanup of other useless imports as well. Thanks!

tchughesiv:
@dmartinol @tokoko thanks for the reviews. i've made the requested changes, including improved docs and additional tests in [test_universal_registry.py](https://github.com/feast-dev/feast/blob/master/sdk/python/tests/integration/registration/test_universal_registry.py)

sdk/python/feast/cli.py
tokoko: Shuchu started doing it in #4203, but I'm sure there's a lot more left.

tokoko (approved):
LGTM

franciscojavierarceo (approved):

jeremyary:
Thanks all for the reviews & feedback!

Issue #4241, opened 2024-05-30
'feast.ui_server' is not a package


mascai:
## Expected Behavior 
feast ui command is starting without errors

## Current Behavior
```
feast ui --host 127.0.0.1 --port 8889
/Users/aamoskalenko/0_root_folder/02_dev/35_feast_ui/venv/lib/python3.11/site-packages/feast/ui_server.py:45: DeprecationWarning: 
        on_event is deprecated, use lifespan event handlers instead.

        Read more about it in the
        [FastAPI docs for Lifespan Events](https://fastapi.tiangolo.com/advanced/events/).
        
  @app.on_event("shutdown")
/Users/aamoskalenko/0_root_folder/02_dev/35_feast_ui/venv/lib/python3.11/site-packages/fastapi/applications.py:4495: DeprecationWarning: 
        on_event is deprecated, use lifespan event handlers instead.

        Read more about it in the
        [FastAPI docs for Lifespan Events](https://fastapi.tiangolo.com/advanced/events/).
        
  return self.router.on_event(event_type)
Traceback (most recent call last):
  File "/Users/aamoskalenko/0_root_folder/02_dev/35_feast_ui/venv/bin/feast", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/Users/aamoskalenko/0_root_folder/02_dev/35_feast_ui/venv/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aamoskalenko/0_root_folder/02_dev/35_feast_ui/venv/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/aamoskalenko/0_root_folder/02_dev/35_feast_ui/venv/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aamoskalenko/0_root_folder/02_dev/35_feast_ui/venv/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #4248, opened 2024-06-01
fix: Feast UI importlib change

privatedumbo:
Starting the UI fails when using the builtin `importlib.resources.files` function; it works with the one from the library `importlib-resources`. 

## Proposed solution
Using the reference to the module's parent (its package) through __spec__ ([docs](https://docs.python.org/3/library/importlib.html#importlib.machinery.ModuleSpec.parent)). We're using `__spec__.parent` as opposed to `__package__` because `__package__` does not seem to be the way according [to this](https://docs.python.org/3/reference/import.html#package__).

According to the python-docs, the function should work with both modules and packages, but that does not seem to be the case ([link to docs](https://docs.python.org/3.11/library/importlib.resources.html#importlib.resources.files)).

Fixes [4241](https://github.com/feast-dev/feast/issues/4241).

### DIFF
--- a/sdk/python/feast/ui_server.py
+++ b/sdk/python/feast/ui_server.py
@@ -51,7 +51,7 @@ def get_app(
 
     async_refresh()
 
-    ui_dir_ref = importlib_resources.files(__name__) / "ui/build/"
+    ui_dir_ref = importlib_resources.files(__spec__.parent) / "ui/build/"  # type: ignore[name-defined]
     with importlib_resources.as_file(ui_dir_ref) as ui_dir:
         # Initialize with the projects-list.json file
         with ui_dir.joinpath("projects-list.json").open(mode="w") as f:


sdk/python/feast/ui_server.py
```diff
@@ -51,7 +51,7 @@ def shutdown_event():
 
     async_refresh()
 
-    ui_dir_ref = importlib_resources.files(__name__) / "ui/build/"
+    ui_dir_ref = importlib_resources.files(__spec__.parent) / "ui/build/"  # type: ignore[name-defined]
```
privatedumbo, line 54: Type ignore due to a mypy issue:
https://github.com/python/mypy/pull/14739

It was merged 2 weeks ago, but we're using `mypy` 1.10 which is latest currently, and the change might not have been released.

tokoko (approved):
LGTM

shuchu (approved):

PR #4252, opened 2024-06-04
feat: Monitoring for Feast Feature Store (python server)

tsisodia10:

This draft pull request introduces monitoring functionality for the Feast feature store(python server). It instruments the feature store to capture system-level metrics, currently focusing on CPU and memory usage. Additionally, this PR includes the setup for a Prometheus server to collect and store these metrics.

Note: This pull request is currently a work in progress.

### DIFF
--- a/infra/charts/feast-feature-server/templates/deployment.yaml
+++ b/infra/charts/feast-feature-server/templates/deployment.yaml
@@ -3,7 +3,7 @@ kind: Deployment
 metadata:
   name: {{ include "feast-feature-server.fullname" . }}
   labels:
-    {{- include "feast-feature-server.labels" . | nindent 4 }}
+    {{- include "feast-feature-server.labels" . | nindent 4 }} 
 spec:
   replicas: {{ .Values.replicaCount }}
   selector:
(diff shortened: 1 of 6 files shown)


shuchu:
Is it possible to use open-telemetry client for instrumentation?

jeremyary:
@shuchu we'll definitely do some investigating! We're not necessarily opposed to proper abstraction that could bridge components or something similarly if need be down the road. I saw that it's the second most-active CNCF project so I'm guessing it's fairly popular :)

shuchu:
> @shuchu we'll definitely do some investigating! We're not necessarily opposed to proper abstraction that could bridge components or something similarly if need be down the road. I saw that it's the second most-active CNCF project so I'm guessing it's fairly popular :)

Thank you Jeremy.  The Open-telemetry  seems to be a "standard" service for many companies. I thought it not only can benefit the Feast but also the Red Hat's OpenShift (https://www.redhat.com/en/technologies/cloud-computing/openshift/observability).

shuchu:
> Thanks @shuchu for the recommendation, that sounds like a great idea and we are looking into the implementation.

Thank you @tsisodia10 !

PR #4253, opened 2024-06-04
feat: Adding type to base feature view

franciscojavierarceo:

# Fixes

### DIFF
--- a/protos/feast/core/FeatureView.proto
+++ b/protos/feast/core/FeatureView.proto
@@ -61,6 +61,7 @@ message FeatureViewSpec {
 
     // Owner of the feature view.
     string owner = 11;
+    string feature_view_type = 13;
 
     // Features in this feature view can only be retrieved from online serving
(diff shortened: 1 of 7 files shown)


tokoko:
I'm really not sure about this.. 😄 looks like a bug waiting to happen imho, a user can technically do something like this `OnDemandFeatureView(..., feature_view_type=FeatureViewType.BATCH)`.

HaoXuAI:
> I'm really not sure about this.. 😄 looks like a bug waiting to happen imho, a user can technically do something like this `OnDemandFeatureView(..., feature_view_type=FeatureViewType.BATCH)`.

that can be avoided by removing the parameter from the constructor i think. but what's the reason of having this field?

franciscojavierarceo:
I wanted to include this metadata as I thought we could use this in `ilst_all_feature_views` and that this would be better for tagging but we can just do the same by using the type so I'll close this instead.

PR #4256, opened 2024-06-05
feat: List all feature views

franciscojavierarceo:

# Fixes

### DIFF
--- a/protos/feast/core/FeatureView.proto
+++ b/protos/feast/core/FeatureView.proto
@@ -61,6 +61,7 @@ message FeatureViewSpec {
 
     // Owner of the feature view.
     string owner = 11;
+    string feature_view_type = 13;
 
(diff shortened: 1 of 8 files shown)


HaoXuAI (approved):
LGTM

tokoko:
@franciscojavierarceo do you think we should keep this `list_all_feature_views` method in the future or is this temporary? Isn't the final goal to change the functionality of existing `list_feature_views` to return all?

franciscojavierarceo:
We'll remove after we fully switch get_feature_view. Let me add some logging alerting of the breaking changes.

tokoko:
in that case, lgtm as well. let's just throw some warning

franciscojavierarceo:
> in that case, lgtm as well. let's just throw some warning

Done.

Issue #4222, opened 2024-05-23
Running `feast materialize-incremental` for an end date far in future breaks incremental materializations up to that date


samhallam-reverb:
## Expected Behavior

When running `feast materialize-incremental` it should not be possible for a materialize to run starting from a date that hasn't occurred yet. Would expect `materialize-incremental` to choose `min(now, most_recent_materialization_date)`

## Current Behavior

When running `feast materialize-incremental` for a future date, say `2025-04-08T00:00:00`, it will set that date as the next start date. This breaks all future `materialize-incremental` commands up until that date, if you run daily materializations they will no longer function until the actual date goes beyond this erroneously entered start date. Because the data is stored as a serialized protobuf in the SQL-based registry this is also non-trivial to change. 

You can see code [here](https://github.com/feast-dev/feast/blob/1f17b9bcd635d1344e1f9be4dcb745d8e206e071/sdk/python/feast/feature_view.py#L444) showing that there is no validation checks for the start date, merely the latest date that has ever run.

## Steps to reproduce

Run a `materialize-incremental` command for a date far in the future. 

Run a second `materialize-incremental` command for a date less than the last start date. Watch Feast always find no date to materialize. 

### Specifications

- Version:  0.34.1
- Platform: Mac OSX
- Subsystem: Sonoma 14.5

## Possible Solution

Refactor the above function to choose `min(now, most_recent_materialization_date)`. Happy to contribute to this fix.

shuchu:
@samhallam-reverb did you try the 0.37.0 version?

also, welcome to help us fix this issue since you almost know the fix.

tokoko:
I'm not sure about this... Currently `materialize` commands only take into account values that are in `event_timestamp` column, never a clock time. Bringing `now` into this equation means we'll have to handle timezone differences between the two. Also, isn't there a possibility someone might want to prepopulate values into the future in the offline store, maybe you're calculating monthly feature values that you want to take effect from the start of the next month, idk...

To me, this feels like something that needs to be handled by the user, the user passes the end date after all and can always cap it with the current time beforehand.

tokoko:
@samhallam-reverb Sure, looks like it always looks for a max value in it's history of materialization timestamps, so it seems impossible to be rectified by normal means. Registry API also only supports `apply_materialization` that adds to the list of timestamp pairs, never removes them.

[...]

PR #4257, opened 2024-06-05
fix: Update Feast object metadata in the registry

msistla96:

This PR is to fix an issue observed when updating a feature view: The`created_timestamp` and `materialization_intervals`  values already present in the SQL registry for the feature view are overwritten by the respective values created when feature view is updated, which is `None`. So when one queries the feature view from the registry, it return `None` for both these fields.

The fix requires updating the `apply_object` function to have the previously stored values of the metadata fields like created_timestamp etc assigned to the feast objects respective fields, before it gets updated by the SQL registry. 
Hence this change has been made for the relevant Feast objects like `Entity`, `SavedDataset`, `FeatureService`,  `FeatureView`,`OnDemandFeatureView`,`StreamFeatureView`.
This change has also been applied for all other registries other than SQL.

# Fixes


tokoko:
Hey, thanks for raising this. Just curious, maybe you have taken a look, is this the case only for sql registry? why isn't the same happening for file-based registries?

msistla96:
@tokoko I noticed this for the SQL registry as our team uses it, but I'm not sure if this is happening for other registry types as well. If other registries do require this change, it would be separate changes to each of their apply methods from what I've seen so far.

sdk/python/tests/integration/registration/test_universal_registry.py
```diff
@@ -840,6 +989,100 @@ def simple_udf(x: int):
...
+    # Register Feature service
+    test_registry.apply_feature_service(fs, project)
+
+    feature_services = test_registry.list_feature_services(project)
+
+    # List Feature Services
+    assert len(feature_services) == 1
+    assert feature_services[0] == fs
+
+    test_registry.delete_feature_service("my_feature_service_1", project)
```
tokoko, line 1062: Should you be doing this? If you delete and apply again, of course `created_timestamp` will be regenerated. Isn't the whole point to assert that it's unchanged even if new object is applied when during apply an object already exists?

sdk/python/tests/integration/registration/test_universal_registry.py
```diff
@@ -840,6 +989,100 @@ def simple_udf(x: int):
...
+    test_registry.apply_feature_service(fs, project)
+
+    feature_services = test_registry.list_feature_services(project)
+
+    # Verify Feature Services
+    assert len(feature_services) == 1
+    assert feature_services[0] == fs
+    # The created_timestamp for the feature service should be set to the created_timestamp value stored from the
+    # previous apply
+    assert feature_services[0].created_timestamp is not None
```
tokoko, line 1081: If you drop `delete` above, then I think this should be an equality check to assert that it's not changed. `is not None` doesn't seem to be checking much, unless I'm missing something

tokoko:
@msistla96 sure, I quickly took a look and this doesn't seem to be handled anywhere. That's why I find it strange that the newly added tests in `test_universal_registry` aren't failing for file-based registries.

sdk/python/tests/integration/registration/test_universal_registry.py
msistla96: My bad, this was supposed to be present for `test_apply_feature_service_success` only. I'll remove this.

msistla96:
> @msistla96 sure, I quickly took a look and this doesn't seem to be handled anywhere. That's why I find it strange that the newly added tests in `test_universal_registry` aren't failing for file-based registries.

So the tests worked because for the other registries, the created_timestamp value does get set technically but it shouldn't for each time an entity gets updated. I've updated the code for all the other registries failing the test and also the tests as well. 
I've also had to include an additional check to see if `materialization_intervals` should be updated or not, based on whether the feature view was materialized before or not.

tokoko:
@msistla96 thanks. this is looking good. Can you add `materialization_intervals` test in `test_universal_registry` as well? You don't have to actually run anything, just call `apply_materialization` on `test_registry`.
@jeremyary Can you take a look why tests aren't running?

msistla96:
@tokoko, while adding the tests you mentioned, I noticed two bugs with **Remote Registry**:
1. During `apply_materialization`, when Remote Registry converts the start and end dates to Timestamp before making the request to Registry Server [here](https://github.com/feast-dev/feast/blob/2478831e8204bc8b9204ba048a74179ac3193367/sdk/python/feast/infra/registry/remote.py#L276-L277), Timestamp doesn't preserve the timezone information. So when RegistryServer converts them back to dates before calling  the proxy `apply_materialization` here, the resultant date is timezone naive, i.e: it assumes the date relative to the local timezone rather than UTC timezone because of the mentioned reason. I made a change to [RegistryServer](https://github.com/feast-dev/feast/blob/770d92a54821002af416e1acfe8322cb44543b0f/sdk/python/feast/registry_server.py#L296-L300) in this PR to fix this as the integration tests were failing for Remote Registry. 
FYI, it was raised as an [issue](https://github.com/protocolbuffers/protobuf/issues/1057) long ago but Timestamp is not meant to store timezone info in the first place.

To reproduce Timestamp's behavior:

```
from google.protobuf.timestamp_pb2 import Timestamp
from datetime import datetime
from pytz import utc

current_date = datetime.utcnow().replace(tzinfo=utc)
print("Current date: ",current_date, ", timezone: ",current_date.tzinfo)
sample_timestamp = Timestamp()
sample_timestamp.FromDatetime(current_date)

converted_date = datetime.fromtimestamp(
    sample_timestamp.seconds + sample_timestamp.nanos / 1e9)

print("Converted date: ", converted_date, ", timezone: ",converted_date.tzinfo)
```
3. `apply_materialization` fails for Stream Feature Views. [ApplyMaterializationRequest](https://github.com/feast-dev/feast/blob/770d92a54821002af416e1acfe8322cb44543b0f/protos/feast/registry/RegistryServer.proto#L98) for RegistryServer doesn't have Stream Feature View as a property even though ApplyFeatureViewRequest does. Fixing this is out of scope for this PR, so for now I've omitted test cases for materializing stream feature views to prevent the integration tests from failing. I added a TODO in the file as well, but let me know if you want me to raise this as a bug as well.

sdk/python/tests/integration/registration/test_universal_registry.py
```diff
@@ -783,6 +784,28 @@ def odfv1(feature_df: pd.DataFrame) -> pd.DataFrame:
     assert (
         updated_feature_view.created_timestamp is not None
         and updated_feature_view.created_timestamp == feature_view.created_timestamp
+        and len(updated_feature_view.materialization_intervals) == 0
+    )
+
+    # Simulate materialization
```
tokoko, line 790: You should put this before `test_registry.apply_feature_view(updated_fv1, project)` to make sure that an apply operation doesn't mess with materialization intervals.

tokoko:
> 2\. `apply_materialization` fails for Stream Feature Views. [ApplyMaterializationRequest](https://github.com/feast-dev/feast/blob/770d92a54821002af416e1acfe8322cb44543b0f/protos/feast/registry/RegistryServer.proto#L98) for RegistryServer doesn't have Stream Feature View as a property even though ApplyFeatureViewRequest does. Fixing this is out of scope for this PR, so for now I've omitted test cases for materializing stream feature views to prevent the integration tests from failing. I added a TODO in the file, but let me know if you want me to raise this as a bug as well.

I had the same bug in `ApplyFeatureViewRequest` and fixed it later on, missed `ApplyMaterializationRequest` apparently. Yup, a separate bug ticket for this would be good. thanks for catching this.

msistla96:
@tokoko Could I have another review please? I realised that I missed a test case when apply_materialization adds the materialization_intervals which shouldn't call update_materialization_intervals, which I add in the last commit.

tokoko:
@msistla96 that's an interesting corner case. Why do you think that should be the desired behavior? That actually would sort of address an issue raised here #4222, but I'm still unsure. We would effectively be allowing materialization intervals to be updates in two ways (apply_materialization and apply_feature_view).

msistla96:
> @msistla96 that's an interesting corner case. Why do you think that should be the desired behavior? That actually would sort of address an issue raised here #4222, but I'm still unsure. We would effectively be allowing materialization intervals to be updates in two ways (apply_materialization and apply_feature_view).

apply_materialization takes care of updating materialization_intervals, but apply_feature_view is to ensure the materialization_interval stored doesn't get erased(because the only time that would make sense is if the feature view was created for the first time and that becomes a separate feature view anyway). I should rename update_materialization_intervals  to copy_materialization_intervals to avoid this confusion. Does that make sense?

EXPEbdodla:
> > 2. `apply_materialization` fails for Stream Feature Views. [ApplyMaterializationRequest](https://github.com/feast-dev/feast/blob/770d92a54821002af416e1acfe8322cb44543b0f/protos/feast/registry/RegistryServer.proto#L98) for RegistryServer doesn't have Stream Feature View as a property even though ApplyFeatureViewRequest does. Fixing this is out of scope for this PR, so for now I've omitted test cases for materializing stream feature views to prevent the integration tests from failing. I added a TODO in the file, but let me know if you want me to raise this as a bug as well.
> 
> I had the same bug in `ApplyFeatureViewRequest` and fixed it later on, missed `ApplyMaterializationRequest` apparently. Yup, a separate bug ticket for this would be good. thanks for catching this.

@tokoko What is the purpose/ use case of calling apply_materialization for StreamFeatureViews? I'm assuming it mostly make sense for Feature Views only.

tokoko:
@EXPEbdodla yeah, you're right, my bad. I've had a similar use (not with feast) case when I needed to update online store both directly and from offline with different datasets, but until we can practically support something like that, there's no point.

msistla96:
@tokoko I see Registry support Stream Feature Views for apply_materialization currently, so I'm a bit confused about your comment on not having it. Are there any changes required for this PR?

tokoko:
@msistla96 Technically it is, but as feast workflow for stream feature views doesn't really contain materialization from offline to online, it's never used. It would be impossible to restrict it in signature with types only as `StreamFeatureView` extends `FeatureView` instead of `BaseFeatureView` (a historical quirk that probably needs to be revisited at some point).

I was still unsure about the update we discussed above, but you're probably right, overwriting with new materialization intervals if provided seems as good as any other option.

tokoko (approved):
LGTM, thanks for the fix

msistla96:
@tokoko Let me know when this would be merged.

HaoXuAI (approved):
LGTM

franciscojavierarceo (approved):

PR #4262, opened 2024-06-07
feat: Remote offline Store 

### LINKED ISSUE #4032: Remote offline feature server deployment

Currently Feast has the capability to deploy both online store (feature server) and registry as standalone services that feast library can access remotely. I'd like to see the same capability with offline store deployment. The primary goal of the feature is to enable fully remote deployment of all feast components and in future enable a common enterprise-level security model for all three components.

There are two possible general solutions I have in mind:

1. Arrow Flight Server Deployment

We can deploy an offline feature server as an Apache Arrow Flight Server that will wrap calls to offline store implementations and expose interface methods as Arrow Flight endpoints. This is also something that Hopsworks seems to have [implemented](https://www.hopsworks.ai/post/python-centric-feature-service-with-arrowflight-and-duckdb) before. They are using duckdb engine behind a flight server.

2. Metadata Sharing Server Deployment (similar to Delta Sharing)

Offline feature server can be deployed as a lightweight rest or grpc server that only handles requests and metadata and utilizes cloud filesystems for data transfer to and from the user. For example in case of s3, feature server would write out the results as a parquet dataset to a folder in an s3 bucket, generate presigned urls for it and return a list of urls to the caller. The idea is to emulate the mechanics of [delta sharing](https://delta.io/sharing/) but avoid it's limitations (in delta-sharing, dataset needs be in `delta` format and metadata responses must be immediate, which is obviously not an option for feast).

In both scenarios, client-side code should be virtually indistinguishable from other offline store implementations. The particulars of the implementations will be hidden behind RemoteOfflineStore class.

redhatHameed:

**This PR added feature for remote offline store contain below changes.**

**Added an Offline Server using Arrow Flight:** 

- Implement the remote OfflineServer similar to the one developed for the Remote [Registry](https://github.com/feast-dev/feast/blob/cf58ebed02eaa8f8da234f9446277bbbb5195681/sdk/python/feast/registry_server.py) or [Feature Server](https://github.com/feast-dev/feast/blob/60756cb4637a7961b6caffef3242e2886e77f78a/sdk/python/feast/feature_server.py)  but using Arrow Flight Server instead of pure gRPC. 
- Added the command into cli for starting the offline server using the `feast serve_offline` command.
- Implemented the the OfflineStore interface methods as arrow flight server methods. 
- Added unit test and integration test functions for the OfflineServer class.

**Added an Offline Remote Client:**

- Implement the offline remote client using arrow flight remote client approach, which supports the type: remote configuration in the offline store configuration for example as seen below.
``` offline_store:
    type: remote 
    host: mystore.feast.svc.cluster.local
     port: 8815
```
- Develop the RemoteOfflineStore class, which implements the existing [OfflineStore](https://github.com/feast-dev/feast/blob/ee51fbfe3f40e637084c5ecf51f7463b840d58b4/sdk/python/feast/infra/offline_stores/offline_store.py#L208) interface in Feast. Utilize the [get_historical_features](https://github.com/feast-dev/feast/blob/ee51fbfe3f40e637084c5ecf51f7463b840d58b4/sdk/python/feast/infra/offline_stores/offline_store.py#L251) method and pass parameters such as entity definition, feature views, offline store config to interact with the Arrow server functions using the Arrow Flight client.
- Client implemented the other methods like  pull_latest_from_table_or_query, write_logged_features , offline_write_batch
- added unit test function to test the remote client.

**Updated Helm Chart for Deploying as a service on Kubernetes:**

- updated chart deployment to install offline server by using `feast_mode` property  based on the deployment choice user can set the property. The online mode is the default and maintains backward compatibility with previous Feast Feature Server implementations.

 Helm install examples:

    helm install feast-feature-server feast-charts/feast-feature-server --set feature_store_yaml_base64=$(base64 > feature_store.yaml)
    helm install feast-offline-server feast-charts/feast-feature-server --set feast_mode=offline  --set feature_store_yaml_base64=$(base64 > feature_store.yaml)
    helm install feast-ui-server feast-charts/feast-feature-server --set feast_mode=ui  --set feature_store_yaml_base64=$(base64 > feature_store.yaml)
    helm install feast-registry-server feast-charts/feast-feature-server --set feast_mode=registry  --set feature_store_yaml_base64=$(base64 > feature_store.yaml)

**Added documentation for remote offline support**

**Added example for remote offline** 

Fixes #4032 

# Fixes


sdk/python/feast/offline_server.py
```diff
@@ -0,0 +1,318 @@
...
+            logger.error(f"Unknown key {key}")
+            return None
+
+        command = json.loads(key[1])
+        api = command["api"]
+        logger.debug(f"get command is {command}")
+        logger.debug(f"requested api is {api}")
+        try:
+            if api == OfflineServer.get_historical_features.__name__:
+                df = self.get_historical_features(command, key).to_df()
```
tokoko, line 136: we can call `to_arrow` here instead of `to_df`

sdk/python/feast/infra/offline_stores/remote.py
```diff
@@ -0,0 +1,413 @@
...
+        command_descriptor = _call_put(
+            api=api_name,
+            api_parameters=api_parameters,
+            client=client,
+            table=table,
+            entity_df=None,
+        )
+        bytes = command_descriptor.serialize()
+
+        client.do_action(pa.flight.Action(api_name, bytes))
```
tokoko, line 283: does this need to happen with two commands? I think a single put should be enough. I get that this way it's more similar to other actions, but we'd be essentially forcing the whole dataset to be kept in memory on server until do_action is called.

tmihalac: You might have a point but in case of `persist()` I need 2 parameters one is the retrieve method to call on the server and the second to know that I need to call persist, do you have an elegant solution without adding specific code to handle this flow?

sdk/python/tests/integration/feature_repos/universal/data_sources/file.py
```diff
@@ -352,3 +357,56 @@ def create_offline_store_config(self):
...
+
+    def create_offline_store_config(self) -> FeastConfigBaseModel:
+        self.remote_offline_store_config = RemoteOfflineStoreConfig(
+            type="remote", host="0.0.0.0", port=self.server_port
+        )
+        return self.remote_offline_store_config
+
+    def teardown(self):
+        super().teardown()
+        if self.proc is None and self.proc.is_alive():
```
tmihalac, line 402: Shouldn't this be `if self.proc is not None and self.proc.is_alive():` ?
@dmartinol

dmartinol: U r right. If you have time, coild you pls fix it? Otherwise I’ll do it tomorrow

sdk/python/feast/infra/offline_stores/remote.py
```diff
@@ -0,0 +1,401 @@
...
+    client: fl.FlightClient,
+):
+    updatedTable: pa.Table
+
+    if entity_df is not None:
+        updatedTable = pa.Table.from_pandas(entity_df)
+    elif table is not None:
+        updatedTable = table
+    else:
+        updatedTable = _create_empty_table()
```
tokoko, line 379: @tmihalac I think this is no longer necessary after your changes in `persist`. `_put_parameters` is always called with either `entity_df` or `table` set.

sdk/python/tests/integration/feature_repos/universal/data_sources/file.py
tmihalac: Will fix it

tokoko:
Just a note, if we decide in the future to use this server for remote materialization engine as well, the name might no longer be a good fit. Just pointing it out.. I don't have any better ideas.

tokoko:
@redhatHameed just curious, are the tests failing only in gh workflows? Do they pass locally?

dmartinol:
> Just a note, if we decide in the future to use this server for remote materialization engine as well, the name might no longer be a good fit. Just pointing it out.. I don't have any better ideas.

Maybe it's too early to kickoff a thread, but in that case, shoildn't we have a separate materialization server with its own codebase? (if you have an issue we can move the discussion there)

tokoko:
> Maybe it's too early to kickoff a thread, but in that case, shouldn't we have a separate materialization server with its own codebase? (if you have an issue we can move the discussion there)

maybe, that would certainly be a cleaner approach, but imho wouldn't really serve any other practical purpose. I'll open a ticket and we probably should pick this up later once remote offline is a bit more fleshed out.

dmartinol:
> @redhatHameed just curious, are the tests failing only in gh workflows? Do they pass locally?

@tokoko  Yes, apart some of them that fail locally. Do you have any idea why from the GH action execution we don't even see the log of the `Test local integration tests` step? Do you have any chance to re-run it and enable the debugging option?

tokoko:
I don't have any privileges there. @franciscojavierarceo @jeremyary any ideas?

sdk/python/tests/integration/feature_repos/universal/data_sources/file.py
```diff
@@ -362,33 +365,34 @@ def create_offline_store_config(self):
...
+        _time_out_sec: int = 180
+
         # Wait for server to start
         wait_retry_backoff(
             lambda: (None, check_port_open(host, self.server_port)),
-            timeout_secs=10,
+            timeout_secs=_time_out_sec,
+            timeout_msg=f"Unable to start the feast server in {_time_out_sec} seconds for remote offline store type, port={self.server_port}",
         )
+        logger.info("Port has been freed.")
```
dmartinol, line 395: isn't this "port has been occupied" instead?

redhatHameed:
> @redhatHameed just curious, are the tests failing only in gh workflows? Do they pass locally?

Yes it does - apart from the one test case -> t`est_spark_materialization_consistency` with error  - `py4j.protocol.Py4JJavaError: An error occurred while calling o114.count`. but if I skip it all are passing locally see the [here](https://gist.github.com/redhatHameed/99cb180d5a607f17840ffcc8ec7fcb38) results

sdk/python/tests/integration/feature_repos/universal/data_sources/file.py
lokeshrangineni: this code likely be triggered multiple times for each test case, isn't it a good idea to add the port number also to log just to differentiate?

sdk/python/tests/integration/feature_repos/universal/data_sources/file.py
```diff
@@ -352,3 +360,73 @@ def create_offline_store_config(self):
...
+    fs = FeatureStore(
+        config=RepoConfig(
+            project=project_name,
+            provider="local",
+            offline_store=parent_offline_config,
+            registry=registry_path,
+            entity_key_serialization_version=2,
+        )
+    )
+    start_server(fs, host, port)
```
tokoko, line 432: I might be grasping for straws, but maybe we can try `fs.serve_offline` instead of `start_server` to avoid an import. If that doesn't work, how about we simply prepopulate some temp folder with feature_store.yaml and then start offline server with cli command pointing to that feature_store.yaml.

tokoko: you can use this `yaml.dump(repo_config.dict(by_alias=True))` to generate file contents

redhatHameed: @tokoko Thanks you mean something like this 

```
    feature_store_configuration = yaml.dump(self.repo_config.dict(by_alias=True))
    fs.repo_path= feature_store_configuration
    fs.serve_offline(host, port)
```

tokoko: No, in that case I meant something like this, use `subprocess` instead of `multiprocessing`:
```
with open('temp.yml', 'w') as outfile:
    yaml.dump(self.repo_config.dict(by_alias=True), outfile)
    
process = subprocess.Popen(['feast', 'serve_offline', ...]) 
```

redhatHameed: @tokoko Thanks. I have updated the [code](https://github.com/feast-dev/feast/pull/4262/commits/08c2616e6e4972ed0ffe3e85e1379ff0e96ac786) and test are passing.

sdk/python/feast/offline_server.py
```diff
@@ -0,0 +1,323 @@
...
+        except Exception as e:
+            logger.exception(e)
+            traceback.print_exc()
+            raise e
+
+        # Get service is consumed, so we clear the corresponding flight and data
+        del self.flights[key]
+        return fl.RecordBatchStream(table)
+
+    def offline_write_batch(self, command, key):
```
lokeshrangineni, line 176: it is not a mandatory to give the types, but it makes it more readable to have types unless it is expected to have any data type.

docs/SUMMARY.md
```diff
@@ -117,6 +117,8 @@
   * [Python feature server](reference/feature-servers/python-feature-server.md)
   * [\[Alpha\] Go feature server](reference/feature-servers/go-feature-server.md)
   * [\[Alpha\] AWS Lambda feature server](reference/feature-servers/alpha-aws-lambda-feature-server.md)
+  * [Remote Offline Store ](reference/feature-servers/remote-offline-store.md)
```
tokoko, line 120: I think it would be more intuitive to have two pages in the docs for it
- remote-offline-store.md for the client under Offline Stores - how to configure the client
- offline-feature-server.md for the server under Feature Servers - how to configure the server

They should of course link to one another

sdk/python/feast/offline_server.py
tmihalac: You mean in the methods parameters or the return value or both?
I guess you mean for all the methods and not only this specific one?

lokeshrangineni: I was specifically about method parameters. but if there is no change in functionality better to add types to return variable also.

sdk/python/tests/integration/feature_repos/universal/data_sources/file.py
lokeshrangineni: seems like it is not yet fixed. just adding comment to avoid any confusion.

lokeshrangineni: also subprocess is not having `is_alive()` method just `is none` check enough.

redhatHameed: Thanks for catching will update in next pr

docs/SUMMARY.md
redhatHameed: updated accordingly. Please give another look

sdk/python/feast/infra/offline_stores/remote.py
redhatHameed: @tmihalac just to confirm is this resolved ?

redhatHameed:
@jeremyary any thoughts I am getting incorrect credentials error one of integration test it was running file perviosuly ```ERROR sdk/python/tests/integration/offline_store/test_push_features_to_offline_store.py::test_push_features_and_read[AWS:Redshift:redis:python_fs:False] - feast.errors.RedshiftCredentialsError: Redshift API failed due to incorrect credentials```

sdk/python/feast/infra/offline_stores/remote.py
tmihalac: Yes

docs/SUMMARY.md
tokoko: Looks good, thanks

sdk/python/feast/offline_server.py
redhatHameed: @lokeshrangineni the change implemented under [commit](https://github.com/feast-dev/feast/pull/4262/commits/bdf0150bd00608079122f41a4ebb71bb432f4ee3) thanks @tmihalac addressing it

redhatHameed:
@tokoko can we get your approval on this if no more comments. Also who will approve this for merging ?

tokoko (approved):
Yup, lgtm. great work, guys. @jeremyary or @franciscojavierarceo can approve and merge

jeremyary (approved):

PR #4272, opened 2024-06-12
chore: Remove serverless feature server deployments

### LINKED ISSUE #4236: Remove feature server remote deployment options (lambda, cloudrun)
Feature server has aws lambda and gcp cloudrun deployment options, lambda being marked alpha and cloudrun undocumented and probably not functional. Both are unmaintained at this point. Their deployment is opinionated and managed by `feast` cli itself, in other words they are deployed with a `feast apply` command, which I think makes them very impractical for a real-world infra deployment. It's practically impossible to deploy them as part of some sort of IaC workflow. 

On the flip side, we have to maintain a whole bunch of code that does a lot of vendor-specific infra management, for example lambda deployment contains uploading docker image do ecr, communicating with aws services using boto client and so on ([aws.py](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/aws.py)). There is parallel work going on supporting an official feast helm chart, which looks like a more stable alternative to me. I propose to remove lambda and cloudrun deployment options from feast.

tokoko:
Removes serverless feature server deployment options (aws lambda, gcp cloudrun)

Fixes #4236


HaoXuAI:
Do you think it's possible to change it to a plugin instead of removing it? Or is it not worth to di it?

tokoko:
I don't think it's worth it, not sure how would one do that even if it was (maybe with Infras). In any case, I don't think there's anyone around willing to maintain that.

tokoko:
@HaoXuAI Can we move ahead here? The code can be always dug up if anyone decides to support something like this as some sort of a plugin.

franciscojavierarceo:
I haven't reviewed the code thoroughly but I am assuming this won't be backwards compatible and it will be a breaking change, yeah?
I'm general, that sounds bad without feedback from the community.

tokoko:
The issue has been up for a few weeks and I posted on slack as well. We can deprecate first and remove later if you think that's necessary.

franciscojavierarceo (approved):

PR #4274, opened 2024-06-13
fix: Handles null values in data during GO Feature retrieval

EXPEbdodla:
Null values are materialized to online store as &types.Value{} instead of nil. Fixing the issue in Go Feature Server code to handle NULL values during retrieval. Without this fix, Go Feature server retrieval fails. 

# Fixes

- Go Feature server retrieval for NULL values in online store.
- Formatting in type_map.py and Log error when NULL values in INT are upcasted to FLOAT64. Logging helps to easily find the issue otherwise it's hard to debug the issue. Dataset has INT values when the TypeError is raised all the values are logged as FLOAT64 in _type_err() function. 

We need to fix NULL values in the array during the retrieval process. Below are my observations:

Null values inside arrays are not supported by either Materialization or Feature Retrieval in GO
- INT32, INT64 columns are upcasted to float64 when Null value exists in array. So it fails with invalid type error during materialization. Logging exception when this happens. 
- String arrays fail materialization when Null value exists
- Materialization works for float, double and timestamp. It's stored as Nan. On Go server side, json encoding not converting NaN to None. Feature server retrieval (GO) doesn't work for float, double but works for timestamp because of the default value set when NULL values are encountered in timestamp array.

### DIFF
--- a/go/types/typeconversion.go
+++ b/go/types/typeconversion.go
@@ -11,6 +11,9 @@ import (
 )
 
 func ProtoTypeToArrowType(sample *types.Value) (arrow.DataType, error) {
+	if sample.Val == nil {
+		return nil, nil
+	}
 	switch sample.Val.(type) {
 	case *types.Value_BytesVal:
 		return arrow.BinaryTypes.Binary, nil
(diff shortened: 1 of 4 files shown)


HaoXuAI (approved):
LGTM!

Issue #4270, opened 2024-06-12
Add `get_online_features` method in `OnlineStore` interface


tokoko:
I'm proposing we refactor online feature retrieval to move most of the logic for `get_online_features` and `get_online_features_async` methods from `FeatureStore` class in `feature_store.py` to new methods in `OnlineStore` class that will have default implementations. This shouldn't be breaking change, nor impact performance in any way, but offers a number of advantages.

- This more closely follows the structure of `OfflineStore` class that has `get_historical_features` as one of it's methods.
- Simplifies the implementation of remote online store as all request info will be exposed to OnlineStore provider. Feature Services are really tricky to implement now.
- Related to the point above, simplifies security checks in remote servers as more of request context is available to be passed to the server.
- Will be a step in a direction to address long-standing issue with our online stores that require us making a single round-trip to the database for each feature view because only a single feature view at a time is exposed to online store provider.

tokoko:
@dmartinol @lokeshrangineni

PR #4278, opened 2024-06-13
feat: Change get_online_features signature, move online retrieval functions to utils

tokoko:

- Changes `get_online_features` signature to also accept entity_rows in columnar format. This allows `get_online_features` to be used from python feature server (previously it was calling `_get_online_features`).
- Refactors most of online retrieval functions from `feature_store.py` to `utils.py`. This is a first step for #4270, functions in utils no longer depend on `FeatureStore` object state.

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -11,12 +11,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import copy
 import itertools
 import logging
 import os
 import warnings
-from collections import Counter, defaultdict
 from datetime import datetime, timedelta
 from pathlib import Path
 from typing import (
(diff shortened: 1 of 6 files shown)


sdk/python/feast/feature_store.py
```diff
@@ -276,42 +265,19 @@ def list_feature_views(self, allow_cache: bool = False) -> List[FeatureView]:
...
-        return self._list_feature_views(allow_cache)
+        return utils._list_feature_views(self._registry, self.project, allow_cache)
 
     def _list_all_feature_views(
         self,
         allow_cache: bool = False,
     ) -> List[Union[FeatureView, StreamFeatureView, OnDemandFeatureView]]:
         all_feature_views = (
-            self._list_feature_views(allow_cache)
+            utils._list_feature_views(self._registry, self.project, allow_cache)
```
dmartinol, line 275: any reasons for not keeping the original name `utils._list_all_feature_views`?

franciscojavierarceo (approved):
nice

PR #4282, opened 2024-06-16
feat: Enabling cached predictions for stateful models

franciscojavierarceo:

This PR creates a simple wrapper `get_online_predictions` using existing classes to allow Feast to calculate and store a score from a model by using a combination of a Feature View and an On Demand Feature View.

This allows users of small models (e.g., Linear/Logistic Regression or XGBoost) to be able to easily run models with Feast alone.

Importantly, this allows the client to make the model *stateful* by storing the model calculation state after execution, which means much faster latency for retrieval (many model scores are recalculated dynamically and this is unnecessary).

The `get_online_predictions` method unlocks the ability for users to easily materialize a model state online and then re-calculate it in real-time. 

There is more left to do (e.g., logging) and this can be extended to broader use cases but this offers a nice, simple way for users of Feast to get started with serving simple models.

More specifically this PR:
- creates `get_online_predictions` and `get_online_inference` as methods in the `FeatureStore` class
- creates `get_unserialized_online_features_response` as modifying the `online_features_response` inside of `get_online_features` required it so
    - I was able to get around this but I think keeping it is worth keeping
- adds `_convert_list_of_dicts_to_dicts_of_lists` function to utils 
- adds `sdk/python/tests/example_repos/example_feature_repo_with_online_predictions.py` as an example repo for a unit test
- adds `sdk/python/tests/unit/online_store/test_online_retrieval.py` to test calculating the score and the cached behavior
- adds `get-online-predictions` endpoint to `sdk/python/feast/feature_server.py`.

The usage would look like so:

```python
result = store.get_online_predictions(
    features=[
        "customer_profile:avg_orders_day",
        "customer_profile:age",
    ],
    entity_rows=[
        {"customer_id": "5"},
        {"customer_id": 5},
    ],
    cached_model_feature_reference="stored_customer_predictions:predictions",
    on_demand_model_feature_reference="risk_score_calculator:predictions",
    force_recompute=False,
    log_features=True,
).to_dict()
```
Where `get_online_predictions` is structured as:
```python
def get_online_predictions(
    self,
    features: List[str],
    entity_rows: List[Dict[str, Any]],
    cached_model_feature_reference: str,
    on_demand_model_feature_reference: str,
    force_recompute: bool,
    log_features: bool,
    full_feature_names: bool = False,
) -> OnlineResponse:
...
```

# Fixes


sdk/python/feast/feature_server.py
```diff
@@ -246,3 +262,48 @@ def start_server(
...
+    body = json.loads(body)
+    # Initialize parameters for FeatureStore.get_online_features(...) call
+    if "feature_service" in body:
+        features = store.get_feature_service(body["feature_service"], allow_cache=True)
+    else:
+        features = body["features"]
+    return features, body
+
+
+def _get_predictions(
```
franciscojavierarceo, line 277: @tokoko i can just move this to `store.infer`

tokoko: Yeah, sounds about right.

sdk/python/feast/feature_store.py
```diff
@@ -2689,6 +2689,53 @@ def get_validation_reference(
...
+           ],
+           mode="python",
+        )
+        def run_inference(data_source_1: dict[str, float], data_source_2: dict[str, float]) -> dict[str, float]:
+            # Run inference
+            output = model.predict({**data_source_1, **data_source_2}).to_dict()
+            return output
+        '''
+        # Get the feature views to use
+        if force_recompute:
```
franciscojavierarceo, line 2718: will need an or statement here if the prediction is missing or a way to error out

sdk/python/feast/feature_server.py
```diff
@@ -160,6 +153,32 @@ def write_to_online_store(body=Depends(get_body)):
             # Raise HTTPException to return the error message to the client
             raise HTTPException(status_code=500, detail=str(e))
 
+    @app.post("/get-online-predictions")
+    def get_predictions_endpoint(body=Depends(get_body)):
```
franciscojavierarceo, line 157:
```suggestion
    def get_online_predictions(body=Depends(get_body)):
```

tokoko:
A couple of early questions:
- Where's the model object? How will the method get ahold of it?
- What's the difference between `prediction_feature_name` and `model_feature_name`?

franciscojavierarceo:
> A couple of early questions:
> 
> - Where's the model object? How will the method get ahold of it?
> 
> - What's the difference between `prediction_feature_name` and `model_feature_name`?

Let me rename those function arguments to be more intuitive but basically one is the feature view and one is the on demand feature view. The ODFV serves as a calculator for now since FVs don't yet support transformations.

franciscojavierarceo:
@tokoko take another look and let me know what you think, I updated the function arguments.

tokoko:
@franciscojavierarceo Putting model inference in a transformation seems like an abuse that I'm not sure we should be encouraging 😄 but leaving that aside for a moment... No matter where we put model invocations (transformation or a separate Model object as in #4288), I think the main concern in the ticket was that we were adding prediction/inference semantics to a FeastureStore, which is also the case in here.

franciscojavierarceo:
> @franciscojavierarceo Putting model inference in a transformation seems like an abuse that I'm not sure we should be encouraging 😄 but leaving that aside for a moment... No matter where we put model invocations (transformation or a separate Model object as in #4288), I think the main concern in the ticket was that we were adding prediction/inference semantics to a FeastureStore, which is also the case in here.

@tokoko I've added a document section where I structured some additional thoughts about my opinion on this. Feel free to share your thoughts!

As a general point, a model is technically a feature transformation, the threshold at which a feature transformation deserves its own service is not a terribly clear boundary but a choice that, in my opinion, should be made by the owners of the service. We're simply offering a starting point for a platform teams to be able to move quickly, be impactful, and make some thoughtful design decisions. 

Adding in the model registry and KServe isn't a requirement for the way I've laid things out here, so I think separating caching a score (which is an enhancement I think is valuable) and the other concerns is worth doing.

tokoko:
@franciscojavierarceo True, technically one can put anything in a transformation and we can't really restrict that, but building additional features on top is a whole different conversation. I have two main issues with the solution outlined here:
- To echo @etirelli's concerns once again, I think this is even more confusing than adding explicit Model abstractions. Sometimes ODFVs will be transformations, in other instances invocations. Putting a model invocation in a transformation comes with a lot of headaches of it's own (how would one correctly load a model object for example or if you're calling an external endpoint, how would you do it asynchronously as transformations really aren't designed for that and so on). I'm not totally opposed to the idea of adding predict semantics somewhere, but I think we at least need to agree on general vision before going forward with it. Should it be a separate project or maybe feast but not `FeatureStore` client as `predict` clearly has no business in Feature Store and so on...
- Does this change really need additional API endpoints and a method in `FeatureStore`? Why don't we instead reframe the whole thing as a optional ODFV caching/materialization feature. When you create an ODFV, you would enable caching and this would trigger a creation of a physical online store table for an ODFV that would be used for caching. This would avoid the confusing method signature that needs two separate feature views to be provided, plus there would be no need for any sort of predict semantics.

HaoXuAI:
Does it impact the get_online_feature flow and creating the on_demand_feature?

What is the flow of getting prediction?
Is it get_online_prediction -> get_online_feature -> call_model -> prediction_response?

franciscojavierarceo:
>Does it impact the get_online_feature flow and creating the on_demand_feature?

No.

>Is it get_online_prediction -> get_online_feature -> call_model -> prediction_response?
The flow will be  `get_online_predictions(force_recompute=True)` and then `force_recompute=False` which would need to be set by the client (which is why `force_recompute` is available in the server. For example, when a customer onboards, you would set the score for the first time by setting this parameter to True. 

I have the opposite example in `test_get_online_predictions()`, to show that the behavior is cached and recalculated. 

>And is the model going to be served/included in the on_demand_feature_view?
Yes, but as I mention in the docs it's meant to be a light-weight serving (e.g., Logistic or Linear regression).

>I think it's a good idea to add the get_prediction api, but I think mixing it with on_demand_feature_view will make it hard to maintain.
Can you elaborate on your concern about maintenance here? I'm not sure I understand it as this just wraps `get_online_features` twice and adds some simple default behavior--so it's pretty light maintenance and I've already added the tests for it but maybe I'm not thinking through something.

franciscojavierarceo:
> @franciscojavierarceo True, technically one can put anything in a transformation and we can't really restrict that, but building additional features on top is a whole different conversation. I have two main issues with the solution outlined here:
> 
> * To echo @etirelli's concerns once again, I think this is even more confusing than adding explicit Model abstractions. Sometimes ODFVs will be transformations, in other instances invocations. Putting a model invocation in a transformation comes with a lot of headaches of it's own (how would one correctly load a model object for example or if you're calling an external endpoint, how would you do it asynchronously as transformations really aren't designed for that and so on). I'm not totally opposed to the idea of adding predict semantics somewhere, but I think we at least need to agree on general vision before going forward with it. Should it be a separate project or maybe feast but not `FeatureStore` client as `predict` clearly has no business in Feature Store and so on...

I'll try to elaborate on the vision more explicitly on the doc and we should discuss the approach more here or in a discussion. As a funny point to note, I found [this discussion topic](https://github.com/feast-dev/feast/discussions/1557) from back in May 2021. In short, a user wasn't sure how how online predictions were meant to work. Willem's original answer was to run Feast side by side with a model serving system. My view is that the transformer integration with KServe doesn't allow for the caching behavior as outlined here and we should revisit that approach grounded in the user experience.

I like having my PR available as it makes what I was discussing before far more concrete. We can definitely revise how we execute the model and whether that's via an ODFV or something else but that won't really change the goal I have here, which is caching the score and being able to recalculate it. That is the *first* problem I wanted to solve.

If we agree on this general pattern of cache and recalculate, we can then move beyond a simple ODFV/FV pairing and into external endpoints and extend the behavior for async calls but first we need to agree that storing predictions and having them orchestrated via Feast is a reasonable option. As mentioned, I think it is and this is something I would have built previously if I had the opportunity to for real customer problems (I would have called our internal inference service) because I care a lot about customer experience.

> * Does this change really need additional API endpoints and a method in `FeatureStore`? Why don't we instead reframe the whole thing as a optional ODFV caching/materialization feature. When you create an ODFV, you would enable caching and this would trigger a creation of a physical online store table for an ODFV that would be used for caching. This would avoid the confusing method signature that needs two separate feature views to be provided, plus there would be no need for any sort of predict semantics.

This would simpler if the `FeatureView` class had transformation support. If we add transformation support to `FeatureView` I would then swap this for it. This was the simplest approach using existing code so that I could highlight what I was talking about before. 

As you may recall, I've been discussing this since April and I've discussed this with my previous team since October 2023, so this is something I've wanted to do for a while. 

I'm very flexible on changing *how we do the caching* here (odfv/fv/kserve or whatever) but I feel strongly that we need to support this capability. I have used this approach in production (not in Feast but elsewhere in other software) since 2014 in almost every company I've worked at and it provides a lot of latency improvements for customers.

HaoXuAI:
> > Does it impact the get_online_feature flow and creating the on_demand_feature?
> 
> No.
> 
> > Is it get_online_prediction -> get_online_feature -> call_model -> prediction_response?
> > The flow will be  `get_online_predictions(force_recompute=True)` and then `force_recompute=False` which would need to be set by the client (which is why `force_recompute` is available in the server. For example, when a customer onboards, you would set the score for the first time by setting this parameter to True.
> 
> I have the opposite example in `test_get_online_predictions()`, to show that the behavior is cached and recalculated.
> 
> > And is the model going to be served/included in the on_demand_feature_view?
> > Yes, but as I mention in the docs it's meant to be a light-weight serving (e.g., Logistic or Linear regression).
> 
> > I think it's a good idea to add the get_prediction api, but I think mixing it with on_demand_feature_view will make it hard to maintain.
> 
> Can you elaborate on your concern about maintenance here? I'm not sure I understand it as this just wraps `get_online_features` twice and adds some simple default behavior--so it's pretty light maintenance and I've already added the tests for it but maybe I'm not thinking through something.

I remember the on_demand_feature_view will need to pickle/serialize the code into binary and register. I'm not sure if adding the model related code into it will make the prediction easy or not, as the user might encounter more issue while defining a prediction inside the on demand feature view.
I guess it's still ok because it depends on the user to define the on demand "transformation", but seems a bit anti-pattern as mixing a feature definition API with a prediction API.
Anyway, since you maintain the backward compatibility, and only introduce new API, I think we can experiment it for now and keep improving it.

franciscojavierarceo:
>I remember the on_demand_feature_view will need to pickle/serialize the code into binary and register. I'm not sure if adding the model related code into it will make the prediction easy or not, as the user might encounter more issue while defining a prediction inside the on demand feature view.

This has variance depending on how it's deployed and how the libraries are loaded along with the ODFV. For a lot of small models, you can just store their model object in the code base and deploy them with that.

>I guess it's still ok because it depends on the user to define the on demand "transformation", but seems a bit anti-pattern as mixing a feature definition API with a prediction API.

Yeah, giving the user the control to define things is kind of the point. I don't see it as an anti-pattern necessarily but I understand why some would see it that way. I think a score is cached as a separate thing or cached as a feature and if it's cached as a separate thing in a service (e.g., in a serving layer), then it needs a database and then it starts to move further into feature store territory. More importantly, when you start to want to use ensemble models or batch models in production, it really starts to look like a use case for a feature store. Batch predictions are probably one of the most popular use cases in production and between Feast and KServe it's not obvious where to use those and the obvious answer is to use Feast.

>Anyway, since you maintain the backward compatibility, and only introduce new API, I think we can experiment it for now and keep improving it.
Thanks! Very open to feedback on this and want to get input from the community.

tokoko:
> I'm very flexible on changing _how we do the caching_ here (odfv/fv/kserve or whatever) but I feel strongly that we need to support this capability.

I'm with you there, I think it makes a lot of sense to use feast for both online caching and storing precomputed values. I just think that we should try to achieve that w/o also incorporating the whole serving layer in feast with it's own sets of challenges (adaptive batching, model load/unloading, handling different serving runtimes and many more). Let me try to come up with something to make it more clear. I don't think it will be all that different from the implementation in this PR except that actual model invocation will be left out of feast (and more importantly, feast registry).

P. S. I think we should continue this in the ticket as the discussion already started there.

franciscojavierarceo:
I'm going to close this for now. I'm going to add some documentation for this though.

Issue #4283, opened 2024-06-16
Feat: New version of entity_key serDe


HaoXuAI:
**Is your feature request related to a problem? Please describe.**

The current entity_key serDe (version 2) is below:
```
def serialize_entity_key(
    entity_key: EntityKeyProto, entity_key_serialization_version=1
) -> bytes:
    """
    Serialize entity key to a bytestring so it can be used as a lookup key in a hash table.

    We need this encoding to be stable; therefore we cannot just use protobuf serialization
    here since it does not guarantee that two proto messages containing the same data will
    serialize to the same byte string[1].

    [1] https://developers.google.com/protocol-buffers/docs/encoding
    """
    sorted_keys, sorted_values = zip(
        *sorted(zip(entity_key.join_keys, entity_key.entity_values))
    )

    output: List[bytes] = []
    for k in sorted_keys:
        output.append(struct.pack("<I", ValueType.STRING))
        output.append(k.encode("utf8"))
    for v in sorted_values:
        val_bytes, value_type = _serialize_val(
            v.WhichOneof("val"),
            v,
            entity_key_serialization_version=entity_key_serialization_version,
        )

        output.append(struct.pack("<I", value_type))

        output.append(struct.pack("<I", len(val_bytes)))
        output.append(val_bytes)

    return b"".join(output)
```

e.g, for `sorted_keys = {tuple: 1} item_id` and `sorted_values = {tuple: 1} int64_val: 1\n` will give output:
`[b'\x02\x00\x00\x00', b'item_id', b'\x04\x00\x00\x00', b'\x08\x00\x00\x00', b'\x01\x00\x00\x00\x00\x00\x00\x00']`

This makes deserialization not doable. In order to deserialize we can append the "length" of value to the `join_key`, such as for the same test key and value we can get the output:
`[b'\x02\x00\x00\x00', b'\x07\x00\x00\x00', b'item_id', b'\x04\x00\x00\x00', b'\x08\x00\x00\x00', b'\x01\x00\x00\x00\x00\x00\x00\x00']`

Then we can deserialize the bytes to proto.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #4284, opened 2024-06-16
feat: Entity key deserialization

HaoXuAI:

* Create new version 3 of entity key serialization
* Create entity key deserialization

To avoid breaking changes the default version of serialization is still set to `2`.

Deserialization function is supported by adding the entity key length in the serialization output:
```
    for k in sorted_keys:
        output.append(struct.pack("<I", ValueType.STRING))
        if entity_key_serialization_version > 2:
            output.append(struct.pack("<I", len(k)))
```

https://github.com/feast-dev/feast/issues/4283

# Fixes

### DIFF
--- a/sdk/python/feast/infra/key_encoding_utils.py
+++ b/sdk/python/feast/infra/key_encoding_utils.py
@@ -66,6 +66,15 @@ def serialize_entity_key(
     serialize to the same byte string[1].
 
     [1] https://developers.google.com/protocol-buffers/docs/encoding
+
+    Args:
+        entity_key_serialization_version: version of the entity key serialization
+        version 1: int64 values are serialized as 4 bytes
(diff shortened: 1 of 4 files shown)


sdk/python/feast/infra/key_encoding_utils.py
```diff
@@ -23,6 +23,22 @@ def _serialize_val(
         raise ValueError(f"Value type not supported for Firestore: {v}")
 
 
+def _deserialize_value(value_type, value_bytes) -> ValueProto:
```
franciscojavierarceo, line 26: Can you add an explicit test of this function?

HaoXuAI: Sure. Will add one for it

franciscojavierarceo (approved):
lgtm!

PR #4298, opened 2024-06-19
chore: Remove existing providers from the repo

tokoko:
This is backwards-compatible change that removes all provider implementations except `PassthroughProvider` from the repo. All other providers become aliases to `PassthroughProvider`. Provider interface is left intact so 3rd party providers should not be affected.

- `file` and `sqlite` become default offline and online stores respectively regardless of the provider.
- `provider` is no longer a mandatory field in `feature_store.yaml`, it defaults to `PassthroughProvider`.


HaoXuAI (approved):

HaoXuAI:
nice. Provider is one of the most confusing things when I started to use feast lol.

Issue #4198, opened 2024-05-13
Security Model


tokoko:
**Is your feature request related to a problem? Please describe.**
This is an offshoot ticket from #4032. Feast is slowly approaching a state in which all major components can be deployed as remote servers. This enables us to start thinking about a comprehensive security model for access to each of them (registry, online store, offline store)

**Describe the solution you'd like**
Here's a high-level overview of what I'm expecting from the security model:

1. I'd avoid incorporating user management into feast as much as possible. We should probably have a pluggable authentication module (LDAP, OIDC, etc...) that takes user/password (or token), validates it and spits out the roles that have been assigned to this particular user. Each server will have to integrate with this module separately, http feature server will get user/pass from basic auth, grpc and flight will get them according to their own standard conventions and pass credentials to the module to get the list of assigned roles. (Somewhat inspired by Airflow [security model](https://developer.squareup.com/blog/secure-apache-airflow-using-customer-security-manager/))

2. (Option 1) We enrich Feature Registry to also contain information about the roles available in the system and each feast object should be annotated with permissions. In other words, the user would run `feast apply` with something like this
```
admin_role = FeastUserRole(name='admin')
reader_role = FeastUserRole(name='reader')

FeatureView(
	name=...
	schema=...
	...
	permissions={
		'read': [role],
		'write': [admin_role] 
	}
)
```
3. (Option 2) Another option is to try to mimic AWS IAM and brush up on our regexes. In this case instead of annotating objects with permissions, you're annotating roles with policies.
```
risk_role = FeastUserRole(
	name='team_risk_role',
	permissions=[
		FeastPermission(
			action='read', //read_online, read_offline, write_online, write_offline
			conditions=[{
				'name': 'very_important_*',
				'team': 'risk'
			}]
		)
	]
)

FeatureView(
	name='very_important_fv',
	schema=...
	...
	tags={
		'team': 'risk'
	}
)
```
The upside of the second approach is that it's a lot less invasive than the first one. You could potentially end up with a setup where permissions and objects are managed with some level of separation between them. I think I'm more in favor of this.

4. Once a server gets ahold of user roles and permission information from the registry, all components will apply the same "rules engine" to authorize the requested action.

dmartinol:
@tokoko we started thinking about a possible solution, that we can share after completing the internal review, but first we'd like to ask a few questions to verify our understanding of the initial requirements (*):
[...]

PR #4300, opened 2024-06-20
feat: Instrument Feast using Prometheus and OpenTelemetry

tsisodia10:
## What this PR does / why we need it:
This draft pull request introduces monitoring functionality for the Feast feature store(python server). It instruments the feature store using Prometheus and OpenTelemetry to capture system-level metrics.


tchughesiv (changes requested):
it would be a good idea to make this optional by adding something like a `--metrics`  flag to the CLI and a similar param to the helm chart as well. imo, this should not be started / configured unless the user chooses to enable it.

infra/charts/feast-feature-server/templates/instrumentation.yml
```diff
@@ -0,0 +1,20 @@
+apiVersion: opentelemetry.io/v1alpha1
+kind: Instrumentation
+metadata:
+  name: python-instrumentation
+  namespace: feast-test
+spec:
+  exporter:
+    endpoint: http://otelcol-collector.opentelemetry.svc.cluster.local:4318
```
redhatHameed, line 8: you can add these configuration into values.yaml file

sdk/python/feast/feature_store.py
```diff
@@ -1906,6 +2169,17 @@ def serve(
         registry_ttl_sec: int = 2,
     ) -> None:
         """Start the feature consumption server locally on a given port."""
+        print("Start Prometheus Server")
```
redhatHameed, line 2172: I think this should be separated from feature server . The metrics server should be start separately then feature server. Currently that is the part of feature server command.
you can add new command `feast serve_metrics` which will run the metrics endpoint

sdk/python/feast/infra/feature_servers/gcp_cloudrun/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM python:3.11-slim
```
redhatHameed, line 1: why needed new docker-file why not using existing one.

sdk/python/feast/feature_store.py
tokoko: Not sure how that would work, that would be a separate process, wouldn't it?

redhatHameed: Yes, will be separate processes as the code starting Prometheus server so I am suggesting to create independent server and users can use the command `feast serve_metrics` to start the server and then be accessed via http://localhost:port/metrics

tokoko: I'm not really a prometheus user myself, but my understanding is the way the client library works is that whatever metrics you send (or set) throughout your application code is picked up and exposed by the prometheus server running in the same process. If an app code and prometheus server are running in different processes, there'd be no way to communicate in order to set metrics. Or maybe I'm misreading the whole thing.

redhatHameed: I got your point, yes it should be in same process to collect the metrics and agree to go with @tchughesiv suggestion.

sdk/python/feast/feature_store.py
```diff
@@ -1895,6 +1905,259 @@ def _retrieve_from_online_store(
             )
         return read_row_protos
 
+    @staticmethod
```
tokoko, line 1908: @tsisodia10 I think this is some sort of merge issue, right? I moved all these static methods to utils in another PR.

infra/charts/feast-feature-server/templates/deployment.yaml
```diff
@@ -52,7 +57,8 @@ spec:
             - "serve"
             - "-h"
             - "0.0.0.0"
-            {{- end }}
+            {{- end }} 
+          command: ["feast", "serve", "-h", "0.0.0.0"]
```
tchughesiv, line 61: this should likely turn into something like -
```yaml
{{- if .Values.metrics }}
command: ["feast", "serve", "--metrics", "-h", "0.0.0.0"]
{{- else }}
command: ["feast", "serve", "-h", "0.0.0.0"]
```

infra/charts/feast-feature-server/templates/instrumentation.yml
```diff
@@ -0,0 +1,20 @@
+apiVersion: opentelemetry.io/v1alpha1
```
tchughesiv, line 1: this template should have a `{{- if .Values.metrics }}` check at the top of it.

infra/charts/feast-feature-server/templates/otel-collector-ls.yml
```diff
@@ -0,0 +1,55 @@
+apiVersion: opentelemetry.io/v1beta1
```
tchughesiv, line 1: same

infra/charts/feast-feature-server/templates/otel-collector.yml
```diff
@@ -0,0 +1,54 @@
+# API reference https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/api.md
```
tchughesiv, line 1: same

infra/charts/feast-feature-server/templates/rbac.yml
```diff
@@ -0,0 +1,69 @@
+apiVersion: v1
```
tchughesiv, line 1: same

infra/charts/feast-feature-server/templates/service-monitor.yml
```diff
@@ -0,0 +1,16 @@
+apiVersion: monitoring.coreos.com/v1
```
tchughesiv, line 1: same

infra/charts/feast-feature-server/templates/service.yaml
```diff
@@ -3,13 +3,17 @@ kind: Service
 metadata:
   name: {{ include "feast-feature-server.fullname" . }}
   labels:
-    {{- include "feast-feature-server.labels" . | nindent 4 }}
+    {{- include "feast-feature-server.labels" . | nindent 4 }}
```
tchughesiv, line 6: spaces were added to the end of this line, should be removed

infra/charts/feast-feature-server/values.yaml
```diff
@@ -6,10 +6,12 @@ replicaCount: 1
 
 image:
   # image.repository -- Docker image for Feature Server repository
-  repository: feastdev/feature-server
+  repository: quay.io/tsisodia10/feastdev/feature-server
```
tchughesiv, line 9: revert this line

infra/charts/feast-feature-server/values.yaml
```diff
@@ -6,10 +6,12 @@ replicaCount: 1
...
   pullPolicy: IfNotPresent
   # image.tag -- The Docker image tag (can be overwritten if custom feature server deps are needed for on demand transforms)
-  tag: 0.39.0
+  tag: dev9
```
tchughesiv, line 12: revert this line

infra/charts/feast-feature-server/values.yaml
```diff
@@ -6,10 +6,12 @@ replicaCount: 1
...
+
+namespace: "default"
```
tchughesiv, line 14: remove

infra/charts/feast-feature-server/values.yaml
```diff
@@ -38,6 +40,11 @@ service:
   type: ClusterIP
   port: 80
 
+otel:
+  exporter:
+    endpoint: "http://otel-collector.feast-test.svc.cluster.local:4317"
+    insecure: "true"
```
tchughesiv, line 46: this needs to be talked through... this endpoint should be more dynamically set correct?

sdk/python/feast/infra/feature_servers/gcp_cloudrun/Dockerfile
tchughesiv: +1 , do we need a separate image?

tchughesiv (changes requested):
getting there...
now, a similar bool `metrics` param should be added to the `feast-feature-server` helm chart. this param should be used to enable all metrics related configs when set. it should default to `false`. be sure to run `make build-helm-docs` afterward.

tokoko:
@tsisodia10 `prometheus-client` should be added to `setup.py` explicitly under an extra, I think (maybe `metrics` or `prometheus`). (Right now it's part of our `ci` dependencies only under `ge` extra "thanks to" a whole mess of unnecessary dependencies `great-expectations` library likes to drag along wherever it goes)

sdk/python/feast/cli.py
```diff
@@ -684,6 +684,12 @@ def init_command(project_directory, minimal: bool, template: str):
     default=5,
     show_default=True,
 )
+@click.option(
+    "--metrics",
+    is_flag=True,
+    show_default=True,
+    help="Disable the Metrics Server",
```
tchughesiv, line 691: shouldn't this be "**Enable** the Metrics Server"?

infra/charts/feast-feature-server/values.yaml
```diff
@@ -14,6 +14,12 @@ image:
 imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: ""
+namespace: ""
```
redhatHameed, line 17: do we need namepace ?? it should set automatically with ` -n, --namespace`

sdk/python/feast/cli.py
```diff
@@ -684,6 +684,13 @@ def init_command(project_directory, minimal: bool, template: str):
     default=5,
     show_default=True,
 )
+@click.option(
+    "--metrics",
+    "-m",
+    is_flag=True,
+    show_default=True,
+    help="Enable the Metrics Server",
```
redhatHameed, line 692: what is the purpose of    ` is_flag=True,
    show_default=True`, sorry if Missed any conversation related to that

infra/charts/feast-feature-server/templates/service.yaml
```diff
@@ -3,13 +3,17 @@ kind: Service
...
-    {{- include "feast-feature-server.labels" . | nindent 4 }}
+    {{- include "feast-feature-server.labels" . | nindent 4 }}  
 spec:
   type: {{ .Values.service.type }}
   ports:
     - port: {{ .Values.service.port }}
       targetPort: {{ .Values.feast_mode }}
       protocol: TCP
       name: http
+    - name: metrics
```
redhatHameed, line 14: I think you need to check here as well if the metics endpoint is enabled

infra/charts/feast-feature-server/values.yaml
redhatHameed: can you try with `{{ .Release.Namespace }}`

infra/charts/feast-feature-server/templates/deployment.yaml
tchughesiv: @tsisodia10 if the `--metrics` flag is NOT specified, prom and telemetry should NOT be enabled. Metrics should only enabled if the flag is used. Based on this understanding, we would need this helm check. Is this not currently how the cli flag is functioning?

tchughesiv: i don't see it being passed anywhere in the helm charts... how it is currently being enabled?

tchughesiv (changes requested):
@tsisodia10 it might be a good idea to add a test w/ this PR... maybe in `sdk/python/tests/integration/online_store`?

infra/charts/feast-feature-server/templates/instrumentation.yml
```diff
@@ -0,0 +1,22 @@
+{{- if .Values.metrics }}
+apiVersion: opentelemetry.io/v1alpha1
+kind: Instrumentation
+metadata:
+  name: feast-instrumentation
+  namespace: {{ .Release.Namespace }}
+spec:
+  exporter:
+    endpoint: http://{{ .Values.service.name }}-collector.{{ .Release.Namespace }}.svc.cluster.local:4318 # This is the default port for the OpenTelemetry Collector
```
dmartinol, line 9: The `svc.cluster.local` part could be omittes, IIUC (in both places, also previous file)
Also, the port numbers, shouldn't they be configurable with a default in the values file, instead?

infra/charts/feast-feature-server/templates/otel-collector-ls.yml
```diff
@@ -0,0 +1,57 @@
...
+              - targets: [ '0.0.0.0:8888' ]
+
+    processors:
+      batch: {}
+
+    exporters:
+      otlp/ls:
+        endpoint: "ingest.lightstep.com:443"
+        headers:
+          "lightstep-access-token": "${LS_TOKEN}"
```
dmartinol, line 39: where does `LS_TOKEN` come from? Do we need a secret to expose this value?

infra/charts/feast-feature-server/templates/deployment.yaml
dmartinol, line 1: I think that in the `command` section you should also add the optional `--metric` flag whenever the `metrics` value is set. Otherwise, how is it enabled at runtime?

dmartinol (commented):
One doubt about the collector deployment: the helm chart is designed to deploy any of the available servers using the `feast_mode` value, but any of them is also going to deploy their own collector. Shouldn't we need a separate chart for the collector instead?

+  @tchughesiv

sdk/python/feast/feature_server.py
dmartinol, line 1: I assume that also the other servers `offline_server` and `registry_server` will be instrumented the same way, correct?

infra/charts/feast-feature-server/templates/deployment.yaml
dmartinol: I mean, if you anabled the `metrics` value, then you should run `feast --metric ...` command, right?

tchughesiv: +1

sdk/python/feast/feature_server.py
tchughesiv: or, we could add a second check around feast_mode to ensure only certain modes trigger metrics deployment?
```yaml
{{- if .Values.metrics && eq .Values.feast_mode "online" }}
```

infra/charts/feast-feature-server/templates/deployment.yaml
tchughesiv: @dmartinol related thread

sdk/python/feast/feature_server.py
dmartinol: We can do anything, but it depends on our requirement: what server do we want to instrument with monitoring metrics? Since it's an internal ticket I suggest to create a new one for the remaining servers where we can also clarify the number of collectors that we want to deploy

infra/charts/feast-feature-server/templates/rbac.yml
```diff
@@ -0,0 +1,71 @@
+{{- if .Values.metrics }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: opentelemetry-targetallocator-sa
+  namespace: {{ .Values.namespace }}
```
redhatHameed, line 6: I think you need to update here to the namespace

infra/charts/feast-feature-server/templates/instrumentation.yml
```diff
@@ -0,0 +1,22 @@
+{{- if .Values.metrics }}
+apiVersion: opentelemetry.io/v1alpha1
+kind: Instrumentation
+metadata:
+  name: feast-instrumentation
+  namespace: {{ .Release.Namespace }}
```
tchughesiv, line 6: this line shouldn't be necessary ... helm should deploy to the appropriate namespace. the existing templates don't specify this line, so we should probably follow suit

infra/charts/feast-feature-server/templates/otel-collector.yml
```diff
@@ -0,0 +1,56 @@
...
+# Refs for v1beta1 config: https://github.com/open-telemetry/opentelemetry-operator/issues/3011#issuecomment-2154118998
+{{- if .Values.metrics }}
+apiVersion: opentelemetry.io/v1beta1
+kind: OpenTelemetryCollector
+metadata:
+  name: otelcol
+  namespace: {{ .Release.Namespace }}
```
tchughesiv, line 8: this line shouldn't be necessary ... helm should deploy to the appropriate namespace. the existing templates don't specify this line, so we should probably follow suit

infra/charts/feast-feature-server/templates/rbac.yml
tchughesiv: this line shouldn't be necessary ... helm should deploy to the appropriate namespace. the existing templates don't specify this line, so we should probably follow suit

tchughesiv (changes requested):

sdk/python/feast/feature_server.py
tokoko: Sorry if I'm digging up older discussions... why did we go with the `feast_mode` route again? This is probably something that would fit much more nicely with a single feast chart approach.

tchughesiv: @tokoko are you suggesting a better approach might be separate charts for each "mode" ? If so, I think we briefly discussed it as an option, and I probably took issue with the added maintenance something like that would introduce. Keeping 4 separate charts _mostly_ in sync, but not quite so... and then the separate releases, etc.

That said, i don't think its too late to pivot to something like that if everyone else prefers that kind of implementation. It certainly has its benefits. @redhatHameed @dmartinol thoughts?

redhatHameed: > @tokoko are you suggesting a better approach might be separate charts for each "mode" ? If so, I think we briefly discussed it as an option, and I probably took issue with the added maintenance something like that would introduce. Keeping 4 separate charts _mostly_ in sync, but not quite so, & vetted... and then the separate releases, etc.
> 
> That said, i don't think its too late to pivot to something like that if everyone else prefers that kind of implementation. It certainly has its benefits. @redhatHameed @dmartinol thoughts?

we also discussed different approached as mentioned in this [document](https://docs.google.com/document/d/e/2PACX-1vSdLMfBkhAhZFPkIXu0MN8_RIM46EUUY8b-5mPkop7S1JvdxLCY4W0B6KcAWCEcwMxMAMR8f9P8acOR/pub)

tokoko: @redhatHameed sorry, don't have access to it.
@tchughesiv No, I was referring to a "single chart for all" approach, not one for each. In that case values file would look something like this:
```
registry:
    enabled: true/false
    ....

online:
    enabled: true/false
    ....

offline:
    enabled: true/false
    ...
```

I remember this also came up in a PR discussion, but don't think there were any decisions made or most likely I just didn't stick around for that part 😄. I just meant that in this case, it would be simpler to have a shared collector for all services, which is hard when different feast services aren't part of a single helm release as @dmartinol pointed out.

redhatHameed: @tokoko sorry just attaching here again [Installing Feast multiple component with helm charts.docx](https://github.com/user-attachments/files/16150684/Installing.Feast.multiple.component.with.helm.charts.docx)

tchughesiv: are you saying each would be deployed separately if enabled, as part of the same install ... but all "mode" templates would live within the same chart? That would still require maintaining 4 different sets of templates... not much different than 4 different charts. The same issue(s) related to 4 _mostly_ similar templates having to be maintained would by present.

we opted for a single set of templates, with the flexibility of multiple installs in different modes ... e.g.
```shell
helm install feast-feature-server feast-charts/feast-feature-server ...
helm install feast-offline-server feast-charts/feast-feature-server --set feast_mode=offline  ...
helm install feast-ui-server feast-charts/feast-feature-server --set feast_mode=ui  ...
```

that said, i think we're asking the wrong question here.
should this feast chart be handling the deployment of the collector anyway? instead, shouldn't there be a set of values under the chart's metrics umbrella that allows the user to pass "collector" configs?

There's no reason for the feast-feature-server chart to be installing collector related deployment, service, clusterrole, etc. IMO, this kind of metrics cluster setup should be on the user to install/maintain. We'd be signing up for OpenTelemetry related maintenance as that product evolves or has bugs, etc. We should have no opinion on how, or if, that product is installed.

infra/charts/feast-feature-server/templates/otel-collector.yml
```diff
@@ -0,0 +1,56 @@
+# API reference https://github.com/open-telemetry/opentelemetry-operator/blob/main/docs/api.md
+# Refs for v1beta1 config: https://github.com/open-telemetry/opentelemetry-operator/issues/3011#issuecomment-2154118998
+{{- if .Values.metrics }}
+apiVersion: opentelemetry.io/v1beta1
+kind: OpenTelemetryCollector
```
tchughesiv, line 5: IMO, this chart should not be handling the installation of the collector. instead, we could offer the user a set collector related values under the `metrics` umbrella which would allow them to point to a collector if they so choose.

infra/charts/feast-feature-server/templates/rbac.yml
```diff
@@ -0,0 +1,71 @@
+{{- if .Values.metrics }}
+apiVersion: v1
+kind: ServiceAccount
```
tchughesiv, line 3: same, should be removed.

tchughesiv (changes requested):
We shouldn't be installing/maintaining the collector deployment. We'd be signing up for OpenTelemetry related maintenance as that product evolves or has bugs, etc. We should have no opinion on how, or if, that product is installed.

sdk/python/feast/feature_server.py
tchughesiv: that would also resolve the shared collector question.

tokoko: I'm not sure I agree on maintenance/user-experience tradeoff. A user would have to manage several different releases, which might be too much (or maybe not). Having said that, I agree on not managing the collector deployment, so that solves the problem, I guess.

dmartinol: I agree. For now, we can isolate the collector deployment. However, in the long run, we should adopt a broader perspective and possibly reconsider our initial decisions. From an operator's standpoint, we should offer options to install the entire Feast infrastructure in one go, including all its components (or 'micro-services,' to use standard terminology) and perhaps optional services like the collector. However, since the collector is 'external,' I agree that it should not be the operator's responsibility to install it: a documented sample deployment can be enough for now.

The aggregated chart deployment (aka, "umbrella chart" or "uber chart") can ensure consistent configuration across all components. This includes the recently developed configurations such as:
- Authorization server configuration (https://github.com/feast-dev/feast/issues/4198): why repeat the configuration for every server if the authorization server remains the same?
- Metrics collector configuration (this PR)

dmartinol: Also (just over-thinking aloud), each server needs its own YAML configuration, which is based on a 'seed YAML' template that was used to initialize and validate the store. An additional benefit of the aggregated deployment approach is that server configurations could be automatically generated from this seed YAML, with adjustments depending on the application's configuration.

E.g., the YAML for the `offline server` could be computed like:
```yaml
auth:
  type: # Defined by the operator depending on the application config
  ...
registry:
  registry_type: remote # Defined by the operator 
  path: my-registry:6570 # Defined by the operator using the registry server host and port
offline_store:
  .... # Same as the seed YAML 
```

While the one of the `registry server` is just like:
```yaml
registry:
  .... # Same as the seed YAML 
```

Finally (over-over-thinking), this approach could also generate a YAML configuration in a Secret/ConfigMap to be used by clients that want to connect the FeatureStore by mounting it as a volume, like:
```yaml
registry:
  registry_type: remote # Defined by the operator 
  path: my-registry:6570 # Defined by the operator using the registry server host and port
offline_store:
  type: remote
  host: my-offline_store # Defined by the operator using the offline server host
  port: 8815 # Defined by the operator using the offline server port
online_store:
  type: remote
  path: my-online_store:6570 # Defined by the operator using the online server host and port
```

tchughesiv: @tokoko @dmartinol if the majority wants a chart that's more extensive, that's ok by me. i do think we should stay out of the collector deployment business though. the chart should only ever deploy _feast_ related items.

dmartinol: agree for `only ever deploy feast related items`

dmartinol: 1- Yes. I would add sample manifests for the monitoring components and mention the deployment instructions in the docs as optional resources, w/o adding them to the chart. I think we don't need a separate chart for that, few hardcoded YAMLs are fine, then they can manually edit them or use their own collector instead.
2- Yes, please add a new task to instrument other servers.

tokoko: Agreed with @dmartinol, I think uber chart will have considerable benefits.

sdk/python/feast/feature_store.py
```diff
@@ -15,7 +15,11 @@
 import logging
 import os
 import warnings
-from datetime import datetime, timedelta
+import threading
```
tokoko, line 18: These imports are no longer necessary

setup.py
```diff
@@ -105,6 +105,18 @@
     "psycopg2-binary>=2.8.3,<3",
 ]
 
+PROMETHEUS_CLIENT_REQUIRED = [
+    "prometheus_client",
+]
+
+GRPCIO_REQUIRED = [
+    "grpcio-reflection",
```
tokoko, line 113: Where is this being used?

sdk/python/feast/feature_server.py
tchughesiv: @tsisodia10 IMO, you could just point to opentelemetry collector deployment docs instead... any example yamls we might include could become antiquated, or might not be considered best practice for the collector.

sdk/python/feast/infra/feature_servers/multicloud/Dockerfile
```diff
@@ -7,8 +7,7 @@ RUN apt update && \
         build-essential
 
 RUN pip install pip --upgrade
-RUN pip install "feast[aws,gcp,snowflake,redis,go,mysql,postgres]"
-
+RUN pip install "feast[aws,gcp,snowflake,redis,go,mysql,postgres,prometheus_client,psutil,grpcio]"
```
tchughesiv, line 10: is a similar change necessary in `Dockerfile.dev`?

infra/charts/feast-feature-server/samples/service-monitor.yml
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.metrics }}
```
tchughesiv, line 1: given that these are samples, no longer instrumented by the chart, these if statements should be removed.

infra/charts/feast-feature-server/samples/otel-sm.yml
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.metrics }}
```
tchughesiv, line 1: given that these are samples, no longer instrumented by the chart, these if statements should be removed.

infra/charts/feast-feature-server/samples/prometheus.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.metrics }}
```
tchughesiv, line 1: given that these are samples, no longer instrumented by the chart, these if statements should be removed.

tchughesiv (changes requested):

infra/charts/feast-feature-server/samples/instrumentation.yml
```diff
@@ -0,0 +1,21 @@
+{{- if .Values.metrics }}
+apiVersion: opentelemetry.io/v1alpha1
+kind: Instrumentation
+metadata:
+  name: feast-instrumentation
+spec:
+  exporter:
+    endpoint: http://{{ .Values.service.name }}-collector.{{ .Release.Namespace }}.svc.cluster.local:4318 # This is the default port for the OpenTelemetry Collector
```
tchughesiv, line 8: do helm value references make sense in these samples anymore? shouldn't it just be an example endpoint instead?

infra/charts/feast-feature-server/samples/otel-sm.yml
```diff
@@ -0,0 +1,18 @@
...
+metadata:
+  labels:
+    app: feast
+  name: otel-sm-1
+spec:
+  endpoints:
+    - port: metrics
+  namespaceSelector:
+    matchNames:
+      - {{ .Release.Namespace }}
```
tchughesiv, line 13: do helm value references make sense in these samples anymore?

infra/charts/feast-feature-server/samples/rbac.yml
```diff
@@ -0,0 +1,70 @@
...
+  name: opentelemetry-targetallocator-rb-1
+  annotations:
+    meta.helm.sh/release-name: "feast-release"
+    meta.helm.sh/release-namespace: "feast-val"
+  labels:
+    app.kubernetes.io/managed-by: "Helm"
+subjects:
+  - kind: ServiceAccount
+    name: opentelemetry-targetallocator-sa
+    namespace: {{ .Release.Namespace }}
```
tchughesiv, line 65: do helm value references make sense in these samples anymore?

infra/charts/feast-feature-server/templates/deployment.yaml
```diff
@@ -33,6 +34,10 @@ spec:
           env:
             - name: FEATURE_STORE_YAML_BASE64
               value: {{ .Values.feature_store_yaml_base64 }}
+            # - name: OTEL_EXPORTER_OTLP_ENDPOINT
+            #   value: http://{{ .Values.service.name }}-collector.{{ .Release.namespace }}.svc.cluster.local:4317
+            # - name: OTEL_EXPORTER_OTLP_INSECURE
+            #   value: "true"
```
tchughesiv, line 40: are we missing some logic here? should these be set if the values are set?

infra/charts/feast-feature-server/values.yaml
```diff
@@ -15,6 +15,16 @@ imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: ""
 
+metrics:
+  enabled: false
+  otelCollector:
+    endpoint: http://{{ .Values.service.name }}-collector.{{ .Release.namespace }}.svc.cluster.local  # sample endpoint: "otel-collector.default.svc.cluster.local:4317"
```
tchughesiv, line 21: do helm value references make sense here? maybe just set to "" as the default.

infra/charts/feast-feature-server/values.yaml
```diff
@@ -15,6 +15,16 @@ imagePullSecrets: []
...
+    port: 4317
+
+
+service:
```
tchughesiv, line 25: is this metrics related? if so, maybe it should live under that umbrella of values to help differentiate it? if not, at the very least the name of the value should better describe its use-case... maybe `otel_service`?

tchughesiv (changes requested):

infra/charts/feast-feature-server/README.md
```diff
@@ -55,4 +55,99 @@ See [here](https://github.com/feast-dev/feast/tree/master/examples/python-helm-d
...
\ No newline at end of file
+| tolerations | list | `[]` |  |
+
+## Adding Monitoring
+To add monitoring to the Feast Feature Server, follow these steps:
+
+### Configuration
+Update your Helm values file to enable monitoring and configure the OpenTelemetry Collector.
+
+Feast instrumentation Using OpenTelemetry and Prometheus - https://lucid.app/lucidchart/b07f9c86-e31b-4e19-8123-0a64ba532096/edit?page=0_0&invitationId=inv_b57cd30b-5741-436f-baac-2e4b547ac2a5#
```
dmartinol, line 66: Can you add a png image instead? Th link says "Lucid is retiring on Feb 24"

infra/charts/feast-feature-server/README.md
```diff
@@ -55,4 +55,99 @@ See [here](https://github.com/feast-dev/feast/tree/master/examples/python-helm-d
...
+
+```
+- name: OTEL_EXPORTER_OTLP_ENDPOINT
+    value: http://{{ .Values.service.name }}-collector.{{ .Release.namespace }}.svc.cluster.local:{{ .Values.metrics.endpoint.port}}
+- name: OTEL_EXPORTER_OTLP_INSECURE
+              value: "true"     
+```
+
+### Deploy manifests -
+Add Instrumentation, OpenTelemetryCollector, ServiceMonitors, Prometheus Instance and RBAC rules as provided in the samples/ directory.
```
dmartinol, line 121: maybe add a link to open the folder as `[samples/](./samples)`?

infra/charts/feast-feature-server/README.md
dmartinol, line 1: Shouldn't we create a separate md file for the collector topic and just link it from the README?

infra/charts/feast-feature-server/templates/deployment.yaml
dmartinol: Pls uncomment and use the configured value like `.Values.metrics.otelCollector.endpoint`, surrounded by the usual:
`{{- if .Values.metrics }}`

setup.py
```diff
@@ -105,6 +105,14 @@
     "psycopg[binary,pool]>=3.0.0,<4",
 ]
 
+PROMETHEUS_CLIENT_REQUIRED = [
```
tokoko, line 108: I'm also a little unsure about these two, feels too granular. What about changing to a single extra named `telemetry` that will install both `prometheus_client` and `psutil`.

tokoko: Can you call it just `opentelemetry` instead of `opentelemetry_required`?

tokoko:
@tsisodia10 Can you take care of a DCO check? I think the easiest way is to rebase all commits and force push a single signed commit (`git rebase HEAD~49 --signoff` and `git push --force-with-lease origin inst-stable`).

Also please run `make lint-python` and `make format-python` before pushing the changes.

infra/charts/feast-feature-server/templates/deployment.yaml
```diff
@@ -48,11 +49,20 @@ spec:
...
             - "serve_registry"
             {{- else }}
+            # Add the metric check if the metrics are enabled
+            # {{- if .Values.metrics }}
+            # - "feast"
+            # - "serve"
+            # - "--metrics"
+            # - "-h"
+            # - "0.0.0.0"
+            # {{- else }}
```
tchughesiv, line 59: shouldn't this block of configs be uncommented and enabled only when metrics are configured in the chart?

sdk/python/feast/infra/feature_servers/multicloud/Dockerfile
```diff
@@ -7,8 +7,7 @@ RUN apt update && \
         build-essential
 
 RUN pip install pip --upgrade
-RUN pip install "feast[aws,gcp,snowflake,redis,go,mysql,postgres]"
-
+RUN pip install "feast[aws,gcp,snowflake,redis,go,mysql,postgres,opentelemetry]"
```
tchughesiv, line 10: as previously discussed, the Dockerfile.dev file should also be modified accordingly

tchughesiv (changes requested):

infra/charts/feast-feature-server/templates/deployment.yaml
```diff
@@ -14,6 +14,7 @@ spec:
     {{- with .Values.podAnnotations }}
       annotations:
         {{- toYaml . | nindent 8 }}
+        # instrumentation.opentelemetry.io/inject-python: "true"
```
tchughesiv, line 17: does this need to stay? a check added?

tchughesiv (changes requested):

infra/charts/feast-feature-server/templates/deployment.yaml
tchughesiv: these chart changes should all be fully vetted in a cluster to ensure it continues to work as expected with or without metrics configured.

infra/charts/feast-feature-server/templates/deployment.yaml
tchughesiv: for sure! i was thinking it made more sense though to incorporate it into the metrics param chart checks. e.g. 
```yaml
{{- if .Values.metrics }}
```

Issue #4260, opened 2024-06-07
Add `online_read_async` method to Postgres Online Store


job-almekinders:
**Is your feature request related to a problem? Please describe.**
With the implementation of `get_online_features_async` to the `FeatureStore` class ([Issue](https://github.com/feast-dev/feast/issues/3927) and [PR](https://github.com/feast-dev/feast/pull/4172/files#diff-c2b25e2070b2be8b73b7574b51fd84470df80368d2746be111c1a3fb8793b2ee)), `online_read_async ` methods are required for each online store in order to be able to retrieve features in an async manner. 
This issue aims to implement this for the Postgres Online Store.

**Describe the solution you'd like**
Add a `online_read_async` method to Postgres Online Store.

However, the current Postgres implementation within `feast` leverages `psycopg2`, which doesn't support async operations. The newer version - `psycopg3` - does however [support](https://www.psycopg.org/psycopg3/docs/advanced/async.html) this.

Therefore, we would like to propose to implement this change with two separate PRs:
1. A PR which upgrades the `psycopg2` dependency to use `psycopg3`. The latter tries to behave as close as possible to the former, however, there are a few differences that need to be accounted for. A full list of difference can be found [here](https://www.psycopg.org/psycopg3/docs/basic/from_pg2.html).
2. A PR which implements the `online_read_async` method on the Postgres Online Store.

Would love to hear your thoughts on this approach before I start working on this! 

**Describe alternatives you've considered**
Another library to handle async postgres operations is [asyncpg](https://github.com/MagicStack/asyncpg). However, this repos latest release was 5 Nov, 2023. Furthermore, the last commit was three months ago, and the last 7 months only 5 commits are made in total. Since this seems quite limited, we don't think it would be a good move to add this library to the dependencies for the postgres extension within feast.

tokoko:
sounds good to me. I'd go with `psycopg3` as well. One thing to note is that postgres offline store also depends on `psycopg2`. I guess we can have both as dependencies as they're different libraries, but maybe we should look into upgrading offline store as well, if it's not too complicated to make a switch. wdyt?

job-almekinders:
I think that makes sense! 

Once I start with the implementation I'll see how much effort it is and whether we can pick it up in one go 👍

Issue #4036, opened 2024-03-23
Discussion: Pushing batches of data to online store: Should `conn.commit()` happen in the for loop or after?


job-almekinders:
[This](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/online_stores/contrib/postgres.py#L80-L101) piece of code in the `online_write_batch` function in the Postgres online store pushes data in batches to the online store. 

I was wondering whether it makes more sense to put the `conn.commit()` inside the for loop, or after the for loop. I would love to hear the different trade-offs between the two options! 

Copy of the code snippet here: 

```python
batch_size = 5000
for i in range(0, len(insert_values), batch_size):
    cur_batch = insert_values[i : i + batch_size]
    execute_values(
        cur,
        sql.SQL(
            """
            INSERT INTO {}
            (entity_key, feature_name, value, event_ts, created_ts)
            VALUES %s
            ON CONFLICT (entity_key, feature_name) DO
            UPDATE SET
                value = EXCLUDED.value,
                event_ts = EXCLUDED.event_ts,
                created_ts = EXCLUDED.created_ts;
            """,
        ).format(sql.Identifier(_table_id(project, table))),
        cur_batch,
        page_size=batch_size,
    )
    conn.commit() # << This is the point of interest. Should this be placed in the loop or after?  
```

tokoko:
@job-almekinders thanks for raising this. Using multiple commits (while not ideal) is not the end of the world here. note that depending on what materiliasation engine you're using you might still end up with multiple commits even if we pull the commit out of the loop (if using spark or bytewax).

Having said that, I think it seems to be impossible to make that change now anyway. Batch size of 5000 is a really small number for even the most common use cases. I have never used this in practice, but I suspect for feature views numbering several million entities and tens of features, the whole process probably takes ages. Leaving the connection uncommited for so long is not ideal either.

We could make batch size configurable first, but increasing it is also too risky. If I'm not mistaken (?) `execute_values` generates a single sql string with all the values as tuples inside. With high enough batch size, the size of that sql query could quickly get out of hand... So maybe it's time to give up and try using [adbc](https://arrow.apache.org/adbc/main/driver/postgresql.html) instead? [bulk writes](https://github.com/apache/arrow-adbc/pull/1093) with `adbc` look a lot more promising. What do you think?

job-almekinders:
Thank you for your reply and additional information, this helps a lot :) 

[...]

PR #4303, opened 2024-06-21
feat: Bump psycopg2 to psycopg3 for all Postgres components

job-almekinders:
This PR upgrades the `psycopg2` dependency to the newer `psycopg3` dependency.  See [here](https://www.psycopg.org/psycopg3/docs/basic/from_pg2.html) for more information on the differences between the two versions. 

This is the 1st out of 2 PRs, required to enable async feature retrieval for the Postgres Online Store. 

While here: 
- Makefile: Remove additional whitespace and exclude Snowflake tests for Postgres offline integration tests
- Make some functions more readable, and add some additional docstrings and typehints.
- Make the `batch_size` argument configurable for the postgres online store materialization function. This Fixes: https://github.com/feast-dev/feast/issues/4036

# Additional remarks

The changes in [this commit](https://github.com/feast-dev/feast/commit/b7de4b2c0fb6fd6e2519b3bb8f9ed976bf499c71) are related to the linter. In psycopg3, stricter type hints on the Cursor object require handling cases where cursor.description might be None. Although psycopg2 could also return None for this, it wasn't previously accounted for.

1st out of 2 PRs required to fix https://github.com/feast-dev/feast/issues/4260


sdk/python/feast/errors.py
```diff
@@ -401,3 +401,13 @@ def __init__(self, input_dict: dict):
...
+
+
+class ZeroRowsQueryResult(Exception):
+    def __init__(self, query: str):
+        super().__init__(f"This query returned zero rows:\n{query}")
+
+
+class ZeroColumnQueryResult(Exception):
+    def __init__(self, query: str):
+        super().__init__(f"This query returned zero columns:\n{query}")
```
job-almekinders, line 413: Exceptions to use for stricter handling of type hints of `psycopg3`

sdk/python/feast/infra/offline_stores/contrib/postgres_offline_store/postgres.py
```diff
@@ -331,16 +333,19 @@ def _get_entity_df_event_timestamp_range(
...
-                cur.execute(
-                    f"SELECT MIN({entity_df_event_timestamp_col}) AS min, MAX({entity_df_event_timestamp_col}) AS max FROM ({entity_df}) as tmp_alias"
-                ),
-            )
+            query = f"""
+                SELECT
+                    MIN({entity_df_event_timestamp_col}) AS min,
+                    MAX({entity_df_event_timestamp_col}) AS max
+                FROM ({entity_df}) AS tmp_alias
+                """
```
job-almekinders, line 343: No updates here, only re-formatting the query

sdk/python/feast/infra/online_stores/contrib/postgres.py
```diff
@@ -64,57 +75,56 @@ def online_write_batch(
             Tuple[EntityKeyProto, Dict[str, ValueProto], datetime, Optional[datetime]]
         ],
         progress: Optional[Callable[[int], Any]],
+        batch_size: int = 5000,
```
job-almekinders, line 78: Make configurable, addressing https://github.com/feast-dev/feast/issues/4036

sdk/python/feast/infra/online_stores/contrib/postgres.py
```diff
@@ -64,57 +75,56 @@ def online_write_batch(
...
+            ON CONFLICT (entity_key, feature_name) DO
+            UPDATE SET
+                value = EXCLUDED.value,
+                vector_value = EXCLUDED.vector_value,
+                event_ts = EXCLUDED.event_ts,
+                created_ts = EXCLUDED.created_ts;
+        """
+        ).format(sql.Identifier(_table_id(config.project, table)))
+
+        # Push data in batches to online store
```
job-almekinders, line 121: No changes here, only moving code further up in the function to make it more readable.

sdk/python/feast/infra/online_stores/contrib/postgres.py
```diff
@@ -64,57 +75,56 @@ def online_write_batch(
...
-            # Control the batch so that we can update the progress
-            batch_size = 5000
+                )
+
+        # Create insert query
+        sql_query = sql.SQL(
+            """
+            INSERT INTO {}
+            (entity_key, feature_name, value, vector_value, event_ts, created_ts)
+            VALUES (%s, %s, %s, %s, %s, %s)
```
job-almekinders, line 111: 1 out of 2 actual changes to the function:

We need to explicitly set the number of placeholder values.

sdk/python/feast/infra/online_stores/contrib/postgres.py
```diff
@@ -64,57 +75,56 @@ def online_write_batch(
...
-                            value = EXCLUDED.value,
-                            vector_value = EXCLUDED.vector_value,
-                            event_ts = EXCLUDED.event_ts,
-                            created_ts = EXCLUDED.created_ts;
-                        """,
-                    ).format(sql.Identifier(_table_id(project, table))),
-                    cur_batch,
-                    page_size=batch_size,
-                )
+                cur.executemany(sql_query, cur_batch)
```
job-almekinders, line 125: 2 out of 2 actual changes to the function:

The `psycopg2.extras.execute_values` functionality is removed in `psycopg3`. The maintainer of `psycopg3` advices to use `executemany`. See https://github.com/psycopg/psycopg/issues/576 and https://github.com/psycopg/psycopg/issues/114

sdk/python/feast/infra/online_stores/contrib/postgres.py
```diff
@@ -172,7 +182,9 @@ def online_read(
             # when we iterate through the keys since they are in the correct order
             values_dict = defaultdict(list)
             for row in rows if rows is not None else []:
-                values_dict[row[0].tobytes()].append(row[1:])
+                values_dict[
+                    row[0] if isinstance(row[0], bytes) else row[0].tobytes()
+                ].append(row[1:])
```
job-almekinders, line 187: Only call `tobytes()` when `row[0]` is not already of bytes type. Otherwise, this will result in Errors.

sdk/python/feast/infra/utils/postgres/connection_utils.py
```diff
@@ -1,50 +1,59 @@
...
+    """Get the additional `kwargs` required for connection objects."""
+    return {
+        "sslmode": config.sslmode,
+        "sslkey": config.sslkey_path,
+        "sslcert": config.sslcert_path,
+        "sslrootcert": config.sslrootcert_path,
+        "options": "-c search_path={}".format(config.db_schema or config.user),
+    }
+
+
```
job-almekinders, line 56: Helper functions to prevent code duplication in the above methods.

sdk/python/feast/infra/utils/postgres/connection_utils.py
```diff
@@ -63,16 +72,14 @@ def df_to_postgres_table(
...
         cur.execute(_df_to_create_table_sql(df, table_name))
-        psycopg2.extras.execute_values(
-            cur,
-            f"""
-            INSERT INTO {table_name}
-            VALUES %s
-            """,
-            df.replace({np.NaN: None}).to_numpy(),
-        )
+        cur.executemany(query, values)
```
job-almekinders, line 82: Moved the parsing of variables further to the top of the function. 

1. Again, we need to replace `execute_values` by `executemany`. 
2. Again, we need to explicitly set the number of placeholders. Since this function should be able to handle a dynamic amount of columns, we use the `placeholders` variable

sdk/python/tests/integration/online_store/test_universal_online.py
```diff
@@ -39,13 +39,18 @@
 
 @pytest.mark.integration
 @pytest.mark.universal_online_stores(only=["postgres"])
+@pytest.mark.parametrize(
+    "conn_type",
+    [ConnectionType.singleton, ConnectionType.pool],
+    ids=lambda v: f"conn_type:{v}",
+)
```
job-almekinders, line 46: Test both ConnectionTypes

job-almekinders (commented):
Some additional clarifications from my side!

sdk/python/requirements/py3.10-requirements.txt
```diff
@@ -20,32 +20,38 @@ charset-normalizer==3.3.2
     # via requests
 click==8.1.7
     # via
+    #   feast (setup.py)
```
tokoko, line 23: Did you use `lock-python-dependencies-all` to generate these filese? `feast (setup.py)` lines shouldn't have been added, I think.

job-almekinders: Yes, I used that command indeed!

sdk/python/tests/integration/registration/test_universal_registry.py
```diff
@@ -149,7 +149,9 @@ def pg_registry():
 
     registry_config = RegistryConfig(
         registry_type="sql",
-        path=f"postgresql://{POSTGRES_USER}:{POSTGRES_PASSWORD}@{container_host}:{container_port}/{POSTGRES_DB}",
+        # The `path` must include `+psycopg` in order for `sqlalchemy.create_engine()`
+        # to understand that we are using psycopg3.
+        path=f"postgresql+psycopg://{POSTGRES_USER}:{POSTGRES_PASSWORD}@{container_host}:{container_port}/{POSTGRES_DB}",
```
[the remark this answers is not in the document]
job-almekinders: Yes :( More context can be found [here](https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#module-sqlalchemy.dialects.postgresql.psycopg) btw!

docs/tutorials/using-scalable-registry.md
```diff
@@ -49,7 +49,7 @@ When this happens, your database is likely using what is referred to as an
 in `SQLAlchemy` terminology. See your database's documentation for examples on
 how to set its scheme in the Database URL.
 
-`Psycopg2`, which is the database library leveraged by the online and offline
+`Psycopg`, which is the database library leveraged by the online and offline
```
franciscojavierarceo, line 52:
```suggestion
`Psycopg3`, which is the database library leveraged by the online and offline
```

franciscojavierarceo: Oh I see

sdk/python/tests/integration/registration/test_universal_registry.py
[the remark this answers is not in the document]
franciscojavierarceo: This is indeed unfortunate. Let's plan on drafting comms about this then.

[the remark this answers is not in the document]
TomSteenbergen: A (somewhat hacky) alternative is to add a `pydantic` [field_validator](https://docs.pydantic.dev/latest/api/functional_validators/#pydantic.functional_validators.field_validator) to the `RegistryConfig` class that replaces `postgresql` with `postgresql+psycopg` in the path if `registry_type` equals `sql` and `path` starts with `postgresql`, but not with `postgresql+psycopg`.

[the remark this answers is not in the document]
TomSteenbergen: A (somewhat hacky) alternative is to add a `pydantic` [field_validator](https://docs.pydantic.dev/latest/api/functional_validators/#pydantic.functional_validators.field_validator) to the `RegistryConfig` class that replaces `postgresql` with `postgresql+psycopg` in the path if `registry_type` equals `sql` and `path` starts with `postgresql`, but not with `postgresql+psycopg`.

What are your thoughts on this @tokoko @franciscojavierarceo ?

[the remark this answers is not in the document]
tokoko: I think that makes sense, but we should also take into account that someone out there might still want to use `psycopg2` behind the registry. So I think the validator should check that only `psycopg3` is installed first and add `+psycopg` to the url with an appropriate warning there as well.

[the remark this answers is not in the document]
TomSteenbergen: > I think that makes sense, but we should also take into account that someone out there might still want to use psycopg2 behind the registry.

Is this still a possible scenario after this PR gets merged? IIUC, all usage of `psycopg2` in `feast` will be replaced by `psycopg`. So `psycopg2` will no longer be a transitive dependency when installing `feast` with extras `postgres`.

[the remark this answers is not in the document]
TomSteenbergen: > there's no reason for the user can't keep using psycopg2

@tokoko But would a user really care that `feast` uses specifically `psycopg2` for the registry instead of the newer and better `psycopg`? 

I'd imagine users to care primarily about using `postgres` for their registry (opposed to some other type of registry), but not really care about whether `feast` is leveraging `psycopg2` or `psycopg` under the hood for its Postgres database operations.

> So I think the validator should check that only psycopg3 is installed first and add +psycopg to the url with an appropriate warning there as well.

This is still something that could be added of course. Just curious if this is really necessary.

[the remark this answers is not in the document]
tokoko: It's not important, I'm probably being pedantic here. I was thinking of a case when some connection param works with only `psycopg2`, that might be confusing.

But since there's a way to explicitly specify `psycopg2` if a user really wants too (`postgresql+psycopg2`), let's just throw a warning if only `postgresql` is specified and silently change to `postgresql+psycopg`.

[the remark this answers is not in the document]
job-almekinders: Thanks for your suggestions! 

I just pushed another commit implementing the suggestion you made. Can you perhaps take another look? :) @tokoko

sdk/python/requirements/py3.10-requirements.txt
tokoko: Not sure honestly, I'll try to look into it. We can still merge regardless, it's not a blocker, just a bunch of extra line changes in the PR.

sdk/python/feast/infra/offline_stores/contrib/postgres_offline_store/postgres.py
```diff
@@ -274,8 +274,10 @@ def to_sql(self) -> str:
     def _to_arrow_internal(self, timeout: Optional[int] = None) -> pa.Table:
         with self._query_generator() as query:
             with _get_conn(self.config.offline_store) as conn, conn.cursor() as cur:
-                conn.set_session(readonly=True)
+                conn.read_only = True
                 cur.execute(query)
+                if not cur.description:
+                    raise ZeroColumnQueryResult(query)
```
tokoko, line 280: I'm not necessarily against these error messages, but we probably want to do the same for all offline stores in that case. I'm also not sure what would actually trigger these errors.

job-almekinders: > I'm also not sure what would actually trigger these errors.

IIUC, these errors can be triggered when
1. The query returns zero rows.
2. The queried table doesn't have any columns in the table definition.

These errors will occur when using either one of `psycopg2` or `psycopg3`. I just did a small test. I spinned up a postgres database and created an empty table with no columns, and queries them using both versions.

```python
# psycopg2
>>> with conn.cursor() as cur: cur.execute("select * from empty_table;"); conn.commit(); cur.description
...
()
>>> with conn.cursor() as cur: cur.execute("select * from empty_table;"); res=cur.fetchone(); conn.commit(); print(res)
...
None

# psycopg3
>>> with conn.cursor() as cur: cur.execute("select * from empty_table;"); conn.commit(); cur.description
...
<psycopg.Cursor [TUPLES_OK] [INTRANS] (host=localhost database=postgres) at 0x15b6a8b40>
[]
>>> with conn.cursor() as cur: cur.execute("select * from empty_table;"); res=cur.fetchone(); conn.commit(); print(res)
...
<psycopg.Cursor [TUPLES_OK] [INTRANS] (host=localhost database=postgres) at 0x15b6a8880>
None

```

The only difference between the two implementations is that the type hints for `psycopg3` 'force' us to handle the case when `cur.description` or `cur.fetch()` (or a similar fetch operation) can be `None` or an empty list/tuple. For the `psycopg2` implementation the same situation can occur, however, the type hinting seems to be less strict, and thus the linter was not complaining. 

For full context, these are the errors that the linter is returning:
```
feast/infra/offline_stores/contrib/postgres_offline_store/postgres_source.py:118: error: Item "None" of "Optional[list[Column]]" has no attribute "__iter__" (not iterable)  [union-attr]
feast/infra/offline_stores/contrib/postgres_offline_store/postgres.py:281: error: Item "None" of "Optional[list[Column]]" has no attribute "__iter__" (not iterable)  [union-attr]
feast/infra/offline_stores/contrib/postgres_offline_store/postgres.py:344: error: Value of type "Optional[Any]" is not indexable  [index]
```

> I'm not necessarily against these error messages, but we probably want to do the same for all offline stores in that case

We could do the same for all offline stores. However, I think this is a peculiarity of psycopg2/3. Furthermore, I don't think these errors will actually be triggered ever, given the fact that the psycopg2 implementation seems to have almost identical behavior as psycopg3. 

Let me know what you think!

sdk/python/tests/integration/registration/test_universal_registry.py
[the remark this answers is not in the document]
tokoko: This looks good, thanks

sdk/python/feast/infra/offline_stores/contrib/postgres_offline_store/postgres.py
tokoko: Fair enough, ideally we'd have some sort of universal exception types that would span other stores as well, but these just seem like ways to suppress linter issues iiuc. It's unlikely that any realistic feast use path would actually hit those.

tokoko (approved):
LGTM

sdk/python/feast/infra/offline_stores/contrib/postgres_offline_store/postgres.py
job-almekinders: Yes exactly! My thoughts as well :)

job-almekinders:
Hey @franciscojavierarceo, would you perhaps be able to do another pass on this PR? :)

HaoXuAI:
Looks like I merged the other pr caused conflicts here. mind fix it then we can merge it?

job-almekinders:
> Looks like I merged the other pr caused conflicts here. mind fix it then we can merge it?

I just pushed the update! @HaoXuAI

job-almekinders:
We tried to install and run this feature branch already in one of our downstream projects. We noticed that there is an (edge) case where the required dependency combination of psycopg and sqlalchemy is not working properly. This is related to [this](https://github.com/feast-dev/feast/pull/4303#discussion_r1649012843) discussion.

One of Feast' requirements is [SQLAlchemy>=1](https://github.com/feast-dev/feast/blob/master/setup.py#L59). This enables feast to install either version 1 or 2. With this feature branch, the psycopg2 dependency is replaced with psycopg3. The `path` parameter of the `RegistryConfig` is being passed to sqlalchemy directly. However, the psycopg3 driver is only available for SQLAlchemy>=2. 
If users in downstream projects have an additional dependency of SQLAlchemy<2, Feast is still able to be installed properly, however, with SQLAlchemy==1. This version doesn't support the psycop3 driver however. Since psycopg2 is no dependency of feast anymore, the user will not be able to connect to the registry out of the box.

The question now is how to handle this. We think there are a number of potential solutions:
1) Update the [validator](https://github.com/feast-dev/feast/pull/4303/files#diff-c90c6781dcec3cdcc26843e013d3d6e9c1ce89d084b3a67e818fe76a0c7d0436R127-R140), add add addional version checks for sqlalchemy and/or psycopg2/3, and based on that decide which driver to use. For example:
    a) If SQLAlchemy>=2, use psycopg3 (which is always installed with feast)
    b) If SQLAlchemy<2, check for availability of psycopg2. 
    	If psycopg2 is available, add this to the path prefix and continue.
    	If psycopg2 is not available, raise an Exception and instruct the user to add it to their own dependencies.
2) Users have to install the psycopg2 dependency themselves in their downstream project, and use the `postgresql+psycopg2` prefix for the `path` parameter of the `RegistryConfig`. Potentially, we could also decide to drop the added validator, and instead communicate to the user how to handle this.

There are more ways to tackle this issue, but I'm curious to hear your thoughts on what you think might be the best way forward. 

@tokoko I would love to hear your thoughts as well on this matter if you have the time :)

HaoXuAI:
> We tried to install and run this feature branch already in one of our downstream projects. We noticed that there is an (edge) case where the required dependency combination of `psycopg` and `sqlalchemy` is not working properly. This is related to [this](https://github.com/feast-dev/feast/pull/4303#discussion_r1649012843) discussion.
> 
> One of Feast' requirements is [SQLAlchemy>=1](https://github.com/feast-dev/feast/blob/master/setup.py#L59). This enables `feast` to install either version 1 or 2. With this feature branch, the `psycopg2` dependency is replaced with `psycopg3`. The `path` parameter of the `RegistryConfig` is being passed to `sqlalchemy` directly. However, the `psycopg3` driver is only available for `SQLAlchemy>=2`. If users in downstream projects have an additional dependency of `SQLAlchemy<2`, Feast is still able to be installed properly, however, with `SQLAlchemy==1`. This version doesn't support the `psycop3` driver however. Since `psycopg2` is no dependency of feast anymore, the user will not be able to connect to the registry out of the box.
> 
> The question now is how to handle this. We think there are a number of potential solutions:
> 
> 1. Update the [validator](https://github.com/feast-dev/feast/pull/4303/files#diff-c90c6781dcec3cdcc26843e013d3d6e9c1ce89d084b3a67e818fe76a0c7d0436R127-R140), add add additional version checks for `sqlalchemy` and/or `psycopg2/3`, and based on that decide which driver to use. For example:
>    a) If `SQLAlchemy>=2`, use `psycopg3` (which is always installed with feast with postgres extras)
>    b) If `SQLAlchemy<2`, check for availability of `psycopg2`.
>    If `psycopg2` is available, add this to the path prefix and continue.
>    If `psycopg2` is not available, raise an Exception and instruct the user to add it to their own dependencies.
> 2. Users have to install the `psycopg2` dependency themselves in their downstream project, and use the `postgresql+psycopg2` prefix for the `path` parameter of the `RegistryConfig`. Potentially, we could also decide to drop the added validator, and instead communicate to the user how to handle this.
> 
> There are more ways to tackle this issue, but I'm curious to hear your thoughts on what you think might be the best way forward.
> 
> @tokoko I would love to hear your thoughts as well on this matter if you have the time :)

any reason we are not able to bump sqlAlchemy to 2.0+?

tokoko:
@job-almekinders Thanks for pointing that out. However, I don't believe this warrants any action from us. Those kinds of diamond dependencies are just part of life in libraries like this, unfortunately. I'm sure there are a few others similar to this lurking around that we haven't noticed before. 

If a user has some hard dependency on sqlalchemy1 and also uses postgres in feast only for the registry, they are free to forgo `postgres` extra, install `psycopg2` instead and set up sql registry path accordingly.

> any reason we are not able to bump sqlAlchemy to 2.0+?

I guess we can, but as long as we don't absolutely need to, it's better to leave it as is not to cause unnecessary diamond dependency problems to downstream libraries.

TomSteenbergen:
> any reason we are not able to bump sqlAlchemy to 2.0+?

@HaoXuAI This will probably break the Snowflake integration, as SQLAlchemy 2 isn't supported yet by the Snowflake Python libraries. See this relevant issue: https://github.com/snowflakedb/snowflake-sqlalchemy/issues/380

job-almekinders:
> @job-almekinders Thanks for pointing that out. However, I don't believe this warrants any action from us. Those kinds of diamond dependencies are just part of life in libraries like this, unfortunately. I'm sure there are a few others similar to this lurking around that we haven't noticed before.
> 
> If a user has some hard dependency on sqlalchemy1 and also uses postgres in feast only for the registry, they are free to forgo `postgres` extra, install `psycopg2` instead and set up sql registry path accordingly.

That makes sense !

If this is no blocker then I think we are good to move forward with this PR, at least from our end :)

HaoXuAI:
> > any reason we are not able to bump sqlAlchemy to 2.0+?
> 
> @HaoXuAI This will probably break the Snowflake integration, as SQLAlchemy 2 isn't supported yet by the Snowflake Python libraries. See this relevant issue: [snowflakedb/snowflake-sqlalchemy#380](https://github.com/snowflakedb/snowflake-sqlalchemy/issues/380)

I think snowflake can use snowflake-python module which doesn't depend on sqlAchemy.
Also since we are on python 3.9+ that shouldn't be many cases which still depending on sqlAchemy 1.4?

tokoko:
> I think snowflake can use snowflake-python module which doesn't depend on sqlAchemy.

I think he meant snowflake-backed sql registry, not a separate snowflake registry.

> Also since we are on python 3.9+ that shouldn't be many cases which still depending on sqlAchemy 1.4?

I'm fine with removing EOL versions if that will get us anything. Probably best to follow up in a different issue, though.

@HaoXuAI This is good to merge, right?

HaoXuAI (approved):

PR #4305, opened 2024-06-22
fix: Using one single function call for utcnow().

### LINKED ISSUE #4240: Retire/update the datetime.utcnow() function
### Specifications

- Version: 0.38
- Platform:
- Subsystem:

shuchu:

Use one single function _utc_now() to represent the datetime.utcnow().

This is a partial fix for deprecating the "datetime.utcnow()".  

# Parital Fixes #4240


shuchu:
Please do not close the issue #4240 .

The new PR will modify the behavior of _utc_now() to deprecate the datetime.uctnow().

PR #4307, opened 2024-06-22
fix: Using one single function call for utcnow().

### LINKED ISSUE #4240: Retire/update the datetime.utcnow() function

shuchu:

The `datetime.utcnow()` is about to be deprecated in Python3.12.  This pr use one single function to wrap the `datetime.utcnow()` for easy update and fix.

Please be aware this PR does not retire the `datetime.utcnow()`. 

# Fixes #4240

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -86,6 +86,7 @@ from feast.repo_config import RepoConfig, load_repo_config
 from feast.repo_contents import RepoContents
 from feast.saved_dataset import SavedDataset, SavedDatasetStorage, ValidationReference
 from feast.stream_feature_view import StreamFeatureView
+from feast.utils import _utc_now
 from feast.version import get_version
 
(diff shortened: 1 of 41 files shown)


tokoko:
Do we really need this? a one-liner util looks odd. Might as well change to the newer version directly.

shuchu:
> Do we really need this? a one-liner util looks odd. Might as well change to the newer version directly.

I was thinking about to provide a single function as an implementation of the utcnow(). It's easy to maintenance in case if someone create a new PR but using their own favor implementation of utcnow().   

The next step is to have a small PR to deprecate the current datetime.utcnow().  But I need to be very careful since some place in our code base will transfer the value of datetime.utcnow() to a string first and change it to a new string value.  The is a inexplicitly "assumption" about the length and format of the str(datetime.utcnow()).

Issue #4325, opened 2024-07-01
new Dask version lock pandas to 2.0+


HaoXuAI:
## Current Behavior
the current dask version:
`    "dask[dataframe]>=2024.4.2",` will lock the pandas to 2.0+. Not sure if this is required?
Related PR: https://github.com/feast-dev/feast/pull/4159
### Specifications

- Version:
- Platform:
- Subsystem:

## Possible Solution

change to `"dask[dataframe]>=2024.2.1"`

PR #4326, opened 2024-07-01
fix: Update dask version to support pandas 1.x

HaoXuAI:
fix issue: https://github.com/feast-dev/feast/issues/4325

# Fixes

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,7 +65,7 @@ REQUIRED = [
     "fastapi>=0.68.0",
     "uvicorn[standard]>=0.14.0,<1",
     "gunicorn; platform_system != 'Windows'",
-    "dask[dataframe]>=2024.4.2",
+    "dask[dataframe]>=2024.2.1",
 ]
 
 GCP_REQUIRED = [
--- a/sdk/python/requirements/py3.10-ci-requirements.txt
+++ b/sdk/python/requirements/py3.10-ci-requirements.txt
@@ -36,6 +36,10 @@ asttokens==2.4.1
     # via stack-data
 async-lru==2.0.4
     # via jupyterlab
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -65,7 +65,7 @@
     "fastapi>=0.68.0",
     "uvicorn[standard]>=0.14.0,<1",
     "gunicorn; platform_system != 'Windows'",
-    "dask[dataframe]>=2024.4.2",
+    "dask[dataframe]>=2024.2.1",
```
franciscojavierarceo, line 68: are we downgrading to support pandas 1?

tokoko:
@HaoXuAI Did you try running tests locally with `dask[dataframe]==2024.2.1`? I vaguely remember that we made some code changes in dask offline store when we introduced that lower bound.

setup.py
HaoXuAI: not downgrading exactly, just lower the bound of dask to relax the pandas version. I think force pandas to be 2.0+ maybe too tight, will cause conflicts with many other projects.

HaoXuAI:
> @HaoXuAI Did you try running tests locally with `dask[dataframe]==2024.2.1`? I vaguely remember that we made some code changes in dask offline store when we introduced that lower bound.

not sure, which test will fail? looks like it passed the ci. from the pr it doesn't require dask to be specific version: https://github.com/feast-dev/feast/pull/4159/files#diff-c91058a32040621ac54d6ea4194b302a420b7849f6e8561920d23a0e97acf69a

tokoko (approved):
sure, but CI isn't testing for lower bounds. Anyway, you're right, even if some feature of dask offline store doesn't work, it's not worth it to block users from using pandas v1.

PR #4328, opened 2024-07-02
feat: Extend `get_feature_view` to return stream and on demand feature views

### LINKED ISSUE #4235: Registry get_feature_view should read stream and on-demand feature views
`BaseRegistry` has a strange non-symmetric behavior with CRUD operations of stream and on-demand feature views. All 3 types of feature views are applied with a `apply_feature_view` method that takes `BaseFeatureView` as an input, but for getting/listing existing feature views, there are 3 separate methods (`get_feature_view`, `get_stream_feature_view`, `get_stream_feature_view` and similar for listing). This is very impractical in practice as normally when a user asks for a FV by name, there's no way of knowing which type of FV the name refers to.

`get_feature_view` (and probably `list_feature_views` as well, but that will be hard to do in a backwards-compatible way) should be rewritten to return `BaseFeatureView` objects.

tokoko:
* Fixes #4235

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -456,7 +456,7 @@ class FeatureStore:
 
     def get_feature_view(
         self, name: str, allow_registry_cache: bool = False
-    ) -> FeatureView:
+    ) -> BaseFeatureView:
         """
         Retrieves a feature view.
 
@@ -477,11 +477,15 @@ class FeatureStore:
         name: str,
         hide_dummy_entity: bool = True,
         allow_registry_cache: bool = False,
-    ) -> FeatureView:
+    ) -> BaseFeatureView:
         feature_view = self._registry.get_feature_view(
             name, self.project, allow_cache=allow_registry_cache
         )
-        if hide_dummy_entity and feature_view.entities[0] == DUMMY_ENTITY_NAME:
+        if (
+            isinstance(feature_view, FeatureView)
+            and hide_dummy_entity
+            and feature_view.entities[0] == DUMMY_ENTITY_NAME
+        ):
             feature_view.entities = []
         return feature_view
 
@@ -686,6 +690,7 @@ class FeatureStore:
                         name, hide_dummy_entity=False
                     )
 
+                assert isinstance(feature_view, FeatureView)
                 if not feature_view.online:
                     raise ValueError(
                         f"FeatureView {feature_view.name} is not configured to be served online."
@@ -1450,9 +1455,11 @@ class FeatureStore:
                 feature_view_name, allow_registry_cache=allow_registry_cache
             )
         except FeatureViewNotFoundException:
-            feature_view = self.get_feature_view(
+            fv = self.get_feature_view(
                 feature_view_name, allow_registry_cache=allow_registry_cache
             )
+            assert isinstance(fv, FeatureView)
+            feature_view = fv
         if df is not None and inputs is not None:
             raise ValueError("Both df and inputs cannot be provided at the same time.")
         if df is None and inputs is not None:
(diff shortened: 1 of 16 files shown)


tokoko:
@dmartinol Ended up having to make changes all over. Maybe you can find time to look through.

dmartinol (commented):
lgtm.
I just see a lot of `assert isinstance(fv, FeatureView)` that I hope are all correct as some of them, like the `OnDemandFV`, are only `BaseFV` instead. I cannot follow the complete flow, but I trust you 👍

tokoko:
That's mostly to accommodate the linter for now, I should probably add more feast-specific exceptions (for example if a user is asking to materialize an odfv, which doesn't make sense)

On a general note, I found that we have a following trade-off here:
- When `get_feature_view` returns all types a lot of most common operations become cleaner to implement, for example when a user is passing a list of fv names that might be any one of them, we no longer need crazy exception handling to figure out which method to call.
- In some other less common cases when only a single fv type makes sense, we will now have to check the type of the object returned, we can no longer rely on the fact that registry method used would do that for us. When we drop py3.9, we should probably start using python [match](https://www.geeksforgeeks.org/python-match-case-statement/) statements for this.

Despite the trade-off, I still think it's a worthwhile change. Still wanted to get your opinion.

sdk/python/feast/feature_store.py
```diff
@@ -477,11 +477,15 @@ def _get_feature_view(
...
         hide_dummy_entity: bool = True,
         allow_registry_cache: bool = False,
-    ) -> FeatureView:
+    ) -> BaseFeatureView:
         feature_view = self._registry.get_feature_view(
             name, self.project, allow_cache=allow_registry_cache
         )
-        if hide_dummy_entity and feature_view.entities[0] == DUMMY_ENTITY_NAME:
+        if (
+            isinstance(feature_view, FeatureView)
```
dmartinol, line 485: `utils._list_feature_views` has a similar logic but misses the `isinstance` check, don't we need it there too?

tokoko: Not yet, it's using `list_feature_views` to get the objects which is still `FeatureView` specific, this PR only makes changes to `get_feature_view`. I'll do that in a follow-up.

sdk/python/feast/infra/registry/proto_registry_utils.py
```diff
@@ -106,6 +106,21 @@ def get_feature_view(
...
+            and feature_view_proto.spec.project == project
+        ):
+            return StreamFeatureView.from_proto(feature_view_proto)
+
+    for on_demand_feature_view in registry_proto.on_demand_feature_views:
+        if (
+            on_demand_feature_view.spec.project == project
+            and on_demand_feature_view.spec.name == name
+        ):
+            return OnDemandFeatureView.from_proto(on_demand_feature_view)
```
dmartinol, line 122: the method is declared to return a `FeatureView`, isn't the linter complaining about that?
I think the reason is that `OnDemandFeatureView.from_proto` does not declare the retuned type 😉

dmartinol:
> * In some other less common cases when only a single fv type makes sense, we will now have to check the type of the object returned, we can no longer rely on the fact that registry method used would do that for us. When we drop py3.9, we should probably start using python [match](https://www.geeksforgeeks.org/python-match-case-statement/) statements for 

It we really have this need, can't we apply an optional filter to return only instances of the requested type with no need to further check?
Really pseudo-code:
```py
fv :  StreamFeatureView = fs.get_feature_view(name=..., accept_type=StreamFeatureView)
...

fvs: list[StreamFeatureView] = fs.list_feature_views(...,  accept_type=StreamFeatureView)
for fv in fvs:
    ...
```

tokoko:
> It we really have this need, can't we apply an optional filter to return only instances of the requested type with no need to further check?

Let me look into this. This is actually probably a better idea since we'll definitely need something like this for `list_feature_views` method anyway, so why not here as well? I was just trying to keep the interface simple, because of the chain of changes this causes in registry server and so on.

P.S. I'm afraid linter issues won't go away even in this case. type annotation will have to be `BaseFeatureView` and some asserts here and there will be necessary.

Issue #4309, opened 2024-06-24
PostgresOnlineStore: Improve materialization


job-almekinders:
**Is your feature request related to a problem? Please describe.**
The current implementation of `PostgreSQLOnlineStoreConfig.online_write_batch()` is quite slow. 

**Describe the solution you'd like**
One of the suggested options discussed [in this issue](https://github.com/feast-dev/feast/issues/4036) is using [bulk writes](https://github.com/apache/arrow-adbc/pull/1093) with [adbc](https://arrow.apache.org/adbc/main/driver/postgresql.html). 

**Describe alternatives you've considered**
Another option could be to use `COPY FROM` command. This [performance benchmark](https://www.timescale.com/blog/psycopg2-vs-psycopg3-performance-benchmark/) shows that the number of rows written per second is significantly larger compared to the batch insert we are doing now. 

This image shows the comparison from the performance benchmark: 
<img width="607" alt="Screenshot 2024-06-24 at 11 48 25" src="https://github.com/feast-dev/feast/assets/55230856/c5976c8c-93c2-484a-89b1-8fd42acfbf9a">

**Additional context**
I think we should benchmark both solutions and pick the one that achieves the best performance.

tokoko:
Performance comparison sounds good. Both use COPY command under the hood, so I wouldn't expect too much of a difference, but stranger things have happened :). `adbc` might have an upper hand in user-friendliness as input is just arrow tables, not sure how easy it will be to prepare the input for `psycopg` COPY command.

TomSteenbergen:
So unfortunately `COPY` (even using `adbc`) is not faster than the current `execute_values` of `psycopg2`. 

`COPY` excels when you are inserting a large file at once. However, the `materialize` method calls the `_materialize_one` method of the `LocalMaterializationJob` class, which loops over many smaller batches ([source](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/materialization/local_engine.py#L169)) using `for batch in table.to_batches(DEFAULT_BATCH_SIZE):`. `DEFAULT_BATCH_SIZE` is set to 10,000, but the `to_batches` method on a `pyarrow.Table` only uses that as a _max_ chunk size. The [docs](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.to_batches) note the following:

> Maximum number of rows for each RecordBatch chunk. Individual chunks may be smaller depending on the chunk layout of individual columns.

[...]

PR #4331, opened 2024-07-04
fix: Remove redundant batching in PostgreSQLOnlineStore.online_write_batch and fix progress bar

TomSteenbergen:

Since the materialization engine already batches the data and calls the `online_write_batch` method of the online store for each batch, we can get rid of the additional batching in the `PostgreSQLOnlineStore.online_write_batch` method, which will speed up materialization logic for Postgres. 

While there, fix the progress bar by updating it with the number of entities in the batch, instead of the number of entity * feature pairs.

https://github.com/feast-dev/feast/issues/4309


tokoko (approved):
lgtm

franciscojavierarceo (approved):

Issue #3823, opened 2023-10-30
google-cloud-bigquery 3.13.0 breaking change


tfusillo:
## Expected Behavior 
Successfully get the pyarrow datatype from a feast ValueType using the `feast_value_type_to_pa` function

## Current Behavior
Currently receive the error `feast.errors.FeastExtrasDependencyImportError: cannot import name 'ARROW_SCALAR_IDS_TO_BQ' from 'google.cloud.bigquery._pandas_helpers'`

## Steps to reproduce
Install feast with the latest GCP dependencies and try to convert a feast schema to pyarrow through the use of the function `feast_value_type_to_pa` 

### Specifications

- Version: 0.34.1
- Platform: Linux
- Subsystem: Debian 11 (bullseye)

## Possible Solution
Downgrade to google-cloud-bigquery 3.12.0

achals:
@tfusillo mind making a PR for this?

qianshi508:
I got the same error with 

> google-cloud-bigquery==0.32.1
> feast[redis,gcp]==0.33.1

sudohainguyen:
yep we should set the upper bound for bigquery client

fsalazarh:
Same issue here!! With `google-cloud-bigquery==3.13.0`.

Resolved by downgrading to 3.12.0

PR #4347, opened 2024-07-12
chore: Change arrow scalar ids usage

camenares:
[PENDING] 2. Run unit tests and ensure that they are passing: https://github.com/feast-dev/feast/blob/master/CONTRIBUTING.md#unit-tests

Changes where `_ARROW_SCALAR_IDS_TO_BQ` is imported, so that `google-cloud-bigquery` can be bumped. This PR only bumps to 3.13.0, but unblocks further bumping. `_ARROW_SCALAR_IDS_TO_BQ` was moved [here](https://github.com/googleapis/python-bigquery/pull/1679/files).

Related to https://github.com/feast-dev/feast/issues/3823, but provides a different solution than downgrading.


tokoko:
This will break all versions below 3.14, won't it? Maybe we should just copy the mapping over instead... Not ideal, but might be better than to depend on an internal helper function. @sudohainguyen wdyt?

sudohainguyen:
good fix!
@camenares could you help re-generate the requirements txt files?
to make sure all other packages are compatible 😄

sudohainguyen:
please also check that the change is backward compatible for those who using version less that 3.12

camenares:
@tokoko that's a good point, I think you're right, it would not work 3.12 (but will double check). 

I think there's three options:

1. Change `setup.py` to also strict pin to `3.13.0`
2. Do a try/catch and try to import from each helper
3. Copy the mapping over.

I will work on regenerating the requirements.txt's

tokoko:
@camenares option 2 also sounds good to me. We can resort to copying the mapping if some other issue arises in the future. Any reason for the upper bound set at `<3.14`. Why can't we change to `<4`? Are there some other known issues?

P.S. You need to sign all your commits (`git commit -s`). Follow the instructions [here](https://github.com/feast-dev/feast/pull/4347/checks?check_run_id=27388388106) to fix the DCO check.

tokoko:
Can you run `make lint-python` and `make format-python` commands?

camenares:
@tokoko yep! Fixing that now, I believe it's related to https://github.com/python/mypy/issues/1153 but testing this locally

camenares:
@tokoko I also noticed that allowing the bound to be `<4` failed the full integration tests (local were fine). Restricting to `<3.14.0` seems to let those pass though.

tokoko:
@camenares integration tests failure was unrelated, sorry about that. We accidentally merged a failing PR to master.

camenares:
@tokoko Oh! Thanks for that, I just figured it was worth trying `<3.14`. Let me bump to `<4.0` again then.

tokoko:
You will have to regenerate requirements files for the bump to take effect in ci.

camenares:
@tokoko `<4.0` worked! I see no additional changes after running `make lock-python-dependencies-all`.

tokoko (approved):
right, looks like `firebase-admin` lib is pinning with an upper bound, I'll look into that later. lgtm otherwise, thanks a lot

camenares:
Thanks a lot for the guidance @tokoko , looking forward to more Feast contributions in the future!

tokoko:
we have some strict rate limits on cloud services that our tests are often hitting, you can ignore it.

PR #4349, opened 2024-07-13
chore: Rename FileOfflineStore to DaskOfflineStore

### LINKED ISSUE #3864: Rename FileOfflineStore to DaskOfflineStore
`FileOfflineStore` uses dask under the hood and is in principle no different from any other offline store implementation that can work with file-based data formats. The name doesn't currently reflect that because the underlying execution engine was switched from pandas to dask later on. While the name is probably not important, I think the name change will be valuable to highlight that feast doesn't require one-to-one relationship between `DataSources` and `OfflineStores`. The related `FileSource` class (which should **NOT** be renamed) can be made to be used from other offline stores, for example `SparkOfflineStore`.

For backwards-compatibility, we can keep all user-facing classes (`FileOfflineStore` and maybe also `FileRetrievalJob`, `FileOfflineStoreConfig`) as aliases pointing to their new counterparts.

tokoko:
Fixes #3864


shuchu (approved):
lgtm

franciscojavierarceo:
This PR may not have been ideal.

tokoko:
No, it shouldn't be. both `file` and `dask` in `feature_store.yaml` still refer to this offline store. Users generally never had to interact directly with the classes that were renamed (`FileOfflineStore` or `FileOfflineStoreConfig`).

tokoko:
The main reason was to drive home the distinction between DataSources and OfflineStores. Despite somewhat confusing naming, DataSource in feast is the concept that controls storage, while OfflineStore only controls computation engine. Sometimes those are closely linked together, but in this case not. Having `FileOfflineStore` doesn't make much sense as file system is not what's doing the computation, it's actually dask. `FileSource`s can also be operated on by other offline store engines like duckdb and in the future spark, so calling dask engine `FileOfflinStore` was not consistent.

franciscojavierarceo:
Yeah that makes sense. I think we probably need to update the documentation to clarify this.

PR #4352, opened 2024-07-14
fix: Retire the datetime.utcnow().

### LINKED ISSUE #4240: Retire/update the datetime.utcnow() function

shuchu:
# Fixes  
Fix #4240


sdk/python/tests/unit/test_datetime.py
```diff
@@ -1,6 +1,19 @@
 # -*- coding: utf-8 -*-
 
+from datetime import datetime
+
+from feast.utils import _utc_now
 
 """
 Test the retirement of datetime.utcnow() function.
 """
+
```
tokoko, line 10: Do we need this file? I don't this we should be testing standard lib functions.

sdk/python/tests/utils/test_log_creator.py
```diff
@@ -117,7 +117,7 @@ def prepare_logs(
                     f"{destination_field}__status"
                 ].mask(
                     logs_df[f"{destination_field}__timestamp"]
-                    < (_utc_now() - view.ttl),
+                    < (_utc_now() - view.ttl).replace(tzinfo=None),
```
tokoko, line 120: just curious, why was this change necessary?

sdk/python/tests/unit/test_datetime.py
shuchu: got it.

sdk/python/tests/utils/test_log_creator.py
shuchu: Without it, a "TypeError: Invalid comparison between dtype=datetime64[ns] and datetime" error will show up. 
Kind of date type mismatch. 

```
sdk/python/tests/utils/test_log_creator.py:119: in prepare_logs
    logs_df[f"{destination_field}__timestamp"]
../venv311/lib/python3.11/site-packages/pandas/core/ops/common.py:76: in new_method
    return method(self, other)
../venv311/lib/python3.11/site-packages/pandas/core/arraylike.py:48: in __lt__
    return self._cmp_method(other, operator.lt)
../venv311/lib/python3.11/site-packages/pandas/core/series.py:6119: in _cmp_method
    res_values = ops.comparison_op(lvalues, rvalues, op)
../venv311/lib/python3.11/site-packages/pandas/core/ops/array_ops.py:330: in comparison_op
    res_values = op(lvalues, rvalues)
../venv311/lib/python3.11/site-packages/pandas/core/ops/common.py:76: in new_method
    return method(self, other)
../venv311/lib/python3.11/site-packages/pandas/core/arraylike.py:48: in __lt__
    return self._cmp_method(other, operator.lt)
../venv311/lib/python3.11/site-packages/pandas/core/arrays/datetimelike.py:985: in _cmp_method
    return invalid_comparison(self, other, op)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

left = <DatetimeArray>
['2024-07-12 21:47:22', '2024-07-13 01:34:44', '2024-07-12 18:00:00',
 '2024-07-13 09:09:28', '2024-07...7-15 10:25:15', '2024-07-15 14:36:33', '2024-07-15 14:36:33',
 '2024-07-15 18:00:00']
Length: 25, dtype: datetime64[ns]
right = datetime.datetime(2024, 7, 13, 18, 36, 34, 282931, tzinfo=datetime.timezone.utc), op = <built-in function lt>

    def invalid_comparison(left, right, op) -> npt.NDArray[np.bool_]:
        """
        If a comparison has mismatched types and is not necessarily meaningful,
        follow python3 conventions by:
    
            - returning all-False for equality
            - returning all-True for inequality
            - raising TypeError otherwise
    
        Parameters
        ----------
        left : array-like
        right : scalar, array-like
        op : operator.{eq, ne, lt, le, gt}
    
        Raises
        ------
        TypeError : on inequality comparisons
        """
        if op is operator.eq:
            res_values = np.zeros(left.shape, dtype=bool)
        elif op is operator.ne:
            res_values = np.ones(left.shape, dtype=bool)
        else:
            typ = type(right).__name__
>           raise TypeError(f"Invalid comparison between dtype={left.dtype} and {typ}")
E           TypeError: Invalid comparison between dtype=datetime64[ns] and datetime

../venv311/lib/python3.11/site-packages/pandas/core/ops/invalid.py:40: TypeError
```

tokoko: cool, thanks

sdk/python/tests/unit/test_datetime.py
shuchu: fair enough.  will delete it.

tokoko (approved):
lgtm

Issue #4066, opened 2024-04-02
Azure MSSQL Backend (and possibly others) Broken SQLAlchemy `Engine` objects interactions with `pandas.read_sql()`


peter-resnick:
## Expected Behavior 

`feast` cli commands like `feast plan` work when using a `mssql` offline store

## Current Behavior

Currently, repos with `mssql` offline store are broken when appying feast cli commands. The following occurs:

```
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/feast/repo_operations.py", line 218, in plan
    data_source.validate(store.config)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/feast/infra/offline_stores/contrib/mssql_offline_store/mssqlserver_source.py", line 215, in validate
    self.get_table_column_names_and_types(config)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/feast/infra/offline_stores/contrib/mssql_offline_store/mssqlserver_source.py", line 243, in get_table_column_names_and_types
    table_schema = pandas.read_sql(columns_query, conn)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/pandas/io/sql.py", line 706, in read_sql
    return pandas_sql.read_query(
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/pandas/io/sql.py", line 2736, in read_query
    cursor = self.execute(sql, params)
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/pandas/io/sql.py", line 2670, in execute
    cur = self.con.cursor()
AttributeError: 'Engine' object has no attribute 'cursor'
```
## Steps to reproduce

with a feature store YAML of:

```
project: myproject
provider: azure
registry:
  registry_type: sql
  path: ${SQL_REGISTRY_CONNECTION_STRING}
offline_store: 
  type: mssql
  connection_string: ${SQL_OFFLINE_CONNECTION_STRING}
online_store:
  type: redis
  redis_type: redis_cluster
  connection_string: ${REDIS_CONNECTION_STRING}
entity_key_serialization_version: 2
```
when we run `feast plan`  - we get the above errors.

This may impact other providers, but I've only confirmed it with `azure` provider
### Specifications

- Version: 0.35.0
- Platform: Mac + Linux (probably others)
- Subsystem:

[...]

PR #4360, opened 2024-07-18
feat: Port mssql contrib offline store to ibis

tokoko:
This PR refactors mssql offline store to use existing functions from `ibis.py` similar to the duckdb implementation. This will be the second ibis-based implementation so far. note that mssql is currently broken anyway because of a sqlalchemy version inconsistency. (#4066)

P.S. I'm still leaving mssql as a contrib. I'll add it to ci and move the code out from the contrib folder in a follow-up PR.


tokoko:
@tsisodia10 fyi

HaoXuAI (approved):
LGTM!

PR #4366, opened 2024-07-25
feat: Instrument Feast using Prometheus and OpenTelemetry

tsisodia10:
This draft pull request introduces monitoring functionality for the Feast feature store(python server). It instruments the feature store using Prometheus and OpenTelemetry to capture system-level metrics.


infra/charts/feast-feature-server/templates/deployment.yaml
```diff
@@ -14,6 +14,7 @@ spec:
     {{- with .Values.podAnnotations }}
       annotations:
         {{- toYaml . | nindent 8 }}
+        #instrumentation.opentelemetry.io/inject-python: "true"
```
tchughesiv, line 17: doesn't this need to be enabled when metrics is set?
```yaml
{{- if .Values.metrics }}
```

infra/charts/feast-feature-server/values.yaml
```diff
@@ -7,14 +7,26 @@ replicaCount: 1
 image:
   # image.repository -- Docker image for Feature Server repository
   repository: feastdev/feature-server
-  pullPolicy: IfNotPresent
+  pullPolicy: Always
```
tchughesiv, line 10: this should be reverted before merge

tchughesiv (changes requested):
looking good so far, needs some integration tests added

sdk/python/feast/feature_server.py
```diff
@@ -18,7 +19,12 @@
 from feast.constants import DEFAULT_FEATURE_SERVER_REGISTRY_TTL
 from feast.data_source import PushMode
 from feast.errors import PushSourceNotFoundException
+from prometheus_client import Gauge, start_http_server
+import psutil
 
+# Define prometheus metrics
```
franciscojavierarceo, line 25: Maybe it's not in scope for this but I would recommend adding latency metrics as well to the write and read endpoints.

infra/charts/feast-feature-server/values.yaml
```diff
@@ -7,14 +7,26 @@ replicaCount: 1
...
 nameOverride: ""
 fullnameOverride: ""
 
+metrics:
+  enabled: false
+  otelCollector:
+    endpoint: ""  # sample endpoint: "otel-collector.default.svc.cluster.local:4317"
+    port: 4317
+
+otel_service:
```
dmartinol, line 24: `otel_service` is duplicated and probably not even in use and can be removed

franciscojavierarceo:
FYI @emgeee

sdk/python/tests/integration/online_store/test_remote_online_store.py
```diff
@@ -217,6 +218,20 @@ def _start_feature_server(repo_path: str, server_port: int):
...
+    logging.debug(f"Starting Feast server with command: {' '.join(cmd)}")
+
+    # Check if metrics are enabled and Prometheus server is running
+    if metrics:
+        logging.debug("Metrics are enabled, checking Prometheus server on port 8000.")
+        wait_retry_backoff(
+            lambda: (None, check_port_open("localhost", 8000)),
+            timeout_secs=_time_out_sec,
+            timeout_msg="Unable to start the Prometheus server in 60 seconds.",
+        )
```
tchughesiv, line 233: can you add an `else` test that ensures its NOT running when metrics are disabled?

tchughesiv (changes requested):

infra/charts/feast-feature-server/templates/deployment.yaml
```diff
@@ -14,6 +14,9 @@ spec:
     {{- with .Values.podAnnotations }}
       annotations:
         {{- toYaml . | nindent 8 }}
+        {{- if .Values.metrics }}
```
tchughesiv, line 17: this should probably include the entire `annotations` block

tchughesiv (changes requested):

infra/charts/feast-feature-server/templates/deployment.yaml
tchughesiv: sorry, i think i was wrong about that. it appears `annotations` was preexisting. let's revert my ask

setup.py
```diff
@@ -58,6 +58,9 @@
     "fastapi>=0.68.0",
     "uvicorn[standard]>=0.14.0,<1",
     "gunicorn; platform_system != 'Windows'",
+    "dask[dataframe]>=2024.4.2",
```
tokoko, line 61: Why add these here? They are already in optional dependencies, I don't think we need any changes in required dependencies.

infra/charts/feast-feature-server/opentelemetry.md
```diff
@@ -0,0 +1,107 @@
+## Adding Monitoring
+To add monitoring to the Feast Feature Server, follow these steps:
+
+### Workflow
+
+Feast instrumentation Using OpenTelemetry and Prometheus -
```
tokoko, line 6: Might be a rookie question, but is both otel and prometheus deployment required in this setup? Won't it work with prometheus server running in feast with just an otel collector?

infra/charts/feast-feature-server/opentelemetry.md
```diff
@@ -0,0 +1,107 @@
+## Adding Monitoring
+To add monitoring to the Feast Feature Server, follow these steps:
+
+### Workflow
+
+Feast instrumentation Using OpenTelemetry and Prometheus - 
+![Workflow](samples/workflow.png)
+
+### Deploy Prometheus Operator
+Navigate to OperatorHub and install the stable version of the Prometheus Operator
```
tokoko, line 10: This sounds too OpenShift 😄

infra/charts/feast-feature-server/opentelemetry.md
tokoko: Sure, got that part. I'm asking if `kind: Prometheus` deployment is also strictly required alongside otel components?

tchughesiv: when the prior PR was closed, we lost a lot of prior discussion / decisions around this topic.
https://github.com/feast-dev/feast/pull/4300#discussion_r1672311834

README.md
```diff
@@ -16,10 +16,6 @@
 [![License](https://img.shields.io/badge/License-Apache%202.0-blue)](https://github.com/feast-dev/feast/blob/master/LICENSE)
 [![GitHub Release](https://img.shields.io/github/v/release/feast-dev/feast.svg?style=flat&sort=semver&color=blue)](https://github.com/feast-dev/feast/releases)
 
-## Join us on Slack!
```
tokoko, line 19: @tsisodia10 these changes are accididental, right?

infra/charts/feast-feature-server/templates/deployment.yaml
```diff
@@ -48,10 +51,18 @@ spec:
             - "feast"
             - "serve_registry"
             {{- else }}
+            {{- if .Values.metrics }}
```
tokoko, line 54: shouldn't this be checking `enabled` flag?

infra/charts/feast-feature-server/values.yaml
tokoko: +1

sdk/python/feast/feature_server.py
```diff
@@ -1,16 +1,19 @@
 import json
 import sys
 import threading
+import time
 import traceback
 from contextlib import asynccontextmanager
 from typing import List, Optional
 
 import pandas as pd
+import psutil
```
tokoko, line 10: These shouldn't be top-level imports here. The goal is to have both `psutil` and `prometheus_client` as optional dependencies, right? 

Maybe best to most of the changes from here to a new `metrics.py` file and import them only during `start_server` after checking `metrics` boolean flag first.

setup.py
```diff
@@ -59,6 +59,8 @@
     "uvicorn[standard]>=0.14.0,<1",
     "gunicorn; platform_system != 'Windows'",
     "dask[dataframe]>=2024.2.1",
+    "prometheus_client",
```
tokoko, line 62: These two lines also can be removed after refactoring `feature_server.py`

tokoko (approved):
lgtm. a couple of points to be handled in a follow-up

PR #4368, opened 2024-07-25
fix: Increment operator to v0.39.0

tchughesiv:
the feast-operator version fell behind, likely because it merged after the `0.37.1` release. because of this, 

# Fixes
matches operator version with the other files in `infra/scripts/release/files_to_bump.txt`. allows future scripted version increments to complete successfully, e.g. 
```shell
python ./infra/scripts/release/bump_file_versions.py 0.39.0 0.39.1
```

### DIFF
--- a/infra/feast-operator/Makefile
+++ b/infra/feast-operator/Makefile
@@ -3,7 +3,7 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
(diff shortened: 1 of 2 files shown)


franciscojavierarceo (changes requested):
I don't think we need this since https://github.com/feast-dev/feast/pull/4369

tchughesiv:
@franciscojavierarceo this one is for the `master` branch. the other is to fix the last release branch. we need both, `master` so that future releases are versioned correctly... and the `v0.39-branch` so that users who deploy the (newly released) operator image in that branch will experience a successful deployment.

tokoko (approved):

franciscojavierarceo (approved):
got it, thanks!

PR #4378, opened 2024-08-02
fix: Null value compatibility for unit timestamp list value type

EXPEbdodla:

# Fixes
1. Fixing ValueType.UNIX_TIMESTAMP_LIST to accept NULL values in data


tokoko (approved):
lgtm, thanks

Issue #4185, opened 2024-05-08
Split integration tests per provider


tokoko:
**Is your feature request related to a problem? Please describe.**
Integration tests are currently run tests for all implementations (offline, online, registry, materialization engine) in a single pytest call. There are several problems with this approach:
- It's easy to miss a required dependency in a particular extra as transitive dependencies from the whole project needs to be installed for tests to run.
- Other implementation dependency might lead to downgrading the dependency for an implementation that's being tested. This means to that we will have `library==0.1` tested in ci, while when user installs only the dependencies required for them, they will get `library==0.2`.
- We run it local tests twice in ci, once only for local tests and then again for all. With each implementation having a separate test job, we could delay only it tests with `ok-to-test` label.
- It's a chore to run tests only for the implementation that you're working on locally.

Cons:
- we will have to compile python environments for each action run instead of now having only ci requirements file in the repo.
- Lots of jobs in GH workflows (Is this a bad practce?)

**Describe the solution you'd like**
I'm thinking of splitting up into runs like this:

- offline tests - will run tests from `/offline_store` and `/materialization` folders. will use default online store (sqlite), registry (file) and all applicable mat. engines whenever necessary.
  - test-python-integration-offline-file
  - test-python-integration-offline-duckdb
  - test-python-integration-offline-spark
  - ...
- online tests - will run tests from `/online_store` and `/materialization` folders. will use default offline store (dask), registry (file) and all applicable mat. engines whenever necessary.
  - test-python-integration-online-sqlite
  - test-python-integration-online-redis
  - ...
- registry tests - will run tests from `/registration`
  - test-python-integration-registry-file
  - test-python-integration-registry-sql
  - test-python-integration-registry-s3
  - ...

P.S. There's also `e2e` folder in integration tests, but at this point I'm not convinced it's actually necessary. One file is testing `usage` functionality which is set to be deleted anyway. `test_universal_e2e.py` seems to be a misplaced materialization test.

PR #4380, opened 2024-08-05
feat: Feast Security Model (aka RBAC)

### LINKED ISSUE #4198: Security Model

dmartinol:
This PR proposes a permission model to enable the store administrator to identify which users are allowed to execute operations on each managed resource.

Granular permission enforcement identifies user roles (*) allowed to execute requested operations on 
resources matched by type, optional name patterns, and tags:

```py
Permission(
    name="reader",
    types=[FeatureView],
    name_pattern=".*risky.*",
    policy=RoleBasedPolicy(roles=["trusted"]),
    actions=[AuthzedAction.QUERY_OFFLINE],
)
```

Such resource partitioning allows teams working on the same set of features to share the same feature store, mitigating the risk of unexpected changes due to unauthorized operations.

The PR also adds support for managing authorization tokens from either OIDC or Kubernetes, in a configurable way.

Permission authorization enforcement is performed when requests are executed through one of the Feast (Python) servers:
* The online feature server (`REST)`
* The offline feature server (`Arrow Flight`)
* The registry server (`grpc`)

```yaml
project: foo
registry: "registry.db"
provider: local
online_store:
  path: foo
entity_key_serialization_version: 2
auth:
  type: kubernetes
```

To avoid backward incompatibility with existing installations, the default authorization is not enabled (e.g., `auth` type
is set to `no-auth`) .

Feast clients can use the same configuration options to automatically retrieve the token from the authorization server
and transparently secure all the remote requests. This pattern allows the client to use the authorization infrastructure without affecting the business code.

By design, only the client requests are validated: once the endpoint execution is permitted for the original client request, 
all the next service-to-service requests generated by the execution flow are automatically allowed.

Validation is typically implemented within the endpoint functions using an assertion-like style:
```py
        assert_permissions(
            resource=FeatureView.from_proto(request.feature_view),
            actions=[AuthzedAction.WRITE_ONLINE],
        )
```

The PR also includes:
- Docs update to provide general description of the feature
- Unit test for all the new modules
- Integration tests to verify the server behavior while adding the authorization components
- New `permission` CLI command with sub-commands to explore and troubleshoot the permissions settings

**Fully working examples both Kubernetes and OIDC authorization will be in another next PR.**

(*) Match by role is the predefined option, but it can be customized to support any other user validation.

# Other features coming soon
- Feature flag to override the store configuration and disable any permission validation
- Export validation metrics to telemetry collectors (related to #4366)

Fixes #4198


tokoko:
This is gonna take a while :smile:

sdk/python/feast/registry_server.py
```diff
@@ -24,31 +35,52 @@ def __init__(self, registry: BaseRegistry) -> None:
         self.proxied_registry = registry
 
     def ApplyEntity(self, request: RegistryServer_pb2.ApplyEntityRequest, context):
-        self.proxied_registry.apply_entity(
-            entity=Entity.from_proto(request.entity),
-            project=request.project,
-            commit=request.commit,
+        assert_permissions(
+            resource=self.proxied_registry.apply_entity(
```
tokoko, line 39: Is this correct? I don't think `apply_entity` returns anything.

setup.py
```diff
@@ -179,6 +180,8 @@
         "pytest-env",
         "Sphinx>4.0.0,<7",
         "testcontainers==4.4.0",
+        "python-keycloak==4.2.2",
+        "firebase-admin>=5.2.0,<6",
```
tokoko, line 184: Any reason this was added back?

redhatHameed: you mean ->  "firebase-admin>=5.2.0,<6", possible that added back due to rebase solving the conflicts.

protos/feast/core/Permission.proto
```diff
@@ -0,0 +1,58 @@
...
+      FEATURE_SERVICE = 5;
+      DATA_SOURCE = 6;
+      VALIDATION_REFERENCE = 7;
+      SAVED_DATASET = 8;
+      PERMISSION = 9;
+  }
+
+  repeated Type types = 3;
+
+  bool with_subclasses = 4;
```
tokoko, line 44: I like the idea of allowing all fv types to be configured together, but tying it to subclasses is gonna bite us in the end. odfvs don't extend `FeatureView` for example, while `StreamFeatureView` does. I think we shouldn't rely on that.

setup.py
tokoko: yeah, I removed it in #4359

protos/feast/core/Permission.proto
dmartinol: Do you mean to use only `FeatureView` type for all the possible fvs, whether they extend `FeatureView` or not?
And also provide a more punctual option for extensions like odfv and similar, or expose only the `FeatureView` type?
It'd be great if you can clarify your ask so we can adapt the code properly.

examples/rbac/client/admin_user_resources.yaml
```diff
@@ -0,0 +1,64 @@
...
+      app: client-admin
+  template:
+    metadata:
+      labels:
+        app: client-admin
+    spec:
+      serviceAccountName: feast-admin-sa
+      initContainers:
+        - name: init-client-admin
+          image: quay.io/ahameed/feature-server:dev
```
redhatHameed, line 47: Just missed to update the image this will set Feast image -> `feastdev/feature-server`

protos/feast/core/Permission.proto
tokoko: I'm leaning towards removing `with_subclasses` and instead adding some sort of `ALL_FEATURE_VIEWS` alias/helper, similar to `WRITE` and `QUERY`. wdyt?

docs/reference/registry/registry-permissions.md
```diff
@@ -0,0 +1,45 @@
...
+| ListValidationReferences | ValidationReference | Read                   | List validation references in the registry      |
+| DeleteValidationReference| ValidationReference | Delete                 | Delete a validation reference from the registry |
+| ApplyPermission          | Permission          | Create, Update, Delete | Apply a permission to the registry           |
+| GetPermission            | Permission          | Read                   | Get a permission from the registry           |
+| ListPermissions          | Permission          | Read                   | List permissions in the registry             |
+| DeletePermission         | Permission          | Delete                 | Delete a permission from the registry        |
+| Commit                   |                     | None                   | Commit changes to the registry               |
+| Refresh                  |                     | None                   | Refresh the registry                         |
+| Proto                    |                     | None                   | Get the proto representation of the registry |
```
redhatHameed, line 42: @tokoko any thoughts how to `assert_permissions` on this missing function.

protos/feast/core/Permission.proto
dmartinol: IMO, if we remove `with_subclasses`, the logic should be as if the flag was set to `True` (`OnDemandFeatureView` would be excluded by `type=FeatureView`, as you said before).
I would prefer a simple, understandable, filtering logic that applies to all the resource types, with no exceptions: consider that `DataSource` is just an abstract class and we don't have the actual classes in `FeastObject` type. To match data sources types we need to include the subclasses, which explains the reasons for my personal preference.

tokoko: Didn't think about DataSources.. yeah, that complicates things. Maybe the solution could be to rewrite `StreamFeatureView` to extend `BaseFeatureView` instead of `FeatureView`. That feels like how it should have been done from the get-go. Let me look into that and circle back.

HaoXuAI:
Hey, this is a great feature and also a big PR. Maybe it's better to walk us through the design, implementation, demo etc. in the next standup to get us understanding it better?

redhatHameed:
> Hey, this is a great feature and also a big PR. Maybe it's better to walk us through the design, implementation, demo etc. in the next standup to get us understanding it better?

sure, this will good idea.
cc @jeremyary @dmartinol

dmartinol:
@tokoko we realized that when the remote registry (transparently) invokes a server API, there's no way to catch a `FeatureViewNotFoundException` as in this [example](https://github.com/feast-dev/feast/blob/125fa49fc700a8e6f1994c710a5076210fb97bf9/sdk/python/feast/feature_store.py#L684), because the server raises an `InactiveRpcError` instead.

Should we create an issue to extend the registry server to add error details and the registry client to convert the grpc error into the original Exception? I don't think this has to be part of the RBAC commit.

@redhatHameed  FYI this is the origin of the error in the demo setup 
```text
Exception calling application: Feature view driver_hourly_stats_fresh does not exist in project server
```

tokoko:
@dmartinol sure, sounds like a bug to me. That should be tested in registry tests as well, I think. please go ahead and create a ticket.

franciscojavierarceo:
In the PR template you say

>"To avoid backward compatibilities with existing installations, the default authorization is not enabled (e.g., auth type is set to no-auth) ."

Did you mean "To avoid backward **incompatibility** with.."

docs/SUMMARY.md
```diff
@@ -16,6 +16,7 @@
   * [Feature retrieval](getting-started/concepts/feature-retrieval.md)
   * [Point-in-time joins](getting-started/concepts/point-in-time-joins.md)
   * [Registry](getting-started/concepts/registry.md)
+  * [Permissions](getting-started/concepts/permissions.md)
```
franciscojavierarceo, line 19: nit: I think "Role Based Access Controls (RBAC)" may be more explicit

docs/getting-started/concepts/permission.md
```diff
@@ -0,0 +1,143 @@
+# Permission
```
franciscojavierarceo, line 1:
```suggestion
# Role Based Access Control (RBAC) / Permissions
```

docs/getting-started/concepts/permission.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Overview
+
+The Feast permissions model allows to configure granular permission policies to all the resources defined in a feature store.
```
franciscojavierarceo, line 5:
```suggestion
The Feast RBAC model allows for the configuration of granular permission policies to all of the resources defined in a feature store.
```

docs/getting-started/concepts/permission.md
```diff
@@ -0,0 +1,143 @@
...
+
+The configured permissions are stored in the Feast registry and accessible through the CLI and the registry APIs.
+
+The permission authorization enforcement is performed when requests are executed through one of the Feast (Python) servers
```
franciscojavierarceo, line 9:
```suggestion
The RBAC enforcement is performed when requests are executed through one of the Feast (Python) servers
```

docs/getting-started/concepts/permission.md
```diff
@@ -0,0 +1,143 @@
...
+- The online feature server (REST)
+- The offline feature server (Arrow Flight)
+- The registry server (grpc)
```
franciscojavierarceo, line 12:
```suggestion
- The registry server (gRPC)
```

docs/getting-started/concepts/permission.md
```diff
@@ -0,0 +1,143 @@
...
+
+On the contrary, there is no permission enforcement when accessing the Feast API with a local provider.
```
franciscojavierarceo, line 14:
```suggestion
Note that there is no permission enforcement when accessing the Feast API with a local provider.
```

docs/getting-started/concepts/permission.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Concepts
+
+The permission model is based on the following components:
+- A `resource` is a Feast object that we want to secure against unauthorized access.
+  - We assume that the resource has a `name` attribute and optional dictionary of associated key-value `tags`.
+- An `action` is a logical operation executed on the secured resource, like:
+  - `create`: Create an instance.
+  - `read`: Access the instance state.
```
franciscojavierarceo, line 23: It is not obvious to me what the difference is between `read` and `query`. What do we mean "access" the instance state? Can you elaborate?

docs/getting-started/concepts/permission.md
```diff
@@ -0,0 +1,143 @@
...
+
+{% hint style="info" %}
+**Note**: Feast resources that do not match any of the configured permissions are not secured by any authorization policy, meaning any user can execute any action on such resources.
+{% endhint %}
+
+## Configuration examples
+This permission configuration allows to access the resource state and query all the stores for any feature view or feature service
+to all users with role `super-reader`:
+```py
+Permission(
```
franciscojavierarceo, line 58: Is this referring to a an instantiation of the FeatureView class? 

I would recommend creating a full [Minimal, Reproducible Example](https://stackoverflow.com/help/minimal-reproducible-example).

docs/getting-started/concepts/permission.md
```diff
@@ -0,0 +1,143 @@
...
+    name="feature-reader",
+    types=[FeatureView, FeatureService],
```
franciscojavierarceo, line 60: Is this referring to a an instantiation of the FeatureView class? 

I would recommend creating a full [Minimal, Reproducible Example](https://stackoverflow.com/help/minimal-reproducible-example).

docs/getting-started/concepts/permission.md
```diff
@@ -0,0 +1,143 @@
...
+teams to meet organizational requirements for service and data sharing, and protection of sensitive information.
+
+The `feast` CLI includes a new `permissions` command to list the registered permissions, with options to identify the matching resources for each configured permission and the existing resources that are not covered by any permission.
+
+{% hint style="info" %}
+**Note**: Feast resources that do not match any of the configured permissions are not secured by any authorization policy, meaning any user can execute any action on such resources.
+{% endhint %}
+
+## Configuration examples
+This permission configuration allows to access the resource state and query all the stores for any feature view or feature service
```
franciscojavierarceo, line 55:
```suggestion
This permission configuration allows access to the resource state and the ability to query all the stores for any feature view or feature service
```

docs/getting-started/concepts/permission.md
```diff
@@ -0,0 +1,143 @@
...
+Permission(
+    name="feature-reader",
+    types=[FeatureView, FeatureService],
+    policy=RoleBasedPolicy(roles=["super-reader"]),
+    actions=[AuthzedAction.READ, QUERY],
+)
+```
+Please note that all sub-classes of `FeatureView` are also included since the default for the `with_subclasses` parameter is `True`.
+
+This example grants permission to write on all the data sources with `risk_level` tag set to `hi` only to users with role `admin` or `data_team`:
```
franciscojavierarceo, line 67: I think you mean "high"

docs/getting-started/concepts/permission.md
```diff
@@ -0,0 +1,143 @@
...
+online_store:
+    path: foo
+entity_key_serialization_version: 2
+auth:
+  type: no_auth
+```
+
+* Kubernetes rbac authentication/authorization example:
+{% hint style="info" %}
+**NOTE**: This configuration will only work if you deploy feast on Openshift or Kubernetes platform.
```
franciscojavierarceo, line 114:
```suggestion
**NOTE**: This configuration will only work if you deploy feast on Openshift or a Kubernetes platform.
```

docs/reference/feast-cli-commands.md
```diff
@@ -24,6 +24,7 @@ Commands:
   init                     Create a new Feast repository
   materialize              Run a (non-incremental) materialization job to...
   materialize-incremental  Run an incremental materialization job to ingest...
+  permissions              Access permissions
```
franciscojavierarceo, line 27: probably we can have an `rbac` command as well that is the same as `permissions`

docs/reference/feature-servers/offline-feature-server.md
```diff
@@ -33,3 +33,20 @@ Please see the detail how to configure offline store client [remote-offline-stor
 
 The set of functionalities supported by remote offline stores is the same as those supported by offline stores with the SDK, which are described in detail [here](../offline-stores/overview.md#functionality).
 
+# Offline Feature Server Permissions and Access Control
```
franciscojavierarceo, line 36:
```suggestion
# Offline Feature Server Permissions and Role Based Access Control (RBAC)
```

docs/reference/feature-servers/python-feature-server.md
```diff
@@ -199,3 +199,19 @@ requests.post(
     "http://localhost:6566/push",
     data=json.dumps(push_data))
 ```
+
+# Online Feature Server Permissions and Access Control
```
franciscojavierarceo, line 203:
```suggestion
# Online Feature Server Permissions and Role Based Access Control (RBAC)
```

docs/reference/offline-stores/remote-offline-store.md
```diff
@@ -25,4 +25,7 @@ The complete example can be find under [remote-offline-store-example](../../../e
 
 ## How to configure the server
 
-Please see the detail how to configure offline feature server [offline-feature-server.md](../feature-servers/offline-feature-server.md)
\ No newline at end of file
+Please see the detail how to configure offline feature server [offline-feature-server.md](../feature-servers/offline-feature-server.md)
+
+## How to configure Authentication and Authorization
```
franciscojavierarceo, line 30:
```suggestion
## How to configure Authentication, Authorization, and Role Based Access Controls (RBAC)
```

docs/reference/online-stores/remote.md
```diff
@@ -11,11 +11,17 @@ The registry is pointing to registry of remote feature store. If it is not acces
...
+  type: remote
+entity_key_serialization_version: 2
+auth:
+  type: no_auth
 ```
-{% endcode %}
\ No newline at end of file
+{% endcode %}
+
+## How to configure Authentication and Authorization
```
franciscojavierarceo, line 25:
```suggestion
## How to configure Authentication, Authorization, and  Role Based Access Controls (RBAC)
```

docs/reference/registry/registry-permissions.md
```diff
@@ -0,0 +1,45 @@
+# Registry Permissions and Access Control
```
franciscojavierarceo, line 1:
```suggestion
# Registry Permissions and Role Based Access Controls (RBAC)
```

lokeshrangineni:
> In the PR template you say
> 
> > "To avoid backward compatibilities with existing installations, the default authorization is not enabled (e.g., auth type is set to no-auth) ."
> 
> Did you mean "To avoid backward **incompatibility** with.."

We no need to modify the existing test cases and as well as once the feast version is upgraded existing customers no need to modify `feature_repo.yaml` or explicitly define the auth settings.

docs/getting-started/concepts/permission.md
lokeshrangineni: I will incorporate it.

docs/getting-started/concepts/permission.md
lokeshrangineni: Incorporating it.

docs/getting-started/concepts/permission.md
lokeshrangineni: incorporating it.

docs/getting-started/concepts/permission.md
lokeshrangineni: Incorporating it.

docs/getting-started/concepts/permission.md
```diff
@@ -0,0 +1,143 @@
...
+- A `resource` is a Feast object that we want to secure against unauthorized access.
+  - We assume that the resource has a `name` attribute and optional dictionary of associated key-value `tags`.
+- An `action` is a logical operation executed on the secured resource, like:
+  - `create`: Create an instance.
+  - `read`: Access the instance state.
+  - `update`: Update the instance state.
+  - `delete`: Delete an instance.
+  - `query`:  Query both online and offline stores.
+  - `query_online`:  Query the online store.
+  - `query_offline`:  Query the o ffline store.
```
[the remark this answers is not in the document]
lokeshrangineni: Incorporating it.

docs/getting-started/concepts/permission.md
lokeshrangineni: I agree it needs to be clarified a bit more.  query is able to query all the feature view meta data(such as name) where as read will allow the user to read actual feature view data from online_store. I will update the PR.

dmartinol:
> @dmartinol sure, sounds like a bug to me. That should be tested in registry tests as well, I think. please go ahead and create a ticket.

@tokoko created https://github.com/feast-dev/feast/issues/4392 
Unfortunately I fear that the same behavior happens to all of the remote clients and, since we have different server protocols, we need different implementations for each of them 🤦

docs/getting-started/concepts/permission.md
dmartinol: ATM we've prepared a reproducible example including a fully distributed environment, in the `examples/rbac` folder, working on a K8s cluster with detailed instructions. 
To support users who are not familiar with K8s, we can prepare another example running all the servers from the local environment and using an OIDC authorization for which we can spin a Keycloak cluster. This is similar to what we did in our POC some time ago.

Would that be enough?

docs/SUMMARY.md
dmartinol: @tokoko do you agree with this naming? If yes, we can udate the docs accordingly (this and the other related comments)

lokeshrangineni (approved):

lokeshrangineni (changes requested):
I have approved the PR by accident. Reverting the approval.

Makefile
```diff
@@ -92,8 +92,8 @@ test-python-integration:
 test-python-integration-local:
 	FEAST_IS_LOCAL_TEST=True \
 	FEAST_LOCAL_ONLINE_CONTAINER=True \
-	python -m pytest -n 8 --color=yes --integration --durations=5 --dist loadgroup \
-		-k "not test_lambda_materialization and not test_snowflake_materialization" \
+	python -m pytest -n 4 --color=yes --integration --durations=20 --timeout=1200 --timeout_method=thread --dist loadgroup \
+		-k "not test_lambda_materialization and not test_snowflake_materialization and not test_remote_online_store_read" \
 		sdk/python/tests
```
redhatHameed, line 97: can you remove `and not test_remote_online_store_read`

lokeshrangineni: Done.

protos/feast/core/Permission.proto
tokoko: I opened #4398 to track this. I think we can for now continue as if that's already the case and we don't have inconsistent class inheritance. In other words, let's remove `with_subclasses` and checks will naturally default to True as you suggested.

docs/SUMMARY.md
tokoko: @dmartinol If we add some other type of policy (meaning other than `RoleBasedPolicy`), we would technically be going beyond rbac, right? From that point of view more generic title (Permission or Access Control) seems more appropriate.

setup.py
```diff
@@ -61,6 +61,9 @@
     "dask[dataframe]>=2024.2.1",
     "prometheus_client",
     "psutil",
+    "bigtree>=0.19.2",
+    "pyjwt",
+    "kubernetes<=20.13.0",
```
tokoko, line 66: This might require a bit of refactor, but this one definitely shouldn't be a mandatory dependency, especially since we already have an identical k8s extra.

tmihalac: Just a small reminder, It was required because the java integration tests were not passing even though kubernetes is already in CI_REQUIRED the java integration tests were still failing.
When you mean refactoring please elaborate

examples/rbac/README.md
```diff
@@ -0,0 +1,169 @@
...
+2. Helm: Ensure you have Helm installed for deploying the Feast components. 
+3. Python environment.
+4. Feast CLI latest version.
+
+### Steps
+
+#### Local Project Setup
+
+#### 1. **Create the Feast feature store project with PostgreSQL.**
+
```
franciscojavierarceo, line 39: It would be ideal to show this demo with Kind

examples/rbac/client/admin_user_resources.yaml
franciscojavierarceo: +1

protos/feast/core/Permission.proto
dmartinol: I will also add a global variable like ALL_FEATURE_VIEWS to explicitely list all of them.
Thanks

sdk/python/feast/registry_server.py
dmartinol: @redhatHameed I think the code should change from:
```py
    def ApplyEntity(self, request: RegistryServer_pb2.ApplyEntityRequest, context):
        assert_permissions(
            resource=self.proxied_registry.apply_entity(
                entity=Entity.from_proto(request.entity),
                project=request.project,
                commit=request.commit,
            ),
            actions=CRUD,
        )
        return Empty()
```
to:
```py
        self.proxied_registry.apply_entity(
            entity=assert_permissions(
                resource=Entity.from_proto(request.entity),
                actions=CRUD,
            ),
            project=request.project,
            commit=request.commit,
        )
```

And the same for all the `apply_XYZ` methods. Can you pls take this? (and also verify why UTs did not fail)

sdk/python/feast/permissions/permission.py
```diff
@@ -0,0 +1,248 @@
...
+        tags: Dictionary of key-value pairs that must match the resource tags. All these tags must
+    """
+
+    _name: str
+    _types: list["FeastObject"]
+    _with_subclasses: bool
+    _name_pattern: Optional[str]
+    _actions: list[AuthzedAction]
+    _policy: Policy
+    _tags: Dict[str, str]
```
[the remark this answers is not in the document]
tmihalac: OK will add required_tags as well

docs/getting-started/concepts/permission.md
franciscojavierarceo: I think we'll need to revisit that naming convention to make it more obvious.

franciscojavierarceo: @tokoko what do you think here?

docs/getting-started/concepts/permission.md
franciscojavierarceo: Yeah I think so!

docs/getting-started/concepts/permission.md
tokoko: This might have been my suggestion :smile: `read` refers to R in CRUD and `query` is the next best name for data retrieval. Maybe we can change `read` to `describe` and we will end up with `describe` and `query`. Should be much less ambiguous.

protos/feast/core/Permission.proto
dmartinol: see [latest commit](https://github.com/RHEcosystemAppEng/feast/pull/74/commits/0033c1fe39e4a9ba26c26c7d4c64cef95d14f981)

docs/getting-started/concepts/permission.md
dmartinol: Yes, @tokoko is definitely the ghost writer behind this design 😉 
We can rename it with low effort but I would keep the utility constant CRUD as is, w/o renaming it to `CDUD`

docs/getting-started/concepts/permission.md
dmartinol: @franciscojavierarceo pls TAL [here](https://github.com/feast-dev/feast/tree/4e778a7f9c755f9b3e0dbae932f199916426a78f/examples/rbac-local), an example driven by notebooks to build a fully operational, distributed system with OIDC authorization.

setup.py
tokoko: The fact that python tests were passing even w/o making them required was because we're doing python tests wrong #4185 :smile: We're installing all dependencies (required + extras) in the env and consequently we have no way of checking whether all required dependencies are actually listed as such or not.

Another way to take care of the failures would be to refactor the code how that kubernetes dependency is no longer top-level, i.e. it will only be imported when `feature_store.yaml` specifies k8s as auth type.

protos/feast/core/Permission.proto
tokoko: great, just a single nit. Probably best to remove this sentence altogether `Please note that all sub-classes of FeatureView are also included.` since `FeatureView` will no longer have any subclasses after the refactor.

docs/getting-started/concepts/permission.md
tokoko: yeah, I'm fine with keeping `CRUD` intact. We can make the change to `describe` if @franciscojavierarceo is also in agreement.

docs/reference/feast-cli-commands.md
```diff
@@ -155,6 +156,144 @@ Load data from feature views into the online store, beginning from either the pr
...
+
+### Describe a permission
+Describes the provided permission
+
+```text
+feast permissions describe permission-name
+name: permission-name
+types:
+- FEATURE_VIEW
+withSubclasses: true
```
tokoko, line 212: @dmartinol missed here

docs/getting-started/concepts/permission.md
```diff
@@ -0,0 +1,152 @@
...
+    name="reader",
+    types=[FeatureView],
+    name_pattern=".*risky.*",
+    policy=RoleBasedPolicy(roles=["trusted"]),
+    actions=[AuthzedAction.QUERY_OFFLINE],
+)
+```
+
+## Authorizing Feast clients
+If you would like to leverage the permissions' functionality then `auth` config block should be added to feature_store.yaml. Currently, feast supports oidc and kubernetes rbac authentication/authorization.
```
tokoko, line 105: Maybe also mention that only one mechanism can be configured on the server?

sdk/python/feast/registry_server.py
redhatHameed: updated, accordingly. Thanks

docs/reference/registry/registry-permissions.md
tokoko: for `Proto`, which is basically "read everything" endpoint, we should peek inside filter out only the objects that the user has `read` permission on. It will be a bit awkward and slow unfortunately as `proto` method returns protobuf object rather than feast python objects, each of which I'm assuming we will have to first convert to python before applying permissions and returning "reduced" proto object back to the user. @dmartinol

btw, this endpoint is used for client-side caching so this will be by far the most commonly called endpoint on the server.

dmartinol: > for Proto, which is basically "read everything" endpoint, we should peek inside and filter out only the objects that the user has read permission on.

Agree, but only when it comes from a direct client request to the `Proto` endpoint.
When the request comes from another service it should be unfiltered and return the full content instead.

BTW: how to allow this service-2-service communication is still to be finalized. We have a solution based on a sort of `super-role` that we automatically create (k8s demo), but it's not optimized as it runs unnecessary validations at every requests. Service-2-service requests should be easily identified and promptly permitted without additional overhead.

tokoko: > BTW: how to allow this service-2-service communication is still to be finalized. We have a solution based on a sort of `super-role` that we automatically create (k8s demo), but it's not optimized as it runs unnecessary validations at every requests. Service-2-service requests should be easily identified and promptly permitted without additional overhead.

I was also thinking of some sort of built-in magic `admin` role. Instead of adding a `Permission` object for it, we can simply check for `admin` role right at the start of `assert_*` methods and sidestep the whole thing if the user is admin w/o any validation.

tokoko: How about we delay Commit, Refresh, Proto handling for a separate PR? This one is large enough as is.

dmartinol: > I was also thinking of some sort of built-in magic `admin` role. Instead of adding a `Permission` object for it, we can simply check for `admin` role right at the start of `assert_*` methods and sidestep the whole thing if the user is admin w/o any validation.

This is one of the options under investigation. We want to mark the FS as a "server" when it starts one of the servers, so that each remote proxy, instead of generating the token from the auth provider, it sends something special (*) that the receiver can identify as a server request and skip the authorization enforcements steps.  
(*) under discussion: header, hashed code ...

docs/SUMMARY.md
dmartinol: @tokoko yes, that was the original idea, even if I'm not sure anyone would ever want to do that. Anyway, it's open-source for a good reason 😉

franciscojavierarceo: Makes sense.

docs/getting-started/concepts/permission.md
franciscojavierarceo: Yeah I do think `describe` is more intuitive. I definitely appreciate the `query` language but I think following `CRUD` is a good enough convention that it'd be better to follow that.

docs/reference/registry/registry-permissions.md
tokoko: I'll resolve this and create a separate ticket

HaoXuAI:
LGTM!
Are we ready to merge? @tokoko @franciscojavierarceo

tokoko:
Still a few comments left that need to be resolved first

franciscojavierarceo (changes requested):
Gone to block this to hold as comments are being addressed. This is great!

examples/rbac/client/admin_user_resources.yaml
dmartinol: @redhatHameed remember to apply this change in the unext PR for RBAC examples

docs/getting-started/concepts/permission.md
dmartinol: Addressing this changein next commit

examples/rbac/client/admin_user_resources.yaml
redhatHameed: Thanks @dmartinol will open the PR separately for examples so this can resolve for now

dmartinol:
@franciscojavierarceo @tokoko 
Next PRs will handle:
* Working examples of RBAC configuration with OIDC and Kubernetes
* Solution for intra-service communication

Apart from that, all comments seem addressed.

sdk/python/feast/permissions/permission.py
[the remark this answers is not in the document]
redhatHameed: @dmartinol @tmihalac is this resolved ?

docs/reference/feast-cli-commands.md
redhatHameed: @dmartinol can you fix that

redhatHameed: nvm fixed that can be resolve now

docs/getting-started/concepts/permission.md
redhatHameed: updated. Please take another look Thanks

tokoko:
@dmartinol required dependencies in `setup.py` is the only one left, I think. But I'm fine with fixing that in a separate PR as well. I'll go through the whole thing one last time and then we should be able to merge.

redhatHameed:
> @dmartinol required dependencies in `setup.py` is the only one left, I think. But I'm fine with fixing that in a separate PR as well. I'll go through the whole thing one last time and then we should be able to merge.

@tokoko Thanks, Are you referring to this [thread / comments](https://github.com/feast-dev/feast/pull/4380#discussion_r1711710084)?

tokoko:
@redhatHameed yup

docs/getting-started/concepts/permission.md
```diff
@@ -0,0 +1,111 @@
...
+  - We assume that the resource has a `name` attribute and optional dictionary of associated key-value `tags`.
+- An `action` is a logical operation executed on the secured resource, like:
+  - `create`: Create an instance.
+  - `describe`: Access the instance state.
+  - `update`: Update the instance state.
+  - `delete`: Delete an instance.
+  - `query`:  Query both online and offline stores.
+  - `query_online`:  Query the online store.
+  - `query_offline`:  Query the offline store.
+  - `write`:  Query on any store.
```
tokoko, line 29: typo: write

docs/getting-started/concepts/permission.md
```diff
@@ -0,0 +1,111 @@
...
+To simplify configuration, several constants are defined to streamline the permissions setup:
+- In module `feast.feast_object`:
+  - `ALL_RESOURCE_TYPES` is the list of all the `FeastObject` types.
+  - `ALL_FEATURE_VIEW_TYPES` is the list of all the feature view types, including those not inheriting from `FeatureView` type like 
+  `OnDemandFeatureView`.
+- In module `feast.permissions.action`:
+  - `ALL_ACTIONS` is the list of all managed actions.
+  - `QUERY` includes all the query actions for online and offline store.
+  - `WRITE` includes all the write actions for online and offline store.
+  - `CRUD` includes all the state management actions to create, read, update or delete a Feast resource.
```
tokoko, line 53: `describe` instead of `read`

sdk/python/feast/feature_server.py
```diff
@@ -265,6 +345,18 @@ def start_server(
         )
         monitoring_thread.start()
 
+    print("start_server called ")
```
tokoko, line 348: Maybe change these print statements to debug logs?

redhatHameed: done

setup.py
redhatHameed: This will be addressed in the next/separate PR
cc @tmihalac @lokeshrangineni

tokoko (approved):
LGTM, great work

docs/getting-started/components/overview.md
```diff
@@ -28,3 +28,4 @@ A complete Feast deployment contains the following components:
...
 * **Online Store:** The online store is a database that stores only the latest feature values for each entity. The online store is either populated through materialization jobs or through [stream ingestion](../../reference/data-sources/push.md).
 * **Offline Store:** The offline store persists batch data that has been ingested into Feast. This data is used for producing training datasets. For feature retrieval and materialization, Feast does not manage the offline store directly, but runs queries against it. However, offline stores can be configured to support writes if Feast configures logging functionality of served features.
+* **Authorization manager**: The authorization manager detects authentication tokens from client requests to Feast servers and uses this information to enforce permission policies on the requested services.
```
franciscojavierarceo, line 31: Can we add a section to the `architecture/rbac.md` to give an overview of what's intended?

docs/getting-started/concepts/permission.md
```diff
@@ -0,0 +1,111 @@
...
+
+The permission model is based on the following components:
+- A `resource` is a Feast object that we want to secure against unauthorized access.
+  - We assume that the resource has a `name` attribute and optional dictionary of associated key-value `tags`.
+- An `action` is a logical operation executed on the secured resource, like:
+  - `create`: Create an instance.
+  - `describe`: Access the instance state.
+  - `update`: Update the instance state.
+  - `delete`: Delete an instance.
+  - `query`:  Query both online and offline stores.
```
franciscojavierarceo, line 26: shouldn't this be `read`?

protos/feast/core/Permission.proto
```diff
@@ -0,0 +1,69 @@
...
+  PermissionMeta meta = 2;
+}
+
+message PermissionSpec {
+  enum AuthzedAction {
+    CREATE = 0;
+    DESCRIBE = 1;
+    UPDATE = 2;
+    DELETE = 3;
+    QUERY_ONLINE = 4;
```
franciscojavierarceo, line 25: Shouldn't this be `READ_ONLINE`?

protos/feast/core/Permission.proto
```diff
@@ -0,0 +1,69 @@
...
+    QUERY_OFFLINE = 5;
```
franciscojavierarceo, line 26: Shouldn't this be `READ_OFFLINE`?

sdk/python/feast/feature_server.py
```diff
@@ -93,23 +101,48 @@ async def lifespan(app: FastAPI):
...
+                        store.registry,
+                        store.project,
+                        features,
+                        allow_cache=True,
+                        hide_dummy_entity=False,
+                    )
+                )
+                for feature_view in all_feature_views:
+                    assert_permissions(
+                        resource=feature_view, actions=[AuthzedAction.QUERY_ONLINE]
```
franciscojavierarceo, line 136: `READ_ONLINE`?

sdk/python/feast/feature_server.py
```diff
@@ -93,23 +101,48 @@ async def lifespan(app: FastAPI):
...
+                    )
+                for od_feature_view in all_on_demand_feature_views:
+                    assert_permissions(
+                        resource=od_feature_view, actions=[AuthzedAction.QUERY_ONLINE]
```
franciscojavierarceo, line 140: `READ_ONLINE`?

sdk/python/feast/permissions/action.py
```diff
@@ -0,0 +1,40 @@
...
+    QUERY_OFFLINE = "query_offline"  # Query the offline store only
+    WRITE_ONLINE = "write_online"  # Write to the online store only
+    WRITE_OFFLINE = "write_offline"  # Write to the offline store only
+
+
+#  Alias for all available actions
+ALL_ACTIONS = [a for a in AuthzedAction.__members__.values()]
+
+#  Alias for all query actions
+QUERY = [
```
franciscojavierarceo, line 23: `READ`

sdk/python/feast/permissions/action.py
```diff
@@ -0,0 +1,40 @@
...
+    AuthzedAction.QUERY_OFFLINE,
```
franciscojavierarceo, line 24: AuthzedAction.READ_OFFLINE,

sdk/python/feast/permissions/action.py
```diff
@@ -0,0 +1,40 @@
...
+    AuthzedAction.QUERY_ONLINE,
```
franciscojavierarceo, line 25: AuthzedAction.READ_ONLINE,

sdk/python/feast/permissions/decision.py
```diff
@@ -0,0 +1,114 @@
...
+        """
+        logger.info(
+            f"Decided grant is {self.grant_decision}, explanations={self.explanations}"
+        )
+        return bool(self.grant_decision), self.explanations
+
+    def add_grant(self, grant: bool, explanation: str):
+        """
+        Add a single vote to the decision computation, with a possible denial reason.
+        If the evalluation process already ended, additional votes are discarded.
```
franciscojavierarceo, line 92:
```suggestion
        If the evaluation process already ended, additional votes are discarded.
```

sdk/python/feast/repo_config.py
```diff
@@ -300,6 +332,30 @@ def batch_engine(self):
...
+    @model_validator(mode="before")
+    def _validate_auth_config(cls, values: Any) -> Any:
+        from feast.permissions.auth_model import AuthConfig
+
+        if "auth" in values:
+            allowed_auth_types = AUTH_CONFIGS_CLASS_FOR_TYPE.keys()
+            if isinstance(values["auth"], Dict):
+                if values["auth"].get("type") is None:
+                    raise ValueError(
+                        f"auth configuration is not having authentication type. Possible values={allowed_auth_types}"
```
franciscojavierarceo, line 344:
```suggestion
                        f"auth configuration is missing authentication type. Possible values={allowed_auth_types}"
```

sdk/python/feast/repo_config.py
```diff
@@ -300,6 +332,30 @@ def batch_engine(self):
...
+                    )
+                elif values["auth"]["type"] not in allowed_auth_types:
+                    raise ValueError(
+                        f'auth configuration is having invalid authentication type={values["auth"]["type"]}. Possible '
```
franciscojavierarceo, line 348:
```suggestion
                        f'auth configuration has invalid authentication type={values["auth"]["type"]}. Possible '
```

sdk/python/feast/repo_config.py
```diff
@@ -300,6 +332,30 @@ def batch_engine(self):
...
+                        f'values={allowed_auth_types}'
+                    )
+            elif isinstance(values["auth"], AuthConfig):
+                if values["auth"].type not in allowed_auth_types:
+                    raise ValueError(
+                        f'auth configuration is having invalid authentication type={values["auth"].type}. Possible '
```
franciscojavierarceo, line 354:
```suggestion
                        f'auth configuration has invalid authentication type={values["auth"].type}. Possible '
```

sdk/python/tests/integration/permissions/test_auth_permission.py
```diff
@@ -0,0 +1,9 @@
+# @pytest.mark.integration
```
franciscojavierarceo, line 1: Is this here on purpose?

sdk/python/tests/integration/registration/test_universal_registry.py
```diff
@@ -1343,3 +1350,138 @@ def validate_project_uuid(project_uuid, test_registry):
     assert len(test_registry.cached_registry_proto.project_metadata) == 1
     project_metadata = test_registry.cached_registry_proto.project_metadata[0]
     assert project_metadata.project_uuid == project_uuid
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize("test_registry", all_fixtures)
+def test_apply_permission_success(test_registry):
```
franciscojavierarceo, line 1357: I don't see a test for READ_ONLINE or READ_OFFLINE permissions.

sdk/python/tests/unit/permissions/auth/server/test_utils.py
```diff
@@ -0,0 +1,61 @@
+import assertpy
+import pytest
+
+from feast import Entity, FeatureView, OnDemandFeatureView, StreamFeatureView
+from feast.permissions.action import AuthzedAction
+from feast.permissions.permission import Permission
+from feast.permissions.policy import RoleBasedPolicy
+from feast.permissions.server.utils import AuthManagerType, str_to_auth_manager_type
+
```
franciscojavierarceo, line 9: Only tests describe?

franciscojavierarceo (changes requested):

docs/getting-started/concepts/permission.md
tokoko: looks like we misunderstood each other. because `read` can be interpreted as both "access state" and "read data", my suggestion was not to use `read` anymore at all. Instead we will have:

- `describe` - this means that a user is allowed to read info from the registry about the feature view.
- `query` - this means that a user is allowed to read dataset from online/offline stores.

cc @dmartinol

franciscojavierarceo: got it. I'm okay with that, it just doesn't fit the CRUD convention, right?

sdk/python/tests/integration/permissions/test_auth_permission.py
redhatHameed: @lokeshrangineni what you think ?

docs/getting-started/concepts/permission.md
tokoko: yeah, it doesn't, but results in less ambiguous naming, which was the goal. plus, describe is also widely used in such cases (in aws iam, at least), so I think that should be fine as well.

sdk/python/tests/integration/registration/test_universal_registry.py
tmihalac: @franciscojavierarceo There are tests for them here sdk/python/tests/integration/online_store/test_remote_online_store.py

sdk/python/tests/unit/permissions/auth/server/test_utils.py
lokeshrangineni: yes, this class tests the functionality written in [utils.py](https://github.com/RHEcosystemAppEng/feast/blob/feast-rbac/sdk/python/feast/permissions/server/utils.py) file. and also some common code related to auth tests.

sdk/python/feast/repo_config.py
redhatHameed: updated.

sdk/python/feast/repo_config.py
redhatHameed: updated.

sdk/python/feast/repo_config.py
redhatHameed: updated.

sdk/python/feast/permissions/decision.py
redhatHameed: updated.

docs/getting-started/components/overview.md
redhatHameed: Added doc under `architecture/rbac.md` Please take a look

redhatHameed:
@franciscojavierarceo The comments/feedback have been addressed. Please let us know if any further changes are required at your convenience.

docs/SUMMARY.md
franciscojavierarceo, line 19: Can you add an RBAC link under Architecture?

docs/getting-started/architecture/overview.md
```diff
@@ -17,3 +17,7 @@ typically your Offline Store). We are exploring adding a default streaming engin
   write patterns](write-patterns.md) to your application
 
 * We recommend [using Python](language.md) for your Feature Store microservice. As mentioned in the document, precomputing features is the recommended optimal path to ensure low latency performance. Reducing feature serving to a lightweight database lookup is the ideal pattern, which means the marginal overhead of Python should be tolerable. Because of this we believe the pros of Python outweigh the costs, as reimplementing feature logic is undesirable. Java and Go Clients are also available for online feature retrieval.
+
+* Role-Based Access Control (RBAC)[rbac.md](rbac.md) is a security mechanism that restricts access to resources based on the roles of individual users within an organization. In the context of the Feast Feature Store, RBAC ensures that only authorized users or groups can access or modify specific resources, thereby maintaining data security and operational integrity.
```
franciscojavierarceo, line 21:
```suggestion
* [Role-Based Access Control (RBAC)](rbac.md) is a security mechanism that restricts access to resources based on the roles of individual users within an organization. In the context of the Feast Feature Store, RBAC ensures that only authorized users or groups can access or modify specific resources, thereby maintaining data security and operational integrity.
```

docs/getting-started/architecture/rbac.md
```diff
@@ -0,0 +1,133 @@
+# Role-Based Access Control (RBAC) in Feast Feature Store
```
franciscojavierarceo, line 1:
```suggestion
# Role-Based Access Control (RBAC) in Feast
```

docs/getting-started/architecture/rbac.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Introduction
+
+Role-Based Access Control (RBAC) is a security mechanism that restricts access to resources based on the roles of individual users within an organization. In the context of the Feast Feature Store, RBAC ensures that only authorized users or groups can access or modify specific resources, thereby maintaining data security and operational integrity.
```
franciscojavierarceo, line 5:
```suggestion
Role-Based Access Control (RBAC) is a security mechanism that restricts access to resources based on the roles of individual users within an organization. In the context of the Feast Feature Store, RBAC ensures that only authorized users or groups can access or modify specific resources, thereby maintaining data security and operational integrity.
```
```suggestion
Role-Based Access Control (RBAC) is a security mechanism that restricts access to resources based on the roles of individual users within an organization. In the context of the Feast, RBAC ensures that only authorized users or groups can access or modify specific resources, thereby maintaining data security and operational integrity.
```

docs/getting-started/architecture/rbac.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Functional Requirements
+
+The RBAC implementation in Feast Feature Store is designed to:
```
franciscojavierarceo, line 9:
```suggestion
The RBAC implementation in Feast is designed to:
```

docs/getting-started/architecture/rbac.md
```diff
@@ -0,0 +1,133 @@
...
+
+- **Assign Permissions**: Allow administrators to assign permissions for various operations and resources to users or groups based on their roles.
+- **Seamless Integration**: Integrate smoothly with existing business code without requiring significant modifications.
+- **Backward Compatibility**: Maintain support for non-authorized models as the default to ensure backward compatibility.
+
+## Business Goals
+
+The primary business goals of implementing RBAC in the Feast Feature Store are:
```
franciscojavierarceo, line 17:
```suggestion
The primary business goals of implementing RBAC in the Feast are:
```

docs/getting-started/architecture/rbac.md
```diff
@@ -0,0 +1,133 @@
...
+
+1. **Feature Sharing**: Enable multiple teams to share the feature store while ensuring controlled access to data partitions. This allows for collaborative work without compromising data security.
```
franciscojavierarceo, line 19:
```suggestion
1. **Feature Sharing**: Enable multiple teams to share the feature store while ensuring controlled access. This allows for collaborative work without compromising data security.
```

docs/getting-started/architecture/rbac.md
```diff
@@ -0,0 +1,133 @@
...
+
+2. **Access Control Management**: Prevent unauthorized access to team-specific resources and spaces, governing the operations that each user or group can perform.
+
+## Reference Architecture
+
+The Feast Feature Store operates as a collection of connected services, each enforcing authorization permissions. The architecture is designed as a distributed microservices system with the following key components:
```
franciscojavierarceo, line 25:
```suggestion
Feast operates as a collection of connected services, each enforcing authorization permissions. The architecture is designed as a distributed microservices system with the following key components:
```

docs/getting-started/architecture/rbac.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Permission Model
+
+The RBAC system in Feast uses a permission model that defines the following concepts:
+
+- **Resource**: An object within Feast that needs to be secured against unauthorized access.
+- **Action**: A logical operation performed on a resource, such as Create, Describe, Update, Delete, query, or write operations.
+- **Policy**: A set of rules that enforce authorization decisions on resources. The default implementation uses role-based policies.
+
+### Configuring Permissions
```
franciscojavierarceo, line 40: Please remove this section as the architecture is meant to give a motivation and not the implementation details.

docs/getting-started/architecture/rbac.md
```diff
@@ -0,0 +1,133 @@
...
+Permission(
+    name="ds-writer",
+    types=[DataSource],
+    required_tags={"risk_level": "hi"},
+    actions=[AuthzedAction.WRITE_ONLINE, AuthzedAction.WRITE_OFFLINE],
+    policy=RoleBasedPolicy(roles=["admin", "data_team"]),
+)
+```
+
+### Enforcing Permission Policy
```
franciscojavierarceo, line 59: Please remove this section as well.

docs/getting-started/architecture/rbac.md
```diff
@@ -0,0 +1,133 @@
...
+```python
+assert_permissions(
+    resource=feature_service,
+    actions=[AuthzedAction.QUERY_ONLINE]
+)
+```
+
+## Use Cases
+
+### Tag-Based Permission
```
franciscojavierarceo, line 73: Please remove this section as well.

docs/getting-started/architecture/rbac.md
```diff
@@ -0,0 +1,133 @@
...
+Permission(
+    name="team1-fvs",
+    types=[FeatureView],
+    required_tags={"team": "team1"},
+    actions=[CRUD, WRITE, QUERY],
+    policy=RoleBasedPolicy(roles=["role1"]),
+)
+```
+
+### Name-Based Permission
```
franciscojavierarceo, line 85: Please remove this section as well.

docs/getting-started/architecture/rbac.md
```diff
@@ -0,0 +1,133 @@
...
+## Authorization Architecture
+
+The authorization architecture in Feast is built with the following components:
+
+1. **Token Extractor**: Extracts the authorization token from the request header.
+2. **Token Parser**: Parses the token to retrieve user details.
+3. **Policy Enforcer**: Validates the secured endpoint against the retrieved user details.
+4. **Token Injector**: Adds the authorization token to each secured request header.
+
+### OIDC Authorization
```
franciscojavierarceo, line 106: Please remove this section as well.

docs/getting-started/architecture/rbac.md
```diff
@@ -0,0 +1,133 @@
...
+  type: oidc
+  client_id: _CLIENT_ID_
+  client_secret: _CLIENT_SECRET_
+  username: _USERNAME_
+  password: _PASSWORD_
+  realm: _REALM_
+  auth_server_url: _OIDC_SERVER_URL_
+```
+
+### Kubernetes Authorization
```
franciscojavierarceo, line 122: Please remove this section as well.

franciscojavierarceo (changes requested):
Last changes. The architecture section should be pretty high level. I believe some of what you added to a PR is included in other documentation. If it's not, can you add it to the components section? I like the documentation, I just think it's better served under the components section. 

After that I think we're good.

docs/getting-started/architecture/rbac.md
```diff
@@ -0,0 +1,133 @@
...
+
+1. **Feature Sharing**: Enable multiple teams to share the feature store while ensuring controlled access to data partitions. This allows for collaborative work without compromising data security.
+
+2. **Access Control Management**: Prevent unauthorized access to team-specific resources and spaces, governing the operations that each user or group can perform.
+
+## Reference Architecture
+
+The Feast Feature Store operates as a collection of connected services, each enforcing authorization permissions. The architecture is designed as a distributed microservices system with the following key components:
+
+- **Service Endpoints**: These enforce authorization permissions, ensuring that only authorized requests are processed.
```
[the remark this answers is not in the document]
redhatHameed: that's correct, simplified it more .

note `Service-to-Service Communication` this feature not yet part of this PR will be implementation next PRs.

redhatHameed:
> Last changes. The architecture section should be pretty high level. I believe some of what you added to a PR is included in other documentation. If it's not, can you add it to the components section? I like the documentation, I just think it's better served under the components section.
> 
> After that I think we're good.

Agree was repetition these are already aded into components section, hence remove from here.

README.md
```diff
@@ -16,9 +16,6 @@
 [![License](https://img.shields.io/badge/License-Apache%202.0-blue)](https://github.com/feast-dev/feast/blob/master/LICENSE)
 [![GitHub Release](https://img.shields.io/github/v/release/feast-dev/feast.svg?style=flat&sort=semver&color=blue)](https://github.com/feast-dev/feast/releases)
 
-## Join us on Slack!
```
franciscojavierarceo, line 19: Can you revert this change?

docs/getting-started/components/overview.md
franciscojavierarceo: Thank you!

README.md
redhatHameed: Done.

docs/getting-started/concepts/permission.md
```diff
@@ -0,0 +1,112 @@
...
+teams to meet organizational requirements for service and data sharing, and protection of sensitive information.
+
+The `feast` CLI includes a new `permissions` command to list the registered permissions, with options to identify the matching resources for each configured permission and the existing resources that are not covered by any permission.
+
+{% hint style="info" %}
+**Note**: Feast resources that do not match any of the configured permissions are not secured by any authorization policy, meaning any user can execute any action on such resources.
+{% endhint %}
+
+## Definition examples
+This permission definition grants access to the resource state and the ability to query all the stores for any feature view or
```
franciscojavierarceo, line 65:
```suggestion
This permission definition grants access to the resource state and the ability to read all of the stores for any feature view or
```

docs/reference/feature-servers/offline-feature-server.md
```diff
@@ -33,3 +33,20 @@ Please see the detail how to configure offline store client [remote-offline-stor
...
+
+## API Endpoints and Permissions
+
+| Endpoint                              | Resource Type    | Permission    | Description                                        |
+| ------------------------------------- |------------------|---------------|----------------------------------------------------|
+| offline_write_batch                   | FeatureView      | Write Offline | Write a batch of data to the offline store         |
+| write_logged_features                 | FeatureService   | Write Offline | Write logged features to the offline store         |
+| persist                               | DataSource       | Write Offline | Persist the result of a query in the offline store |
```
franciscojavierarceo, line 44:
```suggestion
| persist                               | DataSource       | Write Offline | Persist the result of a read in the offline store |
```

docs/reference/feature-servers/offline-feature-server.md
```diff
@@ -33,3 +33,20 @@ Please see the detail how to configure offline store client [remote-offline-stor
...
+| get_historical_features               | FeatureView      | Read Offline  | Retrieve historical features                       |
+| pull_all_from_table_or_query          | DataSource       | Read Offline  | Pull all data from a table or read it              |
+| pull_latest_from_table_or_query       | DataSource       | Read Offline  | Pull the latest data from a table or read it       |
```
franciscojavierarceo, line 47:
```suggestion
| pull_latest_from_table_or_read       | DataSource       | Read Offline  | Pull the latest data from a table or read it       |
```

redhatHameed: the is name of [function/endpoint](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/offline_stores/offline_store.py#L218) in offline_store so we need to keep the same.

docs/reference/feature-servers/offline-feature-server.md
redhatHameed: done

docs/getting-started/concepts/permission.md
```diff
@@ -0,0 +1,112 @@
...
+feature service to all users with role `super-reader`:
```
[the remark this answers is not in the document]
redhatHameed: done

docs/getting-started/concepts/permission.md
redhatHameed: done

docs/reference/feature-servers/offline-feature-server.md
franciscojavierarceo: Got it, thank you!

franciscojavierarceo (approved):
Amazing work here, thanks for taking in all of the changes!

/lgtm

redhatHameed:
@tokoko @franciscojavierarceo for taking time to review and approve this

Next PRs will handle:

1. Working examples of RBAC configuration with OIDC and Kubernetes.
2. Solution for intra-service communication
3. Fix the dependencies to make it not required in setup.py 
4. Refactoring the OIDC authentication to use discovery URL

 Also if there is no any further concern **can you merge this** Thanks

cc @jeremyary @dandawg

PR #4384, opened 2024-08-05
feat: Allow feast snowflake to read in byte string for private-key authentication

arturkolakowski:

Allows the user to pass private_key as a byte string for Snowflake key-pair authentication rather than only limiting the ability to refer to the file path where the private key is stored. 
Gives the ability to pass in `private_key` from a Secret Manager service (i.e: Google Cloud Secret Manager) 

# Fixes

### DIFF
--- a/sdk/python/feast/infra/materialization/snowflake_engine.py
+++ b/sdk/python/feast/infra/materialization/snowflake_engine.py
@@ -67,7 +67,7 @@ class SnowflakeMaterializationEngineConfig(FeastConfigBaseModel):
     authenticator: Optional[str] = None
     """ Snowflake authenticator name """
 
-    private_key: Optional[str | bytes] = None
+    private_key: Optional[Union[str,bytes]] = None
     """ Snowflake private key stored as bytes or file path"""
 
     private_key_passphrase: Optional[str] = None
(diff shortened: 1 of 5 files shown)


tokoko:
just curious, are you able to set bytes value from yaml as well or will this only work in python?

arturkolakowski:
> just curious, are you able to set bytes value from yaml as well or will this only work in python?

yes it will work, for example, `feature_store.yaml` can include `private_key` of format 
```
'-----BEGIN ENCRYPTED PRIVATE KEY-----
.
.
-----END ENCRYPTED PRIVATE KEY-----'
```

tokoko:
Wouldn't that be interpreted as a string and therefore as a file path?

arturkolakowski:
As I am looking through the code base and considering this example: `SnowflakeOfflineStore`
`SnowflakeOfflineStore` establishes a snowflake connection with GetSnowflakeConnection (`/feast/infra/utils/snowflake/snowflake_utils.py`)
```
with GetSnowflakeConnection(config.offline_store) as conn:
            snowflake_conn = conn
```
when `parse_private_key_path` gets called `private_key` is already of type `bytes` so this way only works in python, my apologies.

tokoko:
@arturkolakowski how about we add another field in config instead (something like `private_key_content`) which would simply take precedence?

arturkolakowski:
@tokoko Made changes based on the request

tokoko (approved):
lgtm, thanks

PR #4385, opened 2024-08-06
docs: Adding some content about the push vs pull model

franciscojavierarceo:

This PR adds documentation about why Feast uses a Push Model.

In another PR, I'll outline different write patterns when pushing / writing features to Feast. 

# Fixes


shuchu (approved):
LGTM

PR #4391, opened 2024-08-07
fix: Using get_type_hints instead of inspect signature for udf return annotation 

EXPEbdodla:

# Misc
signature function to get the return annotation, can sometimes return it as a string if the annotation itself was defined using forward references. We imported `from __future__ import annotations` in our definitions file. That will return OnDemandFeatureView UDF return type as **'pd.DataFrame'** instead of **pd.DataFrame**. 

Added test cases to prove that.  

get_type_hints will resolve the forward references by giving you the actual type objects instead of strings.

### DIFF
--- a/sdk/python/feast/on_demand_feature_view.py
+++ b/sdk/python/feast/on_demand_feature_view.py
@@ -3,7 +3,7 @@ import functools
 import inspect
 import warnings
 from types import FunctionType
-from typing import Any, Optional, Union
+from typing import Any, Optional, Union, get_type_hints
 
 import dill
 import pandas as pd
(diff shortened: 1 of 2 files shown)


sdk/python/feast/on_demand_feature_view.py
```diff
@@ -631,7 +631,7 @@ def mainify(obj) -> None:
             obj.__module__ = "__main__"
 
     def decorator(user_function):
-        return_annotation = inspect.signature(user_function).return_annotation
+        return_annotation = get_type_hints(user_function).get("return", inspect._empty)
```
tokoko, line 634: Do we need to default to `inspect._empty` anymore after the change? Maybe just default to `None` and change if conditions accordingly?

EXPEbdodla: None conflicts with return type None. So using inspect._empty when the return type is missing.

tokoko (approved):

PR #4394, opened 2024-08-08
fix: Escape special characters in the Postgres password

### LINKED ISSUE #4393: Connection to Postgres fails when password contains special characters
With the release of [this PR](https://github.com/feast-dev/feast/pull/4303/files#diff-2ac41dbcd2a6b435fdc89b6d303480ed6cc1ae105dacecb7b6ba40b8d163da16), [this method](https://github.com/feast-dev/feast/pull/4303/files#diff-2ac41dbcd2a6b435fdc89b6d303480ed6cc1ae105dacecb7b6ba40b8d163da16R35-R43) was added to construct the connection string for the connection to Postgres with psycopg3. 

However, when a password contains special characters, they are not properly escaped, which leads to connection errors.

## Expected Behavior 

The special characters are escaped properly.

## Current Behavior

The special characters are not escaped properly, leading to connection errors.

## Steps to reproduce

In the [setup file](https://github.com/feast-dev/feast/blob/master/sdk/python/tests/integration/feature_repos/universal/online_store/postgres.py#L1) for the postgres integration tests, update all occurrences of the password `test` to something like `test%` or `test!#$%&()*+,-./:;<=>?@[\]^_`{|}~±`. 

Then run all postgres online store integration tests by running:
```bash
make test-python-universal-postgres-online
```

### Specifications

Version: 0.36.0
Platform: MacOS - M1
Subsystem: Sonoma 14.1.1

## Possible Solution
The potential solution is to use update [these lines of code](https://github.com/feast-dev/feast/pull/4303/files#diff-2ac41dbcd2a6b435fdc89b6d303480ed6cc1ae105dacecb7b6ba40b8d163da16R35-R43) to the following:

```python
from psycopg.conninfo import make_conninfo

def _get_conninfo(config: PostgreSQLConfig) -> str:
    """Get the `conninfo` argument required for connection objects."""
    psycopg_config = {
        "user": config.user,
        "password": config.password,
        "host": config.host,
        "port": int(config.port),
        "dbname": config.database,
    }
    return make_conninfo(**psycopg_config)
```

Instead of using a connection string, this passes the config in key-value pairs. When passing it as key-value pairs the special characters are escaped properly.

More info [here](https://www.psycopg.org/psycopg3/docs/api/connections.html).

job-almekinders:
Update the helper function that creates the connection string to use key-value pairs. This makes sure that the special characters in the password are escaped.

Fixes #4393

# Misc
I updated the setup for the postgres online store to use a password with special characters to verify this keeps working in the future.

### DIFF
--- a/sdk/python/feast/infra/utils/postgres/connection_utils.py
+++ b/sdk/python/feast/infra/utils/postgres/connection_utils.py
@@ -5,6 +5,7 @@ import pandas as pd
 import psycopg
 import pyarrow as pa
 from psycopg import AsyncConnection, Connection
+from psycopg.conninfo import make_conninfo
 from psycopg_pool import AsyncConnectionPool, ConnectionPool
 
 from feast.infra.utils.postgres.postgres_config import PostgreSQLConfig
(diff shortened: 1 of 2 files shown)


franciscojavierarceo (approved):
thank you for this!

franciscojavierarceo:
@job-almekinders did we also include this in the master branch? I just wanted to make sure

job-almekinders:
@franciscojavierarceo yes! I see it over [here](https://github.com/feast-dev/feast/commit/419ca5e9523ff38f27141b79ae12ebb0646c6617).

PR #4396, opened 2024-08-09
fix: Escape special characters in the Postgres password (#4394)

job-almekinders:
This PR cherry-picks a fix commit to properly parse special characters in Postgres passwords.

More information can be found in the [original PR](https://github.com/feast-dev/feast/pull/4394).


job-almekinders:
I don't understand why the java integration test is failing; is this an unrelated issue perhaps?

tokoko:
yeah, it's unrelated #4381. we can merge regardless

PR #4400, opened 2024-08-10
Write patterns

franciscojavierarceo:

# Misc


docs/getting-started/architecture-and-components/write-patterns.md
```diff
@@ -0,0 +1,68 @@
...
+be either raw data where Feast computes and store stores the feature values or precomputed feature values.
+
+## Communication Patterns
+
+There are two ways to *_send_* data to the online store: 
+
+1. Synchronously
+   - Using an API call for a small number of entities or a single entity
+2. Asynchronously 
+   - Using an API call for a small number of entities or a single entity
```
tokoko, line 18: Can you clarify what you're referring to here? Are there multiple ways to push features to online using feature server API?

docs/getting-started/architecture-and-components/write-patterns.md
```diff
@@ -0,0 +1,68 @@
...
+
+
+There are two ways to write *feature values* to the online store:
+
+1. Precomputing the transformations
+2. Computing the transformations On Demand
+3. Hybrid (Precomputed + On Demand)
+
+### 1. Precomputing the transformations
+Precomputed transformations can happen outside of Feast (e.g., via some batch job or streaming application) or inside of the Feast feature server when writing to the online store via the `write-to-online-store` api.
```
tokoko, line 43: How can one precompute values during `write-to-online-store`? Are you referring to stream feature views?

docs/getting-started/architecture-and-components/write-patterns.md
franciscojavierarceo: https://en.wikipedia.org/wiki/Message_passing#Synchronous_versus_asynchronous_message_passing

Do you want me to link to that source?

franciscojavierarceo: >P.S. of course a user might use client-side async handling, but I don't think that's what you're referring to here.

Actually, this is exactly what I am referring to!

Sounds like I need to make this more explicit if it wasn't obvious.

docs/getting-started/architecture-and-components/write-patterns.md
tokoko: sure, but my understanding from that ticket and the discussion on the call was that we were talking about "persisting" values during get_online_features, not during write_to_online_store.

franciscojavierarceo: >Again the write_to_online_store: bool parameter would dictate whether this ODFV would write to the online store and the force_compute: bool would dictate whether the ODFV would always recalculate the features. There's an argument to be made that we could skip the write_to_online_store in the FeatureView declaration but this metadata would be useful to have in the registry for users.

The ODFV computations were meant to happen when *writing* to the online store, as outlined in this document. 

I should have written this document first, I suppose.

Let me know if that makes more sense for you.

docs/getting-started/architecture-and-components/write-patterns.md
franciscojavierarceo: @tokoko take a look, I updated.

tokoko:
Maybe we can move these new docs to a new section called "Design" right after the Architecture one. Mixing these with very concrete feast component definitions (Online Store, Offline Store, ...) feels odd.

franciscojavierarceo:
I like that! I'll move it over to `System Design`

franciscojavierarceo:
Okay I ended up changing it to instead split `architecture-and-components` to `architecture` and `components` and moved the relevant material accordingly. I also updated some broken links I found.

HaoXuAI (approved):
nice, lgtm！

Issue #4239, opened 2024-05-30
Retire Pytz library from the code repo.


shuchu:
### Copied from SonarCloud:

```
In Python 3.9 and later, the zoneinfo module is the recommended tool for handling timezones, replacing the pytz library. This recommendation is based on several key advantages.

First, zoneinfo is part of Python’s standard library, making it readily available without needing additional installation, unlike pytz.

Second, zoneinfo integrates seamlessly with Python’s datetime module. You can directly use zoneinfo timezone objects when creating datetime objects, making it more intuitive and less error-prone than pytz, which requires a separate localize method for this purpose.

Third, zoneinfo handles historical timezone changes more accurately than pytz. When a pytz timezone object is used, it defaults to the earliest known offset, which can lead to unexpected results. zoneinfo does not have this issue.

Lastly, zoneinfo uses the system’s IANA time zone database when available, ensuring it works with the most up-to-date timezone data. In contrast, pytz includes its own copy of the IANA database, which may not be as current.

In summary, zoneinfo offers a more modern, intuitive, and reliable approach to handling timezones in Python 3.9 and later, making it the preferred choice over pytz.

```
- Version: 0.38.0
- Platform:
- Subsystem:

## Possible Solution

```
from datetime import datetime
import pytz

dt = pytz.timezone('America/New_York'').localize(datetime(2022, 1, 1))  # Noncompliant: the localize method is needed to avoid bugs (see S6887)

#====================================
from datetime import datetime
from zoneinfo import ZoneInfo

dt = datetime(2022, 1, 1, tzinfo=ZoneInfo('America/New_York'))  # OK: timezone object can be used safely through the datetime constructor
```

Reference: https://blog.ganssle.io/articles/2018/03/pytz-fastest-footgun.html

franciscojavierarceo:
Is this done @shuchu ?

PR #4406, opened 2024-08-14
fix: Retire pytz library

shuchu:
The Python library `pytz` is not recommended after Python3.9. 

This PR updated most code with "timezone.utc" instead of the **dask.py** file. 

It seems for Pandas's dataframe (after call the Dask.Dateframe.compute()) still have timezone info equals to pytz.UTC. 
It requires further work to investigation and retire the pytz in **dask.py**

Fix: #4239


tokoko (approved):
lgtm

PR #4407, opened 2024-08-14
docs: Reorganize registry docs

tokoko:

- Merges registry explanation from architecture and concepts sections into a single doc under architecture.
- Adds info about each registry implementation under `reference/registries` section similar to how we organize online and offline stores. Tutorial about "scalable registry" is moved here and renamed SQL registry.

I'll add registry server and remote registry docs in a separate PR.


docs/SUMMARY.md
```diff
@@ -15,7 +15,6 @@
   * [Feature view](getting-started/concepts/feature-view.md)
   * [Feature retrieval](getting-started/concepts/feature-retrieval.md)
   * [Point-in-time joins](getting-started/concepts/point-in-time-joins.md)
-  * [Registry](getting-started/concepts/registry.md)
```
franciscojavierarceo, line 18: I think it still makes sense to keep it here in the concepts though.

docs/getting-started/architecture-and-components/registry.md
```diff
@@ -1,15 +1,60 @@
 # Registry
 
-The Feast feature registry is a central catalog of all the feature definitions and their related metadata. It allows data scientists to search, discover, and collaborate on new features.
+Feast uses a registry to store all applied Feast objects (e.g. Feature views, entities, etc). It allows data scientists to search, discover, and collaborate on new features. The registry exposes methods to apply, list, retrieve and delete these objects, and is an abstraction with multiple implementations.
 
-Each Feast deployment has a single feature registry. Feast only supports file-based registries today, but supports four different backends.
+Feast comes with built-in file-based and sql-based registry implementations. By default, Feast uses a file-based registry, which stores the protobuf representation of the registry as a serialized file in the local file system. For more details on which registries are supported, please see [Registries](../../reference/registries/).
```
franciscojavierarceo, line 5: I'm not sure that path will work 🤔

docs/getting-started/architecture-and-components/registry.md
```diff
@@ -1,15 +1,60 @@
...
+  type: file
+```
+
+Instantiating a `FeatureStore` object can then point to this:
+
+```python
+store = FeatureStore(repo_path=".")
+```
+
+<!-- Each Feast deployment has a single feature registry. Feast only supports file-based registries today, but supports four different backends.
```
franciscojavierarceo, line 49: Why comment out?

docs/getting-started/architecture-and-components/registry.md
```diff
@@ -1,15 +1,60 @@
...
 
 * `Local`: Used as a local backend for storing the registry during development
 * `S3`: Used as a centralized backend for storing the registry on AWS
 * `GCS`: Used as a centralized backend for storing the registry on GCP
-* `[Alpha] Azure`: Used as centralized backend for storing the registry on Azure Blob storage.
+* `[Alpha] Azure`: Used as centralized backend for storing the registry on Azure Blob storage. -->
 
-The feature registry is updated during different operations when using Feast. More specifically, objects within the registry \(entities, feature views, feature services\) are updated when running `apply` from the Feast CLI, but metadata about objects can also be updated during operations like materialization.
+
+<!-- The feature registry is updated during different operations when using Feast. More specifically, objects within the registry \(entities, feature views, feature services\) are updated when running `apply` from the Feast CLI, but metadata about objects can also be updated during operations like materialization.
```
franciscojavierarceo, line 57: I believe this will comment it out.

docs/getting-started/concepts/README.md
```diff
@@ -24,10 +24,6 @@
 [point-in-time-joins.md](point-in-time-joins.md)
 {% endcontent-ref %}
 
-{% content-ref url="registry.md" %}
```
franciscojavierarceo, line 27: We should keep this here.

franciscojavierarceo (changes requested):
Some small changes. I think having the Registry in the concepts is important for new users who may not be familiar with having a metadata registry.

docs/getting-started/architecture-and-components/registry.md
tokoko: It was meant to be removed, slipped through, thanks

docs/getting-started/architecture-and-components/registry.md
tokoko: Not sure, I mimicked this [line](https://github.com/feast-dev/feast/blob/master/docs/getting-started/architecture-and-components/offline-store.md?plain=1#L5) from offline store markdown.

tokoko:
I agree it should be explained early on, but the text in both docs was much more like a technical description of feast component rather than a concept explanation. As such, considering that registry is as much an architectural component like online/offline stores, I think it makes sense to keep this specific doc text here.

On the concepts side, I think we should improve [overview](https://docs.feast.dev/getting-started/concepts/overview#feature-registration-and-retrieval) which already mentions registry but only briefly.

docs/getting-started/architecture-and-components/registry.md
franciscojavierarceo: Oh yes it will nvm.

franciscojavierarceo:
Okay, I'm comfortable with adding the current stuff to concepts but, instead, can you add (what is called in my PR a Component) that gives an overview of the registry concept rather than the technical? 

If you're comfortable with that, can you approve my PR so you can rebase?

tokoko:
@franciscojavierarceo rebased to latest now

docs/getting-started/components/registry.md
```diff
@@ -1,31 +1,47 @@
 # Registry
 
-The Feast feature registry is a central catalog of all the feature definitions and their related metadata. It allows data scientists to search, discover, and collaborate on new features.
```
franciscojavierarceo, line 3: I actually like this sentence as it helps give a high level understanding of it. Can we keep it?

docs/getting-started/components/registry.md
```diff
@@ -1,31 +1,47 @@
...
+Feast uses a registry to store all applied Feast objects (e.g. Feature views, entities, etc). It allows data scientists to search, discover, and collaborate on new features. The registry exposes methods to apply, list, retrieve and delete these objects, and is an abstraction with multiple implementations.
```
franciscojavierarceo, line 3: I like this too so maybe just a modification.

docs/getting-started/components/registry.md
```diff
@@ -1,31 +1,47 @@
...
+```yaml
+project: feast_demo_aws
+provider: aws
+registry: s3://feast-test-s3-bucket/registry.pb
+online_store: null
+offline_store:
+  type: file
 ```
 
-{% hint style="info" %}
```
franciscojavierarceo, line 28: Why take out this hint? I think it's useful.

franciscojavierarceo (changes requested):
Some small changes but looks good

docs/getting-started/components/registry.md
tokoko: Because this only mostly applies to file-based registries, sql registry used protobuf but isn't a itself a protobuf object. I suppose I can add this back with some caveats mentioned.

docs/getting-started/components/registry.md
tokoko: Sure, I'll merge the two.

docs/SUMMARY.md
franciscojavierarceo: bump on this. i think in another PR we can add metadata as the registry in the arch section

docs/getting-started/components/registry.md
franciscojavierarceo: Thank you!

docs/SUMMARY.md
tokoko: yeah, agreed on that. Do you think we need to make any related changes in this PR as well?

franciscojavierarceo: Let me clarify. I think we should still keep registry in the components section.

franciscojavierarceo: So in this pr it should be in concepts and in another PR we can add metadata in architecture (which is the construct that registry aims to implement).

franciscojavierarceo: Oh I see you moved it to components. We're good.

franciscojavierarceo (approved):
Nice!!

PR #4413, opened 2024-08-15
fix: Using repo_config parameter in teardown to allow for feature-store-yaml overrides

danbaron63:

* Allows the use of `--feature-store-yaml` when running `teardown` (`feast --feature-store-yaml <some-other-file> teardown`) as this currently does not work as the `.yaml` file is resolved differently to `plan` and `apply`.
* Fixes some issues with linting and formatting that prevent committing & pushing.

tbd

# Misc

### DIFF
--- a/sdk/python/tests/integration/feature_repos/universal/data_sources/file.py
+++ b/sdk/python/tests/integration/feature_repos/universal/data_sources/file.py
@@ -5,6 +5,7 @@ import subprocess
 import tempfile
 import uuid
 from pathlib import Path
+from subprocess import Popen
 from typing import Any, Dict, List, Optional
 
 import pandas as pd
@@ -367,7 +368,7 @@ class RemoteOfflineStoreDataSourceCreator(FileDataSourceCreator):
     def __init__(self, project_name: str, *args, **kwargs):
         super().__init__(project_name)
         self.server_port: int = 0
-        self.proc = None
+        self.proc: Optional[Popen[bytes]] = None
 
     def setup(self, registry: RegistryConfig):
         parent_offline_config = super().create_offline_store_config()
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -16,9 +16,6 @@
 [![License](https://img.shields.io/badge/License-Apache%202.0-blue)](https://github.com/feast-dev/feast/blob/master/LICENSE)
 [![GitHub Release](https://img.shields.io/github/v/release/feast-dev/feast.svg?style=flat&sort=semver&color=blue)](https://github.com/feast-dev/feast/releases)
 
-## Join us on Slack!
```
franciscojavierarceo, line 19: Can you revert this change?

franciscojavierarceo (changes requested):
Thanks for making this! Left a request for some small changes.

README.md
danbaron63: Ah sorry - don't know where that came from - reverted it!

danbaron63: Okay I was able to remove this but only using `--no-verify`

danbaron63:
> Thanks for making this! Left a request for some small changes.

Thanks for reviewing so quickly, wasn't sure if this was intended behaviour on `teardown` but this does appear to fix the issue.

I'm unable to run all tests locally without at least a couple of failures (I'm on M1 mac which may not help :/) - would appreciate if someone could test this as well as I'm also relatively new to this code base!

Thanks again!

sdk/python/tests/integration/feature_repos/universal/data_sources/file.py
```diff
@@ -382,13 +383,13 @@ def setup(self, registry: RegistryConfig):
...
             yaml.dump(config.model_dump(by_alias=True), outfile)
-        repo_path = str(repo_path.resolve())
+        repo_path = Path(str(repo_path.resolve()))
 
         self.server_port = free_port()
         host = "0.0.0.0"
         cmd = [
             "feast",
-            "-c" + repo_path,
+            f"-c {repo_path}",
```
franciscojavierarceo, line 392: it looks like my suggestion here may have broken things

franciscojavierarceo: You can revert it, sorry!

franciscojavierarceo (approved):
lgtm!

franciscojavierarceo:
Looks like this breaks the postgres  `sdk/python/tests/integration/registration/test_universal_registry.py` test. Do you mind digging into this? Running the tests locally should work with the `test-python-integration-local`. You can also just run the following command:
```bash
FEAST_IS_LOCAL_TEST=True \
python -m pytest -n 8 --color=yes --integration --durations=5 --dist loadgroup \
-k "not test_lambda_materialization and not test_snowflake_materialization" \
sdk/python/tests/integration/registration/test_universal_registry.py
```

PR #4420, opened 2024-08-17
fix: Default to pandas mode if not specified in ODFV proto in database

EXPEbdodla:
Used pandas mode as default for the ODFV objects that are already present in database. Without this it complaints that mode is missing to convert from Proto to ODFV. 

# Misc

### DIFF
--- a/sdk/python/feast/on_demand_feature_view.py
+++ b/sdk/python/feast/on_demand_feature_view.py
@@ -346,7 +346,7 @@ class OnDemandFeatureView(BaseFeatureView):
             ],
             sources=sources,
             feature_transformation=transformation,
-            mode=on_demand_feature_view_proto.spec.mode,
+            mode=on_demand_feature_view_proto.spec.mode or "pandas",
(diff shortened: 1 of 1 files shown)


HaoXuAI (approved):
LGTM

franciscojavierarceo (approved):
Hmm this is defaulted to pandas in the class, did this have issues for you?

EXPEbdodla:
> Hmm this is defaulted to pandas in the class, did this have issues for you?

Defaulted in class if we don't pass it. But from_proto() method we are passing None so it's using None instead of default value of the parameter.

franciscojavierarceo:
Got it, in a follow up PR can you add a test to cover this? I did that in a different PR. Thanks for catching and fixing this!!!

EXPEbdodla:
> Got it, in a follow up PR can you add a test to cover this? I did that in a different PR. Thanks for catching and fixing this!!!

I'll check how I can add this. This is the use case where an ODFV is already registered with old code in Database and the proto doesn't deserialize correctly.

PR #4424, opened 2024-08-19
fix: Add feast-operator Makefile to semantic-release script

tchughesiv:
This fixes the issue where past releases are not properly versioning the feast-operator.


tokoko (approved):
lgtm

PR #4433, opened 2024-08-21
feat: Intra server to server communication

tmihalac:
All the server-to-server requests should not be secured. Only client requests are secured.
This PR introduces a mechanism for intra server to server.

Which issue(s) is related to PR:
(https://github.com/feast-dev/feast/issues/4198)


infra/charts/feast-feature-server/templates/deployment.yaml
```diff
@@ -36,6 +36,8 @@ spec:
           env:
             - name: FEATURE_STORE_YAML_BASE64
               value: {{ .Values.feature_store_yaml_base64 }}
+            - name: INTRA_COMMUNICATION_BASE64
+              value: {{ "intra-server-communication" | b64enc }}
```
redhatHameed, line 40: where this values come from  ->  `{{ "intra-server-communication" | b64enc }}`

sdk/python/feast/permissions/security_manager.py
```diff
@@ -102,8 +103,12 @@ def assert_permissions(
     Raises:
         PermissionError: If the current user is not authorized to execute the requested actions on the given resources.
     """
+    intra_communication_base64 = os.getenv("INTRA_COMMUNICATION_BASE64")
     sm = get_security_manager()
-    if sm is None:
+    if sm is None or (
```
tokoko, line 108: If the goal is to reduce the performance penalty in intra-server communication as much as possible, I think we will have to go a bit further and expose a method like `is_auth_necessary` that servers will call directly. In our design, servers sometimes need to fetch resources from the registry simply to pass them to security manager and they will still incur that penalty even with this additional if check.

infra/charts/feast-feature-server/templates/deployment.yaml
tmihalac: Its a hard coded value at the moment

redhatHameed: @tokoko what do you think introducing new environment variable  `INTRA_COMMUNICATION_BASE64` or we can add this property inside feature_store.yaml under auth section. The draw back I am seeing client can configure the same property to hack this.

redhatHameed: The other approach is configure/create the full access roles for the servers but that again will cause the performance issue and server keep generating the token or fetching the roles .

redhatHameed: > Its a hard coded value at the moment

I see will will converted into base64 automatically when will be deployed ?

sdk/python/feast/permissions/security_manager.py
tmihalac: The goal in general is not to check permissions in intra server communication.
I tried to look for the scenario you were saying but I couldn't find any, can you please give me an example in the code?

infra/charts/feast-feature-server/templates/deployment.yaml
tmihalac: That was @dmartinol suggestion

sdk/python/feast/permissions/security_manager.py
tmihalac: > If the goal is to reduce the performance penalty in intra-server communication as much as possible, I think we will have to go a bit further and expose a method like `is_auth_necessary` that servers will call directly. In our design, servers sometimes need to fetch resources from the registry simply to pass them to security manager and they will still incur that penalty even with this additional if check.

@tokoko The goal in general is not to check permissions in intra server communication.
I tried to look for the scenario you were saying but I couldn't find any, can you please give me an example in the code?

tokoko: Looks like there aren't many currently, but that's mostly because we have a bug for apply method permission checking (#4444).

btw, I'm fine with proceeding with the current approach for now and optimizing it later as well. Just pointing it out that this way there still might be a performance impact.

infra/charts/feast-feature-server/templates/deployment.yaml
tokoko: > The draw back I am seeing client can configure the same property to hack this.

I don't really see the difference, but env variable sounds better.

sdk/python/feast/permissions/auth/kubernetes_token_parser.py
```diff
@@ -41,10 +42,14 @@ async def user_details_from_access_token(self, access_token: str) -> User:
         current_user = f"{sa_namespace}:{sa_name}"
         logging.info(f"Received request from {sa_name} in {sa_namespace}")
 
-        roles = self.get_roles(sa_namespace, sa_name)
-        logging.info(f"SA roles are: {roles}")
+        intra_communication_base64 = os.getenv("INTRA_COMMUNICATION_BASE64")
+        if sa_name is not None and sa_name == intra_communication_base64:
+            return User(username=sa_name, roles=[])
```
tokoko, line 47: Wouldn't it be cleaner to introduce a magic role here (`admin` for example) and then security manager will check for that role instead of dealing with a username?

One upside there is that a user will also be able to take advantage of this and assign that `admin` role to normal users if that's the behavior they want. wdyt?

sdk/python/feast/permissions/security_manager.py
dmartinol: @tokoko agree with prepending the asserts with this `is_auth_necessary` flag, now that [you explained](https://github.com/feast-dev/feast/issues/4444#issuecomment-2309822627) the reasons for the double asserts in the `apply` methods

dmartinol: @tmihalac can we introduce the `is_auth_necessary` function and replace this `if` blocks?

sdk/python/tests/unit/permissions/auth/test_token_parser.py
```diff
@@ -63,6 +63,58 @@ def test_oidc_token_validation_failure(mock_oauth2, oidc_config):
...
+    monkeypatch.setattr(
+        "feast.permissions.auth.oidc_token_parser.PyJWKClient.get_signing_key_from_jwt",
+        lambda self, access_token: signing_key,
+    )
+
+    user_data = {
+        "preferred_username": f"{intra_communication_val}",
+    }
+
+    if intra_communication_val != "test1234":
```
dmartinol, line 87: instead of using this check, you can add another pytest parameter with a boolean defining the expected behavior, like `is_intra_server`

sdk/python/tests/unit/permissions/auth/test_token_parser.py
```diff
@@ -120,3 +172,70 @@ def test_k8s_token_validation_failure(mock_jwt, mock_config):
         asyncio.run(
             token_parser.user_details_from_access_token(access_token=access_token)
         )
+
+
+@mock.patch.dict(os.environ, {"INTRA_COMMUNICATION_BASE64": "test1234"})
```
dmartinol, line 177: same comments as above

sdk/python/tests/unit/permissions/conftest.py
```diff
@@ -49,6 +49,7 @@ def users() -> list[User]:
     users.append(User("w", ["writer"]))
     users.append(User("rw", ["reader", "writer"]))
     users.append(User("admin", ["reader", "writer", "admin"]))
+    users.append(User("test1234", []))
```
dmartinol, line 52: I don't think we need this entry, it should mock the users from the IDP server instead. can you try w/o it?

sdk/python/feast/permissions/auth/kubernetes_token_parser.py
dmartinol: You already raised the point in the past ([here](https://github.com/feast-dev/feast/issues/4198#issuecomment-2187122275)), but the reason for not taking this approach is to prevent users from assigning the same role to their own `Permission`s. This could be challenging both to explain and to implement.

tokoko: Fair enough, I just think that the ability to assign superadmin role to specific users might be a feature request from users later on. Still, no reason for us to implement it now.

dmartinol: Agree. ATM the `superadmin` requires a manual definition like:
```py
Permission(
    name="admin_permission",
    types=ALL_RESOURCE_TYPES,
    policy=RoleBasedPolicy(roles=["superadmin"]),
    actions=ALL_ACTIONS
)
```
We should expose a simplified configuration for this user with extended privileges, maybe another field under the `auth` section of `feature_store.yaml`. As you suggested, let's wait for feedback from the early adopters 😉

sdk/python/tests/unit/permissions/auth/test_token_parser.py
```diff
@@ -63,6 +63,58 @@ def test_oidc_token_validation_failure(mock_oauth2, oidc_config):
         )
 
 
+@mock.patch.dict(os.environ, {"INTRA_COMMUNICATION_BASE64": "test1234"})
+@pytest.mark.parametrize("intra_communication_val", ("test1234", "my-name"))
+def test_oidc_inter_server_comm(intra_communication_val, oidc_config, monkeypatch):
```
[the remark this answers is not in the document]
tmihalac: Not necessarily in this case yes

sdk/python/tests/unit/permissions/auth/test_token_parser.py
tmihalac: > instead of using this check, you can add another pytest parameter with a boolean defining the expected behavior, like `is_intra_server`

I need the actual value in the environment variable to compare it its not enough to know that the the environment variable is set but it must be the same in the client and server

sdk/python/tests/unit/permissions/auth/test_token_parser.py
tmihalac: Look for my comment above it is related to this as well

sdk/python/tests/unit/permissions/auth/test_token_parser.py
dmartinol: so you don't need to know the env variable, you set the value in the parameter

dmartinol (approved):

dmartinol (approved):

redhatHameed:
any documentation need for this ?

tmihalac:
> any documentation need for this ?

I don't think so, as far as I understand we decided not to expose the env variable for security purposes.

redhatHameed (approved):

redhatHameed:
@tokoko can we merge this if there is no any further comments .

sdk/python/feast/permissions/security_manager.py
tokoko: I think @dmartinol also meant to use `is_auth_necessary` in the servers to avoid unnecessary calculations, right?

redhatHameed:
@tmihalac @dmartinol @lokeshrangineni also now as username and password not required incase if the server can we make this optional [here](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/permissions/auth_model.py#L14-L15)

tmihalac:
> @tmihalac @dmartinol @lokeshrangineni also now as username and password not required incase if the server can we make this optional [here](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/permissions/auth_model.py#L14-L15)

I am not sure I understand what is the connection between auth_model.py and the intra server communication

redhatHameed:
> > @tmihalac @dmartinol @lokeshrangineni also now as username and password not required incase if the server can we make this optional [here](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/permissions/auth_model.py#L14-L15)
> 
> I am not sure I understand what is the connection between auth_model.py and the intra server communication

this isn’t directly related, but I’ve found below validation error while deploying server it always asks for a username and password, with OIDC, and it throws a validation error if those fields are empty. Maybe we could just make the username and password fields optional WDYT?

```: 2 validation errors for OidcAuthConfig
username
Field required [type=missing, input_value={'type': 'oidc', 'auth_di..., 'realm': 'feast-rbac'}, input_type=dict]
For further information visit https://errors.pydantic.dev/2.8/v/missing
password```

tmihalac:
> > > @tmihalac @dmartinol @lokeshrangineni also now as username and password not required incase if the server can we make this optional [here](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/permissions/auth_model.py#L14-L15)
> > 
> > 
> > I am not sure I understand what is the connection between auth_model.py and the intra server communication
> 
> this isn’t directly related, but I’ve found below validation error while deploying server it always asks for a username and password, with OIDC, and it throws a validation error if those fields are empty. Maybe we could just make the username and password fields optional WDYT?
> 
> ```
> username
> Field required [type=missing, input_value={'type': 'oidc', 'auth_di..., 'realm': 'feast-rbac'}, input_type=dict]
> For further information visit https://errors.pydantic.dev/2.8/v/missing
> password```
> ```

I am not sure, maybe ask @lokeshrangineni but aren't those parameters mandatory when auth type is oidc ?

redhatHameed:
> > > > @tmihalac @dmartinol @lokeshrangineni also now as username and password not required incase if the server can we make this optional [here](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/permissions/auth_model.py#L14-L15)
> > > 
> > > 
> > > I am not sure I understand what is the connection between auth_model.py and the intra server communication
> > 
> > 
> > this isn’t directly related, but I’ve found below validation error while deploying server it always asks for a username and password, with OIDC, and it throws a validation error if those fields are empty. Maybe we could just make the username and password fields optional WDYT?
> > ```
> > username
> > Field required [type=missing, input_value={'type': 'oidc', 'auth_di..., 'realm': 'feast-rbac'}, input_type=dict]
> > For further information visit https://errors.pydantic.dev/2.8/v/missing
> > password```
> > ```
> 
> I am not sure, maybe ask @lokeshrangineni but aren't those parameters mandatory when auth type is oidc ?

AFAIK  for the client yes but incase server these should not mandatory.

lokeshrangineni:
> > > > > @tmihalac @dmartinol @lokeshrangineni also now as username and password not required incase if the server can we make this optional [here](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/permissions/auth_model.py#L14-L15)
> > > > 
> > > > 
> > > > I am not sure I understand what is the connection between auth_model.py and the intra server communication
> > > 
> > > 
> > > this isn’t directly related, but I’ve found below validation error while deploying server it always asks for a username and password, with OIDC, and it throws a validation error if those fields are empty. Maybe we could just make the username and password fields optional WDYT?
> > > ```
> > > username
> > > Field required [type=missing, input_value={'type': 'oidc', 'auth_di..., 'realm': 'feast-rbac'}, input_type=dict]
> > > For further information visit https://errors.pydantic.dev/2.8/v/missing
> > > password```
> > > ```
> > 
> > 
> > I am not sure, maybe ask @lokeshrangineni but aren't those parameters mandatory when auth type is oidc ?
> 
> AFAIK for the client yes but incase server these should not mandatory.

Yes, it is mandatory for client and not for server. May be better to create separate task/issue and handle it as it is not related to this PR.

sdk/python/feast/permissions/security_manager.py
tmihalac: @tokoko I missed adding `is_auth_necessary` in newly added method `assert_permissions_to_update` in `security_manager.py`  is this what you meant?

tokoko: yes, exactly

tmihalac:
@tokoko Can we merge this PR?

tokoko (approved):
lgtm

PR #4435, opened 2024-08-22
feat: Add registry methods for dealing with all FV types

### LINKED ISSUE #4235: Registry get_feature_view should read stream and on-demand feature views

tokoko:

- Supersedes #4328
- Adds new methods `get_any_feature_view` and `list_all_feature_views` to `BaseRegistry` that return `BaseFeatureView` objects.

Fixes #4235


tokoko:
fyi @dmartinol

sdk/python/feast/registry_server.py
```diff
@@ -169,6 +192,27 @@ def GetFeatureView(
             actions=[AuthzedAction.DESCRIBE],
         ).to_proto()
 
+    def GetAnyFeatureView(
+        self, request: RegistryServer_pb2.GetAnyFeatureViewRequest, context
+    ):
+        feature_view = assert_permissions(
```
dmartinol, line 198: 👍

dmartinol (commented):
lgtm!

dmartinol (approved):

protos/feast/registry/RegistryServer.proto
```diff
@@ -31,9 +31,13 @@ service RegistryServer{
...
   // FeatureView RPCs
   rpc ApplyFeatureView (ApplyFeatureViewRequest) returns (google.protobuf.Empty) {}
+  rpc DeleteFeatureView (DeleteFeatureViewRequest) returns (google.protobuf.Empty) {}
+  rpc GetAnyFeatureView (GetAnyFeatureViewRequest) returns (GetAnyFeatureViewResponse) {}
+  rpc ListAllFeatureViews (ListAllFeatureViewsRequest) returns (ListAllFeatureViewsResponse) {}
+
+  // plain FeatureView RPCs
   rpc GetFeatureView (GetFeatureViewRequest) returns (feast.core.FeatureView) {}
   rpc ListFeatureViews (ListFeatureViewsRequest) returns (ListFeatureViewsResponse) {}
-  rpc DeleteFeatureView (DeleteFeatureViewRequest) returns (google.protobuf.Empty) {}
```
franciscojavierarceo, line 36: Don't delete protos, rather mark them as deprecated.

https://protobuf.dev/programming-guides/proto3/

franciscojavierarceo (changes requested):
Blocking PR

protos/feast/registry/RegistryServer.proto
tokoko: it's not a deletion, just a reorder, look at line 34

franciscojavierarceo: My bad, looked at this on my phone and got worried there.

protos/feast/registry/RegistryServer.proto
```diff
@@ -201,6 +205,35 @@ message DeleteFeatureViewRequest {
     bool commit = 3;
 }
 
+message AnyFeatureView {
```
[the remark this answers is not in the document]
franciscojavierarceo: Hmm, this is a single message though so I see.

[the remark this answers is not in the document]
franciscojavierarceo: Why wouldn't we call it FeatureViewType?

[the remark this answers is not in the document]
tokoko: I feel like type is something else, it would refer to one the types, not one of the objects. `AnyFeatureView` basically stands for `AnyTypeOfFeatureView`.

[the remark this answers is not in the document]
franciscojavierarceo: Oh I see what you're doing this for. 

Man this is ugly. 

The `get_any_feature_view` should probably be an intermediate solution and instead we should find a way to make `get_feature_view` include ODFV and Stream FVs.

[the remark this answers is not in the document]
franciscojavierarceo: I get why we can't jump right to that, but maybe we should start a doc on what should be reconciled for release to Feast 1.0?

sdk/python/feast/infra/registry/registry.py
```diff
@@ -557,7 +557,26 @@ def apply_materialization(
                     self.commit()
                 return
 
-        raise FeatureViewNotFoundException(feature_view.name, project)
```
franciscojavierarceo, line 560: Did you mean to drop this from `apply_materialization`?

sdk/python/tests/integration/registration/test_universal_registry.py
```diff
@@ -1154,8 +1167,14 @@ def simple_udf(x: int):
         project, tags=sfv.tags
     )
 
+    all_feature_views = test_registry.list_all_feature_views(project, tags=sfv.tags)
+
+    print(stream_feature_views)
```
franciscojavierarceo, line 1172: Can remove these print statements now

franciscojavierarceo (approved):
Nice, this is great, thanks Tornike! 

As mentioned in a comment, I think we should start jotting down some major clean ups we want to do to release 1.0. I'm experiencing this as I'm working through the writes to the ODFV and I think there's a way for us to clean up a lot more. 

There's a lot of code in between class instantiation and `apply` that is very opaque and makes it hard to reason about what's happening. It's for reasonable choices but I think it warrants a broad review of the online store.

Also, it could be a good time to establish DuckDB as the offline store default.

tokoko:
> As mentioned in a comment, I think we should start jotting down some major clean ups we want to do to release 1.0. I'm experiencing this as I'm working through the writes to the ODFV and I think there's a way for us to clean up a lot more.

Yeah, there's probably a lot of discussion to be had until we finalize these APIs for 1.0 release. We should definitely continue the debate in #4235. The problem with `get_feature_view` is that it might be interpreted as referring to all feature views and also to plain `FeatureView` objects specifically. 

> Also, it could be a good time to establish DuckDB as the offline store default.

Yeah, that's probably a pretty good idea. I recently had to make a simple demo and saw that on small datasets duckdb was a lot faster than the current default, dask. I'll open a ticket for this and let's continue there, namely I think we should agree whether duckdb being the default means that it should be a required dependency or not (I think not...)

franciscojavierarceo:
> > As mentioned in a comment, I think we should start jotting down some major clean ups we want to do to release 1.0. I'm experiencing this as I'm working through the writes to the ODFV and I think there's a way for us to clean up a lot more.
> 
> Yeah, there's probably a lot of discussion to be had until we finalize these APIs for 1.0 release. We should definitely continue the debate in #4235. The problem with `get_feature_view` is that it might be interpreted as referring to all feature views and also to plain `FeatureView` objects specifically.
> 
> > Also, it could be a good time to establish DuckDB as the offline store default.
> 
> Yeah, that's probably a pretty good idea. I recently had to make a simple demo and saw that on small datasets duckdb was a lot faster than the current default, dask. I'll open a ticket for this and let's continue there, namely I think we should agree whether duckdb being the default means that it should be a required dependency or not (I think not...)

I think a default offline store should be required, so in that case we shouldn't make it the default. 

What are your thoughts about us drafting a 1.0 Roadmap? It's September now, so a rational question would be: could we get it to 1.0 by end of the year? I'm not sure.

PR #4436, opened 2024-08-22
chore: Mark tests using keycloak with xdist_group

tokoko:
- Adds a workaround to mark tests with xdist markers using new static method on `DataSourceCreator`
- Increases the number of workers for integration tests back to 8


tokoko:
cc @lokeshrangineni

franciscojavierarceo (approved):
lgtm

Issue #4199, opened 2024-05-13
Semi-isolated subprojects


tokoko:
**Is your feature request related to a problem? Please describe.**

Feast has a concept of a `project` that can be used to host several completely isolated resources in a shared registry. This can be used to co-host registry objects for unrelated teams or environments like dev, test, prod. (honestly, not sure why would anybody do that though, a separate registry for each sounds a lot better, there's not much of an overhead). This doesn't address the use case when a user might want to host a shared pool of resources with various ownerships, for example some core objects (entities) that are managed by a core team, with additional feature views and feature services that are owned by different teams. 

This is hard to achieve with normal `feast apply` workflow as applying a subset of feast objects will lead to the removal of other objects.

**Describe the solution you'd like**

1. Introduce a concept of a `subproject` (or maybe call this something else, happy to consider alternatives) which will be used to mark feast objects that are managed as a whole from a single git repository. The user should be able to run `feast apply --subproject team-1` that will overwrite only existing objects that were part of this subproject. Any attempt to apply a change that will break the integrity of a registry as a whole (trying to remove an entity that's referenced by a feature view that belongs to another subproject) should lead to an error.

2. Another option is to repurpose the existing `project` label for this, but this might be too big of a breaking change.

P.S. Subprojects should eventually also fit into the overall security model #4198.

tokoko:
@EXPEbdodla curious what you think about this

EXPEbdodla:
@tokoko Glad you tagged me to this. I was thinking this as an extension but slightly different. What I'm thinking is, Project object should store RepoConfig object instead of defining feature_store.yaml. We can still use feature_store.yaml as an override. feature_store.yaml can help to override multiple projects if necessary. 

All the Feast Proto objects already have Project name in their spec. We need to update Feast Objects in Python to accept Project name So we can tag Feast Objects to Projects. 

Lets say we have a single GITHUB repo to manage all the projects.

During Registration/ Apply stage, We apply all the projects. GITHUB Repo will be in sync with registry for the specified projects. or we can only apply projects specified. 

During Materialization stage, Materialize the feature views based on project name and feature view name specified. RepoConfig is already registered in Registry. We can use that or accept override. 

[...]

PR #4441, opened 2024-08-24
fix: Added Apply/Get/List/Delete methods to project_metadata to BaseRegistry

EXPEbdodla:
1. Adds Apply/Get/List/Delete methods to project_metadata to BaseRegistry
2. Made project as Optional in list_project_metadata method so we can use it retrieve all projects when project is None. Can be deleted in future if necessary. 
3. Renamed _maybe_init_project_metadata to apply_project_metadata. Registry Server will have use cases to create projects in future. 
4. I'll implement the methods to remote registry in future PRs

# Misc

### DIFF
--- a/sdk/python/feast/infra/registry/base_registry.py
+++ b/sdk/python/feast/infra/registry/base_registry.py
@@ -546,18 +546,68 @@ class BaseRegistry(ABC):
         raise NotImplementedError
 
     @abstractmethod
+    def apply_project_metadata(
+        self,
+        project: str,
+        commit: bool = True,
+    ):
+        """
+        Persist a project metadata with a new uuid
+
+        Args:
+            project: Feast project name
+            commit: Whether the change should be persisted immediately
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def delete_project_metadata(
+        self,
+        project: str,
+        commit: bool = True,
+    ):
+        """
+        Deletes a project metadata or raises ProjectMetadataNotFoundException exception if not found.
+
+        Args:
+            project: Feast project name
+            commit: Whether the change should be persisted immediately
+        """
+        raise NotImplementedError
+
+    @abstractmethod
     def list_project_metadata(
-        self, project: str, allow_cache: bool = False
+        self,
+        project: Optional[str],
+        allow_cache: bool = False,
     ) -> List[ProjectMetadata]:
         """
-        Retrieves project metadata
+        Retrieves project metadata if given project name otherwise all project metadata
+
+        Args:
+            project: Filter metadata based on project name or None to retrieve all project metadata
+            allow_cache: Allow returning feature views from the cached registry
+
+        Returns:
+            List of project metadata
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_project_metadata(
+        self,
+        project: str,
+        allow_cache: bool = False,
+    ) -> Optional[ProjectMetadata]:
+        """
+        Retrieves project metadata if present otherwise None
 
         Args:
             project: Filter metadata based on project name
             allow_cache: Allow returning feature views from the cached registry
 
         Returns:
-            List of project metadata
+            Get project metadata if project exists otherwise None
         """
         raise NotImplementedError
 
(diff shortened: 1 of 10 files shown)


sdk/python/feast/infra/registry/base_registry.py
```diff
@@ -545,22 +545,72 @@ def list_validation_references(
         """
         raise NotImplementedError
 
+    @abstractmethod
+    def apply_project_metadata(
```
tokoko, line 549: Wouldn't it make more sense for this to also accept `ProjectMetadata` object as an input similar to other registry resources? I know the object doesn't hold much information right now, but if we plan for it to be "updatable", an update should happen through this method, right?

tokoko:
This probably means we will have to treat `ProjectMetadata` as another one of `FeastObject` types and should be added in [here](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/feast_object.py#L25). Similarly for rbac to work, we need it [here](https://github.com/feast-dev/feast/blob/master/protos/feast/core/Permission.proto#L38) as well, but that can be deferred until handling in remote registry is implemented. @dmartinol 

Another point... if we treat `apply_project_metadata` as basically project creation and `delete_project_metadata` as deletion, along with all the contents (implementations seem to indicate this), wouldn't better name for this new feast object be `Project` rather than `ProjectMetadata`?

sdk/python/feast/infra/registry/base_registry.py
EXPEbdodla: I was not envisioning this as a updatable. At this point its create only. Fields we have for this are UUID and LastUpdateTimestamp. UUID should be constant. LastUpdateTimestamp should be update when any changes happen to any objects associated to the project.

EXPEbdodla:
> Another point... if we treat apply_project_metadata as basically project creation and delete_project_metadata as deletion, along with all the contents (implementations seem to indicate this), wouldn't better name for this new feast object be Project rather than ProjectMetadata?

@tokoko You are right. We can introduce the Project object with spec and metadata. In SQL Registry, currently it's stored in feast_metadata table, that will eventually of no use. It should be the case in other registries too. 

Do you recommend support for both Project and ProjectMetadata for a while? Or replace Project with ProjectMeatadata?

EXPEbdodla:
> This probably means we will have to treat ProjectMetadata as another one of FeastObject types and should be added in [here](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/feast_object.py#L25).

We won't treat this as an object that users define in Feature Definitions. We can populate project name from feature_store.yaml. At this point, I'm not aware of any use cases for this.

tokoko:
> Defining Project object may be a case for RBAC. I have no idea about this at this point. Looking forward for the suggestions on this.

yup, that's what I think as well. A lot of rbac revolves around tagging, so it makes sense to make it possible to tag projects and consequently we will have to let users update those tags. Maybe we can make it so that users are able to define `Project` objects in the repo, but revert to auto-generating one for them if they don't do it.

@franciscojavierarceo @HaoXuAI Any thoughts/objections about introducing `Project` feast object and phasing out `ProjectMetadata` in favor of it? I think it makes a lot of sense especially considering rbac.

dmartinol:
> This probably means we will have to treat ProjectMetadata as another one of FeastObject types and should be added in [here](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/feast_object.py#L25). Similarly for rbac to work, we need it [here](https://github.com/feast-dev/feast/blob/master/protos/feast/core/Permission.proto#L38) as well, but that can be deferred until handling in remote registry is implemented. @dmartinol

Agreee, while developing the registry_server we'll also need to assert the required permissions.

HaoXuAI:
No objection. Can we have multiple projects per repo where different project use different online/offline store setup?

EXPEbdodla:
> No objection. Can we have multiple projects per repo where different project use different online/offline store setup?

@HaoXuAI We are discussing about the multiple Projects in the issue: https://github.com/feast-dev/feast/issues/4199. Project object can help us enable multiple projects per repo.

tokoko:
@EXPEbdodla this is no longer relevant, right?

franciscojavierarceo:
Is this PR still relevant? Thanks for doing this @EXPEbdodla and sorry for the complications here.

EXPEbdodla:
Not relevant. I'll close it.

PR #4443, opened 2024-08-24
fix: Locate feature_store.yaml from __file__

Atry:

This PR changes tests so that it can properly locate `feature_store.yaml` from `__file__` instead of CWD, which only works when running `cd feature_ropo && python ./test_workflow.py`, but does not work when running `python feature_ropo/test_workflow.py`.

Not reported as an issue

# Misc


tokoko (approved):
lgtm

PR #4449, opened 2024-08-26
fix: Validating permission to update an existing request on both the new and the old instance

### LINKED ISSUE #4444: Update permissions aren't being enforced in registry server
## Current Behavior

`apply_x` methods in registry server check permissions on the object that's passed in the request. This is correct for object creation, but overlooks updates. During updates we should be checking permissions on the existing object as well. With current behavior, user is allowed to overwrite objects even when updates aren't allowed.

## Possible Solution
Pseudocode for a typical apply method should look something like this:
- Check if the object already exists in the registry
- Assert update permission on the existing object (if it exists)
- Assert create permission on the object contained in the request
- Proceed with apply

dmartinol:
Fixes #4444 

# Misc
Will update the code using the new `get_any_feature_view` method once #4435 is merged

### DIFF
--- a/sdk/python/feast/permissions/security_manager.py
+++ b/sdk/python/feast/permissions/security_manager.py
@@ -87,20 +87,19 @@ class SecurityManager:
 
 def assert_permissions_to_update(
     resource: FeastObject,
-    actions: Union[AuthzedAction, List[AuthzedAction]],
     getter: Callable[[str, str, bool], FeastObject],
     project: str,
     allow_cache: bool = True,
 ) -> FeastObject:
     """
-    Verify if the current user is authorized to execute the requested actions on the given resource and also checks the authorization on
-    the existing resource, if any.
+    Verify if the current user is authorized to create or update the given resource.
+    If the resource already exists, the user must be granted permission to execute DESCRIBE and UPDATE actions.
+    If the resource does not exist, the user must be granted permission to execute the CREATE action.
 
     If no permissions are defined, the result is to deny the execution.
 
     Args:
         resource: The resources for which we need to enforce authorized permission.
-        actions: The requested actions to be authorized.
         getter: The getter function used to retrieve the existing resource instance by name.
         The signature must be `get_permission(self, name: str, project: str, allow_cache: bool)`
         project: The project nane used in the getter function.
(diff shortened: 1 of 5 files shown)


sdk/python/feast/permissions/security_manager.py
```diff
@@ -84,6 +85,45 @@ def assert_permissions(
...
+        The signature must be `get_permission(self, name: str, project: str, allow_cache: bool)`
+        project: The project nane used in the getter function.
+        allow_cache: Whether to use cached data. Defaults to `True`.
+    Returns:
+        FeastObject: The original `resource`, if permitted.
+
+    Raises:
+        PermissionError: If the current user is not authorized to execute all the requested actions on the given resource or on the existing one.
+    """
+    resource_to_update = assert_permissions(resource=resource, actions=actions)
```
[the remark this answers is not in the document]
dmartinol: Makes sense (and it has changed to `DESCRIBE` instead of `READ`, after a long discussion 😉 ).
And to take it a step further, I’d also hide the `actions` parameter entirely, since the actions are already implied by the function.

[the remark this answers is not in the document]
tokoko: and here we go.. I'd have totally mixed them up 😄 +1 for dropping actions param as well

sdk/python/feast/permissions/security_manager.py
```diff
@@ -111,16 +110,17 @@ def assert_permissions_to_update(
...
-            name=resource_to_update.name,
+            name=resource.name,
             project=project,
             allow_cache=allow_cache,
         )  # type: ignore[call-arg]
         assert_permissions(resource=existing_resource, actions=actions)
     except FeastObjectNotFoundException:
-        pass
+        actions = [AuthzedAction.CREATE]
+    resource_to_update = assert_permissions(resource=resource, actions=actions)
```
tokoko, line 123: Shouldn't this always be a create action even if the object exists? `update` gives the authorization to overwrite something, while `create` describes what you can overwrite with... That way `update`, `delete`, `describe` actions will always be describing existing objects, while `create` will be applied to only the new one. idk, that makes more sense to me, wdyt?

dmartinol: > while create describes what you can overwrite with
IMO create does not overwrite but generates a new instance. Maybe I'm not following your reasoning, could you give an example?

tokoko: I could, but never mind. this is more intuitive, thanks

tokoko (approved):
lgtm

PR #4453, opened 2024-08-26
fix: Add --chdir to test_workflow.py

Atry:

This PR add `--chdir` flag to test_workflow.py so that `python feature_ropo/test_workflow.py` would work.

Similar to #4443 but this PR fixes subprocesses

# Misc


shuchu (approved):
LGTM

PR #4456, opened 2024-08-27
fix: Check for snowflake functions when setting up materialization engine

benchoncy:

When running `feast apply` using the snowflake materialization engine, feast only checks if the stage in snowflake exists and assumes the functions have been created.  This allows a case where if anything goes wrong during function creation, such as permission issues, a stage can be created while the functions are not. Following attempts to `feast apply` see the created stage and tell the user that the functions exist without creating them. This leads to materialization on snowflake to fail down the line due to missing functions.

This PR aims to change the apply logic to instead check for snowflake functions before deciding if action is required.

# Misc

A current workaround for this is to delete the stage in Snowflake before running the next `feast apply`.


tokoko:
Why remove the existing stage check? Does checking for functions make checking for stage redundant?

tokoko:
thanks, makes sense. I guess this still doesn't account for the changes in function definitions themselves, right? If we add some function to be deployed, a user will have to remove the stage or all functions for them to be correctly recreated. A problem for another day...

tokoko (approved):
lgtm

PR #4458, opened 2024-08-28
fix: Initial commit targetting grpc registry server

### LINKED ISSUE #4392: Remote registry client does not map application errors
## Expected Behavior 
A feast application has a registry of type remote connected to a registry server.
The feast client invokes a an API on the feast registry that thrown an error, e.g. `store.get_feature_service("foo")` which raises
`FeatureServiceNotFoundException` from the registry server because `foo` is not a registered feature service.

The expectation is that the method invoked by the client raises the `FeatureServiceNotFoundException` and the application can detect it with:
```py
try:
    fs = store.get_feature_service("foo")
except FeatureServiceNotFoundException
    ....
````

## Current Behavior
The registy server always raises an `_InactiveRpcError` error instead.

This bug causes unexpected errors when the client invokes services whose implementation needs such expections to execute fall-back logic, like [_get_feature_views_to_materialize](https://github.com/feast-dev/feast/blob/125fa49fc700a8e6f1994c710a5076210fb97bf9/sdk/python/feast/feature_store.py#L684).

## Steps to reproduce
See `Expected Behavior` section.

### Specifications

- Version: 0.40.0
- Platform: any
- Subsystem: registry server

## Possible Solution
The registry server can wrap the original exception in an error status message that the remote registry client can catch and translate into the original exception:
[](https://github.com/grpc/grpc/tree/master/examples/python/errors)

### PR DESCRIPTION
**Pull request by dmartinol**
Intercept registry server failures and adds an additional message to the response with the failure status and details.
The proxy client then rebuilds the original error from the status data.

Partially fixes #4392 for Registry Server

# Misc
Need to add some UT/IT to prove the effectivness

### DIFF
--- a/sdk/python/feast/errors.py
+++ b/sdk/python/feast/errors.py
@@ -1,33 +1,51 @@
 from typing import Any, List, Optional, Set
 
 from colorama import Fore, Style
+from fastapi import status as HttpStatusCode
+from grpc import StatusCode as GrpcStatusCode
 
 from feast.field import Field
 
 
-class DataSourceNotFoundException(Exception):
+class FeastError(Exception):
+    pass
+
+    def rpc_status_code(self) -> GrpcStatusCode:
+        return GrpcStatusCode.INTERNAL
+
+    def http_status_code(self) -> int:
+        return HttpStatusCode.HTTP_500_INTERNAL_SERVER_ERROR
+
+
+class DataSourceNotFoundException(FeastError):
     def __init__(self, path):
         super().__init__(
             f"Unable to find table at '{path}'. Please check that table exists."
         )
 
 
-class DataSourceNoNameException(Exception):
+class DataSourceNoNameException(FeastError):
     def __init__(self):
         super().__init__(
             "Unable to infer a name for this data source. Either table or name must be specified."
         )
 
 
-class DataSourceRepeatNamesException(Exception):
+class DataSourceRepeatNamesException(FeastError):
     def __init__(self, ds_name: str):
         super().__init__(
             f"Multiple data sources share the same case-insensitive name {ds_name}."
         )
 
 
-class FeastObjectNotFoundException(Exception):
+class FeastObjectNotFoundException(FeastError):
     pass
+
+    def rpc_status_code(self) -> GrpcStatusCode:
+        return GrpcStatusCode.NOT_FOUND
+
+    def http_status_code(self) -> int:
+        return HttpStatusCode.HTTP_404_NOT_FOUND
 
 
 class EntityNotFoundException(FeastObjectNotFoundException):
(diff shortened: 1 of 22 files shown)

### CODE REVIEW

File: sdk/python/feast/permissions/server/grpc.py [hunk 1]
```diff
@@ -31,7 +32,7 @@ def grpc_interceptors(
     if auth_type == AuthManagerType.NONE:
         return None
 
-    return [AuthInterceptor()]
+    return [AuthInterceptor(), ErrorInterceptor()]
```
lokeshrangineni, line 35: Will it handle the exception happens in the `AuthInterceptor`?

dmartinol: It won't. Specific code is needed in the `AuthInterceptor` (and maybe also use a Feast-specific error instead)

tmihalac: I think that `ErrorInterceptor` should be called `if auth_type == AuthManagerType.NONE` as well

sdk/python/feast/permissions/server/grpc.py
```diff
@@ -52,3 +53,55 @@ def intercept_service(self, continuation, handler_call_details):
...
+                request_deserializer=handler.request_deserializer,
+                response_serializer=handler.response_serializer,
+            )
+        return handler
+
+
+def _error_to_status_code(error: Exception) -> grpc.StatusCode:
+    if isinstance(error, FeastObjectNotFoundException):
+        return grpc.StatusCode.NOT_FOUND
+    if isinstance(error, FeastObjectNotFoundException):
```
tmihalac, line 105: I think you meant `PermissionNotFoundException`

sdk/python/feast/permissions/server/grpc.py
dmartinol: @tmihalac yes, and also we need to move this interceptor in a different package, as it has nothing to do with the `feast.permission` package.
we'll review this later

sdk/python/feast/permissions/server/grpc.py
dmartinol: Resolved by https://github.com/feast-dev/feast/pull/4465

sdk/python/feast/permissions/server/grpc.py
dmartinol: > Will it handle the exception happens in the AuthInterceptor?

@lokeshrangineni due to the latest refactory, we need to implement our own FeastAuthenticationError, so I would create a separate ticket for that.

dmartinol:
@tokoko @tmihalac @lokeshrangineni ready for review now

tokoko:
@dmartinol I think you need to rebase

sdk/python/feast/registry_server.py
```diff
@@ -668,3 +670,21 @@ def start_server(store: FeatureStore, port: int, wait_for_termination: bool = Tr
         server.wait_for_termination()
     else:
         return server
+
+
+def grpc_interceptors(
+    auth_type: AuthManagerType,
+) -> Optional[list[grpc.ServerInterceptor]]:
+    """
+    A list of the authorization interceptors.
```
tokoko, line 679: remove "authorization" from the line

sdk/python/feast/permissions/server/grpc.py
dmartinol: @lokeshrangineni created https://github.com/feast-dev/feast/issues/4466

tokoko:
I'll force a retry, but that looks like a genuine credentials error

sdk/python/feast/errors.py
```diff
@@ -1,21 +1,71 @@
...
+        m = {
+            "module": f"{type(self).__module__}",
+            "class": f"{type(self).__name__}",
+            "message": f"{str(self)}",
+        }
+        return json.dumps(m)
+
+    @staticmethod
+    def from_error_detail(detail: str) -> Optional["FeastError"]:
+        import importlib
```
tokoko, line 50: These are standard lib libraries, why not import top-level?

tokoko (approved):
lgtm

Issue #4457, opened 2024-08-28
Code Improvement - Marking username and password field of OidcAuthConfig model optional


lokeshrangineni:
This issue is created from the [conversation](https://github.com/feast-dev/feast/pull/4433#issuecomment-2313433513).

The class [OidcAuthConfig](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/permissions/auth_model.py) is having username, password and client_secret as mandatory. OIDC server side code is not using username and password so it is forcing server side code to pass the values. One of the proposal is to create separate model for oidc server and client. We need to evaluate all the util methods if it is going to impact in any other way.

PR #4460, opened 2024-08-28
refactor: Making username and password fields in OidcAuthModel as mandatory onl…

lokeshrangineni:
Added separate model for OidcClient side so that username and password is not required for server side.

class [OidcAuthConfig](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/permissions/auth_model.py) is having username and password as mandatory. OIDC server side code is not using username and password so it is forcing server side code to pass the values. One of the proposal is to create separate model for oidc server and client. We need to evaluate all the util methods if it is going to impact in any other way.

https://github.com/feast-dev/feast/issues/4457

### DIFF
--- a/docs/getting-started/components/authz_manager.md
+++ b/docs/getting-started/components/authz_manager.md
@@ -61,24 +61,36 @@ For example, the access token for a client `app` of a user with `reader` role sh
 }
 ```
 
-An example of OIDC authorization configuration is the following: 
+An example of feast OIDC authorization configuration on the server side is the following: 
 ```yaml
 project: my-project
 auth:
   type: oidc
   client_id: _CLIENT_ID__
-  client_secret: _CLIENT_SECRET__
-  realm: _REALM__  
   auth_discovery_url: _OIDC_SERVER_URL_/realms/master/.well-known/openid-configuration
 ...
 ```
 
-In case of client configuration, the following settings must be added to specify the current user:
+In case of client configuration, the following settings username, password and client_secret must be added to specify the current user:
 ```yaml
 auth:
+  type: oidc
   ...
   username: _USERNAME_
   password: _PASSWORD_
+  client_secret: _CLIENT_SECRET__
+```
+
+Below is an example of feast full OIDC client auth configuration:
+```yaml
+project: my-project
+auth:
+  type: oidc
+  client_id: test_client_id
+  client_secret: test_client_secret
+  username: test_user_name
+  password: test_password
+  auth_discovery_url: http://localhost:8080/realms/master/.well-known/openid-configuration
 ```
 
 ### Kubernetes RBAC Authorization
--- a/sdk/python/feast/permissions/auth_model.py
+++ b/sdk/python/feast/permissions/auth_model.py
@@ -1,4 +1,4 @@
-from typing import Literal, Optional
+from typing import Literal
 
 from feast.repo_config import FeastConfigBaseModel
 
@@ -10,10 +10,12 @@ class AuthConfig(FeastConfigBaseModel):
 class OidcAuthConfig(AuthConfig):
     auth_discovery_url: str
     client_id: str
-    client_secret: Optional[str] = None
+
+
+class OidcClientAuthConfig(OidcAuthConfig):
     username: str
     password: str
-    realm: str = "master"
+    client_secret: str
 
 
 class NoAuthConfig(AuthConfig):
--- a/sdk/python/feast/repo_config.py
+++ b/sdk/python/feast/repo_config.py
@@ -298,6 +298,7 @@ class RepoConfig(FeastBaseModel):
                     self.auth.get("type") == AuthType.OIDC.value
                     and "username" in self.auth
                     and "password" in self.auth
+                    and "client_secret" in self.auth
                 )
                 self._auth = get_auth_config_from_type(
                     "oidc_client" if is_oidc_client else self.auth.get("type")
--- a/sdk/python/feast/repo_config.py
+++ b/sdk/python/feast/repo_config.py
@@ -302,6 +302,10 @@ class RepoConfig(FeastBaseModel):
                 self._auth = get_auth_config_from_type(
                     "oidc_client" if is_oidc_client else self.auth.get("type")
                 )(**self.auth)
+            elif isinstance(self.auth, str):
+                self._auth = get_auth_config_from_type(self.auth)()
+            elif self.auth:
+                self._auth = self.auth
 
         return self._auth
 
--- a/sdk/python/feast/repo_config.py
+++ b/sdk/python/feast/repo_config.py
@@ -87,10 +87,13 @@ FEATURE_SERVER_CONFIG_CLASS_FOR_TYPE = {
     "local": "feast.infra.feature_servers.local_process.config.LocalFeatureServerConfig",
 }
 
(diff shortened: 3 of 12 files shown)


sdk/python/feast/repo_config.py
```diff
@@ -290,12 +293,15 @@ def offline_store(self):
...
         if not self._auth:
+            is_oidc_client = "username" in self.auth and "password" in self.auth
             if isinstance(self.auth, Dict):
-                self._auth = get_auth_config_from_type(self.auth.get("type"))(
-                    **self.auth
-                )
+                self._auth = get_auth_config_from_type(
+                    self.auth.get("type"), is_oidc_client
+                )(**self.auth)
             elif isinstance(self.auth, str):
```
lokeshrangineni, line 301: @tokoko @franciscojavierarceo - I have got this example from online_store config parsing but seems like unnecessary check. Do we really need this check or code block?

sdk/python/feast/permissions/auth_model.py
```diff
@@ -11,9 +11,12 @@ class OidcAuthConfig(AuthConfig):
     auth_discovery_url: str
```
redhatHameed, line 11: rename  `OidcAuthConfig(AuthConfig)` ->  `OidcServerAuthConfig(AuthConfig)` WDYT?

lokeshrangineni: Client is inheriting this class so it may not be appropriate. In case if some other parameters needs then we can add separate class for server.

sdk/python/feast/permissions/auth_model.py
```diff
@@ -11,9 +11,12 @@ class OidcAuthConfig(AuthConfig):
...
     client_id: str
     client_secret: Optional[str] = None
+    realm: str = "master"
+
+
+class OidcClientAuthConfig(OidcAuthConfig):
     username: str
     password: str
```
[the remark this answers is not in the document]
dmartinol: After reviewing the code again, looks like there are other fields that are needed only at the client side.

Clients use these fields to get the token:
```py
        token_request_body = {
            "grant_type": "password",
            "client_id": self.auth_config.client_id,
            "client_secret": self.auth_config.client_secret,
            "username": self.auth_config.username,
            "password": self.auth_config.password,
        }
```
On the other hand, servers use these fields to validate the token:
```py
        self.oidc_discovery_service = OIDCDiscoveryService(
            self._auth_config.auth_discovery_url
        )
...
            client_id = self._auth_config.client_id
```

So, to summarize, servers config is like:
```yaml
auth:
  type: oidc
  auth_discovery_url: http://0.0.0.0:9999/realms/rbac_example/.well-known/openid-configuration
  client_id: app
```

Client config template:
```yaml
  auth_discovery_url: http://0.0.0.0:9999/realms/rbac_example/.well-known/openid-configuration
  client_id: app
  client_secret: Rw4X56WaHODmiYw8duPPTwuH0OoCtVvg
  password: password
  type: oidc
  username: reader
```

`realm` field is no more needed since it is already part of the `auth_discovery_url` (it was needed before when we built the URL programmatically).

docs/getting-started/components/authz_manager.md
```diff
@@ -73,9 +73,10 @@ auth:
 ...
 ```
 
-In case of client configuration, the following settings must be added to specify the current user:
+In case of client configuration, the following settings username and password must be added to specify the current user:
```
dmartinol, line 76: To be reviewed according to the other comment

sdk/python/feast/permissions/auth_model.py
```diff
@@ -11,9 +11,12 @@ class OidcAuthConfig(AuthConfig):
     auth_discovery_url: str
     client_id: str
     client_secret: Optional[str] = None
+    realm: str = "master"
```
dmartinol, line 14: `realm` no more needed

sdk/python/feast/permissions/auth_model.py
```diff
@@ -11,9 +11,12 @@ class OidcAuthConfig(AuthConfig):
     auth_discovery_url: str
     client_id: str
     client_secret: Optional[str] = None
```
dmartinol, line 13: `client_secret` for client config only

sdk/python/tests/unit/infra/scaffolding/test_repo_config.py
```diff
@@ -247,7 +248,36 @@ def test_auth_config():
         expect_error="invalid authentication type=not_valid_auth_type",
     )
 
-    oidc_repo_config = _test_config(
+    oidc_server_repo_config = _test_config(
+        dedent(
+            """
+        project: foo
+        auth:
```
dmartinol, line 255: for servers: only `type`, `client_id` and `auth_discovery_url`

sdk/python/tests/unit/infra/scaffolding/test_repo_config.py
```diff
@@ -247,7 +248,36 @@ def test_auth_config():
...
+    assert oidc_server_repo_config.auth_config.realm == "master"
+    assert (
+        oidc_server_repo_config.auth_config.auth_discovery_url
+        == "http://localhost:8080/realms/master/.well-known/openid-configuration"
+    )
+
+    oidc_client_repo_config = _test_config(
         dedent(
             """
         project: foo
```
dmartinol, line 283: pls remove `realm`

dmartinol (changes requested):
Please update according to suggested changes

sdk/python/feast/permissions/auth_model.py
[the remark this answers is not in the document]
lokeshrangineni: Thanks for all the details. I have updated the PR. PTAL

lokeshrangineni:
> Please update according to suggested changes

I have incorporated all the code review comments. Please take a look again.

redhatHameed (approved):
lgtm

sdk/python/feast/repo_config.py
```diff
@@ -291,11 +294,16 @@ def offline_store(self):
     def auth_config(self):
         if not self._auth:
             if isinstance(self.auth, Dict):
-                self._auth = get_auth_config_from_type(self.auth.get("type"))(
-                    **self.auth
+                is_oidc_client = (
+                    self.auth.get("type") == AuthType.OIDC.value
```
dmartinol, line 298: probably miss to check other client fields after the recent review

sdk/python/tests/unit/infra/scaffolding/test_repo_config.py
```diff
@@ -247,7 +246,32 @@ def test_auth_config():
...
 
-    oidc_repo_config = _test_config(
+    oidc_server_repo_config = _test_config(
+        dedent(
+            """
+        project: foo
+        auth:
+            type: oidc
+            client_id: test_client_id
+            auth_discovery_url: http://localhost:8080/realms/master/.well-known/openid-configuration
```
dmartinol, line 256: for the sake of consistency, `master` should be replaced by `test_client_id` in the URL

dmartinol (approved):
lgtm

sdk/python/feast/repo_config.py
lokeshrangineni: Done.

jeremyary (approved):

PR #4461, opened 2024-08-28
chore: Remove upper bound from fsspec requirement

cburroughs:
In #3922 the fsspec requirement was loosened to allow what was then more recent releases.  Since then Feast has further reduced the number dependencies with upper bounds.  Instead of re-testing each month (when fsspec usually releases) it would now be idiomatic only specify a minimum bound.

The new minimum bound was chosen to match what was last locked in ci, since that is what was tested and we can be less sure older versions also work.


tokoko:
No upper bound seems like trouble. Why doesn't fsspec do semver? Are there any breaking changes we need to be aware of?

cburroughs:
I am unable to find a broader discussion of why `fsspec` [chose](https://github.com/fsspec/filesystem_spec/commit/ada92e503da40673264d45b09f4a25fb32342af7) [calendar versioning](https://calver.org/).

tokoko:
I'm honestly not sure about this. how about we simply bump the upper to latest for now?

cburroughs:
Closing in favor of #4512 then

PR #4438, opened 2024-08-23
build: Set a proper build-system protobuf version

Atry:
# What this PR does / why we need it:

This PR enforces the protobuf version in the build system so that it is compatible with the runtime dependency

# Which issue(s) this PR fixes:
Fixes #4437

# Misc
The runtime version of protobuf is determined at:
https://github.com/feast-dev/feast/blob/0a48f7bb436febb0171c78a559a577eedeff421f/setup.py#L46

I have manually tested this PR with the following command

```
pip install git+https://github.com/Atry/feast.git@build-system-protobuf-version && python -m feast.protos.feast.types.Value_pb2
```

Issue #4437, opened 2024-08-23
ImportError: cannot import name 'runtime_version' from 'google.protobuf'


Atry:
When running Feast installed from `git+https://github.com/feast-dev/feast`, an `ImportError` is raised

## Expected Behavior 

No error

## Current Behavior

```
Traceback (most recent call last):
  File "<frozen runpy>", line 189, in _run_module_as_main
  File "<frozen runpy>", line 112, in _get_module_details
  File "/home/nixos/feast-repository/.venv/lib/python3.11/site-packages/feast/__init__.py", line 4, in <module>
    from feast.infra.offline_stores.bigquery_source import BigQuerySource
  File "/home/nixos/feast-repository/.venv/lib/python3.11/site-packages/feast/infra/offline_stores/bigquery_source.py", line 5, in <module>
    from feast import type_map
  File "/home/nixos/feast-repository/.venv/lib/python3.11/site-packages/feast/type_map.py", line 39, in <module>
    from feast.protos.feast.types.Value_pb2 import (
  File "/home/nixos/feast-repository/.venv/lib/python3.11/site-packages/feast/protos/feast/types/Value_pb2.py", line 9, in <module>
    from google.protobuf import runtime_version as _runtime_version
ImportError: cannot import name 'runtime_version' from 'google.protobuf' (/home/nixos/feast-repository/.venv/lib/python3.11/site-packages/google/protobuf/__init__.py)
```

## Steps to reproduce

```
pip install git+https://github.com/feast-dev/feast && python -m feast.protos.feast.types.Value_pb
```

### Specifications

- Version: `master` branch
- Platform: Linux
- Subsystem: Build System

## Possible Solution

The problem is due to incompatible protobuf versions in build system dependencies and runtime dependencies. We should make the protobuf versions consistant.

PR #4472, opened 2024-08-30
build: Explicit protobuf build version; consistent build/setup deps

cburroughs:
Right now if one downloads `feast-0.40.1-py2.py3-none-any.whl` from PyPi it contains:
```
$ grep 'Protobuf Python Version' feast/protos/feast/registry/RegistryServer_pb2.py
```
Which is outside
```
$ grep 'protobuf<' feast-0.40.1.dist-info/METADATA
Requires-Dist: protobuf<5.0.0,>=4.24.0
```
Leading to runtime errors (#4437).  This was mitigated by #4438.  This change tightens this up further by:
 * Deleting the Makefile command that was trying to do this unsuccessfully.
 * Aligns the setup/build requirements
 * Sets the version of protobuf to match the *minimum* of the range. There is no guarantee that protos generated by `4.X` will work with `4.(X-1)`.

### DIFF
--- a/environment-setup.md
+++ b/environment-setup.md
@@ -13,7 +13,6 @@ pip install cryptography -U
 conda install protobuf
 conda install pymssql
 pip install -e ".[dev]"
-make install-protoc-dependencies PYTHON=3.9
 make install-python-ci-dependencies PYTHON=3.9
 ```
 4. start the docker daemon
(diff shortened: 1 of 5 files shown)


environment-setup.md
tokoko, line 1: Can you please remove the file entirely? up-to-date instructions are in `docs/project/development-guide.md`

tokoko (approved):

PR #4475, opened 2024-09-02
feat: Added Project object to Feast Objects

EXPEbdodla:
1. Based on PR #4425 discussion, Added support for Project object for all registry's in feature definitions. Defaults to project in feature_store.yaml if not specified. Syncs projects content in registry (table) from feast_metadata on initialization. Opens up an option to specify multiple projects in single repository (Needs more work). 
2. Added a registry configuration `purge_feast_metadata` to stop using `feast_metadata` table because list_project_metadata may be used by clients. Instead of deleting that, added a configuration to stop populating table and switch to projects specific methods. 
3. Refactored registry methods `_prepare_registry_for_changes` and `_get_registry_proto` methods in registry.py. Refactored proto() methods in sql.py and snowflake.py. 
4. Corrected RegistryConfig model. __new__() method in registry.py is blocking RegistryConfig inheritance. Removed it and corrected the tests accordingly. Now we can defined configs specific to registry instead of defining all at RegistryConfig class. 

# Misc

### DIFF
--- a/sdk/python/feast/errors.py
+++ b/sdk/python/feast/errors.py
@@ -477,6 +477,16 @@ class PermissionObjectNotFoundException(FeastObjectNotFoundException):
             super().__init__(f"Permission {name} does not exist")
 
 
+class ProjectNotFoundException(Exception):
+    def __init__(self, project):
+        super().__init__(f"Project {project} does not exist in registry")
+
+
+class ProjectObjectNotFoundException(FeastObjectNotFoundException):
+    def __init__(self, name, project=None):
+        super().__init__(f"Project {name} does not exist")
+
+
 class ZeroRowsQueryResult(FeastError):
     def __init__(self, query: str):
         super().__init__(f"This query returned zero rows:\n{query}")
--- a/sdk/python/feast/errors.py
+++ b/sdk/python/feast/errors.py
@@ -477,7 +477,7 @@ class PermissionObjectNotFoundException(FeastObjectNotFoundException):
             super().__init__(f"Permission {name} does not exist")
 
 
-class ProjectNotFoundException(Exception):
+class ProjectNotFoundException(FeastError):
     def __init__(self, project):
         super().__init__(f"Project {project} does not exist in registry")
 
--- a/sdk/python/feast/repo_config.py
+++ b/sdk/python/feast/repo_config.py
@@ -9,6 +9,7 @@ from pydantic import (
     BaseModel,
     ConfigDict,
     Field,
+    StrictBool,
     StrictInt,
     StrictStr,
     ValidationError,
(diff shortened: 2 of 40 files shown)


sdk/python/feast/errors.py
```diff
@@ -477,6 +477,16 @@ def __init__(self, name, project=None):
             super().__init__(f"Permission {name} does not exist")
 
 
+class ProjectNotFoundException(Exception):
```
tokoko, line 480: Why are both of these necessary? what's the difference. btw, All Feast exceptoins should extend `FeastError`.

tokoko:
FYI @dmartinol

sdk/python/feast/errors.py
dmartinol: @tokoko I noticed the same while doing the latest refactory but I didn't want to extend the scope of changes too much. Looks like other duplications are there like `PermissionObjectNotFoundException` and `DataSourceObjectNotFoundException`. Can we remove all of them (in a dedicated ticket maybe)?

EXPEbdodla: I'll extend the FeastError. 
Reasons for having two exceptions: Just followed the current convention. 
ProjectNotFoundException --> Exception thrown when the DELETE calls are made and project is not found. 
ProjectObjectNotFoundException --> Exception thrown when GET calls are made and project is not found. _get_object method passes two args (name, project) when object is not found. I don't want to add another if else logic to filter out project.

tokoko: @dmartinol yeah, this just looks weird. Probably best to defer removal until after this is merged, though.. just to avoid a messy rebase

dmartinol: in case, remember an item in the next Release Notes to remind clients to replace any try/except that may have used those duplicate errors.

sdk/python/feast/repo_config.py
```diff
@@ -129,11 +130,10 @@ class RegistryConfig(FeastBaseModel):
     s3_additional_kwargs: Optional[Dict[str, str]] = None
     """ Dict[str, str]: Extra arguments to pass to boto3 when writing the registry file to S3. """
 
-    sqlalchemy_config_kwargs: Dict[str, Any] = {}
-    """ Dict[str, Any]: Extra arguments to pass to SQLAlchemy.create_engine. """
-
-    cache_mode: StrictStr = "sync"
-    """ str: Cache mode type, Possible options are sync and thread(asynchronous caching using threading library)"""
+    purge_feast_metadata: StrictBool = False
```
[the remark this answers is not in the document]
EXPEbdodla: @tokoko Do you have any recommendation on how we should do these type of changes?  I just don't want to delete project_metadata in one go.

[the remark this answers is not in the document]
tokoko: from an API standpoint, since we are still below 1.0 release mark, the policy is to mark as deprecated and then remove after 2-3 minor releases.

The problem in this case is that the deletion should happen not just from API but rather there needs to be changes made in the existing registry as well. I think we will eventually need a new command for this, something like `feast migrate` (borrowing the name from airflow) which should be run after every version bump?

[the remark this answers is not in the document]
EXPEbdodla: I handled the case for changes to existing registry using `_sync_feast_metadata_to_projects_table` method. Something like `migrate` may help rather than embedding the change in code.

EXPEbdodla:
Fixed Integration tests. Removed [test_remote_online_store.py](https://github.com/feast-dev/feast/pull/4475#diff-beab92b9d6c966ff5f8911a97b744554e64cd9e0c8d66b1a3122a751e2e46967) apply step because Client Store Registry Apply brings back the server and client registry store states to the contents in the repository (Repository missing Permissions. Permissions are applied through store.apply() method). Seems like bug in previous versions which let it work. Did some refactoring to registry so it's showing up as error now.

EXPEbdodla:
@tokoko Please take a look at this PR when you get a chance.

tokoko:
looks good, can you also add usage example in [quickstart](https://github.com/feast-dev/feast/blob/master/docs/getting-started/quickstart.md), just an example of creating a project and specifying some tags.

EXPEbdodla:
> looks good, can you also add usage example in [quickstart](https://github.com/feast-dev/feast/blob/master/docs/getting-started/quickstart.md), just an example of creating a project and specifying some tags.

Updated quick start document and `feast init` functionality to include Project object.

tokoko (approved):
lgtm, thanks

Issue #4344, opened 2024-07-11
Update CI to have a test for main feast dependency before release/deployment


franciscojavierarceo:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
As seen in https://github.com/feast-dev/feast/issues/4293, the PyPi package failed because of a dependency in a feast installation that uses an [extra](https://packaging.python.org/en/latest/tutorials/installing-packages/#installing-setuptools-extras).

```bash
pip install feast[sqlite_vec]
```

This can be added to the existing [build_wheels ci](https://github.com/feast-dev/feast/blob/master/.github/workflows/build_wheels.yml)  by testing a python import. If the import fails, we should fail the build/release.

**Describe alternatives you've considered**
We should have the CI test building the main feast package before release/deployment as well each extra/add-on.

**Additional context**

tokoko:
Not exactly the same, but this can be mostly covered by #4185 I think.

franciscojavierarceo:
We should have the quickstart run before the release as well: https://docs.feast.dev/getting-started/quickstart

dmartinol:
> This can be added to the existing [build_wheels ci](https://github.com/feast-dev/feast/blob/master/.github/workflows/build_wheels.yml) by testing a python import. If the import fails, we should fail the build/release.

What module would you import, how can we know the modules to validate? I tried with the blamed version 0.39.0:
```console
% pip install -q feast==0.39.0
% feast version
Feast SDK Version: "0.39.0"
% python -c "import feast"
% python -c "import sqlite_vect" 
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'sqlite_vect'
```

franciscojavierarceo:
You have a typo in "vec"

franciscojavierarceo:
But this CI validation would ideally run on all of the extras.

dmartinol:
> You have a typo in "vec"
thanks for noticing, anyway it fails (as expected) for version 0.39.0:
```console
% python -c "import sqlite_vec"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'sqlite_vec'
```

> But this CI validation would ideally run on all of the extras.
You mean:
- read all the extras configured in `setup.py`
- for each extra module `m` run `python -c "import <m>" || exit 1`

PR #4478, opened 2024-09-03
chore: Migrate from setup.py to pyproject.toml

tokoko:
- Migrates most of `setup.py` configuration to `pyproject.toml` in line with PEP 518
- Some misc changes to suppress linter issues in tests


environment-setup.md
```diff
@@ -1,22 +0,0 @@
-1. install anaconda, install docker
```
franciscojavierarceo, line 1: Can you remove the reference to this doc in the `development-guide.md` file?

tokoko: done

setup.py
```diff
@@ -22,224 +22,19 @@
 from pathlib import Path
 
 from setuptools import find_packages, setup, Command
```
franciscojavierarceo, line 24: Can we break out this into a separate PR? I'd prefer to merge the pyproject changes in first and then rip out setup, yeah?

tokoko: What are you referring to specifically? the PR moves stuff from one to the other, you can't really have the same things configured in both files at the same time.

franciscojavierarceo: 🤦  yeah nvm

franciscojavierarceo:
It'd probably be ideal to have https://github.com/feast-dev/feast/issues/4344 done before merging this, yeah?

pyproject.toml
```diff
@@ -1,3 +1,189 @@
...
+  "types-protobuf~=3.19.22",
+  "types-python-dateutil",
+  "types-pytz",
+  "types-PyYAML",
+  "types-redis",
+  "types-requests<2.31.0",
+  "types-setuptools",
+  "types-tabulate",
+  "virtualenv<20.24.2",
+  "feast[aws, gcp, redis, k8s, snowflake, spark, postgres, mysql, trino, ge, hbase, cassandra, azure, hazelcast, ibis, grpcio, duckdb, delta, elasticsearch, sqlite_vec, singlestore, opentelemetry]"
```
HaoXuAI, line 176: This can be done by using a "group" for each dependency.

tokoko: Can you provide a link to an example? I googled a bit and I think dependency groups might be exclusive to poetry, not part of a standard.

HaoXuAI: https://github.com/project-ryoma/ryoma/blob/d595b1e3af2caa7e084bef4b04df610f2aa634fb/ryoma_lab/pyproject.toml#L40 I'm doing something like this :), but it's not necessary.

redhatHameed:
@tokoko clarification,I have one task on my to-do list related to this conversation-> https://github.com/feast-dev/feast/pull/4380#discussion_r1711710084 do you think it's still valid after this PR change. Thanks

tokoko:
@redhatHameed yup, that's still valid. this PR only changes where requirements are defined, nothing else.

Issue #4481, opened 2024-09-03
Online Store client does not map application errors


tmihalac:
## Expected Behavior 
In a feast application that has a remote Online Store configured, when a feast client invokes an API on the Online Store that throws an error, the expectation is that the client will raises the same error that the online store threw
## Current Behavior
When an error occurs a response with status code = 500 and error_message=Internal Server Error is returned from the server 
## Steps to reproduce
See Expected Behavior section.
### Specifications

- Version: 0.40.0
- Platform: any
- Subsystem: online store server

## Possible Solution
The online store server can wrap the original exception in an error status message that the remote online store client can catch and translate into the original exception:

PR #4488, opened 2024-09-04
fix: Added Online Store REST client errors handler

tmihalac:
Added a decorator for rest calls to remote online server failures and adds an additional message to the response with the failure status and details.
The proxy client then rebuilds the original error from the status data.

Fixes https://github.com/feast-dev/feast/issues/4481

### DIFF
--- a/sdk/python/feast/errors.py
+++ b/sdk/python/feast/errors.py
@@ -133,7 +133,7 @@ class OnDemandFeatureViewNotFoundException(FeastObjectNotFoundException):
 
 
 class RequestDataNotFoundInEntityDfException(FeastObjectNotFoundException):
-    def __init__(self, feature_name, feature_view_name):
+    def __init__(self, feature_name, feature_view_name=None):
         super().__init__(
             f"Feature {feature_name} not found in the entity dataframe, but required by feature view {feature_view_name}"
         )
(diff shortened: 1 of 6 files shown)


sdk/python/feast/feature_server.py
```diff
@@ -101,187 +105,163 @@ async def lifespan(app: FastAPI):
...
+        request = MaterializeIncrementalRequest(**json.loads(body))
+        for feature_view in request.feature_views:
+            assert_permissions(
+                resource=feature_view, actions=[AuthzedAction.WRITE_ONLINE]
+            )
+        store.materialize_incremental(
+            utils.make_tzaware(parser.parse(request.end_ts)), request.feature_views
+        )
+
+    @app.exception_handler(Exception)
```
franciscojavierarceo, line 250: Very nice, I didn't realize FastAPI had a global exception handler. 

https://fastapi.tiangolo.com/tutorial/handling-errors/#install-custom-exception-handlers

sdk/python/feast/errors.py
```diff
@@ -133,7 +133,7 @@ def __init__(self, name, project=None):
 
 
 class RequestDataNotFoundInEntityDfException(FeastObjectNotFoundException):
```
dmartinol, line 135: Instead of using an optional second parameter, you can update the implementation of `from_error_detail` and replace this line:
```py
                instance = class_reference(message)
```
with:
```py
                instance = class_reference.__new__(class_reference)
```
Sorry for not suggesting this before.

dmartinol (commented):
lgtm apart the comment on `errors` module

tokoko (approved):
lgtm

dmartinol (approved):

PR #4499, opened 2024-09-09
fix: Disable active_timer When registry_ttl_sec is 0

phil-park:
- When using the serve command in cli,
--registry_ttl_sec is defined not to update the registry if it is 0.
However, the active_timer actually runs and updates the registry regardless of the value of registry_ttl_sec.
This PR fixes the above issue and does not run the timer thread if registry_ttl_sec is 0.

# Misc


franciscojavierarceo (approved):
thank you for this!

PR #4500, opened 2024-09-09
fix: Ignore the type check as both functions calls are not belonging to Feast code.

### LINKED ISSUE #4462: mypy argtype failure in ui_server.py with mypy>=1.11
## Expected Behavior 

Type checking should pass!

## Current Behavior

With mypy 1.11

```
feast/ui_server.py:54: error: Argument 1 to "files" has incompatible type "str | None"; expected "str | Module"  [arg-type]
feast/ui_server.py:54: note: Error code "arg-type" not covered by "type: ignore" comment
```

## Steps to reproduce

 * Get your dev environment to the point where you can run `make lint-python`
 * `pip install --upgrade 'mypy>=1.11'`
 * ` `make lint-python` --> the above error

### Specifications

- Version: `main`

## Possible Solution

Presumably mypy has identified a nullness bug, or a bug in itself.

shuchu:
Fix #4462 

If "mypy>=1.11", it will show the following error:
`feast/ui_server.py:54: error: Argument 1 to "files" has incompatible type "str | None"; expected "str | Module"  [arg-type]`

### DIFF
--- a/sdk/python/feast/ui_server.py
+++ b/sdk/python/feast/ui_server.py
@@ -51,7 +51,7 @@ def get_app(
 
     async_refresh()
 
-    ui_dir_ref = importlib_resources.files(__spec__.parent) / "ui/build/"  # type: ignore[name-defined]
+    ui_dir_ref = importlib_resources.files(__spec__.parent) / "ui/build/"  # type: ignore[name-defined, arg-type]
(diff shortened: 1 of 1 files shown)


franciscojavierarceo (approved):
lgtm

Issue #4459, opened 2024-08-28
feat: Provide way to change tag data on Feast objects (e.g. entities & feature views)


tchughesiv:
**Is your feature request related to a problem? Please describe.**
https://github.com/feast-dev/feast/issues/2337

**Describe the solution you'd like**
We should be able to tag objects directly, similar to `kubectl label` for example. With the initial implementation we should be able to do the following:
1. add new labels
2. overwrite existing labels
3. remove labels

For the following object types:
- DataSource
- Entity
- FeatureService
- FeatureView
- OnDemandFeatureView
- SavedDatasetSpec
- StreamFeatureView
- ValidationReference

**Additional context**
I'm thinking something like the following for cli:
```shell
$ feast tag --help
Usage: feast tag [OPTIONS] OBJECT_TYPE NAME

  [Experimental] Tag objects

Options:
  --tags TEXT  Tags (e.g. --tags 'key:value' --tags 'key:value, key:value,
               ...').  [required]
  --overwrite  Overwrite tag(s)
  --help       Show this message and exit.
```
```shell
$ feast tag entity driver --tags release:production
```

tchughesiv:
@tokoko now worries... but couldn't what you just described regarding `feast apply` actually be an argument FOR this kind of `feast tag` capability. currently, how does one update feast object tags without re-instantiating everything? as first mentioned here -
https://github.com/feast-dev/feast/issues/2337

that said, i'll leave the decision to the broader feast community.

franciscojavierarceo:
As mentioned in #2337 
>I'd expect feast apply to make those changes in the registry, but not for example tear down / re-instantiate an online store

I think this approach is reasonable. The alternative approach to get the same result is to modify the apply to have tag specific logic in the apply for tags. 

This is a good PR with a lot of changes incorporated to support this, so I'm inclined to include this as is. We can note revisiting it in the future if the imperative use causes problems.

tchughesiv:
@tokoko @franciscojavierarceo i'm completely fine with pivoting here and implementing whatever is best. what would be a better design for this?
 - we could change the command `feast tag` to mimic `feast apply` but instead only tag objects?
 - we could modify `feast apply` by adding a `--tags-only` or `--tags` option to indicate only update tags.
 - something else?

tokoko:
[...]

PR #4503, opened 2024-09-09
feat: Allow tag updates to Feast objects

tchughesiv:
With this change we will be able to tag feast objects directly, similar to `kubectl label`. New registry endpoints were added, in addition to a new `feast tag` cli command. For this initial implementation users can -
1. Add new tags
2. Overwrite existing tags
3. Remove tags

```bash
$ feast tag --help
Usage: feast tag [OPTIONS] OBJECT_TYPE NAME TAGS...

  [Experimental] Tag objects

  *  A tag key and value must begin with a letter or number, and may contain
  letters, numbers, hyphens, dots, and underscores, up to 63 characters each.

  *  Optionally, the key can begin with a DNS subdomain prefix and a single
  '/', like example.com/my-app. The 'feast.dev/' prefix is reserved.

  Examples:

      # Update data source 'foo' with the tag 'test' and the value 'true'.

      feast tag data-source foo test:true

      # Update feature view 'foo' by removing a tag named 'bar' if it exists.
      # Does not require the --overwrite flag.

      feast tag feature-view foo bar-

Options:
  --overwrite  If true, allow tags to be overwritten, otherwise reject tag
               updates that overwrite existing tags.
  --help       Show this message and exit.
```

The following object types are supported for tagging:
- DataSource
- Entity
- FeatureService
- FeatureView
- OnDemandFeatureView
- Permission
- Project
- SavedDataset
- StreamFeatureView
- ValidationReference

Fixes https://github.com/feast-dev/feast/issues/4459

# Misc
Of note, this change will add regex validation for tag keys and values, which is nearly identical to kubernetes labels. In some environments, existing feast objects with "invalid" tags will break future tagging requests due to validation errors. This can be easily resolved by replacing the problematic tags. For example -
```bash
# here we see a pre-existing feature view with an invalid tag
$ feast feature-views describe driver_hourly_stats | yq .spec.tags
team?: driver_performance

# any attempts to add a new tag will fail due to the pre-existing invalid tag
$ feast tag feature-view driver_hourly_stats new:tag
Invalid tag key: 'team?': name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '^(?=.{1,63}$)(?!feast\.dev/)(?:(?:[a-z0-9]([-a-z0-9]*[a-z0-9])?\.)*[a-z0-9]([-a-z0-9]*[a-z0-9])?\/)?[a-z0-9A-Z]([a-z0-9A-Z._-]*[a-z0-9A-Z])?$'

# this can be resolved with a single command that will,
#  1) remove the invalid tag
#  2) replace it with a similar valid tag
#  3) add the new tag.
$ feast tag feature-view driver_hourly_stats 'team?-' team:driver_performance new:tag
feature-view 'driver_hourly_stats' tagged

# here we view the fixed feature view tags
$ feast feature-views describe driver_hourly_stats | yq .spec.tags
team: driver_performance
new: tag
```

### DIFF
--- a/docs/reference/feast-cli-commands.md
+++ b/docs/reference/feast-cli-commands.md
@@ -12,32 +12,61 @@ Usage: feast [OPTIONS] COMMAND [ARGS]...
   For more information, see our public docs at https://docs.feast.dev/
 
 Options:
-  -c, --chdir TEXT  Switch to a different feature repository directory before
-                    executing the given subcommand.
-
-  --help            Show this message and exit.
+  -c, --chdir TEXT               Switch to a different feature repository
+                                 directory before executing the given
+                                 subcommand.
+  --log-level TEXT               The logging level. One of DEBUG, INFO,
+                                 WARNING, ERROR, and CRITICAL (case-
+                                 insensitive).
+  -f, --feature-store-yaml TEXT  Override the directory where the CLI should
+                                 look for the feature_store.yaml file.
+  --help                         Show this message and exit.
 
 Commands:
   apply                    Create or update a feature store deployment
+  data-sources             Access data sources
+  endpoint                 Display feature server endpoints
   entities                 Access entities
+  feature-services         Access feature services
   feature-views            Access feature views
   init                     Create a new Feast repository
+  listen                   Start a gRPC feature server to ingest...
   materialize              Run a (non-incremental) materialization job to...
-  materialize-incremental  Run an incremental materialization job to ingest...
+  materialize-incremental  Run an incremental materialization job to...
+  on-demand-feature-views  [Experimental] Access on demand feature views
   permissions              Access permissions
+  plan                     Create or update a feature store deployment
+  projects                 Access projects
   registry-dump            Print contents of the metadata registry
+  saved-datasets           [Experimental] Access saved datasets
+  serve                    Start a feature server locally on a given port.
+  serve_offline            Start a remote server locally on a given host,...
+  serve_registry           Start a registry server locally on a given port.
+  serve_transformations    [Experimental] Start a feature consumption...
+  stream-feature-views     [Experimental] Access stream feature views
+  tag                      [Experimental] Tag objects
   teardown                 Tear down deployed feature store infrastructure
+  ui                       Shows the Feast UI over the current directory
+  validate                 Perform validation of logged features...
+  validation-references    [Experimental] Access validation references
   version                  Display Feast SDK version
 ```
 
 ## Global Options
 
-The Feast CLI provides one global top-level option that can be used with other commands
+The Feast CLI provides three global top-level options that can be used with other commands
 
 **chdir \(-c, --chdir\)**
 
 This command allows users to run Feast CLI commands in a different folder from the current working directory.
 
+**log-level \(--log-level\)**
+
+This command allows users to change the logging level.
+
+**feature-store-yaml \(-f, --feature-store-yaml\)**
+
+This command allows users to override the directory where the CLI should look for the `feature_store.yaml` file.
 ```text
 feast -c path/to/my/feature/repo apply
 ```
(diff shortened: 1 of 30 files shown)


sdk/python/feast/errors.py
```diff
@@ -509,3 +509,25 @@ def grpc_status_code(self) -> GrpcStatusCode:
 
     def http_status_code(self) -> int:
         return HttpStatusCode.HTTP_403_FORBIDDEN
+
+
+class TagKeyAlreadyExists(Exception):
```
dmartinol, line 514: All Feast errors should extend `FeastError` to be properly transferred from a server to a client proxy

docs/reference/feast-cli-commands.md
```diff
@@ -292,6 +321,47 @@ reader             driver_hourly_stats_fresh   FeatureView      DESCRIBE
...
+  '/', like example.com/my-app. The 'feast.dev/' prefix is reserved.
+
+  Examples:
+
+      # Update data source 'foo' with the tag 'test' and the value 'true'.
+
+      feast tag data-source foo test:true
+
+      # Update feature view 'foo' by removing a tag named 'bar' if it exists.
+      Does not require the --overwrite flag.
```
dmartinol, line 356: Missed a starting "#"

docs/reference/feast-cli-commands.md
```diff
@@ -292,6 +321,47 @@ reader             driver_hourly_stats_fresh   FeatureView      DESCRIBE
...
+
+      feast tag feature-view foo bar-
+
+Options:
+  --overwrite  If true, allow tags to be overwritten, otherwise reject tag
+               updates that overwrite existing tags.
```
lokeshrangineni, line 362: Can we add the default value in the description? I presume it is False.

tchughesiv: hmm... it should be showing already, but `show_default=True` isn't being honored here for some reason. i'll look into it.

sdk/python/feast/infra/registry/base_registry.py
```diff
@@ -99,6 +99,28 @@ def get_entity(self, name: str, project: str, allow_cache: bool = False) -> Enti
         """
         raise NotImplementedError
 
+    @abstractmethod
+    def tag_entity(
```
lokeshrangineni, line 103: Just a thought, I see the default implementation is duplicated in all registry implementations. Can we mark it is non abstract method and override wherever is necessary. I don't know if we are breaking any code guidelines from feast with this type of implementation.

docs/reference/feast-cli-commands.md
franciscojavierarceo: Actually since this is a doc string, those # signs aren't needed at all.

docs/reference/feast-cli-commands.md
franciscojavierarceo: Can probably just add the default as a comment

docs/reference/feast-cli-commands.md
dmartinol: I think this comes from `kubectl label --help`, which has comments instead:
```
Examples:
...
  # Update pod 'foo' by removing a label named 'bar' if it exists
  # Does not require the --overwrite flag
  kubectl label pods foo bar-
```

sdk/python/feast/infra/registry/base_registry.py
dmartinol: I might be wrong, but I fear there are different hierarchies in the actual implementations preventing to have a common definition. @tokoko ?

tokoko: I think it could work regardless, but it's a moot point, we agreed to close the PR, see #4459.

Issue #4504, opened 2024-09-09
Refactor `auth_client_manager_factory.py` in function `get_auth_client_manager`


tmihalac:
**Is your feature request related to a problem? Please describe.**
The code in `auth_client_manager_factory.py` in function `get_auth_client_manager` is cumbersome and can be simplified.

**Describe the solution you'd like**
Can be simplified by introducing `IntraCommAuthClientManager`

PR #4505, opened 2024-09-09
fix: Refactor auth_client_manager_factory.py in function get_auth_client_m…

tmihalac:
…anager

Refactor auth_client_manager_factory.py` in function get_auth_client_manager

Fixes https://github.com/feast-dev/feast/issues/4504


sdk/python/tests/unit/permissions/auth/client/test_authentication_client_manager_factory.py
```diff
@@ -0,0 +1,70 @@
...
+                "type": "oidc",
+                "client_id": "feast-integration-client",
+                "client_secret": "feast-integration-client-secret",
+                "username": "reader_writer",
+                "password": "password",
+                "auth_discovery_url": "KEYCLOAK_URL_PLACE_HOLDER/realms/master/.well-known/openid-configuration",
+            }
+        ),
+    ],
+)
```
lokeshrangineni, line 39: why can't we use the default auth_config under root directory of `conftest.py`?

tmihalac: To create the AuthConfig object at the beginning of the test we needed an ugly if else if else

lokeshrangineni: I see we are using if else to assert the different conditions based on different auth_config. Not sure if i misunderstood any thing.

franciscojavierarceo (approved):

Issue #4280, opened 2024-06-14
Bug: Regression in BigQuery offline store caused by newer pydantic versions


galen-ft:
## Expected Behavior 

Calls to `store = FeatureStore(repo_path=repo_path)` for a BigQuery offline store should just work.

## Current Behavior

Running `store = FeatureStore(repo_path=repo_path)` for a BigQuery offline store causes:
```shell
Traceback (most recent call last):
  File "<string>", line 1, in <module>
TypeError: 'pydantic_core._pydantic_core.ValidationInfo' object is not subscriptable
```

This happens on line 108 in [bigquery.py#L108](https://github.com/feast-dev/feast/blob/v0.38.0/sdk/python/feast/infra/offline_stores/bigquery.py#L108).

### Likely cause

`pydantic>=2.0.0` changed the  `ValidationInfo` object.

The [setup.py for feast >= v0.36.0](https://github.com/feast-dev/feast/blob/v0.36.0/setup.py#L56) requires that `pydantic>=2.0.0`.

Last known successful configuration:
- pydantic==1.10.16
- feast v.0.35.0 - the [setup.py for feast v0.35.0](https://github.com/feast-dev/feast/blob/v0.35.0/setup.py#L64) allows for `pydantic>=1,<2`.

## Steps to reproduce

Example `feature_store.yaml`:
```yaml
project: my_feature_repo
registry: gs://.../registry.db
# The provider AWS used for the online store.
# Mixing AWS and GCP should be okay provided that offline store type is specified.
provider: aws
offline_store:
    type: bigquery
    billing_project_id: some-project-id
    dataset: latest
    gcs_staging_location: gs://...
    location: EU
    project_id: some-project-id
online_store:
  type: redis
  connection_string: ...

entity_key_serialization_version: 2
```

```python
import os
# Credentials for GCP
os.environ["GOOGLE_APPLICATION_CREDENTIALS"]="GCP_keyfile.json"
from feast import FeatureStore, RepoConfig


if __name__ == "__main__":
    # repo_path must contain the GCP_keyfile.json and feature_store.yaml files
    repo_path = "./"
    assert os.path.exists(repo_path)
    # Create a feature store object
    store = FeatureStore(repo_path=repo_path) # <--- Error occurs here
```

### Specifications

- Version: feast v0.38.0
- Platform: x86-64 Ubuntu 22.04.3 LTS as well as arm64 Mac OS Sonoma 14.4.1
- Subsystem:

## Possible Solution

[...]

PR #4509, opened 2024-09-10
fix: Fixes validator field access for 'project_id' in BigQuery offline Store

jgesche:

Fixes a regression in BigQuery offline store due to newer pydantic version

Fixes https://github.com/feast-dev/feast/issues/4280


HaoXuAI (approved):
LGTM

PR #4513, opened 2024-09-11
fix: Removed temporary directory creation and replaced with feature_repo

redhatHameed:

As Feast takes the feature-store YAMLas Base64 environment variable, and stores it in a [temp directory](https://github.com/feast-dev/feast/blob/4a6b663f80bc91d6de35ed2ec428d34811d17a18/sdk/python/feast/repo_operations.py#L375-L383). Since it uses that temporary path and set feature store, and not way to to run `feast apply` on the remote server for example incase file base datasource used.

# Misc

### DIFF
--- a/sdk/python/feast/infra/materialization/aws_lambda/app.py
+++ b/sdk/python/feast/infra/materialization/aws_lambda/app.py
@@ -1,7 +1,6 @@
 import base64
 import json
 import sys
-import tempfile
 import traceback
 from pathlib import Path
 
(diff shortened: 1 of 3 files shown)


sdk/python/feast/infra/materialization/aws_lambda/app.py
```diff
@@ -27,8 +26,10 @@ def handler(event, context):
 
         config_bytes = base64.b64decode(config_base64)
 
-        # Create a new unique directory for writing feature_store.yaml
-        repo_path = Path(tempfile.mkdtemp())
+        # Create a directory for writing feature_store.yaml
```
franciscojavierarceo, line 29: can we instead change this to make this change only when using remote?

redhatHameed:
@franciscojavierarceo closing this PR I found the solution for that instead of feature-store YAMLas Base64 we can use path of feature store file  `feast -c feature_repo serve_registry` that will fix the issue. I will document these steps and will open another PR

PR #4519, opened 2024-09-13
fix: Removed the k8s dependency from required dependencies

redhatHameed:

The k8s dependency already added into extra dependency section as it's not mandatory so removed from required dependencies.

# Misc


tokoko:
@redhatHameed move this [import](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/permissions/server/utils.py#L14) inside a function so that it's invoked only if k8s is set as an auth type.

redhatHameed:
> @redhatHameed move this [import](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/permissions/server/utils.py#L14) inside a function so that it's invoked only if k8s is set as an auth type.

@tokoko thanks for direction, updated accordingly and all tests are passing now.

tokoko:
This should be good. Can you update the lock files as well?

redhatHameed:
> This should be good. Can you update the lock files as well?

this is the make command for that right `lock-python-dependencies-all`

tokoko:
yup

redhatHameed:
> yup

Thanks updated, please have look

tokoko (approved):
lgtm

Issue #4376, opened 2024-08-02
Add write support for On Demand Feature Views


franciscojavierarceo:
# The problem 

As discussed in https://github.com/feast-dev/feast/issues/4365, we should add the ability to write an On Demand Feature View (ODFV) to store the output of the calculation. 

# The solution
The ideal solution would require a boolean to the ODFV decorator as metadata to control the write behavior and another boolean in the `get_online_features` method to allow for users to force features to be recomputed.

The writes would be done by calling `push()` or `write_to_online_store()` with the underlying raw data (inputs) and storing the transformed feature values (outputs) into the online store. The ODFV would be called before executing the writes.

The change for the ODFV definition would be:
```python
@on_demand_feature_view(
    sources=[
        driver_hourly_stats_view,
        input_request
    ],
    schema=[
        Field(name='conv_rate_plus_val1_python', dtype=Float64),
        Field(name='conv_rate_plus_val2_python', dtype=Float64),
    ],
    mode="python",
    write_to_online_store=True,                #  THIS IS THE FIRST NEW BOOLEAN
)
def transformed_conv_rate_python(inputs: Dict[str, Any]) -> Dict[str, Any]:
    output: Dict[str, Any] = {
        "conv_rate_plus_val1_python": [
            conv_rate + val_to_add
            for conv_rate, val_to_add in zip(
                inputs["conv_rate"], inputs["val_to_add"]
            )
        ],
        "conv_rate_plus_val2_python": [
            conv_rate + val_to_add
            for conv_rate, val_to_add in zip(
                inputs["conv_rate"], inputs["val_to_add_2"]
            )
        ]
    }
    return output
```

And the change for the `get_online_features` call would be:
```python
entity_rows = [
    {
        "driver_id": 1001,
        "val_to_add": 1,
        "val_to_add_2": 2,
    }
]

online_response = store.get_online_features(
    entity_rows=entity_rows,
    features=[
        "driver_hourly_stats:conv_rate",
        "driver_hourly_stats:acc_rate",
        "transformed_conv_rate_python:conv_rate_plus_val1_python",
        "transformed_conv_rate_python:conv_rate_plus_val2_python",
    ],
    force_compute=False,                         #  THIS IS THE SECOND NEW BOOLEAN
).to_dict()
```

[...]

PR #4530, opened 2024-09-17
feat: Updating FeatureViewProjection and OnDemandFeatureView to add batch_source and entities

franciscojavierarceo:

This PR addresses some gaps in the `FeatureViewProjection`, `BaseFeatureView`, and `OnDemandFeatureview` classes. In particular, this PR aims to allow the data source to be included in the `BaseFeatureView` and all data sources in a `FeatureViewProjection`. I make the assumption that a single `BaseFeatureView` can only have one datasource, which is a reasonable assumption.

This structure allows users to express an explicit dependency graph with the `BaseFeatureView` being ultimately tied to one single, foundational data source which I believe is the *right* pattern.

Along with storing the datasources, we also store relevant batch_source data in the Projection. 

More specificially, this PR update the following:
1. `FeatureViewProjection` to include the underlying data sources for `OnDemandFeatureView`. This allows for much richer lineage when constructing metadata about an `OnDemandFeatureView`s and its data sources.
2. `OnDemandFeatureView` to optionally support `entities`, which will be required to write an `OnDemandFeatureView` to the online store
3. `BaseFeatureView` to include sources to optionally include `bath_source`
4. `OnDemandFeatureView` to include `write_to_online_store`, a boolean to be used in a follow-up PR (not used in this PR)

This is a step along the path of solving https://github.com/feast-dev/feast/issues/4376

# Misc
N/A

### DIFF
--- a/sdk/python/feast/infra/materialization/snowflake_engine.py
+++ b/sdk/python/feast/infra/materialization/snowflake_engine.py
@@ -307,7 +307,6 @@ class SnowflakeMaterializationEngine(BatchMaterializationEngine):
                 """
 
                 with GetSnowflakeConnection(self.repo_config.offline_store) as conn:
-                    print(query)
                     entities_to_write = conn.cursor().execute(query).fetchall()[0][0]
 
             if feature_view.batch_source.field_mapping is not None:


sdk/python/feast/feature_view_projection.py
```diff
@@ -57,20 +80,57 @@ def from_proto(proto: FeatureViewProjectionProto):
...
-            features=base_feature_view.features,
-            desired_features=[],
-        )
+        if getattr(base_feature_view, "batch_source", None):
+            return FeatureViewProjection(
+                name=base_feature_view.name,
+                name_alias=None,
+                features=base_feature_view.features,
+                desired_features=[],
+                timestamp_field=base_feature_view.batch_source.created_timestamp_column  # type:ignore[attr-defined]
```
franciscojavierarceo, line 119: This is admittedly a hack and I don't like it but this should be refactored in 1.0.0

My overall learning from this is that FeatureViews should be really derived from the same object with the same class parameters and make them optionally instantiated. 

This should be described more thoroughly for 1.0.0

tokoko:
A few questions from me:

> I make the assumption that a single `BaseFeatureView` can only have one datasource, which is a reasonable assumption.

I'm not sure this will hold true actually... BatchFeatureViews or whatever we will call them will need to rely on multiple sources for example.

> `FeatureViewProjection` to include the underlying data sources for `OnDemandFeatureView`. This allows for much richer lineage when constructing metadata about an `OnDemandFeatureView`s and its data sources.

Can you elaborate on why we need this? My understanding is that `FeatureViewProjection` objects hold **additional** info (request-specific modifications) about a specific FeatureView, what's the point of copying some of those field over here when they can already be accessed from FeatureView objects anyway. Feels like unnecessary duplication to me.

> `OnDemandFeatureView` to optionally support `entities`, which will be required to write an `OnDemandFeatureView` to the online store

I sort of understand the rationale behind this, but wouldn't it better for us to try to keep OnDemandFeatureViews entity-agnostic? I guess I have concerns similar to FeatureViewProjection ones above. odfv already has dependency on feature views and their entities as a result. Relisting them here feels redundant and would break a single of version of truth. What is a user applies an odfv whose entities do not match the entities of relevant feature views?

My idea about a odfv caching was not to make them retrievable by any set of entities, but rather by some sort of key that would hold information about all the input fields.

franciscojavierarceo:
ODFVs today can behave the old way by not using the optional forthcoming optional parameter that configures writes. The benefit of storing the data here for batch will be the backfilling and seeing the full lineage of the data sources that's used to compute the ODFV.

In a simple way:

ODFV = f(Data source 1, data source 2, feature view 3)

In practice, combinations like this may occur and we get this flexibility and lineage. Having the write option  basically acts as a precomputation option for faster retrieval.

franciscojavierarceo:
> sort of understand the rationale behind this, but wouldn't it better for us to try to keep OnDemandFeatureViews entity-agnostic? I guess I have concerns similar to FeatureViewProjection ones above. odfv already has dependency on feature views and their entities as a result. 

This is again, important for persistence and backfilling.

>Relisting them here feels redundant and would break a single of version of truth. What if a user applies an odfv whose entities do not match the entities of relevant feature views?

Not redundant as things can change at different points in time.

HaoXuAI:
Instead of directly updating the `BaseFeatureView` with DataSource, is it possible to change `OndemandFeatureView` in inherit from `FeatureView` which has the `source` attribute already?
I'm not saying this is a better idea since all types of concrete feature views have the source attribute. Just in case updating the BaseFeatureView might cause breaking changes

franciscojavierarceo:
It won't be a breaking change as it's optional. I do think we should revisit all of these for free 1.0.0 release and make StreamFeatureView, FeatureView, and OnDemandFeatureView all inherit from BaseFeatureView with a fixed set of parameters and make things much cleaner. They have lots of duplicate code all over the place and it's very unnecessary...but I'd rather do that full cleanup once we have this functionality done. Wdyt?

HaoXuAI (approved):
LGTM!

tokoko:
Sorry, couldn't follow up sooner.

> Not redundant as things can change at different points in time.

I think that's precisely my point. we are allowing essentially same concepts to be defined twice in unrelated places that can easily diverge from one another.

For example, after this PR what's the difference between `entities` field of an odfvs vs retrieving the list of entities from underlying feature views:

```
odfv = store.get_on_demand_feature_view('transformed_conv_rate')

entities_1 = odfvs.entities
entities_2 = [store.registry.get_any_feature_view(fvp.name).entities for fvp in odfv.source_feature_view_projections]
```

Is there any scenario where it makes sense for `entities_1` and `entities_2` not to be exactly the same? If not, why do we need them defined twice?

P.S. the same applied to the `FeatureViewProjection`. Those fields can easily be retrieved from the underlying `FeatureView` imho.

PR #4531, opened 2024-09-17
fix: Logger settings for feature servers and updated logger for permission flow

redhatHameed:

Added Logger settings for feature servers and updated logger for permission flow

# Misc

### DIFF
--- a/sdk/python/feast/permissions/server/arrow.py
+++ b/sdk/python/feast/permissions/server/arrow.py
@@ -17,7 +17,6 @@ from feast.permissions.security_manager import get_security_manager
 from feast.permissions.user import User
 
 logger = logging.getLogger(__name__)
-logger.setLevel(logging.INFO)
 
 
 class AuthorizationMiddlewareFactory(fl.ServerMiddlewareFactory):
@@ -49,7 +48,7 @@ class AuthorizationMiddleware(fl.ServerMiddleware):
 
     def call_completed(self, exception):
         if exception:
-            logger.error(
+            logger.exception(
                 f"{AuthorizationMiddleware.__name__} encountered an exception: {exception}"
             )
 
@@ -78,7 +77,7 @@ def inject_user_details(context: ServerCallContext):
     if sm is not None:
         auth_middleware = cast(AuthorizationMiddleware, context.get_middleware("auth"))
         current_user = asyncio.run(auth_middleware.extract_user())
-        logger.info(f"User extracted: {current_user}")
+        logger.debug(f"User extracted: {current_user}")
 
         sm.set_current_user(current_user)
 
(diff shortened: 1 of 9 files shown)


dmartinol:
Shoudln't we also remove all the various call to `logger.setLevel` that we recently added to the RBAC code?
Thanks!

redhatHameed:
> Shoudln't we also remove all the various call to `logger.setLevel` that we recently added to the RBAC code? Thanks!

done

tokoko (approved):
lgtm, I'll wait a bit for other comments and merge

sdk/python/feast/permissions/server/arrow.py
```diff
@@ -49,7 +48,9 @@ def __init__(self, access_token: str, *args, **kwargs):
 
     def call_completed(self, exception):
         if exception:
-            print(f"{AuthorizationMiddleware.__name__} received {exception}")
+            logger.error(
+                f"{AuthorizationMiddleware.__name__} encountered an exception: {exception}"
```
lokeshrangineni, line 52: The better way to log exception is `logging.exception` I m not sure what it logs if we do the `{exception}`.

sdk/python/feast/permissions/server/arrow.py
```diff
@@ -69,14 +70,14 @@ def inject_user_details(context: ServerCallContext):
...
-        logger.info("No `auth` middleware.")
+        logger.warning("No `auth` middleware.")
         return
 
     sm = get_security_manager()
     if sm is not None:
         auth_middleware = cast(AuthorizationMiddleware, context.get_middleware("auth"))
         current_user = asyncio.run(auth_middleware.extract_user())
-        print(f"extracted user: {current_user}")
+        logger.info(f"User extracted: {current_user}")
```
lokeshrangineni, line 80: Do we really need to log every user using info?

sdk/python/feast/permissions/server/grpc.py
```diff
@@ -22,11 +21,13 @@ def intercept_service(self, continuation, handler_call_details):
                 metadata=dict(handler_call_details.invocation_metadata)
             )
 
-            print(f"Fetching user for token: {len(access_token)}")
+            logger.info(
+                f"Fetching user details for token of length: {len(access_token)}"
```
lokeshrangineni, line 25: my 2 cents - may be debug log here is more appropriate.

dmartinol (approved):

sdk/python/feast/permissions/server/arrow.py
redhatHameed: I think we don't set as a debug incase if required

sdk/python/feast/permissions/server/grpc.py
redhatHameed: set as debug

sdk/python/feast/permissions/server/arrow.py
redhatHameed: set as logging.exception. These will arrow flight sever related internal exceptions.

PR #4535, opened 2024-09-18
fix: Removed protobuf as a required dependency

EXPEbdodla:
1. Resolves the error when generating sdist. Current error is `pkg_resources.ContextualVersionConflict: (protobuf 4.24.0 (/feast/.eggs/protobuf-4.24.0-py3.8-macosx-14.4-arm64.egg), Requirement.parse('protobuf<6.0dev,>=5.26.1'), {'grpcio-tools'})`
2. Moved grpcio-tools as DEV dependency

# Misc

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,10 +58,7 @@ lock-python-ci-dependencies:
 package-protos:
 	cp -r ${ROOT_DIR}/protos ${ROOT_DIR}/sdk/python/feast/protos
 
-install-protoc-dependencies:
-	pip install "grpcio-tools>=1.56.2,<2" "mypy-protobuf>=3.1"
-
-compile-protos-python: install-protoc-dependencies
+compile-protos-python:
 	python setup.py build_python_protos --inplace
 
 install-python:
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,10 @@ lock-python-ci-dependencies:
 package-protos:
 	cp -r ${ROOT_DIR}/protos ${ROOT_DIR}/sdk/python/feast/protos
 
-compile-protos-python:
+install-protoc-dependencies:
+	pip install "grpcio-tools>=1.56.2,<2" "mypy-protobuf>=3.1"
+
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -58,7 +58,10 @@ lock-python-ci-dependencies:
 package-protos:
 	cp -r ${ROOT_DIR}/protos ${ROOT_DIR}/sdk/python/feast/protos
 
-compile-protos-python:
+install-protoc-dependencies:
```
tokoko, line 61: Do we need this? Aren't they required and dev dependencies already anyway? I don't like this as we will end up mixing uv and pip in ci if we merge like this

EXPEbdodla: If users have already installed ci dependencies, it's of no use.

sdk/python/requirements/py3.10-requirements.txt
```diff
@@ -89,9 +89,7 @@ partd==1.4.2
 prometheus-client==0.20.0
     # via feast (setup.py)
 protobuf==4.25.4
-    # via
-    #   feast (setup.py)
-    #   mypy-protobuf
+    # via mypy-protobuf
```
tokoko, line 92: I still think we should also investigate whether `mypy-protobuf` should also be relegated as a dev dependency or not.. but I'm fine with postponing that as well

EXPEbdodla: We can postpone that.

tokoko (approved):
lgtm

cburroughs:
With the fix from #4472 removed, what is the current mechanism that ensures that the protobuf files are generated with the minimum supported protobuf version, and not the max?

franciscojavierarceo:
@cburroughs my recommendation would be to have integration tests in place validating the serialization and deserialization of feature views. I would also recommend having them in environments before deploying to production to ensure upgrading versions don't break serialization into your online store.

PR #4546, opened 2024-09-20
chore: Commit generated python proto files

### LINKED ISSUE #4541: Commit generated proto files to source control
Currently we don't check generated python proto files in source control, instead they're generated during build by setuptools. We've had problems lately because of this as this process means we need to manage additional "setup requirements" and protobuf versions don't always play well with one another.

We also distribute python library as a wheel with generated proto files inside, so keeping these files out of source control doesn't give us any benefit in terms of client system compatibility. protobuf version used is still determined on our ci servers during build-time.

Another major con of running protoc during build-time is that we won't be able to get rid of `setup.py` file even after a migration to `pyproject.toml`.

I propose to instead add a new target in `Makefile` that will generate proto files and will be independent of setuptools. The generated files will be commited to the source control. We can also add a test in ci that will check that proto definitions and generated code is consistent with one another.

tokoko:
Fixes #4541


.gitignore
```diff
@@ -185,7 +185,7 @@ dmypy.json
 
 # Protos
 sdk/python/docs/html
-sdk/python/feast/protos/
+#sdk/python/feast/protos/
```
franciscojavierarceo, line 188: Can just remove

franciscojavierarceo (commented):
This is a big one

.gitignore
tokoko: done

EXPEbdodla:
Support ends for 4.25.* by end of 2025 Q1. https://protobuf.dev/support/version-support/#python

Conda now has 5.* version packages. https://github.com/conda-forge/protobuf-feedstock/pulls?q=is%3Apr+is%3Aclosed 

https://anaconda.org/conda-forge/protobuf/files

Can we try with Version 5 now and see what happens now? We may need lock Protobuf dependency to a version which conda supports. We can do that in a next PR. So we get benefits of Protobuf v5.

tokoko:
@EXPEbdodla That's a perfect timing :laughing: I'd still prefer to defer it to another PR, because we might have that runtime version warning problem to solve somehow with protobuf 5.

.gitignore
```diff
@@ -185,7 +185,6 @@ dmypy.json
 
 # Protos
 sdk/python/docs/html
-sdk/python/feast/protos/
 sdk/go/protos/
```
franciscojavierarceo, line 188: Can probably removes all of the protos from the git ignore, yeah?

franciscojavierarceo (commented):
We should ask @achals @adchia or @woop  if there was a reason we didn't include these protobuf files to begin with before merging this.

franciscojavierarceo (approved):
Think we can update the git ignore to remove all references to protos but this lgtm

.gitignore
tokoko: I think those are golang ones, I'll look into it later, don't want to push something extra inadvertently.

tokoko:
Merged, let's see what this buys us :laughing: we can always revert back...

franciscojavierarceo:
Can you change this to feature instead of chore? Want to make sure it's included loudly in the release notes 👍

PR #4550, opened 2024-09-20
fix: Deleting data from feast_metadata when we delete project

EXPEbdodla:
Deletes data from feast_metadata table when we delete project

# Misc


HaoXuAI (approved):
LGTM!

Issue #3785, opened 2023-10-09
Get rid of duplicate peer dependencies at Feast UI


usersavchenko:
Currently there is problem, Feast UI Obliges us to download a pile of various peer dependencies. They are duplicates of "dependencies" as you can see from my screenshot. What the use of this approach ? Can we release users from downloading all of these ones ?

![image](https://github.com/feast-dev/feast/assets/76808191/68a8a1da-e8e8-4be2-b13c-be44f6734c33)

lokeshrangineni:
@usersavchenko - i don't consider myself a UI developer so please take my comment with grain of salt. 

Feast UI can be used as a standalone project or a plugin in an existing react project. AFAIK peer dependencies are downloaded if the feast is used as plugin.

If you are using feast as a standalone project then only dependencies list will be downloaded. 

The expectation is only either `peerDepedencies` or `dependencies` used depending on how you are using feast UI components. 

Could you check what do you get with the command `npm list`? see if the dependencies are downloaded twice.

My comment is based on the below stackoverflow [discussion](https://stackoverflow.com/questions/65487385/should-i-duplicate-peerdependencies-in-dependencies-field-of-package-json).

peruukki:
I was also wondering about the large number of peer dependencies, thanks for the clarification.

Maybe the `peerDependencies` list could still be trimmed down a bit. They are meant for framework-like dependencies that should be managed by the "host" app, and usually there needs to be a single instance of them. `dependencies` are always installed regardless of how the package is used.

At least the `@types`, `react-scripts` and `typescript` packages are only required for the Feast UI development, so they don't necessarily need to be installed by the application. From the runtime dependencies, possibly at least `moment`, `prop-types`, `query-string`, `use-query-params` or `zod` could be omitted.

shuchu:
Hi All, the CRA is deprecated. Any suggestion about UI code should move
forward? There is a package named nth-check is a dependent of
react-scripts, and it has a high severity security issue.

peruukki:
I suppose the first step would be to [eject](https://create-react-app.dev/docs/available-scripts#npm-run-eject) and make the setup work with that. Then it would be possible to remove unnecessary depedencies and possibly switch any outdated ones..

peruukki:
[...]

PR #4554, opened 2024-09-21
fix: Remove Feast UI TypeScript dependencies from `peerDependencies` and `dependencies`

peruukki:

Two commits to clean up the Feast UI `dependencies` or `peerDependencies`:
1. Move `@types` packages from `dependencies` to `devDependencies` and remove them from `peerDependencies`:
   - Type declaration files are only used during development and not included in the runtime code, so they should be in `devDependencies`. The [DefinitelyTyped docs](https://definitelytyped.org/) also include `--save-dev` in installation commands for this.
   - We also shouldn't require dependent projects to install them by including them in `peerDependencies`; the projects don't necessarily use TypeScript.
2. Remove `typescript` from `peerDependencies` for similar reasons: it's only used during development and the dependent project is free to use whichever version of TypeScript or even not use it.

This is related to #3785 and partially resolves it, but I wouldn't close the issue still with this. I think [there are still other packages](https://github.com/feast-dev/feast/issues/3785#issuecomment-2350990358) in `peerDependencies` that don't need to, but I wanted to improve the situation with the TypeScript related ones separately with this PR.


franciscojavierarceo (approved):

shuchu:
probably you need the npm website access? https://www.npmjs.com/package/@feast-dev/feast-ui

Issue #3794, opened 2023-10-12
EuiCustomLink causing error on feast-ui NPM package


pcminh0505:
Hi, I'm new to Feast and Feast UI, trying to construct a React app using your `feast-ui` package to display multiple feature repos on the UI.

It seems to me that the current NPM package of `feast-ui` is having some problems with the custom component `EuiCustomLink` that fails to parse the path

## Expected Behavior 
Here's the UI with the `feast ui` SDK inside one feature_repo project. I'm using the one inside your `feast/ui/feature_repo` folder on credit_scoring. There was no problem when I clicked Feature View/credit_history 

It is Worth noticing that the date_added = `2022-02-6` is related to the NPM package issue.
![image](https://github.com/feast-dev/feast/assets/54668379/74ba974d-b2f8-401d-929f-25a5e010f495)

## Current Behavior
When I tried replicating with a React app, managing multiple projects (later will show the `projects-list.json`), it prompted the error that could not query the time in the path. The `Data Sources` and `Entities` tabs are fine, only for the `Feature Views` and `Feature Services` that you're applying some filter.
![2023-10-12 18_38_21-Calendar](https://github.com/feast-dev/feast/assets/54668379/9cf63025-bb57-49c7-9d5b-0e06f8c00bf7)

## Steps to reproduce
1. Setup the boilerplate project like the [official docs](https://www.npmjs.com/package/@feast-dev/feast-ui)
2. Run `feast apply` inside the [feast/ui/feature_repo](https://github.com/feast-dev/feast/tree/master/ui/feature_repo) to get the `registry.db` file
3. Copy the `registry.db` file to the `/public` directory of React app
4. Specify the `projects-list.json` file as:
```json
{
  "projects": [
    {
      "name": "Credit Score Project",
      "description": "Project for credit scoring team and associated models.",
      "id": "credit_scoring_aws",
      "registryPath": "/credit_scoring_aws.db"
    },
    {
      "name": "Driver Ranking Project",
      "description": "Project for driver ranking team and associated models.",
      "id": "driver_ranking",
      "registryPath": "/driver_ranking.db"
    }
  ]
}
```
5. Let the `<FeastUI />` as a default component in `index.js` file and run `yarn start`

### Specifications
I tried both using the latest dependency packages version and downgrade to meet the exact version in the `feast/ui` repo, but none of them worked.
```
"dependencies": {
    "@elastic/datemath": "^5.0.3",
    "@elastic/eui": "^94.1.0",
[...]

PR #4556, opened 2024-09-22
fix: Update react-router-dom to 6.3.0 and restrict its version in Feast UI

peruukki:

As noted in #3794, Feast UI is not compatible with the latest react-router-dom versions, more precisely [from 6.4.0 onwards](https://stackoverflow.com/a/74252839/305436). Limit react-router-dom version to a compatible range to avoid the runtime errors mentioned in the issue when installing app dependencies without specifying exact versions.

After setting the restricted versions in `package.json`, running `yarn install` updated `react-router-dom` to the latest compatible version 6.3.0. It should be a minor upgrade (https://github.com/remix-run/react-router/blob/main/CHANGELOG.md#v630), and I didn't notice anything not working in the UI after it.

#3794 has already been closed since there was another way to resolve it, but otherwise this should have fixed it too.

# Misc

I've noticed that regardless of which versions you define in `peerDependencies`, at least Yarn 1 (Classic) installs the latest versions of those packages in the app project with `yarn add`. But the peer dependency versions still have an impact:

- If I keep the react-router-dom version as `6` in `peerDependencies`, Yarn installs the latest version in the root `node_modules` and uses only that, and I get the error in #3794
- If I restrict react-router-dom version to `<6.4.0` in `peerDependencies`, Yarn installs the latest version in the root `node_modules` and version 6.3.0 in `@feast-dev/feast-ui/node_modules`, so Feast UI uses 6.3.0 and the rest of the app the latest version, and I don't get errors (using different versions of react-router-dom **might** introduce some other problems, but limiting the version in `peerDependencies` is the best we can do to influence the app dependency version)  

There are differences in how different package managers (and their different versions) handle peer dependency versions, and you can sometime influence that with command line arguments. But the main point here was to verify that we indeed should update the version in `peerDependencies` too. 😄

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -20,7 +20,7 @@
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
     "react-query": "^3.34.12",
-    "react-router-dom": "6",
+    "react-router-dom": "<6.4.0",
     "react-scripts": "^5.0.0",
     "use-query-params": "^1.2.3",
     "zod": "^3.11.6"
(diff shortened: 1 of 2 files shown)


shuchu (approved):
lgtm

Issue #4529, opened 2024-09-17
Remote apply


dmartinol:
## Expected Behavior 
Running `feast apply` from a client connected using remote proxies should apply the python definitions.

## Current Behavior
The apply fails. With a project using the `postgres` template and a deployment on `kind` using #4528 it throws the following error:
```bash
% feast feature-views list
NAME                         ENTITIES    TYPE
driver_hourly_stats_fresh    {'driver'}  FeatureView
driver_hourly_stats          {'driver'}  FeatureView
transformed_conv_rate_fresh  {'driver'}  OnDemandFeatureView
transformed_conv_rate        {'driver'}  OnDemandFeatureView

% feast apply
...
  File "/Users/dmartino/.pyenv/versions/3.11.9/bin/feast", line 8, in <module>
    sys.exit(cli())
             ^^^^^
...
  File "/Users/dmartino/.pyenv/versions/3.11.9/lib/python3.11/site-packages/feast/repo_operations.py", line 355, in apply_total
    apply_total_with_repo_instance(
  File "/Users/dmartino/.pyenv/versions/3.11.9/lib/python3.11/site-packages/feast/repo_operations.py", line 313, in apply_total_with_repo_instance
    store.apply(all_to_apply, objects_to_delete=all_to_delete, partial=False)
  File "/Users/dmartino/.pyenv/versions/3.11.9/lib/python3.11/site-packages/feast/feature_store.py", line 903, in apply
    self._make_inferences(
  File "/Users/dmartino/.pyenv/versions/3.11.9/lib/python3.11/site-packages/feast/feature_store.py", line 627, in _make_inferences
    update_feature_views_with_inferred_features_and_entities(
  File "/Users/dmartino/.pyenv/versions/3.11.9/lib/python3.11/site-packages/feast/inference.py", line 174, in update_feature_views_with_inferred_features_and_entities
    _infer_features_and_entities(
  File "/Users/dmartino/.pyenv/versions/3.11.9/lib/python3.11/site-packages/feast/inference.py", line 212, in _infer_features_and_entities
    table_column_names_and_types = fv.batch_source.get_table_column_names_and_types(
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #4542, opened 2024-09-19
Registry server does not apply error conversion with no_auth setting


dmartinol:
## Expected Behavior 
When a Registry Server is launched with `no_auth` setting, the remote client proxy must receive the same original error as generated on the server-side.

## Current Behavior
The clients receives an `_InactiveRpcError` instead.

## Steps to reproduce
- Launch a registry server with `no_auth` setting.
- From a remote client invoke a method that raises an error, e.g. `store.list_entities('invalid_name')`

## Possible Solution
The [interceptor](https://github.com/feast-dev/feast/blob/0fb76e9041885659c68e294b0c033c62050bd374/sdk/python/feast/infra/registry/remote.py#L70-L72) must be added with any auth setting.

PR #4559, opened 2024-09-23
fix: Remote apply using offline store

dmartinol:
Moves methods used to run `feast apply` in the Offline Server.
E.g. `DataSource.get_column_names_and_types`

Fixes https://github.com/feast-dev/feast/issues/4529 and https://github.com/feast-dev/feast/issues/4542

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -602,16 +602,23 @@ class FeatureStore:
 
         # New feature views may reference previously applied entities.
         entities = self._list_entities()
+        provider = self._get_provider()
         update_feature_views_with_inferred_features_and_entities(
-            views_to_update, entities + entities_to_update, self.config
+            provider,
+            views_to_update,
+            entities + entities_to_update,
+            self.config,
         )
         update_feature_views_with_inferred_features_and_entities(
-            sfvs_to_update, entities + entities_to_update, self.config
+            provider,
+            sfvs_to_update,
+            entities + entities_to_update,
+            self.config,
         )
         # We need to attach the time stamp fields to the underlying data sources
         # and cascade the dependencies
         update_feature_views_with_inferred_features_and_entities(
-            odfvs_to_update, entities + entities_to_update, self.config
+            provider, odfvs_to_update, entities + entities_to_update, self.config
         )
         # TODO(kevjumba): Update schema inference
         for sfv in sfvs_to_update:
@@ -1529,9 +1536,12 @@ class FeatureStore:
                 feature_view_name, allow_registry_cache=allow_registry_cache
             )
 
+        provider = self._get_provider()
         # Get columns of the batch source and the input dataframe.
         column_names_and_types = (
-            feature_view.batch_source.get_table_column_names_and_types(self.config)
+            provider.get_table_column_names_and_types_from_data_source(
+                self.config, feature_view.batch_source
+            )
         )
         source_columns = [column for column, _ in column_names_and_types]
         input_columns = df.columns.values.tolist()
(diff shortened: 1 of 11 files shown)


sdk/python/feast/infra/passthrough_provider.py
```diff
@@ -427,3 +439,12 @@ def validate_data_source(
         data_source: DataSource,
     ):
         self.offline_store.validate_data_source(config=config, data_source=data_source)
+
+    def get_table_column_names_and_types_from_data_source(
+        self, config: RepoConfig, data_source: DataSource
+    ) -> Iterable[Tuple[str, str]]:
+        if isinstance(data_source, FileSource):
```
tokoko, line 446: why do we need this? note that FileSource is not necessarily local, it might be an S3 path that's only accessible from offline server in a secure deployment.

dmartinol: main reason is that ITs are failing, probably due to messages like 
```
  sdk/python/feast/feature_view.py:425: UserWarning: There are some mismatches in your feature view: order registered entities. Please check if you have applied your entities correctly.Entities: ['customer_id', 'driver'] vs Entity Columns: [driver_id-Int64]
```
but if S3 is also there, it must be fixed differently. I will investigate further

dmartinol: The root cause of the original issue `RemoteOfflineStoreConfig' object has no attribute 'user'` is the following: in a distributed setup with a remote offline server, `feast apply` cannot run from the client host because some functions assume to be executed using the actual feast `offline_store` configuration, not the remote one.
`DataSource.get_table_column_names_and_types` is one of such functions.

The idea behind this PR was to move the `get_table_column_names_and_types` execution under the offline store responsibilities, so that it could execute within the offline server in case of a distributed setup. 
By passing the data source name to the remote request, the offline server can fetch the data source instance from the registry and delegate execution to it, allowing the data source to run using its actual connection.

This idea fails miserably until the data source has been successfully registered, e.g. `feast apply` was properly executed, which is exactly the issue we want to fix.

So, the only option to let the offline store handle the data source instance without fetching it from the DB seems to pass its entire configuration to the server, maybe using the protos to easily rebuild the instance. WDYT?

dmartinol: @tokoko PR ready: using protos instead of names

dmartinol:
@tokoko @franciscojavierarceo no feedback?

sdk/python/feast/feature_store.py
```diff
@@ -602,11 +602,18 @@ def _make_inferences(
...
         update_feature_views_with_inferred_features_and_entities(
-            views_to_update, entities + entities_to_update, self.config
+            provider,
+            views_to_update,
+            entities + entities_to_update,
+            self.config,
         )
         update_feature_views_with_inferred_features_and_entities(
-            sfvs_to_update, entities + entities_to_update, self.config
+            provider,
```
franciscojavierarceo, line 613: There should be a new one for ODFVs

sdk/python/feast/inference.py
```diff
@@ -95,7 +96,8 @@ def update_data_sources_with_inferred_event_timestamp_col(
 
 
 def update_feature_views_with_inferred_features_and_entities(
-    fvs: Union[List[FeatureView], List[StreamFeatureView], List[OnDemandFeatureView]],
+    provider: Provider,
+    fvs: Union[List[FeatureView], List[StreamFeatureView]],
```
franciscojavierarceo, line 100: this is missing OnDemandFeatureView

franciscojavierarceo (changes requested):
small request for a change

sdk/python/feast/feature_store.py
dmartinol: yep, 3 weeks ago it was not there 😉 
will add it

franciscojavierarceo (approved):
lgtm

Issue #4455, opened 2024-08-27
Provide Feature Serving and Model Inference Diagram


franciscojavierarceo:
**Is your feature request related to a problem? Please describe.**
We have outlined some options for feature serving and model inference in this documentation: https://docs.feast.dev/v/master/getting-started/architecture/model-inference

In it, we briefly discuss communication patterns and their tradeoffs (elaborated further here: https://docs.feast.dev/v/master/getting-started/architecture/write-patterns).

We don't have an architecture diagram to support these pieces.

**Describe the solution you'd like**
For a Fintech article I'm writing, I ended up drafting this diagram. I think we could potentially adjust this to use in our documentation.

![Screenshot 2024-08-27 at 6 38 42 AM](https://github.com/user-attachments/assets/19a5f091-440d-48d2-b616-d6b0c26de91c)

**Describe alternatives you've considered**
N/A

**Additional context**
Good diagrams provide clarity and what not, so I thought this would be useful.

franciscojavierarceo:
@tokoko any thoughts here? Otherwise I'll cut a PR with this to add to the documentation. I think it's missing the registry, so I could add that as well.

tokoko:
a couple of remarks
- should feast box contain `training datasets`? Once they are generated, they are effectively outside of feast at that point.
- What does a line coming from an online store marked `Logs/CDC/ELT` refer to?

franciscojavierarceo:
That's a good call out! 

The logs/CDC/ELT refers to logs that should be sent from the feature server during transformations. This is the suggestion I would make to most folks to ensure feature and model reproducibility/replay.

franciscojavierarceo:
It could be events to SQS or Kafka/Kinesis or through a sidecar.

franciscojavierarceo:
@tokoko I added the training datasets, lmk what you think.

![feast_model_serving_architecture](https://github.com/user-attachments/assets/76ab07de-7825-447f-b2d1-9062fba2e1e9)

PR #4570, opened 2024-09-24
feat: Updated Feast model Inference Architecture

franciscojavierarceo:

This PR updates the documentation to provide a diagram of how an ML Platform can work with Feast.

https://github.com/feast-dev/feast/issues/4455

# Misc


HaoXuAI (approved):
Looks good.
I think the application -> online store can also be async write, just not supported yet.
That's something I think we can explore more, as it gives user more flexibility to write to feast directly.

franciscojavierarceo:
> Looks good. I think the application -> online store can also be async write, just not supported yet. That's something I think we can explore more, as it gives user more flexibility to write to feast directly.

I actually covered some of that in the new architecture section under [Communication Patterns](https://docs.feast.dev/v/master/getting-started/architecture/write-patterns#communication-patterns). 

># Communication Patterns
> There are two ways a client (or Data Producer) can send data to the online store:
> 
> 1. Synchronously
> - Using a synchronous API call for a small number of entities or a single entity (e.g., using the [push or write_to_online_store methods](https://docs.feast.dev/v/master/reference/data-sources/push#pushing-data)) or the Feature Server's [push endpoint](https://docs.feast.dev/v/master/reference/feature-servers/python-feature-server#pushing-features-to-the-online-and-offline-stores))
> 
> 2. Asynchronously
> - Using an asynchronous API call for a small number of entities or a single entity (e.g., using the [push or write_to_online_store methods](https://docs.feast.dev/v/master/reference/data-sources/push#pushing-data)) or the Feature Server's [push endpoint](https://docs.feast.dev/v/master/reference/feature-servers/python-feature-server#pushing-features-to-the-online-and-offline-stores))
> - Using a "batch job" for a large number of entities (e.g., using a [batch materialization engine](https://docs.feast.dev/v/master/getting-started/components/batch-materialization-engine))
> 
> Note, in some contexts, developers may "batch" a group of entities together and write them to the online store in a single API call. This is a common pattern when writing data to the online store to reduce write loads but we would not qualify this as a batch job.

Let me know if you think I should add some info there. In general, if users are going to do an async write, I think that has different tradeoffs that we want people to be thoughtful about, which is what I tried to say more broadly in that [Write Patterns](https://docs.feast.dev/v/master/getting-started/architecture/write-patterns) architecture document.

PR #4571, opened 2024-09-25
chore: Test FeatureView apply to expose entity_columns behavior pre and post apply()

franciscojavierarceo:
Feature Views have the property that during class instantiation, entity columns are not serialized to populate `DUMMY_ENTITY_ID` but are populated after the `FeatureStore.apply()` method is called. This PR tests that behavior explicitly.

N/A

# Misc


shuchu (approved):
lgtm

PR #4577, opened 2024-09-26
fix: Remove unnecessary peer dependencies from Feast UI

### LINKED ISSUE #3785: Get rid of duplicate peer dependencies at Feast UI

peruukki:
npm [peer dependencies](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#peerdependencies) in a consumed package are used for dependencies that should be installed by the consuming package/app and usually shared between the two. Most of Feast UI's current peer dependencies can be removed because installing them is not necessarily required in the consuming app when `@feast-dev/feast-ui` is used as a module.  The less we have peer dependencies, the easier it is to take Feast UI into use as a module.

Fixes #3785.

# Misc
Rationale for the new set of peer dependencies is given in https://github.com/feast-dev/feast/issues/3785#issuecomment-2373098965 and in the commit message)

### DIFF
--- a/docs/reference/alpha-web-ui.md
+++ b/docs/reference/alpha-web-ui.md
@@ -45,11 +45,14 @@ Start with bootstrapping a React app with `create-react-app`
 npx create-react-app your-feast-ui
 ```
 
-Then, in your app folder, install Feast UI and its peer dependencies. Assuming you use yarn
+Then, in your app folder, install Feast UI and optionally its peer dependencies. Assuming you use yarn
 
 ```
 yarn add @feast-dev/feast-ui
-yarn add @elastic/eui @elastic/datemath @emotion/react moment prop-types inter-ui react-query react-router-dom use-query-params zod typescript query-string d3 @types/d3
+# For custom UI using the Elastic UI Framework (optional):
+yarn add @elastic/eui
+# For general custom styling (optional):
+yarn add @emotion/react
 ```
 
 Edit `index.js` in the React app to use Feast UI.
--- a/ui/README.md
+++ b/ui/README.md
@@ -21,11 +21,14 @@ Start with bootstrapping a React app with `create-react-app`
 npx create-react-app your-feast-ui
 ```
 
-Then, in your app folder, install Feast UI and its peer dependencies. Assuming you use yarn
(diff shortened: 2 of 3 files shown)


docs/reference/alpha-web-ui.md
```diff
@@ -45,11 +45,14 @@ Start with bootstrapping a React app with `create-react-app`
 npx create-react-app your-feast-ui
 ```
 
-Then, in your app folder, install Feast UI and its peer dependencies. Assuming you use yarn
+Then, in your app folder, install Feast UI and optionally its peer dependencies. Assuming you use yarn
 
 ```
 yarn add @feast-dev/feast-ui
-yarn add @elastic/eui @elastic/datemath @emotion/react moment prop-types inter-ui react-query react-router-dom use-query-params zod typescript query-string d3 @types/d3
```
franciscojavierarceo, line 52: Why remove all of these?

ui/README.md
```diff
@@ -21,11 +21,14 @@ Start with bootstrapping a React app with `create-react-app`
 npx create-react-app your-feast-ui
 ```
 
-Then, in your app folder, install Feast UI and its peer dependencies. Assuming you use yarn
+Then, in your app folder, install Feast UI and optionally its peer dependencies. Assuming you use yarn
 
 ```
 yarn add @feast-dev/feast-ui
-yarn add @elastic/eui @elastic/datemath @emotion/react moment prop-types inter-ui react-query react-router-dom use-query-params zod typescript query-string d3 @types/d3
```
franciscojavierarceo, line 28: Same question, shouldn't these go below in optional?

docs/reference/alpha-web-ui.md
peruukki: These are no longer in the peer dependencies, so there's no need to install them anymore, at least not for Feast UI to work.

peruukki: So these are now considered internal dependencies of Feast UI, mostly irrelevant for apps using it as a module.

PR #4581, opened 2024-09-27
fix: Fix online pg import

HaoXuAI:

# Misc

### DIFF
--- a/sdk/python/feast/infra/online_stores/helpers.py
+++ b/sdk/python/feast/infra/online_stores/helpers.py
@@ -1,8 +1,8 @@
 import struct
+from datetime import datetime
 from typing import Any, List
 
 import mmh3
-from datetime import datetime
 
 from feast.importer import import_class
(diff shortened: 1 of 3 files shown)


sdk/python/feast/infra/online_stores/helpers.py
```diff
@@ -62,3 +63,10 @@ def compute_entity_id(
             entity_key_serialization_version=entity_key_serialization_version,
         )
     ).hex()
+
+
+def _to_naive_utc(ts: datetime) -> datetime:
```
franciscojavierarceo, line 68: would it make sense to set a defalt timezone instead of a none?

HaoXuAI: not sure, i didn't implement it but just moved it out to helper function. it currently in singlestore module and imported into postgres module lol

franciscojavierarceo (approved):

PR #4600, opened 2024-10-04
feat: Add connection_name field to Snowflake config

### LINKED ISSUE #4592: Use Snowflake connections.toml file
**Problem**
We would like to directly use the `connections.toml` file for the Snowflake connector for the Feast Snowflake integration. Setting a `connection_name` in the feature store configuration file would allow us to use the same connection configuration for the feature store that is also used in other python scripts that access Snowflake.

Currently, the feature store configuration gives the option to add a `config_path`. To my understanding, this file is a SnowSQL config file which is not directly compatible with the Python Snowflake connector that is used under the hood in Feast.

**Proposed Solution**
My understanding is, that if the `connection_name` is passed to the `snowflake.connector` it will look for the default location of the file and load the configuration parameters for the defined connection. https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-connect#connecting-using-the-connections-toml-file

Thus, adding a field with name `connection_name` in https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/offline_stores/snowflake.py#L79 would work for the offline store. The other Snowflake resources need to be adjusted similarly.

I would be open for a contribution.

**Alternatives**
A workaround where the connection is configured in an ini file that matches the config in the `connections.toml` file.

nanohanno:

Add the `connection_name` field to Snowflake config so that the configuration can be loaded from the canonical `connections.toml` file used by Snowflake connector for Python.

Fixes #4592 

# Misc

### DIFF
--- a/sdk/python/feast/infra/materialization/snowflake_engine.py
+++ b/sdk/python/feast/infra/materialization/snowflake_engine.py
@@ -46,7 +46,10 @@ class SnowflakeMaterializationEngineConfig(FeastConfigBaseModel):
     """ Type selector"""
 
     config_path: Optional[str] = os.path.expanduser("~/.snowsql/config")
-    """ Snowflake config path -- absolute path required (Cant use ~)"""
+    """ Snowflake snowsql config path -- absolute path required (Cant use ~)"""
+
+    connection_name: Optional[str] = None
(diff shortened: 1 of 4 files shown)


franciscojavierarceo (approved):
lgtm

PR #4601, opened 2024-10-04
fix: _python_value_to_proto_value does not handle list values

TBourton:

Currently, _python_value_to_proto_value does not properly handle lists. It will always return a list of values of length one, regardless of the number of input values.


TBourton:
@kind/bug

Issue #4605, opened 2024-10-05
Enable feast to use current session based client instead of default one for AWS dynamodb


asingh9530:
**Is your feature request related to a problem? Please describe.**
Hi Team, I want to use feast in a cross aws account setup and I am failing to understand how that would work. Here is the problem statement I am running a feast job in aws account A and the online store (dynamodb ) is in aws account B now even after running assume role operation the feast still picking default dynamodb of aws account A and the job is failing. 

**Describe the solution you'd like**
Instead of creating client from boto itself, we could enable feast to use current session as a option. 

**Describe alternatives you've considered**
I have rewritten default client 

```
boto3.client(
            "dynamodb",
            region_name=region,
            endpoint_url=endpoint_url,
            config=Config(user_agent=get_user_agent()),
        )
```

to this

```
boto3.Session().client(
            "dynamodb",
            region_name=region,
            endpoint_url=endpoint_url,
            config=Config(user_agent=get_user_agent()),
        )
```

**Additional context**
@tokoko @franciscojavierarceo  we have a detailed conversation here https://feastopensource.slack.com/archives/C01M2GYP0UC/p1723995709072149

PR #4606, opened 2024-10-05
feat: Add boto3 session based auth for dynamodb online store for cross account access

asingh9530:
Hi Team,

This PR enables feast to create aws client from either current session or using default boto3.client. This will enable feast to use dynamodb in a cross account setup i.e if feast is running in account A and want to use dynamodb in account B, right now it dosen't offers that and creates a default client which is limits the usage of feast.

https://github.com/feast-dev/feast/issues/4605

# Misc

### DIFF
--- a/sdk/python/feast/infra/online_stores/dynamodb.py
+++ b/sdk/python/feast/infra/online_stores/dynamodb.py
@@ -473,6 +473,7 @@ def _initialize_dynamodb_client(
     endpoint_url: Optional[str] = None,
     session_based_auth: Optional[bool] = False,
 ):
+
     if session_based_auth:
         return boto3.Session().client(
             "dynamodb",
--- a/sdk/python/feast/infra/online_stores/dynamodb.py
+++ b/sdk/python/feast/infra/online_stores/dynamodb.py
@@ -473,7 +473,6 @@ def _initialize_dynamodb_client(
     endpoint_url: Optional[str] = None,
     session_based_auth: Optional[bool] = False,
 ):
-
     if session_based_auth:
         return boto3.Session().client(
             "dynamodb",
--- a/sdk/python/feast/infra/online_stores/dynamodb.py
+++ b/sdk/python/feast/infra/online_stores/dynamodb.py
@@ -73,6 +73,7 @@ class DynamoDBOnlineStoreConfig(FeastConfigBaseModel):
     session_based_auth: bool = False
     """AWS session based client authentication"""
 
+
 class DynamoDBOnlineStore(OnlineStore):
     """
     AWS DynamoDB implementation of the online store interface.
@@ -106,10 +107,14 @@ class DynamoDBOnlineStore(OnlineStore):
         online_config = config.online_store
         assert isinstance(online_config, DynamoDBOnlineStoreConfig)
         dynamodb_client = self._get_dynamodb_client(
-            online_config.region, online_config.endpoint_url, online_config.session_based_auth
+            online_config.region,
+            online_config.endpoint_url,
+            online_config.session_based_auth,
         )
         dynamodb_resource = self._get_dynamodb_resource(
-            online_config.region, online_config.endpoint_url, online_config.session_based_auth
+            online_config.region,
+            online_config.endpoint_url,
+            online_config.session_based_auth,
         )
         # Add Tags attribute to creation request only if configured to prevent
         # TagResource permission issues, even with an empty Tags array.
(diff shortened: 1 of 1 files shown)


sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -443,17 +468,38 @@ def _to_client_batch_get_payload(online_config, table_name, batch):
...
-        region_name=region,
-        endpoint_url=endpoint_url,
-        config=Config(user_agent=get_user_agent()),
-    )
+def _initialize_dynamodb_client(
+    region: str,
+    endpoint_url: Optional[str] = None,
+    session_based_auth: Optional[bool] = False,
+):
+    if session_based_auth is True:
```
franciscojavierarceo, line 476:
```suggestion
    if session_based_auth:
```

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -443,17 +468,38 @@ def _to_client_batch_get_payload(online_config, table_name, batch):
...
 
 
-def _initialize_dynamodb_resource(region: str, endpoint_url: Optional[str] = None):
-    return boto3.resource("dynamodb", region_name=region, endpoint_url=endpoint_url)
+def _initialize_dynamodb_resource(
+    region: str,
+    endpoint_url: Optional[str] = None,
+    session_based_auth: Optional[bool] = False,
+):
+    if session_based_auth is True:
```
franciscojavierarceo, line 497:
```suggestion
    if session_based_auth:
```

franciscojavierarceo (changes requested):
Some small nits but this lgtm!

asingh9530:
> Co-authored-by: Francisco Arceo <farceo@redhat.com>

I think DCO does not work on github commits, could you help.

asingh9530:
> > Co-authored-by: Francisco Arceo [farceo@redhat.com](mailto:farceo@redhat.com)
> 
> I think DCO does not work on github commits, could you help. Thanks

No Need did a force signoff

franciscojavierarceo (approved):

asingh9530:
@franciscojavierarceo added a dummy commit as one of the workflow was stuck.

franciscojavierarceo:
There's also a button in the UI fwiw

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -70,6 +70,9 @@ class DynamoDBOnlineStoreConfig(FeastConfigBaseModel):
     tags: Union[Dict[str, str], None] = None
     """AWS resource tags added to each table"""
 
+    session_based_auth: bool = False
```
tokoko, line 73: This is too much bother, I think. We should simply switch to session based, there's no real upside in keeping old behavior, unless I'm missing something.

franciscojavierarceo: Breaking changes create friction to using later versions. I'm fine for getting rid of it in a subsequent release if we give users notice through a warning.

tokoko: What makes this a breaking change? We're changing an internal detail of how a client is instantiated, it's not a change in an API. can't really imagine how anyone would be depending on the previous behavior of us using a default session.

franciscojavierarceo: Oh I see, this isn't configured through the feature_store.yaml, my bad.

franciscojavierarceo (approved):

sdk/python/feast/infra/online_stores/dynamodb.py
asingh9530: @tokoko I would still argue we should use this setup in upcoming release and completely switch to session in next just to make sure we don't break anything internally or for feast existing customers although I can't really think if this would break anything but just as a caution.

asingh9530: What do you think ?

asingh9530: @tokoko any update ?

tokoko: @asingh9530 sorry. I'm fine with proceeding as is, although I think we're being over-cautious. Ironically when we make changes later on to remove this config, the removal will technically be a breaking change :laughing: anyway, that's fine as well, let's just not forget to remove it though.

asingh9530: Lol let me add a warning though if we use default client.

PR #4608, opened 2024-10-07
fix: Correctly handle list values in _python_value_to_proto_value

TBourtonBumble:
Currently, `_python_value_to_proto_value` does not properly handle lists. It will always return a list of values of length one, regardless of the number of input values.

This PR aims to fix this bug, and also adds a check to ensure that the number of returned proto values is the same as the inputted values.

### DIFF
--- a/sdk/python/feast/type_map.py
+++ b/sdk/python/feast/type_map.py
@@ -372,11 +372,18 @@ def _python_value_to_proto_value(
                 if feast_value_type == ValueType.BYTES_LIST:
                     raise _type_err(sample, ValueType.BYTES_LIST)
 
-                json_value = json.loads(sample)
-                if isinstance(json_value, list):
+                json_sample = json.loads(sample)
+                if isinstance(json_sample, list):
+                    json_values = [json.loads(value) for value in values]
                     if feast_value_type == ValueType.BOOL_LIST:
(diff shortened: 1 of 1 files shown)


sdk/python/feast/type_map.py
```diff
@@ -372,11 +372,15 @@ def _python_value_to_proto_value(
                 if feast_value_type == ValueType.BYTES_LIST:
                     raise _type_err(sample, ValueType.BYTES_LIST)
 
-                json_value = json.loads(sample)
-                if isinstance(json_value, list):
+                json_sample = json.loads(sample)
```
franciscojavierarceo, line 375: Any chance you can add a test here?

TBourtonBumble: Hi, sure thing. Did you have in mind something to check that `json.loads(sample)` does not error out? E.g. if there's something like a json decode error?

franciscojavierarceo (approved):

sdk/python/feast/type_map.py
franciscojavierarceo: Actually, it's fine. There's no test there currently so no need to add additional work here. Thank you!

TBourtonBumble:
Hi sorry, I missed something very dumb with the unit test, i've pushed a fix so they should be ok to re-run

Issue #4479, opened 2024-09-03
OnDemandFeatureViews with RequestSource returns different columns depending on online/offline feature retrieval.


job-almekinders:
## Context

I'm using an `ODFV` with a `PandasTransformation` and a `RequestSource` input. Furthermore, I'm using Postgres offline and online store.

## Expected Behavior 

I would expect both the `get_online_features_async` method and the `get_historical_features` method to return the same set of features. However, when retrieving features with `get_online_features_async`, the input data from the `RequestSource` is **not present** in the output response. On the other hand, when retrieving features with `get_historical_features`, the input data from the `RequestSource` **is present** in the output response. 

I'm not sure which behavior is to be expected. However, I think think that both approaches should return the same columns.

## Current Behavior
When calling the `get_online_features_async` method I do not see the input from the `RequestSource` back in the output response, while I would expect it to be there.

One could also argue that it should not be in the output response. That would also be an option. However, I would assume that the online and offline feature retrieval would return the same output in terms of columns.

## Steps to reproduce

docker-compose.yml
```yml
---
version: "3"
services:
  offline_store:
    image: postgres:16-alpine
    container_name: offline_store
    ports:
      - "6543:5432"
    environment:
      - POSTGRES_DB=offline_store
      - POSTGRES_USER=postgres
      - POSTGRES_PASSWORD=postgres
    volumes:
      - ./postgres_init:/docker-entrypoint-initdb.d
  online_store:
    image: postgres:16-alpine
    container_name: online_store
    ports:
      - "5432:5432"
    environment:
      - POSTGRES_DB=online_store
      - POSTGRES_USER=postgres
      - POSTGRES_PASSWORD=postgres

```

feature_store.yml
```yml
project: feast_tryout
provider: local
registry:
  registry_type: sql
  path: postgresql+psycopg2://postgres:postgres@0.0.0.0:5432/online_store
  cache_ttl_seconds: 60
online_store:
  type: postgres
  host: 0.0.0.0
  port: 5432
  database: online_store
  db_schema: online
  user: postgres
  password: postgres
offline_store:
  type: postgres
  host: 0.0.0.0
  port: 6543
  database: offline_store
  db_schema: offline
  user: postgres
  password: postgres
entity_key_serialization_version: 2
```

Insert into offline store (postgres)
postgres_init/create-offline-store-database.sql
```sql
CREATE SCHEMA offline;

CREATE TABLE offline.features (
  "ENTITY_ID"         VARCHAR,
[...]

PR #4615, opened 2024-10-10
chore: Updated tests to confirm behavior of inconsistent offline/online return values

franciscojavierarceo:
This PR updates `get_historical_features` to return the requested set of features when ODFVs are called as `get_online_features`. 

As outlined in https://github.com/feast-dev/feast/issues/4479, `get_historical_features` returns more features than what is requested when an ODFV is requested. 

This PR updates:
1. The tests to verify that `get_historical_features` when calling an ODFV returns only the fields requested
2. The tests to verify that an ODFV will result in an error when `get_historical_features` is called with an entity that does not have data to calculate an ODFV
3. The code to properly return only the requested features in `get_historical_features`

https://github.com/feast-dev/feast/issues/4479

# Misc
N/A


HaoXuAI (approved):
LGTM

Issue #4593, opened 2024-10-01
Improving the feast quickstart documentation


lokeshrangineni:
This issue we are creating to track the [suggestions](https://feastopensource.slack.com/archives/C01M2GYP0UC/p1727721853614169) recommended by feast community member - @MattDelac 

 I wanted to drop some feedback and notes here about my experience:

1. The generated template doesn't contain .gitignore for feast
2. The generated template does not contain any pyproject.toml or requirements file, this is something I'd expect from a scaffolding script for a python-based codebase
3. The quickstart guide doesn't mention anything about instantiating a new virtual environment, this is also something I'd expect
4. If feast is meant to by an executable, I'd recommend instructing users to install it via [pipx](https://github.com/pypa/pipx) instead of pip.
5. How do I start the feast server? Do I even start it? It's not clear to me
6. The feast init generated templates use absolute paths for some files instead of relative paths. I usually find this to be confusion/bad as people can codify things about their local machines into code bases without realizing it (ex: the example_repo.py) contains path=/Users/matt/code/pnl/feast-test/feast_test/feature_repo/data")
7. not super sure why the generated repo is a python module (contains an _init_.py file but also has a sub directory/module called feature_repo
8. nit: I don't understand why Step 3 has a, b, c, d, and e? Maybe just make it step 3, 4, 5, 6 etc
9. Include a description of what Feast is
10. What feast includes
11. Not defining the fundamental concepts in an easy way
12. Project isn't defined under Concepts>Overview
13. Entity concept documentation was confusing
14. feedback unrelated to docs but would be helpful to add the _str_, _repr_ and maybe even __repr_html__ methods to the FeatureStore class to make debugging easier
15. Another quickstart suggestion: I'd recommend actually including the code itself in the quickstart rather than relying on users to have to run feast init. This reduces friction for new people by allowing them to quickly skim the quickstart and follow along without having to be at their computer and install something (for example if they're on their phone).
It also helps ensure the quickstart guide is at least internally consistent meaning you don't have a situation where the someone updates the generated code but forgets to update the quickstart guide. It also helps ensure the quickstart guide is at least internally consistent meaning you don't have a situation where the someone updates the generated code but forgets to update the quickstart guide.

hadisbabakhani:
Missing .gitignore: The generated template doesn't include a .gitignore file specific to Feast. This would be helpful to avoid committing unnecessary files.

[...]

PR #4624, opened 2024-10-14
fix: Changes template file path to relative path

tmihalac:
Changes template file path to relative path, which is more correct

Fixes https://github.com/feast-dev/feast/issues/4593


franciscojavierarceo (approved):
lgtm

Issue #4623, opened 2024-10-14
Kubernetes authorization manager should not use ClusterRoles


dmartinol:
## Expected Behavior 
Kubernetes authorization manager should rely only on Roles and RoleBindings defined in the namespace of the Feast service.

## Current Behavior
Kubernetes authorization manager also fetches ClusterRoles which means that the Feast service must be granted cluster-wide privileges to inspect these instances.

## Steps to reproduce
Install a Feast service in Kubernetes and enable the `kubernetes` authorization module.

## Possible Solution
Use only Roles and RoleBindings defined in the namespace of the Feast service.

PR #4625, opened 2024-10-14
fix: Replaced ClusterRoles with local RoleBindings

dmartinol:
Avoid assigning cluster-wide privileges to Feast services to handle RBAC enforcement.

Foxes #4623 

# Misc
Tested on cluster with existing "rbac_remote" example (which IMO needs some review)

### DIFF
--- a/examples/rbac-remote/server/k8s/server_resources.yaml
+++ b/examples/rbac-remote/server/k8s/server_resources.yaml
@@ -5,23 +5,23 @@ metadata:
   namespace: feast-dev
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
+kind: Role
 metadata:
-  name: feast-cluster-role
+  name: feast-role
 rules:
   - apiGroups: ["rbac.authorization.k8s.io"]
-    resources: ["roles", "rolebindings", "clusterrolebindings"]
+    resources: ["roles", "rolebindings"]
     verbs: ["get", "list", "watch"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRoleBinding
+kind: RoleBinding
 metadata:
-  name: feast-cluster-rolebinding
+  name: feast-rolebinding
 subjects:
   - kind: ServiceAccount
     name: feast-sa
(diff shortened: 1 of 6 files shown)


dmartinol:
@tchughesiv @redhatHameed

examples/rbac-remote/server/k8s/server_resources.yaml
```diff
@@ -5,23 +5,23 @@ metadata:
   namespace: feast-dev
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
+kind: Role
 metadata:
-  name: feast-cluster-role
+  name: feast-role
```
tchughesiv, line 10: should probably add namespace here, same as above `ServiceAccount` object -
```yaml
  name: feast-role
  namespace: feast-dev
```

examples/rbac-remote/server/k8s/server_resources.yaml
```diff
@@ -5,23 +5,23 @@ metadata:
...
-    resources: ["roles", "rolebindings", "clusterrolebindings"]
+    resources: ["roles", "rolebindings"]
     verbs: ["get", "list", "watch"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRoleBinding
+kind: RoleBinding
 metadata:
-  name: feast-cluster-rolebinding
+  name: feast-rolebinding
```
tchughesiv, line 19: should probably add namespace here, same as above `ServiceAccount` object -
```yaml
  name: feast-rolebinding
  namespace: feast-dev
```

tchughesiv (changes requested):

examples/rbac-remote/server/k8s/server_resources.yaml
dmartinol: I think we cannot: I tried both the same NS as the RoleBinding or another one, but when I tried to export the manifest it was removed. I guess it MUST be on the same NS.

tchughesiv (approved):
lgtm

franciscojavierarceo (approved):

PR #4630, opened 2024-10-16
docs: Credit-risk-end-to-end example

dandawg:
This PR presents a clear step-by-step guide aimed at data science teams who need a basic hands-on introduction to Feast. I use the AI modeling workflow to demonstrate Feast in-context. In this example, the data doesn't randomly appear with `feast init`, and it's not randomly generated (except for timestamps and IDs). The example data comes from a popular OpenML dataset, and the reader progresses through data preparation, model training, and model serving. The example walks through the basic steps of the AI modeling process, and Feast solutions are brought in where they naturally provide value. 

As people of different roles (data scientists, data engineers, DevOps, AppDevs, etc.) need introductions to Feast and how it fits in the AI/ML flow, I've kept the example approachable, so that someone who doesn't know all of the technical disciplines can follow along and get an idea of the full picture.

The quickstart documentation and other examples are great for describing Feast functionality, but they use randomly generated data with no real patterns, and they are not as clear on where things fit in the process of doing AI/ML.

# Misc


examples/credit-risk-end-to-end/01_Credit_Risk_Data_Prep.ipynb
```diff
@@ -0,0 +1,1103 @@
...
+    "# View first records in training data (transposed)\n",
+    "train.head().T"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6825a3ed-2495-4ded-8f5f-1c64041ba5ac",
+   "metadata": {},
+   "source": [
+    "### Feature Engineering"
```
franciscojavierarceo, line 638: I don't see any actual feature engineering here, right? Or did I miss something?

dandawg: Pretty much right. In the example I explain this as we are training a baseline model, and there is opportunity to improve by doing more feature engineering. I throw away the categorical features and keep the numerical; then change the outcome from text to an indicator variable. There's certainly an opportunity to do more; I would just want to make sure we keep the simple feel.

examples/credit-risk-end-to-end/01_Credit_Risk_Data_Prep.ipynb
```diff
@@ -0,0 +1,1103 @@
...
+   "source": [
+    "### Additional Data Preparation (For Demonstrating Feast)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "dc4e7653-b118-44c3-ade3-f1b217b112fc",
+   "metadata": {},
+   "source": [
+    "At this point, it's important to bring up that Feast was developed primarily to work with production data. Feast requires datasets to have IDs and timestamps, which it uses in joins.\n",
```
lokeshrangineni, line 702: Feast requires datasets to have `IDs` and timestamps
May be better to use the word "primary key", ignore my comment if it makes sense to a data engineer, needs a bit more elaboration for new users as feast also supports composite keys. And also why feast mandates it.

examples/credit-risk-end-to-end/01_Credit_Risk_Data_Prep.ipynb
```diff
@@ -0,0 +1,1103 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "af6355dc-ff5b-4b3f-b0bd-3c4020ef67e8",
+   "metadata": {},
+   "source": [
+    "With the feature data prepared, we are ready to setup and deploy the feature store. \n",
+    "\n",
+    "Continue with the [02_Credit_Risk_Model_Training.ipynb](02_Credit_Risk_Model_Training.ipynb) notebook."
```
lokeshrangineni, line 1078: This link is not working for me when i run the notebook on local and also on RHOAI platform.

examples/credit-risk-end-to-end/02_Deploying_the_Feature_Store.ipynb
```diff
@@ -0,0 +1,913 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "2c395ca1-5786-4cee-87a9-40655587ef3c",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# For a clean start (if you have already run this notebook before)\n",
+    "# uncomment and run the following lines to purge the files\n",
```
lokeshrangineni, line 71: code is already uncommented.

examples/credit-risk-end-to-end/02_Deploying_the_Feature_Store.ipynb
```diff
@@ -0,0 +1,913 @@
...
+    "    end_date=dt.datetime(2024,1,7,12,0,0)\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "c053fd7b-336d-4286-a3aa-80efee43d2d6",
+   "metadata": {},
+   "source": [
+    "Now, we can query the SQLite database again and see that we get data in the response! Unfortunately, the data is stored in binary strings."
```
lokeshrangineni, line 714: Should we show this step? Feast internally abstracts this based on the underlying store. Also IMO may not be helpful when the data is binary as a feast user.

examples/credit-risk-end-to-end/03_Credit_Risk_Model_Training.ipynb
```diff
@@ -0,0 +1,765 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Using outcome_timestamp as the event timestamp. To specify a column explicitly, please name it event_timestamp.\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Pull the test data\n",
+    "test_labels = pd.read_parquet(\"Feature_Store/Data/test_y.parquet\")\n",
```
lokeshrangineni, line 520: this step failed on RHOAI, case sensitive - 
```suggestion
    "test_labels = pd.read_parquet(\"Feature_Store/data/test_y.parquet\")\n",
```

examples/credit-risk-end-to-end/03_Credit_Risk_Model_Training.ipynb
```diff
@@ -0,0 +1,765 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "fc5be519-7733-449b-8dc3-411e86371315",
+   "metadata": {},
+   "source": [
+    "This notebook assumes that you have already done the following:\n",
+    "\n",
+    "1. Run the [01_Credit_Risk_Data_Prep.ipynb](01_Credit_Risk_Data_Prep.ipynb) notebook to prepare the data.\n",
+    "2. Run the [02_Deploying_the_Feature_Stores.ipynb](02_Deploying_the_Feature_Stores.ipynb) notebook to configure the feature stores and launch the feature store servers.\n",
```
lokeshrangineni, line 102: for some reason this link seems to be not working for me. If it works for you ignore my comment.

examples/credit-risk-end-to-end/02_Deploying_the_Feature_Store.ipynb
```diff
@@ -0,0 +1,913 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": []
+  },
+  {
+   "cell_type": "markdown",
+   "id": "d2849a99-0193-4f6f-8fb5-3c667c091499",
+   "metadata": {},
+   "source": [
+    "### Server Shutdown"
```
lokeshrangineni, line 827: I m also divided, WDYT of moving this section to separate notebook and keep at the end. short link here. so that if someone is running the entire jupyter notebook cells online/offline server won't shutdown.

dandawg: Hmm. I kind of like this idea. Maybe a "Cleanup" notebook.  With commands to shutdown the servers and get rid of the files that were created. As it is, it could be a little hard to remember to go back to notebook 2 and shutdown.

examples/credit-risk-end-to-end/01_Credit_Risk_Data_Prep.ipynb
```diff
@@ -0,0 +1,1103 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "dc4e7653-b118-44c3-ade3-f1b217b112fc",
+   "metadata": {},
+   "source": [
+    "At this point, it's important to bring up that Feast was developed primarily to work with production data. Feast requires datasets to have IDs and timestamps, which it uses in joins.\n",
+    "\n",
+    "In a real loan application scenario, the application fields (in a database) would be associated with a timestamp, while the actual loan outcome (label) would be determined much later and recorded separately with a different timestamp.\n",
+    "\n",
+    "In order to demonstrate Feast capabilities, such as point-in-time joins, we will mock IDs and timestamps for this data. For IDs, we will use the original dataframe index values. For the timestamps, we will generate random values between \"Tue Sep 24 12:00:00 2024\" and \"Wed Oct  9 12:00:00 2024\".\n",
```
dmartinol, line 706: I think it's 2023 instead of 2024

examples/credit-risk-end-to-end/01_Credit_Risk_Data_Prep.ipynb
dmartinol: I confirm, it should be [02_Deploying_the_Feature_Store](02_Deploying_the_Feature_Store.ipynb)

examples/credit-risk-end-to-end/02_Deploying_the_Feature_Store.ipynb
```diff
@@ -0,0 +1,913 @@
...
+      ]
+     },
+     "execution_count": 33,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "response"
```
dmartinol, line 620: Here's a function to convert it to a DataFrame for better readability
```python
import pandas as pd
import json
def response_to_df(response):
  data = json.loads(response[len(response) -1])
  feature_names = data['metadata']['feature_names']
  df = pd.DataFrame(columns=feature_names)
  for i, result in enumerate(data['results']):
      row = [
          value if status == "PRESENT" else None
          for value, status in zip(result['values'], result['statuses'])
      ]
      df[feature_names[i]] = row
  df['application_timestamps'] = data['results'][0]['event_timestamps']
  return df
```

examples/credit-risk-end-to-end/04_Credit_Risk_Model_Serving.ipynb
```diff
@@ -0,0 +1,960 @@
...
+       "13               1.0              4.0  61.0  "
+      ]
+     },
+     "execution_count": 6,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "# Transform JSON into dataframe\n",
```
dmartinol, line 758: can't you use the same in notebook 03 to display the response of get-online request? it's definitely much more readable as a DF

dmartinol:
👏 👏 👏

examples/credit-risk-end-to-end/01_Credit_Risk_Data_Prep.ipynb
franciscojavierarceo: I would recommend adding even a trivial feature calculation to highlight it.

examples/credit-risk-end-to-end/02_Deploying_the_Feature_Store.ipynb
```diff
@@ -0,0 +1,913 @@
...
+     "output_type": "stream",
+     "text": [
+      "10/14/2024 09:03:34 PM root WARNING: _list_feature_views will make breaking changes. Please use _list_batch_feature_views instead. _list_feature_views will behave like _list_all_feature_views in the future.\n",
+      "10/14/2024 09:03:34 PM root WARNING: Cannot use sqlite_vec for vector search\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Online server query\n",
+    "cmd = \"\"\"http://localhost:6566/get-online-features \\\n",
```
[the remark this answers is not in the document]
dandawg: My intent was to explicitly show that the online db is empty at this point. I agree though that this is probably looking under the hood too much. I'll take it out.

examples/credit-risk-end-to-end/02_Deploying_the_Feature_Store.ipynb
dandawg: I'd like to keep it because it provides a bookend to the first db query (input 6). I did add a sentence explaining why the data is in this format, and I think this helps the user visualize some of the optimization that is happening with online stores.

examples/credit-risk-end-to-end/03_Credit_Risk_Model_Training.ipynb
dandawg: Good catch. I changed the name slightly and this needed to be updated.

examples/credit-risk-end-to-end/01_Credit_Risk_Data_Prep.ipynb
dandawg: Good catch.

examples/credit-risk-end-to-end/01_Credit_Risk_Data_Prep.ipynb
dandawg: I added a couple of ordinal features (converted from categorical columns)

examples/credit-risk-end-to-end/01_Credit_Risk_Data_Prep.ipynb
dandawg: I added a sentence to explain this.

examples/credit-risk-end-to-end/01_Credit_Risk_Data_Prep.ipynb
dandawg: Fixed.

examples/credit-risk-end-to-end/02_Deploying_the_Feature_Store.ipynb
dandawg: Fixed.

examples/credit-risk-end-to-end/02_Deploying_the_Feature_Store.ipynb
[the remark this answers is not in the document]
dandawg: Done.

examples/credit-risk-end-to-end/03_Credit_Risk_Model_Training.ipynb
dandawg: Fixed.

examples/credit-risk-end-to-end/01_Credit_Risk_Data_Prep.ipynb
dandawg: Fixed.

examples/credit-risk-end-to-end/02_Deploying_the_Feature_Store.ipynb
dandawg: I've created a cleanup notebook. It ended up being a little more complicated than I expected, in particular for cleaning up the server processes. Shutting down the processes in a separate notebook was leading to leftover zombie procs. To remedy this, I had to change the commands for setting up the servers slightly.

examples/credit-risk-end-to-end/04_Credit_Risk_Model_Serving.ipynb
dandawg: I did two things. I kept the `request.post` query, but shortened the showing of the output (first few lines, just to show this route). I also added the SDK `get_online_features` call, to show getting features that way (and as you mentioned, the nicer output for notebook display).

I felt it good to still mention the `request.post` route because AppDevs will prefer this in some use cases. From an app perspective, it can make sense to call the features REST API endpoint, then the model REST API endpoint, all without needing to import data science libraries (like sklearn and feast).

examples/credit-risk-end-to-end/02_Deploying_the_Feature_Store.ipynb
```diff
@@ -0,0 +1,825 @@
...
+    "    schema=[\n",
+    "        Field(name=\"residence_since\", dtype=Float32),\n",
+    "        Field(name=\"age\", dtype=Float32),\n",
+    "        Field(name=\"existing_credits\", dtype=Float32),\n",
+    "        Field(name=\"num_dependents\", dtype=Float32),\n",
+    "        Field(name=\"housing_ord\", dtype=Float32)\n",
+    "    ],\n",
+    "    source=data_b\n",
+    ")\n",
+    "features_test = FeatureView(\n",
```
tokoko, line 233: Wouldn't it be more realistic to do train/test split after `get_historical_features`? a separate FV for a test dataset might be more misleading/confusing for someone trying to learn feast.

franciscojavierarceo: +1

franciscojavierarceo: I would say that's not the pattern we want users to adopt

examples/credit-risk-end-to-end/README.md
```diff
@@ -0,0 +1,45 @@
+
+![Feast_Logo](https://raw.githubusercontent.com/feast-dev/feast/master/docs/assets/feast_logo.png)
+
+# Feast Credit Risk Classification End-to-End Example
+
+This example starts with an [OpenML](https://openml.org) credit risk dataset, and walks through the steps of preparing the data, setting up feature store resources, and serving features; this is all done inside the paradigm of an ML workflow, with the goal of helping users understand how Feast fits in the progression from exploratory data analysis, to model training and model serving.
+
+The example is organized in four notebooks:
```
accorvin, line 8: This should say `five` notebooks.

examples/credit-risk-end-to-end/02_Deploying_the_Feature_Store.ipynb
dandawg: Here's the issue I'm seeing. A test set should be split out before EDA and feature engineering occurs to prevent data snooping bias (see [Hands-On Machine Learning with Scikit-Learn, Keras, and TensorFlow, 2nd Ed](https://www.amazon.com/Hands-Machine-Learning-Scikit-Learn-TensorFlow/dp/1492032646) page 51). This is actually pretty important. (I didn't make up data snooping bias -- it's an industry term, just in case anyone was wondering.)

However, to `get_historical_features`, you have presumably already done EDA and feature engineering on the underlying data so that you could set up your FeatureViews. Splitting out test data after that would not be recommended.

For this example, I could change the 'test' set to a 'validation' set (or a cross-validation paradigm), which would be split out after feature engineering. (I think this is in line with what you are thinking). In this case, I would just leave the true test set concept out.

We probably need to come up with a pattern for how Feast deals with a true test set scenario. I'll ping in the Feast chat about this.

tokoko: Oh, I missed the encoder part in the first notebook. I'm no data scientist, but fortunately I have read that page :laughing:. tbh, I dislike the example even more now lol.

Correct me if I'm wrong, but I think there's a difference between the kind of feature engineering that happens before feature store and the one that happens after. Most of the features a data scientist would find in an existing feature store while starting to work a problem will be either "raw" descriptive features or features that have been previously crafted by others for other use cases. true, they are probably a product of an EDA, but done by someone else a long time ago. Also, they are never sitting right next to one another like in this example, so it would be physically impossible to do your model-specific EDA before `get_historical_features`. You would realistically call `get_historical_features` on all entities, do a train/test split and a further EDA would come after that.

Another problem with applying OrdinalEncoder in this order is that although you're correctly fitting it only on the train dataset and applying on the test, the example then goes on to store calculated feature values in feast effectively discarding information stored inside fitted OrdinalEncoder object (in case of OrdinalEncoder, mapping between strings and ints). Practically speaking, even if you find the trained model satisfactory, you won't be able to deploy it as you won't be able to apply an encoder to new values that will come to you during inference.

I guess what I'm trying to say is that not all kinds of feature engineering need to feed into the feature store and encoders fitted to a model-specific training sets are a clear example. They should come after `get_historical_features` and more importantly they should be bundled into the resulting model if the goal is for the end product to be deployable (think of something like a sklearn pipeline here).

dandawg: > Correct me if I'm wrong, but I think there's a difference between the kind of feature engineering that happens before feature store and the one that happens after.

Right, notebook 01 starts in the context of "feature engineering before feature store". 

> true, they are probably a product of an EDA, but done by someone else a long time ago.

Unfortunately, bias is transitive--this is what makes it really hard to fix generalization error. 

> it would be physically impossible to do your model-specific EDA before get_historical_features.

This is routine data science -- munging data together. It's done all the time outside of Feast.

> you won't be able to deploy it as you won't be able to apply an encoder to new values that will come to you during inference.

Actually, this is pretty common. With Sklearn, you pickle the encoder with the model in an [inference pipeline](https://scikit-learn.org/1.5/modules/generated/sklearn.pipeline.Pipeline.html), on which you call "predict" just like a simple model.

Now, if I'm understanding right, you would like to see the example flow like this:

- Feast data sources point to raw data
- "training" data pulled (get_historical_features), split, and EDA and final feature engineering completed. Then model trained and validated.
- serving data (get_online_features) fed to inference pipeline to get prediction.

I can refactor the example in this way. Let me know what you thin @tokoko .

dandawg: I was just noticing that none of the other Feast examples (in the repo, or in Feast.dev parent) even mention a validation dataset. The one's that have models do model.fit, and then go straight to serving. 

As we've just discussed, it's not trivial for a user to determine where validation and test sets come in when using Feast. Just getting a validation set in this example, will be a big win for the examples section.

tokoko: yeah, that's right. at least that's the flow that feels most intuitive for me.

- split happens after get_historical_features.
- final part of feature engineering (stuff like OrdinalEncoder, StandardScaler and so on) are applied after split.

franciscojavierarceo: > Correct me if I'm wrong, but I think there's a difference between the kind of feature engineering that happens before feature store and the one that happens after. 

In the short term, I think this is true. As I mentioned today, I do believe it'd be ideal to get Feast to a state where all transformations happen in Feast and model inference can operate on a DGBMatrix, Torch Tensor, or numpy matrix. It turns out that a lot of skew can happen during matrix serialization and getting that centralized in Feast is a worth endeavor.

franciscojavierarceo: > Actually, this is pretty common. With Sklearn, you pickle the encoder with the model in an inference pipeline, on which you call "predict" just like a simple model.

Yes, this entire pipeline can introduce latency for lots of production use cases and miscellaneous skew issues. It's why I said I think long term we should work to move these things in Feast.

franciscojavierarceo: > Since the validation/test is how a data scientist makes decisions and assesses model performance, its really important--and features play a role here.

I agree it's important. It should also be a simple post processing step after `get_online_features`. People cut their training/test/validation sets differently depending upon their business requirements. That's, of course, something we could implement during some sort of Data Version Control process but we'd need to build that.

examples/credit-risk-end-to-end/01_Credit_Risk_Data_Prep.ipynb
franciscojavierarceo: As @tokoko mentioned I think we should use a different featurization example as one hot encoding or ordinal encoding often happens outside of Feast during an inference pipeline (though we should build for this in the future to handle the important implementation detail of new categories for online serving). 

I think a ratio or product of two other features is sufficient to highlight the goal.

franciscojavierarceo (changes requested):
Let's adjust based on the feedback provided

examples/credit-risk-end-to-end/02_Deploying_the_Feature_Store.ipynb
dandawg: Alright, this has been good feedback. @tokoko I reordered the steps according to our discussion above, and I renamed the 'test' set to a 'validate' set (not a big issue, but just to avoid any confusion). I was able to simplify the notebook flow a bit as well.

dandawg: Just checking back. I believe this is the only outstanding question on this PR. With the changes I made, is this ready to be merged?

dandawg: Checking one more time. I believe this question has been resolved.

dandawg:
@franciscojavierarceo I've adjusted this PR based on the feedback given. I believe it's ready to merge.

examples/credit-risk-end-to-end/04_Credit_Risk_Model_Serving.ipynb
```diff
@@ -0,0 +1,697 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 9,
+   "id": "70203f7b-f1e5-46ba-8623-f10bf3a5abf8",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Get predictions from the model\n",
+    "preds = model.predict(records)"
```
franciscojavierarceo, line 480: The only nit I have here is that probably it'd be good to have the model serving to be a separate endpoint, but we can leave that for another time.

franciscojavierarceo (approved):

infra/feast-operator/dist/install.yaml
```diff
@@ -349,7 +349,7 @@ spec:
         - --leader-elect
         command:
         - /manager
-        image: feastdev/feast-operator:0.40.0
+        image: feastdev/feast-operator:0.41.0
         livenessProbe:
```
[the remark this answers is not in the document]
dandawg: Good catch. I'm not sure how this got off--maybe from a `make test` run or something. It should be fixed now.

franciscojavierarceo (approved):

PR #4633, opened 2024-10-16
fix: Added support for multiple name patterns to Permissions

### LINKED ISSUE #4510: Permission should support multiple name patterns
As of today RBAC permissions can match resources by type and a single `name_pattern` field, as in:
```py
Permission(
    name="reader",
    types=[FeatureView],
    name_pattern=".*risky.*",
    policy=RoleBasedPolicy(roles=["trusted"]),
    actions=[AuthzedAction.QUERY_OFFLINE],
)
```

We should extend this concept by allowing multiple patterns instead, to simplify the customer configuration.

Support a list of `name_pattern`s instead:
```py
Permission(
    name="reader",
    types=[FeatureView],
    name_pattern=[".*risky.*", "critical_data"],
    policy=RoleBasedPolicy(roles=["trusted"]),
    actions=[AuthzedAction.QUERY_OFFLINE],
)
```

In this case, the permission would apply to all the feature views matching any of the configured patterns.

Do not rename the `name_pattern` field for backward compatibility.

dmartinol:
Allowing Permissions to match multiple instance by different name patterns

Fixes #4510 

@franciscojavierarceo

### DIFF
--- a/sdk/python/feast/permissions/permission.py
+++ b/sdk/python/feast/permissions/permission.py
@@ -33,7 +33,7 @@ class Permission(ABC):
         name: The permission name (can be duplicated, used for logging troubleshooting).
         types: The list of protected resource types as defined by the `FeastObject` type. The match includes all the sub-classes of the given types.
         Defaults to all managed types (e.g. the `ALL_RESOURCE_TYPES` constant)
-        name_pattern: A regex to match the resource name. Defaults to None, meaning that no name filtering is applied
+        name_patterns: A possibly empty list of regex patterns to match the resource name. Defaults to empty list, e.g. no name filtering is applied
         be present in a resource tags with the given value. Defaults to None, meaning that no tags filtering is applied.
         actions: The actions authorized by this permission. Defaults to `ALL_ACTIONS`.
         policy: The policy to be applied to validate a client request.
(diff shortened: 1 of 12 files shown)


sdk/python/tests/unit/permissions/conftest.py
```diff
@@ -79,7 +79,7 @@ def security_manager() -> SecurityManager:
         Permission(
             name="special",
             types=FeatureView,
-            name_pattern="special.*",
+            name_patterns="special.*",
```
franciscojavierarceo, line 82: Shouldn't this be a list?

sdk/python/feast/permissions/permission.py
```diff
@@ -54,8 +54,8 @@ class Permission(ABC):
     def __init__(
         self,
         name: str,
-        types: Optional[Union[list["FeastObject"], "FeastObject"]] = None,
-        name_pattern: Optional[str] = None,
+        types: Optional[Union[list["FeastObject"], "FeastObject"]] = [],
+        name_patterns: Optional[Union[str, list[str]]] = [],
```
franciscojavierarceo, line 58: I see, here's the init with list and str

franciscojavierarceo (approved):

franciscojavierarceo:
Thank you for this!

PR #4636, opened 2024-10-16
chore: Change Field.__repr__ print format

emgeee:
This PR updates the Field.__repr__() function to be inline with official python recommendations: https://docs.python.org/3/reference/datamodel.html#object.__repr__

"Called by the [repr()](https://docs.python.org/3/library/functions.html#repr) built-in function to compute the “official” string representation of an object. If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment)."

# Misc

### DIFF
--- a/sdk/python/feast/field.py
+++ b/sdk/python/feast/field.py
@@ -81,7 +81,7 @@ class Field:
         return self.name < other.name
 
     def __repr__(self):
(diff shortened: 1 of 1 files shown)


franciscojavierarceo (approved):
lgtm, welcome to Feast!

PR #4244, opened 2024-05-31
feat: Add online_read_async for dynamodb

robhowley:
# What this PR does / why we need it:

* adds aiobotocore to aws dependencies
* adds online_read_async impl for dynamdb online reader
* the use of the standard blocking boto client prevents the feast sdk from being able to handle real time web traffic
* most of the diff is in the py3.X-[ci-]requirements.txt files

# Which issue(s) this PR fixes:

# Fixes

Issue #4401, opened 2024-08-13
Add a close() method to the PostgreSQLOnlineStore


job-almekinders:
**Is your feature request related to a problem? Please describe.**
When running the online store as a web app, I want to make sure that the connections made to the Postgres database using the PostgreSQLOnlineStore class are terminated gracefully.

**Describe the solution you'd like**
A close() method on the PostgreSQLOnlineStore which closes all open connections and connection pools.

**Describe alternatives you've considered**

**Additional context**

tokoko:
What's the point of adding this to `PostgreSQLOnlineStore` if it's not part of `OnlineStore`? How are you thinking of calling it?

job-almekinders:
> What's the point of adding this to `PostgreSQLOnlineStore` if it's not part of `OnlineStore`? How are you thinking of calling it?

That is a valid point. It is possible as follows: `feature_store._provider.online_store.close()`. Where `feature_store` is an instance of the `FeatureStore` class. However, this is not ideal, given the fact that the `provider` object is private.

Do you think it would be an option to add a `close()` method on the `OnlineStore` class? This would also open the possibility for other online stores to implement a similar method. 
We could do a similar approach as for the `retrieve_online_documents` function [here](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/online_stores/online_store.py#L341-L375). By not making it a mandatory function, but an optional one. 

Let me know what you think! @tokoko

tokoko:
public/private issue can be remedied in some way, I suppose. I think my main concern here is that I don't fully understand what a `close` method on an abstract `OnlineStore` class should mean and how/when it could be used. The implementation details would naturally be left up to individual online stores, but what should be the result of closing an online store? Would it make online store unusable from that point on in that process so that subsequent calls may throw an exception? Or would it simply clean up resources held to save on utilization?

Also just an aside, what's the actual practical use case here? I couldn't really come up with an actual use case where I would need to clean up connections but for some reason keep the process running.

job-almekinders:
> Would it make online store unusable from that point on in that process so that subsequent calls may throw an exception? Or would it simply clean up resources held to save on utilization?

This is IMO up for debate. I think the second option would be the best here. Such that all connections are closed, but if the same object is used again, it would re-create them and work again.

I think the current implementation would honor the second approach, but I would have to test/double check that to be sure.

> Also just an aside, what's the actual practical use case here?

[...]

PR #4642, opened 2024-10-17
perf: Add init and cleanup of long lived resources

robhowley:

* add `initialize` and `close` to `FeatureStore`, `Provider`, `OnlineStore` ~~and `OfflineStore`~~
* they are async and default impl does nothing
* provides a hook for datastores like dynamodb which needs to do some setup in order to create reusable clients and sessions
* added the dynamo setup

_update_: added `pytest-asyncio` to ci which is +555 of the diff

Addresses the performance penalty mentioned [here](https://github.com/feast-dev/feast/pull/4244#discussion_r1623274405) by @tokoko 

# Misc


sdk/python/feast/feature_server.py
```diff
@@ -100,8 +100,10 @@ def async_refresh():
     @asynccontextmanager
     async def lifespan(app: FastAPI):
         async_refresh()
+        await store.initialize()
         yield
         stop_refresh()
+        await store.close()
```
robhowley, line 106: can create and cleanup resources as part of the feature server lifespan

robhowley:
not 100% what to make this error in the py39 unit tests

```
ERROR sdk/python/tests/unit/permissions/auth/server/test_auth_registry_server.py::test_registry_apis[\nauth:\n  type: no_auth\n-applied_permissions0] - RuntimeError: Failed to bind to address [::]:34361; set GRPC_VERBOSITY=debug environment variable to see detailed error message.
```

franciscojavierarceo:
@dmartinol @redhatHameed  could you take a look?

dmartinol:
> not 100% what to make this error in the py39 unit tests
> 
> ```
> ERROR sdk/python/tests/unit/permissions/auth/server/test_auth_registry_server.py::test_registry_apis[\nauth:\n  type: no_auth\n-applied_permissions0] - RuntimeError: Failed to bind to address [::]:34361; set GRPC_VERBOSITY=debug environment variable to see detailed error message.
> ```

I cloned your repo and it succeeded with no issues, maybe a temporary env issue?

dmartinol:
Some questions about this change:
- why adding the init/close methods to `offline_store` if there is no implementation yet? 
- if you really think that `offline_store` deserves them, you probably need to invoke them on the `store` instance in `offline_store` module as well. Probably the `feature_server`, which actually is more an `online_server`, will not have an `offline_store` at all, or may use a remote one.

@franciscojavierarceo @tokoko

robhowley:
> Some questions about this change:
> 
> * why adding the init/close methods to `offline_store` if there is no implementation yet?
> * if you really think that `offline_store` deserves them, you probably need to invoke them on the `store` instance in `offline_store` module as well. Probably the `feature_server`, which actually is more an `online_server`, will not have an `offline_store` at all, or may use a remote one.
> 
> @franciscojavierarceo @tokoko

yea that's a good point. offline store doesn't make a ton of sense in the server case. was just trying to maintain consistency across the two, but that need not be the case. i'll remove.

tokoko:
@robhowley thanks, I'm not too sure about the changes though. can you elaborate on why we need this?

Is the client creation actually that expensive that it needs a separate init method? I think what I was pointing out in the comment you linked above was that `create_client` was being called on every invocation rather than only during the first one and cached afterwards. The first one being slightly slower than others doesn't seem that bad to me.

I get why a separate initializer can be beneficial, but I'm a little uneasy that it pushes us into the mixture of sync/async methods. What if I'm invoking the online read the old-fashioned way w/o asyncio, do I still need to call async initializer first or is that only relevant for async calls?

The same goes for the close method. This has actually come up before in #4401. As I pointed out there, I'm not sure what closing an online store should actually mean. Do calls on a closed online store error out or do they still succeed, but need to recreate the resources? I'd rather not have to answer those questions at all unless you think we absolutely need to.

robhowley:
With a little time to mull this over, I'm curious what everyone's thoughts are on this. I'm not attached to the open/close idea, but it does seem like there should be a way to clean up resources. Most of async support will come from libraries expecting to be used in an aync context manager, so some kind of equivalent mechanism should be available.

tokoko:
@robhowley how would the normal non-async workflow be affected with this? If we put connection creation in an async initializer that won't be called by a non-async user... would we implicitly run an initializer anyway during the first call?

robhowley:
The sync usage shouldn't be impacted. The async code paths need not be touched by the sync code, so I don't think any async initialization would be needed/called.

```
# in sdk mode, this shouldn't touch async code
# and as a result shouldn't require any async connections to be created
feature_df = FeatureStore().get_online_features(entities).to_df()
```

robhowley:
@franciscojavierarceo @tokoko im not exactly sure what to make of the failing integration tests. created an issues #4658 bc they don't seem to have passed in better part of two months. im assuming you're aware, just (1) thought it worth flagging and (2) curious if you had any opinions on how concerned i should be about it

robhowley:
@tokoko @franciscojavierarceo btw, some added context on this mr ... i've been running this in prod for a few days and reusing the client reduces event loop lag in the server since creation of the client is a blocking operation. the scalability and speed improvements are obviously quite good as a result. reusing the client, however, should be paired w the ability to clean up the resources.

robhowley:
would it perhaps be clearer to just go full async context manager? somewhat more explicit that this isnt going to impact the sync code paths?

sdk/python/feast/feature_server.py
```diff
@@ -101,8 +101,10 @@ def async_refresh():
     @asynccontextmanager
     async def lifespan(app: FastAPI):
         async_refresh()
+        await store.initialize()
```
franciscojavierarceo, line 104: I wonder if this is the root cause

robhowley: root cause of what?

sdk/python/feast/feature_server.py
franciscojavierarceo: Can you remove this and let's rerun the integration test? Or maybe create a separate branch without it?

sdk/python/feast/feature_server.py
robhowley: for all but dyanmo this method call is just `pass`

robhowley: it failed integration tests [bc the dynamo client is null](https://github.com/feast-dev/feast/actions/runs/11486758859/job/31969747403?pr=4642#step:15:2352) during the test using the server

franciscojavierarceo: i should have been more explicit, shouldn't initialization happen before the refresh? I took a quick look at the code but wasn't sure.

robhowley: they don't really interact w each other. does read like initialize should be first, but it only creates async clients, doesn't deal w the registry

robhowley:
@franciscojavierarceo just a heads up, [the remaining integration test failure](https://github.com/feast-dev/feast/actions/runs/11488391804/job/31975025552?pr=4642#step:15:907) comes from [here](https://github.com/feast-dev/feast/blob/master/sdk/python/tests/integration/online_store/test_python_feature_server.py#L61). the push works, but the subsequent retrieval to confirm it's there, in the case of dynamo, is async. this is a sync function so there's no running event loop. working on a fix to get this green

sdk/python/feast/feature_server.py
robhowley: swapped the order, does read better that way

sdk/python/tests/integration/online_store/test_python_feature_server.py
```diff
@@ -137,5 +135,5 @@ def python_fs_client(environment, universal_data_sources, request):
     feast_objects.extend([driver(), customer(), location()])
     fs.apply(feast_objects)
     fs.materialize(environment.start_date, environment.end_date)
-    client = TestClient(get_app(fs))
-    yield client
+    with TestClient(get_app(fs)) as client:
+        yield client
```
robhowley, line 139: [using test client as a context manager allows fastapi to incorporate lifespan](https://fastapi.tiangolo.com/advanced/testing-events/). dynamo needs this bc the connection and by extention event loop management happens at startup/shutdown. this fixed the remaining integration test issues.

sdk/python/pytest.ini
```diff
@@ -7,6 +9,8 @@ env =
     IS_TEST=True
 
 filterwarnings =
+    error::_pytest.warning_types.PytestConfigWarning
+    error::_pytest.warning_types.PytestUnhandledCoroutineWarning
```
robhowley, line 13: fail if any async test functions are skipped bc of missing plugins

franciscojavierarceo (approved):
Thank you for working through this and for shipping it!
image

Issue #4451, opened 2024-08-26
Update dependencies in sdk/python/feast/ui


elshize:
**Is your feature request related to a problem? Please describe.**

The dependencies have vulnerabilities. See #4439 for context.

**Describe the solution you'd like**

I would appreciate it if someone who knows JS stuff could upgrade those dependencies.

**Describe alternatives you've considered**

See #4439

elshize:
I'm willing to open a change for this, but I would need some guidance for how to fix vulnerabilities using yarn.

peruukki:
Most of the vulnerabilities come from [react-scripts](https://www.npmjs.com/package/react-scripts) which hasn't been updated in 2.5 years. We'll need to [eject](https://create-react-app.dev/docs/available-scripts/#npm-run-eject) to basically get rid of that dependency and start managing the setup ourselves, and then update or replace related dependencies. I'd be interested to try it out, I actually quickly tried a while ago but hit some errors.

For the rest, usually these commands go far:
```sh
# Upgrade multiple dependencies selectively at once
yarn upgrade-interactive

# Upgrade direct dependency to latest version
yarn add [--dev] <package> # Use --dev if the dependency is in devDependencies

# Upgrade transitive dependencies of a direct dependency to latest versions
yarn upgrade <package>
``` 

Also `yarn why <package>` is useful for finding out why a package is in the project (which packages depend on it).

These are the direct dependencies we'd need to upgrade based on `yarn audit`:
```sh
% yarn audit | grep 'Dependency of' | sort | uniq
│ Dependency of │ @babel/core                                                  │
│ Dependency of │ @babel/preset-env                                            │
│ Dependency of │ msw                                                          │
│ Dependency of │ protobufjs-cli                                               │
│ Dependency of │ react-code-blocks                                            │
│ Dependency of │ react-scripts                                                │
│ Dependency of │ rollup-plugin-copy                                           │
│ Dependency of │ rollup-plugin-svgo                                           │
```

I'd be happy to try upgrading the others too, and would start from them.

PR #4645, opened 2024-10-17
chore: Upgrade react-code-blocks to 0.1.6 in Feast UI

peruukki:

This is the only clear runtime app dependency (at least to me 😄) that had vulnerabilities listed in `yarn audit` output. Upgrading it to the latest version removes 4 high and 5 moderate level vulnerabilities.

Relates to a small part of #4451.

# Misc

We only use `react-code-blocks` in one place, and only one component of it, `CopyBlock`, but I wasn't able to make it show when running Feast UI locally with `yarn start` to verify that it works and looks OK. But the needed changes were so small and clear that I feel quite confident about the upgrade.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -31,7 +31,7 @@
     "moment": "^2.29.1",
     "protobufjs": "^7.1.1",
     "query-string": "^7.1.1",
-    "react-code-blocks": "^0.0.9-0",
+    "react-code-blocks": "^0.1.6",
     "react-query": "^3.39.3",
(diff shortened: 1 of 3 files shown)


franciscojavierarceo (approved):

peruukki:
A couple integration tests had failed, so I rebased on master to see if that helps them pass.

peruukki:
Thanks @franciscojavierarceo for re-running the failed integration tests and making them pass. 🙇

PR #4647, opened 2024-10-18
perf: Default to async endpoints, use threadpool for sync

### LINKED ISSUE #4617: Add support for async path operations in the server

Add opt in support for async path operations on the python feature server. This would allow the server to take advantage of the async libraries that are slowly becoming available for the various data stores. Lookups to multiple feature views with online stores such as dynamo and redis can be done in parallel. This would provide significant performance gains when calling feature services.

If we are willing to move to async methods for path operations, i'd like to to see `run_in_threadpool` used for the sync methods. [example pr here](https://github.com/robhowley/feast/pull/8). This is the recommended approach to sync operations in async methods and is more or less what fastapi does under the hood to sync endpoints anyway.

If there is a preference for a manual opt in, then ...

A cli parameter on `feast serve` to indicate preference for deploying the async paths which will leverage the async feature store sdk methods.

```
# simple flag param
feast serve --async-read
```

Under the hood it would declare

```
@app.post(
    "/get-online-features",
    dependencies=[Depends(inject_user_details)],
)
async def get_online_features(body=Depends(get_body)):
    ...
    # and call
    response_proto = await store.get_online_features(
```

As async `store.push` functionality gets added an `--async-push` flag could be added as well. This approach is an explicit opt-in which is desirable when overriding default behavior. Lets the person deploying the server make informed decisions about performance based upon the backend being used. The caller need not worry about the internals of a service, just the api contract.

* a separate cli `feast-async` but 
  * it's nice having `feast` be the one universal cli
  * it implies everything will be async, which it wouldn't be
* a cli subgroup `feast serve async`
  * adds more layers and indirection in the code without a lot of value
* new base class AsyncOnlineDataStore to differentiate from the default synchronous calls
  * when the server is constructed it will define the path operation based upon type
  * this would require a nontrivial refactor

**Additional context**

Here is a flame graph for a call to a feature service with

* 3 feature views
* 11 total features
* dynamo backend

image

We see default batching behavior and 3 calls per batch all of which are done sequentially. These can be done in parallel.

robhowley:

* add property to the `OnlineStore` indicating if async reads/writes are supported
* make the `/get-online-features` endpoint async
* use the async supported flags to make the appropriate `store.get_online_features[_async]` call
  * if async not supported, runs `store.get_online_features` in fastapi's threadpool helper

Fixes #4617 

# Misc

### DIFF
--- a/sdk/python/feast/feature_server.py
+++ b/sdk/python/feast/feature_server.py
@@ -10,6 +10,7 @@ import pandas as pd
 import psutil
 from dateutil import parser
 from fastapi import Depends, FastAPI, Request, Response, status
+from fastapi.concurrency import run_in_threadpool
 from fastapi.logger import logger
 from fastapi.responses import JSONResponse
 from google.protobuf.json_format import MessageToDict
@@ -100,10 +101,8 @@ def get_app(
     @asynccontextmanager
     async def lifespan(app: FastAPI):
         async_refresh()
-        await store.initialize()
         yield
         stop_refresh()
-        await store.close()
 
     app = FastAPI(lifespan=lifespan)
 
@@ -114,7 +113,7 @@ def get_app(
         "/get-online-features",
         dependencies=[Depends(inject_user_details)],
     )
-    def get_online_features(body=Depends(get_body)):
+    async def get_online_features(body=Depends(get_body)):
         body = json.loads(body)
         full_feature_names = body.get("full_feature_names", False)
         entity_rows = body["entities"]
@@ -147,15 +146,22 @@ def get_app(
                     resource=od_feature_view, actions=[AuthzedAction.READ_ONLINE]
                 )
 
-        response_proto = store.get_online_features(
+        read_params = dict(
             features=features,
             entity_rows=entity_rows,
             full_feature_names=full_feature_names,
-        ).proto
+        )
+
+        if store._get_provider().async_supported.online.read:
+            response = await store.get_online_features_async(**read_params)
+        else:
+            response = await run_in_threadpool(
+                lambda: store.get_online_features(**read_params)
+            )
 
         # Convert the Protobuf object to JSON and return it
         return MessageToDict(
-            response_proto, preserving_proto_field_name=True, float_precision=18
+            response.proto, preserving_proto_field_name=True, float_precision=18
         )
 
     @app.post("/push", dependencies=[Depends(inject_user_details)])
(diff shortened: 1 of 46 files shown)


sdk/python/feast/feature_server.py
```diff
@@ -112,7 +113,7 @@ async def get_body(request: Request):
         "/get-online-features",
         dependencies=[Depends(inject_user_details)],
     )
-    def get_online_features(body=Depends(get_body)):
+    async def get_online_features(body=Depends(get_body)):
```
robhowley, line 116: `/get-online-features` endpoint is now async

sdk/python/feast/feature_server.py
```diff
@@ -145,15 +146,22 @@ def get_online_features(body=Depends(get_body)):
...
             full_feature_names=full_feature_names,
-        ).proto
+        )
+
+        if store._get_provider().async_supported.online.read:
+            response = await store.get_online_features_async(**read_params)
+        else:
+            response = await run_in_threadpool(
+                lambda: store.get_online_features(**read_params)
+            )
```
robhowley, line 160: use async feature lookups if it's supported by the online store, otherwise run the sync method in the threadpool

sdk/python/feast/infra/online_stores/online_store.py
```diff
@@ -36,6 +37,10 @@ class OnlineStore(ABC):
     The interface that Feast uses to interact with the storage system that handles online features.
     """
 
+    @property
+    def async_supported(self) -> SupportedAsyncMethods:
+        return SupportedAsyncMethods()
```
robhowley, line 42: by default an online store _does not_ have async methods.

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -88,6 +89,10 @@ class DynamoDBOnlineStore(OnlineStore):
     _dynamodb_resource = None
     _aioboto_session = None
 
+    @property
+    def async_supported(self) -> SupportedAsyncMethods:
+        return SupportedAsyncMethods(read=True)
```
robhowley, line 94: an online store can opt in to using async read/write calls as they become available.

franciscojavierarceo (approved):
Nice!

robhowley:
the `build-docker-image-java (feature-server-java)` looks stuck, can someone kill it and rerun, please? 🙏

franciscojavierarceo:
> the `build-docker-image-java (feature-server-java)` looks stuck, can someone kill it and rerun, please? 🙏
> 
> fyi @franciscojavierarceo

Actually it's a manual thing, I have to trigger the tests.

robhowley:
ohhhh my apologies. not super familiar w how the pipeline is setup. thanks!

PR #4649, opened 2024-10-18
fix: Bigquery dataset create table disposition

### LINKED ISSUE #4648: Feast will attempt to create a BigQuery dataset regardless of `table_create_disposition`
## Expected Behavior 
When `table_create_disposition` is set to `CREATE_NEVER`, a BigQuery dataset should not be created.

## Current Behavior
`table_create_disposition` is ignored and feast will attempt to create the dataset regardless. This can be a problem for orgs who require to manage warehouse infrastructure outside of Feast.

## Steps to reproduce
Run `get_historical_features` on any BigQuery offline store. Example code:

```
from feast import FeatureStore
from pathlib import Path
from datetime import datetime
import pandas as pd

config = Path("feature_store.yaml")
store = FeatureStore(fs_yaml_file=config)
training_df = store.get_historical_features(
    entity_df=pd.DataFrame.from_dict(
        {
            "id": ["<uid>"],
            "feature_timestamp": [datetime(2020, 1, 1)]
        }
    ),
    features=[
        "<feature_view>:<feature>"
    ]
)

print(training_df.to_df())
```

### Specifications

- Version: feast[gcp,aws]==0.40.1
- Platform: python 3.10.14
- Subsystem: Debian GNU/Linux 11 (bullseye)

## Possible Solution
`table_create_disposition` should be checked in [this method](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/offline_stores/bigquery.py#L668). If the dataset does not exist and exception should be thrown when `table_create_disposition` is `CREATE_NEVER`.

danbaron63:

* Updates `get_historical_features` to not create a BigQuery dataset when `create_table_disposition` is set to `CREATE_NEVER`.
* Cleans up or ignores some linting issues preventing pushing due to pre-commit checks.

Fixes #4648 

# Misc

This PR brings BigQuery dataset creation in line with table creation - this will hopefully be useful to orgs (such as mine!) that have stricter requirements over what can/cannot make infrastructure/warehouse changes in production environments. In our case `get_historical_features` is expected to be read only operation.

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery.py
@@ -242,6 +242,7 @@ class BigQueryOfflineStore(OfflineStore):
             dataset_project,
             config.offline_store.dataset,
             config.offline_store.location,
+            config.offline_store.table_create_disposition,
         )
 
         entity_schema = _get_entity_schema(
@@ -670,6 +671,7 @@ def _get_table_reference_for_new_entity(
     dataset_project: str,
     dataset_name: str,
     dataset_location: Optional[str],
+    table_create_disposition: str,
 ) -> str:
     """Gets the table_id for the new entity to be uploaded."""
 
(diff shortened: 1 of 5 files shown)


tokoko:
@danbaron63 I get the issue, but how would this change help? with or without this additional config, `get_historical_features` will still fail unless you prepopulate the dataset, right?

danbaron63:
> @danbaron63 I get the issue, but how would this change help? with or without this additional config, `get_historical_features` will still fail unless you prepopulate the dataset, right?

Hey @tokoko, this addresses the issue of creating cloud resources during what should be a read only operation when `create_table_disposition` is appropriately configured, effectively widening the safety net offered by `create_table_disposition`. This would be useful when a user or service is misconfigured with the wrong dataset and also happens to be over-permissioned - instead of just creating the dataset it can fail gracefully.

In our org this is a serious issue ([and is actually the reason one of my colleagues authored](https://github.com/feast-dev/feast/pull/3762) `create_table_disposition` 😄) as our dataset resources are strictly managed by IaC and are not expected to be created/amended by Feast.

tokoko (approved):

PR #4650, opened 2024-10-18
perf: Make /push async

robhowley:

* make the /push endpoint async
* if pushing `to` online and async writes for the `OnlineStore` are supported, will call `await store.push`
* offline writes in the async codepaths utilize `run_in_threadpool`
* this pr only adds the calls and interface, there are no online stores that support async write
* added tests for when async vs sync calls of `store.push` happen in the server

# Misc

### DIFF
--- a/sdk/python/feast/feature_server.py
+++ b/sdk/python/feast/feature_server.py
@@ -213,6 +213,7 @@ def get_app(
             and to in [PushMode.ONLINE, PushMode.ONLINE_AND_OFFLINE]
         )
         if should_push_async:
+            print("im in async?")
             await store.push_async(**push_params)
         else:
             store.push(**push_params)
--- a/sdk/python/feast/feature_server.py
+++ b/sdk/python/feast/feature_server.py
@@ -165,7 +165,7 @@ def get_app(
         )
 
     @app.post("/push", dependencies=[Depends(inject_user_details)])
-    def push(body=Depends(get_body)):
+    async def push(body=Depends(get_body)):
         request = PushFeaturesRequest(**json.loads(body))
         df = pd.DataFrame(request.df)
         actions = []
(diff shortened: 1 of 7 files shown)


sdk/python/feast/feature_server.py
```diff
@@ -201,13 +201,22 @@ def push(body=Depends(get_body)):
...
         )
 
+        should_push_async = (
+            store._get_provider().async_supported.online.write
+            and to in [PushMode.ONLINE, PushMode.ONLINE_AND_OFFLINE]
+        )
+        if should_push_async:
+            await store.push_async(**push_params)
+        else:
+            store.push(**push_params)
```
robhowley, line 218: if async writes to the online store are supported and the post body indicates that data will be written to the online store, then attempt an `async storep.push`

sdk/python/feast/feature_store.py
```diff
@@ -1467,22 +1476,42 @@ def push(
...
+
+        if to == PushMode.ONLINE or to == PushMode.ONLINE_AND_OFFLINE:
+            _ = await asyncio.gather(
+                *[
+                    self.write_to_online_store_async(
+                        fv.name, df, allow_registry_cache=allow_registry_cache
+                    )
+                    for fv in fvs
+                ]
+            )
```
robhowley, line 1496: for async push we parallelize all the writes to the online store ...

sdk/python/feast/feature_store.py
```diff
@@ -1467,22 +1476,42 @@ def push(
...
+
+        if to == PushMode.OFFLINE or to == PushMode.ONLINE_AND_OFFLINE:
+
+            def _offline_write():
+                for fv in fvs:
+                    self.write_to_offline_store(
+                        fv.name, df, allow_registry_cache=allow_registry_cache
+                    )
+
+            await run_in_threadpool(_offline_write)
```
robhowley, line 1506: but keep the existing sync impl of offline writes except we'll put them in the threadpool so it doesn't block the server

sdk/python/tests/unit/test_feature_server.py
```diff
@@ -0,0 +1,47 @@
...
+    "online_write,push_mode,async_count",
+    [
+        (True, PushMode.ONLINE_AND_OFFLINE, 1),
+        (True, PushMode.OFFLINE, 0),
+        (True, PushMode.ONLINE, 1),
+        (False, PushMode.ONLINE_AND_OFFLINE, 0),
+        (False, PushMode.OFFLINE, 0),
+        (False, PushMode.ONLINE, 0),
+    ],
+)
```
robhowley, line 24: test ensures that the only time we attempt an `async store.push` is if online async write is supported and pushmode includes online

franciscojavierarceo (approved):
Nice

tokoko (approved):

HaoXuAI:
good feature!

Issue #4490, opened 2024-09-05
OnDemandFeatureView: Allow `python` transformations for `offline` processing. 


job-almekinders:
**Is your feature request related to a problem? Please describe.**
When running a `PythonTransformation` for offline feature retrieval, an exception is raised [here](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/transformation/python_transformation.py#L31-L33). I understand that `python` transformations on very large datasets do not scale well. However, in some cases, a user might still want to apply a `python` transformation to a smaller dataset in the offline store. Due to this error (and the missing implementation of `PythonTransformation().transform_arrow()`), this is currently not possible.

**Describe the solution you'd like**
Instead of raising an exception, we could still run the transformation, and also log a warning indicating that it may not scale well. On top of that we can implement the functionality in `transform_arrow` to make the transformation happen.

**Describe alternatives you've considered**
-

**Additional context**
-

franciscojavierarceo:
Oh we should definitely discuss this because I agree. Also, for scaling Spark should really call the Transformation as a UDF and then it can scale very well.

franciscojavierarceo:
This has been implemented. 👍

job-almekinders:
Thanks a lot for this! @franciscojavierarceo 

FYI @TomSteenbergen :)

PR #4653, opened 2024-10-19
feat: Adding mode='python' for get_historical_features on ODFVs

franciscojavierarceo:
This PR adds support for `mode="python"` for On Demand Feature Views (ODFVs) when calling `get_historical_features` by ensuring   that when `*.to_df()` is called, pyarrow executes the function.

https://github.com/feast-dev/feast/issues/4490

# Misc
@job-almekinders

### DIFF
--- a/sdk/python/feast/transformation/python_transformation.py
+++ b/sdk/python/feast/transformation/python_transformation.py
@@ -26,11 +26,11 @@ class PythonTransformation:
         self.udf_string = udf_string
 
     def transform_arrow(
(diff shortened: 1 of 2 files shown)


tokoko (approved):
lgtm. minor: we could show a warning about performance, not sure if it's necessary or not.

franciscojavierarceo:
It's a good point, I can update the docs on it but some of the feedback we've gotten is that we log too many warnings for stuff. I tend to agree.

PR #4655, opened 2024-10-20
chore: Define Node.js version with Node Version Manager in Feast UI

peruukki:

I noticed a TypeScript error in the `FeastUISansProviders.test.tsx` file in VS Code:

ts-error-registry

For whatever reason, I haven't been able to get the error appear elsewhere, e.g. when running the TypeScript compiler. Nonetheless I noticed that we still use Node.js version 16 TypeScript type definitions in the project, and updating those to the latest version 20 definitions makes the error go away. This got me thinking that we don't explicitly define what Node.js version we are supposed to be using, except in the GitHub actions.

So I decided to make a PR and define the Node.js version in a single file to make it easier to use a consistent version and upgrade it in the future. There are a couple alternatives for this, but [Node Version Manager](https://github.com/nvm-sh/nvm) is very common and one I've used for years, so I chose it.

v20.18.0 is the current latest LTS version, so let's use that. Note that the @types/node package version numbers don't match the related Node.js version numbers, 20.16.13 is the latest package in the major number 20 series.

# Misc

setup-node GitHub action node-version-file documentation: https://github.com/actions/setup-node/blob/v3.8.2/docs/advanced-usage.md


franciscojavierarceo (approved):
Lgtm

Issue #4616, opened 2024-10-11
Feast create empty online table when FeatureView attribute online=False


SuperBo:
## Expected Behavior 

Feast shouldn't create any DynamoDB table when I defined FeatureView with online=False.

## Current Behavior

Feast creates empty DynamoDB table when I defined FeatureView with online=False.

## Steps to reproduce

Define a feature store like this 

```yaml
project: example_feature_sotre

provider: aws

offline_store:
  type: spark
  spark_conf: {}

online_store:
  type: dynamodb
  region: us-east-1

entity_key_serialization_version: 2
```

Feature view def

```python
feature_view = FeatureView(
    name="abc_details_monthly",
    entities=[user_id],
    online=False,
    source=SparkSource(
      table="shcema.abc_details_monthly",
      name="abc_details_monthly",
      timestamp_field="event_date",,
      ttl=datetime.timedelta(days=27),
    # Schema definition
    schema=[
        Field(
            name="feature_1_1d",
            dtype=types.Int32,
            description=""
        )
    ]
)
```

### Specifications

- Version: 0.40.1
- Platform: Linux
- Subsystem:

## Possible Solution

Should distinguish list tables to keep of online store and offline store in this file.

https://github.com/feast-dev/feast/blob/cd87562df7f55dfe888bfbbd9a33b87670681040/sdk/python/feast/infra/passthrough_provider.py#L122

franciscojavierarceo:
Yeah, agree wasted space.

PR #4666, opened 2024-10-22
fix: Feast create empty online table when FeatureView attribute online=False

tmihalac:
Currently Feast create an empty online table even if FeatureView attribute online=False, this PR aims to fix that. 

Fixes https://github.com/feast-dev/feast/issues/4616


sdk/python/feast/infra/passthrough_provider.py
```diff
@@ -148,10 +148,16 @@ def update_infra(
     ):
         # Call update only if there is an online store
         if self.online_store:
+            tables_to_keep_online = [
+                fv
+                for fv in tables_to_keep
+                if not isinstance(fv, FeatureView) or fv.online
```
HaoXuAI, line 154: I'm not sure if everying feature view will have a "online" parameter?

dmartinol (commented):
lgtm

sdk/python/feast/infra/passthrough_provider.py
HaoXuAI: I think the API might change in the future, so to avoid breaking, maybe add a check if it has this attribute.

HaoXuAI (approved):
LGTM!

HaoXuAI (approved):

lokeshrangineni (approved):

PR #4671, opened 2024-10-23
fix: Patch FAISS online return signature

franciscojavierarceo:
Patch FAISS online return signature

N/A

# Misc


tokoko (approved):

HaoXuAI:
Thanks! missed it. :)

PR #4686, opened 2024-10-24
fix: Populates project created_time correctly according to created ti…

EXPEbdodla:
…me in feast_metadata table

Populates project created_time correctly according to created time in feast_metadata table. 

# Misc

### DIFF
--- a/sdk/python/feast/infra/registry/sql.py
+++ b/sdk/python/feast/infra/registry/sql.py
@@ -281,7 +281,7 @@ class SqlRegistry(CachingRegistry):
         )
 
     def _sync_feast_metadata_to_projects_table(self):
-        feast_metadata_projects: set = []
+        feast_metadata_projects: dict = {}
         projects_set: set = []
         with self.read_engine.begin() as conn:
             stmt = select(feast_metadata).where(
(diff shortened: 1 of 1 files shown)


HaoXuAI (approved):
LGTM

EXPEbdodla:
Thank you. Can anyone restart pr-integration-tests? Error not related to PR.

HaoXuAI:
@franciscojavierarceo looks like I'm not in reviewers-and-approvers to approve it. need another stamp

franciscojavierarceo:
I updated your permission but that is surprising as you're a maintainer so you should be able to

sdk/python/feast/infra/registry/sql.py
```diff
@@ -976,7 +986,8 @@ def _apply_object(
                 if hasattr(obj_proto, "meta") and hasattr(
                     obj_proto.meta, "created_timestamp"
                 ):
-                    obj_proto.meta.created_timestamp.FromDatetime(update_datetime)
```
franciscojavierarceo, line 979: This logic seems flipped?

franciscojavierarceo (changes requested):

sdk/python/feast/infra/registry/sql.py
EXPEbdodla: If `created_timestamp` is explicitly set on proto object, use object `created_timestamp`. Otherwise set it with `update_datetime`.

PR #4692, opened 2024-10-25
feat: add extra args of feast serve cli in gunicorn

young-hun-jo:
Python feast feature server is using FastAPI and gunicorn and it is run by command `feast serve`. When I search extra optons in gunicorn of `feast serve`, only few options can be used(for example, `keepalive`, `workers`). So I add extra arguments of other gunicorn options. These are `threads`, `max_requests`, `max_requests_jitter`. I

I would run Python feast feature server on EKS environments on production. In this production environments, so many requests is come to feast feature server but only `workers` options is not enough to run feature server on production. 

Especially, I add these arguments for python memory leaking because of automated python's own garbage collector. Combination of `max_requests` and `max_requests_jitter` is appropriate for preventing circumstances that multi-worker process is simultaneously restarted over `max_requests`.

If these options is added, python feast feature server on production is running when so many requests is come to the server. I looking forward to reviewing yours. thanks.

# Misc


young-hun-jo:
a commit username mistakenly done by my other account. so I close it and I will make PR again

PR #4693, opened 2024-10-25
feat: add extra args of feast serve cli in gunicorn

iamcodingcat:

Python feast feature server is using FastAPI and gunicorn and it is run by command feast serve. When I search extra optons in gunicorn of feast serve, only few options can be used(for example, `keepalive, workers`). So I add extra arguments of other gunicorn options. These are` threads, max_requests, max_requests_jitter`.

I would run Python feast feature server on EKS environments on production. In this production environments, so many requests is come to feast feature server but only workers options is not enough to run feature server on production.

Especially, I add these arguments for python memory leaking because of automated python's own garbage collector. Combination of `max_requests and max_requests_jitter` is appropriate for preventing circumstances that multi-worker process is simultaneously restarted over `max_requests`.

If these options is added, python feast feature server on production is running when so many requests is come to the server. I looking forward to reviewing yours. thanks.

# Misc


sdk/python/feast/feature_store.py
```diff
@@ -1894,6 +1894,9 @@ def serve(
         type_: str = "http",
         no_access_log: bool = True,
         workers: int = 1,
+        threads: int = 1,
+        max_requests: int = 0,
```
lokeshrangineni, line 1898: how the server behaves if someone won't set this field and defaults to 0.

sdk/python/feast/feature_server.py
```diff
@@ -369,6 +372,9 @@ def start_server(
             bind=f"{host}:{port}",
             accesslog=None if no_access_log else "-",
             workers=workers,
+            threads=threads,
+            max_requests=max_requests,
+            max_requests_jitter=max_requests_jitter,
```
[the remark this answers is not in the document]
young-hun-jo: @lokeshrangineni yes. I know it. There is a [condition](https://github.com/young-hun-jo/feast/blob/f6e3ced360f59cf41e1fdf2bc08f8018ba132aad/sdk/python/feast/feature_server.py#L295-L315)

sdk/python/feast/cli.py
```diff
@@ -896,6 +896,28 @@ def init_command(project_directory, minimal: bool, template: str):
     show_default=True,
     help="Number of worker",
 )
+@click.option(
+    "--threads",
+    "-t",
+    type=click.INT,
+    default=1,
+    show_default=True,
+    help="Number of thread"
```
lokeshrangineni, line 905: could we add a bit verbose help especially for the beginners. when to increase threads or how it impacts? or may be reference to unicorn documentation.

sdk/python/feast/feature_store.py
young-hun-jo: @lokeshrangineni 

if `threads` is set to zero(`0`),  the only one thread is run in server.(At reference, `workers` can be set to zero. but this is only running a master process not worker process. so can't process client's request. I test it in FastAPI in localhost)

And if `max_requests` is set to zero(`0`), the automatic restart process is disabled(ref: [gunicorn docs](https://docs.gunicorn.org/en/stable/settings.html#max-requests))

sdk/python/feast/cli.py
young-hun-jo: @lokeshrangineni 
yes. I [commit it](https://github.com/feast-dev/feast/pull/4693/commits/dbd3b3843b0b0b10d8774e9f688ddde17df984b6) more. A documentation link is so long, it may worsen visibility of source code, so I wrote the summary

young-hun-jo: And short-cut of `thread` options doesn't exist in [official gunicorn documentation](https://docs.gunicorn.org/en/stable/settings.html#threads). So, I remove this it, too

PR #4694, opened 2024-10-25
chore: Make feature server test a unit test

robhowley:

* the feature server integration tests are only doing integration testing that's already covered by `test_universal_online`, i.e. feature_store -> data_store
* the code inside the path operation can be tested by unit tests
* this should also avoid some of the event loop flakiness arising from 
  * fastapi test client, not actual server isolation
  * context management w a shared store
  * 3rd party services

# Misc


sdk/python/tests/integration/online_store/test_python_feature_server.py
```diff
@@ -1,150 +0,0 @@
...
-)
-from tests.integration.feature_repos.universal.entities import (
-    customer,
-    driver,
-    location,
-)
-
-
-@pytest.mark.integration
-@pytest.mark.universal_online_stores
```
franciscojavierarceo, line 22: This runs the tests for all online stores and basically tests the feature server end to end.

robhowley: it does, but using the fake server via fastapi testclient so not a "true" integration e2e. what ends up happening is that it's basically a unit test of the server and integration test for the feature store. the integration test of the feature store w the diff online stores already happens in [test_universal_online](https://github.com/feast-dev/feast/blob/35fbdc91cc8a327be2a8cd162f7ccff983b16932/sdk/python/tests/integration/online_store/test_universal_online.py). there isn't, however, a test suite dedicated to working w the server, only that implicit one baked into this file. 

since this construct is proving to be complicated and brittle, seemed much simpler to break the implicit unit test out from the integration test and let the two exist on their own. imo if we really wanted a full e2e we'd start up a server and make calls to it w a regular client, not the fastapi one. reasonable people can def disagree on this file, but figured worth considering this option since integration tests have been failing inconsistently.

franciscojavierarceo: Cool, I can get behind that.

franciscojavierarceo (approved):

Issue #4688, opened 2024-10-25
Syntax Error in BigQuery While Retrieving Columns that Start with a Number


nishantgaurav-dev:
## Expected Behavior 
We have a feature view like this:

```
mbr_visits_fv_v1 = FeatureView(
	...
    schema=[
        Field(name="7_days_visits", dtype=Int64),
        Field(name="30_daysvisits", dtype=Int64),
    ],
    source=mbr_visits_bq_ds_v1
    ...
)
```

While querying this view as offline features I am expecting a data-frame containing desired columns without any error.

## Current Behavior
Feast SDK is giving error like:
`google.api_core.exceptions.BadRequest: 400 Syntax error: Missing whitespace between literal and alias at ...`

I have printed some logs and observed that the generated BQ query contains a segment like this:

```
SELECT
  event_timestamp as event_timestamp,
  mbr_id AS mbr_id,
  7_days_visits as 7_days_visits
  30_days_visits as 30_days_visits
FROM `myproj.my_dataset.mbr_visits`
WHERE event_timestamp <= '2024-06-27T00:00:00'
```

**This is incorrect BQ query.**

It should be like this:
```
SELECT
  event_timestamp as event_timestamp,
  mbr_id AS mbr_id,
  `7_days_visits` as `7_days_visits`
  `30_days_visits` as `30_days_visits`
FROM `myproj.my_dataset.mbr_visits`
WHERE event_timestamp <= '2024-06-27T00:00:00'
```

The BQ expects that if a queried column starts with a number, the name should be enclosed in backtick (`) chars.
## Steps to reproduce
Try querying a feature view with column name starting with number and use BigQuery data source.

### Specifications

- Version: 0.34.1
- Platform: Mac
- Subsystem: 

## Possible Solution
Enclose all the BQ column names with backtick (`) chars. It will reduce the readability a little bit, but won't cause any performance impact.

tmihalac:
"While querying this view as offline features I am expecting a data-frame containing desired columns without any error."
Can you also explain how did you query this view, which api and also a stack trace of the exception will be helpful

PR #4713, opened 2024-10-28
fix: Syntax Error in BigQuery While Retrieving Columns that Start wit…

tmihalac:
Syntax Error in BigQuery While Retrieving Columns that Start with a Number.
The solution is to escape the columns in the queries with `.

Fixes https://github.com/feast-dev/feast/issues/4688


sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -429,6 +437,10 @@ def offline_write_batch(
                 job_config=job_config,
             ).result()
 
+    @staticmethod
+    def escape_query_columns(columns: List[str]) -> List[str]:
+        return ["`" + x + "`" for x in columns]
```
dmartinol, line 442: what about:
```python
return [f"`{x}`" for x in columns]
```

tmihalac: Looks good updated code

dmartinol (approved):

HaoXuAI (approved):
LGTM

PR #4718, opened 2024-10-29
feat: Adding ssl support for registry server.

lokeshrangineni:
Adding SSL(TLS) Support for registry server and remote registry client

### DIFF
--- a/docs/reference/feature-servers/python-feature-server.md
+++ b/docs/reference/feature-servers/python-feature-server.md
@@ -216,10 +216,10 @@ The above command will generate two files
 * `cert.pem`: certificate public key
 
 ### Starting the Online Server in SSL Mode
-To start the feature server in SSL mode, you need to provide the private and public keys using the `--ssl-key-path` and `--ssl-cert-path` arguments with the `feast serve` command.
+To start the feature server in SSL mode, you need to provide the private and public keys using the `--key` and `--cert` arguments with the `feast serve` command.
 
 ```shell
-feast serve --ssl-key-path /path/to/key.pem --ssl-cert-path /path/to/cert.pem
+feast serve --key /path/to/key.pem --cert /path/to/cert.pem
 ```
 
 # Online Feature Server Permissions and Access Control
--- a/docs/reference/feature-servers/python-feature-server.md
+++ b/docs/reference/feature-servers/python-feature-server.md
@@ -219,7 +219,7 @@ The above command will generate two files
 To start the feature server in SSL mode, you need to provide the private and public keys using the `--ssl-key-path` and `--ssl-cert-path` arguments with the `feast serve` command.
 
 ```shell
-feast serve --ssl-key-path key.pem --ssl-cert-path cert.pem
+feast serve --ssl-key-path /path/to/key.pem --ssl-cert-path /path/to/cert.pem
 ```
 
 # Online Feature Server Permissions and Access Control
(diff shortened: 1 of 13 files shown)


dmartinol (approved):
adding test code in a separate PR?

franciscojavierarceo:
> adding test code in a separate PR?

Should be in same PR

lokeshrangineni:
> adding test code in a separate PR?

I have added the tests, please take a look at it and approve. cc: @franciscojavierarceo

docs/reference/feature-servers/python-feature-server.md
```diff
@@ -219,7 +219,7 @@ The above command will generate two files
 To start the feature server in SSL mode, you need to provide the private and public keys using the `--ssl-key-path` and `--ssl-cert-path` arguments with the `feast serve` command.
 
 ```shell
-feast serve --ssl-key-path key.pem --ssl-cert-path cert.pem
+feast serve --ssl-key-path /path/to/key.pem --ssl-cert-path /path/to/cert.pem
```
franciscojavierarceo, line 222: Nit: this name is a little misleading

dmartinol: The standard k8s secrets for TLS have 2 keys `tls.crt` and `tls.key`
These can be generated using the CLI with:
```
kubectl create secret tls tls-secret --cert=path/to/tls.crt --key=path/to/tls.key
```
Can we use the same options? [Refs](https://kubernetes.io/docs/reference/kubectl/generated/kubectl_create/kubectl_create_secret_tls)

lokeshrangineni: @franciscojavierarceo - what do you think of those options? `ssl_filename` can be used for private key but needs to come up with other name for public key. `pem_filename` private key extension can be different from pem as well. If you are ok with Daneil's proposal then i can update the code for this PR and also online server code to keep it consistent across feast.

franciscojavierarceo: +1 to @dmartinol's suggestion

lokeshrangineni:
> > adding test code in a separate PR?
> 
> I have added the tests, please take a look at it and let me know for any comments. cc: @franciscojavierarceo

I realized forgot to push my changes, i just pushed my changes. sorry about the confusion. please review them.

docs/reference/feature-servers/python-feature-server.md
lokeshrangineni: Incorporated this change. Please take a look at it again.

dmartinol:
SSL is old terminology, I would replace it with TLS (everywhere)

franciscojavierarceo:
Yeah please update language

lokeshrangineni:
> SSL is old terminology, I would replace it with TLS (everywhere)

I have updated and some places kept both names to have more context, that is intentional still i m wired to realize if i look at SSL rather than TLS. please let me know if you think it is not necessary.

sdk/python/feast/feature_server.py
```diff
@@ -339,8 +339,8 @@ def start_server(
     workers: int,
     keep_alive_timeout: int,
     registry_ttl_sec: int,
-    ssl_key_path: str,
-    ssl_cert_path: str,
+    tls_key_path: str,
+    tls_cert_path: str,
     metrics: bool,
 ):
     if metrics:
```
tmihalac, line 346: I think its a good idea to do
if (tls_key_path and not tls_cert_path) or (not tls_key_path and tls_cert_path)
as well

tmihalac:
lgtm

franciscojavierarceo (approved):

PR #4719, opened 2024-10-30
fix: Support registry instantiation for read-only users

### LINKED ISSUE #4715: FeatureStore instantiation fails after update to 0.41.3
## Expected Behavior 

We have started using feast (v0.40.0) to create a feature registry on S3. In a script to fetch data from the offline store the feature store is instantiated with configuration from the default config file. Afterwards, data can be fetched in the script. The registry is supposed to be changed only by dedicated users.

## Current Behavior

The script fails with AccessDenied error because it tries to change the registry on S3 but there are no sufficient access credentials configured.

Essentially, feast fails at https://github.com/feast-dev/feast/blob/4a6b663f80bc91d6de35ed2ec428d34811d17a18/sdk/python/feast/infra/registry/registry.py#L232 which was added in https://github.com/feast-dev/feast/pull/4475 .

I do not understand the complete logic but why do we make a commit to the registry when we only want to use the existing registry? Should we have a flag here to disable it in cases where we only want to read the registry, like fetching data or `feast plan`?

## Steps to reproduce

Call `fs = FeatureStore()` with existing feature store config and S3 registry without configuring PutObject policies for current user.

### Specifications

- Version: v0.41.3
- Platform: MacOs
- Subsystem:

EXPEbdodla:
Support registry instantiation for read-only users. I'll add the test case in the next PR. 

Fixes #4715 

# Misc

### DIFF
--- a/sdk/python/feast/infra/registry/registry.py
+++ b/sdk/python/feast/infra/registry/registry.py
@@ -228,7 +228,6 @@ class Registry(BaseRegistry):
                 self._sync_feast_metadata_to_projects_table()
             except FileNotFoundError:
                 logger.info("Registry file not found. Creating new registry.")
-            finally:
                 self.commit()
 
     def _sync_feast_metadata_to_projects_table(self):


dmartinol (approved):

nanohanno:
Thanks for the quick response and this fix. Could we have a patch release soon to be able to use the last updates to the library?

PR #4721, opened 2024-10-30
feat: Add api contract to fastapi docs

### LINKED ISSUE #4720: Addition of /docs endpoint
Creation of /docs endpoint in order to have a place to view easily all the avilable endpoints with:

- a basic endpoint description.
- its development status/maturity
- requiered and optional parameters
- example of request body
- example of a succesful response
- custem errors explanation(if exist)

robhowley:

* FastApi by default creates an openapi docs page which lists out endpoints
* the python feature server doesn't use pydantic models for request bodies, so the sample inputs for the endpoints aren't populated and schema unknown
* remove json request body parsing and replace it w pydantic models
* this mr does _not_ define a robust response schema for `/get-online-features`
  * simply specifies it as dict[str, any]
  * should come up w something better but as a follow up

```
feast serve --port 8080
```
generates the following at [http://localhost:8080/docs](http://localhost:8080/docs)

image

image

Fixes #4720 

# Misc

### DIFF
--- a/sdk/python/feast/feature_server.py
+++ b/sdk/python/feast/feature_server.py
@@ -118,7 +118,7 @@ def get_app(
         "/get-online-features",
         dependencies=[Depends(inject_user_details)],
     )
-    async def get_online_features(request: GetOnlineFeaturesRequest):
+    async def get_online_features(request: GetOnlineFeaturesRequest) -> Dict[str, Any]:
         # Initialize parameters for FeatureStore.get_online_features(...) call
         if request.feature_service:
             feature_service = store.get_feature_service(
@@ -167,7 +167,7 @@ def get_app(
         )
 
     @app.post("/push", dependencies=[Depends(inject_user_details)])
-    async def push(request: PushFeaturesRequest):
+    async def push(request: PushFeaturesRequest) -> None:
         df = pd.DataFrame(request.df)
         actions = []
         if request.to == "offline":
@@ -219,7 +219,7 @@ def get_app(
             store.push(**push_params)
 
     @app.post("/write-to-online-store", dependencies=[Depends(inject_user_details)])
-    def write_to_online_store(request: WriteToFeatureStoreRequest):
+    def write_to_online_store(request: WriteToFeatureStoreRequest) -> None:
         df = pd.DataFrame(request.df)
         feature_view_name = request.feature_view_name
         allow_registry_cache = request.allow_registry_cache
@@ -248,7 +248,7 @@ def get_app(
         )
 
     @app.post("/materialize", dependencies=[Depends(inject_user_details)])
-    def materialize(request: MaterializeRequest):
+    def materialize(request: MaterializeRequest) -> None:
         for feature_view in request.feature_views or []:
             # TODO: receives a str for resource but isn't in the Union. is str actually allowed?
             assert_permissions(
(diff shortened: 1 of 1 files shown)


sdk/python/feast/feature_server.py
```diff
@@ -108,33 +114,26 @@ async def lifespan(app: FastAPI):
...
-
     @app.post(
         "/get-online-features",
         dependencies=[Depends(inject_user_details)],
     )
-    async def get_online_features(body=Depends(get_body)):
-        body = json.loads(body)
-        full_feature_names = body.get("full_feature_names", False)
-        entity_rows = body["entities"]
+    async def get_online_features(request: GetOnlineFeaturesRequest) -> Dict[str, Any]:
```
robhowley, line 121: fastapi is smart about types. surfaces pydantic model in the autogenerated api docs

sdk/python/feast/feature_server.py
```diff
@@ -69,6 +68,13 @@ class MaterializeIncrementalRequest(BaseModel):
     feature_views: Optional[List[str]] = None
 
 
+class GetOnlineFeaturesRequest(BaseModel):
+    entities: Dict[str, List[Any]]
+    feature_service: Optional[str] = None
+    features: Optional[List[str]] = None
```
robhowley, line 74: could (should) add validation on the model that throws if both are set. didnt do it here, bc it would be technically breaking. the code works just picks the `feature_service` value and ignore `features`. this is an unintuitive (successful) result, seems like it would make more sense to return `422` and say, "gotta pick one". maybe flag it for 1.0? @franciscojavierarceo

franciscojavierarceo (commented):
Looks like we're ignoring a lot of types now. Were the type checks nontrivial to add? I've had some challenges in the past so wanted to confirm.

robhowley:
image

im honestly unsure how this passed before. i think bc it was pulling from json it was just marking everything as dict[str, Any]? it gets an inconsistent assignment type error which makes sense since it's assigning two diff things. i can try and add asserts to please the mypy gods, but didnt want to introduce the risk of throwing where one didnt exist before (even though we know it shouldn't matter) just bc of mypy. i'll defer to you @franciscojavierarceo on this: try and add asserts or let it be.

franciscojavierarceo:
We can add asserts and type check in subsequent PR. Thank you for this!

robhowley:
image

flake?

sdk/python/feast/feature_server.py
```diff
@@ -250,11 +248,12 @@ async def health():
         )
 
     @app.post("/materialize", dependencies=[Depends(inject_user_details)])
-    def materialize(body=Depends(get_body)):
-        request = MaterializeRequest(**json.loads(body))
-        for feature_view in request.feature_views:
+    def materialize(request: MaterializeRequest) -> None:
+        for feature_view in request.feature_views or []:
+            # TODO: receives a str for resource but isn't in the Union. is str actually allowed?
```
dmartinol, line 253: it's not allowed, so we should fetch the `feature_view` instance as we do in the `write-to-online-store` endpoint.
let's open a GH issue for that, unless you want to fix it here

dmartinol: BTW: thanks for catching it!

robhowley: nah, not here. that should be a sep issue and `fix:` pr

dmartinol: https://github.com/feast-dev/feast/issues/4726

robhowley: i'll take care of #4726 once this pr is merged

robhowley:
giving this a bump.

franciscojavierarceo (approved):

Issue #4075, opened 2024-04-04
Add support for singleton Dictionary execution for on demand feature views


franciscojavierarceo:
**Is your feature request related to a problem? Please describe.**
Support for Python Native transformations was recently released (https://github.com/feast-dev/feast/pull/4069) but it expects a type = `Dict[str, List[Any]]`. Feast should support `Dict[str, Any]` where `Any` could be a singleton. 

**Describe the solution you'd like**
We should be able to run on demand feature views like so:
```python
@on_demand_feature_view(
    sources=[driver_stats_fv[["conv_rate", "acc_rate"]]],
    schema=[
        Field(name="conv_rate_plus_acc_python_singleton", dtype=Float64)
    ],
    mode="python",
)
def python_singleton_view(inputs: Dict[str, Any]) -> Dict[str, Any]:
    output: Dict[str, Any] = dict(conv_rate_plus_acc_python=float("-inf"))
    output["conv_rate_plus_acc_python_singleton"] = (
        inputs["conv_rate"] + inputs["acc_rate"]
    )
    return output
```

Currently, this feature view result in a type error. 

**Describe alternatives you've considered**
N/A

**Additional context**

PR #4724, opened 2024-10-31
feat: Adding support for native Python transformations on a single dictionary 

franciscojavierarceo:
This PR adds support for native Python transformations on a dictionary of a singleton (i.e., what most Data Scientists would call a single row).

As an example, Data Scientists can now transform features using an On Demand Transform like so:

```python
@on_demand_feature_view(
    sources=[driver_stats_fv[["conv_rate", "acc_rate"]]],
    schema=[
        Field(name="conv_rate_plus_acc_python_singleton", dtype=Float64)
    ],
    mode="python",
    singleton=True,
)
def python_singleton_view(inputs: dict[str, Any]) -> dict[str, Any]:
    output: dict[str, Any] = dict(conv_rate_plus_acc_python=float("-inf"))
    output["conv_rate_plus_acc_python_singleton"] = (
        inputs["conv_rate"] + inputs["acc_rate"]
    )
    return output
``` 

Compare this to the previous logic to do the transformation which looks like:

```python
@on_demand_feature_view(
    sources=[driver_stats_fv[["conv_rate", "acc_rate"]]],
    schema=[
        Field(name="conv_rate_plus_val", dtype=Float64),
    ],
    mode="python",
)
def python_demo_view(inputs: dict[str, Any]) -> dict[str, Any]:
    output: dict[str, Any] = {
        "conv_rate_plus_val": [
            conv_rate + acc_rate
            for conv_rate, acc_rate in zip(
                inputs["conv_rate"], inputs["acc_rate"]
            )
        ],
    }
    return output
```

https://github.com/feast-dev/feast/issues/4075

# Misc
N/A

### DIFF
--- a/sdk/python/feast/on_demand_feature_view.py
+++ b/sdk/python/feast/on_demand_feature_view.py
@@ -74,6 +74,7 @@ class OnDemandFeatureView(BaseFeatureView):
     tags: dict[str, str]
     owner: str
     write_to_online_store: bool
+    singleton: bool
 
     def __init__(  # noqa: C901
         self,
@@ -98,6 +99,7 @@ class OnDemandFeatureView(BaseFeatureView):
         tags: Optional[dict[str, str]] = None,
         owner: str = "",
         write_to_online_store: bool = False,
+        singleton: bool = False,
     ):
         """
         Creates an OnDemandFeatureView object.
@@ -121,6 +123,8 @@ class OnDemandFeatureView(BaseFeatureView):
                 of the primary maintainer.
             write_to_online_store (optional): A boolean that indicates whether to write the on demand feature view to
             the online store for faster retrieval.
+            singleton (optional): A boolean that indicates whether the transformation is executed on a singleton
+                (only applicable when mode="python").
         """
         super().__init__(
             name=name,
@@ -204,6 +208,9 @@ class OnDemandFeatureView(BaseFeatureView):
         self.features = features
         self.feature_transformation = feature_transformation
         self.write_to_online_store = write_to_online_store
+        self.singleton = singleton
+        if self.singleton and self.mode != "python":
+            raise ValueError("Singleton is only supported for Python mode.")
 
     @property
     def proto_class(self) -> type[OnDemandFeatureViewProto]:
@@ -221,6 +228,7 @@ class OnDemandFeatureView(BaseFeatureView):
             tags=self.tags,
             owner=self.owner,
             write_to_online_store=self.write_to_online_store,
+            singleton=self.singleton,
         )
         fv.entities = self.entities
         fv.features = self.features
@@ -247,6 +255,7 @@ class OnDemandFeatureView(BaseFeatureView):
             or self.feature_transformation != other.feature_transformation
             or self.write_to_online_store != other.write_to_online_store
             or sorted(self.entity_columns) != sorted(other.entity_columns)
+            or self.singleton != other.singleton
         ):
             return False
 
@@ -328,6 +337,7 @@ class OnDemandFeatureView(BaseFeatureView):
             tags=self.tags,
             owner=self.owner,
             write_to_online_store=self.write_to_online_store,
+            singleton=self.singleton if self.singleton else False,
         )
 
         return OnDemandFeatureViewProto(spec=spec, meta=meta)
@@ -434,6 +444,9 @@ class OnDemandFeatureView(BaseFeatureView):
             ]
         else:
             entity_columns = []
+        singleton = False
+        if hasattr(on_demand_feature_view_proto.spec, "singleton"):
+            singleton = on_demand_feature_view_proto.spec.singleton
 
         on_demand_feature_view_obj = cls(
             name=on_demand_feature_view_proto.spec.name,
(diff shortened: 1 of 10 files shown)


franciscojavierarceo:
FYI @robhowley

robhowley:
is there no way to do any kind of type inference in the codepath to determine it's a singleton or leave it up to the person implementing the transformation to know what data they're expecting and raise errors accordingly? just kinda lean more into duck typing. strikes me as a little unusual to have to pass a `bool` in to specify what the type of my data is. im assuming there's context on this im missing about how the subsequent transformations will interact w proto stuff

franciscojavierarceo:
I definitely agree this Boolean is not the preferred option but it was the what made the most sense to me given that other sorts of type inference start to become brittle once you allow for different input types.

franciscojavierarceo:
Yeah that could be done today because type annotations can still be ignored by users but I don't think that's a good enough experience for Data Scientists. I think this make the transformation really easy and it has rational defaults.

robhowley:
and defining it as a Union of the two types wasn't sufficient?

sdk/python/feast/on_demand_feature_view.py
```diff
@@ -614,17 +628,19 @@ def transform_dict(
                     feature_dict[full_feature_ref] = feature_dict[feature.name]
                     columns_to_cleanup.append(str(full_feature_ref))
 
-        output_dict: dict[str, Any] = self.feature_transformation.transform(
-            feature_dict
-        )
+        if self.singleton and self.mode == "python":
```
franciscojavierarceo, line 631: @robhowley see here

sdk/python/feast/on_demand_feature_view.py
```diff
@@ -614,17 +628,19 @@ def transform_dict(
...
+            output_dict = self.feature_transformation.transform(feature_dict)
         for feature_name in columns_to_cleanup:
             del output_dict[feature_name]
         return output_dict
 
     def infer_features(self) -> None:
-        inferred_features = self.feature_transformation.infer_features(
-            self._construct_random_input()
-        )
+        random_input = self._construct_random_input(singleton=self.singleton)
```
franciscojavierarceo, line 642: @robhowley and here

sdk/python/feast/on_demand_feature_view.py
```diff
@@ -663,20 +681,25 @@ def _construct_random_input(self) -> dict[str, list[Any]]:
             ValueType.BOOL_LIST: [[True]],
             ValueType.UNIX_TIMESTAMP_LIST: [[_utc_now()]],
         }
+        if singleton:
+            rand_dict_value = {k: rand_dict_value[k][0] for k in rand_dict_value}
```
franciscojavierarceo, line 685: @robhowley and here

sdk/python/feast/utils.py
```diff
@@ -570,6 +570,8 @@ def _augment_response_with_on_demand_transforms(
             proto_values.append(
                 python_values_to_proto_values(
                     feature_vector
+                    if isinstance(feature_vector, list)
+                    else [feature_vector]
```
franciscojavierarceo, line 574: @robhowley additional special logic

franciscojavierarceo:
@robhowley I tagged you in some places that outline some of why I ended up doing it this way.

HaoXuAI:
I'm also not following why it can't be just a type check. Adding this singleton feels complex the logic.
also since it's `dict[str, Any]`, if you enforce the value to be List, it could be list of list of list..., so you can't get the logic abstracted.

franciscojavierarceo:
Depends on the type check. You may have input features that are mixed that you want to create a feature out of. 

For example, if I did a type check to ensure all inputs were *not* lists, they will not accommodate the case I just mentioned.

sdk/python/feast/transformation/python_transformation.py
```diff
@@ -37,24 +37,39 @@ def transform(self, input_dict: dict) -> dict:
         output_dict = self.udf.__call__(input_dict)
         return {**input_dict, **output_dict}
 
-    def infer_features(self, random_input: dict[str, list[Any]]) -> list[Field]:
-        output_dict: dict[str, list[Any]] = self.transform(random_input)
+    def transform_singleton(self, input_dict: dict) -> dict:
+        # This flattens the list of elements to extract the first one
+        # in the case of a singleton element, it takes the value directly
+        # in the case of a list of lists, it takes the first list
+        input_dict = {k: v[0] for k, v in input_dict.items()}
```
HaoXuAI, line 44: v[0] might cause error?

HaoXuAI (approved):
I won't block it to go alive, but I think it would be great if there is some other way for this instead of this "singleton" parameter :)

franciscojavierarceo:
Do you have suggestion? I had dawned discussion with @robhowley indicating that this was not my preferred solution but simply the one I ended up having to settle on due to the tradeoffs.

sdk/python/feast/transformation/python_transformation.py
franciscojavierarceo: It shouldn't but I'll add a test for it

HaoXuAI:
> Do you have suggestion? I had dawned discussion with @robhowley indicating that this was not my preferred solution but simply the one I ended up having to settle on due to the tradeoffs.

I don't have any direct suggestion on this yet. But I believe after refactoring rebuild the Transformation entirely it should work  in some structured way :)

Issue #4561, opened 2024-09-23
Feast Go Operator


tchughesiv:
**Is your feature request related to a problem? Please describe.**
As the Feast project continues to grow, it’s become increasingly difficult to deploy a fully functional Feature Store in a way that also leverages key features, like RBAC. Currently, Feast does not offer a simple means of deployment that one might consider “production-like”. The existing installers require many manual operations that can lead to configuration errors. As a user, I want to be able to use an Operator to install and maintain Feature Store environments.

**Describe the solution you'd like**
RFC - https://docs.google.com/document/d/1vGKMizf3_14IyiF_W_Ik7CR03joFkQfzbKT0jH4PZJM/edit

**Describe alternatives you've considered**
These are laid out in the RFC

tchughesiv:
RFC has been approved and we've merged several initial PRs to implement the Operator... including -

- https://github.com/feast-dev/feast/pull/4596
- https://github.com/feast-dev/feast/pull/4635
- https://github.com/feast-dev/feast/pull/4656
- https://github.com/feast-dev/feast/pull/4723
- https://github.com/feast-dev/feast/pull/4739
- https://github.com/feast-dev/feast/pull/4742

PR #4739, opened 2024-11-05
feat: Added support for setting env vars in feast services in feast controller 

tmihalac:
Allow users to add or override the container env vars for each deployed feast service using the new env list in the CR

For example:

let's say the operator sets the offline container env vars like this by default -

```
env:
    value: <string>
  - name: VAR1
    value: test
```
If the user deployed the following CR -

```
kind: FeatureStore
metadata:
  name: sample
spec:
  feastProject: my_project
  services:
    offlineStore:
      env:
       - name: VAR1
         value: override
       - name: VAR2
         value: demo
```
The resulting offline container env vars would look like this -

```
      env:
          value: <string>
        - name: VAR1
          value: override
        - name: VAR2
          value: demo
```


infra/feast-operator/internal/controller/featurestore_controller_test.go
```diff
@@ -751,6 +733,89 @@ var _ = Describe("FeatureStore Controller", func() {
 			Expect(repoConfig).To(Equal(testConfig))
 		})
 
+		It("should properly set container env variables", func() {
+			By("Reconciling the created resource")
```
tchughesiv, line 737: can we add an env variable test for the following type of envvar?
```yaml
     env:
        - name: MY_POD_NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace
```

tchughesiv (changes requested):

infra/feast-operator/internal/controller/featurestore_controller_test.go
tmihalac: Done

dmartinol: Shouldn't we also support `envFrom` options to read from `Secrets` or `ConfigMap`?
@tchughesiv @tmihalac

tchughesiv: @dmartinol that's what this added test will confirm support for... `EnvVarSource` contains fieldRef, as well as `configMapKeyRef` & `secretKeyRef`

tmihalac: It is supported, there is just no specific test for that
I don't see a lot of added value in adding those tests since the only thing we are doing is overriding the original EnvVar object from the container with the ones from the spec if they have the same name

infra/feast-operator/internal/controller/featurestore_controller_test.go
```diff
@@ -1130,3 +1223,25 @@ func getFeatureStoreYamlEnvVar(envs []corev1.EnvVar) *corev1.EnvVar {
...
+
+	// Check the second array against the map.
+	for _, env := range arr2 {
+		if val, exists := envMap[env.Name]; !exists || val != env.Value {
+			return false
+		}
+	}
+
+	return true
+}
```
tchughesiv, line 1247: i wonder if `reflect.DeepEqual()` might be a better approach? instead of this `areEnvVarArraysEqual` function?

tmihalac: > i wonder if `reflect.DeepEqual()` might be a better approach? instead of this `areEnvVarArraysEqual` function?

It doesn't take into account 2 arrays are the same if they have the same elements but different order:

```package main

import (
	"fmt"
	"reflect"
)

func main() {
	arr1 := [3]int{1, 2, 3}
	arr2 := [3]int{1, 2, 3}
	arr3 := [3]int{3, 2, 1}

	fmt.Println(reflect.DeepEqual(arr1, arr2)) // Output: true
	fmt.Println(reflect.DeepEqual(arr1, arr3)) // Output: false
```

infra/feast-operator/internal/controller/featurestore_controller_test.go
tmihalac: > @dmartinol that's what this added test will confirm support for... `EnvVarSource` contains `fieldRef`, as well as `configMapKeyRef` & `secretKeyRef`

I've added a test for `fieldRef`, do you think we should add  for `configMapKeyRef` & `secretKeyRef` as well? 
see my previous comment

tchughesiv: @tmihalac imo, we don't need a test for the others

infra/feast-operator/internal/controller/featurestore_controller_test.go
tchughesiv: the main problem i see with the `areEnvVarArraysEqual` function is its not checking `env.ValueFrom`. maybe we use `reflect.DeepEqual(env, env)` at the EnvValue level somehow... within this custom function?

tchughesiv (changes requested):

infra/feast-operator/internal/controller/featurestore_controller_test.go
tmihalac: Done

tchughesiv (approved):
lgtm

tchughesiv:
@tmihalac looks like one of the unit tests are failing

infra/feast-operator/internal/controller/featurestore_controller_test.go
tchughesiv: maybe `DeepEqual` isn't even needed? `envMap[env.Name] == env` ??

tchughesiv (approved):

lokeshrangineni (approved):

PR #4740, opened 2024-11-05
feat: Support for nested timestamp fields in Spark Offline store

EXPEbdodla:
1. Supports using nested timestamp fields in the Spark Source using Field Mappings
2. Added Tests related to #4715 

# Misc


sdk/python/feast/utils.py
```diff
@@ -212,6 +213,28 @@ def _run_pyarrow_field_mapping(
     return table
 
 
+def _get_fields_with_aliases(
```
franciscojavierarceo, line 216: probably would be good to add a test for this but i won't block on it.

franciscojavierarceo (approved):
lgtm, thank you for this!

sdk/python/feast/utils.py
EXPEbdodla: Don't have a direct test but pull_latest_query tests will use it. I'll add explicit tests in next PRs.

EXPEbdodla:
Resolved merge conflicts with another commit.

PR #4742, opened 2024-11-06
feat: File persistence definition and implementation

dmartinol:
Add ephemeral persistence configuration to the Feast services deployed with the operator.
A sample manifest is also committed and reported here for your convenience:
```yaml
kind: FeatureStore
metadata:
  name: sample-ephemeral-persistence
spec:
  feastProject: my_project
  services:
    onlineStore:
      persistence:
        file:
          path: /data/online_store.db
    offlineStore:
      persistence:
        file:
          type: dask
    registry:
      local:
        persistence:
          file:
            path: /data/registry.db
```

Next PR will add persistent storage by using mounted volumed (Kubernetes [PVC](https://kubernetes.io/docs/concepts/storage/persistent-volumes/))


dmartinol:
@tchughesiv @tmihalac 
Note: validations will be added to the CRD types as the other options will be integrated.

infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -71,16 +71,60 @@ type FeatureStoreServices struct {
...
+// OfflineStorePersistence configures the persistence settings for the offline store service
+type OfflineStorePersistence struct {
+	// +optional
+	FilePersistence *OfflineStoreFilePersistence `json:"file,omitempty"`
+}
+
+// OfflineStorePersistence configures the file-based persistence for the offline store service
+type OfflineStoreFilePersistence struct {
+	// +optional
+	// +default:value=dask
```
tchughesiv, line 87: we don't want to apply defaults here as they will appear in the `spec`. the application of defaults should occur w/ controller logic to `status.applied`.

infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -71,16 +71,60 @@ type FeatureStoreServices struct {
 // OfflineStore configures the deployed offline store service
 type OfflineStore struct {
 	ServiceConfigs `json:",inline"`
+	// +optional
```
tchughesiv, line 74: the `// +optional` comments aren't necessary as it is equivalent to using`omitempty` in the json field descrip. imo, let's pick one approach and stick w/ that.

infra/feast-operator/internal/controller/services/repo_config.go
```diff
@@ -35,48 +35,75 @@ func (feast *FeastServices) GetServiceFeatureStoreYamlBase64(feastType FeastServ
 }
 
 func (feast *FeastServices) getServiceFeatureStoreYaml(feastType FeastServiceType) ([]byte, error) {
-	return yaml.Marshal(feast.getServiceRepoConfig(feastType))
+	repoConfig, err := feast.getServiceRepoConfig(feastType)
+	if err != nil {
+		return nil, err
+	}
+	return yaml.Marshal(repoConfig)
```
tchughesiv, line 42: what was the reason for this change?

infra/feast-operator/internal/controller/services/repo_config.go
```diff
@@ -35,48 +35,75 @@ func (feast *FeastServices) GetServiceFeatureStoreYamlBase64(feastType FeastServ
...
-			repoConfig.OfflineStore = OfflineStoreConfig{
-				Type: OfflineDaskConfigType,
+			fileType := string(OfflineDaskConfigType)
+			if appliedSpec.Services.OfflineStore.Persistence != nil && appliedSpec.Services.OfflineStore.Persistence.FilePersistence != nil {
+				fileType = appliedSpec.Services.OfflineStore.Persistence.FilePersistence.Type
+			}
+
+			repoConfig.OfflineStore = OfflineStoreConfig{}
+			var err error
+			repoConfig.OfflineStore.Type, err = ParseOfflineConfigType(fileType)
```
tchughesiv, line 64: declaring the `err` var shouldn't be necessary. delete that line and do this instead -
```go
			repoConfig.OfflineStore.Type, err := ParseOfflineConfigType(fileType)
```

infra/feast-operator/internal/controller/services/util.go
```diff
@@ -0,0 +1,94 @@
...
+		// if local registry not set, apply an empty pointer struct
+		if applied.Services.Registry.Local == nil {
+			applied.Services.Registry.Local = &feastdevv1alpha1.LocalRegistryConfig{
+				Persistence: &feastdevv1alpha1.RegistryPersistence{
+					FilePersistence: &feastdevv1alpha1.RegistryFilePersistence{
+						Path: DefaultRegistryPath,
+					},
+				},
+			}
+		}
```
tchughesiv, line 43: no need to file persistence to be applied twice here. instead, need to cover all bases by checking each pointer object separately, but persistence is ultimately only set once -
```suggestion
	// default to registry service deployment
	if applied.Services.Registry == nil {
		applied.Services.Registry = &feastdevv1alpha1.Registry{},
	}
	// if remote registry not set, proceed w/ local registry defaults
	if applied.Services.Registry.Remote == nil {
		// if local registry not set, apply an empty pointer struct
		if applied.Services.Registry.Local == nil {
			applied.Services.Registry.Local = &feastdevv1alpha1.LocalRegistryConfig{},
		}
		if applied.Services.Registry.Local.Persistence == nil {
			applied.Services.Registry.Local.Persistence = &feastdevv1alpha1.RegistryPersistence{
					FilePersistence: &feastdevv1alpha1.RegistryFilePersistence{
						Path: DefaultRegistryPath,
					},
				},
			}
		}
```

tchughesiv: this is pertinent in case the user defines `spec.services.registry.local: {}` for example. we'd still need to set some persistence defaults.

infra/feast-operator/internal/controller/services/util.go
```diff
@@ -0,0 +1,94 @@
...
+				FilePersistence: &feastdevv1alpha1.OfflineStoreFilePersistence{
+					Type: string(OfflineDaskConfigType),
+				},
+			}
+		} else if applied.Services.OfflineStore.Persistence.FilePersistence == nil {
+			applied.Services.OfflineStore.Persistence.FilePersistence = &feastdevv1alpha1.OfflineStoreFilePersistence{
+				Type: string(OfflineDaskConfigType),
+			}
+		}
+	}
```
tchughesiv, line 70: same... this could be improved by checking each pointer separately and then applying file persistence defaults last.

infra/feast-operator/internal/controller/services/util.go
```diff
@@ -0,0 +1,94 @@
...
+				FilePersistence: &feastdevv1alpha1.OnlineStoreFilePersistence{
+					Path: DefaultOnlinePath,
+				},
+			}
+		} else if applied.Services.OnlineStore.Persistence.FilePersistence == nil {
+			applied.Services.OnlineStore.Persistence.FilePersistence = &feastdevv1alpha1.OnlineStoreFilePersistence{
+				Path: DefaultOnlinePath,
+			}
+		}
+	}
```
tchughesiv, line 84: same

tchughesiv (changes requested):

infra/feast-operator/internal/controller/services/util.go
tchughesiv: would be good to have unit tests ensuring that proper defaults are being set in status when varying pointers around service persistence are set in the spec.

infra/feast-operator/internal/controller/services/util.go
tchughesiv: we prob need to incorporate an `if applied.Services.Registry.Local.Persistence.FilePersistence != nil && len(applied.Services.Registry.Local.Persistence.FilePersistence) == 0` check in here too right? in that scenario we'd `applied.Services.Registry.Local.Persistence.FilePersistence.Path = DefaultRegistryPath`

infra/feast-operator/internal/controller/services/suite_test.go
```diff
@@ -0,0 +1,90 @@
...
+		// default path defined in controller-runtime which is /usr/local/kubebuilder/.
+		// Note that you must have the required binaries setup under the bin directory to perform
+		// the tests directly. When we run make test it will be setup and used automatically.
+		BinaryAssetsDirectory: filepath.Join("..", "..", "..", "bin", "k8s",
+			fmt.Sprintf("1.29.0-%s-%s", runtime.GOOS, runtime.GOARCH)),
+	}
+
+	var err error
+	// cfg is defined in this file globally.
+	cfg, err = testEnv.Start()
```
tchughesiv, line 71:
```suggestion
	// cfg is defined in this file globally.
	cfg, err := testEnv.Start()
```

infra/feast-operator/internal/controller/services/repo_config.go
tchughesiv, line 64:
```suggestion
			repoConfig.OfflineStore.Type, err := ParseOfflineConfigType(fileType)
```

tchughesiv (changes requested):

infra/feast-operator/internal/controller/services/repo_config.go
```diff
@@ -35,48 +35,75 @@ func (feast *FeastServices) GetServiceFeatureStoreYamlBase64(feastType FeastServ
...
 }
 
-func (feast *FeastServices) getServiceRepoConfig(feastType FeastServiceType) RepoConfig {
-	appliedSpec := feast.FeatureStore.Status.Applied
+func (feast *FeastServices) getServiceRepoConfig(feastType FeastServiceType) (RepoConfig, error) {
+	return getServiceRepoConfig(feastType, feast.FeatureStore)
+}
 
-	repoConfig := feast.getClientRepoConfig()
+func getServiceRepoConfig(feastType FeastServiceType, featureStore *feastdevv1alpha1.FeatureStore) (RepoConfig, error) {
```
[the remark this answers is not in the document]
tchughesiv: but i also kind of get why you did it... your call

infra/feast-operator/internal/controller/services/repo_config.go
tchughesiv: we shouldn't require a `ParseOfflineConfigType`. we should be able to just set the CRD to use the `OfflineConfigType` directly.

infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -71,16 +71,60 @@ type FeatureStoreServices struct {
...
+	FilePersistence *OfflineStoreFilePersistence `json:"file,omitempty"`
+}
+
+// OfflineStorePersistence configures the file-based persistence for the offline store service
+type OfflineStoreFilePersistence struct {
+	// +optional
+	// +default:value=dask
+	// +kubebuilder:validation:Enum=dask;duckdb
+	Type string `json:"type,omitempty"`
 }
```
tchughesiv, line 90: if possible, would be cleaner to assign `OfflineConfigType` directly in the api. then the parse function and checks go away. this would require moving several constants and a var out of services types... to the api types file.
```suggestion
// OfflineStorePersistence configures the file-based persistence for the offline store service
type OfflineStoreFilePersistence struct {
	// +kubebuilder:validation:Enum=dask;duckdb
	Type OfflineConfigType `json:"type,omitempty"`
}

const (
	OfflineRemoteConfigType OfflineConfigType = "remote"
	OfflineDaskConfigType   OfflineConfigType = "dask"
	OfflineDuckDbConfigType OfflineConfigType = "duckdb"
)

// OfflineConfigType provider name or a class name that implements Offline Store
type OfflineConfigType string
```

tchughesiv (changes requested):

infra/feast-operator/api/v1alpha1/featurestore_types.go
tchughesiv: anyway, its worth trying. hopefully it negates the need for parsing function.

infra/feast-operator/internal/controller/services/util.go
tchughesiv: @dmartinol i think we still need to address an empty `Path` or `Type` depending on the feast service type. see the comment directly above this one.

dmartinol: done

infra/feast-operator/internal/controller/suite_test.go
```diff
@@ -67,8 +65,7 @@ var _ = BeforeSuite(func() {
 	}
 
 	var err error
```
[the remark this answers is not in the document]
dmartinol: done

tchughesiv (approved):
lgtm

infra/feast-operator/api/v1alpha1/featurestore_types.go
dmartinol: That check was wrong, we should not check the value from CRD while converting to the one for RepoConfig, they are different things for now.
I added instead a validation function in the API for the supported file types of offline store, and used it in the ApplyDefaults function. Please TAL at the next commit

infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -71,16 +73,65 @@ type FeatureStoreServices struct {
...
+	FilePersistence *OfflineStoreFilePersistence `json:"file,omitempty"`
+}
+
+// OfflineStorePersistence configures the file-based persistence for the offline store service
+type OfflineStoreFilePersistence struct {
+	// +kubebuilder:validation:Enum=dask;duckdb
+	Type string `json:"type,omitempty"`
+}
+
+var validOfflineStoreFilePersistenceType = []string{
```
tchughesiv, line 90: nit ... maybe make this var name plural `validOfflineStoreFilePersistenceTypes`

infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -71,16 +73,65 @@ type FeatureStoreServices struct {
...
+}
+
+// A function to validate the file persistence types for offline stores
+func IsValidOfflineStoreFilePersistenceType(value string) (bool, error) {
+	for _, v := range validOfflineStoreFilePersistenceType {
+		if v == value {
+			return true, nil
+		}
+	}
+	return false, fmt.Errorf("invalid  file type %s for offline store", value)
```
tchughesiv, line 102: let's move this function to the services section and run it early on during the `feast.Deploy()`

infra/feast-operator/internal/controller/featurestore_controller.go
```diff
@@ -78,10 +77,14 @@ func (r *FeatureStoreReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	currentStatus := cr.Status.DeepCopy()
 
 	// initial status defaults must occur before feast deployment
-	applyDefaultsToStatus(cr)
+	if err := services.ApplyDefaultsToStatus(cr); err != nil {
+		logger.Error(err, "Error updating the FeatureStore status")
+		result = ctrl.Result{Requeue: true, RequeueAfter: RequeueDelayError}
+		return result, err
+	}
```
tchughesiv, line 84: let's revert this change and, instead, add the `IsValidOfflineStoreFilePersistenceType` check to `feast.Deploy()` function... which already returns an error to `deployFeast(ctx, cr)` below

infra/feast-operator/internal/controller/services/util.go
```diff
@@ -0,0 +1,79 @@
...
+	cr.Status.FeastVersion = feastversion.FeastVersion
+	applied := cr.Spec.DeepCopy()
+	if applied.Services == nil {
+		applied.Services = &feastdevv1alpha1.FeatureStoreServices{}
+	}
+
+	// default to registry service deployment
+	if applied.Services.Registry == nil {
+		applied.Services.Registry = &feastdevv1alpha1.Registry{}
+	}
```
tchughesiv, line 23: nit ... since we're doing so much here with services, maybe this section (and what follows) cleans up a little if we -
```suggestion
        services := applied.Services
	if services == nil {
		services = &feastdevv1alpha1.FeatureStoreServices{}
	}

	// default to registry service deployment
	if services.Registry == nil {
		services.Registry = &feastdevv1alpha1.Registry{}
	}
```
etc.

tchughesiv (changes requested):
requested a couple changes. also, i'm noticing we're missing new `FeatureStore` CR checks in `featurestore_controller_test.go` for these new status fields and defaults after reconcile

infra/feast-operator/internal/controller/services/util.go
tchughesiv: i realize this is leftover from one of my prior PRs, but now that we're doing so much more with that pointer i think it makes sense

infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -71,16 +73,65 @@ type FeatureStoreServices struct {
...
+	Type string `json:"type,omitempty"`
+}
+
+var validOfflineStoreFilePersistenceType = []string{
+	"dask",
+	"duckdb",
+}
+
+// A function to validate the file persistence types for offline stores
+func IsValidOfflineStoreFilePersistenceType(value string) (bool, error) {
```
tchughesiv, line 96: let's move this function and its corresponding var to the `services` section of code instead, and run it early on during the `feast.Deploy()`

infra/feast-operator/internal/controller/featurestore_controller.go
tchughesiv, line 84: let's revert this change and, instead, add the `IsValidOfflineStoreFilePersistenceType` check to `feast.Deploy()` function... which already returns an error to `deployFeast(ctx, cr)` below. We want to keep the defaults application function simple... no need to get into error handling there.

tchughesiv (changes requested):

infra/feast-operator/api/v1alpha1/featurestore_types.go
tchughesiv: its also worth noting that this should never be an issue as long as the CRD schema validation remains intact. this new `IsValidOfflineStoreFilePersistenceType` function would be backup validation for an extreme edge case.

infra/feast-operator/internal/controller/services/util.go
tchughesiv: that's a addition, and necessary, but its different from ensuring the status defaults are properly being reconciled in the CR.

dmartinol: I see what you mean, I'll add it in the next PR for PVC so there will be more cases to cover

infra/feast-operator/internal/controller/featurestore_controller.go
dmartinol: done, and also removed negative test case in `repo_config_test.go` because, I don't want to test that RepoConfig has  `offline_store.type==invalid` when the CRD has `offline.persistence.file.type==invalid`

infra/feast-operator/internal/controller/services/util.go
```diff
@@ -10,70 +12,74 @@ func IsLocalRegistry(featureStore *feastdevv1alpha1.FeatureStore) bool {
...
 }
+
+func isValidOfflineStoreFilePersistenceType(value string) (bool, error) {
+	for _, v := range feastdevv1alpha1.ValidOfflineStoreFilePersistenceTypes {
+		if v == value {
+			return true, nil
+		}
+	}
+	return false, fmt.Errorf("invalid  file type %s for offline store", value)
+}
```
tchughesiv, line 85: i appears the `bool` return isn't being used anywhere. we should probably either remove it or use it instead of the `error` return. there doesn't currently appear to be a need for both. thoughts?

tchughesiv: probably makes more sense keep the `error` and remove the `bool`. this would necessitate that the function named changed to something like `checkOfflineStoreFilePersistenceType` instead.

tchughesiv (changes requested):

infra/feast-operator/internal/controller/services/util.go
```diff
@@ -0,0 +1,85 @@
...
+		defaultConfigs.Image = &DefaultImage
+	}
+}
+
+func checkOfflineStoreFilePersistenceType(value string) error {
+	for _, v := range feastdevv1alpha1.ValidOfflineStoreFilePersistenceTypes {
+		if v == value {
+			return nil
+		}
+	}
```
tchughesiv, line 83: fyi, starting w/ the version of go we're using (1.21), there's a now a slices package -
```suggestion
	if slices.Contains(feastdevv1alpha1.ValidOfflineStoreFilePersistenceTypes, value) {
		return nil
	}
```

tchughesiv (approved):
i've left a couple nits, but otherwise lgtm

dmartinol:
> > requested a couple changes. also, i'm noticing we're missing new FeatureStore CR checks in featurestore_controller_test.go for these new status fields and defaults after reconcile
> 
> one more commit to add the missing checks

Latest commit includes new controller tests.
Please note that I had to change the resource name because some resources (svc?) are probably not deleted by previous tests, so it would fail due to an attempt to change ownership:
```
Object default/feast-services-offline is already owned by another Service controller feast-services-offline
```
I did not investigate further. I think we can handle this issue in a separate PR.

```
  2024-11-07T17:30:30+01:00     INFO    Successfully reconciled {"Service": "feast-services-offline", "operation": "updated"}
  2024-11-07T17:30:30+01:00     ERROR   Error deploying the FeatureStore client service {"error": "Object default/feast-services-offline is already owned by another Service controller feast-services-offline"}
  github.com/feast-dev/feast/infra/feast-operator/internal/controller/services.(*FeastServices).setFeastServiceCondition
        /Users/dmartino/projects/AI/feast/feast/infra/feast-operator/internal/controller/services/services.go:290
  github.com/feast-dev/feast/infra/feast-operator/internal/controller/services.(*FeastServices).deployFeastServiceByType
        /Users/dmartino/projects/AI/feast/feast/infra/feast-operator/internal/controller/services/services.go:95
  github.com/feast-dev/feast/infra/feast-operator/internal/controller/services.(*FeastServices).Deploy
```

lokeshrangineni (approved):

Issue #4747, opened 2024-11-07
Remove unused (and out-of-place) FeatureStore.version function


dandawg:
The `FeatureStore` class has a `version` function which returns the Python SDK version. This gives the illusion that it is needed in the class, but it is never actually used. When I first saw it, I was confused why the class would have a version (which is why I dug into the source code). I propose simplifying the FeatureStore class by dropping the `FeatureStore.version` function. 

The benefit will be making the code easier to understand, with less opportunity for error.

## Expected Behavior
I expect the FeatureStore class not to have a version function, as it is not used and not needed.

Even if it were needed, the better way to get the version would be either directly through the function it calls (`version.get_version` -- also part of the SDK), or by calling `feast.__version__`.

## Current Behavior

The `FeatureStore` class defines a `version` function that is never used.
https://github.com/feast-dev/feast/blob/e9cd3733f041da99bb1e84843ffe5af697085c34/sdk/python/feast/feature_store.py#L174

### Specifications

- Version: 0.41.3
- Platform: NA
- Subsystem: NA

PR #4748, opened 2024-11-07
fix: Removed version func from feature_store.py

dandawg:

This PR removes the `version` function from `feature_store.py`, which is a method to the `FeatureStore` class. The `version` method was unused, un-needed, and confusing. See issue #4747 for more details.
 
I categorized this as a "fix" because I see an unused/un-needed function as a bug, but I understand that this could merit a different categorization.

#4747

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -88,7 +88,6 @@ from feast.repo_contents import RepoContents
 from feast.saved_dataset import SavedDataset, SavedDatasetStorage, ValidationReference
 from feast.stream_feature_view import StreamFeatureView
 from feast.utils import _utc_now
-from feast.version import get_version
 
 warnings.simplefilter("once", DeprecationWarning)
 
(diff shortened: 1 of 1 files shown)


shuchu (approved):
Lgtm

dandawg:
I used vscode's 'references' functionality to verify that this function was unused in the repo. To double check this, I also did the following:

```
# search for all mentions of version(
# excluding ones like ..._version(, ...conversion(, etc.
# also excludes lines with venv (my virtual env files)
grep -R '[^A-z_]version(' . | grep -v "venv" 
```
Output:
```
./sdk/python/feast/feature_store.py:    def version(self) -> str:
./sdk/python/feast/version.py:        sdk_version = version("feast")
./sdk/python/feast/cli.py:def version():
./build/lib/feast/feature_store.py:    def version(self) -> str:
./build/lib/feast/version.py:        sdk_version = version("feast")
./build/lib/feast/cli.py:def version():
```

I manually verified that `./sdk/python/feast/version.py` and `./build/lib/feast/version.py` are not using `feature_store.FeatureStore.version`.

franciscojavierarceo (approved):
Yeah I can see how this is convenient but I definitely agree this should just rely on package version.

Thank you!

PR #4750, opened 2024-11-08
feat: PVC configuration and impl

dmartinol:
Adds durable persistence configuration to Feast services deployed with the operator.

PVC configuration allows to either create a new PVC or use an existing one:
```yaml
kind: FeatureStore
metadata:
  name: sample-pvc-persistence
spec:
  feastProject: my_project
  services:
    onlineStore:
      persistence:
        file:
          path: online_store.db
          pvc:
            ref:
              name: online-pvc
            mountPath: /data/online
    offlineStore:
      persistence:
        file:
          type: duckdb
          pvc:
            create:
              storageClassName: standard
              resources:
                requests:
                  storage: 5Gi
            mountPath: /data/offline
    registry:
      local:
        persistence:
          file:
            path: registry.db
            pvc:
              create: {}
              mountPath: /data/registry
```


dmartinol:
@tchughesiv @tmihalac I also splitted the controller tests in multiple files to simplify them

infra/feast-operator/config/samples/v1alpha1_featurestore_ephemeral_persistence.yaml
```diff
@@ -6,15 +6,18 @@ spec:
...
         file:
           path: /data/online_store.db
     offlineStore:
+      image: quay.io/dmartino/feature-server:0.2
       persistence:
         file:
           type: dask
     registry:
       local:
+        image: quay.io/dmartino/feature-server:0.2
```
tchughesiv, line 20: these need to be reverted before we merge

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -135,6 +160,24 @@ func (feast *FeastServices) createDeployment(feastType FeastServiceType) error {
...
 
+func (feast *FeastServices) createPVC(pvcCreate *feastdevv1alpha1.PvcCreate, feastType FeastServiceType) error {
+	logger := log.FromContext(feast.Context)
+	pvc := feast.initPVC(feastType)
+	err := feast.setPVC(pvc, pvcCreate, feastType)
+	if err != nil {
+		return err
+	}
+	if op, err := controllerutil.CreateOrUpdate(feast.Context, feast.Client, pvc, controllerutil.MutateFn(func() error {
+		return nil
```
tchughesiv, line 171:
```suggestion
	if op, err := controllerutil.CreateOrUpdate(feast.Context, feast.Client, pvc, controllerutil.MutateFn(func() error {
		return feast.setPVC(pvc, pvcCreate, feastType)
```

tchughesiv (changes requested):

infra/feast-operator/internal/controller/services/services.go
tchughesiv: the PVC object mods need to happen within the `CreateOrUpdate func ()`

tchughesiv: was this moved because of the immutable nature of the PVC?

tchughesiv: if so, we could be explicit with PVC creation and use `feast.Client.Create()` here instead?

dmartinol: Tried the Create option but creation was not an issue. Deletion seems to fail, at least in the tests (just restored them to see the failure)

infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -118,8 +122,40 @@ type RegistryPersistence struct {
 }
 
 // RegistryFilePersistence configures the file-based persistence for the registry service
+// +kubebuilder:validation:XValidation:rule="(!has(self.pvc) && has(self.path)) ? self.path.startsWith('/') : true",message="Ephemeral stores must have absolute paths."
+// +kubebuilder:validation:XValidation:rule="(has(self.pvc) && has(self.path)) ? !self.path.startsWith('/') : true",message="PVC path must be a file name only, with no slashes."
 type RegistryFilePersistence struct {
-	Path string `json:"path,omitempty"`
+	Path      string     `json:"path,omitempty"`
+	PvcConfig *PvcConfig `json:"pvc,omitempty"`
+}
```
tchughesiv, line 130: `OnlineStoreFilePersistence` and `RegistryFilePersistence` are identical. Any reason we shouldn't make a single `FilePersistence` struct instead?

dmartinol: Registry also supports the `s3_additional_kwargs` options (which will go under the `Registry.Persistence.File` section). Will we add it to a new PR dedicated to object store settings.

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -53,6 +53,11 @@ func (feast *FeastServices) Deploy() error {
 			if err := feast.deployFeastServiceByType(OfflineFeastType); err != nil {
 				return err
 			}
+			if _, shouldCreate := shouldCreatePvc(feast.FeatureStore, OfflineFeastType); !shouldCreate {
+				if err := feast.deleteOwnedFeastObj(feast.initPVC(OfflineFeastType)); err != nil && !apierrors.IsNotFound(err) {
+					return err
+				}
+			}
```
tchughesiv, line 60: let's move this block of code to the `deployFeastServiceByType` function instead... maybe as an `else {` to what you've already added?

tchughesiv (changes requested):

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -63,6 +68,11 @@ func (feast *FeastServices) Deploy() error {
 			if err := feast.deployFeastServiceByType(OnlineFeastType); err != nil {
 				return err
 			}
+			if _, shouldCreate := shouldCreatePvc(feast.FeatureStore, OnlineFeastType); !shouldCreate {
+				if err := feast.deleteOwnedFeastObj(feast.initPVC(OnlineFeastType)); err != nil && !apierrors.IsNotFound(err) {
+					return err
+				}
+			}
```
tchughesiv, line 75: same

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -73,6 +83,11 @@ func (feast *FeastServices) Deploy() error {
 			if err := feast.deployFeastServiceByType(RegistryFeastType); err != nil {
 				return err
 			}
+			if _, shouldCreate := shouldCreatePvc(feast.FeatureStore, RegistryFeastType); !shouldCreate {
+				if err := feast.deleteOwnedFeastObj(feast.initPVC(RegistryFeastType)); err != nil && !apierrors.IsNotFound(err) {
+					return err
+				}
+			}
```
tchughesiv, line 90: same

infra/feast-operator/api/v1alpha1/featurestore_types.go
tchughesiv: ok, so this is in prep for future work... got it, thanks

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -104,6 +124,11 @@ func (feast *FeastServices) removeFeastServiceByType(feastType FeastServiceType)
 	if err := feast.deleteOwnedFeastObj(feast.initFeastSvc(feastType)); err != nil {
 		return err
 	}
+	if _, shouldCreate := shouldCreatePvc(feast.FeatureStore, feastType); shouldCreate {
```
tchughesiv, line 127: we prob don't need the `shouldCreate` check here. this `removeFeastServiceByType` function indicates that everything related to that feast service type should be removed.

tchughesiv (changes requested):

infra/feast-operator/internal/controller/services/services.go
dmartinol: I initially was doing the same, but got worried about mixing a deletion during a deploy. If you are not worried about that, I can do as you suggest

tchughesiv: we're technically already mixing anyway... all good i think

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -94,6 +94,15 @@ func (feast *FeastServices) deployFeastServiceByType(feastType FeastServiceType)
 	if err := feast.createDeployment(feastType); err != nil {
 		return feast.setFeastServiceCondition(err, feastType)
 	}
+	if pvcCreate, shouldCreate := shouldCreatePvc(feast.FeatureStore, feastType); shouldCreate {
+		if err := feast.createPVC(pvcCreate, feastType); err != nil {
+			return feast.setFeastServiceCondition(err, feastType)
+		}
+	} else {
+		if err := feast.deleteOwnedFeastObj(feast.initPVC(feastType)); err != nil && !apierrors.IsNotFound(err) {
```
tchughesiv, line 102: no need for `IsNotFound` check here... we already `return nil` on this in the `deleteOwnedFeastObj` function
```suggestion
		if err := feast.deleteOwnedFeastObj(feast.initPVC(feastType)); err != nil {
```

tchughesiv (changes requested):

infra/feast-operator/internal/controller/services/services.go
tchughesiv: thinking about this more... do we even want to return an error here ? do we care if the delete doesn't succeed? is that worth halting the FeatureStore's reconciliation process or health of the stack? 
maybe we just `_ = feast.deleteOwnedFeastObj(feast.initPVC(feastType))` ??

thoughts?

tchughesiv (approved):
lgtm

dmartinol:
@lokeshrangineni @redhatHameed can you take a look?

infra/feast-operator/config/samples/v1alpha1_featurestore_pvc_persistence.yaml
```diff
@@ -0,0 +1,45 @@
...
+                  storage: 5Gi
+            mountPath: /data/offline
+    registry:
+      local:
+        persistence:
+          file:
+            path: registry.db
+            pvc:
+              create: {}
+              mountPath: /data/registry
```
redhatHameed, line 34: why `create` section empty here, is there related docs added for this configuration. and looks the settings not consistent with each type

For example registry 
```        persistence:
          file:
            path: registry.db
            pvc:
              create: {}
              mountPath: /data/registry
```
online
```        file:
          path: online_store.db
          pvc:
            ref:
              name: online-pvc
            mountPath: /data/online

```

infra/feast-operator/internal/controller/services/util.go
```diff
@@ -6,13 +6,37 @@ import (
...
 
 func isLocalRegistry(featureStore *feastdevv1alpha1.FeatureStore) bool {
 	appliedServices := featureStore.Status.Applied.Services
 	return appliedServices != nil && appliedServices.Registry != nil && appliedServices.Registry.Local != nil
 }
 
+func hasPvcConfig(featureStore *feastdevv1alpha1.FeatureStore, feastType FeastServiceType) (*feastdevv1alpha1.PvcConfig, bool) {
+	services := featureStore.Status.Applied.Services
+	var pvcConfig *feastdevv1alpha1.PvcConfig = nil
+	if feastType == OnlineFeastType && services.OnlineStore != nil && services.OnlineStore.Persistence.FilePersistence != nil {
```
redhatHameed, line 21: not an issue, how about considering `switch` statment here
   ```
 switch feastType {
    case OnlineFeastType:
        if services.OnlineStore != nil {  
```

**Review by redhatHameed — COMMENTED** [review]
Thank `lgtm` some minor comments.

File: infra/feast-operator/config/samples/v1alpha1_featurestore_pvc_persistence.yaml (hunk 9 above)
Comment (tchughesiv) [reply infra/feast-operator/config/samples/v1alpha1_featurestore_pvc_persistence.yaml:34]: i think he's demonstrating 3 different types of pvc configurations. all 3 feast service types can be configured similarly. an empty `create:` tells the operator that the user wants a pvc, but doesn't want to define it. the operator will apply defaults during the creation.

File: infra/feast-operator/internal/controller/services/services.go [hunk 11]
```diff
@@ -220,6 +252,16 @@ func (feast *FeastServices) setService(svc *corev1.Service, feastType FeastServi
...
+func (feast *FeastServices) createNewPVC(pvcCreate *feastdevv1alpha1.PvcCreate, feastType FeastServiceType) (*corev1.PersistentVolumeClaim, error) {
+	pvc := feast.initPVC(feastType)
+
+	pvc.Spec = corev1.PersistentVolumeClaimSpec{
+		AccessModes: []corev1.PersistentVolumeAccessMode{corev1.ReadWriteMany},
+		Resources:   pvcCreate.Resources,
+	}
+	return pvc, controllerutil.SetControllerReference(feast.FeatureStore, pvc, feast.Scheme)
+}
+
```
tchughesiv, line 264: is `StorageClassName` set elsewhere? i can't seem to find where its being set.

infra/feast-operator/internal/controller/services/util.go
dmartinol: @tchughesiv ? I also thought of using the `switch` but I wanted to be consistent with the existing code

tchughesiv: i think that would be a better approach, so def go for it

infra/feast-operator/config/samples/v1alpha1_featurestore_objectstore_persistence.yaml
```diff
@@ -0,0 +1,24 @@
...
+          type: dask
+    registry:
+      local:
+        persistence:
+          file:
+            path: s3://bucket/registry.db
+            s3_additional_kwargs:
+              ServerSideEncryption: AES256
+              ACL: bucket-owner-full-control
+              CacheControl: max-age=3600
```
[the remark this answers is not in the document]
dmartinol: this file should not be in the PR, will remove it

infra/feast-operator/config/samples/v1alpha1_featurestore_pvc_persistence.yaml
dmartinol: There is no doc apart from the CRD definition, which has some comments on it.
wrt your questions:
- pvc can have either create or ref, and create has default fields in case they are omitted
Anyway, to avoid misunderstandings, I will add the other settings too and avoid defaults

tchughesiv: imo, CRD should be the doc of choice for API. @dmartinol i like the file in its current state, shows the user all their options.

infra/feast-operator/internal/controller/services/services.go
dmartinol: good catch. will add a test for it

infra/feast-operator/config/samples/v1alpha1_featurestore_pvc_persistence.yaml
dmartinol: agree, reverted previous change

infra/feast-operator/internal/controller/featurestore_controller_pvc_test.go
```diff
@@ -67,6 +67,8 @@ var _ = Describe("FeatureStore Controller-Ephemeral services", func() {
 		onlineStoreMountPath := "/online"
 		registryMountPath := "/registry"
 
+		storageClassName := "default"
```
tchughesiv, line 70: it's a bit confusing to use "default" as the `storageClassName` here, since you're testing a scenario where a user actually defines the storage class name... as opposed to leaving it blank and using the default. maybe use a different name? "test" ?

tchughesiv (changes requested):

redhatHameed (approved):

PR #4764, opened 2024-11-16
chore: Upgrade react-router-dom and use-query-params to latest in /ui

peruukki:

Upgrades react-router-dom and use-query-params to their latest versions; the latter is upgraded because its older version didn't work properly with the latest react-router-dom.

One small change in EuiCustomLink is done to avoid the error reported in #3794, I tried clicking everywhere in the UI and didn't see the error after the change.

This allows to use the latest react-router-dom in a consuming app in case someone is using it, no longer being restricted to <6.4.0 due to #3794.

# Misc

After the upgrade, react-router-dom writes a warning console message like this:
> ⚠️ React Router Future Flag Warning: Relative route resolution within Splat routes is changing in v7. You can use the `v7_relativeSplatPath` future flag to opt-in early. For more information, see https://reactrouter.com/v6/upgrading/future#v7_relativesplatpath.

I tried enabling the flag but it caused issues with our current custom tab routes, and I couldn't sort it out with reasonable effort at this point. Also, since the feature can require non-trivial changes, enabling it _might_ break something in consuming apps that use react-router-dom, so maybe better to leave it later anyway.

I enabled another future flag that I saw a similar console warning about, `v7_startTransition`, since that seems [much safer](https://reactrouter.com/en/6.28.0/upgrading/future#v7_starttransition), though I did need to add one `await` in the tests after that.


ui/src/components/EuiCustomLink.jsx
```diff
@@ -39,7 +39,7 @@ export default function EuiCustomLink({ to, ...rest }) {
   }
 
   // Generate the correct link href (with basename accounted for)
-  const href = useHref({ pathname: to });
+  const href = useHref(to);
```
peruukki, line 42: Copied from react-router-dom's Link component: https://github.com/remix-run/react-router/blob/dd323b6dec531811ce80595ac0521455b46f0bc2/packages/react-router-dom/index.tsx#L999.

franciscojavierarceo (approved):

peruukki:
Do we need a label from someone to run the last check, build-docker-image-java (feature-server-java)?

peruukki:
Could we get an `ok-to-test` label for this to move it forward? 🙏 Don't hesitate to ask if you have doubts about the upgrade.

peruukki:
A friendly reminder: could someone add a required label to this PR to run the pending check? 🙏

peruukki:
Thanks @franciscojavierarceo!

Issue #4743, opened 2024-11-06
Unable to load UI in 0.41.3


t0a009r:
## Expected Behavior 
I expect to be able to init a feast repo, apply the example changes, and then load up the UI successfully. 

## Current Behavior
The UI is unable to load and returns an error.
![Screenshot 2024-11-06 at 5 26 22 PM](https://github.com/user-attachments/assets/24f6d8b2-019a-4536-8075-8ea881ff3f99)

## Steps to reproduce
```
$ feast version
Feast SDK Version: "0.41.3"
(.venv) [05:25 PM] ~ 
$ feast init demo 

Creating a new Feast repository in /Users/t0a009r/demo.

(.venv) [05:25 PM] ~ 
(.venv) [05:25 PM] ~ 
$ cd demo/feature_repo/
(.venv) [05:25 PM] ~/demo/feature_repo 
(.venv) [05:25 PM] ~/demo/feature_repo 
$ feast apply
Applying changes for project demo
...
RuntimeWarning: On demand feature view is an experimental feature. This API is stable, but the functionality does not scale well for offline retrieval
  warnings.warn(
11/06/2024 05:25:46 PM root WARNING: Cannot use sqlite_vec for vector search
11/06/2024 05:25:46 PM root WARNING: Cannot use sqlite_vec for vector search
11/06/2024 05:25:46 PM root WARNING: Cannot use sqlite_vec for vector search
11/06/2024 05:25:46 PM root WARNING: Cannot use sqlite_vec for vector search
Created project demo
Created entity driver
Created feature view driver_hourly_stats
Created feature view driver_hourly_stats_fresh
Created on demand feature view transformed_conv_rate_fresh
Created on demand feature view transformed_conv_rate
Created feature service driver_activity_v2
Created feature service driver_activity_v3
Created feature service driver_activity_v1

11/06/2024 05:25:46 PM root WARNING: Cannot use sqlite_vec for vector search
11/06/2024 05:25:46 PM root WARNING: Cannot use sqlite_vec for vector search
11/06/2024 05:25:46 PM root WARNING: Cannot use sqlite_vec for vector search
11/06/2024 05:25:46 PM root WARNING: Cannot use sqlite_vec for vector search
Created sqlite table demo_driver_hourly_stats_fresh
Created sqlite table demo_driver_hourly_stats

(.venv) [05:25 PM] ~/demo/feature_repo 
$ feast ui
...
DeprecationWarning: 
[...]

PR #4772, opened 2024-11-20
fix: Read project data from the 'projects' key while loading the registry state in the Feast UI

boliri:
Starting from 0.41.x releases, the Feast UI is no longer able to load the project summary in the projects overview / landing page. The reason why this is happening is that, after some changes from https://github.com/feast-dev/feast/pull/4475 got merged into `master`, the `Registry` class is unable to build and persist the project's metadata to the backend used as the registry through its method `_prepare_registry_for_changes` - this behavior has been observed in, at least, the `file` registry, but it might be applicable to others as well.

Since project's metadata is not generated, and the `useLoadRegistry` function from the UI is expecting those metadata to exist, the UI fails to render the Project Overview page and hints that the issue might be in the `feature_store.yaml`, which is not the case.

The purpose of this PR is pretty simple: make the UI extract the project information from the `projects` array instead of the `projectMetadata` object while it is refreshing the registry state. With this tiny fix in place, the UI renders components as usual.

Fixes https://github.com/feast-dev/feast/issues/4743


boliri:
I regenerated the `registry.db` file locally and, after running `yarn test`, the UI tests passed without issues. However, there's still something wrong in the GitHub Action for these tests as it cannot finish properly:

```
Error: src/FeastUISansProviders.test.tsx(58,5): error TS2533: Object is possibly 'null' or 'undefined'.
Error: src/queries/useLoadRegistry.ts(55,22): error TS2533: Object is possibly 'null' or 'undefined'.
```

This might be related to how the proto representation of the Registry is decoded in the UI tests, but I'm not 100% sure.

Thoughts?

franciscojavierarceo (approved):

boliri:
Just a reminder: seems like the Python integration tests job failed, might be related to an execution timeout.

redhatHameed:
@franciscojavierarceo can we merge this

redhatHameed (approved):

PR #4779, opened 2024-11-21
feat: Improve exception message for unsupported Snowflake data types

nanohanno:

For Snowflake data sources, there it is checked if all columns in a table have data types that are supported. If not an exception is thrown with limited information. 

These changes add more information about which column and datatype are affected. This information will help in fixing the underlying table schema.

# Misc


shuchu (approved):
lgtm

nanohanno:
Who could unblock the pending checks? Seems like the build-docker-image-java task was skipped but has not reported back here? @shuchu are you able to do that?

PR #4781, opened 2024-11-21
fix: Reduce feast-server container image size & fix dev image build

tchughesiv:
Improve the Dockerfile to remove unnecessary packages and do some cleanup. This will reduce the current image size by over 300mb.
```shell
$ podman images
REPOSITORY                         TAG         IMAGE ID      CREATED        SIZE
quay.io/tchughesiv/feature-server  0.41.3      84dc7acaad64  2 minutes ago  1.04 GB
docker.io/feastdev/feature-server  0.41.3      1bbf34c192d9  3 weeks ago    1.37 GB
```

This is done by switching to the official python3 debian 11 slim image (`python:3.11-slim-bullseye`) which handles some of cleanup around python installation. We also tell pip install not to use cache and remove the `build-essential` package. Finally, we do some additional cleanup around apt.

Additionally, this PR fixes `Dockerfile.dev` so that it builds local dev changes into a functional container image, as was its intent.

Fixes https://github.com/feast-dev/feast/issues/4784

### DIFF
--- a/sdk/python/feast/infra/feature_servers/multicloud/Dockerfile.dev
+++ b/sdk/python/feast/infra/feature_servers/multicloud/Dockerfile.dev
@@ -1,24 +1,28 @@
-FROM debian:11-slim
+FROM python:3.11-slim-bullseye
 
-RUN apt update && \
-        apt install -y \
-        jq \
-        python3 \
-        python3-pip \
-        python3-dev \
-        build-essential
+RUN pip install --no-cache-dir pip --upgrade
+RUN pip install --no-cache-dir pip-tools
 
-RUN pip install pip --upgrade
-COPY . .
+RUN apt update && apt install -y -V ca-certificates lsb-release wget make git curl gcc && \
+    curl -sL https://deb.nodesource.com/setup_20.x | bash - && \
+    wget https://apache.jfrog.io/artifactory/arrow/$(lsb_release --id --short | tr 'A-Z' 'a-z')/apache-arrow-apt-source-latest-$(lsb_release --codename --short).deb && \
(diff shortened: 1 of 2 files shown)


sdk/python/feast/infra/feature_servers/multicloud/Dockerfile.dev
```diff
@@ -1,24 +1,28 @@
...
-        python3-dev \
-        build-essential
+RUN pip install --no-cache-dir pip --upgrade
+RUN pip install --no-cache-dir pip-tools
 
-RUN pip install pip --upgrade
-COPY . .
+RUN apt update && apt install -y -V ca-certificates lsb-release wget make git curl gcc && \
+    curl -sL https://deb.nodesource.com/setup_20.x | bash - && \
+    wget https://apache.jfrog.io/artifactory/arrow/$(lsb_release --id --short | tr 'A-Z' 'a-z')/apache-arrow-apt-source-latest-$(lsb_release --codename --short).deb && \
```
dmartinol, line 8: one of these days we should investigate on the reason why these arrow source code bundle for debian is installed.
AFAIK, installing the pyarrow python package is the only thing we need to connect an arrow system.
I will track a GH issue to investigate and in case clear the dockerfile.

dmartinol (approved):

sdk/python/feast/infra/feature_servers/multicloud/Dockerfile.dev
dmartinol: https://github.com/feast-dev/feast/issues/4785

Issue #4765, opened 2024-11-18
RBAC configuration in the Feast operator


dmartinol:
**Is your feature request related to a problem? Please describe.**
Feast services currently support RBAC configuration via the `feature_store.yaml`, which is managed by the Feast Operator in Kubernetes environments. To enable proper authorization setup by feature store administrators, the Feast Operator should also provide capabilities to define and enforce consistent RBAC settings across all services. 
Additionally, sensitive information such as IDP server credentials should be protected using encoded secrets within the deployment namespace.

This issue relates to #4561.

**Describe the solution you'd like**
Authorization settings must be given in the `FeatureStore` custom resource and be propagated by the operator to the deployed Feast services.

**Describe alternatives you've considered**
Since the operator reconciles all changes to the deployed applications to match the latest state of the `FeatureStore` resource, there are no alternatives that we could consider in order to inject the authorization settings, located in the `feature_store.yaml`, into the Feast services.

**Additional context**
* [RBAC definitions in Feast](https://docs.feast.dev/master/getting-started/concepts/permission)
* [Feast authorization settings](https://docs.feast.dev/master/getting-started/components/authz_manager)

dmartinol:
Related PRs:
- #4786 
- #4801

PR #4786, opened 2024-11-22
feat: RBAC Authorization in Feast Operator

dmartinol:
Adding support to define the [kubernetes](https://docs.feast.dev/master/getting-started/components/authz_manager#kubernetes-rbac-authorization) authorization manager with the Feast Operator.
* All services are configured to adopt this authorization manager.
* Services runs with a `ServiceAccount` that is bound to a newly created `Role` allowing to `get, list, watch` the other `Roles` and `RoleBindings` in the same namespace.
* Admins are not requested to perform any manual configuration once the custom resource is installed. 

Sample manifest to configure the deployments:
```yaml
kind: FeatureStore
metadata:
  name: sample-kubernetes-auth
spec:
  feastProject: my_project
  services:
    onlineStore: {}
    offlineStore: {}
    registry: {}
  auth:
    kubernetes:
      roles:
        - reader
        - writer
```

Relates to https://github.com/feast-dev/feast/issues/4765
Next PR will add support for the OIDC authorization.


dmartinol:
@tchughesiv @redhatHameed @tmihalac

infra/feast-operator/internal/controller/services/util.go
```diff
@@ -45,6 +45,14 @@ func shouldCreatePvc(featureStore *feastdevv1alpha1.FeatureStore, feastType Feas
 func ApplyDefaultsToStatus(cr *feastdevv1alpha1.FeatureStore) {
 	cr.Status.FeastVersion = feastversion.FeastVersion
 	applied := cr.Spec.DeepCopy()
+
+	if applied.AuthConfig == nil {
+		applied.AuthConfig = &feastdevv1alpha1.AuthConfig{}
+	}
+	if applied.AuthConfig.KubernetesAuth != nil && applied.AuthConfig.KubernetesAuth.Roles == nil {
+		applied.AuthConfig.KubernetesAuth.Roles = []string{}
```
tchughesiv, line 53: curious why we would default to an empty slice here?

tchughesiv (changes requested):
thanks for this! only a few nits so far

infra/feast-operator/internal/controller/auth/auth_types.go
```diff
@@ -0,0 +1,30 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// FeastAuth is an interface for configuring feast authorization
+type FeastAuth struct {
+	Handler handler.FeastHandler
+}
+
+var (
+	nameLabelKey = feastdevv1alpha1.GroupVersion.Group + "/name"
```
tchughesiv, line 15: i'd like to only have this set once between `auth` and `services` package so we can ensure the labels stay in sync. any ideas on a location where both could import from?

tchughesiv (changes requested):

infra/feast-operator/internal/controller/services/services_types.go
```diff
@@ -67,15 +66,15 @@ var (
 
 	FeastServiceConstants = map[FeastServiceType]deploymentSettings{
 		OfflineFeastType: {
-			Command:    []string{"feast", "serve_offline", "-h", "0.0.0.0"},
+			Command:    []string{"feast", "--log-level", "DEBUG", "serve_offline", "-h", "0.0.0.0"},
```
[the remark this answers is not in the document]
tchughesiv: though we probably need to make this configurable in the CR

infra/feast-operator/internal/controller/services/util.go
dmartinol: just thinking it could help in range loops, but I see it's not causing any harms. removing it, thanks

infra/feast-operator/internal/controller/auth/auth_types.go
dmartinol: Changed this to reuse `services.NameLabelKey` instead. I'm not 100% convinced of that, but the `auth` package is also using other APIs and resources from the `services` package:
```
services.GetFeastName
services.GetFeastServiceName
services.NameLabelKey
```

We can move them in a common package or under `controller`, or just accept that `auth` depends on `services` and try to avoid cyclic dependencies.

infra/feast-operator/internal/controller/services/services_types.go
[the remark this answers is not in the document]
dmartinol: Agree! I already suggested you to add a new story/task under the epic for troubleshooting support, this would definitely help as logging is really poor at the default level.

infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -28,26 +28,29 @@ const (
...
-	OfflineStoreReadyType = "OfflineStore"
-	OnlineStoreReadyType  = "OnlineStore"
-	RegistryReadyType     = "Registry"
-	ReadyType             = "FeatureStore"
+	ClientReadyType         = "Client"
+	OfflineStoreReadyType   = "OfflineStore"
+	OnlineStoreReadyType    = "OnlineStore"
+	RegistryReadyType       = "Registry"
+	ReadyType               = "FeatureStore"
+	KubernetesAuthReadyType = "KubernetesAuth"
```
tchughesiv, line 36: with oidc auth support, will the operator deploy anything? would a separate auth condition be set or would the k8s auth condition just be removed? 

if oidc will deploy anything, we should make this condition type more generic so both can use it. maybe `Authentication`. if this is only used for k8s auth, i think we're good as is.

infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -205,6 +209,15 @@ type OptionalConfigs struct {
...
 }
 
+// AuthConfig defines the authorization settings for the deployed Feast services.
+type AuthConfig struct {
+	KubernetesAuth *KubernetesAuth `json:"kubernetes,omitempty"`
+}
+
+type KubernetesAuth struct {
+	Roles []string `json:"roles,omitempty"`
+}
```
tchughesiv, line 219: This needs a description via comment. `Roles` could use a description as well, along with details on any default we might set (if any). Maybe even explaining to the user that these correlate into deployed `Role` objects?

infra/feast-operator/internal/controller/auth/auth.go
```diff
@@ -0,0 +1,220 @@
...
+		})
+	}
+	if auth.Handler.FeatureStore.Status.Applied.Services.OnlineStore != nil {
+		roleBinding.Subjects = append(roleBinding.Subjects, rbacv1.Subject{
+			Kind:      rbacv1.ServiceAccountKind,
+			Name:      services.GetFeastServiceName(auth.Handler.FeatureStore, services.OnlineFeastType),
+			Namespace: auth.Handler.FeatureStore.Namespace,
+		})
+	}
+	if auth.Handler.FeatureStore.Status.Applied.Services.Registry != nil {
```
tchughesiv, line 149: should this only be configured if a local registry is being deployed? if the user is leveraging a remote registry, it would be on them to handle inter-ns/cluster rbac configs/perms for registry access.
```suggestion
	if servces.isLocalRegistry(auth.Handler.FeatureStore) {
```

tchughesiv (changes requested):

infra/feast-operator/api/v1alpha1/featurestore_types.go
dmartinol: OIDC will not deploy anything but has to validate that the linked secret includes all the required parameters as secret keys. I was thinking of using the condition to mark the completion state of this step, do you think it's only meant to track the actual resource deployments instead?

tchughesiv: @dmartinol no, i think your line of thinking is a fair approach. in that case though, we should probably make the ReadyType more generic.

infra/feast-operator/internal/controller/auth/auth.go
dmartinol: u r right, thanks!

infra/feast-operator/api/v1alpha1/featurestore_types.go
dmartinol: What about a single `ReadyType` and two distinct `AuthFailedReason` and `AuthReadyMessage`?

infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -28,26 +28,29 @@ const (
...
-	OnlineStoreFailedReason  = "OnlineStoreDeploymentFailed"
-	RegistryFailedReason     = "RegistryDeploymentFailed"
-	ClientFailedReason       = "ClientDeploymentFailed"
+	ReadyReason                = "Ready"
+	FailedReason               = "FeatureStoreFailed"
+	OfflineStoreFailedReason   = "OfflineStoreDeploymentFailed"
+	OnlineStoreFailedReason    = "OnlineStoreDeploymentFailed"
+	RegistryFailedReason       = "RegistryDeploymentFailed"
+	ClientFailedReason         = "ClientDeploymentFailed"
+	KubernetesAuthFailedReason = "KubernetesAuthorizationDeploymentFailed"
```
tchughesiv, line 45:
```suggestion
	KubernetesAuthzFailedReason = "KubernetesAuthorizationDeploymentFailed"
```

infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -28,26 +28,29 @@ const (
...
-	OfflineStoreReadyMessage = "Offline Store installation complete"
-	OnlineStoreReadyMessage  = "Online Store installation complete"
-	RegistryReadyMessage     = "Registry installation complete"
-	ClientReadyMessage       = "Client installation complete"
+	ReadyMessage               = "FeatureStore installation complete"
+	OfflineStoreReadyMessage   = "Offline Store installation complete"
+	OnlineStoreReadyMessage    = "Online Store installation complete"
+	RegistryReadyMessage       = "Registry installation complete"
+	ClientReadyMessage         = "Client installation complete"
+	KubernetesAuthReadyMessage = "Kubernetes authorization installation complete"
```
tchughesiv, line 53:
```suggestion
	KubernetesAuthzReadyMessage = "Kubernetes authorization installation complete"
```

infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -205,6 +209,23 @@ type OptionalConfigs struct {
 	Resources       *corev1.ResourceRequirements `json:"resources,omitempty"`
 }
 
+// AuthzConfig defines the authorization settings for the deployed Feast services.
+type AuthzConfig struct {
+	KubernetesAuth *KubernetesAuth `json:"kubernetes,omitempty"`
```
tchughesiv, line 214:
```suggestion
	KubernetesAuthz *KubernetesAuthz `json:"kubernetes,omitempty"`
```

infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -205,6 +209,23 @@ type OptionalConfigs struct {
...
+}
+
+// KubernetesAuth provides a way to define the authorization settings using Kubernetes RBAC resources.
+// https://kubernetes.io/docs/reference/access-authn-authz/rbac/
+type KubernetesAuth struct {
```
tchughesiv, line 219:
```suggestion
// KubernetesAuthz provides a way to define the authorization settings using Kubernetes RBAC resources.
// https://kubernetes.io/docs/reference/access-authn-authz/rbac/
type KubernetesAuthz struct {
```

infra/feast-operator/internal/controller/services/services_types.go
```diff
@@ -136,6 +135,9 @@ var (
 	}
 )
 
+// AuthType defines the authorization type
+type AuthType string
```
tchughesiv, line 139:
```suggestion
type AuthzType string
```

tchughesiv (changes requested):
a few nits ... otherwise lgtm

tchughesiv (approved):

dmartinol:
@feast-dev/reviewers-and-approvers please TAL

lokeshrangineni (approved):

redhatHameed:
@dmartinol I am document this feature using operator, what is the usage of the `roles` here what if the client application deployed in different namespace ?

dmartinol:
> @dmartinol I am document this feature using operator, what is the usage of the `roles` here what if the client application deployed in different namespace ?

The RBAC policy checks that the role, in the Feast namespace, is bound to the SA in the client namespace. This is how it works

redhatHameed:
> > @dmartinol I am document this feature using operator, what is the usage of the `roles` here what if the client application deployed in different namespace ?
> 
> The RBAC policy checks that the role, in the Feast namespace, is bound to the SA in the client namespace. This is how it works

but as per the example these role are created as namespace scope. How it will be bind the client namespace SA ? 

let suppose I have installed feast using operator in namespace `feast` and I have client application called `feast-client` how we will bind these role in this namespace with SA ?

tchughesiv:
@redhatHameed here's how you'd create a roleBinding for an SA in another ns -
```yaml
kind: RoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: feast-authz-client
  namespace: feast
subjects:
  - kind: ServiceAccount
    name: feast-sample-client
    namespace: feast-client
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: feast-sample-kubernetes-auth
```

tchughesiv:
@redhatHameed here's how you'd create the binding via command-line -
```bash
$ kubectl create rolebinding feast-authz-client -n feast --role=feast-sample-kubernetes-auth --serviceaccount=feast-client:feast-sample-client
```

redhatHameed:
@tchughesiv @dmartinol let me try with this. Thanks

Issue #4793, opened 2024-11-26
Leftovers of PR #4771


dmartinol:
This issue is defined to track the leftovers of #4771 that must be closed before the next release:
- infra/feast-operator/config/samples/v1alpha1_featurestore_db_persistence.yaml: add sample secret [link](https://github.com/feast-dev/feast/pull/4771#discussion_r1849756083)
- Can you add simple tests for invalid store type for each of the feast service? [link](https://github.com/feast-dev/feast/pull/4771#discussion_r1849839579)
- Test overriding secret keys [link](https://github.com/feast-dev/feast/pull/4771#discussion_r1850337144)
- Fix examples in the decription (wrong usage of `secretKeyName`)
- Use required fields in CRD type [link](https://github.com/feast-dev/feast/pull/4771#discussion_r1852888570), [other](https://github.com/feast-dev/feast/pull/4771#discussion_r1852889633)
- Describe CRD types [link](https://github.com/feast-dev/feast/pull/4771#discussion_r1852894919)
- Unit test incorrect types [link](https://github.com/feast-dev/feast/pull/4771#discussion_r1856850010)

@tmihalac @tchughesiv

Issue #4783, opened 2024-11-21
Add to Go operator support for db stores


tmihalac:
**Is your feature request related to a problem? Please describe.**
The current feast go operator doesn't support db stores for the various components(online,offline,registry)

**Describe the solution you'd like**
The proposed solution is to add to the operator the ability to read the DB parameters from a secret specified in the CR 

```spec:
kind: FeatureStore
metadata:
  name: example
spec:
  feastProject: my_project
  services:
    onlineStore:
      persistence:
        store:
          type: postgres
          secretRef: 
            name: _SECRET_NAME_ 
          secretKeyName: my-postgres
```

PR #4771, opened 2024-11-19
feat: Added feast Go operator db stores support

tmihalac:
# What this PR does / why we need it:
This PR adds db stores persistence configuration options to the FeatureStore CRD.
DB store para
* Support is provided to the Online service, Offline service and Registry service.
* Follows definitions in [RFC #42](https://docs.google.com/document/d/1vGKMizf3_14IyiF_W_Ik7CR03joFkQfzbKT0jH4PZJM/edit?tab=t.0#heading=h.9gaqqtox9jg6)
* Configuration examples are provided in config/samples
```apiVersion: feast.dev/v1alpha1
kind: FeatureStore
metadata:
  name: sample-s3-registry
spec:
  feastProject: my_project
  services:
    onlineStore:
      persistence:
        store:
          type: postgres
          secretRef: 
            name: _SECRET_NAME_ # Secret keys should be placed as-is under the `type` (e.g postgres) key or under the secretKeyName if specified
          secretKeyName: my-postgres # The secret key the db parameters are in (optional parameter)
    offlineStore:
      persistence:
        store:
          type: postgres
          secretRef: 
            name: _SECRET_NAME_ # Secret keys should be placed as-is under the `type` (e.g postgres) key or under the secretKeyName if specified
          secretKeyName: my-postgres # The secret key the db parameters are in (optional parameter)
    registry:
      local:
        persistence:
          store:
            type: sql
            secretRef: 
              name: _SECRET_NAME_ # Secret keys should be placed as-is under the `type` (e.g sql) key or under the secretKeyName if specified
            secretKeyName: my-postgres # The secret key the db parameters are in (optional parameter)
```

```
kind: Secret
apiVersion: v1
metadata:
  name: my-secret
  namespace: test
  uid: b410b5e5-857b-4611-acc3-1302b8c6c68c
  resourceVersion: '62030'
  creationTimestamp: '2024-11-19T20:25:15Z'
  managedFields:
    - manager: kubectl-create
      operation: Update
      apiVersion: v1
      time: '2024-11-19T20:25:15Z'
      fieldsType: FieldsV1
      fieldsV1:
        'f:data':
          .: {}
          'f:mykey': {}
        'f:type': {}
data:
  postgres: |
    host: DB_HOST
    port: DB_PORT
    database: DB_NAME
    db_schema: DB_SCHEMA
    user: DB_USERNAME
    password: DB_PASSWORD
    sslmode: verify-ca
    sslkey_path: /path/to/client-key.pem
[...]

PR #4802, opened 2024-12-02
feat: Updated feast Go operator db stores

tmihalac:
This PR contains the changes and leftovers from [4771](https://github.com/feast-dev/feast/pull/4771)

Fixes https://github.com/feast-dev/feast/issues/4793
Fixes https://github.com/feast-dev/feast/issues/4783


infra/feast-operator/config/crd/bases/feast.dev_featurestores.yaml
```diff
@@ -297,6 +297,9 @@ spec:
                               the DB store persistence for the offline store service
                             properties:
                               secretKeyName:
+                                description: |-
+                                  By default, the DB parameters should be placed as-is from the feature_store.yaml under the `type` key, if
```
dmartinol, line 301: does `should be placed as-is from the feature_store.yaml under the type key` explain that the secret values must be contained in a secret key with the name same to the value of the `type` field? as it is now, I would be tempted to use a key called `type`, instead.

tmihalac: Can you propose of an alternative phrase ?

tmihalac: How does this sound ?
`// By default, the DB parameters secret values should be placed as-is from the feature_store.yaml in a secret key with the name same to the value of the type field, if
	// SecretKeyName is specified the DB parameters secret values should be placed as-is from the feature_store.yaml under the specified key`

tchughesiv: i think its still a little muddy. 
maybe something like this to start... `By default, the selected store "type" is used as the SecretKeyName and should ...`
also, instead of `DB parameters secret values`, maybe `data store parameters` is sufficient?

those might help, but it could be simpler and more clear overall.

tmihalac: So is this enough ? @tchughesiv @dmartinol 
`By default, the selected store "type" is used as the SecretKeyName, data store parameters should be placed as-is from the feature_store.yaml`

tchughesiv: @tmihalac the `as-is` bit worries me... what if they include `type` or `registry_type`? won't that throw an error?

tmihalac: Yes it will throw an error

tchughesiv: @tmihalac is that documented anywhere? maybe this needs to be split up?

`// By default, the selected store "type" is used as the SecretKeyName` stays put over `SecretKeyName`.

Something like this might go over the `SecretRef` field? `// Data store parameters should be placed as-is from the "feature_store.yaml" under the secret key. "registry_type" & "type" fields should be removed.`

tmihalac: Sounds good to me

tchughesiv: confirmed

tmihalac: Pushed changes

tchughesiv (changes requested):
@tmihalac seems you might need to rebase on an updated master? something got messed up with your last push

tchughesiv (approved):
lgtm

franciscojavierarceo (approved):

franciscojavierarceo:
@tmihalac can you update DCO here? (You can click in the UI)

PR #4807, opened 2024-12-03
docs: Operator example notebook & api reference docs

tchughesiv:

This PR will demonstrate how to install and use Feast on Kubernetes with the Feast Go Operator. It uses the "Real-time Credit Scoring" tutorial as well.
https://github.com/feast-dev/feast-credit-score-local-tutorial

With this PR, we'll also begin generating API reference documentation.


dmartinol:
Very exhaustive showcase of the main operator deployment functions.

- I would just add a short explanation of the feast architecture (a Topology diagram?)
- maybe also clarify what the client pod does under the scenes with the `init-containers`, maybe adding the relevant snippets or short text description:
```bash
until grpcurl -insecure -d '''' -format text feast-example-registry.feast.svc.cluster.local:443 grpc.health.v1.Health/Check; do echo waiting for registry; sleep 2; done
```
```bash
#feast init
with mock.patch("click.prompt", side_effect=["postgres.feast.svc.cluster.local", "5432", "feast", "public", "feast", "feast"]):
  with mock.patch("click.confirm", side_effect=[True]):
    init_repo(project_directory, template)' > postgres-apply.py;
...
feast apply
...
CURRENT_TIME=$(date -u +"%Y-%m-%dT%H:%M:%S");
feast materialize-incremental $CURRENT_TIME
```
image

examples/operator-quickstart/02-Client.ipynb
dmartinol, line 1: can we add the few more `kubectl exec` commands to run sample commands like `feast XYZ list` to prove the repo was correctly created?

examples/operator-quickstart/01-Install.ipynb
dmartinol, line 1: It seems these notebooks have no output. While I agree that maintaining outputs in git can be challenging, having the outputs preserved is indeed very helpful when the notebooks from a browser. It provides immediate insight into the results of the experiment without needing to rerun it. WDYT?

dmartinol (commented):
lgtm!

examples/operator-quickstart/01-Install.ipynb
tchughesiv: oh ok, i hadn't thought of it that way. i'll add them

examples/operator-quickstart/01-Install.ipynb
```diff
@@ -0,0 +1,489 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "secret/feast-data-stores created\n",
+      "featurestore.feast.dev/example created\n"
+     ]
+    }
+   ],
+   "source": [
+    "!kubectl apply -f feast.yaml"
```
redhatHameed, line 233: WDYT if we can mention the namespace here 
`kubectl apply -f feast.yaml -n feast` just to make sure use using right namespace.

tchughesiv: instead, i'll add the namespace to the individual yaml objects.

tchughesiv: done, good call

redhatHameed (approved):
lgtm

PR #4812, opened 2024-12-04
feat: Add online/offline replica support

dandawg:

Adds offline/online replica scaling support to the FeastServices specs, as well as controller functionality for the feast operator to control replicas. A simple unit test was also added.

# Misc


dandawg:
I noticed that one of the operator unit tests was failing from `tls_test.go`, specifically for TLS on the offline store server. If I remember right, this was due to an issue with TLS on offline server, and not due to my changes. Still, I did have to resolve merge conflicts from the TLS commit to push this PR, so let me know if this is related to that.

tchughesiv (changes requested):
At first glance there’s quite a bit of rebase cleanup that needs to be done.

tchughesiv:
@dandawg all of the TLS unit tests should pass, just as they do in master. This is likely due to rebase leftovers.

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -455,23 +471,69 @@ func (feast *FeastServices) createNewPVC(pvcCreate *feastdevv1alpha1.PvcCreate,
 	return pvc, controllerutil.SetControllerReference(feast.Handler.FeatureStore, pvc, feast.Handler.Scheme)
 }
 
-func (feast *FeastServices) getServiceConfigs(feastType FeastServiceType) feastdevv1alpha1.ServiceConfigs {
-	appliedServices := feast.Handler.FeatureStore.Status.Applied.Services
+// <<<<<<< HEAD
```
lokeshrangineni, line 474: looks like some conflicts

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -286,11 +286,18 @@ func (feast *FeastServices) setDeployment(deploy *appsv1.Deployment, feastType F
 		return err
 	}
 	deploy.Labels = feast.getLabels(feastType)
+	// <<<<<<< HEAD
```
dmartinol, line 289: need to fix a merge issue here

infra/feast-operator/internal/controller/services/services.go
dandawg: I commented this out and forgot to remove it. I'll get it fixed.

infra/feast-operator/internal/controller/services/services.go
dandawg: Thanks for catching this. I tested with the conflict commented out, then forgot to remove it.

dandawg:
I addressed the TLS merge conflict. All operator tests passed for me locally.

infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -279,7 +284,14 @@ type DefaultConfigs struct {
...
 // OptionalConfigs k8s container settings that are optional
-type OptionalConfigs struct {
+type RegistryOptionalConfigs struct {
+	Env             *[]corev1.EnvVar             `json:"env,omitempty"`
+	ImagePullPolicy *corev1.PullPolicy           `json:"imagePullPolicy,omitempty"`
+	Resources       *corev1.ResourceRequirements `json:"resources,omitempty"`
+}
+
+type StoreOptionalConfigs struct {
+	Replicas        *int32                       `json:"replicas,omitempty"`
```
tchughesiv, line 294: since we want to set a default of 1 in the applied status, let's instead move this new field to a new `StoreDefaultConfigs` struct... that includes the existing `DefaultConfigs` inline. Then, look into leveraging the existing `ApplyDefaultsToStatus` in services package to apply said default to both stores (but only if nil in the spec).

this should be a relatively small code change. most of the other changes you've made to services functions should be reverted.

tchughesiv (changes requested):

infra/feast-operator/config/manager/kustomization.yaml
```diff
@@ -5,4 +5,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: feastdev/feast-operator
-  newTag: 0.42.0
+  newTag: 0.41.0
```
tchughesiv, line 8: likely needs to be rebased on master again given the recent 0.42.0 release.

tchughesiv (changes requested):

sdk/python/feast/driver_test_data.py
```diff
@@ -2,10 +2,10 @@
 import itertools
 from datetime import timedelta, timezone
 from enum import Enum
-from zoneinfo import ZoneInfo
```
tchughesiv, line 5: let's revert any python changes like this. these appear to be unrelated to this change.

sdk/python/tests/data/data_creator.py
```diff
@@ -1,8 +1,8 @@
 from datetime import datetime, timedelta, timezone
 from typing import Dict, List, Optional
-from zoneinfo import ZoneInfo
```
tchughesiv, line 3: same

tchughesiv (changes requested):

sdk/python/tests/data/data_creator.py
dandawg: These were auto-fixed by precommit hooks. Precommit (and therefore push) fails without them. I can override it though, and we can make this a separate PR.

As a side note, we may need to figure out how these formatting issues are making it into the master branch--they should be caught by CI (and contributors should be running precommit hooks).

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -485,6 +485,25 @@ func (feast *FeastServices) getServiceConfigs(feastType FeastServiceType) feastd
...
+		}
+	case OnlineFeastType:
+		if feast.isOnlinStore() {
+			return appliedServices.OnlineStore.Replicas
+		}
+	case RegistryFeastType:
+		if feast.isLocalRegistry() {
+			return &DefaultReplicas
+		}
+	}
```
tchughesiv, line 503: is this section necessary?
```suggestion
```

dandawg: I guess not, since it gets covered by the default return value at the end. Committed.

infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -297,6 +297,11 @@ type DefaultConfigs struct {
 	Image *string `json:"image,omitempty"`
 }
 
+type StoreServiceConfigs struct {
+	Replicas       *int32 `json:"replicas,omitempty"`
```
tchughesiv, line 301: this new field needs a description, something explaining to the user what it affects. https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#deploymentspec-v1-apps

do we want to encourage the user to use persistence if setting above 1 ?

tchughesiv (changes requested):

infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -297,7 +297,10 @@ type DefaultConfigs struct {
 	Image *string `json:"image,omitempty"`
 }
 
+// StoreServiceConfigs k8s deployment settings
 type StoreServiceConfigs struct {
+	// Replicas determines the number of pods for the feast service.
+	// When Replicas > 1, persistence should be configured to provide central storage
```
tchughesiv, line 303:
```go
// When replicas > 1, persistence is recommended.
```

tchughesiv (changes requested):

tchughesiv (approved):
lgtm

PR #4814, opened 2024-12-05
chore: Remove upper bound for `protobuf`

franco-bocci:
Removing upper bound for `protobuf` dependency.

Not created yet.


franciscojavierarceo (commented):
need to run `make lock-python-dependencies-all`

Erin-Boehmer:
@franciscojavierarceo would love to get this merged! The feast upper bound is causing some issues in our monorepo 🙏

franciscojavierarceo:
@Erin-Boehmer I think there was some concern about this from @chris

PR #4826, opened 2024-12-07
chore: Refactoring release yamls to be more module and not use redund…

franciscojavierarceo:
Refactoring release yamls to be more module and not use redundant code.

# Misc


tchughesiv:
@franciscojavierarceo we probably need a way of ensuring that `.github/workflows/publish_python_sdk.yml` completes successfully before `.github/workflows/publish_images.yml` starts. thoughts?

franciscojavierarceo:
> @franciscojavierarceo we probably need a way of ensuring that `.github/workflows/publish_python_sdk.yml` completes successfully before `.github/workflows/publish_images.yml` starts. thoughts?

100% I'm just saving my code because I wanted to switch to fixing the images. This PR is draft, so feel free to ignore for now. I'll tag you when ready. 👍

franciscojavierarceo:
@tchughesiv @lokeshrangineni  I've refactored the workflows and broken them apart into more reusable components. Can you take a look and see if I missed anything?

.github/workflows/publish_images.yml
```diff
@@ -14,81 +14,17 @@ on:
...
-          echo $HIGHEST_SEMVER_TAG
+    uses: ./.github/workflows/get_semantic_release_version.yaml
+    with:
+      custom_version: ${{ github.event.inputs.custom_version }}
+      token: ${{ github.event.inputs.token }}
 
   build-publish-docker-images:
     runs-on: ubuntu-latest
-    needs: [get-version]
+    needs: [ get-version ]
```
tchughesiv, line 24: should this be added?
```suggestion
    needs: [ get-version, publish-python-sdk ]
```

.github/workflows/publish.yml
```diff
@@ -17,81 +17,23 @@ on:
...
+      custom_version: ${{ github.event.inputs.custom_version }}
+      token: ${{ github.event.inputs.token }}
+
+  publish-python-sdk:
+    uses: ./.github/workflows/publish_python_sdk.yaml
+    with:
+      custom_version: ${{ github.event.inputs.custom_version }}
+      token: ${{ github.event.inputs.token }}
 
   build-publish-docker-images:
```
tchughesiv, line 31: is this section redundant since its also being done in the `publish-images.yml`?

.github/workflows/publish_images.yml
franciscojavierarceo: I kept this out so that if we decided to run this component independently, it wouldn't trigger an SDK update.

Notice in the `publish.yml` file it does have these required (which is the end-to-end flow).

.github/workflows/publish.yml
franciscojavierarceo: So `publish.yml` is meant to orchestrate the whole end-to-end workflow. `publish-images.yml` is just the workflow it is calling.

I separated them out so that each one can be triggered manually for testing if necessary. That'll help maintainability and debuggability.

Let me know if that doesn't make sense though, not a complete expert on github actions.

tchughesiv (approved):
lgtm

PR #4830, opened 2024-12-10
fix: Made fixes to Go Operator DB persistence

tmihalac:
* FeatureStore CR goes into an error state if the user includes the `type` or `registry_type` field in their persistence config Secret.
The fix is to throw an error only if these fields are set and do not match the CR configured persistence type for that service.
* There are some missing online/offline store types validations in the CR.
The fix is to add CR validation for missing types

### DIFF
--- a/infra/feast-operator/api/v1alpha1/featurestore_types.go
+++ b/infra/feast-operator/api/v1alpha1/featurestore_types.go
@@ -99,7 +99,7 @@ type OfflineStorePersistence struct {
 
 // OfflineStoreFilePersistence configures the file-based persistence for the offline store service
 type OfflineStoreFilePersistence struct {
-	// +kubebuilder:validation:Enum=dask;duckdb
+	// +kubebuilder:validation:Enum=file;dask;duckdb
 	Type      string     `json:"type,omitempty"`
 	PvcConfig *PvcConfig `json:"pvc,omitempty"`
 }
@@ -107,11 +107,12 @@ type OfflineStoreFilePersistence struct {
 var ValidOfflineStoreFilePersistenceTypes = []string{
 	"dask",
 	"duckdb",
+	"file",
 }
 
 // OfflineStoreDBStorePersistence configures the DB store persistence for the offline store service
 type OfflineStoreDBStorePersistence struct {
-	// +kubebuilder:validation:Enum=snowflake.offline;bigquery;redshift;spark;postgres;feast_trino.trino.TrinoOfflineStore;redis
+	// +kubebuilder:validation:Enum=snowflake.offline;bigquery;redshift;spark;postgres;trino;redis;athena;mssql
 	Type string `json:"type"`
 	// Data store parameters should be placed as-is from the "feature_store.yaml" under the secret key. "registry_type" & "type" fields should be removed.
 	SecretRef corev1.LocalObjectReference `json:"secretRef"`
@@ -125,8 +126,10 @@ var ValidOfflineStoreDBStorePersistenceTypes = []string{
 	"redshift",
 	"spark",
 	"postgres",
-	"feast_trino.trino.TrinoOfflineStore",
+	"trino",
 	"redis",
+	"athena",
+	"mssql",
 }
 
 // OnlineStore configures the deployed online store service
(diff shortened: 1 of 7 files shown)


infra/feast-operator/internal/controller/featurestore_controller_db_store_test.go
```diff
@@ -506,6 +449,60 @@ var _ = Describe("FeatureStore Controller - db storage services", func() {
...
+
+			By("Referring to a secret that contains parameter named type")
+			resource = &feastdevv1alpha1.FeatureStore{}
+			err = k8sClient.Get(ctx, typeNamespacedName, resource)
+			Expect(err).NotTo(HaveOccurred())
+
+			secret := &corev1.Secret{}
+			err = k8sClient.Get(ctx, onlineSecretNamespacedName, secret)
+			Expect(err).NotTo(HaveOccurred())
+			secret.Data[string(services.OnlineDBPersistenceCassandraConfigType)] = []byte(invalidSecretContainingTypeYamlString)
```
tchughesiv, line 461: in this scenario, does the name `invalidSecretContainingTypeYamlString` match the expectations of the test?

infra/feast-operator/internal/controller/featurestore_controller_db_store_test.go
```diff
@@ -353,39 +328,7 @@ var _ = Describe("FeatureStore Controller - db storage services", func() {
...
-			err = k8sClient.Get(ctx, typeNamespacedName, resource)
-			Expect(err).NotTo(HaveOccurred())
-
-			_, err = controllerReconciler.Reconcile(ctx, reconcile.Request{
-				NamespacedName: typeNamespacedName,
-			})
-			Expect(err).To(HaveOccurred())
-
-			Expect(err.Error()).To(Equal("secret key sql in secret registry-store-secret contains invalid tag named registry_type"))
+			Expect(err.Error()).To(Equal("secret key cassandra in secret online-store-secret contains tag named type with value wrong"))
```
tchughesiv, line 331: nit ... can this error message be modified?
```suggestion
			Expect(err.Error()).To(Equal("secret key cassandra in secret online-store-secret contains tag named type with wrong value"))
```

tchughesiv (changes requested):
looking good... just a couple nits so far

infra/feast-operator/internal/controller/featurestore_controller_db_store_test.go
tchughesiv: nm, disregard

tchughesiv (approved):
otherwise, lgtm

infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -99,19 +99,20 @@ type OfflineStorePersistence struct {
...
-	// +kubebuilder:validation:Enum=dask;duckdb
+	// +kubebuilder:validation:Enum=file;dask;duckdb
 	Type      string     `json:"type,omitempty"`
 	PvcConfig *PvcConfig `json:"pvc,omitempty"`
 }
 
 var ValidOfflineStoreFilePersistenceTypes = []string{
 	"dask",
 	"duckdb",
+	"file",
```
tchughesiv, line 110: it appears this is the same as `dask`... so maybe we don't include it.

tchughesiv: thoughts?

tmihalac: It is supported and not deprecated yet, so I think we should include it as well

infra/feast-operator/internal/controller/featurestore_controller_db_store_test.go
tmihalac: You are right, changed the name

infra/feast-operator/api/v1alpha1/featurestore_types.go
dmartinol: it was removed from the doc since v0.40
https://docs.feast.dev/v0.40-branch/reference/offline-stores/overview

infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -125,8 +126,10 @@ var ValidOfflineStoreDBStorePersistenceTypes = []string{
 	"redshift",
 	"spark",
 	"postgres",
-	"feast_trino.trino.TrinoOfflineStore",
+	"trino",
```
dmartinol, line 129: the docs are using "feast_trino.trino.TrinoOfflineStore" in the example configs. Is there an issue tracking the defect? otherwise, you can simply fix it here 😉

tchughesiv: @dmartinol it appears the docs are out of date ... here's the code where these classes/types are declared -
https://github.com/feast-dev/feast/blob/a36f7e50d97c85595cbaa14165901924efa61cbb/sdk/python/feast/repo_config.py#L87-L100

tchughesiv: ah ... agree @dmartinol

tmihalac: Opened an issue https://github.com/feast-dev/feast/issues/4837

PR #4832, opened 2024-12-10
feat: Go Operator - Parsing the output to go structs

lokeshrangineni:
Parsing the output to go lang structs rather than a Map to simplify the parsing logic.


tchughesiv (approved):
lgtm

Issue #4670, opened 2024-10-23
Make it mandatory to specify value_type for an entity


tokoko:
Currently specifying a value_type for an entity is optional. value types are instead inferred during an online call with a this [function](https://github.com/feast-dev/feast/blob/a1ff1290002376b4c9ac8ad14e60df4622bb47a4/sdk/python/feast/utils.py#L578) that basically goes through all the feature views and checks actual data types where a specific entity is used. This is obviously not very performant.

The possible reason for this design is to allow the same entity to have different types in different feature views (although now this is not allowed anyway). I think even that goal is a bit dubious, making joins across values with different data types could lead to many corner cases and bugs. I think we should instead make it mandatory to specify a value type when the user applies an entity and also make sure that data types align when actual feature views reference entities.

This will obviously be a breaking change, but I think the one that can be rectified easily. Maybe we can add DeprecationWarning in the next release saying that entities must have a type and make it mandatory in the release after that.

tokoko:
@franciscojavierarceo @HaoXuAI

franciscojavierarceo:
Yeah I definitely agree with this.

tmihalac:
@franciscojavierarceo @tokoko Can you elaborate a bit more? I want to fix this issue

PR #4833, opened 2024-12-11
chore: Issue warning announcing entity's value_type as mandatory 

franciscojavierarceo:
Make value_type for Entity issue a warning if not set. Next release we'll make it mandatory.

https://github.com/feast-dev/feast/issues/4670

# Misc

### DIFF
--- a/sdk/python/feast/entity.py
+++ b/sdk/python/feast/entity.py
@@ -11,9 +11,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import warnings
 from datetime import datetime
 from typing import Dict, List, Optional
-import warnings
 
 from google.protobuf.json_format import MessageToJson
 from typeguard import typechecked
--- a/sdk/python/feast/entity.py
+++ b/sdk/python/feast/entity.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 from datetime import datetime
 from typing import Dict, List, Optional
+import warnings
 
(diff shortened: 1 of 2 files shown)


sdk/python/feast/entity.py
```diff
@@ -79,6 +80,13 @@ def __init__(
             ValueError: Parameters are specified incorrectly.
         """
         self.name = name
+        if value_type is None:
+            warnings.warn(
+                "Entity value_type will be mandatory in the next release. "
+                "Please specify a value_type for entity '%s'." % name,
+                DeprecationWarning,
```
dmartinol, line 87: why not FutureWarning instead?
https://docs.python.org/3/library/exceptions.html#FutureWarning

franciscojavierarceo: DeprecationWarning:
>Base class for warnings about deprecated features when those warnings are intended for other Python developers.
>
>Ignored by the default warning filters, except in the __main__ module ([PEP 565](https://peps.python.org/pep-0565/)). Enabling the [Python Development Mode](https://docs.python.org/3/library/devmode.html#devmode) shows this warning.

FutureWarning
>Base class for warnings about deprecated features when those warnings are intended for end users of applications that are written in Python.

I think DeprecationWarning is a reasonable choice but I could see both ways. Devin chose the former. 🤷

dmartinol: I can’t see any deprecation actually, but if Devin will become the main maintainer, that’s fine 🤷🏻‍♂️

franciscojavierarceo: I'm going to force it next release.

dmartinol (approved):

PR #4842, opened 2024-12-11
Feat: Adding ca file support for the offline server

lokeshrangineni:
Adding TLS - ca file support for the offline server


tchughesiv (approved):
lgtm

PR #4861, opened 2024-12-19
chore: Update go targets in Makefile

xaniasd:
This PR fixes or updates some of the go-related targets, specifically addresses the following:
- incorrectly defined dependency targets
- protobuf/GRPC go tools installed globally, thus potentially messing up existing installations
- pip used instead of uv pip
- .PHONY targets not declared
- moved inline python/grpc requirements to separate file

N/A

# Misc
target `build-go` builds a binary in the root directory, it should be moved under `dist/`  (follow-up PR)

### DIFF
--- a/sdk/python/requirements/protoc-requirements.txt
+++ b/sdk/python/requirements/protoc-requirements.txt
@@ -1,0 +1,3 @@
+protobuf>=4.24.0,<5.0.0
+grpcio-tools>=1.56.2,<2
+mypy-protobuf>=3.1
--- a/sdk/python/requirements/protoc-requirements.txt
+++ b/sdk/python/requirements/protoc-requirements.txt
@@ -1,3 +1,0 @@
-protobuf>=4.24.0,<5.0.0
-grpcio-tools>=1.56.2,<2
-mypy-protobuf>=3.1
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ scratch*
 
 ### Local Environment ###
 *local*.env
+tools
 
 ### Secret ###
 **/service_account.json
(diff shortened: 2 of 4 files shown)


sdk/python/requirements/protoc-requirements.txt
shuchu, line 1: Can we avoid to make a file connection from folder "go" to other file folders? We try to make the Go folder self-contained and limit the dependencies with other directories.

franciscojavierarceo: +1

shuchu (approved):
lgtm

sdk/python/requirements/protoc-requirements.txt
xaniasd: Hi @shuchu, thanks. Based on your comment, I think there might be some room for further improvement in our setup. 

I see that we currently call `BuildGoProtosCommand` in [setup.py](https://github.com/feast-dev/feast/blob/master/setup.py#L272) to compile the proto files. This is a duplicate of the python script that compiles protos to Python (`infra/scripts/generate_protos.py`), which n.b. is kept out of `setup.py`.  Is there any reason for `BuildGoProtosCommand` to be in `setup.py`? 

Alternatively, we could install the compiler and run the commands directly from the Makefile instead. This would keep the Python and Go build processes nicely separated and remove a potential source of confusion.

The compiler is already downloaded for target `install-dependencies-proto-docs`, perhaps that could be taken along?

xaniasd: I pushed a commit to showcase how we can separate protobuf compilation for go and python.

shuchu: Thank you for your reply, @xaniasd  :)

for the "setup.py" issue, I didn't change/update it in my last PR by following the "minimum change" rule. What we see about the Go Feature server part in the "setup.py" now are sth in dust.  I would say feel free to improve them.

shuchu (approved):
lgtm

PR #4862, opened 2024-12-19
feat: Add duckdb extra to multicloud release image

tchughesiv:

duckdb is used in our feast-aws-credit-scoring-tutorial example, so it should be available here by default. this would allow us to use that same demo with the operator


lokeshrangineni (approved):

franciscojavierarceo (approved):
lgtm

franciscojavierarceo (approved):

Issue #4848, opened 2024-12-12
Operator - registry grpc healthcheck throws error when feast authorization is enabled


tchughesiv:
## Expected Behavior 
The pods should start without issue

## Current Behavior
The online and offline store pods will not start because the initContainer registry healthcheck never passes.
The registry throws this error - 
```shell
File "/opt/app-root/lib64/python3.11/site-packages/feast/permissions/server/grpc_token_extractor.py", line 34, in extract_access_token
raise AuthenticationError("Missing authorization header")
starlette.authentication.AuthenticationError: Missing authorization header
```

PR #4863, opened 2024-12-19
fix: Refactor Operator to deploy all feast services to the same Deployment/Pod

tchughesiv:

Current Operator approach to Feast deployment relies heavily on the remote provider type and feast server interaction over k8s services. This PR pulls feast services into a single Pod (but still separate containers) which allows for a more traditional Feast stack configuration. Each feast service container is able to share things like storage, feature_repo, and the feature_store.yaml.

Fixes https://github.com/feast-dev/feast/issues/4866 & https://github.com/feast-dev/feast/issues/4848


infra/feast-operator/internal/controller/services/tls.go
```diff
@@ -183,18 +190,19 @@ func (feast *FeastServices) mountTlsConfig(feastType FeastServiceType, podSpec *
...
+				Name:      volName,
+				MountPath: GetTlsPath(feastType),
+				ReadOnly:  true,
+			})
+		}
 	}
 }
 
-func mountTlsRemoteRegistryConfig(feastType FeastServiceType, podSpec *corev1.PodSpec, tls *feastdevv1alpha1.TlsRemoteRegistryConfigs) {
+func mountTlsRemoteRegistryConfig(podSpec *corev1.PodSpec, tls *feastdevv1alpha1.TlsRemoteRegistryConfigs) {
```
lokeshrangineni, line 203: just curious, `TlsRemoteRegistryConfigs` this only for registry? not for other offline and online?

tchughesiv: > just curious, TlsRemoteRegistryConfigs existing only for registry?

correct... in the event a remote registry is specified by the user with TLS configured via `services.registry.remote.tls`, the online/offline containers will need the public cert (or ca) in order to communicate with said remote registry.

> not for other offline and online?

with this function the online/offline containers are configured with the remote registry's TLS volume.

PR #4880, opened 2024-12-27
fix: Make transformation_service_endpoint configuration optional

xaniasd:
Prevents the Go feature server from panicking when `transformation_service_endpoint` configuration is not defined in feature_store.yaml. 

N/A


franciscojavierarceo (approved):
lgtm!

PR #4891, opened 2025-01-03
chore: Add Milvus to pr_integration_tests.yml

franciscojavierarceo:
Add Milvus to pr_integration_tests.yml

N/A

# Misc
Tested the Milvus container in this repo: https://github.com/franciscojavierarceo/testing-github-action-sandbox/blob/main/.github/workflows/milvus.yml

Screen shot for reference of successful build:

Screenshot 2025-01-03 at 2 11 35 PM

### DIFF
--- a/.github/workflows/pr_integration_tests.yml
+++ b/.github/workflows/pr_integration_tests.yml
@@ -45,10 +45,6 @@ jobs:
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
-      milvus:
-        image: milvus
-        ports:
-          - 19530:19530
     steps:
       - uses: actions/checkout@v4
         with:
@@ -101,7 +97,7 @@ jobs:
           docker run -d -p 6001:6379 -p 6002:6380 -p 6003:6381 -p 6004:6382 -p 6005:6383 -p 6006:6384 --name redis-cluster vishnunair/docker-redis-cluster
       - name: Setup Milvus Cluster
         run: |
-          wget https://github.com/milvus-io/milvus/releases/download/v2.4.20/milvus-standalone-docker-compose.yml -O docker-compose.yml
+          wget https://github.com/milvus-io/milvus/releases/download/v2.5.1/milvus-standalone-docker-compose.yml -O docker-compose.yml
           docker compose up -d
       - name: Test python
         if: ${{ always() }}  # this will guarantee that step won't be canceled and resources won't leak
--- a/.github/workflows/pr_integration_tests.yml
+++ b/.github/workflows/pr_integration_tests.yml
@@ -45,6 +45,10 @@ jobs:
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
+      milvus:
+        image: milvus
+        ports:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr_integration_tests.yml
```diff
@@ -95,6 +99,10 @@ jobs:
         run: |
           docker pull vishnunair/docker-redis-cluster:latest
           docker run -d -p 6001:6379 -p 6002:6380 -p 6003:6381 -p 6004:6382 -p 6005:6383 -p 6006:6384 --name redis-cluster vishnunair/docker-redis-cluster
+      - name: Setup Milvus Cluster
+        run: |
+          wget https://github.com/milvus-io/milvus/releases/download/v2.4.20/milvus-standalone-docker-compose.yml -O docker-compose.yml
+          docker compose up -d
```
lokeshrangineni, line 105: Don't we need to tear down the cluster before ending the job?

franciscojavierarceo: Nope, we don't for Redis. It'll happen automatically by the action. You can confirm this in the example I linked here: https://github.com/franciscojavierarceo/testing-github-action-sandbox/actions/runs/12602548011/job/35125851535

lokeshrangineni: not sure why do we need both one in action and services? do we need at both places? Looks like we are following the same pattern for redis as well.

.github/workflows/pr_integration_tests.yml
```diff
@@ -95,6 +99,10 @@ jobs:
         run: |
           docker pull vishnunair/docker-redis-cluster:latest
           docker run -d -p 6001:6379 -p 6002:6380 -p 6003:6381 -p 6004:6382 -p 6005:6383 -p 6006:6384 --name redis-cluster vishnunair/docker-redis-cluster
+      - name: Setup Milvus Cluster
+        run: |
+          wget https://github.com/milvus-io/milvus/releases/download/v2.4.20/milvus-standalone-docker-compose.yml -O docker-compose.yml
```
redhatHameed, line 104: WDYT about using latest release version -> https://github.com/milvus-io/milvus/releases/tag/v2.5.2

franciscojavierarceo: i tested 2.4.20 locally but i can test the latest

franciscojavierarceo: Based on the docs, the latest available image is 2.5.1

https://milvus.io/docs/install_standalone-docker-compose.md#Run-Milvus-with-Docker-Compose-Linux

I tested it and it works. I'll update.

.github/workflows/pr_integration_tests.yml
franciscojavierarceo: There are some benefits for services but for Milvus, the service isn't necessary. I just followed what I thought was the convention but it's not necessary. I'll update the PR. 👍

redhatHameed (approved):

lokeshrangineni (approved):

Issue #4904, opened 2025-01-07
Fixing the integration tests to spin up only one keycloak instance


lokeshrangineni:
## Expected Behavior 
We have to spin up the only one instance of the keycloak for all the integration tests.

## Current Behavior
Currently the keycloak instance is spinning up multiple times even though we are using the dist groups. I have enabled the logs and can see multiple keycloak instances being spinning up. This is causing memory issues and eventual failure of the integration job on the PR. We have to spin up the only one instance of the keycloak.

## Steps to reproduce
Enable the logger info by passing extra arg `-s` to pytest command in make file. You can see the logs `Starting keycloak instance` being printed multiple times.

PR #4905, opened 2025-01-07
feat: Separating the RBAC and Remote related integration tests.

lokeshrangineni:
* Separating the tests related to remote and rbac functionality.
* Added a new test marker to separate the tests related to rbac and remote functionality.
* Added a new github action to perform tests related to rbac and remote functionality.
* Filtered the rbac integration tests in the current github job. and added new make target to run the new tests.
* Added shared state so that same keycloak instance can be reused.

[4904](https://github.com/feast-dev/feast/issues/4904)

### DIFF
--- a/.github/workflows/pr_remote_rbac_integration_tests.yml
+++ b/.github/workflows/pr_remote_rbac_integration_tests.yml
@@ -1,0 +1,58 @@
+name: pr-remote-rbac-integration-tests
+# This runs the integration tests related to rbac functionality and remote registry and online features.
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - labeled
+    paths-ignore:
+      - 'community/**'
+      - 'docs/**'
+      - 'examples/**'
+
+jobs:
+  remote-rbac-integration-tests-python:
+    if:
+      ((github.event.action == 'labeled' && (github.event.label.name == 'approved' || github.event.label.name == 'lgtm' || github.event.label.name == 'ok-to-test')) ||
+      (github.event.action != 'labeled' && (contains(github.event.pull_request.labels.*.name, 'ok-to-test') || contains(github.event.pull_request.labels.*.name, 'approved') || contains(github.event.pull_request.labels.*.name, 'lgtm')))) &&
+      github.event.pull_request.base.repo.full_name == 'feast-dev/feast'
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [ "3.11" ]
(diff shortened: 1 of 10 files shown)


sdk/python/tests/integration/conftest.py
```diff
@@ -9,14 +12,30 @@
 from tests.utils.auth_permissions_util import setup_permissions_on_keycloak
 
 logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
```
redhatHameed, line 15: any reason to set log level as `INFO` as default is `warning` https://github.com/feast-dev/feast/blob/master/sdk/python/feast/cli.py#L84

.github/workflows/pr_remote_rbac_integration_tests.yml
```diff
@@ -0,0 +1,58 @@
...
+  remote-rbac-integration-tests-python:
+    if:
+      ((github.event.action == 'labeled' && (github.event.label.name == 'approved' || github.event.label.name == 'lgtm' || github.event.label.name == 'ok-to-test')) ||
+      (github.event.action != 'labeled' && (contains(github.event.pull_request.labels.*.name, 'ok-to-test') || contains(github.event.pull_request.labels.*.name, 'approved') || contains(github.event.pull_request.labels.*.name, 'lgtm')))) &&
+      github.event.pull_request.base.repo.full_name == 'feast-dev/feast'
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [ "3.11" ]
```
franciscojavierarceo, line 25: only 3.11?

franciscojavierarceo (approved):
one nit but lgtm

.github/workflows/pr_remote_rbac_integration_tests.yml
lokeshrangineni: current integration tests are only running on 3.11. so kept the same for this action also. Should we add another python version for all integration tests?

sdk/python/tests/integration/conftest.py
lokeshrangineni: This log is only for integration tests to display info logs. Keeping it info helped to display the important logs during the test runs. These logs purged after the new run per PR only. It does not harm to keep it Info but also if you want we can change it WARN as well.

redhatHameed: we can keep Feast default log level, until unless it's required to see the info message.

lokeshrangineni: I can update in the next PR as this is merged.

PR #4914, opened 2025-01-10
🍭 ADD - Route definition

ckavili:

We want to expose Feast UI to be able to access it from outside of the cluster and in OpenShift land, creating a Route object is one way to achieve that. That's why we added route template and updated values file accordingly.

# Misc


redhatHameed (approved):

ckavili:
did a force push for DCO 🫠

franciscojavierarceo (approved):

PR #4952, opened 2025-01-22
fix: Adding periodic check to fix the sporadic failures of the operator e2e tests. 

lokeshrangineni:
Adding periodic check to fix the sporadic failures of the operator e2e tests. 

Fixes - [4953](https://github.com/feast-dev/feast/issues/4953)


redhatHameed:
@lokeshrangineni WDYT if we can add retry option also ?

PR #4973, opened 2025-01-27
fix: Update manifest to add feature server image for odh

redhatHameed:

Update manifest to add feature server image for odh

# Misc


tchughesiv (changes requested):
otherwise lgtm

tchughesiv (changes requested):
lgtm

PR #4974, opened 2025-01-27
chore: Remove upper bound for `protobuf`

Erin-Boehmer:

Removing upper bound for protobuf dependency. 

This PR extends https://github.com/feast-dev/feast/pull/4814 to include updated python lockfiles.  

NA


franciscojavierarceo (approved):
Lgtm

PR #4982, opened 2025-01-29
feat: Operator adding OpenShift route for online feature server

redhatHameed:

will create  OpenShift route for online feature server

Screenshot 2025-01-29 at 1 22 56 PM

# Misc


tchughesiv (changes requested):
@redhatHameed this may need to be more thought out... what if a user doesn't want the online server to be exposed publicly? there are security concerns to consider.

redhatHameed:
> @redhatHameed this may need to be more thought out... what if a user doesn't want the online server to be exposed publicly? there are security concerns to consider.

in that case I am fine to close this PR.

Issue #4981, opened 2025-01-29
UI - Error Loading Project Configs


ngoduykhanh:
I've searched for similar issues opened previously here, but didn't see any useful information so I'd like to open a new one.

## Expected Behavior 

The UI loads successfully

## Current Behavior

The UI is accessible but can't load anything

<img width="1299" alt="Image" src="https://github.com/user-attachments/assets/760281ab-aca0-4c46-8593-fe7812dc0b94" />

## Steps to reproduce

I have tried both local setup using the Quickstart guide and deploying to Kubernetes. They have the same issue.

I enabled the DEBUG log of Feast UI container and saw no error.

If I log in into the feast-ui container, I can run feast cli successfully without issue.

```
(app-root) bash-5.1$ feast feature-views list
/opt/app-root/lib64/python3.11/site-packages/feast/feature_view.py:48: DeprecationWarning: Entity value_type will be mandatory in the next release. Please specify a value_type for entity '__dummy'.
  DUMMY_ENTITY = Entity(
Received base64 encoded feature_store.yaml
NAME                         ENTITIES    TYPE
driver_hourly_stats_fresh    {'driver'}  FeatureView
driver_hourly_stats          {'driver'}  FeatureView
transformed_conv_rate_fresh  {'driver'}  OnDemandFeatureView
transformed_conv_rate        {'driver'}  OnDemandFeatureView
```

### Specifications

- Version: 0.43
- Platform: MacOS/Kubernetes
- Subsystem:

itsmano1993:
I am facing the same issue

redhatHameed:
limited knowledge on UI part, but I believe [this](https://github.com/feast-dev/feast/pull/4772/files) PR should fix that. However, it's not part of the `0.43.0` current version.

tchughesiv:
@redhatHameed it doesn't appear to resolve the issue... i just ran the latest code w/ -
```bash
make -C infra/feast-operator install deploy IMG=quay.io/feastdev-ci/feast-operator:develop FS_IMG=quay.io/feastdev-ci/feature-server:develop
kubectl apply -f infra/feast-operator/config/samples/v1alpha1_featurestore_all_services_default.yaml
```

redhatHameed:
@tchughesiv thanks for verifying, can you make sure that while loading the page you  see the [PR](https://github.com/feast-dev/feast/pull/4772/files) code changes.

danbaron63:
[...]

PR #4995, opened 2025-01-31
chore: Upgrade msw to 2.7.0 (and Node.js to 22.13.1) in /ui

peruukki:

The main purpose was to upgrade msw from 0.36.8 to 2.7.0, to get rid of the security alerts of the old version. The upgrade turned out to be quite difficult, and in the end I also had to upgrade Node.js from 20 to 22, see more below.

No specific issue for this, but the idea for the upgrade originated from #4451.

# Misc

I only got the tests to pass with this combination of other dependencies:
- test running environment: [@happy-dom/jest-environment](https://github.com/capricorn86/happy-dom)
    - jsdom (the default) or [jest-fixed-jsdom](https://github.com/mswjs/jest-fixed-jsdom) (suggested in many places around the internet) neither worked due to various different errors.
- Node.js version: 22
    - I don't have an explanation, but with Node 20 the tests often fail to navigate to other pages, without visible errors, though sometimes the tests do pass. I saw the same failing behavior in other running environments than @happy-dom/jest-environment too, but in those there were also some visible errors. With Node 22, the tests seem to run notably faster, which might be related.
    - Version 22 has been in long-term support for quite a while now, so it's good to move to it anyway.


franciscojavierarceo:
Hey @peruukki have you had issues loading project configs? We had someone report issues with the UI since the 0.40 release: https://github.com/feast-dev/feast/issues/4981

peruukki:
> Hey @peruukki have you had issues loading project configs? We had someone report issues with the UI since the 0.40 release: #4981

I have only used the sample projects in the repository, and actually the Empty Registry there fails to load. But as far as I remember, it has always been like that for me, and I assumed it was expected since the project is indeed totally empty. 😄

empty-project-fails-to-load

tchughesiv:
@franciscojavierarceo @peruukki that issue has been resolved for dev builds and will be fully resolved with the next release.

redhatHameed (approved):

redhatHameed:
@franciscojavierarceo we merge this if no concern

PR #5001, opened 2025-02-03
chore: Operator API struct consolidation & organization

tchughesiv:

Removes repeat fields & references, renames a few structs where warranted. None of these changes affect the CRD.


tmihalac:
LGTM

lokeshrangineni:
Does this changes needs some modification to operator example?

tchughesiv:
@lokeshrangineni no, because the CRD isn't changing

PR #5003, opened 2025-02-03
Creating/updating the stable branch after the release.

lokeshrangineni:
Creating/updating the stable branch after the release.


.github/workflows/release.yml
```diff
@@ -179,3 +179,11 @@ jobs:
         if: github.event.inputs.dry_run == 'false'
         run: |
           npx -p @semantic-release/changelog -p @semantic-release/git -p @semantic-release/exec -p semantic-release semantic-release
+
+      - name: Create or update stable branch after release.
+        if: github.event.inputs.dry_run == 'false'
+        run: |
+          git fetch origin
+          git checkout -B stable
```
redhatHameed, line 187: why `-B` here ?

redhatHameed: we have already created the `stable` branch, but it's working no issue we can keep it.

redhatHameed: Got it will checkout locally

.github/workflows/release.yml
```diff
@@ -156,7 +156,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
         with:
-          persist-credentials: false
+          persist-credentials: true
```
franciscojavierarceo, line 159: is this needed?

franciscojavierarceo: oh i see to execute the checkout of the stable branch

.github/workflows/release.yml
```diff
@@ -179,3 +179,11 @@ jobs:
         if: github.event.inputs.dry_run == 'false'
         run: |
           npx -p @semantic-release/changelog -p @semantic-release/git -p @semantic-release/exec -p semantic-release semantic-release
+
+      - name: Create or update stable branch after release.
```
franciscojavierarceo, line 183:
```suggestion
      - name: Updating `stable` branch after release.
```

.github/workflows/release.yml
franciscojavierarceo, line 187:
```suggestion
          # note that this checkout creates a branch called `stable` if it does not exist
          git checkout -B stable
```

.github/workflows/release.yml
lokeshrangineni: because we are doing git push in later steps we need this credentials to be carried to next step in the current job.

franciscojavierarceo: yeah that makes sense, i realized after.

franciscojavierarceo (approved):

PR #5004, opened 2025-02-04
chore: Set router basename to simplify paths in /ui

peruukki:

Sets the React Router [basename](https://reactrouter.com/6.28.2/router-components/browser-router#basename) to avoid repeating the `PUBLIC_URL` environment variable.

This is pretty much a re-do of https://github.com/feast-dev/feast/pull/3541 but with only internal changes. Related short discussion started in https://github.com/feast-dev/feast/pull/3514#issuecomment-1472661945.

# Misc

I verified that the environment variable still works by starting the app with `PUBLIC_URL=/feast-ui yarn start` (also `PUBLIC_URL=http://localhost:3000/feast-ui yarn start` somehow works):
feast-ui-prefix

**Note:** I originally included some changes that affected the behavior (see discussion below), but eventually decided to keep this as purely refactoring, and make other kinds of improvements separately.

### DIFF
--- a/ui/src/FeastUI.tsx
+++ b/ui/src/FeastUI.tsx
@@ -15,10 +15,14 @@ const defaultQueryClient = new QueryClient();
 
 const FeastUI = ({ reactQueryClient, feastUIConfigs }: FeastUIProps) => {
   const queryClient = reactQueryClient || defaultQueryClient;
+  const basename = process.env.PUBLIC_URL ?? '';
 
   return (
     // Disable v7_relativeSplatPath: custom tab routes don't currently work with it
-    <BrowserRouter future={{ v7_relativeSplatPath: false, v7_startTransition: true }}>
+    <BrowserRouter
+      basename={basename}
+      future={{ v7_relativeSplatPath: false, v7_startTransition: true }}
+    >
       <QueryClientProvider client={queryClient}>
         <QueryParamProvider adapter={ReactRouter6Adapter}>
           <FeastUISansProviders feastUIConfigs={feastUIConfigs} />
--- a/ui/src/FeastUISansProviders.tsx
+++ b/ui/src/FeastUISansProviders.tsx
@@ -62,8 +62,6 @@ const FeastUISansProviders = ({
           isCustom: true,
         }
       : { projectsListPromise: defaultProjectListPromise(), isCustom: false };
-    
-  const BASE_URL = process.env.PUBLIC_URL || ""
 
   return (
     <EuiProvider colorMode="light">
@@ -76,9 +74,9 @@ const FeastUISansProviders = ({
           >
             <ProjectListContext.Provider value={projectListContext}>
               <Routes>
-                <Route path={BASE_URL + "/"} element={<Layout />}>
+                <Route path="/" element={<Layout />}>
                   <Route index element={<RootProjectSelectionPage />} />
-                  <Route path={BASE_URL + "/p/:projectName/*"} element={<NoProjectGuard />}>
+                  <Route path="/p/:projectName/*" element={<NoProjectGuard />}>
                     <Route index element={<ProjectOverviewPage />} />
                     <Route path="data-source/" element={<DatasourceIndex />} />
                     <Route
--- a/ui/src/components/FeaturesInServiceDisplay.tsx
+++ b/ui/src/components/FeaturesInServiceDisplay.tsx
@@ -28,9 +28,7 @@ const FeaturesInServiceList = ({ featureViews }: FeatureViewsListInterace) => {
       field: "featureViewName",
       render: (name: string) => {
         return (
-          <EuiCustomLink
-            to={`${process.env.PUBLIC_URL || ""}/p/${projectName}/feature-view/${name}`}
-          >
+          <EuiCustomLink to={`/p/${projectName}/feature-view/${name}`}>
             {name}
           </EuiCustomLink>
         );
--- a/ui/src/components/FeaturesListDisplay.tsx
+++ b/ui/src/components/FeaturesListDisplay.tsx
@@ -21,7 +21,7 @@ const FeaturesList = ({
       field: "name",
       render: (item: string) => (
         <EuiCustomLink
-          to={`${process.env.PUBLIC_URL || ""}/p/${projectName}/feature-view/${featureViewName}/feature/${item}`}
+          to={`/p/${projectName}/feature-view/${featureViewName}/feature/${item}`}
         >
           {item}
         </EuiCustomLink>
--- a/ui/src/components/ObjectsCountStats.tsx
+++ b/ui/src/components/ObjectsCountStats.tsx
@@ -55,7 +55,7 @@ const ObjectsCountStats = () => {
             <EuiFlexItem>
               <EuiStat
                 style={statStyle}
-                onClick={() => navigate(`${process.env.PUBLIC_URL || ""}/p/${projectName}/feature-service`)}
+                onClick={() => navigate(`/p/${projectName}/feature-service`)}
                 description="Feature Services→"
                 title={data.featureServices}
                 reverse
@@ -65,7 +65,7 @@ const ObjectsCountStats = () => {
               <EuiStat
                 style={statStyle}
                 description="Feature Views→"
-                onClick={() => navigate(`${process.env.PUBLIC_URL || ""}/p/${projectName}/feature-view`)}
+                onClick={() => navigate(`/p/${projectName}/feature-view`)}
                 title={data.featureViews}
                 reverse
               />
@@ -74,7 +74,7 @@ const ObjectsCountStats = () => {
               <EuiStat
                 style={statStyle}
                 description="Entities→"
-                onClick={() => navigate(`${process.env.PUBLIC_URL || ""}/p/${projectName}/entity`)}
+                onClick={() => navigate(`/p/${projectName}/entity`)}
                 title={data.entities}
                 reverse
               />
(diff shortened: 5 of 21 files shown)


HaoXuAI:
This is a really nice feature, thx!

HaoXuAI:
yeah might be better to write something here: https://docs.feast.dev/reference/alpha-web-ui

peruukki:
I added a subsection about the new prop, [see it rendered](https://github.com/peruukki/feast/blob/ui-router-basename/docs/reference/alpha-web-ui.md#custom-basename). (I also did tiny other tweaks.)

peruukki:
It now occurred to me that this is actually a breaking change if you were using Feast UI as a module in your app and using the `PUBLIC_URL` environment variable: now you need to explicitly pass its value in the `basename` prop like we do in [`ui/src/index.tsx`](https://github.com/feast-dev/feast/pull/5004/files#diff-e3dadf18c7049ea1787b2c60c668ec96d22256ed05de81fc181fbc23bba8427e) since the `FeastUI` component no longer references the environment variable.

So maybe that should be mentioned in the release notes? Let me know if you want me to change e.g. the PR title prefix from "chore" to something else due to this.

redhatHameed (approved):

franciscojavierarceo:
> > erences the environment variable.
> > So maybe that should be mentioned in the release notes? Let me know if you wan
> 
> Is there a way to make `PUBLIC_URL` still works in environment variables for feast?

+1 @peruukki any chance we can investigate this?

peruukki:
To clarify, this doesn't remove the possibility to use `PUBLIC_URL`. But if you are using Feast UI [as a module](https://github.com/feast-dev/feast/blob/master/docs/reference/alpha-web-ui.md#importing-as-a-module-to-integrate-with-an-existing-react-app), you now need to pass that value to the `FeastUI` component via its `basename` prop.

If you were only defining the `PUBLIC_URL` environment variable to use it inside the `FeastUI` component, you can now remove it. If you want to keep using it, you'll need to make a change like this in your `index.tsx` file (similarly to the [`ui/src/index.tsx` changes](https://github.com/feast-dev/feast/pull/5004/files#diff-e3dadf18c7049ea1787b2c60c668ec96d22256ed05de81fc181fbc23bba8427e) in this PR):

```diff
@@ -3,16 +3,19 @@ import ReactDOM from 'react-dom/client';
 
 import './index.css';
 import FeastUI from '@feast-dev/feast-ui';
 import '@feast-dev/feast-ui/dist/feast-ui.css';
 
+const basename = process.env.PUBLIC_URL ?? '';
+
 const root = ReactDOM.createRoot(document.getElementById('root')!);
 root.render(
   <StrictMode>
     <FeastUI
+      basename={basename}
       feastUIConfigs={{
-        projectListPromise: fetch(`${process.env.PUBLIC_URL ?? ''}/projects-list.json`, {
+        projectListPromise: fetch(`${basename}/projects-list.json`, {
           headers: {
             'Content-Type': 'application/json',
           },
         }).then((res) => {
           return res.json();
```

(And now looking at that, we could use `basename` in [defaultProjectListPromise](https://github.com/feast-dev/feast/blob/f4afcd2e9bb831e412dd31327b8596961422be5a/ui/src/FeastUISansProviders.tsx#L44) too, so you wouldn't always need to provide a custom `projectListPromise` just for this, but maybe that's for another pull request. 😄)

I would prefer not to use environment variables inside the `FeastUI` component because they are quite a hidden way to pass values and give you less flexibility on how you define the value, and I don't see why we couldn't use a regular prop here. But I also understand that we may want to preserve backwards compatibility. In that case, we can either:

1. Not add the `basename` prop to `FeastUI`, and we keep referencing `PUBLIC_URL` inside it. We can still keep the benefits of referencing it only in one place and in a way that doesn't cause an error if you haven't defined a `process.env` global; these were my main motivation for these changes anyway.
2. Have two ways of setting the `basename`, either via this new prop or via the `PUBLIC_URL` environment variable like before. And if both are defined, I would give the prop precedence since it's more local.

HaoXuAI:
> To clarify, this doesn't remove the possibility to use `PUBLIC_URL`. But if you are using Feast UI [as a module](https://github.com/feast-dev/feast/blob/master/docs/reference/alpha-web-ui.md#importing-as-a-module-to-integrate-with-an-existing-react-app), you now need to pass that value to the `FeastUI` component via its `basename` prop.
> 
> If you were only defining the `PUBLIC_URL` environment variable to use it inside the `FeastUI` component, you can now remove it. If you want to keep using it, you'll need to make a change like this in your `index.tsx` file (similarly to the [`ui/src/index.tsx` changes](https://github.com/feast-dev/feast/pull/5004/files#diff-e3dadf18c7049ea1787b2c60c668ec96d22256ed05de81fc181fbc23bba8427e) in this PR):
> 
> ```diff
> @@ -3,16 +3,19 @@ import ReactDOM from 'react-dom/client';
>  
>  import './index.css';
>  import FeastUI from '@feast-dev/feast-ui';
>  import '@feast-dev/feast-ui/dist/feast-ui.css';
>  
> +const basename = process.env.PUBLIC_URL ?? '';
> +
>  const root = ReactDOM.createRoot(document.getElementById('root')!);
>  root.render(
>    <StrictMode>
>      <FeastUI
> +      basename={basename}
>        feastUIConfigs={{
> -        projectListPromise: fetch(`${process.env.PUBLIC_URL ?? ''}/projects-list.json`, {
> +        projectListPromise: fetch(`${basename}/projects-list.json`, {
>            headers: {
>              'Content-Type': 'application/json',
>            },
>          }).then((res) => {
>            return res.json();
> ```
> 
> (And now looking at that, we could use `basename` in [defaultProjectListPromise](https://github.com/feast-dev/feast/blob/f4afcd2e9bb831e412dd31327b8596961422be5a/ui/src/FeastUISansProviders.tsx#L44) too, so you wouldn't always need to provide a custom `projectListPromise` just for this, but maybe that's for another pull request. 😄)
> 
> I would prefer not to use environment variables inside the `FeastUI` component because they are quite a hidden way to pass values and give you less flexibility on how you define the value, and I don't see why we couldn't use a regular prop here. But I also understand that we may want to preserve backwards compatibility. In that case, we can either:
> 
> 1. Not add the `basename` prop to `FeastUI`, and we keep referencing `PUBLIC_URL` inside it. We can still keep the benefits of referencing it only in one place and in a way that doesn't cause an error if you haven't defined a `process.env` global; these were my main motivation for these changes anyway.
> 2. Have two ways of setting the `basename`, either via this new prop or via the `PUBLIC_URL` environment variable like before. And if both are defined, I would give the prop precedence since it's more local.

ok I think that make sense. I like this feature because when I setup Superset it is very hard to have a ui prefix in the public URL. it would be convenient to have `process.env.PUBLIC_URL` works so that users can customize the web url.

HaoXuAI:
> > To clarify, this doesn't remove the possibility to use `PUBLIC_URL`. But if you are using Feast UI [as a module](https://github.com/feast-dev/feast/blob/master/docs/reference/alpha-web-ui.md#importing-as-a-module-to-integrate-with-an-existing-react-app), you now need to pass that value to the `FeastUI` component via its `basename` prop.
> > If you were only defining the `PUBLIC_URL` environment variable to use it inside the `FeastUI` component, you can now remove it. If you want to keep using it, you'll need to make a change like this in your `index.tsx` file (similarly to the [`ui/src/index.tsx` changes](https://github.com/feast-dev/feast/pull/5004/files#diff-e3dadf18c7049ea1787b2c60c668ec96d22256ed05de81fc181fbc23bba8427e) in this PR):
> > ```diff
> > @@ -3,16 +3,19 @@ import ReactDOM from 'react-dom/client';
> >  
> >  import './index.css';
> >  import FeastUI from '@feast-dev/feast-ui';
> >  import '@feast-dev/feast-ui/dist/feast-ui.css';
> >  
> > +const basename = process.env.PUBLIC_URL ?? '';
> > +
> >  const root = ReactDOM.createRoot(document.getElementById('root')!);
> >  root.render(
> >    <StrictMode>
> >      <FeastUI
> > +      basename={basename}
> >        feastUIConfigs={{
> > -        projectListPromise: fetch(`${process.env.PUBLIC_URL ?? ''}/projects-list.json`, {
> > +        projectListPromise: fetch(`${basename}/projects-list.json`, {
> >            headers: {
> >              'Content-Type': 'application/json',
> >            },
> >          }).then((res) => {
> >            return res.json();
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > (And now looking at that, we could use `basename` in [defaultProjectListPromise](https://github.com/feast-dev/feast/blob/f4afcd2e9bb831e412dd31327b8596961422be5a/ui/src/FeastUISansProviders.tsx#L44) too, so you wouldn't always need to provide a custom `projectListPromise` just for this, but maybe that's for another pull request. 😄)
> > I would prefer not to use environment variables inside the `FeastUI` component because they are quite a hidden way to pass values and give you less flexibility on how you define the value, and I don't see why we couldn't use a regular prop here. But I also understand that we may want to preserve backwards compatibility. In that case, we can either:
> > 
> > 1. Not add the `basename` prop to `FeastUI`, and we keep referencing `PUBLIC_URL` inside it. We can still keep the benefits of referencing it only in one place and in a way that doesn't cause an error if you haven't defined a `process.env` global; these were my main motivation for these changes anyway.
> > 2. Have two ways of setting the `basename`, either via this new prop or via the `PUBLIC_URL` environment variable like before. And if both are defined, I would give the prop precedence since it's more local.
> 
> ok I think that make sense. I like this feature because when I setup Superset it is very hard to have a ui prefix in the public URL. it would be convenient to have `process.env.PUBLIC_URL` works so that users can customize the web url.

@franciscojavierarceo seems good to go.

peruukki:
@franciscojavierarceo I reduced the scope of this PR and made it purely refactoring, i.e. no externally visible changes. It's better to make such changes separately after this.

franciscojavierarceo (approved):

PR #5020, opened 2025-02-05
fix: Changed feast operator to set status of featurestore cr to ready based on deployment.status = available

tmihalac:
Changed feast Operator to update FeatureStore CR's status.conditions by checking resulting Deployment's status

Now the status of the feast CR is more accurate

### DIFF
--- a/infra/feast-operator/internal/controller/featurestore_controller.go
+++ b/infra/feast-operator/internal/controller/featurestore_controller.go
@@ -138,6 +138,30 @@ func (r *FeatureStoreReconciler) deployFeast(ctx context.Context, cr *feastdevv1
 			Reason:  feastdevv1alpha1.FailedReason,
 			Message: "Error: " + err.Error(),
 		}
+	} else {
+		deployment, deploymentErr := feast.GetDeployment()
+		if deploymentErr != nil {
+			condition = metav1.Condition{
+				Type:    feastdevv1alpha1.ReadyType,
+				Status:  metav1.ConditionUnknown,
+				Reason:  feastdevv1alpha1.DeploymentNotAvailableReason,
+				Message: feastdevv1alpha1.DeploymentNotAvailableMessage,
+			}
+
+			result = errResult
+		} else {
+			isDeployAvailable := services.IsDeploymentAvailable(deployment.Status.Conditions)
+			if !isDeployAvailable {
+				condition = metav1.Condition{
+					Type:    feastdevv1alpha1.ReadyType,
+					Status:  metav1.ConditionUnknown,
+					Reason:  feastdevv1alpha1.DeploymentNotAvailableReason,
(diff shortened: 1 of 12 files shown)


infra/feast-operator/internal/controller/featurestore_controller.go
```diff
@@ -138,6 +138,30 @@ func (r *FeatureStoreReconciler) deployFeast(ctx context.Context, cr *feastdevv1
...
 			Message: "Error: " + err.Error(),
 		}
+	} else {
+		deployment, deploymentErr := feast.GetDeployment()
+		if deploymentErr != nil {
+			condition = metav1.Condition{
+				Type:    feastdevv1alpha1.ReadyType,
+				Status:  metav1.ConditionUnknown,
+				Reason:  feastdevv1alpha1.DeploymentNotAvailableReason,
+				Message: feastdevv1alpha1.DeploymentNotAvailableMessage,
```
tchughesiv, line 148: on `deploymentErr` should we maybe display the error to the user in the `Message`? similar to what's being done above?

tchughesiv: hmm... maybe not, disregard

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -917,3 +924,13 @@ func getProbeHandler(feastType FeastServiceType, tls *feastdevv1alpha1.TlsConfig
...
 }
+
+func IsDeploymentAvailable(conditions []appsv1.DeploymentCondition) bool {
+	for _, condition := range conditions {
+		if condition.Type == appsv1.DeploymentAvailable {
+			return condition.Status == corev1.ConditionTrue
+		}
+	}
+
+	return false
```
tchughesiv, line 935: does something like this do the same?
```suggestion
        return apimeta.IsStatusConditionTrue(conditions, appsv1.DeploymentAvailable)
```

tchughesiv: if my suggestion does the job, we may not need the new `IsDeploymentAvailable`... this check could be done directly.

tmihalac: You can't do that since `deployment.Status.Conditions` is of type `DeploymentCondition` and not `metav1.Condition` thats the reason I didn't do it in the first place

tmihalac: > if my suggestion does the job, we may not need the new `IsDeploymentAvailable` method... this check could be done directly.

You can't do that since deployment.Status.Conditions is of type DeploymentCondition and not metav1.Condition thats the reason I didn't do it in the first place

tchughesiv: ok, makes sense... thanks

franciscojavierarceo (approved):

Issue #5023, opened 2025-02-06
Operator should only create online/offline remote server containers when necessary


tchughesiv:
**Is your feature request related to a problem? Please describe.**
The operator 
**Describe the solution you'd like**
The operator currently creates separate containers for online and offline remote servers whether the user wants them or not. This should be optional and an explicit request in the CR.

PR #5024, opened 2025-02-06
feat: Operator - make server container creation explicit in the CR

tchughesiv:

Currently, the operator creates a running server/container for each feast service by default. This is unnecessary and should only be done on an as-needed basis. With this change, the user will have to explicitly configure a new `server` subsection for a given service type (online/offline/etc). Otherwise, the operator will not run a separate container w/ remote server for that service. By default, the registry container will continue created.

Fixes https://github.com/feast-dev/feast/issues/5023

# Misc

In the following example the online store, but NOT the offline, would have a separate container running a remote online feature server. The offline persistence configs, however, would still be reflected in the shared `feature_store.yaml`.
```yaml
kind: FeatureStore
metadata:
  name: example
spec:
  feastProject: my_project
  services:
    offlineStore:
      persistence:
        file:
          type: duckdb
    onlineStore:
      server: {}
```
an example of some of the optional fields within this `server` object -
```yaml
      server:
        image: 'feastdev/feature-server:0.45.0'
        logLevel: debug
        tls:
          secretKeyNames:
            tlsCrt: tls.crt
            tlsKey: tls.key
          secretRef:
            name: feast-sample-all-servers-online-tls
        envFrom:
          - secretRef:
              name: postgres-secret
        resources:
          requests:
            cpu: 150m
            memory: 128Mi
```


redhatHameed:
@tchughesiv WDYT think if we can just flag `remote` flag it will be be `false` for offline server, for online always true 

example
```
kind: FeatureStore
metadata:
  name: sample-all-services
spec:
  feastProject: my_project
  services:
    onlineStore:
      remote: true
    offlineStore:
      remote: false
    registry:
      remote:?
```

infra/feast-operator/internal/controller/featurestore_controller_loglevel_test.go
```diff
@@ -246,3 +242,7 @@ var _ = Describe("FeatureStore Controller - Feast service LogLevel", func() {
 
 	})
 })
+
```
tmihalac, line 245: Are you sure a one line function like this is needed?

infra/feast-operator/internal/controller/featurestore_controller_loglevel_test.go
```diff
@@ -246,3 +242,7 @@ var _ = Describe("FeatureStore Controller - Feast service LogLevel", func() {
...
+func strPtr(str string) *string {
```
tmihalac, line 246: Are you sure a one line function like this is needed?

tchughesiv: unfortunately yes... the alternative would be separate vars for each.

redhatHameed:
> @tchughesiv WDYT if we can just flag `remote` it will be be `false` for offline server, for online always true
> 
> example
> 
> ```
> apiVersion: feast.dev/v1alpha1
> kind: FeatureStore
> metadata:
>   name: sample-all-services
> spec:
>   feastProject: my_project
>   services:
>     onlineStore:
>       remote: true
>     offlineStore:
>       remote: false
>     registry:
>       remote:?
> ```
> 
> if this is just for offline server setting we just go with this settings
> 
> ```
> apiVersion: feast.dev/v1alpha1
> kind: FeatureStore
> metadata:
>   name: sample-all-services
> spec:
>   feastProject: my_project
>   services:
>     onlineStore {}
>     offlineStore:
>       remote: false
>     registry:
>       local: {}
> ```

Thanks, @tchughesiv for explaining on this on meet/call, I am good to keep current approach using `sever` section

redhatHameed (approved):
lgtm

PR #5025, opened 2025-02-06
feat: Add Couchbase Columnar as an Offline Store

ejscribner:
- Adds the ability to use [Couchbase Capella Columnar](https://www.couchbase.com/products/analytics/) as an Offline Store

N/A

# Misc


setup.py
```diff
@@ -147,7 +147,10 @@
 
 SINGLESTORE_REQUIRED = ["singlestoredb<1.8.0"]
 
-COUCHBASE_REQUIRED = ["couchbase==4.3.2"]
+COUCHBASE_REQUIRED = [
+    "couchbase==4.3.2",
+    "couchbase-columnar==1.0.0"
```
ntkathole, line 152: Add this in https://github.com/feast-dev/feast/blob/master/pyproject.toml#L58 as well

ejscribner: Thanks, added: [645467b](https://github.com/feast-dev/feast/pull/5025/commits/645467bdae99a0aa14b8a9077ca72f80abfc7680)

franciscojavierarceo (approved):
this is great @ejscribner!! 🚀

Issue #3270, opened 2022-10-04
Fetching online features with the wrong entities is not validated


chhabrakadabra:
## Expected Behavior 

When using `feature_store.get_online_features`, if the wrong entity (one that exists but has nothing to do with the features being requested) is used, it leads to an unexpected crash instead of a validation failure. I would expect this edge case to be caught be some validation logic ahead of time and for the error message that pops up to explain the problem.

## Current Behavior

Currently it leads to this error message that does not explain to the user what the actual problem is:

```
Traceback (most recent call last):
  File "bug_repro.py", line 5, in <module>
    feature_vector = store.get_online_features(
  File "/Users/abhin/src/github.com/chhabrakadabra/feast/.venv38/lib/python3.8/site-packages/feast/usage.py", line 294, in wrapper
    raise exc.with_traceback(traceback)
  File "/Users/abhin/src/github.com/chhabrakadabra/feast/.venv38/lib/python3.8/site-packages/feast/usage.py", line 283, in wrapper
    return func(*args, **kwargs)
  File "/Users/abhin/src/github.com/chhabrakadabra/feast/.venv38/lib/python3.8/site-packages/feast/feature_store.py", line 1588, in get_online_features
    return self._get_online_features(
  File "/Users/abhin/src/github.com/chhabrakadabra/feast/.venv38/lib/python3.8/site-packages/feast/feature_store.py", line 1775, in _get_online_features
    table_entity_values, idxs = self._get_unique_entities(
  File "/Users/abhin/src/github.com/chhabrakadabra/feast/.venv38/lib/python3.8/site-packages/feast/feature_store.py", line 1972, in _get_unique_entities
    unique_entities, indexes = tuple(
ValueError: not enough values to unpack (expected 2, got 0)
```

## Steps to reproduce

- `feast init my_project`
- `cd my_project/feature_repo`
- Change the `example_repo.py` file to the following contents:

```python
from datetime import timedelta

from feast import Entity, FeatureView, Field, FileSource
from feast.types import Float32, Int64

# Define an entity for the driver. You can think of an entity as a primary key used to
# fetch features.
driver = Entity(name="driver", join_keys=["driver_id"])
customer = Entity(name="customer", join_keys=["customer_id"])

# Read data from parquet files. Parquet is convenient for local development mode. For
# production, you can use your favorite DWH, such as BigQuery. See Feast documentation
# for more info.
driver_stats_source = FileSource(
    name="driver_hourly_stats_source",
[...]

PR #5031, opened 2025-02-07
fix: Validate entities when running get_online_features

franciscojavierarceo:

This PR addresses the issue of unexpected crashes when using feature_store.get_online_features with incorrect or missing join keys. 

Specifically, it introduces validation logic in the `_get_unique_entities` function to ensure that all expected join keys are present and non-empty before proceeding. 

**If any required join keys are missing** or have empty values, a KeyError is raised with a descriptive error message. Additionally, new unit tests have been added to cover various edge cases, ensuring that the function behaves as expected under different scenarios.

Changes:
- `sdk/python/feast/utils.py`:
    - Added validation for missing and empty join keys in `_get_unique_entities`.
    - Updated the logic to convert column-oriented `table_entity_values` into row-wise data and handle edge cases.
- `sdk/python/tests/unit/online_store/test_online_retrieval.py`:
    - Added a test case to verify that a `KeyError` is raised when required join keys are missing.
- `sdk/python/tests/unit/test_unit_feature_store.py`:
    - Imported pytest for testing.
    - Renamed `test_get_unique_entities` to `test_get_unique_entities_success` for clarity.
    - Added test cases for missing join keys and handling errors appropriately.

https://github.com/feast-dev/feast/issues/3270

# Misc

It's worth noting, when one key is present and one is not, the retrieval will still succeed as this is expected behavior.

### DIFF
--- a/sdk/python/feast/utils.py
+++ b/sdk/python/feast/utils.py
@@ -687,25 +687,48 @@ def _get_unique_entities(
         entity_name_to_join_key_map,
         join_key_values,
     )
+    # Validate that all expected join keys exist and have non-empty values.
+    expected_keys = set(entity_name_to_join_key_map.values())
+    expected_keys.discard("__dummy_id")
+    missing_keys = sorted(
+        list(set([key for key in expected_keys if key not in table_entity_values]))
+    )
+    empty_keys = sorted(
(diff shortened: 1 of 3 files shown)


HaoXuAI (approved):
LGTM!

Issue #5038, opened 2025-02-10
Postgres Online store does not always manage transactions gracefully


TomSteenbergen:
## Expected Behavior 

Whenever transactions are used with a Postgres online store, changes should be committed. If transactions are not required, we shouldn't open transactions (e.g. when simply reading features).

## Current Behavior

We are using `psycopg3` in case we use Postgres for the online store. If you create connections with `psycopg3` using a context manager (i.e. a `with ...` block), it will automatically commit the transaction once you exit the context. However, in `feast` we are manually creating (and in the case of connection pools, putting) connections, so we need to ensure that we also explicitly call `commit` after firing queries to the database. 

The `commit` statements were missing for (async) read methods of the `PostgresOnlineStore` and the `teardown` method. For the read statements, this wasn't an issue as we just do a `SELECT`. However, it does add warning statements and quite a bit of latency as transactions needed to be rolled back. 

The warning statement: 
```
2025-02-07 14:05:40,301 WARNING [psycopg.pool] rolling back returned connection: <psycopg.AsyncConnection [INTRANS] (host=<REDACTED> port=<REDACTED> user=<REDACTED> database=<REDACTED>) at 0x314498790>

```

### Specifications

- Version: 0.40.1
- Platform: MacOS
- Subsystem: Sequoia 15.2

## Possible Solution

Add missing `commit` calls and enable `autocommit` if a transaction is not needed for improved performance (e.g. for read methods).

Filed a PR here: https://github.com/feast-dev/feast/pull/5039

PR #5039, opened 2025-02-10
fix: Ensure Postgres queries are committed or autocommit is used

TomSteenbergen:

We are using `psycopg3` in case we use Postgres for the online store. If you create connections with `psycopg3` using a context manager (i.e. a `with ...` block), it will automatically commit the transaction once you exit the context. However, in `feast` we are manually creating (and in the case of connection pools, putting) connections, so we need to ensure that we also explicitly call `commit` after firing queries to the database. 

The `commit` statements were missing for (async) read methods of the `PostgresOnlineStore` and the `teardown` method. For the read statements, this wasn't an issue as we just do a `SELECT`. However, it does add warning statements and quite a bit of latency as transactions needed to be rolled back. 

Hence, for those read methods, I introduced an optional `autocommit` to the `_get_conn` and `_get_conn_async` methods that defaults to `False`. We are setting it to `True` in our read methods, so that prevents us from having to call `commit` and further optimizes the performance. In my local tests, `online_read_async` performance improved by 20-30% when calling explicitly commit, and **40-70%** when enabling autocommit (depending on feature view).

Fixes https://github.com/feast-dev/feast/issues/5038

### DIFF
--- a/sdk/python/feast/infra/online_stores/postgres_online_store/postgres.py
+++ b/sdk/python/feast/infra/online_stores/postgres_online_store/postgres.py
@@ -5,6 +5,7 @@ from datetime import datetime
 from typing import (
     Any,
     AsyncGenerator,
+    Bool,
     Callable,
     Dict,
     Generator,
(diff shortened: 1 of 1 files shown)


franciscojavierarceo (approved):
Thank you for this!

PR #5041, opened 2025-02-10
chore: Operator / Feast should have matching Data Source types

tchughesiv:

The operator needs to remain in sync with the feast data source python class types. This PR ensures that's the case long-term.

### DIFF
--- a/infra/feast-operator/api/v1alpha1/featurestore_types.go
+++ b/infra/feast-operator/api/v1alpha1/featurestore_types.go
@@ -114,8 +114,8 @@ var ValidOfflineStoreFilePersistenceTypes = []string{
 
 // OfflineStoreDBStorePersistence configures the DB store persistence for the offline store service
 type OfflineStoreDBStorePersistence struct {
-	// Type of the persistence type you want to use. Allowed values are: snowflake.offline, bigquery, redshift, spark, postgres, trino, redis, athena, mssql
-	// +kubebuilder:validation:Enum=snowflake.offline;bigquery;redshift;spark;postgres;trino;redis;athena;mssql
+	// Type of the persistence type you want to use.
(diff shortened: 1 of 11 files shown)


infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -160,7 +159,7 @@ type OnlineStoreFilePersistence struct {
 
 // OnlineStoreDBStorePersistence configures the DB store persistence for the online store service
 type OnlineStoreDBStorePersistence struct {
-	// Type of the persistence type you want to use. Allowed values are: snowflake.online, redis, ikv, datastore, dynamodb, bigtable, postgres, cassandra, mysql, hazelcast, singlestore, hbase, elasticsearch, qdrant, couchbase, milvus
+	// Type of the persistence type you want to use.
```
redhatHameed, line 162: why this comments section removed

tchughesiv: i guess i was concerned with it getting out of date as there's not way to test that its accurate.

i've added a test to ensure the next line, kubebuilder's enum validation, is accurate... and so i guess i felt like that should be sufficient as it will inform the user which data sources are supported.

that said, if we think the comment is important i can add it back.

tmihalac:
lgtm

infra/feast-operator/api/v1alpha1/featurestore_types.go
redhatHameed: I saw similar comments are there for`OfflineStoreDBStorePersistence`  wether we keep both or removed from OfflineStoreDBStorePersistence too WDYT ?

redhatHameed: Yes fine with we can remove due to out of date issue.

redhatHameed (approved):
lgtm

PR #5044, opened 2025-02-11
fix: Operator - make onlineStore the default service

tchughesiv:

Currently, the default deployed feast server container is the remote registry grpc server. With this change, the default will change to an online feature server container. The user can still run an offline and/or registry server container if they like.

The reasoning behind this change is as follows:
 - The online feature server is the more commonly used / necessary component of a FeatureStore.
 - The online feature server is a read-only service, so is more secure out of the box than a remote registry service. As such, running the grpc registry should be an explicit choice by the user.
 - The remote registry provider type and server is a still relatively new feature and more of an edge use-case


tmihalac (approved):

redhatHameed (approved):
lgtm

PR #5059, opened 2025-02-15
chore: Convert EuiCustomLink to TypeScript in /ui

peruukki:

EuiCustomLink was the last JavaScript component, convert it to TypeScript too.

No related issue.

# Misc

The `href` prop was never used, so we omit it from the props type definition, and no longer pass it to the component. (We define `href` in EuiCustomLink based on the passed `to` prop.)

We could also have chosen to drop the `to` prop and use `href` instead, but `to` is used by [React Router's Link](https://reactrouter.com/6.29.0/components/link), and it makes sense to use the same name in EuiCustomLink since we pass it to React Router's `navigate` function; the naming hints that it's tied to React Router and not just a regular URL path.

These changes partially overlap with #5004, but I'll remove such changes from the other PR if this gets merged first, and vice versa.

### DIFF
--- a/ui/src/components/EuiCustomLink.tsx
+++ b/ui/src/components/EuiCustomLink.tsx
@@ -1,23 +1,26 @@
-// File name: "EuiCustomLink.js".
 import React from "react";
-import { EuiLink } from "@elastic/eui";
-import { useNavigate, useHref } from "react-router-dom";
+import { EuiLink, type EuiLinkAnchorProps } from "@elastic/eui";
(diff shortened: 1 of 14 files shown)


peruukki:
This is still waiting for a review, would someone be available for it? 🙏

franciscojavierarceo (approved):
lgtm.

franciscojavierarceo:
@peruukki feel free to tag me directly next time! Sorry for the delay there! Thank you for your contribution!

PR #5062, opened 2025-02-17
feat: Adding support for SQLite Offline Store

franciscojavierarceo:

# Misc


sdk/python/feast/infra/offline_stores/sqlite_source.py
```diff
@@ -0,0 +1,272 @@
...
+    def get_table_column_names_and_types(
+        self, config: RepoConfig
+    ) -> Iterable[Tuple[str, str]]:
+        """Retrieves a mapping of column names to types for this SQLite data source."""
+        with sqlite3.connect(self._sqlite_options._path) as conn:
+            cursor = conn.cursor()
+            # Get column info using pragma
+            query = f"SELECT * FROM {self.get_table_query_string()} LIMIT 0"
+            cursor.execute(query)
+            columns = cursor.description
```
ntkathole, line 210: columns can be None here if there are no columns in result ?

franciscojavierarceo: I had devin create most of this PR so feel free to ignore it, I was testing 😂

PR #5065, opened 2025-02-18
feat: Adding docling to feast as add on

franciscojavierarceo:
Adding docling to feast as add on

# Misc


setup.py
```diff
@@ -158,6 +158,8 @@
 
 MILVUS_REQUIRED = ["pymilvus"]
 
+RAG_REQUIRED = ["docling"]
```
dmartinol, line 161: unversioned?

franciscojavierarceo: I was going to figure out which version was most appropriate unbounded but this uncovered the broader issue of us having to move to pyproject.toml over the setup.py so we'll have to work on that first.

Issue #4898, opened 2025-01-06
Querying SQLRegistry when `allow_cache=True` hangs when cache expires


aloysius-lim:
## Expected Behavior 

When using SQLRegistry with cache enabled, and a query is run with `allow_cache=True`, the registry should refresh the cache (if necessary) and return the results with minimal latency.

## Current Behavior

When the cache expires, any method call with `allow_cache=True` runs indefinitely, and has to be terminated with Ctrl+C. The stacktrace shows that it always gets stuck on this line:

```
File ~/mambaforge/envs/nudgerank/lib/python3.12/site-packages/feast/infra/registry/caching_registry.py:433, in CachingRegistry._refresh_cached_registry_if_necessary(self)
    431 def _refresh_cached_registry_if_necessary(self):
    432     if self.cache_mode == "sync":
--> 433         with self._refresh_lock:
    434             if self.cached_registry_proto == RegistryProto():
    435                 # Avoids the need to refresh the registry when cache is not populated yet
    436                 # Specially during the __init__ phase
    437                 # proto() will populate the cache with project metadata if no objects are registered
    438                 expired = False
```

This also happens when calling `refresh()`, which means that the registry can never be refreshed, and the cache is not usable at all. Only `allow_cache=False` works, which adds latency to every query.

## Steps to reproduce

Set up an SQLRegistry and enable caching. Then, run `refresh()` or any other method with `allow_cache=True`.

```python
import time

from feast import Entity, FeatureStore, RepoConfig

# Set up Feature Store with SQLRegistry

repo_config = RepoConfig(
    project="my_project",
    registry={
        "registry_type": "sql",
        "path": "<db_url>",
        # Set short TTL for testing.
        "cache_ttl_seconds": 5,
    },
    ...
)
feature_store = FeatureStore(config=repo_config)

# Create and register entity
driver = Entity(name="driver", join_keys=["driver_id"])
feature_store.apply(driver)

# Get entity, and use cache. This succeeds if it is run before the TTL.
feature_store.get_entity("driver", True)

# Let cache expire
time.sleep(6)

# This runs forever
feature_store.get_entity("driver", True)

# Any of these also run forever
feature_store.refresh_registry()
feature_store.get_*(..., True)
```

### Specifications

- Version: 0.40.1
- Platform: macOS 14.6.1
- Subsystem:

## Possible Solution

[...]

PR #5068, opened 2025-02-18
fix: Skip refresh if already in progress or if lock is already held

ntkathole:

This is the fix to prevent deadlocks on registry cache refresh.

- Used `self._refresh_lock.acquire(blocking=False)` to avoid infinite waiting and avoids releasing an unlocked lock.
- Fixed recursive calls by checking `self._refresh_lock.locked()` flag to prevent `self.refresh()` from calling itself indirectly.
- Handled exceptions by wrapping logic inside try to catch errors and always release locks.

#4898 

Tests:
```
# pytest sdk/python/tests/unit/infra/registry/test_registry.py -svv
sdk/python/tests/unit/infra/registry/test_registry.py::test_cache_expiry_triggers_refresh 02/20/2025 03:36:21 PM feast.infra.registry.caching_registry INFO: Registry cache expired, so refreshing
02/20/2025 03:36:21 PM feast.infra.registry.caching_registry INFO: Skipping refresh if already in progress
02/20/2025 03:36:21 PM feast.infra.registry.caching_registry INFO: Registry cache expired, so refreshing
02/20/2025 03:36:21 PM feast.infra.registry.caching_registry INFO: Skipping refresh if already in progress
PASSED
sdk/python/tests/unit/infra/registry/test_registry.py::test_skip_refresh_if_lock_held 02/20/2025 03:36:21 PM feast.infra.registry.caching_registry INFO: Skipping refresh if lock is already held by another thread
PASSED
==================================================== 2 passed in 0.01s ====================================================
```

### DIFF
--- a/sdk/python/feast/infra/registry/caching_registry.py
+++ b/sdk/python/feast/infra/registry/caching_registry.py
@@ -425,12 +425,24 @@ class CachingRegistry(BaseRegistry):
         return self._list_projects(tags)
 
     def refresh(self, project: Optional[str] = None):
-        self.cached_registry_proto = self.proto()
-        self.cached_registry_proto_created = _utc_now()
+        if self._refresh_lock.locked():
+            logger.info("Skipping refresh if already in progress")
+            return
+        try:
+            self.cached_registry_proto = self.proto()
+            self.cached_registry_proto_created = _utc_now()
+        except Exception as e:
+            logger.error(f"Error while refreshing registry: {e}", exc_info=True)
 
     def _refresh_cached_registry_if_necessary(self):
         if self.cache_mode == "sync":
-            with self._refresh_lock:
+            # Try acquiring the lock without blocking
+            if not self._refresh_lock.acquire(blocking=False):
+                logger.info(
+                    "Skipping refresh if lock is already held by another thread"
+                )
+                return
+            try:
                 if self.cached_registry_proto == RegistryProto():
                     # Avoids the need to refresh the registry when cache is not populated yet
                     # Specially during the __init__ phase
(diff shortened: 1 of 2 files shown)


franciscojavierarceo (commented):
it looks like `CachingRegistry` doesn't have tests anywhere at all. It'd be good to add some and start to catch some of this stuff. 

Could you add a test under `sdk/python/tests/units/infra/registry/test_registry.py`?

lokeshrangineni:
integration test failure seems to be legit. may be better to investigate.

sdk/python/feast/infra/registry/caching_registry.py
```diff
@@ -425,12 +426,25 @@ def list_projects(
         return self._list_projects(tags)
 
     def refresh(self, project: Optional[str] = None):
-        self.cached_registry_proto = self.proto()
-        self.cached_registry_proto_created = _utc_now()
+        if self._refreshing:  # Skipping refresh if already in progress
```
lokeshrangineni, line 429: I think it is better to log here. it will be helpful to debug or observe the pattern.

sdk/python/feast/infra/registry/caching_registry.py
```diff
@@ -37,6 +37,7 @@ def __init__(self, project: str, cache_ttl_seconds: int, cache_mode: str):
         )
         self.cached_registry_proto = self.proto()
         self.cached_registry_proto_created = _utc_now()
+        self._refreshing = False  # Prevents reentrant calls to refresh
```
lokeshrangineni, line 40: can we use `_refresh_lock` status instead of  having another property?

ntkathole:
> it looks like `CachingRegistry` doesn't have tests anywhere at all. It'd be good to add some and start to catch some of this stuff.
> 
> Could you add a test under `sdk/python/tests/units/infra/registry/test_registry.py`?

Tests added

sdk/python/feast/infra/registry/caching_registry.py
ntkathole: Done!

sdk/python/feast/infra/registry/caching_registry.py
ntkathole: Done!

ntkathole:
> it looks like `CachingRegistry` doesn't have tests anywhere at all. It'd be good to add some and start to catch some of this stuff.
> 
> Could you add a test under `sdk/python/tests/units/infra/registry/test_registry.py`?

Tests added.

franciscojavierarceo (approved):
Nice thanks for this!

PR #5072, opened 2025-02-19
test: Added upgrade tests

tmihalac:
Added feast go operator upgrade tests

Tests backward compatibility of the CR

There are 2 new test suites with new make targets:
`test-previous-version` - Deploys the previous version operator and feast server, applies the previous version CRs and runs the e2e test
`test-upgrade` -  Deploys the previous version operator and feast server, applies the current CRs and runs the e2e test

Those tests are run in the integration test github action right after the `make test-e2e` target


.github/workflows/operator-e2e-integration-tests.yml
```diff
@@ -55,6 +55,12 @@ jobs:
           # Run the e2e tests
           cd infra/feast-operator/
           make test-e2e
+          docker images --format '{{.Repository}}:{{.Tag}}' | grep 'feast' | xargs -r docker rmi -f
```
lokeshrangineni, line 58: may be you can add it as separate step for more clarity in case if there is a failure? Even if it makes sense to have its own github action?

infra/feast-operator/test/e2e/e2e_test.go
```diff
@@ -17,217 +17,38 @@ limitations under the License.
...
-			By("loading the the feast image on Kind cluster")
-			err = utils.LoadImageToKindClusterWithName(feastLocalImage)
-			ExpectWithOffset(1, err).NotTo(HaveOccurred())
+	featureStoreName := "simple-feast-setup"
+	feastResourceName := utils.FeastPrefix + featureStoreName
+	feastK8sResourceNames := []string{
+		feastResourceName + "-online",
+		feastResourceName + "-offline",
+		feastResourceName + "-ui",
+	}
```
redhatHameed, line 31: just want make sure this is correct, we have `online store/server` default now. so need registry here instead of online. WDYT ?
cc @tchughesiv

tchughesiv: or, just add `feastResourceName + "-registry",`

.github/workflows/operator-e2e-integration-tests.yml
tmihalac: > may be you can add it as separate step for more clarity in case if there is a failure? Even if it makes sense to have its own github action?

I had to do it here otherwise the tests were failing because of out of space error(Github allocated size is only 14gb and can't be changed or affected)

redhatHameed:
@tmihalac few minor questions

1. Are the upgrade tests set to run during the Feast release process, or every pull request?
2. What will happened it test fails will this block releases ?

infra/feast-operator/test/e2e/e2e_test.go
tmihalac: > just want make sure this is correct, we have `online store/server` default now. so need registry here instead of online. WDYT ? cc @tchughesiv

I used the old e2e tests without changing them and in them there is special care of registry, i didn't want to mix updating/fixing e2e tests following latests changes with doing an upgrade test, I think it should be a different task

tmihalac:
> @tmihalac few minor questions
> 
> 1. Are the upgrade tests set to run during the Feast release process, or every pull request?
> 2. What will happened if test fails will this block releases ?

I've added them to the e2e github action which runs on:
```on:
  push:
    branches:
      - main
  pull_request:
    types:
      - opened
      - synchronize
      - labeled```

File: .github/workflows/operator-e2e-integration-tests.yml (hunk 1 above)
Comment (tmihalac) [reply .github/workflows/operator-e2e-integration-tests.yml:58]: > I might be missing something, but could you clarify how the compute/memory can change if it's run in a separate step within the same action?

Maybe I misunderstood what you mean, you want me to split all the lines after `make test-e2e` into different steps, I can do that no problem there

Comment (lokeshrangineni) [reply .github/workflows/operator-e2e-integration-tests.yml:58]: Sorry for the confusion from my original message:

1) Moving it to a separate step in the same action.
2) Moving it to the its own github action. This is helpful especially for the compute/memory reasons so that you no need to share the resources with the existing tests.

#1 makes more sense to do so that easy to identify the failures and logs are also separated into its own step.
#2 you can decide based on pros and cons of each.

Comment (tmihalac) [reply .github/workflows/operator-e2e-integration-tests.yml:70]: > Sorry for the confusion from my original message:
> 
> 1. Moving it to a separate step in the same action.
> 2. Moving it to the its own github action. This is helpful especially for the compute/memory reasons so that you no need to share the resources with the existing tests.
> 
> #1 makes more sense to do so that easy to identify the failures and logs are also separated into its own step. #2 you can decide based on pros and cons of each.

@lokeshrangineni I chose option 1 and updated the PR, please have a look

File: infra/feast-operator/test/e2e/e2e_test.go (hunk 2 above)
Comment (tmihalac) [reply infra/feast-operator/test/e2e/e2e_test.go:31]: > maybe try to change `feastResourceName + "-online"` -> `feastResourceName + "-registry" ` and see where it breaks ? if this required more works then we can open a separate PR for this.

It causes the tests to fail, I think it should be in a different task

Comment (tchughesiv) [reply infra/feast-operator/test/e2e/e2e_test.go:31]: i think we need a different method maybe? one that checks for remote registry and one that checks for local registry service. when remote registry is configured, a svc won't exist for it

Comment (tmihalac) [reply infra/feast-operator/test/e2e/e2e_test.go:31]: > i think we need a different method maybe? one that checks for remote registry and one that checks for local registry service. when remote registry is configured, a svc won't exist for it

Do you think it should be fixed as part of the upgrade tests ? IMO it should be a separate task - "revise existing e2e test and enhance them"

Comment (tchughesiv) [reply infra/feast-operator/test/e2e/e2e_test.go:31]: +1

File: .github/workflows/operator-e2e-integration-tests.yml [hunk 3]
```diff
@@ -67,5 +67,45 @@ jobs:
...
+          make test-upgrade
+
+      - name: Debug KIND Cluster when there is a failure
+        if: failure()
+        run: |
+          kubectl get pods --all-namespaces
+          kubectl describe nodes
+
+      - name: Clean up
+        if: always()
```
lokeshrangineni, line 106: why do we need clean up multiple times? deleting the kind cluster in previous clean up and is the test expected to run without kind cluster or we recreating it?

tmihalac: I clean between each step because in each step an image is being created and loaded

.github/workflows/operator-e2e-integration-tests.yml
```diff
@@ -62,10 +62,44 @@ jobs:
...
+          docker images --format '{{.Repository}}:{{.Tag}}' | grep 'feast' | xargs -r docker rmi -f
+          docker system prune -a -f
+
+      - name: Run Previous version tests
+        run: |
+          # Run the previous version tests
+          cd infra/feast-operator/
+          make test-previous-version
+
+      - name: Debug KIND Cluster when there is a failure
```
lokeshrangineni, line 77: This can be removed. i see it is duplicated.

.github/workflows/operator-e2e-integration-tests.yml
```diff
@@ -62,10 +62,44 @@ jobs:
...
+        if: failure()
+        run: |
+          kubectl get pods --all-namespaces
+          kubectl describe nodes
+
+      - name: Clean up Clean up docker images
```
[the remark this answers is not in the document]
tmihalac: I don't think so, why does it need to be removed ?

.github/workflows/operator-e2e-integration-tests.yml
lokeshrangineni: I have not tested but my understanding is this DEBUG step will execute if any one of the step fails in the current job.

tmihalac: > I have not tested but my understanding is this DEBUG step will execute if any one of the step fails in the current job.

@lokeshrangineni In GitHub Actions, the if: failure() condition is used to determine whether a step should run only if any previous steps have failed in the workflow.

If multiple steps failed I want to print the cluster state after each step that failed and not only the state after the last failure, that why I don't think we should remove the cleanup steps and leave only one in the end

infra/feast-operator/test/e2e/e2e_test.go
tmihalac: Opened an new issues to track this https://github.com/feast-dev/feast/issues/5091

Issue #5081, opened 2025-02-21
Update Sqlite UPDATE and INSERT logic to use ON CONFLICT


franciscojavierarceo:
**Is your feature request related to a problem? Please describe.**
Migrate from this redundant code:

```python
conn.execute(
    f"""
        UPDATE {table_name}
        SET value = ?, event_ts = ?, created_ts = ?
        WHERE (entity_key = ? AND feature_name = ?)
    """,
    (
        # SET
        val.SerializeToString(),
        timestamp,
        created_ts,
        # WHERE
        entity_key_bin,
        feature_name,
    ),
)

conn.execute(
    f"""INSERT OR IGNORE INTO {table_name}
        (entity_key, feature_name, value, event_ts, created_ts)
        VALUES (?, ?, ?, ?, ?)""",
    (
        entity_key_bin,
        feature_name,
        val.SerializeToString(),
        timestamp,
        created_ts,
    ),
)
```

To
```python
conn.execute(
    f"""
    INSERT INTO {table_name} (entity_key, feature_name, value, event_ts, created_ts)
    VALUES (?, ?, ?, ?, ?)
    ON CONFLICT(entity_key, feature_name) DO UPDATE SET
        value = excluded.value,
        event_ts = excluded.event_ts,
        created_ts = excluded.created_ts;
    """,
    (
        entity_key_bin,          # entity_key
        feature_name,            # feature_name
        val.SerializeToString(), # value
        timestamp,               # event_ts
        created_ts               # created_ts
    ),
)
```

**Describe the solution you'd like**
Update with this code

**Describe alternatives you've considered**
N/A

**Additional context**
Should do this for sqlite-vec use cases too.

Issue #5073, opened 2025-02-19
Add SQLite Hybrid Search and BM25 for local vector search


franciscojavierarceo:
**Is your feature request related to a problem? Please describe.**
See this example: https://github.com/liamca/sqlite-hybrid-search/tree/main and the sqlite docs: https://www.sqlite.org/fts5.html

This should be complemented with the SQLite-vec implementation. 

**Describe the solution you'd like**
```python
document_embeddings = FeatureView(
    name="embedded_documents",
    entities=[item, author],
    schema=[
        Field(
            name="vector",
            dtype=Array(Float32),
            # Look how easy it is to enable RAG!
            vector_index=True,
            vector_search_metric="COSINE",
        ),
        Field(name="item_id", dtype=Int64),
        Field(name="author_id", dtype=String),
        Field(name="created_timestamp", dtype=UnixTimestamp),
        Field(name="sentence_chunks", dtype=String),
        Field(name="event_timestamp", dtype=UnixTimestamp),
    ],
    source=rag_documents_source,
    ttl=timedelta(hours=24),
)
```
Somewhere in the FeatureView we should allow the search to be declared explicitly.

Also, need to drop the `query_embedding` as a required input in:

```python
results = store.retrieve_online_documents_v2(
    features=[
        "document_embeddings:Embeddings",
        "document_embeddings:content",
        "document_embeddings:title",
    ],
    query=query_embedding,
    query_string="(content: 5) OR (title: 1) OR (title: 3)",
    top_k=3,
).to_dict()
print(results)
```

**Describe alternatives you've considered**
TBD

**Additional context**

PR #5082, opened 2025-02-22
feat: Updating retrieve online documents v2 to work for other fields for sq…

franciscojavierarceo:

This PR enables full text search for the `retrieve_online_documents/` endpoint for SQLite Vec. It also establishes a new parameter in the SDK method called `query_string` that can be passed to use key word search. There are a number of limitations with this approach as the `top_k` parameter can be misleading (as evident by the example). This offers a good start for keyword search that leverages the existing vector retrieval endpoint. As a next step, enabling hybrid search would be beneficial. 

- `feature_server.py`:
    - Added optional query_string parameter to the GetOnlineFeaturesRequest class.
Updated retrieve_online_documents to support the query_string parameter.
feature_store.py:
    - Added optional query_string parameter to retrieve_online_documents_v2.
Updated related methods to handle query_string.
feature_view.py:
    - Added an assertion to ensure only one vector feature per feature view.

- `milvus.py`:
    - Added optional query_string parameter to retrieve_online_documents_v2.

- `online_store.py`:
    - Added optional query_string parameter to retrieve_online_documents_v2.

- `sqlite.py`:
    - Extensive changes to support text search with BM25, including adding text_search_enabled configuration and handling query_string.
    - Updated SQL operations to support the new functionalities.

- `passthrough_provider.py` and `provider.py`:
    - Updated retrieve_online_documents_v2 to support the query_string.

- `types.py`:
    - Added FEAST_VECTOR_TYPES list for handling vector types.

- `example_feature_repo_1.py`:
    - Added content and title fields to an example feature view.

https://github.com/feast-dev/feast/issues/5081
https://github.com/feast-dev/feast/issues/5073

# Misc

### DIFF
--- a/sdk/python/feast/types.py
+++ b/sdk/python/feast/types.py
@@ -14,7 +14,7 @@
 from abc import ABC, abstractmethod
 from datetime import datetime, timezone
 from enum import Enum
-from typing import Dict, Union
+from typing import Dict, List, Union
 
 import pyarrow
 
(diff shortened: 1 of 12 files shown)


franciscojavierarceo:
@HaoXuAI

sdk/python/feast/types.py
```diff
@@ -196,6 +196,17 @@ def __str__(self):
     UnixTimestamp: pyarrow.timestamp("us", tz=_utc_now().tzname()),
 }
 
+FEAST_VECTOR_TYPES: List[Union[ValueType, PrimitiveFeastType, ComplexFeastType]] = [
```
HaoXuAI, line 199: wonder if this is used somewhere? :)

franciscojavierarceo: https://github.com/feast-dev/feast/pull/5082/files#r1972213724

shuchu (approved):

ejscribner (approved):

franciscojavierarceo:
Technically there's a flaw here I need to resolve because you have to pass in a query embedding for just pure text search, which is silly.

Issue #5037, opened 2025-02-10
Add Docling to Feast


franciscojavierarceo:
**Is your feature request related to a problem? Please describe.**
Feast should support a docling extra to allow easy usage of docling in the Feature Server so that a user can run docling as a transformation.

**Describe the solution you'd like**
An ODFV on write can use docling to chunk the data.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

franciscojavierarceo:
@ntkathole this would be a good one to add as an extra Nikhil if you have bandwidth.

franciscojavierarceo:
I think you can take that one over, the bandwidth is really the challenge that we have to get the CICD to work with pytorch. I started looking into it but it looks like it's going to take a bit and I have to prioritize some other items. :/

ntkathole:
We need to update uv version to fix dependencies. Old version of uv is no longer providing correct dependency resolution for us. Raised https://github.com/feast-dev/feast/pull/5086 to fix existing dependencies, once merged will add docling and torch.

Issue #4890, opened 2025-01-03
Add torch.Tensor as option for online and offline retrieval


franciscojavierarceo:
**Is your feature request related to a problem? Please describe.**
Feast should create an extra to support PyTorch tensors as a first-class citizen. 

**Describe the solution you'd like**

```python
# Fetch features from Feast
 feature_vector = feature_store.get_online_features(
        entity_rows=[entity_row],       
       features=self.feature_refs
 ).to_tensor(kind='torch') # default is torch, could support Tensorflow later
```

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #5089, opened 2025-02-25
feat: Added docling and pytorch as add on

ntkathole:

Adds docling and torch as add on to feast.
Integrating Docling would allow users to efficiently chunk text data within Online Feature Views (ODFVs) during the write process and torch will allow conversion of feature vectors into pytorch tensors.

#5037 #4890

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -527,6 +527,12 @@ build-feast-operator-docker:
 
 # Dev images
 
+build-feature-server-dev-minimal:
+	docker buildx build \
+		-t feastdev/feature-server:dev \
+		-f sdk/python/feast/infra/feature_servers/multicloud/Dockerfile \
+		--load sdk/python/feast/infra/feature_servers/multicloud
+
 build-feature-server-dev:
 	docker buildx build \
 		-t feastdev/feature-server:dev \
--- a/infra/feast-operator/Makefile
+++ b/infra/feast-operator/Makefile
@@ -162,7 +162,7 @@ docker-build: ## Build docker image with the manager.
 ## Build feast docker image.
 .PHONY: feast-ci-dev-docker-img
 feast-ci-dev-docker-img:
-	cd ./../.. && make build-feature-server-dev
+	cd ./../.. && make build-feature-server-dev-minimal
 
 
 .PHONY: docker-push
--- a/setup.py
+++ b/setup.py
@@ -143,6 +143,8 @@ DUCKDB_REQUIRED = ["ibis-framework[duckdb]>=9.0.0,<10"]
 
 DELTA_REQUIRED = ["deltalake"]
 
+DOCLING_REQUIRED = ["docling>=2.23.0"]
+
 ELASTICSEARCH_REQUIRED = ["elasticsearch>=8.13.0"]
 
 SINGLESTORE_REQUIRED = ["singlestoredb<1.8.0"]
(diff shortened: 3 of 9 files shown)


franciscojavierarceo (approved):

setup.py
```diff
@@ -143,6 +143,8 @@
 
 DELTA_REQUIRED = ["deltalake"]
 
+DOCLING_REQUIRED = ["docling>=2.23.0"]
```
franciscojavierarceo, line 146: Can we create an NLP_REQUIRED with TORCH, DOCLING, and Milvus in it? 

The goal would be to have something OOTB that combines these. HuggingFace and Gensim would probably be useful too but can be added later.

ntkathole:
It seems this is causing increase in an image size. Do we have large runners available under our github org?

franciscojavierarceo:
@ntkathole this is what I see available. Anything here work?

Screenshot 2025-02-25 at 12 28 14 PM

infra/feast-operator/Makefile
```diff
@@ -162,7 +162,7 @@ docker-build: ## Build docker image with the manager.
 ## Build feast docker image.
 .PHONY: feast-ci-dev-docker-img
 feast-ci-dev-docker-img:
-	cd ./../.. && make build-feature-server-dev
+	cd ./../.. && make build-feature-server-dev-minimal
```
ntkathole, line 165: @franciscojavierarceo Since we do not have larger runner to run kind + feast with all dependencies, either we need self-hosted runners or we need to use image with minimal dependencies installed in CI. 
Currently moving from [multicloud/Dockerfile.dev](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/feature_servers/multicloud/Dockerfile.dev) to [multicloud/Dockerfile](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/feature_servers/multicloud/Dockerfile) works in CI to run operator e2e test.

franciscojavierarceo: My concern is that we miss potential breaks by using this minimal image. That's the risk here, right?

Cc @lokeshrangineni @tchughesiv @redhatHameed

ntkathole: We will be using this image just for operator e2e test, feast functionality will be tested via integration and unit tests with all dependencies installed. 
@tchughesiv or someone more familiar with operator work can confirm if it's a big risk in operator testing or better way to handle this.

franciscojavierarceo:
looks like that worked!

infra/feast-operator/Makefile
tchughesiv: this change needs to be reverted

tchughesiv: @ntkathole this must remain `make build-feature-server-dev`

Makefile
```diff
@@ -527,6 +527,12 @@ build-feast-operator-docker:
 
 # Dev images
 
+build-feature-server-dev-minimal:
+	docker buildx build \
+		-t feastdev/feature-server:dev \
+		-f sdk/python/feast/infra/feature_servers/multicloud/Dockerfile \
+		--load sdk/python/feast/infra/feature_servers/multicloud
+
```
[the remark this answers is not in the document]
tchughesiv: we'll need to remove this

PR #5092, opened 2025-02-25
feat: Set value_type of entity directly in from_proto

nanohanno:

In the last release a deprecation warning was added to `Entity` if the `value_type` is `None`. Currently, the `from_proto` method creates an `Entity` object first without `value_type` and sets it afterwards. Therefore, for each creation a Warning log is issued.

This change moves the `value_type` into the creator of `Entity` in the `from_proto` method.

# Misc


franciscojavierarceo (approved):
nice lgtm

PR #5094, opened 2025-02-25
feat: Override the udf name when provided as input to an on demand transformation

blaketastic2:
This PR allows the user to define an ODFV function with a name override so we don't have to use the function name if we don't want to. This is also meant to provide feature parody with other ways of defining an ODFV.

N/A

# Misc

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,7 +202,7 @@ The list below contains the functionality that contributors are planning to deve
   * [x] Python Client
   * [x] [Python feature server](https://docs.feast.dev/reference/feature-servers/python-feature-server)
   * [x] [Java feature server (alpha)](https://github.com/feast-dev/feast/blob/master/infra/charts/feast/README.md)
-  * [x] [Go feature server (alpha)](https://github.com/feast-dev/feast/blob/master/go/README.md)
+  * [x] [Go feature server (alpha)](https://docs.feast.dev/reference/feature-servers/go-feature-server)
 * **Data Quality Management (See [RFC](https://docs.google.com/document/d/110F72d4NTv80p35wDSONxhhPBqWRwbZXG4f9mNEMd98/edit))**
   * [x] Data profiling and validation (Great Expectations)
 * **Feature Discovery and Governance**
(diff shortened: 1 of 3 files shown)


sdk/python/tests/unit/test_on_demand_feature_view.py
```diff
@@ -356,3 +357,68 @@ def test_on_demand_feature_view_stored_writes():
...
+            Field(name="output2", dtype=Float32),
+        ],
+    )(transform_features)
+
+    # Verify default name behavior
+    assert odfv.name == "transform_features"
+    assert isinstance(odfv, OnDemandFeatureView)
+
+    proto = odfv.to_proto()
+    assert proto.spec.name == "transform_features"
```
franciscojavierarceo, line 395: nit: can you make the `transform_features` and `custom-function-name` string variables at the top of the test?

README.md
blaketastic2, line 1: This was changed by the pre-commit hook...

sdk/python/tests/unit/test_on_demand_feature_view.py
blaketastic2: I've compromised. See if this works

README.md
franciscojavierarceo: Yeah, it's fine no worries.

sdk/python/tests/unit/test_on_demand_feature_view.py
```diff
@@ -24,9 +24,12 @@
     OnDemandFeatureView,
     PandasTransformation,
     PythonTransformation,
+    on_demand_feature_view,
 )
 from feast.types import Float32
 
+CUSTOM_FUNCTION_NAME = "custom-function-name"
```
franciscojavierarceo, line 31: nit: this could be adding in the scope of the `test_function_call_syntax()`

franciscojavierarceo (approved):
nice, thanks Blake!!!

blaketastic2:
@franciscojavierarceo this failure seems like a flaky test

franciscojavierarceo:
yeah it is

PR #5099, opened 2025-02-26
chore: Change the feature-server release container image to use a `requirements.txt` file for install

tchughesiv:

Changes feature-server image Dockerfile to use a `requirements.txt` file for the installation.

### DIFF
--- a/sdk/python/feast/infra/feature_servers/multicloud/requirements.txt
+++ b/sdk/python/feast/infra/feature_servers/multicloud/requirements.txt
@@ -1,0 +1,2 @@
+# keep VERSION on line #2, this is critical to release CI
+feast[aws,gcp,snowflake,redis,go,mysql,postgres,opentelemetry,grpcio,k8s,duckdb,milvus] == 0.46.0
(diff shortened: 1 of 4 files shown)


tmihalac (approved):

sdk/python/feast/infra/feature_servers/multicloud/requirements.txt
```diff
@@ -0,0 +1 @@
+feast[aws,gcp,snowflake,redis,go,mysql,postgres,opentelemetry,grpcio,k8s,duckdb,milvus] == ${VERSION}
```
ntkathole, line 1: Just a thought : How about having OPERATOR_REQUIRED/MULTICLOUD_REQUIRED in 
https://github.com/feast-dev/feast/blob/master/setup.py#L161 and then make command to install those dependencies in Dockerfile instead ?

redhatHameed (approved):

sdk/python/feast/infra/feature_servers/multicloud/requirements.txt
ntkathole: ah, please ignore, seems it is needed for konflux

redhatHameed:
@lokeshrangineni @tmihalac any thought on operator e2e test failing.

PR #5104, opened 2025-02-27
chore: Making embedding optional in retrieve_online_documents

franciscojavierarceo:
chore: Making embedding optional in retrieve_online_documents

# Misc

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -1863,11 +1863,11 @@ class FeatureStore:
 
     def retrieve_online_documents_v2(
         self,
-        query: Optional[List[float]],
-        top_k: int,
         features: List[str],
-        distance_metric: Optional[str] = "L2",
+        top_k: int,
+        query: Optional[List[float]] = None,
         query_string: Optional[str] = None,
+        distance_metric: Optional[str] = "L2",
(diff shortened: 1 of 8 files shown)


sdk/python/feast/feature_store.py
```diff
@@ -1876,15 +1876,14 @@ def retrieve_online_documents_v2(
             features: The list of features that should be retrieved from the online document store. These features can be
                 specified either as a list of string document feature references or as a feature service. String feature
                 references must have format "feature_view:feature", e.g, "document_fv:document_embeddings".
-            query: The query to retrieve the closest document features for.
+            query: The query to retrieve the closest document features for (optional)
```
dmartinol, line 1879: Probably it's `The embedded query` rather than `The query`

dmartinol: Also missed to doc the `query_string` param

dmartinol (commented):
lgtm

sdk/python/feast/feature_store.py
franciscojavierarceo: Good catch!

franciscojavierarceo:
Actually prowl doesn't work, do you mind clicking the approval?

dmartinol (approved):

PR #5118, opened 2025-03-05
fix: Cant add different type of list types

itay1551:
When pushing a data frame to a store and the data frame have a `List[int64]` (to store an array not used for embedding) I was getting an error of type not exits.

This fixed it.


franciscojavierarceo (commented):
Nice! Would you mind double checking if there's a test using this? Would be helpful!

itay1551:
I checked the tests, and there are no tests that pushing arrays different from type `Array(Float32)`

franciscojavierarceo:
Would you mind adding one? :D

franciscojavierarceo (approved):

Issue #41, opened 2019-01-07
Create proper OWNERS files for each sub-component


woop:
In order to speed up reviews and getting PRs merged, we should have proper OWNERS files defined. These YAML files are used by Prow to assign reviewers and approvers for specific sub-directories of the project. This means the appropriate people are assigned to PRs faster, and they can review and approve the PR faster because they are also more familiar with that area of the code 

The OWNERS files also allow us to give localized `/lgtm` and `/approve` access to new contributors.

Please see [this document](https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md) for more details.

PR #5119, opened 2025-03-05
chore: Consolidate wheels/source builds & artifacts for release

tchughesiv:

Since the `0.42.0` release, our workflow has no longer been uploading the sdist tarball to PyPi. This PR aims to fix this by consolidating the wheels and source builds to a single job and uploading said artifacts to the same temp storage which feeds the PyPi upload job.

It's worth noting that the separate source job appears to have previously been created to resolve some issues that the `make install-go-ci-dependencies` was causing. This make command is no longer in use as it has since been commented out.
https://github.com/feast-dev/feast/blob/75f5a90536f7caa566b38b9c368ec33a90d2bfa5/Makefile#L649-L651

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -29,7 +29,7 @@ on:
 
 jobs:
   build-python-wheel:
-    name: Build wheels
+    name: Build wheels and source
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
@@ -50,7 +50,7 @@ jobs:
         with:
           custom_version: ${{ github.event.inputs.custom_version }}
           token: ${{ github.event.inputs.token }}
-      - name: Build wheels
+      - name: Checkout version and install dependencies
         env:
           VERSION: ${{ steps.get-version.outputs.release_version }}
           PYPI_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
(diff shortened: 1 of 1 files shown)


franciscojavierarceo:
> @franciscojavierarceo in troubleshooting this fix... i'm realizing the `build-docker-images` pre-release job will never pass since it depends on a published feast version in PyPi.

Yeah, that I had made sure of.

.github/workflows/build_wheels.yml
```diff
@@ -54,62 +54,26 @@ jobs:
...
           git checkout ${VERSION}
-          pip install -U pip setuptools wheel twine
+          pip install -U pip setuptools wheel twine build
           make build-ui
           git status
           git restore go.mod go.sum
           git restore sdk/python/feast/ui/yarn.lock
-      - name: Build
-        run: |
-          python3 setup.py sdist
```
redhatHameed, line 102: why this removed ? as we need sdist for Cachi2 
documentation also suggest same reference command for generating sdist
https://docs.python.org/3.10/distutils/sourcedist.html

tchughesiv: using `-sw` will build both source(sdist) and wheels at the same time. i've tested this on my dev server.
```shell
$ python -m build --help
usage: python -m build [-h] [--version] [--verbose] [--sdist] [--wheel] [--outdir PATH] [--skip-dependency-check]
                       [--no-isolation | --installer {pip,uv}] [--config-setting KEY[=VALUE]]
                       [srcdir]

    A simple, correct Python build frontend.

    By default, a source distribution (sdist) is built from {srcdir}
    and a binary distribution (wheel) is built from the sdist.
    This is recommended as it will ensure the sdist can be used
    to build wheels.

    Pass -s/--sdist and/or -w/--wheel to build a specific distribution.
    If you do this, the default behavior will be disabled, and all
    artifacts will be built from {srcdir} (even if you combine
    -w/--wheel with -s/--sdist, the wheel will be built from {srcdir}).

positional arguments:
  srcdir                source directory (defaults to current directory)

options:
  -h, --help            show this help message and exit
  --version, -V         show program's version number and exit
  --verbose, -v         increase verbosity
  --sdist, -s           build a source distribution (disables the default behavior)
  --wheel, -w           build a wheel (disables the default behavior)
```

redhatHameed: Got it, thanks

tchughesiv: yes ... this worked much better

redhatHameed (approved):

tchughesiv:
@tokoko do you know why we were using a macos to separately handle our sdist build?

.github/workflows/build_wheels.yml
```diff
@@ -50,66 +50,26 @@ jobs:
...
-    runs-on: macos-13
-    steps:
-      - uses: actions/checkout@v4
-      - name: Setup Python
-        id: setup-python
-        uses: actions/setup-python@v5
-        with:
-          python-version: "3.11"
-          architecture: x64
-      - name: Setup Node
```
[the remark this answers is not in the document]
tchughesiv: yes, but we're consolidating and it's already being set in the new job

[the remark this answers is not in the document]
tchughesiv: check line #41

franciscojavierarceo (approved):

tokoko:
@tchughesiv nope, it was macos from the get-go, shouldn't make a difference one way or the other. I suspect the point of a separate job was simply to run it in parallel.

franciscojavierarceo:
@tchughesiv the local integration tests ran for like 5+ hours so I just merged it...that sounds like something we should look into.

PR #5124, opened 2025-03-06
feat: Provide callback function for threaded cache refresh failures

zabarn:

The previous implementation of `_start_thread_async_refresh` was constantly creating new background threads to handle refreshing the cache. This exposed the possibility of running into a `Runtime Error: Can't start new thread` when attempting to start one, leading to no refresh thread running a causing a stale cache.

We are switching to keeping the same thread running, but using time.sleep() to control when a refresh occurs.

We are also introducing a callback function that is passed through `SqlRegistry` to `CachingRegistry`, which allows users to be notified whenever cache refreshes fail, and handle them in whatever way they want.

# Misc

@EXPEbdodla

### DIFF
--- a/sdk/python/feast/infra/registry/caching_registry.py
+++ b/sdk/python/feast/infra/registry/caching_registry.py
@@ -1,11 +1,12 @@
 import atexit
 import logging
 import threading
+import time
 import warnings
 from abc import abstractmethod
 from datetime import timedelta
 from threading import Lock
-from typing import List, Optional
+from typing import Callable, List, Optional
 
 from feast.base_feature_view import BaseFeatureView
 from feast.data_source import DataSource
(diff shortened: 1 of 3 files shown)


sdk/python/feast/infra/registry/caching_registry.py
```diff
@@ -475,14 +485,35 @@ def _refresh_cached_registry_if_necessary(self):
...
-        self.registry_refresh_thread = threading.Timer(
-            cache_ttl_seconds, self._start_thread_async_refresh, [cache_ttl_seconds]
-        )
-        self.registry_refresh_thread.daemon = True
-        self.registry_refresh_thread.start()
+
+        def refresh_loop():
+            while not self._stop_event.is_set():
+                try:
+                    time.sleep(cache_ttl_seconds)
```
ntkathole, line 495:
```suggestion
                    self._stop_event.wait(self.cache_ttl_seconds)
```

franciscojavierarceo:
@zabarn looks like it's breaking some of the tests actually.

sdk/python/feast/infra/registry/caching_registry.py
```diff
@@ -475,14 +481,41 @@ def _refresh_cached_registry_if_necessary(self):
...
+                f"Registry cache refresh thread started with TTL {cache_ttl_seconds}"
+            )
+        except Exception as e:
+            logger.exception("Failed to start registry refresh thread: %s", e)
+            raise e
 
     def _exit_handler(self):
-        self.registry_refresh_thread.cancel()
+        logger.info("Exiting, setting stop event for registry cache refresh thread")
+        self._stop_event.set()
```
ntkathole, line 521: I think test is failing due to _exit_handler() does not properly stop the refresh thread. Currently it doesn’t ensure all async operations are cleaned up. 

```
if self.registry_refresh_thread:
        self.registry_refresh_thread.join(timeout=5)  # Ensure thread exits cleanly
```

zabarn: Thanks @ntkathole, that makes sense. Fixed the failing integration test.

Issue #4584, opened 2024-09-28
Unify Feature Transformations and Feature Views


franciscojavierarceo:
**Is your feature request related to a problem? Please describe.**

In Feast, the current On Demand Feature View executes feature transformations at read time. However, this behavior is not immediately obvious from the name "On Demand Feature View." The term "On Demand" doesn't explicitly convey when the transformation occurs, leading to potential confusion for users trying to understand the timing and execution flow of feature transformations. Additionally, there's inconsistency in how different types of feature views (batch, streaming, on-demand) are declared and used, which can make the codebase less intuitive and harder to maintain.

**Describe the solution you'd like**
I propose to refactor and rename the way transformations are defined in Feast to make the execution timing explicit. Introducing a @transform decorator with a type parameter that accepts a FeastTransformation enum can achieve this clarity. The enum would define the transformation types:

```python
from enum import Enum

class FeastTransformation(Enum):
    ON_READ = 1      # Equivalent to the current On Demand Feature View
    ON_WRITE = 2     # New: Transformations executed at write time
    BATCH = 3        # New: Batch processing transformations
    STREAMING = 4    # Migrates the existing stream_feature_view decorator

@transform(
    type=FeastTransformation.ON_WRITE,
    schema=[...],
    entity=[...],
    sources=[...],
)
def my_feature_view(inputs: List[Dict[str, float]]) -> List[Dict[str, float]]:
    # Transformation logic here
    return outputs
```

By explicitly declaring when the transformation should occur (`ON_READ`, `ON_WRITE`, `BATCH`, or `STREAMING`), the code becomes more readable and maintainable. 

This approach unifies the declaration of feature views and transformations, making it easier for users to understand and for developers to extend functionality in the future.

**Describe alternatives you've considered**

* Enhancing Documentation: Improving the existing documentation to better explain the behavior of On Demand Feature View without changing the code structure. However, this doesn't solve the inherent ambiguity in the naming.

* Renaming Existing Decorators: Simply renaming on_demand_feature_view to something like on_read_feature_view to make the execution timing clearer. While this improves clarity, it doesn't address the inconsistency in how different feature views are declared.
* Using Configuration Parameters: Adding parameters to existing decorators to specify execution timing. This could make the decorators more complex and doesn't provide a unified approach.

**Additional context**
Consistency and Extensibility: This change promotes a consistent method of declaring feature transformations, regardless of when they execute. It also sets a foundation for adding more transformation types in the future.

[...]

Issue #4696, opened 2024-10-25
Add Chaining Reusable Transformation Functions part of Feature Views


Vishnu-Rangiah:
**Is your feature request related to a problem? Please describe.**
Feature transformations should be comprised of reusable components in the form of python functions which can be built into a feature transformation pipeline. 

**Describe the solution you'd like**

Chaining transformation into a pipeline which can be reused across FVs would reduce transformation code duplication across the feature repo. 

Here is an example adapted from Tecton's transformation API: https://docs.tecton.ai/docs/defining-features/feature-views/transformations#a-feature-view-that-calls-a-pyspark-transformation-passing-two-pyspark-transformation-outputs

```
@transformation(mode="big_query")
def last_balance_time(transactions, max_user_transaction):
    return f"""SELECT t.user_id, t.current_balance, last_t.last_transaction_date as timestamp
                      FROM {transactions} t
                      INNER JOIN {max_user_transaction} last_t
                      ON t.user_id = last_t.user_id AND t.timestamp = last_t.last_transaction_date;"""

@transformation(mode="big_query")
def user_last_transaction_time(transactions):
    return f"""SELECT user_id, MAX(timestamp) AS last_transaction_date
                      FROM {transactions}
                      GROUP BY user_id"""

@feature_view(
    sources=[credit_data_batch],
    entities=[user],
    mode="pipeline", # creates a DAG from re-useable transformation functions
    batch_schedule=timedelta(days=1),
    schema=[Field("user_id", String), Field("timestamp", Timestamp), Field("current_balance", Float64)],
)
def user_last_balance(transactions):
    user_last_transaction_time = user_last_transaction_time(transactions)
    return last_balance_time(transactions, user_last_transaction_time)
```

**Describe alternatives you've considered**
 

**Additional context**
- https://github.com/feast-dev/feast/issues/4277

Issue #4277, opened 2024-06-13
Add feature transformation support to FeatureView


franciscojavierarceo:
**Is your feature request related to a problem? Please describe.**
FeatureViews should support transformations as well.

**Describe the solution you'd like**
```python
@batch_feature_view(
    sources=[credit_data_batch],
    entities=[user],
    mode="python",
    batch_schedule=timedelta(days=1),
    schema=[Field("user_id", String), Field("timestamp", Timestamp), Field("current_balance", Float64)],
)
def user_last_balance(transactions):
    return transactions[["user_id", "timestamp", "current_balance"]]

```

**Describe alternatives you've considered**
N/A

**Additional context**
Should behave similar to ODFV and Stream Feature Views and support Python and Pandas.

franciscojavierarceo:
I am starting to think having the decorator of feature view type is weird.

Instead we could just do something like:

```python
@transform(
    sources=[credit_data_batch],
    entities=[user],
    mode="python",
    batch_schedule=timedelta(days=1),
    schema=[Field("user_id", String), Field("timestamp", Timestamp), Field("current_balance", Float64)],
)
def user_last_balance(transactions):
    return some_computation(transactions[["user_id", "timestamp", "current_balance"]])

```

And maybe in the decorator we have a way to map to when the computation is happening? 

My view is that we need to provide clarity about when transformations happen (on demand, during a write, or in a stream--i.e., before a write) and I'm not sure the current way does this as obviously. Maybe it does.

franciscojavierarceo:
This will be solved https://github.com/feast-dev/feast/issues/4376

franciscojavierarceo:
Actually this is more general and does not handle the batch transformation use case so keeping this open. The ODFV writes do help for a subset of these items.

Vishnu-Rangiah:
Chaining transformation into a dag which can be reused across FVs would be nice to have an would reduce code duplication in the feature repo.

```
@transformation(mode="big_query")
def last_balance_time(transactions, max_user_transaction):
    return f"""SELECT t.user_id, t.current_balance, last_t.last_transaction_date as timestamp
                      FROM {transactions} t
                      INNER JOIN {max_user_transaction} last_t
                      ON t.user_id = last_t.user_id AND t.timestamp = last_t.last_transaction_date;"""

@transformation(mode="big_query")
def user_last_transaction_time(transactions):
    return f"""SELECT user_id, MAX(timestamp) AS last_transaction_date
[...]

PR #5130, opened 2025-03-09
feat: Kickoff Transformation implementation

HaoXuAI:

Created a Transformation interface. it still works with the current pandas_transformation, python_transformation etc.

The next step is refactor the BatchMaterializationEngine to make it works for both Materialization and Transformation.

https://github.com/feast-dev/feast/issues/4584
https://github.com/feast-dev/feast/issues/4277#issuecomment-2427772928
https://github.com/feast-dev/feast/issues/4696
# Misc


sdk/python/feast/batch_feature_view.py
```diff
@@ -95,3 +119,75 @@ def __init__(
             schema=schema,
             source=source,
         )
+
+    def get_feature_transformation(self) -> Optional[Transformation]:
+        if not self.udf:
+            return None
+        if self.mode == TransformationMode.pandas or self.mode == "pandas":
```
franciscojavierarceo, line 126: Probably can just do a dictionary mapping

sdk/python/feast/transformation/base.py
```diff
@@ -0,0 +1,98 @@
...
+class Transformation(ABC):
+    def __new__(
+        cls,
+        mode: Union[TransformationMode, str],
+        udf: Callable[[Any], Any],
+        name: Optional[str] = None,
+        udf_string: str = "",
+        tags: Optional[Dict[str, str]] = None,
+        description: str = "",
+        owner: str = "",
```
franciscojavierarceo, line 21: Probably can add the singleton parameter too

sdk/python/feast/transformation/mode.py
```diff
@@ -0,0 +1,9 @@
+from enum import Enum
+
+
+class TransformationMode(Enum):
+    PYTHON = "python"
```
franciscojavierarceo, line 5: Nice

franciscojavierarceo (commented):
Very nice!! 🚀🚀🚀

sdk/python/feast/batch_feature_view.py
HaoXuAI: sure will add that

sdk/python/feast/batch_feature_view.py
```diff
@@ -54,11 +61,15 @@ class BatchFeatureView(FeatureView):
...
     materialization_intervals: List[Tuple[datetime, datetime]]
+    udf: Optional[FunctionType]
+    udf_string: Optional[str]
+    feature_transformation: Optional[Transformation]
 
     def __init__(
         self,
         *,
         name: str,
+        mode: Union[TransformationMode, str],
```
franciscojavierarceo, line 72: Nice

sdk/python/feast/transformation/mode.py
```diff
@@ -0,0 +1,9 @@
...
+    PANDAS = "pandas"
+    spark = "spark"
```
franciscojavierarceo, line 7:
```suggestion
    SPARK = "spark"
```

franciscojavierarceo:
Do you think we can scope one type of batch transformation as our MVP? If so, which one would you feel most comfortable doing?

HaoXuAI:
> Do you think we can scope one type of batch transformation as our MVP? If so, which one would you feel most comfortable doing?

yep, maybe Spark SQL. that should be easy to implement, and we can get it live soon.

franciscojavierarceo:
> > Do you think we can scope one type of batch transformation as our MVP? If so, which one would you feel most comfortable doing?
> 
> yep, maybe Spark SQL. that should be easy to implement, and we can get it live soon.

LFG!

franciscojavierarceo:
FYI @HaoXuAI a very selfish goal I have is to use a Spark batch transformation to handling scaling embedding documents. 

So in an ideal world we could do something like:

```python
def create_embeddings(partitionData):  
  tokenizer = AutoTokenizer.from_pretrained("sentence-transformers/all-MiniLM-L6-v2")  
  model = AutoModel.from_pretrained("sentence-transformers/all-MiniLM-L6-v2")  
  for row in partitionData:  
      document = str(row.document)  
      inputs = tokenizer(document, padding=True, truncation=True, return_tensors="pt", max_length=512)  
      result = model(**inputs)  
      embeddings = result.last_hidden_state[:, 0, :].cpu().detach().numpy()  
      lst = embeddings.flatten().tolist()  
      yield [row.id, lst, "", "{}", None]
```
And
```python
embeddings = dataset_df.rdd.mapPartitions(create_embeddings)  
```

Used in a batch feature view and an on demand feature view (on write) to support offline and online processing of docs for RAG. The key thing here is that we'd be able to really scale RAG offline embedding and make the transition seamless to online.

HaoXuAI:
> FYI @HaoXuAI a very selfish goal I have is to use a Spark batch transformation to handling scaling embedding documents.
> 
> So in an ideal world we could do something like:
> 
> ```python
> def create_embeddings(partitionData):  
>   tokenizer = AutoTokenizer.from_pretrained("sentence-transformers/all-MiniLM-L6-v2")  
>   model = AutoModel.from_pretrained("sentence-transformers/all-MiniLM-L6-v2")  
>   for row in partitionData:  
>       document = str(row.document)  
>       inputs = tokenizer(document, padding=True, truncation=True, return_tensors="pt", max_length=512)  
>       result = model(**inputs)  
>       embeddings = result.last_hidden_state[:, 0, :].cpu().detach().numpy()  
>       lst = embeddings.flatten().tolist()  
>       yield [row.id, lst, "", "{}", None]
> ```
> 
> And
> 
> ```python
> embeddings = dataset_df.rdd.mapPartitions(create_embeddings)  
> ```
> 
> Used in a batch feature view and an on demand feature view (on write) to support offline and online processing of docs for RAG. The key thing here is that we'd be able to really scale RAG offline embedding and make the transition seamless to online.
> 
> This example is from the Pinecone docs here: https://docs.pinecone.io/integrations/databricks#3-create-the-vector-embeddings

This seems to be a really good example of the BatchFeatureView + transformation

HaoXuAI:
PR diverged, cherrypick at https://github.com/feast-dev/feast/pull/5181

Issue #5103, opened 2025-02-27
Add sqlite-vec to default installation


franciscojavierarceo:
**Is your feature request related to a problem? Please describe.**
Feast should support sqlite-vec out of the box, it should no longer be an add on.

**Describe the solution you'd like**
`pip install feast` should include `sqlite-vec`

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #5138, opened 2025-03-11
feat: Adding sqlite-vec to default installation

jyejare:
The changes in the PR:
- adds the `sqlite-vec` dependency to the required list of installation from the optional dependencies.
- The changes are made in both pyproject and setup files

#5103 

# Misc

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,6 +43,7 @@ dependencies = [
     "psutil",
     "bigtree>=0.19.2",
     "pyjwt",
+    "sqlite-vec==v0.1.7a2",
 ]
 
 [project.optional-dependencies]
@@ -106,7 +107,6 @@ singlestore = ["singlestoredb<1.8.0"]
 snowflake = [
     "snowflake-connector-python[pandas]>=3.7,<4",
 ]
-sqlite_vec = ["sqlite-vec==v0.1.6"]
 
 ci = [
     "build",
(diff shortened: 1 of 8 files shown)


franciscojavierarceo:
unit tests are failing for torch, thing we need to pin that exact version.

jyejare:
@tchughesiv I think since all the packages are now being built from source, We should now be able to move sqlite-vec to default installation. In short, merge this PR.

tchughesiv (commented):
if memory serves, `sqlite-vec` cannot be built from source, as it uses it's own proprietary build tool. this would very likely break downstream releases/builds.

tchughesiv:
if you decided to go through with this, one way to move forward w/ at least supporting `sqlite-vec` during offline installs would be to add the wheel to these files/locations -
https://github.com/feast-dev/feast/blob/01cd442fb9f76e0fecb6ed332891775b593922f1/sdk/python/feast/infra/feature_servers/multicloud/offline/hermeto-generic-artifacts.sh#L54-L55
https://github.com/feast-dev/feast/blob/01cd442fb9f76e0fecb6ed332891775b593922f1/sdk/python/feast/infra/feature_servers/multicloud/offline/Dockerfile.sdist#L113-L114
https://github.com/feast-dev/feast/blob/01cd442fb9f76e0fecb6ed332891775b593922f1/sdk/python/feast/infra/feature_servers/multicloud/offline/Dockerfile.sdist.release#L113-L114

then execute the `./sdk/python/feast/infra/feature_servers/multicloud/offline/hermeto-generic-artifacts.sh` script and commit the changes.

you would then need to ensure it all works as an offline image build by running this script on a `linux/amd64` machine -
`./sdk/python/feast/infra/feature_servers/multicloud/offline/offline-release-build.sh`

PR #5150, opened 2025-03-16
chore: Add new website

woop:

Creates a new website since Wordpress is broken. This has a much simpler/cleaner design, and is backed by Astro so we can also manage our blog content from the main repo.

image

Cursor 2025-03-15 20 06 35


franciscojavierarceo (approved):
Lgtm

PR #5157, opened 2025-03-17
feat: Feast-festure-server helm support more deployment config

maodahua:

This PR let the feast-festure-server helm chart support the env, volume and volumemount config in values.yaml. And if not provide any values, it will keep the same render result as before.

Why I raise this juse because I am deploy the server in GKE, but in out case we need access the AWS dynamoDB as the online store for test now. After my expriment, I need add the env, volume and volumemount in the deployment manifest, but the config can not support those fields, I have to patch them after helm upgrade. So just made this change to support them.

N/A

# Misc

N/A

### DIFF
--- a/infra/charts/feast-feature-server/README.md
+++ b/infra/charts/feast-feature-server/README.md
@@ -35,6 +35,7 @@ See [here](https://github.com/feast-dev/feast/tree/master/examples/python-helm-d
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | affinity | object | `{}` |  |
+| extraEnvs | list | `[]` | Additional environment variables to be set in the container |
 | feast_mode | string | `"online"` | Feast supported deployment modes - online (default), offline, ui and registry |
 | feature_store_yaml_base64 | string | `""` | [required] a base64 encoded version of feature_store.yaml |
 | fullnameOverride | string | `""` |  |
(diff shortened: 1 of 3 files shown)


redhatHameed:
@maodahua 
Thanks for this PR!  
FYI, We also introduced a Go-based Feast operator: [[feast-operator](https://github.com/feast-dev/feast/tree/master/infra/feast-operator)], which supports volume mounts and environment variables.

redhatHameed (approved):

maodahua:
Hi @redhatHameed , thanks for your approval! 
We still in the POC of this tool, futher more I think we can take a look of that 🍻

redhatHameed:
@maodahua I believe this will part of next release v0.48.0

mozupwork:
ok then the documentation in Readme.md is mis-leading. It shows supported fields in 0.47.0 which will actually be a part of 0.48.0

mozupwork:
@redhatHameed also may I ask when will that release happen? we also need these changes

maodahua:
> ok then the documentation in Readme.md is mis-leading. It shows supported fields in 0.47.0 which will actually be a part of 0.48.0

Hey @mozupwork , I think maybe you refer the main branch, but the PR have to include the doc updates, right? If you go to the helm repo to take a look, the last version is 0.47.0, so my PR not in that release. and if you want to refer the docs, you can  access the repo via the release tag, it will bring you to the right place.

BTW, I also want to know what is the time of next release. We bypass this one used kustomize patch the deployment now.

redhatHameed:
@franciscojavierarceo when we are planing for Feast next release ? Thanks

PR #5169, opened 2025-03-20
feat: Operator mounts the odh-trusted-ca-bundle configmap when deployed on RHOAI or ODH

redhatHameed:
The operator searches for the `odh-trusted-ca-bundle` ConfigMap and mounts it into the Feast containers. In case the standalone feast operator is deployed and the ConfigMap is not present, the operator will not mount it.

This change is required when the Feast operator is deployed with ODH or RHOAI.

example of volume mount. 
Screenshot 2025-03-20 at 2 34 59 PM

# Misc

### DIFF
--- a/infra/feast-operator/internal/controller/services/services_types.go
+++ b/infra/feast-operator/internal/controller/services/services_types.go
@@ -34,12 +34,16 @@ const (
 	DefaultOnlineStorePath    = "online_store.db"
 	svcDomain                 = ".svc.cluster.local"
 
-	HttpPort      = 80
-	HttpsPort     = 443
-	HttpScheme    = "http"
-	HttpsScheme   = "https"
-	tlsPath       = "/tls/"
-	tlsNameSuffix = "-tls"
+	HttpPort              = 80
+	HttpsPort             = 443
+	HttpScheme            = "http"
+	HttpsScheme           = "https"
+	tlsPath               = "/tls/"
+	tlsPathCustomCABundle = "/etc/pki/tls/custom-certs/ca-bundle.crt"
+	tlsNameSuffix         = "-tls"
+
+	caBundleAnnotation = "config.openshift.io/inject-trusted-cabundle"
+	caBundleName       = "odh-trusted-ca-bundle"
 
 	DefaultOfflineStorageRequest  = "20Gi"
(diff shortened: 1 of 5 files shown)


infra/feast-operator/internal/controller/services/services_types.go
```diff
@@ -34,12 +34,14 @@ const (
...
-	tlsPath       = "/tls/"
-	tlsNameSuffix = "-tls"
+	HttpPort              = 80
+	HttpsPort             = 443
+	HttpScheme            = "http"
+	HttpsScheme           = "https"
+	tlsPath               = "/tls/"
+	tlsPathCustomCABundle = "/etc/pki/tls/custom-certs/ca-bundle.crt"
+	tlsNameSuffix         = "-tls"
+	CA
```
tchughesiv, line 44: is this line intentional?
```suggestion
```

redhatHameed: good catch removed, added by mistake.

infra/feast-operator/internal/controller/services/client.go
```diff
@@ -86,3 +99,35 @@ func (feast *FeastServices) initCaConfigMap() *corev1.ConfigMap {
 	cm.SetGroupVersionKind(corev1.SchemeGroupVersion.WithKind("ConfigMap"))
 	return cm
 }
+
+// GetCustomCertificatesBundle retrieves the custom CA bundle ConfigMap if it exists
+func (feast *FeastServices) GetCustomCertificatesBundle() CustomCertificatesBundle {
```
tchughesiv, line 104: does it make more sense to put this with the rest of the TLS methods instead of client.go?

tmihalac:
The code LGTM but I think you should add the basic tests with caBundle present and without assertions

redhatHameed:
> The code LGTM but I think you should add the basic tests with caBundle present and without assertions

Thanks for reviewing added [tests](https://github.com/feast-dev/feast/compare/fda90ed4c9f652fc52ac459b651ad70fbdf99c2f..44a316675c5fcf94a005cc6de10e179830c652a0)

redhatHameed:
@tchughesiv can I get approval on this. Thanks

tchughesiv (approved):

PR #5172, opened 2025-03-21
feat: Enable transformations on PDFs

franciscojavierarceo:

This PR updates feature transformations, online store handling, and support for PDFs during transformation.

- `sdk/python/feast/infra/online_stores/milvus_online_store/milvus.py`
    - Updated `online_read` to include a new condition for updating the `feature_name_feast_primitive_type_map` based on the `table.schema`.
    - Added support for `string_val` in the list of value types.
    - Modified error message to correctly reflect the unsupported value type and feature.

- `sdk/python/feast/feature_store.py`
    - Enhanced `_get_feature_view_and_df_for_online_write` to handle feature view transformations differently for singleton and non-singleton cases.
    - Improved handling of transformed data and input data frame creation.

- `sdk/python/feast/on_demand_feature_view.py`
    - Added support for `PDF_BYTES` in the random input construction.

- `sdk/python/feast/transformation/python_transformation.py`
    - Improved type inference for nested lists and singleton feature views.

- `sdk/python/feast/types.py`
    - Introduced `PDF_BYTES` to the primitive feast types and updated related mappings and enums.

- `sdk/python/feast/value_type.py`
    - Added `PDF_BYTES` to the `ValueType` enum.

- `sdk/python/tests/unit/online_store/test_online_retrieval.py`
    - Added a new test `test_milvus_stored_writes_with_explode` to validate storing and retrieving exploded document embeddings with Milvus online store.
    - Renamed `test_milvus_lite_get_online_documents_v2` to `test_milvus_lite_retrieve_online_documents_v2`.

- `sdk/python/tests/unit/test_on_demand_python_transformation.py`
    - Added platform and version-specific skip condition for tests.
    - Introduced new tests for PDF handling and document transformation with chunking and embedding.

https://github.com/feast-dev/feast/issues/5173

# Misc:
N/A


sdk/python/feast/on_demand_feature_view.py
```diff
@@ -678,6 +678,9 @@ def _construct_random_input(
     ) -> dict[str, Union[list[Any], Any]]:
         rand_dict_value: dict[ValueType, Union[list[Any], Any]] = {
             ValueType.BYTES: [str.encode("hello world")],
+            ValueType.PDF_BYTES: [
```
HaoXuAI, line 681: Is it necessary to have a new type, maybe just use BYTES?

HaoXuAI (commented):
Overall looks good, just not sure if PDF_BYTES is a primitive value or not.

franciscojavierarceo:
We need it for validation of ODFV. Using raw bytes fails when you test with PDF otherwise.

sdk/python/feast/on_demand_feature_view.py
franciscojavierarceo: Yes, running type inference on raw bytes will fail when using a transformation expecting PDF content.

HaoXuAI:
> We need it for validation of ODFV. Using raw bytes fails when you test with PDF otherwise.

sounds good

HaoXuAI (approved):

PR #5175, opened 2025-03-21
feat: Cache get_any_feature_view results

sebastjaeger:

Profiling indicates that a large amount of time is spent in creating features views from the protobuf representations for every call to the online feature retrieval interfaces.

In our testing setup where features are retrieved from 24 features views, 80%+ of the time spent in Feast excluding networking operations is due to the above. On my local machine (M2 Mac), this time can exceed 15ms.

This PR introduces a decorator that caches the results of `get_any_feature_view`, which will remove the need to call `FeatureView.from_proto` in many cases when retrieving online features.

In principle, the decorator could also be used for other functions such as `get_feature_view`.

# Misc

### DIFF
--- a/sdk/python/feast/infra/registry/proto_registry_utils.py
+++ b/sdk/python/feast/infra/registry/proto_registry_utils.py
@@ -71,6 +71,31 @@ def registry_proto_cache_with_tags(func):
     return wrapper
 
 
+def registry_proto_cache_name(func):
+    cache_proto_key = None
+    cache = {}
+
+    @wraps(func)
+    def wrapper(registry_proto: RegistryProto, name: str, project: str):
+        nonlocal cache_proto_key, cache
+
+        proto_key = (id(registry_proto), registry_proto.version_id)
+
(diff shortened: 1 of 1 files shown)


sdk/python/feast/infra/registry/proto_registry_utils.py
```diff
@@ -94,6 +119,7 @@ def get_feature_service(
     raise FeatureServiceNotFoundException(name, project=project)
 
 
+@registry_proto_cache_name
```
franciscojavierarceo, line 122: is there a reason this would be preferred to `functools.cache` or `functools.lru_cache`?

franciscojavierarceo: Awesome, thanks for the thorough response!

sebastjaeger: Of course!

franciscojavierarceo (approved):

PR #5178, opened 2025-03-21
feat: Add async DynamoDB timeout and retry configuration

sebastjaeger:

In some cases it can be useful to be able to control the timeout and retry behavior of (async) DynamoDB connections. This PR exposes the respective functionality of (aio)botocore in the current project. By default, `connect_timeout` and `read_timeout` are set to 60 seconds, which is the default in (aio)botocore (https://botocore.amazonaws.com/v1/documentation/api/latest/reference/config.html). Therefore this value is currently also implicitly used.

The `DynamoDBOnlineStoreRetryConfig` model was introduced to stay close to what the configuration would look like in botocore. If preferred, the respective fields could also be moved into the top level config model, i.e., DynamoDBOnlineStoreConfig.

# Misc

### DIFF
--- a/sdk/python/feast/infra/online_stores/dynamodb.py
+++ b/sdk/python/feast/infra/online_stores/dynamodb.py
@@ -96,9 +96,7 @@ class DynamoDBOnlineStoreConfig(FeastConfigBaseModel):
     total_max_retry_attempts: Union[int, None] = None
     """Maximum number of total attempts that will be made on a single request.
 
-    Maps to `retries.total_max_attempts` in botocore.config.Config. Cf.
-    https://github.com/boto/botocore/blob/dd8406d5fa1df18037d1dd2977aec47334f7e3ce/botocore/args.py#L558
-    as for why `retries.max_attempts` is not exposed here.
+    Maps to `retries.total_max_attempts` in botocore.config.Config.
     """
 
     retry_mode: Union[Literal["legacy", "standard", "adaptive"], None] = None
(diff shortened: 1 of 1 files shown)


franciscojavierarceo (changes requested):
let's move the new fields to the single config

sebastjaeger:
Moved the retry configuration into `DynamoDBOnlineStoreConfig`. 

While doing so, I also did some digging into botocore, which led me to remove the `max_attempts` configuration field. The reason is that `max_attempts` is either ignored when `total_max_attempts` is set or is only used to set `total_max_attempts` if the latter is not provided: https://github.com/boto/botocore/blob/dd8406d5fa1df18037d1dd2977aec47334f7e3ce/botocore/args.py#L558

The retry configuration is provided to AioConfig as a dict (`retries`), where keys are only set if they are not None. This is to ensure that the retry default configuration remains unaffected if the respective values are not set.

Note the naming: 
* `total_max_retry_attempts` maps to `total_max_attempts` in the retries dict of botocore.config.Config
* `retry_mode` maps to `mode` in the aforementioned dict

sebastjaeger:
Force push to fix missing signoffs.

sebastjaeger:
Not clear if test failures are due to the changes. Potential networking issues.

franciscojavierarceo (approved):
thank you for this!!

sebastjaeger:
My pleasure :)

PR #5063, opened 2025-02-17
chore: Clean up go feature server docs

isuyyy:
# What this PR does / why we need it:

Clean up documentation about deploying go feature server using CLI since it is not supported anymore.

# Which issue(s) this PR fixes:

relates to #5051 

# Misc

PR #5194, opened 2025-03-27
fix: Update the doc

redhatHameed:
update the doc to add link for examples
added link offline server and registry server
removed the link for go-feature-server

# Misc

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -144,6 +144,7 @@
   * [Python feature server](reference/feature-servers/python-feature-server.md)
   * [\[Alpha\] Go feature server](reference/feature-servers/go-feature-server.md)
   * [Offline Feature Server](reference/feature-servers/offline-feature-server.md)
+  * [Registry server](reference/feature-servers/registry-server.md)
 * [\[Beta\] Web UI](reference/alpha-web-ui.md)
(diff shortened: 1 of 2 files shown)


docs/SUMMARY.md
```diff
@@ -142,8 +142,8 @@
   * [.feastignore](reference/feature-repository/feast-ignore.md)
 * [Feature servers](reference/feature-servers/README.md)
   * [Python feature server](reference/feature-servers/python-feature-server.md)
-  * [\[Alpha\] Go feature server](reference/feature-servers/go-feature-server.md)
```
franciscojavierarceo, line 145: why remove?

redhatHameed: This just a link the actual file remove in this PR #5063

franciscojavierarceo: actually we should probably revert that PR as @shuchu has been doing work on the go feature server.

redhatHameed: If that's the plan, then I've added it back. But it will be broken until PR [#5063](https://github.com/feast-dev/feast/pull/5063) is reverted.

franciscojavierarceo: @shuchu do you mind reverting?

PR #5202, opened 2025-03-31
feat: Enable users to use Entity Query as CTE during historical retrieval

blaketastic2:
This PR gives users the option to use CTEs for querying historical features instead of the current approach of creating a table, using the table and then dropping the table.

The option sits behind a flag that's configurable on the Offline Store configuration.

When trying to join more than 1 or 2 entities to a feature view, we started noticing performance issues. During this time, we were able to manipulate the query SQL to use a CTE instead of the TABLE created by default, and saw noticeable improvements in getting results back.

We believe this will have limited if any downsides to the existing implementation.

# Misc
There's a note that this SQL template should stay in sync w/ Redshift, but we are now drifting.

### DIFF
--- a/docs/reference/offline-stores/postgres.md
+++ b/docs/reference/offline-stores/postgres.md
@@ -32,6 +32,7 @@ offline_store:
   sslkey_path: /path/to/client-key.pem
   sslcert_path: /path/to/client-cert.pem
   sslrootcert_path: /path/to/server-ca.pem
+  entity_select_mode: temp_table
 online_store:
     path: data/online_store.db
 ```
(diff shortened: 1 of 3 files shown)


franciscojavierarceo (commented):
Awesome, would you mind updating the documentation for Postgres while you're here?

blaketastic2:
@franciscojavierarceo Let me know if the updated docs are what you were looking for.

franciscojavierarceo (approved):
this lgtm, as a follow up you may want to have a way to enforce the temp_table to be unique if two people end up having collisions with the desire to operate on different partitions of the data.

blaketastic2:
I believe in this particular case, the `entity_query` CTE will run without the need to worry about partitioning. Also, it looks like a flaky test failure again.

PR #4971, opened 2025-01-27
feat: Adding support to return additional features from vector retrieval for Milvus db

franciscojavierarceo:
# What this PR does / why we need it:
This PR adds support for returning multiple fields when doing  vector similarity search with Milvus.

The behavior is now as simple as:
```python
# Define your FeatureView
document_embeddings = FeatureView(
    name="embedded_documents",
    entities=[item, author],
    schema=[
       # notice that we define key index and metric search information at the field level
        Field(
            name="vector",
            dtype=Array(Float32),
            vector_index=True,
            vector_search_metric="COSINE",
        ),
        Field(name="item_id", dtype=Int64),
        Field(name="author_id", dtype=String),
        Field(name="sentence_chunks", dtype=String),
        Field(name="created_timestamp", dtype=UnixTimestamp),
        Field(name="event_timestamp", dtype=UnixTimestamp),
    ],
    source=rag_documents_source,
    ttl=timedelta(hours=24),
)

# Get the documents
result = store.retrieve_online_documents_v2(
    features=[
        "embedded_documents:vector",
        "embedded_documents:item_id",
        "embedded_documents:author_id",
        "embedded_documents:sentence_chunks",
    ],
    query=query_embedding,
    top_k=3,
).to_dict()
```

1. **Feature Store Enhancements**
   - **Added** `retrieve_online_documents_v2` method to `FeatureStore` for improved document retrieval using embeddings.
   - **Modified** `retrieve_online_documents` to enhance performance and support additional features.
   - **Introduced** `_retrieve_from_online_store_v2` to support the new document retrieval method internally.
   - **Added** `_get_feature_view_vector_field_metadata` to fetch metadata for vector fields within a `FeatureView`.

2. **Utility Functions**
   - **Added** `_get_unique_entities_from_values` to obtain unique composite entities and their indexes.
   - **Introduced** `_populate_response_from_feature_data_v2` to populate responses with feature data.
   - **Enhanced** `_extract_proto_values_to_dict` to handle serialization to string and vector columns.

3. **Provider and Online Store Updates**
   - **Updated** `PassthroughProvider` and `Provider` interfaces to include `retrieve_online_documents_v2`.
   - **Enhanced** `MilvusOnlineStore` and `OnlineStore` with support for the new document retrieval method and vector field handling.
   - **Modified** `MilvusOnlineStoreConfig` to default to `COSINE` metric type for vector searches.

4. **Testing and Example Repositories**
[...]

PR #5224, opened 2025-04-05
fix: Fixed integration tests for qdrant and milvus

ntkathole:
Fixed integration tests for milvus and qdrant.

```
make test-python-universal-milvus-online

make test-python-universal-qdrant-online
```

- sdk/python/feast/feature_store.py
       - Removed override of distance_metric in _retrieve_from_online_store. This was always overriding distance_metric with default value of vector_search_metric even wrong value is passed. It's required only in v2 method. It was added in https://github.com/feast-dev/feast/pull/4971 . 
     
- sdk/python/feast/infra/online_stores/qdrant_online_store/qdrant.py 
     - Fixed `json.loads() `when `get_list_val_str` returns `None`. In `online_write_batch()`, we are assuming that every value will be a a vector/embedding. But some of the features (like `string_feature: "a"`) are not embeddings. 

- sdk/python/tests/integration/feature_repos/repo_configuration.py
    - Fixed entity_key_serialization_version set logic.

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -2020,10 +2020,6 @@ class FeatureStore:
         """
         Search and return document features from the online document store.
         """
-        vector_field_metadata = _get_feature_view_vector_field_metadata(table)
-        if vector_field_metadata:
-            distance_metric = vector_field_metadata.vector_search_metric
-
         documents = provider.retrieve_online_documents(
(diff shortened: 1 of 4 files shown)


franciscojavierarceo (approved):

YassinNouh21:
@franciscojavierarceo can u merge it please

PR #5229, opened 2025-04-06
fix: Dask pulling of latest data

### LINKED ISSUE #5228: Dask offline store fails to include the most recent data point
While retrieving the latest data from the file source, Dask fails to include the most recent data point.

For an example - If we have dataset with the `event_timestamp` from `2025-03-21 16:00:00` till `2025-04-05 15:00:00` and suppose we ran `feast materialize 2021-04-12T07:00:00 2025-04-05T15:00:00`, the data/event which gets ingested in online store is of timestamp `2025-04-05T14:00:00` instead of `2025-04-05T15:00:00`

patilsuraj767:
Fixes dask condition to pull latest data point. 

Fixes #5228

### DIFF
--- a/sdk/python/feast/infra/offline_stores/dask.py
+++ b/sdk/python/feast/infra/offline_stores/dask.py
@@ -365,13 +365,13 @@ class DaskOfflineStore(OfflineStore):
                 if start_date and end_date:
                     source_df = source_df[
                         source_df[timestamp_field].between(
-                            start_date, end_date, inclusive="left"
+                            start_date, end_date, inclusive="both"
                         )
                     ]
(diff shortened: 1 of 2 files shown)


franciscojavierarceo (approved):

franciscojavierarceo (changes requested):
Looks like tests are failing

franciscojavierarceo:
@patilsuraj767 can you resolve the conflicts on this? Thanks a ton!

franciscojavierarceo (approved):
lgtm

probably could add a test for this in a follow up PR so someone doesn't accidentally break it in the future but appreciate the work here

Issue #5180, opened 2025-03-22
Doc: Remove Feature retrieval for Serialization version 2


jyejare:
Currently we are not supporting `Feature retrieval for Serialization/deserialization for version 2` and hence:

1. We should remove all its references from documentation, examples, etc.
2. We should ask the user to use serialization version 3.

PR #5248, opened 2025-04-08
feat: Serialization version 2 deprecation notice

jyejare:

- [x] Shows deprecation warning for usage of Entity serialization less than version 2
- [x] A util for re-serialization of entity ver 2 to ver 3
- [x] A guide for reserializing the entity keys in an offline/online store
- [x] A test case ensuring successful re-serilization of entity key 

Not fully but in the next release the serialization 2 would be completely deprecated to resolve https://github.com/feast-dev/feast/issues/5180 

# Misc

### DIFF
--- a/docs/how-to-guides/entity-reserialization-of-from-v2-to-v3.md
+++ b/docs/how-to-guides/entity-reserialization-of-from-v2-to-v3.md
@@ -1,0 +1,33 @@
+# Entity Key Re-Serialization from Version 2 to 3
+
+Entity Key Serialization version 2 will soon be deprecated, hence we need to shift the serialization and deserilization to version 3. 
+
+But here comes the challegnge where existing FeatuteViews on stores has written features with version 2. A version 2 serialized entity key cant be retrived using version 3 deserilization algorithm.
+
+## Reserialize the Feature Views entity Keys to version 3
+
+The solution is to reserialize the entity keys from version 2 to version 3.
+
+Follow the following procedures to reserialize the entity key to version 3 in feature View in an offline / online store.
+
+In hosrt, you need to iterate through all the feature views in your Feast repository, retrieve their serialized entity keys (if stored in version 2), reserialize them to version 3, and then update the online/offline store or wherever the serialized keys are stored.
+
+### 1. Initialize the Feature Store
+
+    Load the FeatureStore object to access all feature views in your repository.
+
+### 2. Iterate Through Feature Views
+
+    Use the list_feature_views() method to retrieve all feature views in the repository.
+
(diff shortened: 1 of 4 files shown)


HaoXuAI:
This is nice but needs to be declared more explicitly. Can you update the all docs/Readme that has `entity_key_serialization_version=2` to 2?

ntkathole:
Please update the templates as well https://github.com/feast-dev/feast/blob/master/sdk/python/feast/templates/gcp/feature_repo/feature_store.yaml#L25

jyejare:
@HaoXuAI @ntkathole This PR is just a `notice of deprecation` in the current release and helper function for users, that would help users to move to serialization version 3.

In next release, we will complete deprecating version 2 and below and stick to version 3. That time I will resolve your current queries @HaoXuAI @ntkathole :)

HaoXuAI:
> @HaoXuAI @ntkathole This PR is just a `notice of deprecation` in the current release and helper function for users, that would help users to move to serialization version 3.
> 
> In next release, we will complete deprecating version 2 and below and stick to version 3. That time I will resolve your current queries @HaoXuAI @ntkathole :)

sounds good!

HaoXuAI (approved):
LGTM!

joyneuberger:
If I've updated my entity_key_serialization_version, can't I just run feast apply again rather than using the utility function to reserialize?

jyejare:
@joyneuberger :
IF the data is stored/pushed in stores (offline/online) with version 2, 
AND IF you are attempting to read the features from stores with version 3,
THEN you will have conflicts casue desearilization of version 2 entities wont be possible with version 3.
HENCE Reserializing it the perfect TODO here.


BUT IF You have not yet pushed data to the stores,
AND IF you have changed the version to 3,
THEN you are free to run `feast apply`


Most important thing here is the `feast apply` does not uses entity_key_serialization_version in any means. Why ?

1. Basicaly feast apply is only suppose to create infra(like tables, schema etc) for storing the data in stores, and it does not writes the actual data to the stores.
2. The entity key serialization version is used during runtime operations, not during the apply procedure, like:
**Online Feature Retrieval:** When calling get_online_features(), the version is used to serialize entity keys for lookups in the online store.
**Online Feature Writing:** When writing features to the online store (during materialization or push operations), the version determines how entity keys are serialized.
**Document Retrieval:** For vector-based feature stores, the version affects how entity keys are serialized for similarity searches.


**Finally,** 
- If you are concerned about data in stores that is already pushed, please use reserialize function. 
- Else if you are concern about infra and the data is not already pushed to the stores, you can easily run feast apply.

Hope that answers !!

PR #5257, opened 2025-04-10
fix: Embed Query configuration breaks when switching between DataFrame and SQL

blaketastic2:
This PR fixes a bug introduced w/ the `entity_select_mode`. Currently, if a user sets `entity_select_mode="embed_query` and then passed a DataFrame, we fail. This should not happen as we can no longer be in a mixed state of both SQL queries w/ `entity_select_mode="embed_query` and DataFrames.

N/A

# Misc
N/A


blaketastic2:
@franciscojavierarceo there was a bug in the first iteration of this code and I'm hoping to get this patch into the next release.

Essentially, if the flag was set, you could no longer use DataFrames.

franciscojavierarceo (approved):

PR #5278, opened 2025-04-16
feat: Local compute engine

HaoXuAI:
Create LocalComputeEngine which supports Pandas and Polars as backend.
The source node and output node are going to be done in the next PR. Other nodes are implemented. It needs to update some offline store code to make I/O work.
Also refactor some SparkComputeEngine code

# Misc


sdk/python/feast/infra/common/materialization_job.py
```diff
@@ -0,0 +1,57 @@
...
+
+class MaterializationJobStatus(enum.Enum):
+    WAITING = 1
+    RUNNING = 2
+    AVAILABLE = 3
+    ERROR = 4
+    CANCELLING = 5
+    CANCELLED = 6
+    SUCCEEDED = 7
+
```
franciscojavierarceo, line 34:
```suggestion
PAUSED = 8
RETRYING = 9
```

franciscojavierarceo (approved):
Big PR but lgtm

PR #5281, opened 2025-04-18
feat: Offline store update pull_all_from_table_or_query to make timestampfield optional

HaoXuAI:

The major change in this PR is making the start_date and end_date in pull_all_from_table_or_query optional:
```
    def pull_all_from_table_or_query(
        config: RepoConfig,
        data_source: DataSource,
        join_key_columns: List[str],
        feature_name_columns: List[str],
        timestamp_field: str,
        created_timestamp_field: Optional[str] = None,
        start_date: Optional[datetime] = None,
        end_date: Optional[datetime] = None,
    ) -> RetrievalJob:
```
The reason is the API should be able to let the user skip the time range selection, and return all entities with features.
This will also make the current `materialization` and `get_historical_features` in Compute Engine to be able to use this API.
The change is non-breaking, and the API currently is only used in `retrieve_feature_service_logs`, and `retrieve_saved_dataset`.

# Misc

### DIFF
--- a/sdk/python/feast/infra/common/retrieval_task.py
+++ b/sdk/python/feast/infra/common/retrieval_task.py
@@ -1,6 +1,6 @@
 from dataclasses import dataclass
 from datetime import datetime
-from typing import Union
+from typing import Optional, Union
 
 import pandas as pd
 
@@ -15,5 +15,5 @@ class HistoricalRetrievalTask:
     feature_view: Union[BatchFeatureView, StreamFeatureView]
     full_feature_name: bool
     registry: Registry
-    start_time: datetime
-    end_time: datetime
+    start_time: Optional[datetime] = None
+    end_time: Optional[datetime] = None
--- a/sdk/python/feast/infra/compute_engines/feature_builder.py
+++ b/sdk/python/feast/infra/compute_engines/feature_builder.py
@@ -72,10 +72,10 @@ class FeatureBuilder(ABC):
     def build(self) -> ExecutionPlan:
         last_node = self.build_source_node()
 
-        # PIT join entities to the feature data, and perform filtering
-        if isinstance(self.task, HistoricalRetrievalTask):
-            last_node = self.build_join_node(last_node)
+        # Join entity_df with source if needed
+        last_node = self.build_join_node(last_node)
 
+        # PIT filter, TTL, and user-defined filter
         last_node = self.build_filter_node(last_node)
 
         if self._should_aggregate():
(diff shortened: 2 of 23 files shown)


sdk/python/feast/infra/compute_engines/local/feature_builder.py
```diff
@@ -95,25 +94,3 @@ def build_output_nodes(self, input_node):
         node = LocalOutputNode("output")
         node.add_input(input_node)
         self.nodes.append(node)
-
-    def build(self) -> ExecutionPlan:
```
HaoXuAI, line 99: we can remove it as it is the same with the Parent `build`

sdk/python/feast/infra/common/retrieval_task.py
```diff
@@ -15,5 +15,5 @@ class HistoricalRetrievalTask:
     feature_view: Union[BatchFeatureView, StreamFeatureView]
     full_feature_name: bool
     registry: Registry
-    start_time: datetime
-    end_time: datetime
+    start_time: Optional[datetime] = None
```
HaoXuAI, line 18: retrieval job doesn't have to provide start or end date

sdk/python/feast/infra/compute_engines/feature_builder.py
```diff
@@ -72,10 +72,10 @@ def _should_validate(self):
     def build(self) -> ExecutionPlan:
         last_node = self.build_source_node()
 
-        # PIT join entities to the feature data, and perform filtering
-        if isinstance(self.task, HistoricalRetrievalTask):
-            last_node = self.build_join_node(last_node)
+        # Join entity_df with source if needed
+        last_node = self.build_join_node(last_node)
```
HaoXuAI, line 76: without entity_df provided, the join node is just a pass through node

sdk/python/feast/infra/compute_engines/local/nodes.py
```diff
@@ -15,14 +16,39 @@
...
+        offline_store = context.offline_store
+        (
+            join_key_columns,
+            feature_name_columns,
+            timestamp_field,
+            created_timestamp_column,
+        ) = context.column_info
+
+        # 📥 Reuse Feast's robust query resolver
+        retrieval_job = offline_store.pull_all_from_table_or_query(
```
HaoXuAI, line 41: reuse the `pull_all_from_table_or_query` to read data from all feast offline_store now

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -206,10 +207,13 @@ def pull_all_from_table_or_query(
             + BigQueryOfflineStore._escape_query_columns(feature_name_columns)
             + [timestamp_field]
         )
+        timestamp_filter = get_timestamp_filter_sql(
```
HaoXuAI, line 210: major change: get the timestamp filter from `get_timestamp_filter_sql`

franciscojavierarceo (approved):

Issue #5166, opened 2025-03-20
Eliminate vector length restrictions in the store


itay1551:
## Expected Behavior 
Let's say I have two arrays in a store, the first one of size `k` and the other one of size `m`
In the current behavior of Feast, we define in the feature_store.yaml

```
online_store:
  ...
  vector_len: 64
  ...
```
Here `k == 64` is the size of the embedding vector that will be used in the online store,
If I want to push to the store an array that isn't of size `k == 64` I will get an error even if the array is  defined as `vector_index=False` in the Feature View,

From my perspective, the `vector_len` parameter needs to be defined `per field` in a FeatureView, or even not predefined.

## Current Behavior
When pushing to a store a vector of size `m` when `m != k` where in the feature_repo.yaml defined  ` vector_len: k` 
an error rise
`DataException: expected 64 dimensions, not 10`

## Steps to reproduce
Create feature view and push a vector of size that is different from the defined feature_repo.yaml file

### Specifications

- Version: 0.46.0
- Platform: local 
- Subsystem: Fedora

franciscojavierarceo:
This is a good first issue. The vector_len is now able to be defined at the Field level like so:

```python
document_embeddings = FeatureView(
    name="embedded_documents",
    entities=[item, author],
    schema=[
        Field(
            name="vector",
            dtype=Array(Float32),
            # Look how easy it is to enable RAG!
            vector_index=True,
            vector_search_metric="COSINE",
        ),
        Field(name="item_id", dtype=Int64),
        Field(name="author_id", dtype=String),
        Field(name="created_timestamp", dtype=UnixTimestamp),
        Field(name="sentence_chunks", dtype=String),
        Field(name="event_timestamp", dtype=UnixTimestamp),
    ],
    source=rag_documents_source,
    ttl=timedelta(hours=24),
)
```

So we can remove this dependency.

PR #5289, opened 2025-04-21
feat: Vector length definition moved to Feature View from Config 

jyejare:

- Vector length parameter removed from config
- Field class is modified to have  `Vector length` parameter.
- SQLite Online Store modified for vector length from FV field.
- ElasticSearch Online Store modified for vector length from FV field.
- Qdrant Online Store modified for vector length from FV field.
- Tests are modified for all above changes

Closes https://github.com/feast-dev/feast/issues/5166 

# Misc


franciscojavierarceo:
issues with unpacking data in sqlite-vec i think.

protos/feast/core/Feature.proto
```diff
@@ -42,4 +42,7 @@ message FeatureSpecV2 {
 
     // Metric used for vector similarity search.
     string vector_search_metric = 6;
+
+    // Field indicating the vector length
+    int32 vector_len = 7;
```
franciscojavierarceo, line 47:
```suggestion
    int32 vector_length = 7;
```

franciscojavierarceo: i think the three extra characters are worth it and i'm not sure why we didn't add it to begin with, sorry!

franciscojavierarceo (changes requested):
nit: can you change the protobuf to `vector_length`?

protos/feast/core/Feature.proto
```diff
@@ -44,5 +44,5 @@ message FeatureSpecV2 {
     string vector_search_metric = 6;
 
     // Field indicating the vector length
-    int32 vector_len = 7;
+    int32 vector_length = 7;
```
franciscojavierarceo, line 47: technically this is a breaking change, so we should add a new protobuf here.

franciscojavierarceo (approved):
lgtm

protos/feast/core/Feature.proto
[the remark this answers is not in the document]
franciscojavierarceo: nvm, we haven't released this yet. thanks @jyejare!

jyejare: Yup, the new protobuf is added , in this PR only

PR #5291, opened 2025-04-23
feat: Add feature view tags to dynamo tags

robhowley:

* currently only global online config tags for dynamo are added as table tags
* update the table creation to include feature view specific tags as well
* this is helpful for cost tracking, alerting by ownership, etc

# Misc

### DIFF
--- a/sdk/python/feast/infra/online_stores/dynamodb.py
+++ b/sdk/python/feast/infra/online_stores/dynamodb.py
@@ -140,19 +140,20 @@ class DynamoDBOnlineStore(OnlineStore):
 
     @staticmethod
     def _table_tags(online_config, table_instance) -> list[dict[str, str]]:
+        table_instance_tags = table_instance.tags or {}
         online_tags = online_config.tags or {}
+
         common_tags = [
-            {"Key": key, "Value": value} for key, value in online_tags.items()
+            {"Key": key, "Value": table_instance_tags.get(key) or value}
+            for key, value in online_tags.items()
         ]
-        return common_tags + (
-            [
-                {"Key": key, "Value": value}
-                for key, value in table_instance.tags.items()
-                if key not in online_tags
-            ]
-            if table_instance.tags
-            else []
-        )
+        table_tags = [
+            {"Key": key, "Value": value}
+            for key, value in table_instance_tags.items()
+            if key not in online_tags
+        ]
+
+        return common_tags + table_tags
 
     @staticmethod
     def _update_tags(dynamodb_client, table_name: str, new_tags: list[dict[str, str]]):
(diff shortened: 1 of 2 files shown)


robhowley:
> can you add a test here? :D

sorry i meant to open it as a draft, converted. apologize for the early notice, def going to add a test

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -138,6 +138,45 @@ async def close(self):
...
+
+    @staticmethod
+    def _update_tags(dynamodb_client, table_name: str, new_tags: list[dict[str, str]]):
+        table_arn = dynamodb_client.describe_table(TableName=table_name)["Table"][
+            "TableArn"
+        ]
+        current_tags = dynamodb_client.list_tags_of_resource(ResourceArn=table_arn).get(
+            "Tags"
+        )
+        if current_tags:
```
robhowley, line 165: if there are existing tags, remove those that will be changing w the update. either key no longer exists or the value changed

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -167,40 +198,43 @@ def update(
...
-            dynamodb_client.get_waiter("table_exists").wait(
-                TableName=_get_table_name(online_config, config, table_instance)
-            )
+            table_name = _get_table_name(online_config, config, table_instance)
+            dynamodb_client.get_waiter("table_exists").wait(TableName=table_name)
+            # once table is confirmed to exist, update the tags.
+            # tags won't be updated in the create_table call if the table already exists
+            if do_tag_updates[table_name]:
+                tags = self._table_tags(online_config, table_instance)
+                self._update_tags(dynamodb_client, table_name, tags)
```
robhowley, line 237: if the table already exists, then we should perform tag updates. otherwise we can skip that bc the tags would've been added in the create_table call

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -138,6 +138,38 @@ async def close(self):
...
+
+        common_tags = [
+            {"Key": key, "Value": table_instance_tags.get(key) or value}
+            for key, value in online_tags.items()
+        ]
+        table_tags = [
+            {"Key": key, "Value": value}
+            for key, value in table_instance_tags.items()
+            if key not in online_tags
+        ]
```
robhowley, line 154: the table level tags override the global where applicable. eg

```
# in yaml, default to platform team owns the infra
team: platform
```

```
# in feature view we override for one particular instance
tags={"team": "product-team"}
```

robhowley:
hey there @franciscojavierarceo 👋 any big picture concerns w this change?

robhowley:
thanks for approving, but I can't merge. does this need the ok-to-test tag or something?

robhowley:
awesome thanks 🚢 🥳

PR #5294, opened 2025-04-23
fix: Operator - remove default replicas setting from Feast Deployment

tchughesiv:

With this PR, the Operator will honor whatever number of replicas the Feast Deployment is scaled to (by the user, HPA, etc.).

Currently, the Operator enforces only 1 replica per Feature Store Deployment. We should leave this up to the user. While Feast does not support any kind of HA implementation, certain data store types may support horizontal scaling of feast.

Doing this also gets the Operator out of the business of making any future HA or scaling decisions.

# Misc


tchughesiv:
@jyejare @ntkathole

ntkathole (approved):
lgtm

jyejare (approved):

PR #5298, opened 2025-04-27
feat: Add sqlite offline store

franciscojavierarceo:

# Misc


tchughesiv (changes requested):
Operator unit test is failing because this new file-based data store type needs to be added to the Operator API validation -
https://github.com/feast-dev/feast/blob/0c9f11bf91939822671902da6b02d06d0aafe570/infra/feast-operator/api/v1alpha1/featurestore_types.go#L135-L146

PR #5300, opened 2025-04-28
chore: Refactor transform on write

franciscojavierarceo:
Refactored code to make it easier toread.

# Misc


HaoXuAI (approved):
this looks good to me.
But long term I think it's better to refactor with a factory pattern or an abstraction to avoid so many conditions check.

PR #5309, opened 2025-04-30
fix: Upperbound limit for protobuf generation

HaoXuAI:

# Misc


ntkathole:
Need to modify in https://github.com/feast-dev/feast/blob/master/pyproject.toml#L126
and probably also https://github.com/feast-dev/feast/blob/a714d3af2ac3d61e058228430958a4f9ab23d257/setup.py#L139

tchughesiv:
@HaoXuAI fyi https://github.com/feast-dev/feast/pull/5311

franciscojavierarceo (approved):

PR #5319, opened 2025-05-02
Devin/1745638554 sqlite offline store

franciscojavierarceo:

# Misc


tchughesiv (changes requested):
add here as well -
https://github.com/feast-dev/feast/blob/1af7c701762d617b3ae1f69172179a71f5c40185/infra/feast-operator/api/v1alpha1/featurestore_types.go#L303

PR #5342, opened 2025-05-09
feat: Added rest-apis serving option for registry server

ntkathole:
This PR add 
- option to serve HTTP REST apis for registry server via FastAPI. 
- implements REST endpoints corresponding to gRPC methods.
- REST endpoints mirror the gRPC interface but use intuitive HTTP methods and routes.
- tests for registry REST endpoints.

This will allow easier integration with browser based web apps/UIs. 

# Misc
 
`# feast serve_registry --rest-api`

Screenshot 2025-05-09 at 12 05 51 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,7 @@ The list below contains the functionality that contributors are planning to deve
 * **Feature Serving**
   * [x] Python Client
   * [x] [Python feature server](https://docs.feast.dev/reference/feature-servers/python-feature-server)
+  * [x] [Feast Operator (alpha)](https://github.com/feast-dev/feast/blob/master/infra/feast-operator/README.md)
   * [x] [Java feature server (alpha)](https://github.com/feast-dev/feast/blob/master/infra/charts/feast/README.md)
   * [x] [Go feature server (alpha)](https://docs.feast.dev/reference/feature-servers/go-feature-server)
   * [x] [Offline Feature Server (alpha)](https://docs.feast.dev/reference/feature-servers/offline-feature-server)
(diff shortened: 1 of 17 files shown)


ntkathole:
> @ntkathole when this flag is passed, does the rest API run instead of, or in addition to, the gRPC server?

@tchughesiv rest api run instead of grpc.

redhatHameed:
@ntkathole what about Feast RBAC ? does that also dependent on internal grpc implementation ?

franciscojavierarceo (commented):
I don't see the permissions assertions anywhere here.

I do see the get permissions but it's not clear to me that's asserting the permissions like it's done in the feature server. Am I missing something?

ntkathole:
@redhatHameed @franciscojavierarceo yes, RBAC authorization is same and dependent on gPRC server permissions assertion. Overall it works as :
**Authentication** 
- RBAC authentication is handled by the inject_user_details dependency and the auth_manager. 
- FastAPI has `dependencies=[Depends(inject_user_details)]` i.e. every incoming request passes through inject_user_details.  It also set user context `sm.set_current_user(current_user)` which is globally accessible during request lifecycle using `get_security_manager().current_user`

**Authorization**
- happens outside of gRPC or REST api server and happens within Feast itself. Here we are calling same python method which is called by gRPC handlers (grpc_handler.ListEntities(...)) that already enforce RBAC using the current user from the security manager., which has assert permissions logic. 

That's one of the important benefit of this approach, we do not need to duplicate the logic like RBAC/response structure again for REST APIs, everything works like gRPC maintaining consistency between them.

redhatHameed:
@ntkathole also if you can add an example just demonstrate this approach will be helpful for community, not in this PR maybe you can add once you get chance. Thanks

franciscojavierarceo:
@ntkathole can you make a follow up PR with the documentation?

PR #5364, opened 2025-05-17
feat: Allow to set registry server rest/grpc mode in operator

ntkathole:

This PR adds capability in feast-operator to allow user to deploy registry serving in REST or gRPC mode. 

Setting `mode: rest` will deploy registry with :
`feast serve_registry --rest-api --key /tls/registry/tls.key --cert /tls/registry/tls.crt -p 6571`

```
spec:
  feastProject: example
  services:
    registry:
      local:
        server:
          mode: rest
```
By default, registry serving is kept in rest mode.


infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -499,6 +499,9 @@ type ServerConfigs struct {
 	// required by the Feast components. Ensure that each volume mount has a corresponding
 	// volume definition in the Volumes field.
 	VolumeMounts []corev1.VolumeMount `json:"volumeMounts,omitempty"`
+	// Mode specifies the registry API serving mode.
+	// Allowed values: "grpc", "rest".
+	Mode string `json:"mode,omitempty"`
```
tchughesiv, line 504: is this specific to the registry? if so, you'll want to create a new `RegistryServerConfigs` struct with `ServerConfigs` inline.

also, thoughts on a bool instead? unless there was a reason you went w/ a `string` field?
```suggestion
	// Mode specifies the registry API serving mode.
	// Allowed values: "grpc", "rest".
	RestMode bool `json:"restMode,omitempty"`
```

tchughesiv (changes requested):

infra/feast-operator/api/v1alpha1/featurestore_types.go
ntkathole: Updated!

ntkathole:
> when rest is enabled for registry, does the default port change?

nope, same default port is used, unless changed explicitly by user.

infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -502,6 +502,14 @@ type ServerConfigs struct {
 	VolumeMounts []corev1.VolumeMount `json:"volumeMounts,omitempty"`
 }
 
+// RegistryServerConfigs creates a registry server for the feast service, with specified container configurations.
+type RegistryServerConfigs struct {
+	ServerConfigs `json:",inline"`
+	// RestAPIEnabled determines if the registry should serve using the REST API instead of gRPC.
+	// +kubebuilder:default=true
+	RestAPIEnabled bool `json:"restAPIEnabled,omitempty"`
```
tchughesiv, line 510: Does it make more sense for the field to match the flag/option name? "Enabled" is implited by setting to `true`. Also, this field should not be set by default as this is not the default behaviour of the registry's remote server. This flag should be intentionally set by the user as it is an "option" in the CLI.
```suggestion
	// RestAPI determines if the registry should serve using the REST API instead of gRPC.
	RestAPI bool `json:"restAPI,omitempty"`
```
One more note on this, we don't use kubebuilder to set defaults. If this was a feild where a default should be set, the current approach/design with this CR is to set that default in `status.applied`.

So, if the intention was to force this as a default setting (i think some broader discussions should happen first if so), you'd want to make the field a pointer bool `*bool` and handle the default in status.

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -499,6 +499,13 @@ func (feast *FeastServices) getContainerCommand(feastType FeastServiceType) []st
 	deploySettings := FeastServiceConstants[feastType]
 	targetPort := deploySettings.TargetHttpPort
 	tls := feast.getTlsConfigs(feastType)
+
+	if feastType == RegistryFeastType {
+		registry := feast.Handler.FeatureStore.Spec.Services.Registry
+		if registry != nil && registry.Local != nil && registry.Local.Server != nil && registry.Local.Server.RestAPIEnabled {
```
tchughesiv, line 505: Use the built-in method for checking if Local registry is set & the applied status when verifying whether to act on a field. This is because applied status will also have defaults set, etc.
```suggestion
	if feastType == RegistryFeastType && feast.isLocalRegistry() {
		registry := feast.Handler.FeatureStore.Status.Applied.Services.Registry
		if registry.Local.Server.RestAPIEnabled {
```

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -645,8 +652,10 @@ func (feast *FeastServices) getServerConfigs(feastType FeastServiceType) *feastd
 			return appliedServices.OnlineStore.Server
 		}
 	case RegistryFeastType:
-		if feast.isLocalRegistry() {
-			return appliedServices.Registry.Local.Server
+		if feast.isLocalRegistry() && appliedServices.Registry != nil &&
+			appliedServices.Registry.Local != nil &&
+			appliedServices.Registry.Local.Server != nil {
+			return &appliedServices.Registry.Local.Server.ServerConfigs
```
tchughesiv, line 658: `feast.isLocalRegistry()` already handles those checks.
```suggestion
		if feast.isLocalRegistry() {
			return &appliedServices.Registry.Local.Server.ServerConfigs
```

tchughesiv (changes requested):
@ntkathole please add unit tests to check that the rest option is being properly handled in the resulting Deployment object... and any other tests related to this PR that you think would be valuable. Also, please update the description to match any design changes that have occurred.

tchughesiv (changes requested):
Based on the failing e2e test around remote registry use, it's clear that using rest api mode will break registry's ability to be used as a "remote" provider type. this adds additional complexity that will need to be handled/checked ... likely in this method -
https://github.com/feast-dev/feast/blob/648c53dc64ad88077e49e56f0cfc70756d6a5824/infra/feast-operator/internal/controller/services/services.go#L749-L768
If restAPI is true, we'd want to notify the user in the Status that remote provider registry capabilities cannot be used.

It would probably be good idea to check this in with a unit test as well. Maybe ensure whatever status message you set is being handled correctly.

There are existing remote registry tests in `infra/feast-operator/internal/controller/featurestore_controller_test.go` for reference. Might be a good place to add more.

infra/feast-operator/api/v1alpha1/featurestore_types.go
ntkathole: Most of the operators (like kubeflow has EnableIstio/MetricsCollectorSpec.Enabled) follows Enabled in their naming convention for boolean fields so added enabled earlier, removed Enabled from field now in https://github.com/feast-dev/feast/pull/5364/commits/66fd234f724ca4265dbb70b37962f495b8cc04d2

I do want to apply some default settings like when ui is deployed, registry server served with rest api, but I will do it another PR and agree that can be discussed separately.

tchughesiv: > I do want to apply some default settings like when ui is deployed, registry server served with rest api

it's my understanding that the feast UI has no reliance on the registry remote server (grpc or rest). isn't the UI server a standalone component of feast? all it currently needs, as i understand it, is a valid `feature_store.yaml` with proper data store configs.

infra/feast-operator/internal/controller/featurestore_controller_test.go
```diff
@@ -1213,6 +1213,51 @@ var _ = Describe("FeatureStore Controller", func() {
 			Expect(cond.Message).To(Equal("Error: Remote feast registry of referenced FeatureStore '" + referencedRegistry.Name + "' is not ready"))
 		})
 
+		It("should handle local registry with REST API enabled", func() {
+			By("By properly setting RestAPI registry in status")
```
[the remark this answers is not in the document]
ntkathole: Added in https://github.com/feast-dev/feast/pull/5364/commits/49537546741f9fe76df2da270dde905094e13ebb

ntkathole:
> looking good, but we still need checks added as mentioned in this prior review comment - [#5364 (review)](https://github.com/feast-dev/feast/pull/5364#pullrequestreview-2849033911)

Can you provide an example here, I didn't get how user can set restAPI true in the remote registry ? I think we have only added restAPI flag option for local registry server configs.

ntkathole:
@tchughesiv Ready for review now

ntkathole:
> A user can reference a registry in another `FeatureStore` CR and use it as a `remote` registry -
> 
> https://github.com/feast-dev/feast/blob/6c92447d1507bff02451f77f134df0a24cbd8036/infra/feast-operator/test/testdata/feast_integration_test_crs/v1alpha1_remote_registry_featurestore.yaml#L11-L15
> 
> We already have tests in place (both e2e & unit) around this feature. Given the restAPI breaks this functionality, you'll need to add an additional check in -
> 
> https://github.com/feast-dev/feast/blob/648c53dc64ad88077e49e56f0cfc70756d6a5824/infra/feast-operator/internal/controller/services/services.go#L749-L768
> 
> and unit test(s) verifying the new check.. e.g. -
> 
> https://github.com/feast-dev/feast/blob/6c92447d1507bff02451f77f134df0a24cbd8036/infra/feast-operator/internal/controller/featurestore_controller_test.go#L992-L1010

This is handled in https://github.com/feast-dev/feast/pull/5364/commits/b991093c779a7d155ba2705faf29cc9e115b6bea Thank you.

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -224,8 +224,32 @@ func (feast *FeastServices) deployFeastServiceByType(feastType FeastServiceType)
...
+			registry := feast.Handler.FeatureStore.Status.Applied.Services.Registry
+			if registry.Local.Server.RestAPI != nil && *registry.Local.Server.RestAPI {
+				if err := feast.createRestService(feastType); err != nil {
+					return feast.setFeastServiceCondition(err, feastType)
+				}
+			} else {
+				// Delete REST API service if REST API is disabled
+				_ = feast.Handler.DeleteOwnedFeastObj(&corev1.Service{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      feast.GetFeastServiceName(feastType) + "-rest",
```
tchughesiv: should there be a new method to return the registry's rest object name? since it appears to be needed several times?

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -404,6 +428,16 @@ func (feast *FeastServices) setContainer(containers *[]corev1.Container, feastTy
...
+		if feastType == RegistryFeastType && feast.isRegistryServer() {
+			registry := feast.Handler.FeatureStore.Status.Applied.Services.Registry
+			if registry.Local.Server.RestAPI != nil && *registry.Local.Server.RestAPI {
+				container.Ports = append(container.Ports, corev1.ContainerPort{
+					Name:          name + "-rest",
+					ContainerPort: getTargetRestPort(feastType, tls),
+					Protocol:      corev1.ProtocolTCP,
+				})
+			}
+		}
```
tchughesiv: what if GRPC is set to false? do you remove it from the container ports?

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -499,6 +533,19 @@ func (feast *FeastServices) getContainerCommand(feastType FeastServiceType) []st
...
 	targetPort := deploySettings.TargetHttpPort
 	tls := feast.getTlsConfigs(feastType)
+
+	if feastType == RegistryFeastType && feast.isRegistryServer() {
+		registry := feast.Handler.FeatureStore.Status.Applied.Services.Registry
+		if registry.Local.Server.GRPC != nil && *registry.Local.Server.GRPC {
+			deploySettings.Args = append(deploySettings.Args, "--grpc")
+		} else if registry.Local.Server.GRPC != nil && !*registry.Local.Server.GRPC {
+			deploySettings.Args = append(deploySettings.Args, "--no-grpc")
+		}
```
tchughesiv:
```suggestion
		if registry.Local.Server.GRPC != nil {
		    if *registry.Local.Server.GRPC {
			deploySettings.Args = append(deploySettings.Args, "--grpc")
		    } else {
			deploySettings.Args = append(deploySettings.Args, "--no-grpc")
		    }
		}
```

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -224,8 +224,32 @@ func (feast *FeastServices) deployFeastServiceByType(feastType FeastServiceType)
 		if err := feast.createService(feastType); err != nil {
 			return feast.setFeastServiceCondition(err, feastType)
 		}
+		// Create REST API service if needed
+		if feastType == RegistryFeastType && feast.isRegistryServer() {
+			registry := feast.Handler.FeatureStore.Status.Applied.Services.Registry
+			if registry.Local.Server.RestAPI != nil && *registry.Local.Server.RestAPI {
+				if err := feast.createRestService(feastType); err != nil {
```
tchughesiv:
```suggestion
		if feastType == RegistryFeastType && feast.isRegistryServer() {
			registry := feast.Handler.FeatureStore.Status.Applied.Services.Registry
			if registry.Local.Server.RestAPI != nil && *registry.Local.Server.RestAPI {
				if err := feast.createRestService(); err != nil {
```

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -615,6 +662,70 @@ func (feast *FeastServices) setService(svc *corev1.Service, feastType FeastServi
 	return controllerutil.SetControllerReference(feast.Handler.FeatureStore, svc, feast.Handler.Scheme)
 }
 
+// createRestService creates a separate service for REST API
+func (feast *FeastServices) createRestService(feastType FeastServiceType) error {
+	if feastType != RegistryFeastType || !feast.isRegistryServer() {
+		return nil
+	}
```
tchughesiv:
```suggestion
// createRestService creates a separate service for the Registry REST API
func (feast *FeastServices) createRestService() error {
	if feast.isRegistryServer() {
```

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -615,6 +662,70 @@ func (feast *FeastServices) setService(svc *corev1.Service, feastType FeastServi
...
+		},
+	}
+
+	if err := controllerutil.SetControllerReference(feast.Handler.FeatureStore, svc, feast.Handler.Scheme); err != nil {
+		return err
+	}
+
+	logger := log.FromContext(feast.Handler.Context)
+	if op, err := controllerutil.CreateOrUpdate(feast.Handler.Context, feast.Handler.Client, svc, controllerutil.MutateFn(func() error {
+		return nil // No need to mutate as we set everything above
```
tchughesiv: the problem with this approach is that the operator can't modify the service when changes are necessary to an existing FeatureStore's service. For example, should the user decide to implement TLS, or disable it, in an already deployed `FeatureStore` CR. The `CreateOrUpdate` method won't/can't make the appropriate changes.

tchughesiv: If you're going to create a new `createRestService`, it should look nearly identical to the existing `createService` method ... for example -

```go
func (feast *FeastServices) createRestService() error {
	if feast.isRegistryServer() {
                feastType := RegistryFeastType
        	logger := log.FromContext(feast.Handler.Context)
        	svc := feast.initFeastRestSvc(feastType)
        	if op, err := controllerutil.CreateOrUpdate(feast.Handler.Context, feast.Handler.Client, svc, controllerutil.MutateFn(func() error {
        		return feast.setService(svc, feastType)
        	})); err != nil {
	        	return err
        	} else if op == controllerutil.OperationResultCreated || op == controllerutil.OperationResultUpdated {
	        	logger.Info("Successfully reconciled", "Service", svc.Name, "operation", op)
        	}
	}
        return nil
}
```
Any reason you can't use the existing `setService` method?

infra/feast-operator/internal/controller/services/util.go
```diff
@@ -125,6 +125,17 @@ func ApplyDefaultsToStatus(cr *feastdevv1alpha1.FeatureStore) {
...
+				if services.Registry.Local.Server.GRPC == nil {
+					defaultGRPC := true
+					services.Registry.Local.Server.GRPC = &defaultGRPC
+				}
+
+				// Set default for RestAPI: false if nil
+				if services.Registry.Local.Server.RestAPI == nil {
+					defaultRestAPI := false
+					services.Registry.Local.Server.RestAPI = &defaultRestAPI
+				}
```
tchughesiv: No reason to explicitly set this. It being `nil` is equivalent to `false`.
```suggestion
```

tchughesiv:
Looking good, a few comments.

infra/feast-operator/internal/controller/services/services.go
ntkathole: Good observation, added new method https://github.com/feast-dev/feast/pull/5364/commits/e28a5789fb03dadd4e4544608f86250956d370e3

infra/feast-operator/internal/controller/services/services.go
ntkathole: We do not remove that, should we? If yes, can be done in separate PR?

tchughesiv: considering you've added the ability to disable GRPC in this PR, the operator should reflect a `GRPC: false` configuration everywhere. this PR would be the right place to do so.

infra/feast-operator/internal/controller/services/services.go
ntkathole: I changed rest service method to look nearly identical to the existing `createService`. Also, used existing setService method inside createRestService. https://github.com/feast-dev/feast/pull/5364/commits/63af9dadd1584c1b288b24b0f06c783411c6face

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -583,7 +632,7 @@ func (feast *FeastServices) setInitContainer(podSpec *corev1.PodSpec, fsYamlB64
 	}
 }
 
-func (feast *FeastServices) setService(svc *corev1.Service, feastType FeastServiceType) error {
+func (feast *FeastServices) setService(svc *corev1.Service, feastType FeastServiceType, isRestService ...bool) error {
```
tchughesiv:
```suggestion
func (feast *FeastServices) setService(svc *corev1.Service, feastType FeastServiceType, isRestService bool) error {
```

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -599,6 +648,14 @@ func (feast *FeastServices) setService(svc *corev1.Service, feastType FeastServi
 		port = HttpsPort
 		scheme = HttpsScheme
 	}
+
+	var targetPort int32
+	if len(isRestService) > 0 && isRestService[0] {
```
tchughesiv:
```suggestion
	if isRestService {
```

sdk/python/feast/cli/serve.py
```diff
@@ -232,7 +232,15 @@ def serve_registry_command(
...
-        store.serve_registry(port, tls_key_path, tls_cert_path, rest_api)
+        if grpc:
+            store.serve_registry(port, tls_key_path, tls_cert_path)
+        else:
+            store.serve_registry(
+                port=rest_port,
+                tls_key_path=tls_key_path,
+                tls_cert_path=tls_cert_path,
+                rest_api=rest_api,
+            )
```
tchughesiv, line 243: considering this is an operator related PR, what is the purpose of this change? should it be added to PR description?

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -1102,9 +1119,30 @@ func getTargetRestPort(feastType FeastServiceType, tls *feastdevv1alpha1.TlsConf
...
+		if feast.isRegistryGrpcEnabled() {
+			targetPort := getTargetPort(feastType, tls)
+			return corev1.ProbeHandler{
+				TCPSocket: &corev1.TCPSocketAction{
+					Port: intstr.FromInt(int(targetPort)),
+				},
+			}
+		}
+	} else if feastType == OnlineFeastType {
+		targetPort := getTargetPort(feastType, tls)
```
tchughesiv: `targetPort` should only have to be set once, right? also, no need to change how grpc probe is handled. this is only for rest. lastly, this isn't an `else if` situation with online.
```suggestion
	targetPort := getTargetRestPort(feastType, tls)
	if feastType == RegistryFeastType && feast.isRegistryRestEnabled() {
		probeHandler := corev1.ProbeHandler{
			HTTPGet: &corev1.HTTPGetAction{
				Port: intstr.FromInt(int(targetPort)),
			},
		}
		if tls.IsTLS() {
			probeHandler.HTTPGet.Scheme = corev1.URISchemeHTTPS
		}
		return probeHandler
	}
	if feastType == OnlineFeastType {
```

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -1116,6 +1154,7 @@ func getProbeHandler(feastType FeastServiceType, tls *feastdevv1alpha1.TlsConfig
 		}
 		return probeHandler
 	}
+	targetPort := getTargetPort(feastType, tls)
```
tchughesiv:
```suggestion
```

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -1137,3 +1176,15 @@ func IsDeploymentAvailable(conditions []appsv1.DeploymentCondition) bool {
...
+func (feast *FeastServices) isRegistryGrpcEnabled() bool {
+	registry := feast.Handler.FeatureStore.Status.Applied.Services.Registry
+	return registry.Local.Server.GRPC == nil || *registry.Local.Server.GRPC
+}
+
+// isRegistryRestEnabled checks if REST API is enabled for registry service
+func (feast *FeastServices) isRegistryRestEnabled() bool {
+	registry := feast.Handler.FeatureStore.Status.Applied.Services.Registry
+	return registry.Local.Server.RestAPI != nil && *registry.Local.Server.RestAPI
+}
```
tchughesiv: the check should be all-encompassing so it can be run from anywhere -
```suggestion
// isRegistryRestEnabled checks if REST API is enabled for registry service
func (feast *FeastServices) isRegistryRestEnabled() bool {
    if feast.isRegistryServer() {
	    registry := feast.Handler.FeatureStore.Status.Applied.Services.Registry
            return registry.Local.Server.RestAPI != nil && *registry.Local.Server.RestAPI
    }
    return false
}
```

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -1137,3 +1176,15 @@ func IsDeploymentAvailable(conditions []appsv1.DeploymentCondition) bool {
 func (feast *FeastServices) GetFeastRestServiceName(feastType FeastServiceType) string {
 	return feast.GetFeastServiceName(feastType) + "-rest"
 }
+
+// isRegistryGrpcEnabled checks if gRPC is enabled for registry service
+func (feast *FeastServices) isRegistryGrpcEnabled() bool {
+	registry := feast.Handler.FeatureStore.Status.Applied.Services.Registry
+	return registry.Local.Server.GRPC == nil || *registry.Local.Server.GRPC
+}
```
tchughesiv: the check should be all-encompassing so it can be run from anywhere.
also, since GRPC is being set by default, we should not return true on `== nil`... `nil` in this case implies `false`, which would further confuse users.
```suggestion
func (feast *FeastServices) isRegistryGrpcEnabled() bool {
    if feast.isRegistryServer() {
	    registry := feast.Handler.FeatureStore.Status.Applied.Services.Registry
	    return registry.Local.Server.GRPC != nil && *registry.Local.Server.GRPC
    }
    return false    
}
```

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -420,24 +442,28 @@ func (feast *FeastServices) setContainer(containers *[]corev1.Container, feastTy
...
-		probeHandler := getProbeHandler(feastType, tls)
-		container.Ports = []corev1.ContainerPort{
-			{
+		probeHandler := feast.getProbeHandler(feastType, tls)
+		container.Ports = []corev1.ContainerPort{}
+
+		isRegistry := feastType == RegistryFeastType && feast.isRegistryServer()
+
+		grpcEnabled := !feast.isRegistryServer() || feast.isRegistryGrpcEnabled()
+		if grpcEnabled {
```
tchughesiv: why would grpc be enabled if the registry server wasn't enabled?
```suggestion
		if feast.isRegistryGrpcEnabled() {
```

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -420,24 +442,28 @@ func (feast *FeastServices) setContainer(containers *[]corev1.Container, feastTy
...
-			registry := feast.Handler.FeatureStore.Status.Applied.Services.Registry
-			if registry.Local.Server.RestAPI != nil && *registry.Local.Server.RestAPI {
-				container.Ports = append(container.Ports, corev1.ContainerPort{
-					Name:          name + "-rest",
-					ContainerPort: getTargetRestPort(feastType, tls),
-					Protocol:      corev1.ProtocolTCP,
-				})
-			}
+
+		if isRegistry && feast.isRegistryRestEnabled() {
```
tchughesiv:
```suggestion
		if feastType == RegistryFeastType && feast.isRegistryRestEnabled() {
```

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -54,6 +54,11 @@ func (feast *FeastServices) Deploy() error {
 	if feast.noLocalCoreServerConfigured() {
 		return errors.New("at least one local server must be configured. e.g. registry / online / offline")
 	}
+	if feast.isRegistryServer() {
+		if !feast.isRegistryGrpcEnabled() && !feast.isRegistryRestEnabled() {
+			return errors.New("at least one of gRPC or REST API must be enabled for registry service")
+		}
```
tchughesiv, line 60: this is a good check, but it should **also** be handled w/ CRD validation so (hopefully) the operator never has to handle this at all.
https://book.kubebuilder.io/reference/markers/crd-validation

Maybe some kind of `+kubebuilder:validation:XValidation:rule` that they are never both set to false in the spec? in this situation it shouldn't be a check that at least 1 is true, because we set GRPC by default for the user... so the user could leave them both blank if they wanted. we need to check for a scenario where the user explicitly sets both to `false`.

this crd validation would be added here -
https://github.com/feast-dev/feast/blob/f231f0d30d2da2012034db5ca3a0068484a4d143/infra/feast-operator/api/v1alpha1/featurestore_types.go#L505-L506

examples of using this kind of validation can be seen throughout the same file... e.g. -
https://github.com/feast-dev/feast/blob/f231f0d30d2da2012034db5ca3a0068484a4d143/infra/feast-operator/api/v1alpha1/featurestore_types.go#L353-L360

tchughesiv (changes requested):

infra/feast-operator/internal/controller/services/services.go
tchughesiv: this new error message should also be checked w/ a unit test to ensure it's being set in the registry  status condition message

infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -502,6 +502,17 @@ type ServerConfigs struct {
 	VolumeMounts []corev1.VolumeMount `json:"volumeMounts,omitempty"`
 }
 
+// RegistryServerConfigs creates a registry server for the feast service, with specified container configurations.
+type RegistryServerConfigs struct {
+	ServerConfigs `json:",inline"`
+
+	// Enable REST API registry server. Defaults to false if unset.
```
tchughesiv:
```suggestion
	// Enable REST API registry server.
```

tchughesiv (changes requested):
this whole block needs to change -
https://github.com/feast-dev/feast/blob/f231f0d30d2da2012034db5ca3a0068484a4d143/infra/feast-operator/internal/controller/services/services.go#L447-L465

the issue is that you're breaking the setting of ports for other service types (online, offline, etc)...
something like this should do the trick -
```go
	if feastType == RegistryFeastType {
		if feast.isRegistryGrpcEnabled() {
			container.Ports = append(container.Ports, corev1.ContainerPort{
			    	Name:          name,
			    	ContainerPort: getTargetPort(feastType, tls),
			    	Protocol:      corev1.ProtocolTCP,
			})
		    }
		if feast.isRegistryRestEnabled() {
		    	container.Ports = append(container.Ports, corev1.ContainerPort{
		    		Name:          name + "-rest",
		    		ContainerPort: getTargetRestPort(feastType, tls),
		    		Protocol:      corev1.ProtocolTCP,
		    	})
		    }
        } else {
            container.Ports = append(container.Ports, corev1.ContainerPort{
		    	Name:          name,
		    	ContainerPort: getTargetPort(feastType, tls),
		    	Protocol:      corev1.ProtocolTCP,
            })
        }
```
the above would be done in addition to the other suggestions i've made around the `isRegistryRestEnabled` & `isRegistryGrpcEnabled` method changes

infra/feast-operator/internal/controller/services/services.go
ntkathole: right, targetPort can be defined once. I used `else if` since `getProbeHandler` is called once per container setup and we know only one condition will be true at a time.

sdk/python/feast/cli/serve.py
ntkathole: I found the issue with incorrect port number used when only rest is enabled. Earlier it was using default grpc port instead of default rest port. Updated the description. Thanks.

infra/feast-operator/internal/controller/services/services.go
tchughesiv: i see now you use `getTargetRestPort` for rest, which is a different method. i guess it will need to be called separately.

infra/feast-operator/api/v1alpha1/featurestore_types.go
```diff
@@ -503,10 +503,11 @@ type ServerConfigs struct {
 }
 
 // RegistryServerConfigs creates a registry server for the feast service, with specified container configurations.
+// +kubebuilder:validation:XValidation:rule="self.restAPI == true || self.grpc == true || !has(self.grpc)", message="At least one of restAPI or grpc must be true"
```
tchughesiv, line 506: in this situation it shouldn't be a check that at least 1 is true, because we set GRPC by default for the user... so the user _could_ leave them both blank if they wanted.

here, we **must** to check/disallow a scenario where the user explicitly sets both to `false`.

tchughesiv (changes requested):
looking good! one more change requested.

we also need unit tests for changes made around the rest & grpc container ports, services, probeHandlers, etc.

infra/feast-operator/api/v1alpha1/featurestore_types.go
tchughesiv: something like this _should_ work -
```go
// +kubebuilder:validation:XValidation:rule="!(self.grpc == false && (!has(self.restAPI) || self.restAPI == false))", message="Either restAPI must be true or grpc must not be false"
```

Passes these scenarios -

1. restAPI is true
2. Both true
3. At least one is true
4. grpc is true, restAPI missing is OK
5. Both missing
7. grpc is missing, restAPI is false

Fails these scenarios -

- Both are explicitly false
- grpc is false and restAPI is missing


</body></html>

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -607,14 +691,33 @@ func (feast *FeastServices) setService(svc *corev1.Service, feastType FeastServi
...
 		},
 	}
 
 	return controllerutil.SetControllerReference(feast.Handler.FeatureStore, svc, feast.Handler.Scheme)
 }
 
+// createRestService creates a separate service for the Registry REST API
+func (feast *FeastServices) createRestService(feastType FeastServiceType) error {
+	if feast.isRegistryServer() {
+		if !feast.isRegistryRestEnabled() {
```
tchughesiv, line 705: redundant check
```suggestion
	if !feast.isRegistryRestEnabled() {
```

infra/feast-operator/internal/controller/services/services.go
```diff
@@ -54,6 +54,11 @@ func (feast *FeastServices) Deploy() error {
 	if feast.noLocalCoreServerConfigured() {
 		return errors.New("at least one local server must be configured. e.g. registry / online / offline")
 	}
+	if feast.isRegistryServer() {
+		if !feast.isRegistryGrpcEnabled() && !feast.isRegistryRestEnabled() {
```
tchughesiv, line 58:
```suggestion
	if !feast.isRegistryGrpcEnabled() && !feast.isRegistryRestEnabled() {
```

tchughesiv: i'm wrong... ignore this

infra/feast-operator/test/api/featurestore_types_test.go
```diff
@@ -336,6 +340,90 @@ func registryStoreWithDBPersistenceType(dbPersistenceType string, featureStore *
 	return fsCopy
 }
 
+func registryWithOnlyRestAPI(featureStore *feastdevv1alpha1.FeatureStore) *feastdevv1alpha1.FeatureStore {
```
tchughesiv:
```suggestion
func registryWithRestAPIFalse(featureStore *feastdevv1alpha1.FeatureStore) *feastdevv1alpha1.FeatureStore {
```

tchughesiv: may want to add a `registryWithOnlyRestAPI` separate check w/ rest set to `true` though 
```go
                                 Server: &feastdevv1alpha1.RegistryServerConfigs{
					RestAPI: boolPtr(true),
				},
```

Issue #3146, opened 2022-08-26
error: "Snowflake Data Type is not supported" on "NUMBER" with non-zero scale


cburroughs:
## Expected Behavior 

feast plan/apply completed without error in version `0.23.0` and I was able to query with `get_historical_features`.

## Current Behavior
With the snowflake datatype ` NUMBER(38,6)` I get:
```
NotImplementedError: The following Snowflake Data Type is not supported: DECIMAL -- Convert to DOUBLE
```

## Steps to reproduce

 * `$ feast init -t snowflake`
 * Alter the table like so:
 ```
alter table SNOWY_FEAST.PUBLIC."amazed_dane_feast_driver_hourly_stats" add column temp_conv_rate NUMBER(38,6);
update SNOWY_FEAST.PUBLIC."amazed_dane_feast_driver_hourly_stats" set temp_conv_rate = "conv_rate";
alter table SNOWY_FEAST.PUBLIC."amazed_dane_feast_driver_hourly_stats" drop column "conv_rate";
alter table SNOWY_FEAST.PUBLIC."amazed_dane_feast_driver_hourly_stats" rename column temp_conv_rate to "conv_rate";
 ```
 * `$ feast apply`
 ```
Traceback (most recent call last):
...
    raise NotImplementedError(
NotImplementedError: The following Snowflake Data Type is not supported: DECIMAL -- Convert to DOUBLE
```
 

### Specifications

- Version: `Feast SDK Version: "feast 0.24.0"`
- Platform: x86_64 on Python 3.10.5
- Subsystem: Linux 5.4.203-

achals:
I think this is intentional cc @sfc-gh-madkins

sfc-gh-madkins:
@cburroughs my advice for right now would be to cast this to type double/float, as the error recommends. Let me see if I can get NUMBER with scale to work.

cburroughs:
I installed from `master` to test it out.  Looks good; thank you!

PR #5367, opened 2025-05-19
feat: Add Decimal to allowed python scalar types

nanohanno:

When exporting features to parquet files (via `to_remote_storage()`) columns of type `NUMBER` are converted to the pyarrow `decimal128` type. This type is translated into `decimal.Decimal` in python when trying to convert it into proto values (https://github.com/feast-dev/feast/blob/648c53dc64ad88077e49e56f0cfc70756d6a5824/sdk/python/feast/utils.py#L282).

Eventually, it fails when the conversion into a proto value is done because `decimal.Decimal` is not supported as a valid type in https://github.com/feast-dev/feast/blob/648c53dc64ad88077e49e56f0cfc70756d6a5824/sdk/python/feast/type_map.py#L313. As far as I could see there was nothing immediately against adding `Decimal` to the supported list as it can be converted into Double. I might be overlooking something however.

# Misc

I stumbled upon this issue when using the Lambda materialization engine.

Slightly related to https://github.com/feast-dev/feast/issues/3146

### DIFF
--- a/sdk/python/feast/type_map.py
+++ b/sdk/python/feast/type_map.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import decimal
 import json
 import logging
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


franciscojavierarceo (approved):
lgtm

PR #5371, opened 2025-05-20
Eapc 17414/null sort keys materialization

kpulipati29:
[test-logs.log](https://github.com/user-attachments/files/20328338/test-logs.log)

Currently, the batch materialization spark job pretends like its successful even though there were errors during the materialization (for example, the materialization fails if the value of any primary key column is null) and the materialization was not successful in reality. This is because, the errors are never propagated from online_write_batch method. In this PR, the error is re-thrown in online_write_batch which ultimately fails the job.

https://expediagroup.atlassian.net/browse/EAPC-17414

# Test
Test logs are attached where I tried to materialize features that has null values for sort key columns and as seen in the logs, the job was failed and the error is logged there.

### DIFF
--- a/sdk/python/feast/infra/online_stores/contrib/cassandra_online_store/cassandra_online_store.py
+++ b/sdk/python/feast/infra/online_stores/contrib/cassandra_online_store/cassandra_online_store.py
@@ -398,16 +398,13 @@ class CassandraOnlineStore(OnlineStore):
                       rows is written to the online store. Can be used to
                       display progress.
         """
-        is_error: bool = False
-        ex: Exception = Exception("Exception raised while writing a batch")
+        ex: Optional[Exception] = None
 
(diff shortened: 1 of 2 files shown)


franciscojavierarceo:
hi @kpulipati29 I think you may have included a bit too many commits in this PR.

franciscojavierarceo:
@kpulipati29 did you mean to include all of these changes in your PR? Or did you mean to submit this to Expedia's fork?

kpulipati29:
> @kpulipati29 did you mean to include all of these changes in your PR? Or did you mean to submit this to Expedia's fork?

Sorry for the late reply, yes I meant to submit this to Expedia's fork, I will close this PR as I don't need this anymore!

franciscojavierarceo:
Sounds good! Do let us know if you would like to reconcile more items upstream, would be happy to help there.

PR #5377, opened 2025-05-21
feat: Add max_file_size to Snowflake config

nanohanno:

When using the Snowflake storage integration to export features as parquet files on S3 it can be useful to set a maximum file size. When working with the Lambda materialization engine we faced the issue that the number of rows per file and thus per invoked Lambda function was so large that the Lambda functions timed out. 
With the new `max_file_size` parameter the file size can be reduced so that each Lambda function needs to write less rows to the online store.

# Misc

Snowflake docs https://docs.snowflake.com/en/sql-reference/sql/copy-into-location#copy-options-copyoptions


franciscojavierarceo (approved):
lgtm

PR #5379, opened 2025-05-21
feat: Improvements to Lambda materialization engine

nanohanno:

While working with the Lambda materialization engine we had to make a couple of changes to make it work for us.
1. The read timeout for the Lambda client is increased to the Lambda timeout + 10 s. The default value is 60 s which caused failures when Lambda functions were running longer than the read timeout.
2. The case is handled when no feature updates and thus no files are exported for the given time range. Before, an exception was raised because `ThreadPoolExecutor` does not accept `max_workers=0`.
3. Functionality to retry invocation of Lambda functions is added in case that a function times out. Such timeouts were observed when new features are materialized initially with a large number of historical feature updates. DynamoDB, which we use as online store, cannot handle the write throughput that is generated by 20 concurrent Lambda functions. Therefore, Lambda functions may time out. When these functions are retried, they have the chance to finish successfully when the DynamoDB table is scaled up, which takes some 10 minutes.
4. If an exception is raised in a Lambda function or not all invocations finish the `LambdaMaterializationJob` is unsuccessful and an exception is raised. This behaviour is helpful when running it in automation where failed materialization runs are more visible.

# Misc

### DIFF
--- a/sdk/python/feast/infra/materialization/aws_lambda/lambda_engine.py
+++ b/sdk/python/feast/infra/materialization/aws_lambda/lambda_engine.py
@@ -55,11 +55,16 @@ class LambdaMaterializationEngineConfig(FeastConfigBaseModel):
 
 @dataclass
 class LambdaMaterializationJob(MaterializationJob):
-    def __init__(self, job_id: str, status: MaterializationJobStatus) -> None:
+    def __init__(
+        self,
+        job_id: str,
+        status: MaterializationJobStatus,
+        error: Optional[BaseException] = None,
+    ) -> None:
         super().__init__()
         self._job_id: str = job_id
         self._status = status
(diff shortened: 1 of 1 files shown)


franciscojavierarceo (commented):
can you add a unit test? @HaoXuAI is making some changes and want to make sure we don't break anything.

nanohanno:
> can you add a unit test? @HaoXuAI is making some changes and want to make sure we don't break anything.

Sure, tests would be good. I only see the skipped integration test so far https://github.com/feast-dev/feast/blob/ad69eb58048189c4497e7f44718e18de8fbb6110/sdk/python/tests/integration/materialization/test_lambda.py#L24
I can take another look the next days to get it running or work out a unit test with mocking.

PR #5382, opened 2025-05-22
fix: Update milvus connect function to work with remote instance

Fiona-Waters:

This PR allows for a local remote instance of Milvus to connect, as previously it was bypassed by `provider: local` which remains and is correct but not the only consideration when using local milvus. The PR also updates the variable from `url` to `uri` as expected by the MilvusClient.

N/A — this issue was found independently and not yet tracked.

# Misc


franciscojavierarceo (approved):

Fiona-Waters:
> /hold

/unhold !

jyejare (approved):

PR #5386, opened 2025-05-25
feat: SQLite online retrieval. Add timezone info into timestamp.

wckdman:

In all other places of the codebase we deal with timestamps in the `utc` timezone.
During experiments with online retrieval from SQLite I found that there're timestamps without tz-info.
I want to add it to be consistent.

N/A

# Misc

N/A


franciscojavierarceo:
Looks like integration test failed

PR #5387, opened 2025-05-25
feat: Snowflake source. fetch MAX in a single query

wckdman:

Default precision in Snowflake is 38 ([Docs](https://docs.snowflake.com/en/sql-reference/data-types-numeric#number)).
Current realisation executes a bunch of `MAX(column)` queries to check the exact precision.

This PR introduces a small optimisation to execute a single query for all columns.

N/A

# Misc

N/A


HaoXuAI (approved):
LGTM!

PR #5403, opened 2025-05-30
Droid/redis performance optimizations

darpanKumar29:

# Misc


all protos/Aggregation.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
```
haoxu0, line 1: Do you need those protos? we have protos in the project now:
https://github.com/feast-dev/feast/blob/master/protos/feast/core/Aggregation.proto

franciscojavierarceo:
@darpanKumar29 did you want to fix this PR?

PR #5420, opened 2025-06-04
feat: Allow to use env variable FEAST_FS_YAML_FILE_PATH and FEATURE_REPO_DIR

### LINKED ISSUE #5416: Add env variable support pointing to `feature_store.yaml`

I'm building a solution that uses Feast under the hood, and I want to abstract away infrastructure details from end users — specifically, the location of the `feature_store.yaml` file. Currently, Feast requires the path to be passed via `-f` for every CLI command, or the file must exist in the current working directory.

This tightly couples user workflows to the project structure and forces them to think about internal config paths. I’d prefer to centralize this configuration and avoid exposing it to end users.

I would like Feast to support a new environment variable, e.g., `FEAST_FEATURE_STORE_YAML_PATH`, that allows specifying the path to the `feature_store.yaml` file. If this env var is set, Feast should use it as the default location for the config file — so users don’t have to pass `-f` every time or change directories.

This would allow me to provide a smoother, zero-config CLI experience to users, while still maintaining flexibility and modularity internally.

- Using the `-f` flag repeatedly — functional, but leaks internal file structure to the user and adds CLI overhead.
- Changing the working directory to match the config location — fragile in some execution environments.

**Additional context**

Here’s an example of what this could look like in practice:

```bash
export FEAST_FEATURE_STORE_YAML_PATH=/my/app/internal/feature_store.yaml
feast apply    # No -f needed
feast ui       # Still picks up the right config
```

ntkathole:
Allow user to set chdir and feature-store-yaml cli options via env variables.

Fixes #5416

### DIFF
--- a/sdk/python/feast/cli/cli.py
+++ b/sdk/python/feast/cli/cli.py
@@ -47,6 +47,7 @@ from feast.cli.serve import (
 from feast.cli.stream_feature_views import stream_feature_views_cmd
 from feast.cli.ui import ui
 from feast.cli.validation_references import validation_references_cmd
+from feast.constants import FEAST_FS_YAML_FILE_PATH_ENV_NAME
 from feast.errors import FeastProviderLoginError
 from feast.repo_config import load_repo_config
 from feast.repo_operations import (
@@ -81,7 +82,8 @@ class NoOptionDefaultFormat(click.Command):
 @click.option(
     "--chdir",
     "-c",
-    help="Switch to a different feature repository directory before executing the given subcommand.",
+    envvar="FEATURE_REPO_DIR_ENV_VAR",
+    help="Switch to a different feature repository directory before executing the given subcommand. Can also be set via the FEATURE_REPO_DIR_ENV_VAR environment variable.",
 )
 @click.option(
     "--log-level",
@@ -91,7 +93,7 @@ class NoOptionDefaultFormat(click.Command):
 @click.option(
     "--feature-store-yaml",
     "-f",
-    help="Override the directory where the CLI should look for the feature_store.yaml file.",
+    help=f"Override the directory where the CLI should look for the feature_store.yaml file. Can also be set via the {FEAST_FS_YAML_FILE_PATH_ENV_NAME} environment variable.",
 )
 @click.pass_context
 def cli(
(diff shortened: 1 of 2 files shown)


sdk/python/feast/cli/cli.py
```diff
@@ -81,7 +82,8 @@ def format_options(self, ctx: click.Context, formatter: click.HelpFormatter):
 @click.option(
     "--chdir",
     "-c",
-    help="Switch to a different feature repository directory before executing the given subcommand.",
+    envvar="FEATURE_REPO_DIR",
```
jyejare:
```suggestion
    envvar="FEATURE_REPO_DIR_ENV_VAR",

```

sdk/python/feast/cli/cli.py
```diff
@@ -91,7 +93,8 @@ def format_options(self, ctx: click.Context, formatter: click.HelpFormatter):
 @click.option(
     "--feature-store-yaml",
     "-f",
-    help="Override the directory where the CLI should look for the feature_store.yaml file.",
+    envvar=FEAST_FS_YAML_FILE_PATH_ENV_NAME,
```
jyejare:
```suggestion
    envvar=FS_YAML_FILE_PATH_ENV_VAR,

```
FEAST is implcit here.

jyejare (commented):
Some nitpick comments.

sdk/python/feast/cli/cli.py
ntkathole: @jyejare This is constant value defined https://github.com/feast-dev/feast/blob/master/sdk/python/feast/constants.py#L24

ntkathole: I simply removed it from latest commit.

jyejare: Interesting, How setting up that as EnvVar would be recognized then ?

ntkathole: https://github.com/feast-dev/feast/blob/master/sdk/python/feast/utils.py#L96

jyejare (approved):

sdk/python/feast/cli/cli.py
franciscojavierarceo: oh interesting, i guess we'll follow the FEAST prefix. feels bad 🤷

sdk/python/tests/unit/transformation/test_spark_transformation.py
```diff
@@ -53,7 +53,10 @@ def remove_extra_spaces_sql(df, column_name):
 @pytest.fixture
 def spark_fixture():
     spark = SparkSession.builder.appName("Testing PySpark Example").getOrCreate()
-    yield spark
+    try:
+        yield spark
+    finally:
+        spark.stop()
```
ntkathole, line 59: This helps to cleanup spark orphan process

franciscojavierarceo (approved):

Issue #5205, opened 2025-04-01
Implement online_write_batch for remote


itay1551:
I trying to `push` to an online store 
```
online_store:
  ...
  type: remote
  ...
  vector_enabled: true
  vector_len: 64
```

And I getting `NotImplementedError`

at `/opt/app-root/lib64/python3.11/site-packages/feast/infra/online_stores/remote.py:63`

Thanks!

Issue #5412, opened 2025-06-02
None of the objects returned for allow_registry_cache (Postgres online store, vector_enabled)


jyejare:
## Description

When using the Postgres online store with `vector_enabled`, none of the objects are returned for `allow_registry_cache`. This issue is reproducible and may be due to either:
- The caching mechanism itself being broken
- Retrieval of cached registry items failing

## Steps to Reproduce
1. Use Postgres online store with `vector_enabled`.
2. Attempt to retrieve objects with `allow_registry_cache`.
3. Observe that no objects are returned.

## Expected Behavior
Objects should be returned from the cache if present.

## Actual Behavior
None of the objects are returned.

## Additional Notes
- This could be an issue with the caching logic or the retrieval of cached registry items.
- Needs investigation to determine the root cause.

---
Please investigate and provide a fix or further guidance.

jyejare:
### End-to-End Example to Reproduce the Issue

Below is a minimal example to reproduce the issue with `allow_registry_cache` returning no objects when using the Postgres online store with `vector_enabled`:

```python
from feast import FeatureStore, FeatureView, Entity, Field, FileSource
from feast.types import Float32
from feast.infra.online_stores.postgres import PostgresOnlineStoreConfig

# Define your Postgres config (update with your credentials)
postgres_config = PostgresOnlineStoreConfig(
    host="localhost",
    port=5432,
    database="feast",
    user="feast",
    password="feast",
    vector_enabled=True,
)

# Define a simple entity and feature view
entity = Entity(name="id", join_keys=["id"])
source = FileSource(path="data.csv", event_timestamp_column="event_timestamp")
feature_view = FeatureView(
    name="example_fv",
    entities=["id"],
    ttl=None,
    schema=[Field(name="value", dtype=Float32)],
    online=True,
    source=source,
)

# Apply objects to the registry
store = FeatureStore(repo_path=".")
store.apply([entity, feature_view])

# Try to retrieve objects with allow_registry_cache
objects = store._get_registry().list_feature_views(allow_registry_cache=True)
print("Returned objects:", objects)
```

**Expected:** Objects should be returned from the cache if present.

**Actual:** No objects are returned.

---
- Ensure you have a running Postgres instance and the correct credentials.
- The above code assumes you have a `data.csv` file with appropriate columns.
- This demonstrates the issue for the vector-enabled Postgres online store.

PR #5422, opened 2025-06-04
feat: Remote Write to Online Store completes client / server architecture

jyejare:

Opens and completes the full cycle of remote (Client/Server) architecture by :

- [x] Allows the write features to online store remotely from the client to a feature server
- [x] This takes care of push, materialize and materialize-incremental
- [x] Test is added to validate the write and then read the written features to the store

Closes https://github.com/feast-dev/feast/issues/5205 

# Misc

### DIFF
--- a/sdk/python/feast/infra/online_stores/remote.py
+++ b/sdk/python/feast/infra/online_stores/remote.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 import json
 import logging
+from collections import defaultdict
 from datetime import datetime
 from typing import Any, Callable, Dict, List, Literal, Optional, Sequence, Tuple
 
@@ -20,12 +21,16 @@ import requests
 from pydantic import StrictStr
 
 from feast import Entity, FeatureView, RepoConfig
+from feast.infra.online_stores.helpers import _to_naive_utc
 from feast.infra.online_stores.online_store import OnlineStore
 from feast.protos.feast.types.EntityKey_pb2 import EntityKey as EntityKeyProto
 from feast.protos.feast.types.Value_pb2 import Value as ValueProto
 from feast.repo_config import FeastConfigBaseModel
 from feast.rest_error_handler import rest_error_handling_decorator
-from feast.type_map import python_values_to_proto_values
+from feast.type_map import (
+    feast_value_type_to_python_type,
+    python_values_to_proto_values,
+)
 from feast.value_type import ValueType
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -60,7 +60,77 @@ def online_write_batch(
         ],
         progress: Optional[Callable[[int], Any]],
     ) -> None:
-        raise NotImplementedError
+        """
+        Writes a batch of feature rows to the remote online store via the remote API.
+        """
+        assert isinstance(config.online_store, RemoteOnlineStoreConfig)
+        config.online_store.__class__ = RemoteOnlineStoreConfig
```
ntkathole, line 72: is this line necessary ? we are already asserting isinstance in above line.

sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -60,7 +60,77 @@ def online_write_batch(
...
+
+        # Restructure data into a columnar dictionary format for the 'df' key
+        columnar_data: Dict[str, List[Any]] = {}
```
ntkathole: defaultdict(list) can be used here

sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -60,7 +60,77 @@ def online_write_batch(
...
+
+        # Collect all unique entity key names and feature names
+        all_keys = set()
```
ntkathole: I think all_keys collection is not needed, you can simply populate in the below loop once defaultdict is initialized above.

sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -60,7 +60,77 @@ def online_write_batch(
...
+        }
+
+        response = post_remote_online_write(config=config, req_body=req_body)
+
+        if response.status_code != 200:
+            error_msg = f"Unable to write online store data using feature server API. Error_code={response.status_code}, error_message={response.text}"
+            logger.error(error_msg)
+            raise RuntimeError(error_msg)
+
+        if progress:
```
ntkathole, line 111: Probably good to add info logging like writing len(data) rows to remote store for FeatureView table.name

sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -60,7 +60,77 @@ def online_write_batch(
...
+                )
+
+            # Populate feature values
+            for feature_name, feature_value_proto in feature_values_proto.items():
+                columnar_data[feature_name].append(
+                    self.value_proto_to_python(feature_value_proto)
+                )
+
+            # Populate timestamps
+            columnar_data["event_timestamp"].append(event_ts.isoformat())
```
ntkathole: tz=timezone.utc ? We have `_to_naive_utc` helper

sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -60,7 +60,77 @@ def online_write_batch(
...
+            if "created" in columnar_data:
+                columnar_data["created"].append(
+                    created_ts.isoformat() if created_ts else None
+                )
+
+        req_body = {
+            "feature_view_name": table.name,
+            "df": columnar_data,
```
ntkathole, line 100: Does this request fail for empty data?

sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -60,7 +60,77 @@ def online_write_batch(
...
+
+        if response.status_code != 200:
+            error_msg = f"Unable to write online store data using feature server API. Error_code={response.status_code}, error_message={response.text}"
+            logger.error(error_msg)
+            raise RuntimeError(error_msg)
+
+        if progress:
+            progress(len(data))
+
+    def value_proto_to_python(self, value_proto: ValueProto):
```
ntkathole: can we use any existing helper instead? may be from https://github.com/feast-dev/feast/blob/master/sdk/python/feast/type_map.py

sdk/python/feast/infra/online_stores/remote.py
jyejare: Yes necessary for type safety.

jyejare:
Looks like too much logging in the integration testing job made it to cancel the job. But the tests for `sdk/python/tests/integration/online_store/test_remote_online_store.py::test_remote_online_store_read_write` are passing. 👍

sdk/python/feast/infra/online_stores/remote.py
jyejare: For now, its not failing not only for remote store but for any. I think it's good to identify the empty data at the higher level than at the store class level.

That can be taken care in a different PR btw.

jyejare: I opened https://github.com/feast-dev/feast/issues/5428

sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -60,7 +65,55 @@ def online_write_batch(
...
+            # Populate feature values
+            for feature_name, feature_value_proto in feature_values_proto.items():
+                columnar_data[feature_name].append(
+                    feast_value_type_to_python_type(feature_value_proto)
+                )
+
+            # Populate timestamps
+            columnar_data["event_timestamp"].append(_to_naive_utc(event_ts).isoformat())
+            columnar_data["created"].append(
+                _to_naive_utc(event_ts).isoformat() if created_ts else None
```
ntkathole: should this be created_ts instead of event_ts here?

sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -60,7 +65,55 @@ def online_write_batch(
...
+            )
+
+        req_body = {
+            "feature_view_name": table.name,
+            "df": columnar_data,
+            "allow_registry_cache": False,
```
ntkathole, line 101: allow_registry_cache needs to be always False/hardcoded?

jyejare: Actually there is an issue raised - https://github.com/feast-dev/feast/issues/5412 but its not always reproducible. So to be on the safe side I choosed to be cache free.

franciscojavierarceo:
@jyejare in a follow up PR can you add that this is supported in the documentation?

Issue #5074, opened 2025-02-20
Support for `numpy 2.0`?


kjod:
Is there a plan to support numpy 2.0 sometime in the near future?

franciscojavierarceo:
Yup! Thanks for raising!

TomSteenbergen:
Hi @franciscojavierarceo! Is there any update on `numpy` 2 support?

tchughesiv:
Just a comment... if I remember correctly, it's my understanding that python 3.9 support would need to be dropped in order to do this.

franciscojavierarceo:
Yes,  and I think that's something we should do this year.

CC @ntkathole @jyejare

TomSteenbergen:
Currently `feast` is blocking our ML applications from migrating to `numpy` 2. Not sure how ubiquitous `numpy` usage is in `feast`'s source code, but could we consider supporting both major versions of `numpy` in `feast`? So that the Python 3.9 deprecation is not blocking upgrading `numpy` here? At first glance, the amount of breaking changes in `numpy` seems limited: https://numpy.org/devdocs/numpy_2_0_migration_guide.html

franciscojavierarceo:
It's been on my todo to deprecate and add 3.12 for this exact reason, so happy to do it and just support one numpy version.

franciscojavierarceo:
PR to support 3.12 here btw: https://github.com/feast-dev/feast/pull/5276

tchughesiv:
@franciscojavierarceo nice! do you think we can cut another release before merging the numpy/python changes?

HaoXuAI:
@franciscojavierarceo you can cut the release sooner, maybe on the weekend. I'll see if I can merge one or two PRs before that, otherwise I can cut a patch release if needed

franciscojavierarceo:
I think we have time, just wanted to know when you want to release things is all.

HaoXuAI:
> I think we have time, just wanted to know when you want to release things is all.

cool, I think one week should be enough for me to get a checkpoint for compute engines

franciscojavierarceo:
Latest release is out, next one will drop 3.9

franciscojavierarceo:
https://github.com/feast-dev/feast/pull/5375

PR #5427, opened 2025-06-06
chore: Update numpy

ntkathole:

This PR adds support for numpy>2.

**Key points:**
- 2.2.2 version of PyTorch works only with NumPy 1.x. So, it is important to **upgrade torch>2.2.2**.
- Pytorch macos x86 builds are deprecated, so we have to **remove macos-13 and use macos-14** instead in our CI.
- PyPI has CUDA-enabled PyTorch wheels as the primary distribution. Since our linux github runners need **cpu-only build**, we are installing torch from index https://download.pytorch.org/whl/cpu.
- pyspark needs to be updated as 'np.NaN` was removed in the NumPy 2.0 release.

https://github.com/feast-dev/feast/issues/5074 

# Misc

- MacOS x86_64 builds are no longer supported by official PyTorch wheels.

If you're using an Intel-based Mac (not Apple Silicon) or running Python with Rosetta-translated x86_64 architecture:

You cannot install PyTorch versions 2.3.0+ via pip, as no compatible wheels are published. The last supported version is: torch==2.2.2.

To upgrade to newer versions like torch==2.7.0, you must use macOS 14+ on Apple Silicon (arm64) and run a matching arm64 Python environment. 

To check your architecture:
`python3 -c "import platform; print(platform.machine())"`
If the output is x86_64, you’ll be limited to PyTorch ≤ 2.2.2.
If it’s arm64, you can install the latest PyTorch.


Makefile
```diff
@@ -152,7 +152,7 @@ test-python-integration-local: ## Run Python integration tests (local dev mode)
 test-python-integration-rbac-remote: ## Run Python remote RBAC integration tests
 	FEAST_IS_LOCAL_TEST=True \
 	FEAST_LOCAL_ONLINE_CONTAINER=True \
-	python -m pytest --tb=short -v -n 8 --color=yes --integration --durations=10 --timeout=1200 --timeout_method=thread --dist loadgroup \
+	python -m pytest --tb=short -v -n 4 --color=yes --integration --durations=10 --timeout=1200 --timeout_method=thread --dist loadgroup \
```
franciscojavierarceo, line 155: Why 4?

franciscojavierarceo (approved):
One nit

Makefile
ntkathole: Currently with 8, runner is intermittently throwing `Error: The operation was canceled.` (not related to this PR, happening on all PRs) due to high memory load on node. For stability, reduced to 4.

PR #5432, opened 2025-06-07
fix: Add missing conn.commit() in SnowflakeOnlineStore.online_write_batch

### LINKED ISSUE #5431: Bug: Snowflake online store writes are not committed
## Expected Behavior
Data written using Feast's `SnowflakeOnlineStore.online_write_batch` should be **persisted** in the Snowflake table and be visible in subsequent reads, even across different connections.
## Current Behavior
Data written by Feast is not persisted. It is sometimes visible in the same session (temporary reads succeed), but disappears after the connection is closed.
In Snowflake UI the table shows `0B`, and subsequent reads return empty results.
## Steps to Reproduce
1. Use a StreamFeatureView, which triggers the creation of a Table for that featureview.
2. Use Feast (with `SnowflakeOnlineStore`) and Spark-Kafka Processor to write features via `online_write_batch`.
3. Immediately read back data with `online_read` → values are present.
4. After the connection/session is closed:
* The table shows `0B` in Snowflake UI.
* Reads return no data (entity ids missing).
## Specifications
* Version: Feast `main` branch / latest (please update with your Feast version)
* Platform: Snowflake
* Subsystem: SnowflakeOnlineStore
## Possible Solution
The issue is caused by a **missing `conn.commit()`** in `online_write_batch`.
In `SnowflakeOnlineStore.online_write_batch`, Feast uses:
```python
with GetSnowflakeConnection(config.online_store, autocommit=False) as conn:
write_pandas_binary(...)
execute_snowflake_statement(conn, query) # INSERT OVERWRITE
# Missing: conn.commit()
```
Since `autocommit=False`, the transaction is not committed - and changes are discarded when the connection is closed.
**Fix:** Add an explicit `conn.commit()`:
```python
with GetSnowflakeConnection(config.online_store, autocommit=False) as conn:
write_pandas_binary(...)
execute_snowflake_statement(conn, query)
conn.commit() # required for persistence
```

## Why This Happens
* Tests arent failing as during the same session, uncommitted data is visible (why initial reads worked) => integreation tests with new connection would be helpful.
* Once the connection closes, uncommitted transaction is discarded (why the table is empty and reads fail later)

j-wine:
This PR fixes a critical bug in SnowflakeOnlineStore.online_write_batch.

Currently, the method sets autocommit=False but does not explicitly call conn.commit() after writing data and running the INSERT OVERWRITE query. As a result, data is not persisted - the transaction is discarded when the connection closes - even though no error is raised.

This causes:
Data appears in first reads (same connection)
Data disappears after session ends - table is empty in Snowflake UI and subsequent reads return nothing.

Adding conn.commit() ensures that the transaction is persisted correctly.

Fixes #5431

# Misc
This is a minimal one-line fix, but solves a critical data loss bug.
I will look into tadding a test which would fail for current implementation. Since this is transactional behavior, testing may require an end-to-end integration test.

### DIFF
--- a/sdk/python/feast/infra/online_stores/snowflake.py
+++ b/sdk/python/feast/infra/online_stores/snowflake.py
@@ -149,7 +149,7 @@ class SnowflakeOnlineStore(OnlineStore):
                             "_feast_row" = 1;
                 """
                 execute_snowflake_statement(conn, query)
-
+                conn.commit()
             if progress:
                 progress(len(data))
 


franciscojavierarceo (approved):
Thank you for this!

ntkathole (approved):

Issue #5430, opened 2025-06-07
Bug: On-Demand Feature Views (ODFV) on-write transformations not persisting


j-wine:
## Expected Behavior

On-write transformations in On-Demand Feature Views (ODFV) should be triggered when data is pushed to the online store using:

* `store.push()`
* `store.write_to_online_store()`
* `materialize()`

This worked correctly in Feast version 0.43.

## Current Behavior

Since Feast version 0.43, only the on-read transformations are triggered. On-write transformations do not update the online store as expected - the values remain `None`. The tested versions are  0.43,0.44,0.45,0.46,0.47,0.49 with a Python 3.10.3 venv, that contains only the dependencies of feast[redis,postgres].

## Steps to Reproduce

I have created a minimal repo demonstrating this issue:
 [https://github.com/j-wine/feast-odfv-user-test](https://github.com/j-wine/feast-odfv-user-test)

The repo uses a dockerized environment with:

* Python 3.10
* Feast 0.49

The test script performs:

1. `push()` to FeatureViews
2. `get_online_features()` - shows ODFV on-read works, but on-write values are `None`
3. `write_to_online_store()` - no change for on-write values
4. `materialize()` - still no change

### Example Output on Feast 0.49:

```
step 1: Push to basic FeatureViews

step 2: get_online_features post Push
product_id customer_id price_plus_revenue revenue_plus_one price_plus_rev
0 58 74 None None 50.0

step 3: write_to_online_store into basic feature view
res write_to_online:
product_id customer_id price_plus_revenue revenue_plus_one price_plus_rev
0 58 74 None None 50.0

Step 3: materialization
...
Step 4: get_online_features post materialization
product_id customer_id price_plus_revenue revenue_plus_one price_plus_rev
0 58 74 None None 50.0

Comparison Write-Time vs. Read-Time ODFV
price_plus_revenue: Missing (value: None)
revenue_plus_one: Missing (value: None)
```

### Behavior in Feast 0.43:

Running the same test script with Feast 0.43 produces correct results - on-write transformations populate values properly. (Except for an old bug when using more than one entity in a fsv)

## Specifications

* Version: 0.49 
* Platform: Docker, Python 3.10.13, Ubuntu LTS / Ubuntu WSL
* Subsystem: ODFV on-write transformations

## Possible Solution

Suspected cause (based on discussion with Francisco and my debugging):

* Integration tests might not fully cover this case.
* Potential gap in unit tests for on-write transformations.

## Additional Context
[...]

PR #5436, opened 2025-06-09
feat: Add materialization to Transformation On Writes

franciscojavierarceo:

https://github.com/feast-dev/feast/issues/5430

# Misc


sdk/python/feast/feature_store.py
```diff
@@ -656,7 +656,7 @@ def _make_inferences(
     def _get_feature_views_to_materialize(
         self,
         feature_views: Optional[List[str]],
-    ) -> List[FeatureView]:
+    ) -> List[Union[FeatureView, OnDemandFeatureView]]:
```
ntkathole, line 659: also StreamFeatureView ?

sdk/python/tests/unit/test_on_demand_python_transformation.py
```diff
@@ -1117,6 +1117,163 @@ def python_stored_writes_feature_view(
...
             }
 
+    def test_materialize_with_odfv_writes(self):
+        with tempfile.TemporaryDirectory() as data_dir:
+            self.store = FeatureStore(
+                config=RepoConfig(
+                    project="test_on_demand_python_transformation",
+                    registry=os.path.join(data_dir, "registry.db"),
+                    provider="local",
+                    entity_key_serialization_version=2,
```
ntkathole:
```suggestion
                    entity_key_serialization_version=3,
```

sdk/python/feast/feature_store.py
franciscojavierarceo: Actually no, for Transformation on Writes the transformation happens in the feature server, for stream feature views, the transformation happens in the stream processor (e.g., Flink) the materialization is pure ingestion.

sdk/python/feast/feature_store.py
```diff
@@ -1312,6 +1332,11 @@ def materialize_incremental(
         )
         # TODO paging large loads
         for feature_view in feature_views_to_materialize:
+            from feast.on_demand_feature_view import OnDemandFeatureView
```
haoxu0, line 1335: to avoid import twice, can put this at top of the file?

sdk/python/feast/feature_store.py
jyejare, line 659: Interesting that we are returning only FeatureView/OnDemandFeatureView, but in reality when `feature_views` are not given, we are returning everything including Stream Feature Views.

The return list needs to change here.

sdk/python/feast/feature_store.py
```diff
@@ -669,34 +669,53 @@ def _get_feature_views_to_materialize(
...
+            feature_views_to_materialize.extend(
+                [
+                    odfv
+                    for odfv in on_demand_feature_views_to_materialize
+                    if odfv.write_to_online_store
+                ]
+            )
         else:
             for name in feature_views:
+                feature_view: Union[FeatureView, OnDemandFeatureView]
```
jyejare, line 697: Could we accept one more parameter to `_get_feature_views_to_materialize` for returning the list of different types of feature views depending on `purpose`.

Its really confusing why we are returning different FV types list in `if and else blocks `.

sdk/python/feast/feature_store.py
```diff
@@ -1352,12 +1377,13 @@ def tqdm_builder(length):
                 tqdm_builder=tqdm_builder,
             )
 
-            self._registry.apply_materialization(
-                feature_view,
-                self.project,
-                start_date,
-                end_date,
-            )
+            if not isinstance(feature_view, OnDemandFeatureView):
```
jyejare, line 1380: Since We are continuing the loop at line 1337&1338, we dont need this condition here.

sdk/python/feast/feature_store.py
```diff
@@ -1312,6 +1332,11 @@ def materialize_incremental(
         )
         # TODO paging large loads
         for feature_view in feature_views_to_materialize:
+            from feast.on_demand_feature_view import OnDemandFeatureView
+
+            if isinstance(feature_view, OnDemandFeatureView):
+                continue
```
jyejare, line 1338: For my knowledge:
What does it mean that the ODFV cannot be materialized?

sdk/python/feast/feature_store.py
```diff
@@ -1426,12 +1454,13 @@ def tqdm_builder(length):
                 tqdm_builder=tqdm_builder,
             )
 
-            self._registry.apply_materialization(
-                feature_view,
-                self.project,
-                start_date,
-                end_date,
-            )
+            if not isinstance(feature_view, OnDemandFeatureView):
```
jyejare, line 1457: Rather than waiting until this line for ODFV based skip, `Continue` the looping before getting the provider (line1438).

kind of bringing uniqueness !

sdk/python/feast/infra/passthrough_provider.py
```diff
@@ -420,13 +421,22 @@ def ingest_df_to_offline_store(self, feature_view: FeatureView, table: pa.Table)
...
         config: RepoConfig,
-        feature_view: FeatureView,
+        feature_view: Union[FeatureView, OnDemandFeatureView],
         start_date: datetime,
         end_date: datetime,
         registry: BaseRegistry,
         project: str,
         tqdm_builder: Callable[[int], tqdm],
     ) -> None:
+        from feast.on_demand_feature_view import OnDemandFeatureView
```
jyejare, line 431: We are reapeating the import.

ntkathole:
Fixed via https://github.com/feast-dev/feast/pull/5528

PR #5441, opened 2025-06-10
fix: Increase the Operator Manager memory limits and requests

jyejare:

The operator manager pod is failing with CrashBackLoop due to OOM. 
```
Terminated at Jun 10, 2025, 1:12 PM with exit code 137 (OOMKilled)
```

This indicates that the operator manager pod and manifest needs to have an increased resources.

RHOAIENG-27341

# Misc

### DIFF
--- a/infra/feast-operator/config/manager/manager.yaml
+++ b/infra/feast-operator/config/manager/manager.yaml
@@ -91,8 +91,8 @@ spec:
         # More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
         resources:
           limits:
-            cpu: 500m
-            memory: 128Mi
+            cpu: 1000m
+            memory: 256Mi
           requests:
             cpu: 10m
             memory: 64Mi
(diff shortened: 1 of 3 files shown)


infra/feast-operator/config/manager/manager.yaml
```diff
@@ -91,10 +91,10 @@ spec:
...
           limits:
-            cpu: 500m
-            memory: 128Mi
+            cpu: 1000m
+            memory: 256Mi
           requests:
-            cpu: 10m
-            memory: 64Mi
+            cpu: 50m
+            memory: 128Mi
```
tchughesiv: probably best to leave the requests low (as-is) so the manager can run in very small clusters ... kind, etc.

tchughesiv (changes requested):

jyejare:
@tchughesiv Updated!

tchughesiv:
thanks @jyejare but you'll need to run `make test` and then commit changes as well. unit test will fail otherwise

PR #5454, opened 2025-06-14
feat: Make batch_source optional in PushSource (#5440)

### LINKED ISSUE #5440: Remove batch_source required field for PushSource
I have a `PushSource` which I used at offline time to push features that I created during training (for example, embedding of entity after model training),

The problem is that `PushSource` **requires** me to provide a `batch_source` but I dont have this kind of features before training (Because this features are generated only after training) so I must do this overhead/workaround:
* Create a dummy parquet file to store at least one value (else if I remember right I got an error)
* Create a FileSource  to point the parquet file.
* Point the `batch_source`  to the `FileSource`
All of this need to done for each feature view of a push source that I want,

Expected behavour will be to leave the `batch_source` empty and only create `FV` and `PushSource`

**Additional context**
You can see example of the workaround [here](https://github.com/RHEcosystemAppEng/rec-sys-app/blob/main/feature_repo/data_sources.py#L52)

snehsuresh:
feat: make batch_source optional for PushSource

This PR addresses a common pain point when using `PushSource` without a historical data source. Previously, `batch_source` was required even when not needed, forcing users to create dummy files just to satisfy the schema.

With this change:
- `batch_source` is now optional in `PushSource`
- `from_proto` and `to_proto` updated to safely handle absence of `batch_source`
- Added a test to ensure the functionality works as expected

Fixes #5440

# Misc
- Added unit test: `test_push_source_without_batch_source`
- All tests pass locally (`pytest tests/unit/test_data_sources.py`)
- No breaking changes introduced

### DIFF
--- a/sdk/python/feast/data_source.py
+++ b/sdk/python/feast/data_source.py
@@ -815,7 +815,6 @@ class PushSource(DataSource):
 
     @staticmethod
     def from_proto(data_source: DataSourceProto):
-        # assert data_source.HasField("batch_source")
         batch_source = (
             DataSource.from_proto(data_source.batch_source)
             if data_source.HasField("batch_source")
@@ -831,8 +830,6 @@ class PushSource(DataSource):
         )
 
     def to_proto(self) -> DataSourceProto:
-        # batch_source_proto = None
-
         data_source_proto = DataSourceProto(
             name=self.name,
             type=DataSourceProto.PUSH_SOURCE,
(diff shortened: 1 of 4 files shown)


sdk/python/feast/data_source.py
```diff
@@ -827,19 +831,19 @@ def from_proto(data_source: DataSourceProto):
         )
 
     def to_proto(self) -> DataSourceProto:
-        batch_source_proto = None
-        if self.batch_source:
-            batch_source_proto = self.batch_source.to_proto()
+        # batch_source_proto = None
```
franciscojavierarceo: please remove comment

sdk/python/feast/data_source.py
```diff
@@ -815,8 +815,12 @@ def get_table_column_names_and_types(
 
     @staticmethod
     def from_proto(data_source: DataSourceProto):
-        assert data_source.HasField("batch_source")
-        batch_source = DataSource.from_proto(data_source.batch_source)
+        # assert data_source.HasField("batch_source")
```
franciscojavierarceo: please remove comment

PR #5461, opened 2025-06-19
test: Add RHOAI upgrate test to validate feast apply and materialize functionality

Srihari1192:

Added RHOAI upgrade test scenario of credit-scoring example to verify the functionality of a Feast FeatureStore
after upgrade. 
- Introduced `feast_preupgrade_test` to ensure the FeatureStore CR is correctly set up prior to upgrade
- Added `feast_post_upgrade_test` to validate feast apply and materialize functionality after upgrade
- Implemented `ReplaceNamespaceInYamlFilesInPlace` utility to dynamically update namespaces in YAMLs,
  preventing conflicts with existing namespaces used by other tests
- Removed model training functionality from Feast apply and materialize tests, as it is not relevant to validate

### DIFF
--- a/infra/feast-operator/test/e2e_rhoai/feast_postupgrade_test.go
+++ b/infra/feast-operator/test/e2e_rhoai/feast_postupgrade_test.go
@@ -1,0 +1,56 @@
+/*
+Copyright 2025 Feast Community.
+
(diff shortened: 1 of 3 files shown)


ntkathole (approved):
Looks good 👍

PR #5462, opened 2025-06-21
feat: Support DynamoDB as online store in Go feature server

iamcodingcat:
# What this PR does / why we need it:

For ultimate low latency feature serving, I am using Golang api server for production environment. Moreover, a few months ago, I developed the feature that integrates s3 registry into Golang api server.(#5352 , #5336)

I am using Golang api server with redis as online store. But redis is usually for cache server and have lower SLA than DynamoDB. So I tried to integrate dynamodb as online store in Golang api server and succeed it. I refer the [source code](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/online_stores/dynamodb.py) from integrating dynamodb in python sdk.

Of course, I am not writing test code but I locally finish testing about this feature. After reviewing my PR, I will soon add test code for this. please review about this for me! Thanks.

For reference, I carefully see compatibility between python and go of `mmh3` hash algorithm used for converting entity key to entity id(In feast python sdk, it use [this](https://github.com/feast-dev/feast/blob/8a37bf55fb0297de8d035fbdbb069df8e9a57457/sdk/python/feast/infra/online_stores/helpers.py#L60)). I use this [Golang lib](https://pkg.go.dev/github.com/roberson-io/mmh3#Hashx64_128) and I check the complete compatibility in this lib with python sdk. I found [this issue](https://github.com/hajimes/mmh3/issues/46) while investigating this. In my local environment(64-bit little endian), the test was completed. I will test it on my production linux environment on k8s cluster, asap.

# Which issue(s) this PR fixes:
Golang api server can support DynamoDB as online store. About this, the issue(#3971) is made before. I will support this.

# Misc

@shuchu @jyejare cc. @franciscojavierarceo

PR #5352, opened 2025-05-14
fix: Reopen for integration test about add s3 storage-based registry store in Go feature server

iamcodingcat:
# What this PR does / why we need it:
- I reopened for integration test with master branch in #5336 cc. @franciscojavierarceo 

# Which issue(s) this PR fixes:

# Misc

PR #5336, opened 2025-05-07
feat: Add s3 storage-based registry store in Go feature server

iamcodingcat:
# What this PR does / why we need it:
This pull request integrates S3 storage as a registry store supported by the Go Feature Server, which is currently in alpha. At the moment, the Go Feature Server only supports a registry store based on the local file system. To make it suitable for production use in a service environment, we developed support for a registry store based on AWS S3 — a widely adopted and popular cloud storage solution.

I developed below things
- [x] protobuf version upgraded to latest version
    - [x] arm64 platform and MacOS compatibility
    - [x] arm64 platform and debian OS compatibility
- [x] add feature of s3 based registry store and test-code

*For your reference, the commit username younghun-jo-ilevit-com is also me.

# Which issue(s) this PR fixes:
No. it is just for feature enhancement.

# Misc

Issue #3971, opened 2024-02-27
Golang support for DynamoDb


dawid-laszuk-sp:
**Preface**

Ticket is both for "feature request" and "documentation request". Of course, we'd love the work to be done, but we (Sailpoint) are no strangers to being busy and no one can contribute, then maybe we could give it a shot.

**Problem**

We have a growing stack of tech and langues. One of the solutions is to update feast to support DynamoDB in golang.  From the [documentation](https://docs.feast.dev/reference/feature-servers/go-feature-server) it looks like Go server is both "in alpha", and doesn't support DynamoDB. 

**Describe the solution you'd like**

DynamoDB support in golang.

**Describe alternatives you've considered**

We've considered contributing to make this feature happen. However, as we haven't looked/worked with golang side of the feast, it's difficult to estimate how much effort this change would be. Any pointers on what might need updating would be much appreciated, with the hope that we can contribute.

PR #5464, opened 2025-06-25
feat: Support DynamoDB as online store in Go feature server

iamcodingcat:

**(I re-created from [previous PR](https://github.com/feast-dev/feast/pull/5462) because of twisted commits. sorry 😢 )**
For ultimate low latency feature serving, I am using Golang api server for production environment. Moreover, a few months ago, I developed the feature that integrates s3 registry into Golang api server.(#5352 , #5336)

I am using Golang api server with redis as online store. But redis is usually for cache server and have lower SLA than DynamoDB. So I tried to integrate dynamodb as online store in Golang api server and succeed it. I refer the [source code](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/online_stores/dynamodb.py) from integrating dynamodb in python sdk.

Of course, I am not writing test code but I locally finish testing about this feature. After reviewing my PR, I will soon add test code for this. please review about this for me! Thanks.

For reference, I carefully see compatibility between python and go of `mmh3` hash algorithm used for converting entity key to entity id(In feast python sdk, it use [this](https://github.com/feast-dev/feast/blob/8a37bf55fb0297de8d035fbdbb069df8e9a57457/sdk/python/feast/infra/online_stores/helpers.py#L60)). I use this [Golang lib](https://pkg.go.dev/github.com/roberson-io/mmh3#Hashx64_128) and I check the complete compatibility in this lib with python sdk. I found [this issue](https://github.com/hajimes/mmh3/issues/46) while investigating this. In my local environment(64-bit little endian), the test was completed. I will test it on my production linux environment on k8s cluster, asap.

Golang api server can support DynamoDB as online store. About this, the issue(#3971) is made before. I will support this.

# Misc

@shuchu @jyejare cc. @franciscojavierarceo


go.mod
```diff
@@ -6,15 +6,21 @@ toolchain go1.22.5
...
+	github.com/aws/aws-sdk-go-v2/config v1.29.14
+	github.com/aws/aws-sdk-go-v2/service/dynamodb v1.43.3
+	github.com/aws/aws-sdk-go-v2/service/s3 v1.79.3
 	github.com/ghodss/yaml v1.0.0
 	github.com/golang/protobuf v1.5.4
 	github.com/google/uuid v1.6.0
 	github.com/mattn/go-sqlite3 v1.14.23
 	github.com/pkg/errors v0.9.1
 	github.com/redis/go-redis/v9 v9.6.1
+	github.com/roberson-io/mmh3 v0.0.0-20190729202758-fdfce3ba6225
```
iamcodingcat, line 19: I refer @jyejare 's [comment](https://github.com/feast-dev/feast/pull/5462#discussion_r2163739773) in previous PR  cc. @franciscojavierarceo @tchughesiv 

> @franciscojavierarceo @tchughesiv How are we treating these external/personal repos as a safe repositories to use ? Something like govulncheck Github Action should really help.

PR #5470, opened 2025-06-27
fix: Update generate_answer function to provide correct parameter format for retrieve function

Fiona-Waters:

This PR makes necessary updates to the generate_answer function in the FeastRagRetriever. In testing a jupyter notebook example I noticed that a numpy array was expected by the retrieve function but a tensor was being passed in. This is now fixed, has been tested and works correctly.

N/A

# Misc

@franciscojavierarceo please take a look.


franciscojavierarceo (approved):

ntkathole (approved):
👍

PR #5471, opened 2025-06-27
feat: Populate created and last updated timestamp on data sources

ntkathole:

Currently data sources do not have created and updated timestamp metadata, while other resources in feast has it.
This PR is is to add created and last updated timestamp on data sources.

```
{
      "type": "REQUEST_SOURCE",
      "requestDataOptions": {
        "schema": [
          {
            "name": "val_to_add",
            "valueType": "INT64"
          },
          {
            "name": "val_to_add_2",
            "valueType": "INT64"
          }
        ]
      },
      "name": "vals_to_add",
      "meta": {
        "createdTimestamp": "2025-06-27T16:48:48.904245Z",
        "lastUpdatedTimestamp": "2025-06-27T16:48:48.904247Z"
      }
    },
    {
      "type": "BATCH_FILE",
      "timestampField": "event_timestamp",
      "createdTimestampColumn": "created",
      "fileOptions": {
        "uri": "data/driver_stats.parquet"
      },
      "name": "driver_hourly_stats_source",
      "meta": {
        "createdTimestamp": "2025-06-27T16:48:48.904411Z",
        "lastUpdatedTimestamp": "2025-06-27T16:48:48.904412Z"
      }
    }
```


sdk/python/feast/feature_view.py
```diff
@@ -263,6 +263,7 @@ def __eq__(self, other):
             or self.batch_source != other.batch_source
             or self.stream_source != other.stream_source
             or sorted(self.entity_columns) != sorted(other.entity_columns)
+            or self.materialization_intervals != other.materialization_intervals
```
ntkathole, line 266: Without this check, two FeatureViews with identical definitions but created at different times would now be considered "different" because their data sources had different timestamps. Including materialization intervals in equality ensures that the registry correctly identifies when materialization history needs to be preserved, even if it means more registry updates are triggered. This behavior is verified in `test_modify_feature_views_success` test.

franciscojavierarceo (approved):

ntkathole:
@sudohainguyen please help to review here

sdk/python/feast/feature_view.py
jyejare: > two FeatureViews with identical definitions but created at different times would now be considered "different"

Do feature view name is not enough here to identify them as a `different` ?

sdk/python/feast/infra/registry/registry.py
```diff
@@ -338,10 +338,27 @@ def list_data_sources(
     def apply_data_source(
         self, data_source: DataSource, project: str, commit: bool = True
     ):
+        now = _utc_now()
+        if not data_source.created_timestamp:
+            data_source.created_timestamp = now
+        data_source.last_updated_timestamp = now
```
jyejare, line 344: Should the update time be updated just for registering the data source in registry. I dont see it is an actual update.

wdyt ?

jyejare: Also in what cases do you see the data sources wont have create and update timestamp as null ? Becasue we are setting it at the initialization time itself ?

sdk/python/feast/infra/registry/registry.py
```diff
@@ -338,10 +338,27 @@ def list_data_sources(
...
-                del registry.data_sources[idx]
+                existing_data_source = DataSource.from_proto(existing_data_source_proto)
+                # Check if the data source has actually changed
+                if existing_data_source == data_source:
+                    return
+                else:
+                    # Preserve created_timestamp from existing data source
+                    data_source.created_timestamp = (
+                        existing_data_source.created_timestamp
+                    )
```
jyejare, line 358: Pritty strange why we want to preserve the timestamp of datasource which we are about to delete ?

jyejare (commented):
Looks Good, Have some comments and questions !

sdk/python/feast/feature_view.py
ntkathole: `__eq__` is not about identifying a unique feature view within a registry. The registry uses equality checks to determine if updates are needed.

sdk/python/feast/infra/registry/registry.py
ntkathole: When applying a new data source, initially there won't be any metadata, which is null.         data_source.last_updated_timestamp = now line itself setting last_updated_timestamp on data source.

sdk/python/feast/infra/registry/registry.py
ntkathole: The logic is not about deletion. It finds existing data source in registry, If it's different from the new one, preserve the original created_timestamp (because the data source was first created at that time), delete the old version from registry, add the new version with the preserved created_timestamp. This ensures created_timestamp represents when the data source was first created, not when this particular update happened.

PR #5473, opened 2025-06-29
fix: Fix remote rbac integration tests

### LINKED ISSUE #4904: Fixing the integration tests to spin up only one keycloak instance

ntkathole:
`
test-python-integration-rbac-remote` tests are causing GitHub runners to terminate due to the memory issues.

- The `RemoteOfflineStoreDataSourceCreator` class was missing a teardown() method. When tests ran, feast serve_offline subprocesses were started but never cleaned up. This was the main reason behind github runner termination.
- Also, fixed the keycloak setup to share a single Keycloak instance across all test runners, allowing us tests to run with more pytest runnners. 

Fixes #4904

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,7 +155,7 @@ test-python-integration-local: ## Run Python integration tests (local dev mode)
 test-python-integration-rbac-remote: ## Run Python remote RBAC integration tests
 	FEAST_IS_LOCAL_TEST=True \
 	FEAST_LOCAL_ONLINE_CONTAINER=True \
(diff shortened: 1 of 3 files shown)


franciscojavierarceo (approved):
thank you for this!

PR #5481, opened 2025-06-30
chore: Added ntkathole and jyejare to red hat CI approvers

ntkathole:
Adding myself and @jyejare to red hat CI approvers.


ntkathole:
@redhatHameed @tchughesiv ^

PR #5482, opened 2025-06-30
Draft: Compute engine multi sources and feature view as source

HaoXuAI:

# Misc


sdk/python/feast/batch_feature_view.py
```diff
@@ -53,6 +53,7 @@ class BatchFeatureView(FeatureView):
     entities: List[str]
     ttl: Optional[timedelta]
     source: DataSource
+    sink_source: Optional[DataSource] = None
```
franciscojavierarceo, line 56: Maybe we should just call it sink?

sdk/python/feast/infra/compute_engines/algorithms/topo.py
```diff
@@ -0,0 +1,43 @@
+from typing import List, Set
+
+from feast.infra.compute_engines.dag.node import DAGNode
+
+
+def topo_sort(root: DAGNode) -> List[DAGNode]:
```
franciscojavierarceo, line 6: Why not call it topological_sort?

franciscojavierarceo (approved):
Some small nits but this mostly lgtm


Can you add a page on the docs before merging this PR? Would be great to share with community on how to use it.

sdk/python/feast/batch_feature_view.py
haoxu0: Sounds good, will update it

sdk/python/feast/feature_view.py
```diff
@@ -403,12 +435,21 @@ def from_proto(cls, feature_view_proto: FeatureViewProto):
...
+            if feature_view_proto.spec.HasField("batch_source")
+            else None
+        )
         stream_source = (
             DataSource.from_proto(feature_view_proto.spec.stream_source)
             if feature_view_proto.spec.HasField("stream_source")
             else None
         )
+        source_views = [
+            FeatureView.from_proto(FeatureViewProto(spec=view_spec, meta=None))
```
ntkathole, line 449: `from_proto()` method recursively calls itself for each source view without any depth limit. While cycle detection exists in `FeatureResolver`,  cycle detection only runs when you use the compute engine, but proto deserialization happens much earlier during APIs/ registry loading. 

We might need to handle this in FeatureView.from_proto().

ntkathole: Also, do we not need to store metadata for nested feature views ?  meta=None ?

sdk/python/feast/feature_view.py
```diff
@@ -90,6 +90,7 @@ class FeatureView(BaseFeatureView):
     ttl: Optional[timedelta]
     batch_source: DataSource
     stream_source: Optional[DataSource]
+    source_views: Optional[List["FeatureView"]]
```
ntkathole, line 93: In `__eq__ `, I think we also need to compare compare source_views, else two FeatureViews with different source dependencies will be considered equal.

same for `__copy__ `

sdk/python/feast/feature_view.py
ntkathole: We need both cycle detection and de-duplication during serialization. It may not cause issue for few feature views but if there are many feature views, it could cause slowness.

A -> [B, C]
B -> [D, E]
C -> [D, E]

When serializing FeatureViewA = FeatureViewD and FeatureViewE get serialized twice.

sdk/python/feast/feature_view.py
haoxu0: good catch

sdk/python/feast/feature_view.py
haoxu0: right make sense.
I don't have any meta data required for compute engine at the moment, what do you think something useful?

haoxu0:
Merged it and add new PR with following suggestions

PR #5487, opened 2025-07-01
Enhance remote document retrieval functionality

jyejare:
PR includes:

1. Implements a new request model for remotely retrieving online documents
2. Updates the feature server to handle document retrieval,
3. Both V1 and V2 docvuments retrival is implemented
4. Unit tests to ensure functionality.


sdk/python/feast/feature_server.py
```diff
@@ -88,8 +88,16 @@ class GetOnlineFeaturesRequest(BaseModel):
     feature_service: Optional[str] = None
     features: Optional[List[str]] = None
     full_feature_names: bool = False
-    query_embedding: Optional[List[float]] = None
+
+
+class GetOnlineDocumentsRequest(BaseModel):
```
ntkathole, line 93: All fields are optional  ?

sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -197,12 +427,104 @@ def _construct_online_read_api_json_request(
...
+
+    def _construct_online_documents_v2_api_json_request(
+        self,
+        table: FeatureView,
+        requested_features: Optional[List[str]],
+        embedding: Optional[List[float]],
+        top_k: int,
+        distance_metric: Optional[str] = None,
+        query_string: Optional[str] = None,
+        api_version: Optional[int] = 1,
```
ntkathole: default api_version 2 ?

sdk/python/feast/feature_server.py
```diff
@@ -246,24 +257,26 @@ async def get_online_features(request: GetOnlineFeaturesRequest) -> Dict[str, An
...
-            query_string=request.query_string,
-        )
+        read_params = dict(features=features, query=request.query, top_k=request.top_k)
+        if request.api_version == 2 and request.query_string is not None:
+            read_params["query_string"] = request.query_string
 
-        response = await run_in_threadpool(
-            lambda: store.retrieve_online_documents_v2(**read_params)  # type: ignore
-        )
+        if request.api_version == 2:
```
franciscojavierarceo, line 272: this shouldn't be necessary. v2 vs v1 should be approximately the same where v2 was meant to incorporate more fields. If there is a difference, it's a bug.

franciscojavierarceo (changes requested):

sdk/python/feast/feature_server.py
franciscojavierarceo: we should remove v1 and move to v2 for all of the existing usage.

jyejare: @franciscojavierarceo Yes, I have seen atleast with postgres online store that v1 and v2 are retrieving different dataframes as the logic is completely different for both versions.

I agree for keeping one solid logic across all stores but only after identifying which is the best version `AND THEN` we should adapt the best logic having more fields. Finally I think we can remove the bad one and keep single retrieval version.

But I think the investigatin and removal from all stores should not be part of this PR, we should handle that in spearate PR.

sdk/python/feast/feature_server.py
jyejare: Since we are not passing entities dataframe, yes, everything is optional.

We can make features as non optional though.

sdk/python/feast/infra/online_stores/remote.py
jyejare: I havent tested version 2 in depth for all online stores. BTW Do we have v2 implemented for all stores and confirm stability ?

If not, I think we can go with v1 as stable with opt-in to v2.

sdk/python/feast/feature_server.py
franciscojavierarceo: features should be required

sdk/python/feast/feature_server.py
franciscojavierarceo: yeah that's fine, let's just make an issue to track, yeah?

sdk/python/feast/infra/online_stores/remote.py
franciscojavierarceo: let's set default to v2.

sdk/python/feast/feature_server.py
jyejare: Generated issue based on our discussion: https://github.com/feast-dev/feast/issues/5537

PR #5496, opened 2025-07-07
feat: Added relationship support to all API endpoints

ntkathole:

This enhancement allows users to query relationships and lineage information for all Feast objects (entities, feature views, data sources, feature services, etc.) through a consistent `include_relationships` parameter.

- Include all relationships (both direct and indirect) for the object(s) in the list and get API response. 
- Added `get_object_relationships()` function for single object relationship queries.
- Added `get_relationships_for_objects() `function for bulk relationship queries.
- Added relationship for entity <-> data sources and test coverage for entity-to-data source relationships.
- Updated API endpoint documentation with relationship examples.

Screenshot 2025-07-07 at 3 34 27 PM

### DIFF
--- a/docs/reference/feature-servers/registry-server.md
+++ b/docs/reference/feature-servers/registry-server.md
@@ -32,6 +32,15 @@ Most endpoints support these common query parameters:
 - `allow_cache` (optional, default: `true`): Whether to allow cached data
 - `tags` (optional): Filter results by tags in key=value format
 
+#### Relationship Parameters
+- `include_relationships` (optional, default: `false`): Include all relationships (both direct and indirect) for the object(s) in the response
+
+#### Pagination Parameters (List Endpoints Only)
+- `page` (optional): Page number (starts from 1)
+- `limit` (optional, max: 100): Number of items per page
+- `sort_by` (optional): Field to sort by
+- `sort_order` (optional): Sort order: "asc" or "desc" (default: "asc")
+
 ### Entities
 
 #### List Entities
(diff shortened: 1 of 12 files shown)


docs/reference/feature-servers/registry-server.md
```diff
@@ -32,17 +32,41 @@ Most endpoints support these common query parameters:
 - `allow_cache` (optional, default: `true`): Whether to allow cached data
 - `tags` (optional): Filter results by tags in key=value format
 
+#### Relationship Parameters
+- `include_relationships` (optional, default: `false`): Include all relationships (both direct and indirect) for the object(s) in the response
+
+#### Pagination Parameters (List Endpoints Only)
+- `page` (optional): Page number (starts from 1)
+- `limit` (optional, max: 100): Number of items per page
```
jyejare, line 40: Add the default value as well ?

docs/reference/feature-servers/registry-server.md
```diff
@@ -275,6 +394,140 @@ All endpoints return JSON responses with the following general structure:
...
+- Feature Services → Feature Views (feature services consume feature views)
+- Entities → Data Sources (entities connect to data sources through feature views)
+- Entities → Feature Services (entities connect to feature services through feature views)
+
+#### Pagination
+
+All list endpoints support pagination to improve performance and manageability of large datasets:
+
+- **`page`** - Page number (starts from 1)
+- **`limit`** - Number of items per page (maximum 100)
```
jyejare, line 527: Why do we have a maximum limitation of 100 ?

jyejare (commented):
LGTM, few non blocking comments.

docs/reference/feature-servers/registry-server.md
ntkathole: This is simply to prevent from requesting excessively large datasets and very large result sets are typically not useful for UI display anyway. Also, REST API may be processing relationships (when `include_relationships=true`) which can multiply the data size significantly.

docs/reference/feature-servers/registry-server.md
ntkathole: Pagination is optional, and adding default value will change the existing behavior, which is not good.

jyejare: I mean if limit is optional, there should be some default value we would have `if limit is not given`.

ntkathole: If limit is not given, by default api returns all the values. There is no limit applied. or did I miss something?

jyejare: Considering the default limit is not given and it could attempt to fetch all the data in one request along with relationships(if asked), the API could max limit and hence good to catch that error and ask user to use the `limit` parameter for chunked retrival.

Non blocking for this PR.

jyejare (approved):

franciscojavierarceo (approved):

PR #5497, opened 2025-07-08
fix: Added support to handle empty dataframe

aniketpalu:

Adds validation to `write_to_online_store` and `write_to_online_store_async` to prevent empty dataframes from being written to the online store.

**Changes:**
- Validates completely empty dataframes 
- Validates dataframes with entity data but empty feature columns
- Works with both `df` and `inputs` parameters
- Throws descriptive `ValueError` messages

[RHOAIENG-27681](https://issues.redhat.com/browse/RHOAIENG-27681)

# Misc

Added comprehensive unit tests in `TestEmptyDataFrameValidation` class.

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -1746,8 +1746,6 @@ class FeatureStore:
             allow_registry_cache (optional): Whether to allow retrieving feature views from a cached registry.
             transform_on_write (optional): Whether to transform the data before pushing.
 
-        Raises:
-            ValueError: If the dataframe is empty (has no rows) or if feature columns are empty.
         """
 
         feature_view, df = self._get_feature_view_and_df_for_online_write(
@@ -1793,8 +1791,6 @@ class FeatureStore:
             inputs: Optional the dictionary object to be written
             allow_registry_cache (optional): Whether to allow retrieving feature views from a cached registry.
 
-        Raises:
-            ValueError: If the dataframe is empty (has no rows) or if feature columns are empty.
         """
 
         feature_view, df = self._get_feature_view_and_df_for_online_write(
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -1761,14 +1761,14 @@ class FeatureStore:
         # Validate that the dataframe has meaningful feature data
         if df is not None:
             if df.empty:
-                raise ValueError("Cannot write empty dataframe to online store")
+                warnings.warn("Cannot write empty dataframe to online store")
 
             # Check if feature columns are empty (entity columns may have data but feature columns are empty)
             feature_column_names = [f.name for f in feature_view.features]
             if feature_column_names:
                 feature_df = df[feature_column_names]
                 if feature_df.empty or feature_df.isnull().all().all():
-                    raise ValueError(
+                    warnings.warn(
                         "Cannot write dataframe with empty feature columns to online store"
                     )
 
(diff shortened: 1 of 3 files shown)


sdk/python/feast/feature_store.py
```diff
@@ -1780,6 +1801,21 @@ async def write_to_online_store_async(
...
+        if df is not None:
+            if df.empty:
+                raise ValueError("Cannot write empty dataframe to online store")
+
+            # Check if feature columns are empty (entity columns may have data but feature columns are empty)
+            feature_column_names = [f.name for f in feature_view.features]
+            if feature_column_names:
+                feature_df = df[feature_column_names]
+                if feature_df.empty or feature_df.isnull().all().all():
+                    raise ValueError(
```
ntkathole: I think this is breaking change. If `write_to_online_store()` raises a ValueError due to empty data validation, it will stop the entire materialization process.

You can test the scenario : consider there are 5 feature views and the 2nd one has empty data, the remaining 3 won't be processed.



https://github.com/feast-dev/feast/blob/master/sdk/python/feast/feature_store.py#L1430

aniketpalu: Thanks for pointing out. Addressed your comment.

ntkathole:
@jyejare PTAL

sdk/python/tests/unit/online_store/test_online_writes.py
```diff
@@ -155,6 +159,419 @@ def test_online_retrieval(self):
...
+                "event_timestamp": [current_time] * 2,
+                "created": [current_time] * 2,
+                "conv_rate": [0.5, 0.7],
+                "acc_rate": [0.8, 0.9],
+                "avg_daily_trips": [10, 12],
+            }
+        )
+
+        # This should not raise an exception
+        self.store.write_to_online_store(
```
jyejare, line 310: You are not validating that its not raising any exception using `with not pytest.raises(Exception_class_name_here)`.

sdk/python/tests/unit/online_store/test_online_writes.py
```diff
@@ -155,6 +159,419 @@ def test_online_retrieval(self):
...
+                    "event_timestamp": [current_time] * 2,
+                    "created": [current_time] * 2,
+                    "conv_rate": [0.5, 0.7],
+                    "acc_rate": [0.8, 0.9],
+                    "avg_daily_trips": [10, 12],
+                }
+            )
+
+            # This should not raise an exception
+            await self.store.write_to_online_store_async(
```
jyejare, line 338: Same as above in async test.

sdk/python/tests/unit/online_store/test_online_writes.py
```diff
@@ -155,6 +159,419 @@ def test_online_retrieval(self):
...
+                "event_timestamp": [current_time] * 3,
+                "created": [current_time] * 3,
+                "conv_rate": [0.5, None, 0.7],  # Mixed values
+                "acc_rate": [0.8, 0.9, None],  # Mixed values
+                "avg_daily_trips": [10, 12, 15],  # All valid
+            }
+        )
+
+        # This should not raise an exception because not all feature values are null
+        self.store.write_to_online_store(
```
jyejare, line 365: Same as above here

sdk/python/tests/unit/online_store/test_online_writes.py
jyejare: Or may be just `warnings` context manager.

jyejare (commented):
ACK with nitpick comments.

sdk/python/tests/unit/online_store/test_online_writes.py
aniketpalu: Addressed comments & optimised code to reuse warning checking logic

ntkathole (approved):
Looks good

PR #5498, opened 2025-07-09
fix: Standalone Web UI Publish Workflow

ntkathole:

Earlier UI npm publish was part of release workflow, while all other publish workflows were part of publish.yml. 

All publish operations now follow the same modular pattern. UI npm publishing logic is isolated and reusable and can trigger UI publishing independently or as part of releases.


franciscojavierarceo (approved):
lgtm

PR #5500, opened 2025-07-09
fix: Fix release version without prefix

ntkathole:

Custom version v0.50.0 → strips to 0.50.0 for script
Tagged branch v0.50.0 → strips to 0.50.0 for script
NEXT_VERSION already clean from `version_without_prefix`


.github/workflows/publish.yml
```diff
@@ -66,6 +66,4 @@ jobs:
     needs: [ publish-python-sdk ]
     secrets: inherit
     with:
-      custom_version: ${{ github.event.inputs.custom_version }}
```
franciscojavierarceo: sometimes we have to override the custom version for historical issues.

franciscojavierarceo (approved):

PR #5513, opened 2025-07-14
feat: Added endpoints to allow user to get data for all projects

ntkathole:

This PR enable users to fetch a globally aggregated, paginated, and sorted list of entities, data sources, feature views, feature services, and saved datasets across all Feast projects via the REST API. This is especially useful to get a unified view of all resources, regardless of project boundaries.


HaoXuAI (approved):
LGTM!

ntkathole:
> this is great! can you update the docs that we support this now?

Added docs, Thank you!

PR #5521, opened 2025-07-18
feat: Update doc

HaoXuAI:

Updated docs of compute engine, BatchFeatureView and StreamFeatureView.
Update the compute config to have FeatureView based config overrides the project level config.

# Misc


ntkathole (approved):
👍

PR #5526, opened 2025-07-20
feat: Added Ray Compute Engine and Ray Offline Store Support

ntkathole:

This PR introduces Ray as both a compute engine and an offline store in Feast. It enables scalable, distributed feature engineering and data access in Feast using Ray, making it easier to handle large datasets and complex feature pipelines.

## Key Features
**Ray Compute Engine**

- Distributed DAG execution for feature transformations, aggregations, joins, and materialization.
- Automatic selection between broadcast and distributed join strategies.
- Resource-aware partitioning and parallelism.
- Efficient point-in-time joins for historical feature retrieval.
- Seamless integration with Feast’s feature view and materialization APIs.

**Ray Offline Store**
- High-performance data reading and writing using Ray Datasets.
- Support for local and distributed Ray clusters.
- Batch writing and saved dataset persistence.
- Configurable resource limits for safe local development and testing.

Documentation added for both Ray compute engine and offline store, including configuration, resource management, and usage examples.


sdk/python/feast/infra/compute_engines/ray/feature_builder.py
```diff
@@ -0,0 +1,224 @@
...
+        # For now, validation is handled in the retrieval job
+        # This could be extended to include Ray-specific validation logic
+        logger.debug("Validation node not implemented yet")
+        return input_node
+
+    def build(self) -> ExecutionPlan:
+        """Build execution plan with optimized order for aggregation scenarios."""
+
+        # For historical retrieval with aggregations, use a different execution order
+        if self.is_historical_retrieval and self._should_aggregate(self.feature_view):
```
haoxu0, line 215: interesting to see this. does it still work if pass in FeatureView as source?

ntkathole: I am still testing this, need to handle https://github.com/feast-dev/feast/issues/5535 first

ntkathole: This works now, serde issue handled as well.

sdk/python/feast/infra/compute_engines/ray/feature_builder.py
```diff
@@ -0,0 +1,325 @@
...
+    ):
+        super().__init__(registry, feature_view, task)
+        self.config = config
+        self.is_historical_retrieval = isinstance(task, HistoricalRetrievalTask)
+        self.is_materialization = isinstance(task, MaterializationTask)
+
+    def build_source_node(self, view):
+        """Build the source node for reading feature data."""
+
+        source_info = resolve_feature_view_source_with_fallback(
```
haoxu0: I'm a bit confused about this one.
The node here is already resolved node, so there should always be a `source` for the node.
For example,
```
feature_view_a
        ^
        |
feature_view_b
```
The resolver will always resolve to `feature_view_a` first, and `feature_view_a` will have physical batch_source, read it, and send it to the downstream `feature_view_b`. Then `feature_view_b` will just use it as the source.

ntkathole: right, I kept the similar logic as offline store, didn't realise it's redundant here and DAG don't need that handling.

franciscojavierarceo (changes requested):
Just putting a hold on this because I want to review. This is fantastic @ntkathole!

docs/reference/compute-engine/ray.md
```diff
@@ -0,0 +1,393 @@
...
+    start_date=datetime.now() - timedelta(days=7),
+    end_date=datetime.now(),
+    feature_views=["driver_stats", "customer_stats"]
+)
+
+# The Ray compute engine handles:
+# - Distributed data processing
+# - Optimal join strategies
+# - Resource management
+# - Progress tracking
```
jyejare, line 119: Do you have plans to add examples for this ? Else we can remove it ?

docs/reference/compute-engine/ray.md
```diff
@@ -0,0 +1,393 @@
...
+
+# Optional: Online store configuration
+online_store:
+    type: sqlite
+    path: data/online_store.db
+
+# Optional: Feature server configuration
+feature_server:
+    port: 6566
+    metrics_port: 8888
```
jyejare: Is this the right configuration? I see type is missing and metrics port is not seen anywhere in feast

docs/reference/compute-engine/ray.md
```diff
@@ -0,0 +1,393 @@
...
+- Row-level transformations
+- Column-level transformations
+- Custom transformation functions
+
+### RayWriteNode
+
+Writes results to various targets:
+- Online stores
+- Offline stores
+- Temporary storage
```
jyejare, line 251: How this is useful?

docs/reference/compute-engine/ray.md
```diff
@@ -0,0 +1,393 @@
...
+import ray
+
+# Check cluster resources
+resources = ray.cluster_resources()
+print(f"Available CPUs: {resources.get('CPU', 0)}")
+print(f"Available memory: {resources.get('memory', 0) / 1e9:.2f} GB")
+
+# Monitor job progress
+job = store.get_historical_features(...)
+# Ray compute engine provides built-in progress tracking
```
jyejare, line 322: I think we missed how to track the progress here in python code or from UI ?


or is it just an information ?

docs/reference/offline-stores/ray.md
```diff
@@ -0,0 +1,512 @@
...
+
+| Option | Type | Default | Description |
+|--------|------|---------|-------------|
+| `type` | string | Required | Must be `feast.offline_stores.contrib.ray_offline_store.ray.RayOfflineStore` or `ray` |
+| `storage_path` | string | None | Path for storing temporary files and datasets |
+| `ray_address` | string | None | Address of the Ray cluster (e.g., "localhost:10001") |
+| `use_ray_cluster` | boolean | false | Whether to use Ray cluster mode |
+| `ray_conf` | dict | None | Ray initialization parameters for resource management (e.g., memory, CPU limits) |
+
+#### Ray Compute Engine Options
```
jyejare, line 181: We are duplicating this information in compute engine and this doc.

docs/reference/offline-stores/ray.md
```diff
@@ -0,0 +1,512 @@
...
+```
+
+## Limitations
+
+The Ray offline store has the following limitations:
+
+1. **File Sources Only**: Currently supports only `FileSource` data sources
+2. **No Direct SQL**: Does not support SQL query interfaces
+3. **No Online Writes**: Cannot write directly to online stores
+4. **Limited Transformations**: Complex feature transformations should use the Ray Compute Engine
```
jyejare: Limited word does not justifies.

docs/reference/offline-stores/ray.md
```diff
@@ -0,0 +1,512 @@
...
+
+## Integration with Ray Compute Engine
+
+For complex feature processing operations, use the Ray offline store in combination with the [Ray Compute Engine](../compute-engine/ray.md). See the **Ray Offline Store + Compute Engine** configuration example in the [Configuration](#configuration) section above for a complete setup.
+
+The Ray offline store provides the data I/O foundation, while the Ray compute engine handles:
```
jyejare: I think we are repeating again

docs/reference/offline-stores/ray.md
```diff
@@ -0,0 +1,512 @@
...
+```python
+# Initialize feature store
+store = FeatureStore("feature_store.yaml")
+
+# Get historical features (uses compute engine if configured)
+features = store.get_historical_features(entity_df=df, features=["fv:feature"])
+
+# Direct data access (uses offline store)
+job = RayOfflineStore.pull_latest_from_table_or_query(...)
+df = job.to_df()
```
jyejare, line 480: I think we should add offline write batch as well

infra/feast-operator/bundle/manifests/feast-operator.clusterserviceversion.yaml
```diff
@@ -50,7 +50,7 @@ metadata:
         }
       ]
     capabilities: Basic Install
-    createdAt: "2025-07-21T20:53:09Z"
+    createdAt: "2025-07-25T09:58:54Z"
```
jyejare, line 53: unnecessary change

pyproject.toml
```diff
@@ -167,9 +168,9 @@ ci = [
     "types-setuptools",
     "types-tabulate",
     "virtualenv<20.24.2",
-    "feast[aws, azure, cassandra, clickhouse, couchbase, delta, docling, duckdb, elasticsearch, faiss, gcp, ge, go, grpcio, hazelcast, hbase, ibis, ikv, k8s, mcp, milvus, mssql, mysql, opentelemetry, spark, trino, postgres, pytorch, qdrant, rag, redis, singlestore, snowflake, sqlite_vec]"
+    "feast[aws, azure, cassandra, clickhouse, couchbase, delta, docling, duckdb, elasticsearch, faiss, gcp, ge, go, grpcio, hazelcast, hbase, ibis, ikv, k8s, mcp, milvus, mssql, mysql, opentelemetry, spark, trino, postgres, pytorch, qdrant, rag, ray, redis, singlestore, snowflake, sqlite_vec]"
 ]
-nlp = ["feast[docling, milvus, pytorch, rag]"]
+nlp = ["feast[docling, milvus, pytorch, rag, ray]"]
```
jyejare: I don't think it should be part of nlp. Does ray supports vector store ?

docs/reference/compute-engine/ray.md
ntkathole: It is added in docs as commented code to explain the role of ray compute engine when materialize or get_historical_features is called.

docs/reference/compute-engine/ray.md
ntkathole: Was added as optional online/feature_server config section, removed

docs/reference/compute-engine/ray.md
ntkathole: It can be used when memory is constrained and processing datasets available exceeds memory.

docs/reference/compute-engine/ray.md
ntkathole: It's Ray's dataset and dashboard progress bars, check https://docs.ray.io/en/latest/data/monitoring-your-workload.html#ray-data-progress-bars

infra/feast-operator/bundle/manifests/feast-operator.clusterserviceversion.yaml
ntkathole: it's not, bundle creates it

sdk/python/feast/infra/compute_engines/ray/nodes.py
```diff
@@ -0,0 +1,722 @@
...
+        try:
+            retrieval_job = create_offline_store_retrieval_job(
+                data_source=self.source,
+                column_info=self.column_info,
+                context=context,
+                start_time=self.start_time,
+                end_time=self.end_time,
+            )
+
+            if hasattr(retrieval_job, "to_ray_dataset"):
```
haoxu0, line 66: we probably can make it a first citizen method for retrieval_job in future

sdk/python/feast/infra/compute_engines/ray/feature_builder.py
```diff
@@ -0,0 +1,314 @@
...
+        """Build execution plan with support for derived feature views and sink_source writing."""
+        if self.is_historical_retrieval and self._should_aggregate(self.feature_view):
+            return self._build_aggregation_optimized_plan()
+
+        if self.is_materialization:
+            return self._build_materialization_plan()
+
+        return super().build()
+
+    def _build_materialization_plan(self) -> ExecutionPlan:
```
haoxu0, line 223: I wonder if this can be just leveraging the parent logic, or something need to be handled by ray only.

ntkathole: It needs Ray-specific handling for intermediate sink writes and derived view materialization chains, which are different from the parent's DAG approach.

sdk/python/feast/infra/compute_engines/ray/nodes.py
ntkathole: yup 👍

docs/reference/compute-engine/ray.md
```diff
@@ -0,0 +1,383 @@
...
+```yaml
+project: my_project
+registry: data/registry.db
+provider: local
+offline_store:
+    type: ray
+    storage_path: data/ray_storage
+batch_engine:
+    type: ray.engine
+    max_workers: 4                          # Optional: Maximum number of workers
```
franciscojavierarceo:
```suggestion
    max_workers: 4                         # Optional: Maximum number of workers
```

docs/reference/compute-engine/ray.md
```diff
@@ -0,0 +1,383 @@
...
+    enable_optimization: true               # Optional: Enable performance optimizations
```
franciscojavierarceo:
```suggestion
    enable_optimization: true              # Optional: Enable performance optimizations
```

docs/reference/compute-engine/ray.md
```diff
@@ -0,0 +1,383 @@
...
+    window_size_for_joins: "1H"            # Optional: Time window for distributed joins
+    ray_address: localhost:10001           # Optional: Ray cluster address
+    use_ray_cluster: false                 # Optional: Use Ray cluster mode
+```
+
+### Configuration Options
+
+| Option | Type | Default | Description |
+|--------|------|---------|-------------|
+| `type` | string | Required | Must be `ray.engine` |
```
franciscojavierarceo: Is this table correct? You have "Required" as the value for the "Default" column. I think that means something may be wrong?

docs/reference/compute-engine/ray.md
```diff
@@ -0,0 +1,383 @@
...
+| `max_workers` | int | CPU count | Maximum number of Ray workers |
+| `enable_optimization` | boolean | true | Enable performance optimizations |
+| `broadcast_join_threshold_mb` | int | 100 | Size threshold for broadcast joins (MB) |
+| `max_parallelism_multiplier` | int | 2 | Parallelism as multiple of CPU cores |
+| `target_partition_size_mb` | int | 64 | Target partition size (MB) |
+| `window_size_for_joins` | string | "1H" | Time window for distributed joins |
+| `ray_address` | string | None | Ray cluster address |
+| `use_ray_cluster` | boolean | false | Use Ray cluster mode |
```
franciscojavierarceo: shouldn't we just detect this if ray_address is configured?

docs/reference/compute-engine/ray.md
```diff
@@ -0,0 +1,383 @@
...
+# Automatically selected when feature data < 100MB
+# Features are cached in Ray's object store
+# Entities are distributed across cluster
+# Each worker gets a copy of feature data
+```
+
+### Distributed Windowed Join
+
+Used for large feature datasets:
+```python
```
franciscojavierarceo: why is this markdown as comments in python code format? same for the ohters?

docs/reference/offline-stores/ray.md
```diff
@@ -0,0 +1,510 @@
...
+| write_logged_features           | Yes       |
+
+
+| RetrievalJob Feature            | Supported |
+|----------------------------------|-----------|
+| export to dataframe             | Yes       |
+| export to arrow table           | Yes       |
+| persist results in offline store| Yes       |
+| local execution of ODFVs        | Yes       |
+| remote execution of ODFVs       | No        |
```
franciscojavierarceo: what is the different between local and remove ODFVs?

docs/reference/offline-stores/ray.md
franciscojavierarceo: +1

sdk/python/feast/infra/compute_engines/ray/feature_builder.py
haoxu0: Hmmm I don't see any specific for Ray in the implementation?

ntkathole: The main change is logic to handle the intermediate write nodes for derived views and Ray needs intermediate writes because of its distributed nature. This `_build_materialization_plan` code create a chain of write nodes where each view's write node depended on its parent views' write nodes, ensuring that intermediate results were persisted before being used by derived views.

If we remove this handling, it can leverage the parent logic and functionality will work. But we will loose scalability. Ex:
With intermediate writes:
`View A → [Write A] → View B → [Write B] → View C → [Write C]`
- If View C fails, you can restart from View B's write
- Can materialize View C later without recomputing A and B

Without intermediate writes:
`View A → View B → View C → [Write C only]`
- If View C fails, you lose all progress
- Everything must fit in memory
- Must recompute A and B every time

However, this won't affect much for small datasets, let me know what do you think ? I can update it to use parent logic.

docs/reference/offline-stores/ray.md
ntkathole: I think it may be confusing here to explicitly mention it for offline store. ODFVs can be executed remotely via transformation services not offline store. Removed it from here.

docs/reference/compute-engine/ray.md
ntkathole: That's good point, I initially though ray standalone and ray cluster can be different modes, but it works for both modes, ray handles it internally.

sdk/python/feast/infra/compute_engines/ray/feature_builder.py
haoxu0: Gotcha, I think that makes sense. And also in Spark, you can have caching/persist as well to store intermediate results. So I think in the future we can generalize it to the FeatureBuilder class as well :)

docs/reference/compute-engine/ray.md
jyejare:
```suggestion
# When materialization runs, the Ray compute engine handles:
# - Distributed data processing
# - Optimal join strategies
# - Resource management
# - Progress tracking
```
Ahh good to be explicit then ?

docs/reference/compute-engine/ray.md
jyejare: So its just the information, Can you paste this link there to see how Ray shows it?

sdk/python/feast/infra/compute_engines/ray/compute.py
```diff
@@ -0,0 +1,288 @@
...
+        views_to_delete: Sequence[
+            Union[BatchFeatureView, StreamFeatureView, FeatureView]
+        ],
+        views_to_keep: Sequence[
+            Union[BatchFeatureView, StreamFeatureView, FeatureView, OnDemandFeatureView]
+        ],
+        entities_to_delete: Sequence[Entity],
+        entities_to_keep: Sequence[Entity],
+    ):
+        """Ray compute engine doesn't require infrastructure updates."""
```
jyejare, line 104: I think any compute engine might noe want to do any infra updates. So it may make sense to remove update method OR make it non_abstract method?

wdyt?

Not blocker for this PR!

sdk/python/feast/infra/compute_engines/ray/compute.py
```diff
@@ -0,0 +1,288 @@
...
+        pass
+
+    def teardown_infra(
```
jyejare, line 107: Same here for all Compute engines, not required!

Not blocking for this PR!

sdk/python/feast/infra/compute_engines/ray/compute.py
```diff
@@ -0,0 +1,288 @@
...
+
+            # Convert to Arrow Table and write to online/offline stores
+            arrow_table = retrieval_job.to_arrow()
+
+            # Write to online store if enabled
+            if getattr(feature_view, "online", False):
+                # TODO: Implement proper online store writing with correct data format conversion
+                logger.debug(
+                    f"Online store writing not implemented yet for {arrow_table.num_rows} rows"
+                )
```
jyejare:
```suggestion
                logger.debug(
                    f"Online store writing not implemented yet for Ray offline store."
                )
```

sdk/python/feast/infra/compute_engines/ray/config.py
```diff
@@ -0,0 +1,66 @@
...
+from feast.repo_config import FeastConfigBaseModel
+
+
+class RayComputeEngineConfig(FeastConfigBaseModel):
+    """Configuration for Ray Compute Engine."""
+
+    type: Literal["ray.engine"] = "ray.engine"
+    """Ray Compute Engine type selector"""
+
+    ray_address: Optional[str] = None
```
jyejare, line 17: Maybe provide default address instead of None `ray://127.0.0.1:10001`

sdk/python/feast/infra/compute_engines/ray/config.py
```diff
@@ -0,0 +1,66 @@
...
+    """Maximum number of Ray workers. If None, uses all available cores."""
+
+    enable_optimization: bool = True
+    """Enable automatic performance optimizations."""
+
+    execution_timeout_seconds: Optional[int] = None
+    """Timeout for job execution in seconds."""
+
+    @property
+    def window_size_timedelta(self) -> timedelta:
```
jyejare, line 53: I think this can be moved to parent class. Could useful in all Compute Engines.

Non blocking comment

sdk/python/feast/infra/compute_engines/ray/feature_builder.py
```diff
@@ -0,0 +1,314 @@
...
+            feature_view=view,
+            inputs=[final_node],
+        )
+
+        self.nodes.append(node)
+        return node
+
+    def build_validation_node(self, view, input_node):
+        """Build the validation node for feature validation."""
+        # TODO: Implement validation logic
```
jyejare, line 171: Is this we want to implement yet in this PR or left for future improvements ?

sdk/python/feast/infra/compute_engines/ray/feature_builder.py
```diff
@@ -0,0 +1,314 @@
...
+            raise ValueError(f"Failed to build processing node for {view.name}")
+
+        last_node = self.build_filter_node(view, last_node)
+
+        if self._should_aggregate(view):
+            last_node = self.build_aggregation_node(view, last_node)
+        elif self._should_dedupe(view):
+            last_node = self.build_dedup_node(view, last_node)
+
+        if self._should_validate(view):
```
jyejare, line 208: I think until we implement build validation node, its good to not expose it to frontend.

sdk/python/feast/infra/compute_engines/ray/feature_builder.py
```diff
@@ -0,0 +1,314 @@
...
+                        column_info=self.get_column_info(view),
+                        is_materialization=self.is_materialization,
+                    )
+                    self.nodes.append(derived_read_node)
+
+                    # Then build the rest of the processing chain (filter, aggregate, etc.)
+                    processing_node = self._build(view, [derived_read_node])
+                else:
+                    # Parent not yet built - this shouldn't happen in topo order
+                    raise ValueError(f"Parent views for {view.name} not yet built")
```
jyejare, line 268:
```suggestion
                    raise ValueError(f"Parent write nodes for {view.name} not yet built")
```

sdk/python/feast/infra/compute_engines/ray/job.py
```diff
@@ -0,0 +1,296 @@
...
+        if not self._config.batch_engine.staging_location:
+            raise ValueError("Staging location must be set for remote storage")
+
+        try:
+            self._ensure_executed()
+            assert self._result_dataset is not None, (
+                "Dataset should not be None after execution"
+            )
+            output_uri = (
+                f"{self._config.batch_engine.staging_location}/{str(uuid.uuid4())}"
```
jyejare, line 184: it would be hard for identifying / tracing the parquet with generated UUIDs. Could we find some inique name here for better tracing purpose. Else just log the output URI for reference.

sdk/python/feast/infra/compute_engines/ray/nodes.py
```diff
@@ -0,0 +1,722 @@
...
+            elif agg.function == "min":
+                agg_dict[feature_name] = (agg.column, "min")
+            elif agg.function == "max":
+                agg_dict[feature_name] = (agg.column, "max")
+            elif agg.function == "std":
+                agg_dict[feature_name] = (agg.column, "std")
+            elif agg.function == "var":
+                agg_dict[feature_name] = (agg.column, "var")
+            else:
+                logger.warning(f"Unknown aggregation function: {agg.function}")
```
jyejare: Maybe raise error rather than warning and continuing? That way user would be informed about wrong / newly added aggr. functions avaible and warning gone unnoticed and have wrong aggregated data.

sdk/python/feast/infra/compute_engines/ray/nodes.py
```diff
@@ -0,0 +1,722 @@
...
+                    repo_config,
+                ) = serialized_artifacts.unserialize()
+
+                arrow_table = pa.Table.from_pandas(batch)
+
+                # Write to online store if enabled
+                if getattr(feature_view, "online", False):
+                    # TODO: Implement proper online store writing with correct data format conversion
+                    logger.debug(
+                        f"Online store writing not implemented yet for {len(batch)} rows"
```
jyejare:
```suggestion
                        f"Online store writing not implemented yet for Ray Compute Engine."
```

sdk/python/feast/infra/compute_engines/ray/compute.py
ntkathole: It's abstract method in base class https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/compute_engines/base.py#L71

sdk/python/feast/infra/compute_engines/ray/compute.py
ntkathole: Same as above

sdk/python/feast/infra/compute_engines/ray/feature_builder.py
ntkathole: It's abstract method

sdk/python/feast/infra/compute_engines/ray/feature_builder.py
ntkathole: I think view is correct

sdk/python/feast/infra/compute_engines/ray/compute.py
jyejare: Sorry for not being explicit. I mean we can remove this from base class itself :) But as said its not blocker for this PR.

sdk/python/feast/infra/compute_engines/ray/config.py
ntkathole: Not all compute engine needs that, Spark has it's native window().

sdk/python/feast/infra/compute_engines/ray/config.py
ntkathole: Currently When None, Ray automatically starts a local cluster and When an address is provided, it connects to an existing cluster. Changing it to ray://127.0.0.1:10001 would be problematic.

sdk/python/feast/infra/compute_engines/ray/feature_builder.py
ntkathole: abstract method and consistent with other compute engines

sdk/python/feast/infra/compute_engines/ray/job.py
ntkathole: added

sdk/python/tests/doctest/test_all.py
```diff
@@ -71,48 +71,53 @@ def test_docstrings():
...
+                    except Exception as e:
+                        successful = False
+                        failed_cases.append(
+                            (full_name, str(e) + traceback.format_exc())
+                        )
+                    finally:
+                        if teardown_function:
+                            teardown_function()
+            except DeprecationWarning:  # To catch ray.tune.automl deprecation
+                pass
```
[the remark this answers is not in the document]
ntkathole: Ray does raise warnings as exception - https://github.com/ray-project/ray/blob/master/python/ray/tune/automl/__init__.py

docs/reference/offline-stores/ray.md
franciscojavierarceo: @ntkathole can you remove the duplicate info?

sdk/python/feast/infra/compute_engines/ray/compute.py
franciscojavierarceo: @HaoXuAI what was the motivation for having this method? I agree with @jyejare but that should be outside of scope of this PR.

docs/reference/offline-stores/ray.md
ntkathole: @franciscojavierarceo It's just a reference for compute engine document, duplicate info is removed, do we want to remove reference as well ?

sdk/python/feast/infra/compute_engines/ray/compute.py
```diff
@@ -0,0 +1,288 @@
...
+            if sink_source is not None:
+                logger.debug(
+                    f"Writing derived view {feature_view.name} to sink_source: {sink_source.path}"
+                )
+
+                # Write to sink_source using Ray data
+                try:
+                    # Convert arrow table to pandas then to ray dataset
+                    df = arrow_table.to_pandas()
+                    ray_dataset = ray.data.from_pandas(df)
```
franciscojavierarceo: is this the best way to handle this? will be quite memory intensive, no?

sdk/python/feast/infra/compute_engines/ray/config.py
franciscojavierarceo: Doe spark's windowing use a time delta? I didn't think so but it's not obvious what the unit of their window function operates on: https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.Window.html

sdk/python/feast/infra/compute_engines/ray/feature_builder.py
```diff
@@ -0,0 +1,314 @@
...
+        node = RayWriteNode(
+            name="output",
+            feature_view=view,
+            inputs=[final_node],
+        )
+
+        self.nodes.append(node)
+        return node
+
+    def build_validation_node(self, view, input_node):
```
franciscojavierarceo, line 169: can we log that this is not implemented yet? maybe raise a warning that user is yoloing

sdk/python/feast/infra/compute_engines/ray/feature_builder.py
franciscojavierarceo, line 223: Why keep this implementation at the Ray level? Feesl more generic, no?

sdk/python/feast/infra/compute_engines/ray/job.py
```diff
@@ -0,0 +1,297 @@
...
+            self._ensure_executed()
+            assert self._result_dataset is not None, (
+                "Dataset should not be None after execution"
+            )
+            self._result_dataset.write_parquet(destination_path)
+            return destination_path
+        except Exception as e:
+            raise RuntimeError(f"Failed to persist dataset to {destination_path}: {e}")
+
+    def to_sql(self) -> str:
```
franciscojavierarceo, line 224: Ray doesn't support this natively, so probably we should raise that?

But apparently there is this: https://github.com/datafusion-contrib/ray-sql

which is worth some investigation.

sdk/python/feast/infra/compute_engines/ray/feature_builder.py
ntkathole: https://github.com/feast-dev/feast/pull/5526#discussion_r2255918763

franciscojavierarceo: Yeah I'd prefer we generalize this actually.

franciscojavierarceo: Oh I see. Can we make an Issue to follow up and extend this to make it more generic? As @HaoXuAI mentioned, this approach can be used in Spark as well.

sdk/python/feast/infra/compute_engines/ray/nodes.py
franciscojavierarceo: can we make an issue for that as well. I want to make sure we track the follow ups.

sdk/python/feast/infra/compute_engines/ray/nodes.py
```diff
@@ -0,0 +1,720 @@
...
+        if isinstance(entity_df, pd.DataFrame):
+            entity_dataset = ray.data.from_pandas(entity_df)
+        else:
+            entity_dataset = entity_df
+
+        join_keys = self.column_info.join_keys
+        timestamp_col = self.column_info.timestamp_column
+        requested_feats = getattr(self.column_info, "feature_cols", [])
+
+        # Check if the feature dataset contains aggregated features (from aggregation node)
```
franciscojavierarceo, line 137: nice

sdk/python/feast/infra/compute_engines/ray/nodes.py
```diff
@@ -0,0 +1,720 @@
...
+                feature_cols = sample_data[0].columns.tolist()
+            else:
+                feature_cols = (
+                    list(sample_data[0].keys())
+                    if isinstance(sample_data[0], dict)
+                    else []
+                )
+
+            # Check for aggregated feature column patterns
+            is_aggregated = any(
```
franciscojavierarceo: feels kind of brittle FWIW

sdk/python/feast/infra/offline_stores/contrib/ray_offline_store/ray.py
```diff
@@ -0,0 +1,2146 @@
...
+        return series.astype("boolean")
+    elif value_type == ValueType.STRING:
+        return series.astype("string")
+    elif value_type == ValueType.UNIX_TIMESTAMP:
+        return pd.to_datetime(series, unit="s", errors="coerce")
+    else:
+        return series.astype(target_pandas_type)
+
+
+def _convert_array_column(series: pd.Series, value_type: ValueType) -> pd.Series:
```
franciscojavierarceo: can you move this to the types file? feels like it should go there.

sdk/python/feast/infra/offline_stores/contrib/ray_offline_store/ray.py
```diff
@@ -0,0 +1,2146 @@
...
+                return np.empty(0, dtype=object)
+            else:
+                return np.empty(0, dtype=target_dtype)  # type: ignore
+        else:
+            return item
+
+    return series.apply(convert_array_item)
+
+
+class RayOfflineStoreConfig(FeastConfigBaseModel):
```
franciscojavierarceo, line 309: can you link to the documentation where users can easily find what these things mean?

sdk/python/feast/infra/offline_stores/contrib/ray_offline_store/ray.py
```diff
@@ -0,0 +1,2146 @@
...
+    window_size_for_joins: Optional[str] = "1H"
+
+    # Logging settings
+    enable_ray_logging: Optional[bool] = False
+
+    # Ray configuration for resource management (memory, CPU limits)
+    ray_conf: Optional[Dict[str, Any]] = None
+
+
+class RayResourceManager:
```
franciscojavierarceo, line 337: would link to the appropriate Ray documentation here too

sdk/python/feast/transformation/pandas_transformation.py
```diff
@@ -19,36 +19,61 @@
 class PandasTransformation(Transformation):
     def __new__(
         cls,
-        udf: Callable[[Any], Any],
-        udf_string: str,
+        udf: Optional[Callable[[Any], Any]] = None,
```
franciscojavierarceo, line 22: why make this optional?

franciscojavierarceo (commented):
this looks great some small nits but i think we're almost there. can you make some issues to track the follow up here so we don't forget and I'll merge 👍

sdk/python/feast/transformation/pandas_transformation.py
ntkathole: It's not optional in real use, we are still maintaining strict validation (lines 72-75 enforce that they can't actually be None). We have to make it optional to ensures that PandasTransformation objects can be properly reconstructed across Ray workers without constructor failures.

sdk/python/feast/infra/compute_engines/ray/compute.py
ntkathole: I remember I did it like that to fix some empty data handling with tests, but looks like converting directly from arrow to dataset working fine as well with latest code. Thanks.

sdk/python/feast/infra/compute_engines/ray/config.py
ntkathole: https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/compute_engines/spark/nodes.py#L127

sdk/python/feast/infra/compute_engines/ray/nodes.py
ntkathole: agree, simplified the logic to just detect aggregated flag from aggregation node.

sdk/python/feast/infra/compute_engines/ray/compute.py
haoxu0: For backward compatible. the old materialization engine have these two methods.
In some way i think it make sense, people can use compute engine to update infra directly

franciscojavierarceo:
Thanks for this @ntkathole !!!

PR #5572, opened 2025-08-19
fix: Consistency in allow_cache and savedDatasets api

ntkathole:

This PR adds :
- allow_cache param in GetFeatureRequest API to be consistent with all other apis.
- Rename saved_datasets to savedDatasets in response to be consistent with other apis.


ntkathole:
cc @franciscojavierarceo

sdk/python/tests/unit/api/test_api_rest_registry.py
```diff
@@ -527,11 +527,11 @@ def test_saved_datasets_via_rest(fastapi_test_app):
     response = fastapi_test_app.get("/saved_datasets?project=demo_project")
     assert response.status_code == 200
     response_data = response.json()
-    assert "saved_datasets" in response_data
-    assert isinstance(response_data["saved_datasets"], list)
-    assert len(response_data["saved_datasets"]) == 1
+    assert "savedDatasets" in response_data
```
franciscojavierarceo, line 530: so was this always wrong?

sdk/python/feast/api/registry/rest/saved_datasets.py
```diff
@@ -119,7 +119,7 @@ def list_saved_datasets(
         saved_datasets = response.get("savedDatasets", [])
 
         result = {
-            "saved_datasets": saved_datasets,
```
franciscojavierarceo, line 122: oh i see. why this changed?

Issue #5551, opened 2025-08-02
Milvus Online Store Dimension Mismatch Error in Push API and Materialization


abhijeet-dhumal:
## Summary
Feast's Milvus online store integration has a critical dimension mismatch bug that affects both the push API and materialization approaches. When storing embeddings with correct dimensions (384), Feast internally transforms the data incorrectly, causing Milvus to reject the data with dimension errors.

## Environment
- **Feast version**: 0.51.0
- **Python version**: 3.12.11
- **pymilvus version**: 2.3.0+
- **OS**: macOS (Darwin 24.5.0)
- **Milvus**: milvus-lite (via `path: data/online_store.db`)

### Error Message
```
ERROR:pymilvus.decorators:RPC error: [upsert_rows], <MilvusException: (code=65535, message=the length(7695) of float data should divide the dim(384): )>
```

### Expected Behavior
- Input: 5 embeddings × 384 dimensions = 1920 total elements
- Feast should store these embeddings correctly in Milvus
- Expected elements sent to Milvus: 1920

### Actual Behavior
- Input: 5 embeddings × 384 dimensions = 1920 total elements  
- Feast transforms this to 7695 elements (factor of ~4x)
- Milvus rejects the data because 7695 ÷ 384 = 20.04... (not integer)

### 1. Feature Store Configuration
```yaml
# feast_feature_repo/feature_store.yaml
project: rag
provider: local
registry: data/registry.db
online_store:
  type: milvus
  path: data/online_store.db
  vector_enabled: true
  embedding_dim: 384
  index_type: "FLAT"
  metric_type: "L2"
offline_store:
  type: file
entity_key_serialization_version: 3
auth:
  type: no_auth
```

### 2. Feature Definitions
```python
from feast import Entity, FeatureView, Field, FileSource, PushSource
from feast.types import Array, Float32, String, Int64
from feast.value_type import ValueType
from datetime import timedelta

document = Entity(
    name="document_id",
    value_type=ValueType.STRING,
    description="Unique identifier for document chunks"
)

document_embeddings_source = FileSource(
    name="document_embeddings_source",
    path="data/document_embeddings.parquet",
    timestamp_field="event_timestamp",
    created_timestamp_column="created_timestamp",
)

document_embeddings_push_source = PushSource(
    name="document_embeddings_push_source",
    batch_source=document_embeddings_source,
)

document_embeddings = FeatureView(
    name="document_embeddings",
    entities=[document],
    ttl=timedelta(days=365),
[...]

Issue #5372, opened 2025-05-20
Support Image Search


franciscojavierarceo:
**Is your feature request related to a problem? Please describe.**
https://milvus.io/docs/image_similarity_search.md

**Describe the solution you'd like**
https://milvus.io/docs/image_similarity_search.md

**Describe alternatives you've considered**
n/a

**Additional context**

PR #5577, opened 2025-08-22
feat: Added support for image search

ntkathole:

Added support for image search combining text and image queries and fixed Milvus binary data handling with base64 encoding.

Users can now define `Field(name="image_bytes", dtype=ImageBytes)`

This PR also fixes the vector field inconsistency issue on online_write_batch with milvus. 

- IMAGE_BYTES Type Support
```
sdk/python/feast/value_type.py: Added IMAGE_BYTES = 21
sdk/python/feast/types.py: Added ImageBytes type and mappings
sdk/python/feast/on_demand_feature_view.py: Added sample JPEG data for testing
```
- Image Search Functionality
```
sdk/python/feast/feature_store.py: Enhanced retrieve_online_documents_v2 with image search parameters
sdk/python/feast/vector_store.py: Added query_image_bytes support
sdk/python/feast/image_utils.py: New file with ImageFeatureExtractor and combine_embeddings
```

Fixes https://github.com/feast-dev/feast/issues/5372 and Fixes https://github.com/feast-dev/feast/issues/5551


sdk/python/feast/feature_store.py
```diff
@@ -2239,6 +2239,12 @@ def retrieve_online_documents_v2(
         query: Optional[List[float]] = None,
         query_string: Optional[str] = None,
         distance_metric: Optional[str] = "L2",
+        query_image_bytes: Optional[bytes] = None,
```
franciscojavierarceo, line 2242: is this the right way to do this? See I would have updated the proto values to support ImageBytes like we support PDFBytes and then just queried embedding the standard way. 

You can then enrich the image vector embeddings with text and semantic embeddings and then allow a composite search of both. In that sense, you still would use 1 query but searching across multiple vectors.

franciscojavierarceo: Actually I'm realizing my mistake here. ImageBytes is required for retrieval, what you here is appropriate for image search when the image is passed in for the query.

ntkathole: Added IMAGE_BYTES as well 👍

infra/templates/README.md.jinja2
```diff
@@ -36,7 +36,7 @@ Feast allows ML platform teams to:
 Please see our [documentation](https://docs.feast.dev/) for more information about the project.
 
 ## 📐 Architecture
-![](docs/assets/feast_marchitecture.png)
+![](https://raw.githubusercontent.com/feast-dev/feast/master/docs/assets/feast_marchitecture.png)
```
franciscojavierarceo: why swap these?

sdk/python/feast/infra/online_stores/milvus_online_store/milvus.py
```diff
@@ -43,6 +44,7 @@
 
 PROTO_TO_MILVUS_TYPE_MAPPING: Dict[ValueType, DataType] = {
     PROTO_VALUE_TO_VALUE_TYPE_MAP["bytes_val"]: DataType.VARCHAR,
+    ValueType.IMAGE_BYTES: DataType.VARCHAR,  # IMAGE_BYTES serializes as bytes_val
```
franciscojavierarceo:
```suggestion
    ValueType.IMAGE_BYTES: DataType.VARCHAR,
```

sdk/python/feast/infra/online_stores/milvus_online_store/milvus.py
```diff
@@ -247,7 +249,7 @@ def online_write_batch(
     ) -> None:
         self.client = self._connect(config)
         collection = self._get_or_create_collection(config, table)
-        vector_cols = [f.name for f in table.features if f.vector_index]
+        vector_cols = [f.name for f in table.schema if f.vector_index]
```
franciscojavierarceo: ?

infra/templates/README.md.jinja2
ntkathole: Template change was missed in https://github.com/feast-dev/feast/pull/5579

sdk/python/feast/infra/online_stores/milvus_online_store/milvus.py
ntkathole: Removed. While adding a new test, I used entity as a vector field, to fix it, this check was added but no need of it.

sdk/python/feast/image_utils.py
```diff
@@ -0,0 +1,279 @@
...
+        except Exception as e:
+            raise ValueError(f"Failed to extract embedding from image: {e}")
+
+    def get_embedding_dimension(self) -> int:
+        """
+        Get the dimension of embeddings produced by this model.
+        Returns:
+            Integer dimension of the embedding vector
+        """
+        dummy_input = torch.randn(1, 3, 224, 224)  # Standard input size
```
franciscojavierarceo: this is brittle fyi

sdk/python/feast/image_utils.py
```diff
@@ -0,0 +1,279 @@
...
+                raise ValueError(f"Failed to preprocess image: {e}")
+
+        batch_tensor = torch.stack(images)
+
+        with torch.no_grad():
+            outputs = self.model(batch_tensor)
+
+        for output in outputs:
+            feature_vector = output.numpy()
+            normalized = normalize(feature_vector.reshape(1, -1), norm="l2")
```
franciscojavierarceo, line 155: feel like l2 should be configurable but probably not needed yet

sdk/python/feast/image_utils.py
```diff
@@ -0,0 +1,279 @@
...
+        text_padded = text_embedding + [0.0] * (max_dim - len(text_embedding))
+        image_padded = image_embedding + [0.0] * (max_dim - len(image_embedding))
+
+        combined = [(t + i) / 2.0 for t, i in zip(text_padded, image_padded)]
+        return combined
+
+    else:
+        raise ValueError(
+            f"Unknown combination strategy: {strategy}. "
+            f"Supported strategies: weighted_sum, concatenate, average"
```
franciscojavierarceo: small nit, technically you should just make the strategy an enum so the valueerror can be maintained easier

franciscojavierarceo (approved):
some small nits but this lgtm and i'd like to include it in the next release as this is a killer new feature. great work as always @ntkathole!

sdk/python/feast/image_utils.py
ntkathole: right, most of the usecase get covered via l2, we can expand as needed

sdk/python/feast/image_utils.py
ntkathole: Made the changes

PR #5580, opened 2025-08-25
fix: Updated python packages conflicting with kserve dependencies

aniketpalu:

Updated following python packages conflicting with kserve dependencies.
1. uvicorn dependency version range updated to 0.30.6 to 0.34.0
2. kubernetes dependency version range updated to 20.13.0 to 23.3.0

[RHOAIENG-31060](https://issues.redhat.com/browse/RHOAIENG-31060)

# Misc

https://github.com/issues/created?issue=kserve%7Ckserve%7C4646

### DIFF
--- a/infra/feast-operator/docs/namespace-registry.md
+++ b/infra/feast-operator/docs/namespace-registry.md
@@ -1,0 +1,58 @@
+# Feast Namespace Registry
+
+## Overview
+
+The Feast Namespace Registry is a feature that automatically creates and maintains a centralized ConfigMap containing information about all Feast feature store instances deployed by the operator. This enables dashboard applications and other tools to discover and connect to Feast instances across different namespaces.
+
+## Implementation Details
+
+1. **ConfigMap Creation**: The operator creates a ConfigMap in the appropriate namespace:
+   - **OpenShift AI**: `redhat-ods-applications` namespace (or DSCi configured namespace)
+   - **Kubernetes**: `feast-operator-system` namespace
+
+2. **Access Control**: A RoleBinding is created to allow `system:authenticated` users to read the ConfigMap
+
+3. **Automatic Registration & Cleanup**: When a new feature store instance is created, it automatically registers its namespace and client configuration in the ConfigMap. When deleted, it automatically removes its entry from the ConfigMap
+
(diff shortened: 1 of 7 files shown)


ntkathole:
@aniketpalu Please update the setup.py as well to match pyproject.toml

franciscojavierarceo:
while we're at it, can you add a test to make sure these are in sync? doesn't have to be overly elaborate but probably we should catch this sooner.

sdk/python/tests/unit/infra/test_dependency_conflicts.py
```diff
@@ -0,0 +1,32 @@
...
+logger.setLevel(logging.DEBUG)
+
+
+class TestDependencyConflicts:
+    def test_install_kserve_with_feast(self):
+        """Test installing KServe in the current environment where Feast is already installed.
+        Ensures no dependency conflict errors occur.
+        """
+        # Command to install KServe
+        command = [sys.executable, "-m", "pip", "install", "kserve==0.15.2"]
```
franciscojavierarceo, line 15: oh i actually probably just meant that the pyproject and requirements were in sync

jyejare (approved):
LGTM

sdk/python/tests/unit/infra/test_dependency_conflicts.py
aniketpalu: Hi @franciscojavierarceo 

This test case is not that straightforward.  I will take care of this in separate PR.

aniketpalu: @franciscojavierarceo 

[RHOAIENG-33926](https://issues.redhat.com/browse/RHOAIENG-33926)

franciscojavierarceo: sounds good.

franciscojavierarceo:
unit tests are failing, please resolve. unit tests shouldn't be flakey.

sdk/python/tests/unit/infra/test_dependency_conflicts.py
franciscojavierarceo, line 15: this specific version feels kind of brittle

PR #5587, opened 2025-08-31
fix: Fix the name typo in the local registry file example.

shuchu:
 the current document has a typo in the "local registry" example page.
A reasonable file named is "registry.db"
In the current doc, the file name is "registry.pb"

Fix the typo in the example page.

# Misc


shuchu:
seems .pb is used in many places. nvm

PR #5595, opened 2025-09-03
fix: Correct column list polars materialization engine

danbaron63:
Currently materialization fails when using the `PolarsBackend` as the `columns` method has no implementation. Worse still, rather than raising a `NotImplementedError` it fails silently by returning nothing which gets interpreted as `None`.

This provides a valid implementation that allows the `feast materialize ...` command to run successfully.

N/a

# Misc
N/a

### DIFF
--- a/sdk/python/feast/infra/compute_engines/local/backends/polars_backend.py
+++ b/sdk/python/feast/infra/compute_engines/local/backends/polars_backend.py
@@ -7,8 +7,8 @@ from feast.infra.compute_engines.local.backends.base import DataFrameBackend
 
 
 class PolarsBackend(DataFrameBackend):
-    def columns(self, df):
-        pass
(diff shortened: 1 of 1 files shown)


haoxu0:
Thanks for contributing!

haoxu0:
Need to fix the DCO

danbaron63:
> Need to fix the DCO

Apologies! Recommitted the last commit, should pass now 🤞

danbaron63:
@HaoXuAI any idea why the integration tests are failing for this change? I feel this should be a pretty trivial change 🤔 Thanks!

haoxu0:
> @HaoXuAI any idea why the integration tests are failing for this change? I feel this should be a pretty trivial change 🤔 Thanks!

seems to be a flaky test, rerunning now.

haoxu0:
Thanks for your contributing!

Issue #5601, opened 2025-09-09
Feature Registry not refreshing


scottmelhop:
## Expected Behavior 
When a new object is inserted into the registry, the UI should show it.

## Current Behavior
When applying new objects to the Feature Registry the cache is not refreshed, so in the UI they new objects are not shown

## Steps to reproduce

Config:
```
project: feast_experiment
registry:
    registry_type: sql
    path: postgresql+psycopg://registry_user:registry_pass@feast_registry_db:5432/feast_registry
    sqlalchemy_config_kwargs:
        echo: false
        pool_pre_ping: true
provider: local
online_store:
    type: redis
    connection_string: feast_online_store:6379
offline_store:
    type: postgres
    host: feast_offline_db
    port: 5432
    user: offline_user
    password: offline_pass
    database: feast_offline
entity_key_serialization_version: 3
auth:
    type: no_auth
```

Docker Compose:
```
services:
  feature_server:
    container_name: feast_ui
    image: quay.io/feastdev/feature-server:latest
    command: feast --log-level=DEBUG -c /feature_repo ui -h 0.0.0.0
    volumes:
      - ./configs/docker_compose:/feature_repo
    ports:
     - 8888:8888
  feast_offline_db:
    image: postgres:15.6
    container_name: feast_offline_db
    restart: unless-stopped
    environment:
      POSTGRES_USER: offline_user
      POSTGRES_PASSWORD: offline_pass
      POSTGRES_DB: feast_offline
    ports:
      - "5428:5432"
    volumes:
      - ./backend/offline_db:/var/lib/postgresql/data

  feast_registry_db:
    image: postgres:15.6
    container_name: feast_registry_db
    restart: unless-stopped
    environment:
      POSTGRES_USER: registry_user
      POSTGRES_PASSWORD: registry_pass
      POSTGRES_DB: feast_registry
    ports:
      - "5439:5432"
    volumes:
      - ./backend/regitry_db:/var/lib/postgresql/data

  feast_online_store:
      image: redis:7.2
      container_name: feast_online_store
      restart: unless-stopped
      ports:
        - "6375:6379"
      volumes:
        - ./backend/online_db:/data

```

### Specifications

- Version: feast==0.53.0
- Platform: amd64
- Subsystem: linux

## Possible Solution

Not sure what is happening in the refresh logic.

scottmelhop:
[...]

PR #5604, opened 2025-09-10
fix: Fixed Registry Cache Refresh Issues

ntkathole:

This PR fix the issues with the Feast registry cache refresh mechanism:

- Fixed Circular Dependency in sql.py and snowflake.py. The `proto()` method was calling cached methods (list_*) which would trigger cache refresh, creating a circular dependency.
- Improved the readability of logic to handle cache expiry. Added `is_cache_expired()` function with ttl handling. 
- Added test coverage for cache ttl. 

While testing the fix, identified issue with metric api, that's also fixed. 

#5601


sdk/python/feast/api/registry/rest/metrics.py
```diff
@@ -321,20 +319,43 @@ async def recently_visited(
...
-            visits_json = (
-                server.registry.get_project_metadata(project_val, key)
-                if server
-                else None
-            )
-            visits = json.loads(visits_json) if visits_json else []
-        except Exception:
-            visits = []
+        visits = []
+        if project:
```
ntkathole, line 324: Earlier logic was not considering all projects while showing metrics. Fixed it by listing projects first and get the entries for all projects.

franciscojavierarceo:
Test are failing

PR #5617, opened 2025-09-15
fix: Fixed tls issue when running both grpc and rest servers

ntkathole:

When both gRPC and REST registry servers are enabled with TLS, the Feast operator was generating separate TLS certificates for each service, causing hostname validation failures. The gRPC client would receive a "Peer name not in peer certificate" error because it was trying to connect to the gRPC service hostname but the certificate only contained the REST service hostname.

Error:

```
_InactiveRpcError: <_InactiveRpcError of RPC that terminated with: 
status = StatusCode.UNAVAILABLE 
details = "failed to connect to all addresses; last error: UNKNOWN: ipv4:172.xx.xx.xx:443: 
Peer name feast-example-registry.default.svc.cluster.local is not in peer certificate"
```

This PR implemented a unified TLS certificate approach using Subject Alternative Names (SANs). When both gRPC and REST are enabled, use the gRPC service name as the primary certificate and add the REST service hostname as a Subject Alternative Name in the certificate.


franciscojavierarceo (approved):
stamp

PR #5618, opened 2025-09-16
feat: Add support for extra options for stream source

Prathap-P:

# Misc


franciscojavierarceo:
@Prathap-P can you fix the DCO? Just have to click the UI.

Prathap-P:
This pr exposes runtime options that users can give before ingesting stream data.

- Our team have certificate way of authn/authz set for kafka topic and currently feast doesn't accept those options while registering a kafka source.
- For ex: "kafka.security.protocol", "kafka.ssl.truststore.location", "kafka.ssl.truststore.password", "kafka.ssl.keystore.location", "kafka.ssl.keystore.password", "kafka.ssl.key.password"
- So I thought we can expose an optional stream source options that users can provide while ingesting, so that they dont have to register again multiple times when options change.



# Misc
I think this addition will be essential as spark supports numerous options which are used in real world scenarios.

Prathap-P:
Hi @franciscojavierarceo DCO check is passed.
Let me know if you need anything from my side.

Issue #5624, opened 2025-09-20
Do not require event_timestamp for materialization


franciscojavierarceo:
**Is your feature request related to a problem? Please describe.**
Do not require event_timestamp for materialization

`materialize` shouldn't always require a timetsamp and the feature views shouldn't require a date to materialize as well.

**Describe the solution you'd like**
We should allow users to override with a flag `--disable-event-timestamp` should set the default timestamp for materilalization to `datetime.now()`.

**Describe alternatives you've considered**
N/A

**Additional context**

PR #5625, opened 2025-09-22
feat: Enable ingestion without event timestamp

franciscojavierarceo:
This PR enables materialization without a timestamp by just using the current timestamp. 

As updated in the documentation, this can be run with:

```bash
feast materialize --disable-event-timestamp
```

Fixes https://github.com/feast-dev/feast/issues/5624

# Misc

### DIFF
--- a/sdk/python/feast/api/registry/rest/search.py
+++ b/sdk/python/feast/api/registry/rest/search.py
@@ -167,6 +167,7 @@ def get_search_router(grpc_handler) -> APIRouter:
                             "name": feature.get("name", ""),
                             "description": feature.get("description", ""),
                             "project": current_project,
+                            "featureView": feature.get("featureView", ""),
                             "tags": feature.get("tags", {}),
                         }
                     )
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -109,11 +109,26 @@ print(training_df.head())
...
+CURRENT_TIME=$(date -u +"%Y-%m-%dT%H:%M:%S")
+feast materialize 2021-04-12T00:00:00 $CURRENT_TIME
+```
+
+**Option 3: Simple materialization without timestamps**
+```commandline
+feast materialize --disable-event-timestamp
+```
+
+The `--disable-event-timestamp` flag allows you to materialize all available feature data using the current datetime as the event timestamp, without needing to specify start and end timestamps. This is useful when your source data lacks proper event timestamp columns.
```
ntkathole, line 130: I think we do not need this flag, we can keep default end time as current time and start time as (current time - 24hr/a month) ? 
Same for materialize-incremental command to have end date as current time by default.

franciscojavierarceo: my concern is the significant change that has for users, right? making it optional would silently ingest data for users without telling them about it. i'd rather do more code work to make them opt in rather than have users accidentally ingest data.

ntkathole: That's fair 👍

PR #5626, opened 2025-09-22
Add 'featureView' in global search api result for features.

aniketpalu:

When search api is called & result has features, UI needs workaround to fetch information of feature without featureView name access. To overcome this, added 'featureView' key-value pair in results of feature. 

[RHOAIENG-34599](https://issues.redhat.com/browse/RHOAIENG-34599)

# Misc


ntkathole:
Please include assertion of feature view in some existing search test.

Issue #5633, opened 2025-09-25
Add Kuberay Support for Ray Offline Store and Compute engine


ntkathole:
**Describe the solution you'd like**

Enable seamless integration with Kubernetes-managed Ray clusters through the [CodeFlare SDK](https://github.com/project-codeflare/codeflare-sdk/). This enhancement is for Feast users running on Kubernetes and wants to use Kuberay.

Currently, Feast's Ray integration supports:

- Local Ray clusters
- Remote Ray clusters (via ray_address)

We need to add support for Kubernetes-managed Ray clusters (KubeRay):

- Authenticate with CodeFlare SDK using token
- Generate TLS certificates for cluster access
- Connect to KubeRay cluster via cluster.cluster_uri()
- Initialize Ray with runtime environment
 
Example:

```
offline_store:
  type: ray
  use_kuberay: true
  kuberay_conf:
    cluster_name: "ray-cluster"
    namespace: "ray-system"

batch_engine:
  type: ray.engine
  use_kuberay: true
  kuberay_conf:
    cluster_name: "ray-cluster"
    namespace: "ray-system"
```

PR #5637, opened 2025-09-26
feat: Added RemoteDatasetProxy that executes Ray Data operations remotely

ntkathole:

This PR adds support to run Ray Datasets operations on remote Kuberay cluster. Implemented `RemoteDatasetProxy` class, that wraps all Ray Dataset operations in `@ray.remote` functions, ensuring they execute on KubeRay cluster workers.

https://github.com/feast-dev/feast/issues/5633


ntkathole:
> This is fine, the syntax of is_ray_dataset instead of is instance is ugly though

Tried to make it little better with `is_ray_data(data)`

PR #5638, opened 2025-09-26
feat: Added feast[client] dependency group to support feast client

aniketpalu:

Added feast[client] dependency group to support feast client

[RHOAIENG-33709](https://issues.redhat.com/browse/RHOAIENG-33709)

# Misc


tokoko:
Most of this are already required dependencies. What's the point of adding an extra unless you also want to drop them from required?

PR #5640, opened 2025-09-27
feat: Added codeflare-sdk to requirements

ntkathole:

This PR adds codeflare-sdk to requirements for managing ray.  Codeflare-sdk requires python 3.11+ and kubernetes > 27.2.0.

https://github.com/feast-dev/feast/issues/5633

### DIFF
--- a/docs/reference/beta-on-demand-feature-view.md
+++ b/docs/reference/beta-on-demand-feature-view.md
@@ -55,7 +55,8 @@ When defining an ODFV, you can control when the transformation is applied using 
 #### Example 1: On Demand Transformation on Read Using Pandas Mode
 
 ```python
-from feast import Field, RequestSource, on_demand_feature_view
+from feast import Field, RequestSource
+from feast.on_demand_feature_view import on_demand_feature_view
 from feast.types import Float64, Int64
 import pandas as pd
 


franciscojavierarceo:
I thought the codeflare SDK was being dropped

ntkathole:
> I thought the codeflare SDK was being dropped

It is due to be renamed, but it's not going away.

franciscojavierarceo (approved):
lgtm

franciscojavierarceo:
unit tests failing

haoxu0:
Hey is this necessary dependency?

franciscojavierarceo:
@ntkathole can we outline the benefits of having the codeflare SDK here?

haoxu0:
Or maybe create another dep like "Ray-clodeflakre"

ntkathole:
> Hey is this necessary dependency?

Codeflare-sdk is required for handling authentication, certs and making connection to Kuberay. I am going to add support for this in upcoming PRs by adding codeflare wrapper above current implementation. This will be completely optional as a end user, User can use simply use ray address to connect to remote cluster as well, if ray cluster do not need authentication or tls certs. 

> can we outline the benefits of having the codeflare SDK here?

Codeflare-sdk allows us:

- To find existing KubeRay cluster and connect without need to perform manual steps such as setting up tls certs configuration. 
```
from codeflare_sdk import TokenAuthentication
from codeflare_sdk import generate_cert, get_cluster
```

- User don't need to know cluster details, just name and namespace of Ray cluster. 
- Eliminates need to manually deploy and configure Ray clusters in Kubernetes. 

I will add this in our documentation once implementation is done. 

> Or maybe create another dep like "Ray-clodeflakre"

We can do this, but it could be redundant. I am thinking anyhow codeflare-sdk will pull ray as it's dependency and usually ray needs same version for both server and client. If we did it like this, we don't need to maintain the ray version and let codeflare-sdk handle the compatibility.

franciscojavierarceo:
That makes sense to me.

Please make sure to document the benefits on the usage to the users.

@HaoXuAI codeflare and the auth @ntkathole mentioned is how we recommend users securely use Ray clusters at Red Hat. See here: https://developers.redhat.com/articles/2025/07/24/submit-remote-rayjobs-ray-cluster-codeflare-sdk

haoxu0:
> That makes sense to me.
> 
> Please make sure to document the benefits on the usage to the users.
> 
> @HaoXuAI codeflare and the auth @ntkathole mentioned is how we recommend users securely use Ray clusters at Red Hat. See here: https://developers.redhat.com/articles/2025/07/24/submit-remote-rayjobs-ray-cluster-codeflare-sdk

I don't have double on the benefits. Just concerned that the codeflare is not available in our team's artifacts. Though we don't use Ray at the moment but possibly in the future

franciscojavierarceo:
@HaoXuAI it's also open source so can be installed. It was a projected started by IBM I believe.

PR #5650, opened 2025-10-05
feat: Support table format

HaoXuAI:

# Misc


protos/feast/core/DataSource.proto
```diff
@@ -231,6 +231,9 @@ message DataSource {
 
     // Date Format of date partition column (e.g. %Y-%m-%d)
     string date_partition_column_format = 5;
+
+    // Table Format (e.g. iceberg, delta, etc)
+    string table_format = 6;
```
haoxu0: TODO, create TableFormat proto, consolidate with FileFormat proto

tokoko:
+1 on the inclusion of all 3 formats. Still I think we might be able to better design data-source side such that data source definitions don't tie the sources to specific offline stores. For example right now `FileSource` already supports specifying delta as a format... I don't like that design either because delta and especially iceberg might require more configuration that plain files, but the upside in that design was that Source class wasn't tied to a single offline store and it is actually usable from both dask and duckdb offline stores. Hiding formats behind SparkSource will mean that when a user stores a data source instance in the registry only spark offline store can be used to retrieve the data.

I think we can have best of both worlds if we instead go for adding all these formats as separate independent data sources (`DeltaSource`, `IcebergSource`, `HudiSource`) and then simply teach spark engine (and in the future others) how to read it.

sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark_source.py
```diff
@@ -58,7 +60,9 @@ def __init__(
             table: The name of a Spark table.
             query: The query to be executed in Spark.
             path: The path to file data.
-            file_format: The format of the file data.
+            file_format: The underlying file format (parquet, avro, csv, json).
```
franciscojavierarceo, line 62: why not consolidate now?

franciscojavierarceo:
>I think we can have best of both worlds if we instead go for adding all these formats as separate independent data sources (DeltaSource, IcebergSource, HudiSource) and then simply teach spark engine (and in the future others) how to read it.

+1

haoxu0:
@franciscojavierarceo @tokoko consolidation with FilleFormat and new data sources could break the backward compatibility, so I want to do it pace by pace.

franciscojavierarceo:
That makes sense

tokoko:
@HaoXuAI Why would new data sources break backwards compatibility though?

haoxu0:
> @HaoXuAI Why would new data sources break backwards compatibility though?

There will be some proto changes, no 100% sure if there will be API changes exposed to users but I think might be the case

haoxu0:
@franciscojavierarceo @ntkathole mind take a look

haoxu0:
> @HaoXuAI i don't see use actually using or testing Spark Table, Iceberg, or Hudi format's outside of our definitions, can you add that?
> 
> Can you also add documentation that these formats are now supported?
> 
> Otherwise lgtm.

haoxu0:
> > @HaoXuAI i don't see use actually using or testing Spark Table, Iceberg, or Hudi format's outside of our definitions, can you add that?
> > Can you also add documentation that these formats are now supported?
> > Otherwise lgtm.

Gonna update to add the TableFormat proto in the next PR, after that I'll add the docs. And I think the test will need to be changed as well.

haoxu0:
@franciscojavierarceo mind take another look?

PR #5655, opened 2025-10-10
feat: Add HDFS as a feature registry

chimeyrock999:
1. Implemented a new feature registry that reads and writes feature data directly on HDFS using `pyarrow.fs.HadoopFileSystem`.
2. Added integration tests using two containers (1 Namenode and 1 Datanode) via Testcontainers and [bde2020](https://hub.docker.com/u/bde2020)’s Hadoop images.
3. To use `pyarrow.fs.HadoopFileSystem` in client or CI environments, you must properly configure the environment by following the [PyArrow HDFS documentation](https://arrow.apache.org/docs/python/filesystems.html#hadoop-distributed-file-system-hdfs).
.


chimeyrock999:
> Thank you for this contribution!!
> 
> Can we update the registry documentation to include this?

Thank @franciscojavierarceo. I’ve fixed the integration test and updated the docs for the HDFS registry.

.github/workflows/pr_local_integration_tests.yml
```diff
@@ -42,6 +42,8 @@ jobs:
           enable-cache: true
       - name: Install dependencies
         run: make install-python-dependencies-ci
+      - name: Instal Hadoop dependencies
```
ntkathole:
```suggestion
      - name: Install Hadoop dependencies
```

Makefile
```diff
@@ -87,6 +87,11 @@ install-python-dependencies-ci: ## Install Python CI dependencies in system envi
 	uv pip sync --system sdk/python/requirements/py$(PYTHON_VERSION)-ci-requirements.txt
 	uv pip install --system --no-deps -e .
 
+# Used in github actions/ci
+install-hadoop-dependencies-ci: ## Install Hadoop dependencies
+	wget -q https://dlcdn.apache.org/hadoop/common/hadoop-3.4.2/hadoop-3.4.2.tar.gz
```
ntkathole: Should we need to document this for users trying hdfs registry ?

`To use HDFS registry, you need Hadoop 3.4+ installed and HADOOP_HOME set`

docs/reference/registries/hdfs.md
```diff
@@ -0,0 +1,39 @@
...
+
+An example of how to configure this would be:
+
+{% code title="feature_store.yaml" %}
+```yaml
+project: feast_hdfs
+registry:
+  path: hdfs://[YOUR NAMENODE HOST]:[YOUR NAMENODE PORT]/[PATH TO REGISTRY]/registry.pb
+  cache_ttl_seconds: 60
+online_store: null
```
ntkathole, line 38: remove instead of specifying null

.github/workflows/pr_local_integration_tests.yml
```diff
@@ -42,6 +42,8 @@ jobs:
...
+        run: make install-hadoop-dependencies-ci
```
ntkathole, line 51: This downloads ~400 MB eveytime, we should cache it so that it won't download everytime 

```
- name: Cache Hadoop tarball
        uses: actions/cache@v4
        with:
          path: ~/hadoop-3.4.2.tar.gz
          key: hadoop-3.4.2
```

docs/reference/registries/hdfs.md
```diff
@@ -0,0 +1,39 @@
...
+## Example
+
+An example of how to configure this would be:
+
+{% code title="feature_store.yaml" %}
+```yaml
+project: feast_hdfs
+registry:
+  path: hdfs://[YOUR NAMENODE HOST]:[YOUR NAMENODE PORT]/[PATH TO REGISTRY]/registry.pb
+  cache_ttl_seconds: 60
```
ntkathole, line 37: cache_ttl_seconds seems not supported yet with your implementation

chimeyrock999: I think `cache_ttl_seconds` might already be handled at the `Registry` class (`feast/infra/registry/registry.py`). Given that `HDFSRegistryStore` is already registered in `REGISTRY_STORE_CLASS_FOR_SCHEME`, the caching should work automatically. Not entirely sure though — does that sound correct?

Makefile
chimeyrock999: I’ve tested this with Hadoop 3.3.6 as well, and it worked fine — so I think we can safely say it requires Hadoop 3.3+ instead of 3.4+.

PR #5661, opened 2025-10-13
chore: Loosen pyarrow

### LINKED ISSUE #5609: The dependency "pyarrow<=17.0.0"  does not work with Python 3.13
## Expected Behavior 

Pyarrow can be compiled and installed by "uv pip install" command

## Current Behavior

failed with errors during the cmake configuration.

### Specifications

- Version: 0.53.0
- Platform: Linux/Debian 13
- Subsystem:

## Possible Solution
1, It works well with Python 3.12.11
2, evaluate the possible of adapting higher version of Pyarrow

dlovell:
This PR loosens the pyarrow requirement to allow for more recent pyarrow versions

fixes #5609


franciscojavierarceo:
hmm looks like macos-14 doesn't like it @ntkathole can you provide some guidance here?

dlovell:
i'm not sure i created my venv with `make install-python-dependencies-dev`

i'll start over with a venv created that way

dlovell:
looks like `nvidia-*` causes macos to fail and `nvidia-*` comes from `torch` / `image` comes from `ci`

ntkathole:
> looks like `nvidia-*` causes macos to fail and `nvidia-*` comes from `torch` / `image` comes from `ci`

The requirement file included in PR has nvidia-* dependencies which is only meant for GPU supported systems. You need to make sure `make lock-python-dependencies-all` do not include those dependencies, I think probably using `PIP_EXTRA_INDEX_URL=https://download.pytorch.org/whl/cpu `

ntkathole:
Also, fyi - we have APACHE_ARROW_VERSION="17.0.0" added in feast-operator images, which needs to be updated as well. 
https://github.com/search?q=repo%3Afeast-dev%2Ffeast%20APACHE_ARROW_VERSION&type=code

dlovell:
@ntkathole FYI: i've modified `lock-python-dependencies-all` to pass the args necessary to use the pytorch cpu index

its not clear that `APACHE_ARROW_VERSION` needs to change: its usages seem to be pointing to `minimal` installs of feast which still refer to `17.0.0`

ntkathole:
> @ntkathole FYI: i've modified `lock-python-dependencies-all` to pass the args necessary to use the pytorch cpu index

Looks like it is appending `+cpu `to the requirement explicitly causing issues. I am not able to push to your branch, so created https://github.com/feast-dev/feast/pull/5662 

> its not clear that `APACHE_ARROW_VERSION` needs to change: its usages seem to be pointing to `minimal` installs of feast which still refer to `17.0.0`

I see, right. No need to update since it's still pin to 17.0.0

dlovell:
@ntkathole FYI: this branch also includes a change to the ray / codeflare dependency environment marker. `uv lock` fails due to its use of `python_full_version` if this change is not present.

PR #5663, opened 2025-10-14
fix: Denial by default to all resources when no permissions set 

jyejare:

We observed Inconsistent permission response behavior across Feature Store API routes. 
- Fixing the  behaviour by changes in filter_only and named_patterns 

Inconsistent permission response behavior across Feature Store API routes. [RHOAIENG-35929](https://issues.redhat.com/browse/RHOAIENG-35929) 

# Misc

### DIFF
--- a/sdk/python/feast/permissions/enforcer.py
+++ b/sdk/python/feast/permissions/enforcer.py
@@ -72,19 +72,14 @@ def enforce_policy(
                 if evaluator.is_decided():
                     grant, explanations = evaluator.grant()
                     if not grant:
-                        if not filter_only:
-                            logger.error(f"Permission denied: {','.join(explanations)}")
-                            raise FeastPermissionError(",".join(explanations))
-                        elif filter_only and not p.name_patterns:
-                            logger.error(f"Permission denied: {','.join(explanations)}")
-                            raise FeastPermissionError(",".join(explanations))
-                        else:
+                        if filter_only and p.name_patterns:
                             continue
-                    if grant:
-                        logger.debug(
-                            f"Permission granted for {type(resource).__name__}:{resource.name}"
-                        )
-                        _permitted_resources.append(resource)
+                        logger.error(f"Permission denied: {','.join(explanations)}")
+                        raise FeastPermissionError(",".join(explanations))
+                    logger.debug(
+                        f"Permission granted for {type(resource).__name__}:{resource.name}"
+                    )
+                    _permitted_resources.append(resource)
                     break
         else:
             if not filter_only:
--- a/sdk/python/tests/unit/permissions/auth/server/test_auth_registry_server.py
+++ b/sdk/python/tests/unit/permissions/auth/server/test_auth_registry_server.py
@@ -142,6 +142,11 @@ def _test_get_historical_features(client_fs: FeatureStore):
 
 
 def _test_get_entity(client_fs: FeatureStore, permissions: list[Permission]):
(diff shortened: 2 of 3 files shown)


sdk/python/feast/permissions/enforcer.py
```diff
@@ -71,17 +71,47 @@ def enforce_policy(
...
+                            logger.error(f"Permission denied: {','.join(explanations)}")
+                            raise FeastPermissionError(",".join(explanations))
+                        else:
+                            continue
                     if grant:
                         logger.debug(
                             f"Permission granted for {type(resource).__name__}:{resource.name}"
                         )
                         _permitted_resources.append(resource)
                     break
```
jameerpathan111, line 83: @jyejare, what do you think about following suggestion?
```suggestion
                    if not grant:
                        if filter_only or p.name_patterns:
                            continue
                        logger.error(f"Permission denied: {','.join(explanations)}")
                        raise FeastPermissionError(",".join(explanations))
                    logger.debug(
                            f"Permission granted for {type(resource).__name__}:{resource.name}"
                        )
                    _permitted_resources.append(resource)
                    break
```

jyejare: @jameerpathan111 Accepting the suggestiong with `if filter_only and p.name_patterns:` instead of you proposed `if filter_only or p.name_patterns:`.

The and satisfies to cotinue when both filter_only and p.name_patterns are given when grant is not True.

sdk/python/feast/permissions/enforcer.py
```diff
@@ -71,17 +71,42 @@ def enforce_policy(
...
+                # filter_only=True: Check if there are permissions for this resource type
+                resource_type_permissions = [
+                    p
+                    for p in permissions
+                    if any(isinstance(resource, t) for t in p.types)  # type: ignore
+                ]
+                if not resource_type_permissions:
+                    # No permissions exist for this resource type - should raise error
+                    message = f"No permissions defined to manage {actions} on {type(resource)}/{resource.name}."
+                    logger.exception(f"**PERMISSION NOT GRANTED**: {message}")
```
ntkathole, line 99: should this be logger.error() ?

jyejare: Yeah thats error with exception :)

sdk/python/feast/permissions/enforcer.py
```diff
@@ -43,9 +43,9 @@ def enforce_policy(
         # If no permissions are defined, deny access to all resources
         # This is a security measure to prevent unauthorized access
         logger.warning("No permissions defined - denying access to all resources")
-        if not filter_only:
-            raise FeastPermissionError("No permissions defined - access denied")
-        return []
+        raise FeastPermissionError(
```
franciscojavierarceo, line 46: wait, am i correct in understanding that if no permissions are enabled we'll raise the permission error for everyone? is this forcing the functionality to everyone then?

jyejare: Yeah forcing the functionality for everyone, when kubernetes auth is deployed. So in short if kubernetes / RBAC is enabled but if the permissions itself are not granted then no would see anything.

franciscojavierarceo: okay cool, just wanted to confirm.

ntkathole (approved):
Looks good

PR #5665, opened 2025-10-14
feat: Added health endpoint for the UI

blaketastic2:

Adds a `/health` endpoint to the ui server for Load Balancer checks. Also uses `status` for HTTP codes.

None

# Misc

None


franciscojavierarceo:
Probably good to have a simple test here

PR #5666, opened 2025-10-15
feat: Support aggregation in odfv

HaoXuAI:

1. The pr moved the backend dataframe module to parent level rather than `local`, as it doesn't have to be local abstraction, and spark can inherit it as well.
2. Enable the `aggregations` config into the OnDemandFeatureView. When transforming the features, it can run either `aggregations` or `feature_transform`, not both. The reason is to make the the schema inference simple. But in future this can be redesigned to support both.

# Misc

### DIFF
--- a/docs/reference/data-sources/trino.md
+++ b/docs/reference/data-sources/trino.md
@@ -20,8 +20,8 @@ from feast.infra.offline_stores.contrib.trino_offline_store.trino_source import 
 )
 
 driver_hourly_stats = TrinoSource(
-    event_timestamp_column="event_timestamp",
-    table_ref="feast.driver_stats",
+    timestamp_field="event_timestamp",
+    table="feast.driver_stats",
     created_timestamp_column="created",
 )
 ```


sdk/python/feast/utils.py
```diff
@@ -605,7 +676,29 @@ def _augment_response_with_on_demand_transforms(
     for odfv_name, _feature_refs in odfv_feature_refs.items():
         odfv = requested_odfv_map[odfv_name]
         if not odfv.write_to_online_store:
-            if odfv.mode == "python":
+            # Apply aggregations if configured.
+            if odfv.aggregations:
+                if odfv.mode == "python":
```
haoxu0, line 681: Currently if this is a python input, we don't have aggregation function to use OOTD, so it use the `pandas` Dataframe backend to aggregate. I think in long term we need to create a `real time compute engine`, where we support python native compute operations such as aggregations, filtering, joining etc.

haoxu0: @franciscojavierarceo

franciscojavierarceo: what does OOTD mean ?

haoxu0: OOTB :)

sdk/python/feast/utils.py
```diff
@@ -561,6 +562,76 @@ def construct_response_feature_vector(
...
+
+    # Convert aggregations to agg_ops format
+    agg_ops = _get_aggregate_operations(aggregations)
+
+    # Apply aggregations using PandasBackend
+    if group_keys:
+        result_df = backend.groupby_agg(df, group_keys, agg_ops)
+    else:
+        # No grouping - aggregate over entire dataset
+        result_df = backend.groupby_agg(df, [], agg_ops)
```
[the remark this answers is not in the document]
haoxu0: Inconsistent with the API

docs/reference/beta-on-demand-feature-view.md
```diff
@@ -35,10 +35,40 @@ When defining an ODFV, you can specify the transformation mode using the `mode`
...
-write transformation functions that operate on a single row at a time, making the code more intuitive and aligning with 
+Native Python mode supports transformations on singleton dictionaries by setting `singleton=True`. This allows you to
+write transformation functions that operate on a single row at a time, making the code more intuitive and aligning with
 how data scientists typically think about data transformations.
 
+## Aggregations
+
+On Demand Feature Views support aggregations that compute aggregate statistics over groups of rows. When using aggregations, data is grouped by entity columns (e.g., `driver_id`) and aggregated before being passed to the transformation function.
+
+**Important**: Aggregations and transformations are mutually exclusive. When aggregations are specified, they replace the transformation function.
```
ntkathole, line 46: Any reason we are keeping aggregations and transformations mutually exclusive? I think if we change if else conditions from utils.py, we can achieve both in pipeline ?

Raw Data → Aggregations → Transformations → Output

franciscojavierarceo: +1 

aggregation is a transformation

haoxu0: refer to https://github.com/feast-dev/feast/issues/5689

ntkathole (approved):
Thank you

Issue #5641, opened 2025-09-29
Ondemand Feature with STRING type input fails on apply


OptimeeringBigya:
## Expected Behavior 
New on demand feature view created.

## Current Behavior
Attempting to apply a feature view with STRING type field fails with an exception below. 
```
/home/user/source_code/feast_exp/.venv/bin/python /home/user/source_code/feast_exp/src/od.py 
/home/user/source_code/feast_exp/.venv/lib/python3.11/site-packages/feast/feature_store.py:580: RuntimeWarning: On demand feature view is an experimental feature. This API is stable, but the functionality does not scale well for offline retrieval
  warnings.warn(
Traceback (most recent call last):
  File "/home/user/source_code/feast_exp/src/od.py", line 23, in <module>
    STORE.apply([r,transform])
  File "/home/user/source_code/feast_exp/.venv/lib/python3.11/site-packages/feast/feature_store.py", line 936, in apply
    self._make_inferences(
  File "/home/user/source_code/feast_exp/.venv/lib/python3.11/site-packages/feast/feature_store.py", line 643, in _make_inferences
    odfv.infer_features()
  File "/home/user/source_code/feast_exp/.venv/lib/python3.11/site-packages/feast/on_demand_feature_view.py", line 654, in infer_features
    inferred_features = self.feature_transformation.infer_features(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/source_code/feast_exp/.venv/lib/python3.11/site-packages/feast/transformation/pandas_transformation.py", line 98, in infer_features
    python_type_to_feast_value_type(
  File "/home/user/source_code/feast_exp/.venv/lib/python3.11/site-packages/feast/type_map.py", line 214, in python_type_to_feast_value_type
    raise ValueError(
ValueError: Value with native type object cannot be converted into Feast value type

Process finished with exit code 1
```

## Steps to reproduce
Attempt to apply the feature view below.
```
from feast import on_demand_feature_view
from feast import RequestSource
from feast import Field
from feast.types import String,Int32

from feature_store.configs.store import STORE

r=RequestSource(
    name="r",
    timestamp_field="event_ts",
    schema=[
        Field(name="input_1", dtype=String)
    ]
)


@on_demand_feature_view.on_demand_feature_view(name="a",schema=[],entities=[],sources=[r])
def transform(df):
    return df

if __name__ == "__main__":
    STORE.apply([r,transform])
```

[...]

PR #5669, opened 2025-10-15
fix: Fix STRING type handling in on-demand feature views

ntkathole:

Pandas DataFrames use dtype=object for strings by default (not dtype=string), but Feast's type inference system didn't handle "object" dtype, causing it to fail when trying to convert string columns.

Fixes https://github.com/feast-dev/feast/issues/5641

### DIFF
--- a/sdk/python/tests/unit/infra/test_inference_unit_tests.py
+++ b/sdk/python/tests/unit/infra/test_inference_unit_tests.py
@@ -96,14 +96,6 @@ def test_on_demand_features_valid_type_inference():
 
     python_native_test_view.infer_features()
 
-
-def test_on_demand_features_invalid_type_inference():
-    # Create Feature Views
-    date_request = RequestSource(
-        name="date_request",
-        schema=[Field(name="some_date", dtype=UnixTimestamp)],
-    )
-
     @on_demand_feature_view(
         sources=[date_request],
         schema=[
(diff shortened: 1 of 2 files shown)


sdk/python/tests/unit/infra/test_inference_unit_tests.py
```diff
@@ -96,29 +96,27 @@ def python_native_test_view(input_dict: dict[str, Any]) -> dict[str, Any]:
...
         data["output"] = features_df["some_date"]
         data["object_output"] = features_df["some_date"].astype(str)
         return data
 
-    with pytest.raises(ValueError, match="Value with native type object"):
-        invalid_test_view.infer_features()
+    object_string_test_view.infer_features()
+
+
+def test_on_demand_features_invalid_type_inference():
```
franciscojavierarceo, line 115: if this is invalid, shouldn't it raise?

ntkathole: it's a negative test.

ntkathole: There is `with pytest.raises()` sections which catch exceptions raised when mode is mismatched or feature is missing

franciscojavierarceo: ooh nice i  see on line 138, thank you.

franciscojavierarceo (approved):

PR #5686, opened 2025-10-21
fix: Date wise remote offline store historical data retrieval 

aniketpalu:

When using a remote offline store with PostgreSQL backend, calling get_historical_features() with start_date and end_date parameters fails with:

> TypeError: Got an unexpected keyword argument 'start_date' / 'end_date' for RemoteOfflineStore

The implementation follows the same pattern used by local PostgreSQL based offline stores ensuring consistency across the codebase:

- Client-side changes (remote.py):
  -   Add **kwargs parameter to RemoteOfflineStore.get_historical_features() method signature
  -   Extract start_date and end_date from kwargs and serialize them to ISO format strings
  -   Include serialized dates in api_parameters for transmission over Arrow Flight

- Server-side changes (offline_server.py):
  - Extract start_date and end_date from the command dict (if present)
  - Deserialize ISO format strings back to timezone-aware datetime objects
  - Pass extracted dates as **kwargs to the underlying offline store

# Misc

### DIFF
--- a/sdk/python/feast/infra/offline_stores/remote.py
+++ b/sdk/python/feast/infra/offline_stores/remote.py
@@ -219,7 +219,7 @@ class RemoteOfflineStore(OfflineStore):
             "full_feature_names": full_feature_names,
             "name_aliases": name_aliases,
         }
-        
+
         # Extract and serialize start_date/end_date for remote transmission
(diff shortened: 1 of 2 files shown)


ntkathole (approved):
lgtm

PR #5688, opened 2025-10-22
feat: Added support for filtering multi-projects

ntkathole:

This PR adds comprehensive multi-project support to the Feast UI, enabling users to view and filter data across multiple projects when using a shared registry. Previously, the UI would display all objects from all projects without distinction, making it difficult to work with multi-project setups.

**Features**:

- Project-Based Filtering
- Data on all list pages is now filtered based on the selected project
- Cross-Project Global Search
- Search results display the project name for each item
- "All Projects" View
- Appears automatically when multiple projects exist in the registry
- Maintains current page context when switching projects
- Projects are automatically detected from registry

https://github.com/user-attachments/assets/a9061600-44f6-48d8-86bc-efda8da31b71

### DIFF
--- a/ui/src/queries/useLoadRegistry.ts
+++ b/ui/src/queries/useLoadRegistry.ts
@@ -22,11 +22,12 @@ interface Feature {
   name: string;
   featureView: string;
   type: string;
+  project?: string;
 }
 
-const useLoadRegistry = (url: string) => {
+const useLoadRegistry = (url: string, projectName?: string) => {
   return useQuery(
-    `registry:${url}`,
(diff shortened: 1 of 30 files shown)


ui/src/queries/useLoadRegistry.ts
```diff
@@ -55,6 +55,121 @@ const useLoadRegistry = (url: string) => {
             objects.featureViews = [];
           }
 
+          // Debug: Log the first few objects to see what project values they have
```
franciscojavierarceo: you can probably clean up the logs after testing it.

PR #5691, opened 2025-10-24
fix: Improve trino to feast type mapping with (real,varchar,timestamp,decimal)

cutoutsy:
Like trino_to_pa_value_type() function, improve trino_to_feast_value_type() function
- Split decimal type into decimal32/decimal64 based on precision
- Add support for real type mapping to ValueType.FLOAT
- Add support for varchar(length) type
- Add support for timestamp(p) type


sdk/python/feast/infra/offline_stores/contrib/trino_offline_store/trino_type_map.py
```diff
@@ -14,13 +14,37 @@ def trino_to_feast_value_type(trino_type_as_str: str) -> ValueType:
...
         "varchar": ValueType.STRING,
         "boolean": ValueType.BOOL,
+        "real": ValueType.FLOAT,
     }
-    return type_map[trino_type_as_str.lower()]
+    _trino_type_as_str: str = trino_type_as_str
+    trino_type_as_str = trino_type_as_str.lower()
+
+    if trino_type_as_str.startswith("decimal"):
+        search_precision = re.search(
```
shuchu, line 29: calling "re" is expensive. since we only have "32" and "64" as the substring. can we do a simple Array based trick? like `trino_type_as_str[-2:] == "32"`

cutoutsy: not only have "32" and "64", it like  DECIMAL(10,3), DECIMAL(20)

trino data type: [https://trino.io/docs/current/language/types.html#decimal](https://trino.io/docs/current/language/types.html#decimal)

shuchu: got it. Thank you for the explanation.

shuchu (approved):
lgtm

PR #5690, opened 2025-10-24
feat: Batch Embedding at scale for RAG with Ray

ntkathole:
# What this PR does / why we need it:

This PR introduces `TransformationMode.RAY` for Native Ray Dataset processing and a new Ray RAG template that demonstrates distributed embedding generation for Retrieval-Augmented Generation (RAG) applications using Feast with Ray compute engine and Milvus vector store.

- Data ingestion → Distributed embedding generation → Vector search
- Distributed UDF execution with Ray Datasets

```
# feast init -t ray_rag
   🎬 Downloading real IMDB movie data for RAG demonstration...
   📥 Attempting to download IMDB dataset...
   🔑 Kaggle API found, checking authentication...
Dataset URL: https://www.kaggle.com/datasets/yashgupta24/48000-movies-dataset
   📥 Dataset downloaded successfully!
   📄 Found dataset file: final_data.csv
   📊 Dataset shape: (48513, 18)
   📋 Columns: ['id', 'url', 'Name', 'PosterLink', 'Genres', 'Actors', 'Director', 'Description', 'DatePublished', 'Keywords', 'RatingCount', 'BestRating', 'WorstRating', 'RatingValue', 'ReviewAurthor', 'ReviewDate', 'ReviewBody', 'duration']
   ✅ Successfully downloaded Kaggle dataset with 48513 movies
   📁 Copied CSV to: /feast-projects/handy_wildcat/feature_repo/data/final_data.csv
🚀 Ray RAG template initialized successfully!

🎯 To get started:
  1. cd handy_wildcat/feature_repo
  2. feast apply
  3. feast materialize --disable-event-timestamp
  4. python test_workflow.py

Creating a new Feast repository in /feast-projects/handy_wildcat.
```

```
# python test_workflow.py
Feature views: 1
Vector similarity search with Feast ...
   Query: 'Science fiction movie'
   📊 Top 3 results:
      1. Science Fiction
         Director: Danny Deprez | Genres: Adventure,Family,Mystery,Sci-Fi,Thriller
      2. The Scientist
         Director: Zach LeBeau | Genres: Drama,Sci-Fi
      3. Life
         Director: Daniel Espinosa | Genres: Horror,Sci-Fi,Thriller

   Query: 'Action movie with explosions and car chases'
   📊 Top 3 results:
      1. Blastfighter
         Director: Lamberto Bava | Genres: Action,Crime,Drama,Mystery
      2. Blast
         Director: Anthony Hickox | Genres: Action,Thriller
[...]

PR #5693, opened 2025-10-25
fix: Add mode field to Transformation proto for proper serialization

ntkathole:

This fixes a bug where transformation mode (pandas, python, spark, sql)
was stored in Python objects but not serialized to proto, causing
mode information to be lost during save/load cycles.

- Mode is stored in UserDefinedFunctionV2.mode (inside the Transformation proto)
- FeatureViewSpec has feature_transformation field which contains the transformation with mode
- Add mode at the FeatureView level for easy access (unification of Feature View)

### DIFF
--- a/protos/feast/core/FeatureView.proto
+++ b/protos/feast/core/FeatureView.proto
@@ -36,7 +36,7 @@ message FeatureView {
     FeatureViewMeta meta = 2;
 }
 
-// Next available id: 16
+// Next available id: 17
 // TODO(adchia): refactor common fields from this and ODFV into separate metadata proto
 message FeatureViewSpec {
     // Name of the feature view. Must be unique. Not updated.
@@ -51,17 +51,8 @@ message FeatureViewSpec {
     // List of specifications for each feature defined as part of this feature view.
     repeated FeatureSpecV2 features = 4;
 
-    // List of specifications for each entity defined as part of this feature view.
-    repeated FeatureSpecV2 entity_columns = 12;
-
-    // Description of the feature view.
-    string description = 10;
-
     // User defined metadata
     map<string,string> tags = 5;
-
-    // Owner of the feature view.
-    string owner = 11;
 
     // Features in this feature view can only be retrieved from online serving
     // younger than ttl. Ttl is measured as the duration of time between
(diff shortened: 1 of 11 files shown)


sdk/python/feast/protos/feast/core/FeatureView_pb2.pyi
```diff
@@ -137,7 +137,7 @@ class FeatureViewSpec(google.protobuf.message.Message):
     def source_views(self) -> google.protobuf.internal.containers.RepeatedCompositeFieldContainer[global___FeatureViewSpec]: ...
     @property
     def feature_transformation(self) -> feast.core.Transformation_pb2.FeatureTransformationV2:
-        """Feature transformation (UDF or Substrait) for batch feature views"""
```
franciscojavierarceo, line 140: Think you forgot to commit the featurieview protobuf

franciscojavierarceo: Still missing the FeatureView.proto

ntkathole: ah, ok .you mean the change for unification

ntkathole: Added now

franciscojavierarceo (approved):

protos/feast/core/FeatureView.proto
```diff
@@ -71,20 +62,33 @@ message FeatureViewSpec {
...
     // Whether these features should be written to the offline store
     bool offline = 13;
 
     repeated FeatureViewSpec source_views = 14;
 
     // Feature transformation for batch feature views
     FeatureTransformationV2 feature_transformation = 15;
+
+    // The transformation mode (e.g., "python", "pandas", "spark", "sql", "ray")
+    string mode = 16;
```
haoxu0, line 91: Sorry missed reviewing this. Do we need this to be declared in FeatureView? Can't we infer from the transformation definition? @ntkathole

franciscojavierarceo: Yes long term we do. We will consolidate everything to just feature view and so it belongs here. Batch feature view inherits from feature view.

ntkathole: yeah, I missed to link the context in this PR https://github.com/feast-dev/feast/pull/5690#discussion_r2462910416 related to https://github.com/feast-dev/feast/issues/4584

PR #5699, opened 2025-10-28
feat: Add batch commit mode for MySQL OnlineStore

chimeyrock999:
Adds optional batch insert and commit mode to the MySQLOnlineStore to improve materialization performance for TiDB and MySQL backends.
By default, Feast writes each row individually and commits per record, which causes large overhead on MySQL compatible Raft-based databases like TiDB.
This PR introduces a configurable batch mode to group inserts and commit per batch, improving write throughput.
## Changes

- Added config flags:
  ```
  batch_mode: true
  batch_size: 1000
  ```
- Implemented batched insert using `executemany()` / multi-row SQL.
- Each batch is committed atomically.
- Backward compatible (default = per-row mode).

# Misc
- Verified on TiDB v7.5.0 and MySQL 8.0.44.
- Thanks @ntkathole confirming this direction.

### DIFF
--- a/docs/reference/online-stores/mysql.md
+++ b/docs/reference/online-stores/mysql.md
@@ -28,6 +28,28 @@ online_store:
 
 The full set of configuration options is available in [MySQLOnlineStoreConfig](https://rtd.feast.dev/en/master/#feast.infra.online_stores.mysql_online_store.MySQLOnlineStoreConfig).
 
+## Batch write mode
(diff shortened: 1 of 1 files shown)


ntkathole:
@chimeyrock999 Thank you! Can we also please mention this support in our docs ? may be at https://github.com/feast-dev/feast/blob/master/docs/reference/online-stores/mysql.md

chimeyrock999:
@ntkathole Thank you for the feedback! I’ve added documentation to cover this support of MySQL Online Store.

PR #5701, opened 2025-10-28
feat: Add annotations to cronjob CRDs

FredericKayser:

This PR adds support for custom annotations on the CronJob resource in the FeatureStore CRD. This enhancement allows users to specify Kubernetes annotations for CronJobs deployed by the Feast operator, which is useful for:
- Integration with monitoring and observability tools
- Policy enforcement and compliance tooling
- Service mesh configuration
- Custom operator integrations that use annotations for configuration

The annotations field has been added to both the spec and status sections of the CronJob definition in the FeatureStore CRD.


FredericKayser:
@franciscojavierarceo do you have any idea why the CI fails now? The failure doesn't really seem to be related to my changes.

PR #5703, opened 2025-10-29
fix: Support historical feature retrieval with start_date/end_date in RemoteOfflineStore

aniketpalu:

**Problem**

- get_historical_features() with start_date/end_date parameters (no entity_df) works in local mode but fails in remote mode with:
  - EntityTimestampInferenceException: Please provide an entity_df with a column named event_timestamp

**Root Cause**

- Remote client sends mock table {"key": "mock_key"} when entity_df is None
- Server treats mock table as valid entity_df, which lacks event_timestamp column
- Server doesn't forward start_date/end_date to underlying offline store

**Changes**

- Client (remote.py):
  - Handle None entity_df in _create_retrieval_metadata() and _get_entity_schema()
- Server (offline_server.py):
  - Detect and discard mock tables (single 'key' column)
  - Create minimal entity_df with event_timestamp from start_date/end_date
  - Forward start_date/end_date kwargs to underlying offline store
  - Use pd.to_datetime() and utils.make_tzaware() for proper timestamp types

# Misc

### DIFF
--- a/sdk/python/feast/offline_server.py
+++ b/sdk/python/feast/offline_server.py
@@ -8,7 +8,6 @@ from datetime import datetime
 from typing import Any, Dict, List, Optional, cast
 
 import click
-import pandas as pd
 import pyarrow as pa
 import pyarrow.flight as fl
 from google.protobuf.json_format import Parse


ntkathole (approved):
Looks good! @aniketpalu can you please rebase to resolve conflicts ?

sdk/python/feast/offline_server.py
```diff
@@ -431,6 +432,20 @@ def get_historical_features(self, command: dict, key: Optional[str] = None):
...
+                entity_df = None
+
+        # If no entity_df provided, create minimal one with timestamps
+        if entity_df is None and "start_date" in command and "end_date" in command:
+            # Create minimal entity_df with event_timestamp when using start_date/end_date mode
+            start_date = utils.make_tzaware(
+                datetime.fromisoformat(command["start_date"])
+            )
+            end_date = utils.make_tzaware(datetime.fromisoformat(command["end_date"]))
+            entity_df = pd.DataFrame(
```
jyejare: We should not need to create dummy entity_df at all. We should handle empty(`entity_df=None`) entity_df for and in `offline_store.get_historical_features` which we are returning in this function.

aniketpalu: You are right, its not needed to create mock dataframe with timestamps, just had to make sure that empty_df is being passed.

Issue #5707, opened 2025-10-31
Slow materialization


miroslawas:
## Expected Behavior 

Feature materialization used to call "get latest feature values" function and return only the latest values - this was much faster

## Current Behavior

Materializatio now pulls all features now and deduplicates in pandas - this makes materialization much slower (almost unusable for some cases) 

## Steps to reproduce

Try materialize

### Specifications

- Version: 0.56
- Platform: 
- Subsystem:

offline_store: Clickhouse, online_store: Redis

## Possible Solution

Give ability to pull only latest values by introducing flag/env variable

miroslawas:
@franciscojavierarceo @HaoXuAI

miroslawas:
FYI, @astronautas

astronautas:
We could contribute this param ourselves, no issue at all, just wanted to see what maintainers think.

franciscojavierarceo:
was this always the case or recently changed? It'd be helpful to know what exposed this. If some users like this behavior than probably the exercise is for us to make this configurable. As a general point, I do agree with you that having pandas execute deduplication is wildly expensive.

astronautas:
> was this always the case or recently changed? It'd be helpful to know what exposed this. If some users like this behavior than probably the exercise is for us to make this configurable. As a general point, I do agree with you that having pandas execute deduplication is wildly expensive.

Seems from 0.50.0, the [version which broke Clickhouse offline store ](https://github.com/feast-dev/feast/issues/5493), since the core had started calling ```get_all_...``` instead of ```get_latest_...``` from offline store (which eventually requires deduplication taken care by online store e.g. Redis does this in pure Python / Pandas).

haoxu0:
It's probably a bug, even it changed to pull_all, default will still dedup and get the latest features only

astronautas:
> It's probably a bug, even it changed to pull_all, default will still dedup and get the latest features only

@HaoXuAI I don't think it's a bug. It's a feature. Functionally, it does the job - online store receives only the latest values. The issue is not about this.

Non-functionally, it's much slower. That is the issue. Why? It dedups just before pushing to the online store. Pure Python is used for that. For large periods between `start_ts` and `end_ts`, that pulls in a significant amount of data from offline store to be deduped in later downstream. Before it would dedup using SQL via `get_latest`, so it was multiple times faster and used many times less memory. Basically, it did predicate pushdown, in some sense.

[...]

PR #5709, opened 2025-11-03
feat: Add deduplicate pushdown to clickhouse - improve materialize performance

astronautas:

https://github.com/feast-dev/feast/issues/5707. We have observed significant slowdown in heavy materialization jobs. Heavy = pulling in significant period of {start_ts, end_ts}, when essentially you end up with many values to be deduplicated by Feast's compute engine. Unfortunately, Polars local engine did not show any significant speed-up. Spark is out of question due to sheer complexity of adding yet another cluster engine. We also observed nearly twice increased memory usage...

Not being a fan of altering the core, I would like to introduce a local change to Clickhouse provider to pushdown the deduplication logic to offline store. This assumes you don't have any Feast compute transformations or aggregation. I would like to just use a flag for now and not make it overcomplicated i.e. try to infer it's Pandas, if it has compute engine transformations. Keeping it simple, for now, and disabled by default.

https://github.com/feast-dev/feast/issues/5707

# Misc

### DIFF
--- a/sdk/python/feast/infra/offline_stores/contrib/clickhouse_offline_store/clickhouse.py
+++ b/sdk/python/feast/infra/offline_stores/contrib/clickhouse_offline_store/clickhouse.py
@@ -201,12 +201,17 @@ class ClickhouseOfflineStore(OfflineStore):
         join_key_columns: List[str],
         feature_name_columns: List[str],
         timestamp_field: str,
-        start_date: datetime,
-        end_date: datetime,
         created_timestamp_column: Optional[str] = None,
+        start_date: Optional[datetime] = None,
+        end_date: Optional[datetime] = None,
     ) -> RetrievalJob:
         assert isinstance(config.offline_store, ClickhouseOfflineStoreConfig)
         assert isinstance(data_source, ClickhouseSource)
+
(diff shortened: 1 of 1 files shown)


astronautas:
@HaoXuAI Let me know what you think. I really wouldn't like to make it more complicated than above, for now 😁 .

astronautas:
@franciscojavierarceo cc

franciscojavierarceo:
I guess I'm curious who wouldn't want this behavior natively? Based on the other thread, it sounds like we degraded functionality and probably it would make sense to remove the pandas deduplication and make that optional. 

What do you think @HaoXuAI?

astronautas:
> Thanks for contributing, but I think this will make it inconsistent with other offline store. Instead of changing at the store level, I would suggest to make the materialize an option config `pull_latest=true`, this will explicitly let the user to be able to either pull all or pull latest. @franciscojavierarceo The reason we change to pull_all is because pull_latest doesn't work with the transformation such as aggregation and in some cases feature_transformation as well.

Will adjust, makes sense.

astronautas:
https://github.com/feast-dev/feast/pull/5713 closing in favor of this

PR #5710, opened 2025-11-04
fix: Thread safe Clickhouse offline store

astronautas:

PR adjusts the Clickhouse offline store to be thread safe. Issue with the current implementation results in the issue as follows:

```raise ProgrammingError('Attempt to execute concurrent queries within the same session.' ```

if you bombard it - `feast serve` - with concurrent requests, the under the hood threadpool is more than 1 thread by default (arbitrary number decided by uvicorn).

as per Clickhouse docs [here](https://clickhouse.com/docs/integrations/language-clients/python/advanced-usage#multithreaded-multiprocess-and-asyncevent-driven-use-cases):

> Because each query or insert executed maintains state in its own QueryContext or InsertContext object, respectively, these helper objects are not thread-safe, and they should not be shared between multiple processing streams.

Solution - using threading.local() we can initialize a per-thread container for variables, to make sure we re-use the client on subsequent calls. Can we just drop the caching? Not really, unless the store is heavily refactored.

No issue raised yet.

# Misc

### DIFF
--- a/sdk/python/feast/infra/utils/clickhouse/connection_utils.py
+++ b/sdk/python/feast/infra/utils/clickhouse/connection_utils.py
@@ -8,8 +8,6 @@ from feast.infra.utils.clickhouse.clickhouse_config import ClickhouseConfig
 thread_local = threading.local()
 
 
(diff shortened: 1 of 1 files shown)


franciscojavierarceo (approved):
lgtm

PR #5713, opened 2025-11-05
feat: Add possibility to materialize only latest values, to increase performance

astronautas:

Adds an option to materialize only the latest values (essentially pushes down deduplication to offline store), to reduce client memory consumption and reduce e2e duration. Especially noticeable for large-scale materialization - think hundreds of thousands of rows with ~150 feature views, with latency-critical materializations - as we observed in our ML project at cast.ai.

https://github.com/feast-dev/feast/issues/5707#issuecomment-3480031768

# Misc

This will be configured via feature store (repo) config file:
```
registry: registry.db
project: credit_scoring_aws
provider: local
online_store:
  type: sqlite
offline_store:
  type: file
entity_key_serialization_version: 3
materialization: # <- new option here
  pull_latest_features: true | false (default)
```


astronautas:
@HaoXuAI

astronautas:
@franciscojavierarceo

astronautas:
Let's re-run tests? Random issue, but no changes to dependency management :/

astronautas:
image

Dependency install was skipped :what:

astronautas:
seems like aws creds have expired @franciscojavierarceo

franciscojavierarceo:
@ntkathole @jeremyary can you investigate?

astronautas:
> I think it might be better to add the config to the fs.materialize API? So that you can customize the materialize process that materialize the FeatureView if you need pushdown filter, and some other process you don't need.

Why not indeed. I'll check it out and tag you back.

PR #5716, opened 2025-11-11
fix: Materialize API - ODFV views not looked-up (thinks views non existant)  - crashes materialize

astronautas:

On-demand feature views not checked when materialize is called via a web-endpoint `/materialize`. Which means, ODFV cannot be materialized via an endpoint - FeatureViewNotFoundException gets raised...

Solution: check `get_on_demand_feature_view` too!

On a bigger scale, this is caused by apparently 3 different implementations of checking whether a feature view exists - logging, server and feature_store (Python SDK). I wouldn't venture into refactoring yet, since we're quite busy, but at least this bugfix unblocks ODFV materialization (which we need a lot).

Also re-organized a bit - imho we should prioritize checking features, odfvs and then stream.

### DIFF
--- a/sdk/python/feast/feature_server.py
+++ b/sdk/python/feast/feature_server.py
@@ -344,18 +344,27 @@ def get_app(
     async def _get_feast_object(
         feature_view_name: str, allow_registry_cache: bool
     ) -> FeastObject:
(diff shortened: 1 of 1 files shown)


astronautas:
@franciscojavierarceo

PR #5718, opened 2025-11-12
fix: add option to force full_feature_names for materialize

astronautas:

We stumbled upon a peculiar issue while working on materializing some ODFVs. It appears deep within the flow, ODFV materialization calls get_historical_features. Why it is done, not sure. In any case, for projects where feature names collide and feature_views act as namespaces, this causes the following issue:

`/materialize`

```
    raise FeatureNameCollisionError(collided_feature_refs, full_feature_names)
feast.errors.FeatureNameCollisionError: Duplicate features named <my_feature_1>
To resolve this collision, either use the full feature name by setting 'full_feature_names=True', or ensure that the features in question have different names.
```

I propose we expose an optional parameter to `/materialize`, to specify whether full_feature_names are passed-in. This way, for projects that really use full_feature_names, the materialize flow won't break.

### DIFF
--- a/sdk/python/feast/feature_server.py
+++ b/sdk/python/feast/feature_server.py
@@ -78,6 +78,7 @@ class MaterializeRequest(BaseModel):
     end_ts: Optional[str] = None
     feature_views: Optional[List[str]] = None
     disable_event_timestamp: bool = False
+    full_feature_names: bool = False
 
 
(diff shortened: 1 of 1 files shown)


franciscojavierarceo:
wait was this an issue?

astronautas:
Yes, premature PR though (yes, it's an issue, just didn't submit a formal issue!)

PR #5722, opened 2025-11-15
fix: Fixed file-based registry cache sync

ntkathole:

This PR fixes an issue where file-based registries, where registry servers in operator deployments, were not refreshing immediately after `feast apply` even when `cache_mode: sync` was set. The changes ensure that registry servers detect file modifications immediately, providing immediate consistency as expected with sync mode.

## Problem

1. **Python File Registry**: File writes were not being synced to disk, causing `refresh_registry()` to potentially read stale data immediately after `commit()`.

2. **Python Registry Sync Mode**: The `Registry` class only checked TTL expiration, not file modification time. This meant that even with `cache_mode: sync`, separate registry server processes would only see changes after the TTL expired (default 600 seconds), not immediately.

### 1. File Registry Disk Sync (Python)
- Added `f.flush()` and `os.fsync(f.fileno())` to `FileRegistryStore._write_registry()` to ensure writes are synced to disk before returning.

### 2. Python Registry File Modification Time Checking
- Enhanced `Registry` class to:
  - Store `cache_mode` from registry config
  - Track file path and modification time for file-based registries
  - Check file modification time in `_get_registry_proto()` when `cache_mode == "sync"` and `allow_cache=True`
  - Refresh immediately if file modification time changed, regardless of TTL expiration

### DIFF
--- a/sdk/python/feast/infra/registry/file.py
+++ b/sdk/python/feast/infra/registry/file.py
@@ -1,4 +1,5 @@
 import json
+import os
 import uuid
 from pathlib import Path
 from typing import Optional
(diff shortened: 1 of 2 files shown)


HaoXuAI (approved):
lgtm！

PR #5725, opened 2025-11-17
feat: Add annotations for pod templates

FredericKayser:

Adds support for applying user-defined annotations to the CronJob JobTemplate PodTemplate (PodTemplateAnnotations), enabling cases like disabling Istio sidecar injection (e.g. sidecar.istio.io/inject: "false"), integrating with monitoring, policy, and service mesh tooling.


ntkathole (approved):
Looks good

FredericKayser:
@franciscojavierarceo

Issue #5727, opened 2025-11-18
Materialize via feature server for on-demand feature views crashes


astronautas:
## Expected Behavior 
I propose we expose an optional parameter to `/materialize`, to specify whether full_feature_names are passed-in. This way, for projects that really use full_feature_names, the materialize flow won't break.

## Current Behavior
We stumbled upon a peculiar issue while working on materializing some ODFVs. It appears deep within the flow, ODFV materialization calls get_historical_features. Why it is done, not sure. In any case, for projects where feature names collide and feature_views act as namespaces, this causes the following issue:

`/materialize`

```
    raise FeatureNameCollisionError(collided_feature_refs, full_feature_names)
feast.errors.FeatureNameCollisionError: Duplicate features named <my_feature_1>
To resolve this collision, either use the full feature name by setting 'full_feature_names=True', or ensure that the features in question have different names.
```

## Steps to reproduce

Call `/materialize` for any arbritrary ODFV.

### Specifications

- Version:
- Platform:
- Subsystem:

## Possible Solution
[PR.](https://github.com/feast-dev/feast/pull/5728)

astronautas:
@franciscojavierarceo

PR #5728, opened 2025-11-18
fix: Add possibility to force full features names for materialize ops

astronautas:

Adds an optional param to materialize to force use full_feature_names. This ensures ODFV materialization does not crash - since it's currently hardcoded to full_feature_names=False - in case ODFVs have multiple features with same name.

E.g.

```
# /materialize
{
    "feature_views": [
        "my_odfv",
    ],
    "start_ts": "2025-11-10T15:52:46.962218",
    "end_ts": "2025-11-13T15:52:46.962218",
    "full_feature_names": true # <- optional
}
```

https://github.com/feast-dev/feast/issues/5727

# Misc


astronautas:

sdk/python/feast/feature_store.py
ntkathole, line 1581: nitpick, it would be great if we update full_feature_names in docstrings as well

ntkathole (approved):
nitpick, looks good

sdk/python/feast/feature_store.py
astronautas: Will do!

astronautas:
@ntkathole fixed!

PR #5729, opened 2025-11-18
feat: Added batching to feature server /push to offline store (#5683)

### LINKED ISSUE #5683: Feature Request: Add configurable batching for offline store writes in Feature Server push API
## Is your feature request related to a problem? Please describe.

Currently, when using the Feature Server's push API with streaming ingestion, offline store writes are not automatically batched. Users must either:
1. Manually batch calls to `write_to_offline_store()` in their application code
2. Set up a separate stream processing pipeline (e.g., Spark with Kafka) to handle batching

For users receiving high-frequency feature updates (e.g., 100 requests/second) who want to write to both online and offline stores, this creates unnecessary complexity and can lead to inefficient writes to offline stores like BigQuery or Parquet.

## Describe the solution you'd like

Add configurable automatic batching for offline store writes in the Feature Server, similar to how the Spark-Kafka processor handles batching via the `processingTime` parameter.

**Proposed configuration options:**
- `offline_write_batch_size`: Number of records to buffer before flushing to offline store
- `offline_write_batch_interval`: Time interval (seconds) to wait before flushing to offline store
- Flush on whichever threshold is reached first

**Example configuration in `feature_store.yaml`:**
```yaml
online_store:
  type: redis
  ...
offline_store:
  type: bigquery
  ...
feature_server:
  offline_write_batching:
    enabled: true
    batch_size: 1000
    batch_interval_seconds: 30
```

## Describe alternatives you've considered

1. **Current approach with stream processor**: Users can set up a Spark Structured Streaming job that reads from Kafka and uses `processingTime` to control batching. This works but requires additional infrastructure and complexity.

2. **Manual batching in application code**: Users can implement their own buffering logic before calling `write_to_offline_store()`, but this duplicates effort across different applications.

3. **Use only online store**: Skip offline store writes entirely during streaming and rely on periodic materialization from batch sources, but this loses the real-time historical logging capability.

## Additional context

This feature request came from a community discussion about streaming ingestion patterns. When using `PushMode.ONLINE_AND_OFFLINE` with high-frequency updates, the lack of automatic batching can result in:
- Excessive API calls to offline stores (one per feature update)
- Higher costs for cloud data warehouses like BigQuery
- Increased latency for write operations

**Current implementation reference:**
The Spark-Kafka processor already implements batching at `sdk/python/feast/infra/contrib/spark_kafka_processor.py:134-173` using `foreachBatch`. Similar batching logic could be added to the Feature Server's push endpoint.

**Benefits:**
- Simplified architecture for users who want real-time serving + historical logging
- Reduced costs for offline store writes
- Better alignment with best practices for data warehouse ingestion
- No breaking changes (feature would be opt-in via configuration)

Related code:
- Feature Server: `sdk/python/feast/feature_store.py:1786-1828` (`write_to_offline_store`)
- Spark batching example: `sdk/python/feast/infra/contrib/spark_kafka_processor.py:134-173`

cc @feast-dev/maintainers

jfw-ppi:
Added batching configuration for feature_servers /push endpoint for offline store writes

Fixes #5683


sdk/python/feast/infra/feature_servers/base_config.py
```diff
@@ -1,9 +1,36 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 from typing import Optional
 
 from pydantic import StrictBool, StrictInt
 
 from feast.repo_config import FeastConfigBaseModel
 
+class OfflinePushBatchingConfig(FeastConfigBaseModel):
```
franciscojavierarceo: i think having a config is fine but do we actually need it? we probably could have just passed these as optional args, no?

jfw-ppi: You're right, I just noticed the FeatureLoggingConfig with 5 fields and decided that for 3 fields it would be justified to also add a config. Do you want me to refactor it to use optional args?

jfw-ppi: > i think having a config is fine but do we actually need it? we probably could have just passed these as optional args, no?

I refactored it as you wanted, so that there is no config. @franciscojavierarceo

sdk/python/feast/feature_server.py
```diff
@@ -732,3 +823,170 @@ def start_server(
...
+    def _flush_locked(self, key: _OfflineBatchKey) -> None:
+        """
+        Flush a single buffer; caller must hold self._lock.
+        """
+        dfs = self._buffers.get(key)
+        if not dfs:
+            return
+
+        batch_df = pd.concat(dfs, ignore_index=True)
+        self._buffers[key].clear()
```
ntkathole: Does it make sense to move clear inside `try: self._store.push` so that buffer gets cleared only after the write succeeds ?

sdk/python/feast/feature_server.py
```diff
@@ -732,3 +823,170 @@ def start_server(
...
+
+        logger.debug(
+            "Flushing offline batch for push_source=%s with %s rows",
+            key.push_source_name,
+            len(batch_df),
+        )
+
+        # NOTE: offline writes are currently synchronous only, so we call directly
+        try:
+            self._store.push(
```
ntkathole: What about splitting `_flush_locked` into two methods: one that extracts data (with lock) and one that does I/O (without lock) ? 

Something like:
- Extracting the batch data while holding the lock
- Releasing the lock before doing I/O
- Re-enqueueing data if the write fails

sdk/python/tests/unit/test_feature_server.py
```diff
@@ -209,6 +225,244 @@ def test_materialize_request_model():
...
+        enabled=batching_enabled,
+        batch_size=1,  # flush immediately on a single offline request
+        batch_interval_seconds=60,
+    )
+
+    client = TestClient(get_app(fs))
+
+    # use a multi-row payload to ensure we test non-trivial dfs
+    resp = client.post("/push", json=push_body_many(push_mode, count=2, id_start=100))
+    assert resp.status_code == 200
```
ntkathole: optional but I think it's good to return 202 when batching is enabled and offline writes are involved

jfw-ppi: Nice catch! It's done.

sdk/python/feast/feature_server.py
jfw-ppi: Thanks for the suggestion! 
I split _flush_locked into _drain_locked (extract under lock) and _flush (I/O without lock). I also added _inflight to prevent concurrent flushes per key. On failure the drained batch is re‑enqueued so we don’t drop data.

sdk/python/feast/feature_server.py
jfw-ppi: Totally makes sense, it's done. Thanks for seeing that.

PR #5738, opened 2025-11-25
feat: Offline Store historical features retrieval based on datetime range in Ray

aniketpalu:

- Add support for entity_df=None in RayOfflineStore.get_historical_features with start_date/end_date.
  -- Derives entity set by reading distinct join keys from each FeatureView source within the time window, applies field mappings and join_key_map, filters by timestamp, and unions aligned schemas.
  -- Adds stable event_timestamp = end_date for PIT joins.

- Signature change: get_historical_features accepts entity_df: Optional[Union[pd.DataFrame, str]] and **kwargs.
  -- Why: Match base interface and support date-only retrieval.

[RHOAIENG-38643](https://issues.redhat.com/browse/RHOAIENG-38643)

# Misc

### DIFF
--- a/sdk/python/feast/infra/offline_stores/contrib/ray_offline_store/tests/test_ray_integration.py
+++ b/sdk/python/feast/infra/offline_stores/contrib/ray_offline_store/tests/test_ray_integration.py
@@ -1,3 +1,5 @@
+from datetime import timedelta
+
 import pandas as pd
 import pytest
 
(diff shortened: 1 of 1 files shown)


sdk/python/feast/infra/offline_stores/contrib/ray_offline_store/ray.py
```diff
@@ -1197,6 +1197,152 @@ def schema(self) -> pa.Schema:
             return pa.Table.from_pandas(df).schema
 
 
+def _compute_non_entity_dates_ray(
```
jyejare, line 1200: I think we should have make a common utility function for this, so that it can be used in all stores without repeating the code.

wdyt ?

sdk/python/feast/infra/offline_stores/contrib/ray_offline_store/ray.py
```diff
@@ -1197,6 +1197,152 @@ def schema(self) -> pa.Schema:
...
+def _make_filter_range(timestamp_field: str, start_date: datetime, end_date: datetime):
+    # Why: factory function for time-range filtering in Ray map_batches
+    def _filter_range(batch: pd.DataFrame) -> pd.Series:
+        ts = pd.to_datetime(batch[timestamp_field], utc=True)
+        return (ts >= start_date) & (ts <= end_date)
+
+    return _filter_range
+
+
+def _make_select_distinct_keys(join_keys: List[str]):
```
jyejare: I think we should not drop rows with duplicate IDs, because there could be multiple transactions per ID and we need to choose the row based on timestamp while joining the colums from another table/view. I think this is the same case with your spark PR.

Please check the postgres implementation to understand the case.

Or Am I misreading this ?

aniketpalu: Testing the case after discussion

aniketpalu: Previously, when entity_df=None was passed to get_historical_features(), the Ray offline store would extract only distinct entity keys and assign a single fixed timestamp (end_date) to all entities. This broke point-in-time joins for cases where multiple transactions exist per entity ID in date-time range. 

Now extracts distinct (entity_keys, event_timestamp) combinations, aligning with Postgres based offline store's behaviour.

sdk/python/feast/infra/offline_stores/contrib/ray_offline_store/ray.py
```diff
@@ -1197,6 +1197,171 @@ def schema(self) -> pa.Schema:
...
+    def _filter_range(batch: pd.DataFrame) -> pd.Series:
+        ts = pd.to_datetime(batch[timestamp_field], utc=True)
+        return (ts >= start_date) & (ts <= end_date)
+
+    return _filter_range
+
+
+def _make_select_distinct_entity_timestamps(join_keys: List[str], timestamp_field: str):
+    # Why: factory function for distinct (entity_keys, event_timestamp) projection in Ray map_batches
+    # This preserves multiple transactions per entity ID with different timestamps for proper PIT joins
```
jyejare: I think we can remove / update this ?

ntkathole (approved):

PR #5743, opened 2025-11-27
feat: Production ready feast operator with v1 apiversion

jyejare:

This PR now switches the feast operator to API version from `v1alpha1` to `v1` as a production ready operator with no more alpha tag.

Closes Jira https://issues.redhat.com/browse/RHOAIENG-39324

# Misc


infra/feast-operator/config/crd/patches/v1alpha1_deprecated.yaml
```diff
@@ -0,0 +1,45 @@
+# This patch adds v1alpha1 as a deprecated version to support migration from previous versions
```
ntkathole, line 1: @jyejare rather than using rename + patch approach, I think we should use kubebuilder standard approach for API version upgrades, where we keep both v1 and v1aplha1 like:
```
api/
├── v1/
│   ├── featurestore_types.go      (+kubebuilder:storageversion)
│   ├── groupversion_info.go
│   └── zz_generated.deepcopy.go
└── v1alpha1/
    ├── featurestore_types.go      (+kubebuilder:deprecatedversion)
    ├── groupversion_info.go
    └── zz_generated.deepcopy.go
```

and CRD will look like:

```
versions:
- name: v1
  storage: true           # Primary storage version
  # Full schema...
  
- name: v1alpha1
  storage: false          # Deprecated, not stored
  deprecated: true
  deprecationWarning: "feast.dev/v1alpha1 FeatureStore is deprecated, use feast.dev/v1 instead"
  # Full schema (for validation)
```

jyejare: Yeah, I tried that with latest update but I think there is something wrong with our tests. I am closing this PR as I have  messed up the changes.

jyejare: Opened https://github.com/feast-dev/feast/pull/5771

Issue #5689, opened 2025-10-22
Make OnDemandFeatureView work for having both aggregations and transformations


HaoXuAI:
**Is your feature request related to a problem? Please describe.**

With PR https://github.com/feast-dev/feast/pull/5666, OnDemandFeatureView supports `aggregations`. But it is exclusive with `transformation`.
the reason is the infer_schema will break if having both defined in feature views.
This can be solved by reworking the infer_schema.
```
    def infer_features(self) -> None:
        random_input = self._construct_random_input(singleton=self.singleton)
        inferred_features = self.feature_transformation.infer_features(
            random_input=random_input, singleton=self.singleton
        )
```
  Schema inference doesn't run the aggregation step - it only:
  1. Looks at the source feature view's schema
  2. Creates fake random data matching those columns
  3. Passes it directly to the UDF
  But at runtime, the UDF receives aggregated columns, not the original source columns.
  So the UDF would need to be written for aggregated input:
```
  def my_odfv(inputs):
      return {"trips_doubled": inputs["sum_trips"] * 2}  # expects sum_trips
```
  But schema inference passes it:
```
  {"trips": [10], "rating": [4.5]}
```
**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

franciscojavierarceo:
@ntkathole @jyejare FYI

astronautas:
Also seems ODFV materialize has a completely different compute engine flow, be careful there. Maybe legacy split or smth.

franciscojavierarceo:
yeah we need to think through how to align the compute engine with ODFVs. ODFVs basically use 2 compute engines: one batch one and the feature server. We have to work out how best to make that cleaner.

haoxu0:
yep. So  we need to refactor OnDemandFeatureView to:
1. Inherit from FeatureView
2. use the current feature resolver/builder

haoxu0:
another tech debt found: https://github.com/feast-dev/feast/pull/5716, forcing to refactor the OnDemandFeatureView

PR #5283, opened 2025-04-19
feat: Make transform on writes configurable for ingestion

franciscojavierarceo:
# What this PR does / why we need it:

This PR introduces a new `transform_on_write` flag to various functions and classes in the Feast SDK. The flag allows users to enable or disable transformations of data when writing to the online or offline store. This feature improves flexibility in managing data transformations during feature ingestion.

This feature is particularly useful if you want to batch transform a set of data (e.g., using Spark) and upload that transformed data to the online store **_and_** then want to also offer transformation on writes with the Feast Feature Server using an On Demand transformation.

#### Changes Summary:

- **`sdk/python/feast/feature_server.py`**
    - Added `transform_on_write` attribute to `WriteToFeatureStoreRequest` and `PushFeaturesRequest` classes.
    - Updated `push` and `write_to_online_store` functions to pass the `transform_on_write` parameter to downstream calls.
- **`sdk/python/feast/feature_store.py`**
    - Added `transform_on_write` parameter to the `push` method and its docstring.
    - Updated the `push` method to pass the `transform_on_write` parameter to the `write_to_online_store` and `write_to_offline_store` methods.
    - Modified `_get_feature_view_and_df_for_online_write` and `write_to_online_store` functions to support the `transform_on_write` parameter.
- **`sdk/python/tests/unit/test_on_demand_python_transformation.py`**
    - Fixed a typo in the constant `ODFV_OTHER_STRING_CONSTANT`.
    - Introduced new test cases for `transform_on_write` functionality to verify behavior when the flag is disabled (`transform_on_write=False`).
    - Added constants and test data for validating untransformed writes to the online store.

This enhancement provides better control over data transformations when pushing features, which can be particularly useful for scenarios where pre-transformed data is being ingested.

# Which issue(s) this PR fixes:
https://github.com/feast-dev/feast/issues/5196

# Misc

PR #5747, opened 2025-11-28
feat: Unify Feature Transformations and Feature Views

franciscojavierarceo:

This PR refactors the transformation system with a cleaner architecture that separates transformation logic from execution. Transformations should focus on HOW to transform data, while FeatureViews handle WHEN and WHERE to execute.

Changes:
  - Added: `feature_transformation` field to FeatureView for transformation logic
  - Added: `transform: bool = True` parameter to API methods for per-call control
  - Added: Schema validation for pre-transformed data (`transform=False`)
  - Added: Transformation execution logic in online/offline feature retrieval

```python
  # Define transformation logic
  @transformation(mode="python")
  def driver_features(inputs):
      return {"conv_plus_acc": [c + a for c, a in zip(inputs["conv_rate"], inputs["acc_rate"])]}

  # Create FeatureView with transformation
  fv = FeatureView(
      name="driver_computed_features",
      source=driver_source,
      entities=[driver],
      schema=[Field(name="conv_plus_acc", dtype=Float64)],
      feature_transformation=driver_features,  # ← NEW
      online=True,
  )

  # API-level transformation control
  # Execute transformations (default)
  features = store.get_online_features(
      features=["driver_computed_features:conv_plus_acc"],
      entity_rows=entity_rows,
      transform=True  # ← NEW (default)
  )

  # Skip transformations for external batch jobs
  features = store.get_online_features(
      features=["driver_computed_features:conv_plus_acc"],
      entity_rows=spark_transformed_rows,
      transform=False  # ← NEW: skip + validate schema
  )
```
#4584, #5716, #5689

# Misc

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -1279,17 +1279,16 @@ class FeatureStore:
                     view_name, self.project, allow_cache=True
                 )
 
-                # For historical retrieval, keep unified FeatureViews as FeatureViews
-                # (they'll have their transformations applied post-retrieval)
-                if (
+                # For historical retrieval, categorize appropriately
+                # Check for ODFV first since ODFVs also have feature_transformation
+                if isinstance(fv, OnDemandFeatureView):
+                    all_on_demand_feature_views.append(fv)
+                elif (
                     hasattr(fv, "feature_transformation")
                     and fv.feature_transformation is not None
                 ):
+                    # Unified FeatureViews - will have transformations applied post-retrieval
                     all_feature_views.append(cast(FeatureView, fv))
-                elif hasattr(fv, "__class__") and "OnDemandFeatureView" in str(
-                    type(fv)
-                ):
-                    all_on_demand_feature_views.append(cast(OnDemandFeatureView, fv))
                 else:
                     all_feature_views.append(cast(FeatureView, fv))
             except Exception:
--- a/sdk/python/feast/infra/offline_stores/dask.py
+++ b/sdk/python/feast/infra/offline_stores/dask.py
@@ -281,7 +281,9 @@ class DaskOfflineStore(OfflineStore):
                     join_key = feature_view.projection.join_key_map.get(
                         entity_column.name, entity_column.name
                     )
-                    join_keys.append(join_key)
+                    # Skip dummy entity for entityless feature views - it's not in the data source
+                    if join_key != DUMMY_ENTITY_ID:
+                        join_keys.append(join_key)
 
                 right_entity_key_columns = [
                     timestamp_field,
--- a/sdk/python/feast/infra/offline_stores/duckdb.py
+++ b/sdk/python/feast/infra/offline_stores/duckdb.py
@@ -157,6 +157,7 @@ class DuckDBOfflineStore(OfflineStore):
         registry: BaseRegistry,
         project: str,
         full_feature_names: bool = False,
+        **kwargs,
     ) -> RetrievalJob:
         return get_historical_features_ibis(
             config=config,
@@ -170,6 +171,7 @@ class DuckDBOfflineStore(OfflineStore):
             data_source_writer=_write_data_source,
             staging_location=config.offline_store.staging_location,
             staging_location_endpoint_override=config.offline_store.staging_location_endpoint_override,
+            **kwargs,
         )
 
     @staticmethod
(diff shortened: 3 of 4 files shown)


sdk/python/feast/transformation/base.py
```diff
@@ -130,18 +162,109 @@ def mainify(obj):
...
+                raise ValueError("At least one source must be provided for FeatureView")
+            elif len(sources) == 1:
+                # Single source - pass directly (works for DataSource or FeatureView)
+                source_param = sources[0]
+            else:
+                # Multiple sources - pass as list (must be List[FeatureView])
+                from feast.feature_view import FeatureView as FV
+
+                for src in sources:
+                    if not isinstance(src, (FV, type(src).__name__ == "FeatureView")):
```
ntkathole:
```suggestion
                    if not isinstance(src, FeatureView):
```

sdk/python/feast/feature_view.py
```diff
@@ -123,6 +127,9 @@ def __init__(
         tags: Optional[Dict[str, str]] = None,
         owner: str = "",
         mode: Optional[Union["TransformationMode", str]] = None,
+        feature_transformation: Optional[Transformation] = None,
+        when: Optional[Union[TransformationTiming, str]] = None,
```
ntkathole: Need updates to proto ? when and online_enabled not serialized in `from_proto`. 

Also, modify `__eq__()` and `__copy__()` for consider these new attributes ?

haoxu0:
I think `when` can be a bit confusing. even for 'batch', it can be 'on_read' or 'on_write'. I think on a different angle, should `transformation` be in the scope of `writing`, and isn't `transformation` only responsible for transforming the data?

FeatureView as the actually data assets, I think makes more sense to having the API.

franciscojavierarceo:
> I think `when` can be a bit confusing. even for 'batch', it can be 'on_read' or 'on_write'. I think on a different angle, should `transformation` be in the scope of `writing`, and isn't `transformation` only responsible for transforming the data?
> 
> FeatureView as the actually data assets, I think makes more sense to having the API.

Yeah, that's fair. @HaoXuAI I updated the PR, lmk what you think.

sdk/python/feast/feature_view.py
```diff
@@ -157,6 +166,36 @@ def __init__(
...
+        # Don't override feature_transformation if it's already set by subclass (e.g., BatchFeatureView)
+        if (
+            not hasattr(self, "feature_transformation")
+            or self.feature_transformation is None
+        ):
+            self.feature_transformation = feature_transformation
+        self.transform_when = transform_when
+
+        # Auto-infer online setting based on transform_when pattern
+        if transform_when in [
```
haoxu0: So why batch on read/write forces online to be true?
I think we need to think about if we really need a new tranform_when config. And if we need it, what is its relation with online/offline.
I think it might be counterintuitive that if this featureview is only used by batch or stream or on demand. what if the user defines it today as one mode and wants to change it to another one or allow both mode in the future?
My personal pint of view is removing any config for this. As long as you define the transformation, it will be applied to run before writing to destination. Use online/offline to control the destination only, if both are off, then no write. For any transformation needed on read, this should be configured at API level. Fo example, you can put get_online_feature(transform=true) to force transformation.

franciscojavierarceo: >I think we need to think about if we really need a new tranform_when config. And if we need it, what is its relation with online/offline.

Here's how things actually work:

1. Batch
 - Sometimes data can be transformed online
    1. On Reads
    2. On Writes
  - Sometimes not, and that's basically a batch transform with no feature server transformation and pure materialization
2. Streaming
 - Handles historical data the same way it handles online serving 

that's really what the `transform_when` indicates. 

We need something other  than `online` because the rest will be executed by the server and you can't orchestrate the transformation on write during retrieval.

franciscojavierarceo: I'm fine with renaming something but we need something equivalent to `write_to_online_store` that's more obvious.

haoxu0: online = true means write to online store.

`transform_on_write` should be removed from the server. it's a hacky way. https://github.com/feast-dev/feast/pull/5283. this PR doesn't get enough reviews I think.
https://github.com/feast-dev/feast/pull/5300. I think this one is also not ideal implementation as well.

When you use `get_online_feature`, you should assign it a `transform=true` to run transformation in "retrieval".

haoxu0: I've read through the feature server code and on demand feature view implementation. I think a lot code needs to be refactored. And if you think there is some lines of code won't work you can tag me. I'll try to explain it

franciscojavierarceo: Yes we agreed on this direction and I'm going to implement things that way. Just need to finish the work.

Issue #5776, opened 2025-12-15
ODFVs with write_to_online_store=True crash offline stores (they get queried from offline store, but shouldn't)


astronautas:
## Expected Behavior 
On-demand feature views (ODFVs) with `write_to_online_store=True` should be materialized to the online store, but they should never be queried from the offline store (ODFVs, no data for them in the offline store).

## Current Behavior
ODFVs with write_to_online_store=True are being queried from offline stores. In the code [here](https://github.com/feast-dev/feast/blob/27ec8ec26b2884d09df3ee3e648b96824e1f2b4c/sdk/python/feast/utils.py#L505)
, if an ODFV has the `write_to_online_store` flag set, it is treated as a regular feature view. As a result, it gets queried from the offline store.

However, ODFVs are not historical feature views and do not have any data in offline stores, so they should never be queried there.

## Steps to reproduce
Run `get_historical_features` for an ODFV with `write_to_online_store=True`.

### Specifications

- Version: 0.57.0
- Clickhouse offline store.

## Possible Solution
Not sure yet, but basically remove the condition? Is it even needed here? Seems incorrect...

astronautas:
@franciscojavierarceo You added this condition, could you elaborate why it's needed? It's crashing our offline store after upgrade, but we really need `write_to_online_store=True` for inference efficiency :D.

franciscojavierarceo:
It shouldn't be, let me try and take it out. Unless you can quickly, I'm working on something else for the pytorch announcement. If you can cut a quick PR we can get it in the next release (which I'm hoping to do today or tomorrow).

CC @ntkathole

astronautas:
@franciscojavierarceo I can do it myself, no worries ;). Just wanted to see why it was done that way and if it makes sense to remove it.

franciscojavierarceo:
Fixed, thanks @astronautas

astronautas:
no worries!

PR #5786, opened 2025-12-17
fix: Do not pull ODFVs from offline store (for write_to_online_store=True)

astronautas:

## Expected Behavior
On-demand feature views (ODFVs) with write_to_online_store=True should be materialized to the online store, but they should never be queried from the offline store (ODFVs, no data for them in the offline store).

## Current Behavior
ODFVs with write_to_online_store=True are being queried from offline stores. In the code [here](https://github.com/feast-dev/feast/blob/27ec8ec26b2884d09df3ee3e648b96824e1f2b4c/sdk/python/feast/utils.py#L505)
, if an ODFV has the write_to_online_store flag set, it is treated as a regular feature view. As a result, it gets queried from the offline store.

However, ODFVs are not historical feature views and do not have any data in offline stores, so they should never be queried there.

As discussed with @franciscojavierarceo, we remove the special condition to treat these materialized ODFVs and treat them as regular ODFVs, when retrieving offline features.

https://github.com/feast-dev/feast/issues/5776

# Misc

### DIFF
--- a/sdk/python/feast/utils.py
+++ b/sdk/python/feast/utils.py
@@ -499,8 +499,15 @@ def _group_feature_refs(
 
     # on demand view to on demand view proto
     on_demand_view_index: Dict[str, "OnDemandFeatureView"] = {}
+
+    # FIXME: broken for get_historical_features, for them, we need to treat ODFVs as ODFVs
+    # not like FVs (that's only convenient for online request)
     for view in all_on_demand_feature_views:
-        on_demand_view_index[view.projection.name_to_use()] = view
+        if view.projection and not view.write_to_online_store:
+            on_demand_view_index[view.projection.name_to_use()] = view
+        elif view.projection and view.write_to_online_store:
+            # we insert the ODFV view to FVs for ones that are written to the online store
+            view_index[view.projection.name_to_use()] = view
 
     # view name to feature names
     views_features = defaultdict(set)


astronautas:
@franciscojavierarceo

astronautas:
FAILED sdk/python/tests/unit/test_on_demand_python_transformation.py::TestOnDemandPythonTransformation::test_stored_writes - AssertionError: assert ['counter', '...put_datetime'] == ['conv_rate_p...put_datetime']
  At index 0 diff: 'counter' != 'conv_rate_plus_acc'
  Right contains 2 more items, first extra item: 'driver_id'
  Full diff:
    [
  -  'conv_rate_plus_acc',
     'counter',
  -  'current_datetime',
     'driver_id',
     'input_datetime',
    ]
FAILED sdk/python/tests/unit/online_store/test_online_retrieval.py::test_milvus_stored_writes_with_explode - KeyError: 'Feature document_id not found in projection milvus_explode_feature_view'

lots of unit tests failing. Will need to check :/

franciscojavierarceo:
So maybe I misunderstood, but ODFVs should be queryable from offline store. They should either execute the transformation at query time or be precomputed and stored already.

astronautas:
First, here's what the docs say:

https://docs.feast.dev/reference/beta-on-demand-feature-view

By setting write_to_online_store=True, transformations are applied during data ingestion, and the transformed features are stored in the online store. This can improve online feature retrieval performance by reducing computation during reads. Conversely, if write_to_online_store=False (the default if omitted), transformations are applied during feature retrieval.

----

Here's my thinking and experience using this new feature in prod. You call `store.materialize` an an ODFV, and it does get stored in online store (tested this ourselves).

ODFVs are precomputed and stored **within the online store** if `write_to_online_store=True`, or computed on request time otherwise. At least that's how their materialization is implemented within Feast code (we tested it in prod!). Offline stores are populated by offline jobs, outside of Feast e.g. Clickhouse, Spark, whatnot, which get_historical_features merely then slices off features from to materialize into online store e.g. materialize from Clickhouse into Redis.

@franciscojavierarceo `write_to_online_store` flag that enables materializing (precomputing) ODFVs has online, not offline store, within its semantical definition as well.

-----

Lastly:

the logic I propose to change in the PR above within `get_historical_features` from an offline store:

    elif view.projection and view.write_to_online_store:
has nothing to do with offline stores, but rather with online stores - it seems flawed to call it in offline_store.get_historical_features

sorry long text, but got to explain :D The test is missing `materialize` step, which loads ODFV values into online store, to be queried (think, cached).

----

Here's the gist. Hope it makes sense.

astronautas:
@franciscojavierarceo

astronautas:
the logic I propose to change in the PR above:

        elif view.projection and view.write_to_online_store:

has nothing to do with offline store, but rather with online store - that seems flawed.

sorry long text, but got to explain :D

sdk/python/feast/utils.py
```diff
@@ -500,11 +500,7 @@ def _group_feature_refs(
     # on demand view to on demand view proto
     on_demand_view_index: Dict[str, "OnDemandFeatureView"] = {}
     for view in all_on_demand_feature_views:
-        if view.projection and not view.write_to_online_store:
```
franciscojavierarceo, line 503: we need projections IIRC in the materialization

franciscojavierarceo: so if you want to avoid the transformation on get_historical_features then i would put that logic there.

astronautas: 👍 I'll check and adjust.

astronautas:
@franciscojavierarceo Closing PR. Will open up a related fix only targeting get_historical_features, need to re-think this.

PR #5787, opened 2025-12-19
feat: Enable static artifacts for feature server that can be used in Feature Transformations

franciscojavierarceo:
Improves speed of embedding transformation so static artifacts are loaded at build time.

Enables loading models, lookup tables, and other static resources once at server startup instead of per-request, improving performance for real-time inference.

Changes
  - Core: Adds `load_static_artifacts()` function to feature server with FastAPI lifespan integration
  - Template: Updated PyTorch NLP template with `static_artifacts.py` module and pre-loaded artifact usage
  - Tests: Added 6 comprehensive unit tests covering success, error handling, and persistence
  - Docs: Added static artifacts section to Python feature server documentation

N/A

# Misc


franciscojavierarceo:
CI was passing before the AWS issue. Going to merge.

PR #5802, opened 2025-12-30
fix: Remove upper bound on uvicorn to resolve dependency conflict wit…

### LINKED ISSUE #5677: Unsolvable dependency conflict between uvicorn and uvicorn-worker
### **Describe the bug**

When installing `feast>=0.54.0`, the package resolver fails due to an unsolvable dependency conflict.

  - `feast` requires `uvicorn[standard]<=0.34.0`.
  - `feast` also requires `uvicorn-worker`.
  - Recent versions of `uvicorn-worker` (e.g., `0.4.0`) require `uvicorn>=0.36.0`.

This creates a situation where no valid version of `uvicorn` can be found, causing the installation to fail.

### **To Reproduce**

1.  Create a `requirements.txt` file with the following content:
    ```
    feast==0.54.0
    ```
2.  Attempt to install the dependencies using `pip` or `uv`:
    ```bash
    pip install -r requirements.txt
    ```
3.  The installation will fail with a dependency resolution error.

### **Expected behavior**

The installation should complete successfully by finding a compatible set of packages.

### **Actual behavior**

The package installation fails with a resolution error similar to this:

```
ERROR: Could not produce a pip compatible constraints file.

[...]

* No single version of uvicorn meets all requirements
  uvicorn[standard]<=0.34.0,>=0.30.6 matches ['0.34.0']
    feast==0.54.0
  uvicorn>=0.36.0 matches ['0.37.0']
    uvicorn-worker==0.4.0
```

### **System Information**

  - Feast Version: `0.54.0`
  - Python Version: `3.12`
  - OS: `RHEL 9.6`

### **Proposed Solution / Workaround**

A temporary workaround is to manually constrain the version of `uvicorn-worker`:

```
uvicorn-worker<=0.3.0
```

I think the long-term fix would be for Feast to remove any upper version caps.

u-sheikh:
…h uvicorn-worker

The previous constraint `uvicorn[standard]>=0.30.6,<=0.34.0` conflicted with uvicorn-worker 0.4.0 which requires `uvicorn>=0.36.0`. This made it impossible to install Feast with the latest uvicorn-worker version.

Removes the upper bound since there are no breaking changes in uvicorn that affect Feast's usage of uvicorn.run() or uvicorn-worker.UvicornWorker.

Fixes #5677


ntkathole:
@u-sheikh Please run make lock-python-dependencies-all to update dependencies in requirements

u-sheikh:
> @ntkathole please let me know if this looks good, Thanks!

PR #5803, opened 2025-12-31
feat: Pass requested On-Demand Feature Views to Provider `get_historical_features`

Shizoqua:
### Summary
This PR fixes an interface/plumbing gap in historical retrieval by explicitly passing the requested On-Demand Feature Views (ODFVs) from FeatureStore.get_historical_features() into the provider’s get_historical_features() implementation.

### What changed
- Updated Provider.get_historical_features(...) signature to include:
   - `on_demand_feature_views: List[OnDemandFeatureView]`
- Updated FeatureStore.get_historical_features() to pass the computed on_demand_feature_views list into the provider call.
- Updated PassthroughProvider.get_historical_features() to accept the new argument (ignored for passthrough delegation).
- Updated the test FooProvider implementation to match the new provider interface.
- Added a unit test that asserts the provider receives the expected ODFVs list:
   - sdk/python/tests/unit/test_feature_store_passes_odfvs_to_provider.py

### Why
Providers may need the explicit set of requested ODFVs to correctly handle or route historical retrieval logic. Previously, only the combined feature_views list was passed, which made it awkward/impossible for providers to distinguish which ODFVs were requested.

### Testing
- Added unit test test_feature_store_passes_on_demand_feature_views_to_provider.

---

<a href="https://app.devin.ai/review/feast-dev/feast/pull/5803">
  <picture>
    Open with Devin
  </picture>
</a>


Shizoqua:
Good day @franciscojavierarceo 
I would appreciate it if my PR is looked at and reviewed if there is any comment you'd want to give in relations to the PR would be highly appreciated.

sdk/python/feast/infra/passthrough_provider.py
```diff
@@ -459,13 +459,15 @@ def get_historical_features(
         self,
         config: RepoConfig,
         feature_views: List[Union[FeatureView, OnDemandFeatureView]],
```
franciscojavierarceo, line 467: There shouldn't be a need to pass a separate `on_demand_feature_views` when `feature_views` containers the `OnDemandFeatureView` type. Is there a reason why we need it?

franciscojavierarceo: Oh I see the comment in the description here:

>Providers may need the explicit set of requested ODFVs to correctly handle or route historical retrieval logic. Previously, only the combined feature_views list was passed, which made it awkward/impossible for providers to distinguish which ODFVs were requested.


Can you elaborate a bit more about this? Which provider encountered this problem? The reason this was collapsed into `feature_views` was intentional (and where we want to consolidate things long term).

Shizoqua: Good question. While the Provider signature allows `feature_views` to include `OnDemandFeatureView`, at the FeatureStore call site we actually use `feature_views` to mean “offline-backed FeatureViews used for PIT joins” and we compute the “requested ODFVs” separately via `_group_feature_refs`.

Some provider implementations (esp. custom/downstream providers) need the explicit set of *requested* ODFVs to decide how to route/execute historical retrieval (e.g. whether to push down ODFV computation, whether to run a separate compute step, etc). Without this, providers have to re-derive “requested ODFVs” by re-parsing `feature_refs` / duplicating FeatureStore logic around FeatureService projections/full_feature_names, which is brittle.

Totally agree the long-term direction should be consolidation. If preferred, an alternative would be to make `feature_views` always include *only the requested* ODFVs (not just be typed to allow them) and drop the extra param. But that would require changing existing semantics at the call site and likely more refactoring/tests. This PR keeps current semantics and makes the requested ODFVs explicit.

Shizoqua:
Good day @franciscojavierarceo
I would appreciate it if my PR is looked at and reviewed.

Shizoqua:
Good day @franciscojavierarceo
I would appreciate it if my PR is looked at and reviewed.

Shizoqua:
Hello,
Good day @franciscojavierarceo
Just a quick reminder of my PR getting reviewed.

haoxu0:
Can you explain why you need this argument?

Shizoqua:
> Can you explain why you need this argument?

The new `on_demand_feature_views` argument is needed because `FeatureStore.get_historical_features()` already resolves which OnDemandFeatureViews are actually requested (via `_group_feature_refs`), but historically we were not passing that information to the Provider.
Without this argument, the Provider only sees feature_views + feature_refs and has to:
- either re-resolve ODFVs from the feature refs (i.e. duplicating logic), or
- conservatively treat ODFVs as part of the full graph and potentially execute transformations for ODFVs that aren’t requested, which can add unnecessary joins/transforms and overhead.

Passing `on_demand_feature_views` makes the API explicit so provider implementations that can optimize or need this for correct planning like compute-engine based retrieval can do so without re-deriving it.

haoxu0:
> > Can you explain why you need this argument?
> 
> The new `on_demand_feature_views` argument is needed because `FeatureStore.get_historical_features()` already resolves which OnDemandFeatureViews are actually requested (via `_group_feature_refs`), but historically we were not passing that information to the Provider. Without this argument, the Provider only sees feature_views + feature_refs and has to:
> 
> * either re-resolve ODFVs from the feature refs (i.e. duplicating logic), or
> * conservatively treat ODFVs as part of the full graph and potentially execute transformations for ODFVs that aren’t requested, which can add unnecessary joins/transforms and overhead.
> 
> Passing `on_demand_feature_views` makes the API explicit so provider implementations that can optimize or need this for correct planning like compute-engine based retrieval can do so without re-deriving it.

Although I don’t fully support this ad hoc change, since it’s not user-facing, I’m okay with approving it. Could you please add a TODO and create a follow-up issue to flag it for a proper long-term fix?

Shizoqua:
> Although I don’t fully support this ad hoc change, since it’s not user-facing, I’m okay with approving it. Could you please add a TODO and create a follow-up issue to flag it for a proper long-term fix?

Okay, I'm on it now

Shizoqua:
> Although I don’t fully support this ad hoc change, since it’s not user-facing, I’m okay with approving it. Could you please add a TODO and create a follow-up issue to flag it for a proper long-term fix?

Added a TODO in FeatureStore.get_historical_features to flag this as an interim interface change. I’ve also open a follow-up issue to track a proper long-term cleanup of the historical retrieval Provider API (request object / clearer contract) and will link it here once created.

Shizoqua:
Hello,
Good day @franciscojavierarceo
Just a quick reminder of my PR getting reviewed or processed

Shizoqua:
Hello,
Good day
Please can my PR get reviewed?

PR #5818, opened 2026-01-06
feat: Add GCS registry store in Go feature server

samuelkim7:

This PR implements the GCS registry store for the Go feature server. While the Go feature server previously supported only local file-based and S3-based registry stores, this update enables seamless integration for users operating on Google Cloud Platform (GCP). By adding GCS support, we are providing greater flexibility to users and making the Go feature server production-ready for a wider range of enterprise use cases.

None. This pull request implements a new feature rather than fixing an existing bug or issue.

# Misc


franciscojavierarceo (approved):
nice, thank you for this!

PR #5825, opened 2026-01-09
fix: Update model_validator to use instance method signature (Pydantic v2.12 deprecation)

WhiteLotusLA:
## Summary

This PR fixes a Pydantic v2.12+ deprecation warning in `OidcClientAuthConfig._validate_credentials`.

**The warning:**
```
DeprecationWarning: Using `@model_validator` with mode='after' on a classmethod is deprecated
```

## Changes

In Pydantic v2.12, using `@model_validator(mode="after")` with a classmethod-style signature (`cls, values`) was deprecated in favor of an instance method signature (`self`).

**Before:**
```python
@model_validator(mode="after")
def _validate_credentials(cls, values):
    d = values.__dict__ if hasattr(values, "__dict__") else values
    has_user_pass = bool(d.get("username")) and bool(d.get("password"))
    # ...
    return values
```

**After:**
```python
@model_validator(mode="after")
def _validate_credentials(self):
    has_user_pass = bool(self.username) and bool(self.password)
    # ...
    return self
```

## Benefits

- Eliminates deprecation warning spam in user logs
- Uses direct attribute access instead of dict gymnastics
- Cleaner, more idiomatic Pydantic v2 code
- No functional changes - same validation logic

## Testing

This is a non-breaking change that maintains identical validation behavior. The only difference is the method signature and attribute access pattern.

## References

- [Pydantic v2 model_validator documentation](https://docs.pydantic.dev/latest/concepts/validators/#model-validators)
- Pydantic changelog noting this deprecation


franciscojavierarceo:
Can you fix DCO?

PR #5827, opened 2026-01-10
feat: Add dbt integration for importing models as FeatureViews

### LINKED ISSUE #3335: Feast <> dbt integration
- Users commonly want to leverage their existing tables created by dbt models within Feast. Right now, this requires manual creation
- When materializing features to the online store, often, it's desirable to have dbt run first to create the upstream table prior to materializing an interval. This requires manual work today.

- A mechanism to import dbt models (e.g. `feast dbt-import`) that lets users import from a directory of models, or configure a directory and enable individual models

- The above mechanism should generate a data source + feature view with pre-populated tags to track the origin dbt model. Something like
```python
driver_stats_source = BigQuerySource(
    name="driver_hourly_stats_source",
    table="AUTO-IMPORTED-NAME",
    tags={ 
       "feast.dev/dbt-model-path": "INSERT MODEL PATH" 
       "feast.dev/dbt-model-serialized": "<serialized dbt model to be shown in UI>" 
    }
)
```

- Then these models could potentially be shown directly in the Feast Web UI too (there was prior art here e.g. https://github.com/feast-dev/feast/blob/v0.24-branch/ui/src/pages/data-sources/DataSourceDbt.tsx 

- Potentially, if configured, materialize can also run dbt run for the upstream model

**Additional context**

YassinNouh21:
Adds CLI commands to import dbt models as Feast FeatureViews, enabling automatic generation of Feast objects from dbt manifest.json files.

- `feast dbt list` - Discover dbt models available for import
- `feast dbt import` - Create Feast FeatureViews from dbt models
- `--output` option to generate Python files instead of applying to registry
- Supports BigQuery, Snowflake, and File data sources
- Maps 38 dbt data types to Feast types (including ARRAY and NUMBER with precision)
- Preserves dbt metadata (tags, descriptions) in generated Feast objects

Closes #3335

## Does this PR introduce a user-facing change
Yes. Users can now import dbt models as Feast FeatureViews using:
```bash
# List models with 'feast' tag
feast dbt list -m target/manifest.json --tag feast

# Import models to registry
feast dbt import -m target/manifest.json -e driver_id --tag feast

# Generate Python file instead
feast dbt import -m target/manifest.json -e driver_id --output features.py
```

## Test plan
- [x] Unit tests for parser, mapper, and codegen modules
- [x] Tested with real dbt projects at 3 complexity levels
- [x] Verified generated Python files are syntactically correct and importable

Signed-off-by: yassinnouh21 <yassinnouh21@gmail.com>


sdk/python/feast/dbt/parser.py
```diff
@@ -0,0 +1,304 @@
...
+            else:
+                columns = []
+
+            return DbtModel(
+                name=model_name,
+                unique_id=node_id,
+                database=getattr(node, "database", "") or "",
+                schema=getattr(node, "schema_", "")
+                or getattr(node, "schema", "")
+                or "",
```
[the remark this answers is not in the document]
YassinNouh21: This code was removed in the recent refactor that simplified the parser to rely solely on `dbt-artifacts-parser` (commit 5db19523). The triple fallback is no longer needed as we now exclusively use typed objects from the library.

YassinNouh21:
@franciscojavierarceo it passed the CI and I reviewed your comments

franciscojavierarceo:
Can you add documentation for this? 🙏

.github/workflows/unit_tests.yml
```diff
@@ -32,6 +32,8 @@ jobs:
           enable-cache: true
       - name: Install dependencies
         run: make install-python-dependencies-ci
+      - name: Install dbt extra dependencies
```
ntkathole: This isn't required since earlier step is install-python-dependencies-ci, which installs all dependencies. Include dbt in feast[ci] and run make lock-python-dependencies-all to update the requirements files.

setup.py
```diff
@@ -167,6 +167,8 @@
 
 MILVUS_REQUIRED = ["pymilvus==2.4.15", "milvus-lite==2.4.12", "setuptools>=60,<81"]
 
+DBT_REQUIRED = ["dbt-artifacts-parser>=0.6.0,<1"]
```
ntkathole, line 168: Add same to [pyproject.toml](https://github.com/feast-dev/feast/blob/master/pyproject.toml)

YassinNouh21:

Thanks for the reviews! I've addressed all the feedback:

- Added comprehensive documentation for the dbt integration (commit 53932ff)
- Added `dbt-artifacts-parser` to both `setup.py` and `pyproject.toml` under `feast[ci]` (commit fb40e93)

All CI checks are passing and there are no merge conflicts. The PR should be ready for another review when you have a chance.

Let me know if there's anything else needed! 🙏

docs/how-to-guides/dbt-integration.md
```diff
@@ -0,0 +1,370 @@
...
+    Tags: feast
+```
+
+### 5. Import models as Feast definitions
+
+Generate a Python file with Feast object definitions:
+
+```bash
+feast dbt import target/manifest.json \
+    --entity-column driver_id \
```
franciscojavierarceo, line 126: what happens when a user has multiple dbt models with multiple entities? ideally we can go from some sort of metadata tag to autogenerating the Entity.

sdk/python/feast/cli/dbt_import.py
```diff
@@ -0,0 +1,379 @@
...
+    help="TTL (time-to-live) in days for feature views",
+)
+@click.option(
+    "--dry-run",
+    is_flag=True,
+    default=False,
+    help="Preview what would be created without applying changes",
+)
+@click.option(
+    "--exclude-columns",
```
franciscojavierarceo, line 77: probably we should allow users to use either exclude or include because sometimes users can have really big models

sdk/python/feast/cli/dbt_import.py
```diff
@@ -0,0 +1,379 @@
...
+    default=None,
+    help="Comma-separated list of columns to exclude from features",
+)
+@click.option(
+    "--output",
+    "-o",
+    type=click.Path(),
+    default=None,
+    help="Output Python file path (e.g., features.py). Generates code instead of applying to registry.",
```
franciscojavierarceo, line 86: In the general case of multiple tables -> multiple feature views, we should probably generate one python file per dbt model

sdk/python/feast/cli/dbt_import.py
```diff
@@ -0,0 +1,379 @@
...
+        feast dbt import -m target/manifest.json -e driver_id --tag feast --output features.py
+    """
+    from feast.dbt.mapper import DbtToFeastMapper
+    from feast.dbt.parser import DbtManifestParser
+
+    # Parse manifest
+    click.echo(f"{Fore.CYAN}Parsing dbt manifest: {manifest_path}{Style.RESET_ALL}")
+
+    try:
+        parser = DbtManifestParser(manifest_path)
```
franciscojavierarceo, line 129: it makes sense to do the manifest but allow users to interactively test a single model could have utility as well so that they can quickly iterate and run feast apply until they get it to work (good devx)

sdk/python/feast/cli/dbt_import.py
franciscojavierarceo: would make sense to raise if users tried to use both

sdk/python/feast/cli/dbt_import.py
```diff
@@ -0,0 +1,379 @@
...
+
+    # Apply to Feast
+    click.echo(f"\n{Fore.CYAN}Applying to Feast registry...{Style.RESET_ALL}")
+
+    repo = ctx.obj["CHDIR"]
+    fs_yaml_file = ctx.obj["FS_YAML_FILE"]
+    cli_check_repo(repo, fs_yaml_file)
+    store = create_feature_store(ctx)
+
+    store.apply(all_objects)
```
franciscojavierarceo, line 292: i feel the user should generate the code first and then separately run feast apply manually

sdk/python/feast/dbt/codegen.py
```diff
@@ -0,0 +1,385 @@
...
+    # Map type objects to their names
+    type_map = {
+        String: "String",
+        Int32: "Int32",
+        Int64: "Int64",
+        Float32: "Float32",
+        Float64: "Float64",
+        Bool: "Bool",
+        UnixTimestamp: "UnixTimestamp",
+        Bytes: "Bytes",
```
franciscojavierarceo, line 154: i guess Image and PdfBytes wouldn't be sensible here

docs/how-to-guides/dbt-integration.md
```diff
@@ -0,0 +1,370 @@
+# Importing Features from dbt
+
+This guide explains how to use Feast's dbt integration to automatically import dbt models as Feast FeatureViews. This enables you to leverage your existing dbt transformations as feature definitions without manual duplication.
+
+## Overview
```
franciscojavierarceo, line 16: we should probably highlight we have only supported Snowflake, BigQuery, and file source, yeah? We should probably call this an Alpha feature

haoxu0:
looks good to me

YassinNouh21:
Thanks for catching this! The current code already handles this correctly at lines 368-370:

```python
desc = model.description[:80] + (
    "..." if len(model.description) > 80 else ""
)
```

The ellipsis is only added when `len(model.description) > 80`.

Examples:
```python
# Short (40 chars)
"This table contains driver features" → "This table contains driver features"

# Long (120 chars)  
"This is a very long description..." → "This is a very long description that goes on and on and on and contains lots..."
```

The conditional already checks length before adding ellipsis.

---

Re: dbt_import.py:368 description truncation comment

YassinNouh21:
You're right - ImageBytes and PdfBytes wouldn't make sense here. While these types exist in Feast, dbt manifests only expose generic `BYTES` type without semantic information about whether bytes represent images, PDFs, or other binary data.

Example:
```sql
-- dbt model
SELECT 
    user_id,
    profile_photo,  -- BigQuery type: BYTES
    resume_pdf      -- BigQuery type: BYTES  
FROM users
```

dbt manifest only shows `"data_type": "BYTES"` for both columns - no way to distinguish images from PDFs. The mapping only includes types that can actually appear in dbt artifacts.

---

Re: codegen.py:152 ImageBytes/PdfBytes comment

sdk/python/feast/cli/dbt_import.py
franciscojavierarceo: @YassinNouh21 this is one change I do want to include here. Can we add a parameter instead that says "apply" to the CLI then? by default we shouldn't apply in the dbt code gen. 

It should be:

```bash
feast dbt-import --apply=True
```

Or something like that.

franciscojavierarceo (approved):
okay, going to ship this. @YassinNouh21 it'd be great if we can get a proper integration test in a follow up PR that sets up a light weight local dbt project that then does the compilation but i don't want to block that here

franciscojavierarceo:
> okay, going to ship this. @YassinNouh21 it'd be great if we can get a proper integration test in a follow up PR that sets up a light weight local dbt project that then does the compilation but i don't want to block that here

@copilot can you make a github issue for this?

YassinNouh21:
## Code Review - Issues Created

I've reviewed PR #5827 and created the following issues to track improvements and edge cases:

### High Priority
1. **#5869** - Add integration tests for dbt import with local dbt project
   - As mentioned by @franciscojavierarceo, need proper e2e tests

2. **#5870** - Improve error handling for missing dbt-artifacts-parser (good first issue)
   - CLI doesn't catch ImportError from parser

3. **#5874** - FileSource generates invalid hardcoded paths
   - `/data/{model}.parquet` likely doesn't exist

### Validation & Safety
4. **#5871** - Add validation for timestamp field data type
   - Should verify timestamp field is actually a timestamp type

5. **#5876** - Validate entity column data type is appropriate
   - Should warn if entity column is FLOAT, BYTES, etc.

### Feature Enhancements
6. **#5872** - Support multiple entities per FeatureView
   - Real-world use cases often need composite keys

7. **#5873** - Make TTL default configurable and increase default value
   - 1 day default is very aggressive, consider 7 days

8. **#5875** - Add support for data source connection configuration
   - Need way to specify BigQuery project, Snowflake account, etc.

## Overall Assessment

The PR is well-structured with good test coverage for the core functionality. The main gaps are:
- Integration tests with actual dbt projects
- Validation of input data types
- Configuration flexibility for production use cases

Great work on getting this feature merged! 🎉

PR #5881, opened 2026-01-20
feat: Optimize container infrastructure for production

franciscojavierarceo:
- Add multi-worker configuration with auto-scaling (CPU * 2 + 1)
- Add worker connections, max-requests, and jitter parameters
- Optimize registry TTL from 2s/5s to 60s for reduced refresh overhead
- Support --workers=-1 for automatic worker count calculation
- Add worker recycling to prevent memory leaks

Expected Impact:
- 300-500% throughput increase with proper worker scaling
- Reduced registry refresh overhead
- Better resource utilization in containerized environments

# What this PR does / why we need it:
make serving faster

# Which issue(s) this PR fixes:

# Misc

PR #5926, opened 2026-01-30
feat: Added online server worker config support in operator

ntkathole:

This PR adds support for configuring the new serving options introduced in PR #5881 via the Feast Operator. These options allow users to optimize the online store (feature server) for production deployments.

Example Usage:

```
apiVersion: feast.dev/v1
kind: FeatureStore
metadata:
  name: my-feast
spec:
  feastProject: my_project
  services:
    onlineStore:
      server:
        workerConfigs:
          workers: -1              # Auto-calculate based on CPU cores
          workerConnections: 2000
          maxRequests: 1000
          maxRequestsJitter: 50
          keepAliveTimeout: 30
          registryTTLSeconds: 60
```

---

<a href="https://app.devin.ai/review/feast-dev/feast/pull/5926">
  <picture>
    Open with Devin
  </picture>
</a>


Makefile
```diff
@@ -639,6 +639,12 @@ build-feast-operator-docker: ## Build Feast Operator Docker image
 	VERSION=$(VERSION) \
 	$(MAKE) docker-build
 
+build-feast-operator-docker-on-mac: ## Build Feast Operator Docker image on Mac
```
franciscojavierarceo, line 642: Why can't we just Donna if statement for the OS?

franciscojavierarceo (approved):
One small nit please but otherwise lgtm

Makefile
ntkathole: Adding OS auto-detection would force `--platform linux/amd64` on all Mac builds, breaking users who intentionally want native ARM images for local development/testing. Keeping separate _on_mac targets gives explicit control: native arch by default, cross-platform when explicitly requested.

Alternatively, we could use an environment variable (e.g., DOCKER_PLATFORM) or configure Docker's default platform at the system level, but these add extra configuration overhead for contributors to remember.

franciscojavierarceo: I have to remember to use this command though so still same problem but okay

Issue #5426, opened 2025-06-05
Create DocEmbedder class along with RAGRetriever


jyejare:
Relates to Jira issue: RHOAIENG-26770

patelchaitany:
Hey I like to work On this issue.

patelchaitany:
Hey I thought of Implementing this plan and i am looking for any feedback Like is there something which i missed or Which i can do the better. 

Implementation Plan:

1. Document Embedder Class - A high-level orchestrator that coordinates the entire ingestion pipeline.

2. Chunker - Handles chunking for different types of data. Users can extend or override methods to handle data in their own way.

3. Logical Layer - Defines how to represent the data in the FeatureView schema. This is written by the Data Scientist or Data Engineer.

4. Embedding Class - Handles embedding generation. This is designed as a separate class so it can support different modalities (text, images, etc.) by simply extending the base class.

5. Storage - Write the processed data to both the offline and online store.


Here is the Diagram 

<img width="859" height="425" alt="Image" src="https://github.com/user-attachments/assets/7da0d7af-df43-4335-af78-af80ed6728a1" />

I have some concern Regarding this - "The RAGRetriever uses the DocEmbedder object for adding it to offline / online store."

DocEmbedder and FeastRAGRetriever are independent classes. The FeastRAGRetriever provides retrieval functionality at query time. The DocEmbedder is only needed during data ingestion. Keeping them separate maintains clean code and clear responsibilities.

jyejare:
Hey @patelchaitany , Thanks for putting this all together. Initially, when I put this together (and I am open to have any modifications in the architecture), I saw DocEmbedder as a mediator between the raw documents and embeddings. So the mindmap I had in mind for RAGRetriever was like:

Raw Documents -> Chunking -> embeddings -> Offload to Store -> read embeddings based on the query -> Top-K retrievals.

Its like one shot architecture from raw docs to embeddings. But even then, I wanted to keep the Embedder class separate, which would take care of pulling the docs and saving them in the offline store.

So, I think we are on the same page, just only extra thing I had in mind was to plug DocEmbedder with RAGRetreiver as described in mind map.

Let me know your thoughts.

PR #5973, opened 2026-02-16
feat: Created DocEmbedder class

patelchaitany:

This PR adds a Document Embedder capability to Feast, allowing users to go from raw documents to embeddings stored in the online vector store in a single step. It handles chunking, embedding generation, and writing the results to the online store — providing an end-to-end ingestion pipeline for RAG workflows within Feast.

### sdk/python/feast/chunker.py

Defines the document chunking layer. Provides:

- A BaseChunker abstract class that users can extend with custom chunking strategies
- A built-in TextChunker that splits plain text by word count with configurable size, overlap, and thresholds
- A chunk_dataframe() convenience method to chunk all documents in a DataFrame

Currently only basic text chunking is implemented. There is room for improvement — future iterations can support more advanced strategies like semantic chunking, sentence-aware splitting, or format-specific chunkers (PDF, HTML, etc.).

### sdk/python/feast/embedder.py

Defines the embedding generation layer. Provides:

- A BaseEmbedder abstract class with a modality-routing registry for extensibility
- A MultiModalEmbedder with built-in support for text (via sentence-transformers) and image (via CLIP) embeddings
- Lazy model loading and configurable batch processing

### sdk/python/feast/doc_embedder.py

The high-level orchestrator that coordinates chunking, embedding, and storage. Provides:

- A DocEmbedder class that runs the full pipeline: chunk -> embed -> logical layer -> write to online store
- A user-defined logical layer function to map output to the FeatureView schema
- Auto-generation of a default FeatureView if the user does not define their own (controlled by create_feature_view). This is a basic implementation intended as a starting point

### sdk/python/feast/__init__.py

Updated to export DocEmbedder, LogicalLayerFn, BaseChunker, TextChunker, ChunkingConfig, BaseEmbedder, MultiModalEmbedder, and EmbeddingConfig as part of Feast's public API.

Create DocEmbedder class along with RAGRetriever #5426 

# Misc

---

<a href="https://app.devin.ai/review/feast-dev/feast/pull/5973" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


sdk/python/feast/embedder.py
```diff
@@ -0,0 +1,235 @@
...
+
+                processed = self.image_processor(images=images, return_tensors="pt")
+            finally:
+                for img in opened:
+                    img.close()
+
+            embeddings = self.image_model.get_image_features(**processed)
+            all_embeddings.append(embeddings.detach().numpy())
+
+        return np.concatenate(all_embeddings, axis=0)
```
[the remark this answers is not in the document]
patelchaitany: No, then it is not intended use case as If the Image is empty then it should not call the _embed_image or Image is are corrupted then it should throw the error or we handle those error?

patelchaitany:
@ntkathole @jyejare can You pls review this PR and let me know if any changes is needed.

sdk/python/tests/unit/test_doc_embedder.py
```diff
@@ -0,0 +1,654 @@
...
+    )
+
+
+class TestTextChunker:
+    def test_basic_chunking(self):
+        """Given ~200 words, returns chunks with correct keys and count."""
+        chunker = TextChunker()
+        text = " ".join([f"word{i}" for i in range(200)])
+
+        chunks = chunker.load_parse_and_chunk(source=text, source_id="doc1")
```
jyejare: I think chunker should decide the text in source to chunk the text, we should not need to manually feed that.

sdk/python/tests/unit/test_doc_embedder.py
```diff
@@ -0,0 +1,654 @@
...
+    """BaseEmbedder is abstract and cannot be instantiated directly."""
+    with pytest.raises(TypeError):
+        BaseEmbedder()
+
+
+class TestMultiModalEmbedder:
+    def test_supported_modalities(self):
+        """After init, supported_modalities returns text and image."""
+        embedder = MultiModalEmbedder()
+        modalities = embedder.supported_modalities()
```
jyejare: Supported modalities can be set as a property

sdk/python/tests/unit/test_doc_embedder.py
```diff
@@ -0,0 +1,654 @@
...
+            ["hello", "world"],
+            batch_size=64,
+            show_progress_bar=True,
+        )
+        assert result.shape == (2, 3)
+
+    def test_text_embed_custom_config(self):
+        """Text embedding respects custom EmbeddingConfig values."""
+        config = EmbeddingConfig(batch_size=16, show_progress=False)
+        embedder = MultiModalEmbedder(config=config)
```
jyejare, line 180: Any reasons we r calling it as multimodal embedder ?

sdk/python/tests/unit/test_doc_embedder.py
```diff
@@ -0,0 +1,654 @@
...
+
+    def test_lazy_loading(self):
+        """Models are None until first accessed."""
+        embedder = MultiModalEmbedder()
+
+        assert embedder._text_model is None
+        assert embedder._image_model is None
+        assert embedder._image_processor is None
+
+    def test_custom_modality_registration(self):
```
jyejare: Repeat test ?

sdk/python/tests/unit/test_doc_embedder.py
patelchaitany: Yes, Chunker can decide the which field is text in the DataFrame We Just need to pass the column name in the chunk_dataframe function of the Chunker class.

This test only for testing that load_parse_and_chunk function return the match the required return type.

sdk/python/tests/unit/test_doc_embedder.py
patelchaitany: This MultiModalEmbedder Supports the Text and Image Both Modality.

sdk/python/tests/unit/test_doc_embedder.py
patelchaitany: This is for when we register new Modality then It will correctly route to the New Modality.

But I agree we can remove this test.

sdk/python/feast/embedder.py
```diff
@@ -0,0 +1,238 @@
...
+        Args:
+            modality: The modality to query (e.g. "text", "image").
+
+        Returns:
+            The embedding dimension, or None if unknown.
+        """
+        if modality == "text":
+            return self.text_model.get_sentence_embedding_dimension()
+        elif modality == "image":
+            return self.image_model.config.vision_config.hidden_size
```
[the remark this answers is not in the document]
patelchaitany: No this this is new works and the projection_dim dose not work.

ntkathole:
@patelchaitany filename typo - `examples/rag-retriever/rag_feast_docebedder.ipynb` should be `rag_feast_docembedder.ipynb`

sdk/python/feast/doc_embedder.py
```diff
@@ -0,0 +1,348 @@
...
+        repo_path: Path to the feature repo directory.
+        feature_view_name: Name of the feature view to create.
+        vector_length: Dimension of the embedding vectors. Should match the
+            output dimension of the embedding model being used. Defaults to
+            384 (matching the default all-MiniLM-L6-v2 model).
+
+    Returns:
+        Path to generated file.
+    """
+    import os
```
ntkathole: import os already at the top

sdk/python/feast/doc_embedder.py
```diff
@@ -0,0 +1,348 @@
...
+
+        return _DEFAULT_VECTOR_LENGTH
+
+    def save_to_online_store(self, df: pd.DataFrame, feature_view_name: str) -> None:
+        """
+        Save the embedded documents to the online store.
+        """
+        from feast.feature_store import FeatureStore
+
+        store = FeatureStore(repo_path=self.repo_path)
```
ntkathole: Each call to save_to_online_store creates a fresh `FeatureStore(repo_path=...)`. This should be cached or created once during initialization.

sdk/python/feast/doc_embedder.py
```diff
@@ -0,0 +1,348 @@
...
+            df = current_logical_layer_fn(df)
+        elif (
+            current_logical_layer_fn is default_logical_layer_fn
+            and source_column != "text"
+            and column_mapping[source_column][1] != "text_embedding"
+        ):
+            raise ValueError(
+                f"source_column='{source_column}' is not compatible with "
+                f"default_logical_layer_fn. Provide a custom logical_layer_fn."
+                f"Mapping coulmn should be text: (text, text_embedding)"
```
ntkathole:
```suggestion
                f"Mapping column should be text: (text, text_embedding)"
```

sdk/python/feast/embedder.py
```diff
@@ -0,0 +1,238 @@
...
+        return None
+
+    # Text Embedding
+    @property
+    def text_model(self):
+        if self._text_model is None:
+            from sentence_transformers import SentenceTransformer
```
ntkathole, line 179: we have not added sentence-transformers to project dependencies or feast[rag]. Probably we should check if import works fine or add it as dependencies

sdk/python/feast/chunker.py
```diff
@@ -0,0 +1,150 @@
...
+
+        Returns:
+            List of chunk dicts with keys:
+                - chunk_id: str
+                - original_id: str
+                - text: str
+                - chunk_index: int
+                - (any additional metadata)
+        """
+        raise NotImplementedError("Subclasses must implement load_parse_and_chunk")
```
ntkathole: This is abstractmethod so no need of NotImplementedError, decorator already prevents instantiation

sdk/python/feast/embedder.py
```diff
@@ -0,0 +1,238 @@
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from typing import Any, Callable, List, Optional
+
+import numpy as np
```
ntkathole: Consider lazy-loading numpy too

sdk/python/feast/embedder.py
```diff
@@ -0,0 +1,238 @@
...
+
+                processed = self.image_processor(images=images, return_tensors="pt")
+            finally:
+                for opened_img in opened:
+                    opened_img.close()
+
+            embeddings = self.image_model.get_image_features(**processed)
+            all_embeddings.append(embeddings.detach().numpy())
+
+        return np.concatenate(all_embeddings, axis=0)
```
ntkathole, line 241: `_embed_image` returns raw get_image_features output which is not normalized, but `_embed_text` returns the output of SentenceTransformer.encode which produces normalized embeddings by default. This inconsistency will cause problems when computing cosine similarity across modalities or even within image embeddings.

sdk/python/feast/embedder.py
patelchaitany: we cannot do the lazy of numpy as it is required for the Type checking.

sdk/python/feast/doc_embedder.py
```diff
@@ -0,0 +1,353 @@
...
+            if original_cwd is not None:
+                os.chdir(original_cwd)
+
+    def embed_documents(
+        self,
+        documents: pd.DataFrame,
+        id_column: str,
+        source_column: str,
+        type_column: Optional[str] = None,
+        column_mapping: Optional[dict[str, tuple[str, str]]] = None,
```
ntkathole: column_mapping I think simply can be tuple, column_mapping[source_column] is only used

sdk/python/feast/doc_embedder.py
```diff
@@ -0,0 +1,306 @@
...
+# Source
+{feature_view_name.replace(" ", "_").replace("-", "_")}_source = FileSource(
+    name="{feature_view_name}_source",
+    path="data/{feature_view_name}.parquet",
+    timestamp_field="event_timestamp",
+)
+
+# FeatureView
+{feature_view_name.replace(" ", "_").replace("-", "_")} = FeatureView(
+    name="{feature_view_name}",
```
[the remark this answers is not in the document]
ntkathole: @patelchaitany this is valid

sdk/python/feast/doc_embedder.py
```diff
@@ -0,0 +1,308 @@
...
```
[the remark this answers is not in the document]
ntkathole: needs to resolve

sdk/python/feast/embedder.py
ntkathole: Gate this behind type check
```
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    import numpy as np
```
importing feast will fail if numpy is not installed

sdk/python/feast/doc_embedder.py
```diff
@@ -0,0 +1,353 @@
...
+                "logical_layer_fn must be a LogicalLayerFn or a function that takes a DataFrame and returns a DataFrame"
+            )
+        if create_feature_view:
+            resolved_vector_length = self._resolve_vector_length(vector_length, "text")
+            generate_repo_file(
+                repo_path=repo_path,
+                feature_view_name=feature_view_name,
+                vector_length=resolved_vector_length,
+            )
+        self.apply_repo()
```
ntkathole: Consider making this explicit, may be a separate setup() method or a auto_apply=True parameter. The constructor always calls apply_repo(), which invokes apply_total, a heavy operation that touches the registry

sdk/python/feast/doc_embedder.py
```diff
@@ -0,0 +1,353 @@
...
+        vector_length: Optional[int] = None,
+    ):
+        self.repo_path = repo_path
+        self.yaml_path = os.path.join(Path(repo_path).resolve(), yaml_file)
+        self.feature_view_name = feature_view_name
+        self.chunker = chunker or TextChunker()
+        self.embedder = embedder or MultiModalEmbedder()
+        self.store: Optional[FeatureStore] = None
+
+        if isinstance(logical_layer_fn, LogicalLayerFn):
```
ntkathole: what is `isinstance` check for ? it's not validating anything

pyproject.toml
```diff
@@ -112,6 +112,7 @@ qdrant = ["qdrant-client>=1.12.0"]
 rag = [
     "transformers>=4.36.0",
     "datasets>=3.6.0",
+    "sentence-transformers>=3.0.0"
```
ntkathole:
```suggestion
    "sentence-transformers>=3.0.0",
```

sdk/python/feast/chunker.py
```diff
@@ -0,0 +1,150 @@
...
+        """
+
+        chunks_per_row = df.apply(
+            lambda row: self.load_parse_and_chunk(
+                row[source_column],
+                str(row[id_column]),
+                source_column,
+                row[type_column] if type_column else None,
+            ),
+            axis=1,
```
ntkathole: def.apply with axis=1 is going to be slow for large dataframes, instead use simple list compression

```
all_chunks = []
type_values = df[type_column] if type_column else [None] * len(df)
for src, doc_id, doc_type in zip(df[source_column], df[id_column], type_values):
    chunks = self.load_parse_and_chunk(src, str(doc_id), source_column, doc_type)
    all_chunks.extend(chunks)

if not all_chunks:
    return pd.DataFrame(
        columns=["chunk_id", "original_id", source_column, "chunk_index"]
    )
return pd.DataFrame(all_chunks)
```

sdk/python/feast/doc_embedder.py
```diff
@@ -0,0 +1,368 @@
...
+        }
+    )
+
+
+def generate_repo_file(
+    repo_path: str,
+    feature_view_name: str = "text_feature_view",
+    vector_length: int = 384,
+) -> str:
+    from feast.repo_operations import is_valid_name
```
ntkathole: import this after docstring

sdk/python/feast/doc_embedder.py
```diff
@@ -0,0 +1,368 @@
...
+        self.yaml_path = os.path.join(Path(repo_path).resolve(), yaml_file)
+        self.feature_view_name = feature_view_name
+        self.chunker = chunker or TextChunker()
+        self.embedder = embedder or MultiModalEmbedder()
+        self.store: Optional[FeatureStore] = None
+        sig = inspect.signature(logical_layer_fn)
+        params = list(sig.parameters.values())
+        if (
+            len(params) != 1
+            or params[0].name != "df"
```
ntkathole: Probabably name check can be removed?

sdk/python/feast/doc_embedder.py
```diff
@@ -0,0 +1,368 @@
...
+        df = self.embedder.embed_dataframe(
+            df, column_mapping={source_column: column_mapping}
+        )
+
+        if (
+            column_mapping[0] == "text"
+            or current_logical_layer_fn is not default_logical_layer_fn
+        ):
+            df = current_logical_layer_fn(df)
+        elif (
```
ntkathole: This condition seems never pass, and pragmatically elif ValueError never be raised

ntkathole (approved):
lgtm

sdk/python/feast/doc_embedder.py
```diff
@@ -0,0 +1,380 @@
...
+
+from feast.chunker import BaseChunker, TextChunker
+from feast.embedder import BaseEmbedder, MultiModalEmbedder
+
+if TYPE_CHECKING:
+    from feast.feature_store import FeatureStore
+
+
+@runtime_checkable
+class LogicalLayerFn(Protocol):
```
franciscojavierarceo, line 17: Probably we should rename this as it's not particularly intuitive IMO

patelchaitany: So i have two name in the mind - SchemaMapperFn or the FeatureViewMapperFn i think this two name are more intuitive if you have particular name in mind then let me know.

PR #5979, opened 2026-02-17
fix: added ray in the minimal in the pyproject.toml

patelchaitany:

This PR adds the ray in the minimal in the pyproject.toml

Add Ray dependencies to Feast Operator and feature server images to support Ray Offline Store and Ray Compute Engine

Feast now supports Ray but upon configuring services with Ray getting runtime error 

ModuleNotFoundError: No module named 'ray'
FeastModuleImportError: Could not import module 'feast.infra.offline_stores.contrib.ray_offline_store.ray'

# Misc

---

<a href="https://app.devin.ai/review/feast-dev/feast/pull/5979" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


ntkathole:
@patelchaitany Please run make lock-python-dependencies-all and commit the updated requirements files

patelchaitany:
@franciscojavierarceo, the issue is that feast-operator uses the minimal image, so using the Ray offline store requires us to maintain a custom Docker image. I believe adding Ray to the minimal build would solve this and improve the developer experience. Do you think that is a valid approach?

ntkathole:
> I don't think Ray should be in minimal

@franciscojavierarceo We will need these compute engines in some image so that users do not need to build their custom images. We can either include in ray/spark in minimal or have new feast[server] group to build image, but having separate group means having 2 different images as well. Some slack context - https://feastopensource.slack.com/archives/C01M2GYP0UC/p1768568651039479

franciscojavierarceo:
@ntkathole i think the custom image path is the right option or we deploy images for each support compute engine (e.g., Ray, Spark, etc.).

PR #5999, opened 2026-02-21
fix: Check duplicate names for feature view across types

Prathap-P:

- This PR enforces cross-type feature view name uniqueness to fix a bug (feast-dev/feast#5995).
- Previously, `get_online_features` resolved feature view names via `get_any_feature_view`, which checks registry tables in a fixed order: FeatureView → StreamFeatureView → OnDemandFeatureView. 
- If a project registered both a FeatureView and StreamFeatureView with the same name, every get_online_features call would silently return the FeatureView, even when the StreamFeatureView contained fresher data

Fixes https://github.com/feast-dev/feast/issues/5995

  # Misc

---

<a href="https://app.devin.ai/review/feast-dev/feast/pull/5999" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/sdk/python/tests/unit/cli/test_cli_apply_duplicates.py
+++ b/sdk/python/tests/unit/cli/test_cli_apply_duplicates.py
@@ -152,9 +152,7 @@ def test_cli_apply_imported_featureview_with_duplication() -> None:
         rc, output = runner.run_with_output(["apply"], cwd=repo_path)
 
         assert rc != 0
-        assert (
-            b"Multiple FeatureViews with name 'driver_hourly_stats' found." in output
-        )
+        assert b"Multiple FeatureViews with name 'driver_hourly_stats' found." in output
 
 
 def test_cli_apply_duplicated_featureview_names_multiple_py_files() -> None:
@@ -195,6 +193,5 @@ def test_cli_apply_duplicated_featureview_names_multiple_py_files() -> None:
 
         assert (
             rc != 0
-            and b"Feature view names must be case-insensitively unique"
-            in output
+            and b"Feature view names must be case-insensitively unique" in output
         )


sdk/python/feast/infra/registry/base_registry.py
```diff
@@ -263,6 +267,61 @@ def apply_feature_view(
...
+        def _check_conflict(getter, not_found_exc, existing_type: str):
+            try:
+                getter(name, project, allow_cache=allow_cache)
+                raise ConflictingFeatureViewNames(name, existing_type, new_type)
+            except not_found_exc:
+                pass
+
+        # Check StreamFeatureView before FeatureView since StreamFeatureView is a subclass
+        # Note: All getters raise FeatureViewNotFoundException (not type-specific exceptions)
+        if isinstance(feature_view, StreamFeatureView):
```
haoxu0, line 295: Does this mean it will call to retrieve feature view many times? Can we optimize it?

Prathap-P: If I understand correctly, you're concerned that when creating 100 feature views, the `apply_feature_view` function calls the getters for the other two types each time. That results in 2 database calls per feature view — 200 total — and you’re wondering if we should optimize this part.

Prathap-P: @HaoXuAI 
I have a plan. We need to validate in two places:

1. In the incoming request — ensure the new feature view name doesn’t exist in the other two types within the request (and vice versa).
2. In the database — check against existing records (for the SQL registry case).

Currently, the cache util doesn’t have an `exists` function to check if a name is already in the registry. Even if it did, calling it 100 times wouldn’t be efficient.

So I’m planning to create a util function per feature view type that takes a list of names and a project, and checks that none of the names exist in the DB.

For example, if 50 feature views come in:

* I’ll call the new util once with those 50 names against the stream and on-demand tables.
* Do the same vice versa for the other types.

This way, instead of calling per feature view, we make only two DB calls per type.

That’s my current idea. Do you see any better optimization?

Prathap-P: @HaoXuAI I can go ahead to implement, if you find we are good with above logic 😄

haoxu0: i think that makes sense. Feel free to create a new issue for this. We can go with the current solution for now

sdk/python/feast/feature_store.py
```diff
@@ -3255,18 +3256,25 @@ def _print_materialization_log(
...
+    """
+    fv_by_name: Dict[str, BaseFeatureView] = {}
     for fv in feature_views:
         case_insensitive_fv_name = fv.name.lower()
-        if case_insensitive_fv_name in fv_names:
-            raise ValueError(
-                f"More than one feature view with name {case_insensitive_fv_name} found. "
-                f"Please ensure that all feature view names are case-insensitively unique. "
-                f"It may be necessary to ignore certain files in your feature repository by using a .feastignore file."
+        if case_insensitive_fv_name in fv_by_name:
```
ntkathole, line 3269: `_validate_feature_views()` in feature_store.py uses case-insensitive names (fv.name.lower()), while
`_ensure_feature_view_name_is_unique()` relies on the registry getters, which are typically case-sensitive. Worth aligning the behavior.

sdk/python/feast/infra/registry/base_registry.py
```diff
@@ -263,6 +266,32 @@ def apply_feature_view(
...
+        new_type = type(feature_view).__name__
+        for existing_feature_view in self.list_all_feature_views(
+            project, allow_cache=allow_cache
+        ):
+            if existing_feature_view.name.lower() != case_insensitive_name:
+                continue
+
+            existing_type = type(existing_feature_view).__name__
+            if existing_type != new_type:
+                raise ConflictingFeatureViewNames(name, existing_type, new_type)
```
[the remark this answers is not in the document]
Prathap-P: let me update in sometime

sdk/python/feast/feature_store.py
Prathap-P: @ntkathole  Thanks for your comment.
On trying to work on that, I have a found an inconsistency (might be a potential issue).
I'm holding my progress on above issue, till I get clarified on this.

In current Feast master, FeatureStore.apply() behaves as an incremental upsert (not a full registry refresh): it validates only the objects in the current request and writes those objects, while existing unrelated records remain untouched. Because of this, duplicate-name protection is incomplete for previously registered objects—especially when names differ only by case (for example, existing driver and new Driver). The in-request validation checks only the incoming payload, and the SQL registry uniqueness constraints are per table and case-sensitive by default, so this scenario can pass and create ambiguous registry state. Expected behavior (per my understanding) is case-insensitive uniqueness for feature view names, so this appears to be a gap.

ntkathole: @Prathap-P I think it's fine if you want to handle it separately and we keep scope of this PR for sql registry duplicate names

Prathap-P: @ntkathole Solving the test cases took a lot of time, finally was able to fix it. Do you have any other points to discuss?

Prathap-P: I know Devin is reporting these changes are not implemented in snowflake, but as you said to scope this pr to sql, ignoring it...

PR #6011, opened 2026-02-24
feat: Making transformer optional for ODFV

nquinn408:
---

<a href="https://app.devin.ai/review/feast-dev/feast/pull/6011" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


docs/reference/beta-on-demand-feature-view.md
```diff
@@ -69,6 +76,37 @@ def driver_aggregated_stats(inputs):
...
+
+# Define an ODFV for pre-transformed features without a transformation function
+@on_demand_feature_view(
+    sources=[driver_hourly_stats_view],
+    schema=[
+        Field(name="conv_rate_adjusted", dtype=Float64),
+    ],
+    write_to_online_store=True,
+)
+def pre_transformed_conv_rate(inputs):
```
franciscojavierarceo: this usage probably feels less intuitive then:

```python
pre_transformed_conv_rate = OnDemandFeatureView(
    name="pre_transformed_conv_rate",
    sources=[driver_hourly_stats_view],
    schema=[
        Field(name="conv_rate_adjusted", dtype=Float64),
    ],
    write_to_online_store=True,
)
```

This example is confusing to them though as I don't see a calculation used during ingestion so why is this not just a regular feature view?

nquinn408: I completely agree. I will update.

haoxu0:
Hey @franciscojavierarceo mind adding @nquinn408 into the team after this PR? He's gonna contribute more to the project and want to reduce some frictions :)

docs/reference/beta-on-demand-feature-view.md
```diff
@@ -39,6 +39,13 @@ Native Python mode supports transformations on singleton dictionaries by setting
 write transformation functions that operate on a single row at a time, making the code more intuitive and aligning with
 how data scientists typically think about data transformations.
 
+## Optional Transformations
+
+Transformation functions in On Demand Feature Views are optional. You can define an ODFV without a transformation function in the following scenarios:
+
+1. **When using aggregations**: Aggregations compute aggregate statistics and don't require a separate transformation function.
+2. **When materializing pre-transformed data**: If your data is already transformed by an external process, you can define an ODFV to serve those features without additional transformation logic.
```
franciscojavierarceo, line 47: @nquinn408 can you explain a use case where you would do this example? 

ODFVs are either transforms on reads or on writes so I'm a little confused on why you'd omit a transformation for an ODFV. For that use case, why wouldn't you just use a regular Feature View?

Let me know if I misunderstood something.

franciscojavierarceo:
> Hey @franciscojavierarceo mind adding @nquinn408 into the team after this PR? He's gonna contribute more to the project and want to reduce some frictions :)

yeah definitely happy to do so (sorry I'm blocking the PR here, want to get an understanding about things).

docs/reference/beta-on-demand-feature-view.md
nquinn408: @franciscojavierarceo, of course.  To understand our feature store usage, we have a number of unpublished features that only exist online.  When these features are served, some services write to them and others read them.  They are synced offline through CDC pipeline. They do not fit the push source model because our online feature store does not exist offline.  

I don't think this is such a usual setup. These on demand features do not have an offline source so they are not FeatureViews which require a source. They do not fit the transform on read/write model because the value may be only written by one service and only read by others.

nquinn408: Another concern would be if we are using a Java based or Go feature server which means that having to rely on python based transformation service for every on demand feature view read request would be extremely expensive and highly inefficient.

haoxu0: I think after refactoring OnDemandFeatureView to inherit from FeatureView this will natively wokrs. For now it is a adhoc changes.

franciscojavierarceo:
fyi @nquinn408 I've sent you an invite to the team

docs/reference/beta-on-demand-feature-view.md
franciscojavierarceo: >To understand our feature store usage, we have a number of unpublished features that only exist online. 

Oh interesting! Does this mean you are just logging features in production before testing?

>When these features are served, some services write to them and others read them. 

Are you referring to features in general or ODFVs?


>They do not fit the push source model because our online feature store does not exist offline.

It is not a requirement that you use the same feature view offline and online, so I'm confused a bit.


>These on demand features do not have an offline source so they are not FeatureViews which require a source. 

So to me, this is the part that stands out as something good for us to enhance by supporting ODFVs without a source, no?  Would that solve your problem?

franciscojavierarceo: >I think after refactoring OnDemandFeatureView to inherit from FeatureView this will natively wokrs. For now it is a adhoc changes.

Yeah I can't quite make that comment confidently as I'm not sure I understand the original use case.

nquinn408: > Oh interesting! Does this mean you are just logging features in production before testing?

They are simulated offline for training, but the freshness of the feature is important so yes. 

> Are you referring to features in general or ODFVs?

ODFVs

> It is not a requirement that you use the same feature view offline and online, so I'm confused a bit.

You were asking why we didn't just use FeatureViews and since FeatureViews all have a source, this is why we don't use FVs with push source. 

> So to me, this is the part that stands out as something good for us to enhance by supporting ODFVs without a source, no? Would that solve your problem?

Yes but I would prefer to have options in expressing complex use cases so I wonder why we cannot have both.

nquinn408:
@franciscojavierarceo , would you prefer I close this and submit a PR for `FeatureView` with optional `source`?

franciscojavierarceo:
> @franciscojavierarceo , would you prefer I close this and submit a PR for `FeatureView` with optional `source`?

@nquinn408 would you mind taking that approach? Apologies for being slow to respond on this I was traveling this week for work. 

I think ODFVs are fundamentally transform centric and after reviewing our thread again I worry making transformations optional for On Demand Feature Views will cause intense confusion for users. Especially because we address this in our docs explicitly https://docs.feast.dev/getting-started/architecture/write-patterns#id-2.-computing-transformations-on-demand

nquinn408:
@franciscojavierarceo , my only concern about pursuing this is that when you don't have a source and your entity comes from a request. Can your source for a non ODFV be a request source?

nquinn408:
Going to close this PR and issue and switch to working on FeatureView with optional source.

PR #6017, opened 2026-02-24
feat: Add Oracle DB as Offline store in python sdk & operator

aniketpalu:

## Oracle DB Offline Store for Feast (ibis-based)
This PR adds a native Oracle Database offline store for Feast, built on the ibis-framework's Oracle backend (ibis.oracle). It follows the same proven pattern used by the existing MSSQL ibis offline store — delegating to the shared ibis.py functions for point-in-time joins, deduplication, and feature retrieval.
The implementation is ~300 lines total (vs 1096 lines in the SQLAlchemy-based approach), with no Jinja2 templates, no dialect-specific SQL workarounds, and no manual query generation. ibis generates Oracle SQL automatically and provides native Arrow transfer via to_pyarrow().
### Oracle Column Name Handling
Oracle stores unquoted identifiers in UPPERCASE. For example, CREATE TABLE t (user_id INT) stores the column as USER_ID. This store uses a passthrough approach — column names are returned exactly as Oracle stores them, with no automatic case transformation.
This means:
Unquoted columns (the standard Oracle convention): The user references them in UPPERCASE, matching what they see in Oracle tools like SQL Developer, DBeaver, or DESCRIBE table.
Quoted columns (e.g., "CamelCase"): The user references them with the exact casing used when creating the table.
This design ensures predictable behavior — what you see in Oracle is what you use in Feast. No hidden transformations, no conventions to learn.

### How to Use

`feature_store.yaml`
```
project: my_project
provider: local
registry: data/registry.db

offline_store:
  type: oracle
  user: feast
  password: feast
  host: localhost
  port: 1521
  service_name: FREEPDB1

online_store:
  type: sqlite
  path: data/online_store.db
```

### What's Tested
Tested against a live Oracle 23ai database (Oracle Free 23.26.1.0.0):

#### Core Feast Operations:

- Oracle connectivity via ibis.oracle.connect() with schema introspection
- feast apply — registering entities, feature views, and feature services
- get_historical_features — point-in-time correct feature retrieval with entity DataFrame
- materialize — pulling latest features from Oracle offline store into SQLite online store
- get_online_features — retrieving materialized features from the online store
- pull_all_from_table_or_query — direct offline store batch retrieval with date range filtering
- to_pyarrow() — native Arrow transfer via ibis (no intermediate pandas conversion)

#### Non-Entity Retrieval (entity_df=None):

- With both start_date and end_date provided
- With only end_date (start calculated from feature view TTL)
- With only start_date (end defaults to current time)
- With neither (end defaults to current time, start calculated from TTL)

#### Oracle Column Name Handling:

- Unquoted identifiers (UPPERCASE) — e.g., USER_ID, FEATURE_TIME, TXN_LAST_24H
- Quoted mixed-case identifiers — e.g., "User_ID" and "user_iD" in the same test run
- Full Feast flow (apply → historical → materialize → online) verified for both casing styles

https://github.com/feast-dev/feast/issues/6018

# Misc

---

<a href="https://app.devin.ai/review/feast-dev/feast/pull/6017" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


Srihari1192:
@aniketpalu Tested oracle DB configuration with feast operator and feast SDK  , LGTM
Thanks

sdk/python/feast/infra/offline_stores/ibis.py
```diff
@@ -413,7 +413,7 @@ def point_in_time_join(
             else:
                 alias = "".join(random.choices(string.ascii_uppercase, k=10))
 
-                feature_table = feature_table.alias(alias=alias).sql(
+                feature_table = feature_table.alias(alias).sql(
```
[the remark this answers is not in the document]
aniketpalu: In ibis 12.0.0, Table.alias() declares alias as positional-only (signature: (self, alias: str, /)). Passing it as a keyword argument raises TypeError: Table.alias() got some positional-only arguments passed as keyword arguments: 'alias'. This is a runtime crash fix.

sdk/python/feast/infra/offline_stores/contrib/oracle_offline_store/oracle.py
```diff
@@ -0,0 +1,309 @@
...
+                    (table[ts_col] >= ibis.literal(start_date))
+                    & (table[ts_col] <= ibis.literal(end_date))
+                ).select(cols)
+                sub = sub.rename({"event_timestamp": ts_col})
+                entity_dfs.append(sub.execute())
+
+            entity_df = pd.concat(entity_dfs, ignore_index=True).drop_duplicates()
+
+        # If entity_df is a SQL string, execute it to get a DataFrame
+        if type(entity_df) == str:
```
tokoko: have you tried passing this directly w/o converting to pandas dataframe here?

aniketpalu: Thanks for the review @tokoko!

If the concern is the overhead of converting to pandas here — agreed, the current path goes Oracle ibis table → pandas → ibis.memtable(pandas_df) which makes pandas an unnecessary middleman. However, eliminating the pandas conversion entirely would require changes to the shared get_historical_features_ibis() in ibis.py, since it and its helpers (_get_entity_schema, _to_utc, _get_entity_df_event_timestamp_range) all operate on pd.DataFrame. That would also impact the MSSQL offline store which follows the same pattern, so it might be better suited as a separate issue/PR.

tokoko: got it, same applies to "no entity_df" code path, I think. both would be better integrated into shared implementation, but that can wait.

sdk/python/feast/infra/offline_stores/contrib/oracle_offline_store/oracle.py
```diff
@@ -0,0 +1,309 @@
...
+    return _write_data_source
+
+
+class OracleOfflineStoreConfig(FeastConfigBaseModel):
+    """Offline store config for Oracle Database"""
+
+    type: Literal["oracle"] = "oracle"
+    """Offline store type selector"""
+
+    user: StrictStr = "system"
```
ntkathole: Default credentials shouldn't be shipped in a config class. We can make user and password required (no defaults), or at minimum use empty string defaults.

sdk/python/feast/infra/offline_stores/contrib/oracle_offline_store/oracle.py
ntkathole:
```suggestion
        if isinstance(entity_df, str):
```

sdk/python/feast/infra/offline_stores/contrib/oracle_offline_store/oracle.py
```diff
@@ -0,0 +1,309 @@
...
+
+def _build_data_source_reader_for_retrieval(config: RepoConfig):
+    """Build a reader that materializes Oracle data into an in-memory table.
+
+    Used by ``get_historical_features`` which joins feature tables with an
+    in-memory entity table (``ibis.memtable``).  Both sides must be on the
+    same backend for computed columns like ``entity_row_id`` to survive the
+    join — converting to memtable ensures this.
+    """
+    con = get_ibis_connection(config)
```
ntkathole: I see we are creating this connection multiple times, can't we simply pass con object so that there isn't separate Oracle connections for a single `get_historical_features` call

sdk/python/feast/infra/offline_stores/contrib/oracle_offline_store/oracle.py
```diff
@@ -0,0 +1,309 @@
...
+
+    def _read_data_source(data_source: DataSource, repo_path: str = "") -> Table:
+        table = _read_oracle_table(con, data_source)
+        return ibis.memtable(table.execute())
```
ntkathole: reading entire table in memory? Isn't there a better way to read filtered table based on timestamps may be?

sdk/python/feast/infra/offline_stores/contrib/oracle_offline_store/oracle.py
```diff
@@ -0,0 +1,309 @@
...
+    """Oracle database password"""
+
+    host: StrictStr = "localhost"
+    """Oracle database host"""
+
+    port: StrictInt = 1521
+    """Oracle database port"""
+
+    service_name: Optional[StrictStr] = None
+    """Oracle service name (mutually exclusive with sid and dsn)"""
```
ntkathole, line 132: probably we can add validation for mutually exclusive fields

ntkathole:
Can we add documentation as well for offline store ?

sdk/python/feast/__init__.py
```diff
@@ -5,6 +5,9 @@
 from feast.infra.offline_stores.contrib.athena_offline_store.athena_source import (
     AthenaSource,
 )
+from feast.infra.offline_stores.contrib.oracle_offline_store.oracle_source import (
+    OracleSource,
+)
```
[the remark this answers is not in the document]
aniketpalu: Valid concern, committed the fix

sdk/python/feast/infra/offline_stores/contrib/oracle_offline_store/oracle.py
tokoko: pretty sure `_build_data_source_reader_for_retrieval` function is completely redundant, it just repackages `_build_data_source_reader` for no good reason.

aniketpalu: @ntkathole @tokoko 

Added pre-filter to avoid reading the whole table. Thanks for the catch

sdk/python/feast/infra/offline_stores/contrib/oracle_offline_store/oracle.py
aniketpalu: Done

sdk/python/feast/infra/offline_stores/contrib/oracle_offline_store/oracle.py
aniketpalu: Makes sense. Done this.

sdk/python/feast/infra/offline_stores/contrib/oracle_offline_store/oracle.py
aniketpalu: Removed default values for credentials.

ntkathole:
@aniketpalu Need rebase

sdk/python/feast/infra/offline_stores/contrib/oracle_offline_store/oracle.py
tokoko: to be clear, I don't understand why we need `_build_data_source_reader_for_retrieval` at all. can't you use `_build_data_source_reader` in `get_historical_features_ibis` call? that way there will be no materialization

sdk/python/feast/infra/offline_stores/contrib/oracle_offline_store/oracle.py
```diff
@@ -0,0 +1,376 @@
...
+            config=config,
+            data_source=data_source,
+            join_key_columns=join_key_columns,
+            feature_name_columns=feature_name_columns,
+            timestamp_field=timestamp_field,
+            created_timestamp_column=created_timestamp_column,
+            start_date=start_date,
+            end_date=end_date,
+            data_source_reader=_build_data_source_reader(config),
+            data_source_writer=_build_data_source_writer(config),
```
ntkathole:
```suggestion
            data_source_writer=_build_data_source_writer(config, con=con),
```

Define  con = get_ibis_connection(config)

sdk/python/feast/infra/offline_stores/contrib/oracle_offline_store/oracle.py
aniketpalu: @tokoko 

Thanks for clarification, I misunderstood your earlier comment. You are right, materialization can be avoided by using `_build_data_source_reader`. 

After this fix, needed to do a little change in building `entity_row_id` in ibis as `r = ibis.literal("")` in `_generate_row_id()` creates broken `entity_row_id` due to Oracle DB casting it as NULL.

PR #6024, opened 2026-02-25
perf: Parallelize DynamoDB batch reads in sync online_read

abhijeet-dhumal:
## Summary
Execute DynamoDB BatchGetItem requests in parallel using ThreadPoolExecutor instead of sequentially. This significantly reduces latency when reading features for many entities that span multiple batches.

## Changes
- Pre-split entity IDs into batches upfront
- Use `ThreadPoolExecutor` to execute batch requests concurrently
- Skip parallelization for single batch (no overhead)
- Merge results in original order after parallel fetch

## Expected Behavior
For multiple batches, DynamoDB BatchGetItem requests should execute in parallel, reducing total latency from `N × network_latency` to approximately `1 × network_latency`.

## Current Behavior
The sync `online_read` method executes batch requests sequentially in a while loop:

```python
while True:
    batch = list(itertools.islice(entity_ids_iter, batch_size))
    if len(batch) == 0:
        break
    response = dynamodb_resource.batch_get_item(RequestItems=...)  # Sequential!
    result.extend(batch_result)
```

For 500 entities with `batch_size=100`, this makes 5 sequential network calls.

## Steps to Reproduce
1. Configure DynamoDB online store with `batch_size=100`
2. Call `get_online_features` for 500 entities
3. Profile network latency - observe 5 sequential calls

## Specifications
- **Version:** 0.47.0+
- **Platform:** All
- **Subsystem:** `sdk/python/feast/infra/online_stores/dynamodb.py`

## Performance Impact
For 500 entities with batch_size=100 (5 batches):
- **Before:** 5 sequential network calls = 50-150ms
- **After:** 5 parallel network calls = 10-30ms
- **Estimated savings:** 40-120ms for large entity sets

## Possible Solution
Already implemented in this PR using `ThreadPoolExecutor`:

```python
with ThreadPoolExecutor(max_workers=min(len(batches), batch_size)) as executor:
    responses = list(executor.map(fetch_batch, batches))
```

## Related
- RHOAIENG-46061 (60ms p99 SLA target for online feature serving)
- Note: The async path (`online_read_async`) already uses `asyncio.gather()` for parallel execution

---

<a href="https://app.devin.ai/review/feast-dev/feast/pull/6024" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/docs/reference/online-stores/dynamodb.md
+++ b/docs/reference/online-stores/dynamodb.md
@@ -46,7 +46,7 @@ online_store:
 | ------ | ---- | ------- | ----------- |
 | `region` | string | | AWS region for DynamoDB |
 | `table_name_template` | string | `{project}.{table_name}` | Template for table names |
-| `batch_size` | int | `40` | Number of items per BatchGetItem/BatchWriteItem request (max 100) |
+| `batch_size` | int | `100` | Number of items per BatchGetItem/BatchWriteItem request (max 100) |
 | `max_read_workers` | int | `10` | Maximum parallel threads for batch read operations. Higher values improve throughput for large batch reads but increase resource usage |
 | `consistent_reads` | bool | `false` | Whether to use strongly consistent reads (higher latency, guaranteed latest data) |
 | `tags` | dict | `null` | AWS resource tags added to each table |


sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -479,33 +480,57 @@ def online_read(
             online_config.endpoint_url,
             online_config.session_based_auth,
         )
-        table_instance = dynamodb_resource.Table(
-            _get_table_name(online_config, config, table)
-        )
+        table_name = _get_table_name(online_config, config, table)
+        table_instance = dynamodb_resource.Table(table_name)
```
ntkathole: table_instance is only used in the single-batch path, and seems unnecessary since table_name can be used instead of table_instance.name

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -479,33 +480,57 @@ def online_read(
...
-                response,
-                batch,
+            batch_entity_ids = self._to_resource_batch_get_payload(
+                online_config, table_name, batch
             )
+            return thread_resource.batch_get_item(RequestItems=batch_entity_ids)
+
+        # Use ThreadPoolExecutor for parallel I/O
+        # Cap at 10 workers to avoid excessive thread creation
+        max_workers = min(len(batches), 10)
```
ntkathole: I think this should be configurable (`max_read_workers`) in online store configs

sdk/python/feast/infra/online_stores/dynamodb.py
```diff
@@ -479,33 +480,57 @@ def online_read(
...
             )
-            response = dynamodb_resource.batch_get_item(
-                RequestItems=batch_entity_ids,
+            response = dynamodb_resource.batch_get_item(RequestItems=batch_entity_ids)
+            return self._process_batch_get_response(table_name, response, batches[0])
+
+        # Execute batch requests in parallel for multiple batches
+        # Note: boto3 resources are NOT thread-safe, so we create a new resource per thread
+        def fetch_batch(batch: List[str]) -> Dict[str, Any]:
+            thread_resource = _initialize_dynamodb_resource(
```
ntkathole: I think creating here we are creating new session on each thread. Instead we can share a single `_dynamodb_client` across all threads.

```
def fetch_batch(batch: List[str]) -> Dict[str, Any]:
    batch_entity_ids = self._to_client_batch_get_payload(
        online_config, table_name, batch
    )
    return dynamodb_client.batch_get_item(RequestItems=batch_entity_ids)
```

https://docs.aws.amazon.com/boto3/latest/guide/clients.html#multithreading-or-multiprocessing-with-clients

sdk/python/feast/infra/online_stores/dynamodb.py
abhijeet-dhumal: Good catch! Removed table_instance entirely - now using table_name directly in the single-batch path.

sdk/python/feast/infra/online_stores/dynamodb.py
abhijeet-dhumal: Done! Added max_read_workers config option to DynamoDBOnlineStoreConfig with default value of 10. Users can now tune parallelism based on their workload.

sdk/python/feast/infra/online_stores/dynamodb.py
abhijeet-dhumal: Great suggestion! Switched to using a shared _dynamodb_client across all threads since boto3 clients are thread-safe. This avoids the overhead of creating new sessions per thread. Also added the TypeDeserializer for parsing client responses (same approach as the async path).

sdk/python/feast/infra/online_stores/dynamodb.py
ntkathole: can we also add documentation about this config in dynamodb online store docs?

docs/reference/online-stores/dynamodb.md
```diff
@@ -22,6 +22,47 @@ online_store:
...
+```
+{% endcode %}
+
+### Configuration Options
+
+| Option | Type | Default | Description |
+| ------ | ---- | ------- | ----------- |
+| `region` | string | | AWS region for DynamoDB |
+| `table_name_template` | string | `{project}.{table_name}` | Template for table names |
+| `batch_size` | int | `40` | Number of items per BatchGetItem/BatchWriteItem request (max 100) |
```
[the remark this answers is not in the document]
ntkathole: @abhijeet-dhumal ^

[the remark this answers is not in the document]
abhijeet-dhumal: Fixed ✅

PR #6039, opened 2026-03-01
fix: Correct inverted default_value in _construct_random_input

### LINKED ISSUE #6008: Fix inverted default_value logic in OnDemandFeatureView._construct_random_input
In _construct_random_input(), the default_value fallback logic is inverted:

``` 
# current (wrong)
default_value = None if not singleton else [None]
```
When `singleton=True`, the fallback should be scalar None, but it gets [None] (list). When `singleton=False`, it should be [None], but it gets None. This can break UDF schema inference for feature types not present in sample_values.

Found during review of #5951.

Br1an67:

In `OnDemandFeatureView._construct_random_input`, the `default_value` assignment for missing source columns was inverted. When `singleton=True` (single row), the fallback should be a scalar `None`; when `singleton=False` (batch), it should be `[None]`. The existing code had these swapped, causing downstream errors when the random input was consumed.

Fixes #6008

# Misc

One-line fix in `sdk/python/feast/on_demand_feature_view.py`.

---

<a href="https://app.devin.ai/review/feast-dev/feast/pull/6039" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/sdk/python/feast/on_demand_feature_view.py
+++ b/sdk/python/feast/on_demand_feature_view.py
@@ -1011,7 +1011,7 @@ class OnDemandFeatureView(BaseFeatureView):
             sample_values = {k: v[0] for k, v in sample_values.items()}
 
         # Default value for missing types
-        default_value = None if not singleton else [None]
+        default_value = [None] if not singleton else None
 
         feature_dict = {}
 


haoxu0:
LGTM. please fix the DCO. also might be better to add a test

ntkathole:
@Br1an67 Will you please add a test ?

PR #6041, opened 2026-03-01
chore: Run duckdb offline store tests seperately

tokoko:
- Runs duckdb offline store tests in a separate workflow, full integration suite is really hard to run. we can pull other tests into backend-specific workflows as well to simplify local runs/debugging.
- relies on on pixi env defined inside `pyproject.toml` rather than our custom patched-together lockfiles.

---

<a href="https://app.devin.ai/review/feast-dev/feast/pull/6041" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -282,7 +282,17 @@ python = "~=3.10.0"
 feast = { path = ".", editable = true, extras = ["duckdb", "delta", "grpcio", "test"] }
 
 [tool.pixi.feature.duckdb-tests.tasks]
-test = { cmd = "python -m pytest -n 8 --integration --ignore=sdk/python/tests/integration/offline_store/test_dqm_validation.py sdk/python/tests/integration/offline_store", env = { PYTHONPATH = ".", FULL_REPO_CONFIGS_MODULE = "sdk.python.tests.integration.feature_repos.duckdb_repo_configuration", FEAST_IS_LOCAL_TEST = "True" } }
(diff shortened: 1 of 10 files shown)


sdk/python/tests/integration/feature_repos/repo_configuration.py
```diff
@@ -72,21 +58,6 @@
     create_order_feature_view,
     create_pushable_feature_view,
 )
-from tests.integration.feature_repos.universal.online_store.bigtable import (
-    BigtableOnlineStoreCreator,
```
franciscojavierarceo, line 76: Why remove these?

tokoko: moved them down to lazily import only when actually used. this way I would have to pull in extra dependencies just for pytest collection stage.

pyproject.toml
```diff
@@ -267,3 +269,30 @@ exclude = [
 dev = [
     "pytest-xdist>=3.8.0",
 ]
+
+# Pixi configuration
```
ntkathole, line 273: Is there a reason to use pixi ? I think we can use uv similar to other workflows for consistency ?

ntkathole: I see the description about lock files, may be we can add uv.lock file instead of pixi env

tokoko: the problem is you can't have multiple "solve groups" in uv, that's why we are running multiple uv compile commands for each type of environment right now. pixi acts as a thin uv wrapper in this case, allowing you to have independent solve groups for different python versions.

PR #6043, opened 2026-03-02
chore: Run registry tests separately

tokoko:
# What this PR does / why we need it:
- Runs registry tests in a separate workflow
- both integration and integration-local now ignore registration folder
- other misc changes to concentrate all tests with universal_offline_stores marker under offline_store and with universal_online_stores marker under online_store.

---

<a href="https://app.devin.ai/review/feast-dev/feast/pull/6043" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

PR #6045, opened 2026-03-03
fix: Ray offline store tests are duplicated across 3 workflows

ntkathole:

Ray offline store tests were running 3 times across different workflows, and Ray compute engine tests were running in the main CI instead of the dedicated Ray workflow.

- Removed the try/except block that conditionally appended RayDataSourceCreator to AVAILABLE_OFFLINE_STORES. This was the root cause - Ray was being auto-included in every workflow where it was installable.
- Changed online store from RedisOnlineStoreCreator to {"type": "sqlite"}. The compute engine tests focus on Ray compute, not online store behavior, so sqlite is sufficient and avoids needing redis in the pixi environment.
- Split the single test task into test-offline and test-compute, then combined them with depends-on. The Ray workflow now runs both sdk/python/tests/integration/offline_store (with Ray config) and sdk/python/tests/integration/compute_engines/ray_compute.
- Added `--ignore=sdk/python/tests/integration/compute_engines/ray_compute` to test-python-integration (main CI target).
- Replaced test-python-universal-ray-offline and test-python-ray-compute-engine with a single `test-python-ray-integration` target.

---

<a href="https://app.devin.ai/review/feast-dev/feast/pull/6045" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/.github/workflows/pr_ray_integration_tests.yml
+++ b/.github/workflows/pr_ray_integration_tests.yml
@@ -8,7 +8,7 @@ on:
       - labeled
 
 jobs:
-  integration-test-ray-offline:
+  integration-test-ray:
     if:
       ((github.event.action == 'labeled' && (github.event.label.name == 'approved' || github.event.label.name == 'lgtm' || github.event.label.name == 'ok-to-test')) ||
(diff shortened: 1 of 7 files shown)


tokoko:
#6043 partly fixes this as well

ntkathole:
> #6043 partly fixes this as well

let's merge yours first and then will rebase

tokoko (approved):
lgtm. you can ignore registry test fail, not sure how that slipped through. let's take care of other issues separately.

PR #6067, opened 2026-03-05
fix: Fix non-specific label selector on metrics service

ntkathole:

The `feast-operator-controller-manager-metrics-service` uses a non-specific label selector (`control-plane: controller-manager`) that matches pods from any operator using the standard kubebuilder scaffolding label. In shared-namespace environments, where multiple operators are deployed in the same namespace, this causes the Service to incorrectly select pods belonging to other operators.

Added `app.kubernetes.io/name: feast-operator` to all selectors and pod template labels so that the metrics Service, Deployment, and ServiceMonitor uniquely identify Feast operator pods.

---

<a href="https://app.devin.ai/review/feast-dev/feast/pull/6067" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/infra/feast-operator/config/manager/manager.yaml
+++ b/infra/feast-operator/config/manager/manager.yaml
@@ -19,6 +19,7 @@ metadata:
 spec:
   selector:
     matchLabels:
+      app.kubernetes.io/name: feast-operator
       control-plane: controller-manager
   replicas: 1
   template:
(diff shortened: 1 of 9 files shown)


infra/feast-operator/config/manager/manager.yaml
astefanutti, line 23: That's also a problem here. We need to be careful for the upgrade path.

ntkathole: yes, it's not added there due to upgrade failure. It's immutable. But it doesn't cause the bug.

ntkathole: Added, and documented the upgrade path

Issue #6088, opened 2026-03-10
Extract groups and namespaces claims from JWT token


aniketpalu:
**Is your feature request related to a problem? Please describe.**

`OidcTokenParser` only extracts `username` and `roles` from the JWT. The `groups` and `namespaces` claims are ignored, so `GroupBasedPolicy` and `NamespaceBasedPolicy` never match for OIDC-authenticated users.

**Describe the solution you'd like**

Extract `groups` and `namespaces` claims from the decoded JWT in `OidcTokenParser.user_details_from_access_token()` and pass them to the `User` object. Default to empty list if the claims are absent.

**Describe alternatives you've considered**

None — the `User` model, policy classes, and enforcer already support groups and namespaces. Only the OIDC parser is missing the extraction.

**Additional context**

Affected file: `sdk/python/feast/permissions/auth/oidc_token_parser.py`

PR #6089, opened 2026-03-10
feat: Extended OIDC support to extract groups & namespaces and token injection with multiple methods

aniketpalu:

Extends OIDC authentication in Feast to support `GroupBasedPolicy` and `NamespaceBasedPolicy` for OIDC users, adds flexible client token injection, and wires up operator support for OIDC deployments.

Previously, the OIDC token parser only extracted `preferred_username` and `resource_access` roles from the JWT. `GroupBasedPolicy`, `NamespaceBasedPolicy`, and `CombinedGroupNamespacePolicy` could never grant access for OIDC users because the `User` object was always created with empty groups and namespaces.

## Server Side Changes

**Files:** `oidc_token_parser.py`, `oidc_service.py`, `utils.py`

When `auth.type: oidc`, the Feast server now handles two types of incoming tokens:

* **Keycloak JWTs** (human users via UI, Swagger, or notebooks) are validated against the OIDC provider's JWKS. The parser extracts `preferred_username` (with `upn` fallback for Azure AD / Entra ID), `groups` claim, and `resource_access.<client_id>.roles`. `GroupBasedPolicy` uses the groups. `client_id` is optional; when absent, roles default to empty and groups still work.
* **Kubernetes SA tokens** (workbench pods) are detected via the `kubernetes.io` claim in an initial unverified decode. These are delegated to a lightweight TokenReview that validates the token and extracts the namespace from the SA identity. `NamespaceBasedPolicy` uses the namespace. No RBAC queries are performed, only `tokenreviews/create` is needed.

Additional server improvements:

* `verify_ssl` field (default `true`) controls TLS verification for OIDC discovery and JWKS endpoints. Set to `false` for self-signed certificates.
* `ca_cert_path` field allows specifying a custom CA certificate for the OIDC provider when `verify_ssl: true`.
* SSL certificate errors now produce clear, actionable log messages instead of generic "Invalid token" errors.
* `PyJWKClientError` is now caught (previously only `InvalidTokenError` was caught), preventing 500 responses when JWKS endpoints are unreachable or keys are not found.
* Malformed tokens in the intra-communication check no longer crash the parser.

### Server Configuration (`feature_store.yaml`)

```yaml
auth:
  type: oidc
  auth_discovery_url: https://keycloak.example.com/realms/myrealm/.well-known/openid-configuration
  client_id: feast-server       # optional, needed only for role extraction
  verify_ssl: true              # optional, default true
  ca_cert_path: /path/to/ca.crt # optional, for custom CA certs
```

### Keycloak Setup Required

The OIDC client in Keycloak must have a **Groups claim mapper** configured:
Client > Mappers > Add "Group Membership" mapper > claim name: `groups`

Without this mapper, JWTs will not contain the `groups` claim and `GroupBasedPolicy` will deny all access.

## Client Side Changes

**Files:** `oidc_authentication_client_manager.py`, `auth_model.py`, `repo_config.py`

`OidcAuthClientManager.get_token()` now supports multiple token sources with a clear priority:

1. `token` field in config: static JWT, returned directly, no network calls
2. `token_env_var` field: reads from the named environment variable on every call (supports token refresh). If the variable is configured but missing, raises `PermissionError` immediately with no fallback.
3. `client_secret` field (with `auth_discovery_url` and `client_id`): fetches token from the IDP via client_credentials or ROPG flow
4. `FEAST_OIDC_TOKEN` environment variable: last resort fallback for bare `{type: oidc}` configs
5. Kubernetes SA token file at `/var/run/secrets/kubernetes.io/serviceaccount/token`: for workbench pods running inside Kubernetes

`token`, `token_env_var`, and `client_secret` are mutually exclusive (enforced by a Pydantic validator at config load time).

### Client Configuration Examples

**Zero config (workbench pods or kube-authkit users):**
```yaml
auth:
  type: oidc
```
Picks up `FEAST_OIDC_TOKEN` env var if set, otherwise reads the mounted SA token.

**Named environment variable (supports external token refresh):**
```yaml
auth:
  type: oidc
  token_env_var: MY_APP_TOKEN
```

**Static token (testing / CI):**
```yaml
auth:
  type: oidc
  token: "eyJhbGciOiJSUzI1NiIs..."
```

**Client credentials flow (service to service):**
```yaml
auth:
  type: oidc
  auth_discovery_url: https://keycloak.example.com/realms/myrealm/.well-known/openid-configuration
  client_id: feast-client
  client_secret: my-secret
```

## Operator Changes

**Files:** `featurestore_types.go`, `repo_config.go`, `tls.go`, `services_types.go`, `authz.go`

The operator now generates separate server and client OIDC configs:

* **Server config** gets `auth_discovery_url` (resolved from three sources), optional `client_id`/`client_secret` from Secret, `verify_ssl`, and `ca_cert_path`.
* **Client config** gets only `{type: oidc}` with an optional `token_env_var`. Client pods never see server credentials.

### New CRD Fields on `OidcAuthz`

| Field | Type | Description |
|-------|------|-------------|
| `issuerUrl` | string | OIDC issuer URL. Operator appends `/.well-known/openid-configuration` to derive the discovery endpoint. |
| `secretRef` | LocalObjectReference (optional) | Secret with OIDC properties (`auth_discovery_url`, `client_id`, `client_secret`). |
| `secretKeyName` | string | Key in the Secret containing all OIDC properties as a single YAML value. |
| `tokenEnvVar` | string | Env var name for client pods to read an OIDC token from. |
| `verifySSL` | bool | Verify SSL certificates for the OIDC provider. Defaults to true. |
| `caCertConfigMap` | object (`name`, `key`) | ConfigMap with the CA certificate for self-signed OIDC providers. |

### Discovery URL Resolution Priority

1. CR `issuerUrl` field (derives discovery URL by appending `/.well-known/openid-configuration`)
2. Secret `auth_discovery_url` key (used verbatim)
3. `OIDC_ISSUER_URL` environment variable on the operator pod (for RHOAI/ODH zero config)

### CA Certificate Resolution Priority

1. CRD `caCertConfigMap` field (user specified ConfigMap, mounted at `/etc/pki/tls/oidc-ca/ca.crt`)
2. Auto detected `odh-trusted-ca-bundle` ConfigMap with `odh-ca-bundle.crt` key (RHOAI/ODH clusters)
3. System CA store (default when no custom CA is configured)

### RBAC

When `authz: oidc` is configured, the operator provisions a ClusterRole and ClusterRoleBinding granting the Feast server SA `tokenreviews/create` permission. This enables the SA token delegation path for workbench pods.

### Operator CR Example

**Upstream (with issuerUrl):**
```yaml
spec:
  authz:
    oidc:
      issuerUrl: "https://keycloak.example.com/realms/myrealm"
      secretRef:
        name: feast-oidc-secret
      verifySSL: true
      caCertConfigMap:
        name: keycloak-ca
        key: ca.crt
```

**ODH (OpenDataHub) zero config (platform injects OIDC_ISSUER_URL):**
```yaml
spec:
  authz:
    oidc: {}
```

## Permission Examples

```python
from feast.permissions.permission import Permission
from feast.permissions.policy import GroupBasedPolicy, NamespaceBasedPolicy
from feast.permissions.action import AuthzedAction, ALL_ACTIONS
from feast.feast_object import ALL_RESOURCE_TYPES
from feast.feature_view import FeatureView

# Grant full access to users in the "feast-admins" Keycloak group
admin_perm = Permission(
    name="admin_permissions",
    types=ALL_RESOURCE_TYPES,
    policy=GroupBasedPolicy(groups=["feast-admins"]),
    actions=ALL_ACTIONS,
)

# Grant read access to users in the "data-readers" Keycloak group
reader_perm = Permission(
    name="reader_permissions",
    types=[FeatureView],
    policy=GroupBasedPolicy(groups=["data-readers"]),
    actions=[AuthzedAction.DESCRIBE, AuthzedAction.READ_ONLINE],
)

# Grant full access to workbench pods in the "feast" namespace
workbench_perm = Permission(
    name="workbench_permissions",
    types=ALL_RESOURCE_TYPES,
    policy=NamespaceBasedPolicy(namespaces=["feast"]),
    actions=ALL_ACTIONS,
)
```

#6088

# Misc

---

<a href="https://app.devin.ai/review/feast-dev/feast/pull/6089" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/infra/feast-operator/bundle/manifests/feast.dev_featurestores.yaml
+++ b/infra/feast-operator/bundle/manifests/feast.dev_featurestores.yaml
@@ -2536,6 +2536,12 @@ spec:
                             type: object
                         type: object
                     type: object
+                  podAnnotations:
+                    additionalProperties:
+                      type: string
+                    description: PodAnnotations are annotations to be applied to the
+                      Deployment's PodTemplate metadata.
+                    type: object
                   podDisruptionBudgets:
                     description: PodDisruptionBudgets configures a PodDisruptionBudget
                       for the FeatureStore deployment.
@@ -8290,6 +8296,12 @@ spec:
                                     type: integer
                                 type: object
                             type: object
+                        type: object
+                      podAnnotations:
+                        additionalProperties:
+                          type: string
+                        description: PodAnnotations are annotations to be applied
+                          to the Deployment's PodTemplate metadata.
                         type: object
                       podDisruptionBudgets:
                         description: PodDisruptionBudgets configures a PodDisruptionBudget
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -934,7 +934,7 @@
         "filename": "infra/feast-operator/api/v1/featurestore_types.go",
         "hashed_secret": "44e17306b837162269a410204daaa5ecee4ec22c",
         "is_verified": false,
-        "line_number": 756
+        "line_number": 761
       }
     ],
     "infra/feast-operator/api/v1/zz_generated.deepcopy.go": [
@@ -943,21 +943,21 @@
         "filename": "infra/feast-operator/api/v1/zz_generated.deepcopy.go",
         "hashed_secret": "f914fc9324de1bec1ad13dec94a8ea2ddb41fc87",
         "is_verified": false,
-        "line_number": 686
+        "line_number": 693
       },
       {
         "type": "Secret Keyword",
         "filename": "infra/feast-operator/api/v1/zz_generated.deepcopy.go",
         "hashed_secret": "44e17306b837162269a410204daaa5ecee4ec22c",
         "is_verified": false,
-        "line_number": 747
+        "line_number": 754
       },
       {
         "type": "Secret Keyword",
         "filename": "infra/feast-operator/api/v1/zz_generated.deepcopy.go",
         "hashed_secret": "c2028031c154bbe86fd69bef740855c74b927dcf",
         "is_verified": false,
-        "line_number": 1293
+        "line_number": 1300
       }
     ],
     "infra/feast-operator/api/v1alpha1/featurestore_types.go": [
@@ -1539,5 +1539,5 @@
       }
     ]
   },
-  "generated_at": "2026-04-07T09:25:31Z"
+  "generated_at": "2026-04-07T15:56:56Z"
 }
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -70,6 +70,7 @@
   * [Multi-Team Feature Store Setup](how-to-guides/federated-feature-store.md)
 * [Running Feast in production (e.g. on Kubernetes)](how-to-guides/running-feast-in-production.md)
 * [Feast on Kubernetes](how-to-guides/feast-on-kubernetes.md)
+* [Online Server Performance Tuning](how-to-guides/online-server-performance-tuning.md)
 * [Customizing Feast](how-to-guides/customizing-feast/README.md)
   * [Adding a custom batch materialization engine](how-to-guides/customizing-feast/creating-a-custom-materialization-engine.md)
   * [Adding a new offline store](how-to-guides/customizing-feast/adding-a-new-offline-store.md)
--- a/docs/how-to-guides/online-server-performance-tuning.md
+++ b/docs/how-to-guides/online-server-performance-tuning.md
@@ -1,0 +1,991 @@
+# Online Server Performance Tuning
+
+This guide covers end-to-end performance tuning for the Feast Python online feature server, with a focus on Kubernetes deployments using the [Feast Operator](./feast-on-kubernetes.md). It brings together worker configuration, registry caching, online store selection, horizontal scaling, observability, and network optimization into a single operational playbook.
+
+## Request lifecycle
+
+Understanding where time is spent during a `get_online_features()` call helps you focus tuning efforts on the right layer:
+
+1. **Registry lookup** — resolve feature references to metadata (cached locally).
+2. **Online store read** — fetch feature values from the backing database (network I/O).
+3. **On-demand transformation** — execute any ODFVs attached to the request (CPU).
+4. **Response serialization** — encode the result as JSON or protobuf.
+
+In most production deployments, **step 2 (online store read)** dominates latency. Registry refresh (step 1) can cause periodic latency spikes if not configured for background refresh. Step 3 matters only when ODFVs are present.
+
+---
+
+## Feature view and feature service structuring
+
+How you organize features into feature views directly affects online serving latency, because **each distinct feature view in a request produces a separate online store read**. This is a structural decision that no amount of runtime tuning can fix after the fact.
+
+### Fewer feature views = fewer store round-trips
+
+When the server processes a `get_online_features()` call, it groups the requested features by their source feature view and issues one `online_read()` per group. For sync stores, these reads are **sequential**; for async stores (DynamoDB, MongoDB), they run **concurrently** via `asyncio.gather()`, but each is still a separate network round-trip.
+
+| Request shape | Store reads | Impact |
+| ------------- | ----------- | ------ |
+| 10 features from **1** feature view | 1 read | Minimal latency |
+| 10 features from **5** feature views | 5 reads | 5× the network round-trips (sequential) or 5 concurrent reads competing for connections (async) |
+| 10 features from **10** feature views | 10 reads | Latency dominated by store reads, not transformation |
+
+**Guideline:** For features that share the same entity key and are frequently requested together, consolidate them into a **single feature view**. This reduces the number of store round-trips per request. Split feature views only when features have different entities, different materialization schedules, or different data source update frequencies.
+
+### Feature services are free
+
+A [Feature Service](../getting-started/concepts/feature-retrieval.md) is a named collection of feature references — it's a convenience grouping, not a separate storage or execution unit. Using a feature service adds only a registry lookup (cached) compared to listing features individually. There is no performance penalty for using feature services, and they are the recommended way to define stable, versioned feature sets for production models.
+
+### ODFV overhead is additive
+
+Regular feature views incur only **store read** cost. On-demand feature views add **CPU-bound transformation** cost on top:
+
+```
+Total request latency ≈ store_read_time + sum(odfv_transformation_times)
+```
+
+Each ODFV in the request runs its transformation function after all store reads complete. The overhead depends on:
+
+- **Number of ODFVs** in the request — each one adds its transformation time.
+- **Transformation complexity** — a simple arithmetic operation is microseconds; ML model inference can be milliseconds.
+- **Transformation mode** — `mode="python"` is 3–10× faster than `mode="pandas"` for online serving (see [ODFV optimization](#on-demand-feature-view-odfv-optimization)).
+- **Entity count** — transformations run across all entity rows in the request. More entities = more work.
+
+### Hidden store reads from ODFV source dependencies
+
+ODFVs declare source feature views. When you request features from an ODFV, the server must also read **all source feature views** that the ODFV depends on, even if you didn't explicitly request features from those views. This can add unexpected store reads:
+
+```python
+@on_demand_feature_view(
+    sources=[driver_stats_fv, vehicle_stats_fv, request_source],
+    schema=[Field(name="combined_score", dtype=Float64)],
+    mode="python",
+)
+def combined_score(inputs: dict[str, Any]) -> dict[str, Any]:
+    return {"combined_score": [d + v for d, v in zip(inputs["driver_rating"], inputs["vehicle_rating"])]}
+```
+
+Requesting just `combined_score` triggers reads from **both** `driver_stats_fv` and `vehicle_stats_fv`. If those are already needed by other features in the same request, there's no extra cost (the server deduplicates). But if they're only needed by the ODFV, these are hidden reads you should account for.
+
+### Structuring recommendations
+
+| Practice | Why |
+| -------- | --- |
+| Group co-accessed features into the same feature view | Reduces store round-trips per request |
+| Use feature services to define stable production feature sets | No performance cost; improves governance and versioning |
+| Minimize the number of ODFVs per request | Each ODFV adds CPU-bound transformation latency |
+| Prefer `write_to_online_store=True` for ODFVs that don't need request-time data | Moves compute from serving to materialization path |
+| Audit ODFV source dependencies | Avoid pulling in unnecessary store reads via unused source feature views |
+| Use `track_metrics=True` on ODFVs during profiling | Identifies which transforms are the bottleneck |
+
+---
+
+## Worker and connection tuning
+
+The Python feature server uses Gunicorn with async workers. Tuning workers, connections, and timeouts directly impacts throughput and tail latency.
+
+### Feast Operator CR configuration
+
+```yaml
+apiVersion: feast.dev/v1
+kind: FeatureStore
+metadata:
+  name: my-feast
+spec:
+  feastProject: my_project
+  services:
+    onlineStore:
+      server:
+        workerConfigs:
+          workers: 4
+          workerConnections: 2000
+          maxRequests: 5000
+          maxRequestsJitter: 500
+          keepAliveTimeout: 30
+          registryTTLSeconds: 300
+        resources:
+          requests:
+            cpu: "2"
+            memory: 2Gi
+          limits:
+            cpu: "4"
+            memory: 4Gi
+```
+
+### CLI equivalent
+
+```bash
+feast serve \
+  --workers 4 \
+  --worker-connections 2000 \
+  --max-requests 5000 \
+  --max-requests-jitter 500 \
+  --keep-alive-timeout 30 \
+  --registry_ttl_sec 300
+```
+
+### Tuning recommendations
+
+| Parameter | Default | Recommended starting point | Notes |
+| --------- | ------- | -------------------------- | ----- |
+| `workers` | 1 | `2 × CPU cores + 1` (use `-1` for auto) | Each worker is a separate process. More workers = more parallel requests, but also more memory. |
+| `workerConnections` | 1000 | 1000–2000 | Max simultaneous connections **per worker**. Increase for high-concurrency workloads behind a load balancer. |
+| `maxRequests` | 1000 | 5000–10000 | Recycles a worker after N requests to guard against memory leaks. Higher values reduce recycle overhead. |
+| `maxRequestsJitter` | 50 | 200–500 | Randomizes recycle timing to avoid a thundering herd of simultaneous worker restarts. |
+| `keepAliveTimeout` | 30 | 30–60 | Seconds to keep an idle connection open. Match this to your load balancer's idle timeout. |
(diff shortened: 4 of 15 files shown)


khmelevskiy:
Thanks a lot for this 🙌  

We’re currently using a bit of a temporary hack by overriding `oidc_token_parser`, so really looking forward to this landing in a release and finally getting rid of it 🙂

Really appreciate the work here! 💜

botriki:
Thank you for this work.

Could you add an override for the "current_user" parameter? Our OIDC server uses the UPN instead of preferred_username and I'm currently working on a monkey patch solution for this. But it would be much more convenient to have this in the official release.

```
      if "upn" in data:
          current_user = data["upn"]
      elif "preferred_username" in data:
          current_user = data["preferred_username"]
      else:
          raise AuthenticationError(
              "Missing upn or preferred_username field in access token."
          )
```

sdk/python/feast/permissions/client/oidc_authentication_client_manager.py
```diff
@@ -10,31 +10,69 @@
...
+        if os.path.isfile(SA_TOKEN_PATH):
+            with open(SA_TOKEN_PATH) as f:
+                token = f.read().strip()
+            if token:
+                return token
+        return None
+
+    def _fetch_token_from_idp(self) -> str:
+        """Obtain an access token via client_credentials or ROPG flow."""
+        assert self.auth_config.auth_discovery_url is not None
```
ntkathole: raise ValueError instead of assert

sdk/python/feast/permissions/auth/oidc_token_parser.py
```diff
@@ -66,56 +148,79 @@ async def user_details_from_access_token(self, access_token: str) -> User:
...
+        """Validate a K8s SA token via TokenReview and extract the namespace.
+
+        Lightweight alternative to full KubernetesTokenParser — only validates
+        the token and extracts the namespace from the authenticated identity.
+        No RBAC queries (RoleBindings, ClusterRoleBindings) are performed,
+        so the server SA needs only ``tokenreviews/create`` permission.
+        """
+        from kubernetes import client, config
+
+        config.load_incluster_config()
```
ntkathole: re-creates the API client on every SA token validation, may be initialized once and reused ?

sdk/python/feast/permissions/client/oidc_authentication_client_manager.py
```diff
@@ -10,31 +10,69 @@
...
+            raise PermissionError(
+                "No OIDC token source configured. Provide one of: "
+                "'token', 'token_env_var', 'client_secret' (with "
+                "'auth_discovery_url' and 'client_id'), set the "
+                "FEAST_OIDC_TOKEN environment variable, or run inside "
+                "a Kubernetes pod with a mounted service account token."
+            )
+
+    @staticmethod
+    def _read_sa_token() -> str | None:
```
ntkathole:
```suggestion
    def _read_sa_token() -> Optional[str]:
```

sdk/python/feast/permissions/auth/oidc_token_parser.py
```diff
@@ -50,12 +57,87 @@ async def _validate_token(self, access_token: str):
...
+    @staticmethod
+    def _extract_username_or_raise_error(data: dict) -> str:
+        """Extract the username from the decoded JWT. Raises if missing — identity is mandatory.
+
+        Checks ``preferred_username`` first (Keycloak default), then falls back
+        to ``upn`` (Azure AD / Entra ID).
+        """
+        if "preferred_username" in data:
+            return data["preferred_username"]
+        if "upn" in data:
```
ntkathole, line 71: `_get_intra_comm_user` also need to read upn ?

infra/feast-operator/config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/feastdev/feast-operator
-  newTag: 0.61.0
+  newName: quay.io/aniket-redhat/feast-operator
```
ntkathole: remove

docs/getting-started/components/authz_manager.md
```diff
@@ -97,6 +132,12 @@ auth:
   auth_discovery_url: http://localhost:8080/realms/master/.well-known/openid-configuration
 ```
 
+When using client credentials or ROPC flows, the `verify_ssl` setting also applies to the discovery and token endpoint requests.
+
+#### Multi-Token Support (OIDC + Kubernetes Service Account)
+
+When the Feast server is configured with OIDC auth and deployed on Kubernetes, the `OidcTokenParser` can handle both Keycloak JWT tokens and Kubernetes service account tokens. Incoming tokens that contain a `kubernetes.io` claim are validated via the Kubernetes Token Access Review API and the namespace is extracted from the authenticated identity — no RBAC queries are performed, so the server service account only needs `tokenreviews/create` permission. All other tokens follow the standard OIDC/Keycloak JWKS validation path. This enables `NamespaceBasedPolicy` enforcement for service account tokens while using `GroupBasedPolicy` and `RoleBasedPolicy` for OIDC user tokens.
```
jyejare, line 139: Do we support OIDC token passing from ODH Notebook instead of service account token ? This we alreayd support for kubernetes token.

Not a blocker though, just for consistency.

infra/feast-operator/bundle/manifests/feast.dev_featurestores.yaml
```diff
@@ -76,6 +80,15 @@ spec:
                             type: string
                         type: object
                         x-kubernetes-map-type: atomic
+                      tokenEnvVar:
+                        description: The name of the environment variable that client
+                          pods will use to read a pre-existing OIDC token.
```
jyejare: Could you clarify which client pods we are talking about here ?

infra/feast-operator/bundle/manifests/feast.dev_featurestores.yaml
```diff
@@ -5695,7 +5712,6 @@ spec:
                 type: object
             required:
             - feastProject
-            - replicas
```
jyejare, line 5704: Is this change from auto-scaling commit. Does this PR need rebase ?

sdk/python/feast/permissions/auth/oidc_token_parser.py
```diff
@@ -66,56 +148,79 @@ async def user_details_from_access_token(self, access_token: str) -> User:
...
+                username=current_user,
+                roles=roles,
+                groups=groups,
+            )
         except jwt.exceptions.InvalidTokenError:
             logger.exception("Exception while parsing the token:")
             raise AuthenticationError("Invalid token.")
 
+    @staticmethod
+    async def _validate_k8s_sa_token_and_extract_namespace(access_token: str) -> User:
```
jyejare: I think we should create a common function for k8s sa token extraction to be used between both kubeterntes auth and OIDC auth.

sdk/python/feast/permissions/auth_model.py
```diff
@@ -1,71 +1,73 @@
...
+    """Validate that at most one named group is configured, and completely.
+
+    Each *group* is a tuple of field values.
+    A group is **active** only when *all* its values are truthy.
+    A group is **partial** (error) when *any* but not *all* values are truthy.
+    At most one active group may exist.
+    """
+    partial = [name for name, vals in groups.items() if any(vals) and not all(vals)]
+    if partial:
+        raise ValueError(
```
jyejare, line 20: The current message is generic and doesn't specify which fields are actually required. A more helpful error would list the incomplete group's field names. For example:


```Python
raise ValueError(
    f"Incomplete configuration for '{partial[0]}': "
    f"all required fields for this group must be set together. "
    f"Check the documentation for valid credential combinations."
)
```
Or even better, you could inspect the groups parameter to show exactly what's missing:

```Python
raise ValueError(
    f"Incomplete configuration for '{partial[0]}': "
    f"configure all of these fields together, or none at all."
)
```
This would give users clearer guidance on what went wrong and how to fix it.

sdk/python/feast/permissions/auth/oidc_token_parser.py
```diff
@@ -50,12 +57,87 @@ async def _validate_token(self, access_token: str):
...
+        )
+
+    @staticmethod
+    def _is_kubernetes_token(access_token: str) -> bool:
+        """Check if the token contains the ``kubernetes.io`` claim (a dict with namespace, pod, serviceaccount)."""
+        try:
+            unverified = jwt.decode(access_token, options={"verify_signature": False})
+        except jwt.exceptions.DecodeError as e:
+            raise AuthenticationError(f"Failed to decode token: {e}")
+        return isinstance(unverified.get("kubernetes.io"), dict)
```
jyejare: This runs on every non-intra-comm request before the normal OIDC path. The decoded payload is then discarded and the token is decoded again via _decode_token() for standard OIDC tokens. Consider caching the unverified decode result or restructuring to decode once.

sdk/python/feast/permissions/auth/oidc_token_parser.py
aniketpalu: Not needed here. Intra-communication tokens are self-generated by Feast's get_token() method which always encodes preferred_username as the key (see OidcAuthClientManager.get_token() -> jwt.encode({"preferred_username": intra_communication_base64, ...})). These tokens never come from a real IDP, so upn will never appear.

docs/getting-started/components/authz_manager.md
aniketpalu: Yes, this PR enables that. When a workbench notebook uses auth: {type: oidc}, the SDK's get_token() supports multiple token sources:

- FEAST_OIDC_TOKEN env var: the user (or kube-authkit) sets this with a Keycloak JWT. The human user's identity, groups, and roles flow through to the Feast server.
- SA token file fallback: if no env var is set, the workbench pod's SA token at var/run/secrets/kubernetes.io/serviceaccount/token is read automatically.

The server handles both. OIDC JWTs are validated against Keycloak JWKS, and SA tokens are detected via the kubernetes.io claim and validated via TokenReview. So both human OIDC tokens and SA tokens work from ODH Notebooks.

infra/feast-operator/bundle/manifests/feast.dev_featurestores.yaml
aniketpalu: "Client pods" refers to any pod running the Feast SDK as a client that connects to the Feast server. Would you like me to change description?

infra/feast-operator/bundle/manifests/feast.dev_featurestores.yaml
aniketpalu: Yes, rebased

sdk/python/feast/permissions/auth/oidc_token_parser.py
aniketpalu: This is good idea, but I thought to not touch k8s auth for this PR & this can be takes code improvement.

sdk/python/feast/permissions/auth_model.py
aniketpalu: Done, Thanks

sdk/python/feast/permissions/auth/oidc_token_parser.py
aniketpalu: Thanks for the catch

Fixed, the code now performs exactly one unverified decode at the top of user_details_from_access_token()

infra/feast-operator/bundle/manifests/feast.dev_featurestores.yaml
```diff
@@ -13920,6 +13983,10 @@ spec:
                     x-kubernetes-validations:
                     - message: One selection required.
                       rule: '[has(self.local), has(self.remote)].exists_one(c, c)'
+                  runFeastApplyOnInit:
```
ntkathole, line 13998: regenerate and resolving conflicts

aniketpalu: Done

sdk/python/feast/permissions/auth/oidc_token_parser.py
```diff
@@ -50,79 +59,201 @@ async def _validate_token(self, access_token: str):
...
             logger.debug("Token successfully validated.")
         except Exception as e:
+            if self._is_ssl_error(e):
+                logger.error(
+                    "OIDC provider SSL certificate verification failed. "
+                    "If using a self-signed certificate, set verify_ssl: false "
+                    "or provide a CA certificate via ca_cert_path."
+                )
             logger.error(f"Token validation failed: {e}")
             raise AuthenticationError(f"Invalid token: {e}")
```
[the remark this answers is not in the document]
aniketpalu: The unverified decode is intentional for routing (SA token detection, intra-comm check)
OAuth2AuthorizationCodeBearer.__call__() does not make network calls, so there is no SSL mismatch
The _validate_token being a lightweight header check is pre-existing behavior, not introduced by this PR, and is tracked as a future improvement

infra/feast-operator/bundle/manifests/feast-operator.clusterserviceversion.yaml
```diff
@@ -259,6 +270,7 @@ spec:
                   value: quay.io/feastdev/feature-server:0.61.0
                 - name: RELATED_IMAGE_CRON_JOB
                   value: quay.io/openshift/origin-cli:4.17
+                - name: OIDC_ISSUER_URL
```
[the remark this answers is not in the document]
aniketpalu: operator-sdk generate bundle strips value: "" from env vars by design. This is a known operator-sdk behavior as it considers value: "" equivalent to absent and removes it during YAML serialization. The source manager.yaml has value: "", kustomize build (install.yaml) preserves it, but operator-sdk (CSV) strips it.

infra/feast-operator/internal/controller/authz/authz.go
```diff
@@ -25,6 +25,16 @@ func (authz *FeastAuthorization) Deploy() error {
 	_ = authz.Handler.DeleteOwnedFeastObj(authz.initFeastRole())
 	_ = authz.Handler.DeleteOwnedFeastObj(authz.initFeastRoleBinding())
 	apimeta.RemoveStatusCondition(&authz.Handler.FeatureStore.Status.Conditions, feastKubernetesAuthConditions[metav1.ConditionTrue].Type)
+
+	if authz.isOidcAuth() {
+		if err := authz.createFeastClusterRole(); err != nil {
```
ntkathole, line 30: createFeastClusterRoleis using setFeastClusterRole, which gives OIDC-only deployments unnecessary exctra permissions, but since those are only read-only, considering it as non-blocker, can be handled as followup

ntkathole: this also leave the ClusterRole and ClusterRoleBinding are orphaned when a user switches from OIDC auth to no auth (or to Kubernetes auth)

ntkathole:
@aniketpalu Please make sure to have followup changes as required for non blocking issues

infra/feast-operator/internal/controller/authz/authz.go
aniketpalu: Will address this in follow up PR. Thanks for review.

Issue #2728, opened 2022-05-23
Improved feature view and model versioning


adchia:
# Problem
When using Feast, it can be hard to version features with respect to models in production. 
- Users modifying a re-used feature view (e.g. in response to upstream schema change) can accidentally impact other models in production
- Hard to tie versions of features together or versions of models (~`FeatureService`)

As Feast adds batch transformations or #2597, transformation logic will also need to be versioned. This is already an issue today with on demand feature views too.  

# Potential solution
One way to address this is to double down on:
### 1. Feature services as the central way to retrieve features + track models. 
  - Registering a feature service signifies dependent features / feature views are in "production"
  - Allow `feature_service.get_historical_features` or `feature_service.get_online_features`
### 2. Improvements to `feast plan` and `feast apply`
  - `feast plan` that prevents deletion or editing of feature views that feature services in prod depend on (Modifying https://github.com/feast-dev/feast/blob/master/sdk/python/feast/feature_store.py#L594)
    - If you remove a feature service, then the underlying feature views could be removed if no other feature service depend on it
  - `feast apply` that fails in above case by default (overridable though)
     - e.g. modifying the logic at https://github.com/feast-dev/feast/blob/master/sdk/python/feast/feature_store.py#L676
### 3. Feature view versioning
  - Not super sure here on the best practice, but a way to say that feature view B is a new version of feature view A.

franciscojavierarceo:
Keeping this open until all sub issues are closed for online stores.

PR #6101, opened 2026-03-12
feat: Add version tracking to FeatureView

franciscojavierarceo:
## Summary

Adds automatic version tracking to Feast feature views (`FeatureView`, `StreamFeatureView`, `OnDemandFeatureView`). Every `feast apply` that changes a feature view's schema or UDF creates a versioned snapshot in the registry. Users can list version history, pin serving to a prior version, query specific versions via `@v<N>` syntax, and stage new versions without promoting them.

Resolves https://github.com/feast-dev/feast/issues/2728

## Key Features

- **Automatic version history** — every schema/UDF change is tracked; metadata-only changes (description, tags, TTL) update in place without creating a new version
- **Version pinning** — declaratively pin to a historical version: `FeatureView(name="driver_stats", version="v2")`
- **Version-qualified online reads** — query specific versions: `driver_stats@v2:trips_today` (SQLite online store)
- **Staged publishing (`--no-promote`)** — `feast apply --no-promote` saves a new version without promoting it to active, enabling phased rollouts without a transition window
- **Version-aware materialization** — `feast materialize --views fv --version v2` targets a specific version's online store table
- **Feature service support** — feature services automatically resolve versioned FVs when `enable_online_feature_view_versioning` is enabled
- **Pin conflict detection** — prevents applying a schema change and version pin simultaneously
- **Concurrent apply safety** — SQL registry uses unique constraints with retry; file registry is last-write-wins (pre-existing limitation)
- **Full backward compatibility** — `version` is optional; omitting it is identical to `version="latest"`

## Changes

**Proto layer:**
- New `FeatureViewVersion.proto` with `FeatureViewVersionRecord` / `FeatureViewVersionHistory`
- Added `version` field to `FeatureViewSpec`, `StreamFeatureViewSpec`, `OnDemandFeatureViewSpec`
- Added `current_version_number` to meta messages, `version_tag` to `FeatureViewProjection`
- Added `feature_view_version_history` to `Registry` proto

**Python SDK — core versioning:**
- `version_utils.py` — parses `"latest"`, `"v2"`, `"version2"` (case-insensitive)
- `version` parameter on `FeatureView`, `StreamFeatureView`, `OnDemandFeatureView`
- Version-aware `apply_feature_view` in file registry, SQL registry (with concurrent apply retry)
- `list_feature_view_versions(name, project)` on registries and `FeatureStore`
- `get_feature_view_by_version(name, project, version_number)` for explicit version lookup
- `_schema_or_udf_changed()` to distinguish version-significant changes from metadata updates

**Python SDK — `--no-promote` flag:**
- `feast apply --no-promote` CLI option threaded through `apply_total()` → `apply_total_with_repo_instance()` → `store.apply()` / `store._apply_diffs()` → `registry.apply_feature_view()`
- File registry: saves version snapshot, restores old active proto
- SQL registry: saves version snapshot, skips active row UPDATE
- Python SDK: `store.apply([...], no_promote=True)`

**Python SDK — version-qualified reads:**
- `_parse_feature_ref()` and `_strip_version_from_ref()` utilities for `fv@v<N>:feature` parsing
- Version-aware online feature retrieval with `@v<N>` syntax
- Version metadata injection in online feature responses
- Feature service version resolution with `version_tag` on projections

**Python SDK — feature ref parsing audit & fixes:**
- Fixed ibis offline store prefix matching to use `name_to_use()` for versioned refs
- Fixed `passthrough_provider.py` saved dataset column naming to strip `@v<N>`
- Fixed Go `ParseFeatureReference` to strip `@v<N>` from feature view name

**CLI:**
- `feast feature-views versions <name>` — list version history
- `feast apply --no-promote` — stage versions without promoting
- `feast materialize --views <name> --version v<N>` — version-aware materialization

**UI:**
- Version display on feature view detail pages (Regular, Stream, OnDemand)
- New "Versions" tab showing version history with timestamps

**Documentation:**
- RFC: `docs/rfcs/feature-view-versioning.md` — full design, concurrency, feature services, staged publishing
- Alpha reference: `docs/reference/alpha-feature-view-versioning.md`
- Updated feature view concepts page with `[Alpha] Versioning` section
- Updated CLI reference, retrieval docs, and SUMMARY.md

## Test plan

- [x] 47 unit tests: version parsing, normalization, proto roundtrip, equality, copy, feature ref parsing, table ID generation
- [x] 37 integration tests: apply/version creation, idempotency, pinning, pin conflicts, forward declaration, schema vs metadata changes, stream FV UDF changes, version metadata, feature service gates, `--no-promote` (5 tests)
- [x] Pre-commit hooks pass (format, lint, detect-secrets)
- [x] mypy type checking passes on all modified files
- [ ] CI: existing unit tests pass (no regressions)

---

UI Preview

Screenshot 2026-03-19 at 8 57 59 AM

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<a href="https://app.devin.ai/review/feast-dev/feast/pull/6101" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/go/internal/feast/onlineserving/serving.go
+++ b/go/internal/feast/onlineserving/serving.go
@@ -21,6 +21,8 @@ import (
 	"github.com/feast-dev/feast/go/types"
 )
 
+var versionTagRegex = regexp.MustCompile(`^[vV]\d+$`)
+
 /*
 FeatureVector type represent result of retrieving single feature for multiple rows.
 It can be imagined as a column in output dataframe / table.
@@ -494,13 +496,15 @@ func ParseFeatureReference(featureRef string) (featureViewName, featureName stri
 		featureName = parsedFeatureName[1]
 	}
 
-	// Reject @v<N> version qualifier — Go feature server does not support versioned reads
+	// Handle @version qualifier on feature view name
 	if atIdx := strings.Index(featureViewName, "@"); atIdx >= 0 {
 		suffix := featureViewName[atIdx+1:]
-		matched, _ := regexp.MatchString(`^[vV]\d+$`, suffix)
-		if matched {
+		if versionTagRegex.MatchString(suffix) {
 			e = fmt.Errorf("versioned feature refs (@%s) are not supported by the Go feature server", suffix)
 			return
+		}
+		if strings.EqualFold(suffix, "latest") {
+			featureViewName = featureViewName[:atIdx]
 		}
 	}
 	return


tokoko:
I haven't been able to go through the whole thing yet, but iiuc the expectation is that online stores would be expected to handle this by essentially treating fv1@v1 and fv1@v2 as two different feature views, right? (sqlite example infers table name from the version). Just wondering, maybe that's to high a bat for online store implementations. for example, if you have fv1@v1 and fv1@v2 both containing 100 features that differ by a single feature only, the online store would have to keep redundant copies of the other 99 features or do some complicated logic to diff values and deduplicate.

wdyt about concentrating on introducing versioning on feature level instead? tbh, that makes more intuitive sense to me. there's a feature called `income` that one found a fault with. we enable them to introduce income@v2 and somehow deprecate income@v1. online store can handle income@v1 and income@v2 as separate features, but all the rest of the features are unaffected.

franciscojavierarceo:
> I haven't been able to go through the whole thing yet, but iiuc the expectation is that online stores would be expected to handle this by essentially treating fv1@v1 and fv1@v2 as two different feature views, right? (sqlite example infers table name from the version). Just wondering, maybe that's to high a bat for online store implementations. for example, if you have fv1@v1 and fv1@v2 both containing 100 features that differ by a single feature only, the online store would have to keep redundant copies of the other 99 features or do some complicated logic to diff values and deduplicate.
> 
> wdyt about concentrating on introducing versioning on feature level instead? tbh, that makes more intuitive sense to me. there's a feature called `income` that one found a fault with. we enable them to introduce income@v2 and somehow deprecate income@v1. online store can handle income@v1 and income@v2 as separate features, but all the rest of the features are unaffected.

I spent some time reasoning about feature-level versioning with Claude. My initial reaction was that it's too large of a change and it only works today in a broken sense. 

By "in a broken sense" I mean that today we don't really version feature views or features, if someone changes the feature view or the feature, we just overwrite it and lose history. Moreover, we essentially force the materialization to be out of sync. Today, that just ignores the behavior silently. 

I like how we've implemented it here (i.e., creating a new table and storing the history of the metadata but allowing for callers to specify exact versions and defaulting to the existing behavior) because it is far more explicit about the challenges of materialization consistency issues when you change feature versions.

 So, I don't recommend we do feature-level versioning as I worry it makes materialization very unreliable. We can, in fact, declare feature view level versioning because transformations are a collection of features mapped one-to-one with a table.

franciscojavierarceo:
## Response to Devin Review Findings

Thanks for the thorough automated review! Here's a summary of where all the findings stand after the latest commits (`66c280b2f`, `cfc038b6f`):

### Already Resolved (confirmed by Devin itself in follow-up passes)

The following were real issues that have been fixed in prior commits:

- **Config field name mismatch** (`enable_feature_view_versioning` vs `enable_online_feature_view_versioning`) — fixed in `66c280b2f`
- **`StreamFeatureView.from_proto` missing `current_version_number` restoration** — fixed with full if/elif/else block
- **`SqlRegistry` read_engine vs write_engine** issues (snapshot detection, `_get_next_version_number`) — fixed to use `write_engine`
- **`delete_feature_view` not cleaning up version history** (both file and SQL registries) — fixed
- **Metadata-only updates not committed** — fixed with `commit()` in early return path
- **`_update_metadata_fields` not updating `spec.version`** — fixed
- **Version-qualified refs breaking `split(":")` calls** (`_augment_response_with_on_demand_transforms`, `retrieve_online_documents`, `_resolve_feature_counts`) — all fixed to use `_parse_feature_ref`
- **Version-qualified features dropped from response with `full_feature_names=True`** — fixed with version-aware `requested_result_row_names`
- **SqlRegistry snapshot protos missing `spec.project`** — fixed
- **SqlRegistry `IntegrityError` on re-creating deleted FV** — fixed with version history cleanup in `delete_feature_view`
- **Stale `current_version_number` in snapshot protos** (both registries) — fixed by setting version number before serialization

### Not Bugs / By Design

The remaining open findings are either false positives or by-design behavior:

| Finding | Verdict |
|---------|---------|
| SQLite `_table_id` routes to versioned tables | **By design** — this is the core versioning mechanism; `version > 0` guard preserves backward compat for v0 |
| `_validate_feature_refs` collision with versioned refs | **Not a bug** — `@v1` and `@v2` refs are distinct strings in `Counter`, no false collision |
| `_get_any_feature_view` missing in SqlRegistry | **Wrong** — method exists (line 422) |
| `version_tag` not persisted in proto serialization | **By design** — `version_tag` is a runtime-only field set during query resolution, not during FeatureService definition |
| `BatchFeatureView` missing `version` constructor param | **Acceptable** — version is set post-construction via `from_proto`, and the proto round-trip path works correctly |
| `_check_versioned_read_support` allows v0 through | **Not a bug** — `0 is not None` evaluates to `True`, so `@v0` refs ARE checked |
| `_schema_or_udf_changed` asymmetric comparison | **Not a bug** — `sorted()` is symmetric |
| Early return without updating proto list | **Not a bug** — in-place protobuf mutation + commit works correctly |
| Elasticsearch `retrieve_online_documents` signature | **Not a bug** — signature matches base class |
| `OnDemandFeatureView.from_proto` version before projection | **No functional impact** — `version` field is independent of projection creation |
| `StreamFeatureView.__eq__` redundant version check | **Already resolved** — current code doesn't have redundant check |
| ODFV proto `Next available id` comment | **Not a bug** — comment value is correct |
| `retrieve_online_documents` `_parse_feature_ref` rejects bare names | **Already fixed** by Devin |
| `version_tag` not set when projection is None | **Not a real issue** — FeatureService always provides projections; string refs set it at line 1331-1333 |
| SQL registry `apply_feature_view` snapshot bytes vs final proto | **Not a bug** — snapshot bytes are re-serialized after setting `current_version_number` |
| File registry skips version history for pin/revert | **By design** — pin/revert restores a snapshot, shouldn't create new versions |

docs/getting-started/concepts/feature-retrieval.md
```diff
@@ -78,15 +78,19 @@ feature_store.get_historical_features(features=feature_service, entity_df=entity
...
 
 ```python
 online_features = fs.get_online_features(
     features=[
-        'driver_locations:lon',
-        'drivers_activity:trips_today'
+        'driver_locations:lon',                # latest version (default)
+        'drivers_activity:trips_today',        # latest version (default)
+        'drivers_activity@v2:trips_today',     # specific version
+        'drivers_activity@latest:trips_today', # explicit latest
```
haoxu0, line 93: although unlikely, it can still happen that a feature name has a `@`, will the apply be able to identify and abort safely?

sdk/python/feast/infra/registry/registry.py
```diff
@@ -530,6 +811,54 @@ def apply_feature_view(
                     del existing_feature_views_of_same_type[idx]
                     break
 
+        # Version history tracking
+        if is_latest:
```
ntkathole, line 815:
```suggestion
        if self.enable_online_versioning and is_latest:
```

I think it's better to keep version history writes behind the flag as well and make this feature fully opt-in. May be we can make it default in future commits/releases ?

franciscojavierarceo: Disagree, I think having the history is important. Having the retrieval is optional.

franciscojavierarceo: Today we silently delete data and that's **wrong**.

ntkathole: That's fair 👍

docs/getting-started/concepts/feature-retrieval.md
franciscojavierarceo: I added validation for this so that new `feast apply` with `@` in the feature view names are rejected but old ones are used. 

For the latter case we have to make some assumptions about format because we support `@vN`, `@versionN`, and `@latest` so those rare edge cases could cause headaches but we've documented it. Thanks for calling it out!

sdk/python/feast/feature_view.py
```diff
@@ -460,6 +494,7 @@ def to_proto_spec(
             feature_transformation=feature_transformation_proto,
             mode=mode_to_string(self.mode),
             enable_validation=self.enable_validation,
+            version=self.version,
```
ntkathole, line 497:
```suggestion
            version=self.version if self.version and self.version.lower() != "latest" else "",,
```

After upgrade from old version, registry will treat every feature-view/odfv as changed ?

sdk/python/feast/infra/registry/registry.py
```diff
@@ -530,6 +811,54 @@ def apply_feature_view(
...
+                feature_view_proto = feature_view.to_proto()
+                feature_view_proto.spec.project = project
+                new_proto_bytes = feature_view_proto.SerializeToString()
+                self._save_version_record(
+                    feature_view.name, project, next_ver, fv_type_str, new_proto_bytes
+                )
+            else:
+                # New FV: save as v0
+                feature_view.current_version_number = 0
+                feature_view_proto = feature_view.to_proto()
```
ntkathole, line 855: Even when online versioning is off, the registry sets current_version_number on every feature view ? ( can version number only be set on the snapshot copy that goes into version history ?)

sdk/python/feast/feature_view.py
franciscojavierarceo: Here's what happens on upgrade:
  1. Old FV in registry has `spec.version = ""`
  2. `feast apply` runs → `_update_metadata_fields` writes `version="latest"` to the proto
  3. Registry gets committed with the updated bytes

That's it. There's no spurious version bump, no "UPDATED" in feast plan output (because `__eq__` normalizes both to `latest`), and no user-visible side effect. The registry bytes change once on upgrade, which is
  fine, it's just a one-time migration.

The` _schema_or_udf_changed` and `__eq__` comparisons both handle the `""` → `"latest"`
  normalization correctly. The proto bytes update silently in the background, which is harmless.

sdk/python/feast/infra/registry/registry.py
franciscojavierarceo: The current_version_number on the active definition is intentional, it records which version is currently deployed (e.g. "this is v3"). It's not something users interact with, it's metadata that reflects reality.

Users never need to set or declare a version number. The version parameter defaults to `"latest"`, so a normal `feast apply` just works—it auto-increments the version number when the schema changes and stamps it on the active definition. The only time a user specifies a version is if they want to explicitly revert to an old snapshot (e.g. `version="v2"`).

Keeping the version number on the active definition also means that if `enable_online_feature_view_versioning` is turned on later, the online store immediately knows which version table to use, no migration step needed.

sdk/python/feast/errors.py
```diff
@@ -128,6 +128,39 @@ def __init__(self, name, project=None):
...
+        else:
+            super().__init__(f"Version {version} of feature view {name} does not exist")
+
+
+class VersionedOnlineReadNotSupported(FeastError):
+    def __init__(self, store_name: str, version: int):
+        super().__init__(
+            f"Versioned feature reads (@v{version}) are not yet supported by {store_name}. "
+            f"Currently only SQLite supports version-qualified feature references. "
+            f"See https://github.com/feast-dev/feast/issues/6200"
```
ntkathole: This issue doesn't exists, need to change

sdk/python/feast/infra/registry/snowflake.py
```diff
@@ -259,8 +259,13 @@ def apply_feature_service(
...
+        self,
+        feature_view: BaseFeatureView,
+        project: str,
+        commit: bool = True,
+        no_promote: bool = False,
     ):
+        feature_view.ensure_valid()
         fv_table_str = self._infer_fv_table(feature_view)
         fv_column_name = fv_table_str[:-1]
         return self._apply_object(
```
YassinNouh21, line 276: `no_promote` is accepted here but never used — `_apply_object()` is called unconditionally, so the parameter is silently ignored. More importantly, the Snowflake registry inherits the base `list_feature_view_versions` / `get_feature_view_by_version` which both raise `NotImplementedError`. So a user on Snowflake sees `--no-promote` in the CLI, but versioning doesn't actually work — `feast feature-views list-versions` just says "not supported" and `@v2` refs will blow up at runtime.

Two options:
1. Add a version history table to Snowflake (same as the SQL registry did).
2. If that's out of scope, have `apply_feature_view` raise explicitly when `no_promote=True`, so users get a clear error instead of silent no-op behavior.

go/internal/feast/onlineserving/serving.go
```diff
@@ -492,6 +493,15 @@ func ParseFeatureReference(featureRef string) (featureViewName, featureName stri
 		featureViewName = parsedFeatureName[0]
 		featureName = parsedFeatureName[1]
 	}
+
+	// Strip @v<N> version qualifier from feature view name
+	if atIdx := strings.Index(featureViewName, "@"); atIdx >= 0 {
+		suffix := featureViewName[atIdx+1:]
+		matched, _ := regexp.MatchString(`^[vV]\d+$`, suffix)
+		if matched {
+			featureViewName = featureViewName[:atIdx]
```
YassinNouh21, line 507: This silently drops the version qualifier and reads from the unversioned (latest) table. A request for `driver_stats@v2:trips_today` returns whatever the current version is, not v2's data — and the caller has no way to know they got the wrong thing.

Since the Go feature server doesn't have a versioned registry lookup, it'd be safer to return an error here ("versioned feature refs are not supported by the Go feature server") rather than silently serving stale/wrong data. A TODO + doc note would also work if you want to keep forward-compat parsing.

sdk/python/feast/utils.py
```diff
@@ -60,6 +60,55 @@
...
+    version_str = fv_part[at_idx + 1 :]
+
+    if not version_str or version_str.lower() == "latest":
+        return (fv_name, None, feature_name)
+
+    # Parse version number from formats like "v2", "V2"
+    match = re.match(r"^[vV](\d+)$", version_str)
+    if not match:
+        # Not a recognized version format — treat entire fv_part as the name
+        return (fv_part, None, feature_name)
```
YassinNouh21, line 98: Falling back to treating the whole `fv_part` as the name means `driver_stats@abc:trips` silently becomes a lookup for a feature view literally named `driver_stats@abc`, which almost certainly doesn't exist. The user gets a confusing "feature view not found" error much later instead of a clear parse-time error.

Since `@` is now a reserved separator for versioning, unrecognized suffixes should raise immediately — something like `ValueError(f"Unrecognized version format '{version_str}' in '{ref}'. Expected @v<N> or @latest.")`. That way typos like `@v2a` or `@beta` fail fast with an actionable message.

sdk/python/feast/infra/registry/sql.py
```diff
@@ -540,6 +572,13 @@ def delete_feature_view(self, name: str, project: str, commit: bool = True):
             )
         if deleted_count == 0:
             raise FeatureViewNotFoundException(name, project)
+        # Clean up version history for the deleted feature view
```
YassinNouh21: The FV delete (via `_delete_object`) and the version history cleanup run in separate transactions. If the second `DELETE` fails (connection drop, constraint error, etc.), the feature view is already gone but its version history rows are orphaned. If someone later creates a new FV with the same name, the old version snapshots reappear — that's a data integrity bug.

Since `_delete_object` opens its own `write_engine.begin()` block, a clean fix would be to move the version history delete *inside* `delete_feature_view`'s loop body right after `_delete_object`, or refactor `_delete_object` to accept an open connection/transaction so both deletes share the same commit boundary.

sdk/python/feast/utils.py
franciscojavierarceo: This is intentional, the fallback preserves backward compatibility for pre-existing feature view names that happen to contain `@`. We already reject `@` in _**new**_ FV names at `feast apply` time, so this path only fires for legacy names. Raising here would be a breaking change for anyone with those names.

If the FV doesn't exist, the user will still get a clear `FeatureViewNotFoundException` downstream, it just happens at lookup time rather than parse time.

sdk/python/feast/infra/registry/sql.py
franciscojavierarceo: fixed

go/internal/feast/onlineserving/serving.go
franciscojavierarceo: fixed

sdk/python/feast/infra/registry/snowflake.py
franciscojavierarceo: fixed

Issue #5615, opened 2025-09-15
Support OpenAI Search API


franciscojavierarceo:
**Is your feature request related to a problem? Please describe.**
Feast introduced the standard language in the AI/ML era on feature stores, particularly the language around "online features".

In the Agentic era, OpenAI has defined the API for inference with it's Chat Completions API. It has not introduced the Responses API (their answer for an agent-native API) and this API supports "File Search". 

Many frameworks have empowered the open source community by being compatible with OpenAI's inference APIs. Compatibility with the  Search API could empower Data Scientists and Machine Learning Engineers to drive impact in their RAG and Agentic applications.

**Describe the solution you'd like**
We would support the Vector Store APIs and Search APIs.

**Describe alternatives you've considered**
N/A

**Additional context**
We've introduced this functionality in another community I am working with and I think it could be impactful here as well. 

Before choosing to support this, it'd be great to hear from the community if there is interest.

franciscojavierarceo:
CC @shuchu @HaoXuAI @ntkathole

shuchu:
I agree. We need to support this feature.

haoxu0:
LGTM
The only thing is making sure the dependencies are separate from the main sdk to avoid library size explosion:)

patelchaitany:
Hey @franciscojavierarceo I have One doubt here what do mean by the vector store API here dose it going to be used to create vector store (Described in the OpenAI vector store API docs) or it is going to be used for the file upload.

franciscojavierarceo:
@HaoXuAI yeah agree but I don't think t his will warrant a blow up in dependencies as it's just honoring the REST API layer.

patelchaitany:
I have some question regarding this - 


1. Currently, Feast does not have a built-in text embedding module in its serving path. The existing retrieve_online_documents and retrieve_online_documents_v2 APIs require the caller to provide pre-computed embedding vectors — Feast does not convert raw text queries into embeddings itself.

2. API Layer vs. Retrieval Layer Compatibility

- There are two possible approaches for OpenAI compatibility:

   1. REST API translation layer only: Add new FastAPI endpoints in feature_server.py that accept OpenAI shaped requests, translate them into calls to the existing retrieve_online_documents_v2, and reshape the response to match OpenAI's format. The retrieval itself remains unchanged the compatibility is purely at the HTTP boundary.
   
[...]

PR #6121, opened 2026-03-17
feat: Making feast vector store with open ai search api compatible

patelchaitany:

This PR making the feast vector store api with open ai search api compatible so.

This are the changes are made.

1. **New OpenAI-compatible endpoint**

- Added `POST /v1/vector_stores/{vector_store_id}/search` that matches the OpenAI vector store search API
- `vector_store_id` just maps to a feature view name
- Takes a query string, embeds it via LiteLLM, calls `retrieve_online_documents_v2`, and returns results in OpenAI's `vector_store.search_results.page` format
- LiteLLM embedding config goes in `feature_store.yaml` under a new `embedding_model` section (model, api_key, api_base, api_version, dimensions

2. **Metadata filtering**

- New `filter_models.py` with two Pydantic models: `ComparisonFilter` (eq, ne, gt, gte, lt, lte, in, nin) and `CompoundFilter` (and/or, nestable)
- Threaded filters through the entire retrieval stack down to each online store
- Each store translates filters into its native query language:
  - Elasticsearch: Query DSL clauses (term, range, terms, bool)
  - Milvus: boolean expressions like `field == 'value'`
  - Postgres: parameterized SQL subqueries with `entity_key IN (SELECT ...)`
  - SQLite: same approach as Postgres, SQLite syntax

3. **Numeric storage fix**

- Without this change, all values are stored as text, so '9' > '100' is true in filters
- New `enable_openai_compatible_store` config flag on every store backend
- When enabled, adds a `value_num` column that stores int, float, double, and bool values natively alongside the existing `value_text

4. **Bug fixes picked up along the way**

- SQLite BM25 search was reading raw `value` instead of `value_text`
- SQLite's `query` param renamed to `embedding` since that's what it actually is
- Added input escaping for Milvus query strings

5. **Tests**

- 160 lines of unit tests for filter models (valid/invalid operators, value types, nested compounds)
- ~320 lines of integration tests covering filtered vector search, filtered text search, OpenAI response shape, and error cases (no embedding config, nonexistent feature view, empty results)

#5615

# Misc

---

<a href="https://app.devin.ai/review/feast-dev/feast/pull/6121" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -326,20 +329,29 @@ def retrieve_online_documents_v2(
...
-                else 0
-            )
+            first_result_values = results[0].get("values") or []
+            num_results = len(first_result_values)
             result_tuples = []
-
+            if requested_features is None:
+                requested_features = []
+            if "distance" not in requested_features:
+                requested_features.append("distance")
```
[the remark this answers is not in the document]
patelchaitany: the request feature is modified after getting response as it is not returning the distance field in the response of the remoteonlinestore

sdk/python/feast/infra/online_stores/sqlite.py
```diff
@@ -585,29 +866,33 @@ def retrieve_online_documents_v2(
...
                 ) f
-                left join {table_name} fv
+                left join {_quote_id(table_name)} fv
                     on f.rowid = fv.rowid
-                left join {table_name} fv2
+                left join {_quote_id(table_name)} fv2
                     on fv.entity_key = fv2.entity_key
-                where fv2.feature_name != "{vector_field}"
+                where {where_sql}
                 """,
```
jyejare, line 798: The vector_field name is interpolated directly into SQL strings via f-strings (e.g., f'fv2.feature_name != "{vector_field}"'). While this was pre-existing, the PR perpetuates it. The _quote_id helper was added but not applied here.

sdk/python/feast/infra/passthrough_provider.py
```diff
@@ -326,19 +327,21 @@ def retrieve_online_documents_v2(
...
-                requested_features,
-                query,
-                top_k,
-                distance_metric,
-                query_string,
-                include_feature_view_version_metadata,
+                config=config,
+                table=table,
+                requested_features=requested_features,
+                embedding=query,
```
jyejare: The base class OnlineStore.retrieve_online_documents_v2 and Provider.retrieve_online_documents_v2 use query as the parameter name, while the concrete implementations (Postgres, ES, Milvus, SQLite) now use embedding. The PassthroughProvider bridges this with embedding=query (line 339), but this change creates a fragile contract.

If anyone calls a store directly with a positional argument, it works, but the naming mismatch between the abstract interface and implementations is confusing and could break if the abstract method adds keyword-only enforcement.

sdk/python/feast/infra/online_stores/postgres_online_store/postgres.py
```diff
@@ -105,57 +155,123 @@ def online_write_batch(
...
+                    VALUES (%s, %s, %s, %s, %s, %s, %s)
+                    ON CONFLICT (entity_key, feature_name) DO
+                    UPDATE SET
+                        value = EXCLUDED.value,
+                        value_text = EXCLUDED.value_text,
+                        vector_value = EXCLUDED.vector_value,
+                        event_ts = EXCLUDED.event_ts,
+                        created_ts = EXCLUDED.created_ts;
+                """
+                ).format(sql.Identifier(table_name))
```
jyejare: The Postgres online_write_batch has a similarly duplicated pattern -- two different tuple shapes for insert_values (with/without value_num) and two different SQL queries. The tuple construction itself is duplicated across the has_value_num_col branches.

sdk/python/feast/infra/online_stores/elasticsearch_online_store/elasticsearch.py
```diff
@@ -45,10 +46,56 @@ class ElasticSearchOnlineStoreConfig(FeastConfigBaseModel, VectorStoreConfig):
...
+                .get("dynamic_templates", [])
+            )
+            for tmpl in templates:
+                for _, tmpl_body in tmpl.items():
+                    props = tmpl_body.get("mapping", {}).get("properties", {})
+                    if "value_num" in props:
+                        self._index_value_num_cache[index_name] = True
+                        return True
+        except Exception:
+            pass
```
jyejare: swallows all exceptions (network errors, authentication failures, etc.) when checking the index mapping. It should catch a more specific exception type or at least log a warning.

sdk/python/feast/infra/online_stores/sqlite.py
```diff
@@ -835,13 +1161,14 @@ def _generate_bm25_search_insert_query(
     """
     _string_fields = ", ".join(string_field_list)
     query = f"INSERT INTO search_table (entity_key, fv_rowid, {_string_fields})\nSELECT\n\tDISTINCT fv0.entity_key,\n\tfv0.rowid as fv_rowid"
-    from_query = f"\nFROM (select rowid, * from {table_name} where feature_name = '{string_field_list[0]}') fv0"
+    quoted_table = _quote_id(table_name)
+    from_query = f"\nFROM (select rowid, * from {quoted_table} where feature_name = '{string_field_list[0]}') fv0"
 
     for i, string_field in enumerate(string_field_list):
-        query += f"\n\t,fv{i}.value as {string_field}"
+        query += f"\n\t,fv{i}.value_text as {string_field}"
```
jyejare: n the original code, value_text wasn't written for non-vector stores. The PR adds value_text to the table schema and writes, but only when the online_store.vector_enabled or enable_openai_compatible_store is set. The BM25 text search in SQLite now reads fv{i}.value_text instead of fv{i}.value.

This is a breaking change for existing SQLite users who have BM25 search enabled but haven't re-applied. Previously it read binary value column (which somehow worked), now it reads value_text which may be NULL for old data.

patelchaitany: It worked as SQlite is actively casting those binary blobs to strings.

patelchaitany:
Hey @franciscojavierarceo, @ntkathole can you pls review this PR.

sdk/python/feast/infra/online_stores/postgres_online_store/postgres.py
```diff
@@ -678,25 +931,36 @@ def retrieve_online_documents_v2(
...
                     distance_metric_sql=sql.SQL(distance_metric_sql),
                     table_name=sql.Identifier(table_name),
+                    outer_where=outer_where,
                     top_k=sql.Literal(top_k),
                 )
-                params = (embedding, requested_features)
+                base_params = [embedding, requested_features]
+                if has_filters:
+                    base_params.extend(filter_params)
+                params = tuple(base_params)
```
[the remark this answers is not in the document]
patelchaitany: I know about this and why i chose this way because when we do meta data filtering first then we need to retrieve more than top K results then after we need to apply vector similarity search on that which is going to do filter of top K but the thing is that in that process we might loose the closest search according to the vector similarity.

and in the Elastic search and Milvus the metadata filtering happens during the vector search so it is combined step over there not like here which is step by step process and this is limitation of the SQL and POSTGRES.

patelchaitany:
Hi @HaoXuAI, @shuchu when you get a chance, could you take a look at this PR? Happy to address any feedback. Thanks!

sdk/python/feast/feature_store.py
```diff
@@ -2954,9 +2961,181 @@ def retrieve_online_documents_v2(
...
+                "Add an 'embedding_model' section with at least a 'model' "
+                "field to use retrieve_online_documents_openai.\n"
+                "Example:\n"
+                "  embedding_model:\n"
+                "    model: text-embedding-3-small\n"
+                "    api_key: sk-..."
+            )
+
+        try:
+            from litellm import embedding as litellm_embedding
```
ntkathole: This has to be properly abstracted, allow user to pass EmbeddingProvider

Define it like:
```
class EmbeddingProvider(Protocol):
    def embed(self, texts: List[str]) -> List[List[float]]: ...
    async def aembed(self, texts: List[str]) -> List[List[float]]: ...
```

sdk/python/feast/infra/online_stores/elasticsearch_online_store/elasticsearch.py
```diff
@@ -339,6 +400,95 @@ def retrieve_online_documents(
                 )
         return result
 
+    def _translate_filters(
```
ntkathole: Filter translation is duplicated across all stores. 

Again can be defined with:

```
class FilterTranslator(ABC):
    @abstractmethod
    def translate(self, filters: FilterType) -> Any:
        """Return the backend-native filter expression."""
        ...
```

sdk/python/feast/feature_store.py
patelchaitany: I've created EmbeddingProvider class in the embedder.py  which do the runtime checks whether method have same function name but it dose not check the function argument type or return type so for that we need to rely on the CI test to catch the difference in the signature.

sdk/python/feast/infra/online_stores/elasticsearch_online_store/elasticsearch.py
patelchaitany: I've defined this base abstract class which removed the repeated pattern, and each Online Store Implement it's translate filter to query logic.

sdk/python/feast/feature_store.py
```diff
@@ -2954,9 +2994,148 @@ def retrieve_online_documents_v2(
...
+        (when ``query_embedding_model`` is configured in feature_store.yaml),
+        and returns results in OpenAI's ``vector_store.search_results.page``
+        format.
+
+        Args:
+            vector_store_id: Feature view name (maps to the OpenAI
+                ``vector_store_id`` path parameter).
+            query: Natural language query string, or list of strings.
+            max_num_results: Maximum number of results to return.
+            filters: OpenAI-compatible filters (accepted but not yet
```
ntkathole: filters are accepted and applied as well

sdk/python/feast/feature_server.py
```diff
@@ -110,7 +112,42 @@ class GetOnlineDocumentsRequest(BaseModel):
...
+    content_field: Optional[str] = None
+
+
+class OpenAIComparisonFilter(BaseModel):
+    key: str
+    type: str
+    value: Union[str, int, float, bool, List[Union[str, int]]]
+
+
+class OpenAICompoundFilter(BaseModel):
```
ntkathole: I think these all are not needed now, it's duplicate of filter_models.py definitions

sdk/python/feast/embedder.py
```diff
@@ -1,10 +1,124 @@
...
+    def embed(self, texts: List[str]) -> List[List[float]]:
+        """Return one embedding vector per input text."""
+        ...
+
+    async def aembed(self, texts: List[str]) -> List[List[float]]:
+        """Async variant of :meth:`embed`."""
+        ...
+
+
+class LiteLLMEmbeddingProvider:
```
ntkathole:
```suggestion
class LiteLLMEmbeddingProvider(EmbeddingProvider):
```

sdk/python/feast/feature_server.py
```diff
@@ -444,6 +485,51 @@ async def retrieve_online_documents(
...
+    )
+    async def openai_vector_store_search(
+        vector_store_id: str,
+        request: OpenAISearchRequest,
+    ) -> JSONResponse:
+        with feast_metrics.track_request_latency(
+            "/v1/vector_stores/{vector_store_id}/search"
+        ):
+            try:
+                result = await run_in_threadpool(
```
ntkathole: The entire call including the embedding network I/O is wrapped in `run_in_threadpool`, which blocks a thread waiting for the OpenAI/Ollama HTTP response. LiteLLMEmbeddingProvider.aembed() exists precisely to avoid this. Make retrieve_online_documents_openai async, call await self.embedding_provider.aembed(...) inside it, and then await run_in_threadpool only for the DB retrieval step. The endpoint can then await store.retrieve_online_documents_openai(...) directly without an outer run_in_threadpool.

sdk/python/feast/feature_server.py
```diff
@@ -444,6 +485,51 @@ async def retrieve_online_documents(
...
+                    lambda: store.retrieve_online_documents_openai(
+                        vector_store_id=vector_store_id,
+                        query=request.query,
+                        max_num_results=request.max_num_results or 10,
+                        filters=(
+                            request.filters.model_dump() if request.filters else None
```
ntkathole: request.filters.model_dump() can be removed once line 146 uses ComparisonFilter/CompoundFilter directly

infra/website/docs/blog/feast-openai-compatible-api.md
```diff
@@ -0,0 +1,344 @@
+---
+title: "Making Feast Speak OpenAI: Vector Search Without the Glue Code"
+description: "Feast now exposes an OpenAI-compatible vector store search endpoint. Send a plain text query, get results back in the standard OpenAI format. No client-side embeddings required."
+date: 2026-04-28
```
ntkathole: update date before merge

ntkathole:
Also add documentation for OpenAI-compatible Vector search endpoint in docs/reference/alpha-vector-database.md and docs/reference/feature-servers/python-feature-server.md

Also, missing docs for how user can use different embedding provider other than litellm

sdk/python/feast/feature_server.py
patelchaitany: Done. retrieve_online_documents_openai is now async with await aembed(...). The retrieve_online_documents_v2 call is still in run_in_threadpool since it has no async variant.

ntkathole (approved):
looks good to me!

@franciscojavierarceo will you be able to take a look once?

patelchaitany:
@ntkathole, The CI for the [mcp-feature-server-runtime](https://github.com/feast-dev/feast/actions/runs/26014391698/job/76461342471#logs) was failing because fastapi_mcp does not handle circular references properly. To fix this, I made changes in mcp_server.py where we inserted a custom schema resolver that handles how reference schemas are resolved. I also verified its logic against the original fastapi_mcp logic -- both produce the same output when there are no self-referencing schemas.

patelchaitany:
@franciscojavierarceo, can you please review it.

franciscojavierarceo (approved):

franciscojavierarceo (commented):
can you generate an RFC out of this please? 🙏

infra/website/docs/blog/feast-openai-compatible-api.md
```diff
@@ -0,0 +1,344 @@
...
+authors: ["Chaitanya Patel", "Nikhil Kathole"]
+---
+
+<div class="hero-image">
+  <img src="/images/blog/feast-openai-compat-flow.png" alt="Sequence diagram showing a client sending a text query to Feast, which embeds and searches server-side" loading="lazy">
+</div>
+
+If you've tried to connect an AI agent to Feast's vector search, you've probably hit this wall: the agent needs to search your feature store, but Feast expects a raw embedding vector. The agent doesn't have one. It has a question in English.
+
+Until now, the workaround was ugly. You'd call an embedding provider (OpenAI, Ollama, whatever) to turn the text into a float array, then pass that array to Feast's `retrieve-online-documents` endpoint. Every client had to know both APIs, carry both sets of credentials, and run glue code whose only job was bridging the gap.
```
franciscojavierarceo: we should deprecate the `retrieve-online-documents` API and rename it as `search` API to highlight the more generic primitive we want to support, which will offer a lot more long term flexibility.

CC @ntkathole

infra/website/docs/blog/feast-openai-compatible-api.md
```diff
@@ -0,0 +1,344 @@
...
+)
+```
+
+No embedding SDK. No raw vectors. The request and response match OpenAI's format, so anything that already talks to OpenAI can talk to Feast.
+
+### What happens under the hood
+
+When Feast receives this request, it:
+
+1. Embeds the query server-side using the model configured in `feature_store.yaml` (via [LiteLLM](https://docs.litellm.ai/), which supports OpenAI, Ollama, Azure, Cohere, HuggingFace, and 100+ other providers).
```
franciscojavierarceo: yeah this seems acceptable if it's opt-in but we should also support sentence transformers.

infra/website/docs/blog/feast-openai-compatible-api.md
patelchaitany: Okay I have added the new endpoint /search and deprecated the retrieve-online-documents as it is still available for the backward compatibility.

patelchaitany:
@franciscojavierarceo, I have created an RFC as you requested [Docs](https://docs.google.com/document/d/1oWHO7DJKSxyzilvkelcxiDi2vo452I_YY7M_tIt5ZlY/)

patelchaitany:
@franciscojavierarceo, let me know if you need any more changes, or if this is good to go.

sdk/python/feast/feature_server.py
```diff
@@ -495,6 +518,72 @@ async def retrieve_online_documents(
...
+    )
+    async def vector_store_search(
+        vector_store_id: str,
+        request: OpenAISearchRequest,
+    ) -> JSONResponse:
+        with feast_metrics.track_request_latency(
+            "/v1/vector_stores/{vector_store_id}/search"
+        ):
+            try:
+                result = await store.retrieve_online_documents_openai(
```
franciscojavierarceo: i think there's a bug here: this new endpoint bypasses `_get_features`, so auth-enabled servers never run the existing `READ_ONLINE` check for the target feature view. `/search` goes through `_get_features()` before retrieval, but this path calls `store.retrieve_online_documents_openai(...)` directly. Can we resolve the feature view, assert `AuthzedAction.READ_ONLINE`, and add an auth regression test for this endpoint?

sdk/python/feast/feature_store.py
```diff
@@ -3774,9 +3816,154 @@ def retrieve_online_documents_v2(
...
+            num_rows = len(response_dict.get(first_key, []))
+            for i in range(num_rows):
+                score = 0.0
+                attributes: Dict[str, Any] = {}
+                content_parts: List[Dict[str, str]] = []
+
+                for key, values in response_dict.items():
+                    val = values[i] if i < len(values) else None
+                    if key == "distance":
+                        score = float(val) if val is not None else 0.0
```
franciscojavierarceo: i think this is returning the wrong OpenAI semantics. `distance` is lower-is-better for the default pgvector L2 path, but the OpenAI response field is `score`, which clients treat as higher-is-better relevance and use with `score_threshold`. We should either normalize per metric before putting it in `score`, or avoid claiming OpenAI score semantics here.

sdk/python/feast/feature_server.py
```diff
@@ -495,6 +518,72 @@ async def retrieve_online_documents(
...
+                    vector_store_id=vector_store_id,
+                    query=request.query,
+                    max_num_results=request.max_num_results or 10,
+                    filters=request.filters,
+                    ranking_options=(
```
franciscojavierarceo, line 619: This accepts and forwards `ranking_options`, including `score_threshold`, but `retrieve_online_documents_openai()` never applies it. That makes clients think thresholding is enforced when it is silently ignored. Can we either implement `score_threshold` now or reject unsupported `ranking_options`/`rewrite_query` with a 4xx until they are wired?

sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -464,20 +467,29 @@ def retrieve_online_documents_v2(
...
```
franciscojavierarceo, line 494: This still mutates the caller-owned `requested_features` list. When this is called through `FeatureStore._retrieve_from_online_store_v2`, Feast later builds `features_to_request = requested_features + ["distance"]`, so `distance` can appear twice in the response metadata. Please copy before appending, or keep synthetic fields owned by one layer only.

sdk/python/feast/feature_store.py
```diff
@@ -3774,9 +3816,154 @@ def retrieve_online_documents_v2(
...
+                    max_num_results=3,
+                    features_to_retrieve=["name", "description"],
+                )
+        """
+        feature_view = self.get_feature_view(vector_store_id)
+
+        if features_to_retrieve:
+            feature_names = features_to_retrieve
+        else:
+            feature_names = [f.name for f in feature_view.features]
```
franciscojavierarceo: By default this requests every feature from the feature view, which includes the embedding/vector field itself. The response conversion below then copies every non-entity, non-distance field into `attributes`, so OpenAI-compatible search responses can include the raw embedding vector as metadata. Can we exclude vector fields by default and only return scalar metadata/content fields unless explicitly requested?

docs/reference/alpha-vector-database.md
```diff
@@ -32,6 +32,109 @@ backwards compatibility and the adopt industry standard naming conventions.
...
+{
+  "type": "and",
+  "filters": [
+    {"type": "eq", "key": "category", "value": "Electronics"},
+    {"type": "gte", "key": "rating", "value": 4.5}
+  ]
+}
+```
+
+String equality filters work on all backends. Numeric and boolean filters require `enable_openai_compatible_store: true` in the online store config.
```
franciscojavierarceo: This doc says string equality filters work on all backends without the new store flag, but the Postgres and SQLite implementations reject any filter when `enable_openai_compatible_store` is false. Either the runtime should allow text-only filters on existing schemas, or the docs should say all filtering requires the flag for those backends.

franciscojavierarceo (changes requested):
Thanks for pushing this through. I think we need another pass before merge. The main blockers are that the new OpenAI-compatible endpoint bypasses the existing online-read permission path, exposes Feast distances as OpenAI scores, and accepts OpenAI-shaped request fields that are not actually applied. I also left one existing remote-store mutation issue because it still reproduces on this head.

CI note: the latest `unit-test-python (3.12, macos-14)` check is still failing with two 120s CLI/apply timeouts, so this is not green yet.

sdk/python/feast/feature_server.py
patelchaitany: Okay I will add those test as well as the authentication checks for this newly added api endpoints

sdk/python/feast/feature_store.py
patelchaitany: Sure, I remove the vector field from the feature_names

sdk/python/feast/feature_server.py
patelchaitany: Auth regression test is out of scope for this PR, will address in a follow-up

sdk/python/feast/feature_store.py
patelchaitany: Added _distance_to_score(distance, metric) in utils.py and applied it at the store level in Postgres, SQLite, and Milvus. Elasticsearch and Qdrant already return higher-is-better scores, so no changes needed there

infra/website/docs/blog/feast-openai-compatible-api.md
```diff
@@ -0,0 +1,346 @@
...
+
+<div class="hero-image">
+  <img src="/images/blog/feast-openai-compat-flow.png" alt="Sequence diagram showing a client sending a text query to Feast, which embeds and searches server-side" loading="lazy">
+</div>
+
+If you've tried to connect an AI agent to Feast's vector search, you've probably hit this wall: the agent needs to search your feature store, but Feast expects a raw embedding vector. The agent doesn't have one. It has a question in English.
+
+Until now, the workaround was ugly. You'd call an embedding provider (OpenAI, Ollama, whatever) to turn the text into a float array, then pass that array to Feast's vector search endpoint (`POST /search`, formerly `retrieve-online-documents`). Every client had to know both APIs, carry both sets of credentials, and run glue code whose only job was bridging the gap.
+
+Feast now has a new endpoint: `POST /v1/vector_stores/{feature_view}/search`. It follows the [OpenAI Vector Store Search API](https://platform.openai.com/docs/api-reference/vector-stores-search) format. You send text, Feast handles the embedding internally, and you get results back in the same JSON shape that OpenAI returns. No float arrays, no extra SDK.
```
franciscojavierarceo: but this isn't openai compatible.

infra/website/docs/blog/feast-openai-compatible-api.md
```diff
@@ -0,0 +1,346 @@
...
+
+## One endpoint, standard format
+
+With the new endpoint, that same search looks like this:
+
+```python
+import requests
+
+result = requests.post(
+    "http://feast-server:6566/v1/vector_stores/product_catalog/search",
```
franciscojavierarceo: the `product_catalog` is not vector store compatible. it should be some `vs_{hash}` identifier.

sdk/python/feast/infra/online_stores/postgres_online_store/postgres.py
```diff
@@ -811,7 +1044,9 @@ def retrieve_online_documents_v2(
                         "vector_distance"
                     ] != float("inf"):
                         dist_val = ValueProto()
-                        dist_val.double_val = entity_data["vector_distance"]
+                        dist_val.double_val = _distance_to_score(
```
franciscojavierarceo: i think this fix landed in the wrong layer. `retrieve_online_documents_v2()` and `/search` still expose this synthetic field as `distance`, but Postgres now stores a higher-is-better OpenAI score in it. That changes the native Feast contract for non-OpenAI callers, and the same pattern appears in SQLite/Milvus. Can we keep the stores returning raw distances and convert to OpenAI `score` only in `retrieve_online_documents_openai()`?

sdk/python/feast/feature_store.py
```diff
@@ -3887,9 +3929,172 @@ def retrieve_online_documents_v2(
...
+                    features_to_retrieve=["name", "description"],
+                )
+        """
+        unsupported: List[str] = []
+        if ranking_options:
+            if ranking_options.get("score_threshold") is not None:
+                unsupported.append("ranking_options.score_threshold")
+            if ranking_options.get("ranker") is not None:
+                unsupported.append("ranking_options.ranker")
+        if rewrite_query is not None:
```
franciscojavierarceo: this rejects `rewrite_query: false`, even though that is the no-op/default shape an OpenAI-compatible client can send. OGX keeps this field for API compatibility and ignores false at the store layer. Can we only reject `rewrite_query=True` until query rewriting is implemented?

docs/reference/alpha-vector-database.md
```diff
@@ -32,6 +32,109 @@ backwards compatibility and the adopt industry standard naming conventions.
...
+```
+
+### Request fields
+
+| Field | Type | Default | Description |
+|-------|------|---------|-------------|
+| `query` | `string` or `list[string]` | (required) | Plain text search query. Lists are joined with spaces before embedding. |
+| `max_num_results` | `int` | `10` | Maximum number of results to return. |
+| `filters` | `object` | `null` | OpenAI-style filters (see below). |
+| `ranking_options` | `object` | `null` | Accepted but not yet applied. |
```
franciscojavierarceo: the docs now say `ranking_options` and `rewrite_query` are accepted but not applied, but the implementation raises 422 when `score_threshold`, `ranker`, or any `rewrite_query` value is present. Can we make the docs match the actual behavior, or wire the supported no-op/default cases?

franciscojavierarceo (changes requested):
Thanks for the update. CI is green now and some of the earlier concerns are fixed, but I still don't think this is ready to merge. The main blocker is that the OpenAI score fix is currently changing the existing Feast `distance` contract for native `/search` / `retrieve_online_documents_v2` callers. I also left notes on the remaining OpenAI-compat/doc mismatches.

infra/website/docs/blog/feast-openai-compatible-api.md
patelchaitany: I initially went with feature view names for simplicity from implementation and user perspective but you're right it should follow the OpenAI vs_{hash} format for proper compatibility.

it is an easy addition, plan is to hash the  project:feature_view_name and cache it at server startup time and it refresh at every registry TTL,  Will also add GET /v1/vector_stores so user can discover the vs_ IDs for their feature views

working on it

patelchaitany: Now added this endpoint Give it another look when you get time

docs/reference/alpha-vector-database.md
```diff
@@ -32,6 +32,130 @@ backwards compatibility and the adopt industry standard naming conventions.
...
+
+### Requirements
+
+- An `embedding_model` section in your `feature_store.yaml`. Two providers are supported:
+
+  **LiteLLM** (default) — API-backed models via [LiteLLM](https://docs.litellm.ai/) (OpenAI, Azure, Ollama, Cohere, etc.):
+
+  ```yaml
+  embedding_model:
+    provider: litellm          # default; can be omitted
```
franciscojavierarceo: i'm a little worried about requiring litellm for embeddings, why wouldn't we embed within feast? in fact, i think this is a good usage of this. even though it's not as efficient as using a GPU, it's good enough. 

CC @ntkathole

infra/website/docs/blog/feast-openai-compatible-api.md
```diff
@@ -0,0 +1,382 @@
+---
+title: "Making Feast Speak OpenAI: Vector Search Without the Glue Code"
```
franciscojavierarceo:
```suggestion
title: "Using Feast's OpenAI Compatible Search API"
```

are we limiting this only to vector search? or are we supporting hybrid as well?

docs/reference/alpha-vector-database.md
patelchaitany: We want to support wide range of the embedding model and using litellm for embedding looks feasible option to me but in feast we support the sentence-transformers for doing it entirely in the feast.

infra/website/docs/blog/feast-openai-compatible-api.md
patelchaitany: Currently we are doing only the vector search

docs/reference/alpha-vector-database.md
franciscojavierarceo: yeah so then let's not add litellm to scope

franciscojavierarceo (approved):
we can address this in a follow up PR but right now the diagram includes LiteLLM still but we took that out, so we should remove that from the diagram but let's get this merged.

franciscojavierarceo:
@patelchaitany can you also document this feature as an alpha API in the docs?

patelchaitany:
@franciscojavierarceo, I had question regarding adding the /search endpoint should we Include the retrieve_online_documents_openai() under that as well or should it remain only accessible through this endpoint 
`/v1/vector_stores/{vector_store_id}/search`

franciscojavierarceo:
>@franciscojavierarceo, I had question regarding adding the /search endpoint should we Include the retrieve_online_documents_openai() under that as well or should it remain only accessible through this endpoint

We should just call it `search()` or `search_online_store()` or `openai_search()`. 

I like `search()` as the public API that's non-OpenAI and then we could have a `search()` method and `openai_search()`. method.

patelchaitany:
> > @franciscojavierarceo, I had question regarding adding the /search endpoint should we Include the retrieve_online_documents_openai() under that as well or should it remain only accessible through this endpoint
> 
> We should just call it `search()` or `search_online_store()` or `openai_search()`.
> 
> I like `search()` as the public API that's non-OpenAI and then we could have a `search()` method and `openai_search()`. method.

Thanks @franciscojavierarceo ,so it's a change for the name, but there is not any public method for search() in the Python SDK yet (search is available as an API endpoint). If renaming the method / introducing something like search could cost more changes, so I could rename `retrieve_online_documents_openai` to `openai_search` as that is introduced in this change. But about the search API, I could simply wrap it in a wrapper, but couldn't change it completely as it might be a big change.

franciscojavierarceo:
yeah just rename to openai_search sounds good to me

sdk/python/feast/infra/online_stores/milvus_online_store/milvus.py
```diff
@@ -96,6 +104,87 @@
...
+    DataType.BOOL,
+}
+
+
+def _milvus_escape_string(s: str) -> str:
+    """Escape a string for safe use inside a Milvus single-quoted literal.
+
+    Backslashes must be escaped first; otherwise a trailing backslash in the
+    input would combine with the escaped quote to break out of the literal.
+    """
```
jyejare, line 125: **[Critical] SQL Injection vulnerability in Milvus string escaping**

The Milvus string escaping function only escapes backslashes and single quotes, but doesn't handle other potential injection vectors. Milvus boolean expressions could be vulnerable to injection if field names or other components aren't properly validated.

**Suggested:**
```suggestion
def _milvus_escape_string(s: str) -> str:
    """Escape a string for safe use inside a Milvus single-quoted literal."""
    # More comprehensive escaping for Milvus
    return s.replace("\\", "\\\\").replace("'", "\\'")
        .replace('"', '\\"').replace('\n', '\\n').replace('\r', '\\r')
```

jyejare (commented):
This PR adds OpenAI-compatible vector store search endpoints to Feast, enabling text-based search with server-side embedding. The implementation is comprehensive with good documentation and multi-backend support. However, there are several security concerns around input validation and potential injection attacks that need to be addressed before merging.

patelchaitany:
@franciscojavierarceo let me know if any changes are required.

franciscojavierarceo (approved):

PR #6146, opened 2026-03-24
fix: support absolute paths in feast init CLI argument

### LINKED ISSUE #6134: feast init /tmp/test # does not work
## Expected Behavior 
Just work (tm).

## Current Behavior
Error: Invalid value for PROJECT_DIRECTORY: Name should be alphanumeric values, underscores, and hyphens but not start with an underscore or hyphen

## Steps to reproduce
feast init /tmp/test

### Specifications
Linux
Python 3.11.14
feast 0.61.0

## Possible Solution
Support absolute paths.

joaquinhuigomez:
## Summary

`feast init /tmp/test` fails because the entire path is validated as a project name. When the argument contains a path separator, the CLI now uses the full path as the repo directory and extracts the basename as the project name.

Fixes #6134

### DIFF
--- a/sdk/python/feast/cli/cli.py
+++ b/sdk/python/feast/cli/cli.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 import json
 import logging
+import os
 from datetime import datetime
 from importlib.metadata import version as importlib_version
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


ntkathole:
@joaquinhuigomez wondering if you have seen comment in issue https://github.com/feast-dev/feast/issues/6134#issuecomment-4096918475 ?

joaquinhuigomez:
Yeah, --repo-path works. This just makes the bare `feast init /tmp/test` path work the same way without needing the flag — seemed like a natural UX expectation.

franciscojavierarceo:
thanks for this @joaquinhuigomez yeah by convention feast has always used feast init as the project name so probably we will keep it as is. thank you for opening this up though and definitely appreciate you taking the time on it.

joaquinhuigomez:
Got it, makes sense. Thanks for taking a look.

PR #6235, opened 2026-04-08
feat: Feast-MLflow Integration

Vperiodt:

https://github.com/user-attachments/assets/684adb69-c5b1-466b-8268-dc4507999ea4

Native MLflow integration for Feast that provides automatic feature lineage tracking alongside ML experiments. When enabled via feature_store.yaml, every feature retrieval is logged to the active MLflow run.

# Key Capabilities
1. `Auto-logging: ` Feature retrieval metadata is tagged on the active MLflow run (feast.feature_refs, feast.feature_views, feast.feature_service, feast.entity_count, etc.)
2. `Entity DataFrame archival:` Optionally saves the training entity DataFrame as an MLflow artifact (entity_df.parquet) for full reproducibility
3. `Model-to-feature-service resolution:` store.mlflow.resolve_features() maps any MLflow model URI back to its Feast feature service, enabling serving pipelines to auto-discover which features a model needs
4. `Entity DataFrame reconstruction:` store.mlflow.get_training_entity_df() rebuilds the exact entity DataFrame from a past run's artifacts, enabling training reproducibility
5. `Model lifecycle lineage:` log_model auto-attaches feast_features.json, register_model propagates feast.feature_service to model versions, load_model links prediction runs to training runs
6. `Operations audit trail:` feast apply and feast materialize logged to a dedicated {project}-feast-ops experiment
7. `Dataset tracking:` log_training_dataset() logs training DataFrames as MLflow dataset inputs
8. `Two access patterns: `store.mlflow (explicit, multi-store safe) and feast.mlflow (drop-in replacement for import mlflow)
9. `Configuration: `Controlled entirely via feature_store.yaml under a new mlflow: block
# Demo
Full end-to-end demo notebook with all capabilities: [integration-example](https://github.com/Vperiodt/integration-example)

# Checks

## Testing Strategy
- [ ] Unit tests
- [ ] Integration tests
- [x] Manual tests

# Misc

---

<a href="https://app.devin.ai/review/feast-dev/feast/pull/6235" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


docs/reference/mlflow.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Overview
+
+When enabled, the integration provides:
+
+- **Historical feature retrieval** -- `get_historical_features()` tags the run with feature refs, feature views, entity count, and retrieval duration
+- **Online feature retrieval** -- `get_online_features()` tags the run with the same metadata
+- **Entity DataFrame archival** -- optionally saves the training entity DataFrame as an MLflow artifact for full reproducibility
+- **Execution context tagging** -- tags runs with where they ran (workbench, KFP pipeline, feature server, or standalone)
+- **Operation logging** -- optionally logs `feast apply` and `feast materialize` to a separate MLflow experiment
```
franciscojavierarceo: not sure this is actually useful

docs/reference/mlflow.md
```diff
@@ -0,0 +1,208 @@
...
+- **Model-to-Feature resolution** -- map any MLflow model URI back to its Feast feature service
```
franciscojavierarceo: nice

docs/reference/mlflow.md
```diff
@@ -0,0 +1,208 @@
...
+- **Training reproducibility** -- reconstruct the exact entity DataFrame from a past MLflow run
```
franciscojavierarceo: 👏

docs/reference/mlflow.md
```diff
@@ -0,0 +1,208 @@
...
+- **Training-to-prediction linkage** -- `FeastMlflowClient.load_model()` links prediction runs back to their training runs
```
franciscojavierarceo: very nice

docs/reference/mlflow.md
```diff
@@ -0,0 +1,208 @@
...
+- **Feast MLflow Client** -- a thin wrapper that eliminates direct `import mlflow` in user code
```
franciscojavierarceo: makes sense

sdk/python/feast/mlflow_integration/logger.py
```diff
@@ -0,0 +1,282 @@
...
+        _logger.debug("%s (failures=%d): %s", msg, _consecutive_failures, exc)
+
+
+def _report_success() -> None:
+    """Reset failure counter on successful logging."""
+    global _consecutive_failures
+    _consecutive_failures = 0
+
+
+def log_feature_retrieval_to_mlflow(
```
franciscojavierarceo, line 74: i guess if we have this enabled why do we need to even have the wrapper? can't we full handle this for the user?

franciscojavierarceo: I guess my question is, can't we just hide all of the mlflow client usage to the user?

docs/reference/mlflow.md
Vperiodt: actually, this is not updated one, thought of this one for backtracking from mllfow run to the workbench experiment but now it is omitted

sdk/python/feast/mlflow_integration/logger.py
Vperiodt: the auto-logging in `logger.py` fully handled for users who already use `mlflow.start_run()` or `client.start_run()` considering our FeastMlflowClient wrapper which is for users who want to avoid import mlflow entirely and get additional Feast-specific behavior on top ex: logging a model, registering it, loading it for inference etc so` FeastMlflowClient ` is for users who want the extra lineage features without touching MLflow directly

docs/reference/mlflow.md
```diff
@@ -0,0 +1,284 @@
...
+
+### Tracking URI resolution
+
+The tracking URI is resolved in this order:
+
+1. `tracking_uri` field in `feature_store.yaml`
+2. `MLFLOW_TRACKING_URI` environment variable
+3. MLflow's default (`./mlruns` local directory)
+
+This means you can omit `tracking_uri` from the YAML and set `MLFLOW_TRACKING_URI` in your environment instead.
```
jyejare:
```suggestion
This means you can omit `tracking_uri` from the YAML and set `MLFLOW_TRACKING_URI` in your environment instead or it would be pulled from /.mlruns automatically when both are not set.
```
wdyt ?

docs/reference/mlflow.md
```diff
@@ -0,0 +1,284 @@
...
+
+| Artifact | Description |
+|----------|-------------|
+| `entity_df.parquet` | Full entity DataFrame used in the retrieval |
+
+When a model is logged via `client.log_model()`:
+
+| Artifact | Description |
+|----------|-------------|
+| `required_features.json` | JSON list of feature references the model was trained on |
```
jyejare: Could we rename to relate it with feast better and so that it would be self describing in MLFLow ecosystem, something like:
```
required_feast_features.json
```
OR
```
feast_features.json
```

docs/reference/mlflow.md
```diff
@@ -0,0 +1,284 @@
...
+| `feast.materialize.duration_sec` | `12.3456` |
+
+## Usage
+
+### Automatic logging 
+
+With the configuration above, feature metadata is logged whenever there is an active MLflow run:
+
+```python
+import mlflow
```
jyejare: I think explicit import is no more required as feast module gives you the ability to:
```
store.mlflow.start_run(....
```

docs/reference/mlflow.md
```diff
@@ -0,0 +1,284 @@
...
+
+### FeastMlflowClient (zero `import mlflow`)
+
+The `FeastMlflowClient` wraps MLflow so user code never needs `import mlflow`. All configuration is inherited from `feature_store.yaml`:
+
+```python
+from feast import FeatureStore
+
+store = FeatureStore(".")
+client = store.get_mlflow_client()
```
jyejare: Even this is changed now. Lets keep only `store.mlflow.` examples.

docs/reference/mlflow.md
```diff
@@ -0,0 +1,324 @@
...
+In most cases, simply creating a `FeatureStore(...)` is enough — no `init()` needed.
+
+#### Error handling
+
+`feast.mlflow` raises clear errors on first use if something is misconfigured:
+
+| Condition | Error |
+|-----------|-------|
+| No `feature_store.yaml` in cwd and no store created | `RuntimeError` with guidance to call `feast.mlflow.init(store)` |
+| `mlflow.enabled` is `false` or missing | `RuntimeError` with guidance to set `mlflow.enabled=true` |
```
jyejare: Even if user wants to set it to false ?

docs/reference/mlflow.md
```diff
@@ -0,0 +1,324 @@
...
+
+### Feast-only functions
+
+These are unique to `feast.mlflow` and have no `mlflow` equivalent:
+
+| Function | Description |
+|----------|-------------|
+| `feast.mlflow.init(store)` | Explicitly bind to a `FeatureStore` (optional) |
+| `feast.mlflow.resolve_features(model_uri)` | Resolve model URI to Feast feature service name |
+| `feast.mlflow.get_training_entity_df(run_id, ...)` | Recover entity DataFrame from a past MLflow run |
```
jyejare: So no need of phase 2 ?

sdk/python/feast/mlflow_integration/entity_df_builder.py
```diff
@@ -0,0 +1,104 @@
...
+    Returns:
+        A ``pd.DataFrame`` suitable for passing to
+        ``store.get_historical_features(entity_df=...)``.
+
+    Raises:
+        FeastMlflowEntityDfError: If mlflow is not installed, run not found,
+            or no entity data is available on the run.
+    """
+    try:
+        import mlflow
```
jyejare: Can we use store object here instead, we are promising to not ise mlflow directly ? :)

sdk/python/feast/mlflow_integration/entity_df_builder.py
```diff
@@ -0,0 +1,104 @@
...
+    )
+
+
+def _try_artifact(client, run_id: str, artifact_name: str, fmt: str):
+    """Try to download and load an artifact as a DataFrame."""
+    try:
+        local_path = client.download_artifacts(run_id, artifact_name)
+        if fmt == "parquet":
+            return pd.read_parquet(local_path)
+        return pd.read_csv(local_path)
```
jyejare: We may have an exception here if the file is not CSV (after checking if not parquet). Can we explitely check `csv` extention, if not warn user about different extention.

sdk/python/feast/mlflow_integration/logger.py
jyejare: I would rather push all these functions to integration client as everything here is a part of mlflow feast integration?

sdk/python/feast/mlflow_integration/logger.py
```diff
@@ -0,0 +1,369 @@
...
+        tracking_uri: MLflow tracking URI.
+        ops_experiment_suffix: Suffix for the ops experiment name.
+        transition_types: Optional mapping of object name to transition
+            type string (``"CREATE"``, ``"UPDATE"``, ``"DELETE"``).
+            When provided, per-transition tags are logged alongside
+            the aggregate ``*_changed`` tags.
+    """
+    mlflow = _get_mlflow()
+    if mlflow is None:
+        return False
```
jyejare: This seems making it very complex to unnecessary initiate lots of mlflow objects in multiple functions.

I would rather push all these function to integration client/wrapper.

jyejare: If you would still want to keep it separate, i would rather create a class for each which takes store object in param ,and the class is initiated in the integration client.

This way we would end up having single source of truth object and dont need to initiate it everywhere.

sdk/python/feast/mlflow_integration/model_resolver.py
```diff
@@ -0,0 +1,150 @@
...
+        tracking_uri: Optional MLflow tracking URI. When not provided,
+            falls back to the MLFLOW_TRACKING_URI environment variable,
+            then to MLflow's built-in default.
+
+    Raises:
+        FeastMlflowModelResolutionError: If mlflow is not installed, URI is
+            invalid, resolution fails, or validation against the store fails.
+    """
+    try:
+        import mlflow
```
jyejare: Same here. Create a class and initiate in integration object/class that takes store object as param.

sdk/python/feast/feature_store.py
```diff
@@ -194,6 +219,177 @@ def __init__(
...
+        This bridges Feast config with MLflow so that user calls like
+        ``mlflow.start_run()`` automatically target the correct tracking
+        server and experiment without any extra setup in training scripts.
+        """
+        try:
+            mlflow_cfg = self.config.mlflow
+            if mlflow_cfg is None or not mlflow_cfg.enabled:
+                return
+
+            import mlflow
```
jyejare: The store object is available to you here as `self`, so why not :
```
self.mlflow.set_tracking_uri(....
```

sdk/python/feast/feature_store.py
```diff
@@ -194,6 +219,175 @@ def __init__(
...
+    def _rebuild_fs_name_index(self) -> None:
+        """Rebuild the {frozenset(refs) → service_name} index from the registry."""
+        index: Dict[frozenset, str] = {}
+        for fs in self.registry.list_feature_services(self.project, allow_cache=True):
+            fs_refs = frozenset(
+                f"{p.name}:{f.name}"
+                for p in fs.feature_view_projections
+                for f in p.features
+            )
+            index[fs_refs] = fs.name
```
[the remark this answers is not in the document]
jyejare: This ^

sdk/python/feast/mlflow.py
```diff
@@ -0,0 +1,141 @@
...
+
+    client = _ensure_client()
+
+    client_attr = getattr(client, name, None)
+    if client_attr is not None:
+        return client_attr
+
+    mlflow_attr = getattr(client._mlflow, name, None)
+    if mlflow_attr is not None:
+        return mlflow_attr
```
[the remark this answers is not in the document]
jyejare: This ^

sdk/python/tests/integration/test_mlflow_integration.py
```diff
@@ -0,0 +1,1181 @@
...
+            tracking_uri=tracking_uri,
+        )
+        assert result is False
+
+    @pytest.mark.integration
+    def test_feature_views_sorted_and_deduped(self, tracking_uri):
+        client = MlflowClient(tracking_uri=tracking_uri)
+        refs = ["z_view:f1", "a_view:f2", "z_view:f3", "a_view:f4"]
+
+        with mlflow.start_run() as run:
```
jyejare, line 259: The tests should also be focusing only on `store.mlflow...` culture.

jyejare (commented):
Comments inline:

1. We also need to enable mlflow configuratins from feast operator else admin has to do it explicitly.
2. Also, we need to remove duplicate mechanisms for initiating/using mlflow client and we should provide a clean way that is `store.mlflow` whenever possible.

sdk/python/feast/mlflow_integration/logger.py
```diff
@@ -0,0 +1,388 @@
...
+        try:
+            from feast import Entity, FeatureService
+            from feast.feature_view import FeatureView
+
+            project = self._store.project
+            mlflow_cfg = self._store.config.mlflow
+            ops_suffix = mlflow_cfg.ops_experiment_suffix
+
+            effective_uri = self._tracking_uri or self._mlflow.get_tracking_uri()
+            client = self._mlflow.MlflowClient(tracking_uri=effective_uri)
```
ntkathole: instead of creating new client everytime, can we initialise once and reuse ?

sdk/python/feast/feature_store.py
```diff
@@ -194,6 +197,127 @@ def __init__(
         # Initialize feature service cache for performance optimization
         self._feature_service_cache = {}
 
+        # Cache for _resolve_feature_service_name lookups
```
ntkathole, line 201: the MLflow auto-log could resolve to a deleted or renamed feature service. `_fs_name_cache` and `_fs_name_index` caches (used by `_resolve_feature_service_name`) are never cleared when `refresh_registry()` or `_clear_feature_service_cache()` is called

sdk/python/feast/feature_store.py
```diff
@@ -194,6 +197,127 @@ def __init__(
...
+                f"{p.name_to_use()}:{f.name}"
+                for p in fs.feature_view_projections
+                for f in p.features
+            )
+            index[fs_refs] = fs.name
+        self._fs_name_index = index
+        self._fs_name_cache = {}
+        self._fs_name_index_ts = time.monotonic()
+
+    def _resolve_feature_service_name(self, feature_refs: List[str]) -> Optional[str]:
```
ntkathole, line 275: can't we use existing `_feature_service_cache` from FeatureStore?

sdk/python/feast/feature_store.py
```diff
@@ -2681,6 +2963,55 @@ async def get_online_features_async(
             include_feature_view_version_metadata=include_feature_view_version_metadata,
         )
 
+        try:
+            if self.mlflow is not None and self.config.mlflow.auto_log:
+                import asyncio
```
ntkathole: already imported at top

sdk/python/feast/feature_store.py
```diff
@@ -194,6 +197,127 @@ def __init__(
         # Initialize feature service cache for performance optimization
         self._feature_service_cache = {}
 
+        # Cache for _resolve_feature_service_name lookups
+        self._fs_name_cache: Dict[frozenset, Optional[str]] = {}
+        self._fs_name_index: Dict[frozenset, str] = {}
+        self._fs_name_index_ts: float = -self._FS_NAME_INDEX_TTL_SECONDS
+
+        self._mlflow_client: Optional[Any] = None
+        self._init_mlflow()
```
ntkathole: I think better to do lazy loading here, first check the config and then load module.

sdk/python/feast/mlflow_integration/client.py
```diff
@@ -0,0 +1,314 @@
...
+        import mlflow as _mlflow_mod
+
+        self._mlflow = _mlflow_mod
+        self._store = store
+        self._tracking_uri = store.config.mlflow.get_tracking_uri()
+        self._client = _mlflow_mod.MlflowClient(tracking_uri=self._tracking_uri)
+        self._default_experiment = store.config.project
+
+        if self._tracking_uri:
+            _mlflow_mod.set_tracking_uri(self._tracking_uri)
```
ntkathole: remove `set_tracking_uri()` from `__init__ `and instead pass `tracking_uri` explicitly when needed (the MlflowClient already receives it)

sdk/python/feast/ui_server.py
```diff
@@ -115,6 +118,266 @@ def health():
             else Response(status_code=status.HTTP_503_SERVICE_UNAVAILABLE)
         )
 
+    @app.get("/api/mlflow-runs")
+    def get_mlflow_runs(max_results: int = 50):
```
ntkathole, line 107: should this be a registry API? Not a blocker for this PR, but architecture wise, registry is source of truth

ntkathole (changes requested):

sdk/python/feast/feature_store.py
Vperiodt: `_feature_service_cache` lookup is : s`ervice_name → [feature_refs]`and is populated lazily on demand
`_fs_name_index` lookup is reverse  `{feature_refs} → service_name`, populated from all registry services needed for subset matching when users pass raw refs instead of a FeatureService

sdk/python/feast/ui_server.py
Vperiodt: Yes, will add this one will open to other benefits too !

franciscojavierarceo:
linter is failing

Vperiodt:
> linter is failing

fixed

pyproject.toml
ntkathole: include mlflow here in ci dependencies

ui/src/pages/feature-views/RegularFeatureViewOverviewTab.tsx
```diff
@@ -59,12 +63,33 @@ const RegularFeatureViewOverviewTab = ({
...
+  const lastUsedLabel =
+    lastUsed != null ? new Date(lastUsed).toLocaleDateString() : "N/A";
+
   return (
     <React.Fragment>
       <EuiFlexGroup>
         <EuiFlexItem>
           <EuiStat title={`${numOfFs}`} description="Consuming Services" />
         </EuiFlexItem>
+        <EuiFlexItem>
```
ntkathole: This should be conditionally rendered, users who don't use MLflow will wonder what these stats mean and why they're there

ui/src/pages/feature-views/FeatureViewUsagePanel.tsx
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  if (isError || !data) {
+    return null;
+  }
+
+  const usage: FeatureUsageEntry | undefined =
+    data.feature_usage?.[featureViewName];
+
+  if (!usage || usage.run_count === 0) {
```
ntkathole, line 69: same here, if mlflow is not configured, this needs to be hidden

ntkathole:
@Vperiodt Guard the UI components behind a check on whether MLflow data is available. The simplest approach is to have the backend return a `mlflow_enabled: false` flag in the API responses (or a new lightweight /api/mlflow-status endpoint), and conditionally render MLflow components only when enabled.

Vperiodt:
> @Vperiodt Guard the UI components behind a check on whether MLflow data is available. The simplest approach is to have the backend return a `mlflow_enabled: false` flag in the API responses (or a new lightweight /api/mlflow-status endpoint), and conditionally render MLflow components only when enabled.

hello ! @ntkathole, Addressed all reviews

sdk/python/feast/ui_server.py
ntkathole: @Vperiodt going to handle this in this PR or different one?

Vperiodt: yess, will address this in a follow up PR

ntkathole:
@Vperiodt need to update pixi.lock file, also please rebase

ntkathole (approved):
Looks good to me! Thank you @Vperiodt , This is great

franciscojavierarceo:
lgtm

PR #6259, opened 2026-04-10
fix: Ambiguous truth value of array during materialization

### LINKED ISSUE #6255: `ValueError: The truth value of an empty array is ambiguous` during materialization
## Expected Behavior

`feast materialize` should complete successfully even when the source DataFrame
contains an **empty numpy array** (`np.array([])`) in a scalar feature column.
The empty array should be treated as a null / missing value and produce an empty
`ProtoValue()`, consistent with how `None` and `np.nan` are already handled.

## Current Behavior

`feast materialize` crashes with:

```
ValueError: The truth value of an empty array is ambiguous.
Use `array.size > 0` to check that an array is not empty.
```

Full stack trace:

```
  File "/opt/app-root/bin/feast", line 10, in <module>
    sys.exit(cli())
             ^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/click/core.py", line 1485, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/click/core.py", line 1406, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/click/core.py", line 1873, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/click/core.py", line 1269, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/click/core.py", line 824, in invoke
    return callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/click/decorators.py", line 34, in new_func
    return f(get_current_context(), *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/feast/cli/cli.py", line 393, in materialize_command
    store.materialize(
  File "/opt/app-root/lib64/python3.11/site-packages/feast/feature_store.py", line 1816, in materialize
    provider.materialize_single_feature_view(
  File "/opt/app-root/lib64/python3.11/site-packages/feast/infra/passthrough_provider.py", line 456, in materialize_single_feature_view
    raise e
  File "/opt/app-root/lib64/python3.11/site-packages/feast/infra/compute_engines/local/compute.py", line 84, in _materialize_one
    plan.execute(context)
  File "/opt/app-root/lib64/python3.11/site-packages/feast/infra/compute_engines/dag/plan.py", line 51, in execute
    output = node.execute(context)
             ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/feast/infra/compute_engines/local/nodes.py", line 274, in execute
    rows_to_write = _convert_arrow_to_proto(
                    ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/feast/utils.py", line 281, in _convert_arrow_to_proto
    return _convert_arrow_fv_to_proto(table, feature_view, join_keys)  # type: ignore[arg-type]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/feast/utils.py", line 298, in _convert_arrow_fv_to_proto
    proto_values_by_column = {
                             ^
  File "/opt/app-root/lib64/python3.11/site-packages/feast/utils.py", line 299, in <dictcomp>
    column: python_values_to_proto_values(
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/feast/type_map.py", line 840, in python_values_to_proto_values
    proto_values = _python_value_to_proto_value(value_type, values)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.11/site-packages/feast/type_map.py", line 772, in _python_value_to_proto_value
    elif not pd.isnull(value):
ValueError: The truth value of an empty array is ambiguous. Use `array.size > 0` to check that an array is not empty.
```

The root cause is in [`sdk/python/feast/type_map.py`](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/type_map.py),
function `_convert_scalar_values_to_proto` (around line 968):

```python
# Generic scalar conversion
out = []
for value in values:
    if isinstance(value, ProtoValue):
        out.append(value)
    elif not pd.isnull(value):          # ← crashes here
        out.append(ProtoValue(**{field_name: func(value)}))
    else:
        out.append(ProtoValue())
```

`pd.isnull()` is **vectorised**: when `value` is a numpy array (including an empty
one), it returns a numpy array of booleans instead of a scalar boolean. Applying
Python's `not` to that array raises `ValueError`. The same pattern exists a few lines
above in the `ValueType.BOOL` path (`if not pd.isnull(value)`).

## Steps to reproduce

```python
import numpy as np
from feast.type_map import python_values_to_proto_values
from feast.value_type import ValueType

# A scalar column where one row contains an empty array
python_values_to_proto_values([np.array([]), 1.0, 2.0], ValueType.DOUBLE)
# → ValueError: The truth value of an empty array is ambiguous
```

### Specifications

- Version: `v0.60.0` (also reproducible on `main` as of 2026-04-10)
- Platform: Linux (Python 3.11), macOS (Python 3.11)
- Subsystem: `feast/type_map.py` – `_convert_scalar_values_to_proto`

## Possible Solution

The fix belongs in [`sdk/python/feast/type_map.py`](https://github.com/feast-dev/feast/blob/85f5cef78/sdk/python/feast/type_map.py), specifically the **generic scalar conversion loop** at [line 963–975](https://github.com/feast-dev/feast/blob/85f5cef78/sdk/python/feast/type_map.py#L963-L975) (`elif not pd.isnull(value)`).

Before calling `not pd.isnull(value)`, check whether the value is array-like.
`pd.isnull()` is vectorised and returns an `np.ndarray` for array inputs, so
calling `not` on it raises `ValueError`. The fix must handle three sub-cases:

| Value | Expected outcome |
|---|---|
| Empty array (`size == 0`) | null → `ProtoValue()` |
| Non-empty array containing any null | null → `ProtoValue()` |
| Non-empty array with all valid data | convert → `ProtoValue(**{field_name: func(value)})` |
| Plain scalar null | null → `ProtoValue()` |
| Plain scalar non-null | convert → `ProtoValue(**{field_name: func(value)})` |

```python
# Generic scalar conversion
out = []
for value in values:
    if isinstance(value, ProtoValue):
        out.append(value)
    elif isinstance(value, np.ndarray) or (
        hasattr(value, "__len__") and not isinstance(value, (str, bytes))
    ):
        # Array-like value in a scalar column
        if hasattr(value, "size") and value.size == 0:
            # Empty numpy array – treat as null
            out.append(ProtoValue())
        else:
            is_null = pd.isnull(value)
            if hasattr(is_null, "any"):
                # pd.isnull returned an array; null if any element is null
                out.append(ProtoValue() if is_null.any() else ProtoValue(**{field_name: func(value)}))
            elif not is_null:
                out.append(ProtoValue(**{field_name: func(value)}))
            else:
                out.append(ProtoValue())
    elif not pd.isnull(value):
        out.append(ProtoValue(**{field_name: func(value)}))
    else:
        out.append(ProtoValue())
return out
```

alan-gauthier-jt:

`feast materialize` crashes with `ValueError: The truth value of an empty array is ambiguous` when a scalar feature column contains an empty numpy array (e.g. `np.array([])`). This is a real-world scenario when a DataFrame row has a missing value represented as an empty array rather than `None` or `np.nan`.

**Root cause:** In `_convert_scalar_values_to_proto` ([`sdk/python/feast/type_map.py`](sdk/python/feast/type_map.py)), the null check uses `not pd.isnull(value)` for every value in the loop. `pd.isnull()` is vectorised — when `value` is a numpy array, it returns a boolean array instead of a scalar. Applying Python's `not` operator to that array raises `ValueError`. The same issue exists in:
- the `BOOL` scalar path (`not pd.isnull(value)` in a list comprehension)
- the `UNIX_TIMESTAMP` early-return path (`_python_datetime_to_int_timestamp(values)` called with the raw values list, including any array-like values)
- the sample type-validation check (`sample == 0`)

**Fix:** Before calling `pd.isnull()`, guard both scalar conversion loops (generic and BOOL) and the sample type-validation with an explicit `isinstance(value, np.ndarray)` check. Any array-like value in a scalar feature column is unmappable to a protobuf scalar field anyway, so it is safely treated as null → `ProtoValue()`.

| Input value | Behaviour before | Behaviour after |
|---|---|---|
| `np.array([])` (empty) | `ValueError` crash | `ProtoValue()` (null) |
| `np.array([np.nan, 1.0])` | `ValueError` crash | `ProtoValue()` (null) |
| `np.array([1.0, 2.0])` | `ValueError` crash | `ProtoValue()` (null) |
| `None` | `ProtoValue()` (null) | unchanged |
| scalar non-null | `ProtoValue(field=value)` | unchanged |

Fixes #6255

# Checks

## Testing Strategy
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

# Misc

---

<a href="https://app.devin.ai/review/feast-dev/feast/pull/6259" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/sdk/python/feast/type_map.py
+++ b/sdk/python/feast/type_map.py
@@ -929,16 +929,35 @@ def _convert_scalar_values_to_proto(
         return [ProtoValue()] * len(values)
 
     if feast_value_type == ValueType.UNIX_TIMESTAMP:
-        int_timestamps = _python_datetime_to_int_timestamp(values)
-        return [ProtoValue(unix_timestamp_val=ts) for ts in int_timestamps]  # type: ignore
+        out = []
+        for value in values:
+            if isinstance(value, np.ndarray) or (
+                hasattr(value, "__len__") and not isinstance(value, (str, bytes))
+            ):
+                # Array-like value in a scalar UNIX_TIMESTAMP column: treat as null.
+                out.append(ProtoValue())
+            elif value is None:
+                out.append(ProtoValue())
+            else:
+                (ts,) = _python_datetime_to_int_timestamp([value])
+                out.append(ProtoValue(unix_timestamp_val=ts))  # type: ignore
+        return out
 
     field_name, func, valid_scalar_types = PYTHON_SCALAR_VALUE_TYPE_TO_PROTO_VALUE[
         feast_value_type
     ]
 
-    # Validate scalar types
-    if valid_scalar_types:
-        if (sample == 0 or sample == 0.0) and feast_value_type != ValueType.BOOL:
+    # Validate scalar types — skip for array-like samples (they will be treated
+    # as null or raw values in the conversion loop below).
+    if valid_scalar_types and not (
+        isinstance(sample, np.ndarray)
+        or (hasattr(sample, "__len__") and not isinstance(sample, (str, bytes)))
+    ):
+        try:
+            is_zero = sample == 0 or sample == 0.0
+        except (ValueError, TypeError):
+            is_zero = False
+        if is_zero and feast_value_type != ValueType.BOOL:
             # Numpy converts 0 to int, but column type may be float
             allowed_types = {np.int64, int, np.float64, float, decimal.Decimal}
             assert type(sample) in allowed_types, (
@@ -951,20 +970,39 @@ def _convert_scalar_values_to_proto(
 
     # Handle BOOL specially due to np.bool_ conversion requirement
     if feast_value_type == ValueType.BOOL:
-        return [
-            ProtoValue(
-                **{field_name: func(bool(value) if type(value) is np.bool_ else value)}
-            )  # type: ignore
-            if not pd.isnull(value)
-            else ProtoValue()
-            for value in values
-        ]
+        out = []
+        for value in values:
+            if isinstance(value, np.ndarray) or (
+                hasattr(value, "__len__") and not isinstance(value, (str, bytes))
+            ):
+                # Array-like value in a scalar BOOL column: treat as null.
+                out.append(ProtoValue())
+            elif not pd.isnull(value):
+                out.append(
+                    ProtoValue(
+                        **{
+                            field_name: func(
+                                bool(value) if type(value) is np.bool_ else value
+                            )
+                        }
+                    )  # type: ignore
+                )
+            else:
+                out.append(ProtoValue())
+        return out
 
     # Generic scalar conversion
     out = []
     for value in values:
         if isinstance(value, ProtoValue):
             out.append(value)
+        elif isinstance(value, np.ndarray) or (
+            hasattr(value, "__len__") and not isinstance(value, (str, bytes))
+        ):
+            # Array-like value in a scalar column: always treat as null.
+            # pd.isnull() is vectorised and would return an ndarray here,
+            # making `not pd.isnull(value)` raise ValueError.
+            out.append(ProtoValue())
         elif not pd.isnull(value):
             out.append(ProtoValue(**{field_name: func(value)}))
         else:
--- a/sdk/python/feast/type_map.py
+++ b/sdk/python/feast/type_map.py
@@ -909,6 +909,16 @@ def _convert_list_values_to_proto(
     ]
 
 
+def _is_array_like(value: Any) -> bool:
+    """Return True if *value* is array-like (numpy array or any sized,
+    non-string, non-bytes container).  Array-like values in a scalar
+    feature column cannot be mapped to a protobuf scalar field and are
+    therefore always treated as null."""
+    return isinstance(value, np.ndarray) or (
+        hasattr(value, "__len__") and not isinstance(value, (str, bytes))
+    )
+
+
 def _convert_scalar_values_to_proto(
     feast_value_type: ValueType,
     values: List[Any],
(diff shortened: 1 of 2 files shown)


sdk/python/feast/type_map.py
```diff
@@ -929,16 +929,35 @@ def _convert_scalar_values_to_proto(
         return [ProtoValue()] * len(values)
 
     if feast_value_type == ValueType.UNIX_TIMESTAMP:
-        int_timestamps = _python_datetime_to_int_timestamp(values)
-        return [ProtoValue(unix_timestamp_val=ts) for ts in int_timestamps]  # type: ignore
+        out = []
+        for value in values:
+            if isinstance(value, np.ndarray) or (
```
ntkathole: let's extract this into small helper and use it everywhere to avoid duplication:

```
def _is_array_like(value: Any) -> bool:
    return isinstance(value, np.ndarray) or (
        hasattr(value, "__len__") and not isinstance(value, (str, bytes))
    )
```

sdk/python/feast/type_map.py
```diff
@@ -929,16 +929,35 @@ def _convert_scalar_values_to_proto(
...
+                hasattr(value, "__len__") and not isinstance(value, (str, bytes))
+            ):
+                # Array-like value in a scalar UNIX_TIMESTAMP column: treat as null.
+                out.append(ProtoValue())
+            elif value is None:
+                out.append(ProtoValue())
+            else:
+                (ts,) = _python_datetime_to_int_timestamp([value])
+                out.append(ProtoValue(unix_timestamp_val=ts))  # type: ignore
```
ntkathole: if else logic is going through all rows, better to pre filter first:

```
if feast_value_type == ValueType.UNIX_TIMESTAMP:
    out = [None] * len(values)
    clean_indices = []
    clean_values = []
    for i, value in enumerate(values):
        if _is_array_like(value) or value is None:
            out[i] = ProtoValue()
        else:
            clean_indices.append(i)
            clean_values.append(value)
    if clean_values:
        timestamps = _python_datetime_to_int_timestamp(clean_values)
        for i, ts in zip(clean_indices, timestamps):
            out[i] = ProtoValue(unix_timestamp_val=ts)
    return out
```

ntkathole:
@alan-gauthier-jt I think even better if we fix this at https://github.com/feast-dev/feast/blob/master/sdk/python/feast/type_map.py by adding logic for scalar columns, skip array-like values when picking a sample.

sdk/python/feast/type_map.py
alan-gauthier-jt: Thanks, I implemented your solution

sdk/python/feast/type_map.py
alan-gauthier-jt: Done

sdk/python/feast/type_map.py
```diff
@@ -1107,12 +1150,18 @@ def _python_value_to_proto_value(
...
+    # Scalar types — pick a sample that is not array-like so that the type
+    # validation in _convert_scalar_values_to_proto always receives a plain
+    # scalar (array-like values in a scalar column are treated as null).
     if (
         feast_value_type in PYTHON_SCALAR_VALUE_TYPE_TO_PROTO_VALUE
         or feast_value_type == ValueType.UNIX_TIMESTAMP
     ):
-        return _convert_scalar_values_to_proto(feast_value_type, values, sample)
+        scalar_sample = next(
+            (v for v in values if _non_empty_value(v) and not _is_array_like(v)),
```
alan-gauthier-jt, line 1161: @ntkathole is this what you had in mind regarding the samples?

ntkathole (approved):
Thanks @alan-gauthier-jt looks good

PR #6263, opened 2026-04-11
fix: use ValueError and fix f-string interpolation in yaml_loader

korbonits:

Fixes two issues in `sdk/python/feast/loaders/yaml.py`:

1. **Broken f-string interpolation**: `${yaml_strings}` in the error message on line 25 does not interpolate in Python — the `$` prefix is invalid and the variable value was never included in the error string. Fixed to `{yaml_strings}`.
2. **Over-broad exception types**: Both `yaml_loader` and `_get_yaml_contents` raised bare `Exception` instead of `ValueError`, making it harder for callers to catch specific errors.

N/A

# Checks

## Testing Strategy
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

### DIFF
--- a/sdk/python/feast/loaders/yaml.py
+++ b/sdk/python/feast/loaders/yaml.py
@@ -21,8 +21,8 @@ def yaml_loader(yml, load_single=False):
     # Return a single resource dict
     if load_single:
         if len(yaml_strings) > 1:
-            raise Exception(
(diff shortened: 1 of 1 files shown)


korbonits:
Sorry this is dead code, going to close this PR. Maybe I will raise another to delete it.

PR #6275, opened 2026-04-14
fix: Fix five bugs in milvus online store

korbonits:
## What does this PR do?

Fixes five bugs in `sdk/python/feast/infra/online_stores/milvus_online_store/milvus.py` and adds regression tests for three of them.

### 1. `update()` overwrites the collection cache (`_collections`) instead of updating it

`_get_or_create_collection()` already updates `self._collections[collection_name]` as a side effect and returns that entry. The assignment in `update()` replaced the entire `_collections` dict with the raw `describe_collection()` dict for the last table, corrupting all subsequent cache lookups for any other collection.

**Fix:** drop the assignment; the side effect in `_get_or_create_collection()` is sufficient.

### 2. `plan()` raises `NotImplementedError` instead of returning `[]`

The `OnlineStore` base class default is `return []`. Raising `NotImplementedError` caused `feast plan` to crash for any project using the Milvus store.

### 3. `online_read()` has an extra `full_feature_names` parameter not in the base class

The parameter was unused in the method body and violates the `OnlineStore` interface contract (`online_read` takes `requested_features` as the last argument).

### 4. `retrieve_online_documents_v2()` crashes with `TypeError` when the composite key is absent from a hit

```python
# before — bytes.fromhex(None) raises TypeError
entity_key_bytes = bytes.fromhex(
    hit.get("entity", {}).get(composite_key_name, None)
)

# after
raw_key = hit.get("entity", {}).get(composite_key_name)
entity_key_bytes = bytes.fromhex(raw_key) if raw_key else None
```

### 5. `_connect()` uses `print()` instead of `logging`

Connection messages now use `logger.info()` so they respect the application's logging configuration instead of unconditionally writing to stdout.

## Tests

Three regression tests added in `tests/unit/online_store/test_online_retrieval.py`:

| Test | Bug covered |
|------|-------------|
| `test_milvus_update_preserves_collection_cache` | `_collections` cache corruption in `update()` |
| `test_milvus_plan_returns_empty_list` | `plan()` raising `NotImplementedError` |
| `test_milvus_retrieve_online_documents_v2_missing_entity_key` | `TypeError` from `bytes.fromhex(None)` |

Full Milvus unit test results:

```
tests/unit/online_store/test_online_retrieval.py -k milvus
→ 8 passed, 1 skipped  (7.90s)
```

`ruff check` passes with no errors.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


franciscojavierarceo (approved):

korbonits:
Hmm, looks like the unit test error for python 3.12 ubuntu was:

```python
________________ tests/unit/local_feast_tests/test_e2e_local.py ________________
[gw6] linux -- Python 3.12.13 /home/runner/work/feast/feast/.venv/bin/python3
worker 'gw6' crashed while running 'tests/unit/local_feast_tests/test_e2e_local.py::test_e2e_local'
=============================== warnings summary ==============================
```

Maybe re-running would resolve?

Issue #5456, opened 2025-06-16
Support Data Labeling and LabelViews


franciscojavierarceo:
**Is your feature request related to a problem? Please describe.**
Historically, Feast has played a key role in feature development. Particularly around dataset preparation for model development and feature serving for online inference. 

Pictorially, you can think of it like this:

![Image](https://github.com/user-attachments/assets/a1d2a6e7-c661-4e5c-9f0e-ccad9ff22e4f)

Yet labels are the core piece of a training dataset that makes model training successful. Without labels, features are a waste of time (excluding semi/self-supervised learning). 

Given the work with compute engine, my proposal is to expand Feast to include the entire Training Dataset preparation life cycle which would include labels and their correction.

A proof of concept was developed in the UI to highlight educate users about this here: https://github.com/feast-dev/feast/pull/5410

We should expand this properly so that users can define a `LabelView` in the online store that can be used to store labels explicitly.

**Describe the solution you'd like**
A `LabelView` that can be used to write data to the online and offline store. 

**Describe alternatives you've considered**
A clear and concise description of any alternative solutions or features you've considered. It could look something like:

```python
customer = Entity(name="customer_id", dtype=Int64)

# 2) Point to your label data in e.g. Parquet
label_source = FileSource(
    path="gs://my-bucket/churn_labels/*.parquet",
    event_timestamp_column="label_timestamp",
    created_timestamp_column="created_ts",
)

# 3) Declare the LabelView
customer_churn = LabelView(
    name="customer_churn",
    entities=[customer],
    schema=[
        Field(name="churned", dtype=ValueType.BOOL),
        Field(name="risk_score", dtype=ValueType.FLOAT),
    ],
    batch_source=label_source,
    ttl=timedelta(days=90),
    description="Customer churn flag and risk score for training/monitoring.",
)

```

**Additional context**

haoxu0:
This make sense. One suggestion, it would be better to put it in a separate module, since it's more domain specific feature view now. And I think the module will grow much broader as time goes by.

franciscojavierarceo:
Yeah, agree.

franciscojavierarceo:
Also this is a useful example of logging but it omits the label for sure. https://whylogs.readthedocs.io/en/latest/examples/integrations/Feature_Stores_and_whylogs.html

ntkathole:
https://docs.feast.dev/master/getting-started/concepts/label-view

PR #6292, opened 2026-04-17
feat: Feast First-Class LabelView Implementation

ntkathole:

RFC - https://docs.google.com/document/d/1mcJM0sHeBk269oMIsLsLnAlrMtgnxn5Pws0oEeogc48/edit?usp=sharing 

## Summary

Introduces `LabelView`, a new alpha-stage Feast primitive that manages mutable labels and annotations separately from immutable feature data stored in regular `FeatureView`s. This supports multi-labeler workflows (human reviewers, automated safety scanners, reward models) where different sources independently write labels for the same entity keys.

LabelView is fully opt-in — existing workflows are unaffected unless a user explicitly defines one.

https://github.com/user-attachments/assets/4812e034-cdee-408c-9fba-3d9f081cc357

### What's included

**Core implementation (`feast.labeling` module)**
- `LabelView` class inheriting from `BaseFeatureView`, with label-specific attributes: `labeler_field`, `conflict_policy`, `retain_history`, `reference_feature_view`
- `ConflictPolicy` enum (`LAST_WRITE_WINS`, `LABELER_PRIORITY`, `MAJORITY_VOTE`)
- Protobuf definition (`LabelView.proto`) with `LabelViewSpec`, `LabelViewMeta`, and `ConflictResolutionPolicy`

**Full SDK integration**
- `FeatureStore.apply()` — registers LabelViews, auto-registers their PushSource, validates name uniqueness across all view types
- `FeatureStore.push()` — routes push data to LabelView when its PushSource matches
- `FeatureStore.get_historical_features()` — works via `batch_source`/`stream_source` properties on LabelView
- `FeatureStore.teardown()` — includes LabelView tables
- `FeatureStore.list_label_views()` / `get_label_view()` — new public API methods
- `FeatureService` composability — LabelViews can be included alongside regular feature views

**Registry support (all backends)**
- File registry — full CRUD, apply_materialization, name-conflict checks
- SQL registry — `label_views` table, `_infer_fv_table`/`_infer_fv_classes` support, proto() builder
- Remote (gRPC) registry — `label_view` arm added to `ApplyFeatureViewRequest` oneof, full client + server wiring
- Snowflake registry — DDL for `LABEL_VIEWS` table, correct column-name mapping in `delete_feature_view`

**CLI**
- `feast label-views list` — lists all label views with name, entities, conflict policy
- `feast label-views describe <name>` — shows full label view details
- `feast feature-views list` unchanged (label views are only in their dedicated command)

**Permissions**
- `LABEL_VIEW = 11` added to `Permission.proto` Type enum
- `_PERMISSION_TYPES` map updated for `from_proto`/`to_proto` roundtrip

**Regression safety**
- No changes to existing `feature-views list` output
- All registry `proto()` builders include `label_views` so cached/exported blobs are complete
- Cross-type name uniqueness enforced via `_ensure_feature_view_name_is_unique`

**Documentation**
- Docstrings with `[Alpha]` tags and Google-style Args blocks on all public classes/methods
- Sphinx/RST API reference entries for `LabelView` and `ConflictPolicy`
- Concept guide (`docs/getting-started/concepts/label-view.md`) with usage examples, conflict policies, and alpha limitations
- Alpha limitations clearly documented: `conflict_policy` stored but not enforced at read time (requires online-store query-path changes); `retain_history` stored but not enforced at write time (requires online-store write-path changes); batch materialization not supported

**Tests**
- 30 unit tests covering creation, defaults, proto roundtrip, copy/equality, validation, FeatureService composition, registry proto roundtrip, batch_source/stream_source properties, name-conflict detection, and MaterializationTask acceptance

## Alpha Limitations

- `conflict_policy` — persisted in registry metadata but not enforced during `get_online_features`. Online store returns last-written row regardless of policy.
- `retain_history` — persisted but not acted on. Online store always overwrites previous value.

## Test Plan

- [x] All 30 `test_label_view.py` unit tests pass
- [x] All 5 `test_registry_diff.py` tests pass (including `test_diff_registry_objects_permissions`)
- [x] No linter errors introduced
- [x] Existing feature view / entity / feature service workflows unaffected when no LabelView is defined
- [x] Manual: `feast apply` with a repo containing a LabelView definition
- [x] Manual: `feast label-views list` / `feast label-views describe`
- [x] Manual: `FeatureStore.push()` to a LabelView's PushSource

https://github.com/feast-dev/feast/issues/5456


JiwaniZakir (commented):
In `docs/getting-started/concepts/label-view.md`, the `LabelView` definition example includes `interaction_id` both as the entity join key (`Entity(join_keys=["interaction_id"])`) and as an explicit `Field` inside the `schema` list. Standard Feast `FeatureView` definitions typically exclude join-key columns from the schema since they're inferred from the entity definition — duplicating them here either reflects an inconsistency with how the underlying implementation handles schema validation, or it will silently produce duplicate columns. This should either be removed from the schema to match Feast conventions or explicitly documented as a required pattern specific to `LabelView`.

On the alpha-limitations transparency: documenting that `conflict_policy` and `retain_history` are persisted but not enforced is the right call, but consider surfacing a runtime warning (e.g., a `UserWarning` on `apply()` or `push()`) when a non-default `ConflictPolicy` or `retain_history=True` is configured. As written, users who set `ConflictPolicy.MAJORITY_VOTE` will silently get last-write-wins behavior, which is a subtle correctness hazard that the docs alone may not prevent.

ntkathole:
> In `docs/getting-started/concepts/label-view.md`, the `LabelView` definition example includes `interaction_id` both as the entity join key (`Entity(join_keys=["interaction_id"])`) and as an explicit `Field` inside the `schema` list. Standard Feast `FeatureView` definitions typically exclude join-key columns from the schema since they're inferred from the entity definition — duplicating them here either reflects an inconsistency with how the underlying implementation handles schema validation, or it will silently produce duplicate columns. This should either be removed from the schema to match Feast conventions or explicitly documented as a required pattern specific to `LabelView`.

Thanks for review @JiwaniZakir. Can you please re-confirm - this is how regular feature view works, it isn't duplication. 

> 
> On the alpha-limitations transparency: documenting that `conflict_policy` and `retain_history` are persisted but not enforced is the right call, but consider surfacing a runtime warning (e.g., a `UserWarning` on `apply()` or `push()`) when a non-default `ConflictPolicy` or `retain_history=True` is configured. As written, users who set `ConflictPolicy.MAJORITY_VOTE` will silently get last-write-wins behavior, which is a subtle correctness hazard that the docs alone may not prevent.

valid point but the next PR will have support for conflict_policy and retain_history and required changes in online store.

JiwaniZakir:
The Python 3.12 worker crash in `test_e2e_local.py` is worth investigating before merging — a `gw6` worker crash typically signals a segfault or unhandled exception during collection or teardown, which could be unrelated to LabelView itself but should be confirmed. Can you share the full traceback from that run? Also, given the `retain_history` and `conflict_policy` attributes on `LabelView`, it would be worth confirming the batched write tests cover the `MAJORITY_VOTE` conflict resolution path specifically, since that's the most stateful and error-prone of the three policies.

franciscojavierarceo (approved):

PR #6310, opened 2026-04-21
fix(spark): BatchFeatureView with TransformationMode.PYTHON now reads all source columns

abhijeet-dhumal:
## What
FeatureBuilder.get_column_info() sets feature_cols=[] to signal "read all source columns" for transformation modes that compute output features from raw input. This signal was only applied for "ray" and "pandas" modes — "python" was missing.
TransformationMode.PYTHON is the default mode for @batch_feature_view. This means every default @batch_feature_view using SparkComputeEngine was broken out of the box.

## Why it breaks
With feature_cols populated from the view's output schema (e.g. ["user_avg_rating", "user_review_count"]), the Spark offline store read step generates:
```
SELECT user_id, user_avg_rating, user_review_count, event_timestamp
FROM s3a://bucket/processed/reviews/
WHERE event_timestamp BETWEEN ...
```
These output feature names don't exist in the raw source data — the UDF is supposed to compute them. Spark's analyzer throws:
```
AnalysisException: [UNRESOLVED_COLUMN.WITH_SUGGESTION] A column or function
parameter with name 'user_avg_rating' cannot be resolved.
```
## Fix
Added "python" to the existing mode guard in get_column_info()

[x] Verified end-to-end on OpenShift AI with Feast 0.62.0 + PySpark 3.5.3 + MinIO + Redis — 2M+ feature vectors materialized successfully after fix

### DIFF
--- a/sdk/python/feast/infra/compute_engines/feature_builder.py
+++ b/sdk/python/feast/infra/compute_engines/feature_builder.py
@@ -158,12 +158,14 @@ class FeatureBuilder(ABC):
         # we need to read ALL source columns, not just the output feature columns.
         # This is specifically for transformations that create new columns or need raw data.
         mode = getattr(getattr(view, "feature_transformation", None), "mode", None)
-        if mode in ("ray", "pandas") or getattr(mode, "value", None) in (
+        if mode in ("ray", "pandas", "python") or getattr(mode, "value", None) in (
             "ray",
             "pandas",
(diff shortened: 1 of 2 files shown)


ntkathole (approved):

ntkathole:
@abhijeet-dhumal seems pushed the different commits in this PR ?

PR #6324, opened 2026-04-24
fix: Handle array of strings columns in Athena materialization

### LINKED ISSUE #6325: Bug: TypeError / ValueError when materializing Array(String) feature views with Athena offline store
## Description

Materializing feature views that contain `Array(String)` columns using the Athena offline store fails intermittently with one of two errors:

### Error 1: `ValueError: The truth value of an empty array is ambiguous`

Triggered when an entity row has no values set for an array column (e.g. `tags = []`).

```
File ".../feast/type_map.py", line 772, in _python_value_to_proto_value
    elif not pd.isnull(value):
ValueError: The truth value of an empty array is ambiguous. Use `array.size > 0` to check that an array is not empty.
```

### Error 2: `TypeError: bad argument type for built-in operation`

Triggered when any user has values in an array column.

```
File ".../feast/type_map.py", line 905, in <listcomp>
    ProtoValue(**{field_name: proto_type(val=value)})
TypeError: bad argument type for built-in operation
```

## Root Cause

Arrow/Athena deserializes `Array(String)` feature columns as `numpy.ndarray` with `object` dtype rather than plain Python lists. Two code paths in `type_map.py` do not handle this:

1. **Scalar null-check** (`_convert_scalar_values_to_proto`): The line `elif not pd.isnull(value)` calls `pd.isnull()` on a numpy array, which returns an array of bools — then `not <array>` raises `ValueError` because the truth value is ambiguous.

2. **Generic list conversion** (`_convert_list_values_to_proto`): The call `proto_type(val=value)` passes the raw `numpy.ndarray` directly to the protobuf constructor. Protobuf rejects non-list types and raises `TypeError`. Additionally, Arrow nullable columns can produce `None` elements inside the ndarray, which protobuf `StringList` also rejects.

3. **Type validation** (`_validate_collection_item_types`): `None` elements inside an ndarray fail the `type(item) in valid_types` check before they can be sanitized downstream.

## Steps to Reproduce

1. Define a `FeatureView` with an `Array(String)` field:
   ```python
   Field(name="tags", dtype=Array(String))
   ```
2. Materialize from Athena where some rows have non-empty arrays, some have empty arrays, and some have NULL values in array elements.
3. Observe `ValueError` or `TypeError` in `feast/type_map.py`.

## Expected Behavior

Materialization completes successfully. Array columns from Arrow/Athena are converted to proto-safe Python lists, with `None` elements replaced by empty string.

## Environment

- Feast version: (any version with the generic list proto conversion)
- Offline store: Athena
- Python: 3.11
- Feature column type: `Array(String)` (maps to `ValueType.STRING_LIST`)

## Fix

I opened a PR with a fix for this issue: https://github.com/feast-dev/feast/pull/6324

alan-gauthier-jt:

Fixes two related bugs that cause `TypeError` and `ValueError` when materializing
feature views with array-typed columns (e.g. `Array(String)`, `Array(Int64)`) using
the Athena offline store.

Arrow/Athena deserializes array columns as `numpy.ndarray` (object dtype) instead of
plain Python lists. This breaks two code paths in `type_map.py`:

1. `_convert_scalar_values_to_proto`: `pd.isnull(ndarray)` returns an array of bools,
   and `not <array>` raises `ValueError: The truth value of an empty array is ambiguous`.
   → Already guarded by `_is_array_like` in newer Feast versions; no change needed here.

2. `_convert_list_values_to_proto` (generic list path): `proto_type(val=ndarray)` passes
   the raw numpy array to the protobuf constructor, which only accepts Python lists →
   `TypeError: bad argument type for built-in operation`. Additionally, Arrow nullable
   columns can yield `None` elements inside the ndarray, which protobuf repeated fields
   also reject.

3. `_validate_collection_item_types`: `None` elements inside an ndarray failed the
   `type(item) in valid_types` check before reaching the sanitization step.

### `feast/type_map.py`

- **Add module-level `_LIST_NONE_DEFAULTS`** dict mapping each list `ValueType` to a
  type-appropriate zero/empty default value used to replace `None` elements:
  - `STRING_LIST`, `UUID_LIST`, `TIME_UUID_LIST`, `DECIMAL_LIST` → `""`
  - `BYTES_LIST` → `b""`
  - `INT32_LIST`, `INT64_LIST` → `0`
  - `FLOAT_LIST`, `DOUBLE_LIST` → `0.0`
  - `BOOL_LIST` → `False`
  - `UNIX_TIMESTAMP_LIST` → `NULL_TIMESTAMP_INT_VALUE`

- **Add module-level `_sanitize_list_value(value, feast_value_type)`** helper that:
  - Calls `.tolist()` on any `numpy.ndarray` to produce a plain Python list
    (empty ndarray → `None`, treated as a missing row)
  - Replaces `None` elements with the type-appropriate default from `_LIST_NONE_DEFAULTS`
  - Is a no-op for plain Python lists without `None` and for scalar values

- **Apply sanitization upfront in `_convert_list_values_to_proto`**: both `values` and
  `sample` are normalised via `_sanitize_list_value` before any type-checking or proto
  conversion, removing the need for per-path ndarray handling.

- **Remove the old `_to_proto_safe_list` / `_DROP_NONE` / `_LIST_TYPE_NONE_REPLACEMENT`**
  module-level helpers, which have been superseded by the above.

- **Skip `None` elements in `_validate_collection_item_types`** — `None` entries are
  valid in nullable Arrow columns and are sanitized upstream; raising a `TypeError` on
  them before that point was incorrect.

## Testing

Added `TestArrowArrayStringListMaterialization` in
`sdk/python/tests/unit/test_type_map.py` covering:

| Test | Scenario |
|---|---|
| `test_sanitize_list_value_ndarray` | ndarray → plain list |
| `test_sanitize_list_value_empty_ndarray` | empty ndarray → `None` (missing row) |
| `test_sanitize_list_value_ndarray_with_none` | `None` elements in STRING_LIST replaced with `""` |
| `test_sanitize_list_value_plain_list` | plain list passthrough |
| `test_sanitize_list_value_plain_list_with_none` | `None` in plain STRING_LIST list replaced with `""` |
| `test_sanitize_list_value_numeric_none_replaced` | `None` in numeric/bool lists replaced with zero default |
| `test_sanitize_list_value_bytes_none_replaced` | `None` in BYTES_LIST replaced with `b""` |
| `test_sanitize_list_value_scalar_passthrough` | non-list, non-ndarray values unchanged |
| `test_string_list_from_ndarray` | full round-trip via `python_values_to_proto_values` |
| `test_string_list_from_empty_ndarray` | empty ndarray no longer raises `ValueError` |
| `test_string_list_from_ndarray_with_none_elements` | `None` in ndarray no longer raises `TypeError` |
| `test_string_list_null_row_produces_empty_proto` | `None` rows produce empty `ProtoValue` |
| `test_mixed_batch_simulating_athena_chunk` | full simulation of a failing Athena materialization batch |

```
pytest sdk/python/tests/unit/test_type_map.py::TestArrowArrayStringListMaterialization -v
```

## Which issues this PR fixes

Fixes #6325 

## Does this PR introduce a user-facing change?

Yes — materialization of array-typed feature columns from Athena no longer fails with
`TypeError` or `ValueError` when a batch contains empty arrays, `None` rows, or `None`
elements inside arrays. `None` elements inside an array are now stored as the
type-appropriate zero/empty value (e.g. `""` for strings, `0` for integers).

```
Previously:
  TypeError: bad argument type for built-in operation
  ValueError: The truth value of an empty array is ambiguous

After this fix:
  Materialization completes successfully.
  None elements inside arrays are replaced with type-appropriate defaults.
```

### DIFF
--- a/sdk/python/feast/type_map.py
+++ b/sdk/python/feast/type_map.py
@@ -739,7 +739,7 @@ def _validate_collection_item_types(
     """
     if sample is None:
         return
-    if all(type(item) in valid_types for item in sample):
+    if all(type(item) in valid_types for item in sample if item is not None):
         return
 
     # to_numpy() upcasts INT32/INT64 with NULL to Float64 automatically
@@ -750,6 +750,8 @@ def _validate_collection_item_types(
         ValueType.INT64_SET,
     ]
     for item in sample:
+        if item is None:
+            continue  # None elements in STRING_LIST are replaced with ""; for other types they are dropped
         if type(item) not in valid_types:
             if feast_value_type in int_collection_types:
                 # Check if the float values are due to NULL upcast
@@ -868,6 +870,46 @@ def _python_set_to_proto_values(
     ]
 
 
+# Sentinel value used by _to_proto_safe_list to indicate that None elements
+# should simply be filtered (dropped) rather than replaced with a default.
+_DROP_NONE = object()
+
+# Per-type default values substituted for None elements inside list columns.
+# Only STRING_LIST uses ""; numeric/bytes types drop None entirely because
+# there is no meaningful in-band sentinel (protobuf rejects wrong scalar types).
+_LIST_TYPE_NONE_REPLACEMENT: Dict[ValueType, Any] = {
+    ValueType.STRING_LIST: "",
+}
+
+
+def _to_proto_safe_list(
+    value: Any, feast_value_type: ValueType = ValueType.STRING_LIST
+) -> Any:
+    """Convert an array/list column value to a proto-safe Python list.
+
+    Arrow/Athena returns Array columns as numpy.ndarray (object dtype).
+    Protobuf repeated fields reject ndarrays and (for non-string types) None
+    elements, so we:
+    1. Call .tolist() to convert any numpy.ndarray to a plain Python list.
+    2. For STRING_LIST: replace None elements with "" (empty string).
+       For all other list types: drop None elements, since there is no valid
+       in-band default for numeric/bytes protobuf fields.
+
+    Args:
+        value: The raw column value (ndarray, list, or scalar).
+        feast_value_type: The Feast ValueType of the list column. Controls how
+            None elements are handled. Defaults to STRING_LIST.
+    """
+    if isinstance(value, np.ndarray):
+        value = value.tolist()
+    if isinstance(value, list):
+        none_replacement = _LIST_TYPE_NONE_REPLACEMENT.get(feast_value_type, _DROP_NONE)
+        if none_replacement is _DROP_NONE:
+            return [x for x in value if x is not None]
+        return [x if x is not None else none_replacement for x in value]
+    return value
+
+
 def _convert_list_values_to_proto(
     feast_value_type: ValueType,
     values: List[Any],
@@ -946,8 +988,14 @@ def _convert_list_values_to_proto(
         ]
 
     # Generic list conversion
+    # Arrow/Athena deserializes Array columns as numpy.ndarray (object dtype).
+    # _to_proto_safe_list converts to a plain Python list and sanitizes None
+    # elements in a type-appropriate way (replaced with "" for STRING_LIST,
+    # dropped for numeric/bytes types).
     return [
-        ProtoValue(**{field_name: proto_type(val=value)})  # type: ignore[arg-type]
+        ProtoValue(
+            **{field_name: proto_type(val=_to_proto_safe_list(value, feast_value_type))}  # type: ignore[arg-type]
+        )
         if value is not None
         else ProtoValue()
         for value in values
--- a/sdk/python/feast/type_map.py
+++ b/sdk/python/feast/type_map.py
@@ -897,8 +897,6 @@ def _sanitize_list_value(value: Any, feast_value_type: ValueType) -> Any:
     """
     if isinstance(value, np.ndarray):
         value = value.tolist()
-        if isinstance(value, list) and len(value) == 0:
-            return None
     none_default = _LIST_NONE_DEFAULTS.get(feast_value_type)
     if none_default is not None and isinstance(value, list):
         value = [none_default if v is None else v for v in value]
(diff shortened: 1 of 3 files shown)


sdk/python/feast/type_map.py
```diff
@@ -823,6 +825,46 @@ def convert_set_to_list(value: Any) -> Any:
...
 
 
+# Sentinel value used by _to_proto_safe_list to indicate that None elements
+# should simply be filtered (dropped) rather than replaced with a default.
+_DROP_NONE = object()
+
+# Per-type default values substituted for None elements inside list columns.
+# Only STRING_LIST uses ""; numeric/bytes types drop None entirely because
+# there is no meaningful in-band sentinel (protobuf rejects wrong scalar types).
+_LIST_TYPE_NONE_REPLACEMENT: Dict[ValueType, Any] = {
```
ntkathole: @alan-gauthier-jt The approach used in https://github.com/feast-dev/feast/pull/6327/changes seems safer and preserve list length

sdk/python/feast/type_map.py
```diff
@@ -823,6 +825,46 @@ def convert_set_to_list(value: Any) -> Any:
...
+        feast_value_type: The Feast ValueType of the list column. Controls how
+            None elements are handled. Defaults to STRING_LIST.
+    """
+    if isinstance(value, np.ndarray):
+        value = value.tolist()
+    if isinstance(value, list):
+        none_replacement = _LIST_TYPE_NONE_REPLACEMENT.get(feast_value_type, _DROP_NONE)
+        if none_replacement is _DROP_NONE:
+            return [x for x in value if x is not None]
+        return [x if x is not None else none_replacement for x in value]
```
ntkathole: if feast_value_type in _LIST_TYPE_NONE_REPLACEMENT instead?

sdk/python/feast/type_map.py
alan-gauthier-jt: I changed the implementation to a similar solution than #6327

franciscojavierarceo (approved):

sdk/python/feast/type_map.py
alan-gauthier-jt: This was removed following the re-implementation

ntkathole (approved):

PR #6360, opened 2026-05-02
fix(bigquery): Prefer query over table in get_table_query_string

### LINKED ISSUE #6200: BigQuerySource.get_table_query_string() silently ignores query when table is also set
## Summary

When both `table` and `query` are provided to `BigQuerySource`, `get_table_query_string()` always returns `table`, silently ignoring `query`. This makes it impossible to use a custom query (e.g., for deduplication) on a `PushSource` batch source, since `PushSource` requires `table` for offline writes via `offline_write_batch()`.

## Expected Behavior

When both `table` and `query` are set on a `BigQuerySource`:
- **Reads** (`get_table_query_string()`) should use `query` — it's more specific and intentionally provided
- **Writes** (`offline_write_batch()`) should continue using `.table` directly as the write destination

## Current Behavior

`get_table_query_string()` in `bigquery_source.py` always prefers `table`:

```python
def get_table_query_string(self) -> str:
    if self.table:
        return f"`{self.table}`"
    return f"({self.query})"
```

This means any custom `query` (e.g., deduplication logic) is silently ignored when `table` is also present.

## Use Case

Streaming (push) sources often produce duplicate rows in BigQuery. The natural solution is:

```python
batch_source = BigQuerySource(
    name="my_batch_source",
    table="project.dataset.my_table",        # needed for push writes
    query="""
        SELECT * FROM `project.dataset.my_table`
        QUALIFY ROW_NUMBER() OVER (PARTITION BY entity_id, event_time) = 1
    """,                                      # needed for deduplicated reads
    timestamp_field="event_time",
)
push_source = PushSource(name="my_source", batch_source=batch_source)
```

But because `get_table_query_string()` ignores `query` when `table` is set, reads return duplicates. And removing `table` to force `query` usage breaks `offline_write_batch()`, which accesses `.table` directly (`bigquery.py:449`).

## Environment

- Feast version: 0.58.0 (also confirmed unresolved on 0.61.0 / current main)
- Offline store: BigQuery

Jwrede:
## Summary

Fixes #6200

When both `table` and `query` are set on a `BigQuerySource`, `get_table_query_string()` silently ignores `query` and always returns the table reference. This makes it impossible to use a custom read query (e.g., for deduplication via `QUALIFY`) on a `PushSource`, since `PushSource` requires `table` for offline writes.

**Root cause:** `get_table_query_string()` checks `if self.table` first — since `table` is always truthy when set, `query` is never reached.

**Fix:** Invert the priority — prefer `query` when present (it's more specific and intentionally provided), fall back to `table`. The write path (`offline_write_batch()`) accesses `.table` directly and is unaffected.

Also applies the same fix to `get_table_column_names_and_types()` so schema inference uses the query when both are set, matching the actual read path.

## Changes

- `bigquery_source.py`: Swap condition order in `get_table_query_string()` and `get_table_column_names_and_types()` to prefer `query` over `table`
- `test_bigquery.py`: Add 4 unit tests covering table-only, query-only, both-set, and write-path-unaffected scenarios

## Test plan

- [x] All 11 existing + new unit tests pass (`pytest sdk/python/tests/unit/infra/offline_stores/test_bigquery.py`)
- [ ] Integration test with BigQuery PushSource (requires GCP credentials)

### DIFF
--- a/sdk/python/feast/infra/offline_stores/bigquery_source.py
+++ b/sdk/python/feast/infra/offline_stores/bigquery_source.py
@@ -156,10 +156,10 @@ class BigQuerySource(DataSource):
 
     def get_table_query_string(self) -> str:
         """Returns a string that can directly be used to reference this table in SQL"""
-        if self.table:
-            return f"`{self.table}`"
-        else:
+        if self.query:
             return f"({self.query})"
+        else:
+            return f"`{self.table}`"
 
     @staticmethod
     def source_datatype_to_feast_value_type() -> Callable[[str], ValueType]:
@@ -185,14 +185,14 @@ class BigQuerySource(DataSource):
             location=config.offline_store.location,
             client_info=http_client_info.ClientInfo(user_agent=get_user_agent()),
         )
-        if self.table:
-            schema = client.get_table(self.table).schema
-            if not isinstance(schema[0], bigquery.schema.SchemaField):
-                raise TypeError("Could not parse BigQuery table schema.")
-        else:
+        if self.query:
             bq_columns_query = f"SELECT * FROM ({self.query}) LIMIT 0"
             query_res = client.query(bq_columns_query).result()
             schema = query_res.schema
+        else:
+            schema = client.get_table(self.table).schema
+            if not isinstance(schema[0], bigquery.schema.SchemaField):
+                raise TypeError("Could not parse BigQuery table schema.")
 
         name_type_pairs: List[Tuple[str, str]] = []
         for field in schema:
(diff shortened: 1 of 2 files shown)


Jwrede:
The failing test (`test_online_write_batch_async_skip_dedup_single_pipeline` in `test_redis.py`) is unrelated to this PR — it's an event-loop issue introduced in 2e50da0 that affects macOS and Python 3.12. My changes only touch `bigquery_source.py` and `test_bigquery.py`.

The 3.10 and 3.11 Ubuntu runs pass cleanly.

ntkathole:
@Jwrede init from bigquery_source.py says `Exactly one of 'table' and 'query' must be specified.` ?

Jwrede:
Good catch, the docstring says "Exactly one of 'table' and 'query' must be specified" but the actual validation (line 67) only enforces **at least one**:

```python
if table is None and query is None:
    raise ValueError('No "table" or "query" argument provided.')
```

It has never rejected both being reads. I'll push an update to fix the docstring to match reality.

ntkathole (approved):

PR #6361, opened 2026-05-03
fix: Fix intermittent async test failures for DynamoDB and Redis

ntkathole:

This PR fixes intermittent failures in integration tests failing with `botocore.exceptions.ClientError: An error occurred (UnrecognizedClientException) when calling the BatchWriteItem operation: The security token included in the request is invalid.`

This error is a credential staleness issue in the cached aiobotocore client.

Also fixes redis tests eventloop issue.


ntkathole:
Merging to fix test failures on all PRs

PR #6076, opened 2026-03-07
feat: Utilize date partition column in BigQuery

Anarion-zuo:
# What this PR does / why we need it:

- Some BigQuery util does not use the date_partition_column as filtering condition, disabling partioning by those columns. Adjusted the function calls to make use of date_partition_column.
- Added related unit tests and benchmark cases.

# Which issue(s) this PR fixes:

https://github.com/feast-dev/feast/issues/2530

# Misc

To run the newly added unit test:
```
python -m pytest sdk/python/tests/unit/infra/offline_stores/test_bigquery.py
```

To run the newly added benchmark case, first you must be able to access your BigQuery.
```
export FEAST_BQ_BENCH_PROJECT="my-gcp-project"
export FEAST_BQ_BENCH_TABLE="my-gcp-project:my_dataset.my_table"
export FEAST_BQ_BENCH_TIMESTAMP_FIELD="event_timestamp"
export FEAST_BQ_BENCH_PARTITION_COLUMN="_PARTITIONDATE"
export FEAST_BQ_BENCH_LOCATION="US"
export FEAST_BQ_BENCH_REQUIRE_REDUCTION=true
```

Then run
```
python -m pytest sdk/python/tests/benchmarks/test_bigquery_partition_pruning_benchmark.py --benchmark
```

---

<a href="https://app.devin.ai/review/feast-dev/feast/pull/6076" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

PR #6362, opened 2026-05-03
feat(bigquery): Support DATE-type event timestamp columns

### LINKED ISSUE #2530: Handle BigQuery partitions when event_timestamp is not the partition column, deal with event_timestamp columns of DATE type
At a high level, I'd like to be able to install Feast, describe my sources and FVs, and immediately start working to get historical data and materialize features. However, by not being partition-aware and by not accounting for non-timestamp date columns, a large number of our tables are not usable without modification.

This feature request is specific to the GCP provider and BigQuery offline store. It has two parts: 1) relating to respecting partition when constructing date ranges in SQL, 2) relating to source tables with "timestamp" column not of TIMESTAMP type.

1) In any table were the partition column is not the same column as the event_timestamp, Feast will ignore partitions and do a full table scan on each query. For large tables, this causes BQ to fail due to "insufficient resources" - not a slot/quote issue, but having to do with needing to marshal too much transient data. A common example with BQ is when doing DAY partitioning using ingestion time, the magic column _PARTITIONTIME is the partition column. Each record will contain its own actual event_timestamp column, but this column is not the partition column.

2) For tables with an "event_timestamp" that is not a TIMESTAMP type, the SQL can break when there's a type mismatch, e.g. when the column is a DATE type. A common use case is a DAY partitioned summary table that performs aggregations over the day or possibly just takes the final value for each entity/date. These tables have a DATE column on each record. In this case, the DATE column is usually the partition column, but it is not the correct format for the date conditions in the Feast template SQL. I would prefer not to have to create a new column or a view on each table just to make Feast work.

I'd like to be able to specify information about the partition column such that Feast could formulate SQL optimized to my table definition taking advantage of partitioning. This information would probably need to include the partition column name, type, and partition type at a minimum. I believe optimized SQL would not be much different from the existing queries but would require, in cases where the partition column is not the same as the event_timestamp column, to add an additional range condition using the partition column.

I'd also like to be able to indicate that my "event_timestamp" column is not a TIMESTAMP type - and have Feast adjust the SQL accordingly. Specifically, I'd like to specify in the Feast BigQuerySource definition that the "event_timestamp" column is the partition column and is of DATE type and have Feast make the corresponding adjustments to the templatized SQL.

To deal with (1), the only alternatives I can think of are a) creating a view that changes the partition column to the event_timestamp column for every table, b) creating a duplicate of each table using the event_timestamp column.

To deal with (2), similarly, either a) create a view with a new timestamp typed column to use as the event_timestamp, b) add a new timestamp column to the existing tables.

All these solutions require a good amount of work and/or creating duplicate tables/views. I would really like Feast to just do the right thing after I create a Source.

**Additional context**
Happy to provide any further details if the above is not clear.

Jwrede:
## What this PR does / why we need it:

When the `event_timestamp` column in BigQuery is a `DATE` type (not `TIMESTAMP`), the generated SQL wraps comparison values in `TIMESTAMP()`, causing a type mismatch error. This makes DATE-partitioned summary tables unusable without creating views or duplicate tables.

This PR adds an optional `timestamp_field_type` parameter to `BigQuerySource`. When set to `"DATE"`, SQL generation uses `DATE('YYYY-MM-DD')` comparisons instead of `TIMESTAMP('...')`, both in direct queries (`pull_latest_from_table_or_query`, `pull_all_from_table_or_query`) and in the point-in-time join Jinja template.

Usage:
```python
BigQuerySource(
    table="project:dataset.daily_features",
    timestamp_field="event_date",
    timestamp_field_type="DATE",
)
```

Changes:
- Proto: add `timestamp_field_type` string field (field 28) to `DataSource`
- `DataSource` base class: add `timestamp_field_type` attribute, equality check, and `__init__` parameter
- `BigQuerySource`: wire `timestamp_field_type` through `__init__`, `from_proto`, and `_to_proto_impl`
- `get_timestamp_filter_sql()`: add `"date_func"` cast style that generates `DATE('YYYY-MM-DD')`
- `BigQueryOfflineStore`: select cast style based on `timestamp_field_type`
- Jinja template: conditional `DATE()` comparisons for DATE-type timestamp fields
- `FeatureViewQueryContext`: propagate `timestamp_field_type` to template context

Backward-compatible: when `timestamp_field_type` is unset, behavior is unchanged.

## Which issue(s) this PR fixes:

Fixes https://github.com/feast-dev/feast/issues/2530 (part 2 -- DATE type event_timestamp support; part 1 was addressed by #6076)

## How to test:

```
python -m pytest sdk/python/tests/unit/infra/offline_stores/test_bigquery.py -v
```

4 new tests added:
- `test_pull_latest_date_type_timestamp_field` -- verifies `DATE()` cast in pull_latest
- `test_pull_all_date_type_timestamp_field` -- verifies `DATE()` cast in pull_all
- `test_pull_latest_date_type_with_partition_column` -- DATE type combined with partition pruning
- `test_bigquery_source_date_type_proto_roundtrip` -- proto serialization roundtrip


franciscojavierarceo (approved):

Jwrede:
Friendly ping @sudohainguyen -- this and #6365 have been open ~10 days. Happy to address feedback.

ntkathole:
@Jwrede please resolve the conflicts and fix linting

Jwrede:
@ntkathole Rebased onto master and verified ruff check/format passes. All conflicts resolved.

ntkathole:
@Jwrede can you please regenerate protos using correct mypy-protobuf version pinned in CI ?

Jwrede:
@ntkathole Done -- reset all non-DataSource generated files to match master, then hand-edited DataSource_pb2.pyi to add only the timestamp_field_type field in the existing import style. Only DataSource_pb2.py and DataSource_pb2.pyi differ from master now.

ntkathole (approved):

PR #6367, opened 2026-05-05
fix(operator): Set appProtocol: grpc on registry gRPC Service

tmvfb:
## Summary

The registry `Service` (the use case is local registry + `server.grpc` enabled, i.e. SQLite) is always created with `name: http` and no `appProtocol`. This causes service meshes (notably Istio) to misclassify the registry gRPC traffic as HTTP/1.1, breaking connections in two places simultaneously:

- **Client-side envoy**: sees `name: http` → downgrades to HTTP/1.1. gRPC requires HTTP/2.
- **Server-side envoy**: builds its inbound listener for the registry port as HTTP/1.1 → receives gRPC (HTTP/2) frames → sends RST with "protocol error".

Both failures stem from the same missing field. The fix is minimal: set `appProtocol: grpc` on the registry gRPC `Service` port.

## Changes

- Added `getServiceAppProtocol` helper that returns `ptr.To("grpc")` for the registry gRPC Service and `nil` for all other services.
- Set `AppProtocol` on the `ServicePort` in `setService` via this helper.
- Added `k8s.io/utils/ptr` import to `services.go` (already a transitive dependency, used in tests).

## Effect

With `appProtocol: grpc` set, Istio and other service meshes correctly identify the port as gRPC and handle HTTP/2 and mTLS normally, no workarounds required.

## Testing

`go vet ./internal/...` passes. Existing operator unit tests are unaffected (they assert on port number and name, not `appProtocol`). Manual verification on a cluster with Istio sidecar injection confirmed gRPC connectivity works end-to-end after this change with no additional mesh configuration.

## *Workaround*

- **Client-side envoy** downgrades to HTTP/1.1. Fixed by creating an alt-Service with `appProtocol: grpc`.
- **Server-side envoy** builds its inbound listener for port 6570 as HTTP/1.1 and rejects incoming HTTP/2 with a protocol error. The only workaround is `traffic.sidecar.istio.io/excludeInboundPorts: "6570"` on the feast-server pod to bypass it entirely.
- With the server-side envoy bypassed there is nothing to terminate mTLS, so a `DestinationRule` with `tls: DISABLE` is also required to stop the client from attempting it.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1156,7 +1156,7 @@
         "filename": "infra/feast-operator/internal/controller/services/services.go",
         "hashed_secret": "36dc326eb15c7bdd8d91a6b87905bcea20b637d1",
         "is_verified": false,
-        "line_number": 179
+        "line_number": 180
       }
     ],
     "infra/feast-operator/internal/controller/services/tls_test.go": [
@@ -1539,5 +1539,5 @@
       }
     ]
   },
-  "generated_at": "2026-05-20T07:55:48Z"
+  "generated_at": "2026-05-22T11:36:48Z"
 }
--- a/infra/feast-operator/internal/controller/services/services.go
+++ b/infra/feast-operator/internal/controller/services/services.go
@@ -32,6 +32,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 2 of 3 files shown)


tmvfb:
@ntkathole and @shuchu — would appreciate a review when you have a moment. This is a one-liner fix (missing appProtocol: grpc on the registry Service port) that breaks Istio users. Happy to add tests if there's a pattern to follow for operator Service tests.

ntkathole:
@tmvfb can you please add unit test for this? 

Also, fix the linting

tmvfb:
> @tmvfb can you please add unit test for this?
> 
> Also, fix the linting

@ntkathole, thanks for the prompt answer! The linting failure is not caused by this branch and comes from main, this PR only touches the go code.

I added 7 unit tests, first 5 are pretty generic but considering the other unit tests are rather verbose, I think it is alright. The unit test failures also seem unrelated to this branch:

```
=========================== short test summary info ============================

FAILED
  sdk/python/tests/unit/permissions/auth/client/
  test_authentication_client_manager_factory.py::
  test_authentication_client_manager_factory[
    auth:
      type: kubernetes
  ]

  jwt.exceptions.InvalidKeyError:
    HMAC key must not be empty.


FAILED
  sdk/python/tests/unit/permissions/auth/client/
  test_authentication_client_manager_factory.py::
  test_authentication_client_manager_factory[
    auth:
      type: oidc
      client_id: feast-integration-client
      client_secret: feast-integration-client-secret
      username: reader_writer
      password: password
      auth_discovery_url:
        KEYCLOAK_URL_PLACE_HOLDER/realms/master/.well-known/openid-configuration
  ]

  jwt.exceptions.InvalidKeyError:
    HMAC key must not be empty.


==== 2 failed, 1730 passed, 134 skipped, 227 warnings in 489.99s (0:08:09) ====

make: *** [Makefile:170: test-python-unit] Error 1
```

ntkathole:
@tmvfb ignore, unit tests - those will get fixed by https://github.com/feast-dev/feast/pull/6427 
But linting issue is related to this branch, since it touches services.go, secrets baseline failed due to code line number changed

```
 "filename": "infra/feast-operator/internal/controller/services/services.go",
         "hashed_secret": "36dc326eb15c7bdd8d91a6b87905bcea20b637d1",
         "is_verified": false,
-        "line_number": 179
+        "line_number": 180
```

you can either run `detect-secrets scan --baseline .secrets.baseline` or `pre-commit run detect-secrets --all-files` to fix

ntkathole (approved):

PR #6369, opened 2026-05-05
fix: Scope feature view name conflict check to current project in file-based registry

Abhishek8108:
## What this PR does

Fixes #6209.

`_check_conflicting_feature_view_names` built its lookup map from all feature views in `cached_registry_proto` without filtering by project. In a shared file-based registry with multiple projects, this caused false `ConflictingFeatureViewNames` errors when two different projects defined feature views with the same name.

## Root cause

`_existing_feature_view_names_to_fvs` iterated over all three collections (feature views, stream feature views, on-demand feature views) in the entire `RegistryProto` without scoping to the current project. The dict is keyed by name only, so a `FeatureView` named `driver_stats` in `project_a` and a `FeatureView` also named `driver_stats` in `project_b` would collide, triggering a spurious error on the second `feast apply`.

## Fix

- Added a `project: str` parameter to both `_check_conflicting_feature_view_names` and `_existing_feature_view_names_to_fvs`
- Each collection is now filtered by `fv.spec.project == project` so only feature views belonging to the current project are included in the conflict map
- The call site in `apply_feature_view` passes the `project` argument that is already in scope

## Tests

- Added `test_cross_project_feature_view_names_do_not_conflict` in `tests/unit/local_feast_tests/test_local_feature_store.py`: two `FeatureStore` instances sharing a registry file but different projects both apply a `FeatureView` named `driver_stats` — asserts no `ConflictingFeatureViewNames` is raised
- Existing `test_apply_conflicting_feature_view_names` continues to pass — same-project type conflicts are still detected correctly

### DIFF
--- a/infra/feast-operator/bundle/manifests/openlineage-secret_v1_secret.yaml
+++ b/infra/feast-operator/bundle/manifests/openlineage-secret_v1_secret.yaml
@@ -3,4 +3,4 @@ kind: Secret
 metadata:
   name: openlineage-secret
 stringData:
-  api_key: your-marquez-api-key
+  api_key: your-marquez-api-key  # pragma: allowlist secret
--- a/sdk/python/feast/infra/registry/registry.py
+++ b/sdk/python/feast/infra/registry/registry.py
@@ -764,7 +764,7 @@ class Registry(BaseRegistry):
         self._prepare_registry_for_changes(project)
         assert self.cached_registry_proto
 
-        self._check_conflicting_feature_view_names(feature_view)
+        self._check_conflicting_feature_view_names(feature_view, project)
         existing_feature_views_of_same_type: RepeatedCompositeFieldContainer
         if isinstance(feature_view, StreamFeatureView):
             existing_feature_views_of_same_type = (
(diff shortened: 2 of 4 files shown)


Abhishek8108:
Hi — wanted to flag that the remaining CI failure is on the `unit-test-python (3.12, macos-14)` job and appears to be a runner environment issue unrelated to this PR.

The root cause visible in the job logs is a `psutil` version conflict on the macOS runner:

```
ImportError: version conflict:
'.venv/lib/python3.12/site-packages/psutil/_psutil_osx.abi3.so'
C extension module was built for another version of psutil (5.9.8 instead of 5.9.0)
```

Because the `feast` CLI subprocess cannot even complete its import, every test that spawns a CLI subprocess exits immediately with code `-1` and empty stdout. This explains all three failures:

- `test_cli_apply_imported_featureview` — expects exit 0, gets -1
- `test_cli_apply_duplicated_featureview_names` — expects the error message in output, gets empty bytes
- `test_milvus_retrieve_online_documents_v2_missing_entity_key` — unrelated Milvus internal error on the same runner

None of these tests are in files changed by this PR, and the same macOS 3.12 job passes on `master`. The Ubuntu jobs (3.10, 3.11, 3.12) and macOS 3.11 all pass cleanly.

Would you be able to re-trigger the macOS 3.12 job, or confirm you're happy to merge with the understanding that this failure is a transient runner issue? Happy to do anything else needed on my end — thank you for your time!

ntkathole (approved):
lgtm

PR #6381, opened 2026-05-07
fix(trino): Clean up temporary entity tables after retrieval

### LINKED ISSUE #6306: `TrinoOfflineStore` does not cleanup temporary tables
## Expected Behavior

Temporary entity tables created during `get_historical_features()` should be cleaned up after the retrieval job completes, consistent with how BigQuery, Redshift, and Athena handle it.

## Current Behavior

Each call to `get_historical_features()` with a pandas DataFrame `entity_df` creates a permanent table (`{catalog}.{dataset}.feast_entity_df_<uuid>`) that is never dropped, leaking tables in the Trino catalog indefinitely.

## Steps to reproduce

1. This is a known issue in `TrinoOfflineStore` https://github.com/feast-dev/feast/blob/master/sdk/python/feast/infra/offline_stores/contrib/trino_offline_store/trino.py#L486 	

### Specifications

- Version: 0.50+
- Platform: Any
- Subsystem: `TrinoOfflineStore`

## Possible Solution

Refactor `TrinoRetrievalJob` to accept a context manager (`query_generator`) instead of a plain query string, following the pattern used by BigQuery/Redshift/Athena, and add a `DROP TABLE IF EXISTS` in the `finally` block.

Jwrede:
## Summary

`TrinoOfflineStore.get_historical_features()` creates a temporary table
(`feast_entity_df_<uuid>`) for the entity DataFrame but never drops it,
leaking tables in the Trino catalog indefinitely.

This applies the same context manager pattern already used by the
BigQuery, Redshift, and Athena offline stores: wrap the query in a
generator that issues `DROP TABLE IF EXISTS` in a `finally` block.

The cleanup only runs when the entity_df is a DataFrame (when a temp
table was actually created), not when it is a SQL string.

Fixes #6306

## How Has This Been Tested?

- Verified the change follows the same pattern as BigQuery
  (`bigquery.py:304`), Redshift (`redshift.py:236`), and Athena
  (`athena.py:224`).
- All callers of `TrinoRetrievalJob` that pass a plain query string
  (e.g. `pull_latest_from_table_or_query`) continue to work via the
  fallback `contextmanager` wrapper.

### DIFF
--- a/sdk/python/feast/infra/offline_stores/contrib/trino_offline_store/trino.py
+++ b/sdk/python/feast/infra/offline_stores/contrib/trino_offline_store/trino.py
@@ -1,6 +1,8 @@
+import contextlib
+import logging
 import uuid
 from datetime import date, datetime
-from typing import Any, Dict, List, Literal, Optional, Tuple, Union
+from typing import Any, Callable, ContextManager, Dict, Iterator, List, Literal, Optional, Tuple, Union
 
 import numpy as np
 import pandas as pd
@@ -36,6 +38,8 @@ from feast.infra.registry.base_registry import BaseRegistry
 from feast.on_demand_feature_view import OnDemandFeatureView
 from feast.repo_config import FeastConfigBaseModel, RepoConfig
 from feast.saved_dataset import SavedDatasetStorage
+
+logger = logging.getLogger(__name__)
 
 
 class BasicAuthModel(FeastConfigBaseModel):
@@ -177,14 +181,23 @@ class TrinoOfflineStoreConfig(FeastConfigBaseModel):
 class TrinoRetrievalJob(RetrievalJob):
     def __init__(
         self,
-        query: str,
+        query: Union[str, Callable[[], ContextManager[str]]],
         client: Trino,
         config: RepoConfig,
         full_feature_names: bool,
         on_demand_feature_views: Optional[List[OnDemandFeatureView]] = None,
         metadata: Optional[RetrievalMetadata] = None,
     ):
-        self._query = query
+        if not isinstance(query, str):
+            self._query_generator = query
+        else:
+
+            @contextlib.contextmanager
+            def query_generator() -> Iterator[str]:
+                assert isinstance(query, str)
+                yield query
+
+            self._query_generator = query_generator
         self._client = client
         self._config = config
         self._full_feature_names = full_feature_names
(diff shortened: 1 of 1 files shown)


Jwrede:
Friendly ping @woop -- this has been open ~6 days. Happy to address any feedback.

ntkathole:
@Jwrede can you please fix linting here as well?

Jwrede:
@ntkathole Fixed -- rebased onto master and resolved the ruff import sorting issue.

sdk/python/feast/infra/offline_stores/contrib/trino_offline_store/trino.py
```diff
@@ -201,17 +225,19 @@ def on_demand_feature_views(self) -> List[OnDemandFeatureView]:
...
 
     def _to_arrow_internal(self, timeout: Optional[int] = None) -> pyarrow.Table:
         """Return payrrow dataset as synchronously including on demand transforms"""
         return pyarrow.Table.from_pandas(self._to_df_internal(timeout=timeout))
 
     def to_sql(self) -> str:
         """Returns the SQL query that will be executed in Trino to build the historical feature table"""
-        return self._query
+        with self._query_generator() as query:
+            return query
```
ntkathole: Anyone who calls `to_sql()` for logging, debugging, or query inspection before materializing results will now silently destroy the temp table, causing subsequent calls to fail with a "table not found" error.

```
job = store.get_historical_features(entity_df=my_dataframe, ...)
sql = job.to_sql()       # table gets dropped here on context exit
df = job.to_df()         # FAILS — temp table no longer exists
```

ntkathole: same for 

```
First to_df() → succeeds → drops table
Second to_df() → query references a table that's gone → fails
```

Jwrede: Good catch -- redesigned the cleanup approach. The context manager pattern was wrong here since every method exit triggered a table drop.

Changes:
- Query is now stored as a plain string; to_sql() returns it directly with no side effects
- Cleanup moved to a dedicated _drop_temp_table() method, called only after actual query execution (to_df, to_trino)
- _cleaned_up flag makes the drop idempotent -- safe to call multiple times
- __del__ fallback ensures cleanup if execution methods are never called

ntkathole (approved):

PR #6395, opened 2026-05-12
fix(compute-engine/local): Honor field_mapping on join keys in dedup + join nodes

### LINKED ISSUE #5942: Materialization fails when field mappings are used to rename entity join keys (SnowflakeSource)
## Expected Behavior 
Features materialized to online store successfully, with features persisted on the online store using `field_mapping` *values* for the entity join key as names.

## Current Behavior
Materialization fails before the offline store is queried: `KeyError: Index(['USERID'], dtype='object')`
Only tested on `SnowflakeSource`

## Steps to reproduce
Define a feature source as 
```sql
create view schema_name.test as (
    select 
        current_timestamp() as EVENT_TIMESTAMP,
        1 as USERID,
        0.75 as FEATURE_1,
        300 as FEATURE_2
);
```

```python
user = Entity(name="user", join_keys=["user_id"], value_type=ValueType.STRING)

feature_source = SnowflakeSource(
    table="TEST",
    name="test_features",
    timestamp_field="EVENT_TIMESTAMP",
    database=dbname,
    schema=schema,
    field_mapping={
        "USERID": "user_id",
        "FEATURE_1": "feature_1",
        "FEATURE_2": "feature_2",
    },
)

feature_view = FeatureView(
    name="test_feature_view",
    entities=[user],
    schema=[
        Field(name="feature_1", dtype=types.Float32),
        Field(name="feature_2", dtype=types.Int32),
    ],
    online=True,
    source=feature_source,
)
```

### Specifications

- Version: 0.58.0
- Platform: macOS
- Subsystem: SQLite registry, SnowflakeSource

## Inconsistencies

* `get_historical_features` already works with the above configuration. Materialization not working in the same way causes inconsistent and confusing behavior.
  * Using `user_id` (field_mapping value) as an input to `get_historical_features`) The offline store is queried with "USERID" (field_mapping key), and the return value once again contains `user_id`. 

## Discussion

`join_keys` arg when defining an `Entity` already provides a way to map offline store column names to feast registry field/key names, for entity keys. If this is considered the de facto way to manage source column names, than this issue can be closed. 

However I think its worth exploring. Having the renaming done at entity level means all tables in the offline store needs to use the same column name for that entity key. There might be cases where different teams refer to the same entity with different names, just because the business context is different.

1fanwang:
**What this PR does**

Fixes #5942.

The bug isn't actually Snowflake-specific despite the issue title — it lives in the local compute engine's DAG nodes.

Path:

1. `_get_column_names()` returns the *reverse-mapped* join keys (e.g. `["USERID"]`).
2. `pull_latest_from_table_or_query` is called with those raw names — correct, the offline store needs them.
3. `LocalSourceReadNode.execute` renames the result columns via `field_mapping.get(col, col)` — the table now has `user_id`.
4. `LocalDedupNode` and `LocalJoinNode` then look up `column_info.join_keys` (still `["USERID"]`) on a df whose columns are `user_id`. pandas raises `KeyError: Index(['USERID'], dtype='object')` — the exact error in the issue.

The existing `ColumnInfo` class already has mapped properties for `timestamp_column` and `created_timestamp_column` (precedent: PR #4886, which fixed the analogous bug for `get_historical_features`). This PR adds the missing `join_keys_columns` mapped property and uses it in `LocalDedupNode` and `LocalJoinNode`.

**How was this tested?**

- New regression test `test_local_dedup_node_with_field_mapping_on_join_key` in `tests/unit/infra/compute_engines/local/test_nodes.py`. Reproduces the exact `KeyError(['USERID'])` on master; passes with this PR.
- All 9 existing tests in `test_nodes.py` still pass.
- All 62 tests under `tests/unit/infra/compute_engines/` still pass.
- `ruff check`, `ruff format --check`, `mypy` all clean.

**Scope**

Local-engine only. The Spark, Ray, AWS Lambda, k8s, and Snowflake compute engines have the same `column_info.join_keys` lookup pattern in their nodes and likely the same latent bug. Each is a separate <50 LoC follow-up — happy to file those if maintainers want them in this PR or as stacked.

### DIFF
--- a/sdk/python/feast/infra/compute_engines/dag/context.py
+++ b/sdk/python/feast/infra/compute_engines/dag/context.py
@@ -42,6 +42,18 @@ class ColumnInfo:
         """
         return self._get_mapped_column(self.created_ts_col)
 
+    @property
+    def join_keys_columns(self) -> List[str]:
+        """
+        Get the join keys, mapped through field_mapping to their post-rename
+        column names. Use this when looking up columns on a DataFrame that has
+        already had its source columns renamed (e.g. inside DAG nodes that
+        consume the output of a source-read node).
+        """
+        if not self.field_mapping:
+            return list(self.join_keys)
+        return [self.field_mapping.get(key, key) for key in self.join_keys]
+
     def _get_mapped_column(self, column: Optional[str]) -> Optional[str]:
         """
         Helper method to get the mapped column name if it exists in field_mapping.
--- a/sdk/python/feast/infra/compute_engines/local/nodes.py
+++ b/sdk/python/feast/infra/compute_engines/local/nodes.py
@@ -79,6 +79,10 @@ class LocalJoinNode(LocalNode):
         for val in input_values:
             val.assert_format(DAGFormat.ARROW)
 
+        # The upstream source-read node has already renamed columns via
+        # field_mapping, so use the mapped join keys for joining (see #5942).
+        join_keys = self.column_info.join_keys_columns
+
         # Convert all upstream ArrowTables to backend DataFrames
         joined_df = self.backend.from_arrow(input_values[0].data)
         for val in input_values[1:]:
@@ -86,7 +90,7 @@ class LocalJoinNode(LocalNode):
             joined_df = self.backend.join(
                 joined_df,
                 next_df,
-                on=self.column_info.join_keys,
+                on=join_keys,
                 how=self.how,
             )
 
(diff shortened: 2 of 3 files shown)


haoxu0:
LGTM
The initial work didn't handle the field mapping because I was thinking to unify the field mapping interface. But this never happened.

HaoXuAI (approved):

PR #6414, opened 2026-05-18
feat: REST API-backed UI for RBAC compatibility and per-page lazy loading

ntkathole:

The Feast UI currently serves the entire registry as a single serialized protobuf blob
via `GET /registry`. The frontend downloads this monolithic blob on every page load,
decodes it client-side, and extracts the slice it needs. This approach has two
fundamental problems:
- RBAC is completely bypassed
- Performance degrades with registry size

This PR replaces the protobuf blob approach entirely, making the REST API the **only**
data source for the Feast UI. **`feast ui`** now mounts the REST registry API at `/api/v1` (via `RegistryServer` +
  `register_all_routes`) and the frontend fetches data exclusively through these
  endpoints.

### DIFF
--- a/sdk/python/feast/cli/ui.py
+++ b/sdk/python/feast/cli/ui.py
@@ -2,6 +2,8 @@ import click
 
 from feast.repo_operations import create_feature_store, registry_dump
 
+VALID_MODES = ("proto", "rest", "rest-external")
+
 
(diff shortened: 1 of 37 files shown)


Vperiodt:
tested locally, working !

PR #6415, opened 2026-05-18
fix: Replace selector label strip patch with migration Job for upgrade-safe selector uniqueness

ntkathole:

- Replaces the remove_selector_label_patch.yaml workaround (which stripped app.kubernetes.io/name from the Deployment selector) with a pre-upgrade migration Job that deletes the existing feast-operator-controller-manager Deployment if its spec.selector doesn't match the expected labels.
- Preserves app.kubernetes.io/name: feast-operator label in the Deployment selector for label uniqueness, while ensuring clean upgrades from all prior versions.


Srihari1192 (approved):
lgtm

PR #6420, opened 2026-05-20
fix: Replace registry proto dump with RBAC-compliant APIs in UI and feature servers

ntkathole:

- Replace `store.registry.proto() `usage across UI server, feature server, and registry gRPC server with granular, RBAC-aware registry APIs (list_projects, etc.)
- Add permission checks (UPDATE) to the Commit and Refresh gRPC RPCs in the registry server
- Remove redundant async_refresh timer from the UI server (registry's built-in cache TTL handles freshness)
- Clean up unused parameters (get_registry_dump, --registry_ttl_sec) from the UI server path

## Changes

**UI server:** Use list_projects(allow_cache=True) for project discovery and health checks instead of fetching the full registry proto
**Feature server:** Simplified health endpoint to use list_projects; removed proto caching from the refresh loop
**Registry gRPC server:** Removed Proto RPC; added authorization to Commit and Refresh RPCs
**CLI:** Removed --registry_ttl_sec flag from feast ui (no longer applicable)


aniketpalu (approved):

sdk/python/feast/infra/feature_servers/multicloud/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/python-312-minimal:1
+FROM registry.access.redhat.com/ubi9/python-312-minimal:latest
```
ntkathole, line 1: CI Operator E2E tests are failing across all PRs since ~May 19 with deployment timeouts.
The root cause is a runtime `ImportError` in the `feast-apply` init container.

The Dockerfiles use `ubi9/python-312-minimal:1` which resolves to a RHEL 9.6 image
with Python 3.12.9 and `sqlite-libs-3.34.1-8.el9_6`. When `python3.12-devel` is
installed via `microdnf`, it upgrades `python3.12-libs` to 3.12.13 (from RHEL 9.8
repos). This newer build's `_sqlite3.so` was compiled expecting `sqlite3_deserialize`
(SQLite 3.36.0+), but the 9.6 `sqlite-libs` (3.34.1) doesn't provide it.

The `:latest` / `:9.8` image ships with Python 3.12.13 and a matching
`sqlite-libs-3.34.1-10.el9_8` that are compatible. Installing `python3.12-devel`
on this image does not trigger any upgrade since the versions already match.

PR #6432, opened 2026-05-22
fix(go): Skip registry refresh when cache_ttl_seconds <= 0 and avoid NewTicker panic

samuelkim7:

The Go feature server panics on startup when `cache_ttl_seconds <= 0` is set in `feature_store.yaml`:

```
panic: non-positive interval for NewTicker

goroutine 84 [running]:
time.NewTicker(...)
github.com/feast-dev/feast/go/internal/feast/registry.(*Registry).RefreshRegistryOnInterval
    go/internal/feast/registry/registry.go:83
```

Because the ticker is created in a goroutine started during `InitializeRegistry`, an unrecovered panic there terminates the entire Go process — the feature server pod enters `CrashLoopBackOff` and serves no traffic.

Python Feast treats `cache_ttl_seconds <= 0` as "do not refresh the cache", and the Go side's `getRegistryProto` already guards on `cachedRegistryProtoTtl > 0` for expiration checks. However, `RefreshRegistryOnInterval` passed the value directly to `time.NewTicker`, which panics for non-positive durations.

This PR skips the refresh loop when TTL is non-positive, preventing the `time.NewTicker` panic and aligning Go behavior with Python.

# Checks

## Testing Strategy
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

Added `TestRefreshRegistryOnIntervalNonPositiveTTL` covering `ttl = 0` and `ttl = -1s` via `assert.NotPanics`. Verified the test fails with the original panic message when the fix is reverted.

Also verified on a live Go feature server pod running with `cache_ttl_seconds: 0`: the pod starts cleanly and serves traffic without crashing.

# Misc

### DIFF
--- a/go/internal/feast/registry/registry.go
+++ b/go/internal/feast/registry/registry.go
@@ -81,6 +81,10 @@ func (r *Registry) InitializeRegistry() error {
 }
 
 func (r *Registry) RefreshRegistryOnInterval() {
+	if r.cachedRegistryProtoTtl <= 0 {
+		log.Info().Msg("Registry cache TTL is non-positive; skipping periodic refresh")
+		return
+	}
 	ticker := time.NewTicker(r.cachedRegistryProtoTtl)
 	for ; true; <-ticker.C {
 		err := r.refresh()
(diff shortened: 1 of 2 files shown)


samuelkim7:
Python CI failure seems pre-existing, unrelated to this Go-only change.

ntkathole (approved):

PR #6440, opened 2026-05-27
feat(spark): SparkSource query+path and pre-computed offline read for BatchFeatureView

abhijeet-dhumal:

`get_historical_features()` on a `BatchFeatureView` re-runs the full UDF on raw data every call. For embedding pipelines, that's 20–40 min of compute per training run even though features already exist from the last `materialize`.

**Fix:** Route `get_historical_features()` to read pre-computed parquet from `batch_source.path` instead of re-executing the UDF.

To support this, `SparkSource` now accepts `query + path` together:
- `query` — raw data read during `materialize()`
- `path` — write-back target and pre-computed read source for `get_historical_features()`

```python
SparkSource(
    query="SELECT id, text, event_timestamp FROM bronze.documents",
    path="s3://my-bucket/feast/features/document_embeddings/",
)
```

Also allows `BatchFeatureView` with `online=False, offline=True` (offline-only) to skip the online validation check in `get_historical_features()`, so it can be used purely for training data without configuring an online store.

Falls back to live query if `path` doesn't exist yet (first run before any materialization).

N/A. Enables efficient training data retrieval for `BatchFeatureView` embedding pipelines without re-running UDFs.

# Checks

## Testing Strategy
- [x] Unit tests — offline path routing, SparkSource constraint, graceful fallback
- [x] Manual tests — `get_historical_features()` reads from parquet, not UDF, after materialization

### DIFF
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -1296,9 +1296,10 @@ class FeatureStore:
                         f"Enable it before materializing."
                     )
                 if hasattr(feature_view, "online") and not feature_view.online:
-                    raise ValueError(
-                        f"FeatureView {feature_view.name} is not configured to be served online."
-                    )
+                    if not getattr(feature_view, "offline", False):
+                        raise ValueError(
+                            f"FeatureView {feature_view.name} is not configured to be served online."
+                        )
                 elif (
                     hasattr(feature_view, "write_to_online_store")
                     and not feature_view.write_to_online_store
--- a/sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
+++ b/sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
@@ -131,10 +131,17 @@ def _apply_bfv_transformations_for_historical(
                 ctx = replace(ctx, table_subquery=tmp_view)
                 new_contexts.append(ctx)
                 continue
-            except Exception:
+            except (FileNotFoundError, PermissionError) as e:
                 warnings.warn(
-                    f"Offline path '{fv.batch_source.path}' not readable for "
-                    f"'{ctx.name}'; falling back to source query.",
+                    f"Offline path '{fv.batch_source.path}' not accessible for "
+                    f"'{ctx.name}': {e}; falling back to source query.",
+                    RuntimeWarning,
+                    stacklevel=2,
+                )
+            except Exception as e:
+                warnings.warn(
+                    f"Unexpected error loading offline path '{fv.batch_source.path}' "
+                    f"for '{ctx.name}': {e}; falling back to source query.",
                     RuntimeWarning,
                     stacklevel=2,
                 )
--- a/sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
+++ b/sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
@@ -92,6 +92,72 @@ class SparkFeatureViewQueryContext(offline_utils.FeatureViewQueryContext):
     max_date_partition: Optional[str]
 
 
+def _apply_bfv_transformations_for_historical(
+    spark_session: SparkSession,
+    feature_views: List[FeatureView],
+    query_context: List[offline_utils.FeatureViewQueryContext],
+) -> List[offline_utils.FeatureViewQueryContext]:
+    """
+    For BatchFeatureViews, redirect get_historical_features to read from the
+    pre-materialized offline store (batch_source.path) when available, avoiding
+    expensive UDF re-execution on raw data.
+
+    Precedence:
+      1. offline=True + batch_source.path set -> read pre-computed parquet
+      2. Python/pandas UDF present -> execute UDF on raw source (fallback)
+      3. Otherwise -> pass through unchanged
+    """
+    from dataclasses import replace
+
+    fv_by_name = {fv.projection.name_to_use(): fv for fv in feature_views}
+    new_contexts = []
+
+    for ctx in query_context:
+        fv = fv_by_name.get(ctx.name)
+        if fv is None or not isinstance(fv, BatchFeatureView):
+            new_contexts.append(ctx)
+            continue
+
+        if (
+            getattr(fv, "offline", False)
+            and isinstance(fv.batch_source, SparkSource)
+            and fv.batch_source.path
+        ):
+            tmp_view = f"__feast_offline_{ctx.name}_{uuid.uuid4().hex[:8]}"
+            file_format = fv.batch_source.file_format or "parquet"
+            df = spark_session.read.format(file_format).load(fv.batch_source.path)
+            df.createOrReplaceTempView(tmp_view)
+            ctx = replace(ctx, table_subquery=tmp_view)
+        elif (
+            hasattr(fv, "feature_transformation")
+            and fv.feature_transformation is not None
+            and (
+                getattr(fv.feature_transformation, "mode", None)
+                in ("python", "pandas")
+                or getattr(
+                    getattr(fv.feature_transformation, "mode", None), "value", None
+                )
+                in ("python", "pandas")
+            )
+        ):
+            udf = getattr(fv.feature_transformation, "udf", None) or getattr(
+                fv, "udf", None
+            )
+            if udf is not None:
+                temp_view_name = f"__feast_bfv_{ctx.name}_{uuid.uuid4().hex[:8]}"
+                spark_session.conf.set("spark.sql.runSQLOnFiles", "true")
+                raw_df = spark_session.sql(
+                    f"SELECT * FROM {ctx.table_subquery}"
+                )
+                transformed_df = udf(raw_df)
+                transformed_df.createOrReplaceTempView(temp_view_name)
+                ctx = replace(ctx, table_subquery=temp_view_name)
+
+        new_contexts.append(ctx)
+
+    return new_contexts
+
+
 class SparkOfflineStore(OfflineStore):
     @staticmethod
     def pull_latest_from_table_or_query(
@@ -298,8 +364,10 @@ class SparkOfflineStore(OfflineStore):
             entity_df_event_timestamp_range,
         )
 
-        query_context = _apply_bfv_transformations(
-            spark_session, feature_views, query_context
+        query_context = _apply_bfv_transformations_for_historical(
+            spark_session=spark_session,
+            feature_views=feature_views,
+            query_context=query_context,
         )
 
         spark_query_context = [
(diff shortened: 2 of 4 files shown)


sdk/python/feast/feature_store.py
```diff
@@ -1053,9 +1053,10 @@ def _get_feature_views_to_materialize(
                         f"Enable it before materializing."
                     )
                 if hasattr(feature_view, "online") and not feature_view.online:
-                    raise ValueError(
-                        f"FeatureView {feature_view.name} is not configured to be served online."
-                    )
+                    if not getattr(feature_view, "offline", False):
+                        raise ValueError(
+                            f"FeatureView {feature_view.name} is not configured to be served online."
+                        )
```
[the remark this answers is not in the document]
abhijeet-dhumal: This is actually handled correctly - the online/offline branching doesn't happen at the provider level, it happens inside SparkComputeEngine's execution plan in sdk/python/feast/infra/compute_engines/spark/nodes.py

abhijeet-dhumal:
@ntkathole May I request your review here too ?

sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
```diff
@@ -79,6 +79,80 @@ class SparkFeatureViewQueryContext(offline_utils.FeatureViewQueryContext):
...
+            and isinstance(fv.batch_source, SparkSource)
+            and fv.batch_source.path
+        ):
+            tmp_view = f"__feast_offline_{ctx.name}_{uuid.uuid4().hex[:8]}"
+            file_format = fv.batch_source.file_format or "parquet"
+            try:
+                df = spark_session.read.format(file_format).load(fv.batch_source.path)
+                df.createOrReplaceTempView(tmp_view)
+                ctx = replace(ctx, table_subquery=tmp_view)
+                new_contexts.append(ctx)
```
jyejare: **[Critical] Path traversal vulnerability in file loading**

The code directly loads files from fv.batch_source.path without any validation. This creates a path traversal security vulnerability where malicious paths could access unauthorized files on the system.


**Suggested:**
```suggestion
+            # Validate and sanitize the path
+            import os
+            normalized_path = os.path.normpath(fv.batch_source.path)
+            if '..' in normalized_path or normalized_path.startswith('/'):
+                warnings.warn(f"Invalid path '{fv.batch_source.path}' for '{ctx.name}'", RuntimeWarning)
+                new_contexts.append(ctx)
+                continue
+            try:
+                df = spark_session.read.format(file_format).load(normalized_path)
+                df.createOrReplaceTempView(tmp_view)
+                ctx = replace(ctx, table_subquery=tmp_view)
+                new_contexts.append(ctx)
+                continue
```

sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
```diff
@@ -79,6 +79,80 @@ class SparkFeatureViewQueryContext(offline_utils.FeatureViewQueryContext):
...
+                continue
+            except Exception:
+                warnings.warn(
+                    f"Offline path '{fv.batch_source.path}' not readable for "
+                    f"'{ctx.name}'; falling back to source query.",
+                    RuntimeWarning,
```
jyejare: **[Warning] Overly broad exception handling masks specific errors**

Catching all exceptions with 'except Exception:' is too broad and could mask important errors like permission issues, file corruption, or configuration problems. This makes debugging difficult and could hide security issues.
**Suggested:**
```suggestion
+            except (FileNotFoundError, PermissionError) as e:
+                warnings.warn(
+                    f"Offline path '{fv.batch_source.path}' not accessible for "
+                    f"'{ctx.name}': {str(e)}; falling back to source query.",
+                    RuntimeWarning,
+                    stacklevel=2,
+                )
+            except Exception as e:
+                # Log unexpected errors but continue with fallback
+                import logging
+                logging.warning(f"Unexpected error loading '{fv.batch_source.path}': {e}")
+                warnings.warn(
+                    f"Failed to load offline path for '{ctx.name}'; falling back to source query.",
+                    RuntimeWarning,
+                    stacklevel=2,
+                )
```

sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
```diff
@@ -79,6 +79,80 @@ class SparkFeatureViewQueryContext(offline_utils.FeatureViewQueryContext):
...
+                in ("python", "pandas")
+            )
+        ):
+            udf = getattr(fv.feature_transformation, "udf", None) or getattr(
+                fv, "udf", None
+            )
+            if udf is not None:
+                temp_view_name = f"__feast_bfv_{ctx.name}_{uuid.uuid4().hex[:8]}"
+                spark_session.conf.set("spark.sql.runSQLOnFiles", "true")
+                raw_df = spark_session.sql(f"SELECT * FROM {ctx.table_subquery}")
```
jyejare: **[Warning] UDF execution without sandboxing poses security risks**

Executing user-defined functions without proper sandboxing or validation could allow arbitrary code execution. This is a significant security risk in production environments.

**Suggested:**
```suggestion
+                # Add UDF validation and execution controls
+                if not hasattr(udf, '__call__'):
+                    raise ValueError(f"Invalid UDF for {ctx.name}")
+                raw_df = spark_session.sql(f"SELECT * FROM {ctx.table_subquery}")
+                # Consider adding timeout and resource limits here
+                transformed_df = udf(raw_df)
+                transformed_df.createOrReplaceTempView(temp_view_name)
```

jyejare (commented):
This PR adds support for SparkSource with combined query+path configuration and pre-computed offline reads for BatchFeatureView. The changes enable reading from materialized offline stores to avoid expensive UDF re-execution. While the feature is useful, there are several security vulnerabilities and error handling gaps that need attention.

sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
abhijeet-dhumal: Thanks for flagging this. fv.batch_source.path is a trusted configuration value set by the feature store admin at definition time — it's not runtime user input flowing from an API boundary, so path traversal in the traditional web security sense doesn't apply here.

abhijeet-dhumal: Also, the suggested fix using os.path.normpath() would silently corrupt S3/GCS URIs (s3://bucket/path - s3:/bucket/path) since normpath collapses double slashes. Happy to add a lightweight guard for local paths only if you'd like, but I'd keep it separate from object storage paths.

sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
abhijeet-dhumal: Good catch — updated to catch `FileNotFoundError` and `PermissionError` explicitly for the expected fallback cases, with a separate `except Exception` block emitting a distinct warning so unexpected errors aren't silently swallowed.

sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
abhijeet-dhumal: The UDF here is `feature_transformation.udf` registered in the Feast registry by the ML engineer at feature view definition time — same function that runs during `materialize()`. It's not arbitrary user input. The `hasattr(udf, '__call__')` check doesn't add a meaningful boundary since any Python object can implement `__call__`. Executor sandboxing is a cluster-level concern (Spark resource limits, executor isolation) — outside Feast's scope.

sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
jyejare: Make Sense.

jyejare (approved):

sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
```diff
@@ -261,8 +335,10 @@ def get_historical_features(
             entity_df_event_timestamp_range,
         )
 
-        query_context = _apply_bfv_transformations(
-            spark_session, feature_views, query_context
+        query_context = _apply_bfv_transformations_for_historical(
```
ntkathole: so previous `_apply_bfv_transformations` helper not removed from code?

sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
```diff
@@ -79,6 +79,80 @@ class SparkFeatureViewQueryContext(offline_utils.FeatureViewQueryContext):
...
+            except Exception:
+                warnings.warn(
+                    f"Offline path '{fv.batch_source.path}' not readable for "
+                    f"'{ctx.name}'; falling back to source query.",
+                    RuntimeWarning,
+                    stacklevel=2,
+                )
+
+        if (
+            hasattr(fv, "feature_transformation")
```
ntkathole: Use has_transformation() and get_transformation_function() instead

sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
```diff
@@ -79,6 +79,80 @@ class SparkFeatureViewQueryContext(offline_utils.FeatureViewQueryContext):
     max_date_partition: str
 
 
+def _apply_bfv_transformations_for_historical(
```
ntkathole: I think instead of writing new helper, better approach is to extend the existing `_apply_bfv_transformations` with a pre-computed-path shortcut (adding the "read from parquet if offline=True and path exists" branch at the top), rather than creating a parallel function that reimplements UDF execution.

sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
abhijeet-dhumal: Good catch 🙌 
the previous version had a separate _apply_bfv_transformations_for_historical that duplicated UDF execution logic. Removed it and folded the pre-computed path shortcut directly into _apply_bfv_transformations as the first branch. The call site now uses the single function. 
Please review !

sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
abhijeet-dhumal: Done ✅ 
The UDF detection now uses has_transformation() and get_transformation_function() from feast.feature_view_utils, same as the existing _apply_bfv_transformations already did. The raw hasattr/getattr chains for feature_transformation are gone.

sdk/python/feast/infra/offline_stores/contrib/spark_offline_store/spark.py
abhijeet-dhumal: Agreed 🙌 
Refactored exactly as you suggested. _apply_bfv_transformations_for_historical is deleted. _apply_bfv_transformations now has two branches:

1. Pre-computed path shortcut (offline=True + batch_source.path) which reads parquet directly, falls back on error
2. UDF execution via has_transformation() / get_transformation_function() which is unchanged from upstream

No duplication now.. Please review !

ntkathole (approved):

PR #6454, opened 2026-05-29
fix: Detect duplicate feature view/data source names early in parse_repo to fix flaky test (#6417)

### LINKED ISSUE #6417: Flaky test: test_cli_apply_duplicated_featureview_names fails with empty output on subprocess timeout
## Summary

`test_cli_apply_duplicated_featureview_names` is intermittently failing in CI (observed on parallel pytest-xdist workers):

```
FAILED sdk/python/tests/unit/cli/test_cli_apply_duplicates.py::test_cli_apply_duplicated_featureview_names
  AssertionError: assert (-1 != 0 and b'Feature view names must be case-insensitively unique' in b'')
```

The return code `-1` and empty output `b''` indicate the subprocess timed out and was killed before producing any output.

## Root Cause

Duplicate feature view name detection happens too late in the pipeline — inside `store.plan()` / `store.apply()` via `_validate_feature_views()` in `feature_store.py`. By that point, `FeatureStore` and its dependencies (Dask, registry, provider) have already been initialized.

The call chain is:

1. CLI `apply_total_command` calls `apply_total()`, which calls `_get_repo_contents()` (parses repo), then `_get_store_and_registry()` (creates FeatureStore), then `apply_total_with_repo_instance()`, then `store.plan()`, then `_validate_all_feature_views()`, which finally raises `ConflictingFeatureViewNames`.
2. The CLI only catches `FeastProviderLoginError`, so `ConflictingFeatureViewNames` propagates as an **unhandled exception** (raw traceback).
3. During process shutdown, slow atexit handlers (Dask thread pool, PySpark JVM) can block indefinitely.
4. The test helper `CliRunner.run_with_output()` has a 60s timeout — when exceeded, it sends SIGKILL and the output is lost.

This creates two compounding problems:
- **The error is raised too late** — after heavy initialization that can trigger slow cleanup on exit
- **The CLI does not handle the error** — producing a traceback instead of a clean message, and not exiting promptly

## Proposed Fix

1. **Detect duplicate names at parse time in `parse_repo()`** (repo_operations.py): Track feature view names (across FeatureView, BatchFeatureView, StreamFeatureView, OnDemandFeatureView) and data source names during collection. Raise `ConflictingFeatureViewNames` / `DataSourceRepeatNamesException` immediately when a collision is found — before FeatureStore or any heavy dependencies are initialized.

2. **Catch `FeastError` in CLI plan/apply commands** (cli/cli.py): Add `except FeastError` handlers (after the existing `FeastProviderLoginError` handler) to print a clean error message and exit with code 1, instead of letting validation errors propagate as unhandled tracebacks.

## Environment

- Observed on macOS CI runners with pytest-xdist parallel execution ([gw2])
- Python 3.11.9

## Affected Tests

- `test_cli_apply_duplicated_featureview_names`
- `test_cli_apply_duplicate_data_source_names` (same pattern, same risk)
- `test_cli_apply_duplicated_featureview_names_multiple_py_files` (same pattern, same risk)

obielin:

Fixes the flaky test `test_cli_apply_duplicated_featureview_names` (and similar tests) that intermittently fails in CI with empty output due to subprocess timeout.

**Root cause:** Duplicate feature view name detection happened too late in the pipeline — inside `store.plan()` / `store.apply()` via `_validate_feature_views()`, after `FeatureStore` and heavy dependencies (Dask, registry, provider) had already been initialized. When the validation error was raised, slow `atexit` handlers (Dask thread pool) could block process exit indefinitely, causing the 60s timeout to trigger and the output to be lost.

### `sdk/python/feast/repo_operations.py`
- Added `ConflictingFeatureViewNames` and `DataSourceRepeatNamesException` imports from `feast.errors`
- - Added **early duplicate detection** in `parse_repo()` before any `FeatureStore` initialization:
-   - Validates that all feature view names (across `FeatureView`, `StreamFeatureView`, `OnDemandFeatureView`) are case-insensitively unique → raises `ConflictingFeatureViewNames` immediately if collision found
-   - Validates that all data source names are case-insensitively unique → raises `DataSourceRepeatNamesException` immediately if collision found
### `sdk/python/feast/cli/cli.py`
- Added `import sys`
- - Added `FeastError` to the import from `feast.errors`
- - Added `except FeastError` handler (after existing `FeastProviderLoginError`) in both `plan_command` and `apply_total_command` — prints a clean error message and exits with code 1 instead of letting validation errors propagate as unhandled tracebacks
## How has this been tested

- The existing integration tests in `sdk/python/tests/integration/cli/test_cli_apply_duplicates.py` cover the fix:
-   - `test_cli_apply_duplicated_featureview_names`
-   - `test_cli_apply_duplicate_data_source_names`
-   - `test_cli_apply_duplicated_featureview_names_multiple_py_files`
These tests assert `rc != 0` and the expected error message in the output, which is now guaranteed since the error is raised before any heavy initialization.

## Closes

Closes #6417

### DIFF
--- a/sdk/python/feast/cli/cli.py
+++ b/sdk/python/feast/cli/cli.py
@@ -12,8 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import json
-import sys
 import logging
+import sys
 from datetime import datetime
 from importlib.metadata import version as importlib_version
 from pathlib import Path
--- a/sdk/python/feast/cli/cli.py
+++ b/sdk/python/feast/cli/cli.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import json
+import sys
 import logging
 from datetime import datetime
 from importlib.metadata import version as importlib_version
@@ -49,7 +50,7 @@ from feast.cli.stream_feature_views import stream_feature_views_cmd
 from feast.cli.ui import ui
 from feast.cli.validation_references import validation_references_cmd
 from feast.constants import FEAST_FS_YAML_FILE_PATH_ENV_NAME
-from feast.errors import FeastProviderLoginError
+from feast.errors import FeastError, FeastProviderLoginError
 from feast.repo_config import load_repo_config
 from feast.repo_operations import (
     apply_total,
@@ -258,6 +259,9 @@ def plan_command(
         plan(repo_config, repo, skip_source_validation, skip_feature_view_validation)
     except FeastProviderLoginError as e:
         print(str(e))
+    except FeastError as e:
+        print(str(e))
+        sys.exit(1)
 
 
 @cli.command("apply", cls=NoOptionDefaultFormat)
@@ -316,6 +320,9 @@ def apply_total_command(
         )
     except FeastProviderLoginError as e:
         print(str(e))
+    except FeastError as e:
+        print(str(e))
+        sys.exit(1)
 
 
 @cli.command("teardown", cls=NoOptionDefaultFormat)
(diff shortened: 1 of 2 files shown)


franciscojavierarceo:
@Coded review

obielin:
Closing in favour of a new PR (#6455) that has proper `Signed-off-by` on all commits to satisfy the DCO requirement. The code changes are identical.

PR #6455, opened 2026-05-30
fix: Add checks for registry and feature server handlers

ntkathole:

For consistency across all registry and feature server endpoints, added missing inline auth checks that their sibling handlers already enforce. 

FeastObjectNotFoundException -> Yes → pass -> return empty list

FeastPermissionError -> No → propagates → 403

### DIFF
--- a/sdk/python/feast/api/registry/rest/rest_registry_server.py
+++ b/sdk/python/feast/api/registry/rest/rest_registry_server.py
@@ -235,28 +235,35 @@ class RestRegistryServer:
                     else:
                         object_type = None
                         object_name = None
-                    visit = {
-                        "path": path,
-                        "timestamp": _utc_now().isoformat(),
-                        "project": project,
-                        "user": user,
-                        "object": object_type,
-                        "object_name": object_name,
-                        "method": method,
-                    }
-                    try:
-                        visits_json = self.registry.get_project_metadata(project, key)
-                        visits = json.loads(visits_json) if visits_json else []
-                    except Exception:
-                        visits = []
-                    visits.append(visit)
-                    visits = visits[-self.recent_visits_limit :]
-                    try:
-                        self.registry.set_project_metadata(
-                            project, key, json.dumps(visits)
-                        )
-                    except Exception as e:
-                        logger.warning(f"Failed to persist recent visits: {e}")
+
+                    response = await call_next(request)
+
+                    if response.status_code < 400:
+                        visit = {
+                            "path": path,
+                            "timestamp": _utc_now().isoformat(),
+                            "project": project,
+                            "user": user,
+                            "object": object_type,
+                            "object_name": object_name,
+                            "method": method,
+                        }
+                        try:
+                            visits_json = self.registry.get_project_metadata(
+                                project, key
+                            )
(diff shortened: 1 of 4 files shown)


franciscojavierarceo (approved):

jyejare (commented):
## AI Code Review

This PR adds important authorization checks and improves error handling across registry and feature server handlers. The changes properly implement permission validation using the existing security framework and handle edge cases like non-existent projects gracefully.

---
*1 comments across 1 files*

---

**[Warning] Permission check should happen before accessing project metadata**
> File: `sdk/python/feast/api/registry/rest/metrics.py`
> Lines: 439-448

The permission check is performed after the project existence check but before accessing project metadata. However, the metadata access happens outside this block. The permission check should be moved to protect the entire metadata access operation.

**Current code:**
```diff
+            if server:
+                try:
+                    project_obj = server.registry.get_project(
+                        name=project, allow_cache=True
+                    )
+                    assert_permissions(
+                        resource=project_obj, actions=[AuthzedAction.DESCRIBE]
+                    )
+                except FeastObjectNotFoundException:
+                    pass
             try:
                 visits_json = (
```

**Suggested replacement:**
```suggestion
            if server:
                try:
                    project_obj = server.registry.get_project(
                        name=project, allow_cache=True
                    )
                    assert_permissions(
                        resource=project_obj, actions=[AuthzedAction.DESCRIBE]
                    )
                    # Only access metadata if permission check passes
                    visits_json = server.registry.get_project_metadata(project, key)
                    if visits_json:
                        visits = json.loads(visits_json)
                except FeastObjectNotFoundException:
                    pass
                except Exception:
                    pass
            elif project:
                # Handle case when server is None but project is specified
                pass
```

sdk/python/feast/api/registry/rest/rest_registry_server.py
```diff
@@ -235,28 +235,35 @@ async def dispatch(self, request: Request, call_next):
...
+                            visits = []
+                        visits.append(visit)
+                        visits = visits[-self.recent_visits_limit :]
+                        try:
+                            self.registry.set_project_metadata(
+                                project, key, json.dumps(visits)
+                            )
+                        except Exception as e:
+                            logger.warning(f"Failed to persist recent visits: {e}")
+                    return response
```
jyejare, line 266: **[Critical] Unreachable code after return statement**

There are three lines of unreachable code after the return statement. The response handling and call_next() execution will never be reached, which could break the middleware functionality.

**Current code:**
```diff
+                    return response
                 response = await call_next(request)
                 return response
```

**Suggested:**
```suggestion
                    return response
                # Remove the unreachable code below
```

jyejare (changes requested):
This PR adds important security and authorization checks to registry and feature server handlers, improving access control for projects and operations. The changes properly implement permission assertions and handle edge cases like missing projects, though there are some areas that need attention around error handling and code duplication.

sdk/python/feast/api/registry/rest/rest_registry_server.py
ntkathole: The suggestion to remove those lines would break the middleware for all non-GET requests.

PR #6459, opened 2026-06-01
fix: Remove selector migration job - no longer needed

ntkathole:

We can remove the selector_migration_job.yaml from the Feast operator overlays. The ODH operator fix makes it redundant - The `migrateDeploymentSelector` action added in opendatahub-operator runs before any manifests are applied, so the stale Deployment is already deleted.


Srihari1192 (approved):
LGTM Thanks

PR #6478, opened 2026-06-08
feat(redis): Add retrieve_online_documents_v2 support with Redis vect…

### LINKED ISSUE #6461: feat: Add retrieve_online_documents_v2 support to Redis online store

The Redis online store currently only supports online_read / online_write operations via get_online_features. There is no support for retrieve_online_documents or retrieve_online_documents_v2, which prevents users from performing vector similarity search (ANN) through the Feast SDK when using Redis as their online store.

 Redis 8 introduced Vector Sets as a native data structure (via the VSIM command), making it a first-class vector store without requiring additional modules like RedisSearch. Despite this, Feast users are forced to bypass the SDK and call Redis directly for ANN queries, losing the abstraction and consistency that Feast provides.

  Implement retrieve_online_documents_v2 in the Redis online store (sdk/python/feast/infra/online_stores/redis.py) with support for:

  - Vector similarity search using Redis 8 native Vector Sets (VSIM command)
  - COSINE and L2 distance metrics
  - top_k results
  - Optional metadata/payload filtering by entity key
  - Backward compatibility with older Redis versions (graceful error if Vector Sets unavailable)

  Expected usage:

```python
  results = store.retrieve_online_documents(
      feature="user_semantic_memory:perfil_embedding",
      query=query_embedding,  # np.ndarray float32
      top_k=10,
      distance_metric="COSINE",
  )
````

  - PostgreSQL + pgvector: Supports retrieve_online_documents_v2 natively but has higher latency and lower throughput compared to Redis for online serving.
  - Milvus: Full support via PR #4971, but adds significant operational overhead.
  - Direct Redis calls (current workaround): Bypasses the Feast SDK entirely, losing feature logging, consistency guarantees, and SDK abstractions.

  **Additional context**

  - Redis 8 Vector Sets docs: https://redis.io/blog/announcing-vector-sets-a-new-redis-data-type-for-vector-similarity
  - Existing retrieve_online_documents_v2 implementations that can serve as reference: milvus.py, mongodb.py, postgres.py
  - Related issue: #6445 (Qdrant support)
  - Redis 8 is GA and widely adopted in production environments, making this a high-impact addition for the Feast community.

### PR DESCRIPTION
**Pull request by falloficaruss**
Adds Redis 8 Vector Set support to the Redis online store for `retrieve_online_documents_v2`, so Feast can do ANN/vector similarity search without bypassing the SDK.

This PR:
- indexes vector fields into Redis Vector Sets using `VADD`
- queries vectors through `VSIM`
- supports `COSINE` and `L2` metrics from Feast vector metadata
- keeps vector-set state aligned with hash writes and TTL behavior
- cleans up vector-set members when entities/tables are deleted
- adds unit coverage for indexing, deletion, and retrieval behavior

Fixes #6461

# Checks

## Testing Strategy
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

### CODE REVIEW

**Review by franciscojavierarceo — COMMENTED** [review]
Left a few inline comments on the Redis vector-set implementation. The main blockers are around the claimed L2 semantics and Redis < 8 capability detection.

File: sdk/python/feast/infra/online_stores/redis.py [hunk 1]
```diff
@@ -114,24 +118,199 @@ class RedisOnlineStore(OnlineStore):
...
     def async_supported(self) -> SupportedAsyncMethods:
         return SupportedAsyncMethods(read=True, write=True)
 
+    def _check_vadd_supported(self, client: Union[Redis, RedisCluster]) -> bool:
+        """Check if the connected Redis server supports VADD (Redis 8+ Vector Sets)."""
+        if self._vadd_supported is not None:
+            return self._vadd_supported
+        try:
+            info = client.execute_command("COMMAND", "INFO", "VADD")
+            self._vadd_supported = info is not None and len(info) > 0
```
franciscojavierarceo: This can report VADD as supported on Redis versions that do not have it. `COMMAND INFO <name>` returns a nil entry for unknown commands, so a response shaped like `[None]` still has `len(info) > 0`; writes would then proceed into `VADD` and fail instead of cleanly skipping/raising the intended unsupported-server path. Can we inspect the actual command entry/value, not just the container length?

sdk/python/feast/infra/online_stores/redis.py
```diff
@@ -114,24 +118,199 @@ class RedisOnlineStore(OnlineStore):
...
+                    continue
+                python_vector = feast_value_type_to_python_type(vec_val)
+                if not isinstance(python_vector, Sequence) or isinstance(
+                    python_vector, (bytes, str)
+                ):
+                    continue
+                floats = [float(v) for v in python_vector]
+                if not floats:
+                    continue
+                if metric in {"COSINE", "L2"}:
```
franciscojavierarceo: This makes the advertised `L2` mode behave like cosine/angular search. Normalizing indexed vectors and the query discards magnitude, so vectors like `[2, 0]` and `[100, 0]` become equivalent even though true L2 distances are very different. Since `VSIM WITHSCORES` returns Redis's similarity score, Redis should either reject `L2` here or use a backend/query path that preserves real L2 semantics.

sdk/python/feast/infra/online_stores/redis.py
```diff
@@ -181,6 +360,10 @@ def delete_table(self, config: RepoConfig, table: FeatureView):
                 deleted_count += 1
             pipe.execute()
 
+        # Also delete the Vector Set key if it exists
```
franciscojavierarceo: This cleanup is skipped when `delete_table()` returns early because no hash keys were found. That can leave the vector set behind after hash TTL expiry or partial cleanup. Could we delete `vs_key` before the `if not all_keys: return` path, or otherwise ensure table cleanup always removes the vector set?

sdk/python/feast/infra/online_stores/redis.py
falloficaruss: If we prefer to try preserving true L2 semantics, we'd need to investigate whether Redis VSIM natively supports L2 and how to not normalize vectors in L2 mode, which is more complex. So, I will fix it by rejecting L2 here.

sdk/python/feast/infra/online_stores/redis.py
falloficaruss: I've moved the `vs_key` deletion before the early-return guard so the vector set is always cleaned up when `delete_table` is called, regardless of whether hash keys are still present. `DELETE` on a non-existent key is a no-op, so it's safe when no vector set exists either

Issue #6445, opened 2026-05-28
feat: Add retrieve_online_documents_v2 support to Qdrant online store


patelchaitany:
## What is missing

- `QdrantOnlineStore` has `retrieve_online_documents` (v1) only — no `retrieve_online_documents_v2`
- `FeatureStore.retrieve_online_documents_v2()` raises `NotImplementedError` for `online_store.type: qdrant`
- v1 ignores `requested_features`; returns a 5-tuple, not `Dict[str, ValueProto]`
- No `query_string` / hybrid path in Feast's Qdrant integration
- `create_collection` configures dense vectors only (no sparse vectors for keyword search)

**File:** `sdk/python/feast/infra/online_stores/qdrant_online_store/qdrant.py`

## What is required (v2 contract)

Implement `retrieve_online_documents_v2` per `OnlineStore` interface:

- **Input:** `requested_features`, `embedding` and/or `query_string`, `top_k`, `distance_metric`, `table`, `config`
- **Output:** `List[(event_ts, EntityKeyProto, Dict[str, ValueProto])]` with requested features + `distance` (+ `text_rank` when text search is used)

Reference: `milvus.py`, `mongodb.py`, `postgres.py`

## What to include

1. **Dense vector search** — `query_points` / Query API; honor `requested_features` in `feature_dict`; map score → `distance`
2. **Multi-feature per hit** — handle Qdrant's one-feature-per-point write model (join/fetch or storage change)
3. **Hybrid / `query_string`** — sparse vector index + write path; encode `query_string` to sparse vector; Query API prefetch (dense + sparse) + fusion (RRF); support embedding-only, text-only, and combined queries
4. **Tests** — unit and/or integration under `sdk/python/tests/`
5. **Docs** — update `docs/reference/alpha-vector-database.md` and `docs/reference/online-stores/qdrant.md`

rehan243:
fwiw, we ran into the same NotImplementedError for `retrieve_online_documents_v2` when switching Feast's online store to Qdrant — the docs barely mention v2's contract and so it's easy to miss that `requested_features` actually matters. The thing is, Qdrant's data model is super rigid (one feature per point) so doing multi-feature retrieval means either some nasty join logic or changing how you write points in the first place. For hybrid queries, we ended up hacking together a sparse vector field for keyword search using Qdrant's `payload_index`, then fused dense/sparse scores with an RRF ranker. Here’s a stub we used for the v2 method — honestly, it’s messy, but covers the basics:

```python
[...]

PR #6484, opened 2026-06-09
feat: Add retrieve_online_documents_v2 support to Qdrant online store

patelchaitany:

This PR adds retrieve_online_documents_v2 for the Qdrant online store so vector search returns all requested features per document, not just the embedding hit. It implements dense embedding search (with a join across Qdrant’s one-point-per-feature layout), plus optional hybrid search when text_search_enabled: true. v1 is unchanged; hybrid is off by default so existing setups don’t need migration. Files touched: qdrant.py, qdrant.md, alpha-vector-database.md, test_qdrant_online_retrieval.py, test_qdrant_retrieve_online_documents_v2.py, and manual_qdrant_v2.py.

[#6445](https://github.com/feast-dev/feast/issues/6445)

# Checks

## Testing Strategy
- [x] Unit tests
- [x] Integration tests
- [x] Manual tests

# Misc

---

<a href="https://app.devin.ai/review/feast-dev/feast/pull/6484" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/docs/reference/alpha-vector-database.md
+++ b/docs/reference/alpha-vector-database.md
@@ -14,8 +14,8 @@ Below are supported vector databases and implemented features:
 | Milvus          | [x]       | [x]      | [x]         | [x]         |
 | Faiss           | [ ]       | [ ]      | []          | []          |
 | SQLite          | [x]       | [ ]      | [x]         | [x]         |
-| Qdrant          | [x]       | [x]      | []          | []          |
 | ScyllaDB        | [x]       | [x]      | [x]         | [x]         |
+| Qdrant          | [x]       | [x]      | [x]         | [x]         |
 
 *Note: V2 Support means the SDK supports retrieval of features along with vector embeddings from vector similarity search.
 
@@ -31,7 +31,7 @@ Beyond that, we will then have `retrieve_online_documents` and `retrieve_online_
 backwards compatibility and the adopt industry standard naming conventions.
 {% endhint %}
 
-**Note**: Milvus, SQLite, and ScyllaDB implement the v2 `retrieve_online_documents_v2` method in the SDK. This will be the longer-term solution so that Data Scientists can easily enable vector similarity search by just flipping a flag.
+**Note**: Milvus, SQLite, PostgreSQL, Elasticsearch, MongoDB, ScyllaDB and Qdrant implement the v2 `retrieve_online_documents_v2` method in the SDK. This will be the longer-term solution so that Data Scientists can easily enable vector similarity search by just flipping a flag.
 
 ## Examples
 
--- a/sdk/python/feast/infra/online_stores/qdrant_online_store/qdrant.py
+++ b/sdk/python/feast/infra/online_stores/qdrant_online_store/qdrant.py
@@ -11,6 +11,7 @@ from qdrant_client import QdrantClient, models
 
 from feast import Entity, FeatureView, RepoConfig
 from feast.infra.key_encoding_utils import (
+    deserialize_entity_key,
     get_list_val_str,
     serialize_entity_key,
 )
@@ -19,11 +20,13 @@ from feast.infra.online_stores.vector_store import VectorStoreConfig
 from feast.protos.feast.types.EntityKey_pb2 import EntityKey as EntityKeyProto
 from feast.protos.feast.types.Value_pb2 import Value as ValueProto
 from feast.repo_config import FeastConfigBaseModel
+from feast.types import PrimitiveFeastType
 from feast.utils import (
     _build_retrieve_online_document_record,
     _get_feature_view_vector_field_metadata,
     to_naive_utc,
 )
+from feast.value_type import ValueType
 
 SCROLL_SIZE = 1000
 
@@ -33,6 +36,8 @@ DISTANCE_MAPPING = {
     "dot": models.Distance.DOT,
     "l1": models.Distance.MANHATTAN,
 }
+
+logger = logging.getLogger(__name__)
 
 
 class QdrantOnlineStoreConfig(FeastConfigBaseModel, VectorStoreConfig):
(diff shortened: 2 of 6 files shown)


docs/reference/alpha-vector-database.md
SIDDHESH1564: @patelchaitany I think alpha-vector-database.md matrix table is inconsistent with the note text.

docs/reference/alpha-vector-database.md
```diff
@@ -30,7 +30,7 @@ Beyond that, we will then have `retrieve_online_documents` and `retrieve_online_
 backwards compatibility and the adopt industry standard naming conventions.
 {% endhint %}
 
-**Note**: Milvus and SQLite implement the v2 `retrieve_online_documents_v2` method in the SDK. This will be the longer-term solution so that Data Scientists can easily enable vector similarity search by just flipping a flag.
+**Note**: Milvus, SQLite, PostgreSQL, Elasticsearch, MongoDB, and Qdrant implement the v2 `retrieve_online_documents_v2` method in the SDK. This will be the longer-term solution so that Data Scientists can easily enable vector similarity search by just flipping a flag.
```
SIDDHESH1564: The note at line 33 was updated to include Qdrant, but the table at line 17 still shows [] for V2 Support. Need to update the table row to [x] for V2 Support (and Online Read).

patelchaitany: Done, Thanks for Pointing out

sdk/python/feast/infra/online_stores/qdrant_online_store/qdrant.py
```diff
@@ -64,6 +69,107 @@ class QdrantOnlineStoreConfig(FeastConfigBaseModel, VectorStoreConfig):
...
+    if not texts:
+        return []
+    if not getattr(client, "_FASTEMBED_INSTALLED", False):
+        raise ValueError(
+            "Qdrant hybrid text search requires fastembed. "
+            "Install with: pip install fastembed"
+        )
+    client.set_sparse_model(config.sparse_embedding_model)
+    sparse_vectors: List[models.SparseVector] = []
+    for embedding in client._sparse_embed_documents(texts):
```
SIDDHESH1564: @patelchaitany _encode_sparse_texts has a likely bug and uses deprecated internals

client._sparse_embed_documents(texts) at line 155 is called without passing embedding_model_name, so it defaults to "BAAI/bge-small-en" (a dense model). The official usage in qdrant-client's own add() method explicitly passes self.sparse_embedding_model_name. This means the sparse encoding at write time would attempt to use a dense model name to initialize a sparse model, which will either error or produce wrong results.

Additionally, _sparse_embed_documents, _FASTEMBED_INSTALLED, and the surrounding fastembed APIs were deprecated in qdrant-client. The modern approach is to use models.Document objects and let the client handle embedding internally.

SIDDHESH1564: Suggested fix: if you must use this private API, pass the model name explicitly:

```client._sparse_embed_documents(texts, embedding_model_name=config.sparse_embedding_model)```

Or better yet, consider using the non-deprecated models.Document approach for sparse vector creation (as the retrieval path already does at line 559).

Comment (patelchaitany) [reply sdk/python/feast/infra/online_stores/qdrant_online_store/qdrant.py]: Done

File: sdk/python/feast/infra/online_stores/qdrant_online_store/qdrant.py [hunk 3]
```diff
@@ -256,6 +363,271 @@ def teardown(
...
+                hit_order.append(entity_key_bin)
+            hit_payload_by_entity.setdefault(entity_key_bin, payload)
+            if point.score is not None:
+                score = float(point.score)
+                feature_name = payload.get("feature_name")
+                if is_hybrid:
+                    if feature_name == vector_field_name:
+                        dense_score_by_entity.setdefault(entity_key_bin, score)
+                    elif feature_name == text_feature_name:
+                        sparse_score_by_entity.setdefault(entity_key_bin, score)
```
SIDDHESH1564: **RRF fused scores are mislabeled as dense/sparse scores**

In hybrid mode, Qdrant's RRF fusion query returns a single **fused rank score** per point — not the original dense cosine similarity or sparse BM25 relevance. This loop assigns the fused score to `dense_score_by_entity` or `sparse_score_by_entity` based on `feature_name`, but the score from an RRF-fused result set doesn't represent the raw similarity for that specific vector type.

Downstream, `distance` and `text_rank` are surfaced to users who'll naturally interpret them as cosine similarity and keyword relevance. This could lead to incorrect ranking decisions in application code.

**Options:**
- Return only a single `distance` field in hybrid mode representing the fused RRF score, and document it clearly
- Run dense and sparse queries independently (like qdrant-client's own `query()` method does — see `qdrant_fastembed.py` L680-703), fuse manually with `reciprocal_rank_fusion()`, and keep the individual scores

For non-hybrid paths (dense-only or text-only), the current behavior is correct.

sdk/python/feast/infra/online_stores/qdrant_online_store/qdrant.py
```diff
@@ -256,6 +363,271 @@ def teardown(
...
+        if query_string is not None and not online_store_config.text_search_enabled:
+            raise ValueError(
+                "Text search requires text_search_enabled: true in the Qdrant "
+                "online store config. Recreate the collection with feast apply after "
+                "enabling it."
+            )
+
+        metric = (distance_metric or online_store_config.similarity or "cosine").lower()
+        if metric not in DISTANCE_MAPPING:
+            raise ValueError(f"Unsupported distance metric: {metric}")
```
SIDDHESH1564, line 530: **`distance_metric` is validated but never actually used in the query**

This line validates the metric against `DISTANCE_MAPPING`, but it's never passed to `client.query_points()`. Qdrant uses whatever distance metric was configured when the collection was created (in `create_collection`). If a user passes `distance_metric="l2"` but the collection uses cosine, they'll silently get cosine results.

Suggestions:
- Log a warning if the requested metric differs from the configured `online_store_config.similarity`
- Or document in the docstring that this parameter is for validation only and the collection's configured metric is always used

sdk/python/feast/infra/online_stores/qdrant_online_store/qdrant.py
```diff
@@ -256,6 +363,271 @@ def teardown(
...
+        distance_metric: Optional[str] = None,
+        query_string: Optional[str] = None,
+        include_feature_view_version_metadata: bool = False,
+    ) -> List[
+        Tuple[
+            Optional[datetime],
+            Optional[EntityKeyProto],
+            Optional[Dict[str, ValueProto]],
+        ]
+    ]:
```
SIDDHESH1564, line 473: **Missing docstring** — All other online store implementations (Milvus, MongoDB, PostgreSQL, SQLite) include a docstring on `retrieve_online_documents_v2` documenting the parameters, return format, and any backend-specific behavior (e.g., the one-point-per-feature join, hybrid search semantics). Please add one here for consistency.

SIDDHESH1564 (commented):
Thanks for addressing the earlier feedback, @patelchaitany. The `_encode_sparse_texts` → `models.Document` rewrite is a big improvement, the alpha-vector-database.md table is now consistent, `_parse_timestamp` handles both ISO formats, and the v1 path properly decodes base64 entity keys. All 14 unit tests pass locally.

### What's good

- **`models.Document` for sparse vectors** — The write path now uses the public Qdrant API (`models.Document`) instead of the private `_sparse_embed_documents`, consistent with the retrieval path. Much more maintainable.
- **Robust timestamp parsing** — `_parse_timestamp` now handles both `%Y-%m-%dT%H:%M:%S.%f` and `%Y-%m-%dT%H:%M:%S` with a clear error for unsupported formats.
- **v1 fix** — `retrieve_online_documents` now uses `_entity_key_bytes_from_payload` and `_parse_timestamp`, fixing silent bugs with raw entity key strings and brittle timestamp parsing.
- **New test coverage** — `test_online_write_batch_uses_document_for_sparse_vectors`, `test_retrieve_online_documents_v1_decodes_base64_entity_key`, and `test_parse_timestamp_*` add solid coverage for the new code paths.

### Remaining issues (see inline comments)

1. **RRF fused scores mislabeled as dense/sparse** — In hybrid mode, the scores from RRF fusion are not the original cosine/BM25 scores but fused rank scores. Assigning them to `distance` and `text_rank` is semantically misleading.
2. **`distance_metric` param validated but never applied** — The parameter has no effect on the Qdrant query; the collection uses the metric configured at creation time.
3. **Missing docstring on `retrieve_online_documents_v2`** — Other online store implementations include one.
4. **No integration test for hybrid search** — Only dense and error-case integration tests exist.

sdk/python/feast/infra/online_stores/qdrant_online_store/qdrant.py
```diff
@@ -256,6 +363,271 @@ def teardown(
...
+                ),
+            ]
+            # One Qdrant point per feature: the same entity may appear twice in
+            # fused results (embedding + text_field). Request extra points so we
+            # can dedupe to top_k unique entities.
+            response = client.query_points(
+                collection_name=collection_name,
+                prefetch=prefetches,
+                query=models.FusionQuery(fusion=models.Fusion.RRF),
+                limit=top_k * 2,
```
SIDDHESH1564, line 588: **Nit:** `top_k * 2` works when only embedding + text_field points exist per entity, but if a feature view has more features stored as points, the same entity could consume more than 2 slots in the fused results. A comment documenting the assumption (2 point types participate in fusion: the embedding point and the text_field point) would help future maintainers.

sdk/python/tests/integration/online_store/test_qdrant_retrieve_online_documents_v2.py
```diff
@@ -0,0 +1,166 @@
...
+        assert all(
+            isinstance(c, str) and c.startswith("cat-") for c in result["category"]
+        )
+        assert all(isinstance(i, int) for i in result["doc_id"])
+        assert all(
+            isinstance(e, list) and len(e) == vector_dim for e in result["embedding"]
+        )
+    finally:
+        shutil.rmtree(repo_dir, ignore_errors=True)
+
```
SIDDHESH1564, line 117: **Missing hybrid search integration test** — The integration tests cover dense search and the error case for `text_search_enabled`, but there's no end-to-end test exercising hybrid search (`embedding` + `query_string` together with `text_search_enabled: true`). This is the most complex code path and the one most likely to break. Even if it requires `fastembed` as a dependency, a guarded test (with `pytest.importorskip("fastembed")`) would add strong confidence.

SIDDHESH1564 (approved):
@patelchaitany Look good to me!

PR #6504, opened 2026-06-11
docs: Add end-to-end registry deletion lifecycle example and unit test (#5360)

### LINKED ISSUE #5360: Doc: Document feature store objects deletion from feast
Feast currently supports the objects deletion from registry, but its not documented yet. The object deletion is done via : `feast.apply(objects_to_delete=[feature_view1])`

1. Document the object deletion using a jupyter notebook that:
- Creates feature store objects.
- Read the feature store objects.
- Deletes the objects.
- Validate the deleted objects are not available.
2. [Optional] Write an unit test.

obielin:
## What this PR does

Closes #5360.

The existing registry docs already cover the `feast apply` deletion warning, the CLI command, the individual Python SDK `delete_*` methods, and the `partial=False` note. The one remaining gap, as agreed in the issue discussion, is that there is no single, self-contained example showing the complete deletion lifecycle in one place.

This PR addresses that gap with a minimal, copy-pasteable addition.

## Changes

- **`docs/getting-started/components/registry.md`**: Adds an "End-to-end example" section demonstrating the full feature view deletion lifecycle — list/verify the object exists, delete it, then list again to confirm removal — and notes that the same pattern applies to other registry objects via their corresponding `delete_*` methods.
- **`sdk/python/tests/unit/local_feast_tests/test_local_feature_store.py`**: Adds `test_apply_delete_feature_view`, which registers an entity and feature view, deletes it using `apply(objects_to_delete=[...], partial=False)`, and asserts the feature view is gone and that `get_feature_view` raises `FeatureViewNotFoundException`.

## Notes

This keeps the change intentionally small and documentation-focused, per the consensus reached in the issue thread between the maintainer and contributors. Thanks to @Henildiyora for the original snippet draft and review, and to @jyejare for the guidance.

### DIFF
--- a/sdk/python/tests/unit/local_feast_tests/test_local_feature_store.py
+++ b/sdk/python/tests/unit/local_feast_tests/test_local_feature_store.py
@@ -11,7 +11,7 @@ from feast.aggregation import Aggregation
 from feast.data_format import AvroFormat, ParquetFormat
 from feast.data_source import KafkaSource
 from feast.entity import Entity
-from feast.errors import ConflictingFeatureViewNames
+from feast.errors import ConflictingFeatureViewNames, FeatureViewNotFoundException
 from feast.feast_object import ALL_RESOURCE_TYPES
 from feast.feature_store import FeatureStore
 from feast.feature_view import DUMMY_ENTITY_ID, DUMMY_ENTITY_NAME, FeatureView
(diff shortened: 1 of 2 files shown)


sdk/python/tests/unit/local_feast_tests/test_local_feature_store.py
```diff
@@ -441,6 +441,64 @@ def test_apply_permissions(test_feature_store):
...
+        # Register entity and feature view
+        test_feature_store.apply([entity, driver_fv])
+
+        # Verify feature view exists
+        fvs = test_feature_store.list_batch_feature_views()
+        assert len(fvs) == 1
+        assert fvs[0].name == driver_fv.name
+
+        # Delete the feature view using objects_to_delete
+        test_feature_store.apply(
```
ntkathole, line 547: test is using feast apply to delete while docs and docstrings says objects_to_delete or delete_feature_view.

obielin: Good catch, you're right — thanks. Fixed in `92523ec38`.

The docs snippet demonstrates `store.delete_feature_view(name)`, but the only test I added went through `apply(objects_to_delete=[...], partial=False)`, so the API the example actually teaches had no coverage at all.

Rather than swap one for the other, I split them, since `registry.md` legitimately documents both routes:

- **`test_delete_feature_view`** — mirrors the end-to-end snippet step for step: list, delete by name, list again, then assert `get_feature_view` raises `FeatureViewNotFoundException`.
- **`test_delete_feature_view_raises_when_missing`** — covers the `FeatureViewNotFoundException` that `delete_feature_view`'s own docstring promises for a name that was never registered.
- **`test_apply_delete_feature_view`** — keeps the `objects_to_delete` / `partial=False` path from the hint block above the example, with a docstring noting it is deliberately the `feast apply` route and distinct from the one above.

I also lifted the shared source frame and registration into two small helpers, so the two lifecycle tests don't duplicate ~30 lines of setup.

Verified locally: the 3 deletion tests pass, and the file is at 24 passed. The 2 failures in `test_apply_stream_feature_view` / `test_apply_stream_feature_view_udf` are pre-existing on `master` — identical with my changes stashed. `ruff check` and `ruff format --check` are both clean.

The branch was 151 commits behind, so I rebased onto current `master` in the same push; that clears the previous out-of-date state.

ntkathole (approved):

PR #6512, opened 2026-06-13
feat: Warn when the registry server starts with authentication disabled

AgentGymLeader:
Add a startup warning when the gRPC registry server is launched with authentication disabled (`auth` type `no_auth`).

## Why

The registry server accepts control-plane writes such as applying feature views. Materializing those definitions loads user-provided transformation code on the registry host. When authentication is disabled the server accepts these requests unauthenticated, so it is only safe to run inside a trusted network boundary. Today nothing in the startup path signals this to the operator — the only startup logs are info-level TLS/non-TLS messages.

## What

- Add `_warn_if_auth_disabled(auth_manager_type)`, which emits a single `logger.warning` when the auth manager type is `NONE`.
- Call it from `start_server` right after the auth type is resolved.
- Add a unit test covering that the warning fires for `NONE` and stays silent for an authenticated type.

Behavior is unchanged; this only adds an operator-facing warning. This is a defense-in-depth / operator-visibility improvement, not a behavior change.

### DIFF
--- a/sdk/python/feast/registry_server.py
+++ b/sdk/python/feast/registry_server.py
@@ -175,6 +175,25 @@ def _build_any_feature_view_proto(feature_view: BaseFeatureView):
     )
 
 
+def _warn_if_auth_disabled(auth_manager_type: AuthManagerType) -> None:
+    """Emit a startup warning when the registry server runs without authentication.
+
+    The registry server accepts control-plane writes (e.g. applying feature
+    views), which materialize user-provided transformation code on this host.
+    When authentication is disabled it accepts those requests unauthenticated,
+    so it should only run inside a trusted network boundary.
+    """
+    if auth_manager_type == AuthManagerType.NONE:
+        logger.warning(
+            "Registry server is starting with authentication disabled "
+            "(auth type 'no_auth'). It will accept unauthenticated requests, "
+            "including feature-view apply operations that load user-provided "
+            "transformation code on this host. Only run the registry server in "
+            "this mode inside a trusted network boundary, and enable "
+            "authentication before exposing it on an untrusted network."
+        )
+
+
 class RegistryServer(RegistryServer_pb2_grpc.RegistryServerServicer):
     def __init__(self, registry: BaseRegistry) -> None:
         super().__init__()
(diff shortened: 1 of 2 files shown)


suzunn:
I like that this keeps the behavior unchanged and only raises operator visibility. One regression I would add is a thin `start_server`-level test for the `no_auth` config path, with the security/server setup calls monkeypatched out. The helper tests prove the warning logic, but a startup-path test would also catch a future refactor that resolves `auth_manager_type` correctly and then accidentally stops calling `_warn_if_auth_disabled` before the server is exposed.

AgentGymLeader:
@suzunn good call — added `test_start_server_warns_when_auth_disabled` in `test_registry_server.py` (c1997047dc).

It calls `start_server` with a `no_auth` store config and monkeypatches `init_security_manager` (the step `start_server` invokes right after auth-type resolution) to assert the `no_auth` warning has *already* been emitted, then raises a sentinel to halt before anything binds. So a future refactor that resolves `auth_manager_type` correctly but stops calling `_warn_if_auth_disabled` before the server is exposed would fail this test — exactly the startup-path gap you flagged. The helper-level tests still cover the warning logic itself; behavior is otherwise unchanged.

AgentGymLeader:
Hi @suzunn! Thanks for the helpful feedback — I've added the test and rebased onto the latest master. It looks like only the DCO check has run so far. Could you help trigger the CI workflows when you get a chance? Really appreciate your time! 🙏

AgentGymLeader:
@suzunn quick heads up, I rebased onto latest master to clear the behind state. Didn't touch any of the reviewed code, it's just replayed onto the new base (only thing that came along is an unrelated upstream UI-build fix). The no_auth start-server test from your earlier suggestion is still there.

AgentGymLeader:
Hi @suzunn, sorry to ping again. The start_server-level test you suggested for the no_auth path (`test_start_server_warns_when_auth_disabled`) is in. It looks like the CI workflows past DCO still need a maintainer trigger — could you kick those off when you get a chance? Happy to rebase or adjust anything if needed.

PR #6527, opened 2026-06-15
fix: Do not pass undeclared feature view columns to ODFV UDFs

### LINKED ISSUE #6158: OnDemandFeatureView (pandas/python mode) UDF receives all response columns, not just declared sources
## Expected Behavior
When an OnDemandFeatureView (ODFV) is defined with a specific sources list, the UDF passed to it should only receive columns that belong to the declared sources. Accessing columns outside the declared sources should either raise an error or be impossible. 

## Current Behavior
In pandas and python transformation modes, the UDF receives all feature columns present in the offline response at call time — including features from feature views that are not declared in sources. This means the UDF can silently depend on undeclared sources without any warning or error.

## Steps to reproduce
1. Define two feature views fv1 and fv2.
2. Define an ODFV with sources=[fv1] (mode pandas or python).
3. Write a UDF that reads a column belonging to fv2.
4. Call get_online_features requesting features from both fv1, fv2, and the ODFV simultaneously.
5. Observe that the UDF runs without error and silently uses the undeclared fv2 column.

### Specifications

- Version: latest
- Platform: all
- Subsystem: OnDemandFeatureView, PandasTransformation, PythonTransformation

Vedant-Agarwal:
## What

An OnDemandFeatureView's UDF (pandas and python modes) was receiving every feature column in the online response, including features from feature views that the ODFV did not list in its sources. This allowed a UDF to silently depend on an undeclared source.

This filters the UDF input down to the ODFV's declared sources before the transform runs. Join keys, request-source fields and the declared features are kept; only columns from undeclared feature views are removed. substrait already restricts its inputs through its query plan, so it is left as is.

Fixes #6158.

## Tests

Three tests were added in `test_on_demand_pandas_transformation.py`:

- pandas mode: the UDF sees its declared feature and the entity join key, but not another requested feature view's column.
- python mode: the same check with the dict input.
- two ODFVs: one ODFV's UDF does not see the other ODFV's source column, even when that feature view is present only as the other ODFV's input.

Each of these fails on the current code and passes with this change.

Functional test results:

```
$ pytest tests/unit/test_on_demand_pandas_transformation.py \
    -k "undeclared or other_requested_odfv"
3 passed in 0.52s
```

`ruff` and `mypy` are clean.

## Scope

This covers the online path (`get_online_features`), which is what the issue reports. The offline path (`get_historical_features`) has the same shape and can be addressed as a follow-up.

### DIFF
--- a/sdk/python/feast/utils.py
+++ b/sdk/python/feast/utils.py
@@ -812,17 +812,58 @@ def _augment_response_with_on_demand_transforms(
             # Apply transformation. Note: aggregations and transformation configs are mutually exclusive
             # TODO: Fix to make it work for having both aggregation and transformation
             #  ticket: https://github.com/feast-dev/feast/issues/5689
+            # A UDF should only get features from the sources it declared. Drop
+            # features that belong to other feature views so it can't read them by
+            # accident (#6158). Join keys, request data and the declared features
+            # stay. substrait filters on its own, so leave it alone.
+            declared_source_names = {
+                projection.name
+                for projection in odfv.source_feature_view_projections.values()
+            }
+            undeclared_columns: set[str] = set()
+            # features the caller asked for that aren't one of our sources
+            for feature_ref in feature_refs:
+                ref_view_name, _, ref_feature_name = _parse_feature_ref(feature_ref)
+                if ref_view_name in requested_odfv_feature_names:
+                    continue
+                if ref_view_name not in declared_source_names:
+                    undeclared_columns.add(ref_feature_name)
+                    undeclared_columns.add(f"{ref_view_name}__{ref_feature_name}")
+            # features that are only in the response because another requested
+            # ODFV needs them as input
+            for other_odfv in requested_on_demand_feature_views:
+                for projection in other_odfv.source_feature_view_projections.values():
+                    if projection.name in declared_source_names:
+                        continue
+                    for feature in projection.features:
+                        undeclared_columns.add(feature.name)
+                        undeclared_columns.add(f"{projection.name}__{feature.name}")
+
             if odfv.mode == "python":
                 if initial_response_dict is None:
                     initial_response_dict = initial_response.to_dict()
+                odfv_input_dict = {
+                    k: v
+                    for k, v in initial_response_dict.items()
+                    if k not in undeclared_columns
+                }
                 transformed_features_dict: Dict[str, List[Any]] = odfv.transform_dict(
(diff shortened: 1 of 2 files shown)


sdk/python/feast/utils.py
```diff
@@ -811,17 +811,58 @@ def _is_metrics_active():
             # Apply transformation. Note: aggregations and transformation configs are mutually exclusive
             # TODO: Fix to make it work for having both aggregation and transformation
             #  ticket: https://github.com/feast-dev/feast/issues/5689
+            # A UDF should only get features from the sources it declared. Drop
+            # features that belong to other feature views so it can't read them by
+            # accident (#6158). Join keys, request data and the declared features
+            # stay. substrait filters on its own, so leave it alone.
+            declared_source_names = {
+                projection.name
+                for projection in odfv.source_feature_view_projections.values()
```
Sanjays2402, line 821: `FeatureViewProjection` can carry a `name_alias` or version tag, but these comparisons and column names use `.name`. For an ODFV sourced from an aliased projection, won’t the alias-qualified input be classified as undeclared while `alias__feature` also escapes the second filter? Should these use `projection.name_to_use()` instead?

Vedant-Agarwal: Thanks @Sanjays2402 — fair question, so I verified it against this branch (checked out this PR's head) rather than guess.

With an ODFV sourced from an aliased projection:

```python
aliased = driver_stats.with_name("aliased_stats")
# projection.name = 'driver_stats', name_to_use() = 'aliased_stats'
```

the columns feeding the transform come through as `['driver_stats__conv_rate', 'conv_rate']` — `.name`-qualified, not alias-qualified. That's consistent with how source refs are built throughout `on_demand_feature_view.py` (`f"{source_fv_projection.name}__{feature.name}"`, e.g. the loops around lines 987/1055/1151/1261). So `.name` matches the real column names here; switching to `name_to_use()` would build `aliased_stats__conv_rate`, which isn't present, and would drop the declared feature.

On `version_tag`: it's only set via `from_proto` (there's no user-facing setter), and the retrieval path keys source columns by `.name` regardless — so even a versioned projection's column is still `driver_stats__conv_rate`.

I'll add a test with an aliased source to lock this behavior in. If you know a path (offline retrieval?) where the input comes through alias-qualified, point me at it and I'll handle that too — I exercised online retrieval here.

ntkathole (approved):

PR #6532, opened 2026-06-16
feat: Add OnlineStore for Aerospike

vkagamlyk:

Adds a first-class Aerospike online store integration at
`feast.infra.online_stores.aerospike_online_store`, backed by the official
`aerospike>=19` Python client and targeting Aerospike server **6.0+**
(developed and tested against CE **8.x**).

Aerospike is a low-latency distributed key-value store widely deployed for
real-time ML feature serving. This integration gives Feast users a managed path
to Aerospike for online retrieval.

## Storage layout

Each entity is stored as a **single Aerospike record** keyed by the serialized
entity key. Multiple feature views for the same entity share one record using
Map CDT bins:

```text
namespace: <config.namespace>              # default: "feast"
set:       <set_name_template>             # default: "{project}_{collection_suffix}" → e.g. my_repo_latest
key:       bytearray(serialize_entity_key(entity_key))
bins:
  features   : map<fv_name, map<feature_name, native_value>>
  event_ts   : map<fv_name, epoch_ms>
  created_ts : epoch_ms                    # optional, single top-level value
record TTL: config.ttl_seconds             # record-level (-1 = never, 0 → never, unset → namespace default)
```

**Default:** one set per project (`{project}_latest`). Per-feature-view isolation
is opt-in via `set_overrides` (see below).

Writes use Aerospike **`batch_write`** with Map CDT `map_put_items` /
`map_put` ops so concurrent writers touching different feature views on the same
entity never clobber each other. Reads use **`batch_operate`** with server-side
Map-get projection when `requested_features` is provided.

## Implementation highlights

- **Sync and async paths** — implements `online_write_batch`, `online_read`,
  `online_write_batch_async`, `online_read_async`, `initialize()`, and
  `close()`. Async methods wrap the blocking client via `run_in_executor`.
  Native asyncio client adoption is tracked as a follow-up.
- **`requested_features` projection** — read path builds a Map CDT
  `map_get_by_key_list` nested into the feature-view submap via `cdt_ctx_map_key`
  so only requested columns are returned from the server.
- **Per-record error surfacing** — `batch_write` / `batch_operate` only raise
  when the whole request is rejected; this store inspects per-record result
  codes and raises on transient failures (`RECORD_NOT_FOUND` and
  `OP_NOT_APPLICABLE` are treated as missing features).
- **Binary key safety** — user keys are passed to the Aerospike C client as
  `bytearray`, not `bytes`. The client hashes only the first byte of `bytes`
  keys during digest computation, causing silent collisions; `bytearray` uses
  the intended binary-key digest.
- **Batch-read ordering** — `batch_operate` preserves input order; responses
  are paired back via `zip` rather than re-parsing `br.key[2]`.
- **MAP_KEY_ORDERED** — Map CDTs are created with ordered-map policy for
  O(log N) lookups on wide feature views and on the `update()` background scan.
- **KEY_DIGEST writes** — writes use the client default digest policy; the
  serialized entity key is not stored redundantly on the server.
- **Composite entity keys** — entities go through `serialize_entity_key`; no
  schema changes for multi-entity keys.
- **TTL sentinels** — `ttl_seconds`: unset → namespace default, `0` → never
  expire, `>0` → explicit seconds.
- **Per-FV namespace / set overrides** — `namespace_overrides` and
  `set_overrides` pin individual feature views to different namespaces or sets
  without splitting the project across stores. `update()` and `teardown()`
  honour resolved `(namespace, set)` pairs.
- **Prewriting hook** — optional import-string-resolved callable on
  `online_write_batch` for PII masking, encryption, coercion, etc.
- **Timeout surface** — `read_timeout_ms`, `write_timeout_ms`,
  `batch_total_timeout_ms` (per-batch total deadline including retries), and
  optional `socket_timeout_ms` (per-attempt, propagates to all three policies).
- **Enterprise-only features** — `user` / `password` auth and TLS
  (`tls` / `tls_ca_file`) are plumbed through the config but are **config-only
  in v1** on Aerospike CE (not exercised in GitHub Actions). Documented in the
  reference page.

## Tests

- **~57 pure-Python unit tests** (no Docker required) covering:
  - Datetime helpers, TTL sentinels, timeout policy wiring
  - Raw-doc → proto conversion, result ordering
  - `_build_batch_writes` shape: Map bins, TTL meta, `bytearray` keys,
    `MAP_KEY_ORDERED` policy, empty-batch short-circuit
  - `online_read` projection, not-found / op-not-applicable / transient errors
  - Per-record write error surfacing
  - `update(tables_to_delete=...)` scan coalescing by `(namespace, set)`
  - `teardown` truncates all resolved pairs and closes the client
  - Namespace/set overrides on read, write, update, teardown
  - Prewriting hook resolution, caching, transform, and short-circuit
  - Async wrappers delegate to sync path; `initialize` / `close` lifecycle
  - Config-type guards reject non-Aerospike configs

- **3 Docker-backed integration tests** via `testcontainers` (Aerospike CE 8 —
  `aerospike/aerospike-server:8.0.0.10_1`):
  - `test_aerospike_online_features` — end-to-end write + read + projection +
    missing-key → None across int, string, and composite entity keys
  - `test_aerospike_cross_fv_map_cdt_upsert` — two feature views on the same
    entity coexist (Map CDT upsert semantics)
  - `test_aerospike_update_strips_dropped_feature_view` — `update()` background
    scan removes a dropped FV slot while the surviving FV stays intact

- **Universal online-store suite wiring** — ships `AerospikeOnlineStoreCreator`
  plus a `FULL_REPO_CONFIGS` module for opt-in via
  `FULL_REPO_CONFIGS_MODULE=feast.infra.online_stores.aerospike_online_store.aerospike_repo_configuration`.
  Default-matrix registration is deferred to a follow-up PR (MongoDB, Couchbase,
  Cassandra, MySQL pattern).

## Documentation

- `docs/reference/online-stores/aerospike.md` — reference page (config, data
  model, CE vs EE matrix, overrides, prewriting hook, tuning).
- `docs/reference/online-stores/README.md` + `docs/SUMMARY.md` — linked.
- `docs/roadmap.md` — Aerospike marked `[x]` under Online Stores.
- `docs/how-to-guides/online-server-performance-tuning.md` — Aerospike row and
  tuning subsection.
- `examples/online_store/aerospike_overrides_and_hooks/` — sample hook +
  override config.

## feast-operator

Adds `aerospike` to supported online store types in feast-operator CRDs and
bundled manifests.

## Out of scope for v1 (follow-ups)

- Feature-view versioning (raw FV names used as Map keys — matches MongoDB).
- Native asyncio Aerospike client (currently `run_in_executor` wrappers).
- Override `async_supported` on `AerospikeOnlineStore` so the feature server
  routes to `online_read_async` / `online_write_batch_async` (methods exist;
  property still inherits the base `read=False, write=False` default).
- Participation in the default universal-suite CI matrix (opt-in only today).
- Vector search / Aerospike Vector Search integration.

## Known follow-ups before upstream PR

- [ ] Remove or retain preview hint in `aerospike.md` per feast-dev convention
      (reviewer note: same pattern as MongoDB preview banner).
- [ ] Optional: add feast-benchmarks Aerospike sweep results to PR / blog once
      upstream benchmarks PR lands.

## Checks

- [x] Tests passing locally (`pytest` unit + Docker integration tests).
- [x] Commits signed off (`git commit -s`).
- [x] PR title follows conventional commits format.

## Testing strategy

- [x] Unit tests
- [x] Integration tests (testcontainers / Aerospike CE 8)

## Misc

Aerospike client dependency is gated behind the `aerospike` optional extra
(`pip install 'feast[aerospike]'`), so existing installs that don't need it pay
no import cost.


franciscojavierarceo:
@cocdex review

lyndonbauto (approved):

lyndonbauto:
Hi @franciscojavierarceo, are you able to take a look at this PR? We are really excited to have a feast integration, customers have been making custom ones and we want to get something official! 

About us:
Aerospike is a real-time data platform, with our primary usage being KV. We have uncompromising and consistent performance, with sub-millisecond reads.

ntkathole:
@vkagamlyk Need to generate pixi.lock file to update dependencies

ntkathole:
@vkagamlyk looks good overall,

1. Thread-unsafe lazy client initialization - potential connection leak

`_get_client` has no locking. In a threaded feature server (the default deployment), two concurrent requests can both create a client, one gets silently leaked (never closed). Other online stores like Redis and MongoDB both handle this correctly.

2. No batch size chunking - will crash on large materializations

Aerospike's server-side batch-max-requests defaults to 5,000 records ? If online_write_batch receives more than that (common during feast materialize), the entire batch fails with a server error. (DynamoDB's store chunks at 25, Redis pipelines in segments.)

vkagamlyk:
> @vkagamlyk looks good overall,
> 
>     1. Thread-unsafe lazy client initialization - potential connection leak
> 
> 
> `_get_client` has no locking. In a threaded feature server (the default deployment), two concurrent requests can both create a client, one gets silently leaked (never closed). Other online stores like Redis and MongoDB both handle this correctly.
> 
>     2. No batch size chunking - will crash on large materializations
> 
> 
> Aerospike's server-side batch-max-requests defaults to 5,000 records ? If online_write_batch receives more than that (common during feast materialize), the entire batch fails with a server error. (DynamoDB's store chunks at 25, Redis pipelines in segments.)

Thank you for review @ntkathole. 

Addressed in https://github.com/feast-dev/feast/pull/6532/commits/256d65d2fc8a014cc75e69e5aac2d8341eaa196e

ntkathole (approved):
Looks good, Thank you @vkagamlyk for contribution!

PR #6549, opened 2026-06-23
feat: Add OpenLineage Consumer to Feast - receive, store, and visualize cross-producer lineage

ntkathole:

Feast currently only produces OpenLineage events. Teams running multi-tool pipelines (Airflow → Spark → dbt → Feast) must deploy a separate Marquez instance to see how data flows across producers. OpenLineage could be consumed without Marquez.

This PR makes Feast a **first-class OpenLineage consumer**. When the consumer is not enabled, the Feast UI remains unchanged.

Screenshot 2026-06-23 at 10 58 40 AM

## Screen Capture

https://github.com/user-attachments/assets/efa49e72-06d1-4b13-ba22-3e807ba478ac


franciscojavierarceo (approved):

franciscojavierarceo:
probably in the UI we should make feast only lineage the default? Maybe not ?

ntkathole:
> probably in the UI we should make feast only lineage the default? Maybe not ?

May be not. It's an opt-in feature. If someone went through the effort of configuring it, that means they want to see the cross-producer lineage.

PR #6550, opened 2026-06-24
feat: Bring Your Own Spark - SparkApplication

aniketpalu:

### Core changes
**Upstream (feature_store.py):**
- Refactored \`materialize()\` and \`materialize_incremental()\` to collect all \`MaterializationTask\`s and pass them to the engine in a single \`batch_engine.materialize(registry, tasks)\` call. This enables batching engines to process all feature views in one job. Existing engines are unaffected — \`ComputeEngine.materialize()\` base class loops tasks internally via \`_materialize_one()\`.
- Added public \`get_provider()\` method on \`FeatureStore\` (wraps existing private \`_get_provider()\`).
**New engine (spark_application/):**
- \`config.py\`: Pydantic config with Spark resource specs, K8s options, validation (\`env\` entries must have \`name\`, \`staging_location\` warning).
- \`compute.py\`: Creates a K8s Secret with \`feature_store.yaml\` + task list, builds SparkApplication CR, polls for completion, retrieves driver logs on failure.
- \`job.py\`: Maps all 14 SparkApplication states to \`MaterializationJobStatus\`, with exponential backoff retry on transient API errors.
- \`main.py\`: Driver script that runs inside the pod — loads config from Secret, materializes each feature view via \`provider.materialize_single_feature_view()\`. Supports concurrent FV processing via ThreadPoolExecutor + FAIR scheduler.
- \`Dockerfile\`: Based on \`apache/spark:4.0.1\`, installs \`feast[redis]==0.64.0\`, includes Hadoop AWS JARs for S3A support.
### Design decisions
- **Two config rewrites in driver pod**: \`batch_engine\` → \`spark.engine\` (prevents recursive CRD creation, uses local SparkSession), \`registry\` → \`remote\` when \`registry_address\` is set (pod can't access server filesystem).
- **Driver calls \`provider.materialize_single_feature_view()\`**, not \`store.materialize()\` — avoids FeatureViewState transition conflicts (MATERIALIZING→MATERIALIZING is invalid) and duplicate OpenLineage/MLflow emissions.
- **SQLite online store rejected at init** — data written inside a pod is lost on termination.
- **Kueue integration** via \`queue_name\` config field → \`kueue.x-k8s.io/queue-name\` label.
### Validated on
- OpenShift AI (ROSA) with Spark Operator v2.5.0, MinIO (S3A), Redis
- 5 feature views, 4800 rows each, 2 executors: 2.4x speedup with concurrent mode
## Test plan
- [x] 12 unit tests: config defaults, SQLite rejection, registry validation, config rewrites, CR structure, state mapping (14 states), cleanup 404 handling, timeout behavior, job naming
- [x] Existing tests unaffected: 29 feature_view_state + 22 local_feature_store tests pass
- [x] Integration test on cluster with real Spark Operator"

# Checks

## Testing Strategy
- [x] Unit tests
- [x] Integration tests
- [x] Manual tests

# Misc


jyejare:
@aniketpalu lint checks are failing

sdk/python/feast/feature_store.py
```diff
@@ -424,6 +424,113 @@ def _get_provider(self) -> Provider:
         # TODO: Bake self.repo_path into self.config so that we dont only have one interface to paths
         return self.provider
 
+    def get_provider(self) -> Provider:
```
jyejare: **[Warning] Breaking API change - public method exposure**

Adding a public get_provider() method breaks encapsulation and exposes internal implementation details. This could lead to misuse by external code and makes future refactoring more difficult.

sdk/python/feast/feature_store.py
```diff
@@ -424,6 +424,113 @@ def _get_provider(self) -> Provider:
         # TODO: Bake self.repo_path into self.config so that we dont only have one interface to paths
```
jyejare, line 433: **[Critical] Race condition in state management**

The state transition and rollback logic has potential race conditions when multiple materialization jobs run concurrently. The previous_states dictionary and feature view state updates are not atomic, which could lead to inconsistent states or lost rollbacks.

**Suggested:**
```suggestion
        # Use locks or atomic operations for state management
        import threading
        state_lock = threading.Lock()
        
        with state_lock:
            for fv, job in zip(regular_fvs, jobs):
                fv_status = job.status()

                if fv_status == MaterializationJobStatus.ERROR:
                    failed_fvs.append(fv)
                    if first_error is None and job.error():
                        first_error = job.error()
                else:
                    succeeded_fvs.append(fv)

            if failed_fvs:
                self._rollback_fv_states(failed_fvs, previous_states)
```

sdk/python/feast/infra/compute_engines/spark_application/main.py
```diff
@@ -0,0 +1,190 @@
...
+
+        # Get the active SparkSession (created by spark-submit)
+        spark = SparkSession.getActiveSession()
+        if spark is None:
+            spark = SparkSession.builder.getOrCreate()
+        logger.info(f"SparkSession: {spark.sparkContext.applicationId}")
+
+        if concurrency > 1:
+            _original_get_active = SparkSession.getActiveSession
+            SparkSession.getActiveSession = staticmethod(lambda: spark)
```
jyejare: **[Critical] Thread safety violation with SparkSession monkey patching**

Monkey patching SparkSession.getActiveSession globally affects all threads and could cause issues if other code relies on the original behavior. This is a dangerous pattern that could lead to hard-to-debug concurrency issues.

**Suggested:**
```suggestion
            # Use thread-local storage instead of global monkey patching
            import threading
            thread_local = threading.local()
            thread_local.spark_session = spark
            
            def get_thread_spark():
                return getattr(thread_local, 'spark_session', spark)
            
            # Pass spark session explicitly to worker functions
            logger.info("Using thread-local SparkSession management")
```

sdk/python/feast/infra/compute_engines/spark_application/compute.py
```diff
@@ -0,0 +1,420 @@
...
+            )
+
+        registry_type = getattr(repo_config.registry, "registry_type", "")
+        if registry_type in _FILE_BASED_REGISTRY:
+            raise ValueError(
+                f"spark_application engine cannot use '{registry_type}' registry. "
+                f"File-based registries ({', '.join(sorted(_FILE_BASED_REGISTRY))}) "
+                "store data on the local filesystem, which is inaccessible from the "
+                "SparkApplication pod. Use a network-accessible registry: sql, "
+                "snowflake.registry, etc."
```
jyejare, line 91: **[Warning] Overly restrictive store validation**

The validation logic rejects all file-based stores but doesn't account for cases where users might have network-mounted filesystems or custom configurations that make file-based stores accessible from pods.

**Suggested:**
```suggestion
        online_type = getattr(repo_config.online_store, "type", "")
        if online_type in _FILE_BASED_ONLINE:
            import warnings
            warnings.warn(
                f"Using '{online_type}' online store with spark_application engine. "
                f"Ensure the store is accessible from SparkApplication pods via "
                f"network mounts or shared storage, otherwise data will be lost.",
                UserWarning,
                stacklevel=2
            )
```

sdk/python/feast/infra/compute_engines/spark_application/job.py
```diff
@@ -0,0 +1,115 @@
...
+                last_exc = e
+                wait = _RETRY_BACKOFF_BASE**attempt
+                logger.warning(
+                    f"API poll attempt {attempt + 1}/{_MAX_POLL_RETRIES} failed "
+                    f"(HTTP {e.status}), retrying in {wait}s"
+                )
+                time.sleep(wait)
+        self._error = Exception(
+            f"Failed to poll SparkApplication after {_MAX_POLL_RETRIES} attempts: "
+            f"{last_exc.reason if last_exc else 'unknown'}"
```
jyejare, line 150: **[Warning] Insufficient retry logic for transient failures**

The retry logic only applies to API polling but not to other Kubernetes operations. Network timeouts, temporary API server issues, or resource constraints could cause legitimate operations to fail unnecessarily.

**Suggested:**
```suggestion
    def _get_cr_with_retry(self) -> Optional[dict]:
        """Fetch SparkApplication CR with exponential backoff on transient errors."""
        last_exc = None
        for attempt in range(_MAX_POLL_RETRIES):
            try:
                return self.custom_api.get_namespaced_custom_object(
                    group="sparkoperator.k8s.io",
                    version="v1beta2",
                    namespace=self.namespace,
                    plural="sparkapplications",
                    name=f"feast-sa-{self._job_id}",
                )
            except ApiException as e:
                if e.status == 404:
                    self._error = Exception(
                        f"SparkApplication feast-sa-{self._job_id} not found"
                    )
                    return None
                # Retry on server errors (5xx) and rate limiting (429)
                if e.status < 500 and e.status != 429:
                    self._error = Exception(f"API error: {e.reason}")
                    return None
                last_exc = e
                wait = min(_RETRY_BACKOFF_BASE**attempt, 60)  # Cap at 60s
                logger.warning(
                    f"API poll attempt {attempt + 1}/{_MAX_POLL_RETRIES} failed "
                    f"(HTTP {e.status}), retrying in {wait}s"
                )
                time.sleep(wait)
```

sdk/python/feast/infra/compute_engines/spark_application/compute.py
```diff
@@ -0,0 +1,420 @@
...
+                f"feast-sa-{job_id}",
+            ),
+            lambda: self.core_v1.delete_namespaced_config_map(
+                f"feast-sa-{job_id}",
+                self.config.namespace,
+            ),
+        ]:
+            try:
+                fn()
+            except ApiException as e:
```
jyejare, line 509: **[Warning] Resource cleanup may fail silently**

The cleanup method swallows all non-404 errors with just a warning, which could lead to resource leaks if there are permission issues or other persistent problems preventing cleanup.

**Suggested:**
```suggestion
    def _cleanup(self, job_id: str):
        cleanup_errors = []
        for resource_type, fn in [
            ("SparkApplication", lambda: self.custom_api.delete_namespaced_custom_object(
                "sparkoperator.k8s.io",
                "v1beta2", 
                self.config.namespace,
                "sparkapplications",
                f"feast-sa-{job_id}",
            )),
            ("ConfigMap", lambda: self.core_v1.delete_namespaced_config_map(
                f"feast-sa-{job_id}",
                self.config.namespace,
            )),
        ]:
            try:
                fn()
                logger.info(f"Cleaned up {resource_type} feast-sa-{job_id}")
            except ApiException as e:
                if e.status == 404:
                    logger.debug(f"{resource_type} feast-sa-{job_id} not found (already deleted)")
                else:
                    error_msg = f"Failed to cleanup {resource_type}: {e.reason}"
                    logger.error(error_msg)
                    cleanup_errors.append(error_msg)
        
        if cleanup_errors:
            logger.warning(f"Some cleanup operations failed: {'; '.join(cleanup_errors)}")
```

sdk/python/feast/infra/compute_engines/spark_application/config.py
```diff
@@ -0,0 +1,67 @@
...
+                "get_historical_features (not yet supported by this engine). "
+                "It will be ignored for materialize operations.",
+                stacklevel=2,
+            )
+        for i, entry in enumerate(self.env):
+            if "name" not in entry:
+                raise ValueError(
+                    f"env[{i}] is missing required 'name' key. "
+                    f"Each env entry must have at least 'name' and 'value'. Got: {entry}"
+                )
```
jyejare: **[Suggestion] Improve configuration validation**

The env validation only checks for 'name' field but doesn't validate that 'value' exists or that the overall structure is correct for Kubernetes environment variables.

**Suggested:**
```suggestion
        for i, entry in enumerate(self.env):
            if not isinstance(entry, dict):
                raise ValueError(
                    f"env[{i}] must be a dict, got {type(entry).__name__}: {entry}"
                )
            if "name" not in entry:
                raise ValueError(
                    f"env[{i}] is missing required 'name' key. "
                    f"Each env entry must have at least 'name' and 'value'. Got: {entry}"
                )
            if "value" not in entry and "valueFrom" not in entry:
                raise ValueError(
                    f"env[{i}] must have either 'value' or 'valueFrom'. Got: {entry}"
                )
```

sdk/python/feast/infra/compute_engines/spark_application/main.py
```diff
@@ -0,0 +1,190 @@
...
+                SparkSession.getActiveSession = _original_get_active
+                logger.info("Restored SparkSession.getActiveSession")
+
+        total_elapsed = time.time() - total_start
+        logger.info(
+            f"Materialization batch complete: "
+            f"{succeeded} succeeded, {failed} failed, {total} total, "
+            f"elapsed={total_elapsed:.1f}s"
+        )
+        if failed > 0 and succeeded == 0:
```
jyejare: **[Warning] Incomplete error handling for partial failures**

The logic exits with status 1 only if all feature views failed, but partial failures (some succeeded, some failed) exit with status 0. This could mask important issues in monitoring systems.

**Suggested:**
```suggestion
        # Exit with non-zero status for any failures to ensure visibility
        if failed > 0:
            logger.error(f"Materialization completed with {failed} failures")
            sys.exit(1 if succeeded == 0 else 2)  # 1=total failure, 2=partial failure
```

jyejare (changes requested):
While the core architecture is solid, there are several critical issues around thread safety, error handling, and resource management that need addressing before merge.

sdk/python/feast/feature_store.py
aniketpalu: Agreed, removed get_provider() entirely. The existing FeatureStore.provider property (line 418) already provides public access to the provider instance. No new API surface needed.

sdk/python/feast/infra/compute_engines/spark_application/main.py
aniketpalu: Agreed, the global monkey-patch was an ugly pattern. Replaced it with Spark's official per-thread session binding mechanism.

The new _bind_spark_session_to_thread(spark) function constructs a SparkSession(spark.sparkContext, spark._jsparkSession) from the existing SparkContext + JVM session handle. PySpark's __init__ internally calls Java setActiveSession for the current thread. This registers the session per-thread without mutating any class-level state.
Each worker thread calls this at the start of _materialize_one_fv() before doing any Spark work.

sdk/python/feast/infra/compute_engines/spark_application/compute.py
aniketpalu: This is intentional, we treat file-based stores as a hard error, not a warning.

The reason: SparkApplication pods have ephemeral filesystems. SQLite/faiss online stores write data inside the pod and that data is permanently lost when the pod terminates (no partial success, no recovery). File-based offline stores (dask, file, duckdb) read from the local filesystem, which doesn't exist inside the pod. A warning would let users proceed to a guaranteed failure after waiting minutes for pod scheduling + Spark startup, resulting in worse user experience than failing upfront.

Most production use cases that need Spark-level compute use network-accessible stores. This validation saves users from a confusing failure 5 minutes into execution.

sdk/python/feast/infra/compute_engines/spark_application/job.py
aniketpalu: Agreed, expanded the retry logic in two ways:

Smarter retryability: Only retry on genuinely transient errors (429, 500, 502, 503, 504). Fail fast on 401/403 with an actionable RBAC hint: "Check that the Feast server ServiceAccount has the required Role/RoleBinding for sparkapplications and configmaps in this namespace." Previously these would retry 3 times and then give a generic error.

Retry on create paths: Added _create_with_retry() for ConfigMap and SparkApplication creation (not just polling). Same exponential backoff, same retryable status codes. If a 502 hits during CR creation, we retry rather than failing the entire materialization.

The retry helpers (_is_retryable, _rbac_hint, _MAX_RETRIES, _RETRY_BACKOFF_BASE) are shared between polling and creation paths.

sdk/python/feast/infra/compute_engines/spark_application/compute.py
aniketpalu: Improved the cleanup logging

sdk/python/feast/infra/compute_engines/spark_application/main.py
aniketpalu: Agreed, fixed. The driver now exits with status 1 if any FV fails (if failed > 0: sys.exit(1)), not just when all fail.

After the SparkApplication completes (or fails), _build_per_fv_jobs() reads each FV's state from the registry independently.

This means if 3/5 FVs succeeded and 2 failed: the SparkApp CR shows FAILED (exit 1), but the 3 succeeded FVs correctly report SUCCEEDED with their own independent job objects. Only the 2 failed FVs report ERROR. No "pollution" of success status from a shared job object.

sdk/python/feast/feature_store.py
aniketpalu: The threading race condition described isn't present because previous_states is a local variable (line 2551) and all loops (for fv, job in zip(...)) are sequential Python for loops, not threaded. Each materialize() call creates its own dict.

However, reviewing this code path revealed a shared-object bug in our engine's _build_per_fv_jobs. Previously, all succeeded FVs received the same SparkApplicationMaterializationJob reference. During _wait_for_completion, polling sets _error on that object when the SparkApp transitions to FAILED. Later, when feature_store.py calls job.status() for each FV, they all hit the if self._error is not None: return ERROR early-return, even for the FVs that actually succeeded and wrote AVAILABLE_ONLINE to the registry.

sdk/python/feast/feature_store.py
```diff
@@ -2249,7 +2352,21 @@ def materialize_incremental(
...
+            provider = self._get_provider()
+            end_date_tz = utils.make_tzaware(end_date) or _utc_now()
+
+            def tqdm_builder(length):
+                return tqdm(total=length, ncols=100)
+
+            tasks: list = []
+            regular_fvs: list = []
+            previous_states: dict = {}
+            fv_start_dates: dict = {"__end_date__": end_date_tz}
```
ntkathole:
```
@dataclass
class MaterializationDateRange:
    end_date: datetime
    fv_start_dates: dict  # fv_name -> start_date
```

sdk/python/feast/feature_store.py
```diff
@@ -424,6 +424,109 @@ def _get_provider(self) -> Provider:
...
+            jobs = provider.batch_engine.materialize(self.registry, tasks)
+        except Exception:
+            self._rollback_fv_states(regular_fvs, previous_states)
+            raise
+
+        first_error = None
+        succeeded_fvs = []
+        failed_fvs = []
+
+        for fv, job in zip(regular_fvs, jobs):
```
ntkathole, line 513: If the engine returns a different number of jobs than tasks (a bug, but possible), zip would silently truncate. A defensive check would help:


```
assert len(jobs) == len(regular_fvs), \
    f"Engine returned {len(jobs)} jobs for {len(regular_fvs)} tasks"
```

sdk/python/feast/infra/compute_engines/spark_application/main.py
```diff
@@ -0,0 +1,195 @@
...
+        config=thread_config,
+        feature_view=fv,
+        start_date=start,
+        end_date=end,
+        registry=thread_store.registry,
+        project=thread_store.project,
+        tqdm_builder=lambda length: tqdm(total=length, ncols=100),
+    )
+
+    thread_store.registry.apply_materialization(fv, thread_store.project, start, end)
```
ntkathole, line 93: apply_materialization is called twice for each succeeded FV - once by the driver pod, once by the server after the job completes.

sdk/python/feast/infra/compute_engines/spark_application/compute.py
```diff
@@ -0,0 +1,485 @@
...
+        if registry_type in _FILE_BASED_REGISTRY:
+            raise ValueError(
+                f"spark_application engine cannot use '{registry_type}' registry. "
+                f"File-based registries ({', '.join(sorted(_FILE_BASED_REGISTRY))}) "
+                "store data on the local filesystem, which is inaccessible from the "
+                "SparkApplication pod. Use a network-accessible registry: sql, "
+                "snowflake.registry, etc."
+            )
+
+        k8s_config.load_config()
```
ntkathole: If the engine is created but never used for materialization (e.g., during feast apply), this will fail in environments without kubeconfig

sdk/python/feast/infra/compute_engines/spark_application/compute.py
```diff
@@ -0,0 +1,485 @@
...
+                    f"SparkApplication creation failed: HTTP {e.status} {e.reason}."
+                    f"{_rbac_hint(e.status)}"
+                ),
+            )
+            return [job for _ in tasks]
+
+        job = SparkApplicationMaterializationJob(
+            job_id, self.config.namespace, self.custom_api
+        )
+        self._wait_for_completion(job)
```
ntkathole: This should be async.

This blocks the Feast server thread for the entire duration of the Spark job (up to job_timeout_seconds, default 1 hour). For a server handling multiple clients, this is a scalability concern.

sdk/python/feast/infra/compute_engines/spark_application/job.py
```diff
@@ -0,0 +1,164 @@
...
+    "FAILED": MaterializationJobStatus.ERROR,
+    "SUBMISSION_FAILED": MaterializationJobStatus.ERROR,
+    "PENDING_RERUN": MaterializationJobStatus.WAITING,
+    "INVALIDATING": MaterializationJobStatus.WAITING,
+    "SUCCEEDING": MaterializationJobStatus.RUNNING,
+    "FAILING": MaterializationJobStatus.RUNNING,
+    "SUSPENDING": MaterializationJobStatus.WAITING,
+    "SUSPENDED": MaterializationJobStatus.WAITING,
+    "RESUMING": MaterializationJobStatus.WAITING,
+    "UNKNOWN": MaterializationJobStatus.RUNNING,
```
ntkathole: Mapping to WAITING and eventually timing out would be safer instead of mapping to RUNNING.  The poller will keep polling indefinitely.

sdk/python/feast/infra/compute_engines/spark_application/compute.py
```diff
@@ -0,0 +1,485 @@
...
+            )
+            if status == MaterializationJobStatus.ERROR:
+                logs = self._get_driver_logs(job._job_id)
+                if logs:
+                    logger.error(f"Driver logs (last 50 lines):\n{logs}")
+                return
+            if status == MaterializationJobStatus.SUCCEEDED:
+                return
+            time.sleep(self.config.poll_interval_seconds)
+        self._cleanup(job._job_id)
```
ntkathole: This is only called on TIMEOUT, every successful (or failed-but-not-timed-out) materialization run leaves behind a ConfigMap.  Add a finally block or call _cleanup() after _wait_for_completion regardless of outcome, or at minimum delete the ConfigMap on the success path.

sdk/python/feast/infra/compute_engines/spark_application/compute.py
aniketpalu: Agreed this is a real scalability concern for the sync path: `_wait_for_completion` holds the caller until the SparkApplication finishes.

Making the engine poll itself async would redesign the `ComputeEngine.materialize()`. The intended non-blocking path is remote materialization in #6590 (`POST /materialize-async` + registry polling). We'll address true submit-without-wait server behavior there so the async path does not hold a worker thread for the full Spark job.

Would it be okay to leaving this as-is in #6550 and tracked against #6590?

sdk/python/feast/infra/compute_engines/spark_application/compute.py
aniketpalu: Addressed. Lazy-init: create ApiClient / CoreV1 / CustomObjects on first materialize/cleanup use

sdk/python/feast/feature_store.py
aniketpalu: Agreed, replaced the `__end_date__` dict sentinel with a small `_MaterializationDateRange` dataclass (`end_date` + `fv_start_dates`). Same behavior, clearer ownership of the date range passed into `_submit_and_process_materialization_jobs`.

sdk/python/feast/feature_store.py
aniketpalu: Agreed. After `batch_engine.materialize()` we now check `len(jobs) == len(regular_fvs)`. On mismatch we roll back FV states and raise `RuntimeError` instead of silently truncating via `zip`. Used an explicit check rather than `assert` so it still runs under `-O`.

sdk/python/feast/infra/compute_engines/spark_application/job.py
aniketpalu: Addressed. Mapped SparkApplication `UNKNOWN` to `WAITING` instead of `RUNNING`, so we don’t treat an unknown state as healthy progress. The existing `job_timeout_seconds` deadline in `_wait_for_completion` still bounds how long we poll.

sdk/python/feast/infra/compute_engines/spark_application/compute.py
aniketpalu: Agreed.  `_cleanup()` now runs in a `try/finally` after `_wait_for_completion`, so ConfigMap + SparkApplication are deleted on success, error, and timeout. The timeout-only cleanup path inside `_wait_for_completion` was removed to avoid double-delete.

sdk/python/feast/infra/compute_engines/spark_application/main.py
ntkathole: @aniketpalu This still exists

sdk/python/feast/feature_store.py
```diff
@@ -2295,78 +2466,85 @@ def materialize_incremental(
...
-                        registry=self.registry,
-                        project=self.project,
-                        tqdm_builder=tqdm_builder,
-                    )
-                except Exception:
-                    fv_success = False
-                    # Roll back state to previous value on failure.
+            # batch_engine is on PassthroughProvider (concrete); same access as
+            # _submit_and_process_materialization_jobs via untyped provider.
+            if getattr(provider, "batch_engine").supports_batch:
```
ntkathole: fragile

```
batch_engine = getattr(provider, "batch_engine", None)
if batch_engine and getattr(batch_engine, "supports_batch", False):
```

sdk/python/feast/infra/compute_engines/spark_application/main.py
aniketpalu: Agreed, the double write was duplicating materialization intervals. The driver pod remains the owner of apply_materialization (needed for per-FV AVAILABLE_ONLINE / partial success). The server batch path now skips the second call when the engine sets applies_materialization=True (SparkApplication only). Other engines still apply on the server as before.

sdk/python/feast/feature_store.py
aniketpalu: Done

ntkathole (approved):
looks good

PR #6552, opened 2026-06-25
feat: Implement RegistryServer.Proto RPC with RBAC-filtered response (#6558)

nquinn408:
The remote registry client (RemoteRegistry.proto) calls stub.Proto(Empty()) to fetch the full registry, but the RegistryServer service never implemented the Proto RPC, so remote clients received UNIMPLEMENTED. Return proxied_registry.proto().


sdk/python/feast/registry_server.py
```diff
@@ -180,6 +181,9 @@ def __init__(self, registry: BaseRegistry) -> None:
         super().__init__()
         self.proxied_registry = registry
 
+    def Proto(self, request: Empty, context) -> RegistryProto:
+        return self.proxied_registry.proto()
```
ntkathole: @nquinn408 it was intentionally removed, so that it won't return all information to the users not having access. Probably proper rbac needs to be added if we need it for any purpose.

nquinn408: I understand @ntkathole but then it breaks RemoteRegistry.proto, so I don't know what the solution is. I guess a different implementation for RemoteRegistry.proto?

nquinn408: https://github.com/feast-dev/feast/blob/a834126b674356ea1efeafd7006f579c9148c3a1/sdk/python/feast/infra/registry/remote.py#L691

ntkathole: I think we can construct the response by calling each typed List* method (have RBAC checks) rather than forwarding the raw proto().

nquinn408: Is there an open issue to address? I will give it a shot.

ntkathole: created https://github.com/feast-dev/feast/issues/6558

nquinn408: @ntkathole I restructured this PR to address your issue: https://github.com/feast-dev/feast/issues/6558.  Can you kick off LLM review?

sdk/python/feast/registry_server.py
```diff
@@ -180,6 +181,91 @@ def __init__(self, registry: BaseRegistry) -> None:
...
+                registry_proto.validation_references.append(
+                    validation_reference.to_proto()
+                )
+
+            for permission in describable(
+                self.proxied_registry.list_permissions(project=project_name)
+            ):
+                registry_proto.permissions.append(permission.to_proto())
+
+        registry_proto.last_updated.FromDatetime(datetime.now(timezone.utc))
```
ntkathole: shouldn't this be actual registry timestamp ?

ntkathole (approved):
looks good, pending comment

PR #6566, opened 2026-06-29
fix: Use LONGBLOB for SQL registry proto columns on MySQL

### LINKED ISSUE #5800: Feast SQL registry fails to start due to truncated FeatureView protobuf in MySQL (BLOB size limit)
Hi Feast maintainers! I have observed what i consider a bug, or at least something to be improved, i also shared this in the slack community [channel](https://feastopensource.slack.com/archives/C01M2GYP0UC/p1767026747833689).

I think this is related with issue https://github.com/feast-dev/feast/issues/3703

 Here you have the details.

## Expected Behavior

Feast should be able to run long-lived, periodically materialized FeatureViews using the `MySQL` registry without the registry growing unbounded or becoming unreadable. The registry should not reach a state where feast serve or any CLI command fails due to accumulated internal state.

## Current Behavior

Feast records every materialization interval for a FeatureView in the registry. When using the MySQL registry, each `materialize` / `materialize-incremental` run appends a new interval record for each feature view `feature_view_proto` column , causing the serialized FeatureView registry protobuf to grow monotonically (~30 bytes per materialization in our tests).

With enough materializations, the serialized proto exceeds the MySQL `BLOB` column size limit (65,535 bytes). When this happens MySQL silently truncates the blob, after which Feast fails to start with:

```txt
google.protobuf.message.DecodeError:
Error parsing message with type 'feast.core.FeatureView'
```

At this point:
- feast `serve` fails.
- any Feast CLI command that loads the registry fails.
- recovery requires manual MySQL registry intervention: Which is deleting the featureview row in the `feature_views` table.
- No matter the FeatureView definition itself is small; growth comes from accumulated materialization interval history.

## Steps to reproduce

1. Configure Feast with a SQL registry backed by `MySQL`. It uses the `BLOB` MySQL type for the `feature_view_proto` column in the `feature_views` table.
2. Define a FeatureView and run `feast apply`.
3. Run `materialize-incremental` repeatedly over time (e.g. via a cron job).
4. Observe that the registry entry for the FeatureView grows on each run (~30 bytes). For that you can use the following query:

```sql
SELECT
  feature_view_name,
  project_id,
  OCTET_LENGTH(feature_view_proto) AS proto_bytes
FROM feature_views;
```

To get an output like:

| feature_view_name | project_id | proto_bytes |
| ------ | ------ | ------ |
|    my_test_fv    |    my_test_project    |   65535     |

5. Once the stored proto reaches 65,535 bytes (like in the example), MySQL truncates it.
6.  Run `feast serve`, it will fail to load with protobuf `DecodeError`.

### Specifications

- Version: `v0.58.0`
- Platform: docker. 
- Subsystem: both arm and amd images.

## Possible Solution

Clarify and/or improve SQL registry behavior for long-running deployments, for example by:

- SQL registry `BLOLB` columns must be unbounded (`MEDIUMBLOB` / `LONGBLOB`), but this is just a temporal solution.
- Introducing pruning of materialization interval history.

Please tell me if you need more information. Thanks in advance!

nikolauspschuetz:
## What

The SQL registry stores each Feast object as a serialized protobuf in a binary column. On MySQL/MariaDB, SQLAlchemy's `LargeBinary` maps to `BLOB`, which caps at **64 KB**. A single `FeatureView` proto routinely exceeds that, so MySQL **silently truncates** the write and the registry later fails to load with a protobuf `DecodeError` (e.g. `feast serve` failing to start). PostgreSQL and SQLite were never affected.

Fixes #5800.

## How

- Introduce a dialect-aware `ProtoBytes` type that emits `LONGBLOB` on **MySQL and MariaDB** while keeping `LargeBinary`'s default mapping on every other dialect (`BLOB` on SQLite, `BYTEA` on PostgreSQL), and apply it to all binary proto/metadata columns.
  - MariaDB is registered separately because SQLAlchemy 2.x reports `dialect.name == "mariadb"`.
  - The variants are **chained** (not variadic) so the expression also works on SQLAlchemy 1.4.x, which Feast still supports (`SQLAlchemy>=1.4.19`).
- `metadata.create_all` only creates missing tables, so **existing MySQL registries are not migrated automatically**. Added a best-effort startup check (`_warn_if_narrow_blob_columns`) that, on MySQL/MariaDB, logs an **error** naming any columns still typed `BLOB` and points at the documented migration. It is logged at ERROR (so monitoring pipelines that filter below ERROR still catch it) but **deliberately does not refuse to start** — a registry whose protos all fit in 64 KB is unaffected and a routine upgrade should not break it. (Happy to switch to fail-fast-with-escape-hatch if maintainers prefer.)
- Documented the migration in the SQL registry reference, including the metadata-lock risk, a mandatory "stop writers first" checklist, and online-schema-change guidance (`pt-online-schema-change` / `gh-ost`). Note: `BLOB`→`LONGBLOB` is a column type change, which InnoDB performs with `ALGORITHM=COPY` (rebuild) — `INPLACE` is not supported.

## Tests

- **Unit** (`tests/unit/infra/registry/test_sql_registry.py`): assert the compiled DDL emits `LONGBLOB` on MySQL/MariaDB, `BLOB` on SQLite, `BYTEA` on PostgreSQL; cover the startup-check paths (errors on stale, silent when migrated, skips non-MySQL, never raises on query failure, runs on both read+write engines).
- **Integration** (`tests/integration/registration/test_universal_registry.py`): `test_apply_feature_view_large_proto_roundtrip_mysql` applies a `FeatureView` with a >64 KB proto on the live `mysql_registry` / `mysql_registry_async` fixtures and reads it back from the DB (`allow_cache=False`), asserting byte-identity — the regression guard for the truncation bug.

## Docs / agent guidance

- `docs/reference/registries/sql.md` — new "serialized-proto columns use LONGBLOB" section + migration runbook.
- Updated the in-repo agent rules/skills (`feast-components` rule pair, `feast-architecture` and `feast-testing` SKILLs) so the `ProtoBytes` convention and test patterns are discoverable.

### DIFF
--- a/.claude/rules/feast-components.md
+++ b/.claude/rules/feast-components.md
@@ -24,6 +24,7 @@ For testing patterns and debugging, also read `skills/feast-testing/SKILL.md`.
 
 - **Unit tests**: add or update tests in `sdk/python/tests/unit/infra/<subsystem>/`
 - **Integration tests**: run `make test-python-integration-local`; add a universal test case in `sdk/python/tests/integration/` if the change affects retrieval or materialization behavior
+- **SQL registry binary columns**: in `infra/registry/sql.py`, a new column that stores a serialized proto or blob metadata must use `ProtoBytes`, not `LargeBinary` directly — `LargeBinary` maps to MySQL `BLOB` (64 KB cap) and silently truncates large protos
 - **Protos**: if you add a field to a proto message, recompile with `make protos` and update serialization helpers in `proto_registry_utils.py`
 - **Both SDKs**: if the change affects online serving, check whether the Go server (`go/`) also needs updating
 - **Skills/Rules**: if the change introduces new patterns, interfaces, or conventions that agents should follow, update the relevant section in `skills/feast-architecture/SKILL.md` (and `skills/feast-testing/SKILL.md` if testing patterns changed)
--- a/.cursor/rules/feast-components.mdc
+++ b/.cursor/rules/feast-components.mdc
@@ -20,6 +20,7 @@ For testing patterns and debugging, also read `skills/feast-testing/SKILL.md`.
 
 - **Unit tests**: add or update tests in `sdk/python/tests/unit/infra/<subsystem>/`
 - **Integration tests**: run `make test-python-integration-local`; add a universal test case in `sdk/python/tests/integration/` if the change affects retrieval or materialization behavior
+- **SQL registry binary columns**: in `infra/registry/sql.py`, a new column that stores a serialized proto or blob metadata must use `ProtoBytes`, not `LargeBinary` directly — `LargeBinary` maps to MySQL `BLOB` (64 KB cap) and silently truncates large protos
 - **Protos**: if you add a field to a proto message, recompile with `make protos` and update serialization helpers in `proto_registry_utils.py`
 - **Both SDKs**: if the change affects online serving, check whether the Go server (`go/`) also needs updating
 - **Skills/Rules**: if the change introduces new patterns, interfaces, or conventions that agents should follow, update the relevant section in `skills/feast-architecture/SKILL.md` (and `skills/feast-testing/SKILL.md` if testing patterns changed)
--- a/docs/reference/registries/sql.md
+++ b/docs/reference/registries/sql.md
@@ -83,7 +83,84 @@ If you are running Feast in Kubernetes, set the `image.repository` and
 There are some things to note about how the SQL registry works:
 - Once instantiated, the Registry ensures the tables needed to store data exist, and creates them if they do not.
 - Upon tearing down the feast project, the registry ensures that the tables are dropped from the database.
-- The schema for how data is laid out in tables can be found . It is intentionally simple, storing the serialized protobuf versions of each Feast object keyed by its name.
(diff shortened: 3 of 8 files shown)


nikolauspschuetz:
@franciscojavierarceo @ntkathole — would appreciate a review when you have a moment. 🙏

This fixes #5800: on MySQL/MariaDB, SQLAlchemy's `LargeBinary` maps to `BLOB` (64 KB cap), so a serialized `FeatureView` proto over 64 KB is **silently truncated** on write and later fails to load with a `DecodeError` (e.g. `feast serve` won't start). PostgreSQL/SQLite are unaffected.

Approach:
- Dialect-aware `ProtoBytes` type → `LONGBLOB` on MySQL/MariaDB, default mapping elsewhere; applied to all serialized-proto columns. Variants are chained (not variadic) to stay compatible with the supported SQLAlchemy 1.4.x floor.
- Best-effort startup diagnostic that logs (at ERROR) any registry proto column still typed `BLOB` and points at a documented `ALTER TABLE` migration — `create_all` never widens existing columns, so existing MySQL registries aren't auto-migrated.
- Tests: DDL-compile unit tests (LONGBLOB on MySQL/MariaDB, BLOB/BYTEA elsewhere) + a live >64 KB round-trip integration test on the `mysql_registry` fixtures + diagnostic-path coverage.
- Docs: SQL-registry reference gains a migration runbook (lock-safety + online-schema-change guidance). Agent skills/rules updated per the repo's `.claude/rules` convention.

Happy to adjust anything — including whether the startup check should hard-fail (with an env-var escape hatch) instead of logging, if you'd prefer that posture.

PR #6569, opened 2026-06-29
feat: Retrieve historical features from BigQuery without entity_df

### LINKED ISSUE #5830: BigQuery - historical retrieval without entity dataframe

**Additional context**

singhhimanshu0811:
-Adds non-entity mode to `BigQueryOfflineStore.get_historical_features`
- When `entity_df=None`, synthesizes a left table via `UNION DISTINCT` of entity keys observed in all feature view sources within the time window, using `end_date` as the as-of timestamp for PIT joins
- Adds helpers: `_gather_all_entities` and `_bq_create_entity_union_table`

Fixes #5830 

# Checks

## Testing Strategy
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual tests

# Misc


singhhimanshu0811:
Hi @sudohainguyen i have tested code on my local manually and it seems to be working fine - will be adding more tests in the meanwhile. Can you review code once.

singhhimanshu0811:
@ntkathole @franciscojavierarceo can you trigger tests

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -297,36 +300,79 @@ def get_historical_features(
...
+                start_date=kwargs.get("start_date"),
+                end_date=kwargs.get("end_date"),
+            )
+            entity_df_event_timestamp_range = (start_date, end_date)
 
-        @contextlib.contextmanager
-        def query_generator() -> Iterator[str]:
-            _upload_entity_df(
+            # Pre-compute query contexts to collect entity column names per feature view.
+            fv_query_contexts_pre = offline_utils.get_feature_view_query_context(
```
ntkathole, line 316: with this call added now, `get_feature_view_query_context` is called twice in non-entity mode with the same arguments: once at line 316 (fv_query_contexts_pre) and again at line 378 inside query_generator. The second call is redundant. Reuse fv_query_contexts_pre directly inside the generator for the non-entity branch to avoid the duplicate work.

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -297,36 +300,79 @@ def get_historical_features(
...
+                registry,
+                project,
+                entity_df_event_timestamp_range,
+            )
+            all_entities = offline_utils.gather_all_entities(fv_query_contexts_pre)
+            event_timestamp_col = "entity_ts"
+            entity_schema_keys: KeysView[str] = cast(
+                KeysView[str],
+                {k: None for k in (all_entities + [event_timestamp_col])}.keys(),
+            )
```
ntkathole, line 328:
```suggestion
            entity_schema_keys = list(all_entities) + [event_timestamp_col]
```

sdk/python/feast/infra/offline_stores/bigquery.py
```diff
@@ -567,6 +613,62 @@ def clear_monitoring_baseline(
...
+        from_expression = fv.batch_source.get_table_query_string()
+        timestamp_field = ctx.timestamp_field
+
+        ctx_entities_set = set(ctx.entities)
+        select_entities: List[str] = []
+        for col in all_entities:
+            if col in ctx_entities_set:
+                select_entities.append(f"`{col}`")
+            else:
+                select_entities.append(f"NULL AS `{col}`")
```
ntkathole:
```suggestion
            else:
                # Explicit STRING cast prevents BigQuery type-mismatch errors
                # when UNION DISTINCT combines branches with heterogeneous entity types.
                select_entities.append(f"CAST(NULL AS STRING) AS `{col}`")
```

ntkathole: better to do cast both:

```
        select_entities: List[str] = []
        for col in all_entities:
            if col in ctx_entities_set:
                select_entities.append(f"CAST(`{col}` AS STRING) AS `{col}`")
            else:
                select_entities.append(f"CAST(NULL AS STRING) AS `{col}`")
```

singhhimanshu0811: @ntkathole what if the entity value is not string? join will fail!!

ntkathole: This will fail at runtime where:

- Feature view A has entity columns [customer_id, item_id]
- Feature view B has entity columns [customer_id] (no item_id)
- item_id in the source table is type STRING

The generated SQL would be:

```
SELECT DISTINCT `customer_id`, `item_id`        -- item_id is STRING
FROM `project.dataset.orders`
WHERE ...
UNION DISTINCT
SELECT DISTINCT `customer_id`, NULL AS `item_id` -- item_id is INT64 (default)
FROM `project.dataset.customers`
WHERE ...
```
BigQuery will reject this with: "Column 2 in UNION DISTINCT has incompatible types: STRING, INT64"

singhhimanshu0811: done

sdk/python/feast/infra/offline_stores/bigquery.py
singhhimanshu0811: done

sdk/python/feast/infra/offline_stores/bigquery.py
singhhimanshu0811: hey @ntkathole other functions like build_point_in_time_query want KeysView only, list wont work.

sdk/python/feast/infra/offline_stores/bigquery.py
singhhimanshu0811: @ntkathole this cast should only be done for null case right? not the non null case? (the if block)?

ntkathole: I think both - check hpw it's done in spark.py

singhhimanshu0811: I think its working fine there because of spark magic. for bigquery its fails if we cast as string.

singhhimanshu0811: Spark does automatically coerce (or implicitly cast) data types in its SQL engine during operations like joins, unions, and aggregations. for bigquery it fails

singhhimanshu0811: so i registered these feature views 

`
user_activity_source = BigQuerySource(
    name="user_activity_source",
    table="wmt-gtp-featurestore-dev.feast_dataset.fs_clone",
    timestamp_field="creation_date"
)

user_entity = Entity(name="id", join_keys=["id"], value_type=ValueType.INT64)
user_entity1 = Entity(name="age", join_keys=["age"], value_type=ValueType.STRING)

user_activity_fv = FeatureView(
    name="feature_set_check",
    entities=[user_entity],
    ttl=timedelta(days=30),
    schema=[
        Field(name="display_name", dtype=String),
        Field(name="about_me", dtype=String),
    ],
    online=True,
    source=user_activity_source,
)

user_activity_fv1 = FeatureView(
    name="feature_set_check1",
    entities=[user_entity, user_entity1],
    ttl=timedelta(days=30),
    schema=[
        Field(name="display_name", dtype=String),
        Field(name="about_me", dtype=String),
    ],
    online=True,
    source=user_activity_source,
)

training_data = store.get_historical_features(
    features=[
        "feature_set_check:display_name",
        "feature_set_check:about_me",
    ],
    start_date = datetime(1980, 1, 1),
    end_date = datetime(2023, 1, 1)
).to_df()

print(training_data.head(10))

training_data = store.get_historical_features(
    features=[
        "feature_set_check1:display_name",
        "feature_set_check1:about_me",
    ],
    start_date = datetime(1980, 1, 1),
    end_date = datetime(2023, 1, 1)
).to_df()

print(training_data.head(10))`


I got correct results. 

@ntkathole can you kindly tell the reason why cast to string is being done in spark, especially in non null case?

singhhimanshu0811: i tried this also 

`training_data = store.get_historical_features(
    features=[
        "feature_set_check:display_name",
        "feature_set_check1:about_me"
    ],
    start_date = datetime(1980, 1, 1),
    end_date = datetime(2023, 1, 1)
).to_df()

print(training_data.head(10))`

and this also works fine. I think cast as string is only needed for NULL case only. It should not be done for non null case. even though its being done in spark, spark auto coercive datatype is handling it . but bigquery is much more strict and is not working with cast!!

ntkathole: I see, bigquery automatically coerces it https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/conversion_rules. Seems Spark handles the type mismatch implicitly.

singhhimanshu0811: Yeah for null cases it's needed. (Second message example)

singhhimanshu0811:
Hey @ntkathole are changes are done on my side, and it seems to be working fine. Can you review it once? Please let me know in case of any other issues!

ntkathole (approved):

lvijnck:
When should this come in? Getting the error below, even on the latest pre-release:

```
feast-features         | TypeError: BigQueryOfflineStore.get_historical_features() got an unexpected 
feast-features         | keyword argument 'start_date'
```

**edit**

Nevermind, my UV workspace was pinning versions.

PR #6570, opened 2026-06-29
fix: Unblock nightly UI build

franciscojavierarceo:
## Summary
- remove unused UI imports that fail the release wheel build under `CI=true`
- keep the change scoped to the two current warnings reproduced on `master`

## Root cause
The nightly Python SDK release calls the reusable wheel build workflow, which runs `make build-ui`. React treats ESLint warnings as errors when `CI=true`, so unused imports in the UI stopped the release before wheel creation or publishing.

## Validation
- `CI=true make build-ui`


ntkathole:
@copilot We need to add check that would catch these issues (npm run build) before they reach master.

Also, fix :
- 'setAllowOverwrite' assigned but unused  from CreateDatasetForm.tsx 
- 'isSuccess' assigned but unused  from DatasetOverviewTab.tsx 
- 'EuiHorizontalRule' imported but unused from DatasetUsageTab.tsx

PR #6587, opened 2026-07-07
feat: Apply Intermediate TLS defaults on API fallback and handle transient errors

ugiordan:

Follow-up to #6567. Fixes two issues in the TLS profile integration:

1. **Else-only pattern**: `NewTLSConfigFromProfile` was only called in the success branch of the TLS profile fetch. On any error path (non-OpenShift cluster, resource not found), no TLS config was applied, leaving the operator running with Go's bare defaults (no MinVersion, no cipher restrictions). This PR moves `NewTLSConfigFromProfile` outside the if/else and explicitly falls back to `configv1.TLSProfiles[TLSProfileIntermediateType]` on all error paths.

2. **No transient error handling**: The error switch only handled `IsNotFound` and `IsNoMatchError`. Transient API errors (`ServiceUnavailable`, `Timeout`, `ServerTimeout`, `TooManyRequests`, `context.DeadlineExceeded`) hit the default case and crashed the operator with `os.Exit(1)`. This PR adds a dedicated case for these errors with a graceful Intermediate fallback. `tlsProfileFetched` is set to `true` so the `SecurityProfileWatcher` self-heals when the API recovers.

Follow-up to #6567

## Testing

- `go build ./...` in `infra/feast-operator`
- `go test ./...` in `infra/feast-operator`

### DIFF
--- a/infra/feast-operator/cmd/tls_bootstrap.go
+++ b/infra/feast-operator/cmd/tls_bootstrap.go
@@ -1,0 +1,125 @@
+/*
+Copyright 2024 Feast Community.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"crypto/tls"
+	"errors"
+	"fmt"
+	"time"
+
+	configv1 "github.com/openshift/api/config/v1"
+	tlspkg "github.com/openshift/controller-runtime-common/pkg/tls"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
+	tlsFetchTimeout = 10 * time.Second
+	alpnHTTP11      = "http/1.1"
+)
+
+type tlsBootstrapResult struct {
+	TLSOpts            []func(*tls.Config)
+	ProfileFetched     bool
+	ProfileSpec        configv1.TLSProfileSpec
+	AdherenceFetched   bool
+	AdherencePolicy    configv1.TLSAdherencePolicy
+	UnsupportedCiphers []string
+}
+
+func fetchTLSProfile(ctx context.Context, k8sClient client.Client) (configv1.TLSProfileSpec, bool, error) {
+	fetchCtx, cancel := context.WithTimeout(ctx, tlsFetchTimeout)
+	defer cancel()
+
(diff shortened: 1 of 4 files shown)


franciscojavierarceo (changes requested):
The added Notebook ConfigMap controller, validation/polling helpers, RBAC/startup behavior, and `os.Exit(0)` restart path are unrelated to the TLS fallback fix and substantially enlarge its operational risk. Please split that controller work into a separate PR so the TLS change can be reviewed and reverted independently. The TLS error-classification/default-profile behavior also needs focused unit coverage in this PR; the current diff changes only `cmd/main.go` and adds no tests.

ugiordan:
> The added Notebook ConfigMap controller, validation/polling helpers, RBAC/startup behavior, and `os.Exit(0)` restart path are unrelated to the TLS fallback fix and substantially enlarge its operational risk. Please split that controller work into a separate PR so the TLS change can be reviewed and reverted independently. The TLS error-classification/default-profile behavior also needs focused unit coverage in this PR; the current diff changes only `cmd/main.go` and adds no tests.

Thanks for the review @franciscojavierarceo. I've addressed both points:
1. Removed the Notebook ConfigMap controller code from this PR and will open it as a separate PR. 

2. Extracted the TLS bootstrap logic into `cmd/tls_bootstrap.go` and added comprehensive unit tests in `cmd/tls_bootstrap_test.go` covering all error classification edge cases.

The diff is now a single commit touching only TLS error handling.

infra/feast-operator/cmd/tls_bootstrap.go
```diff
@@ -0,0 +1,125 @@
...
+
+	tlsConfigFn, unsupported := tlspkg.NewTLSConfigFromProfile(profile)
+	result.UnsupportedCiphers = unsupported
+	if len(unsupported) > 0 {
+		logger.Info("TLS profile contains ciphers unsupported by Go", "unsupported", unsupported)
+	}
+	result.TLSOpts = append(result.TLSOpts, tlsConfigFn)
+
+	adherence, adherenceFetched, err := fetchTLSAdherencePolicy(ctx, k8sClient)
+	if err != nil {
```
ntkathole, line 115: dead code - function returns error but can never return a non-nil error.

infra/feast-operator/cmd/tls_bootstrap.go
```diff
@@ -0,0 +1,125 @@
...
+		return nil, err
+	}
+	result.AdherenceFetched = adherenceFetched
+	result.AdherencePolicy = adherence
+
+	result.TLSOpts = append(result.TLSOpts, func(c *tls.Config) {
+		c.NextProtos = []string{"h2", alpnHTTP11}
```
ntkathole: alpnHTTP11 is defined as a constant but "h2" is hardcoded inline. Either both should be constants or neither.

infra/feast-operator/cmd/tls_bootstrap.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+}
+
+func fetchTLSAdherencePolicy(ctx context.Context, k8sClient client.Client) (configv1.TLSAdherencePolicy, bool, error) {
+	fetchCtx, cancel := context.WithTimeout(ctx, tlsFetchTimeout)
+	defer cancel()
+
+	policy, err := tlspkg.FetchAPIServerTLSAdherencePolicy(fetchCtx, k8sClient)
+	if err != nil {
+		return configv1.TLSAdherencePolicy(""), false, nil
```
ntkathole: fetchTLSProfile has proper error classification but here silently swallows all errors

ntkathole (approved):

Issue #4526, opened 2024-09-17
Remote materialization


tokoko:
**Is your feature request related to a problem? Please describe.**
We already have an option to run online/offline store queries remotely through feature server and offline server, respectively. This way rbac rules will be applied on operations. One piece that's missing is materialization. There are several ways to do this:

- Keep materialization local, but rely on remote online/offline engines to apply rbac rules. This is currently impossible because remote online client doesn't implement `online_write_batch` method. Even if we did implement it in the feature server itself, we would essentially be using a lightweight fastapi server to transport batches and batches of potentially huge datasets.

- Create remote materialization engine that will defer the whole materialization call to a backend server and apply rbac rules there. We can create another server component MaterializationServer that will receive these requests.

- The same as above but instead of creating a new component, we can reuse OfflineServer to do the request handling. This is slightly awkward from the naming perspective, but probably makes the most sense in term of usage/maintenance.

I'd probably go with option 3 as a starting point.

dmartinol:
> The same as above but instead of creating a new component, we can reuse OfflineServer to do the request handling. This is slightly awkward from the naming perspective, but probably makes the most sense in term of usage/maintenance.

ATM the `OfflineServer` was designed to implement the `OfflineStore` interface only. Adding a method to write to the online stores would introduce an unplanned dependency and raise some concerns.
- How the actual method should be implemented on the server side, w/o a reference to the (likely `remote`) `online_store`, which would raise again the original issue?
- If we instead think to implement it in the `OfflineServer`, then we'd need an actual `online_store` configuration as well, which was not planned to have in such server (a `remote` online store was designed instead). Not sure about any possible side effects.

Why aren't we using the `/materialize-incremental` endpoint on the `FeatureServer` instead? (and add a new endpoint for non-incremental jobs)
This would avoid any "transport batches and batches of potentially huge datasets." as it would work on the server itself (and would use the remote `offline_store` to `pull_latest_from_table_or_query` using the flight protocol) .

Otherwise, I'd be in favour of a dedicated `MaterializationServer` (with remote `offline_store` and provided `online_store`), which can still be designed as a "lightweight fastapi server" if I understood the materialization flow.

dmartinol:
@tokoko do we want to evaluate this one? Any further comments on what solution to apply?

tmihalac:
@tokoko we would like to see how do we progress with the remote materialization, how do we progress from here ?

PR #6590, opened 2026-07-08
feat: Add remote materialization support via feature server

aniketpalu:

Allow store.materialize() and materialize_incremental() to delegate execution to the feature server instead of running the batch engine locally. This eliminates the need for engine-specific dependencies (PySpark, kubernetes client, etc.) and RBAC on the client pod. Components:
- New shared poll_materialization_status() function
- Client-side _remote_materialize() with registry-based FV state polling
- POST /materialize-async and /materialize-incremental-async endpoints (202)
- materialize_mode, url, timeout, poll_interval config fields
- 19 unit tests + 2 E2E integration tests Addresses feast-dev/feast#4526 (remote materialization).

# Checks

## Testing Strategy
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

# Misc


sdk/python/feast/infra/feature_servers/local_process/config.py
```diff
@@ -9,3 +11,25 @@ class LocalFeatureServerConfig(BaseFeatureServerConfig):
 
     # The endpoint definition for transformation_service
     transformation_service_endpoint: str = "localhost:6569"
+
+    materialize_mode: Literal["local", "remote"] = "local"
+    """When 'remote', store.materialize() delegates to the feature server's
+    async endpoint instead of running the batch engine locally."""
```
jyejare: I think we dont need to provide the materialize_mode config. We should strictly feature server take care of materialization delegation to compute engine. Leaving no choice to do materialization on client side.

ntkathole: A cleaner approach might be: a RemoteComputeEngine (like RemoteOnlineStore/RemoteOfflineStore) that POSTs to the existing /materialize endpoint, returning a job handle with a dedicated /materialize-status/{job_id} endpoint - keeping the materialization delegation at the engine layer rather than the FeatureStore layer.

aniketpalu: if you mean something like 

```
batch_engine:
  type: remote.engine
  path: https://feast-online:443   # same host as online
  cert: /var/run/secrets/.../service-ca.crt
```

I thought about this before choosing current approach. I see two points which made me not take this decision.
1. This adds another config section which has to indicate only one thing, i.e. trigger materialization on feature server. For this, it add 1 config section & user has to configure 2 redundant fields.
2. Once client is initialized, user wont have option to try normal materialization flow if remote materialization fails for any reason. 

My proposed approach is

```
store.materialize(remote=True, wait=False) --> pre-process check & return 202 Accepted response
store.poll_materialization() --> returns status
```

This approach has 
1. Backward compatibility - User can fall back without re-initializing client.
2. Flexibility of Status polling - Full control to user to set interval of polling. 
3. No need of Operator changes. 
4. The same API can work for UI as well

ntkathole: @aniketpalu Ok, so, I think this PR is mixing three orthogonal concerns: async execution, remote delegation, and job tracking.

- Remote delegation already works today - POST /materialize runs materialization server-side. A client with `online_store: remote` can call this endpoint directly. As @jyejare mentioned, the feature server should strictly handle materialization - no per-call remote=True flag needed; the deployment topology already decides this.

- Async execution - The real gap is that long-running engines (SparkApplication) block the HTTP caller. The existing /materialize endpoint just needs an `async` mode (?async=true → 202 with job_id) rather than entirely new `/materialize-async` endpoints. This serves both UI on-demand requests and SDK wait=False use cases, while CronJob continues using the existing sync path unchanged.

- Job tracking - Polling FeatureViewState from the registry is fragile (race conditions, no job_id, ambiguous failure, breaks FeatureView.__eq__). If we need job tracking, it should be a dedicated MaterializationJob entity in the registry with proper RPCs - not piggybacking on FV metadata. For SparkApplication, the SparkApp CR status on K8s is already the source of truth.

May be we can scope this PR to just remote delegation and async execution ? `?async=true` on the existing `/materialize` endpoint (fire-and-forget, returns 202. Completion is already observable via existing mechanisms like materialization_intervals on the FV, SparkApp CR status) + SDK support for async with remote delegation. 

Job tracking with dedicated status endpoints / registry entities can come as a separate effort.

aniketpalu: > Remote delegation already works today - POST /materialize runs materialization server-side. A client with online_store: remote can call this endpoint directly. As @jyejare mentioned, the feature server should strictly handle materialization - no per-call remote=True flag needed; the deployment topology already decides this.
> Async execution - The real gap is that long-running engines (SparkApplication) block the HTTP caller. The existing /materialize endpoint just needs an async mode (?async=true → 202 with job_id) rather than entirely new /materialize-async endpoints. This serves both UI on-demand requests and SDK wait=False use cases, while CronJob continues using the existing sync path unchanged.

agree on both point

sdk/python/feast/feature_server.py
```diff
@@ -708,6 +710,152 @@ async def materialize_incremental(request: MaterializeIncrementalRequest) -> Non
...
+        )
+        for fv in feature_views_to_materialize:
+            assert_permissions(
+                resource=fv,
+                actions=[AuthzedAction.WRITE_ONLINE],
+            )
+        return [fv.name for fv in feature_views_to_materialize]
+
+    def _parse_materialize_timestamps(request: MaterializeRequest):
+        """Parse and validate timestamps from a materialize request."""
```
jyejare: **[Critical] Function defined inside endpoint handler scope**

The function `_authorize_materialize_views` is defined inside the endpoint handler scope but references `store` from the outer scope. This creates a closure that captures the `store` variable, which could lead to unexpected behavior if the store changes or if this function is called in different contexts. This pattern violates the principle of explicit dependencies and makes testing difficult.

**Suggested:**
```suggestion
def _authorize_materialize_views(
    store: FeatureStore, request_feature_views: Optional[List[str]]
) -> List[str]:
    """Shared authz + FV name resolution for async materialize endpoints."""
    feature_views_to_materialize = store._get_feature_views_to_materialize(
        request_feature_views
    )
    for fv in feature_views_to_materialize:
        assert_permissions(
            resource=fv,
            actions=[AuthzedAction.WRITE_ONLINE],
        )
    return [fv.name for fv in feature_views_to_materialize]
```

sdk/python/feast/feature_server.py
```diff
@@ -708,6 +710,152 @@ async def materialize_incremental(request: MaterializeIncrementalRequest) -> Non
...
+                if getattr(fv, "state", None) == FeatureViewState.MATERIALIZING:
+                    return JSONResponse(
+                        status_code=409,
+                        content={
+                            "error": f"Feature view '{fv_name}' is already materializing"
+                        },
+                    )
+            except Exception:
+                pass
+        return None
```
jyejare: **[Critical] Function defined inside endpoint handler with closure dependencies**

Similar to the previous issue, `_check_already_materializing` and other helper functions are defined inside the endpoint handler and capture variables from outer scope. This makes the code harder to test and reason about. These functions should be extracted as proper module-level functions with explicit parameters.

**Suggested:**
```suggestion
def _check_already_materializing(
    store: FeatureStore, fv_names: List[str]
) -> Optional[JSONResponse]:
    """Return 409 if any requested FV is already materializing."""
    for fv_name in fv_names:
        try:
            fv = store.registry.get_feature_view(
                fv_name, store.project, allow_cache=False
            )
            if getattr(fv, "state", None) == FeatureViewState.MATERIALIZING:
                return JSONResponse(
                    status_code=409,
                    content={
                        "error": f"Feature view '{fv_name}' is already materializing"
                    },
                )
        except Exception:
            pass
    return None
```

sdk/python/feast/feature_server.py
```diff
@@ -708,6 +710,152 @@ async def materialize_incremental(request: MaterializeIncrementalRequest) -> Non
...
+
+            conflict = _check_already_materializing(fv_names)
+            if conflict:
+                return conflict
+
+            _update_fv_state(fv_names, FeatureViewState.MATERIALIZING)
+
+            def _run_materialize():
+                try:
+                    store.materialize(
```
jyejare: **[Warning] Materialize async endpoint lacks proper error handling**

The `/materialize-async` endpoint catches all exceptions in the background thread but only logs them. If the materialization fails during the initial setup (before the background thread), the state won't be properly reset. The endpoint should have a try-catch around the initial setup and reset states appropriately.

**Suggested:**
```suggestion
def _run_materialize():
    try:
        store.materialize(
            start_date,
            end_date,
            request.feature_views,
            disable_event_timestamp=request.disable_event_timestamp,
            full_feature_names=request.full_feature_names,
            version=request.version,
            _force_local=True,
        )
        # Set state to AVAILABLE_ONLINE on success
        _update_fv_state(fv_names, FeatureViewState.AVAILABLE_ONLINE)
    except Exception as e:
        logger.error(
            f"Async materialization failed for {fv_names}: {e}",
            exc_info=True,
        )
        _update_fv_state(fv_names, FeatureViewState.GENERATED)
```

sdk/python/feast/feature_store.py
```diff
@@ -454,13 +460,18 @@ def _transition_fv_to_materializing(
         feature_view,
         already_transitioned: list,
         previous_states: dict,
+        _force_local: bool = False,
```
jyejare: **[Warning] Empty return in state transition function breaks existing logic**

Adding an early return when `_force_local=True` means the `previous_states[feature_view.name] = previous_state` line is never reached, which could break rollback logic elsewhere. The function should still record the previous state even when skipping the transition.

**Suggested:**
```suggestion
Skip when _force_local: the async handler already set MATERIALIZING.
"""
previous_state = getattr(feature_view, "state", None)
previous_states[feature_view.name] = previous_state
if _force_local:
    return  # Skip state transition but preserve previous_state recording
```

sdk/python/feast/feature_store.py
```diff
@@ -2359,12 +2382,265 @@ def _materialize_odfv(
...
+            return FVMaterializationStatus.RUNNING
+        if state == FeatureViewState.GENERATED:
+            if seen_materializing is not None and fv_name in seen_materializing:
+                return FVMaterializationStatus.FAILED
+            return FVMaterializationStatus.PENDING
+        return FVMaterializationStatus.PENDING
+
+    def poll_materialization(
+        self,
+        feature_views: Optional[List[str]] = None,
```
jyejare: **[Warning] State mapping logic could be more explicit**

The `_fv_state_to_materialization_status` function uses a mutable `seen_materializing` set that's passed by reference. This side-effect pattern makes the function harder to reason about and test. Consider returning a tuple or using a more functional approach.

**Suggested:**
```suggestion
def _fv_state_to_materialization_status(
    self, state, *, seen_materializing: Optional[set] = None, fv_name: str = ""
) -> Tuple[FVMaterializationStatus, bool]:
    """Map FeatureViewState to FVMaterializationStatus.
    
    Returns:
        Tuple of (status, was_materializing) where was_materializing indicates
        if this FV was seen in MATERIALIZING state.
    """
    from feast.materialization_status import FVMaterializationStatus

    if state == FeatureViewState.AVAILABLE_ONLINE:
        return FVMaterializationStatus.SUCCEEDED, False
    if state == FeatureViewState.MATERIALIZING:
        if seen_materializing is not None:
            seen_materializing.add(fv_name)
        return FVMaterializationStatus.RUNNING, True
    if state == FeatureViewState.GENERATED:
        was_materializing = seen_materializing is not None and fv_name in seen_materializing
        status = FVMaterializationStatus.FAILED if was_materializing else FVMaterializationStatus.PENDING
        return status, was_materializing
    return FVMaterializationStatus.PENDING, False
```

sdk/python/feast/feature_store.py
```diff
@@ -2359,12 +2382,265 @@ def _materialize_odfv(
...
+        disable_event_timestamp: bool = False,
+        full_feature_names: bool = False,
+        version: Optional[str] = None,
+        timeout: float = 3600.0,
+        poll_interval: float = 5.0,
+        wait: bool = True,
+    ) -> None:
+        """Delegate materialization to the feature server, poll via registry."""
+        fv_names = feature_views or [
```
jyejare: **[Warning] HTTP session not properly closed in all paths**

The HTTP session is closed in a finally block, but if an exception occurs during session creation or configuration, the session might never be assigned, leading to a NameError. The session creation and closure should be more robust.

**Suggested:**
```suggestion
session = None
try:
    session = self._get_remote_http_session()
    post_kwargs: dict = {
        "json": payload,
        "timeout": 30,
    }
    cert = getattr(self.config.online_store, "cert", None)
    if cert:
        post_kwargs["verify"] = cert
    resp = session.post(f"{server_url}{endpoint}", **post_kwargs)
    resp.raise_for_status()
except Exception as e:
    raise Exception(
        f"Failed to trigger remote materialization at "
        f"{server_url}{endpoint}: {e}"
    ) from e
finally:
    if session is not None:
        session.close()
```

jyejare (commented):
This PR adds remote materialization support via feature server, allowing clients to delegate materialization to a remote server instead of running it locally. The implementation includes async endpoints in the feature server, client-side remote delegation logic, status polling mechanisms, and comprehensive test coverage. The feature is well-architected with proper error handling and state management.

jyejare:
@aniketpalu Please fix DCO check for incorrect signoff. Also please update docs for remote materialization support.

sdk/python/feast/infra/feature_servers/local_process/config.py
aniketpalu: @ntkathole @jyejare 

To keep the changes and commit history clean, I have raised separate PR. Would love your feedback on that. https://github.com/feast-dev/feast/pull/6649

PR #6591, opened 2026-07-09
fix(operator): Only deploy an online store when it is declared

### LINKED ISSUE #6586: Feast operator always deploys an online store, regardless of CR configuration
## Expected Behavior 

The Feast operator should deploy only the services declared in a FeatureStore CR. It should be possible to deploy a standalone registry or a standalone offline store, with no online store, by simply not specifying spec.services.onlineStore. This mirrors how registry and offlineStore already behave and is required by the documented [federated / shared registry pattern](https://docs.feast.dev/how-to-guides/feast-snowflake-gcp-aws/federated-feature-store).

## Current Behavior

The operator always deploys an online store service (including a serving pod), even when the CR does not declare spec.services.onlineStore. Root cause is in ApplyDefaultsToStatus ([internal/controller/services/util.go](https://github.com/feast-dev/feast/blob/master/infra/feast-operator/internal/controller/services/util.go#L166)):

```go
// default to onlineStore service deployment
if services.OnlineStore == nil {
    services.OnlineStore = &feastdevv1.OnlineStore{}
}
// ...
if services.OnlineStore.Server == nil {
    services.OnlineStore.Server = &feastdevv1.ServerConfigs{}   // forces a serving pod
}
```

Unlike registry `(if services.Registry != nil)` and `offlineStore (if services.OfflineStore != nil)`, which are only defaulted when declared, and whose `.Server` is only defaulted when already present, the onlineStore block runs unconditionally and creates Server from nil.

Downstream:

1. `isOnlineStore()` is always true → `reconcileServices()` always deploys the online store, and `isOnlineServer()` is always true. There's no opt-out, the `OnlineStore` CRD type has no `disabled` field. (This also makes the "at least one local server must be configured" guard in `Deploy()` unreachable.)
2. The scaling admission validation is asymmetric ([api/v1/featurestore_types.go](https://github.com/feast-dev/feast/blob/master/infra/feast-operator/api/v1/featurestore_types.go#L148)): the offline rule allows the store to be absent (`!has(self.services.offlineStore) || ...`), but the online rule requires it to exist and be DB-backed. So scaling a registry-only/offline-only store fails admission with `Scaling requires DB-backed persistence for the online store. Configure services.onlineStore.persistence.store when using replicas > 1 or autoscaling.`, forcing DB persistence for an online store the user never asked for.

## Steps to reproduce

1. Apply a CR declaring only a registry (or only an offline store), with no `spec.services.onlineStore`.
2. `kubectl get deploy -l feast.dev/name=<name>` → an online/feature-server container is deployed alongside the registry (expected: registry only).
3. Add `spec.replicas: 2` (or `services.scaling.autoscaling`) to that registry-only CR → rejected at admission with the online-store DB-persistence error, despite no online store being declared.

### Specifications

- Version: 0.64.0
- Platform: Kubernetes
- Subsystem: feast-operator — internal/controller/services

## Possible Solution

1. Guard the online-store block in `ApplyDefaultsToStatus` with if `services.OnlineStore != nil { ... }` (and only default Server when already present), making it symmetric with registry/offline. Preferable to adding a new opt-out API field.
2. Add an "online store absent" escape hatch to the online CEL rule (`!has(self.services.onlineStore) || (...)`), matching the offline rule.
3. Keep `noLocalCoreServerConfigured()` enforcing that at least one core server is deployed.

Happy to open a PR! 😄

nikolauspschuetz:
Closes #6586.

### Problem

The Feast operator always deployed an online store service (including a serving pod), even when the `FeatureStore` CR did not declare `spec.services.onlineStore`. This makes it impossible to run a standalone registry or offline store (e.g. the documented [federated / shared-registry pattern](https://docs.feast.dev/how-to-guides/feast-snowflake-gcp-aws/federated-feature-store)), and is inconsistent with how `registry` and `offlineStore` behave — both are only defaulted when declared.

### Cause

In `ApplyDefaultsToStatus` (`infra/feast-operator/internal/controller/services/util.go`), the online-store block force-created the service when nil:

```go
if services.OnlineStore == nil {
    services.OnlineStore = &feastdevv1.OnlineStore{}
}
// ...defaults always applied...
```

whereas `registry` and `offlineStore` are gated behind `if services.X != nil`.

### Fix

Gate the online-store defaulting behind `services.OnlineStore != nil`, mirroring the two sibling services. A declared online store still receives all of its defaults (persistence, path, PVC, server, container configs).

### Tests

Added `util_test.go` with two specs for `ApplyDefaultsToStatus`:
- a CR without an online store → `Status.Applied.Services.OnlineStore` stays `nil` (fails on `master`, passes with this change),
- a CR that declares an online store → it still receives its defaults.

Verified with `KUBEBUILDER_ASSETS=... go test ./internal/controller/services/` (envtest 1.31.0); `go build ./...`, `go vet`, and `gofmt` are clean. The existing controller tests are unaffected because they declare `onlineStore` explicitly.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -957,7 +957,7 @@
         "filename": "infra/feast-operator/api/v1/featurestore_types.go",
         "hashed_secret": "44e17306b837162269a410204daaa5ecee4ec22c",
         "is_verified": false,
-        "line_number": 937
+        "line_number": 942
       }
     ],
     "infra/feast-operator/api/v1/zz_generated.deepcopy.go": [
--- a/infra/feast-operator/api/v1/featurestore_types.go
+++ b/infra/feast-operator/api/v1/featurestore_types.go
@@ -544,6 +544,11 @@ type OnlineStore struct {
 	// Controls metrics granularity, offline push batching, and MCP.
 	// +optional
 	Serving *ServingConfig `json:"serving,omitempty"`
+	// Disabled skips deploying the online store service entirely, including its
+	// serving pod and persistence. Omitting the online store block, or setting
+	// this to false, deploys the online store with defaults as before.
+	// +optional
+	Disabled bool `json:"disabled,omitempty"`
 }
 
 // ServingConfig configures the feature_server section of the generated feature_store.yaml.
--- a/infra/feast-operator/config/crd/bases/feast.dev_featurestores.yaml
+++ b/infra/feast-operator/config/crd/bases/feast.dev_featurestores.yaml
@@ -2258,6 +2258,11 @@ spec:
                   onlineStore:
                     description: OnlineStore configures the online store service
                     properties:
+                      disabled:
+                        description: |-
+                          Disabled skips deploying the online store service entirely, including its
+                          serving pod and persistence.
+                        type: boolean
                       persistence:
                         description: OnlineStorePersistence configures the persistence
                           settings for the online store service
@@ -8562,6 +8567,11 @@ spec:
                       onlineStore:
                         description: OnlineStore configures the online store service
                         properties:
+                          disabled:
+                            description: |-
+                              Disabled skips deploying the online store service entirely, including its
+                              serving pod and persistence.
+                            type: boolean
                           persistence:
                             description: OnlineStorePersistence configures the persistence
                               settings for the online store service
--- a/infra/feast-operator/dist/install.yaml
+++ b/infra/feast-operator/dist/install.yaml
@@ -2266,6 +2266,11 @@ spec:
                   onlineStore:
                     description: OnlineStore configures the online store service
                     properties:
+                      disabled:
+                        description: |-
+                          Disabled skips deploying the online store service entirely, including its
+                          serving pod and persistence.
+                        type: boolean
                       persistence:
                         description: OnlineStorePersistence configures the persistence
                           settings for the online store service
@@ -8570,6 +8575,11 @@ spec:
                       onlineStore:
                         description: OnlineStore configures the online store service
                         properties:
+                          disabled:
+                            description: |-
+                              Disabled skips deploying the online store service entirely, including its
+                              serving pod and persistence.
+                            type: boolean
                           persistence:
                             description: OnlineStorePersistence configures the persistence
                               settings for the online store service
--- a/infra/feast-operator/docs/api/markdown/ref.md
+++ b/infra/feast-operator/docs/api/markdown/ref.md
@@ -654,6 +654,9 @@ _Appears in:_
 | `persistence` _[OnlineStorePersistence](#onlinestorepersistence)_ |  |
 | `serving` _[ServingConfig](#servingconfig)_ | Serving configures the Feast feature_server section written into feature_store.yaml for the online serve pod.
 Controls metrics granularity, offline push batching, and MCP. |
+| `disabled` _boolean_ | Disabled skips deploying the online store service entirely, including its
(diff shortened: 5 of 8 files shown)


nikolauspschuetz:
The red \`unit-test-python (3.11, macos-14)\` leg is an unrelated CI flake, not caused by this change (which is Go-only, scoped to \`infra/feast-operator/\`):

\`\`\`
ERROR sdk/python/tests/unit/test_feature_server.py::test_push_and_get
  AssertionError: stdout: b''
  stderr: b"Command timed out after 120s: [\x27apply\x27]"
2259 passed, 47 skipped, 1 error
\`\`\`

A \`feast apply\` subprocess timed out after 120s on the macOS-14 runner. The macOS-3.12 leg and all Ubuntu legs (3.10/3.11/3.12) passed on this same commit, and \`operator-test\` is green. I do not have permission to re-run the job from the fork — a maintainer re-run of that single leg should clear it.

ntkathole:
@nikolauspschuetz instead I think fix should be to allow online server deployment to opt-out.

nikolauspschuetz:
Thanks @ntkathole — that makes sense for backward compatibility. Omitting `onlineStore` today implicitly deploys one, so flipping that default (what this PR currently does) is a breaking change on upgrade for anyone relying on the implicit default.

To make it opt-out instead, I'd keep the current default-on behavior and add an explicit disable switch on the online store spec, mirroring the existing `DisableInitContainers` idiom already in the CRD:

```go
type OnlineStore struct {
    // Disabled skips deploying the online store service.
    // +optional
    Disabled bool `json:"disabled,omitempty"`

    Server      *ServerConfigs          `json:"server,omitempty"`
    Persistence *OnlineStorePersistence `json:"persistence,omitempty"`
    // ...
}
```

`ApplyDefaultsToStatus` would skip the onlineStore defaulting/deployment when `disabled` is true; omitting the block keeps deploying it exactly as before. The standalone-registry / federated pattern from #6586 would then be expressed explicitly as `spec.services.onlineStore.disabled: true`.

One clarification so I build the right thing: do you want the opt-out at the whole online-store level (skip both persistence and the serving pod), or only the online **server** (serving pod), keeping the online store config? Happy to rework the PR either way once you confirm.

ntkathole:
> opt-out at the whole online-store level (skip both persistence and the serving pod), or only the online **server** (serving pod), keeping the online store config?

opt-out at the whole onlineStore level (disabled: true), not server-only.

nikolauspschuetz:
Thanks — reworked it to opt-out. The default is back to deploying the online store, so nothing changes on upgrade for existing setups. Setting `spec.services.onlineStore.disabled: true` now skips it entirely (serving pod and persistence), which covers the registry-only / offline-only case from #6586. It's gated through `isOnlineStore()`, so metrics and the service monitor drop off with it too.

ntkathole:
@nikolauspschuetz Please resolve conflicts

nikolauspschuetz:
Rebased onto `master` and resolved the conflict, @ntkathole. The only conflict was in `.secrets.baseline` (a shifted `line_number` and the `generated_at` timestamp — my `disabled` field pushed the tracked line in `featurestore_types.go` from 937 to 941); the operator code merged cleanly.

Ran the operator suite locally against the envtest harness (k8s 1.31.0) after the rebase — `internal/controller/services` passes, including the `isOnlineStore` / `disabled: true` cases in `util_test.go`. Ready for another look.

ntkathole:
@nikolauspschuetz need to fix linting

ntkathole (approved):

PR #6603, opened 2026-07-13
fix: Preserve all join keys in remote online store read requests

### LINKED ISSUE #6488: get_online_features on remote stores drops all join keys but the first
## Expected Behavior

Feast would use all the entity_keys passed by the user to fetch the proper rows.

## Current Behavior

Feast only uses the first entity_key.

```python
        entity_values = []
        entity_key = ""
        for row in entity_keys:
            entity_key = row.join_keys[0]
            entity_values.append(
                getattr(row.entity_values[0], row.entity_values[0].WhichOneof("val"))  # type: ignore[arg-type]
            )
```
The error is evident by the use of 0 index on the row.join_keys.

## Steps to reproduce

Fetch a multi-key online feature from a remote online store

### Specifications

- Version: 0.63.0
- Platform: FeatureStore CR / K8s

## Possible Solution

Something like this:
```python
    entities = {}
    for row in entity_keys:
        if len(row.join_keys) != len(row.entity_values):
            raise ValueError(
                "EntityKeyProto has mismatched join_keys and entity_values lengths."
            )

        for join_key, entity_value in zip(row.join_keys, row.entity_values):
            value_attr = entity_value.WhichOneof("val")
            if value_attr is None:
                entities[join_key].append(None)
            else:
                entities[join_key].append(getattr(entity_value, value_attr))

    return {
        "features": api_requested_features,
        "entities": dict(entities),
    }
```

AmosBunde:

`RemoteOnlineStore._construct_online_read_api_json_request` built the request body from `row.join_keys[0]` and `row.entity_values[0]` only. For a feature view with a composite entity key, every join key after the first was dropped before the request left the client. The feature server then joined on a partial key, matched a row, and returned it with status `PRESENT`. No exception was raised anywhere in the path, so callers silently received features belonging to the wrong entity.

Reproduced on 0.64.0 with a two-entity feature view keyed on `(driver_id, customer_id)`. The request that reached the server was:

```
{'features': ['driver_customer_stats:conv_rate'],
 'entities': {'driver_id': [1001, 1002]}}
```

`customer_id` is not null or truncated. It is absent.

Two notes for reviewers:

1. The issue title attributes this to `get_online_features`, but that method already builds a correct columnar request across all keys. The defect is one layer down, in the `online_read` path taken when a local `FeatureStore` is configured with an online store of `type: remote`.

2. `get_online_features` is covered by `test_columnar_entity_rows`. The `online_read` request builder had no test coverage at all, which is why the divergence between the two went unnoticed.

This PR builds the request from every `(join_key, entity_value)` pair, keyed on join key name rather than position, and raises on a length mismatch between `join_keys` and `entity_values` rather than silently truncating.

Fixes #6488

# Checks

## Testing Strategy
- [x] Unit tests
- [ ] Integration tests

### DIFF
--- a/sdk/python/feast/infra/online_stores/remote.py
+++ b/sdk/python/feast/infra/online_stores/remote.py
@@ -598,17 +598,21 @@ class RemoteOnlineStore(OnlineStore):
             for requested_feature in requested_features:
                 api_requested_features.append(f"{table.name}:{requested_feature}")
 
-        entity_values = []
-        entity_key = ""
+        entity_columns: Dict[str, List[Any]] = {}
         for row in entity_keys:
-            entity_key = row.join_keys[0]
-            entity_values.append(
-                getattr(row.entity_values[0], row.entity_values[0].WhichOneof("val"))  # type: ignore[arg-type]
-            )
+            if len(row.join_keys) != len(row.entity_values):
+                raise ValueError(
+                    f"Entity key has {len(row.join_keys)} join keys but "
+                    f"{len(row.entity_values)} values; these must be the same length."
+                )
+            for join_key, value in zip(row.join_keys, row.entity_values):
+                entity_columns.setdefault(join_key, []).append(
+                    getattr(value, value.WhichOneof("val"))  # type: ignore[arg-type]
+                )
 
         return {
             "features": api_requested_features,
-            "entities": {entity_key: entity_values},
+            "entities": entity_columns,
         }
 
     def _construct_online_documents_api_json_request(


sdk/python/feast/infra/online_stores/remote.py
```diff
@@ -598,17 +598,21 @@ def _construct_online_read_api_json_request(
...
-                getattr(row.entity_values[0], row.entity_values[0].WhichOneof("val"))  # type: ignore[arg-type]
-            )
+            if len(row.join_keys) != len(row.entity_values):
+                raise ValueError(
+                    f"Entity key has {len(row.join_keys)} join keys but "
+                    f"{len(row.entity_values)} values; these must be the same length."
+                )
+            for join_key, value in zip(row.join_keys, row.entity_values):
+                entity_columns.setdefault(join_key, []).append(
+                    getattr(value, value.WhichOneof("val"))  # type: ignore[arg-type]
```
Sanjays2402, line 610: `WhichOneof("val")` returns `None` for an unset `ValueProto`, so `getattr(value, None)` raises `TypeError` here. The issue specifically calls out null entity values; this should append `None` instead.

PR #6604, opened 2026-07-14
fix: Correct Flink PyArrow dependency constraints

### LINKED ISSUE #6583: `extra !=` does not work as expected
#6476 has recently added (https://github.com/feast-dev/feast/pull/6476/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L26) the following lines to pyproject.toml:

https://github.com/feast-dev/feast/blob/f296d4ba14c5d512429219b2b7845673e0fe524d/pyproject.toml#L26-L27

The `extra != 'flink'"` of this does not work as expected: Neither pip or uv support `extra !=` markers, with uv first evaluating `feast` on its own without the base package and pip always evaluating it to true. As we're in the process of formally removing `extra !=` due to being broken by design (https://discuss.python.org/t/ban-negative-extras-for-extras-marker/108022), we found this one usage and wanted give you a heads up about this feature being broken.

When you try to `[uv] pip install feast[flink]`, it currently downgrades to installing 0.63.0, to avoid the unsatisfiable 0.64.0.

CC @XuananLe

zelinewang:
The Flink extra currently combines a base `pyarrow>=21.0.0` requirement with a
`pyarrow<21.0.0` extra requirement. The attempted `extra != 'flink'` exclusion
does not prevent pip or uv from evaluating both requirements, so
`feast[flink]` is unsatisfiable.

This change puts the shared `pyarrow>=16.1.0` lower bound in the base
dependencies and adds only `pyarrow<21.0.0` to the Flink extra. A normal Feast
install can still select the latest PyArrow release, while a Flink install gets
the compatible `>=16.1.0,<21.0.0` intersection.

The dependency regression check now lives in a focused test module that does
not import the Flink runtime.

Fixes #6583.

# Checks

- [x] My commits are signed off (`git commit -s`).
- [x] My PR title follows conventional commits format.

## Testing Strategy

- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

Validated with:

- the focused dependency test on Python 3.10;
- dry-run uv resolution for both `.[flink]` and the default project;
- a wheel build and inspection of its `Requires-Dist` metadata;
- Ruff check and format verification.

## Release note

```release-note
fix: Make the Flink extra installable by applying its PyArrow upper bound only when the extra is selected.
```

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -2424,8 +2424,8 @@ packages:
   requires_python: '>=3.10'
 - pypi: ./
   name: feast
-  version: 0.64.1.dev40+g82a6918c1.d20260717
-  sha256: 5ab6cffee504478fb0d04132e90e2ea5afa7776ff5e0cd695853e25e71239657
+  version: 0.64.1.dev43+g62f787425
+  sha256: c70a5aa8ddb601588c2aebe6ee791f55b2797d6e2d77a822de53ca3c71874f03
   requires_dist:
   - attrs
   - click>=7.0.0,<9.0.0
@@ -2436,8 +2436,7 @@ packages:
   - mmh3
   - numpy>=2.0.0,<3
   - pandas>=1.4.3,<3
-  - pyarrow>=21.0.0 ; extra != 'flink'
-  - pyarrow>=16.1.0,<21.0.0 ; extra == 'flink'
+  - pyarrow>=16.1.0
   - pydantic>=2.10.6
   - pygments>=2.12.0,<3
   - pyyaml>=5.4.0,<7
@@ -2477,6 +2476,7 @@ packages:
   - elasticsearch>=8.13.0 ; extra == 'elasticsearch'
   - faiss-cpu>=1.7.0,<=1.10.0 ; extra == 'faiss'
   - apache-flink>=2.2.1,<3 ; extra == 'flink'
+  - pyarrow<21.0.0 ; extra == 'flink'
   - google-api-core>=1.23.0,<3 ; extra == 'gcp'
   - googleapis-common-protos>=1.52.0,<2 ; extra == 'gcp'
   - google-cloud-bigquery[pandas]>=2,<4 ; extra == 'gcp'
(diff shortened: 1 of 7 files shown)


shuchu (approved):
lgtm

shuchu:
the CI pipeline fails, @zelinewang  can you try to update the "pixi.lock" file by running the following command?
```
pixi lock
```

ntkathole:
Also run `make lock-python-dependencies-all` to update requirements

zelinewang:
done, ran pixi lock plus the requirements recompile for all three python versions. heads up the pybuild-deps step fails because faiss-cpu==1.14.3 has no sdist on pypi and that pin is already in master's minimal-sdist files, so I left the three -build.txt files untouched. also the fresh resolve moves pyarrow to 21.0.0 on py3.10 (24 on 3.11/3.12), both within the corrected floor

shuchu:
here is the error info: 
```
pixi install -e duckdb-tests --locked
  /home/runner/.pixi/bin/pixi install -e duckdb-tests --locked --manifest-path pyproject.toml --color always
  Error:   × Lock-file version 7 is newer than supported
    help: Maximum supported version: 6 (pixi v0.63.1)
          Cannot continue with --locked or --frozen mode as the lock-file cannot be read.
          Try running `pixi self-update` to update to the latest version.
```

pixi.lock
```diff
@@ -1,17 +1,31 @@
-version: 6
+version: 7
```
shuchu: version 7 is not supported at this moment.

zelinewang:
oops, my local pixi was 0.73 which writes the v7 format. redid it with an older one, back to v6 now and the diff is way smaller. also fixed the missing DCO sign-off

ntkathole:
@zelinewang needs rebase on master, there are conflicts

zelinewang:
rebased on master and redid the lockfiles there, conflicts are gone. the diff also picks up pyiceberg since the iceberg change hadn't updated requirements yet

ntkathole (approved):

PR #6608, opened 2026-07-16
fix: Bump decommissioned Snowflake Python UDF runtime from 3.9 to 3.10 (#6606)

### LINKED ISSUE #6606: Snowflake offline store provider uses decommissioned Python 3.9 runtime, blocking feast apply
We are using the Snowflake offline store/provider which is currently failing on a `feast apply` because it hardcodes RUNTIME_VERSION = '3.9' (see [here](https://github.com/search?q=repo%3Afeast-dev%2Ffeast+RUNTIME_VERSION&type=code)). As of April 30 2026, Snowflake has officially decommissioned the Python 3.9 runtime environment. Snowflake's query engine now strictly rejects any SQL statements attempting to register or execute objects using this runtime, resulting in a hard block that cannot be bypassed via Snowflake account configuration:
```sh
snowflake.connector.errors.ProgrammingError: 391569 (22023): 01c5be5e-040b-1f4a-0003-890423b60d6b: SQL compilation error: Python runtime version 3.9 is decommissioned. Please update your code to Python runtime version 3.10 or later.
```

## Expected Behavior 

`feast apply` to complete successfully.

## Current Behavior

Snowflake returns a compilation error stating that the requested Python runtime version is decommissioned or unsupported, blocking any `feast apply`.
```sh
snowflake.connector.errors.ProgrammingError: 391569 (22023): 01c5be5e-040b-1f4a-0003-890423b60d6b: SQL compilation error: Python runtime version 3.9 is decommissioned. Please update your code to Python runtime version 3.10 or later.
```

## Steps to reproduce

Use Snowflake as offline store.

Example feature view/data source:
```python
driver_stats_source = SnowflakeSource(
    database="DRIVER_DEV",
    schema="FEAST_TEST_RAW",
    table="DRIVER_EXAMPLE",
    name="driver_stats",
    timestamp_field="TIMESTAMP_UTC",
)
    name="driver_hourly_stats",
    entities=[driver],
    ttl=timedelta(days=1),
    schema=[
        Field(name="CONV_RATE", dtype=Float32),
        Field(name="ACC_RATE", dtype=Float32),
        Field(name="AVG_DAILY_TRIPS", dtype=Int64, description="Average daily trips"),
    ],
    online=True,
    source=driver_stats_source,
    tags={"team": "driver_performance"},
)
```
Run `feast apply`.

### Specifications

- Version: 0.64.0

## Possible Solution

Feast to use a supported Python runtime version (such as 3.12+) when creating server-side objects in Snowflake. Snowflake's Python runtime support roadmap can be seen [here](https://docs.snowflake.com/en/developer-guide/python-runtime-support-policy). Ideally either default to a newer version (3.12 seems to be supported up to 02 Oct 2028) or be exposed as a configurable property in `feature_store.yaml`?

zerafachris:
Fixes #6606

Snowflake decommissioned the Python 3.9 UDF runtime environment as of
April 30, 2026, so any `feast apply` against a Snowflake offline store
that deploys Feast's materialization UDFs now fails with:
`Python runtime version 3.9 is decommissioned. Please update your code
to Python runtime version 3.10 or later.`

## What changed
`RUNTIME_VERSION = '3.9'` was hardcoded 16 times in
`snowflake_python_udfs_creation.sql` (the SQL template actually executed
against Snowflake by `SnowflakeComputeEngine.deploy_infra`) and mirrored
in the reference docstrings of `snowflake_udfs.py`. Both are bumped to
`'3.10'`, matching Feast's own minimum supported Python version
(`requires-python = ">=3.10.0"` in pyproject.toml) and a runtime
Snowflake still supports.

## Test evidence
Added `sdk/python/tests/unit/infra/utils/snowflake/test_snowflake_udfs.py`,
which statically asserts neither file references the decommissioned
runtime version.

Before fix:
```
test_udf_creation_sql_does_not_use_decommissioned_runtime_version FAILED
test_udf_module_docstrings_do_not_reference_decommissioned_runtime_version FAILED
(AssertionError: ... still requests the decommissioned Snowflake Python runtime '3.9')
```

After fix:
```
test_udf_creation_sql_does_not_use_decommissioned_runtime_version PASSED
test_udf_module_docstrings_do_not_reference_decommissioned_runtime_version PASSED
```

Full snowflake unit suite: 28 passed. Broader `sdk/python/tests/unit/infra`
suite: 397 passed (1 pre-existing, unrelated environment failure due to
missing pip in the test venv). `ruff check`/`format` and `mypy` clean on
touched files.

Prepared with AI assistance (Claude, agent: claude-sonnet-5), reviewed for correctness before submission.


ntkathole:
@zerafachris is it possible to make it user configurable ?

zerafachris:
@ntkathole Good call — done. Pushed a follow-up commit (c5ae39b) that makes the Python UDF runtime version user-configurable instead of hardcoded.

Added a new field `python_udf_runtime_version` to `SnowflakeComputeEngineConfig` (the `batch_engine` config, alongside `account`, `warehouse`, `database`, etc.):

```yaml
batch_engine:
  type: snowflake.engine
  ...
  python_udf_runtime_version: "3.10"  # optional, defaults to "3.10"
```

- Defaults to `"3.10"` (matching Feast's own `requires-python = ">=3.10.0"` floor), so no config change is needed for existing users picking up this PR.
- Validated as a Python version string (e.g. `"3.10"` or `"3.11.2"`) — an invalid value raises a clear `ValidationError` instead of failing later as an opaque Snowflake SQL error.
- Plumbed through to the executed SQL the same way `PROJECT_NAME`/`STAGE_HOLDER` already are: `snowflake_python_udfs_creation.sql` now uses a `RUNTIME_VERSION_HOLDER` placeholder, which `SnowflakeComputeEngine.update()` fills in from `python_udf_runtime_version` at deploy time.

So the next time Snowflake decommissions a runtime, it's a one-line config change (e.g. `python_udf_runtime_version: "3.11"`) rather than another hardcoded-value PR. Also documented in `docs/reference/compute-engine/snowflake.md`, and added tests covering the default, a custom override, invalid-value validation, and the rendered SQL template.

ntkathole (approved):

sdk/python/feast/infra/utils/snowflake/snowpark/snowflake_python_udfs_creation.sql
```diff
@@ -1,127 +1,127 @@
 CREATE FUNCTION IF NOT EXISTS feast_PROJECT_NAME_snowflake_binary_to_bytes_proto(df BINARY)
   RETURNS BINARY
   LANGUAGE PYTHON
-  RUNTIME_VERSION = '3.9'
+  RUNTIME_VERSION = 'RUNTIME_VERSION_HOLDER'
   PACKAGES = ('protobuf', 'pandas')
   HANDLER = 'feast.infra.utils.snowflake.snowpark.snowflake_udfs.feast_snowflake_binary_to_bytes_proto'
   IMPORTS = ('@STAGE_HOLDER/feast.zip');
 
 CREATE FUNCTION IF NOT EXISTS feast_PROJECT_NAME_snowflake_varchar_to_string_proto(df VARCHAR)
```
ntkathole: I also noticed that there is migration gap, existing deployed UDFs won't be updated as there is check https://github.com/zerafachris/feast/blob/a327346b876c4fdaeab05c0e5f14a4f99817f72c/sdk/python/feast/infra/compute_engines/snowflake/snowflake_engine.py#L161

If a user already has UDFs deployed with RUNTIME_VERSION = '3.9', running feast apply will detect the existing functions and skip re-deployment entirely. The user gets the "already detected" message and the old broken 3.9 UDFs remain in place.

Even if this early return were bypassed, the SQL template uses CREATE FUNCTION IF NOT EXISTS, which means Snowflake wouldn't overwrite existing functions.

we would need to either:
- Remove the early-return guard and change the SQL from CREATE FUNCTION IF NOT EXISTS to CREATE OR REPLACE FUNCTION in snowflake_python_udfs_creation.sql
- Or document that users must run feast teardown then feast apply

zerafachris: Good catch, thank you — fixed in 21cb098.

Went with your first option: removed the `SHOW USER FUNCTIONS` early-return guard in `SnowflakeComputeEngine.update()`, and switched every statement in `snowflake_python_udfs_creation.sql` from `CREATE FUNCTION IF NOT EXISTS` to `CREATE OR REPLACE FUNCTION`. So `update()` now always (re)deploys the materialization UDFs on every `feast apply`, using whichever `python_udf_runtime_version` is currently configured — no more "already detected" skip leaving stale UDFs (e.g. still on `3.9`) in place.

Added unit tests (`tests/unit/infra/compute_engines/snowflake/test_snowflake_compute_engine.py`) asserting `update()` no longer issues a `SHOW USER FUNCTIONS` check and that repeated calls always redeploy with the currently configured runtime version. Ran the full snowflake + compute_engines unit suite (138 passed), ruff, and mypy on the touched files — all clean.

PR #6617, opened 2026-07-19
feat: Add opt-in filter_by_created_timestamp cutoff to get_historical_features

### LINKED ISSUE #6615: feat: Opt-in `created_timestamp <= entity_timestamp` cutoff in point-in-time joins (as-of-known-time retrieval)

Picking this back up from [a Slack thread a while ago](https://feastopensource.slack.com/archives/C01MSKCMB37/p1744885223539189), where the dedup-only handling of `created_timestamp_column` was confirmed.

`get_historical_features` enforces point-in-time correctness on the event timestamp only. When a feature view has a `created_timestamp_column`, it is used purely as a dedup tiebreaker (`MAX(created_timestamp)` per event timestamp), never as a filter. There is no `created_timestamp <= entity_timestamp` predicate anywhere in the join templates (verified on current master, BigQuery template as reference; the other offline stores follow the same pattern).

As a result, a training row at entity timestamp `T` can be served a value that did not exist yet at `T`: its `event_timestamp <= T` but its `created_timestamp > T` (backfills, late corrections, delayed batch writes). This causes:

- Feature leakage: training rows see values that were written after the event.
- Non-reproducible training sets: re-running after a backfill silently changes past feature vectors.
- Offline/online skew: the online model at `T` served the older value, offline retrieval returns the newer one.

Concrete example: two rows for the same entity and event timestamp, value `4000` created before the entity timestamp and `6000` created after it. `get_historical_features` returns `6000`.

An opt-in, per-query flag, default off:

```python
store.get_historical_features(
    entity_df=...,
    features=...,
    filter_by_created_timestamp=True,  # only use feature versions known as of each entity row's timestamp
)
```

When enabled and the feature view has a `created_timestamp_column`, add a single predicate to the point-in-time join:

```sql
AND subquery.created_timestamp <= entity_dataframe.entity_timestamp
```

Default `False` produces identical SQL and results to today. The flag can be threaded through the same optional-kwargs mechanism as the recently added `start_date`/`end_date`, so offline stores that have not implemented it are unaffected unless a user explicitly opts in.

- A per-`FeatureView` or config-level setting. Per-query is preferred since the same view may want both leak-free training and "latest known" analysis.
- #2980 proposed a global `max_created_time` fixed at query time (closed by the stale bot). This proposal is the per-row variant: each entity row is cut off at its own timestamp, which is what you need to emulate what the online store would have served at each event.
- We currently run a patched BigQuery join template with this predicate in production. It works but has to be re-synced on every Feast upgrade.

**Additional context**

- Rows with a NULL `created_timestamp` would be excluded by the predicate; suggest documenting the column must be non-null when the mode is enabled.
- Orthogonal to `ttl`: TTL bounds by event time, this bounds by known time.
- I have a PR ready implementing this across the offline stores (SQL templates, ibis-based stores, dask) with tests. Happy to iterate on naming and API surface.

addenergyx:

Adds an opt-in `filter_by_created_timestamp: bool = False` parameter to `get_historical_features` that enforces `created_timestamp <= entity_timestamp` in point-in-time joins.

Today `created_timestamp_column` is used only as a dedup tiebreaker, never as a filter, so a training row at entity timestamp `T` can be served a value that did not exist yet at `T` (backfills, late corrections). With `filter_by_created_timestamp=True`, retrieval reflects what was actually known as of each entity row's timestamp, i.e. what the online store would have served at that moment.

```python
store.get_historical_features(
    entity_df=entity_df,
    features=features,
    filter_by_created_timestamp=True,  # default False: existing behavior
)
```

## Design

- Threaded through `FeatureStore.get_historical_features` using the same optional-kwargs pattern as `start_date`/`end_date`: only passed to the offline store when `True`, so the default path is untouched and third-party stores are unaffected unless a user explicitly opts in.
- For SQL-template stores, the shared `build_point_in_time_query` gains a `filter_by_created_timestamp` template variable, and each template's `__base` join gains one guarded predicate:
  ```jinja
  {% if filter_by_created_timestamp and featureview.created_timestamp_column %}
  AND subquery.created_timestamp <= entity_dataframe.entity_timestamp
  {% endif %}
  ```
  Everything else (event-timestamp window, TTL, dedup, `ROW_NUMBER`) is unchanged.
- Feature views without a `created_timestamp_column` are unaffected.

## Store coverage

Implemented: BigQuery, Redshift, Spark, Trino, Athena, Postgres, ClickHouse, Couchbase (SQL templates); DuckDB, MSSQL, Oracle (via the shared ibis `point_in_time_join`); Dask/file (pandas filter mirroring `_filter_ttl`); Hybrid (delegates, re-checking the resolved child store).

Stores declare support via a `supports_filter_by_created_timestamp` class attribute on `OfflineStore` (default `False`). A single guard in the passthrough provider raises `NotImplementedError` when the flag is set for a store that does not declare support, so no store can silently ignore it and new stores are covered automatically. Unsupported today: Snowflake (its ASOF JOIN dedups before the join and `MATCH_CONDITION` takes a single predicate, so the cutoff cannot be expressed without restructuring the query), Remote (needs protocol support to forward the flag), Ray, and MongoDB.

## Notes

- Rows with a NULL `created_timestamp` are excluded by the predicate in all implementations; the docstring notes the column should be non-null when the mode is used. (In the dask store, unmatched entity rows from the left join are still preserved.)
- In the dask store, an entity row whose matching feature rows are all filtered out disappears from the result instead of surviving with null features. This matches the store's existing TTL filtering behavior and is a pre-existing trait of its incremental join, not something this PR changes.
- Orthogonal to `ttl`: TTL bounds by event time, this bounds by known time.

Fixes #6615

# Misc

Tests added in `sdk/python/tests/unit/infra/offline_stores/test_filter_by_created_timestamp.py`:
- Template rendering: predicate present iff `filter_by_created_timestamp=True` and a `created_timestamp_column` is set; default rendering unchanged.
- Dask and ibis (DuckDB) end-to-end: same event timestamp with two versions created before/after the entity timestamp; default serves the later-created value, `filter_by_created_timestamp=True` serves the version known at the entity timestamp.

`ruff check`/`ruff format` clean; `mypy feast` reports the same 17 pre-existing errors as `master` (none introduced).


franciscojavierarceo (changes requested):
The Dask implementation can drop entity rows. It performs the left merge first and then `_filter_created_timestamp()` removes rows whose matched feature has `created_timestamp > entity_timestamp`. If every candidate for an entity is too new, the merge produced no unmatched/null row to fall back to, so filtering removes that entity entirely. `get_historical_features()` should preserve entity-dataframe cardinality and return null features in this case. Please apply the cutoff in/before the join or reconstruct unmatched rows, and add a regression where all matching candidates are future-created.

sdk/python/feast/infra/offline_stores/ibis.py
```diff
@@ -434,6 +437,14 @@ def point_in_time_join(
             feature_table[timestamp_field] <= entity_table[event_timestamp_col],
         )
 
+        if filter_by_created_timestamp and created_timestamp_field:
```
ntkathole, line 445: This predicate casts the created timestamp to UTC but the existing event-timestamp comparison on the same join doesn't cast. Compare with the existing predicate at line 433-434 which does no cast.

addenergyx:
Thanks @franciscojavierarceo should be fixed now, it now nulls out the feature columns instead of dropping the row, so the entity row survives with null features

sdk/python/feast/infra/offline_stores/ibis.py
addenergyx: Thanks @ntkathole, made a change so read_fv now normalizes the created timestamp on read, as it already did for the event timestamp, so the predicate no longer casts. Made it unconditional since deduplicate() reads that column regardless of the flag.

ntkathole (approved):
Looks good

PR #6140, opened 2026-03-23
fix: ODFV output projection in offline retrieval (#6099)

jyejare:
## Summary

Fixes #6099 - Ensures offline retrieval honors ODFV feature projection, matching online retrieval behavior.

## Problem

When requesting a subset of features from an OnDemandFeatureView:
- **Online retrieval** ✅ Returns only requested features
- **Offline retrieval** ❌ Returns ALL ODFV output features (before this fix)

This caused schema mismatches between training and serving pipelines.

## Solution

Modified `RetrievalJob.to_arrow()` in `offline_store.py` to:
1. Parse requested features from `metadata.features`
2. Build a mapping of ODFV name → requested feature names
3. Filter ODFV transformation output to only include requested columns

## Example

**Before this fix:**
```python
features = ["my_odfv:feature_a"]
offline_result = store.get_historical_features(features=features, ...)
# Columns: driver_id, event_timestamp, feature_a, feature_b, feature_c ❌
```

**After this fix:**
```python
features = ["my_odfv:feature_a"]
offline_result = store.get_historical_features(features=features, ...)
# Columns: driver_id, event_timestamp, feature_a ✅
```

### Modified: `sdk/python/feast/infra/offline_stores/offline_store.py`
- Updated `RetrievalJob.to_arrow()` method (lines 140-184)
- Added filtering logic for ODFV output projection
- Maintains backward compatibility

### Added: Test in `sdk/python/tests/integration/offline_store/test_universal_historical_retrieval.py`
- `test_odfv_projection()` - Comprehensive test verifying:
  - Single feature request returns only that feature
  - Multiple feature request returns only requested features
  - Unrequested features are NOT included
  - Offline and online retrieval have consistent behavior
- Parametrized for both `full_feature_names=True` and `False`

## Testing

The new test `test_odfv_projection` verifies:
1. ✅ Requesting 1 out of 3 ODFV features → returns only that 1 feature
2. ✅ Requesting 2 out of 3 ODFV features → returns only those 2 features
3. ✅ Unrequested features are NOT included in the result
4. ✅ Offline and online retrieval return consistent schemas

## Backward Compatibility

- ✅ Falls back to old behavior if metadata is unavailable
- ✅ No breaking changes to existing functionality
- ✅ Only affects ODFV feature projection

## Impact

This fix ensures:
- ✅ Consistent behavior between online and offline retrieval
- ✅ No schema mismatches in ML pipelines
- ✅ More efficient - doesn't compute/return unnecessary features
[...]

Issue #6099, opened 2026-03-12
get_historical_features returns all ODFV output columns even when a single ODFV feature is requested


andrevus:
## Expected Behavior 

When requesting a single feature from an `OnDemandFeatureView`, both online and offline retrieval should return only the explicitly requested output feature.

Example:

```python
features=["odfv:feature_a"]
```

Expected result:

online returns only feature_a

offline returns only feature_a

The requested feature list should define the output schema consistently across both retrieval paths.

## Current Behavior

For OnDemandFeatureViews, offline retrieval via get_historical_features() returns all output columns defined by the ODFV, even when only a single ODFV feature is requested.

Example:
```python
features=["odfv:feature_a"]
```

Actual result:

online returns only feature_a

offline returns feature_a, feature_b, feature_c, etc. (all ODFV outputs)

This creates a schema inconsistency between online and offline retrieval.

Important clarification:
this is not about extra input columns from RequestSource or source FeatureViews being present in offline retrieval.
This issue is specifically about offline retrieval not honoring projection of ODFV output features.

## Steps to reproduce

1. Define an OnDemandFeatureView with multiple output features, for example:

- feature_a
- feature_b
- feature_c

2. Request a single feature from that ODFV in offline retrieval:

```python
store.get_historical_features(
    entity_df=entity_df,
    features=["odfv:feature_a"],
)
```
3. Observe that the offline result contains all output columns from the ODFV instead of only feature_a.
4. Compare with online retrieval for the same requested feature set and observe that online returns only the requested feature.

### Specifications

- Version: latest
- Platform: na
- Subsystem: Offline retrieval / OnDemandFeatureView / get_historical_features

## Possible Solution

get_historical_features() should honor the requested output feature subset for OnDemandFeatureViews in the same way as online retrieval.

If this is intentional behavior, the documentation should explicitly state that offline retrieval for ODFVs returns the full ODFV output schema even when only a subset of ODFV features is requested.

However, from a user perspective, consistent projection behavior between online and offline retrieval would be much more intuitive and would prevent schema mismatches in downstream training and serving pipelines.

### Related

This issue appears related to, but different from, a few existing Feast topics:

#4479 — OnDemandFeatureViews with RequestSource returns different columns depending on online/offline feature retrieval

That issue discusses extra input columns (for example from RequestSource) appearing in offline retrieval.
My issue is different: even when I request a single ODFV output feature, offline retrieval returns all ODFV output features, while online retrieval returns only the requested one.

[...]

PR #6622, opened 2026-07-22
fix: Merge shared ODFV source projections in feature resolution

### LINKED ISSUE #6621: when calling for several ODFVs at once, only fetches one ODFV's column-list per shared source view (last request wins),
## Expected Behavior 

When several on demand feature views share a source feature view but use different input columns, feature resolution should produce one entry for the shared source view whose projection contains the union of the required input columns.

## Current Behavior

 _get_feature_views_to_use appends the shared source view once per ODFV, each copy carrying only that ODFV's projection. The membership check if source_fv not in fvs_to_use never   matches because BaseFeatureView.__eq__ compares projections, and the freshly fetched view (full projection) never equals a stored narrowed copy. Downstream, _group_feature_refs keys   its view index by projection.name_to_use(), so the last duplicate wins and the other ODFVs' source projections are silently dropped.

This corrupts get_historical_features output: with a request containing only ODFV refs, the losing ODFVs' input columns are never fetched, so transforms receive missing inputs (NaN in pandas mode). We hit this in production on 0.46, where a downstream fillna(0) turned the missing inputs into silently wrong feature values on about 10% of rows.

Since 0.62 the retrieval path is masked by the ref injection loop added in #6140, but the resolution helper still returns duplicate entries with conflicting projections to its other callers (_get_online_request_context, retrieve_online_documents, retrieve_online_documents_v2).

## Steps to reproduce

```
  src_fv = FeatureView(name="src_fv", entities=[driver], schema=[Field(name="a", dtype=Float64), Field(name="b", dtype=Float64)], source=...)

  @on_demand_feature_view(sources=[src_fv[["a"]]], schema=[Field(name="a_out", dtype=Float64)], mode="pandas")
  def odfv_a(inputs):
  @on_demand_feature_view(sources=[src_fv[["b"]]], schema=[Field(name="b_out", dtype=Float64)], mode="pandas")
  def odfv_b(inputs):
      return pd.DataFrame({"b_out": inputs["b"] + 100})

  fvs, _ = _get_feature_views_to_use(store.registry, store.project, ["odfv_a:a_out", "odfv_b:b_out"])
  # Actual: two src_fv entries with projections ['a'] and ['b']
  # Expected: one src_fv entry with projection ['a', 'b']
```

### Specifications

- Version: master (also reproduced on 0.46; wrong retrieval values on 0.61 and earlier)
- Platform: all
- Subsystem: historical retrieval

## Possible Solution

In _get_feature_views_to_use, match existing entries by projection.name_to_use() and merge projections (union of features, deduplicated by name) instead of appending a duplicate. 

Related: #6099, #6140

sanskar-singh-2403:
What this does
Fixes #6621.

When several on demand feature views share a source feature view but select different input columns, _get_feature_views_to_use appended the shared source view once per ODFV, each copy carrying only that ODFV's projection. The membership guard if source_fv not in fvs_to_use never matched, because BaseFeatureView.__eq__ compares projections and the freshly-fetched full-projection view never equals an already-stored narrowed copy.

Downstream, _group_feature_refs keys its view index by projection.name_to_use(), so the duplicate entries collide and the last one wins, the other ODFVs' source projections are silently dropped. During get_historical_features this means the losing ODFV's input columns are never fetched, so transforms receive missing inputs (NaN in pandas mode).

The offline retrieval path is masked on 0.62+ by the ref-injection loop from #6140, but the resolution helper still returns duplicate entries with conflicting projections to its other callers (_get_online_request_context, retrieve_online_documents, retrieve_online_documents_v2).

The fix
In _get_feature_views_to_use, index accumulated source views by projection.name_to_use() and, on a repeat hit, merge the projections (union of features, deduped by name) instead of appending a duplicate. A shared source view now resolves to a single entry carrying the union of the required input columns.

The merge helper reassigns projection.features to a fresh list rather than mutating in place, because the projections in play are shallow copies (copy.copy) that share their features list with the registry object.

Tests
Added TestGetFeatureViewsToUseSharedSource in tests/unit/test_utils.py:

- test_shared_source_projections_are_merged two ODFVs (odfv_a on src_fv[["a"]], odfv_b on src_fv[["b"]]) resolve to a single src_fv entry with projection ["a", "b"].
- test_shared_source_ref_order_independent result is the same regardless of ODFV request order.

Both fail on master (assert 2 == 1 the duplicate entries) and pass with this change. ruff check and ruff format are clean.

Related
#6099, #6140

### DIFF
--- a/sdk/python/feast/utils.py
+++ b/sdk/python/feast/utils.py
@@ -55,6 +55,7 @@ if typing.TYPE_CHECKING:
     from feast.base_feature_view import BaseFeatureView
     from feast.feature_service import FeatureService
     from feast.feature_view import FeatureView
+    from feast.feature_view_projection import FeatureViewProjection
     from feast.infra.registry.base_registry import BaseRegistry
     from feast.on_demand_feature_view import OnDemandFeatureView
 
@@ -1193,6 +1194,19 @@ def _list_feature_views(
     return feature_views
 
 
+def _merge_projection_features(
+    target: "FeatureViewProjection",
+    other: "FeatureViewProjection",
+) -> None:
+    existing_names = {feature.name for feature in target.features}
+    merged = list(target.features)
+    for feature in other.features:
+        if feature.name not in existing_names:
+            existing_names.add(feature.name)
+            merged.append(feature)
+    target.features = merged
+
+
 def _get_feature_views_to_use(
     registry: "BaseRegistry",
     project,
@@ -1223,6 +1237,25 @@ def _get_feature_views_to_use(
         feature_views = parsed  # type: ignore[assignment]
 
     fvs_to_use, od_fvs_to_use = [], []
+    fvs_by_projection_key: Dict[str, "FeatureView"] = {}
+
+    def _append_or_merge_source_fv(fv_with_projection: "FeatureView") -> None:
+        # Index every (source) feature view by its effective projection key so
+        # that a regular FeatureView and an ODFV source projection resolving to
+        # the same key are merged into a single entry instead of duplicated.
+        # Without this, requesting ["src_fv:a", "odfv_b:b_out"] where odfv_b
+        # sources src_fv appends two src_fv entries and later raises
+        # "KeyError: Feature a not found in projection src_fv".
+        projection_key = fv_with_projection.projection.name_to_use()
+        existing = fvs_by_projection_key.get(projection_key)
+        if existing is None:
+            fvs_by_projection_key[projection_key] = fv_with_projection
+            fvs_to_use.append(fv_with_projection)
+        else:
+            _merge_projection_features(
+                existing.projection, fv_with_projection.projection
+            )
+
     for name, version_num, projection in feature_views:
         if version_num is not None:
             if not getattr(registry, "enable_online_versioning", False):
@@ -1286,10 +1319,8 @@ def _get_feature_views_to_use(
                     source_fv.entities = []  # type: ignore[attr-defined]
                     source_fv.entity_columns = []  # type: ignore[attr-defined]
 
-                if source_fv not in fvs_to_use:
-                    fvs_to_use.append(
-                        source_fv.with_projection(copy.copy(source_projection))
-                    )
+                new_source_fv = source_fv.with_projection(copy.copy(source_projection))
+                _append_or_merge_source_fv(new_source_fv)
         else:
             if (
                 hide_dummy_entity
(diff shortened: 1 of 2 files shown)


sanskar-singh-2403:
hey, @oavdeev @ankurs just a friendly reminder... waiting for your review(s) 🙏

sanskar-singh-2403:
> hey, @oavdeev @ankurs just a friendly reminder... waiting for your review(s) 🙏

hey @oavdeev @ankurs @ches just a friendly reminder to review this PR, thanks!

cc: @addenergyx

sdk/python/feast/utils.py
```diff
@@ -1286,9 +1301,17 @@ def _get_feature_views_to_use(
                     source_fv.entities = []  # type: ignore[attr-defined]
                     source_fv.entity_columns = []  # type: ignore[attr-defined]
 
-                if source_fv not in fvs_to_use:
-                    fvs_to_use.append(
-                        source_fv.with_projection(copy.copy(source_projection))
+                projection_key = source_projection.name_to_use()
+                existing_source_fv = fvs_by_projection_key.get(projection_key)
```
addenergyx: fvs_by_projection_key is only populated for ODFV sources. regular FeatureView is appended but not indexed, so ["src_fv:a", "odfv_b:b_out"] creates duplicate src_fv entries and later raises KeyError: Feature a not found in projection src_fv.

Could both paths use the same append-or-merge helper keyed by projection.name_to_use()? I tested that approach in both orders, it produces one merged entry and both requests passed.

sanskar-singh-2403: Good catch @addenergyx , you're right. `fvs_by_projection_key` was only written in the ODFV branch, so a regular FeatureView got appended but never indexed. `["src_fv:a", "odfv_b:b_out"]` then appended a second, partial `src_fv` projection and blew up later with `KeyError: Feature a not found in projection src_fv`.

I pulled the logic into a single append-or-merge helper keyed by `projection.name_to_use()` and routed both paths (regular FV and ODFV source) through it, as you suggested. Added two regression tests covering the mixed regular-FV + ODFV-source case in both orders; confirmed they fail on the old code and pass with the fix. Full `test_utils.py` is green.

addenergyx: Perfect! Thanks for the fix @sanskar-singh-2403

addenergyx (approved):
Thanks! lgtm

sanskar-singh-2403:
@franciscojavierarceo @ntkathole requesting your reviews in this, Thanks!

ntkathole (approved):

PR #6628, opened 2026-07-22
fix: tear down demo FeatureStore in-process (Windows WinError 32)

### LINKED ISSUE #6626: fix: feast teardown fails on Windows (WinError 32) when test_workflow keeps SQLite open
## Description

Running the generated local demo `python test_workflow.py` on Windows often ends with:

```text
PermissionError: [WinError 32] ... online_store.db
  File ".../feast/infra/online_stores/sqlite.py", line ..., in teardown
    os.unlink(db_path)
```

The demo otherwise succeeds (`apply`, historical features, materialize, online fetch, push).

## Environment

- Feast: 0.65.0
- OS: Windows 10
- Python: 3.13.7

## Steps to reproduce

```bash
pip install feast==0.65.0
feast init bug_repro
cd bug_repro/feature_repo
python test_workflow.py
```

## Likely root cause

`test_workflow.py` keeps a live `FeatureStore` (and thus an open SQLite online-store connection) in the parent process, then calls:

```python
subprocess.run(["feast", "teardown"])
```

The CLI process tries to delete `online_store.db` while the parent still holds the file lock. `SqliteOnlineStore.teardown` already retries `PermissionError`, but that cannot unlock a handle owned by another process.

## Suggested fix

In the local (and similar) `test_workflow.py` templates, tear down via the same store instance instead of a subprocess, e.g. `store.teardown()`, and/or explicitly close online-store connections before deleting the DB.

## Notes

Happy to help with a follow-up PR after the `feast init` Windows path fix.

lipereis:
## Summary
- Generated `test_workflow.py` demos called `subprocess.run([\"feast\", \"teardown\"])` while the parent still held an open SQLite connection via `FeatureStore`.
- On Windows this causes `PermissionError: WinError 32` when deleting `online_store.db`.
- Switch templates to `store.teardown()` so the same process closes the connection before unlinking.

Fixes #6626

## Test plan
- [ ] On Windows: `feast init repro && cd repro/feature_repo && python test_workflow.py` completes without WinError 32 at teardown
- [ ] Demo steps before teardown still succeed (apply, historical, materialize, online, push)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/sdk/python/feast/templates/aws/feature_repo/test_workflow.py
+++ b/sdk/python/feast/templates/aws/feature_repo/test_workflow.py
@@ -65,7 +65,7 @@ def run_demo():
     fetch_online_features(store, source="push")
 
     print("\n--- Run feast teardown ---")
-    subprocess.run(["feast", "teardown"])
+    store.teardown()
 
 
 def fetch_historical_features_entity_sql(store: FeatureStore, for_batch_scoring):
--- a/sdk/python/feast/templates/cassandra/feature_repo/test_workflow.py
+++ b/sdk/python/feast/templates/cassandra/feature_repo/test_workflow.py
@@ -54,7 +54,7 @@ def run_demo():
     fetch_online_features(store, source="push")
 
     print("\n--- Run feast teardown ---")
-    subprocess.run(["feast", "teardown"])
+    store.teardown()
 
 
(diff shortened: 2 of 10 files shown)


lipereis:
Closing — contributor prefers not to open PRs against the official Feast repo for now.

Issue #4934, opened 2025-01-19
Add Support for Azure AD appRoles Claims in OIDC Authentication


amitej25:
**Is your feature request related to a problem? Please describe.**
Feast currently supports OIDC-based authentication and expects the resource_access claim in the access token to manage roles and permissions. However, Azure AD issues tokens with appRoles instead of resource_access, causing a mismatch between the token structure and Feast's expectations.

This leads to challenges when integrating Feast with Azure AD for authentication, as Feast cannot interpret the appRoles claim.

**Describe the solution you'd like**
Enhance Feast's OIDC implementation to support multiple token formats by:

Checking for the appRoles claim when resource_access is not found.

**Additional context**
Observed Behavior:
When using Azure AD for OIDC-based authentication, the access token contains the following structure:

```
{
  "appRoles": [
    {
      "allowedMemberTypes": ["User", "Application"],
      "displayName": "feast.reader",
      "value": "feast.reader"
    },
    {
      "allowedMemberTypes": ["User", "Application"],
      "displayName": "feast.writer",
      "value": "feast.writer"
    }
  ]
},
```
Feast, however, expects the token to contain the resource_access claim:

```
{
  "resource_access": {
    "app": {
      "roles": ["reader", "writer"]
    }
  }
}
```
This results in Feast failing to validate roles and permissions from the Azure AD token.

larrysingleton007:
For anyone still watching this: the top-level `roles` claim gap raised here is fixed in #6631 (along with the app-only username gap that also blocked Entra ID), which supersedes this issue.

PR #6631, opened 2026-07-22
fix: Support Entra ID (Azure AD) token claims in OIDC auth

### LINKED ISSUE #6630: OIDC auth rejects Entra ID (Azure AD) tokens: username claims and top-level app roles unsupported
## Expected Behavior

Feature-server OIDC auth (`auth: type: oidc`) authenticates tokens issued by Microsoft Entra ID (Azure AD), including client-credentials (app-only) tokens, and `RoleBasedPolicy` matches Entra app roles.

## Current Behavior

Two independent Keycloak-shape assumptions in `sdk/python/feast/permissions/auth/oidc_token_parser.py` (unchanged 0.64.0 -> 0.65.0):

1. `_extract_username_or_raise_error` requires `preferred_username` or `upn`. Entra client-credentials tokens carry neither, so every machine-to-machine caller fails with `AuthenticationError("Missing preferred_username or upn field in access token.")`.
2. Roles are extracted only from Keycloak's nested `resource_access.<client_id>.roles`. Entra emits app roles in the top-level `roles` claim, so roles are always `[]` and `RoleBasedPolicy` can never match.

#4934 previously requested the roles half and was closed stale without a fix.

## Steps to reproduce

1. Configure the feature server with `auth: type: oidc` against an Entra ID tenant.
2. Acquire a client-credentials token for an app registration granted an app role.
3. Call any authenticated endpoint: the token is rejected at username extraction. With a user (auth-code) token, authentication passes but extracted roles are `[]`.

### Specifications

- Version: 0.64.0 (parser identical in 0.65.0)
- Platform: Python feature server, Microsoft Entra ID (Azure AD)
- Subsystem: `permissions/auth`

## Possible Solution

- Username: when the human claims are absent, fall back to the calling application's identity: `azp` (v2 tokens), then `appid` (v1), then `sub`.
- Roles: merge the top-level `roles` claim with the existing `resource_access` extraction so both issuer shapes authorize.

I'm taking this one: we already run 0.64 behind Entra with exactly this change applied as a build-time patch, and I'll submit it as a PR.

larrysingleton007:

The OIDC token parser is written against Keycloak's token shape, and that breaks Microsoft Entra ID (Azure AD) in two independent places.

Username extraction accepted only `preferred_username` or `upn`. Entra client-credentials (app-only) tokens carry neither, so every machine-to-machine caller was rejected at authentication. This falls back to the calling application's own identity when no human claim is present: `azp` on v2 tokens, `appid` on v1, then `sub`, which every issuer sets. A claim that is present but null or non-string is skipped rather than returned, and a token providing none of the five claims as a string still raises `AuthenticationError`.

Roles were read only from Keycloak's nested `resource_access.<client_id>.roles`. Entra puts app roles in the top-level `roles` claim, so the extracted list was always empty and `RoleBasedPolicy` never matched. This merges the top-level claim into the existing extraction, preserving order and dropping duplicates.

Keycloak's default behavior is unchanged: the username fallbacks only fire when `preferred_username` and `upn` are both absent, the merge only adds roles, and token validation (signature and expiry) is untouched. One behavior change worth calling out for existing deployments: role resolution now also reads the top-level `roles` claim for every issuer, not only Entra. An installation whose IdP already emits a top-level `roles` claim (for example a Keycloak realm with a custom mapper) will see those roles merged into the user's Feast roles. The Keycloak `resource_access.<client_id>.roles` path itself is unchanged.

We run this exact change against an Entra tenant in production as a build-time patch on 0.64.

Fixes #6630. Supersedes #4934, which requested the roles half and was closed stale.

# Checks

## Testing Strategy
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

New unit tests in `sdk/python/tests/unit/permissions/auth/test_token_parser.py` cover: `preferred_username` and `upn` precedence over the application claims (including directly against `appid`/`sub`); `azp`-only, `appid`-only, and `sub`-only tokens each authenticating with that value; tokens whose only identity claims are null or non-string still raising; top-level roles extracted on their own; Keycloak `resource_access` roles still working; the de-duplicated merge when both shapes are present, including duplicates within the top-level claim; and that identity and roles are read from the signature-verified decode rather than the unverified routing decode.

# Misc

Release note: OIDC authentication now accepts Microsoft Entra ID (Azure AD) tokens. The username falls back to `azp`, `appid`, or `sub` for client-credentials (app-only) tokens, and app roles in the top-level `roles` claim are merged with Keycloak's `resource_access.<client_id>.roles`.

The PR template here does not have an auto-generated release-note field, so the note is inline above. I can't apply labels as an outside contributor, so a maintainer will need to add `kind/bug` (and `ok-to-test` for CI).

The last two `test:` commits are self-contained (a fidelity test plus a fixture extraction) and can be dropped without affecting the fix if a smaller diff is preferred.

### DIFF
--- a/docs/getting-started/components/authz_manager.md
+++ b/docs/getting-started/components/authz_manager.md
@@ -47,7 +47,7 @@ Some assumptions are made in the OIDC server configuration:
 * The roles are exposed in the access token under `resource_access.<client_id>.roles` (Keycloak) or in the top-level `roles` claim (Entra ID app roles). Roles found in both are merged.
 * The JWT token is expected to have a verified signature and not be expired. The Feast OIDC token parser logic validates for `verify_signature` and `verify_exp` so make sure that the given OIDC provider is configured to meet these requirements.
 * The username is read from the first of `preferred_username`, `upn`, `azp`, `appid`, `sub` present in the token. Entra ID client-credentials (app-only) tokens carry no user claim, so they authenticate as the calling application.
-* For `GroupBasedPolicy` support, the `groups` claim should be present in the access token (requires a "Group Membership" protocol mapper in Keycloak).
+* For `GroupBasedPolicy` support, the `groups` claim should be present in the access token (requires a "Group Membership" protocol mapper in Keycloak). Entra ID emits group object IDs (GUIDs) in this claim rather than names, and omits it once a principal exceeds Entra's group overage limit, so on Entra a `GroupBasedPolicy` must reference those IDs and cannot rely on the claim for very large group memberships.
 
 (*) Please note that **the role match is case-sensitive**, e.g. the name of the role in the OIDC server and in the `Permission` configuration
 must be exactly the same.
@@ -65,6 +65,16 @@ For example, the access token for a client `app` of a user with `reader` role an
   },
   "groups": [
     "data-team"
+  ]
+}
+```
+
+A Microsoft Entra ID (Azure AD) client-credentials (app-only) token has no user claim; the application authenticates as itself, and its app roles arrive in the top-level `roles` claim:
+```json
+{
+  "azp": "11111111-2222-3333-4444-555555555555",
+  "roles": [
+    "reader"
   ]
 }
 ```
--- a/docs/getting-started/components/authz_manager.md
+++ b/docs/getting-started/components/authz_manager.md
@@ -44,9 +44,9 @@ The server, in turn, uses the same OIDC server to validate the token and extract
 
 Some assumptions are made in the OIDC server configuration:
 * The OIDC token refers to a client with roles matching the RBAC roles of the configured `Permission`s (*)
-* The roles are exposed in the access token under `resource_access.<client_id>.roles`
+* The roles are exposed in the access token under `resource_access.<client_id>.roles` (Keycloak) or in the top-level `roles` claim (Entra ID app roles). Roles found in both are merged.
 * The JWT token is expected to have a verified signature and not be expired. The Feast OIDC token parser logic validates for `verify_signature` and `verify_exp` so make sure that the given OIDC provider is configured to meet these requirements.
-* The `preferred_username` should be part of the JWT token claim.
+* The username is read from the first of `preferred_username`, `upn`, `azp`, `appid`, `sub` present in the token. Entra ID client-credentials (app-only) tokens carry no user claim, so they authenticate as the calling application.
 * For `GroupBasedPolicy` support, the `groups` claim should be present in the access token (requires a "Group Membership" protocol mapper in Keycloak).
 
 (*) Please note that **the role match is case-sensitive**, e.g. the name of the role in the OIDC server and in the `Permission` configuration
--- a/docs/getting-started/components/authz_manager.md
+++ b/docs/getting-started/components/authz_manager.md
@@ -47,7 +47,8 @@ Some assumptions are made in the OIDC server configuration:
 * The roles are exposed in the access token under `resource_access.<client_id>.roles` (Keycloak) or in the top-level `roles` claim (Entra ID app roles). Roles found in both are merged.
 * The JWT token is expected to have a verified signature and not be expired. The Feast OIDC token parser logic validates for `verify_signature` and `verify_exp` so make sure that the given OIDC provider is configured to meet these requirements.
 * The username is read from the first of `preferred_username`, `upn`, `azp`, `appid`, `sub` present in the token. Entra ID client-credentials (app-only) tokens carry no user claim, so they authenticate as the calling application.
-* For `GroupBasedPolicy` support, the `groups` claim should be present in the access token (requires a "Group Membership" protocol mapper in Keycloak). Entra ID emits group object IDs (GUIDs) in this claim rather than names, and omits it once a principal exceeds Entra's group overage limit, so on Entra a `GroupBasedPolicy` must reference those IDs and cannot rely on the claim for very large group memberships.
+* For `GroupBasedPolicy` support, the `groups` claim should be present in the access token (requires a "Group Membership" protocol mapper in Keycloak).
+* **Entra ID limitation**: Group claims use object IDs (GUIDs) instead of names, and are omitted entirely when a user exceeds the group overage threshold. GroupBasedPolicy must reference GUIDs and cannot be used for principals with large group memberships.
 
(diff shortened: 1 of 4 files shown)


larrysingleton007:
I validated this against real Microsoft Entra ID tokens from a production tenant, running them through the parser and a `RoleBasedPolicy` check.

Confirmed:

- v1.0 client-credentials (app-only) tokens carry `appid`, with no `azp` and no `preferred_username`/`upn`, put app roles in a top-level `roles` array, and have no `resource_access`. This is the shape the unpatched parser rejects at username extraction. With this PR the username resolves to `appid` and the top-level roles merge in.
- v2.0 app-only tokens are the same except the caller identity is in `azp` rather than `appid`, which exercises that fallback.
- Delegated user tokens carry `preferred_username` and keep using the existing path unchanged.
- `RoleBasedPolicy` matches on the top-level role string, and a policy requiring a role the token lacks correctly denies.

One interop detail others may hit: it is valid to set `auth_discovery_url` to the v2.0 endpoint while the tenant still issues v1.0 tokens (`iss: https://sts.windows.net/...`). It works because the parser uses discovery only to fetch JWKS and Entra shares signing keys across v1 and v2, and because it runs with `verify_aud=False` and no issuer check. If stricter `iss`/`aud` validation is added later, it would need to allow v1.0 tokens against a v2.0 discovery document.

larrysingleton007:
/kind bug
/cc @ntkathole @franciscojavierarceo

Small, self-contained fix for a filed issue (#6630): the OIDC parser now accepts Microsoft Entra ID (Azure AD) tokens, with a username fallback for app-only tokens and support for the top-level `roles` claim. The diff is the two parser changes plus tests and a docs note, and unit tests pass.

As an outside contributor I can't self-label or start CI. If the `/kind bug` above doesn't take, could a maintainer add the label and run `ok-to-test` when convenient?

ntkathole:
@aniketpalu Can you please take a look ?

larrysingleton007:
Heads up on the red `unit-test-python (3.12, ubuntu-latest)` check here: it is a pre-existing flaky collection error in `test_mcp_server.py`, a `KeyError` importing `feast.infra.mcp_servers.mcp_server` under parallel collection. It is unrelated to the changes in this PR. I addressed it separately at its root in #6634 (issue #6633) to clear the blocker. The tests in this PR pass on the other CI matrices and locally.

sdk/python/tests/unit/permissions/auth/conftest.py
```diff
@@ -62,6 +64,22 @@ def oidc_config() -> OidcAuthConfig:
     )
 
 
+@pytest.fixture
+def discovery_data() -> dict:
```
aniketpalu:
```suggestion
def discovery_data() -> Dict[str, str]:
```

docs/getting-started/components/authz_manager.md
```diff
@@ -44,10 +44,10 @@ The server, in turn, uses the same OIDC server to validate the token and extract
...
 * The JWT token is expected to have a verified signature and not be expired. The Feast OIDC token parser logic validates for `verify_signature` and `verify_exp` so make sure that the given OIDC provider is configured to meet these requirements.
-* The `preferred_username` should be part of the JWT token claim.
-* For `GroupBasedPolicy` support, the `groups` claim should be present in the access token (requires a "Group Membership" protocol mapper in Keycloak).
+* The username is read from the first of `preferred_username`, `upn`, `azp`, `appid`, `sub` present in the token. Entra ID client-credentials (app-only) tokens carry no user claim, so they authenticate as the calling application.
+* For `GroupBasedPolicy` support, the `groups` claim should be present in the access token (requires a "Group Membership" protocol mapper in Keycloak). Entra ID emits group object IDs (GUIDs) in this claim rather than names, and omits it once a principal exceeds Entra's group overage limit, so on Entra a `GroupBasedPolicy` must reference those IDs and cannot rely on the claim for very large group memberships.
```
aniketpalu:
```suggestion
* For `GroupBasedPolicy` support, the `groups` claim should be present in the access token (requires a "Group Membership" protocol mapper in Keycloak). 
* **Entra ID limitation**: Group claims use object IDs (GUIDs) instead of names, and are omitted entirely when a user exceeds the group overage threshold. GroupBasedPolicy must reference GUIDs and cannot be used for principals with large group memberships.
```

sdk/python/feast/permissions/auth/oidc_token_parser.py
```diff
@@ -61,17 +61,26 @@ async def _validate_token(self, access_token: str):
...
         """
-        if "preferred_username" in data:
-            return data["preferred_username"]
-        if "upn" in data:
-            return data["upn"]
+        for claim in ("preferred_username", "upn", "azp", "appid", "sub"):
+            value = data.get(claim)
+            if isinstance(value, str):
+                return value
         raise AuthenticationError(
```
aniketpalu, line 84: Suggestion - consider adding debug log. 
```suggestion
        logger.debug("Found claims: {list(data.keys())}")
        raise AuthenticationError(
```

aniketpalu (commented):
PR looks good. Few nitpicks. Thanks for the important contribution @larrysingleton007

sdk/python/tests/unit/permissions/auth/conftest.py
larrysingleton007: Done, typed as `Dict[str, str]`.

docs/getting-started/components/authz_manager.md
larrysingleton007: Applied, thanks.

sdk/python/feast/permissions/auth/oidc_token_parser.py
larrysingleton007: Added, thanks. Made it an f-string so the claim names interpolate, and it logs only the keys, not the values.

larrysingleton007:
Thanks for the review @aniketpalu. All three suggestions are applied in 25c364f00: typed the `discovery_data` fixture as `Dict[str, str]`, split the Entra group-claim caveat into its own bullet, and added the debug log of the token's claim keys before the missing-username raise. Threads resolved, ready for another look when you have a moment.

aniketpalu:
Thanks @larrysingleton007, LGTM. Please rebase.

aniketpalu (approved):

larrysingleton007:
Rebased onto the latest `master` (clean, no conflicts). Thanks for the review @aniketpalu.

ntkathole (approved):

PR #6637, opened 2026-07-23
fix: Return False from __eq__ on cross-type comparison

### LINKED ISSUE #6636: `__eq__` raises TypeError on cross-type comparison instead of returning NotImplemented
## Expected Behavior

Comparing two Feast objects of different concrete types (e.g. `FileSource` vs `SnowflakeSource`)
returns `False`, per Python's data model. In particular, changing a feature view's data source
type and re-running `feast apply` over an existing registry succeeds.

## Current Behavior

`__eq__` raises `TypeError` on a cross-type comparison instead of returning `False`. Because the
apply/diff path compares registered objects against incoming ones, changing a source type and
re-applying crashes:

```
TypeError: Comparisons should only involve FileSource class objects.
```

Minimal repro:

```python
from feast import FileSource
from feast.infra.offline_stores.snowflake_source import SnowflakeSource
a = FileSource(name="x", path="/tmp/x.parquet", timestamp_field="ts")
b = SnowflakeSource(name="x", database="D", schema="S", table="T", timestamp_field="ts")
a == b   # TypeError — should be False
```

Root cause: the affected `__eq__` methods do
`if not isinstance(other, X): raise TypeError("Comparisons should only involve X class objects.")`.
Python's data model requires `__eq__` to return `NotImplemented` for an unrecognized operand (so
the interpreter can try the reflected op and fall back to identity → `False`), never raise.

This is **systemic** — the pattern appears in ~35 `__eq__` methods across the object model:
`DataSource` and its `Kafka`/`Request`/`Kinesis` subclasses; all offline sources (`file`,
`snowflake`, `bigquery`, `redshift`, and the `ray`/`couchbase`/`trino`/`oracle`/`athena`/`mssql`/
`postgres` contrib sources); plus `Entity`, `FeatureView`, `BaseFeatureView`, `OnDemandFeatureView`,
`StreamFeatureView`, `FeatureService`, `SavedDataset`, `Project`, `ProjectMetadata`, the four
`transformation/*` classes, `Aggregation`, `LabelView`, `Permission`, and the `RoleBasedPolicy` /
`GroupBasedPolicy` / `NamespaceBasedPolicy` / `CombinedGroupNamespacePolicy` policies.
`PushSource.__eq__` already does the right thing (`return False`), so the correct pattern is
already in-tree.

## Steps to reproduce

Run the snippet above, or: define a feature view on a `FileSource` and `feast apply`; change that
feature view's source to a `SnowflakeSource` (same name) and `feast apply` again → crash.

### Specifications

- Version: 0.64.0 and 0.65.0 / master (verified; pattern unchanged)
- Subsystem: core object model (`sdk/python/feast`)

## Possible Solution

Replace each `raise TypeError("Comparisons should only involve …")` in these `__eq__` methods with
`return False`, matching `PushSource.__eq__`, which already handles a cross-type comparison this
way. `PushSource` is the one `__eq__` that does this correctly today, and there is no
`return NotImplemented` anywhere in the codebase, so `return False` keeps a single consistent
convention rather than introducing a new one. (`NotImplemented` would be the stricter data-model
answer and would also let a reflected/matcher `__eq__` participate, but the codebase already
forgoes that everywhere.)

Add a unit test asserting cross-type equality returns `False` (not a raise). This can be one sweep
across all 35 sites (uniform and mechanical) or scoped to the `DataSource` hierarchy (where the
crash reproduces) with a follow-up for the rest — happy to do whichever the maintainers prefer.
PR incoming.

larrysingleton007:

Every registry object's `__eq__` raised `TypeError` when compared against an object of a different type:

```python
if not isinstance(other, X):
    raise TypeError("Comparisons should only involve X class objects.")
```

Comparing objects of different types should be `False`, not an error. Because the apply/diff path compares registered objects against incoming ones, this meant changing a feature view's data source type and re-running `feast apply` over an existing registry crashed with `TypeError: Comparisons should only involve <X> class objects.`

This replaces the raise with `return False` across all 35 affected `__eq__` methods (`DataSource` and its `Kafka`/`Request`/`Kinesis` subclasses, the offline and contrib sources, `Entity`, `FeatureView`, `BaseFeatureView`, `OnDemandFeatureView`, `StreamFeatureView`, `FeatureService`, `SavedDataset`, `Project`, `ProjectMetadata`, the transformations, `Aggregation`, `LabelView`, `Permission`, and the policies). `PushSource.__eq__` already handled a cross-type comparison this way, so this makes every other `__eq__` consistent with it rather than introducing a new pattern (there is no `return NotImplemented` anywhere in the codebase).

Fixes #6636

# Checks

## Testing Strategy
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

Added cross-type regression tests asserting `False` (not a raise) for `DataSource`, `Entity`, and a policy object, and updated `test_label_view.py::test_equality_type_check`, which previously asserted the removed `TypeError`. Full unit suite passes locally.

# Misc

Release note: Comparing two Feast objects of different types now returns `False` instead of raising `TypeError`, so changing a feature view's data source type and re-applying no longer crashes.

Scope: this sweeps all 35 sites across the object model. Happy to narrow to just the `DataSource` hierarchy (where the crash reproduces), with a follow-up for the rest, if a smaller diff is preferred.

As an outside contributor I can't apply labels, so a maintainer will need to add a `kind/bug` label and run `ok-to-test`.

### DIFF
--- a/sdk/python/feast/aggregation/__init__.py
+++ b/sdk/python/feast/aggregation/__init__.py
@@ -90,7 +90,7 @@ class Aggregation:
 
     def __eq__(self, other):
         if not isinstance(other, Aggregation):
-            raise TypeError("Comparisons should only involve Aggregations.")
+            return False
 
         if (
             self.column != other.column
--- a/sdk/python/feast/base_feature_view.py
+++ b/sdk/python/feast/base_feature_view.py
@@ -172,9 +172,7 @@ class BaseFeatureView(ABC):
 
     def __eq__(self, other):
         if not isinstance(other, BaseFeatureView):
-            raise TypeError(
-                "Comparisons should only involve BaseFeatureView class objects."
-            )
+            return False
 
         if (
             self.name != other.name
--- a/sdk/python/feast/data_format.py
+++ b/sdk/python/feast/data_format.py
@@ -33,6 +33,8 @@ class FileFormat(ABC):
         pass
 
     def __eq__(self, other):
+        if not isinstance(other, FileFormat):
+            return False
         return self.to_proto() == other.to_proto()
 
     @classmethod
@@ -95,6 +97,8 @@ class StreamFormat(ABC):
         pass
 
     def __eq__(self, other):
+        if not isinstance(other, StreamFormat):
+            return False
         return self.to_proto() == other.to_proto()
 
     @classmethod
--- a/sdk/python/feast/data_source.py
+++ b/sdk/python/feast/data_source.py
@@ -279,7 +279,7 @@ class DataSource(ABC):
             return False
 
         if not isinstance(other, DataSource):
-            raise TypeError("Comparisons should only involve DataSource class objects.")
+            return False
 
         if (
             self.name != other.name
@@ -499,9 +499,7 @@ class KafkaSource(DataSource):
 
     def __eq__(self, other):
         if not isinstance(other, KafkaSource):
-            raise TypeError(
-                "Comparisons should only involve KafkaSource class objects."
-            )
+            return False
 
         if not super().__eq__(other):
             return False
@@ -639,9 +637,7 @@ class RequestSource(DataSource):
 
     def __eq__(self, other):
         if not isinstance(other, RequestSource):
-            raise TypeError(
-                "Comparisons should only involve RequestSource class objects."
-            )
+            return False
 
         if not super().__eq__(other):
             return False
@@ -801,9 +797,7 @@ class KinesisSource(DataSource):
 
     def __eq__(self, other):
         if not isinstance(other, KinesisSource):
-            raise TypeError(
-                "Comparisons should only involve KinesisSource class objects."
-            )
+            return False
 
         if not super().__eq__(other):
             return False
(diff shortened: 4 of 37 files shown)


larrysingleton007:
@sudohainguyen - awaiting your review please

larrysingleton007:
@ntkathole could you take a look at this one or reassign? The auto-assigned reviewer doesn't seem to be active lately, so it's been sitting since the 23rd. Checks are green and there are no open review threads.

larrysingleton007:
@ntkathole @franciscojavierarceo @jyejare @aniketpalu sorry for the broad ping, but this one still has no active reviewer. The auto-assigned reviewer hasn't reviewed anything here since October, and a second user just confirmed hitting the same bug over on #6636 (swapping a FeatureView source from FileSource to SparkSource). I've since extended the fix to cover their case as well. Checks are green, could one of you pick it up or assign someone?

franciscojavierarceo (approved):
The cross-type guards are applied consistently across the affected domain objects, preserve same-type comparison behavior, and include broad regression coverage. I found no blocking issue in the remote diff.

ntkathole (approved):

PR #6644, opened 2026-07-24
fix: Normalize SQL registry read_path to the psycopg3 driver like path

### LINKED ISSUE #6643: SQL registry read_path is not normalized to the psycopg3 driver like path
## Description

For a SQL registry, `RegistryConfig.validate_path` rewrites a bare `postgresql://` URL in `path` to `postgresql+psycopg://` so SQLAlchemy uses the psycopg3 driver, and logs a warning explaining the rewrite. `read_path` gets no equivalent treatment. It has no validator, so a bare `postgresql://` read_path reaches `create_engine` unchanged and SQLAlchemy falls back to the psycopg2 driver, diverging from `path` (which used psycopg3) with no warning.

If psycopg2 isn't installed, or a project has standardized on psycopg3, this fails at connection time with nothing pointing at read_path as the odd one out.

## Reproduction

```python
from feast.infra.registry.sql import SqlRegistryConfig

config = SqlRegistryConfig(
    path="postgresql://user@host/db",
    read_path="postgresql://user@host/replica",
)
print(config.path)       # postgresql+psycopg://user@host/db   (rewritten + warned)
print(config.read_path)  # postgresql://user@host/replica      (untouched, no warning)
```

## Expected

read_path should normalize the same way path does: rewrite a leading `postgresql://` to `postgresql+psycopg://` and log the same migration warning. An explicit `postgresql+psycopg2://` should be left as-is, matching the existing path behavior.

## Details

`validate_path` is a `@field_validator("path")` on `RegistryConfig` in `sdk/python/feast/repo_config.py`, gated on `registry_type == "sql"`. `read_path` is declared on `SqlRegistryConfig` in `sdk/python/feast/infra/registry/sql.py`; it inherits the path validator but has none of its own, so it flows straight into `create_engine(registry_config.read_path, ...)`.

Affects 0.64.0 and current master. A PR follows.

larrysingleton007:

For a SQL registry, `RegistryConfig.validate_path` rewrites a bare `postgresql://` URL in `path` to `postgresql+psycopg://` (so SQLAlchemy uses the psycopg3 driver) and logs a warning. `read_path` had no equivalent validator, so a bare `postgresql://` read_path reached `create_engine` unchanged and SQLAlchemy fell back to the psycopg2 driver, diverging from `path` (which used psycopg3) with no warning.

This factors the rewrite and warning into a shared `RegistryConfig._normalize_postgres_scheme` static helper and adds a `@field_validator("read_path")` on `SqlRegistryConfig` that mirrors `validate_path`. The `path` warning text is unchanged. The helper now rewrites only the leading scheme rather than every occurrence, so a `postgresql://` appearing later in the URL is left alone.

Fixes #6643

# Checks

## Testing Strategy
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

Added `test_sql_registry.py` cases: a bare `postgresql://` read_path is rewritten to `postgresql+psycopg://` (both with an explicit `registry_type="sql"` and with it left to default), an explicit `postgresql+psycopg2`/`postgresql+psycopg` scheme and a non-postgres scheme are left untouched, `None` stays `None`, and the migration warning is emitted for read_path. Full unit suite passes locally.

# Misc

Release note: A bare `postgresql://` `read_path` on a SQL registry is now rewritten to the psycopg3 driver (`postgresql+psycopg://`) with a warning, the same way `path` already is, instead of silently falling back to psycopg2.

As an outside contributor I can't apply labels, so a maintainer will need to add a `kind/bug` label and run `ok-to-test`.

### DIFF
--- a/sdk/python/feast/infra/registry/sql.py
+++ b/sdk/python/feast/infra/registry/sql.py
@@ -6,7 +6,7 @@ from enum import Enum
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Union, cast
 
-from pydantic import StrictInt, StrictStr
+from pydantic import StrictInt, StrictStr, ValidationInfo, field_validator
 from sqlalchemy import (  # type: ignore
     BigInteger,
     Column,
@@ -305,6 +305,17 @@ class SqlRegistryConfig(RegistryConfig):
 
     thread_pool_executor_worker_count: StrictInt = 0
     """ int: Number of worker threads to use for asynchronous caching in SQL Registry. If set to 0, it doesn't use ThreadPoolExecutor. """
+
+    @field_validator("read_path")
+    def validate_read_path(
+        cls, read_path: Optional[str], values: ValidationInfo
+    ) -> Optional[str]:
+        # Mirror `RegistryConfig.validate_path`: a bare `postgresql://` read_path
+        # must be rewritten to the psycopg3 driver too, otherwise it silently
+        # falls back to psycopg2 while `path` uses psycopg3.
+        if read_path is not None and values.data.get("registry_type") == "sql":
+            return cls._normalize_postgres_scheme(read_path, "read_path")
+        return read_path
 
 
 class SqlRegistry(CachingRegistry):
--- a/sdk/python/feast/infra/registry/sql.py
+++ b/sdk/python/feast/infra/registry/sql.py
@@ -6,7 +6,7 @@ from enum import Enum
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Union, cast
 
-from pydantic import StrictInt, StrictStr, ValidationInfo, field_validator
+from pydantic import StrictInt, StrictStr, field_validator
 from sqlalchemy import (  # type: ignore
     BigInteger,
     Column,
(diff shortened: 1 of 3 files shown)


larrysingleton007:
@ntkathole could you take a look at this one too? It never picked up a reviewer assignment, so it's been sitting since the 24th. Checks are green and there are no open review threads.

sdk/python/feast/infra/registry/sql.py
```diff
@@ -306,6 +306,17 @@ class SqlRegistryConfig(RegistryConfig):
...
     """ int: Number of worker threads to use for asynchronous caching in SQL Registry. If set to 0, it doesn't use ThreadPoolExecutor. """
 
+    @field_validator("read_path")
+    def validate_read_path(
+        cls, read_path: Optional[str], values: ValidationInfo
+    ) -> Optional[str]:
+        # Mirror `RegistryConfig.validate_path`: a bare `postgresql://` read_path
+        # must be rewritten to the psycopg3 driver too, otherwise it silently
+        # falls back to psycopg2 while `path` uses psycopg3.
+        if read_path is not None and values.data.get("registry_type") == "sql":
```
ntkathole: seems values.data.get("registry_type") == "sql" check is redundant, `SqlRegistryConfig` is only ever used for SQL registries

sdk/python/feast/repo_config.py
```diff
@@ -205,19 +205,33 @@ class RegistryConfig(FeastBaseModel):
...
+        feast standardizes on psycopg3 (``postgresql+psycopg``). Pass an explicit
+        ``postgresql+psycopg2`` to keep psycopg2. Shared by the ``path`` and
+        ``read_path`` validators so both endpoints normalize identically.
+        """
+        if value.startswith("postgresql://"):
+            _logger.warning(
+                f"The `{field_name}` of the `RegistryConfig` starts with a plain "
+                "`postgresql` string. We are updating this to `postgresql+psycopg` "
+                "to ensure that the `psycopg3` driver is used by `sqlalchemy`. If "
+                f"you want to use `psycopg2` pass `postgresql+psycopg2` explicitely "
```
ntkathole: typo explicitly

sdk/python/feast/infra/registry/sql.py
larrysingleton007: Good call, removed in fe6ce6c8c along with the now-unused ValidationInfo parameter and import.

sdk/python/feast/repo_config.py
larrysingleton007: Fixed in fe6ce6c8c, both occurrences.

ntkathole (approved):

PR #6648, opened 2026-07-27
fix: Defer feature-freshness thread to post-fork to avoid Gunicorn deadlock

### LINKED ISSUE #6647: Intermittent deadlock on `feast serve` startup under gunicorn: feature-freshness thread starts pre-fork and can freeze an import lock forever
## Describe the bug

`feast serve` (running under gunicorn, `uvicorn_worker.UvicornWorker`) can hang forever at startup with **no error, no traceback, and no log output** beyond `"Waiting for application startup."`. It's intermittent — most starts succeed, but occasionally a worker freezes permanently and never becomes ready.

## Root cause

`feast_metrics.start_metrics_server()` is called from `feature_server.start_server()`, which runs in the **gunicorn master process**, before `FeastServeApplication(...).run()` triggers gunicorn's `fork()` of the worker process(es).

`start_metrics_server()` correctly defers CPU/memory resource monitoring to *after* the fork (there's an explicit `start_resource_monitoring=not uses_gunicorn` check, plus a `post_worker_init` hook that calls `init_worker_monitoring()`). However, it does **not** apply the same treatment to the feature-freshness thread:

```python
if _config.freshness:
    freshness_thread = threading.Thread(
        target=monitor_freshness, args=(store, 30), daemon=True
    )
    freshness_thread.start()
```

This thread is started unconditionally in the master (whenever `feature_server.metrics.freshness` is enabled), and its very first action — with no delay — is `update_feature_freshness(store)` → `store.list_feature_views(...)`, which lazily builds the registry for the first time. For registry backends that need a lazy DBAPI import (e.g. the SQL registry importing `pymysql` via SQLAlchemy's `create_engine()`), this means a background thread in the gunicorn master may be **mid-import**, holding CPython's per-module import lock, at the exact moment gunicorn forks a worker.

POSIX `fork()` only duplicates the calling thread into the child process — every other thread that existed in the parent (including this freshness thread) simply ceases to exist in the worker. If the fork happens while that thread holds a module's import lock, the lock is left permanently in the "locked"/`_initializing` state inside the new worker, and there is no longer any thread anywhere that can finish that import and release it. When the worker's own FastAPI startup later tries to build its own registry and hits the same import, it deadlocks forever.

This matches a known general class of hazard (fork() + threads + an in-flight import lock — see e.g. https://bugs.python.org/issue6380), and it explains the intermittency exactly: it only manifests if the fork lands inside that narrow timing window.

We confirmed this live by attaching `py-spy` to a genuinely frozen worker process and dumping its stack — the main thread was blocked acquiring CPython's per-module import lock inside `SQLAlchemy`'s lazy `import_dbapi()` (dialects/mysql/pymysql.py), called from `SqlRegistry.__init__` → `feature_store.py`'s `registry` property → `refresh_registry()` → `feature_server.py`'s `async_refresh()` → the FastAPI `lifespan`.

We also verified against CPython's actual `importlib._bootstrap._find_and_load` that the per-module import lock is **only** acquired if a module is missing from `sys.modules` or still mid-initialization — once a module has fully finished importing, every later import statement, from any thread, before or after any number of forks, returns the cached module without ever touching the lock again. This is why the freshness thread's *pre-fork* timing is what makes it dangerous, and it points at the fix.

## To Reproduce

1. Configure a registry backend that performs a lazy import on first use (e.g. `registry_type: sql` with a MySQL connection string, which lazily imports `pymysql` via SQLAlchemy).
2. Enable `feature_server.metrics.freshness: true` in `feature_store.yaml`.
3. Run `feast serve --metrics` under gunicorn (Linux, non-Windows path) with 1+ workers.
4. Restart the process repeatedly (e.g. in a container orchestrator that recycles the pod/process on every deploy or crash). Occasionally — timing-dependent, not consistently reproducible on demand — a worker will hang forever at `"Waiting for application startup."` with zero output.

## Expected behavior

The feature-freshness thread (and ideally anything else `start_metrics_server()` spawns pre-fork that can trigger a lazy import) should be deferred to run after gunicorn forks its workers, the same way resource monitoring already is — e.g. via the existing `post_worker_init` hook — rather than starting unconditionally in the master before the fork.

## Workaround

Force the DBAPI driver (e.g. `pymysql`) to finish importing once, synchronously, as the very first statement of the process — before `feast`'s CLI runs, before any thread exists, and before gunicorn forks anything. A tiny wrapper around the CLI entrypoint works:

```python
import pymysql  # must be first — see explanation above
import sys
from feast.cli.cli import cli

if __name__ == "__main__":
    sys.exit(cli())
```

...invoked instead of the `feast` console script (e.g. `python entrypoint.py serve ...`). This is unrelated to thread timing luck — once the module is fully imported, the import lock is never touched again for it, regardless of what any other thread does afterward.

## Environment

- feast version: 0.62.0
- Python: 3.11
- OS: Linux (containerized), gunicorn + `uvicorn_worker.UvicornWorker`
- Registry: SQL registry (`registry_type: sql`, MySQL via pymysql)

casaar97:
Fixes #6647

## Summary

`feast_metrics.start_metrics_server()` starts a feature-freshness thread in the Gunicorn master process, before Gunicorn forks its worker(s). That thread's first action, with no delay, is `update_feature_freshness()` → `store.list_feature_views()`, which lazily builds the registry for the first time. For registry backends needing a lazy DBAPI import (e.g. the SQL registry importing `pymysql` via SQLAlchemy's `create_engine()`), this means a thread in the master can be mid-import, holding CPython's per-module import lock, at the exact moment Gunicorn forks a worker.

POSIX `fork()` only duplicates the calling thread into the child process; every other thread in the parent, including this one, simply ceases to exist in the worker. If the fork lands while that thread holds a module's import lock, the lock stays permanently held in the new worker, since there is no longer any thread that can finish the import and release it. The worker's own later attempt to build its registry then deadlocks forever with no error - the process just hangs at `"Waiting for application startup."`. This is intermittent by nature: it only manifests if the fork lands inside that narrow timing window (see #6647 for the full write-up, including a live `py-spy` stack trace confirming the exact mechanism).

Resource monitoring already avoids this correctly (`start_resource_monitoring=not uses_gunicorn` plus the `post_worker_init` hook calling `init_worker_monitoring()`), but the freshness thread wasn't given the same treatment. This PR applies the identical, already-established pattern:

- `start_metrics_server()` gains a `start_freshness_monitoring` flag, deferred exactly like `start_resource_monitoring` already is.
- A new `init_worker_freshness_monitoring(store)` mirrors `init_worker_monitoring()`.
- `FeastServeApplication`'s `post_worker_init` hook now also calls `init_worker_freshness_monitoring(store)` after the fork (via a closure capturing `store`, since the freshness thread needs the store reference, unlike resource monitoring).

## Test plan

- Added `TestInitWorkerFreshnessMonitoring` covering that the new function starts/doesn't start the thread based on the `freshness` config flag, mirroring the existing `init_worker_monitoring` behavior.
- Added `test_freshness_monitoring_deferred_same_as_resource_monitoring` to `TestMetricsYamlConfig`, asserting `start_freshness_monitoring` is always passed identically to `start_resource_monitoring` from `start_server()` - a regression guard so this can't silently drift out of sync again.
- `sdk/python/tests/unit/test_metrics.py` (89 tests) and `sdk/python/tests/unit/test_feature_server.py` / `test_feature_server_utils.py` (92 tests) pass locally.
- `ruff check` / `ruff format --check` pass on all changed files.

### DIFF
--- a/sdk/python/feast/feature_server.py
+++ b/sdk/python/feast/feature_server.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import asyncio
+import functools
 import os
 import sys
 import threading
@@ -1004,6 +1005,7 @@ if sys.platform != "win32":
                 store=store,
                 registry_ttl_sec=options["registry_ttl_sec"],
             )
+            self._store = store
             self._options = options
             self._metrics_enabled = metrics_enabled
             super().__init__()
@@ -1015,15 +1017,19 @@ if sys.platform != "win32":
 
             self.cfg.set("worker_class", "uvicorn_worker.UvicornWorker")
             if self._metrics_enabled:
-                self.cfg.set("post_worker_init", _gunicorn_post_worker_init)
+                self.cfg.set(
+                    "post_worker_init",
+                    functools.partial(_gunicorn_post_worker_init, self._store),
+                )
                 self.cfg.set("child_exit", _gunicorn_child_exit)
 
         def load(self):
             return self._app
 
-    def _gunicorn_post_worker_init(worker):
-        """Start per-worker resource monitoring after Gunicorn forks."""
+    def _gunicorn_post_worker_init(store: "feast.FeatureStore", worker):
+        """Start per-worker resource and freshness monitoring after Gunicorn forks."""
         feast_metrics.init_worker_monitoring()
+        feast_metrics.init_worker_freshness_monitoring(store)
 
     def _gunicorn_child_exit(server, worker):
         """Clean up Prometheus metric files for a dead worker."""
@@ -1061,6 +1067,7 @@ def start_server(
             store,
             metrics_config=flags,
             start_resource_monitoring=not uses_gunicorn,
+            start_freshness_monitoring=not uses_gunicorn,
         )
 
     logger.debug("start_server called")
--- a/sdk/python/feast/metrics.py
+++ b/sdk/python/feast/metrics.py
@@ -572,11 +572,23 @@ def init_worker_monitoring():
         t.start()
 
 
+def init_worker_freshness_monitoring(store: "FeatureStore"):
+    """Start feature-freshness monitoring inside a Gunicorn worker process.
+
+    Called from the ``post_worker_init`` hook so that each worker starts
+    its own freshness monitoring after the fork, not before.
+    """
+    if _config.freshness:
+        t = threading.Thread(target=monitor_freshness, args=(store, 30), daemon=True)
+        t.start()
+
+
 def start_metrics_server(
     store: "FeatureStore",
     port: int = 8000,
     metrics_config: Optional["_MetricsFlags"] = None,
     start_resource_monitoring: bool = True,
+    start_freshness_monitoring: bool = True,
 ):
     """
     Start the Prometheus metrics HTTP server and background monitoring threads.
@@ -593,6 +605,10 @@ def start_metrics_server(
             monitoring thread.  Set to ``False`` when Gunicorn will
             fork workers — the ``post_worker_init`` hook starts
             per-worker monitoring instead.
+        start_freshness_monitoring: Whether to start the feature-freshness
+            thread here.  Set to ``False`` when Gunicorn will fork
+            workers — the ``post_worker_init`` hook starts per-worker
+            freshness monitoring instead.
     """
     global _config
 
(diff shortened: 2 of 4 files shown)


ntkathole:
@jyejare ^

jyejare (approved):

PR #6649, opened 2026-07-28
feat(server): Remote Materialization

aniketpalu:
## Summary
Supersedes #6590 (clean single-commit history with proper DCO).
Add `?async=true` query parameter support to the existing `/materialize` and `/materialize-incremental` endpoints. When set, materialization runs in a background thread and the endpoint returns **202 Accepted** immediately. Concurrent requests for an already-MATERIALIZING FV are rejected with **409**. Also adds `?force=true` to allow operators to override stuck MATERIALIZING state (e.g. after a crashed SparkApplication pod).
### Client-side (SDK)
- `store.materialize(..., remote=True)` delegates to the feature server
  (URL/TLS derived from `online_store` config)
- `wait=False` returns immediately after triggering
- `store.poll_materialization()` for status polling via registry state
### Server-side
- `?async=true` on existing `/materialize` and `/materialize-incremental`
- `?force=true` to skip the 409 concurrency guard
- Four module-level helpers extracted for testability
### Registry
- SQL and Snowflake `apply_materialization()` now set FV state to
  `AVAILABLE_ONLINE` (parity with file-based registry)
Addresses https://github.com/feast-dev/feast/issues/4526
# Checks
- [ ] My commits are signed off (`git commit -s`)

## Testing Strategy
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual tests

# Misc


sdk/python/feast/feature_view.py
```diff
@@ -416,6 +416,7 @@ def __eq__(self, other):
             or normalize_version_string(self.version)
             != normalize_version_string(other.version)
             or self.org != other.org
+            or self.state != other.state
```
ntkathole: why this change ?

sdk/python/feast/feature_store.py
```diff
@@ -2427,12 +2426,96 @@ def _materialize_odfv(
...
+            session.close()
+
+    def _delegate_remote_materialize(
+        self,
+        endpoint: str,
+        payload: Dict[str, Any],
+        force: bool = False,
+    ) -> None:
+        """Fire-and-forget POST to feature server with ?async=true."""
+        query_params = {"async": "true"}
```
ntkathole: may be allow user to pass param, There's no way for an SDK user to do synchronous materialization now.

franciscojavierarceo (changes requested):
I found two blocking correctness issues:

1. The async concurrency guard is check-then-act. `_check_already_materializing()` reads registry state, but `MATERIALIZING` is not written until the executor task starts. Two concurrent `async=true` requests can therefore both pass the check and enqueue duplicate materializations. Please atomically reserve the feature views before returning 202 (and roll the reservation back if submission fails), or use a registry compare-and-set/lock.
2. `MaterializeIncrementalRequest.version` is accepted and sent by the SDK, but both server paths call `store.materialize_incremental()` without `version=request.version`; authorization resolution also ignores it. A versioned request can therefore materialize a different feature-view version than requested. Please thread `version` through both sync and async paths and cover it in the server tests.

The existing review thread about the SDK no longer exposing synchronous mode also remains unresolved.

sdk/python/feast/feature_view.py
aniketpalu: not required for this change. Reverting; will push shortly.

sdk/python/feast/feature_store.py
```diff
@@ -510,7 +510,7 @@ def _transition_fv_to_materializing(
 
         Rolls back all already-transitioned FVs if this one can't transition.
         """
-        previous_state = getattr(feature_view, "state", None)
+        previous_states[feature_view.name] = getattr(feature_view, "state", None)
```
jyejare: **[Critical] Race condition in state tracking initialization**

The line `previous_states[feature_view.name] = getattr(feature_view, "state", None)` was moved before the state transition logic, but this creates a critical bug. If the transition fails and we need to rollback other FVs, the previous_states dict won't have entries for FVs that failed before this one was reached.

**Suggested:**
```suggestion
        previous_state = getattr(feature_view, "state", None)
        if (
            hasattr(feature_view, "state")
            and feature_view.state != FeatureViewState.STATE_UNSPECIFIED
        ):
            # ... validation logic ...
            feature_view.state = FeatureViewState.MATERIALIZING
            self.registry.apply_feature_view(feature_view, self.project, commit=True)
        previous_states[feature_view.name] = previous_state
```

sdk/python/feast/feature_server.py
```diff
@@ -798,36 +917,49 @@ async def chat_ui():
...
+                            start_date,
+                            end_date,
+                            fv_names,
+                            disable_event_timestamp=request.disable_event_timestamp,
+                            full_feature_names=request.full_feature_names,
+                            version=request.version,
+                        )
+                    except Exception as e:
+                        logger.error(
+                            f"Async materialization failed for {fv_names}: {e}",
```
jyejare: **[Critical] Background task exception handling insufficient**

The async materialization runs in a background thread but doesn't handle critical failures properly. If materialization fails, only the FV state is reset to GENERATED, but there's no way for the client to know the operation failed since the API already returned 202. This could lead to silent failures in production.

**Suggested:**
```suggestion
                    except Exception as e:
                        logger.error(
                            f"Async materialization failed for {fv_names}: {e}",
                            exc_info=True,
                        )
                        _update_fv_state(store, fv_names, FeatureViewState.GENERATED)
                        # TODO: Consider implementing a status endpoint or webhook callback
                        # for clients to check materialization status
```

sdk/python/feast/feature_server.py
```diff
@@ -333,6 +337,121 @@ async def load_static_artifacts(app: FastAPI, store):
...
+                    f"views are already in MATERIALIZING state: {conflicting}. "
+                    f"Use ?force=true to override."
+                ),
+                "feature_views": conflicting,
+            },
+        )
+    return None
+
+
+def _update_fv_state(
```
jyejare, line 409: **[Warning] Silent exception handling could hide real errors**

The `_reset_stuck_materializing_to_generated` and `_check_already_materializing` functions silently ignore all exceptions when accessing feature views. This could hide legitimate errors like registry corruption or network issues, making debugging difficult.

**Suggested:**
```suggestion
        except (FeatureViewNotFoundException, KeyError):
            # Expected when FV doesn't exist
            pass
        except Exception as e:
            logger.warning(f"Unexpected error checking state for {fv_name}: {e}")
            pass
```

sdk/python/feast/feature_view.py
jyejare: **[Suggestion] State comparison added to __eq__ affects hash consistency**

Adding state to the equality comparison means two otherwise identical FeatureViews will be considered different if they're in different states (e.g., GENERATED vs MATERIALIZING). This could break set operations, dict lookups, and caching logic that expects state changes to not affect object identity.

**Suggested:**
```suggestion
# Consider if state should be included in equality. If FeatureViews should be
# considered equal regardless of materialization state, remove this line.
# If state is semantically important for equality, ensure __hash__ is also updated
# or make the class unhashable.
```

sdk/python/feast/feature_server.py
```diff
@@ -94,12 +96,14 @@ class MaterializeRequest(BaseModel):
...
 
 
 class MaterializeIncrementalRequest(BaseModel):
     end_ts: str
     feature_views: Optional[List[str]] = None
     full_feature_names: bool = False
+    version: Optional[str] = None
 
 
 class GetOnlineFeaturesRequest(BaseModel):
```
jyejare, line 122: **[Suggestion] Missing version parameter documentation and validation**

The version parameter was added to both MaterializeRequest and MaterializeIncrementalRequest but there's no documentation about what values are valid or how it affects the materialization behavior.

**Suggested:**
```suggestion
    version: Optional[str] = Field(
        None, 
        description="Optional version to materialize (e.g., 'v2'). Requires feature_views with exactly one entry."
    )
```

sdk/python/feast/feature_server.py
```diff
@@ -839,27 +971,49 @@ async def materialize(request: MaterializeRequest) -> None:
             )
```
jyejare, line 1016: **[Nitpick] Inconsistent parameter passing in sync mode**

In the synchronous code path for materialize, the version parameter is passed but the feature_views parameter uses the original request.feature_views instead of the resolved fv_names list.

**Suggested:**
```suggestion
            await run_in_threadpool(
                store.materialize,
                start_date,
                end_date,
                fv_names,  # Use resolved names for consistency
                disable_event_timestamp=request.disable_event_timestamp,
                full_feature_names=request.full_feature_names,
                version=request.version,  # Don't forget version parameter
            )
```

jyejare (changes requested):
This PR implements remote materialization support, allowing clients with a remote online store topology to delegate materialization requests to a feature server via async HTTP endpoints. The implementation adds proper conflict detection, state management, and force override capabilities.

sdk/python/feast/feature_server.py
aniketpalu: Thanks — agreed that fire-and-forget 202 means clients don’t get the failure on the HTTP response. That’s intentional for this PR: async acceptance + observe completion via FeatureView.state / materialization_intervals (and run_async=False when the client needs the sync HTTP path to block/error).

A dedicated status endpoint / job handle / webhook is out of scope here and planned as a follow-up (along the RemoteComputeEngine / job-tracking direction discussed earlier). I’ve added a TODO in the failure path comment pointing to that follow-up.

sdk/python/feast/feature_view.py
aniketpalu: Follow-up: reverted in d8e7a9974 - `FeatureView.state` is no longer part of `__eq__`.

sdk/python/feast/feature_store.py
aniketpalu: Agreed — added `run_async: bool = True` on `materialize()` / `materialize_incremental()`.
- Default `run_async=True`: remote topology POSTs with `?async=true`, returns after 202
- `run_async=False`: POSTs without async and blocks until the server finishes sync materialization
- `force=True` still requires `run_async=True` (force only applies on the async path)

aniketpalu:
@franciscojavierarceo 

Thanks — both blocking items addressed in d8e7a9974:
1. **Race / reserve before 202:** async path now sets `MATERIALIZING` before returning 202. Concurrent requests hit 409. `store.materialize()` treats already-`MATERIALIZING` as a no-op (idempotent transition; rollback target `GENERATED` for the reserved case).
2. **`version` threading:** `version` is passed through `_authorize_materialize_views(...)` and into both sync and async calls for `/materialize` and `/materialize-incremental`.
Also resolved the SDK sync gap via `run_async` (see related thread).
Still TODO on my side: add server unit coverage that asserts `version` is forwarded on both paths — will follow up shortly.

sdk/python/feast/feature_store.py
aniketpalu: Thanks for the careful look — walked through both placements with the caller loop in mind; they have the same observable outcome (including failure cases).
Caller (simplified):
```python
for feature_view, fv_start in fv_with_dates:
    self._transition_fv_to_materializing(feature_view, regular_fvs, previous_states)
    regular_fvs.append(feature_view)  # only after a successful return
```

On a failed transition, the failing FV is not in already_transitioned / regular_fvs. Rollback only restores FVs that already succeeded.

sdk/python/feast/feature_view.py
aniketpalu: Agreed

sdk/python/feast/feature_server.py
aniketpalu: Done

sdk/python/feast/feature_server.py
aniketpalu: sync path now passes resolved `fv_names` and `version=request.version` (aligned with async).

sdk/python/feast/feature_server.py
aniketpalu: Accepted

sdk/python/feast/feature_server.py
```diff
@@ -798,70 +936,115 @@ async def chat_ui():
...
+                        )
+                    except Exception as e:
+                        logger.error(
+                            f"Async materialize-incremental failed for {fv_names}: {e}",
+                            exc_info=True,
+                        )
+                        _update_fv_state(store, fv_names, FeatureViewState.GENERATED)
+
+                loop = asyncio.get_running_loop()
+                loop.run_in_executor(None, _run_materialize_incremental)
```
ntkathole: non-blocking but I think it's better if we used dedicated executor, instead of default shared pool, for materialization so that it won't block other operations if multiple executors in progress

sdk/python/feast/feature_store.py
```diff
@@ -2427,12 +2433,116 @@ def _materialize_odfv(
...
+                return {"status": "accepted", "raw": response.text}
+        finally:
+            session.close()
+
+    def _delegate_remote_materialize(
+        self,
+        endpoint: str,
+        payload: Dict[str, Any],
+        force: bool = False,
+        run_async: bool = True,
```
ntkathole: should this be by default False ? since it's change in behavior for existing users - the call returns successfully even if the server-side materialization fails later.

aniketpalu: Agree, updating soon

sdk/python/feast/feature_server.py
aniketpalu: Agreed, added separate executor

ntkathole (approved):

PR #6657, opened 2026-07-29
feat: Implemented registry refresh action feature 

### LINKED ISSUE #6650: Feast UI fails to display newly created project after feast apply when using SQL registry
## Expected Behavior
After running feast apply to register a new project and its features, the Feast UI should display the new project and allow searching/browsing its features.

## Current Behavior
When using a SQL-based registry (e.g., SQLite), feast apply successfully creates the project and persists all feature metadata to the database. However, the Feast UI cannot discover the newly created project — searching by the project name returns no results, and the project does not appear in the project list.

## Steps to Reproduce
1. Initialize a Feast project: feast init -t local
2. Configure feature_store.yaml to use a SQL registry:
   ```
   registry:
     registry_type: sql
     path: sqlite:///data/registry_sqlite.db
   ```
3. Define features under project picked_elk in feature_definitions.py
4. Run feast apply — project picked_elk and all features are created successfully
5. Start Feast UI: feast ui — project picked_elk is visible ✅
6. Modify feature_definitions.py to change the project name to picked_elk2
7. Run feast apply again — project picked_elk2 is created successfully, DB tables are created
Image
8. Search for picked_elk2 in Feast UI — project not found ❌### Specifications
Image

- Version: Feast 0.62.0
- Platform: Windows (also likely affects Linux/macOS)
- Subsystem: Feast UI + SQL Registry integration

AdityaPatil22:
`projects-list.json` was written as a static file once at server startup and never refreshed. When a user ran `feast apply` to create a new project while the UI was running, the new project never appeared in the UI's project list.

This PR replaces the static file write with a dynamic FastAPI route (`GET /projects-list.json`) that queries the registry on each request, so newly created projects appear immediately on page refresh.

Fixes #6650

# Checks

## Testing Strategy
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

# Misc
Existing `test_projects_list_*` tests updated to hit the dynamic endpoint via `TestClient` instead of reading the static file from disk. New test `test_projects_list_dynamic_refresh` verifies that mutating the registry after `get_app` is reflected in subsequent requests — the core assertion for #6650.

### DIFF
--- a/docs/how-to-guides/online-server-performance-tuning.md
+++ b/docs/how-to-guides/online-server-performance-tuning.md
@@ -266,15 +266,6 @@ The `registryTTLSeconds` field on the Operator CR (or `--registry_ttl_sec` CLI f
 | Production (low-latency) | `thread` | 300 |
 | Production (frequent schema changes) | `thread` | 60 |
 
-### UI refresh
-
-The Feast UI caches the project list using the same registry cache. After running `feast apply` to add a new project, it may take up to `cache_ttl_seconds` before the project appears in the UI.
-
-To see new projects faster:
-
-- **Lower the TTL**: Set `cache_ttl_seconds: 10` (or similar) in your `feature_store.yaml` registry config. This makes all registry consumers — including the UI — pick up changes within 10 seconds.
-- **Refresh on demand**: The project selection page has a **Refresh** button that explicitly invalidates the server-side registry cache (`POST /api/registry/refresh`) and reloads the project list without a full page refresh.
-
 ---
 
 ## Online store selection
--- a/docs/how-to-guides/online-server-performance-tuning.md
+++ b/docs/how-to-guides/online-server-performance-tuning.md
@@ -266,6 +266,15 @@ The `registryTTLSeconds` field on the Operator CR (or `--registry_ttl_sec` CLI f
 | Production (low-latency) | `thread` | 300 |
 | Production (frequent schema changes) | `thread` | 60 |
 
+### UI refresh
+
+The Feast UI caches the project list using the same registry cache. After running `feast apply` to add a new project, it may take up to `cache_ttl_seconds` before the project appears in the UI.
+
+To see new projects faster:
+
+- **Lower the TTL**: Set `cache_ttl_seconds: 10` (or similar) in your `feature_store.yaml` registry config. This makes all registry consumers — including the UI — pick up changes within 10 seconds.
+- **Refresh on demand**: The project selection page has a **Refresh** button that explicitly invalidates the server-side registry cache (`POST /api/registry/refresh`) and reloads the project list without a full page refresh.
+
 ---
 
 ## Online store selection
--- a/docs/reference/alpha-web-ui.md
+++ b/docs/reference/alpha-web-ui.md
@@ -153,3 +153,12 @@ const tabsRegistry = {
 ```
 
 Examples of custom tabs can be found in the `ui/custom-tabs` folder.
+
+## Refreshing the project list
+
+The Feast UI caches the project list using the same registry cache. After running `feast apply` to add a new project, it may take up to `cache_ttl_seconds` before the project appears in the UI.
+
+To see new projects faster:
+
+- **Lower the TTL**: Set `cache_ttl_seconds: 10` (or similar) in your `feature_store.yaml` registry config. This makes all registry consumers — including the UI — pick up changes within 10 seconds.
+- **Refresh on demand**: The project selection page has a **Refresh** button that explicitly invalidates the server-side registry cache (`POST /api/v1/registry/refresh`) and reloads the project list without a full page refresh.
--- a/docs/reference/alpha-web-ui.md
+++ b/docs/reference/alpha-web-ui.md
@@ -154,11 +154,11 @@ const tabsRegistry = {
 
 Examples of custom tabs can be found in the `ui/custom-tabs` folder.
 
-## Refreshing the project list
+## Refreshing the registry
 
-The Feast UI caches the project list using the same registry cache. After running `feast apply` to add a new project, it may take up to `cache_ttl_seconds` before the project appears in the UI.
+The Feast UI caches registry data (projects, feature views, entities, etc.) using the registry cache. After running `feast apply` to make changes, it may take up to `cache_ttl_seconds` before the updates appear in the UI.
 
-To see new projects faster:
+To see changes faster:
 
 - **Lower the TTL**: Set `cache_ttl_seconds: 10` (or similar) in your `feature_store.yaml` registry config. This makes all registry consumers — including the UI — pick up changes within 10 seconds.
-- **Refresh on demand**: The project selection page has a **Refresh** button that explicitly invalidates the server-side registry cache (`POST /api/v1/registry/refresh`) and reloads the project list without a full page refresh.
+- **Refresh on demand**: The UI has a **Refresh** button that explicitly invalidates the server-side registry cache (`POST /api/v1/registry/refresh`) and reloads the UI without a full page refresh.
--- a/sdk/python/feast/api/registry/rest/__init__.py
+++ b/sdk/python/feast/api/registry/rest/__init__.py
@@ -1,7 +1,7 @@
 import logging
 from typing import Any, Optional
 
-from fastapi import FastAPI
+from fastapi import FastAPI, Response, status
 
 from feast.api.registry.rest.compute_engines import get_compute_engine_router
 from feast.api.registry.rest.data_sources import get_data_source_router
@@ -43,6 +43,21 @@ def register_all_routes(app: FastAPI, grpc_handler, server=None, store=None):
     app.include_router(get_saved_dataset_router(grpc_handler))
     app.include_router(get_monitoring_router(grpc_handler, store=resolved_store))
     app.include_router(get_compute_engine_router(grpc_handler, store=resolved_store))
+
+    if resolved_store:
+
+        @app.post("/registry/refresh")
+        def refresh_registry():
+            try:
+                resolved_store.refresh_registry()
+                return Response(status_code=status.HTTP_200_OK)
+            except Exception:
+                logger.exception("Registry refresh failed")
+                return Response(
+                    content='{"detail":"Registry refresh failed. Check server logs for details."}',
+                    status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+                    media_type="application/json",
+                )
 
     _register_openlineage_consumer(app, resolved_store)
 
--- a/sdk/python/feast/api/registry/rest/__init__.py
+++ b/sdk/python/feast/api/registry/rest/__init__.py
@@ -49,6 +49,14 @@ def register_all_routes(app: FastAPI, grpc_handler, server=None, store=None):
         @app.post("/registry/refresh")
         def refresh_registry():
             try:
+                from feast.permissions.action import AuthzedAction
+                from feast.permissions.security_manager import assert_permissions
+
+                project = resolved_store.registry.get_project(
+                    name=resolved_store.project, allow_cache=True
+                )
+                assert_permissions(resource=project, actions=[AuthzedAction.UPDATE])
+
                 resolved_store.refresh_registry()
                 return Response(status_code=status.HTTP_200_OK)
             except Exception:
--- a/sdk/python/feast/api/registry/rest/__init__.py
+++ b/sdk/python/feast/api/registry/rest/__init__.py
@@ -1,7 +1,7 @@
 import logging
 from typing import Any, Optional
 
-from fastapi import FastAPI, Response, status
+from fastapi import FastAPI
 
 from feast.api.registry.rest.compute_engines import get_compute_engine_router
 from feast.api.registry.rest.data_sources import get_data_source_router
@@ -14,7 +14,7 @@ from feast.api.registry.rest.lineage import get_lineage_router
 from feast.api.registry.rest.metrics import get_metrics_router
 from feast.api.registry.rest.monitoring import get_monitoring_router
 from feast.api.registry.rest.permissions import get_permission_router
-from feast.api.registry.rest.projects import get_project_router
+from feast.api.registry.rest.projects import get_project_router, get_registry_router
 from feast.api.registry.rest.saved_datasets import get_saved_dataset_router
 from feast.api.registry.rest.search import get_search_router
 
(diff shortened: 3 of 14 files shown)


franciscojavierarceo (commented):
Moving `projects-list.json` behind a route avoids freezing the response at startup, but `_build_projects_list()` still calls `list_projects(allow_cache=True)`. With the default registry cache TTL of 600 seconds, a newly added project can remain absent for up to ten minutes—the exact startup-empty case this is intended to fix. Please consider an uncached read (or explicit refresh) for this low-frequency UI endpoint and add the regression test described in the PR body; the current diff changes only `ui_server.py` and contains no test update.

AdityaPatil22:
@franciscojavierarceo - Updated _build_projects_list to use allow_cache=False so new projects are visible immediately. Also added the regression test (test_projects_list_dynamic_refresh) and updated the existing tests to hit the dynamic endpoint via TestClient instead of reading the static file from disk.

ntkathole:
@AdityaPatil22 This turns every browser refresh into a full database round-trip, completely bypassing the caching layer that Feast deliberately built. 

The fix should keep `allow_cache=True` and let the existing TTL handle staleness. Users who need faster UI refresh can already set cache_ttl_seconds: 10 (or whatever suits them) in their config.  

I think we can :
- Document that users can lower cache_ttl_seconds for faster UI refresh.
- Add a Refresh button in the UI that calls `store.registry.refresh()` (via a new POST /api/registry/refresh endpoint). This lets users explicitly invalidate the cache on demand without adding hidden performance costs on every page load.

AdityaPatil22:
@ntkathole - I have added a refresh button in the UI which lets the users invalidate the cache on demand. Please let me know if there needs to be any change in the UI for the button.

Screenshot 2026-07-30 at 3 31 41 PM

docs/how-to-guides/online-server-performance-tuning.md
```diff
@@ -266,6 +266,15 @@ The `registryTTLSeconds` field on the Operator CR (or `--registry_ttl_sec` CLI f
 | Production (low-latency) | `thread` | 300 |
 | Production (frequent schema changes) | `thread` | 60 |
 
+### UI refresh
+
```
ntkathole: I think better place to add these docs is at  docs/reference/alpha-web-ui.md

sdk/python/feast/ui_server.py
```diff
@@ -889,9 +889,16 @@ def get_app(
 
     ui_dir_ref = importlib_resources.files(__spec__.parent) / "ui/build/"  # type: ignore[name-defined, arg-type]
     with importlib_resources.as_file(ui_dir_ref) as ui_dir:
-        projects_dict = _build_projects_list(store, project_id, root_path)
```
ntkathole, line 892: this change is not required and can be kept same as original

sdk/python/feast/ui_server.py
```diff
@@ -889,9 +889,16 @@ def get_app(
...
-        with ui_dir.joinpath("projects-list.json").open(mode="w") as f:
-            f.write(json.dumps(projects_dict))
+        pass
+
+    @app.get("/projects-list.json")
+    def get_projects_list():
+        return _build_projects_list(store, project_id, root_path)
+
+    @app.post("/api/registry/refresh")
```
ntkathole: this endpoint should be on REST API server

ui/src/pages/RootProjectSelectionPage.tsx
```diff
@@ -48,9 +74,23 @@ const RootProjectSelectionPage = () => {
...
-          <h1>Welcome to Feast</h1>
-        </EuiTitle>
+        <EuiFlexGroup alignItems="center" justifyContent="spaceBetween">
+          <EuiFlexItem grow={false}>
+            <EuiTitle size="s">
+              <h1>Welcome to Feast</h1>
+            </EuiTitle>
+          </EuiFlexItem>
+          <EuiFlexItem grow={false}>
+            <EuiButtonIcon
```
ntkathole: This refresh button should be accessible from any page, also you can use a proper labeled button for refresh

docs/how-to-guides/online-server-performance-tuning.md
AdityaPatil22: Moved the docs to docs/reference/alpha-web-ui.md file

sdk/python/feast/ui_server.py
AdityaPatil22: Reverted the changes to original state

sdk/python/feast/ui_server.py
AdityaPatil22: Moved the endpoint from the main app to rest_app inside _setup_rest_mode, so it's now served at /api/v1/registry/refresh on the REST API server.

ui/src/pages/RootProjectSelectionPage.tsx
AdityaPatil22: Moved the refresh button to header bar for it to be accessible from every page.

AdityaPatil22: Screenshot 2026-07-30 at 9 58 55 PM

ui/src/pages/Layout.tsx
```diff
@@ -242,6 +267,17 @@ const Layout = () => {
...
                   {!data && <EuiFlexItem />}
 
+                  <EuiFlexItem grow={false}>
+                    <EuiButton
+                      iconType="refresh"
+                      onClick={handleRefresh}
+                      isLoading={refreshing}
+                      size="s"
+                    >
+                      Refresh Registry
```
ntkathole: just Refresh would work, "Refresh Registry" - too technical for end user

ui/src/pages/Layout.tsx
```diff
@@ -146,6 +156,21 @@ const Layout = () => {
...
+    try {
+      await fetch(`${basename}/api/v1/registry/refresh`, { method: "POST" });
+      const res = await fetch(`${basename}/projects-list.json`, {
+        headers: { "Content-Type": "application/json" },
+      });
+      const json = await res.json();
+      const parsed = ProjectsListSchema.parse(json);
+      queryClient.setQueryData("feast-projects-list", parsed);
+    } finally {
+      setRefreshing(false);
```
ntkathole: No user feedback on success or failure, Add EuiToast / EuiGlobalToastList with message like "Refresh successful/failed"

ui/src/pages/Layout.tsx
```diff
@@ -242,6 +267,17 @@ const Layout = () => {
                   )}
                   {!data && <EuiFlexItem />}
 
+                  <EuiFlexItem grow={false}>
+                    <EuiButton
```
ntkathole: Use EuiButtonEmpty or EuiButtonIcon with a tooltip instead of a filled button
Or use color="text" to make it less prominent than default primary

docs/reference/alpha-web-ui.md
```diff
@@ -153,3 +153,12 @@ const tabsRegistry = {
 ```
 
 Examples of custom tabs can be found in the `ui/custom-tabs` folder.
+
+## Refreshing the project list
```
ntkathole: it's not just refreshing project list, it's refreshing complete registry

ui/src/pages/Layout.tsx
```diff
@@ -1,6 +1,7 @@
-import React, { useState, useRef, useEffect } from "react";
+import React, { useState, useRef, useEffect, useContext } from "react";
```
ntkathole: If a user lands on the project selection page and doesn't see their project, the Refresh button in the Layout header isn't visible - they have to first pick a project, then refresh.

sdk/python/feast/ui_server.py
```diff
@@ -158,6 +158,11 @@ async def feast_object_not_found_handler(
 
     register_all_routes(rest_app, grpc_handler, store=store)
 
+    @rest_app.post("/registry/refresh")
```
ntkathole: Add a `/registry/refresh` route in `register_all_routes()` (in feast/api/registry/rest/) so it's available on all REST registry servers, not just ui server

ui/src/pages/Layout.tsx
AdityaPatil22: Updated the button label  to "Refresh"

ui/src/pages/Layout.tsx
AdityaPatil22: Added EuiGlobalToastList in Layout.tsx with "Refresh successful" (green) and "Refresh failed" (red) toasts.

ui/src/pages/Layout.tsx
AdityaPatil22: Changed from EuiButton to EuiButtonEmpty with iconType="refresh". Also added isDisabled={refreshing} to prevent double-click race conditions since isLoading alone only shows a spinner but does not block clicks.

docs/reference/alpha-web-ui.md
AdityaPatil22: Updated the section heading

ui/src/pages/Layout.tsx
AdityaPatil22: Added a Refresh button directly on RootProjectSelectionPage.tsx (next to the "Welcome to Feast" heading) that shares the same refresh logic via RegistryRefreshContext. The Layout header button is gated with `{projectName &&`  so only one button is visible at a time, the page-level button on the project selection screen, and the header button everywhere else. Toasts are rendered in Layout so they appear regardless of which button is clicked.

sdk/python/feast/ui_server.py
AdityaPatil22: Moved the /registry/refresh endpoint into register_all_routes()

sdk/python/feast/ui_server.py
```diff
@@ -893,6 +893,10 @@ def get_app(
         with ui_dir.joinpath("projects-list.json").open(mode="w") as f:
             f.write(json.dumps(projects_dict))
 
+    @app.get("/projects-list.json")
```
ntkathole: no need to change this

sdk/python/tests/unit/test_ui_server.py
```diff
@@ -148,11 +148,12 @@ def test_catch_all_route(ui_app_with_registry):
 # ---------- projects-list.json tests ----------
 
 
-def _read_projects_list(temp_dir):
-    """Read the projects-list.json written by get_app via the mock (ui_dir = temp_dir)."""
-    projects_file = os.path.join(temp_dir, "projects-list.json")
-    with open(projects_file) as f:
-        return json.load(f)
+def _get_projects_list(app):
```
ntkathole: can be kept as same

sdk/python/feast/api/registry/rest/__init__.py
```diff
@@ -44,6 +44,21 @@ def register_all_routes(app: FastAPI, grpc_handler, server=None, store=None):
     app.include_router(get_monitoring_router(grpc_handler, store=resolved_store))
     app.include_router(get_compute_engine_router(grpc_handler, store=resolved_store))
 
+    if resolved_store:
+
+        @app.post("/registry/refresh")
+        def refresh_registry():
+            try:
+                resolved_store.refresh_registry()
```
ntkathole: Authorization check is missing, somethign like can be added 

```
       # Require UPDATE permission on the project to refresh the registry
        project = resolved_store.registry.get_project(
            name=resolved_store.project, allow_cache=True
        )
        assert_permissions(resource=project, actions=[AuthzedAction.UPDATE])
```

ntkathole:
Also, please add unit test for the rest api added

ui/src/hooks/useRegistryRefresh.ts
```diff
@@ -0,0 +1,72 @@
...
+  const basename = projectListCtx?.basename || "";
+
+  const removeToast = useCallback((removedToast: { id: string }) => {
+    setToasts((prev) => prev.filter((t) => t.id !== removedToast.id));
+  }, []);
+
+  const handleRefresh = useCallback(async () => {
+    setRefreshing(true);
+    try {
+      const refreshRes = await fetch(`${basename}/api/v1/registry/refresh`, {
```
ntkathole, line 31: no auth headers passed here, uses configured `fetchOptions` (from FeastUISansProviders.tsx) that include auth headers.

sdk/python/feast/ui_server.py
AdityaPatil22: The dynamic route is needed for the Refresh button workflow to work correctly. 

When a user clicks Refresh:
POST /api/v1/registry/refresh → invalidates the registry cache
GET /projects-list.json → needs to return the updated project list

If we keep only the static file (written once at startup), the Refresh button would invalidate the cache but /projects-list.json would still serve stale data, new projects wouldn't appear in the project selector until the server is restarted.

sdk/python/tests/unit/test_ui_server.py
AdityaPatil22: Removed the changes

sdk/python/feast/api/registry/rest/__init__.py
AdityaPatil22: Added the Authorization check.

ui/src/hooks/useRegistryRefresh.ts
AdityaPatil22: Added the Auth headers for the call

sdk/python/feast/api/registry/rest/__init__.py
```diff
@@ -44,6 +44,29 @@ def register_all_routes(app: FastAPI, grpc_handler, server=None, store=None):
...
+                from feast.permissions.security_manager import assert_permissions
+
+                project = resolved_store.registry.get_project(
+                    name=resolved_store.project, allow_cache=True
+                )
+                assert_permissions(resource=project, actions=[AuthzedAction.UPDATE])
+
+                resolved_store.refresh_registry()
+                return Response(status_code=status.HTTP_200_OK)
+            except Exception:
```
ntkathole: `FeastPermissionError` caught by generic except Exception, returns 500 instead of 403, Permission denial will look like a server error to the user

sdk/python/feast/ui_server.py
ntkathole: so, now instead of writing static file which is not being used, can the static file write be removed and all subsequent code be moved inside the with block ?

sdk/python/feast/api/registry/rest/__init__.py
```diff
@@ -44,6 +44,29 @@ def register_all_routes(app: FastAPI, grpc_handler, server=None, store=None):
     app.include_router(get_monitoring_router(grpc_handler, store=resolved_store))
     app.include_router(get_compute_engine_router(grpc_handler, store=resolved_store))
 
+    if resolved_store:
+
+        @app.post("/registry/refresh")
```
ntkathole: Also, please follow the same pattern as every other endpoint in the REST API. The current inline approach works functionally but doesn't match the codebase conventions.

You can add to project.py existing file:

```
def get_registry_router(store) -> APIRouter:
    router = APIRouter()

    @router.post("/registry/refresh")
    def refresh_registry():
        from feast.permissions.action import AuthzedAction
        from feast.permissions.security_manager import assert_permissions

        project = store.registry.get_project(
            name=store.project, allow_cache=True
        )
        assert_permissions(resource=project, actions=[AuthzedAction.UPDATE])
        store.refresh_registry()
        return Response(status_code=status.HTTP_200_OK)

    return router
```

sdk/python/feast/api/registry/rest/__init__.py
AdityaPatil22: Fixed by switching to the clean router pattern and removing the try/except block. FeastPermissionError now propagates to the registered exception handler, which correctly returns a 403 response.

sdk/python/feast/api/registry/rest/__init__.py
AdityaPatil22: Moved the endpoint into projects.py

sdk/python/feast/ui_server.py
ntkathole: @AdityaPatil22 ^ ?

ntkathole:
Overall looks good now, pending 1 comment and unit tests failures

sdk/python/feast/ui_server.py
AdityaPatil22: Removed the static file write and moved all subsequent code inside the with block.

ntkathole (approved):
Looks good

PR #6659, opened 2026-07-29
fix: Use matching proto class per feature view list in SqliteOnlineStore.plan()

### LINKED ISSUE #6658: sqlite.py's OnlineStore.plan() raises typeguard.TypeCheckError for stream feature views
## Expected Behavior

`feast plan` (or any `OnlineStore.plan()` implementation that follows the same pattern as `SqliteOnlineStore.plan()`) should work the same way whether the registry contains regular feature views, streaming feature views, or both.

## Current Behavior

Feast has two kinds of feature views: regular ones (`FeatureView`) and streaming ones (`StreamFeatureView`, for Kafka-style sources). Under the hood, each is a distinct protobuf message type — `FeatureViewProto` vs `StreamFeatureViewProto`. They're similar in shape, but not interchangeable.

`SqliteOnlineStore.plan()` loops over every feature view in the registry — both kinds together — and reconstructs each one with the same call ([sdk/python/feast/infra/online_stores/sqlite.py](https://github.com/feast-dev/feast/blob/v0.65.0/sdk/python/feast/infra/online_stores/sqlite.py#L318-L337)):

```python
def plan(
    self, config: RepoConfig, desired_registry_proto: RegistryProto
) -> List[InfraObject]:
    ...
    infra_objects: List[InfraObject] = [
        SqliteTable(
            path=self._get_db_path(config),
            name=_table_id(project, FeatureView.from_proto(view), versioning),
        )
        for view in [
            *desired_registry_proto.feature_views,
            *desired_registry_proto.stream_feature_views,
        ]
    ]
```

`FeatureView.from_proto()` is guarded by a runtime type checker (`typeguard`, via `@typechecked` on the class — [sdk/python/feast/feature_view.py#L110-L111](https://github.com/feast-dev/feast/blob/v0.65.0/sdk/python/feast/feature_view.py#L110-L111)) that only accepts a `FeatureViewProto`. It works fine for the `feature_views` list. But when the loop reaches an entry from `stream_feature_views`, it hands over a `StreamFeatureViewProto` — the wrong type for that function — and typeguard raises:

```
typeguard.TypeCheckError: argument "feature_view_proto" (feast.core.StreamFeatureView_pb2.StreamFeatureView) is not an instance of feast.core.FeatureView_pb2.FeatureView
```

In short: `plan()` calls the wrong reconstruction function on the wrong kind of feature view, so it crashes for any repo that has at least one `StreamFeatureView` registered — a fairly common setup, not an edge case.

We ran into this adding `plan()`/`InfraObject` support to a custom DynamoDB online store (the stock `DynamoDBOnlineStore.plan()` is just the no-op default and never triggers this) and initially copied `sqlite.py`'s pattern as the reference implementation, since it's the only in-tree store that implements `plan()`.

## Steps to reproduce

```python
from feast import FileSource
from feast.data_format import JsonFormat
from feast.data_source import KafkaSource
from feast.stream_feature_view import StreamFeatureView
from feast.infra.online_stores.sqlite import SqliteOnlineStore
from feast.protos.feast.core.Registry_pb2 import Registry as RegistryProto

config = RepoConfig(project="p", registry="/tmp/x.db", provider="local", online_store="sqlite")

stream_fv = StreamFeatureView(
    name="driver_dropoffs_stream",
    entities=[],
    schema=[],
    source=KafkaSource(
        name="k",
        message_format=JsonFormat(schema_json=""),
        kafka_bootstrap_servers="localhost:9092",
        topic="t",
        batch_source=FileSource(path="dummy.parquet", timestamp_field="event_timestamp"),
    ),
)

registry_proto = RegistryProto()
registry_proto.stream_feature_views.append(stream_fv.to_proto())

SqliteOnlineStore().plan(config, registry_proto)  # raises typeguard.TypeCheckError
```

Full traceback:

```
  File "repro.py", line 28, in <module>
    SqliteOnlineStore().plan(config, registry_proto)
  File ".../feast/infra/online_stores/sqlite.py", line 321, in plan
    infra_objects: List[InfraObject] = [
                                       ^
  File ".../feast/infra/online_stores/sqlite.py", line 326, in <listcomp>
    FeatureView.from_proto(view),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../feast/feature_view.py", line 523, in from_proto
    def from_proto(cls, feature_view_proto: FeatureViewProto) -> "FeatureView":
  File ".../typeguard/_functions.py", line 180, in check_argument_types_internal
    check_type_internal(value, annotation, memo)
  File ".../typeguard/_checkers.py", line 994, in check_type_internal
    raise TypeCheckError(f"is not an instance of {qualified_name(origin_type)}")
typeguard.TypeCheckError: argument "feature_view_proto" (feast.core.StreamFeatureView_pb2.StreamFeatureView) is not an instance of feast.core.FeatureView_pb2.FeatureView
```

### Specifications

- Version: reproduced on 0.62.0 (installed via `pip install feast==0.62.0`); confirmed the same code pattern is still present on `v0.65.0` (latest tag) as of this report.
- Platform: macOS, Python 3.11
- Subsystem: online store / `feast plan` / registry

## Possible Solution

Use the matching class for each list instead of `FeatureView.from_proto()` for both:

```python
views = [
    FeatureView.from_proto(view) for view in desired_registry_proto.feature_views
] + [
    StreamFeatureView.from_proto(view) for view in desired_registry_proto.stream_feature_views
]
```

Opened a PR with this fix: #6659

casaar97:

Fixes #6658.

`SqliteOnlineStore.plan()` applies `FeatureView.from_proto()` uniformly to both `desired_registry_proto.feature_views` and `desired_registry_proto.stream_feature_views`. `FeatureView` is `@typechecked`, and `from_proto`'s parameter is annotated as `FeatureViewProto`, so passing a `StreamFeatureViewProto` (from the `stream_feature_views` list) raises `typeguard.TypeCheckError` at runtime -- `plan()` (and therefore `feast plan`) fails outright on any repo with at least one `StreamFeatureView` registered.

This uses the matching class per list instead:

```python
views = [
    FeatureView.from_proto(view) for view in desired_registry_proto.feature_views
] + [
    StreamFeatureView.from_proto(view)
    for view in desired_registry_proto.stream_feature_views
]
```

`_table_id()` / `compute_table_id()` / `compute_versioned_name()` only touch `.name`, `.projection`, and `.current_version_number` (the latter two via `getattr` with defaults), so a `StreamFeatureView` instance works the same as a `FeatureView` instance there -- no other change needed.

We ran into this adding `plan()`/`InfraObject` support to a custom online store and initially copied this exact pattern from `sqlite.py` as the only in-tree reference implementation of `plan()`.

Fixes #6658

## Does this PR introduce a user-facing change?

```release-note
Fixed `feast plan` raising a typeguard.TypeCheckError for any repo with a registered StreamFeatureView.
```

## Test plan

- Added `sdk/python/tests/unit/infra/online_store/test_sqlite_plan.py` with two cases: a registry with only a `StreamFeatureView`, and one with a batch `FeatureView` and a `StreamFeatureView` together.
- Confirmed both fail with the exact traceback from #6658 on `master` before this change, and pass after.
- Ran the full `sdk/python/tests/unit/infra/online_store/` directory (206 tests) -- no regressions.
- `ruff check` and `ruff format --check` pass on the changed files.

### DIFF
--- a/sdk/python/feast/infra/online_stores/sqlite.py
+++ b/sdk/python/feast/infra/online_stores/sqlite.py
@@ -60,6 +60,7 @@ from feast.protos.feast.core.SqliteTable_pb2 import SqliteTable as SqliteTablePr
 from feast.protos.feast.types.EntityKey_pb2 import EntityKey as EntityKeyProto
 from feast.protos.feast.types.Value_pb2 import Value as ValueProto
 from feast.repo_config import FeastConfigBaseModel, RepoConfig
+from feast.stream_feature_view import StreamFeatureView
 from feast.type_map import feast_value_type_to_python_type
 from feast.types import FEAST_VECTOR_TYPES, PrimitiveFeastType
 from feast.utils import (
@@ -465,20 +466,23 @@ class SqliteOnlineStore(OnlineStore):
             config.online_store, "enable_openai_compatible_store", False
         )
 
+        # FeatureView.from_proto() is @typechecked and only accepts a
+        # FeatureViewProto, so it can't be applied to stream_feature_views
+        # (StreamFeatureViewProto) too -- each list needs its matching class.
+        views = [
+            FeatureView.from_proto(view)
+            for view in desired_registry_proto.feature_views
+        ] + [
+            StreamFeatureView.from_proto(view)
+            for view in desired_registry_proto.stream_feature_views
+        ]
         infra_objects: List[InfraObject] = [
             SqliteTable(
                 path=self._get_db_path(config),
-                name=_table_id(
-                    project,
-                    FeatureView.from_proto(view),
-                    versioning,
-                ),
+                name=_table_id(project, view, versioning),
                 include_value_num=include_value_num,
             )
-            for view in [
-                *desired_registry_proto.feature_views,
-                *desired_registry_proto.stream_feature_views,
-            ]
+            for view in views
         ]
 
         for lv_proto in desired_registry_proto.label_views:
--- a/sdk/python/tests/unit/infra/online_store/test_sqlite_plan.py
+++ b/sdk/python/tests/unit/infra/online_store/test_sqlite_plan.py
@@ -1,0 +1,88 @@
+from datetime import timedelta
+
+from feast.data_format import AvroFormat
+from feast.data_source import KafkaSource
+from feast.feature_view import FeatureView
+from feast.field import Field
+from feast.infra.offline_stores.dask import DaskOfflineStoreConfig
+from feast.infra.offline_stores.file_source import FileSource
+from feast.infra.online_stores.sqlite import SqliteOnlineStore, SqliteOnlineStoreConfig
+from feast.protos.feast.core.Registry_pb2 import Registry as RegistryProto
+from feast.repo_config import RepoConfig
+from feast.stream_feature_view import StreamFeatureView
+from feast.types import String
+
+
+def _repo_config() -> RepoConfig:
+    return RepoConfig(
+        registry="/tmp/unused_registry.db",
+        project="test_project",
+        provider="local",
+        online_store=SqliteOnlineStoreConfig(),
+        offline_store=DaskOfflineStoreConfig(),
+        entity_key_serialization_version=3,
+    )
+
+
+def _feature_view(name: str) -> FeatureView:
+    return FeatureView(
+        name=name,
+        entities=[],
+        schema=[Field(name="value", dtype=String)],
+        ttl=timedelta(days=1),
+        online=True,
+        source=FileSource(path="dummy.parquet", timestamp_field="event_timestamp"),
+    )
+
+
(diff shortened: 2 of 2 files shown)


casaar97:
The \`unit-test-go\` failure looks unrelated to this change (only Python files are touched here): \`open .../go/internal/test/feature_repo/data/registry.db: no such file or directory\` in \`TestNewFeatureStore\`. That same job passed on the base commit this branch is built from (104ad10), so it looks like environmental flakiness in that job rather than something introduced by this PR. Happy to look further if it's actually related -- let me know.

franciscojavierarceo (approved):
Using the matching class for each proto list fixes the typeguard failure while preserving the table-ID path, and the new tests cover stream-only and mixed registries. I found no blocking issue in the remote diff.

ntkathole (approved):

PR #6670, opened 2026-07-30
feat: Add optional OIDC token audience and issuer verification

### LINKED ISSUE #6669: OIDC token parser never verifies audience or issuer
## The observation

The feature server's OIDC token parser verifies the token signature (JWKS via discovery) and expiry, but audience verification is explicitly off and there is no issuer check (`sdk/python/feast/permissions/auth/oidc_token_parser.py`, `_decode_token`):

```python
audience="account",
options={
    "verify_aud": False,
    "verify_signature": True,
    "verify_exp": True,
},
```

The consequence is that any validly-signed, unexpired token from the configured IdP authenticates against the feature server, regardless of which resource it was minted for. In a multi-app tenant, a token issued to a completely different application passes authentication, and role matching in `RoleBasedPolicy` is the only remaining gate. Authorization ends up doing all of the work; authentication is only proving the token came from the right IdP, not that it was meant for Feast.

## Why the current default is reasonable

This looks deliberate, and changing the default would break real deployments. Identity providers commonly put claim values in tokens that differ from their discovery metadata. The clearest case is Microsoft Entra ID, which issues v1.0 tokens (`iss: https://sts.windows.net/<tenant-id>/`, `aud: api://<app-id-uri>`) that are routinely validated against a v2.0 discovery URL. That works today precisely because discovery is only used to source JWKS keys and no `aud`/`iss` comparison runs. #6631 added Entra support on top of that behavior, so this is proposing configurability, not calling the default a bug.

## Proposal

Two optional `OidcAuthConfig` fields, `audience` and `issuer`, both unset by default:

- When set, the corresponding token claim must match or the token is rejected at authentication.
- When unset, the decode options are identical to today's, so nobody is affected without opting in.
- Operators state the expected values explicitly rather than the code inferring them from the discovery document, which is what keeps the v1-token-against-v2-discovery setup working.

The accompanying PR includes a test that pins the v1/v2 Entra setup so a future refactor cannot silently regress it.

One scope note: the feast-operator copies a whitelist of OIDC secret keys into the generated `feature_store.yaml` (`OidcOptionalSecretProperties`), so exposing these fields through the operator CRD would be a small follow-up if there is interest. The PR keeps to the Python SDK and docs.

larrysingleton007:

Adds optional audience and issuer verification to the OIDC token parser. Today the parser verifies signature and expiry only, so any validly-signed, unexpired token from the configured IdP authenticates regardless of which resource it was minted for; RBAC role matching is the only remaining gate.

Two new `OidcAuthConfig` fields, `audience` and `issuer`, both unset by default:

```yaml
auth:
  type: oidc
  client_id: _CLIENT_ID_
  auth_discovery_url: https://login.example.com/.well-known/openid-configuration
  audience: api://feast-feature-server
  issuer: https://login.example.com/realms/master
```

When set, `jwt.decode` runs with `verify_aud` / `verify_iss` enabled against the configured values, and a mismatched or missing claim rejects the token at authentication. When unset, the decode options are identical to before, so existing deployments see no behavior change; all pre-existing tests are untouched and passing.

Opt-in rather than strict-by-default because IdPs commonly mint tokens whose claims differ from their discovery metadata. Entra ID issues v1.0 tokens (`iss` under `sts.windows.net`, `api://` audience) that are routinely validated against a v2.0 discovery URL, which works because discovery is only used to source JWKS keys. `test_oidc_default_supports_v1_tokens_against_v2_discovery` pins that setup so a future refactor cannot silently regress it. Operators state expected values explicitly rather than the code inferring them from the discovery document, for the same reason.

Testing: 11 new unit tests. Config-to-decode fidelity (the verified decode enables exactly the checks the config provides), end-to-end RS256-signed tokens through the real `jwt.decode` (match accepted, mismatch rejected, missing claim rejected, default permissive), and the v1/v2 pin. `tests/unit/permissions`: 307 passing; ruff and mypy clean on the changed files.

Docs: the OIDC section of `authz_manager.md` documents the new options and warns that expected values must match the token's claims, not the discovery document.

Not included: the feast-operator's OIDC secret-key whitelist (`OidcOptionalSecretProperties`) does not yet pass `audience`/`issuer` through to the generated `feature_store.yaml`. Happy to follow up separately if there is interest in exposing these via the operator CRD.

Fixes #6669

### DIFF
--- a/docs/getting-started/components/authz_manager.md
+++ b/docs/getting-started/components/authz_manager.md
@@ -45,7 +45,7 @@ The server, in turn, uses the same OIDC server to validate the token and extract
 Some assumptions are made in the OIDC server configuration:
 * The OIDC token refers to a client with roles matching the RBAC roles of the configured `Permission`s (*)
 * The roles are exposed in the access token under `resource_access.<client_id>.roles` (Keycloak) or in the top-level `roles` claim (Entra ID app roles). Roles found in both are merged.
-* The JWT token is expected to have a verified signature and not be expired. The Feast OIDC token parser logic validates for `verify_signature` and `verify_exp` so make sure that the given OIDC provider is configured to meet these requirements.
+* The JWT token is expected to have a verified signature and not be expired. The Feast OIDC token parser logic validates for `verify_signature` and `verify_exp` so make sure that the given OIDC provider is configured to meet these requirements. The token's audience and issuer claims are **not** verified by default; both checks can be enabled with the `audience` and `issuer` options (see [Server-Side Configuration](#server-side-configuration)).
 * The username is read from the first of `preferred_username`, `upn`, `azp`, `appid`, `sub` present in the token. Entra ID client-credentials (app-only) tokens carry no user claim, so they authenticate as the calling application.
 * For `GroupBasedPolicy` support, the `groups` claim should be present in the access token (requires a "Group Membership" protocol mapper in Keycloak).
 * **Entra ID limitation**: Group claims use object IDs (GUIDs) instead of names, and are omitted entirely when a user exceeds the group overage threshold. GroupBasedPolicy must reference GUIDs and cannot be used for principals with large group memberships.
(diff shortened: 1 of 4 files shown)


ntkathole (approved):
Let's have it in operator as well

larrysingleton007:
Thanks for the quick review. Operator follow-up PR coming right up: adding audience and issuer to the OIDC secret property whitelist so they flow into the generated feature_store.yaml auth section, plus tests.

larrysingleton007:
Operator follow-up is tracked in #6676 and the change is ready on my side (whitelist additions for audience and issuer in the OIDC Secret, tests, and docs; verify_ssl and ca_cert_path stay CR fields since they already flow through the CRD). One question on sequencing: happy for this one to merge now with the operator PR following right behind, or would you rather hold it and land the two together? Either works for me.

ntkathole:
will merge this, @aniketpalu  fyi

PR #6675, opened 2026-07-30
fix: Reset the global security manager after the permissions test fixture

### LINKED ISSUE #6674: Permissions test fixture leaks the global security manager, 401-failing later in-process tests
## Symptom

Running the permissions unit tests and the feature server unit tests in one pytest process fails 26 feature server tests on current master, with every failing request returning 401 Unauthorized. Each suite passes on its own.

Minimal reproduction (current master, no other changes):

```bash
cd sdk/python
python -m pytest tests/unit/permissions/test_security_manager.py "tests/unit/test_feature_server.py::test_get_online_features_async_supported"
# 2 failed: assert 0 == 1 on the store mock call count; the request log shows
# POST /get-online-features "HTTP/1.1 401 Unauthorized"
```

Broader form:

```bash
python -m pytest tests/unit/permissions/ tests/unit/test_feature_server.py
# 26 failed, 306 passed
```

## Cause

The `security_manager` fixture in `tests/unit/permissions/conftest.py` calls `set_security_manager(sm)`, which populates the module-level singleton in `feast/permissions/security_manager.py`, and returns without ever resetting it. Once any test using that fixture has run, every later test in the same process that serves a request is evaluated against the leaked mock security manager (a mock registry with a fixed permission set) and gets rejected.

CI doesn't catch this because pytest-xdist distributes the modules across workers, so the poisoned ordering rarely occurs there. Anyone running a scoped selection locally in a single process hits it, which makes for confusing local-only failures.

## Fix

Convert the fixture to a yield fixture and reset via `no_security_manager()` on teardown. That's the same cleanup the integration auth tests already do (`tests/integration/permissions/auth/server/test_auth_registry_server.py`). PR attached.

larrysingleton007:

The `security_manager` fixture in `tests/unit/permissions/conftest.py` sets the module-level security manager singleton and never resets it, so any later test in the same pytest process that serves a request runs against the leaked mock manager and gets 401 Unauthorized. On master, running `tests/unit/permissions/` together with `tests/unit/test_feature_server.py` in one process fails 26 feature server tests; CI masks it because xdist distributes the modules across workers, but scoped local runs hit it.

This converts the fixture to yield and resets via `no_security_manager()` on teardown, the same cleanup the integration auth tests already use. Test-only change.

After the fix, the combined run passes (332 passed), and the permissions suite alone is unchanged (307 passed).

Fixes #6674

### DIFF
--- a/sdk/python/tests/unit/permissions/conftest.py
+++ b/sdk/python/tests/unit/permissions/conftest.py
@@ -1,3 +1,4 @@
+from typing import Iterator
 from unittest.mock import MagicMock, Mock
 
 import pytest
@@ -11,6 +12,7 @@ from feast.permissions.permission import AuthzedAction, Permission
 from feast.permissions.policy import RoleBasedPolicy
 from feast.permissions.security_manager import (
     SecurityManager,
+    no_security_manager,
     set_security_manager,
 )
 from feast.permissions.user import User
@@ -63,7 +65,7 @@ def users() -> list[User]:
 
 
 @pytest.fixture
-def security_manager() -> SecurityManager:
+def security_manager() -> Iterator[SecurityManager]:
     permissions = []
     permissions.append(
         Permission(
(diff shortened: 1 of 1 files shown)


larrysingleton007:
@ntkathole @franciscojavierarceo quick test-only one when you have a minute: the permissions security_manager fixture leaves the global security manager set, so any single-process run combining the permissions suite with the feature server tests fails 26 tests with 401s. Reproduces on clean master; CI never sees it because xdist splits the modules across workers. Fix is a yield fixture with the same no_security_manager() cleanup the integration auth tests already use. CI green.

ntkathole (approved):

PR #6677, opened 2026-07-30
feat: Pass optional OIDC audience and issuer through the operator

### LINKED ISSUE #6676: Expose optional OIDC audience and issuer verification through the feast-operator
Follow-up to #6669. PR #6670 added optional `audience` and `issuer` fields to `OidcAuthConfig` for opt-in token claim verification, and review there asked for operator support as well.

The operator copies a whitelist of OIDC Secret keys (`OidcOptionalSecretProperties`) into the generated `feature_store.yaml`, so the two new fields need to be added there before operator deployments can use them. The plan is to extend that whitelist so `audience` and `issuer` set in the referenced OIDC Secret flow into the auth section the same way `client_id` and `client_secret` already do, plus a docs update in the operator security guide. When the keys are absent nothing changes.

An alternative would be first-class CRD fields, but the existing `issuerUrl` field on `OidcAuthz` already means the IdP base URL for discovery rather than the token `iss` claim, so Secret keys avoid a confusing name collision. Happy to go the CRD route instead if that's preferred.

PR to follow.

larrysingleton007:

Follow-up to #6670, requested in review there: expose the new optional OIDC `audience` and `issuer` verification fields through the feast-operator.

Adds `audience` and `issuer` to `OidcOptionalSecretProperties`, so operators can set them in the referenced OIDC Secret and they flow into the generated `feature_store.yaml` auth section the same way `client_id`, `client_secret`, `username`, and `password` already do. When the keys are absent, nothing changes.

Also documents the two optional Secret keys in the operator security guide, with a pointer to the main OIDC authorization page for the caveat about token claims versus discovery metadata.

Scope note: `verify_ssl` and `ca_cert_path` are deliberately not added to the whitelist. They're already exposed as first-class CR fields (`verifySSL`, `caCertConfigMap`) that feed the same generated auth section, so whitelisting them would create two sources for the same keys. The Secret bag carries IdP-coupled values (`auth_discovery_url`, `client_id`, ...), which is where `audience` and `issuer` belong; a first-class CR field for `issuer` was considered and rejected because the existing `issuerUrl` field already means the discovery base URL, not the token `iss` claim.

Testing: extended the repo config suite to include both keys in the mocked Secret and assert they arrive in `OidcParameters` with their values intact, plus the reconcile-level envtest fixtures so the keys go through real Secret extraction and YAML serialization. `go build ./...` and `go test ./internal/controller/...` pass.

A pre-submit review pass surfaced two things worth calling out:

- Secret values are YAML-parsed on extraction, so an all-digits `audience` (a numeric IdP application ID) would arrive as an int, render unquoted, and fail the SDK's `Optional[str]` validation at server startup. The copy now coerces the two claim keys to strings, with a regression test.
- If a referenced Secret already contains keys named `audience` or `issuer` (previously ignored), they take effect after this change. That and two other operational caveats (the IdP must mint matching claims for Feast's own client token flows, and Secret edits apply on the next reconcile rather than immediately) are documented in the security guide.

Deliberately left out: `ui_client_id` is similarly absent from the whitelist and consumed by the UI server; left for a separate change if wanted, to keep this one to the review request.

Fixes #6676

Related to #6669 (main change in #6670).

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1172,7 +1172,7 @@
         "filename": "infra/feast-operator/internal/controller/services/repo_config.go",
         "hashed_secret": "e2fb052132fd6a07a56af2013e0b62a1f510572c",
         "is_verified": false,
-        "line_number": 224
+        "line_number": 235
       }
     ],
     "infra/feast-operator/internal/controller/services/services.go": [
@@ -1564,5 +1564,5 @@
       }
     ]
   },
-  "generated_at": "2026-07-30T09:40:48Z"
+  "generated_at": "2026-07-30T16:22:49Z"
 }
--- a/docs/how-to-guides/feast-operator/05-security.md
+++ b/docs/how-to-guides/feast-operator/05-security.md
@@ -62,7 +62,11 @@ stringData:
   client_secret: <your-client-secret>
   username: <service-account-username>     # used for client-credentials flow
   password: <service-account-password>
+  audience: <expected-aud-claim>           # optional: reject tokens whose aud claim differs
+  issuer: <expected-iss-claim>             # optional: reject tokens whose iss claim differs
 ```
+
+The optional `audience` and `issuer` keys enable strict claim verification on the feature server; when omitted, only signature and expiry are verified. Set them to the values your IdP puts in the token itself, which are not always the ones in the discovery document (see [OIDC Authorization](../../getting-started/components/authz_manager.md#oidc-authorization)).
 
 Reference the Secret from the CR:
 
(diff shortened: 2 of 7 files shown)


larrysingleton007:
@ntkathole this is the operator follow-up you asked for on #6670, ready for review. audience and issuer flow from the referenced OIDC Secret into the generated feature_store.yaml like the existing keys. A pre-submit review caught one real bug worth knowing about: an all-digits audience would arrive as a YAML int and fail the SDK's string validation at server startup, so the copy now coerces the two claim keys to strings, with a regression test. Also documented the operational caveats (pre-existing Secret keys take effect on upgrade, the IdP has to mint matching claims for feast's own client flows, Secret edits apply on next reconcile). CI is green.

ntkathole (approved):

PR #6683, opened 2026-07-31
fix: Reuse the OIDC JWKS client across requests

### LINKED ISSUE #6682: OIDC auth builds a new JWKS client per request, adding a network fetch to every authenticated call
## The observation

`OidcTokenParser._decode_token` constructs a fresh `PyJWKClient` on every call (`sdk/python/feast/permissions/auth/oidc_token_parser.py`):

```python
def _decode_token(self, access_token: str) -> dict:
    ...
    jwks_client = PyJWKClient(
        self.oidc_discovery_service.get_jwks_url(), headers=..., ssl_context=ssl_ctx,
    )
    signing_key = jwks_client.get_signing_key_from_jwt(access_token)
```

`PyJWKClient` caches the fetched JWK set, but the cache lives on the instance, so a per-request instance starts cold every time and `get_signing_key_from_jwt` performs a full HTTPS round trip to the IdP's JWKS endpoint, TLS handshake included, on every authenticated request.

Reproduced on master with a mock JWKS endpoint and a hit counter: 25 authenticated decodes produce 25 JWKS fetches.

## Measured impact

On a production-shaped deployment (containerized feature server, Entra ID OIDC, Redis online store):

- The JWKS fetch alone measures 124-196 ms against `login.microsoftonline.com`.
- Server-side latency for single-entity online reads sat at p50 247 ms / p95 314 ms while the Redis engine idled at 0.5% CPU; the unauthenticated `/health` floor on the same connection was ~50 ms. Nearly the entire request budget goes to re-fetching a document that hasn't changed.
- Because the fetch is blocking I/O in the request path, concurrent requests serialize behind it: 3 parallel readers measured p50 900 ms (an exact 3x of single-request latency), and 10 parallel readers produced 502s from a perfectly healthy process. A deployment could easily misdiagnose this as "the feature server cannot handle load" when it is one cold cache constructor.

Both the latency inflation and the concurrency collapse are the same defect.

## Fix

Build the `PyJWKClient` lazily once per parser. The parser is constructed once per process (`init_auth_manager`), so PyJWT's built-in JWK-set cache (default `cache_jwk_set=True`, `lifespan=300`) then does its job. Key rotation stays safe: `PyJWKClient.get_signing_key` refreshes the set and retries once whenever it sees an unknown `kid` (verified in PyJWT 2.13, the currently resolved version - feast pins no upper bound).

After the change the same 25-decode benchmark performs exactly 1 JWKS fetch. To be precise about the concurrency half: this removes the accidental serialization behind per-request network I/O; it does not add parallelism to the server.

PR to follow shortly.

larrysingleton007:

`OidcTokenParser._decode_token` builds a fresh `PyJWKClient` on every call. The JWK-set cache lives on the instance, so every authenticated request starts cold and pays a full HTTPS fetch of the JWKS document (TLS handshake included) before signature verification. On a production-shaped deployment that fetch measured 124-196 ms against the IdP, dominating single-entity online reads (server-side p50 247 ms while the online store idled), and because it's blocking I/O on the event loop, concurrent requests serialize behind it: 3 parallel readers measured an exact 3x of single-request latency, and 10 produced load-balancer 502s from a healthy process.

The fix builds the client lazily once per parser (the parser is a process singleton created by `init_auth_manager`), letting PyJWT's JWK-set cache work. Reproducible benchmark against a counting mock JWKS endpoint with real RS256 tokens: 25 authenticated decodes performed 25 JWKS fetches before, 1 after.

Details worth reviewer attention:

- `pyjwt` now has a `>=2.13.0` floor: the reuse makes PyJWT's cache semantics load-bearing, and only 2.13+ keeps the cache through transient fetch failures (2.5-2.12 wipe it on one failed fetch; pre-2.5 has no cache at all, silently reverting the fix). The lockfiles already resolve 2.13.0, so no lock changes.
- `lifespan=300` and `timeout=10` are passed explicitly so upgrades can't silently change behavior. The lifespan bounds two staleness windows that were zero with per-request clients and are documented in the code: a key the IdP removed keeps validating, and a rotation that reuses a `kid` keeps failing, each for at most 300 s. Rotations that mint a new `kid` (the common case) recover immediately because `PyJWKClient.get_signing_key` refreshes and retries on a cache miss.
- Tests pin the behaviors a refactor could silently break: client built with the right JWKS URL and an SSL context matching `verify_ssl` (mutation testing showed the SSL branch previously had zero coverage), laziness (no construction before the first request, keeping the blocking discovery fetch out of server boot), per-parser scoping, reuse across requests, and recovery after a failed first construction (the parser is a process singleton, so a wedged client would require a restart).

Known accepted residuals, for transparency: on the threaded Arrow Flight server a cold-start burst can still build the client more than once (benign, self-heals, and PyJWT's own cold-fetch path has the same property); PyJWT's JWK-set cache is unlocked so simultaneous expiry can trigger redundant refetches (bounded by concurrency). Both are strictly better than the guaranteed fetch-per-request they replace. The client-side token fetch (`client_secret` flow) has the same per-request pattern and is a natural follow-up.

To be precise about the concurrency claim: this removes the accidental serialization behind per-request network I/O; it does not add parallelism to the server.

Testing: 311 permissions unit tests pass (including the new lifecycle tests); full unit suite green (2488 passed); ruff and mypy clean.

Fixes #6682

### DIFF
--- a/sdk/python/feast/permissions/auth/oidc_token_parser.py
+++ b/sdk/python/feast/permissions/auth/oidc_token_parser.py
@@ -41,6 +41,34 @@ class OidcTokenParser(TokenParser):
             ca_cert_path=self._auth_config.ca_cert_path,
         )
         self._k8s_auth_api = None
+        self._jwks_client: Optional[PyJWKClient] = None  # Initialize it lazily.
+
+    def _get_jwks_client(self) -> PyJWKClient:
+        """Lazily build and cache a parser-lifetime ``PyJWKClient``.
+
+        A per-request client starts with a cold JWK-set cache, so every
+        authenticated request pays a full HTTPS fetch of the JWKS document
+        (TLS handshake included) before signature verification, and
+        concurrent requests serialize behind that blocking I/O. Reusing one
+        client lets PyJWT's built-in JWK-set cache do its job; on IdP key
+        rotation the client refetches automatically when it sees an unknown
+        ``kid`` (``PyJWKClient.get_signing_key`` refreshes and retries once).
+        """
+        if self._jwks_client is None:
+            ssl_ctx = ssl.create_default_context()
+            if not self._auth_config.verify_ssl:
+                ssl_ctx.check_hostname = False
+                ssl_ctx.verify_mode = ssl.CERT_NONE
+            elif self._auth_config.ca_cert_path and os.path.exists(
+                self._auth_config.ca_cert_path
+            ):
+                ssl_ctx.load_verify_locations(self._auth_config.ca_cert_path)
+            self._jwks_client = PyJWKClient(
+                self.oidc_discovery_service.get_jwks_url(),
+                headers={"User-agent": "custom-user-agent"},
+                ssl_context=ssl_ctx,
+            )
+        return self._jwks_client
 
     async def _validate_token(self, access_token: str):
         """
@@ -125,21 +153,7 @@ class OidcTokenParser(TokenParser):
         metadata (e.g. Entra ID v1.0 tokens validated against a v2.0
         discovery document).
         """
-        optional_custom_headers = {"User-agent": "custom-user-agent"}
-        ssl_ctx = ssl.create_default_context()
-        if not self._auth_config.verify_ssl:
-            ssl_ctx.check_hostname = False
-            ssl_ctx.verify_mode = ssl.CERT_NONE
-        elif self._auth_config.ca_cert_path and os.path.exists(
-            self._auth_config.ca_cert_path
-        ):
-            ssl_ctx.load_verify_locations(self._auth_config.ca_cert_path)
-        jwks_client = PyJWKClient(
-            self.oidc_discovery_service.get_jwks_url(),
-            headers=optional_custom_headers,
-            ssl_context=ssl_ctx,
-        )
-        signing_key = jwks_client.get_signing_key_from_jwt(access_token)
+        signing_key = self._get_jwks_client().get_signing_key_from_jwt(access_token)
         expected_audience = self._auth_config.audience
         expected_issuer = self._auth_config.issuer
         return jwt.decode(
--- a/sdk/python/feast/permissions/auth/oidc_token_parser.py
+++ b/sdk/python/feast/permissions/auth/oidc_token_parser.py
@@ -49,12 +49,12 @@ class OidcTokenParser(TokenParser):
         A per-request client starts with a cold JWK-set cache, forcing a
         full HTTPS fetch of the JWKS document on every authenticated
         request. Reusing one client lets PyJWT cache the JWK set for
-        ``lifespan`` seconds, which also bounds two staleness windows: a
-        key the IdP has removed keeps validating, and a rotation that
-        reuses an existing ``kid`` keeps failing, for at most that long.
-        Rotations that introduce a new ``kid`` recover immediately
-        (``PyJWKClient.get_signing_key`` refreshes and retries once on a
-        cache miss).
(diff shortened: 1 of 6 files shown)


larrysingleton007:
Adding a deployment-shaped before/after to complement the mock benchmark in the description.

Setup: this branch built from source (Dockerfile.dev at 602feb009) versus the pinned 0.65.0 release image, both serving the same feature repository against the same Redis online store on the same machine, authenticating real client-credentials tokens from a live Entra ID tenant. The only variable is this change.

| | 0.65.0 (per-request client) | this branch |
|---|---|---|
| sequential p50 | 250.7 ms | 2.3 ms |
| sequential p95 | 347.7 ms | 3.4 ms |
| 10 concurrent readers | fails (502s behind an ALB) | all 200, p50 16.7 ms, p95 25.6 ms |

The 250 ms baseline matches what we measured server-side on a real deployment (ALB TargetResponseTime p50 247 ms), where the JWKS endpoint fetch cost 124-196 ms per request.

One expected behavior worth noting: a cold-started process pays its single JWKS fetch on the first request, and a burst arriving before that completes queues briefly behind it; after that it is steady state, with a refresh at most once per lifespan (300 s). We saw no failures at 10 concurrent readers on a warm process.

sdk/python/feast/permissions/auth/oidc_token_parser.py
```diff
@@ -41,6 +41,41 @@ def __init__(self, auth_config: OidcAuthConfig):
...
+            ):
+                ssl_ctx.load_verify_locations(self._auth_config.ca_cert_path)
+            self._jwks_client = PyJWKClient(
+                self.oidc_discovery_service.get_jwks_url(),
+                headers={"User-agent": "custom-user-agent"},
+                ssl_context=ssl_ctx,
+                # Explicit so upgrades cannot silently change the staleness
+                # window documented above, and so a hung IdP bounds how long
+                # a fetch can block the serving path.
+                lifespan=300,
```
ntkathole: can we make lifespan and timeout configurable via OidcAuthConfig ?

larrysingleton007: Good call, done in c62e9467e. Both are now OidcAuthConfig options: jwks_cache_lifespan_seconds (default 300) and jwks_request_timeout_seconds (default 10), with docs in the OIDC authorization page.

One thing worth knowing about the lifespan: it also bounds how long a key the IdP has revoked keeps validating tokens, so making it tunable means operators whose provider rotates or revokes aggressively can tighten it, at the cost of proportionally more JWKS fetches. Rotations that introduce a new kid are picked up immediately either way, since an unknown kid forces a refetch.

Both fields reject non-positive values at config load. A zero or negative lifespan would expire the cache immediately and silently put a JWKS fetch back on every request, which would undo the PR without any visible error.

ntkathole:
@larrysingleton007 looks good, want to handle operator support in same PR or different one?

larrysingleton007:
@ntkathole 
Separate PR, if that works for you.

These two are tuning knobs rather than IdP-coupled values, so they don't really fit the Secret whitelist that audience and issuer went through in #6677. They belong next to verifySSL as CR fields on OidcAuthz, which means a CRD change and regenerated manifests, so a different review surface than this Python diff.

There's also a sequencing argument: operator support is only useful once this lands in a release the operator can point at, same as #6670 and #6677 last week.

I'll open an issue and follow with the PR. Happy to fold it in here instead if you'd rather have it as one changeset.

larrysingleton007:
Filed as #6686 with the CR-field plan spelled out.

ntkathole (approved):

PR #6687, opened 2026-07-31
fix: Reuse IdP-issued client tokens until near expiry

### LINKED ISSUE #6684: Client-side OIDC auth pays a discovery fetch and a token fetch per outbound request
## The observation

This is the client-side sibling of #6682. On the `client_secret` (client credentials / ROPC) branch, every outbound RPC builds a fresh auth manager and discards the token it fetched:

- `get_auth_token` constructs a new `AuthenticationClientManagerFactory`, manager, and `OIDCDiscoveryService` per call (`sdk/python/feast/permissions/client/client_auth_token.py`).
- All three interceptors invoke it per request: the gRPC interceptor on every intercepted call, the Arrow Flight middleware in `sending_headers`, and the HTTP wrapper even on its cached-session path.

So each outbound request pays a discovery GET plus a token POST against the IdP, for a token that is typically valid for an hour. Measured with counting mocks on current master: 5 outbound calls perform 5 discovery GETs and 5 token POSTs.

Beyond the latency, a batch loop (e.g. many `get_online_features` calls against a remote server, or materialization against a remote registry) multiplies IdP load by request count, which can trip IdP rate limits.

## Fix

Cache issued tokens keyed by the token-request identity (discovery URL, client id/secret, username, password) and reuse each until shortly before its expiry, read from the token's own `exp` claim with the token response's `expires_in` as a fallback. A token whose expiry cannot be determined is not cached, preserving today's per-call behavior for opaque tokens. The static-token, env-var, and mounted service-account sources are unaffected (they were already cheap).

After the change the same 5-call measurement performs 1 discovery GET and 1 token POST.

PR to follow shortly.

larrysingleton007:

This is the client-side sibling of #6683. On the `client_secret` (client credentials / ROPC) branch, every outbound RPC builds a fresh auth manager and throws away the token it fetched:

- `get_auth_token` constructs a new factory, manager, and `OIDCDiscoveryService` per call ([client_auth_token.py](https://github.com/feast-dev/feast/blob/master/sdk/python/feast/permissions/client/client_auth_token.py)).
- All three interceptors call it per request: the gRPC interceptor on every intercepted call, the Arrow Flight middleware in `sending_headers`, and the HTTP wrapper on every session reuse.

So each request pays a discovery GET plus a token POST against the IdP for a token that is typically valid for an hour. Measured with counting mocks on master: 5 outbound calls produce 5 discovery GETs and 5 token POSTs. Beyond the added latency, a batch loop multiplies IdP load by request count, which can trip provider rate limits.

The fix caches issued tokens keyed by the token-request identity (discovery URL, client id and secret, username, password) and reuses each until shortly before expiry, taken from the token's own `exp` claim with the token response's `expires_in` as a fallback. After the change the same measurement performs 1 discovery GET and 1 token POST.

The cache is module level because the interceptors construct a fresh manager per RPC, so instance state cannot survive between calls. Only the `client_secret` branch is affected; static tokens, `token_env_var`, and mounted service account tokens were already cheap and are untouched.

A few details worth reviewer attention:

- `token_refresh_margin_seconds` (default 30) is exposed on `OidcClientAuthConfig` rather than hardcoded, following the same request you made on #6683. It rejects non-positive values, since a zero or negative margin would allow reuse right up to or past expiry. It sits on the client config rather than the shared `OidcAuthConfig` because only clients fetch tokens from the IdP.
- A token whose expiry cannot be determined (opaque, and no `expires_in`) is deliberately not cached, preserving today's per-call behavior rather than guessing a lifetime.
- Errors from the IdP propagate before any cache write, so a failed fetch never poisons the cache.
- Concurrent misses for the same identity may fetch in parallel and the last write wins. That costs at most a redundant fetch and never returns a wrong-identity token, which is still strictly better than the guaranteed fetch per request it replaces.

One interaction worth flagging: `HttpSessionManager.get_session` re-calls `get_auth_token` on every session cache hit, with a comment from #5895 explaining it does so "in case it expired". After this change that call returns a cached token rather than a freshly fetched one. The intent still holds, because the cache never returns a token with less than the margin remaining, but the mechanism changes and it seemed better to say so than to let a reviewer find it.

Testing: 7 new unit tests covering reuse until expiry, refusal to cache inside the refresh margin, the `expires_in` fallback for opaque tokens, no caching when expiry is unknowable, cache keying across distinct configs, the configurable margin changing reuse behavior, and rejection of non-positive margins. An autouse fixture resets the module cache around every test so no test inherits or leaks a cached token. 315 permissions tests pass; ruff and mypy are clean.

Fixes #6684

### DIFF
--- a/sdk/python/feast/permissions/client/oidc_authentication_client_manager.py
+++ b/sdk/python/feast/permissions/client/oidc_authentication_client_manager.py
@@ -1,6 +1,8 @@
 import logging
 import os
-from typing import Optional
+import threading
+import time
+from typing import Dict, Optional, Tuple
 
 import jwt
 import requests
@@ -12,6 +14,17 @@ from feast.permissions.oidc_service import OIDCDiscoveryService
 logger = logging.getLogger(__name__)
 
 SA_TOKEN_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+
+# IdP-issued tokens keyed by the token-request identity, stored with their
+# true expiry. The auth interceptors build a fresh manager for every outbound
+# RPC, so instance state would not survive between calls; without this
+# cache every RPC pays a discovery GET plus a token POST against the IdP.
+# The refresh margin is applied on read, not baked into the stored value, so
+# configs sharing IdP credentials but setting different margins can share
+# tokens while each still honors its own margin.
+# Concurrent misses may fetch in parallel (benign: last write wins).
+_token_cache: Dict[Tuple, Tuple[str, float]] = {}
+_token_cache_lock = threading.Lock()
 
 
 class OidcAuthClientManager(AuthenticationClientManager):
@@ -67,7 +80,96 @@ class OidcAuthClientManager(AuthenticationClientManager):
         return None
 
     def _fetch_token_from_idp(self) -> str:
-        """Obtain an access token via client_credentials or ROPG flow."""
+        """Return a cached IdP token, or obtain a fresh one.
+
+        The cache stores each token's true expiry (its ``exp`` claim, falling
+        back to the token response's ``expires_in``), and this config's
+        ``token_refresh_margin_seconds`` is applied when reading. Keeping the
+        margin out of the stored value lets configs that share IdP credentials
+        but set different margins share tokens while each honors its own
+        margin. A token whose expiry is unknowable is not cached, preserving
+        the previous per-call behavior for opaque tokens.
+        """
+        cache_key = self._cache_key()
+        with _token_cache_lock:
+            cached = _token_cache.get(cache_key)
+        if cached is not None:
+            cached_token, cached_expiry = cached
+            margin = self.auth_config.token_refresh_margin_seconds
+            if time.time() < cached_expiry - margin:
+                return cached_token
+
+        access_token, expires_in = self._request_token_from_idp()
+
+        expiry = self._token_expiry(access_token, expires_in)
+        if expiry is not None:
+            now = time.time()
+            with _token_cache_lock:
+                # Prune on miss. Entries are keyed by credential identity, so
+                # the cache is bounded by the number of distinct configs, but
+                # a long-lived process that rotates credentials would otherwise
+                # keep every retired identity forever.
+                for key in [k for k, (_, exp) in _token_cache.items() if exp <= now]:
+                    del _token_cache[key]
+                _token_cache[cache_key] = (access_token, expiry)
+        return access_token
+
+    def _cache_key(self) -> Tuple:
+        """Identity of the token request: same credentials, same token."""
+        return (
+            self.auth_config.auth_discovery_url,
+            self.auth_config.client_id,
+            self.auth_config.client_secret,
+            self.auth_config.username,
+            self.auth_config.password,
+        )
+
+    def invalidate_token(self) -> bool:
+        """Drop this config's cached token so the next call refetches.
+
+        Returns whether an entry was actually removed.
+
+        Reuse means a token the IdP revokes mid-life keeps being presented
+        until its own expiry, where fetching per call self-corrected. Callers
+        that can observe an authentication failure should invalidate on it, so
+        the staleness costs one rejected request rather than the remaining
+        lifetime of the token.
+        """
+        with _token_cache_lock:
+            return _token_cache.pop(self._cache_key(), None) is not None
+
+    @staticmethod
+    def _token_expiry(
+        access_token: str, expires_in: Optional[float]
+    ) -> Optional[float]:
+        """Epoch expiry of *access_token*, or ``None`` when it is unknowable.
+
+        Prefers the token's own ``exp`` claim (authoritative); falls back to
+        the token endpoint's ``expires_in``.
+
+        The refresh margin is deliberately not subtracted here. Storing one
+        caller's deadline would let another config with a wider margin reuse
+        the token past its own safety window, since the margin is not part of
+        the cache key.
+        """
+        exp: Optional[float] = None
+        try:
+            claims = jwt.decode(access_token, options={"verify_signature": False})
+            claim = claims.get("exp")
+            if isinstance(claim, (int, float)):
+                exp = float(claim)
+        except jwt.exceptions.DecodeError:
+            pass
+        if exp is None and isinstance(expires_in, (int, float)):
+            exp = time.time() + float(expires_in)
+        return exp
+
+    def _request_token_from_idp(self) -> Tuple[str, Optional[float]]:
+        """Obtain an access token via client_credentials or ROPG flow.
+
+        Returns the token and the token response's ``expires_in`` (seconds),
+        when the IdP provides one.
+        """
         if self.auth_config.auth_discovery_url is None:
             raise ValueError(
                 "auth_discovery_url is required for IDP token fetch "
@@ -106,13 +208,17 @@ class OidcAuthClientManager(AuthenticationClientManager):
         )
 
         if token_response.status_code == 200:
-            access_token = token_response.json()["access_token"]
+            response_body = token_response.json()
+            access_token = response_body["access_token"]
             if not access_token:
                 logger.debug(
                     f"access_token is empty for the client_id=${self.auth_config.client_id}"
                 )
                 raise RuntimeError("access token is empty")
-            return access_token
+            expires_in = response_body.get("expires_in")
+            if not isinstance(expires_in, (int, float)):
+                expires_in = None
+            return access_token, expires_in
         else:
             raise RuntimeError(
                 f"""Failed to obtain oidc access token:url=[{token_endpoint}] {token_response.status_code} - {token_response.text}"""
--- a/sdk/python/feast/permissions/client/oidc_authentication_client_manager.py
+++ b/sdk/python/feast/permissions/client/oidc_authentication_client_manager.py
@@ -23,10 +23,6 @@ SA_TOKEN_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/token"
 _token_cache: Dict[Tuple, Tuple[str, float]] = {}
 _token_cache_lock = threading.Lock()
 
-# Stop reusing a token this many seconds before its expiry, so a reused
-# token cannot expire between header injection and server-side validation.
-_TOKEN_REFRESH_MARGIN_SECONDS = 30
-
 
 class OidcAuthClientManager(AuthenticationClientManager):
     def __init__(self, auth_config: OidcClientAuthConfig):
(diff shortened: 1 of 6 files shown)


franciscojavierarceo (changes requested):
The cache key omits `token_refresh_margin_seconds`, but the stored deadline is calculated using that value. If two configs share the same IdP/client credentials and use different margins—the new option is explicitly configurable—the stricter config can reuse a token until the looser config's deadline. That violates the requested safety margin. We should include the margin in the key or store the token's true expiry and apply each caller's margin when reading; add a regression that uses both configs without manually clearing the cache.

larrysingleton007:
@franciscojavierarceo 
Fixed by storing the token's true expiry and applying each caller's margin on read, the second option you suggested. The cache key is unchanged, so configs sharing credentials still share tokens, but a wider margin can no longer ride a narrower one's deadline. `_token_refresh_deadline` became `_token_expiry` and no longer takes a margin.

One behavior change worth flagging: a token already inside the margin now gets cached instead of skipped. Every read still rejects it for that caller so nothing observable changes, and a config with a smaller margin can legitimately use it.

The regression exercises both configs against the shared entry without clearing the cache, and also checks the reverse direction stays cheap, with the narrow-margin config reusing what the wide one wrote. I verified it fails on the old code, with the wide config receiving the narrow config's token. I also dropped the manual cache clear from `test_refresh_margin_is_configurable`, since that clear is what hid this; it now uses distinct credentials to isolate the two margins.

326 permissions tests pass.

larrysingleton007:
Addressed the cache key issue yesterday. The cache now stores the token's true expiry and each config applies its own margin on read, so a config with a wider margin can no longer ride a narrower one's deadline. The regression drives both configs through the shared entry without clearing the cache, and I verified it fails on the pre-fix code.

Green and ready for another look. It also still needs a `kind/` label, which I can't add as an outside contributor.

larrysingleton007:
Still open and ready whenever you have a moment. The cache-key fix has been in since 5 August: the cache now stores the token's true expiry and each config applies its own margin on read, with a regression that I verified fails against the pre-fix code.

All checks green, no conflicts with current master. It also still needs a `kind/` label, which I can't add as an outside contributor.

ntkathole:
@larrysingleton007 please resolve the conflicts

larrysingleton007:
@ntkathole done, master merged and pushed.

The only conflict was `.secrets.baseline`, which is generated: master had picked up newer entries while this branch carried two shifted line numbers for `test_oidc_auth_client.py` plus a newer `generated_at`. I took master's baseline and re-ran the `detect-secrets` hook rather than hand-merging, which reapplied exactly the two shifts this branch needs (29 to 44, 31 to 46). Hand-editing generated line numbers would have been guesswork.

326 permissions tests pass on the merged tree, with ruff and mypy clean. The cache-key fix and its regression are unchanged.

The other two, #6689 and #6690, merge cleanly against current master, so nothing needed there.

sdk/python/feast/permissions/client/oidc_authentication_client_manager.py
```diff
@@ -13,6 +15,17 @@
...
 
+# IdP-issued tokens keyed by the token-request identity, stored with their
+# true expiry. The auth interceptors build a fresh manager for every outbound
+# RPC, so instance state would not survive between calls; without this
+# cache every RPC pays a discovery GET plus a token POST against the IdP.
+# The refresh margin is applied on read, not baked into the stored value, so
+# configs sharing IdP credentials but setting different margins can share
+# tokens while each still honors its own margin.
+# Concurrent misses may fetch in parallel (benign: last write wins).
+_token_cache: Dict[Tuple, Tuple[str, float]] = {}
```
ntkathole, line 26: Expired entries are never evicted - may be add a lightweight prune on cache miss: drop entries whose stored expiry has passed before inserting the new one ?

sdk/python/feast/permissions/client/oidc_authentication_client_manager.py
```diff
@@ -67,7 +80,71 @@ def _read_sa_token() -> Optional[str]:
...
+        """Return a cached IdP token, or obtain a fresh one.
+
+        The cache stores each token's true expiry (its ``exp`` claim, falling
+        back to the token response's ``expires_in``), and this config's
+        ``token_refresh_margin_seconds`` is applied when reading. Keeping the
+        margin out of the stored value lets configs that share IdP credentials
+        but set different margins share tokens while each honors its own
+        margin. A token whose expiry is unknowable is not cached, preserving
+        the previous per-call behavior for opaque tokens.
+        """
```
ntkathole, line 92: If the IdP revokes a token mid-flight, the client keeps presenting the cached (now-rejected) token for up to ~59 min. Standard OAuth2 practice is to evict the cache entry and retry once on a 401/UNAUTHENTICATED response. Can you add a _invalidate_cached_token(cache_key) helper and wire it in workflow ?

larrysingleton007:
@ntkathole both addressed in 2a73cb7.

**Pruning.** Inserting on a miss now first drops entries whose stored expiry has passed. Keys are credential identities, so the cache was already bounded by the number of distinct configs, but a long-lived process that rotates credentials would have retained every retired one.

**Revocation.** Added `OidcAuthClientManager.invalidate_token()` plus a transport-agnostic `invalidate_auth_token(auth_config)` in `client_auth_token.py`, and wired it into the gRPC interceptor's existing error path: an `UNAUTHENTICATED` status drops the cached token, so the next call refetches. That bounds the staleness to the one request that was rejected rather than the remainder of the token's life.

One deliberate deviation from what you asked, and I'd rather flag it than quietly do less. I did not retry the rejected call. All four interceptor methods share `_handle_call`, and for the stream variants the `request_iterator` may already be partially consumed, so a retry there could replay a partial stream. Retry seems like it belongs per-transport, where the caller knows whether the request is replayable, rather than in the shared handler. The invalidation is the part that fixes the correctness problem; the retry only saves the caller one visible error. Happy to add it for the unary paths in this PR if you'd prefer, or as a follow-up covering all three transports, since the Arrow Flight and requests wrappers have no 401 handling at all today.

Two new tests: one asserting invalidation forces a refetch and is idempotent, one asserting an expired entry is pruned while an unexpired one survives. 328 permissions tests pass, ruff and mypy clean.

ntkathole (approved):

ntkathole:
@larrysingleton007 not able to rebase and merge, can you rebase and possibly squash the commit?

larrysingleton007:
@ntkathole rebased onto current master and squashed to a single commit, 202e30d11. The branch is no longer behind, and the three merge commits are gone.

One note on how I did it, since it affects what you're reviewing. Master had moved three commits past this branch's last merge, so I brought those in first and then collapsed everything onto `upstream/master`. A straight squash of the old branch tip would have reverted those three. The net diff is unchanged: 6 files, +414/-9, all of it this PR's own work.

Re-verified after the rebase, since master's newer commits touched `permissions/user.py`: 329 permissions tests pass, ruff and mypy clean.

sdk/python/feast/permissions/client/oidc_authentication_client_manager.py
larrysingleton007: Done. Inserting on a miss now drops entries whose stored expiry has passed, before writing the new one.

Keys are credential identities, so the cache was already bounded by the number of distinct configs rather than growing per request, but a long-lived process that rotates credentials would have retained every retired identity. Covered by a test that asserts an expired entry is pruned while an unexpired one survives, so the prune can't silently become a clear.

sdk/python/feast/permissions/client/oidc_authentication_client_manager.py
larrysingleton007: Added the eviction. `OidcAuthClientManager.invalidate_token()` plus a transport-agnostic `invalidate_auth_token(auth_config)` in `client_auth_token.py`, wired into the gRPC interceptor's existing error path so an `UNAUTHENTICATED` response drops the cached entry and the next call refetches. That bounds the exposure to the single rejected request instead of the token's remaining life.

I did not add the retry, as noted in the conversation before you approved. All four interceptor methods share `_handle_call`, and for the stream variants the `request_iterator` may already be partially consumed, so retrying there could replay a partial stream. Retry seems to belong per-transport, where the caller knows whether the request is replayable. Still happy to add it for the unary paths here, or as a follow-up covering all three transports, since Arrow Flight and the requests wrapper have no 401 handling at all today. Say which and I'll open it.

larrysingleton007:
@ntkathole both review threads are resolved now. I'd replied in the conversation rather than inside the threads, which is why they stayed open — my mistake.

Summary of where they landed: pruning on cache miss is implemented and covered by a test; eviction on `UNAUTHENTICATED` is implemented via `invalidate_token()` and `invalidate_auth_token()`, wired into the gRPC interceptor. The retry is the one piece I left out, for the stream-replay reason in the thread, and I'm happy to add it either here for the unary paths or as a follow-up across all three transports.

Two things still block the merge, neither of them code in this PR.

The failing check is `unit-test-python (3.10, ubuntu-latest)` on `test_http_session_manager::test_thread_safety`, which asserts `len(sessions) == 10` and got 9. It can't be this PR: that test builds its session from `NoAuthConfig()`, which takes the `AuthType.NONE` branch and never calls `get_auth_token`, so the token cache isn't in its code path. Its own `assert len(errors) == 0` passed, so no thread raised — one thread just didn't record a result, which is a race in the test's own accounting. The test came in with #5895. Could you re-run that job?

@franciscojavierarceo the `reviewDecision` is still CHANGES_REQUESTED from your 3 August review of the cache key. That was fixed on 5 August: the cache now stores the token's true expiry and applies each config's margin on read, with a regression that I verified fails against the pre-fix code. Would you mind taking another look so the review state clears?

larrysingleton007:
Following up on the red check with evidence rather than just asking for a re-run.

`unit-test-python` fails intermittently **on master**, on a different test each time:

| run | date | failing test |
| --- | --- | --- |
| 31768603920 | 14 Aug | `test_key_encoding_utils.py::test_performance_bounds_single_entity` — `assert 0.206 < 0.2` |
| 31501371702 | 11 Aug | `test_feature_server.py::test_push_batched_matrix` — `importlib_metadata.MetadataNotFound` |

Both are on master with no PR involved. The failure here, `test_http_session_manager::test_thread_safety`, is a third instance of the same class: tests that assert on wall-clock timing or thread scheduling, which are sensitive to how loaded the runner is when 8 xdist workers share it.

For this PR specifically, that test builds its session from `NoAuthConfig()`, which takes the `AuthType.NONE` branch and never calls `get_auth_token`, so nothing in this change is in its code path. I also ran it 88 times locally, 40 in isolation and 48 under deliberate CPU contention, without a single failure — consistent with it needing a loaded shared runner rather than being deterministic.

So a re-run should clear it. Separately, `test_performance_bounds_single_entity` looks worth fixing on its own: a 0.2s wall-clock bound cannot be made reliable by raising it a little, since the runner's available CPU isn't bounded. Happy to open an issue and PR for that if it's useful.

larrysingleton007:
One more CI follow-up: the DuckDB failure in the latest run is unrelated to this OIDC diff. I found the same extension-install lock race in 2 other PR runs, filed #6743, and put the fix in #6744.

#6744 installs DuckDB's `delta` extension once before the 8 pytest workers start. The cold-cache suite passes locally with 84 tests passed and 5 skipped, and all 18 ordinary PR checks are green. Its DuckDB CI job is waiting for `ok-to-test`.

No change is needed in this PR for that failure.

Issue #6688, opened 2026-07-31
What is _validate_token meant to verify in the OIDC token parser?


larrysingleton007:
## Question

What is `OidcTokenParser._validate_token` intended to verify? As written it cannot reject any token, but it reads like a security gate and logs as though one passed, so I would rather ask than assume it is dead code.

## What it does today

[`_validate_token`](https://github.com/feast-dev/feast/blob/3a6a1031ac501c715b6d7aaf6d6aa3cee25c5367/sdk/python/feast/permissions/auth/oidc_token_parser.py#L45-L60) builds a mock request, sets `Authorization: Bearer <token>` on it from the token it was handed, and awaits `OAuth2AuthorizationCodeBearer`. That scheme only parses the header and returns the token; it never inspects the value. Since the method supplies the header itself, the check always succeeds.

Verified against the installed fastapi (0.139.2) — every one of these is accepted, including the empty string:

```
''            -> ACCEPTED ''
'   '         -> ACCEPTED ''
'garbage'     -> ACCEPTED 'garbage'
'a.b.c'       -> ACCEPTED 'a.b.c'
```

The caller then logs success:

```python
await self._validate_token(access_token)
logger.debug("Token successfully validated.")
```

Real verification happens afterwards in `_decode_token`, which fetches the JWKS signing key and calls `jwt.decode`. So the token is genuinely validated, just not here. My concern is that the method name, the docstring ("Validate the token ... against the OAuth2 server"), and that log line together suggest a verification step that does not exist, which is the kind of thing that misleads a future reader auditing the auth path.

## The one thing it does affect

Constructing the scheme reads `token_endpoint` and `authorization_endpoint` from the discovery document, so a discovery document **missing** `token_endpoint` raises a `ValidationError` and the request fails (an empty string passes). That is the only input that can make this method fail.

Worth noting the server-side validation path itself never needs a token endpoint — it only needs `jwks_uri`. So this effectively requires a discovery field that Feast's own token validation does not use. It also does not trigger anything that would otherwise be skipped: `_decode_token` reaches the same lazily-fetched discovery data via `get_jwks_url()`.

## Why I am asking rather than sending a patch

Removing the call would be a behavior change for anyone whose provider omits `token_endpoint` (they would go from failing to working), and it is the only `await` on this branch of the OIDC path, which matters if the blocking JWKS work is ever moved to a thread. Both seem like your call rather than mine.

Roughly the options I see:

[...]

PR #6689, opened 2026-07-31
docs: Correct what _validate_token claims to do

larrysingleton007:

`OidcTokenParser._validate_token` is named validate, its docstring says it validates the token against the OAuth2 server, and the caller logs "Token successfully validated" afterwards. None of that is accurate. The bearer scheme only parses an `Authorization` header, and the method constructs that header itself from the token it was handed, so every value passes. Verified against fastapi 0.139.2: the empty string, whitespace, and arbitrary garbage are all accepted.

The token is genuinely verified a few lines later in `_decode_token`, which fetches the JWKS signing key and validates the signature and claims. So authentication is sound. The problem is that anyone auditing this path reads a verification step that does not exist, and a debug log that confirms it.

This changes the docstring to describe what the call actually checks (that the discovery document exposes the OAuth2 endpoints, which is the only thing that can make it fail) and changes the log line to match.

No behavior change: docstring and one log string only. Existing tests pass unmodified.

Whether the call should exist at all is a separate question, asked in #6688, since removing it would change behavior for providers whose discovery document omits a token endpoint. This PR deliberately does not pre-empt that answer.

Relates to #6688

### DIFF
--- a/sdk/python/feast/permissions/auth/oidc_token_parser.py
+++ b/sdk/python/feast/permissions/auth/oidc_token_parser.py
@@ -43,8 +43,17 @@ class OidcTokenParser(TokenParser):
         self._k8s_auth_api = None
 
     async def _validate_token(self, access_token: str):
-        """
-        Validate the token extracted from the header of the user request against the OAuth2 server.
+        """Check that the provider's discovery document exposes the OAuth2 endpoints.
+
+        This does **not** verify *access_token*, despite taking it: the bearer
+        scheme below only parses an ``Authorization`` header, and this method
+        supplies that header itself, so any token value passes. The token is
+        genuinely verified in ``_decode_token``, which checks the signature
+        against the provider's JWKS and validates the claims.
+
+        What can fail here is constructing the scheme, which reads the token
+        and authorization endpoints from the discovery document. A document
+        missing either one raises before any token is inspected.
         """
         # FastAPI's OAuth2AuthorizationCodeBearer requires a Request type but actually uses only the headers field
         # https://github.com/tiangolo/fastapi/blob/eca465f4c96acc5f6a22e92fd2211675ca8a20c8/fastapi/security/oauth2.py#L380
@@ -198,7 +207,7 @@ class OidcTokenParser(TokenParser):
         # Standard OIDC / Keycloak flow
         try:
             await self._validate_token(access_token)
-            logger.debug("Token successfully validated.")
+            logger.debug("OIDC discovery document exposes the expected endpoints.")
         except Exception as e:
             if self._is_ssl_error(e):
                 logger.error(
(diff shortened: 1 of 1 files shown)


larrysingleton007:
This has been green since Friday with no reviewer assigned. It's a small docs correction: `_validate_token` doesn't validate anything beyond decoding, so the docstring and the guide both overstate what it does.

Could one of you take a look, or route it to whoever owns this area? It also needs a `kind/` label, which I can't add as an outside contributor.

sdk/python/feast/permissions/auth/oidc_token_parser.py
```diff
@@ -78,8 +78,17 @@ def _get_jwks_client(self) -> PyJWKClient:
         return self._jwks_client
 
     async def _validate_token(self, access_token: str):
-        """
-        Validate the token extracted from the header of the user request against the OAuth2 server.
+        """Check that the provider's discovery document exposes the OAuth2 endpoints.
```
ntkathole, line 81: isn't the function name misleading then ?

larrysingleton007:
@ntkathole
It was, yes. Renamed it to `_check_discovery_endpoints`, which is what the docstring's first line already said it does. Leaving the old name meant the docstring had to open by correcting the name, which is a smell in itself.

It's private with a single caller in the same file, so the rename is contained. I kept the `access_token` parameter because the bearer scheme still needs a well-formed `Authorization` header, and the caller has the token to hand. Happy to drop it and pass a placeholder if you'd rather make that explicit too, since the value provably can't affect the outcome.

317 permissions tests pass.

larrysingleton007:
Still open. Your naming question was addressed on 5 August: `_validate_token` became `_check_discovery_endpoints`, which is what the docstring's first line already said it does.

The red check is worth explaining, because it isn't this PR and it looks alarming. `unit-test-python (3.12, ubuntu-latest)` failed on `test_3rd_party_providers` with:

```
ModuleNotFoundError: No module named 'psutil'
```

That's a race inside the test suite, not a missing dependency. `tests/unit/infra/test_dependency_conflicts.py` runs `pip install kserve==0.15.2` against the live interpreter with no isolation. feast pins `psutil==5.9.0` and kserve requires `psutil<6.0.0,>=5.9.6`, so pip cannot leave the existing version in place: it uninstalls psutil, then installs 5.9.8. I reproduced that locally and the pip log shows the window plainly.

The suite runs under xdist with 8 workers sharing one environment, so any concurrently running test that shells out to the feast CLI can land in that window and find no psutil. `test_3rd_party_providers` does exactly that. In this run the kserve test finished at 12:25:54 and the failure came at 12:29:15 in the same job. It's timing-dependent, which is why it's intermittent rather than reproducible on demand.

Two things worth flagging beyond this PR. Mutating the interpreter that's running the suite makes any test that shells out nondeterministic, so a re-run will probably go green without anything having changed. And separately, #6713 loosens `psutil==5.9.0` to `>=5.9.0` for unrelated reasons; if the lock then resolves psutil into kserve's range, pip would have nothing to uninstall and this race would close as a side effect.

Happy to open an issue for the isolation problem if that's useful. Otherwise this PR is a one-line rename plus a docstring, with no conflicts against current master.

PR #6690, opened 2026-07-31
feat: Expose the OIDC JWKS tunables through the operator

### LINKED ISSUE #6686: Expose the OIDC JWKS cache lifespan and fetch timeout through the feast-operator
Follow-up to #6682. PR #6683 adds two OIDC options to the Python SDK, `jwks_cache_lifespan_seconds` and `jwks_request_timeout_seconds`, which tune how long the server reuses a fetched JWK set and how long a JWKS fetch may block request serving. Requested in the #6683 review: expose them through the operator too.

Unlike `audience` and `issuer` (#6676/#6677), these are not IdP-coupled values and don't belong in the OIDC Secret. They're operational knobs in the same family as `verifySSL` and `caCertConfigMap`, which the operator already exposes as CR fields on `OidcAuthz` and copies into the generated `feature_store.yaml`. So the plan is to follow that pattern:

- Add two optional fields to `OidcAuthz` in `api/v1/featurestore_types.go`, with the same defaults as the SDK (300 seconds and 10 seconds) and a positive-value constraint matching the SDK's validation.
- Copy them into `oidcParameters` in `repo_config.go` when set, the way `VerifySSL` already is.
- Regenerate the CRD manifests, and add tests plus a note in the operator security guide.

Worth noting for reviewers that the cache lifespan is not purely a performance setting: it also bounds how long a key the IdP has revoked continues to validate tokens, so operators with aggressive rotation may want to lower it.

Best done after #6683 merges and ships, since the operator would otherwise generate config that older feature server images reject.

larrysingleton007:

Follow-up to #6683, which you asked for in that review. #6683 added `jwks_cache_lifespan_seconds` and `jwks_request_timeout_seconds` to `OidcAuthConfig`; this exposes them through the operator so they are reachable from a FeatureStore CR.

They are added as CR fields on `OidcAuthz` rather than OIDC Secret keys, because they are non-secret operational knobs and belong with `verifySSL` and `caCertConfigMap` rather than in the Secret that carries IdP credentials. Wiring follows the existing `VerifySSL` pattern.

```yaml
authz:
  oidc:
    secretRef:
      name: oidc-secret
    jwksCacheLifespanSeconds: 300
    jwksRequestTimeoutSeconds: 10
```

Both are optional pointers with a `Minimum=1` constraint mirroring the SDK's validation, and both are omitted from the generated `feature_store.yaml` when unset so the SDK's own defaults apply rather than the operator asserting a competing set.

Worth flagging for review: `jwksCacheLifespanSeconds` is not purely a performance setting. It also bounds how long a key the provider has revoked continues to validate tokens, so the docs describe that tradeoff rather than presenting it as a throughput dial.

Regenerated artifacts: deepcopy, CRD bases, `dist/install.yaml`, the API reference, and the OLM bundle. The bundle is a separate `make bundle` step that `make manifests` does not cover and no PR workflow runs, so it is easy to miss; without it an OLM install silently prunes the new fields rather than failing.

Testing: unit coverage asserts the keys are absent when unset and forwarded with correct values when set, plus that the client config omits them (it inherits the same strict validation). CRD validation tests cover the `Minimum=1` constraints. `go build ./...`, `gofmt`, and all four operator test packages pass.

One docs fix included: the `caCertConfigMap` example in the same block was written as a bare string, but the CRD requires an object with a `name` key, so that snippet failed to apply as written.

Fixes #6686


franciscojavierarceo (changes requested):
The new JWKS fields are added only to the v1 API and generated v1 artifacts, while v1alpha1 remains a served API version in the CRD. A client submitting the still-served v1alpha1 form cannot express these settings and the fields will be unavailable/pruned during conversion. We should either add the fields and generated schema/deepcopy for v1alpha1 as well, or explicitly document and enforce that these new OIDC settings are unsupported through the served deprecated version.

larrysingleton007:
@franciscojavierarceo
Took the second option and documented it, because the first would make v1alpha1 less consistent rather than more.

The mechanism you describe is real and I confirmed it against the generated CRD: v1 exposes 8 fields under `authz.oidc`, v1alpha1 exposes exactly one (`secretRef`), both are served, and there is no conversion stanza, so a v1alpha1 submission is validated against the v1alpha1 schema and anything outside it is pruned without error.

The part worth flagging is that this isn't specific to the JWKS settings. v1alpha1 has never carried `issuerUrl`, `secretKeyName`, `tokenEnvVar`, `verifySSL` or `caCertConfigMap` either; those all landed v1-only in 7c04026ab. So adding just the two new fields to v1alpha1 would leave it able to tune JWKS caching but still unable to set `verifySSL`, which seems worse for a user than a clean version boundary.

The warning I added covers the whole Advanced OIDC options section rather than the two new fields, and says plainly that applying that example as v1alpha1 leaves OIDC configured by Secret alone with nothing in the output to say so.

On enforcement: the v1alpha1 schema already enforces it, since the fields can't be expressed there at all, so I didn't add a validation path. Happy to backport the full OIDC field set to v1alpha1 if you'd rather close the gap properly, though that feels like its own PR against a deprecated version.

larrysingleton007:
Replied yesterday on the v1alpha1 point. I documented it rather than backporting, because v1alpha1 exposes only `secretRef` under `authz.oidc` while v1 exposes eight fields, and the other five have been v1-only since 7c04026ab. Adding just the two new ones would leave v1alpha1 able to tune JWKS caching but unable to set `verifySSL`. Happy to backport the full OIDC set as its own PR if you'd rather close the gap properly.

Green and ready for another look, and it needs a `kind/` label.

@ntkathole this is the operator follow-up you asked about on #6683.

larrysingleton007:
The red check here is `mcp-feature-server-runtime`, and it isn't from this PR. This branch touches the operator's Go API, the generated CRD manifests and the security guide, nothing in the Python MCP path.

It's failing repo-wide. The run on the unrelated `credentials` branch yesterday (run 30890414540) fails in the same job with the identical error:

```
File "fastapi_mcp/server.py", line 144, in setup_server
    mcp_server: Server = Server(self.name, self.description)
TypeError: Server.__init__() takes 2 positional arguments but 3 were given
```

The cause looks like a constraint that the pinned requirements have drifted past. `pyproject.toml` declares `mcp = ["fastapi_mcp", "mcp>=1.0,<2"]`, but `sdk/python/requirements/py3.11-ci-requirements.txt` pins `mcp==2.0.0`. `fastapi-mcp==0.4.0` still calls `Server(name, description)`, which mcp 2.x no longer accepts, so the feature server logs "MCP support was requested but could not be enabled" and the job's later curl checks fail.

Only PRs carrying `ok-to-test` run this job, which is probably why it hasn't been noticed more widely. Happy to open a separate issue and PR for it if that's useful, rather than fold an unrelated dependency fix into this one.

larrysingleton007:
Still open and ready for another look. The v1alpha1 point was addressed on 5 August: documented rather than backported, since v1alpha1 exposes only `secretRef` under `authz.oidc` while v1 exposes eight, and the other five have been v1-only since 7c04026ab. The offer to backport the full OIDC set separately still stands if you'd prefer that.

The one red check is `mcp-feature-server-runtime`, which is the repo-wide breakage in #6706, not this PR. #6713 fixes it and is still open. No conflicts with current master.

ntkathole (approved):

PR #6697, opened 2026-08-01
fix: Add remote registry client extra

### LINKED ISSUE #6666: Remote registry client fails with ModuleNotFoundError: No module named 'grpc' unless the unrelated [grpcio] extra is installed
## Expected Behavior

Installing Feast with any extra that supports a remote registry, and configuring
`registry_type: remote`, should work — `grpcio` should be present because the
remote registry client cannot run without it.

## Current Behavior

`sdk/python/feast/infra/registry/remote.py` does an unconditional module-level
`import grpc`. `grpcio` is declared only in the `grpcio` optional-dependency
group in `pyproject.toml`, and no other extra depends on it. So an install such
as `pip install feast[redis]` or `feast[postgres]` resolves cleanly and then
fails at runtime as soon as the remote registry is constructed:

```
ModuleNotFoundError: No module named 'grpc'
```

The `grpcio` extra reads as being about serving gRPC (it also pulls
`grpcio-reflection` and `grpcio-health-checking`, which a client does not need),
so there is nothing to suggest a registry *client* requires it. The failure
surfaces only at runtime, in a code path unrelated to anything the user
installed.

## Steps to reproduce

1. `pip install "feast[postgres]==0.64.0"` in a clean virtualenv.

2. `feature_store.yaml`:

   ```yaml
   project: demo
   registry:
     registry_type: remote
     path: registry.example:80
   provider: local
   ```

3. ```python
   from feast import FeatureStore
   FeatureStore(repo_path=".").list_feature_views()
   ```

4. Fails with `ModuleNotFoundError: No module named 'grpc'`.

### Specifications

- Version: 0.64.0; unchanged on 0.65.0 (latest release) and `master`. On 0.65.0
  the `redis` extra pulls only `redis` and `hiredis`, and `grpcio` remains
  declared solely in the `grpcio` extra.
- Platform: Linux, Python 3.12
- Subsystem: packaging / registry — `feast.infra.registry.remote`

## Possible Solution

The dependency has to become reachable — the only question is how.

1. **Move `grpcio` into the core dependencies.** The remote registry is a
   first-class registry type, and `grpcio` ships a wheel on every supported
   platform. This is the only option that fixes an existing
   `pip install feast[postgres]` without the user changing anything.
2. **Add a client-only extra** (for example `remote`) that pulls just `grpcio`,
   and document it as required for `registry_type: remote`. Users still have to
   edit their install line, but there is then a correct one to point them at. The
   existing `grpcio` extra is a poor substitute: it also pulls
   `grpcio-reflection` and `grpcio-health-checking`, which a registry *client*
   does not need, and its name reads as being about serving gRPC.

A lazy import that raises a clear Feast error naming the extra would be an
improvement on the bare `ModuleNotFoundError`, but it is worth being explicit
that it does **not** fix this: the remote registry still cannot run, the call
still fails, and the user is still stuck until the dependency is installed. Good
diagnostics are worth having alongside (1) or (2), not instead of them.

Our workaround is to add `grpcio` to the dependency set of the module that builds
the store, which is invisible to anyone reading `pyproject.toml` for what a
remote registry needs.

yaodong-shen:
## What this PR does

Fixes #6666. `RemoteRegistry` imports `grpc` unconditionally, but `grpcio` was only available through the unrelated server-oriented `grpcio` extra. Users installing a normal provider extra could therefore configure `registry_type: remote` and receive `ModuleNotFoundError` at runtime.

This adds a focused `remote` optional dependency containing only `grpcio` and documents `pip install "feast[remote]"` next to the remote registry configuration. The existing `grpcio` extra remains unchanged for registry/feature-server deployments that need reflection and health-checking packages.

## Validation

- Verified the final branch diff contains only `pyproject.toml` and `docs/reference/registries/remote.md`.
- Verified the TOML extra and documentation text from the master blobs.
- Full Feast dependency installation and unit test suite were not runnable locally because the network clone did not complete; CI should exercise package metadata and docs checks.

## Release notes

NONE

### DIFF
--- a/docs/reference/registries/remote.md
+++ b/docs/reference/registries/remote.md
@@ -4,6 +4,16 @@
 
 The Remote Registry is a gRPC client for the registry that implements the `RemoteRegistry` class using the existing `BaseRegistry` interface.
 
+## Installing the client dependency
+
+The remote registry client requires `grpcio`. Install the dedicated client extra before using `registry_type: remote`:
+
+```bash
+pip install "feast[remote]"
+```
+
+The existing `grpcio` extra remains available when running the registry server and includes its reflection and health-checking dependencies.
+
 ## How to configure the client
 
 User needs to create a client side `feature_store.yaml` file, set the `registry_type` to `remote` and provide the server connection configuration.
--- a/pixi.lock
+++ b/pixi.lock
@@ -2387,6 +2387,7 @@ packages:
   - grpcio>=1.56.2 ; extra == 'grpcio'
   - grpcio-reflection>=1.56.2 ; extra == 'grpcio'
   - grpcio-health-checking>=1.56.2 ; extra == 'grpcio'
+  - grpcio>=1.56.2 ; extra == 'remote'
   - hazelcast-python-client>=5.1 ; extra == 'hazelcast'
   - happybase>=1.2.0,<3 ; extra == 'hbase'
   - ibis-framework>=10.0.0 ; extra == 'ibis'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,6 +86,7 @@ grpcio = [
     "grpcio-reflection>=1.56.2",
     "grpcio-health-checking>=1.56.2",
 ]
+remote = ["grpcio>=1.56.2"]
 hazelcast = ["hazelcast-python-client>=5.1"]
 hbase = ["happybase>=1.2.0,<3"]
 ibis = [


yaodong-shen:
The implementation is ready and the branch is mergeable. Per the repository contribution guide, could a maintainer add the required `kind/bug` and `ok-to-test` labels so the Prow checks can run? The DCO check still correctly awaits the contributor's explicit sign-off authorization.

franciscojavierarceo (approved):
The focused extra cleanly separates the client-only `grpcio` dependency and documents the installation path without changing the existing server extra. I found no blocking issue in the remote diff.

ntkathole (approved):

PR #6699, opened 2026-08-03
fix: Give RemoteRegistry a client-side cache so allow_cache reads skip the RPC

### LINKED ISSUE #6672: RemoteRegistry keeps no client-side cache, so allow_cache=True still issues one RPC per read, unlike every CachingRegistry implementation
## Expected Behavior

`allow_cache=True` on a registry read should mean the read may be served from a
cache without contacting the registry backend, for every registry
implementation. A caller passing `allow_cache=True` in a loop should not issue
one backend round-trip per iteration.

## Current Behavior

`RemoteRegistry` holds no client-side cache. Each read builds a request, calls
the stub and converts the response
(`sdk/python/feast/infra/registry/remote.py`, line 428 on `master`):

```python
def get_feature_view(
    self, name: str, project: str, allow_cache: bool = False
) -> FeatureView:
    request = RegistryServer_pb2.GetFeatureViewRequest(
        name=name, project=project, allow_cache=allow_cache
    )
    response = self.stub.GetFeatureView(request)
    return FeatureView.from_proto(response)
```

`allow_cache` is forwarded as a field on the request message. It selects whether
the *server* may answer from its cache; it has no effect on whether the client
issues the RPC. The pattern is uniform: on `master` the class has 28 methods
taking `allow_cache` and 50 `self.stub.<Rpc>(...)` call sites, and the word
"cache" appears in the file only as that passthrough parameter. No response is
ever retained.

`CachingRegistry` behaves differently for the same argument
(`caching_registry.py`, line 235):

```python
def get_stream_feature_view(
    self, name: str, project: str, allow_cache: bool = False
) -> StreamFeatureView:
    if allow_cache:
        self._refresh_cached_registry_if_necessary()
        return proto_registry_utils.get_stream_feature_view(
            self.cached_registry_proto, name, project
        )
    return self._get_stream_feature_view(name, project)
```

Here `allow_cache=True` is an in-process lookup against
`cached_registry_proto`. `RemoteRegistry` is the outlier: every other registry
implementation has a client-side cache, whether inherited or hand-rolled —
`SqlRegistry(CachingRegistry)` (`sql.py:319`), and `Registry(BaseRegistry)`
(`registry.py:174`, the local-file/S3/GCS registry) and
`SnowflakeRegistry(BaseRegistry)` (`snowflake.py:125`), which both maintain their
own `cached_registry_proto` and `_refresh_cached_registry_if_necessary`.
`RemoteRegistry(BaseRegistry)` (`remote.py:97`) has neither.

The server does honour the field — `registry_server.py:410-418` forwards
`allow_cache=request.allow_cache` to the proxied registry — so for the remote
client `allow_cache` is strictly a server-side hint. It never suppresses the
call.

So the same argument means "may skip the backend" on four registries and
"always one full round-trip" on the fifth. `FeatureStore` methods that default
`allow_registry_cache=True` — `push` (lines 3183, 3221) and
`write_to_offline_store` (3435) — are written on the assumption that repeated
resolution is cheap; that assumption silently does not hold for the remote
registry. (`get_feature_view` and `get_entity` default to `False`, so this is
about the write/push paths specifically.)

**Observed impact.** Long-running batch jobs resolve feature view metadata per
batch. On a remote registry every such resolution is a gRPC round-trip even with
`allow_registry_cache=True`, so registry traffic scales with batch count. It also
removes the usual mitigation: holding one `FeatureStore` for the whole run does
not make the reads cheap, because there is no feature-view cache in the client to
warm.

Two things already in the tree are worth naming, since neither closes this:

- **A partial client cache exists, for feature services only.** `55c2f185`
  ("perf: Cache feature view resolution in get_online_features") added
  `_feature_service_cache`, which `FeatureStore.__init__` attaches to the registry
  instance by `setattr` (`feature_store.py:431-436`) and `utils.py:1190-1192`
  reads. It memoises feature-service→feature-ref resolution only. `get_feature_view`
  and the other object getters are untouched and still pay per call. Notably that
  commit touched `base_registry.py`, `caching_registry.py`, `registry.py` and
  `utils.py`, and deliberately not `remote.py`.
- **`is_cache_valid()` lets a caller probe, but only if they know to.**
  `BaseRegistry.is_cache_valid` (`base_registry.py:1004`) returns `False` with the
  docstring *"Registries without caching always return False (every read goes to
  the backing store)"*; `CachingRegistry` overrides it (`caching_registry.py:509`)
  and `RemoteRegistry` does not. So the divergence is discoverable at runtime.
  That is an unadvertised probe, not a documented contract, and it does not help a
  caller who simply wants `allow_cache` to mean one thing across registries.

#4710 reported the same shape of problem — repeated per-call metadata resolution
from the registry dominating `get_online_features`. It shows as closed/completed,
but it was closed by `stale[bot]` on 2025-05-08 with nothing landed at the time,
so it should not be read as a resolution. The caching that did land later
(`55c2f185`, above) skipped `remote.py`. In that thread a maintainer states the
premise of this issue directly:

> I guess with your wrapper you're overcoming the fact that RemoteRegistry
> doesn't extend CachingRegistry. (and even if it was the way CachingRegistry
> maintains a cache is less that ideal, but that's a different story)
>
> — @tokoko, #4710

## Steps to reproduce

1. Configure a remote registry:

   ```yaml
   project: demo
   registry:
     registry_type: remote
     path: registry.example:80
   ```

2. Read the same object repeatedly with the cache explicitly allowed:

   ```python
   for _ in range(100):
       store.get_feature_view("my_fv", allow_registry_cache=True)
   ```

3. Observe 100 `GetFeatureView` RPCs. Count them in the registry server's access
   log, or wrap `store.registry.stub` and tally calls.

4. Run the same loop against a `registry_type: sql` registry. It issues at most
   one backend read, then serves from `cached_registry_proto` until the TTL
   expires.

### Specifications

- Version: 0.62.0; verified present on 0.65.0 (latest release) and `master`
- Platform: Linux, Python 3.12, Kubernetes; remote registry over gRPC
- Subsystem: registry — `feast.infra.registry.remote.RemoteRegistry`

## Possible Solution

Two directions; the first is the smaller change.

1. Make `RemoteRegistry` extend `CachingRegistry`, implementing the abstract
   `_get_*` / `_list_*` hooks with the existing stub calls. The caching, TTL and
   refresh behaviour then come from the shared base, and `allow_cache` acquires
   the same meaning it already has everywhere else.

   The refresh contract is already satisfied. `CachingRegistry.__init__` line 39
   populates the cache with `self.cached_registry_proto = self.proto()`, and
   `RemoteRegistry.proto` exists (`remote.py:706-707`,
   `return self.stub.Proto(Empty())`) against the `Proto` RPC restored by #6558
   (`RegistryServer.proto:100`, handler `registry_server.py:185`). So no
   reassembly from `list_*` RPCs is needed — the one-shot refresh works as-is.

   One scoping detail: 28 `RemoteRegistry` methods take `allow_cache` but
   `CachingRegistry` declares 26 `_get_*`/`_list_*` hooks. `get_registry_lineage`
   and `get_object_relationships` have no corresponding hook and would need
   separate handling.

2. If unifying is too large, document the divergence — that `allow_cache` on the
   remote registry is a server-side hint only and never suppresses the RPC — so
   callers can size their own caching. `is_cache_valid()` returning `False`
   already encodes this, but nothing points a caller at it; the registry docs say
   nothing about which registry types honour `allow_cache` client-side, and the
   shared signature implies all of them do.

Related: #6671, where this cost is multiplied — a write path issues a
guaranteed-miss lookup before the real one, so a remote registry pays two RPCs
per batch where a caching registry pays roughly none.

adarshsm:
## What this does

Fixes #6672.

`RemoteRegistry` held no client-side cache, so `allow_cache=True` was forwarded to the server as a hint and still issued **one gRPC round-trip per read**. Every other registry (`SqlRegistry`, `Registry`, `SnowflakeRegistry`) serves `allow_cache=True` from an in-process cache, so the same argument meant "may skip the backend" on four registries and "always one round-trip" on the fifth. Long-running batch jobs paid a `GetFeatureView` RPC per batch even with `allow_registry_cache=True` and a single long-lived `FeatureStore`.

This takes the smaller of the two directions in the issue: **make `RemoteRegistry` extend `CachingRegistry`**.

## Changes

**`infra/registry/remote.py`** — `RemoteRegistry(CachingRegistry)`. The 28 `allow_cache` read methods become `_get_*`/`_list_*` hooks that issue the fresh-read RPC; the shared base serves `allow_cache=True` from `cached_registry_proto` (populated once by the server's `Proto` RPC, refreshed on TTL). `allow_cache` now means the same thing on every registry, and `is_cache_valid()` reports a real cache. `refresh()` reloads the client cache and, on an explicit project-scoped call, still pokes the server-side `Refresh` RPC (preserving the historical side effect); internal TTL refreshes pass no project and skip it, so a server refresh failure can't abort the client-cache reload.

Two supporting fixes were required for the client cache to actually resolve reads:

**`registry_server.py`** — `RegistryServer.Proto` rebuilt the registry from `list_*() + obj.to_proto()` for RBAC filtering, but `to_proto()` does not serialize the project, so the assembled proto contained **project-less objects** that `proto_registry_utils` (which filters by project) could never find. This was latent because the aggregate `Proto` response wasn't previously used for per-object lookups. Stamp the known `project` back onto each object during the rebuild so the response matches `proxied_registry.proto()` (most objects at `spec.project`; `DataSource`/`ValidationReference` as a top-level field). RBAC filtering is unchanged.

**`infra/registry/caching_registry.py`** — the eager cache warm-up in `__init__` made a `RemoteRegistry` unconstructable when the server is unreachable at construction time (all other backends are reachable locally). Guard the initial load, mirroring `refresh()`'s existing error handling: an empty cache is left behind (`is_cache_valid()` stays `False`) and the first cache-allowed read refreshes it.

## Tests

- `test_remote_registry.py`: `allow_cache=True` serves 100 reads from the client cache with **0** `GetFeatureView` RPCs; `allow_cache=False` still issues one RPC per read; `is_cache_valid()` is `True`; explicit vs internal `refresh()` behaviour.
- `test_registry_server_proto.py`: each rebuilt object carries its project (`spec.project` and top-level shapes).
- Existing remote/mTLS, SQL-registry, server-`Proto`, and permissions/RBAC unit tests pass unchanged.

Verified end-to-end by driving the real `RegistryServer` over an in-memory channel: construction tolerates an unreachable server, then `allow_cache=True` reads for feature view, entity, feature service, and data source all resolve from the client cache.

### DIFF
--- a/sdk/python/feast/infra/registry/caching_registry.py
+++ b/sdk/python/feast/infra/registry/caching_registry.py
@@ -36,8 +36,19 @@ class CachingRegistry(BaseRegistry):
         self.cached_registry_proto_ttl = timedelta(
             seconds=cache_ttl_seconds if cache_ttl_seconds is not None else 0
         )
-        self.cached_registry_proto = self.proto()
         self.cached_registry_proto_created = _utc_now()
+        try:
+            self.cached_registry_proto = self.proto()
+            self.cached_registry_proto_created = _utc_now()
+        except Exception as e:
+            # A backing store that is unreachable at construction time must not
+            # make the registry unconstructable. Leaving the cache empty keeps
+            # is_cache_valid() False, so the first cache-allowed read triggers a
+            # refresh once the backend is reachable. Mirrors refresh()'s handling.
+            logger.debug(
+                f"Error while performing initial registry cache load: {e}",
+                exc_info=True,
+            )
         if cache_mode == "thread":
             self._start_thread_async_refresh(cache_ttl_seconds)
             atexit.register(self._exit_handler)
--- a/sdk/python/feast/infra/registry/remote.py
+++ b/sdk/python/feast/infra/registry/remote.py
@@ -175,17 +175,36 @@ class RemoteRegistry(CachingRegistry):
         if self.channel:
             self.channel.close()
 
+    def _refresh_cache_after_mutation(self, commit: bool = True):
+        # Now that RemoteRegistry caches the registry proto client-side, a
+        # mutation issued over gRPC would otherwise leave the local cache stale
+        # until the TTL expires -- so a subsequent allow_cache read on the same
+        # client (e.g. apply_feature_view() then get_feature_view(...,
+        # allow_cache=True)) would return the pre-mutation object. Mirror
+        # SqlRegistry and reload the client-side cache after the change lands.
+        #
+        # Only refresh once the write is committed on the server. Uncommitted
+        # (commit=False) writes -- e.g. the batched applies in
+        # FeatureStore.apply() -- are flushed by the trailing commit(), which
+        # refreshes then, so we avoid a full Proto() round-trip per object.
+        # In "thread" cache mode the background refresh thread owns freshness,
+        # matching the SqlRegistry contract.
+        if commit and self.cache_mode == "sync":
+            self.refresh()
+
     def apply_entity(self, entity: Entity, project: str, commit: bool = True):
         request = RegistryServer_pb2.ApplyEntityRequest(
             entity=entity.to_proto(), project=project, commit=commit
         )
         self.stub.ApplyEntity(request)
+        self._refresh_cache_after_mutation(commit)
 
     def delete_entity(self, name: str, project: str, commit: bool = True):
         request = RegistryServer_pb2.DeleteEntityRequest(
             name=name, project=project, commit=commit
         )
         self.stub.DeleteEntity(request)
+        self._refresh_cache_after_mutation(commit)
 
     def _get_entity(self, name: str, project: str) -> Entity:
         request = RegistryServer_pb2.GetEntityRequest(name=name, project=project)
@@ -208,12 +227,14 @@ class RemoteRegistry(CachingRegistry):
             data_source=data_source.to_proto(), project=project, commit=commit
         )
         self.stub.ApplyDataSource(request)
+        self._refresh_cache_after_mutation(commit)
 
     def delete_data_source(self, name: str, project: str, commit: bool = True):
         request = RegistryServer_pb2.DeleteDataSourceRequest(
             name=name, project=project, commit=commit
         )
         self.stub.DeleteDataSource(request)
+        self._refresh_cache_after_mutation(commit)
 
     def _get_data_source(self, name: str, project: str) -> DataSource:
         request = RegistryServer_pb2.GetDataSourceRequest(name=name, project=project)
@@ -238,12 +259,14 @@ class RemoteRegistry(CachingRegistry):
             feature_service=feature_service.to_proto(), project=project, commit=commit
         )
         self.stub.ApplyFeatureService(request)
+        self._refresh_cache_after_mutation(commit)
 
     def delete_feature_service(self, name: str, project: str, commit: bool = True):
         request = RegistryServer_pb2.DeleteFeatureServiceRequest(
             name=name, project=project, commit=commit
         )
         self.stub.DeleteFeatureService(request)
+        self._refresh_cache_after_mutation(commit)
 
     def _get_feature_service(self, name: str, project: str) -> FeatureService:
         request = RegistryServer_pb2.GetFeatureServiceRequest(
@@ -302,12 +325,14 @@ class RemoteRegistry(CachingRegistry):
         )
 
         self.stub.ApplyFeatureView(request)
+        self._refresh_cache_after_mutation(commit)
 
     def delete_feature_view(self, name: str, project: str, commit: bool = True):
         request = RegistryServer_pb2.DeleteFeatureViewRequest(
             name=name, project=project, commit=commit
         )
         self.stub.DeleteFeatureView(request)
+        self._refresh_cache_after_mutation(commit)
 
     def _get_stream_feature_view(self, name: str, project: str) -> StreamFeatureView:
         request = RegistryServer_pb2.GetStreamFeatureViewRequest(
@@ -457,6 +482,7 @@ class RemoteRegistry(CachingRegistry):
             commit=commit,
         )
         self.stub.ApplyMaterialization(request)
+        self._refresh_cache_after_mutation(commit)
 
     def apply_saved_dataset(
         self,
@@ -468,12 +494,14 @@ class RemoteRegistry(CachingRegistry):
             saved_dataset=saved_dataset.to_proto(), project=project, commit=commit
         )
         self.stub.ApplyFeatureService(request)
+        self._refresh_cache_after_mutation(commit)
 
     def delete_saved_dataset(self, name: str, project: str, commit: bool = True):
         request = RegistryServer_pb2.DeleteSavedDatasetRequest(
             name=name, project=project, commit=commit
         )
         self.stub.DeleteSavedDataset(request)
+        self._refresh_cache_after_mutation(commit)
 
     def _get_saved_dataset(self, name: str, project: str) -> SavedDataset:
         request = RegistryServer_pb2.GetSavedDatasetRequest(name=name, project=project)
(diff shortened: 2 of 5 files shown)


franciscojavierarceo (changes requested):
Once `RemoteRegistry` becomes a `CachingRegistry`, its mutation methods still only issue the RPC and never invalidate or refresh `cached_registry_proto`. For example, `apply_feature_view()` followed by `get_feature_view(..., allow_cache=True)` on the same client can return the pre-apply object until the cache TTL expires; the old implementation forwarded the read to the server instead. We should invalidate/refresh the client cache after successful apply/delete/commit operations and add a same-client mutation-then-cached-read regression.

adarshsm:
Good catch — fixed in bb678b7.

Mutation methods now reload the client-side cache after a committed write, mirroring `SqlRegistry` (which already does `if cache_mode == "sync": self.refresh()` after each apply/delete). So `apply_feature_view(commit=True)` followed by `get_feature_view(..., allow_cache=True)` on the same client now returns the applied object instead of the pre-apply snapshot.

Two design points worth calling out:

- **The refresh is gated on `commit=True`.** `FeatureStore.apply()` stages every object with `commit=False` and calls `commit()` once at the end, so refreshing on each uncommitted write would fire a full `Proto()` round-trip per object (and read pre-commit state). Instead, `commit=False` writes are left alone and the trailing `commit()` refreshes the client cache once. Direct `apply_*/delete_*` calls default to `commit=True`, so the reported scenario is covered.
- **`thread` cache mode is unchanged** — the background refresh thread continues to own freshness there, same as `SqlRegistry`.

Added same-client mutation-then-cached-read regression tests: `test_apply_with_commit_refreshes_client_cache` (the reported case; fails without the fix), `test_commit_false_defers_refresh_until_commit` (batch path refreshes once on `commit()`), and `test_thread_cache_mode_skips_refresh_on_mutation`.

adarshsm:
@franciscojavierarceo this is ready for another look — the cache-invalidation gap you flagged was fixed in `bb678b7` (now `ce3c176` after a rebase onto current master).

Every mutation method that reaches the server refreshes the client-side cache when the write is committed, mirroring `SqlRegistry`'s `if cache_mode == "sync": self.refresh()`. `delete_label_view` inherits it by delegating to `delete_feature_view`, and `apply_user_metadata` is a no-op on this registry, so those two are the only mutation methods without an explicit call.

The refresh is gated on `commit=True` so the `commit=False` batch in `FeatureStore.apply()` does not fire a `Proto()` round-trip per object — the trailing `commit()` refreshes once instead. In `thread` cache mode the background thread keeps owning freshness, same as `SqlRegistry`.

Regression tests cover the three paths: committed apply then `allow_cache=True` read on the same client, the deferred-until-`commit()` batch, and the thread-mode contract.

adarshsm:
Rebased onto current `master` (`5ad5592`) — it had gone conflicting. Force-pushed as `dc67875`; no behaviour change from the revision you reviewed, but two things are worth flagging because the rebase was not purely mechanical.

**1. A conflict where a union merge would have been wrong.** `apply_saved_dataset` conflicted between your fix to call the correct RPC and my added cache refresh:

```python
<<<<<<< master
        self.stub.ApplySavedDataset(request)      # your fix
=======
        self.stub.ApplyFeatureService(request)    # the old call my branch was based on
        self._refresh_cache_after_mutation(commit)
```

Resolved as your corrected RPC **plus** the refresh — taking both sides verbatim would have issued two RPCs.

**2. Your new test needed a fixture update.** `test_apply_saved_dataset_calls_apply_saved_dataset_rpc` failed after the rebase with `AttributeError: 'RemoteRegistry' object has no attribute 'cache_mode'`. The `remote_registry` fixture patches `__init__` away and only sets `stub`, which was complete before but is not now that `RemoteRegistry` is a `CachingRegistry` — the mutation path reads `cache_mode`. I gave the fixture a `CachingRegistry.__init__` call and a `Proto` return value so it is a faithful stand-in. Test-only; your test passes unchanged otherwise.

`sdk/python/tests/unit/infra/registry/` is 64 passed (snowflake excluded, missing optional dep) and `ruff check`/`format` are clean.

The cache-invalidation gap you flagged remains fixed — every mutation that reaches the server refreshes the client cache on a committed write, mirroring `SqlRegistry`.

PR #6700, opened 2026-08-03
fix: Report single-feature-view spark_application materialization success

### LINKED ISSUE #6673: spark_application engine reports successful single-feature-view materializations as failure
## Expected Behavior

When a SparkApplication reaches `COMPLETED`, Feast should report the materialization as successful.

## Current Behavior

Every successful single-feature-view materialization using the `spark_application` batch engine is reported as a failure:

```text
Exception: SparkApplication feast-sa-<job-id> not found
```

This is deterministic: Feast deletes the SparkApplication CR before performing the status check that fails.

`SparkApplicationComputeEngine.materialize()` waits for the job to reach a terminal status and then unconditionally deletes the CR:

```python
try:
    self._wait_for_completion(job)
    return self._build_per_fv_jobs(registry, tasks, job_id, job)
finally:
    self._cleanup(job_id)
```

After `materialize()` returns, `FeatureStore._submit_and_process_materialization_jobs()` calls `job.status()` again. For a single feature view, `_build_per_fv_jobs()` returns the live polling job:

```python
if len(tasks) <= 1:
    return [job for _ in tasks]
```

The second status check therefore queries the CR that `_cleanup()` just deleted. The resulting 404 sets `_error`, changes the status to `ERROR`, and is subsequently raised by `FeatureStore`.

By this point, the Spark driver has already materialized the features and updated the registry. Automation calling `feast materialize` or the feature server's `/materialize` endpoint therefore sees a failure and may retry work that already succeeded.

This is not caused by the Spark Operator's `timeToLiveSeconds`; Feast explicitly deletes the CR immediately after `_wait_for_completion()` returns.

The relevant flow is still present on `master`:

- [Cleanup after completion](https://github.com/feast-dev/feast/blob/ca355cb73fdd2e8d955274b9e7e65ada060ddfe3/sdk/python/feast/infra/compute_engines/spark_application/compute.py#L224-L228)
- [Single-FV path returns the polling job](https://github.com/feast-dev/feast/blob/ca355cb73fdd2e8d955274b9e7e65ada060ddfe3/sdk/python/feast/infra/compute_engines/spark_application/compute.py#L266-L267)
- [Status re-queries the Kubernetes API](https://github.com/feast-dev/feast/blob/ca355cb73fdd2e8d955274b9e7e65ada060ddfe3/sdk/python/feast/infra/compute_engines/spark_application/job.py#L94-L115)
- [404 changes the job to an error](https://github.com/feast-dev/feast/blob/ca355cb73fdd2e8d955274b9e7e65ada060ddfe3/sdk/python/feast/infra/compute_engines/spark_application/job.py#L129-L134)
- [FeatureStore performs the second status check](https://github.com/feast-dev/feast/blob/ca355cb73fdd2e8d955274b9e7e65ada060ddfe3/sdk/python/feast/feature_store.py#L563-L569)

## Steps to reproduce

1. Configure a Feast project with `batch_engine.type: spark_application`.
2. Apply the feature definitions.
3. Materialize exactly one feature view.
4. Wait for the SparkApplication to reach `COMPLETED`.
5. Observe Feast delete the CR and report:

   ```text
   SparkApplication feast-sa-<job-id> not found
   ```

The feature-view count matters. For multiple feature views, `_build_per_fv_jobs()` returns `CompletedMaterializationJob` instances for successful views. Their `status()` method returns `SUCCEEDED` without querying Kubernetes, so the successful multi-FV path does not exhibit this bug.

The failure can also be reproduced without a cluster:

```python
from kubernetes.client.exceptions import ApiException

from feast.infra.compute_engines.spark_application.job import (
    SparkApplicationMaterializationJob,
)


class Api:
    """Return COMPLETED once, then simulate the CR having been deleted."""

    def __init__(self):
        self.calls = 0

    def get_namespaced_custom_object(self, **kwargs):
        self.calls += 1
        if self.calls == 1:
            return {
                "status": {
                    "applicationState": {
                        "state": "COMPLETED",
                    }
                }
            }
        raise ApiException(status=404, reason="Not Found")


job = SparkApplicationMaterializationJob("abc123", "default", Api())

print(job.status())  # SUCCEEDED: observed by _wait_for_completion
print(job.status())  # ERROR: observed by FeatureStore after cleanup
print(job.error())   # SparkApplication feast-sa-abc123 not found
```

### Specifications

- Version: Feast v0.65.0; also present on `master` at `ca355cb`
- Platform: Kubernetes/EKS with the Kubeflow Spark Operator (`sparkoperator.k8s.io/v1beta2`)
- Subsystem: `spark_application` batch compute engine and materialization

## Possible Solution

Cache the first terminal status (`SUCCEEDED` or `ERROR`) in `SparkApplicationMaterializationJob` and return it on subsequent calls instead of polling Kubernetes again. Non-terminal states should continue polling normally.

An alternative is to return a `CompletedMaterializationJob` from the successful single-FV path, matching the existing multi-FV behavior. However, caching terminal state is more robust because it prevents any future caller from losing a known result after cleanup.

Regression tests should verify that:

- `COMPLETED` followed by a 404 remains `SUCCEEDED`, with `error()` returning `None`.
- `FAILED` remains `ERROR` and preserves its `errorMessage`.
- Non-terminal states continue polling.

I have the terminal-status caching fix and regression tests working locally and can open a PR

adarshsm:
## What this does

Fixes #6673.

Every **single**-feature-view materialization on the `spark_application` batch engine is reported as a failure:

```text
Exception: SparkApplication feast-sa-<job-id> not found
```

even though the Spark driver already materialized the features and updated the registry. Automation calling `feast materialize` or the feature server's `/materialize` endpoint sees a failure and may retry work that already succeeded. It is deterministic for exactly one feature view.

## Root cause

`materialize()` deletes the SparkApplication CR in its `finally` immediately after `_build_per_fv_jobs()` returns:

```python
try:
    self._wait_for_completion(job)
    return self._build_per_fv_jobs(registry, tasks, job_id, job)
finally:
    self._cleanup(job_id)          # deletes the CR
```

For a single feature view, `_build_per_fv_jobs()` short-circuited and returned the **live polling job**:

```python
if len(tasks) <= 1:
    return [job for _ in tasks]
```

`FeatureStore._submit_and_process_materialization_jobs()` then re-checks `job.status()`. On success the job has no captured error, so `status()` re-queries the CR that `_cleanup()` just deleted → 404 → `_error` set → `ERROR` raised. (On genuine failure `job._error` is already set and `status()` short-circuits, so only the **success** path 404s.)

## Fix

Resolve the per-FV outcome from **registry state** for every task count — exactly what the multi-FV path already does. The driver pod sets each materialized FV to `AVAILABLE_ONLINE`, so:

- materialized FV → `CompletedMaterializationJob`, which reports `SUCCEEDED` with **no Kubernetes call** and is safe after cleanup;
- unmaterialized FV → `ERROR` (`was not materialized …`), consistent with the existing multi-FV behavior; the Spark driver's error detail is already logged by `_wait_for_completion`.

The now-unused `job` parameter to `_build_per_fv_jobs()` is dropped.

## Tests

The existing single-task test asserted the buggy behavior (return the live job, skip the registry). Rewrote it to require resolution from registry state with **no CR query**, and added a single-**failed**-FV case. All `spark_application` unit tests pass; the fix is exercised entirely with mocked K8s/registry — no live cluster needed.

Verified the before/after on the real classes: a live job re-checked after CR deletion returns `ERROR "… not found"` (the reported symptom), while the fixed single-FV path returns `SUCCEEDED` and makes no `get_namespaced_custom_object` call.

### DIFF
--- a/sdk/python/feast/infra/compute_engines/spark_application/compute.py
+++ b/sdk/python/feast/infra/compute_engines/spark_application/compute.py
@@ -223,7 +223,7 @@ class SparkApplicationComputeEngine(ComputeEngine):
         )
         try:
             self._wait_for_completion(job)
-            return self._build_per_fv_jobs(registry, tasks, job_id, job)
+            return self._build_per_fv_jobs(registry, tasks, job_id)
         finally:
             self._cleanup(job_id)
 
@@ -252,7 +252,6 @@ class SparkApplicationComputeEngine(ComputeEngine):
         registry: BaseRegistry,
         tasks: List[MaterializationTask],
         job_id: str,
-        job: SparkApplicationMaterializationJob,
     ) -> List[MaterializationJob]:
         """Build one independent job object per FV from registry state.
 
@@ -262,10 +261,15 @@ class SparkApplicationComputeEngine(ComputeEngine):
 
         Each returned job is an independent object so that a failed
         SparkApplication does not pollute the status of succeeded FVs.
-        """
-        if len(tasks) <= 1:
-            return [job for _ in tasks]
 
+        The per-FV outcome is resolved from registry state — never from the live
+        polling job — for every task count. ``materialize()`` deletes the
+        SparkApplication CR in its ``finally`` immediately after this returns, and
+        ``FeatureStore`` then re-checks ``status()``; a job that re-queried the
+        now-deleted CR would 404 and report a false failure for a materialization
+        that actually succeeded (#6673). ``CompletedMaterializationJob`` reports
+        SUCCEEDED without any Kubernetes call, so it is safe after cleanup.
+        """
         jobs: List[MaterializationJob] = []
         for task in tasks:
             fv = registry.get_feature_view(task.feature_view.name, task.project)
--- a/sdk/python/tests/unit/infra/compute_engines/test_spark_application.py
+++ b/sdk/python/tests/unit/infra/compute_engines/test_spark_application.py
@@ -313,8 +313,7 @@ def test_build_per_fv_jobs_all_succeeded():
     task2.feature_view.name = "fv_2"
     task2.project = "test"
 
-    parent_job = SparkApplicationMaterializationJob("job1", "default", MagicMock())
-    jobs = engine._build_per_fv_jobs(mock_registry, [task1, task2], "job1", parent_job)
+    jobs = engine._build_per_fv_jobs(mock_registry, [task1, task2], "job1")
 
     assert len(jobs) == 2
     assert all(isinstance(j, CompletedMaterializationJob) for j in jobs)
@@ -343,10 +342,7 @@ def test_build_per_fv_jobs_partial_failure():
     task_fail.feature_view.name = "fv_fail"
     task_fail.project = "test"
 
-    parent_job = SparkApplicationMaterializationJob("job1", "default", MagicMock())
-    jobs = engine._build_per_fv_jobs(
-        mock_registry, [task_ok, task_fail], "job1", parent_job
-    )
+    jobs = engine._build_per_fv_jobs(mock_registry, [task_ok, task_fail], "job1")
 
     assert len(jobs) == 2
     assert isinstance(jobs[0], CompletedMaterializationJob)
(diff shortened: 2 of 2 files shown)


franciscojavierarceo (approved):
The single-task path now resolves the outcome from registry state before the SparkApplication CR is deleted, and the tests cover both successful and failed single-feature-view materializations without a post-cleanup Kubernetes lookup. I found no blocking issue in the remote diff.

aniketpalu:
Thanks @adarshsm for raising the fix. Please fix the CI. LGTM

aniketpalu (approved):

adarshsm:
@aniketpalu I rebased onto current master (`abf7dc5`), but the red check is not coming from this PR — master is red on its own.

The failure is always in `sdk/python/tests/unit/test_feature_server.py`, and it is the test environment rather than the code:

- this PR, `unit-test-python (3.11)`: `ImportError: cannot import name 'bcat' from 'psutil._common'` — a half-installed psutil in the runner venv
- this PR before the rebase, `unit-test-python (3.10)`: `importlib.metadata.PackageNotFoundError: No package metadata was found for protobuf`
- master `9affee5` ([run 31501371702](https://github.com/feast-dev/feast/actions/runs/31501371702)), `unit-test-python (3.10)`: `importlib_metadata.MetadataNotFound: No package metadata was found` — same file, 1 failed / 2529 passed

This PR only touches `infra/compute_engines/spark_application/compute.py` and its own test, and nothing in it imports the feature server. Every other job is green, including `unit-test-python (3.12)` on both ubuntu and macOS.

Looks like a flaky dependency resolution in the CI install step rather than anything a PR can fix. Happy to re-run once master is green if you would rather see a clean board first.

ntkathole (approved):

PR #6704, opened 2026-08-05
feat: Added SQL registry schema_mode and registry create command

### LINKED ISSUE #6678: Pre-create feature registry schema
Whenever we start up connection to SQL based feature registry, it attempts to create schema. Allowing app to modify schema is not recommended, if it comes with a bug it might corrupt something or worse case delete it. 

Schema should be pre-created but since new version of feature registry might need new schema, the sql based registry connection should read schema to see if schema is compatible first, then recommend running a schema tool to create it. 

We run the registry service with admin and then redeploy with read-only privileges but that is very manual and time consuming.

**Additional context**
Not necessary for non-sql based feature registry.

AdityaPatil22:
### What this PR does / why we need it:
Adds a schema_mode configuration option to the SQL registry that controls whether Feast creates database tables on startup. This addresses production environments where the application database user should not have DDL (CREATE TABLE) privileges.

Three modes are supported:
- auto (default) - current behavior, creates tables if they don't exist. No breaking change.
- verify - skips DDL but checks that all expected tables exist on startup. Raises a clear error listing missing tables with remediation instructions.
- skip - skips both creation and verification. For environments where schema is fully managed externally (e.g. migration tools like Alembic/Flyway).

Also adds a feast registry create-schema CLI command to pre-create the schema with a privileged user, so the runtime user only needs DML access.

### Which issue(s) this PR fixes:
Fixes #6678

### Checks
- [x] My commits are signed off (git commit -s)
- [x] My PR title follows conventional commits (https://www.conventionalcommits.org/) format

### Testing Strategy
- [x] Unit tests
- [ ] Integration tests
- [x] Manual tests

Unit tests - 5 new tests in TestSchemaMode covering: auto creates tables, verify raises on missing tables, verify passes with pre-created schema, skip works with existing schema, invalid values rejected by Pydantic.

Manual tests - Verified all three modes end-to-end against SQLite: auto creates and applies, verify succeeds with existing schema and fails without, skip works, east registry create-schema creates schema for verify/skip to use.

### DIFF
--- a/docs/reference/feast-cli-commands.md
+++ b/docs/reference/feast-cli-commands.md
@@ -28,6 +28,7 @@ Commands:
   materialize              Run a (non-incremental) materialization job to...
   materialize-incremental  Run an incremental materialization job to ingest...
   permissions              Access permissions
+  registry                 Manage the feature registry
   registry-dump            Print contents of the metadata registry
   teardown                 Tear down deployed feature store infrastructure
   version                  Display Feast SDK version
@@ -482,6 +483,18 @@ reader             driver_hourly_stats_fresh   FeatureView      DESCRIBE
 ...
 ```
 
+
+## Registry
+
+### create-schema
+
+Pre-create the SQL registry schema so the application does not need DDL privileges at runtime. Use this with `schema_mode: verify` or `schema_mode: skip` in your `feature_store.yaml`.
+
+```text
+feast registry create-schema
+```
+
+This command only applies to SQL-based registries (`registry_type: sql`). It is safe to run multiple times — existing tables are not modified.
 
 ## Teardown
 
--- a/docs/reference/feature-store-yaml.md
+++ b/docs/reference/feature-store-yaml.md
@@ -66,6 +66,7 @@ registry:
 |-------|------|---------|-------------|
 | `registry_type` | string | `file` | Registry backend (`file`, `sql`, etc.) |
 | `path` | string | — | Connection string or file path |
+| `schema_mode` | string | `auto` | SQL registry only. `auto`: create tables on startup; `verify`: check tables exist, error if missing; `skip`: no DDL or verification. See [SQL Registry docs](registries/sql.md#schema-management-schema_mode). |
 | `mcp.enabled` | bool | `false` | Enable MCP (Model Context Protocol) on the REST registry server |
 
 When `registry.mcp.enabled` is `true`, the REST registry server exposes registry
--- a/docs/reference/registries/sql.md
+++ b/docs/reference/registries/sql.md
@@ -94,7 +94,7 @@ registry:
 | Value | Behavior |
 |---|---|
 | `auto` (default) | Creates tables if they don't exist. Current behavior, no breaking change. |
-| `verify` | Skips DDL. Checks that all expected tables exist on startup; raises an error listing missing tables if any are absent. Note: this is a table-level check only — it does not verify individual columns. A schema created by an older Feast version (missing newer columns) will pass verification but may fail at query time. |
+| `verify` | Skips DDL. Checks that all expected tables exist on startup; raises an error listing missing tables if any are absent. When a separate `read_path` is configured, the read replica is also verified — a lagging replica (e.g. mid-migration) will block startup. Note: this is a table-level check only — it does not verify individual columns. A schema created by an older Feast version (missing newer columns) will pass verification but may fail at query time. |
 | `skip` | Skips both creation and verification. Use when schema is managed entirely outside Feast (e.g. by a migration tool). |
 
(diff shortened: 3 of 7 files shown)


nquinn408:
@AdityaPatil22 Thanks for this — it's a focused, backward-compatible change that fills a real gap (running the SQL registry against a database where the runtime role shouldn't hold DDL grants). Overall looks good; a few minor points below, none blocking.
  
## Strengths

  - Default auto preserves existing behavior — no migration needed for current users.
  - FeastRegistrySchemaError is actionable: it names the missing tables and the remediation command.
  - The CLI engine construction mirrors the write-engine build (path + sqlalchemy_config_kwargs), so the pre-created schema matches what the runtime expects.
  - Good docs coverage (yaml reference, CLI reference, dedicated SQL-registry section), and idempotency is called out.

## Suggestions (minor / follow-up)

  1. verify checks table existence only, not columns. _verify_schema compares table-name sets. A schema created by an older Feast version (missing a newer column) would pass verify but fail at query  time. _warn_if_narrow_blob_columns already inspects column metadata, so there's precedent — worth either a doc caveat that verify is table-level only, or a follow-up that also checks expected columns.
  2. verify skips the read replica. __init__ only calls _verify_schema(self.write_engine). When read_path points at a separate replica, a missing table there won't be caught. The existing code runs _warn_if_narrow_blob_columns against both engines — mirroring that (verify the read engine too when read_engine is not write_engine) would be consistent.
  3. teardown() still issues DDL. The stated motivation is "the app doesn't need DDL privileges at runtime," but teardown() drops tables regardless of schema_mode. That's defensible since teardown is an explicit admin action, but a one-line doc note would prevent surprise for a verify/skip deployment whose runtime role genuinely lacks DROP.
  4. test_schema_mode_skip_does_not_run_ddl doesn't fully prove the claim. It pre-creates the schema, so it can't distinguish skip from auto/verify. A stronger assertion would run skip against an empty DB and assert it (a) does not raise and (b) leaves the DB with no registry tables — that directly proves "no DDL, no verification."

## Nits

  - The new from sqlalchemy import (inspect as sa_inspect,) is a separate import block directly below the existing large from sqlalchemy import (...) — worth folding in.
  - FeastRegistrySchemaError.__init__(self, missing_tables: list[str]) uses list[str] while the rest of the file uses typing.List. Cosmetic, but matching the file convention would be nice.
  - The test tempfiles from mkstemp() are never removed — teardown() drops the tables but the sqlite file leaks in the temp dir. Consider cleanup or a tmp_path fixture.

nquinn408 (commented):
See feedback

AdityaPatil22:
@nquinn408   Thanks for the thorough review! 

Addressed all points:
verify checks table existence only, not columns - Added a doc caveat to the verify row in the mode table noting it's table-level only and that a schema from an older Feast version may pass verification but fail at query time. Column-level verification can be a follow-up.

verify skips the read replica - _verify_schema now runs against read_engine too when it differs from write_engine, mirroring the existing _warn_if_narrow_blob_columns pattern.

teardown() still issues DDL - On closer inspection, teardown() actually runs DELETE FROM (DML), not DROP TABLE (DDL), so it doesn't need DDL privileges. Updated the doc to correct this pre-existing inaccuracy: "the registry deletes all rows from the registry tables (it does not drop the tables themselves). This runs regardless of schema_mode and requires only DML (DELETE) privileges, not DDL."

test_schema_mode_skip_does_not_run_ddl - Rewritten to mock metadata.create_all and SqlRegistry._verify_schema, asserting neither is called. This directly proves the "no DDL, no verification" claim — the old test pre-created the schema so it couldn't distinguish skip from auto/verify.

Nits:
- list[str] → List[str] to match the file's typing.List convention.
- Folded inspect as sa_inspect into the main sqlalchemy block; ruff re-separates it because the alias can't share the # type: ignore comment - this is as merged as the linter allows.
- All tests switched from tempfile.mkstemp() to pytest tmp_path fixture for automatic cleanup.

nquinn408 (approved):
### One residual consideration (non-blocking)

  Adding `_verify_schema(self.read_engine)` means verify mode now hard-fails startup if the read replica is missing a table — even transiently. There's already a comment nearby noting "a read replica can be on a different schema version (e.g. mid-migration)." For verify that strictness is arguably the intent (you asked it to assert the schema), so I'd keep it — but it's worth the author being aware that a lagging replica mid-migration will now block startup in verify mode. A one-line doc mention would be nice; not required.

AdityaPatil22:
@nquinn408 - I have updated the doc as suggested

nquinn408:
@AdityaPatil22 Where are we with this PR? There is a test failure. Thanks for your work on this!

PR #6709, opened 2026-08-06
fix: Feast apply silently ignoring ttl updates to None or timedelta(0)

### LINKED ISSUE #6703: `feast apply` silently ignores ttl updates when the new ttl is `None` or `timedelta(0)`
## Expected Behavior

Re-applying an existing `FeatureView` with `ttl=None` or `ttl=timedelta(0)` should clear the ttl in the registry. After `feast apply`, `store.get_feature_view("foo_fv").ttl` should reflect the new "no expiry" value.

## Current Behavior

The old ttl is silently retained. `store.get_feature_view("foo_fv").ttl` still returns the previous finite value, with no warning or error, while `feast apply` keeps printing the same "changed" diff on every run.

`Registry.apply_feature_view()` routes ttl changes through `_update_metadata_fields()` instead of creating a new version, and that method gates the ttl copy behind a truthiness check on the raw `timedelta`:

```python
if (
    hasattr(existing_proto.spec, "ttl")
    and hasattr(updated_fv, "ttl")
    and updated_fv.ttl          # falsy for both None and timedelta(0)
):
```

Since `None` and `timedelta(0)` are both falsy — and both are the documented way to express "no ttl" — the update is dropped exactly in the case it should support. `get_ttl_duration()` already handles `None` correctly; only the outer guard is wrong. Still present on `master` as of 2026-08.

Practical impact: a `FeatureView` intended to never expire silently kept a stale finite ttl, causing `get_historical_features` to drop ~11.6% of otherwise-valid rows from a point-in-time join, with no error or warning.

## Steps to reproduce

```python
from feast import Entity, FeatureView, Field, FileSource, Project

project = Project(name="ttl_repro")
entity = Entity(name="foo", join_keys=["foo_id"])
source = FileSource(
    name="foo_source",
    path="./data/foo.parquet",
)

# 1. Apply with a finite ttl -> fv.ttl == 10 days (correct)
fv = FeatureView(
    name="foo_fv",
    entities=[entity],
    ttl=timedelta(days=10),
    schema=[Field(name="bar", dtype=Float32)],
    source=source,
)

# 2. Change ttl to "no expiry" and `feast apply` again
fv_v2 = FeatureView(
    name="foo_fv",
    entities=[entity],
    ttl=None,  # or timedelta(days=0) - same result
    schema=[Field(name="bar", dtype=Float32)],
    source=source,
)
```

Then check `store.get_feature_view("foo_fv").ttl` — it is still `timedelta(days=10)`.

### Specifications

- Version: 0.65.0 (PyPI); also confirmed on `master` as of 2026-08
- Platform: 
- Subsystem: Registry — `sdk/python/feast/infra/registry/registry.py`, `Registry._update_metadata_fields()`

## Possible Solution

Drop the outer truthiness check and always sync ttl on update, relying on `get_ttl_duration()`'s existing `None`-handling and defaulting to an explicit zero `Duration` when unset:

```python
if hasattr(existing_proto.spec, "ttl") and hasattr(updated_fv, "ttl"):
    if isinstance(updated_fv, FeatureView):
        existing_proto.spec.ttl.CopyFrom(updated_fv.get_ttl_duration() or Duration())
    elif isinstance(updated_fv, LabelView):
        ttl_duration = Duration()
        if updated_fv.ttl is not None:
            ttl_duration.FromTimedelta(updated_fv.ttl)
        existing_proto.spec.ttl.CopyFrom(ttl_duration)
```

Happy to open a PR with this plus a regression test (apply with finite ttl, re-apply with `ttl=None`, assert the registry reflects it)

saket3395:
## Summary
Fixes #6703 — re-applying an existing `FeatureView`/`LabelView` with `ttl=None` or `ttl=timedelta(0)` (both the documented way to express "no ttl") silently keeps the old finite ttl in the registry. `feast apply` reports the change on every run, but nothing actually updates.

**Root cause (as diagnosed in the issue):** `Registry._update_metadata_fields()` routes ttl changes through a truthiness check:

```python
if (
    hasattr(existing_proto.spec, "ttl")
    and hasattr(updated_fv, "ttl")
    and updated_fv.ttl          # falsy for both None and timedelta(0)
):
```

Since `None` and `timedelta(0)` are both falsy, the whole ttl-update block is skipped exactly when it should clear the ttl.

**There's a second bug in the same block that the outer-gate fix alone doesn't cover:** for the `FeatureView` branch, `get_ttl_duration()` returns Python `None` when `self.ttl is None`:

```python
def get_ttl_duration(self):
    ttl_duration = None
    if self.ttl is not None:
        ttl_duration = Duration()
        ttl_duration.FromTimedelta(self.ttl)
    return ttl_duration
```

and the existing inner check in `_update_metadata_fields`:

```python
ttl_duration = updated_fv.get_ttl_duration()
if ttl_duration:
    existing_proto.spec.ttl.CopyFrom(ttl_duration)
```

would *still* skip `CopyFrom` when `ttl_duration is None` — so removing only the outer gate fixes the `timedelta(0)` case but not `ttl=None` for `FeatureView`.

## Fix
- Removed the outer truthy gate; kept only the `hasattr` checks.
- For `FeatureView`: always `CopyFrom` — using an explicit empty `Duration()` when `get_ttl_duration()` returns `None`, instead of skipping the write. An empty `Duration` decodes back to `timedelta(0)` via `FeatureView.from_proto`'s existing `ToNanoseconds() == 0` check, so this correctly round-trips as "no ttl".
- For `LabelView`: guarded `FromTimedelta()` against `ttl is None` (previously unreachable because the outer gate prevented `ttl=None` from reaching this branch at all; now that the gate is gone, it needs its own guard) — falls through to the default zero `Duration()` in that case.

## Verification
Traced all three cases (`None`, `timedelta(0)`, a finite value) through the new branch logic in isolation (a minimal stand-in mirroring the real `Duration`/`get_ttl_duration` control flow, since `protobuf` isn't installed in this environment and I didn't want to add it just to verify branch logic):

| ttl input | FeatureView branch | LabelView branch |
|---|---|---|
| `None` | zero `Duration` | zero `Duration` |
| `timedelta(0)` | zero `Duration` | zero `Duration` |
| `timedelta(days=10)` | 864000s `Duration` | 864000s `Duration` |

Both branches now resolve identically for all three cases, and confirmed via `feature_view.py`'s `from_proto` that a zero-value `Duration` round-trips to `timedelta(0)`, matching the codebase's existing "no ttl" convention.

## Test plan
- [x] Traced all three ttl cases through the new logic in isolation, for both `FeatureView` and `LabelView`.
- [ ] Didn't run this against feast's actual test suite / a live `feast apply` round-trip in this environment (`protobuf` not installed here, didn't want to pull in the dependency just to verify this). Flagging for CI/maintainer verification — happy to iterate if anything surfaces.

### DIFF
--- a/sdk/python/feast/infra/registry/registry.py
+++ b/sdk/python/feast/infra/registry/registry.py
@@ -597,9 +597,7 @@ class Registry(BaseRegistry):
                     try:
                         ttl_duration.FromTimedelta(updated_fv.ttl)
                     except (ValueError, OverflowError) as e:
-                        raise ValueError(
-                            f"Invalid TTL value: {updated_fv.ttl}"
-                        ) from e
+                        raise ValueError(f"Invalid TTL value: {updated_fv.ttl}") from e
                 existing_proto.spec.ttl.CopyFrom(ttl_duration)
         if hasattr(existing_proto.spec, "online") and hasattr(updated_fv, "online"):
             existing_proto.spec.online = getattr(updated_fv, "online")
--- a/sdk/python/feast/infra/registry/registry.py
+++ b/sdk/python/feast/infra/registry/registry.py
@@ -580,20 +580,22 @@ class Registry(BaseRegistry):
             existing_proto.spec.version = getattr(updated_fv, "version")
 
         # Configuration fields (FeatureView / LabelView TTL)
-        if (
-            hasattr(existing_proto.spec, "ttl")
-            and hasattr(updated_fv, "ttl")
-            and updated_fv.ttl
-        ):
+        # Note: don't gate this on `updated_fv.ttl` being truthy -- None and
+        # timedelta(0) are both the documented way to express "no ttl", and
+        # are falsy, so that check would silently drop the update exactly
+        # when a user clears an existing ttl.
+        if hasattr(existing_proto.spec, "ttl") and hasattr(updated_fv, "ttl"):
+            from google.protobuf.duration_pb2 import Duration
+
             if isinstance(updated_fv, FeatureView):
                 ttl_duration = updated_fv.get_ttl_duration()
-                if ttl_duration:
-                    existing_proto.spec.ttl.CopyFrom(ttl_duration)
+                existing_proto.spec.ttl.CopyFrom(
+                    ttl_duration if ttl_duration is not None else Duration()
+                )
             elif isinstance(updated_fv, LabelView):
-                from google.protobuf.duration_pb2 import Duration
-
                 ttl_duration = Duration()
-                ttl_duration.FromTimedelta(updated_fv.ttl)
+                if updated_fv.ttl is not None:
+                    ttl_duration.FromTimedelta(updated_fv.ttl)
                 existing_proto.spec.ttl.CopyFrom(ttl_duration)
         if hasattr(existing_proto.spec, "online") and hasattr(updated_fv, "online"):
             existing_proto.spec.online = getattr(updated_fv, "online")
--- a/sdk/python/feast/infra/registry/registry.py
+++ b/sdk/python/feast/infra/registry/registry.py
@@ -19,6 +19,7 @@ from threading import Lock
 from typing import Any, Dict, List, Optional, Union
 from urllib.parse import urlparse
 
+from google.protobuf.duration_pb2 import Duration
 from google.protobuf.internal.containers import RepeatedCompositeFieldContainer
 from google.protobuf.message import Message
 
@@ -585,8 +586,6 @@ class Registry(BaseRegistry):
         # are falsy, so that check would silently drop the update exactly
         # when a user clears an existing ttl.
         if hasattr(existing_proto.spec, "ttl") and hasattr(updated_fv, "ttl"):
-            from google.protobuf.duration_pb2 import Duration
-
             if isinstance(updated_fv, FeatureView):
                 ttl_duration = updated_fv.get_ttl_duration()
                 existing_proto.spec.ttl.CopyFrom(
@@ -595,7 +594,12 @@ class Registry(BaseRegistry):
             elif isinstance(updated_fv, LabelView):
                 ttl_duration = Duration()
                 if updated_fv.ttl is not None:
-                    ttl_duration.FromTimedelta(updated_fv.ttl)
+                    try:
+                        ttl_duration.FromTimedelta(updated_fv.ttl)
+                    except (ValueError, OverflowError) as e:
+                        raise ValueError(
+                            f"Invalid TTL value: {updated_fv.ttl}"
+                        ) from e
                 existing_proto.spec.ttl.CopyFrom(ttl_duration)
         if hasattr(existing_proto.spec, "online") and hasattr(updated_fv, "online"):
             existing_proto.spec.online = getattr(updated_fv, "online")
--- a/sdk/python/tests/unit/infra/registry/test_update_metadata_fields.py
+++ b/sdk/python/tests/unit/infra/registry/test_update_metadata_fields.py
@@ -1,0 +1,53 @@
+"""Unit tests for Registry._update_metadata_fields TTL handling (issue #6703).
+
+Re-applying a FeatureView with its ttl cleared to ``None`` or
+``timedelta(0)`` (both the documented ways to express "no ttl") used to be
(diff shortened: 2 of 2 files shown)


sdk/python/feast/infra/registry/registry.py
```diff
@@ -580,20 +580,22 @@ def _update_metadata_fields(
...
-            hasattr(existing_proto.spec, "ttl")
-            and hasattr(updated_fv, "ttl")
-            and updated_fv.ttl
-        ):
+        # Note: don't gate this on `updated_fv.ttl` being truthy -- None and
+        # timedelta(0) are both the documented way to express "no ttl", and
+        # are falsy, so that check would silently drop the update exactly
+        # when a user clears an existing ttl.
+        if hasattr(existing_proto.spec, "ttl") and hasattr(updated_fv, "ttl"):
+            from google.protobuf.duration_pb2 import Duration
```
jyejare: **[Suggestion] Consider moving import to top of file**

The Duration import is now used in both branches of the conditional. Moving it to the top of the file would be more conventional and slightly more efficient.

**Suggested:**
```suggestion
# Move this import to the top of the file with other imports
from google.protobuf.duration_pb2 import Duration
```

sdk/python/feast/infra/registry/registry.py
```diff
@@ -580,20 +580,22 @@ def _update_metadata_fields(
...
-                    existing_proto.spec.ttl.CopyFrom(ttl_duration)
+                existing_proto.spec.ttl.CopyFrom(
+                    ttl_duration if ttl_duration is not None else Duration()
+                )
             elif isinstance(updated_fv, LabelView):
-                from google.protobuf.duration_pb2 import Duration
-
                 ttl_duration = Duration()
-                ttl_duration.FromTimedelta(updated_fv.ttl)
+                if updated_fv.ttl is not None:
```
jyejare, line 596: **[Suggestion] Consider error handling for invalid timedelta**

FromTimedelta could potentially raise an exception for invalid timedelta values. Consider adding basic validation or error handling to provide better user feedback.

**Suggested:**
```suggestion
ttl_duration = Duration()
if updated_fv.ttl is not None:
    try:
        ttl_duration.FromTimedelta(updated_fv.ttl)
    except (ValueError, OverflowError) as e:
        raise ValueError(f"Invalid TTL value: {updated_fv.ttl}") from e
```

jyejare (approved):
LGTM, some nitpicks.

saket3395:
Thanks for the review, @jyejare! Addressed both: moved the `Duration` import to the top of the file with the other `google.protobuf` imports, and wrapped `FromTimedelta()` in a try/except that re-raises as a `ValueError` naming the offending value.

ntkathole:
@saket3395 can we have test coverage for this ?

saket3395:
@ntkathole added unit coverage in `sdk/python/tests/unit/infra/registry/test_update_metadata_fields.py`:

- clearing a finite ttl to `None` or `timedelta(0)` now writes a zero `Duration` (parametrized over both) — this is the regression that was silently dropped before
- a finite→finite ttl update is preserved

Since `_update_metadata_fields` uses no instance state, the test exercises it directly via the class, mirroring the `FeatureView`/`FileSource`/`Field` construction used elsewhere in the unit tests. I wasn't able to run the suite in my local environment, so I'd appreciate CI confirming it — happy to adjust if anything needs tweaking.

ntkathole:
@saket3395 Please sign the commit and fix the linting checks

saket3395:
@ntkathole done — all commits are now DCO signed-off (DCO check is green), and I fixed the ruff format issue (the multi-line `raise ValueError` in the ttl block collapses to a single line under the 88-char limit). Ran `ruff check` and `ruff format --check` locally on both the changed source and the new test file — all clean now. Let me know if anything else is needed.

saket3395:
Fixed the `lint-pr` failure too — the "Validate PR title" job wanted a conventional-commit type, so I changed the title from `Fix:` to lowercase `fix:` (commitlint flagged `type-case` + `type-enum`). The re-run is showing `action_required` (pending maintainer approval to run on this fork PR) — should go green once it's approved or re-triggered.

ntkathole (approved):

PR #6712, opened 2026-08-07
fix: Fix mypy TorchTensor type alias error

### LINKED ISSUE #5563: torch tensor type error with mypy
## Current Behavior
feast/online_response.py:104: error: Variable "feast.online_response.TorchTensor" is not valid as a type  [valid-type]

### Specifications

- Version:
- Platform:
- Subsystem:

nithin42:
In `online_response.py`, `TorchTensor` is defined using `TypeAlias` inside both conditional branches of `if TYPE_CHECKING:`. Mypy complains: `Variable "feast.online_response.TorchTensor" is not valid as a type` because it parses it as a variable inside the runtime fallback branch.
This PR fixes the typing syntax in `online_response.py` by defining `TorchTensor` cleanly without `TypeAlias` in the fallback branch so mypy parses the type correctly.

Fixes #5563

## Checks

### DIFF
--- a/sdk/python/feast/online_response.py
+++ b/sdk/python/feast/online_response.py
@@ -25,7 +25,6 @@ from feast.type_map import feast_value_type_to_python_type
 from feast.value_type import ValueType
 
 if TYPE_CHECKING:
-    import torch
     from torch import Tensor as TorchTensor
 else:
     TorchTensor = Any
--- a/sdk/python/feast/online_response.py
+++ b/sdk/python/feast/online_response.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import uuid as uuid_module
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, TypeAlias, Union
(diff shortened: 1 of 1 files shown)


sdk/python/feast/online_response.py
```diff
@@ -26,10 +26,9 @@
 
 if TYPE_CHECKING:
     import torch
```
jyejare: This must not be needed anymore.

jyejare (approved):
LGTM, just nitpick.

nithin42:
Good catch — removed the redundant `import torch` in commit [`2704e66`](https://github.com/nithin42/feast/commit/2704e6647e29c11d79a0d1f828e88042373a060e). Since `TorchTensor` is the only torch symbol used in type annotations, the standalone module import was indeed no longer needed after the fix.

jyejare (approved):

nithin42:
@jyejare @ntkathole @franciscojavierarceo — the nitpick is fixed and DCO is green. Could you please approve the workflow runs for this PR so CI can finish and it can be merged? Thank you!

Issue #6389, opened 2026-05-09
Proposal: Version-qualified feature refs in get_historical_features (offline store versioning)


Abhishek8108:
## Summary

Feature view versioning (#6101) added `@v<N>` syntax for version-qualified online reads and `feast materialize --version v<N>` for writing to version-specific online tables. The offline retrieval path (`get_historical_features`) has no equivalent: version qualifiers in feature refs are silently stripped during ibis prefix matching but never used to route to a different source table.

This means a team that pins their online serving to `driver_stats@v2` cannot generate training data from the same v2 schema — breaking training-serving consistency, which is one of the primary motivations for versioning.

## The gap

**Online store (works today):**
```python
store.get_online_features(features=["driver_stats@v2:trips_today"], ...)
# → routes to version-specific table: {project}_driver_stats_v2
```

**Offline store (silently ignores version today):**
```python
store.get_historical_features(entity_df=..., features=["driver_stats@v2:trips_today"])
# → routes to the same source table regardless of @v2
# → version tag stripped in _parse_feature_ref / ibis prefix matching
```

## Design questions

I'd like to get your input on the right approach before writing any code, since there are a few viable directions with different tradeoffs:

**Option A — Version-specific source tables (mirrors online store pattern)**
When `enable_online_feature_view_versioning` is enabled and a versioned ref is used, route to a version-suffixed offline source table (e.g. `{project}_driver_stats_v2`). Requires that versioned offline tables exist — either pre-created by an upstream pipeline or written by a future `feast materialize-offline --version v2`.

Pros: mirrors online store semantics exactly, strong training-serving consistency guarantee.
Cons: offline source data is usually externally managed; teams would need to maintain versioned source tables alongside their online tables, which may be impractical.

**Option B — Schema-level version resolution (read from current table, project v<N> schema)**
Keep reading from the same source table, but apply the schema of the pinned version when selecting columns. Useful when the underlying data hasn't changed but the feature view's field definitions have evolved.

Pros: no need to maintain separate offline tables, works with externally managed sources.
Cons: weaker guarantee — if the source data itself changed between versions, this doesn't help.

**Option C — Raise `VersionedOfflineReadNotSupported` (explicit error instead of silent ignore)**
At minimum, detect `@v<N>` refs in `get_historical_features` and raise a clear error rather than silently stripping the version tag. This prevents silent training-serving skew today while the longer-term design is decided.

## Questions for you

[...]

PR #6718, opened 2026-08-09
feat: Version-pinning for FeatureService (online + offline)

### LINKED ISSUE #6717: FeatureService cannot pin a specific FeatureView/OnDemandFeatureView version
### Problem

A `FeatureService` exists to freeze the exact feature definitions a model was
trained and served on. Today it cannot pin a specific feature view **version** —
it always resolves to the active (promoted) version. Feast's own docs name this
limitation verbatim on the [alpha feature-view-versioning page](https://docs.feast.dev/reference/alpha-feature-view-versioning):

> **Feature services** — Feature services always resolve to the active (promoted)
> version. `--no-promote` versions are not served until promoted.

Building a service from a version-pinned `FeatureView` silently falls back to the
promoted version:

```python
fv  = FeatureView(name="driver_stats", version="v2", ...)
svc = FeatureService(name="model_v1", features=[fv[["conv_rate"]]])
svc.feature_view_projections[0].version_tag    # -> None   (should be 2)
svc.feature_view_projections[0].name_to_use()  # -> "driver_stats"  (should be "driver_stats@v2")
```

### Root cause

1. `FeatureService.__init__` appends the source view's projection without
   translating its `version` string into `projection.version_tag` (the field
   `name_to_use()` checks to render `fv@v2`).
2. `utils._get_feature_views_to_use` hard-codes the version as `None` for the
   `FeatureService` branch, so retrieval always falls through to the promoted
   snapshot via `get_any_feature_view`.

While fixing this, a second pre-existing bug surfaced: **offline** retrieval of a
version-pinned `OnDemandFeatureView` fails. Offline stores re-fetch ODFVs from the
registry via an unversioned `registry.list_on_demand_feature_views`, so a pinned
ODFV ref (`odfv@v1:feat`) raises `ValueError: Could not find feature view from
reference ...` or silently drops the ODFV's features. Every offline backend is
affected (they funnel through `OnDemandFeatureView.get_requested_odfvs` or
`offline_utils.get_feature_view_query_context`, plus dask's inline duplicate).

### Proposal

- Capture the pinned version into the service's projection, and read it back at
  retrieval time (fixes both online and offline for plain FeatureViews, which
  share `_get_feature_views_to_use`).
- Add a clean pinning API: allow plain string refs as `FeatureService.features`
  entries (`"driver_stats@v2:trips_today"`), resolved against the registry at
  `feast apply` time — no importing/reconstructing historical view objects.
- Fix offline ODFV resolution with a shared version-aware helper.

Gated by the existing `registry.enable_online_feature_view_versioning` flag,
which (despite its name) gates both online and offline versioned resolution.

HSarwat:
Closes #6717.
Related to #6389 (feature versioning in offline retrieval).

### What

Lets a `FeatureService` pin a specific historical FeatureView / OnDemandFeatureView version, honored by both `get_online_features` and `get_historical_features`. Feast's docs currently list this as a known limitation ("Feature services always resolve to the active (promoted) version"); this closes that gap.

### Commits

1. **fix: Pin FeatureView version in FeatureService projections** — stamp `projection.version_tag` from the source view's `version` in `FeatureService.__init__`; read `projection.version_tag` (instead of `None`) in `utils._get_feature_views_to_use`. `version="latest"` is unaffected.
2. **feat: Support string feature references in FeatureService** — `features` accepts `"<fv>[@<version>][:<feature>]"` strings, resolved once in `FeatureStore.apply` via `FeatureService.resolve_pending_refs`. Factors `_parse_feature_or_view_ref` out of `_parse_feature_ref` (optional `:feature`).
3. **fix: Resolve version-pinned OnDemandFeatureViews in offline retrieval** — shared `utils._get_requested_on_demand_feature_views` used by all three unversioned ODFV call sites; pinned ODFV refs now resolve to the pinned snapshot offline. This is the offline-retrieval versioning gap tracked in #6389.
4. **docs: Document FeatureService version pinning** — alpha versioning page + `enable_online_feature_view_versioning` docstring clarified to note it gates offline too; Known Limitations updated.

### Problem

Building a service from a version-pinned `FeatureView` silently falls back to the promoted version:

```python
fv  = FeatureView(name="driver_stats", version="v2", ...)
svc = FeatureService(name="model_v1", features=[fv[["conv_rate"]]])
svc.feature_view_projections[0].version_tag    # -> None   (should be 2)
svc.feature_view_projections[0].name_to_use()  # -> "driver_stats"  (should be "driver_stats@v2")
```

Two bugs cause this: `FeatureService.__init__` never translates the view's `version` string into `projection.version_tag`, and `utils._get_feature_views_to_use` hard-codes the version as `None` for the `FeatureService` branch. A related pre-existing bug (#6389) breaks **offline** retrieval of version-pinned `OnDemandFeatureView`s: offline stores re-fetch ODFVs via an unversioned `registry.list_on_demand_feature_views`, so a pinned ODFV ref raises `ValueError: Could not find feature view from reference ...` or silently drops features across every offline backend.

### Design notes for reviewers

- **Priority ordering (commit 2):** a pinned string ref always resolves from `registry.get_feature_view_by_version`, never from `fvs_to_update` (the apply batch only ever holds "latest" objects and would substitute the wrong schema). Covered by `test_pinned_ref_ignores_fvs_to_update`.
- **Backward compatible:** unversioned refs / `version="latest"` behave exactly as before; every existing (unversioned) FeatureService is unaffected.
- **Flag name:** `enable_online_feature_view_versioning` is misleadingly scoped to "online" — it gates both paths. Clarified in the docstring/docs here rather than renamed (rename would be a breaking config change; happy to follow up if maintainers prefer).

### Tests

New unit tests: `test_feature_service_versioning.py` (object + string-ref pinning, proto round-trip, backward compat, error guards, priority ordering) and `test_odfv_offline_versioning.py` (version-aware ODFV resolution, v0 fallback, non-ODFV skip, dedup). All pass; ruff (0.16.0) + mypy clean on changed files.


ntkathole:
@HSarwat Please fix CI

HSarwat:
@ntkathole This PR doesnt introduce any dependency change, yet the CI is failing because it cant resolve greenlet 3.5.5 . I am not really sure where the error is coming from.

Rebased my PR with the latest changed from main in case this error was resolved in a later commit and to trigger the CI again

haoxu0 (changes requested):
String refs introduce an unresolved lifecycle state that retrieval and serialization do not handle. A freshly constructed `FeatureService(name="svc", features=["driver_stats@v2:conv_rate"])` has no projections until apply/plan. Passing that object directly to `get_online_features` or `get_historical_features` is a supported Feast pattern, but `_get_features()` consults the registered service while `_get_feature_views_to_use()` consumes the unresolved passed object, producing inconsistent resolution; `to_proto()` before apply also silently serializes an empty service.

Please either resolve the passed service from the registry in all retrieval paths, or reject unresolved services with a clear error. Add an integration-style test that reconstructs the string-based FeatureService in a fresh process after apply and passes that object to both online and historical retrieval.

HSarwat:
@HaoXuAI 

I resolved the passed FeatureService from the registry in _get_feature_views_to_use when it has unresolved string refs, so it matches what _get_features already does and both retrieval paths agree instead of one seeing empty projections. 

Also added an integration test in test_versioning.py that reconstructs the string-ref service in a fresh store after apply and passes it to both online and historical retrieval.

haoxu0 (approved):
LGTM

sdk/python/feast/feature_service.py
```diff
@@ -95,8 +109,80 @@ def __init__(
...
+                # FeatureView(version="v2")), stamp that version onto the
+                # projection so name_to_use() renders "fv@v2" and retrieval
+                # resolves the pinned snapshot. The default version ("latest")
+                # leaves version_tag as None, preserving existing behavior for
+                # every unversioned feature service.
+                fv_version = getattr(feature_grouping, "version", None)
+                if projection.version_tag is None and fv_version:
+                    is_latest, version_num = parse_version(fv_version)
+                    if not is_latest:
+                        projection.version_tag = version_num
```
ntkathole, line 128: This mutates the source FeatureView's own projection. `FeatureView(version="v2")` does not set `projection.version_tag` itself, so after `FeatureService(features=[fv])` the caller's `fv.projection.version_tag` becomes `2`.

Online table naming (`compute_versioned_name`) reads that field, so a later materialize/get on the same object can unexpectedly hit `*_v2`. The string-ref path already `copy.copy`s for this reason — please copy here too before stamping `version_tag`.

sdk/python/feast/feature_store.py
```diff
@@ -1226,6 +1226,11 @@ def _make_inferences(
             ]
         }
         for feature_service in feature_services_to_update:
+            # Resolve string feature refs (e.g. "driver_stats@v2") before
+            # inference. No-op for object-only services.
+            feature_service.resolve_pending_refs(
```
ntkathole, line 1275: `fvs_to_update_map` only includes `odfvs_to_write` (`write_to_online_store=True`). A normal ODFV is absent, and `_make_inferences` runs before `apply_feature_view`, so the registry lookup fails too.

The docs recommend string refs, but this fails:
```python
odfv = on_demand_feature_view(...)(src_fv)
svc = FeatureService(name="svc", features=["my_odfv"])
store.apply([src_fv, odfv, svc])  # FeatureViewNotFound
```
Object form `features=[odfv]` still works. Please pass `odfvs_to_update` (not just `odfvs_to_write`) into `resolve_pending_refs`.

sdk/python/feast/feature_service.py
```diff
@@ -95,8 +109,80 @@ def __init__(
...
+
+    def resolve_pending_refs(
+        self,
+        project: str,
+        registry: "BaseRegistry",
+        fvs_to_update: Optional[Dict[str, Union[FeatureView, BaseFeatureView]]] = None,
+    ) -> None:
+        """Resolve string feature refs (see ``__init__``) into projections.
+
+        Called automatically by ``FeatureStore.apply``/``plan`` so the pin is
```
ntkathole, line 139: Docstring says apply/plan resolve these refs, but `prepare_for_apply` does not call `resolve_pending_refs`, and its early return is `if self._features and all(p.features for p in self.feature_view_projections)`.

`all([])` is `True`, so a service with only string refs and empty projections returns immediately. `to_proto()` then serializes zero features.

`FeatureStore.apply` happens to call this first, so the happy path is OK. Direct `registry.apply_feature_service(...)` or any proto round-trip before apply will persist an empty service with no error.

Please reject (or resolve) unresolved `_pending_feature_refs` in `prepare_for_apply` and `to_proto()`.

ntkathole (changes requested):
Thanks for the follow-up on the unresolved string-ref retrieval path — that looks correct.

A few remaining issues that I think should land before merge:

1. Object-based pinning mutates the source FeatureView's projection (string-ref path already copies).
2. Recommended string refs cannot resolve an ODFV created in the same `apply` (`fvs_to_update_map` only includes `odfvs_to_write`).
3. `prepare_for_apply` / `to_proto()` still treat an unresolved string-ref service as a valid empty service (`all([])` is `True`).

Inline comments have details. Would also like one integration test that actually pins v0 vs a later v1 and asserts retrieval, not only the unversioned reconstructed-service path.

sdk/python/feast/feature_service.py
HSarwat: Copied

sdk/python/feast/feature_service.py
HSarwat: Resolved during prepare_for_apply, raised error for to_proto

sdk/python/feast/feature_store.py
HSarwat: Built a second map, fvs_to_resolve_map, that includes the full odfvs_to_update list (all ODFVs in the apply, write-enabled or not), and passed that to resolve_pending_refs. I left infer_features on the original fvs_to_update_map so inference/online-write behavior is untouched.

HSarwat:
@ntkathole Made 4 new commits, each for the three changes you requested + the integration test.

ntkathole:
@HSarwat Please sign the commit and handle the CI failure, the object-path copy introduced a regression for schema-less FeatureViews used as a feature subset in a FeatureService

HSarwat:
@ntkathole Can you check the CI now? I ran the unit and integration tests, at least the ones that I could, locally, and should be working

HSarwat:
@ntkathole I've signed off on the commits, fixed branch history to be rebased with main, tested latest changes from main locally, both unit and integration, and everything passed. Can you do one final check, and if everything is working we can merge this PR

ntkathole (approved):

PR #6722, opened 2026-08-10
fix: Prevent mutation of cached feature resolution results

junwoo-choi-rapportlabs:
`_get_cached_request_context()` caches `grouped_refs`, including the mutable
requested-feature lists nested inside it.

The Redis online store appends an internal timestamp key to these lists while
preparing HMGET fields. Because cache hits returned the same nested lists,
timestamp keys could accumulate across requests, increasing Redis fields and
response post-processing work over time.

This change copies each requested-feature list at the request preparation
boundary before passing it to an online store. The cached resolution remains
shared, while downstream mutations become request-local.

image

N/A

# Checks

## Testing Strategy

- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests

### DIFF
--- a/sdk/python/feast/utils.py
+++ b/sdk/python/feast/utils.py
@@ -1568,6 +1568,13 @@ def _prepare_entities_to_read_from_online_store(
         entityless_case,
     ) = _get_cached_request_context(registry, project, features, full_feature_names)
 
+    # Online stores may append internal fields to the requested feature lists.
+    # Keep those request-local so the cached resolution result remains unchanged.
+    grouped_refs = [
+        (feature_view, list(requested_features))
+        for feature_view, requested_features in grouped_refs
+    ]
+
     # Mutable copy — downstream code adds join keys to this set.
     requested_result_row_names = set(requested_result_row_names_frozen)
 
--- a/sdk/python/tests/unit/test_feature_resolution_cache.py
+++ b/sdk/python/tests/unit/test_feature_resolution_cache.py
@@ -6,6 +6,7 @@ from unittest.mock import MagicMock, patch
 import pytest
 
 import feast.utils as utils
+from feast.protos.feast.types.Value_pb2 import Value as ValueProto
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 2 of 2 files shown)


sdk/python/feast/utils.py
```diff
@@ -1568,6 +1568,13 @@ def _prepare_entities_to_read_from_online_store(
         entityless_case,
     ) = _get_cached_request_context(registry, project, features, full_feature_names)
 
+    # Online stores may append internal fields to the requested feature lists.
+    # Keep those request-local so the cached resolution result remains unchanged.
+    grouped_refs = [
+        (feature_view, list(requested_features))
+        for feature_view, requested_features in grouped_refs
+    ]
```
jyejare, line 1576: **[Suggestion] Consider using copy.deepcopy for more robust copying**

While the current list comprehension works for this case, using copy.deepcopy would be more explicit about the intent and handle any future changes to the data structure more robustly.

**Suggested:**
```suggestion
    import copy
    grouped_refs = copy.deepcopy(grouped_refs)
```

sdk/python/tests/unit/test_feature_resolution_cache.py
```diff
@@ -140,6 +141,36 @@ def test_returned_result_row_names_is_mutable_copy(self, mock_ctx):
...
+            "proj",
+            ["fv:feat1"],
+            entity_values,
+            native_entity_values=False,
+        )
+        second_requested_features = second_result[1][0][1]
+
+        assert second_requested_features == ["feat1"]
+        assert second_requested_features is not first_requested_features
+        mock_ctx.assert_called_once()
```
jyejare: **[Suggestion] Add assertion for first_requested_features content**

The test verifies the second request wasn't affected but doesn't assert that the first request's mutation was successful. Adding this assertion would make the test more complete.

**Suggested:**
```suggestion
+        assert first_requested_features == ["feat1", "_ts:fv"]
+        assert second_requested_features == ["feat1"]
+        assert second_requested_features is not first_requested_features
+        mock_ctx.assert_called_once()
```

jyejare (approved):
This PR fixes a caching bug where mutations to feature lists were leaking between cached requests. The fix creates defensive copies of grouped_refs to prevent shared mutable state issues. The implementation is correct and includes comprehensive test coverage.

sdk/python/feast/utils.py
junwoo-choi-rapportlabs: I am not sure if a deep copy is necessary specifically for this issue. I will leave the decision to you.

jyejare: Not necessary but would be future proof. Its a non blocking comment.

shuchu:
The DCO warning needs to be handled. please check the warning message.

junwoo-choi-rapportlabs:
> The DCO warning needs to be handled. please check the warning message.

My bad. corrected it.

ntkathole (approved):

PR #6726, opened 2026-08-11
fix: Remote feastRef FeatureStore fails first apply for a new feastProject

ntkathole:

When a FeatureStore CR uses `spec.services.registry.remote.feastRef` to share another CR’s registry and sets a **new** `feastProject`, `feast apply` fails during init with:

`feast.errors.ProjectObjectNotFoundException: Project does not exist`

The remote reference itself is fine (registry hostname resolves). The failure is on the registry server during first apply of a project that does not exist yet.

## Expected behavior
Remote `feastRef` should:
1. Not deploy a local registry for the secondary CR
2. Reuse the primary registry
3. Allow the secondary CR’s own `feastProject` to be created on that shared registry on first apply

## Actual behavior
`feast apply` → `plan()` fails before the project can be registered.

### Failure sequence (after partial fix)
1. `Refresh(project)` — originally failed on hard `get_project` + `UPDATE`
2. After Refresh fix: `GetInfra(project)` — same pattern (`get_project` + `DESCRIBE`)
3. `UpdateInfra` / `Commit` also had overly strict project existence / all-projects `UPDATE` checks

## Root cause
RBAC hardening on registry-server RPCs calls `get_project` **before** apply can create the project:

- `Refresh` / `GetInfra` / `UpdateInfra` assume the project already exists
- Local registry apply can auto-create missing projects; remote path cannot get past these RPCs

### DIFF
--- a/sdk/python/feast/registry_server.py
+++ b/sdk/python/feast/registry_server.py
@@ -1376,10 +1376,12 @@ class RegistryServer(RegistryServer_pb2_grpc.RegistryServerServicer):
         return Empty()
 
     def UpdateInfra(self, request: RegistryServer_pb2.UpdateInfraRequest, context):
-        project = self.proxied_registry.get_project(
-            name=request.project, allow_cache=True
+        # Create-or-update so first remote apply can write infra for a new project.
+        assert_permissions_to_update(
+            resource=Project(name=request.project),
+            getter=self.proxied_registry.get_project,
+            project=request.project,
         )
-        assert_permissions(resource=project, actions=[AuthzedAction.UPDATE])
         self.proxied_registry.update_infra(
             infra=Infra.from_proto(request.infra),
             project=request.project,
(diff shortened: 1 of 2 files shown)


sdk/python/feast/registry_server.py
```diff
@@ -1388,10 +1390,19 @@ def UpdateInfra(self, request: RegistryServer_pb2.UpdateInfraRequest, context):
...
+        # plan() calls get_infra before the project is created on a shared remote
+        # registry. Mirror ListProjectMetadata: authorize when present, and for a
+        # missing project require CREATE (or allow when auth is off).
+        try:
+            project = self.proxied_registry.get_project(
+                name=request.project, allow_cache=True
+            )
+            assert_permissions(resource=project, actions=[AuthzedAction.DESCRIBE])
+        except FeastObjectNotFoundException:
+            assert_permissions(
```
jyejare, line 1402: Why dont we assert create permission first rather than splitting it in try except ?

jyejare (commented):
LGTM, pending question.

sdk/python/feast/registry_server.py
ntkathole: first we need to make sure project exists already or not - request can be for create new or update existing - user may have describe permission but not create if project exists already, in that case, check should not fail with create perm

jyejare: But cant we have have/define that checks if user have any of describe/update/create permission? I mean create a helper function that checks for either of these permissions ?

ntkathole: There isn't any repeated pattern - just 2 different choices within different functions - so inline make sense. 

GetInfra - either project exists or return as not found
UpdateInfra - either projects exists or create new one

jyejare (approved):

PR #6731, opened 2026-08-12
docs: Fix broken GitHub star history chart in README

OctoBored:
The GitHub star history chart shown near the bottom of the README is currently broken due to restrictions on the GitHub stargazer API, so the chart no longer renders for readers. This change points the chart (and its surrounding link) at a working alternative so the star history is visible again. The underlying SVG chart endpoint and the hash-based link are both updated consistently in the README and the template used to generate it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,11 +283,11 @@ Feast is a community project and is still under active development. Please have 
 
 ## 🌟 GitHub Star History
 <p align="center">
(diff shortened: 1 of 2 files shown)


ntkathole:
@OctoBored please sign the commit to fix DCO check

shuchu (approved):
lgtm

PR #6733, opened 2026-08-13
fix: Resolve kserve with pip --dry-run instead of installing it

### LINKED ISSUE #6732: Unit test installs kserve into the live interpreter, causing nondeterministic failures in unrelated tests
## Expected Behavior

Running the unit suite leaves the environment it ran in unchanged, and a test's outcome does not depend on what other tests are doing concurrently.

## Current Behavior

`sdk/python/tests/unit/infra/test_dependency_conflicts.py` installs a package into the interpreter that is running the suite:

```python
command = [sys.executable, "-m", "pip", "install", "kserve==0.15.2"]
```

There is no isolation and no cleanup, so this mutates the environment every other test is using. Two consequences.

**Unrelated tests fail nondeterministically in the same run.** `pyproject.toml` pins `psutil==5.9.0`; kserve 0.15.2 requires `psutil<6.0.0,>=5.9.6`. pip therefore cannot leave the installed version in place — it uninstalls psutil, then installs 5.9.8:

```
Attempting uninstall: psutil
  Found existing installation: psutil 5.9.0
  Uninstalling psutil-5.9.0:
    Successfully uninstalled psutil-5.9.0
Installing collected packages: … psutil …
```

`make test-python-unit` runs `pytest -n 8`, so eight workers share one environment. Any test running during that window which imports psutil, directly or through a subprocess, fails. `feast/metrics.py` imports psutil at module scope, and `feast/cli/cli.py` reaches it through `feast/__init__.py`, so every test that shells out to the CLI is exposed.

Observed on an unrelated docs PR, in `unit-test-python (3.12, ubuntu-latest)`:

```
FAILED sdk/python/tests/unit/cli/test_cli.py::test_3rd_party_providers
AssertionError: Expected <b'Traceback (most recent call last):
  File ".../feast/cli/cli.py", line 27, in <module>
    from feast import utils
  ...
  File ".../feast/metrics.py", line 57, in <module>
    import psutil
ModuleNotFoundError: No module named \'psutil\'
'> to contain item <b"Provider \'feast123\' is not implemented">
```

In that job `test_install_kserve_with_feast` completed at 12:25:54 and the failure landed at 12:29:15. Because it depends on worker timing, a re-run usually passes, so it presents as flake and the blame lands on whichever PR happens to be running.

**A second consecutive local run fails.** kserve also pulls protobuf down to 4.25.x, and the installed `grpcio-health-checking` ships protobuf 6.x generated code that imports `google.protobuf.runtime_version`. The first run still passes, because pytest imports every module during collection before the mid-run install lands; the next run fails at collection with `ImportError: cannot import name 'runtime_version' from 'google.protobuf'`. CI does not see this because its environments are ephemeral.

## Steps to reproduce

```
make test-python-unit    # passes
make test-python-unit    # collection errors on runtime_version
```

For the race, run the suite with `-n 8` and observe that any test shelling out to the CLI can fail while the kserve install is between uninstall and install.

### Specifications

- Version: master
- Platform: any; observed on `ubuntu-latest` Python 3.12 in CI and locally on macOS

## Possible Solution

The test checks that kserve and feast can be resolved together. `pip` can answer that without installing anything, using `--dry-run`. Verified both directions:

- no conflict: exits 0, performs full resolution (`Would install … psutil-5.9.8 …`), and leaves the environment untouched — psutil stays at 5.9.0 and kserve is not importable afterwards
- real conflict: `pip install --dry-run kserve==0.15.2 psutil==5.9.0` exits 1 with `Cannot install kserve==0.15.2 and psutil==5.9.0 because these package versions have conflicting dependencies`

So the detection is unchanged and the mutation disappears. Happy to open a PR for this.

Separately, the current assertion looks inverted:

```python
conflict_occured = "dependency conflicts" in err and "ERROR" not in err
```

That is only true when pip reports conflicts *without* an error, so a run that fails loudly sets it to `False`. The `exit_code == 0` assertion above it is what actually catches failures today.

larrysingleton007:

`test_install_kserve_with_feast` installs kserve into the interpreter running the unit suite, with no isolation and no cleanup, so it mutates the environment every other test is using.

Feast pins `psutil==5.9.0` and kserve 0.15.2 requires `psutil<6.0.0,>=5.9.6`, so pip cannot leave the installed version in place — it uninstalls psutil, then installs 5.9.8:

```
Attempting uninstall: psutil
  Found existing installation: psutil 5.9.0
  Uninstalling psutil-5.9.0:
    Successfully uninstalled psutil-5.9.0
Installing collected packages: … psutil …
```

`make test-python-unit` runs `pytest -n 8`, so eight workers share one environment. Any test importing psutil during that window fails, including every test that shells out to the CLI, since `feast/metrics.py` imports it at module scope. That is how an unrelated docs PR picked up a red `unit-test-python (3.12, ubuntu-latest)` on `test_3rd_party_providers` with `ModuleNotFoundError: No module named 'psutil'`. Because it depends on worker timing, a re-run usually passes and it reads as flake.

kserve also pulls protobuf down to 4.25.x, which leaves the environment broken for the *next* run: the installed `grpcio-health-checking` ships protobuf 6.x generated code importing `google.protobuf.runtime_version`. The first run passes, because collection imports precede the mid-run install; the second fails at collection.

`--dry-run` performs the same resolution and still exits non-zero when the versions cannot be satisfied together, which is what the test guards against, without installing anything.

This also replaces the conflict assertion, which was inverted:

```python
conflict_occured = "dependency conflicts" in err and "ERROR" not in err
```

That is only true when pip reports conflicts *without* an error, so a run failing loudly set it to `False`. The `exit_code == 0` assertion was already doing the real work; it now carries pip's output for diagnosis.

Closes #6732

## Misc

Verified both directions, so the detection is unchanged rather than merely quieter:

- clean resolution exits 0, performs the full resolve (`Would install … psutil-5.9.8 …`), and leaves psutil at 5.9.0 with kserve not importable afterwards
- `pip install --dry-run kserve==0.15.2 psutil==5.9.0` exits 1 with `Cannot install kserve==0.15.2 and psutil==5.9.0 because these package versions have conflicting dependencies`

The full unit suite now passes twice in a row — 2529 passed, 20 skipped both times, with psutil still 5.9.0 and protobuf still 6.33.6 after the first. Previously the second run produced 31 failures and 8 collection errors. The test also drops from about 13s to under 2s.

This needs a `kind/` label, which I cannot add as an outside contributor.

### DIFF
--- a/sdk/python/tests/unit/infra/test_dependency_conflicts.py
+++ b/sdk/python/tests/unit/infra/test_dependency_conflicts.py
@@ -8,11 +8,29 @@ logger.setLevel(logging.DEBUG)
 
 class TestDependencyConflicts:
     def test_install_kserve_with_feast(self):
-        """Test installing KServe in the current environment where Feast is already installed.
-        Ensures no dependency conflict errors occur.
+        """Resolve KServe against the environment Feast is installed in.
+
+        ``--dry-run`` makes pip resolve the full dependency set and report what
+        it would install, without installing anything. Resolution is the part
+        this test cares about: pip exits non-zero when the versions cannot be
+        satisfied together, which is the conflict being guarded against.
+
+        Installing for real would mutate the interpreter running the suite.
+        Feast pins ``psutil==5.9.0`` while KServe requires ``psutil>=5.9.6``, so
+        pip has to uninstall psutil before reinstalling it, and the unit suite
+        runs under ``pytest -n 8`` against a single environment. Any test that
+        imports psutil during that window fails, including every test that
+        shells out to the CLI, since ``feast.metrics`` imports it at module
+        scope. It also left the environment inconsistent for the next run.
         """
-        # Command to install KServe
-        command = [sys.executable, "-m", "pip", "install", "kserve==0.15.2"]
+        command = [
+            sys.executable,
+            "-m",
+            "pip",
+            "install",
+            "--dry-run",
+            "kserve==0.15.2",
+        ]
 
         process = subprocess.Popen(
             command, stdout=subprocess.PIPE, stderr=subprocess.PIPE
@@ -26,7 +44,10 @@ class TestDependencyConflicts:
         logger.debug(out)
         logger.debug(err)
 
-        # Assertions
-        assert exit_code == 0
-        conflict_occured = "dependency conflicts" in err and "ERROR" not in err
-        assert not conflict_occured, "Dependency conflict detected during installation"
+        # pip reports an unsatisfiable set on stdout and exits non-zero, so the
+        # exit code is the assertion that matters; the message is for the
+        # failure output.
+        assert exit_code == 0, (
+            f"pip could not resolve kserve==0.15.2 against the current "
+            f"environment:\n{out}\n{err}"
+        )


ntkathole (approved):

larrysingleton007:
@ntkathole thanks for the review.

Now that `ok-to-test` triggered the full suite, the checks have settled at 26 passing, 3 skipped, and one failure: `mcp-feature-server-runtime`. That one is #6706, the repo-wide mcp pin breakage, not this PR. #6713 fixes it and is still open, currently red at the `Setup pixi` step because it changes `pyproject.toml` without a regenerated `pixi.lock`.

Worth noting the other five integration jobs pass here, including `integration-test-ray` and `integration-test-registration-ci`, which are the ones that share that pixi setup step. This PR touches only a test file, so nothing in the lock moves. That's a useful control: it confirms the pixi failures on #6713 come from its `pyproject.toml` edit rather than being pre-existing.

GitHub also reports this branch as BEHIND. Happy to merge master in if you'd like it current before merging, though I've left it alone since it's conflict-free and a merge commit would re-run everything for no change.

PR #6736, opened 2026-08-13
fix: Remove inert subjectaccessreviews and reorganize RBAC rules

ntkathole:
## Summary
Remove the unused subjectaccessreviews from the operator ClusterRole and reorganize the RBAC rules for clusterrolebindings and clusterroles to follow the principle of least privilege.

## Changes
- Remove subjectaccessreviews from both role.yaml and authz.go
- Separate clusterrolebindings with its own verb set
- Split clusterroles into general create and named resource management

## Test Plan
- [x] Unit tests passing

### DIFF
--- a/infra/feast-operator/config/rbac/role.yaml
+++ b/infra/feast-operator/config/rbac/role.yaml
@@ -163,10 +163,35 @@ rules:
   - rbac.authorization.k8s.io
   resources:
   - clusterrolebindings
+  verbs:
+  - create
+  - delete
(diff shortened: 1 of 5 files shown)


haoxu0 (approved):

haoxu0 (approved):
The RBAC split matches the controller's actual CreateOrUpdate and cleanup operations, and removing the inert subjectaccessreviews rules is correct. No blocking findings from my review.

PR #6749, opened 2026-08-18
fix: Regenerate operator bundle CSV to include missing RBAC permissions

ntkathole:
## Summary

- Regenerated the operator bundle CSV (`make bundle`) to sync it with the RBAC annotations in Go source and `config/rbac/role.yaml`. The CSV was missing permissions for `kubeflow.org/notebooks`, `apiextensions.k8s.io/customresourcedefinitions`, and several core resources — causing `feast-operator-controller-manager` to CrashLoopBackOff with `Forbidden` errors during DSC bootstrap.
- Added a new `operator-bundle-verify` CI job to `operator_pr.yml` that regenerates the bundle on every PR and fails if the committed CSV diverges from what `make bundle` produces.

## Root Cause

The `+kubebuilder:rbac` annotations in `notebook_configmap_controller.go` and the generated `config/rbac/role.yaml` correctly include `kubeflow.org/notebooks` (get/list/watch). However, `make bundle` was never run after these were added, so the CSV (which OLM uses to create the actual ClusterRole on the cluster) was stale.

This was a regression from the `rhoai-3.5` branch where the CSV was correctly populated.

## What changed

| File | Change |
|------|--------|
| `bundle/manifests/feast-operator.clusterserviceversion.yaml` | Regenerated — adds `kubeflow.org/notebooks`, `apiextensions.k8s.io/customresourcedefinitions`, splits resource blocks to match `role.yaml` |
| `bundle.Dockerfile` | Cosmetic — operator-sdk builder version label updated |
| `bundle/metadata/annotations.yaml` | Cosmetic — operator-sdk builder version label updated |
| `.github/workflows/operator_pr.yml` | New `operator-bundle-verify` job to prevent future CSV drift |

## Test plan

- [ ] Verify `make bundle` produces no diff on a clean checkout of this branch
- [ ] Deploy RHOAI with Feature Store enabled and confirm `feast-operator-controller-manager` starts without CrashLoopBackOff
- [ ] Confirm the new `operator-bundle-verify` CI job passes on this PR
- [ ] Confirm the CI job would fail if someone modifies RBAC annotations without running `make bundle`


ntkathole:
Closing - branch was based on wrong upstream. Recreating from correct base.

PR #6756, opened 2026-08-18
fix: detect duplicate feature view/data source names at parse time

### LINKED ISSUE #6417: Flaky test: test_cli_apply_duplicated_featureview_names fails with empty output on subprocess timeout

mayuriphad:
## Problem

`test_cli_apply_duplicated_featureview_names` (and related tests) intermittently fail in CI with:

```
AssertionError: assert (-1 != 0 and b'Feature view names must be case-insensitively unique' in b'')
```

Root cause: duplicate feature view/data source name detection happens deep inside `store.plan()`/`store.apply()` via `_validate_feature_views()`/`_validate_data_sources()` in `feature_store.py`, **after** `FeatureStore` and its dependencies (Dask, provider, registry) have already been initialized. On process shutdown, slow atexit handlers (Dask thread pool, PySpark JVM) can block long enough that the test harness's subprocess timeout fires and kills the process with `SIGKILL` before any output is flushed — so the real error message is lost and the test sees only an empty, non-zero exit.

The CLI also didn't handle `ConflictingFeatureViewNames`/`DataSourceRepeatNamesException` (both `FeastError` subclasses) at all — only `FeastProviderLoginError` was caught — so these errors propagated as unhandled tracebacks.

## Fix

1. **`sdk/python/feast/repo_operations.py`**: call the existing `_validate_feature_views()` / `_validate_data_sources()` validators (imported from `feature_store.py`, no logic duplicated) at the end of `parse_repo()`, right after the repo contents are collected — before any `FeatureStore`/Dask/PySpark initialization happens. This fails fast, close to where the actual problem (duplicate names in the repo) exists.
2. **`sdk/python/feast/cli/cli.py`**: add `except FeastError` handling (after the existing `FeastProviderLoginError` handler) in `plan_command` and `apply_total_command`, so any `FeastError` (including these) is surfaced as a clean message via `click.ClickException` with a non-zero exit code, rather than an unhandled traceback.

## Testing

Ran the directly affected tests locally (all pass):

```
tests/integration/cli/test_cli_apply_duplicates.py::test_cli_apply_duplicated_featureview_names PASSED
tests/integration/cli/test_cli_apply_duplicates.py::test_cli_apply_duplicate_data_source_names PASSED
tests/integration/cli/test_cli_apply_duplicates.py::test_cli_apply_imported_featureview PASSED
tests/integration/cli/test_cli_apply_duplicates.py::test_cli_apply_imported_featureview_with_duplication PASSED
tests/integration/cli/test_cli_apply_duplicates.py::test_cli_apply_duplicated_featureview_names_multiple_py_files PASSED
```

Fixes #6417

### DIFF
--- a/sdk/python/feast/cli/cli.py
+++ b/sdk/python/feast/cli/cli.py
@@ -52,7 +52,7 @@ from feast.cli.stream_feature_views import stream_feature_views_cmd
 from feast.cli.ui import ui
 from feast.cli.validation_references import validation_references_cmd
 from feast.constants import FEAST_FS_YAML_FILE_PATH_ENV_NAME
-from feast.errors import FeastProviderLoginError
+from feast.errors import FeastError, FeastProviderLoginError
 from feast.repo_config import load_repo_config
 from feast.repo_operations import (
     apply_total,
@@ -261,6 +261,8 @@ def plan_command(
         plan(repo_config, repo, skip_source_validation, skip_feature_view_validation)
     except FeastProviderLoginError as e:
         print(str(e))
+    except FeastError as e:
+        raise click.ClickException(str(e))
 
 
 @cli.command("apply", cls=NoOptionDefaultFormat)
@@ -319,6 +321,8 @@ def apply_total_command(
         )
     except FeastProviderLoginError as e:
         print(str(e))
+    except FeastError as e:
+        raise click.ClickException(str(e))
 
 
 @cli.command("teardown", cls=NoOptionDefaultFormat)
--- a/sdk/python/feast/feature_store.py
+++ b/sdk/python/feast/feature_store.py
@@ -1134,7 +1134,7 @@ class FeatureStore:
                 "This API is stable, but the functionality does not scale well for offline retrieval",
                 RuntimeWarning,
             )
-        _validate_feature_views(
+        validate_feature_views(
             [
                 *views_to_update,
                 *odfvs_to_update,
@@ -1425,7 +1425,7 @@ class FeatureStore:
                 desired_repo_contents.stream_feature_views,
                 desired_repo_contents.label_views,
             )
-        _validate_data_sources(desired_repo_contents.data_sources)
+        validate_data_sources(desired_repo_contents.data_sources)
         self._make_inferences(
             desired_repo_contents.data_sources,
             desired_repo_contents.entities,
@@ -5020,7 +5020,7 @@ def _print_materialization_log(
         )
 
 
-def _validate_feature_views(feature_views: List[BaseFeatureView]):
+def validate_feature_views(feature_views: List[BaseFeatureView]):
     """Verify feature views have case-insensitively unique names across all types.
 
     This validates that no two feature views (of any type: FeatureView,
@@ -5042,7 +5042,7 @@ def _validate_feature_views(feature_views: List[BaseFeatureView]):
             fv_by_name[case_insensitive_fv_name] = fv
 
 
-def _validate_data_sources(data_sources: List[DataSource]):
+def validate_data_sources(data_sources: List[DataSource]):
     """Verify data sources have case-insensitively unique names."""
     ds_names = set()
     for ds in data_sources:
--- a/sdk/python/feast/repo_operations.py
+++ b/sdk/python/feast/repo_operations.py
@@ -22,7 +22,11 @@ from feast.data_source import DataSource, KafkaSource, KinesisSource
 from feast.diff.registry_diff import extract_objects_for_keep_delete_update_add
 from feast.entity import Entity
 from feast.feature_service import FeatureService
-from feast.feature_store import FeatureStore
+from feast.feature_store import (
+    FeatureStore,
+    _validate_data_sources,
+    _validate_feature_views,
+)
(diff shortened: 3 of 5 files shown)


mayuriphad:
Hi maintainers — just checking in on this PR since it hasn't had a maintainer review yet. Happy to make any changes needed; let me know if anything's unclear or if I should split/rework it. Thanks for your time!

ntkathole:
**Potential issues with this approach:**

1. **`--skip-feature-view-validation` is now bypassed** — `parse_repo()` validates unconditionally, but the CLI exposes `--skip-feature-view-validation` (documented as an escape hatch for overly strict validation). Since `parse_repo()` does not receive this flag, users who pass it will still get validation errors they explicitly opted out of. Consider moving the early validation to `_get_repo_contents()` or `plan()`/`apply_total()` where the flag is available.

2. **Double validation** — The existing validation in `store.plan()`/`store.apply()` (`feature_store.py` L1454-1461) is not removed, so validation runs twice per invocation. Harmless but wasteful.

3. **Importing private functions across modules** — `_validate_feature_views` and `_validate_data_sources` are `_`-prefixed internals of `feature_store.py`. Importing them in `repo_operations.py` creates a coupling to private API that could break without notice. Consider either making them public or extracting them to a shared validation module.

mayuriphad:
Thanks for the thorough review, @ntkathole — pushed fixes for all three:

1. **`--skip-feature-view-validation` bypass**: moved the fail-fast validation out of `parse_repo()` (which has no access to CLI flags) and into `_get_repo_contents()`, gated by `skip_feature_view_validation`. Both `plan()` and `apply_total()` now thread the flag through to `_get_repo_contents()`, so passing `--skip-feature-view-validation` correctly skips this early check too, consistent with how `store.plan()`/`store.apply()` already honor it.
2. **Double validation**: with the check now conditional on the same flag used by `store.plan()`/`store.apply()`, it's intentional double-checking (fail fast before heavy deps, then defense-in-depth inside the store) rather than always-on redundant work — and it's skippable in one place via the flag.
3. **Private cross-module imports**: renamed `_validate_feature_views`/`_validate_data_sources` in `feature_store.py` to public `validate_feature_views`/`validate_data_sources`, and updated all internal callers plus the existing unit/integration tests that imported the private names.

Verified locally: the early check still fires by default (before `FeatureStore` is constructed) and is bypassed when `skip_feature_view_validation=True` is passed through `_get_repo_contents()`; existing `validate_feature_views`/`validate_data_sources` unit tests pass; `ruff check`/`ruff format --check` are clean on the touched files.

Let me know if you'd like anything else adjusted.

PR #6330, opened 2026-04-25
feat: Add versioning support to Milvus online store

makinzm:
# What this PR does / why we need it:

Adds feature view versioning support to the Milvus online store. When `enable_online_feature_view_versioning` is enabled in the registry config, the Milvus collection name now includes a version suffix (e.g. `project_driver_stats_v2`), allowing version-qualified feature references (e.g. `driver_stats@v2:trips_today`) to resolve to the correct versioned data.

Changes:
- Updated `_table_id()` in `milvus.py` to accept `enable_versioning` and delegate to the shared `compute_table_id()` helper (same pattern as SQLite, Redis, PostgreSQL, etc.)
- Updated all 5 call sites of `_table_id` to pass `config.registry.enable_online_feature_view_versioning`
- Added `MilvusOnlineStore` to the supported types list in `_check_versioned_read_support()` so versioned reads no longer raise `VersionedOnlineReadNotSupported`
- Added unit tests covering versioned/non-versioned table name generation and versioned read support

# Which issue(s) this PR fixes:

Fixes #6177

# Checks
- [x] I've made sure the tests are passing.
- [x] My commits are signed off (`git commit -s`)
- [x] My PR title follows [conventional commits](https://www.conventionalcommits.org/) format

## Testing Strategy
- [x] Unit tests

# Misc

Implementation follows the existing pattern from `sqlite.py` and `dynamodb.py`. The `compute_table_id()` helper in `helpers.py` handles the version suffix logic, so no additional logic was needed in the Milvus store itself.

---

<a href="https://app.devin.ai/review/feast-dev/feast/pull/6330" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

PR #6256, opened 2026-04-10
feat: Add feature view versioning support to FAISS online store

YassinNouh21:
## Summary
- When `enable_online_feature_view_versioning` is enabled, FAISS indices are namespaced by versioned table keys (e.g. `project_driver_stats_v2`) so multiple feature view versions coexist in memory
- Refactored singleton `_index`/`_in_memory_store` to per-table dictionaries (`_indices`/`_in_memory_stores`)
- Reuses shared `compute_table_id()` from `helpers.py` for consistency with PostgreSQL and MySQL stores (#6193)
- Registered `FaissOnlineStore` in `_check_versioned_read_support()` gate

## What this PR does / why we need it
Closes #6173

Feature view versioning was introduced in #6101, but version-qualified feature references (e.g. `driver_stats@v2:trips_today`) only worked with SQLite, PostgreSQL, and MySQL online stores. All other online stores raised `VersionedOnlineReadNotSupported` when a versioned ref was used.

This PR adds versioned read/write support to the FAISS online store, following the same pattern as the PostgreSQL/MySQL implementation (#6193):

- **Write path**: when `enable_online_feature_view_versioning` is enabled, `_table_id` appends a `_v{N}` suffix to the in-memory key namespace, routing materialization to the correct versioned index
- **Read path**: version-qualified lookups (`@v2`) set `projection.version_tag` on the feature view; `_table_id` picks this up and routes to the correct index. `projection.version_tag` takes priority over `current_version_number`
- **Multi-table support**: replaced single class-level `_index`/`_in_memory_store` with per-table `_indices`/`_in_memory_stores` dicts
- **Gate**: `_check_versioned_read_support` in `online_store.py` now allows `FaissOnlineStore` through

## Which issue(s) this PR fixes
Part of #2728

## Checks
- [x] I've made sure the tests are passing.
- [x] My commits are signed off (`git commit -s`)
- [x] My PR title follows [conventional commits](https://www.conventionalcommits.org/) format

## Testing Strategy
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual tests
- [ ] Testing is not required for this change

Added 16 unit tests covering `_table_id` with versioning disabled/enabled, `projection.version_tag` priority over `current_version_number`, version 0 edge case, versioned read support gate, versioned write/read round trips, namespace isolation between versions, missing index handling, and teardown cleanup.

---

<a href="https://app.devin.ai/review/feast-dev/feast/pull/6256" target="_blank">
  <picture>
[...]

PR #6757, opened 2026-08-18
feat: Add feature view versioning to Hazelcast online store

### LINKED ISSUE #6174: Add feature view versioning support to Hazelcast online store
## Summary

Add versioned feature view read/write support to the `Hazelcast` online store, so that version-qualified feature references (e.g., `driver_stats@v2:trips_today`) resolve to the correct versioned data.

## Context

Feature view versioning was introduced in #6101. Currently, only the SQLite online store supports versioned reads/writes. All other online stores raise `VersionedOnlineReadNotSupported` when a versioned ref is used.

The implementation pattern (see `sqlite.py` for reference) involves:
1. **Write path**: When `enable_online_feature_view_versioning` is enabled, materialize data into a version-specific table/key namespace (e.g., `{project}_{fv_name}_v{version}`)
2. **Read path**: Route version-qualified lookups to the correct versioned table/key namespace

## Reference implementation

- SQLite online store: `sdk/python/feast/infra/online_stores/sqlite.py`
- Store file: `sdk/python/feast/infra/online_stores/hazelcast_online_store/hazelcast_online_store.py`

## Related

Part of #2728

arose26:
Closes #6174. Part of #2728.

## What this does

Threads `registry.enable_online_feature_view_versioning` through `_map_name`, which now
delegates to `compute_table_id`, so that with versioning enabled each feature view version
gets its own Hazelcast map (`test_project_driver_stats_v2`) instead of all versions sharing
one. Applied to every call site — `online_write_batch`, `online_read`, `update` (both the
keep and delete loops) and `teardown`.

Hazelcast already namespaced as `{project}_{table.name}`, so this is the same
`compute_table_id` convention the merged Milvus (#6330) and FAISS (#6256) stores use — no
deviation. With versioning disabled the map name is byte-identical to today's, pinned by the
`versioning=False` cases.

## One small refactor, and why it is necessary

`update()` and `teardown()` previously called `_map_name(...)` inline inside f-strings:

```python
client.sql.execute(f"DELETE FROM {_map_name(config.project, table)}").result()
```

With the extra argument that call no longer fits on one line, and **line breaks inside an
f-string replacement field are a Python 3.12 feature** (PEP 701) while this package declares
`python_requires=">=3.10"`. So the resolved name is bound to a local first and the f-strings
just interpolate it. `ruff` with the project's `py310` target rejects the inline form, which
is how I caught it.

## Versioned-read allowlist: added, after checking

`HazelcastOnlineStore` **is** added to `OnlineStore._is_versioned_read_supported()`. I read the
path rather than assuming: `online_read` builds `entity_keys_str` from the requested keys in
order, then iterates it appending `(event_ts, entry)` for a hit and `(None, None)` for a miss —
so it returns one aligned entry per requested key, which is the contract `sqlite.py` implements
and the property versioned reads depend on. `test_read_returns_one_entry_per_requested_key`
pins it directly rather than taking my word for it.

Write and read also both serialize entity keys at version 3, so they agree on keys and the
change does not desynchronise them.

## Drive-by: a stale error message

`VersionedOnlineReadNotSupported` listed "SQLite, PostgreSQL, MySQL, FAISS, Redis, and
DynamoDB", but `_is_versioned_read_supported()` has included `MilvusOnlineStore` since #6330,
which updated `online_store.py` without touching `errors.py`. So the message has been telling
Milvus users their store is unsupported while the code supported it. Added Milvus alongside
Hazelcast — happy to split that into its own PR if you'd prefer.

## Tests

New `sdk/python/tests/unit/infra/online_store/test_hazelcast_versioning.py`, 15 tests,
`MagicMock`-based in the same style as the merged `test_milvus_versioning.py`, and
`importorskip`-guarded on `hazelcast`.

- naming: unversioned unchanged; version ignored while the flag is off; `_v2` when on;
  `projection.version_tag` takes priority over `current_version_number`; version `0` gets no
  suffix; two versions never collide
- routing: write, read, `update` (create *and* drop) and `teardown` all operate on the
  versioned map — write is parametrized so the unversioned case is asserted as the control
- allowlist: `_is_versioned_read_supported()` is true, `_check_versioned_read_support` no
  longer raises, and the alignment property the allowlist rests on is asserted

Verified red-before/green-after: with the three source files reverted, seven of these fail,
while the `versioning=False` control and the read-alignment test still pass — the latter
correctly, since alignment is a pre-existing property this PR relies on rather than introduces.

Regression check across the 55 unit test files touching online stores or versioning: the set of
failing and erroring tests is **identical** with and without this change (`diff` of the sorted
`FAILED`/`ERROR` lines is empty). Those pre-existing failures are missing optional dependencies
in my environment. `ruff check`, `ruff format` and `mypy` are clean on all four files.

---
🤖 Written with [Claude Code](https://claude.com/claude-code) (Claude Opus 5), reviewed by @arose26.


haoxu0 (approved):
LGTM. thanks for the contribution

PR #6765, opened 2026-08-20
fix: Feature freshness logic updated

jyejare:

The monitoring UI **Freshness** column previously used `metric_date` (the DQM computation window start) as a proxy for data freshness. This was misleading — `metric_date` reflects *when metrics were computed*, not *how fresh the underlying data is*.

This PR introduces a new `max_event_timestamp` field that stores `MAX(event_timestamp)` from the actual data source. The UI now displays age relative to the newest source event, giving an accurate picture of data staleness.

### Changes

**Backend — `monitoring_service.py`**
- Added `_as_utc_datetime()` helper to normalize timestamp-like values to timezone-aware UTC datetimes.
- Added `_newest_event_in_window()` to clamp the source's max timestamp to the computation window.
- All metric save paths (`_save_computed_metrics`, `_save_log_metrics`) now propagate `max_event_timestamp`.
- Feature service metrics aggregate `max_event_timestamp` across contributing feature views.

**Backend — `monitoring_utils.py`**
- Added `MONITORING_TIMESTAMP_FIELDS` constant to DRY-up timestamp field lists across offline stores.
- Added `max_event_timestamp` to all three column definition lists (feature, feature_view, feature_service).
- `normalize_monitoring_row` now handles pandas NaT/NaN for the new field.

**Schema — all offline stores** (BigQuery, Snowflake, Redshift, Postgres, Spark, Oracle, Dask, DuckDB)
- Added `max_event_timestamp` column to CREATE TABLE DDLs for all three monitoring tables.
- Added `ALTER TABLE … ADD COLUMN` migration for existing tables.

**UI — `FeatureMetricsTable.tsx` / `useMonitoringApi.ts`**
- Freshness column now uses `max_event_timestamp` with a fallback to `metric_date`.
- Fixed `render` function signature (`(_val, item)`) so the full `FeatureMetric` object is passed to the freshness logic.
- Added `Number.isNaN` guards for robustness.
- Updated tooltip to show "Newest event" instead of "Data from".

**Docs**
- Updated `feature-monitoring.md` with `max_event_timestamp` in the example response and clarified the freshness semantics.

Fixes inaccurate freshness display in the monitoring UI — the column previously showed `metric_date` (computation window) instead of actual data recency.

# Checks

## Testing Strategy
- [x] Unit tests
- [x] Integration tests
- [x] Manual tests

### Tests added
- **`tests/unit/monitoring/test_feature_freshness.py`** — unit tests for `_newest_event_in_window` (inside window, clamped, before window) and end-to-end `auto_compute` verification that `max_event_timestamp` is persisted.
- **`tests/integration/monitoring/test_monitoring_integration.py`** — extended `test_auto_compute_uses_pushdown_for_max_timestamp` to verify `max_event_timestamp` in saved metrics.
- **Manual** — verified on a live cluster with Dask offline store that freshness correctly reflects `MAX(event_timestamp)` from source parquet files.

# Misc

The new column is nullable and added via `ADD COLUMN IF NOT EXISTS` migrations, so this is backward-compatible with existing monitoring tables.

### DIFF
--- a/sdk/python/feast/monitoring/monitoring_service.py
+++ b/sdk/python/feast/monitoring/monitoring_service.py
@@ -29,6 +29,37 @@ GRANULARITY_WINDOWS = {
     "quarterly": timedelta(days=90),
 }
 
+
+def _as_utc_datetime(val: Any) -> Optional[datetime]:
+    """Parse a timestamp-like value to a timezone-aware UTC datetime."""
+    if val is None:
+        return None
+    if isinstance(val, datetime):
+        return val if val.tzinfo else val.replace(tzinfo=timezone.utc)
+    if isinstance(val, date):
+        return datetime.combine(val, datetime.min.time(), tzinfo=timezone.utc)
+    if isinstance(val, str):
+        parsed = datetime.fromisoformat(val.replace("Z", "+00:00"))
+        return parsed if parsed.tzinfo else parsed.replace(tzinfo=timezone.utc)
+    return None
+
+
+def _newest_event_in_window(
+    max_ts: Optional[datetime],
+    start_dt: datetime,
+    end_dt: datetime,
+) -> Optional[datetime]:
+    """Newest source event that falls inside ``[start_dt, end_dt]``.
+
+    ``max_ts`` is MAX(event_timestamp) over the whole source. If that
+    timestamp is after the window, the window end is used as an upper bound.
+    """
+    aware_max = _as_utc_datetime(max_ts)
+    if aware_max is None or aware_max < start_dt:
+        return None
+    return min(aware_max, end_dt)
+
+
 _FLOAT_FIELDS = frozenset(
     {
         "null_rate",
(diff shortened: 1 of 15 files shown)


sdk/python/feast/monitoring/monitoring_service.py
```diff
@@ -950,6 +993,11 @@ def _compute_for_feature_view(
             granularity=granularity,
             set_baseline=set_baseline,
             now=now,
+            max_event_timestamp=_newest_event_in_window(
+                self._get_max_timestamp(feature_view),
```
ntkathole: `_compute_for_feature_view` already receives start_dt and end_dt that were computed from a prior _get_max_timestamp call in the caller. Consider passing the already-known max_ts (or end_dt) into this method to avoid the redundant query.

jyejare: Good catch! Fixed in 21a5936 — `_compute_for_feature_view` now accepts an optional `max_event_timestamp` parameter, and `compute_metrics` fetches `max_ts` once per feature view via `_get_max_timestamp()` and passes it down, eliminating the redundant source query.

ntkathole (approved):

